PR #8842, opened 2020-05-20
[batch] fix worker activate timeout

jigold:


batch/batch/worker.py
```diff
@@ -861,7 +861,7 @@ async def run(self):
             await site.start()
 
             try:
-                await asyncio.wait_for(self.activate(), MAX_IDLE_TIME_MSECS * 1000)
+                await asyncio.wait_for(self.activate(), MAX_IDLE_TIME_MSECS)
```
danking, line 864: Either I'm misreading the docs for `wait_for` or MAX_IDEL_TIME_MSECS is not in milliseconds, which is it?

jigold: I was being stupid. We need to divide by 1000.

PR #8844, opened 2020-05-20
[batch] add local SSDs

jigold:
CHANGELOG: Changed cost per instance from $0.02170 to $0.02208 from switching to using local SSDs.

- Added 1 local SSD (375 GB) and formatted it in the worker run script.
- Changed the resource for boot-disk to just disk and modified the worker config. I figured there was no reason to have a separate boot disk in the resources as long as all disks are assumed to be fractions of the instance based on the number of cores being used.
- Changed the worker boot disk from 100 GB to 20 GB
- Changed the worker to move all docker files and batch files to the Local SSD from the boot disk.

Can you double check my math for the documentation?

Is it possible it takes longer for an instance to boot up with a local SSD? One of my earlier tests had workers stuck in STAGING. This resolved itself later on so I'm assuming it was a Google error.


jigold:
Actually, the change log doesn't make sense as the version of the service isn't related to the documentation/hail pip version. Thoughts on removing the change log here and just having client interface changes?

batch/sql/insert_local_ssd_resource.py
```diff
@@ -0,0 +1,43 @@
...
+from gear import Database
+from hailtop.utils import rate_gib_month_to_mib_msec
+
+
+async def main():
+    # https://cloud.google.com/compute/disks-image-pricing#localssdpricing
+    # https://cloud.google.com/compute/disks-image-pricing#persistentdisk
+    rates = [
+        ('disk/local-ssd/1', rate_gib_month_to_mib_msec(0.048)),
+        ('disk/pd-ssd/1', rate_gib_month_to_mib_msec(0.17))
```
danking, line 13: Just confirming: you're certain we're using zonal persistent ssds, *not* regional persistent ssds?

batch/batch/worker_config.py
```diff
@@ -112,11 +114,22 @@ def resources(self, cpu_in_mcpu, memory_in_bytes):
...
+            size = disk['size'] * worker_fraction_in_1024ths
+
+            if name not in quantities:
+                quantities[name] = size
+            else:
+                quantities[name] += size
+
+        for name, quantity in quantities.items():
+            resources.append({'name': name,
+                              'quantity': quantity})
```
danking, line 132: Can we put the dimensions in the variable names? so `disk_size_in_mib = ...`

batch/batch/worker_config.py
```diff
@@ -112,11 +114,22 @@ def resources(self, cpu_in_mcpu, memory_in_bytes):
...
 
         resources.append({'name': f'memory/{self.instance_family}-{preemptible}/1',
-                          'quantity': memory_in_bytes / 1024 / 1024})
-
-        # the factors of 1024 cancel between GiB -> MiB and fraction_1024 -> fraction
-        resources.append({'name': f'boot-disk/{self.boot_disk_type}/1',
-                          'quantity': self.boot_disk_size_gb * worker_fraction_in_1024ths})
+                          'quantity': math.ceil(memory_in_bytes / 1024 / 1024)})
+
+        quantities = {}
```
danking, line 119: this is a great use case for `defaultdict`:
```python3
from collections import defaultdict

...

quantities = defaultdict(lambda: 0)

....

    quantitites[name] += size
```

hail/python/hailtop/batch/docs/service.rst
```diff
@@ -55,29 +55,41 @@ Billing
...
+  - Local SSD
+
+    .. code-block:: text
+
+        Cost per GB per month = $0.048
+
+        Cost per core per hour = $0.048 * 375 / 30.4375 / 24 / 16
+
+
+   = $0.00183 per core per hour
```
danking, line 84: so this seems right:
```python3
>>> 0.17 * 20 / 30.4375 / 24 / 16 + 0.048 * 375 / 30.4375 / 24 / 16
0.001830937713894593
```

But why was the old value different from this?
```python3
>>> 0.17 * 20 / 30.4375 / 24 / 16
0.0002908966461327858
```

danking, line 84: this gets displayed as a block quote, maybe the equals sign needs to be one indent further to the left?

hail/python/hailtop/batch/docs/service.rst
```diff
@@ -55,29 +55,41 @@ Billing
...
 
 - IP network cost
    = $0.00025 per core per hour
 
 - Service cost
    = $0.01 per core per hour
 
-The sum of these costs is **$0.02170** per core per hour.
+The sum of these costs is **$0.02208** per core per hour.
```
danking, line 92: seems right to me.

batch/batch/worker_config.py
```diff
@@ -112,11 +114,22 @@ def resources(self, cpu_in_mcpu, memory_in_bytes):
                           'quantity': cpu_in_mcpu})
 
         resources.append({'name': f'memory/{self.instance_family}-{preemptible}/1',
-                          'quantity': memory_in_bytes / 1024 / 1024})
-
-        # the factors of 1024 cancel between GiB -> MiB and fraction_1024 -> fraction
-        resources.append({'name': f'boot-disk/{self.boot_disk_type}/1',
-                          'quantity': self.boot_disk_size_gb * worker_fraction_in_1024ths})
+                          'quantity': math.ceil(memory_in_bytes / 1024 / 1024)})
```
danking, line 117: memory_in_bytes should always be divisible by 1024 * 1024, right? Let's change this to:
```
assert memory_in_bytes % (1024 * 1024) == 0
...
    'quantity': memory_in_bytes // 1024 // 1024
```

batch/sql/insert_local_ssd_resource.py
jigold: Yes. They can't be boot disks. https://cloud.google.com/compute/docs/disks/regional-persistent-disk

hail/python/hailtop/batch/docs/service.rst
jigold: You need `100` instead of `20`. I think it might also be wrong. The total number was definitely correct. I may have made a previous mistake on the disk cost rate.

jigold: I just folded the average number of days into the text blocks. Otherwise, it looked weird.

jigold:
Maybe we should make the boot disk 10 GB instead of 20 GB.

hail/python/hailtop/batch/docs/service.rst
danking: Now the whole thing looks like a block quote? Does it look fine when you build the docs locally? https://github.com/hail-is/hail/blob/091a979e9a2fea8129007ca87f6e533a5dbe73a8/hail/python/hailtop/batch/docs/service.rst

jigold: Screen Shot 2020-05-29 at 11 54 42 AM

jigold:
I think I already changed that in the python database change script and in the documentation.

PR #8847, opened 2020-05-21
[query] Replace recursive BTree foreach with iterative version

chrisvittal:
Allocate a fixed stack, as it was the simplest thing I could do.  128 stack
slots should be more than enough for this implementation as it requires one
stack slot per level of the tree.

There are many improvements that can be made here, but hopefully this should
unblock some amount of the method splitting work.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/agg/AppendOnlyBTree.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/agg/AppendOnlyBTree.scala
@@ -201,30 +201,70 @@ class AppendOnlyBTree(kb: EmitClassBuilder[_], key: BTreeKey, region: Value[Regi
     getF.invokeCode(root, km, kv)
 
   def foreach(cb: EmitCodeBuilder)(visitor: (EmitCodeBuilder, Code[Long]) => Unit): Unit = {
-    val f = kb.genEmitMethod("btree_foreach", FastIndexedSeq[ParamType](typeInfo[Long]), typeInfo[Unit])
-    val node = f.getCodeParam[Long](1)
-    val i = f.newLocal[Int]("aobt_foreach_i")
-
+    // This emit should be removable once method splitting has been done properly
+    val f = kb.genEmitMethod("btree_foreach", FastIndexedSeq(), typeInfo[Unit])
     f.emitWithBuilder { cb =>
+
+    val stackI = cb.newLocal[Int]("btree_foreach_stack_i", -1)
+    val nodeStack = cb.newLocal("btree_foreach_node_stack", Code.newArray[Long](const(128)))
+    val idxStack = cb.newLocal("btree_foreach_index_stack", Code.newArray[Int](const(128)))
+
+    def stackPush(node: Code[Long]) = {
+      cb.assign(stackI, stackI + 1)
+      cb += nodeStack.update(stackI, node)
+      cb += idxStack.update(stackI, -1)
+    }
+    def stackUpdateIdx(newIdx: Code[Int]) = {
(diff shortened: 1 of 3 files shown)


chrisvittal:
For what it's worth, I kinda hate this code. But without easier generation of closures I think this should be sufficient for now.

hail/src/main/scala/is/hail/expr/ir/agg/AppendOnlyBTree.scala
```diff
@@ -201,30 +201,70 @@ class AppendOnlyBTree(kb: EmitClassBuilder[_], key: BTreeKey, region: Value[Regi
...
+        stackUpdateIdx(0)
+        stackPush(loadChild(node, -1))
+        cb.goto(Lstart)
       })
-      cb.assign(i, 0)
-      val Lexit = CodeLabel()
       (0 until maxElements).foreach { i =>
+        cb.define(labels(i))
         cb.ifx(hasKey(node, i), {
           visitor(cb, loadKey(node, i))
```
cseed, line 251: So I don't think this is going to work for splitting.  You're generating a new method and emitting `visitor` inside that method, but the `visitor` is going to be part of the IR code generation that exclusively uses locals within the toplevel apply method.  The options are:

 - fully inline foreach
 - have foreach return an iterator object that contains the traversal state (including the stack), and generate a loop with visitor inside of it at the call site
 - construct a proper closure for `visitor` that captures the local state it needs

In other words, you need to: not use a function, pass an object out, or a pass an object in.

chrisvittal:
@cseed bump

hail/src/main/scala/is/hail/expr/ir/agg/AppendOnlyBTree.scala
```diff
@@ -201,30 +201,61 @@ class AppendOnlyBTree(kb: EmitClassBuilder[_], key: BTreeKey, region: Value[Regi
...
+    def stackUpdateIdx(newIdx: Code[Int]) = {
+      cb += idxStack.update(stackI, newIdx)
+    }
+    def stackPop() = {
+      cb.assign(stackI, stackI - 1)
+    }
 
-    f.emitWithBuilder { cb =>
+    stackPush(root)
+    cb.whileLoop(stackI >= 0, { (Lstart) =>
```
cseed, line 221: Not for this PR, but do we want to merge Lend into whileLoop?  I think of `Lstart` as continue (go back to the beginning of the loop).  Similar, `Lend` is break.

hail/src/main/scala/is/hail/expr/ir/agg/AppendOnlyBTree.scala
```diff
@@ -201,30 +201,61 @@ class AppendOnlyBTree(kb: EmitClassBuilder[_], key: BTreeKey, region: Value[Regi
...
+      val node = cb.newLocal("btree_foreach_node", nodeStack(stackI))
+      val idx = cb.newLocal("btree_foreach_idx", idxStack(stackI))
+      val Lend = CodeLabel()
+      val Lminus1 = CodeLabel()
+      val labels = Array.fill[CodeLabel](maxElements)(CodeLabel())
+      // this should probably be a switch, don't know how to make it one though
+      // furthermore, we should be able to do the lookups at runtime
+      // FIXME, clean this up once we have fixed arrays
```
cseed, line 229: Not for this PR, but let's talk about the interface for this.  `cb.switch(i: Int, default: CodeLabel, cases: Array[CodeLabe])` should work for this case, and should straightforwardly follow from `Code.switch`.

PR #8850, opened 2020-05-21
[batch] Add standing worker to instance pool

jigold:
CHANGELOG: Changed the billing parameters to include the possibility of running on 4 core workers instead of 16 core workers.


jigold:
I made the timeout for the standing workers to be 5 minutes for the test scope and 2 hours for the deploy scope.

jigold:
I had an idea to save money -- limit the standing worker to only be during business hours.

danking:
Eh, as you point out in the billing changes it's 0.02208 USD per core per hour, that's 2 USD per day. It's probably a bit more if we have a full SSD for a 4 core, but still, this is extremely far away from our biggest cost center. I vote for simplicity.

batch/batch/batch_configuration.py
```diff
@@ -12,3 +12,6 @@
 WORKER_LOGS_BUCKET_NAME = os.environ['HAIL_WORKER_LOGS_BUCKET_NAME']
 HAIL_SHA = os.environ['HAIL_SHA']
 HAIL_SHOULD_PROFILE = os.environ.get('HAIL_SHOULD_PROFILE') is not None
+ENABLE_STANDING_WORKER = bool(int(os.environ['ENABLE_STANDING_WORKER']))
```
danking, line 15: I prefer a consistent style for boolean environment variables. Let's do what HAIL_SHOULD_PROFILE did, the variable merely needs to be set to anything at all.

batch/batch/driver/instance_pool.py
```diff
@@ -51,18 +53,21 @@ def __init__(self, app, machine_name_prefix):
 
         # pending and active
         self.live_free_cores_mcpu = 0
+        self.live_total_cores_mcpu = 0
```
danking, line 56: the comment above is now unclear. let's move it to an end-of-line comment to line 55.

hail/python/hailtop/batch/docs/service.rst
```diff
@@ -79,6 +79,10 @@ machines with 100 GB of disk total. The costs are as follows:
 
 The sum of these costs is **$0.02170** per core per hour.
 
+Some jobs may be run on 4 core machines instead of 16 cores. In that case,
```
danking, line 82: Let's be a bit more specific in an attempt to head off people asking us questions:

> At any given moment as many as four cores of the cluster may come from a 4 core machine. If a job is 
> scheduled on this machine, ...

batch/batch/front_end/front_end.py
```diff
@@ -1417,13 +1417,14 @@ async def on_startup(app):
...
         '''
-SELECT worker_type, worker_cores, worker_disk_size_gb,
+SELECT worker_type, worker_cores, worker_disk_size_gb, standing_worker_cores,
   instance_id, internal_token, n_tokens FROM globals;
 ''')
 
     app['worker_type'] = row['worker_type']
     app['worker_cores'] = row['worker_cores']
     app['worker_disk_size_gb'] = row['worker_disk_size_gb']
+    app['standing_worker_cores'] = row['standing_worker_cores']
```
danking, line 1427: This will become out-of-date if someone edits this on the driver. It also appears unused in the front-end?

batch/batch/driver/main.py
```diff
@@ -640,12 +648,15 @@ async def on_startup(app):
...
-        'SELECT worker_type, worker_cores, worker_disk_size_gb, instance_id, internal_token FROM globals;')
+    row = await db.select_and_fetchone('''
+SELECT worker_type, worker_cores, worker_disk_size_gb,
+  standing_worker_cores, instance_id, internal_token FROM globals;
+''')
 
     app['worker_type'] = row['worker_type']
     app['worker_cores'] = row['worker_cores']
     app['worker_disk_size_gb'] = row['worker_disk_size_gb']
+    app['standing_worker_cores'] = row['standing_worker_cores']
```
danking, line 659: This never gets updated, right? Is it used anywhere?

batch/deployment.yaml
```diff
@@ -182,11 +193,22 @@ spec:
...
            value: hail-test-dmk9z
+         - name: STANDING_WORKER_MAX_IDLE_TIME_SECS
+           value: "300"
+{% endif %}
+{% if deploy or scope == "test" %}
+         - name: ENABLE_STANDING_WORKER
+           value: "1"
+{% else %}
+         - name: ENABLE_STANDING_WORKER
+           value: "0"
```
danking, line 211: I'm not convinced the front end needs to know about this.

batch/batch/driver/instance_pool.py
jigold: Why is it unclear? They're both computed from pending and active instances.

jigold:
Depending on which PR of #8844 goes in first, I'll need to update the docs.

batch/batch/driver/instance_pool.py
danking: Ah, yes, I left this comment before I figured out what live_total_cores_mcpu was, my bad.

PR #8851, opened 2020-05-21
remove learn more

akotlar:
CHANGELOG: remove learn more, which should not have gone in until the learn more content was finalized.


tpoterba:
I don't think stuff like this should have a changelog entry. This isn't really something we would expect users to care about, docs changelog entries should be for big milestones like addition of cheat sheets.

akotlar:
Got it

PR #8853, opened 2020-05-21
Lower Force Count

johnc1231:
Added a `Consume` IR that just emits its body and then returns a 1. I think I did the `IEmitCode` bit right such that it should emit all of the inner code, then ignore it and return 1, but I'd appreciate a second set of eyes. The lowering rule is trivial given that `Consume` exists. 

In the future, may be better to delete `ForceCountTable` entirely, and just use `Consume` from Python. 

cc @tpoterba

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -675,6 +675,13 @@ class Emit[C](
         presentC(const(true))
       case False() =>
         presentC(const(false))
+      case Consume(value) => {
+        val iec = emitI(value)
+        iec.map(cb){pc =>
+          // Ignore pc, just return a 1
+          PCode(ir.pType, 1L)
+        }
+      }
 
       case Cast(v, typ) =>
         val iec = emitI(v)
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -676,13 +676,12 @@ class Emit[C](
       case False() =>
         presentC(const(false))
       case Consume(value) => {
-        val iec = emitI(value)
-        iec.map(cb){pc =>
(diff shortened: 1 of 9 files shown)


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -675,6 +675,13 @@ class Emit[C](
         presentC(const(true))
       case False() =>
         presentC(const(false))
+      case Consume(value) => {
+        val iec = emitI(value)
+        iec.map(cb){pc =>
+          // Ignore pc, just return a 1
```
tpoterba, line 681: I actually think maybe this should memoize pc, then ignore the result. That seems like a safe way to ensure it's being evaluated.

tpoterba:
Super fast turnaround, thanks John!

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -675,6 +675,13 @@ class Emit[C](
         presentC(const(true))
       case False() =>
         presentC(const(false))
+      case Consume(value) => {
+        val iec = emitI(value)
+        iec.map(cb){pc =>
```
chrisvittal, line 680: There's no need to bind `iec` to a variable here. `emitI(value).map` would be sufficient.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
chrisvittal: Seconded. Still not sure that it will actually work.

I don't know if we'll still fully have the ability to put things in fields in the future, but I would recommend using `cb.memoizeField`, I think the JVM will optimize less around field stores since it's probably impossible to prove that public fields (which I think all of our fields are) will never be read.

johnc1231:
Alright, I've memoized `pc` now. Thanks for input

PR #8857, opened 2020-05-22
[query] fix converting iterator/buffer to stream

patrick-schultz:
In several places constructing a stream from in iterator or buffer, the side effect of advancing the input was packaged with the value passed to the stream consumer. If that value was never used, the input wasn't advanced properly.

I fixed this pattern every place I could think of: the two `PartitionReader`s, and the `In` node allowing a java iterator to be passed as a stream argument to a compiled function.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/GenericTableValue.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/GenericTableValue.scala
@@ -43,20 +43,22 @@ class PartitionIteratorLongReader(
       assert(contextPC.pt.isInstanceOf[PStruct])
 
       val it = mb.newLocal[Iterator[java.lang.Long]]("pilr_it")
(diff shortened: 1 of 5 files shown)


hail/src/main/scala/is/hail/expr/ir/GenericTableValue.scala
```diff
@@ -7,7 +7,7 @@ import is.hail.backend.spark.SparkBackend
 import is.hail.expr.ir.EmitStream.SizedStream
 import is.hail.expr.ir.lowering.TableStage
 import is.hail.types.TableType
-import is.hail.types.physical.{PStruct, PType}
+import is.hail.types.physical.{PCode, PInt64Optional, PStruct, PType}
```
cseed, line 10: These imports are no longer used.

PR #8859, opened 2020-05-22
[query] Add missing-nonequal staged comparisons

patrick-schultz:
Joins should not equate missing key fields from the left and right inputs, e.g. the join of `[{key: null, leftVal: 1}]` and `[{key: null, rightVal: 5}]` should be
```
[{key: null, leftVal: 1, rightVal: null}, {key: null, leftVal: null, rightVal: 5}]
```
not
```
[{key: null, leftVal: 1, rightVal: 5}]
```

To support this in lowered joins, this PR adds non-symmetric staged comparisons which treat nulls on the left as less than nulls on the right, i.e. `compare(null, null) = -1`.

### DIFF
--- a/hail/src/test/scala/is/hail/expr/ir/OrderingSuite.scala
+++ b/hail/src/test/scala/is/hail/expr/ir/OrderingSuite.scala
@@ -43,55 +43,111 @@ class OrderingSuite extends HailSuite {
   }
 
   @Test def testMissingNonequalComparisons() {
+    def getStagedOrderingFunctionWithMissingness(
+      t: PType,
+      op: CodeOrdering.Op,
+      r: Region,
+      sortOrder: SortOrder = Ascending
+    ): AsmFunction5[Region, Boolean, Long, Boolean, Long, op.ReturnType] = {
+      implicit val x = op.rtti
+      val fb = EmitFunctionBuilder[Region, Boolean, Long, Boolean, Long, op.ReturnType](ctx, "lifted")
+      val m1 = fb.getCodeParam[Boolean](2)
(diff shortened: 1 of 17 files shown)


hail/src/test/scala/is/hail/expr/ir/OrderingSuite.scala
```diff
@@ -41,6 +42,63 @@ class OrderingSuite extends HailSuite {
     fb.resultWithIndex()(0, r)
   }
 
+  @Test def testMissingNonequalComparisons() {
```
catoverdrive, line 45: This test might be more complete if you compared the values to themselves, since the missing values being unequal case will only arise if the values being compared are identical up until the first set of fields where both are missing (not sure how likely this is to come up; we're only running 10 tests). If you have a test comparing values to themselves, you're at least forcing the missingness logic to come up once per test.

patrick-schultz:
Thanks for catching that. The current behavior seems strange to me, but I think we have to preserve it.

hail/src/test/scala/is/hail/expr/ir/OrderingSuite.scala
patrick-schultz: Good idea!

patrick-schultz: Actually, without needing to test deep missingness behavior, it's even easier.

patrick-schultz:
I backed off the support for treating deep NAs as nonequal. That makes the change  simpler, and also easier to test.

PR #8863, opened 2020-05-26
[vcf-combiner] Set <NON_REF> alleles to missing in transform_gvcf

tpoterba:
CHANGELOG: Fixes VCF combiner to successfully import GVCFs with alleles called as `<NON_REF>`.


danking:
I think your CHANGELOG should say NON_REF, right?

tpoterba:
ah yeah will fix

hail/python/hail/experimental/vcf_combiner/vcf_combiner.py
```diff
@@ -136,7 +136,7 @@ def transform_gvcf(mt, info_to_keep=[]) -> Table:
                             GQ=e.GQ,
                             LA=hl.range(0, alleles_len - hl.cond(has_non_ref, 1, 0)),
                             LAD=hl.cond(has_non_ref, e.AD[:-1], e.AD),
-                            LGT=e.GT,
+                            LGT=hl.case().when(~(has_non_ref & (e.GT.unphased_diploid_gt_index() >= (alleles_len * (alleles_len - 1)) / 2)), e.GT).or_missing(),
```
cseed, line 139: Why not just `hl.or_missing()` here?

If it doesn't have non_ref, shouldn't the index always be true?  That is, should that be an error (die) rather than set to missing?

tpoterba: this generates the same IR as `or_missing`. I don't like `or_missing`, we've had a lot of people confused by it in the past.

tpoterba: I can add the error though, yes

cseed: Users, or us?

tpoterba: Users. I want to deprecate and remove or_missing entirely, because I don't think it adds much, and when people see code that contains or_missing, they can miss the point that we're generating a conditional.

This isn't something I feel super strongly about, though.

cseed: I think or_missing is pretty natural.  I thought using a case (with or_missing!) for a single branch is overly noisy.  I think whether it is user visible or we use it are separate questions.  I don't care much if it is user visible.

hail/python/hail/expr/functions.py
```diff
@@ -139,6 +139,28 @@ def null(t: Union[HailType, str]):
     """
     return construct_expr(ir.NA(t), t)
 
+@typecheck(msg=expr_str, t=hail_type)
+def die(msg, t: Union[HailType, str]):
```
chrisvittal, line 143: This should be a separate PR

PR #8865, opened 2020-05-26
[query] Add `hl.die` function for error checking.

tpoterba:
CHANGELOG: Add `hl.die` function that can be used to generate errors. Useful in data validation.

### DIFF
--- a/hail/python/hail/docs/functions/core.rst
+++ b/hail/python/hail/docs/functions/core.rst
@@ -12,6 +12,7 @@ Core language functions
     bind
     rbind
     null
+    die
     str
     is_missing
     is_defined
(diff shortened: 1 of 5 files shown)


chrisvittal:
I'm fine with calling it whatever. `die` could be mistaken for something that will shut down hail, at the same time, that's what the IR node is called.

konradjk:
> slight nitpick---can we call this "error" or "raise_error" or something less dire sounding than "die"?

or some version of `except` for parallelism with python?

konradjk:
Btw, I actually don't understand how one would use this function (not clear to me from the docs nor the test)

johnc1231:
I think it's useful if you want to write something like:

```
hl.if_else(x < y, ...., hl.die("Error: x must be less than y"))
```

Otherwise you can't do input validation on expressions.

konradjk:
Yeah that's what I had initially thought, but then realized it also takes a type though? Anyway, some example in the docs would be good either way

tpoterba:
gonna close this for now.

PR #8866, opened 2020-05-26
[query] fix get_gene_intervals

konradjk:
Not sure what happened to `info` but this used to work. Happy to switch to something else if preferred.


tpoterba:
what error do you get now?

konradjk:
`info not defined`

konradjk:
wait, what? i see the import at the top. maybe my rig had something screwy. i'll close for now unless i hit it again

tpoterba:
this should still work. We recently made some under-the-hood changes to info, but it shouldn't have affected your usage.

PR #8867, opened 2020-05-27
BlockMatrix Transpose Sparsity Fix

johnc1231:
CHANGELOG: Fix a major correctness bug ocurring when calling  `BlockMatrix.transpose` on sparse BlockMatrices. Symmetric matrices are not affected.

It seems like `BlockMatrix.transpose` has been broken for a while when the matrix is sparse. 

I added `PerBlockMatrixSparsifier` as a way to sparsify particular blocks of a `BlockMatrix` from python. This is just so we can write tests / diagnose user errors based on sparsity patterns. I also wrote a helper function to sparsify numpy matrices for testing purposes. 

The crucial fix here is to `GridPartitioner.transpose`. That function is supposed to return a pair of the form `(GridPartitioner, Int => Int)`, where the first of the pair is the new `GridPartitioner` for the transposed thing, and the second of the pair is a function that takes in a partition number and returns the partition number of its parent partition. Crucially, it's a function from new partition ids to old partition ids. I believe that code I'm removing did the opposite.

Refresher on `GridPartitioner`: There are 3 coordinate systems:

There's "coordinate", which is (row, column)

There's "blockIndex", which is the column major numbering of all blocks

And there's "partitionIndex", which is similar to numbering by blockIndex but it skips the sparse blocks

### DIFF
--- a/hail/python/test/hail/linalg/test_linalg.py
+++ b/hail/python/test/hail/linalg/test_linalg.py
@@ -1028,6 +1028,12 @@ class Tests(unittest.TestCase):
         bm = bm._sparsify_blocks(block_list)
         sparse_numpy = sparsify_numpy(np_square, block_size, block_list)
         assert np.array_equal(bm.to_numpy(), sparse_numpy)
+        assert np.array_equal(
+            sparse_numpy,
+            np.array([[0,  0,  3, 4],
+                      [0,  0,  7, 8],
+                      [9,  10, 0, 0],
+                      [13, 14, 0, 0]]))
 
         block_list = [4, 8, 10, 12, 13, 14]
         np_square = np.arange(225, dtype=np.float64).reshape((15, 15))
--- a/hail/python/test/hail/linalg/test_linalg.py
+++ b/hail/python/test/hail/linalg/test_linalg.py
@@ -1039,8 +1039,32 @@ class Tests(unittest.TestCase):
 
     @skip_unless_spark_backend()
     def test_sparse_transposition(self):
+        block_list = [1, 2]
+        np_square = np.arange(16, dtype=np.float64).reshape((4, 4))
+        block_size = 2
+        bm = BlockMatrix.from_numpy(np_square, block_size=block_size)
+        sparse_bm = bm._sparsify_blocks(block_list).T
+        sparse_np = sparsify_numpy(np_square, block_size, block_list).T
+        assert np.array_equal(sparse_bm.to_numpy(), sparse_np)
+
         block_list = [4, 8, 10, 12, 13, 14]
         np_square = np.arange(225, dtype=np.float64).reshape((15, 15))
+        block_size = 4
+        bm = BlockMatrix.from_numpy(np_square, block_size=block_size)
+        sparse_bm = bm._sparsify_blocks(block_list).T
+        sparse_np = sparsify_numpy(np_square, block_size, block_list).T
+        assert np.array_equal(sparse_bm.to_numpy(), sparse_np)
+
+        block_list = [2, 5, 8, 10, 11]
+        np_square = np.arange(150, dtype=np.float64).reshape((10, 15))
+        block_size = 4
+        bm = BlockMatrix.from_numpy(np_square, block_size=block_size)
+        sparse_bm = bm._sparsify_blocks(block_list).T
+        sparse_np = sparsify_numpy(np_square, block_size, block_list).T
+        assert np.array_equal(sparse_bm.to_numpy(), sparse_np)
+
+        block_list = [2, 5, 8, 10, 11]
+        np_square = np.arange(165, dtype=np.float64).reshape((15, 11))
         block_size = 4
         bm = BlockMatrix.from_numpy(np_square, block_size=block_size)
         sparse_bm = bm._sparsify_blocks(block_list).T
--- a/hail/python/test/hail/linalg/test_linalg.py
+++ b/hail/python/test/hail/linalg/test_linalg.py
@@ -1030,10 +1030,10 @@ class Tests(unittest.TestCase):
         assert np.array_equal(bm.to_numpy(), sparse_numpy)
         assert np.array_equal(
             sparse_numpy,
-            np.array([[0,  0,  3, 4],
-                      [0,  0,  7, 8],
-                      [9,  10, 0, 0],
-                      [13, 14, 0, 0]]))
+            np.array([[0,  0,  2, 3],
+                      [0,  0,  6, 7],
+                      [8,  9,  0, 0],
+                      [12, 13, 0, 0]]))
 
         block_list = [4, 8, 10, 12, 13, 14]
         np_square = np.arange(225, dtype=np.float64).reshape((15, 15))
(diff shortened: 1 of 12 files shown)


johnc1231:
Transpose was broken here, but my fix back then was insufficient: #7170

danking:
Before merge, the Changelog will need to be more precise about what could have given bad answers.

johnc1231:
Updated changelog

hail/src/main/scala/is/hail/linalg/GridPartitioner.scala
```diff
@@ -117,14 +118,15 @@ case class GridPartitioner(blockSize: Int, nRows: Long, nCols: Long, maybeBlocks
...
       case Some(bis) =>
-        val (biTranspose, piTranspose) = bis.map(transposeBI).zipWithIndex.sortBy(_._1).unzip
-        val inverseTransposePI = piTranspose.zipWithIndex.sortBy(_._1).map(_._2)
+        val (biTranspose, newPIToOldPI) = bis.map(transposeBI).zipWithIndex.sortBy(_._1).unzip
         
-        (GridPartitioner(blockSize, nCols, nRows, Some(biTranspose)), inverseTransposePI)
-      case None => (gpT, inverseTransposeBI)
+        (GridPartitioner(blockSize, nCols, nRows, Some(biTranspose)), newPIToOldPI.apply)
+      case None => {
+        def newBIToOldBI(bi: Int) = this.coordinatesBlock(gpT.blockBlockCol(bi), gpT.blockBlockRow(bi))
```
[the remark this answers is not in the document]
danking: erm this is just `untransposeBI`, right? It removes the transposition?

hail/src/main/scala/is/hail/linalg/GridPartitioner.scala
```diff
@@ -117,14 +118,15 @@ case class GridPartitioner(blockSize: Int, nRows: Long, nCols: Long, maybeBlocks
   def transpose: (GridPartitioner, Int => Int) = {
     val gpT = GridPartitioner(blockSize, nCols, nRows)
     def transposeBI(bi: Int): Int = gpT.coordinatesBlock(this.blockBlockCol(bi), this.blockBlockRow(bi))
-    def inverseTransposeBI(bi: Int) = this.coordinatesBlock(gpT.blockBlockCol(bi), gpT.blockBlockRow(bi))
     maybeBlocks match {
       case Some(bis) =>
-        val (biTranspose, piTranspose) = bis.map(transposeBI).zipWithIndex.sortBy(_._1).unzip
-        val inverseTransposePI = piTranspose.zipWithIndex.sortBy(_._1).map(_._2)
+        val (biTranspose, newPIToOldPI) = bis.map(transposeBI).zipWithIndex.sortBy(_._1).unzip
```
danking, line 123: this is `transposedBIToUntranposedBI`, right? `newPIToOldPI` has nothing to do with partition indices.

johnc1231: Are you commenting on the first or second of the tuple elements. `biTranspose` is the result of mapping the `blockIndices` from the old `GridPartitioner` to the new transposed one. `newPIToOldPI` is the mapping from new partition indices to old partition indices. Maybe better to call it something about "parent" partition indices?

hail/src/main/scala/is/hail/linalg/GridPartitioner.scala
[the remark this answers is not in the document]
johnc1231: It takes a block index in the transposed gridpartitioner and maps it to the original gridpartitioner

johnc1231:
I tried adding clearer names, but I thought that talking about "new" vs "old" was relatively clear. I may be too deep in it to judge these days.

johnc1231:
Alright, think names are clear now.

hail/src/main/scala/is/hail/linalg/BlockMatrix.scala
```diff
@@ -1559,7 +1559,13 @@ private class BlockMatrixTransposeRDD(bm: BlockMatrix)
```
danking, line 1559: inverseTransposePI is also the wrong name

hail/src/main/scala/is/hail/linalg/GridPartitioner.scala
```diff
@@ -111,43 +112,46 @@ case class GridPartitioner(blockSize: Int, nRows: Long, nCols: Long, maybeBlocks
...
+    partitionIndexToBlockIndex match {
       case Some(bis) =>
-        val (biTranspose, piTranspose) = bis.map(transposeBI).zipWithIndex.sortBy(_._1).unzip
-        val inverseTransposePI = piTranspose.zipWithIndex.sortBy(_._1).map(_._2)
-        
-        (GridPartitioner(blockSize, nCols, nRows, Some(biTranspose)), inverseTransposePI)
-      case None => (gpT, inverseTransposeBI)
+        def transposeBI(bi: Int): Int = gpT.coordinatesBlock(this.blockBlockCol(bi), this.blockBlockRow(bi))
+
+        val (partIdxToBlockIdxT, partIdxTToPartIdx) = bis.map(transposeBI).zipWithIndex.sortBy(_._1).unzip
```
danking, line 124: this is `partIdxT...` It's the transposed partition index to the transposed block index.

The `partIdxToBlockIdxT` would be:
```
bis.map(transposeBi)
```

hail/python/test/hail/linalg/test_linalg.py
```diff
@@ -5,12 +5,38 @@
...
 from hail.expr.expressions import ExpressionException
 
 setUpModule = startTestHailContext
 tearDownModule = stopTestHailContext
 
 
+def sparsify_numpy(np_mat, block_size, blocks_to_sparsify):
+    n_rows, n_cols = np_mat.shape
+    target_mat = np.zeros((n_rows, n_cols))
+    n_block_rows = math.ceil(n_rows / block_size)
```
danking, line 18: for future reference, we usually write this as `(n_rows + block_size - 1) // block_size`

hail/python/test/hail/linalg/test_linalg.py
```diff
@@ -991,3 +1017,56 @@ def test_filtering(self):
...
+
+    @skip_unless_spark_backend()
+    def test_sparsify_blocks(self):
+        block_list = [1, 2]
+        np_square = np.arange(16, dtype=np.float64).reshape((4, 4))
+        block_size = 2
+        bm = BlockMatrix.from_numpy(np_square, block_size=block_size)
+        bm = bm._sparsify_blocks(block_list)
+        sparse_numpy = sparsify_numpy(np_square, block_size, block_list)
+        assert np.array_equal(bm.to_numpy(), sparse_numpy)
```
danking, line 1030: I think it's worth adding this line for this small case as it aids others who come to these tests trying to understand what's happening:
```python3
assert np.array_equal(
    sparse_numpy,
    np.array([[0,  0,  3, 4],
              [0,  0,  7, 8],
              [9,  10, 0, 0],
              [13, 14, 0, 0]]))
```

catoverdrive:
@danking fyi- looks like the test case you added is failing?

danking:
I failed at counting!

PR #8868, opened 2020-05-27
[query] Fix MakeNDArray prune rules

tpoterba:
Also add more typechecking throughout. Might need to disable that.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
@@ -1083,10 +1083,13 @@ object PruneDeadFields {
           bodyEnv.deleteEval(valueName),
           memoizeValueIR(a, TStream(valueType), memo)
         )
-      case MakeNDArray(data, _, _) =>
-        val dataType = data.typ.asInstanceOf[TArray]
+      case MakeNDArray(data, shape, rowMajor) =>
         val elementType = requestedType.asInstanceOf[TNDArray].elementType
-        memoizeValueIR(data, TArray(elementType), memo)
+        unifyEnvs(
+          memoizeValueIR(data, TArray(elementType), memo),
+          memoizeValueIR(shape, shape.typ, memo),
+          memoizeValueIR(rowMajor, rowMajor.typ, memo)
+        )
       case NDArrayMap(nd, valueName, body) =>
         val ndType = nd.typ.asInstanceOf[TNDArray]
         val bodyEnv = memoizeValueIR(body, requestedType.asInstanceOf[TNDArray].elementType, memo)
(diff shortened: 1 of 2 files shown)


hail/src/main/scala/is/hail/expr/ir/Pretty.scala
```diff
@@ -252,7 +252,7 @@ object Pretty {
             case AggLet(name, _, _, isScan) => prettyIdentifier(name) + " " + prettyBooleanLiteral(isScan)
             case TailLoop(name, args, _) => prettyIdentifier(name) + " " + prettyIdentifiers(args.map(_._1).toFastIndexedSeq)
             case Recur(name, _, t) => prettyIdentifier(name) + " " + t.parsableString()
-            case Ref(name, _) => prettyIdentifier(name)
+            case Ref(name, _) => prettyIdentifier(name) + " " + ir.typ.toString
```
catoverdrive, line 255: Let's back this out for legibility in the logs?

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -1429,6 +1429,8 @@ case class TableMapRows(child: TableIR, newRow: IR) extends TableIR {
   override def partitionCounts: Option[IndexedSeq[Long]] = child.partitionCounts
 
   protected[ir] override def execute(ctx: ExecuteContext): TableValue = {
+    val e = Env[Type]("row" -> child.typ.rowType, "global" -> child.typ.globalType)
```
catoverdrive, line 1432: I think we want to back this out in favor of the typechecks you've already added in lowering passes.

hail/src/main/scala/is/hail/expr/ir/lowering/LoweringPipeline.scala
```diff
@@ -12,12 +12,16 @@ case class LoweringPipeline(lowerings: IndexedSeq[LoweringPass]) {
...
+    TypeCheck.withContext(x, "initial check")
+
+    if (optimize) {
       x = Optimize(x, noisy = true, context = "Lowerer, initial IR", ctx)
+    }
 
     lowerings.foreach { l =>
       try {
         x = l.apply(ctx, x)
+        TypeCheck.withContext(x, l.context)
```
catoverdrive, line 24: Do we want to do this once for each pass, after the optional Optimize? Doesn't give us the granularity we necessarily want, but I'm thinking maybe for non-testing purposes the most important typecheck is right before execution?

hail/src/main/scala/is/hail/expr/ir/Pretty.scala
tpoterba: oh, this is invalid anyway

tpoterba:
I'm gonna back out all of the typechecking stuff and introduce that in a separate PR

PR #8875, opened 2020-05-27
[batch] upgrade disk image and add dockerd debug mode

jigold:
- changed worker image from batch-worker-7 to batch-worker-8
- upgraded version of ubuntu from ubuntu-minimal-1804-bionic-v20191024 to ubuntu-minimal-1804-bionic-v20200520
- upgraded docker 
- added debug = true mode to docker daemon

### DIFF
--- a/batch/build-batch-worker-image-startup.sh
+++ b/batch/build-batch-worker-image-startup.sh
@@ -6,6 +6,7 @@ apt-get install -y \
     apt-transport-https \
     ca-certificates \
     curl \
+    jq \
     software-properties-common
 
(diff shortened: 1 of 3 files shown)


batch/build-batch-worker-image-startup.sh
```diff
@@ -6,6 +6,7 @@ apt-get install -y \
     apt-transport-https \
     ca-certificates \
     curl \
+    jq \
```
danking, line 9: Thought: we could store the metadata as JSON now instead of in a long list of metadata keys. That might make the batch worker start up code easier to follow. Also reduces the number of network requests when a worker starts.

PR #8876, opened 2020-05-27
[query] make BGzipOutputStream close idempotent by only writing empty footer …

cseed:
…block once


tpoterba:
fix looks good, want a changelog message tho

PR #8878, opened 2020-05-28
[query] add valid memory checks

cseed:
I suspect some of the inconsistent behavior we're seeing could be due to memory corruption.  So I put in another debugging allocator.

What does this do?  It makes sure all memory accesses in `Memory` are valid.  Also, for each allocation, it puts a sentinel values before and after the allocation, and verifies they are undisturbed on free.

How will this work normally?  Obviously, this will slow things down.  This checked `Memory` will be stored outside the main source, and can be copied over `Memory` to run with checked memory.  Once this is passing, I will organize it that way.  We should probably always run a version of the tests with memory checking enabled.

Am I seeing failures?  Yes, a handful.  Unfortunately, the failures themselves don't seem context dependent, and when I run all the tests things fail, but when I run the isolated test, they pass.

Getting this on the board while we debug it.

### DIFF
--- a/hail/src/main/scala/is/hail/annotations/Memory.java
+++ b/hail/src/main/scala/is/hail/annotations/Memory.java
@@ -3,172 +3,257 @@ package is.hail.annotations;
 import sun.misc.Unsafe;
 
 import java.lang.reflect.Field;
+import java.util.Map;
+import java.util.TreeMap;
 
 @SuppressWarnings("sunapi")
-public final class Memory {
+public class Memory {
     private static final Unsafe unsafe;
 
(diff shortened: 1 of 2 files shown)


cseed:
Failures:
 - a Java test I created that should fail to make sure the memory checks are happening
 - Python test_ld_score
 - A docs test

johnc1231:
@cseed The only test failing now is that Java test, so we should be at a point where the next thing to think about is how to have a separate "debug" build of Hail that uses the checked allocator. And then we'd specify that the failing test should only run on the checked allocator.

cseed:
I'm going to close this in favor of @johnc1231's work in progress.

PR #8879, opened 2020-05-28
[query] Include file in metadata parse error

tpoterba:


cseed:
Not necessary anymore.

PR #8883, opened 2020-05-28
[hail] remove maybeGZipAsBGZip

cseed:
The build will fail until TextMatrixReader is fixed.


cseed:
This should be ready for review now.

PR #8884, opened 2020-05-28
[query] Use GenericLines for TextMatrixReader

tpoterba:


hail/src/main/scala/is/hail/io/TextMatrixReader.scala
```diff
@@ -197,9 +198,9 @@ object TextMatrixReader {
     val separator = params.separatorStr.charAt(0)
     val rowFields = params.rowFieldsStr.mapValues(IRParser.parseType(_))
     val entryType = TStruct("x" -> IRParser.parseType(params.entryTypeStr))
-    val resolvedPaths = fs.globAll(params.paths)
+    val fileStatuses = fs.globAll(params.paths).flatMap(fs.listStatus)
```
cseed, line 201: Use `globalAllStatuses`.  This hits the FS twice unnecessarily.

hail/src/main/scala/is/hail/io/TextMatrixReader.scala
```diff
@@ -211,27 +212,29 @@ object TextMatrixReader {
...
-    val lines = HailContext.maybeGZipAsBGZip(fs, params.gzipAsBGZip) { () =>
-      val localOpts = opts
-      SparkBackend.sparkContext("TextMatrixReader.fromJValue").textFilesLines(resolvedPaths, params.nPartitions.getOrElse(backend.defaultParallelism))
-        .filter(line => line.value.nonEmpty && !localOpts.isComment(line.value))
-    }
+
+    val lines = GenericLines.read(fs, fileStatuses, params.nPartitions, None, None, params.gzipAsBGZip, false)
+      .toRDD()
+      .filter { line =>
+        val l = line.toString
```
cseed, line 235: So `toString` is relatively expensive (constructs the string).  In my code, I pushed the filter into the the next step.  If that's not possible, options might be:
 - make GenericLine memoize toString (this is a bit subtle because it is mutable),
 - map the RDD to a (GenericLine, String) pair first
I probably prefer the first because it avoids the allocation.  Let me know if you want help with this.

hail/src/main/scala/is/hail/io/TextMatrixReader.scala
```diff
@@ -310,27 +313,58 @@ class TextMatrixReader(
       PType.canonical(requestedType.globalType, required = true).asInstanceOf[PStruct]
   }
 
+  def executeGeneric(ctx: ExecuteContext): GenericTableValue = {
+    val fs = ctx.fs
+
+    val tt = fullMatrixType.toTableType(LowerMatrixIR.entriesFieldName, LowerMatrixIR.colsFieldName)
+
+    val lines = GenericLines.read(fs, fileStatuses, params.nPartitions, None, None, params.gzipAsBGZip, false)
```
cseed, line 321: It looks like we're reading twice?  Is that necessary?

hail/src/main/scala/is/hail/io/TextMatrixReader.scala
tpoterba: > make GenericLine memoize toString (this is a bit subtle because it is mutable),

Tried this without realizing it was mutated -- took it out.

toString is only used once here, except in `checkHeaders` which will only call it on one line per partition anyway.

hail/src/main/scala/is/hail/io/TextMatrixReader.scala
tpoterba: fixed.

hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
```diff
@@ -107,7 +107,7 @@ object GenericLines {
             return
           }
 
-          line.offset = offset
+          line.setOffset(offset)
```
cseed, line 110: I just hoisted the set offset to the top, but the real logic is when we set lineLength.  I'd propose something like:

```
class GenericLine {
  def setLine(newOffset: Int, newLength: Int): Unit = {
    _offset = newOffset
    _lineLength = newLength
    _str = null
  }
```

And then call it (including with the offset) in the code paths below.

You'll probably still need to capture the offset here:

```
  val startOffset = offset
```

because decoding the line will change the offset.

hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
```diff
@@ -296,12 +296,23 @@ object GenericLines {
...
+  private var strMemo: String = null
+
+  def setOffset(newOffset: Long): Unit = {
+    strMemo = null
+    _offset = newOffset
+  }
+
+  def offset: Long = _offset
+
+  override def toString: String = if (strMemo != null)
```
cseed, line 313: Just me, but I prefer:

```
  if (_str == null) {
    _str = ...
  }
  _str
```

You save a `_str`.

hail/src/main/scala/is/hail/io/TextMatrixReader.scala
cseed: It's used again in the actual parsing of the line:

 > linesBody(lc).filter(_.toString.nonEmpty)

tpoterba: yeah,  realized that. I'm memoizing the string

PR #8888, opened 2020-05-29
[query] Query Batch backend

cseed:
It lives!

I couldn't get P12 to work for the trust store, so I switched back to JKS like your shuffle code and create the trust stores with keytool.


hail/python/hail/utils/java.py
```diff
@@ -40,7 +40,14 @@ def hc():
...
                 from hail.context import init_service
-                init_service()
+                billing_project = os.environ['HAIL_BILLING_PROJECT']
+                if billing_project is None:
+                    raise ValueError('''\
+Cannot initialize Hail Query service backend without a billing
+project.  Call hl.init_service explicitly or set the
+HAIL_BILLING_PROJECT environment variable.
+''')
+                init_service(billing_project)
```
danking, line 50: I think you should just remove these changes, right? `ServiceBackend.__init__` will display a useful message.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -51,16 +125,84 @@ class ServiceBackend() extends Backend {
 
   def defaultParallelism: Int = 10
 
-  def broadcast[T: ClassTag](_value: T): BroadcastValue[T] = new BroadcastValue[T] {
+  def broadcast[T: ClassTag](_value: T): BroadcastValue[T] = new BroadcastValue[T] with Serializable {
     def value: T = _value
   }
```
danking, line 130: I don't think `BroadcastValue` should be serializable itself, right?

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -51,16 +125,84 @@ class ServiceBackend() extends Backend {
...
+    log.info(s"parallelizeAndComputeWithIndex: token $token: running job")
+
+    val batchClient = BatchClient.fromSessionID(backendContext.sessionID)
+    // FIXME check return
+    val batch = batchClient.run(
+      JObject(
+        "billing_project" -> JString(backendContext.billingProject),
+        "n_jobs" -> JInt(n),
+        "token" -> JString(token)),
+      jobs)
```
danking, line 196: seems like we should assert that every job succeeded, right?

hail/src/main/scala/is/hail/io/fs/GoogleStorageFS.scala
```diff
@@ -107,12 +107,11 @@ class GoogleStorageFS(serviceAccountKey: String) extends FS {
 
   @transient private var codecs: IndexedSeq[hadoop.io.compress.CompressionCodec] = _
 
-  private val defaultHadoopConf = new hadoop.conf.Configuration()
-
   def createCodecs(): Unit = {
     if (codecs != null)
       return
 
+    val defaultHadoopConf = new hadoop.conf.Configuration()
```
danking, line 114: what's this about?

hail/src/main/scala/is/hail/services/batch_client/BatchClient.scala
```diff
@@ -32,19 +33,50 @@ class ClientResponseException(
...
     if (body != null)
       req.asInstanceOf[HttpEntityEnclosingRequest].setEntity(body)
-    Tokens.get.addServiceAuthHeaders("batch", req)
+
+    tokens.addServiceAuthHeaders("batch", req)
 
     retryTransientErrors {
       using(httpClient.execute(req)) { resp =>
         val statusCode = resp.getStatusLine.getStatusCode
+        log.info(s"request: response $statusCode")
```
danking, line 79: If we intend to keep this I think we should at least note the URI

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -1,25 +1,88 @@
...
+      os.write(result)
+    }
+  }
+}
+
+class ServiceBackendContext(
+  val username: String,
+  val sessionID: String,
+  val billingProject: String
+) extends BackendContext with Serializable
```
danking, line 79: Should this be serializable? It contains the most secret thing a user has: the sessionID. I almost want to override toString to specifically not print the sessionID just to head off someone unintentionally doing that in the future.

hail/src/main/scala/is/hail/services/tls/package.scala
```diff
@@ -0,0 +1,73 @@
...
+
+package object tls {
+  lazy val log: Logger = LogManager.getLogger("is.hail.tls")
+
+  private[this] lazy val _getSSLConfig: SSLConfig = {
+    var configFile = System.getenv("HAIL_SSL_CONFIG_FILE")
+    if (configFile == null)
+      configFile = "/ssl-config/ssl-config.json"
+    if (!new File(configFile).isFile)
+      throw new NoSSLConfigFound(s"no java ssl config file found at $configFile")
```
danking, line 39: strike java, it's just a standard hail ssl config, right?

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
cseed: It needs to be.

Right now, I'm just serializing broadcast values.  The idiom is:

```
  val valueBc = backend.broadcast(value)
  ...  { x => (fx, valueBc) } ...
```

where the closure is executed on the workers.  So the broadcast needs to be serializable.

In the future, it will be serializable, but the deserializer will be custom and it will load the through a broadcast mechanism (Google storage or some distributed memory store or whatever).

Since we currently storing the serialized function Google Storage and running 1 core per batch job, there's actually no benefit to a broadcast mechanism at the moment.

hail/build.gradle
```diff
@@ -274,7 +274,6 @@ tasks.withType(ShadowJar) {
     relocate 'org.codehaus.jackson', 'is.hail.relocated.org.codehaus.jackson'
     relocate 'org.apache.commons.lang3', 'is.hail.relocated.org.apache.commons.lang3'
     relocate 'org.apache.commons.io', 'is.hail.relocated.org.apache.commons.io'
-    relocate 'org.apache.httpcomponents', 'is.hail.relocated.org.apache.httpcomponents'
```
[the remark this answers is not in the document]
cseed: As far as I can tell, nothing lives at `org.apache.httpcomponents`.  `httpcomponents` is the group of the dependency, but it lives in the Java package `org.apache.http`.  In particular, `httpcomponents` does not appear in our codebase outside of build.gradle (e.g. not in Scala import statements).

hail/src/main/scala/is/hail/io/fs/GoogleStorageFS.scala
cseed: The codec interface had some legacy dependence on hadoop.  That was just ripped out by a separate PR, so this is gone and the GSFS is "clean" now.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
cseed: Yeah, you're right.  I'll try without it and see if/where it fails.  I think you're right tho, it shouldn't have to leave the driver.

It isn't a case class so toString won't print the contents:

```
scala> class A(val sensitive: Int)
defined class A

scala> val a = new A(5)
a: A = A@5fb97279
```

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
cseed: I throw `RuntimeException` if `state != "success"`.  This is bare bones, there's going to be more work (1) getting a nice error message to the user (file not found, array index exception, etc.) and (2) logging well so we can debug system errors.

PR #8890, opened 2020-05-29
[query] fix incorrect uses of RVD key in TabeIR

patrick-schultz:
In a TableValue, the RVD key may be longer than the TableType key, so it's wrong for the row type of the result of a TableIR execute to depend on the RVD key. I tried to find all cases of this error in TableIR execute methods, and only found these two.


patrick-schultz:
I wasn't able to reproduce, but I didn't try writing. I'll try that.

PR #8892, opened 2020-05-29
Lower TableKeyByAndAggregate

johnc1231:
This PR lowers `TableKeyByAndAggregate`. The tests are set to only run in the interpreter though, since method splitting issues cause them to fail currently. If I turn off method splitting, the tests I've added pass.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/lowering/LowerDistributedSort.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/lowering/LowerDistributedSort.scala
@@ -37,7 +37,7 @@ object LowerDistributedSort {
 
     val ord: Ordering[Annotation] = ExtendedOrdering.rowOrdering(sortColIndexOrd).toOrdering
     val rows = rowsAndGlobal.getAs[IndexedSeq[Annotation]](0)
-    val kType = TStruct(sortFields.takeWhile(_.sortOrder == Ascending).map(f => (f.field, rowType.virtualType.fieldType(f.field))): _*)
+    val kType = TStruct(sortFields.map(f => (f.field, rowType.virtualType.fieldType(f.field))): _*)
     val kIndex = kType.fieldNames.map(f => rowType.fieldIdx(f))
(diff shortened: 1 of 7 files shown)


hail/src/main/scala/is/hail/expr/ir/lowering/LowerDistributedSort.scala
```diff
@@ -37,19 +37,25 @@ object LowerDistributedSort {
...
+    val partitioner = new RVDPartitioner(partitionerKeyType,
       sortedRows.grouped(itemsPerPartition).map { group =>
-        val first = group.head.asInstanceOf[Row].select(kIndex)
-        val last = group.last.asInstanceOf[Row].select(kIndex)
+        val first = group.head.asInstanceOf[Row].select(partitionerKeyIndex)
+        val last = group.last.asInstanceOf[Row].select(partitionerKeyIndex)
         Interval(first, last, includesStart = true, includesEnd = true)
       }.toArray,
-      allowedOverlap = kType.size)
+      allowedOverlap = partitionerKeyType.size)
```
patrick-schultz, line 58: This should just leave out the allowedOverlap argument.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -298,8 +295,56 @@ object LowerTableIR {
...
+
+          val withNewKeyFields = loweredChild.mapPartition { partition =>
+            mapIR(partition) { partitionElement =>
+              Let("row",
+                partitionElement,
+                bindIR(newKey) { newKeyRef =>
+                  val getKeyFields = newKeyType.fieldNames.map(fieldName => fieldName -> GetField(newKeyRef, fieldName)).toIndexedSeq
+                  val getOtherFields = oldRowFieldsNotInNewKey.map(fieldName => fieldName -> GetField(partitionElement, fieldName)).toIndexedSeq
+                  MakeStruct( getKeyFields ++ getOtherFields)
+                }
```
patrick-schultz, line 318: I think the most direct way to write this is using `InsertFields`, which handles overwriting existing fields for you:
```
InsertFields(partitionElement, getKeyFields)
```

PR #8894, opened 2020-05-30
[hail] add hl.shuffle

cseed:
CHANGELOG: added shuffle function

### DIFF
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -5711,3 +5711,28 @@ def _values_similar(left, right, tolerance=1e-6, absolute=False):
 def _locus_windows_per_contig(coords, radius):
     rt = hl.ttuple(hl.tarray(hl.tint32), hl.tarray(hl.tint32))
     return _func("locus_windows_per_contig", rt, coords, radius)
+
+
+@typecheck(a=expr_array(),
+           seed=nullable(int))
+def shuffle(a, seed: int = None) -> ArrayExpression:
+    """Randomly permute an array
+
+    Example
+    -------
+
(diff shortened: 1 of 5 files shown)


hail/python/hail/expr/functions.py
```diff
@@ -5711,3 +5711,28 @@ def _values_similar(left, right, tolerance=1e-6, absolute=False):
...
+    a : :class:`.ArrayExpression`
+        Array to permute.
+    seed : :obj:`int`, optional
+        Random seed.
+
+    Returns
+    -------
+    :class:`.ArrayExpression`
+    """
+    return sorted(a, key=lambda _: hl.rand_unif(0.0, 1.0, seed=seed))
```
konradjk, line 5738: `hl.sorted`

tpoterba: it's defined in this file

tpoterba: you'd need to use `builtins.sorted` to get python base `sorted`.

konradjk: oh i see. isn't it kinda bad form to override the builtins? i guess you're doing that a lot anyway...

tpoterba: we only override the builtins in our code, not user code (unless people do `from hail import *`, which is definitely bad form). I think the predictability is a big win: `len` -> `hl.len`, `sorted` -> `hl.sorted`, etc

tpoterba: (except for all the things we can't override, like and/or/if, of course)

konradjk: yeah agreed it's a big win. i just wasn't sure if there was a better way to do this (like call it _sorted, but then be able to put it as hl.sorted), but i guess not

cseed:
Thanks, fixed!

PR #8903, opened 2020-06-01
Fix type annotations and checks in functions.py

nawatts:
hail/python/hail/expr/functions.py defines functions `str`, `int`, etc. There are a few uses of `str` or `int` in functions.py that refer to `hail.expr.functions.str` and `hail.expr.functions.int` where they are intended to refer to `builtins.str` or `builtins.int`.

Empty collection constructors are currently documented as accepting either a Hail type object or a `hail.expr.functions.str` instead of a `builtins.str` for the type name.
https://hail.is/docs/0.2/functions/constructors.html#hail.expr.functions.empty_array

In the case of `shuffle`, the type check as well as the type annotation is incorrect. Currently, attempting to call `hl.shuffle` with a `seed` argument throws an exception `TypeError: int() missing 1 required positional argument: 'x'`.

---

There is also a more widespread but much less impactful issue with the docstrings in functions.py.   Sphinx interprets ``:obj:`str` `` in a docstring in functions.py to refer to `hail.expr.functions.str`, so the documentation for many parameters of builtin types links to `hail.expr.functions` functions. For example, `str` in the description of the `reference_genome` parameter for `hl.locus` links to `hl.expr.functions.str`, though that parameter is actually of type `builtin.str`.
https://hail.is/docs/0.2/functions/genetics.html#hail.expr.functions.locus

### DIFF
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -5717,8 +5717,8 @@ def _locus_windows_per_contig(coords, radius):
 
 
 @typecheck(a=expr_array(),
-           seed=nullable(int))
-def shuffle(a, seed: int = None) -> ArrayExpression:
+           seed=nullable(builtins.int))
+def shuffle(a, seed: builtins.int = None) -> ArrayExpression:
     """Randomly permute an array
 
     Example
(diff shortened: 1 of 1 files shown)


danking:
CI doesn't recognize your GitHub account, I'm looking into why.

danking:
Should be fixed shortly. Apologies.

nawatts:
Thanks @danking. And no worries, I was mainly interested in being able to see test output vs having tests automatically start.

PR #8911, opened 2020-06-02
[Batch] Add cookbook for ld-based clumping

jigold:
I can send you a tar file with how it looks when Sphinx builds it if needed.


hail/python/hailtop/batch/docs/cookbook/clumping.rst
```diff
@@ -0,0 +1,293 @@
...
+A couple of things to note about this function:
+
+ - We use the image `'hailgenetics/genetics:0.2.37'` which is a publicly available Docker
+   image from Dockerhub maintained by the Hail team that contains many useful bioinformatics
+   tools including PLINK.
+
+ - We explicitly tell PLINK to only use 1Gi of memory because otherwise PLINK tries to
+   reserve half of the memory available on the machine which is not the same amount as what Batch
+   actually gives the job. Not specifying resource requirements correctly can cause performance
+   degradations with PLINK.
```
danking, line 197: I would phrase this differently:

> We explicitly tell PLINK to only use 1Gi of memory because PLINK defaults to using half of the machine's memory. PLINK's memory-available detection mechanism is unfortunately unaware of the memory limit imposed by Batch. ...

hail/python/hailtop/batch/docs/cookbook/clumping.rst
```diff
@@ -0,0 +1,293 @@
...
+ - PLINK creates a hard-coded file `plink.clumped`. We have to move that file to a temporary
+   Batch file `{c.clumped}` in order to use that file in downstream jobs.
+
+
+Merge Clumping Results
+~~~~~~~~~~~~~~~~~~~~~~
+
+The third function concatenates all of the clumping results per chromosome into a single file
+with one header line. The inputs are the :class:`.Batch` for which to create a new :class:`.Job`
+and a list containing all of the individual clumping results files. We use the `'ubuntu:18.04'`
```
danking, line 208: I think I'd use double-backticks here as well.

hail/python/hailtop/batch/docs/cookbook/clumping.rst
```diff
@@ -0,0 +1,293 @@
...
+
+    m = merge(batch, results)
+    batch.write_output(m.ofile, 'gs://<MY_BUCKET>/batch-clumping/1kg-caffeine-consumption.clumped')
+
+The last thing we do is submit the Batch to the service and then close the Backend:
+
+.. code-block:: python
+
+    batch.run(open=True, wait=False)  # doctest: +SKIP
+    backend.close()
```
danking, line 293: we should include the Dockerfile and complete script at the bottom for easy copy-pasting.

PR #8917, opened 2020-06-03
[query] Some refactoring of TableStage

patrick-schultz:
~Stacked on #8874, not by necessity, but because the following PR lowering TableJoin depends on both this and that.~

* Before, `TableStage` defined it's per-partition behavior in an abstract method `def partition(ctxRef: Ref): IR`. This makes defining a new `TableStage` a bit heavy syntactically, and means we can't inspect the type of the partition IR without apply the function to something.

  This PR replaces the abstract method with two fields `ctxRefName: String` and `partitionIR: IR`. It defines a method `def partition(ctx: IR): IR` using these, which is more ergonomic to use because the context isn't forced to be a `Ref`.

* With the type of the partition result accessible, this PR requires that type to be a `TStream<TStruct>`. Without this requirement, there is no clear connection between the partitioner and the rest of the `TableStage`. The only violation of this requirement was mapping to some other type right before collecting; this use is accommodated by a `mapCollect` method which combines the steps.

* The binding structure of `TableStage` has been slightly reorganized. The `letBindings`, which are used on the master, and the `broadcastVals` which are used on the driver (previously, they had to also be usable on master), have been teased apart. In the new structure:
  * `letBindings` are as before: a sequence of bindings which are evaluated in sequence on the master, whose bindings are visible in the `contexts` expression and in the `broadcastVals`
  * `broadcastVals` are now a separate sequence of bindings, which are evaluated on the master in parallel (each broadcast binding sees only the `letBindings`, no previous `broadcastVals`), and whose bindings are visible only in the `partitionIR`
  * `globals` is now required to be a `Ref`, which is defined in `letBindings` and redefined in `broadcastVals`, so that the `Ref` is valid in later `letBindings`, in `contexts`, as well as in the `partitionIR`. This does mean that `globals` is always broadcast, even when it's just an empty struct. But since we're generating a `CollectDistributedArray` which always broadcasts a struct, adding a nested required empty struct shouldn't have any overhead.

  I think this more layered organization has a clearer semantics than before, where `broadcastVals` were a subset of the `letBindings`, and `globals` was an arbitrary IR in the `broadcastVals` scope. It's also closer to the structure of the resulting `CollectDistributedArray`.

  I tried to pull `letBindings` out of `TableStage`, letting them be built up imperatively in the lowering pass, similar to what we do in `LowerMatrixIR`. I had it mostly working, but was blocked by `localSort`, which compiles and executes an intermediate `TableStage`. This needs to be able to compile only those bindings needed in that intermediate stage, and to throw out those used bindings, neither of which is easy when the bindings are accumulated in a single list throughout the lowering pass.

@catoverdrive Would appreciate if you could take a quick look as well.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
@@ -1,11 +1,10 @@
 package is.hail.expr.ir.lowering
 
 import is.hail.expr.ir._
-import is.hail.types
 import is.hail.types.virtual._
 import is.hail.methods.{ForceCountTable, NPartitionsTable}
 import is.hail.rvd.{AbstractRVDSpec, RVDPartitioner}
-import is.hail.types.RTable
+import is.hail.types.{RTable, TableType}
 import is.hail.utils._
 import org.apache.spark.sql.Row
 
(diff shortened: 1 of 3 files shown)


tpoterba:
my initial sense is that this is too big to review effectively -- there are reorgs mixed in with things like abstracting `zipPartitions`. Can we split it up a bit?

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -19,86 +19,152 @@ object TableStage {
...
+  assert(kType.fields.forall(f => rowType.field(f.name).typ == f.typ))
+  assert(broadcastVals.exists { case (name, value) => name == globals.name && value == globals})
+
+  def partition(ctx: IR): IR = {
+    require(ctx.typ == ctxType)
+    Let(ctxRefName, ctx, partitionIR)
+  }
 
-  def partition(ctxRef: Ref): IR
+  require(kType.fields.forall(f => rowType.field(f.name).typ == f.typ))
```
johnc1231, line 79: This is redundant given the `assert` on line 71, no? Unless `require` and `assert` are different in a way I don't understand.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -19,86 +19,152 @@ object TableStage {
...
-      def partition(ctxRef: Ref): IR = self.partition(ctxRef)
-    }
-  }
+  def collectWithGlobals(): IR =
+    MakeStruct(FastSeq(
+      "rows" -> ToArray(flatMapIR(ToStream(mapCollect(ToArray)))(ToStream)),
+      "global" -> globals))
+
+  def changePartitionerNoRepartition(newPartitioner: RVDPartitioner): TableStage =
+    new TableStage(letBindings, broadcastVals, globals, newPartitioner, contexts, ctxRefName, partitionIR)
```
johnc1231, line 163: Could just do  `this.copy(rvdPartitioner = newPartitioner)` here I think, slightly less noisy.

patrick-schultz:
Sure, no problem.

tpoterba:
Thanks Patrick! I think that'll help us get this in faster.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
tpoterba: require and assert just throw different exception types. Pretty similar for our purposes.

patrick-schultz: Oops, yeah, completely redundant.

I think of require as stating preconditions for a function or class constructor, so a failure is the fault of the caller, while an assertion states an assumption about an internal state, so a failure is the fault of the immediate code. I like require for documentation purposes, explaining what a function assumes about its inputs in a self-checking way. But we've never discussed that as a team, and I think we tend to use assert for everything.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
patrick-schultz: It could if `TableStage` was a Case class, or if we manually define a `copy` method. Now that there's no abstract `partition` method, we could do that, and you're right that would tidy up some things.

patrick-schultz:
Okay, I think all the changes are now directly related to the main changes described in the initial PR message.

johnc1231:
So the PR this is stacked on went in, but I still see changes from that PR in this diff. Maybe needs a rebase?

patrick-schultz:
Yup, sorry about that. I'll get it ready.

patrick-schultz:
rebased

tpoterba:
will get to this today.

tpoterba:
sorry, not quite done review. Will finish Thursday morning.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -19,86 +19,121 @@ object TableStage {
...
   val partitioner: RVDPartitioner,
-  val contexts: IR) { self =>
-
-  def partition(ctxRef: Ref): IR
-
-  def wrapInBindings(body: IR): IR = {
-    letBindings.foldRight(body) { case ((name, binding), soFar) => Let(name, binding, soFar) }
+  val contexts: IR,
+  val ctxRefName: String,
+  val partitionIR: IR) { self =>
```
tpoterba, line 63: this should probably be private.

tpoterba: `def partition(ctx)` should be the interface consumers use to construct the partition stream.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -19,86 +19,121 @@ object TableStage {
...
+  val globals: Ref,
   val partitioner: RVDPartitioner,
-  val contexts: IR) { self =>
-
-  def partition(ctxRef: Ref): IR
-
-  def wrapInBindings(body: IR): IR = {
-    letBindings.foldRight(body) { case ((name, binding), soFar) => Let(name, binding, soFar) }
+  val contexts: IR,
+  val ctxRefName: String,
```
tpoterba, line 62: should be private.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -19,86 +19,121 @@ object TableStage {
...
+    copy(
+      letBindings = letBindings :+ globalsRef.name -> newGlobals,
+      broadcastVals = broadcastVals :+ globalsRef.name -> globalsRef,
+      globals = globalsRef)
   }
 
-  def collect(bind: Boolean = true): IR = {
-    val ctx = Ref(genUID(), types.coerce[TStream](contexts.typ).elementType)
-    val broadcastRefs = MakeStruct(letBindings.filter { case (name, _) => broadcastVals.contains(name) })
+  def mapCollect(f: IR => IR): IR = {
```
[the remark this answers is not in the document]
patrick-schultz: I guess I figured if any user is forced to call `wrapInBindings`, why not just be consistent and always have to. One of the things that I first found confusing about `TableStage` was that it wasn't obvious what added bindings and what didn't. It seemed cleaner in some ways to just always make it explicit. (I really wanted to get rid of `letBindings` and `wrapInBindings` entirely, but I don't think that's possible with local sort as it is currently.)

I refactored it so that `wrapInBindings` can be private, by adding `getGlobals` and `getNumPartitions`, that return self-contained IRs. Still not completely satisfactory, but I think fine for now.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -20,95 +19,138 @@ object TableStage {
...
-  def collectWithGlobals(): IR = {
-    val mapped = mapPartition(ToArray)
-    mapped.wrapInBindings(
-      MakeStruct(FastIndexedSeq(
-        "rows" -> ToArray(flatMapIR(ToStream(mapped.collect(bind = false))) { elt => ToStream(elt) }),
-        "global" -> mapped.globals)))
+  def mapCollectWithGlobals(mapF: IR => IR)(body: (IR, IR) => IR): IR =
+    mapCollectWithContextsAndGlobals((part, ctx) => mapF(part))(body)
+
+  def mapCollectWithContextsAndGlobals(mapF: (IR, Ref) => IR)(body: (IR, IR) => IR): IR = {
```
patrick-schultz, line 124: I had to add this during the rebase, to support the lowered TableWriter without using `wrapInBindings`.

PR #8918, opened 2020-06-03
Clarify usage of hailctl dataproc modify

nawatts:
`hailctl dataproc modify`'s `--update-hail-version` and `--wheel` arguments cannot be used together. Using argparse's [mutually exclusive group](https://docs.python.org/3/library/argparse.html#mutual-exclusion) to enforce that makes the usage/help text clearer.

```
usage: hailctl dataproc modify [-h] [--num-workers NUM_WORKERS]
                               [--num-preemptible-workers NUM_PREEMPTIBLE_WORKERS]
                               [--graceful-decommission-timeout GRACEFUL_DECOMMISSION_TIMEOUT]
                               [--max-idle MAX_IDLE] [--dry-run] [--zone ZONE]
                               [--update-hail-version | --wheel WHEEL]
                               name
```

### DIFF
--- a/hail/python/hailtop/hailctl/dataproc/modify.py
+++ b/hail/python/hailtop/hailctl/dataproc/modify.py
@@ -7,7 +7,6 @@ import pkg_resources
 
 def init_parser(parser):
     parser.add_argument('name', type=str, help='Cluster name.')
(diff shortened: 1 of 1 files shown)


johnc1231:
Neat, didn't know about that

PR #8923, opened 2020-06-04
[site] Move non-docs website files into site; deploy independently

danking:
1. Move `hail/www` to `site/www` and associated build commands into `site/Makefile` and a new `build.yaml` step.
2. Prepare for a simpler docs deployment by supporting both the current 0.2 structure (top-level `www` containing `docs/0.2` and `docs/0.1`) and a future, simpler structure (top-level `docs` containing `0.2` and `0.1`).
3. Fix `site/Makefile` which had bit-rotted. `test` doesn't really work anymore so I removed it.

We could restore `make test` by figuring out a local SSL story. I went down this route but couldn't get NGINX to respond to my HTTPS requests. `make deploy` is rather fast now anyway.

Currently deployed at https://internal.hail.is/dking/site/index.html.

There are two known issues with dev deploy, those are resolved at https://github.com/hail-is/hail/pull/8922.


site/Dockerfile
```diff
@@ -21,5 +21,8 @@ ADD poll.sh /poll.sh
 ADD poll-0.1.sh /poll-0.1.sh
 ADD poll-0.2.sh /poll-0.2.sh
 ADD site.sh /site.sh
+ADD site-www.tar.gz /var/www/html
+RUN chown -R www-data /var/www/html && \
+    chmod -R 555 /var/www/html
```
akotlar, line 26: I think this allows execution of any file in html/, which may not be what we want, and we probably only need it on www-data. I think we want something like chmod -R u+rX,g+rX /var/www/html, where X sets execute on dirs only.

site/poll-0.1.sh
```diff
@@ -12,16 +12,21 @@ if [[ $DEPLOYED_SHA = $LATEST_SHA ]]; then
...
-# just in case
-rm -rf /var/www/0.1-old
+old=$(mktemp -d)
+mkdir -p $old
 
-mv /var/www/0.1 /var/www/0.1-old || true
-mv /var/www/0.1-new /var/www/0.1
-rm -rf /var/www/0.1-old
+chown -R www-data $new
+chmod -R 555 $new
```
akotlar, line 26: same as above permission

akotlar:
Another issue I noticed: viz.min.js isn't found in the deployed internal copy. Assuming the vendors directory was copied, you may have deployed before adding execute bit to that folder.

danking:
Hmm. this works: https://internal.hail.is/dking/site/vendors/vanta/viz.min.js, where is it not getting loaded correctly? I probably need another rule in dev deploy's site's nginx to fix the path.

danking:
Good point about the permissions. I went with `=rX` since I also want to remove any write permissions.

danking:
I updated the dev deploy too.

akotlar:
403 forbidden on that resource for me

akotlar:
To be clear, not the link you provided, what the page pulls.

akotlar:
Ah this is obvious: https://internal.hail.is/vendors/vanta/viz.min.js

akotlar:
Or may be easier: provide a symlink'd folder at the original location

danking:
OK, so. Dev deployed `site` uses the same docs and www as non-dev-deployed site. Unfortunately, this means we have a root-URL issue. I use [an nginx `sub_filter`](https://github.com/hail-is/hail/blob/master/site/site.sh) rule to fix links and anchors. I didn't add any rules for javascript modules.

The right fix is to move to a system that can build for different environments and set the root URL properly. Until we switch to that hypothetical new system, I'll add a rule that properly rewrites JS module imports.

We can't fix this with symlinks or any kind of redirection. The root issue is that we, the users, are outside of the system and the way we specify to whom we're talking is `internal.hail.is/NAMESPACE/SERVICE/`. Cotton wanted to do `service.namespace.internal.hail.is` but there were some TLS wildcard issues. It is perhaps worth revisiting this at some point because it is a perennial issue for us.

akotlar:
I see, right we aren’t changing the url, I forgot that we used namespace/service here, hence site. Yep having something that filters .js requests is right. Do we also need .ico and .png (for header icon, which isn’t in an anchor tag).

danking:
It currently fixes anything with the form `href=['"]/` and `src=['"]/`.

danking:
OK, so that was way more pain than I expected. Apparently when you're importing a JS module which needs authentication, you must specify the `crossorigin` attribute to the `script` tag. If you lack that attribute, no headers are sent. Since dev deploys are limited to developer access only (even though there is no in-dev-namespace authentication), this obviously doesn't work.

Why does the word `crossorigin` mean send headers to the same origin? Who knows! Anyway. I don't think that bug should block this PR. That bug is an underlying issue with dev deploy. The fix for the module import bug is here: https://github.com/hail-is/hail/pull/8928

danking:
I also pushed a commit to make it harder to accidentally `make deploy` into the main namespace. `site`'s `make deploy` now requires a `NAMESPACE` argument to be set.

akotlar:
Sounds good, the url's look correct, and as you say the resource is present (but blocked due to CORS)

akotlar:
Thanks for putting in the work!

PR #8926, opened 2020-06-04
[query][lowering] lower TableWrite with native reader

catoverdrive:
This implementation lowers TableWrite with a TableNativeReader.

Punting on the `stageLocally` path for now (it'll throw a lowering error) since our current implementation adds a task listener to the spark task to clean up files, and I'm not sure how we want to handle that in the general case.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
@@ -169,17 +169,17 @@ case class PartitionNativeWriter(spec: AbstractTypedCodecSpec, partPrefix: Strin
             { pc =>
               val row = pc.memoize(cb, "row")
               if (hasIndex) {
-                val annotation = IEmitCode.present(cb, PCode(+PCanonicalStruct(), 0L))
                 val keyRVB = new StagedRegionValueBuilder(mb, keyType)
-                val key = IEmitCode.present(cb, {
+                indexWriter.add(cb, {
                   cb += keyRVB.start()
                   keyType.fields.foreach { f =>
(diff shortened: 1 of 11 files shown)


hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
```diff
@@ -101,7 +148,111 @@ case class PartitionNativeWriter(spec: AbstractTypedCodecSpec, partPrefix: Strin
     eltType: PStruct,
     mb: EmitMethodBuilder[_],
     region: Value[Region],
-    stream: SizedStream): EmitCode = ???
+    stream: SizedStream): EmitCode = {
```
chrisvittal, line 151: Feel free to push back, this looks good, but is there any reason to not use the `scoped` methods for EmitCodeBuilder for a lot of this since we have the necessary EmitMethodBuilder to create the EmitCodeBuilder

hail/src/main/scala/is/hail/io/index/IndexWriter.scala
```diff
@@ -380,4 +386,38 @@ class StagedIndexWriter(branchingFactor: Int, keyType: PType, annotationType: PT
     internalBuilder.create(cb)
     internalBuilder.store(cb, utils, 0)
   }
+
+  class NonImperativeIndexWriter(newInit: EmitMethodBuilder[_], newAdd: EmitMethodBuilder[_], newClose: EmitMethodBuilder[_]) {
```
chrisvittal, line 390: Why is this necessary? It seems like we can create a code builder and use the other version of the index writer.

hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
catoverdrive: No, good call. I forgot that that existed. I'll refactor this.

PR #8928, opened 2020-06-04
[site] when dev deploying also update js module imports

danking:
Not sure why crossorigin is necessary to make authenticated imports work, but it works. No information at MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script.

https://internal.hail.is/dking/site/ demonstrates this working.

### DIFF
--- a/site/site.sh
+++ b/site/site.sh
@@ -14,8 +14,9 @@ EOF
 else
     prefix="/$HAIL_DEFAULT_NAMESPACE/site"
     cat >$rewrite_conf_location <<EOF
-subs_filter href="/([^/]) href="/$HAIL_DEFAULT_NAMESPACE/site/\$1 gr;
-subs_filter src="/([^/]) src="/$HAIL_DEFAULT_NAMESPACE/site/\$1 gr;
+subs_filter href=(['"])/([^/]) href=\$1/$HAIL_DEFAULT_NAMESPACE/site/\$2 gr;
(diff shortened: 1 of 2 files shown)


danking:
grr. that's not even sufficient, apparently I have to also explicitly import it

danking:
ok this is still broken. I'll have to investigate further tomorrow.

akotlar:
Yeah, it appears when site's js requests it, blocked, curl ok

danking:
OK, so, this is apparently an issue where browsers have not yet really implemented the standard correctly. From my reading of [HTML 2.6.4](https://html.spec.whatwg.org/#cors-settings-attributes), `crossorigin="anonymous"` ought to be sufficient for requests to the same origin as the page containing the script tag.

Jake Archibald has an informative [blog post](https://jakearchibald.com/2017/es-modules-in-browsers/) about modules. He links to a [demo](https://module-script-tests-sreyfhwvpq.now.sh/cookie-page) of three cross origin configurations. The three options are:
```
<script type="module" src="cookie-script"></script>
<script type="module" crossorigin="" src="cookie-script?1"></script>
<script type="module" crossorigin="use-credentials" src="cookie-script?2"></script>
```
I'm using Safari Version 13.1 (14609.1.20.111.8). I usually only see the very last script working. However, inexplicably, I have seen the first one very rarely work. All I've been doing is refreshing here and there as I try to understand this. The Safari inspector confirms that the cookie is only sent with thee last option.

So, anyway, I'm adding `crossorigin="use-credentials"`.

danking:
My dev deploy now has the light up graph 🥳

PR #8929, opened 2020-06-05
[query] Make VEP on Dataproc not require a config

johnc1231:
VEP will now automatically look for a config file in environment variable `VEP_CONFIG_URI` if one isn't specified. This environment variable is prepopulated on dataproc clusters started with `hailctl dataproc`


hail/python/hail/methods/qc.py
```diff
@@ -552,6 +553,8 @@ def vep(dataset: Union[Table, MatrixTable], config, block_size=1000, name='vep',
      - ``GRCh37``: ``gs://hail-us-vep/vep85-loftee-gcloud.json``
      - ``GRCh38``: ``gs://hail-us-vep/vep95-GRCh38-loftee-gcloud.json``
 
+     If no config file is specified, this function will check the to see if `VEP_CONFIG_URI` is set with a path to a config file.
```
nawatts, line 556: > will check the to see

It looks like a word may have been left out here.

johnc1231: Fixed, thank you.

PR #8933, opened 2020-06-08
[batch] event is now json, treat it as such

cseed:


batch/batch/driver/instance_pool.py
```diff
@@ -419,18 +419,18 @@ async def handle_call_delete_event(self, instance, timestamp):
...
 
         timestamp = event.timestamp.timestamp() * 1000
-        payload = event.payload
         version = payload['version']
         if version != '1.2':
             log.warning('unknown event verison {version}')
             return
 
-        resource_type = event.resource.type
+        resource_type = event['resource']['type']
```
jigold, line 433: The JSON you sent has `instance` for the resource_type. Is that an expected change based on the example message? Or is it because that message was from a GKE instance?

cseed: That's the jsonPayload resource type!  There are two resource types, and they are different!

```
>>> event['resource']['type']
'gce_instance'
>>> event['jsonPayload']['resource']['type']
'instance'
```

So the existing code looks like it is still correct.

PR #8934, opened 2020-06-08
[batch] change how files are named in the user interface

jigold:
There might be additional PRs, but for now this PR does the following:

- Names folders for jobs based on the job_id computed after ordering the jobs
- Gets rid of double slashes in path names
- Names JobResourceFiles based on the name of the attribute.
Example: j.ofile => /batch/1/ofile
- Names resource groups from jobs as follows:
Example: j.declare_resource_group(ofile={'bed': '{root}.bed'})
/batch/1/ofile.bed
- Names input resource files based on the file name
/batch/inputs/{uid}/my_vcf.bgz
- Names input resource groups two-ways
input = batch.declare_resource_group(vcf=...)
/batch/inputs/{uid}/my_vcf.bgz
/batch/inputs/{uid}.vcf

I tested this locally and on the service to make sure it's doing what I want.

### DIFF
--- a/hail/python/hailtop/batch/resource.py
+++ b/hail/python/hailtop/batch/resource.py
@@ -52,7 +52,6 @@ class ResourceFile(Resource, str):
         self._source = None
         self._output_paths = set()
         self._resource_group = None
-        self._has_extension = False
 
     def _get_path(self, directory):
         raise NotImplementedError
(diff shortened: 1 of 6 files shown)


hail/python/hailtop/batch/resource.py
```diff
@@ -167,10 +132,48 @@ class JobResourceFile(ResourceFile):
...
+        super().__init__(value)
+        self._has_extension = False
+
     def _get_path(self, directory):
         assert self._source is not None
         assert self._value is not None
-        return shq(directory + '/' + self._source._uid + '/' + self._value)
+        return f'{directory}/{self._source._job_id}/{self._value}'
+
+    def add_extension(self, extension):
```
danking, line 144: in other places (e.g. `command`) we don't add a verb, we just state what's being changed. let's stick to that system.

jigold: That would be a breaking change. Let's do it in another PR.

PR #8935, opened 2020-06-08
[query] Add StreamMerge node

patrick-schultz:


hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
```diff
@@ -515,6 +515,75 @@ object Stream {
...
 
+  def merge(
+    mb: EmitMethodBuilder[_],
+    lElemType: PType, left: Stream[EmitCode],
+    rElemType: PType, right: Stream[EmitCode],
+    outElemType: PType, region: Value[Region],
+    comp: (EmitValue, EmitValue) => Code[Int]
+  ): Stream[EmitCode] = new Stream[EmitCode] {
+    def apply(eos: Code[Ctrl], push: EmitCode => Code[Ctrl])(implicit ctx: EmitStreamContext): Source[EmitCode] = {
+      val pulledRight = mb.newLocal[Boolean]()
```
johnc1231, line 526: Talked about this on Zulip. I think renaming `pulledRight` to something like `firstPull` (though I guess that would be the negation) would be helpful to indicate it's something that happens just once.

patrick-schultz: What I find slightly confusing about `firstPull` is that this bit needs to be flipped after the first pull from the left stream, and before the first pull from the right stream. It's true all that happens during the first pull from the merged stream, but having `firstPull` be false while we're comparing the first two values, deciding which is the first value to push up to the consumer, seems confusing. Let me know if the comments I added do enough to clarify things.

PR #8937, opened 2020-06-09
[query] store job configuration in user bucket

cseed:
The problem was query was writing the job configuration to the query bucket, but workers only get the user gsa, so they were unable to read the configuration.  This worked in the tests because the query and user account are both the test service account.

I can remove the query-gsa-key and the hail-query bucket after this goes in.

### DIFF
--- a/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
@@ -21,18 +21,6 @@ import org.json4s.jackson.JsonMethods
 import scala.collection.mutable
 import scala.reflect.ClassTag
 
-object ServiceBackendUtils {
-  def getQueryFS(): FS = {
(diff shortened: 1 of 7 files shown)


hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -76,14 +66,15 @@ object Worker {
...
   val username: String,
   @transient val sessionID: String,
-  val billingProject: String
+  val billingProject: String,
+  val bucket: String
 ) extends BackendContext
 
 object ServiceBackend {
-  lazy val log = LogManager.getLogger("ServiceBackend")
+  lazy val log = LogManager.getLogger("is.hail.backend.service.ServiceBackend")
```
danking, line 74: why this change?

cseed: Consistency.  I used the fully qualified class name elsewhere.  I don't have a huge preference, although there could be name collisions otherwise.

PR #8941, opened 2020-06-09
[query] Fix `import_plink` with n_partitions > 1

tpoterba:
CHANGELOG: Fixed `hl.import_plink` with multiple data partitions.


johnc1231:
Can you briefly explain what the problem was for context?

tpoterba:
we intended to be comparing the locusAlleles struct, but were getting class cast exceptions trying to compare the PlinkVariant class. With 1 partition, this code was never called 😬

tpoterba:
The test I changed catches this now. I think that's sufficient, we shouldn't need a new test if an existing test can be improved to have better coverage.

PR #8949, opened 2020-06-11
[ci] fix clone to actually retry

danking:
I do not understand how `set -e` works in bash functions, but I know what
currently exists does not retry properly. I have verified this works as intended.

It deletes *all* files in the current folder, then tries to clone into this folder.
If the clone fails, becasue it is the last command in the function, the function
will return the same exit code. The retry function will then retry `clone`. Because
`clone` removes all files in the directory before attempting a clone, it is idempotent.

Users of the checkout script all use it by creating an empty directory and cd'ing into it.

### DIFF
--- a/ci/ci/github.py
+++ b/ci/ci/github.py
@@ -140,8 +140,8 @@ def clone_or_fetch_script(repo):
 { RETRY_FUNCTION_SCRIPT }
 
 function clone() {{
-    rm -rf ./{*,.*}
-    git clone {shq(repo)} ./
+    rm -rf ./{{*,.*}}
+    git clone { shq(repo) } ./
 }}
 
 if [ ! -d .git ]; then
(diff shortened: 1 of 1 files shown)


ci/ci/github.py
```diff
@@ -140,8 +140,8 @@ def clone_or_fetch_script(repo):
 { RETRY_FUNCTION_SCRIPT }
 
 function clone() {{
-    rm -rf ./{*,.*}
-    git clone {shq(repo)} ./
+    rm -rf ./{{*,.*}}
```
jigold, line 143: For my education, can you explain what this is doing? Is it matching hidden files?

danking: Yeah. `{..}` is a bash-ism for kind of map-expanding some values:
```
# ls a{bb,cc}d
ls: abbd: No such file or directory
ls: accd: No such file or directory
```

And then `.*` means all the hidden files

PR #8950, opened 2020-06-12
[docs] remove ?highlight

akotlar:
Algolia crawls anchor tags, and treats query strings as distinct urls (since apriori there is no way to know that that doesn't represent a unique page, since query strings can be used for routing).

We don't use ?highlight consistently anyway, and it frankly doesn't add anything useful (it highlights what the page already scrolls to).


akotlar:
I found an alternate solution for Algolia. It may still be worth doing, as any crawler that follows hyperlinks may run into the issue of indexing duplicate pages, but this is a very minor issue.

PR #8952, opened 2020-06-12
EColumnMajorNDArray

johnc1231:
This PR:

- Switches the unused `ENDArray` to be called `EColumnMajorNDArray`, always writes out a column major thing, never writes out strides

- Makes the `fundamentalType` of `PNDArray` be `PNDArray`, and makes required patches throughout code base to facilitate this.


hail/src/main/scala/is/hail/types/encoded/ENDArray.scala
```diff
@@ -1,78 +0,0 @@
-package is.hail.types.encoded
```
tpoterba, line 1: safe to delete because it wasn't being generated. in the future we'll want to move to legacy folder (I know you know this, but want to clarify in case others read)

tpoterba:
Sorry about the slow review, John. Nice work!

hail/src/main/scala/is/hail/annotations/StagedRegionValueBuilder.scala
```diff
@@ -65,14 +71,18 @@ class StagedRegionValueBuilder private (val mb: EmitMethodBuilder[_], val typ: P
     this(er.mb, rowType, er.region, null)
   }
 
-  private val ftype = typ.fundamentalType
+  //private val ftype = typ.fundamentalType
+  private val stype = typ match {
+    case t: PNDArray => t.representation
```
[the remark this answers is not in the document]
johnc1231: Ah, yeah, if an `NDArray`'s `elementType` was not fundamental, the data array would not have gotten fundamentalized. Thanks.

johnc1231:
Addressed the comments, but encountered a small bug when trying to generate the files. Will address shortly

hail/src/main/scala/is/hail/annotations/StagedRegionValueBuilder.scala
```diff
@@ -65,14 +71,17 @@ class StagedRegionValueBuilder private (val mb: EmitMethodBuilder[_], val typ: P
     this(er.mb, rowType, er.region, null)
   }
 
-  private val ftype = typ.fundamentalType
+  private val stype = typ match {
+    case t: PNDArray => t.representation.fundamentalType
```
tpoterba, line 75: wait, what if there's another ndarray in there? I think we need to recursively replace ndarrays with their representation, and otherwise fundamental types? Crap.

johnc1231: Agh, you're right. So I could just write a function called`to_rvb_supported_type` that recursively replaces NDArray's with their structs. Alternatively, I could just make sure every function is ready to accept ndarrays instead of this current thing where I'm trying to replace all the ndarrays at the beginning.

johnc1231: Decided to get rid of this `stype` business. Instead of trying to just replace the `PNDArray` with a `PStruct`, I went ahead and added a `PNDArray` case to all the match statements. No more recursion issues.

johnc1231:
Ok, so the problem of recursively replacing ndarrays with structs prompted me to redo my RVB and SRVB code. Now instead of trying to replace all the `PNDArray`s with `PStruct`s, I handle them explicitly in each match statement. Admittedly the match statements look quite similar to the struct match statements, but I think this was the easier/cleaner path. 

I still have to generate the files. Before doing so, was just wondering if I have to bump the file version first?

johnc1231:
@tpoterba , should be good to go now.

johnc1231:
I also had to increment `n` in `test_backward_compatibility` from `60` to `72`. Seems like it just counts how many tables and matrix tables we read.

johnc1231:
Ugh, looks like somehow we are hitting situations where the length of the data array is not equal to the product of the shape tuple. That's bad, looking into it.

hail/src/main/scala/is/hail/annotations/StagedRegionValueBuilder.scala
```diff
@@ -77,6 +85,7 @@ class StagedRegionValueBuilder private (val mb: EmitMethodBuilder[_], val typ: P
     case t: PArray =>
       elementsOffset = mb.genFieldThisRef[Long]("srvb_array_addr")
       idx = mb.genFieldThisRef[Int]("srvb_array_idx")
+    case t: PNDArray => elementsOffset - mb.genFieldThisRef[Long]("srvb_ndarray_struct_addr")
```
johnc1231, line 88: This should be an equal sign, not a minus sign. Not the source of my problem though.

johnc1231:
Aha, definite reuse: 
```
SRVB.start: Starting array +PCArray[+PInt32] of length 0
SRVB.start: elementsOffset = 5234999304
SRVB.start: Starting array +PCArray[+PInt32] of length 1
SRVB.start: elementsOffset = 5234999304
```

tpoterba:
is that a reuse? If the region is freed in between, one might expect the value in the second row to be allocated in the same spot.

johnc1231:
Yeah, I suppose that's possible too. This just seemed consistent with the fact that I'm seeing the data array of a previous row getting overwritten by the next row.

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
```diff
@@ -218,11 +220,24 @@ final case class PCanonicalNDArray(elementType: PType, nDims: Int, required: Boo
 
   def setRequired(required: Boolean) = if(required == this.required) this else PCanonicalNDArray(elementType, nDims, required)
 
-  def constructAtAddress(mb: EmitMethodBuilder[_], addr: Code[Long], region: Value[Region], srcPType: PType, srcAddress: Code[Long], deepCopy: Boolean): Code[Unit] =
-    this.fundamentalType.constructAtAddress(mb, addr, region, srcPType.fundamentalType, srcAddress, deepCopy)
+  def constructAtAddress(mb: EmitMethodBuilder[_], addr: Code[Long], region: Value[Region], srcPType: PType, srcAddress: Code[Long], deepCopy: Boolean): Code[Unit] = {
+    val srcNDArray = srcPType.asInstanceOf[PNDArray]
+    val srcLocal = mb.newLocal[Long]()
+    val addrLocal = mb.newLocal[Long]()
+    val er = EmitRegion(mb, region)
```
johnc1231, line 227: Delete

tpoterba:
I remember you said this was hitting some error/segfault -- anything you need from me to unblock?

johnc1231:
This one isn't a segfault, it's a test failure where somehow for each ndarray in the table, row i + 1 is getting the data array from row i. Cloud workshop and ndarray aggregator pushed this down my priority list, but I will continue to try and fix it and then get back to you.

tpoterba:
oh! yes, I remember.

johnc1231:
Closed in favor of #9254

PR #8954, opened 2020-06-12
[hail] changes to facilitate changing master to trunk

danking:
We should merge this first to fix CI. It will only break some links while we rename master.

As discussed, let's leave this open until Tuesday lab meeting when we'll all make the switch.


tpoterba:
closing to get off my scorecard until GH makes a decision on naming.

PR #8955, opened 2020-06-13
[site] add documentation search bar, search page

akotlar:
cc @mkveerapen @tpoterba 

1) Adds natural language documentation search bar with autocompletion
2) Adds paginated search page to browse all search results
3) Makes navbar scripts async, to allow inclusion of navbar scripts at top of file without impacting performance (since we include the navbar.html content in other templates we cannot easily locate those scripts at bottom of the includers' pages)
4) Increases hero content width on mobile.
5) Hides sphinx search bar

http://34.207.246.132/

### DIFF
--- a/site/www/template.xslt
+++ b/site/www/template.xslt
@@ -15,6 +15,7 @@
                 <title><xsl:call-template name="page-title"/></title>
                 <link rel='shortcut icon' href='/hail_logo_sq-sm-opt.ico' type='image/x-icon'/>
                 <xsl:call-template name="meta-description"/>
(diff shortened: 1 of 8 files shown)


site/www/template.xslt
```diff
@@ -66,19 +68,20 @@
                             </ul>
                         </div>
                         <script>
-                            <xsl:text disable-output-escaping="yes" >
+                            <xsl:text disable-output-escaping="yes" async="true">
```
danking, line 71: What does it mean for an `xsl:text` to be async?

akotlar:
bump

Issue #8940, opened 2020-06-09
make docs-no-test does not work


nawatts:
The [documented](https://hail.is/docs/0.2/getting_started_developing.html#building-the-docs-and-website) process for building documentation is:
```
cd hail
make docs-no-test
```

That now fails with
```
Warning, treated as error:
html_extra_path entry '/path/to/hail/hail/build/docs/../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../www/hail-logo-cropped.png' does not exist
make[1]: *** [html] Error 2
make: *** [docs-no-test] Error 2
```

It looks like the source of the problem is that docs/conf.py can't find the `www` directory.
https://github.com/hail-is/hail/blob/0b3823af5310a735bc9544fb73308f82426292be/hail/python/hail/docs/conf.py#L225-L232

I'm guessing this is related to changes in #8923.

nawatts:
Looks like this is working again. Thanks @akotlar.

PR #8956, opened 2020-06-13
[docs] fix make-docs

akotlar:
Addresses https://github.com/hail-is/hail/issues/8940

### DIFF
--- a/hail/python/hail/docs/conf.py
+++ b/hail/python/hail/docs/conf.py
@@ -224,7 +224,14 @@ html_static_path = ['_static']
 #
 www_root = os.path.dirname(os.path.abspath(__file__))
 _i = 0
-while _i < 50 and not os.path.isdir(os.path.join(www_root, 'www')):
+while _i < 50:
+    if(os.path.isdir(os.path.join(www_root, 'www'))):
+        break
+
+    if(os.path.isdir(os.path.join(www_root, 'site'))):
+        www_root = os.path.join(www_root, 'site')
+        break
+
     www_root = os.path.join(www_root, '..')
     _i += 1
 html_extra_path = [os.path.join(www_root, 'www/hail-logo-cropped.png'),
--- a/hail/python/hail/docs/conf.py
+++ b/hail/python/hail/docs/conf.py
@@ -222,21 +222,7 @@ html_static_path = ['_static']
 # .htaccess) here, relative to this directory. These files are copied
 # directly to the root of the documentation.
 #
-www_root = os.path.dirname(os.path.abspath(__file__))
-_i = 0
-while _i < 50:
-    if(os.path.isdir(os.path.join(www_root, 'www'))):
-        break
-
-    if(os.path.isdir(os.path.join(www_root, 'site'))):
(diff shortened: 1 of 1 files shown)


hail/python/hail/docs/conf.py
```diff
@@ -224,7 +224,14 @@
 #
 www_root = os.path.dirname(os.path.abspath(__file__))
 _i = 0
-while _i < 50 and not os.path.isdir(os.path.join(www_root, 'www')):
+while _i < 50:
+    if(os.path.isdir(os.path.join(www_root, 'www'))):
+        break
+
+    if(os.path.isdir(os.path.join(www_root, 'site'))):
```
akotlar, line 231: since build depends on www we need 2 paths, or a much larger commit in which we rename site to www

danking: These files aren't actually referenced in the docs. The docs refer directly to `/navbar.css`. It seems like we should just remove lines 225 through 228 and reduce `html_extra_path` to just `misc/`. If we want the navbar to be present when viewing the docs locally, we should create a `hail/Makefile` target that sets up the environment correctly or a `site/Makefile` target that runs the whole site locally.

IMO, when I look at the docs locally it's just to check that I got my RST syntax right, I don't think the nabber is so important to that.

danking: I did try to get a local site working correctly but it was kind of annoying to parameterize nginx's TLS setting and I found `make -C site deploy NAMESPACE=dking` pretty fast anyway

danking: But I'm definitely here for a good local site preview.

akotlar:
> These files aren't actually referenced in the docs. The docs refer directly to /navbar.css

This is used in layout.html. Doc shares the site navbar.

hail/python/hail/docs/conf.py
akotlar: Can we use python -m http.server? I don't think nginx is needed for site.

danking:
I'm not sure I understand. `hail/python/hail/docs/_templates/layout.html` references `/navbar.css` which should be present on the deployed site. I might misunderstand `conf.py`, but, AFAICT, this makes unused copies of navbar.css and the PNG: https://hail.is/docs/0.2/navbar.css and https://hail.is/docs/0.2/hail-logo-cropped.png. The page at https://hail.is/docs/0.2/ loads `hail.is/navbar.css`.

akotlar:
> I'm not sure I understand. `hail/python/hail/docs/_templates/layout.html` references `/navbar.css` which should be present on the deployed site. I might misunderstand `conf.py`, but, AFAICT, this makes unused copies of navbar.css and the PNG: https://hail.is/docs/0.2/navbar.css and https://hail.is/docs/0.2/hail-logo-cropped.png. The page at https://hail.is/docs/0.2/ loads `hail.is/navbar.css`.

Sorry, ignore, I misread the comment. Agreed

akotlar:
There are also bootstrap files used in the project, but no longer in site, so those can be moved over.

PR #8957, opened 2020-06-15
[query] various minor fixes

cseed:
Summary:
 - remove debug statement (?)  in PartitionIteratorLongReader
 - LowerTableIR needs to truncate the partitioner when truncating the keys.  Otherwise, the partitioner will refer to rows that can be modified elsewhere.
 - Added a missing ToStream in TableExplode lowering

Unfortunately, the local backend tests are now segfaulting.  How's the debugging allocator fix coming @tpoterba @johnc1231?

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Compile.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Compile.scala
@@ -10,8 +10,6 @@ import is.hail.types.physical.PType
 import is.hail.types.virtual.Type
 import is.hail.utils._
 
-import scala.reflect.{ClassTag, classTag}
-
 case class CodeCacheKey(aggSigs: IndexedSeq[AggStateSig], args: Seq[(String, PType)], body: IR)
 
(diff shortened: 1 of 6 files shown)


johnc1231:
I am not sure what you're referencing, I'm not currently working on a debugging allocator. Didn't you start making one here? https://github.com/hail-is/hail/pull/8878

tpoterba:
Cotton and I discussed several weeks ago that we should build a debugging allocator for testing/debugging issues like the above. You were a natural candidate to work on this, given how you were digging pretty deeply into region stuff, and even started working on something similar.

Shall we set up a chat in the next day or two to talk about this? Sounds like it's now a high-prio need.

cseed:
bump

cseed:
Thanks!  To clarify your other question, while the local backend segfault with this, I think it is caused by existing memory  issues that show up in the Spark backend tests.  I conjecture when you fix those, the local backend failure will go away.  If not, we can flip to see who debugs it :-)

PR #8958, opened 2020-06-15
[lir] add code splitter, currently unused

cseed:
Due to bytecode verification rules, an allocated but uninitalized object cannot be stored into a field, so the NEW and INVOKESPECIAL constructor call bytecodes cannot be split across methods.  Therefore, I modified newInstance to fuse those operations together.

I broke out control simplification and made it a stronger.

Added method splitting.  Currently, method splitting splits out basic blocks into their own, straight-line methods and all the control flow remains in the original method.  All locals are spilled to fields which is terrible, but what we're doing now.  I expect two changes in the future: recover the structured control flow (there are standard algorithms for this) so we can split out control flow, and use the dataflow analysis from InitializeLocals to only spill locals split across method boundaries.

I will make a stacked PR on this that removes method wrapping from Emit and enables lir method splitting.

### DIFF
--- a/hail/src/main/scala/is/hail/lir/SimplifyControl.scala
+++ b/hail/src/main/scala/is/hail/lir/SimplifyControl.scala
@@ -137,10 +137,12 @@ class SimplifyControl(m: Method) {
     }
 
     val rootFinalTarget = mutable.Map[Int, Block]()
-    blocks.zipWithIndex.foreach { case (b, i) =>
+    blocks.indices.foreach { i =>
       val r = u.find(i)
-      val t = finalTarget(blocks(r))
-      rootFinalTarget(r) = t
+      if (r == i) {
+        val t = finalTarget(blocks(r))
+        rootFinalTarget(r) = t
+      }
     }
 
     for (b <- blocks) {
--- a/hail/src/main/scala/is/hail/lir/SimplifyControl.scala
+++ b/hail/src/main/scala/is/hail/lir/SimplifyControl.scala
@@ -1,0 +1,184 @@
+package is.hail.lir
+
+import is.hail.utils.UnionFind
+
+import scala.collection.mutable
+
+object SimplifyControl {
+  def apply(m: Method): Unit = {
+    new SimplifyControl(m).simplify()
+  }
+}
+
+class SimplifyControl(m: Method) {
+  private val uses = mutable.Map[Block, mutable.Set[(ControlX, Int)]]()
+
+  private val q = mutable.Set[Block]()
+
+  def removeUse(M: Block, c: ControlX, i: Int): Unit = {
+    val r = uses(M).remove((c, i))
+    assert(r)
+  }
+
+  def addUse(M: Block, c: ControlX, i: Int): Unit = {
+    val r = uses(M).add((c, i))
+    assert(r)
+  }
+
+  def finalTarget(b0: Block): Block = {
+    var b = b0
+    while (b.first != null &&
+      b.first.isInstanceOf[GotoX])
+      b = b.first.asInstanceOf[GotoX].L
+    b
+  }
+
+  def simplifyBlock(L: Block): Unit = {
+    val last = L.last.asInstanceOf[ControlX]
+
+    if (uses(L).isEmpty && (L ne m.entry)) {
+      var i = 0
+      while (i < last.targetArity()) {
+        val M = last.target(i)
+        removeUse(M, last, i)
+        last.setTarget(i, null)
+
+        q += M
+        if (uses(M).size == 1) {
+          val (u, _) = uses(M).head
+          q += u.parent
+        }
+
+        i += 1
+      }
+
+      // just popped off q
+      uses -= L
+
+      return
+    }
+
+    var i = 0
+    while (i < last.targetArity()) {
+      val M = last.target(i)
+      val newM = finalTarget(M)
+      if (M ne newM) {
+        removeUse(M, last, i)
+        last.setTarget(i, newM)
+        addUse(newM, last, i)
+
+        q += M
+        if (uses(M).size == 1) {
+          val (u, _) = uses(M).head
+          q += u.parent
+        }
+        q += L
+      }
+      i += 1
+    }
+
+    last match {
+      case x: IfX =>
+        val M = x.Ltrue
+        if (M eq x.Lfalse) {
+          x.remove()
+          removeUse(x.Ltrue, x, 0)
+          x.Ltrue = null
+          removeUse(x.Lfalse, x, 1)
+          x.Lfalse = null
+
+          val g = goto(M)
+          addUse(M, g, 0)
+
+          L.append(g)
+
+          // if there is one parent, it is L
+          q += L
+        }
+
+      case x: GotoX =>
+        val M = x.L
+        if ((M ne L) && uses(M).size == 1 && (m.entry ne M)) {
+          x.remove()
+          removeUse(x.L, x, 0)
+          x.L = null
+
+          while (M.first != null) {
+            val z = M.first
+            z.remove()
+            L.append(z)
+          }
+
+          q -= M
+          uses -= M
+
+          q += L
+        }
+
+      case _ =>
+    }
+  }
+
+  def unify(): Unit = {
+    val (blocks, blockIdx) = m.findAndIndexBlocks()
+
+    val u = new UnionFind(blocks.length)
+    blocks.indices.foreach { i =>
+      u.makeSet(i)
+    }
+
+    for (b <- blocks) {
+      if (b.first != null &&
+        b.first.isInstanceOf[GotoX]) {
+        u.sameSet(
+          blockIdx(b),
+          blockIdx(b.first.asInstanceOf[GotoX].L))
+      }
+    }
+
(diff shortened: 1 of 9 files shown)


hail/src/main/scala/is/hail/lir/SplitMethod.scala
```diff
@@ -0,0 +1,204 @@
...
+      L.append(x)
+      x = n
+    }
+
+    val newM = c.newMethod(genName("m", "wrapped"), FastIndexedSeq.empty[TypeInfo[_]], returnTI)
+    L.method = newM
+    newM.setEntry(L)
+
+    def invokeNewM(): ValueX =
+      methodInsn(INVOKEVIRTUAL, newM, Array(load(m.getParam(0))))
```
patrick-schultz, line 144: I'm not clear on the semantics of `ValueX`. From reading `SplitMethod`, it seems
* a `ValueX` may have arbitrary side effects,
* `X.children` must execute in exactly that order.

Is that right?

hail/src/main/scala/is/hail/lir/SplitMethod.scala
```diff
@@ -0,0 +1,204 @@
...
+
+      size += x.approxByteCodeSize()
+      val n = x.next
+      x.remove()
+      L.append(x)
+      x = n
+    }
+
+    val newM = c.newMethod(genName("m", "wrapped"), FastIndexedSeq.empty[TypeInfo[_]], returnTI)
+    L.method = newM
```
patrick-schultz, line 140: This (I think) ends up wrapping every block, no matter how small. It looks like we could also choose to move `L` back into `b` if `size` is below some threshold. Do you think it matters?

hail/src/main/scala/is/hail/lir/X.scala
```diff
@@ -466,93 +441,135 @@ abstract class StmtX extends X {
...
-class GotoX extends ControlX {
-  var _L: Block = _
 
-  def L: Block = _L
+  def remove(): Unit = {
+    var i = 0
+    while (parent.children(i) ne this)
+      i += 1
+    parent.setChild(i, null)
+    assert(parent == null)
```
patrick-schultz, line 501: This can't be true, can it?

hail/src/main/scala/is/hail/lir/SimplifyControl.scala
```diff
@@ -0,0 +1,184 @@
...
+          blockIdx(b),
+          blockIdx(b.first.asInstanceOf[GotoX].L))
+      }
+    }
+
+    val rootFinalTarget = mutable.Map[Int, Block]()
+    blocks.zipWithIndex.foreach { case (b, i) =>
+      val r = u.find(i)
+      val t = finalTarget(blocks(r))
+      rootFinalTarget(r) = t
```
patrick-schultz, line 143: Am I right this can be guarded by `if (r == i)`, to avoid traversing every chain of gotos once per contained block (or rather, traversing some representative chain in each connected component, once per contained block)?

Otherwise, what's the point of the union-find? You could just
```
    for (b <- blocks) {
      val last = b.last.asInstanceOf[ControlX]
      for (i <- 0 until last.targetArity()) {
        last.setTarget(i, finalTarget(last.target(i)))
      }
    }
```

cseed: Yep, added.  `r == i` is basically just `u.isRoot(i)`.

Your alternative is fine, except it is quadratic.  Even without the guard, this implementation still has quasi-linear complexity.

hail/src/main/scala/is/hail/lir/SplitMethod.scala
cseed: I just wrote the simplest thing to get it to work.  I have some more complicated plans that I think will subsume this.  Basically, I want to peel off structured chunks of code below the target size, update our estimate of the current function, and stop as soon as the current function is below the size.  But we should add this too.  I'll do it in the sequel.

I have splitting working in the stacked PR on this branch.  I ran the performance tests, and it definitely got much worse for large functions/deep loops:

```
                                         sample_qc    2426.5%     29.964    727.083
                             variant_and_sample_qc    1881.2%     39.905    750.713
                                 genetics_pipeline    1425.9%    108.001   1539.998
                         linear_regression_rows_nd     515.4%    212.164   1093.535
...
Geometric mean: 117.9%
Median:  102.8%
```

So we're going to need a slightly better split strategy before that can go in.  (Also, one test that creates 1,200 SeqOps is also failing with method too large.  The skeleton of the control flow is still too large.)

hail/src/main/scala/is/hail/lir/SplitMethod.scala
cseed: 1. Yes, `ValueX` can have side effects.  There is `MethodX` which represents a function call that returns a value.

2. I don't think we should rely on the order of execution of the children.  Let's call the order implementation defined (although in practice they're probably always executed in order.)

hail/src/main/scala/is/hail/lir/SimplifyControl.scala
cseed: Oops, sorry, you're 100% right, the guard is necessary!  Good catch.

hail/src/main/scala/is/hail/lir/SplitMethod.scala
patrick-schultz: What made me ask is this method:
```
      def splitLargeStatement(c: StmtX): Unit = {
        def visit(x: ValueX): Int = {
          // FIXME this doesn't handle many moderate-sized children
          val size = 1 + x.children.map(visit).sum

          if (size > SplitMethod.TargetMethodSize / 2) {
            val l = m.newLocal("spill_large_expr", x.ti)
            x.replace(load(l))
            c.insertBefore(store(l, x))
            1
          } else
            size
        }

        c.children.foreach(visit)
      }
```
For this to preserve the order of any side effects, doesn't it assume that the children of `c` are evaluated in the order of the `children` array? (And also that any side effects performed by `c` itself happen after the children are evaluated.)

hail/src/main/scala/is/hail/lir/X.scala
patrick-schultz: I'm still confused by this. We just called a method on `parent`, how can `parent` be null?

cseed: `this` is the `i`th child.  `setParent` removes the existing child, which means setting `parent = null`, because the node is no longer a child of the parent.

patrick-schultz: Sorry, I see it now.

cseed:
Yes, all children are evaluated left-to-right.

cseed:
Actually, a slightly longer answer:

The parent relationship is essentially a representation of the JVM stack.  Each X corresponds to 1 JVM bytecode (except NewInstanceX which is fused).  During emit, children are pushed on the stack left-to-right by necessity.

Slightly more generally, there are two distinct questions here: what the implementation does, and what it guarantees.  For what it guarantees, there are two options: fixed (like Java, which evaluates things left-to-right), and undefined (like C).

In general, I think it is safest not to rely on the order of evaluation.  I'm not aware of where we do. 
 Our backend is simple enough and the mapping to the JVM concrete enough that I don't see any reason why we'd have reason to deviate from left-to-right.  So I guess I fall somewhere in between (don't rely on it, but violate it).

patrick-schultz:
> In general, I think it is safest not to rely on the order of evaluation. I'm not aware of where we do.

Unless I'm missing something, this code splitter does rely on the evaluation order, which is why I asked about it. If a statement `x` has large children, it moves them to execute before `x`, evaluated in the `children` array order (more precisely, in the post-order traversal order), storing their results in locals.

I think it's possible to avoid relying on order of evaluation, if instead of only splitting out sequences of statements, we can also split out a `ValueX` directly, replacing it by a call to new method which returns the result.

I don't have strong feelings about whether the order should be guaranteed, just want to be clear which it is.

cseed:
> Unless I'm missing something, this code splitter does rely on the evaluation order

I don't understand.  I think I must be missing something.  Maybe there is a bug I don't see.  I'm particularly confused by your use of "rely".

I think splitter, by lifting things out into separate statements (where Block has a clear order of evaluation) implements an order of evaluation.  Now, you could ask if the order implemented by splitting is the same as the order of evaluation of the children of the parent.  What I'm saying is the the order of evaluation of the children should be undefined, so any choice of splitting out large children is valid (but all orderings being valid, we should prefer left-to-right).

Not, I'm just talking about the order of the evaluation of the children.  Children are always evaluated before parents (as necessitated by the data flow), and blocks are executed linearly in order.

patrick-schultz:
> What I'm saying is the the order of evaluation of the children should be undefined, so any choice of splitting out large children is valid (but all orderings being valid, we should prefer left-to-right).

Okay, I understand now, thanks for explaining. I was reading "undefined" as "unspecified", but still thinking it would be wrong for a code transformation to change the order. So the rule is, children `ValueX`s may produce side effects, but they may occur in any order; if the order is semantically relevant, lift them out to assignment statements.

cseed:
Yep, exactly!

PR #8960, opened 2020-06-15
[batch] add gcsfuse support

jigold:
This PR only adds support in the worker and exposes the functionality in the batch client. Another PR will be added to make these changes user facing in the user version of batch.

I added the WIP tag because I want to make sure the test passes once before I comment it out. That's because when this PR deploys, the workers will still be the old workers.


jigold:
Also, FYI - @cseed I had to change the permissions of the worker container to privileged and SYS_ADMIN. This doesn't apply to the user containers. But still we should double check the security implications of this change.

danking:
And we plan to enable the test long-term in the follow up PR, right?

jigold:
I added a front-end batch test. I think we should delete the test I added in this PR in favor of the one in #8961 once we know it's working. This PR as it stands passed the tests.

PR #8961, opened 2020-06-15
[batch] expose gcsfuse to user

jigold:
Stacked on #8960

CHANGELOG: Added support for gcsfuse when using the ServiceBackend.

### DIFF
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -71,6 +71,7 @@ class Job:
         self._image = None
         self._always_run = False
         self._timeout = None
+        self._gcsfuse = []
         self._command = []
 
         self._resources = {}  # dict of name to resource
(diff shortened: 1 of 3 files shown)


hail/python/hailtop/batch/job.py
```diff
@@ -521,6 +522,43 @@ def timeout(self, timeout):
...
+        bucket: :obj:`str`
+            Name of the google storage bucket to mount.
+        mount_point: :obj:`str`
+            The path for which the bucket should be mounted to in the Docker
+            container.
+
+        Returns
+        -------
+        :class:`.Job`
+            Same job object set with a timeout.
```
danking, line 553: I think this needs to be updated

hail/python/hailtop/batch/job.py
```diff
@@ -521,6 +522,43 @@ def timeout(self, timeout):
...
+        >>> j = b.new_job()
+        >>> (j.gcsfuse('my-bucket', '/my-bucket')
+        ...   .command(f'cat /my-bucket/my-file'))
+
+        Parameters
+        ----------
+        bucket: :obj:`str`
+            Name of the google storage bucket to mount.
+        mount_point: :obj:`str`
+            The path for which the bucket should be mounted to in the Docker
```
danking, line 547: at which the bucket

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -447,6 +448,19 @@ def test_verbose(self):
...
+    def test_gcsfuse(self):
+        b = self.batch()
+
+        head = b.new_job()
+        head.command(f'echo head > /{self.bucket_name}{self.gcs_output_path}/gcsfuse_test')
+        head.gcsfuse(self.bucket_name, f'/{self.bucket_name}')
+
+        tail = b.new_job()
+        tail.command(f'cat /{self.bucket_name}{self.gcs_output_path}/gcsfuse_test')
+        tail.gcsfuse(self.bucket_name, f'/{self.bucket_name}')
```
danking, line 460: I think this will race with head, right? Maybe use two batches?

jigold:
Adding the WIP label. Will untag it once all instances are version 5 (standing worker is version 4).

jigold:
I'm going to temporarily close this while I debug with dev deploy.

PR #8963, opened 2020-06-15
[query/lir] use lir method splitting instead of method wrapping

cseed:
Summary of changes:
 - rip out method wrapping logic in ir.Emit
 - add lir.{Blocks, Locals} for enumerating and indexing blocks and locals
 - add SplitLargeBlocks to break up large blocks (the minimum unit of splitting in SplitMethod is the block)
 - add PST to compute our non-standard variant of the program structure tree (see the comment in PST)
 - add SplitMethod to split large methods based on the PST


cseed:
This is still 2-5% slower than the current main line branch.  I have a few more things to try.  Benchmarks where the change is >20%:

```
$ hail-bench compare ./0.2.47-3f8cff262dcf.json ./0.2.47-b2acae4f478f.json 
                 matrix_table_entries_table_no_key     580.3%     56.955    330.503
                         table_aggregate_array_sum     528.0%      9.103     48.066
           table_big_aggregate_compile_and_execute     185.0%     13.325     24.648
                           per_row_stats_star_star     154.0%      8.474     13.046
                            table_aggregate_linreg     148.6%     48.593     72.212
              matrix_table_filter_entries_unfilter     140.8%      8.581     12.085
                     matrix_table_array_arithmetic     135.8%     10.208     13.862
                   matrix_table_many_aggs_col_wise     133.7%     34.847     46.576
...
                               full_combiner_chr22      33.4%   1644.907    548.608
----------------------
Geometric mean: 104.8%
Median:  101.9%
```

FYI @tpoterba @chrisvittal The combiner improvement persists across multiple runs and is real.  That's quite nice!

cseed:
Phew!  I finally got to parity.  I'm impressed how well the old method wrapping logic worked.  Benchmarks with >20% change:

```
$ hail-bench compare ./0.2.47-3f8cff262dcf-1.json 0.2.47-63dccdda2a44.json 
Failed benchmarks in run 1:
    pc_relate_big
    large_range_matrix_table_sum
Failed benchmarks in run 2:
    pc_relate_big
    large_range_matrix_table_sum
           table_big_aggregate_compile_and_execute     334.8%     13.325     44.608
                     matrix_table_array_arithmetic     133.9%     10.208     13.665
                   matrix_table_many_aggs_col_wise     131.4%     34.847     45.781
                           table_aggregate_counter     126.4%     13.136     16.606
                           per_row_stats_star_star     124.5%      8.474     10.553
              matrix_table_filter_entries_unfilter     123.7%      8.581     10.616
...
                            shuffle_key_rows_by_mt      82.8%     35.932     29.742
                               full_combiner_chr22      23.6%   1644.907    388.414
----------------------
Geometric mean: 100.0%
Median:  99.2%
```

tpoterba:
oh, man, this is super exciting. 3x on the combiner? yes please!

We can probably make incremental performance improvements to the LIR method splitting code to bring the compile and execute back down, and that one I consider a little less critical anyway.

cseed:
FYI @patrick-schultz Tim came up in scorecard, but I know you had some interest in this and thought you might like to look at my solution.

hail/src/main/scala/is/hail/lir/InitializeLocals.scala
```diff
@@ -13,7 +13,7 @@ object InitializeLocals {
     var i = entryUsedIn.nextSetBit(0)
     while (i >= 0) {
       val l = locals(i)
-      if (!l.isInstanceOf[Parameter]) {
+      if (!l.isInstanceOf[Parameter] && l.name != "spills") {
```
tpoterba, line 16: 🤔

hail/src/main/scala/is/hail/lir/SplitMethod.scala
```diff
@@ -1,97 +1,212 @@
...
+      blockSize(i) = size
+      methodSize += size
+      i += 1
+    }
+  }
 
-  private val paramFields = m.parameterTypeInfo.zipWithIndex.map { case (ti, i) =>
-    c.newField(genName("f", s"arg$i"), ti)
+  private val spillsClass = new Classx(genName("C", s"${ m.name }Spills"), "java/lang/Object")
+  private val spillsCtor = {
```
tpoterba, line 62: this is for empirical performance reasons, right?

hail/src/main/scala/is/hail/lir/SplitMethod.scala
```diff
@@ -101,104 +216,347 @@ class SplitMethod(c: Classx[_], m: Method) {
...
+    (start to end).foreach { i =>
+      blockPartitions.union(start, i)
+    }
+    updatedBlocks(blockPartitions.find(start)) = newL
+
+    val returnTI = Lend.last match {
       case _: GotoX => UnitInfo
-      case _: IfX => BooleanInfo
+      case x: IfX =>
+        if (!regionBlocks(x.Ltrue) && !regionBlocks(x.Lfalse))
```
tpoterba, line 314: could you explain this bit?

tpoterba: ah, I see.

hail/src/main/scala/is/hail/lir/SplitMethod.scala
```diff
@@ -101,104 +216,347 @@ class SplitMethod(c: Classx[_], m: Method) {
...
+            newLtrue.append(returnx(ldcInsn(1, BooleanInfo)))
+            x.setLtrue(newLtrue)
+
+            val newLfalse = new Block()
+            newLfalse.method = splitM
+            newLfalse.append(returnx(ldcInsn(0, BooleanInfo)))
+            x.setLfalse(newLfalse)
+          }
+        }
+      case x: SwitchX => IntInfo
```
tpoterba, line 416: stray dead code?

hail/src/main/scala/is/hail/lir/SplitMethod.scala
```diff
@@ -101,104 +216,347 @@ class SplitMethod(c: Classx[_], m: Method) {
...
-    L.method = newM
-    newM.setEntry(L)
+      val newsubr = mutable.ArrayBuffer[SizedRange]()
 
-    def invokeNewM(): ValueX =
-      methodInsn(INVOKEVIRTUAL, newM, Array(load(m.getParam(0))))
+      i = sortedsubr.length - 1
+      while (i >= 0) {
+        val ri = sortedsubr(i)
+        if (ri.size > 20 &&
```
tpoterba, line 491: should this be a parameter to include in benchmarking?

cseed: I don't think so.

This is just to make sure that we don't split something that had just been split, which would go into an infinite loop.  It is paired with this:

```
          splitSlice(ri.start, ri.end)
          val s = blockSize(blockPartitions.find(ri.start))
          assert(s < 20)
```

to make sure split things are sufficiently small that we won't be tempted to resplit them.

I don't think this is a performance issue, but there is potentially a pathological case where splitting could fail to reduce the size of the method: you have a massive region, but is made up of size 15 chlid regions which are all independent with no substructure.  Then nothing would get split, and we might blow the method size limit.  It's actually hard to imagine what such a function might look like, a bunch of switches that all jump to each other.  I'm not quite sure how to split such a function, actually.

hail/src/main/scala/is/hail/lir/SplitMethod.scala
```diff
@@ -1,97 +1,212 @@
...
-object SplitMethod {
-  val TargetMethodSize: Int = 2000
+import scala.collection.mutable
+
+class SplitUnreachable() extends Exception()
 
-  def apply(c: Classx[_], m: Method): Unit = {
-    new SplitMethod(c, m).split()
+object SplitMethod {
+  val TargetMethodSize: Int = 500
```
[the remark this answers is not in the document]
cseed: I ran some experiments, but also while I was changing the code, so further tuning might be valuable.  Spilling can increase method size, and I've seem methods nearly 2x the target size.  The JVM max is 8K.  I started with 2K, but found 500 works better.  I think I tried 250, but it was not better (I don't remember how much worse).

hail/src/main/scala/is/hail/lir/PST.scala
```diff
@@ -0,0 +1,581 @@
...
+// this without linearization, but I found this conceptually simple to
+// implement and it will be more than fast enough.  Most steps are linear,
+// except:
+//  - it is O(N log N) in the back edges (a minority of edges) because of
+//    the use of TreeSet when computing minTargetLE and maxSourceGE,
+// - and O(N^2) to identify seqential regions when checking regionStarts,
+//   although that is only non-trivial when the CFG is irreducible, which
+//   is relatively rare.
+
+class Region(
```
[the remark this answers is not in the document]
cseed: I renamed it PSTRegion.

hail/src/main/scala/is/hail/lir/InitializeLocals.scala
cseed: Oops, that wasn't supposed to make it in.  Deleted!

PR #8964, opened 2020-06-15
[site] fix line thickness on low dpi displays

akotlar:
Pixel ratio of 1 makes the lines too thick. Site looks worse for it on low dpi displays:

before:
Screenshot 2020-06-14 16 02 32

after:
Screenshot 2020-06-14 16 02 40


site/www/vendors/vanta/viz.js
```diff
@@ -346,7 +350,7 @@ class Viz {
     }
 
     this.renderer.setSize(this.el.offsetWidth, this.el.offsetHeight)
-    this.renderer.setPixelRatio(window.devicePixelRatio)
+    this.renderer.setPixelRatio(getPixelRatio())
```
danking, line 353: Isn't the central issue that the lines are too thick with pixel ratio is 1?

On a screen with pixel ratio of 1, the lines and, e.g., the font use the same definition of pixel: CSS Pixel, which is 1:1 with device pixel.

On a screen with a pixel ratio of 2, the lines and the font, I presume, use different definitions of pixel. The font still uses CSS pixel. However, thanks to `setPixelRatio`, I assume, the `material.linewidth` is now interpreted in terms of device pixels which are 2:1::device:css. Ergo, the lines look thinner compared to the font (or anything else measured in device pixels) on a higher DPI screen than on a lower DPI screen.

It seems to me that we ought to scale `material.linewidth` with the pixel ratio?

akotlar: Does not work

akotlar: Also line width is inconsistently applied, differs between platforms. Apparently windows fixes to 1 regardless of setting, not that this affects us, since we want < 1

akotlar:
Additionally, I think this is actually the correct/intended use of setPixelRatio. Note that if we set linewidth = .5 successfully, which we cannot, devices with devicePixelRatio = 1 would still look twice as thick as those with ratio 2. 

Finally, I use min(2, device) rather than 2, because in tests of larger pixel ratios (tested 3: in Firefox you can specify this), a fixed 2 value produced lines that appeared slightly too thin.

danking:
If line width < 1 is invalid, why do we set `material.linewidth` to 0.25 on line 328/332 of viz.js?

akotlar:
Needs to be fixed.

akotlar:
The documentation is incorrect; linewidth .25 is needed for Safari:

On a pixelratio == 2 display:
linewidth 1:
Screenshot 2020-06-16 06 40 30

linewidth .25:
Screenshot 2020-06-16 06 40 18

Note this is still orthogonal to the issue of devicePixelRatio

akotlar:
There may also be a way of handling the intended effect using alpha blending, but this works and is a simple fix.

akotlar:
Better experiment below, you can see in the right hand side the computed values. Note: I don't see any visual difference between 1 and < 1 in safari, but for the sake of not complicating this further, I want to keep .25 unless there is evidence this causes issues. Otherwise I think this issue is solved.

Screenshot 2020-06-16 06 58 11
Screenshot 2020-06-16 06 58 28
Screenshot 2020-06-16 06 58 45
Screenshot 2020-06-16 06 59 03

akotlar:
bump

akotlar:
bump

danking:
This change achieves the effect you desire. What I additionally need from this PR is a mental model of why this change works. This change specifies a pixel ratio of two when the device's physical:css pixel ratio is one, but uses the device's pixel ratio otherwise. This breaks my conceptual model. It's important for us all to share compatible models of what the code does so that we all are able to manipulate the code in the future. I think there's a few moving pieces and if we can get a handle on them all, we'll all agree on what the right fix is.

AFAICT, three js is built on WebGL. The MDN recommends not using a WebGL `lineWidth` other than one because of inconsistent (or lack of) support for line widths other than one. You observe that `pixelRatio` affects the visual behavior, at least in Safari, when the lineWidth is set to 0.25. In particular, when the pixel ratio is set to `2` and the lineWidth is set to `0.25`, the lines appear thinner.

If the pixel ratio is higher than the device pixel ratio, something must be interpolating to device pixels. It seems to me that relying on this interpolation behavior will lead to code that is more difficult to understand and manipulate.

If the intention is to make the lines less striking, can we apply an alpha filter instead? Is there another simple & consistent-across-platforms solution?

akotlar:
> This change achieves the effect you desire. What I additionally need from this PR is a mental model of why this change works.

I think I communicated this above: as far as I understand it, the effective resolution of the graphic is dependent on both the viewport resolution and the pixel ratio. The observed resolution is too low with pixelRatio 1, gives thick lines.

> If the intention is to make the lines less striking, can we apply an alpha filter instead? Is there another simple & consistent-across-platforms solution?

Not that I know of. Some people use cone geometry to get around line thickness issues.

> It seems to me that relying on this interpolation behavior will lead to code that is more difficult to understand and manipulate.

I don't understand. Is there another fix that seems easier?
 
> If the intention is to make the lines less striking, can we apply an alpha filter instead? Is there another simple & consistent-across-platforms solution?

Potentially, but this will obviously be a significantly more complicated fix. The proposed fix is effectively 1 line, and takes care of the observed issue in a wide range of browsers.

akotlar:
If you're actually seeing some inconsistent results in some browser, I agree the solution isn't sufficient. Else, why not get the easy solution in, and do more work when it's needed. I see a potential problem statement, no clear reason why the present solution is problematic, and a desire to move to a different solution. 

I've looked into the relationship between fractional line widths and alpha a few days ago, when I wrote the comment suggesting that alpha blending could be an alternative solution. Antialiased fractional line width (which is also what you would get if the effective resolution is larger than the viewport resolution and then objects scaled), will perform the equivalent of alpha blending.

akotlar:
I'd also like to mention that linewidth in general seems to have cross-browser issue (https://github.com/mrdoob/three.js/issues/10357). Most of the issues detailed are for line width > 1, which is not an issue for us. The fact that safari requires this  (< 1) to normalize its behavior with other browsers is unfortunate, but presumably when fixed, it would just truncate to 1 anyhow (this is what every other browser appears to do).

So we have 2 ideas on the table: 1) use the fix proposed, 2) use an entirely different solution, that. does not rely on linewidth. #2 adds a lot of work, and as far as I can tell, absolutely no benefit, except to make us feel more comfortable that, as browser implementations change, the solution will be more stable. That may be true, and at the same time every tested browser (edge, safari, chrome, brave, Firefox) has the correct behavior with this fix.

If you want #2, I am happy to close this request, and have someone else PR an entirely new background geometry. I don't have time for that unfortunately, and it seems too much work without a visible problem to address.

danking:
We seem to be miscommunicating. My issue is that this change uses a knob (pixel ratio) to control something seemingly unrelated to the knob (opacity).

This isn't a browser thing, it's a monitor thing. I have two screens one with device pixel ratio (DPR) 1 and one with DPR 2. Regardless of browser, on the DPR 1 monitor, the lines appear thicker than on the DPR 2 monitor. They must appear thicker! The pixels are bigger and less dense.

My suggestion is this: https://github.com/hail-is/hail/compare/master...danking:viz

That change sets line width to 1 and uses opacity to reduce the brightness of the lines. My proposal looks, to me, exactly the same as this PR (which I've dev deployed into my namespace: https://internal.hail.is/dking/site/). I've compared both on my DPR 2 monitor and my DPR 1 monitor. 

The reason I prefer the change linked above is that it is self-describing: we want the thinest possible lines that the monitor supports and we want them to be a bit faint.

akotlar:
> My suggestion is this: master...danking:viz

This doesn't work, afaict, even at pixel ratio 2.

brave:
Screenshot 2020-06-17 21 03 08


safari:
Screenshot 2020-06-17 21 03 02

akotlar:
> This isn't a browser thing,

It is a browser thing as well as a monitor thing, as far as I can see. I linked to a chromium bug that specifically identified the issue as a browser thing. There is also some difference that I cannot find a bug report for Safari that makes line width unequal between Safari and Chromium browsers, unless I set Safari's line width to <1.

danking:
I believe you're referencing [Chrome Bug 675308](https://bugs.chromium.org/p/chromium/issues/detail?id=675308) which suggests that lines thicker than 1 are broken in some versions of Chrome. My change uses line width 1.

The screenshots you shared seem to suggest that Brave is making the lines somewhat fainter than Safari. I can't explain that. The visuals look the same to me across Chrome and Safari. I don't have Firefox installed to check a non-WebKit renderer.

At this pointI feel that I have said my piece. It feels brittle to fiddle with device pixel ratio and line width to try and simulate line widths smaller than a pixel. If you think the only acceptable solution is the proposed changes in this PR, then let's do it.

akotlar:
> I believe you're referencing Chrome Bug 675308 which suggests that lines thicker than 1 are broken in some versions of Chrome. My change uses line width 1.

My point was that there are browser inconsistencies.

> The screenshots you shared seem to suggest that Brave is making the lines somewhat fainter than Safari. I can't explain that. 

I also can't explain it, and haven't found a Safari bug that would suggest why this is the case, and yet it is clear from that screenshot that there is a difference.

>  It feels brittle to fiddle with device pixel ratio and line width to try and simulate line widths smaller than a pixel

My current mental model is that we're just hitting a resolution issue. Internally threes I believe sets the webgl buffer to pixelRatio * width/height. With resolution 1, resolution seems too low, and this causes fuzzy lines. I don't know why it doesn't work better. To test this hypothesis, I've tried setting pixelRatio to 1 manually on a hidpi display, and it gave a similar fuzzy/thick result, which of course doesn't make any sense if linewidth actually did what it seems it should, and so I agree that not relying on linewidth would be nicer, but would also be additional work.

PR #8966, opened 2020-06-15
[query] split out metadata writers

catoverdrive:
Split out the TableNativeWriter metadata writers into three separate writers. I use this to make the MatrixWriter lowering cleaner.

(see #8967)

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
@@ -10,12 +10,11 @@ import is.hail.expr.ir.lowering.{LowererUnsupportedOperation, TableStage}
 import is.hail.io.fs.FS
 import is.hail.io.index.StagedIndexWriter
 import is.hail.types.virtual._
-
 import is.hail.io.{AbstractTypedCodecSpec, BufferSpec, OutputBuffer, TypedCodecSpec}
 import is.hail.rvd.{AbstractRVDSpec, IndexSpec, RVDPartitioner, RVDSpecMaker}
 import is.hail.types.{RTable, TableType}
 import is.hail.types.encoded.EType
-import is.hail.types.physical.{PArray, PCanonicalString, PCanonicalStruct, PCode, PInt64, PStream, PString, PStruct, PType}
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
```diff
@@ -217,56 +217,89 @@ case class PartitionNativeWriter(spec: AbstractTypedCodecSpec, partPrefix: Strin
...
-      s"in ${ partitionCounts.length } ${ plural(partitionCounts.length, "partition") } " +
-      s"to $path")
+  def writeMetadata(
+    writeAnnotations: => IEmitCode,
+    cb: EmitCodeBuilder,
+    region: Value[Region]): Unit = {
+    cb += cb.emb.getFS.invoke[String, Unit]("mkDir", path)
+    writeAnnotations.consume(cb, {
+      cb._fatal("write annotations can't be missing!")
+    }, { case pc: PIndexableCode =>
```
patrick-schultz, line 280: You can also use the pattern
```
val pc = writeAnnotations.handle(cb, cb._fatal(...))
...
```
for this, which I think is clearer, and reduces indentation.

hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
```diff
@@ -217,56 +217,89 @@ case class PartitionNativeWriter(spec: AbstractTypedCodecSpec, partPrefix: Strin
...
+case class RVDSpecWriter(path: String, spec: RVDSpecMaker) extends MetadataWriter {
+  def annotationType: Type = TArray(TString)
+  def writeMetadata(
+    writeAnnotations: => IEmitCode,
+    cb: EmitCodeBuilder,
+    region: Value[Region]): Unit = {
+    cb += cb.emb.getFS.invoke[String, Unit]("mkDir", path)
+    writeAnnotations.consume(cb, {
+      cb._fatal("write annotations can't be missing!")
+    }, { case pc : PIndexableCode =>
```
patrick-schultz, line 229: Can use `handle` here too.

PR #8968, opened 2020-06-15
[batch] Add type annotations for hailtop version of batch

jigold:


danking:
It looks like pylint/flake8 don't like the import structure. I say we just import them! I don't think it affects performance in any meaningful way.

jigold:
They're circular imports I thought.

danking:
I'll have to think about the circular thing, but I learned that
```
    from __future__ import annotations
```
let's you use names directly rather than as strings.

jigold:
I'm running into this same problem. Going to try upgrading flake8 to see if that helps.
https://github.com/PyCQA/pyflakes/issues/290

danking:
great!

danking:
@jigold this should pass now. I also learned that Mypy checks each command line argument as an independent module. Instead of specifying individual files, we have to tell it to check `batch`. I also added `google_storage.py` in an ill-fated attempt to let batch use that. I think we can only type check modules that depend on one another by installing them first. We'll leave that for future work.

jigold:
There's some major test errors.

danking:
worker.py was missing for some reason. I restored it to the version from master.

danking:
OK! I think I finally squashed all the bugs

danking:
There were version incompatibilities between dev-requirements.txt and docker/requirements.txt. I aligned all the versions and pulled some purest things up into docker/requirements.txt.

PR #8969, opened 2020-06-15
[docs] prefix match on all words in sentence

akotlar:
1) Adds "/" focus on query bar

2) Adds prefixAll queries 
  * This has some drawbacks - https://www.algolia.com/doc/api-reference/api-parameters/queryType/ - but allows "filter allele frequency" to match "filtering allele frequency". We should keep an eye on drop in relevance.


tpoterba:
oh, sorry, didn't realize this wasn't stacked anymore

PR #8970, opened 2020-06-16
Don't use a 0 pointer in TableWriter

johnc1231:
Working on the safe memory allocator that checks if we are accessing invalid memory, it's catching this 0 pointer. 

There are already tests that hit `TableWriter`, and this should result in no functionality change.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
@@ -180,7 +180,11 @@ case class PartitionNativeWriter(spec: AbstractTypedCodecSpec, partPrefix: Strin
                   IEmitCode.present(cb, PCode(keyType, keyRVB.offset))
                 },
                   ob.invoke[Long]("indexOffset"),
-                  IEmitCode.present(cb, PCode(+PCanonicalStruct(), 0L)))
+                  {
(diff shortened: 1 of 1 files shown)


cseed:
So I don't think this is quite correct.

malloc(0) is implementation defined, it either returns a valid pointer which cannot be dereferenced, or 0: https://stackoverflow.com/a/2022402

memcpy with a 0 source is undefined: https://stackoverflow.com/questions/5243012/is-it-guaranteed-to-be-safe-to-perform-memcpy0-0-0

There is some debate in the comments to the first answer about whether any reasonable memcpy implementation would do anything if the length is 0.  However, see the second answer in the link above: gcc improves optimization by assuming the undefined behavior cannot happen.

So our options seem to be:

1. Leave this change, but change Memory.malloc to allocate at least one byte (guaranteeing it never returns 0).

2. Revert this change, and assume memcpy with 0 length is safe and weaken the memcpy check.

3. Revert this change, don't assume memcpy of 0 length is safe, and have Memory.memcpy check the length before calling into the Unsafe memcpy implementation.

I'm inclined to do 3 in this case.

johnc1231:
Ok, I did not realize`malloc(0)` could return 0 pointer. I'll fix it when I'm back on Thursday. Thanks for catching this.

PR #8971, opened 2020-06-16
[batch] unmount gcsfuse in worker when cleaning up job

jigold:


jigold:
fusermount needs root privileges which the container has. I tested this with dev deploy and made sure things were being unmounted properly.

jigold:
This isn't the right solution. The bucket should be mounted in a different place from the scratch directory we blow away.

PR #8973, opened 2020-06-16
[batch] Use weights for cpu and blkio in container config

jigold:


batch/batch/worker.py
```diff
@@ -220,6 +220,10 @@ async def __aexit__(self, exc_type, exc, tb):
         self.timing['duration'] = finish_time - start_time
 
 
+def worker_fraction_in_1024ths(cpu_in_mcpu):
+    return 1024 * cpu_in_mcpu // (CORES * 1000)
```
danking, line 224: I prefer parentheses around the (1024 * cpu_in_mcpu) for clarity.

PR #8983, opened 2020-06-17
release 0.2.46

akotlar:
Do we want to highlight anything else? I don't see batch-specific notes, and besides that the 0L pointer could be added to "bugfix", but I'm not certain if that's worth pointing out as it's not user-facing functionality, and based on cotton's comments the original was not strictly incorrect.

stacked on #8981


hail/python/hail/docs/change_log.md
```diff
@@ -22,6 +22,13 @@ Please note that **forward compatibility should not be expected, especially
 relating to file formats**: this means that it may not be possible to use
 an earlier version of Hail to read files written in a later version.
 
+## Version 0.2.46
+
+Release 2020-06-17
```
tpoterba, line 27: Released (fix the typo down below too)

hail/python/hail/docs/change_log.md
```diff
@@ -22,6 +22,13 @@ Please note that **forward compatibility should not be expected, especially
...
+
+### site
```
tpoterba, line 29: Capitalize?

hail/python/hail/docs/change_log.md
```diff
@@ -22,6 +22,13 @@ Please note that **forward compatibility should not be expected, especially
...
+- (hail#8955) Natural language documentation search
+
```
tpoterba, line 31: missing a `---`

PR #8987, opened 2020-06-18
[query] Fix partitioner memory error

cseed:
The new generic lines coerce code could produce a partitioner with unsafe values.  Those unsafe values ended up in the Compile cache, which become invalid when owning region was cleared.

This fixes the memory errors I was seeing when running with the local backend.  It is possible it will fix (some?) of the errors you were investigating, @johnc1231.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/IR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/IR.scala
@@ -1,6 +1,6 @@
 package is.hail.expr.ir
 
-import is.hail.annotations.{Annotation, Region}
+import is.hail.annotations.{Annotation, Region, UnsafeRow}
 import is.hail.asm4s.Value
 import is.hail.expr.ir.ArrayZipBehavior.ArrayZipBehavior
 import is.hail.expr.ir.EmitStream.SizedStream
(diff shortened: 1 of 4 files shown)


hail/src/main/scala/is/hail/expr/ir/IR.scala
```diff
@@ -88,6 +88,7 @@ object Literal {
 final case class Literal(_typ: Type, value: Annotation) extends IR {
   require(!CanEmit(_typ))
   require(value != null)
+  // require(SafeRow.isSafe(value))
```
johnc1231, line 91: I see the main fix here about creating a `SafeRow` instead of an `UnsafeRow`. I don't see why this and the `assert` in `RVDPartitioner` are now commented out. If you intend to leave them commented out, I think I'd want a line comment above explaining why they are currently commented out.

tpoterba: These are pretty expensive checks to be doing every time the node is constructed, I think that's why. But agree that a comment would be good.

johnc1231: Ah, I read the diff wrong, thought these were things we previously did that are now commented out. Now see that they are newly added lines altogether. Agree that it would be an expensive check. I think I would still add a comment like "For debugging purposes" or something above the lines.

PR #8989, opened 2020-06-18
[site] prevent background from encroaching all-hero-content on wide views

akotlar:
Points shouldn't show through central section, too noisy, triggers trichophobia. This fixes a regression (previously there was enough padding to make the space between the title and panel to block point/line visibility), and improves on that previous state (by making the entire central section uniformly opaque, gives more visual consistency and will mean future padding changes don't break opaqueness). 

After:
Screenshot 2020-06-18 11 45 37

### DIFF
--- a/site/www/style.css
+++ b/site/www/style.css
@@ -3,6 +3,7 @@ html {
   font-size: 100%;
   line-height: 1.15;
   -webkit-text-size-adjust: 100%;
+  background: #fff;
 }
 
 * {
@@ -15,6 +16,7 @@ body {
   line-height: 1.6;
   margin: auto;
   border-color: #ddd;
+  background: inherit;
 }
 
 a:hover {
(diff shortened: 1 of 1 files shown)


site/www/style.css
```diff
@@ -3,6 +3,7 @@ html {
   font-size: 100%;
   line-height: 1.15;
   -webkit-text-size-adjust: 100%;
+  background: #fff;
```
akotlar, line 6: by specifying background at top level and inheriting, we can specify background color once (this will be useful for future dark mode)

site/www/style.css
```diff
@@ -824,6 +827,14 @@ pre code {
   }
 }
 
+@media only screen and (min-width: 1521px) {
```
akotlar, line 830: only applied to wide view, where we have the graph panel. on narrow views hero-content has a background, and all-hero-content stretches the width of the screen, so we don't want to make it opaque.

PR #8991, opened 2020-06-18
[query] Lower TableAggregate (take 1)

tpoterba:
This pull request lowers the TableAggregate node. It also reimplements
the AggStateValue and CombOpValue nodes.

However, this is not the final design -- I think that there should also
be an InitFromSerializedValue node that is used to prevent us from
initializing twice (and inlining aggs.init in the lowering).

feel free to push back on merging this before that refactoring, but it works and I have a preference for making that change in a followup.

After that, I'll move on to lowering TableMapRows with scans, as well as designing the pass to lift out nodes as relational lets that we discussed.

Woohoo!


hail/src/main/scala/is/hail/expr/ir/IR.scala
```diff
@@ -427,7 +427,7 @@ final case class SeqOp(i: Int, args: IndexedSeq[IR], aggSig: PhysicalAggSig) ext
 final case class CombOp(i1: Int, i2: Int, aggSig: PhysicalAggSig) extends IR
 final case class ResultOp(startIdx: Int, aggSigs: IndexedSeq[PhysicalAggSig]) extends IR
 final case class CombOpValue(i: Int, value: IR, aggSig: PhysicalAggSig) extends IR
-final case class AggStateValue(i: Int, aggSig: AggStateSig) extends IR
+final case class AggStateValue(i: Int, aggSig: PhysicalAggSig) extends IR
```
catoverdrive, line 430: I don't think this should be a PhysicalAggSig---the aggOp should have no bearing on how you serialize the state.

tpoterba: yeah, had the same thought this morning. CombOpValue too, I think.

PR #8995, opened 2020-06-19
[docs] overview titles, 

akotlar:
Will remove the seemingly-duplicate "Table", "Expressions", "MatrixTable", "Types" search results, and improve similarity between "GWAS" vs "Genome-Wide Association Study" queries


akotlar:
bump

PR #8997, opened 2020-06-20
[build] Add makefile target that creates a Hail egg archive

tpoterba:


tpoterba:
The target creates an egg in `hail/hail/build/deploy/dist/`, alongside the wheel.

PR #9000, opened 2020-06-21
[query/lir] add Block.replace

cseed:
I need this for upcoming improvements to SplitMethod.

Summary:
 - add Block.replace
 - have Block track uses
 - makes SimplifyControl simpler
 - Method.findBlocks prunes dead uses


cseed:
bump

cseed:
Thanks!

PR #9006, opened 2020-06-23
Fix Memory Leak in ExportBgen

johnc1231:
CHANGELOG: Fix memory leak in ExportBgen

If you look a dozen or so lines under the `boundary` I added, you'll see a :

```
        it.foreach { ptr =>
          val (b, d) = bpw.emitVariant(ptr)
          out.write(b)
          dropped += d
        }
```

So that thing is stepping through the iterator without freeing. The `boundary` is to do a `clear` after each read from the `it`.


johnc1231:
Trying to fix: https://discuss.hail.is/t/limit-memory-usage/1468/3

PR #9005, opened 2020-06-23
[site] tutorial page

akotlar:
Adds the tutorial page (http://34.207.246.132/tutorial.html) and Learn More button bottom-right justified on home page

cc @mkveerapen includes some further content changes

minor updates to style (mostly width) found in the subsequent pr: #9007

PR #9007, opened 2020-06-23
[site] reference styling, more consistent styling elsewhere

akotlar:
Last todo to round out the first phase of web work.

1) Styles references page by adding header and making sure content is centered on wide views
2) Makes get help .about content consistent with tutorial page and home page (using .content)
3) Sets max width to 1521px on non-home pages

stacked on #9005

I will look at any comments on July 1


akotlar:
@mkveerapen assigned you since you reviewed tutorial page. this will be a small pr once that goes in (< 100 lines)

PR #9009, opened 2020-06-23
[query][bugfix] clear region in rvd.countPerPartition

catoverdrive:


julirsch:
Wow. This will make things so much cheaper!

johnc1231:
Yeah I can't believe we didn't notice this sooner, should help a lot.

PR #9010, opened 2020-06-23
[query] test we do not blow memory

danking:
1024 seems pretty generous, but I didn't rigorously search for a value.

I do know that when setting 400 million bytes, memset is way faster.


catoverdrive:
Looks like this does the thing, yay!

PR #9011, opened 2020-06-23
[query] fuse StreamMaps in StreamZip streams

catoverdrive:
I've not written a test for this because we use this pattern a lot, and we haven't been writing explicit test cases for all of our simplify rules, but let me know if you want one.


chrisvittal:
Do we have a benchmark or similar to show the improvement here?

johnc1231:
What's going on with this. Should this still be open?

chrisvittal:
I'm so sorry about this. It seems to have fallen off my plate. Thanks for bringing it up John.

Arcturus, what would you like to do here?

danking:
Trying to debug a CI issue and this PR's status has exceeded the number of CI updates due to it being rather old. I'll reopen once I've figured out CI's issue.

PR #9014, opened 2020-06-24
[site] order buttons, appropriate svg path

akotlar:


akotlar:
bump @tpoterba could you accept this since Kumar is on vacation?

akotlar:
thank you John

PR #9021, opened 2020-06-26
[batch] expose gcsfuse to user

jigold:
CHANGELOG: Added gcsfuse option for the Service Backend.

The only thing I wasn't sure about was whether the default for the read_only should be True or False.


jigold:
This should have all instances be version 7 in the default namespace! Adding the WIP tag until we verify all instances have been converted over.

PR #9023, opened 2020-06-29
[query] fix bug in LoweredTableReader

patrick-schultz:
`LoweredTableReader.makeCoercer` was using the same aggregator to find both the min and max keys in a partition, hence creating degenerate partitioners. This manifested in incorrect results in joins. This PR fixes it by adding a `SortOrder` option to the `TakeBy` aggregator, and using a descending `TakeBy` to compute the max key.

### DIFF
--- a/hail/src/test/scala/is/hail/expr/ir/Aggregators2Suite.scala
+++ b/hail/src/test/scala/is/hail/expr/ir/Aggregators2Suite.scala
@@ -315,7 +315,6 @@ class Aggregators2Suite extends HailSuite {
         perm <- permutations;
         so <- FastIndexedSeq(Ascending, Descending)
     ) {
-      println(s"take $n, reverse=$so")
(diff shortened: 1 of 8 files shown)


hail/src/test/scala/is/hail/expr/ir/Aggregators2Suite.scala
```diff
@@ -303,16 +305,18 @@ class Aggregators2Suite extends HailSuite {
...
 
     // test counts and data input orderings
     for (
       n <- FastIndexedSeq(0, 1, 4, 100);
-        perm <- permutations
+        perm <- permutations;
+        so <- FastIndexedSeq(Ascending, Descending)
     ) {
-      test(n, perm, t, identity[IR], identity[Row], TInt32, GetField(_, "b"))
+      println(s"take $n, reverse=$so")
```
tpoterba, line 318: println

PR #9024, opened 2020-06-29
[query] Refactor agg lowering to use new InitFromSerializedState node

tpoterba:
Stacked on #8991


tpoterba:
Ready for review.

hail/src/main/scala/is/hail/lir/X.scala
```diff
@@ -116,6 +116,7 @@ class Classx[C](val name: String, val superName: String) {
     }
      */
 
+    // println(Pretty(this))
```
tpoterba, line 119: I think I took this out by mistake in the last PR.

hail/src/main/scala/is/hail/expr/ir/agg/AggregatorState.scala
```diff
@@ -30,15 +30,15 @@ trait AggregatorState {
   def deserialize(codec: BufferSpec): (EmitCodeBuilder, Value[InputBuffer]) => Unit
 
   def deserializeFromBytes(cb: EmitCodeBuilder, t: PBinary, address: Code[Long]): Unit = {
-    val lazyBuffer = kb.getOrDefineLazyField[MemoryBufferWrapper](Code.newInstance[MemoryBufferWrapper](), (this, "buffer"))
+    val lazyBuffer = kb.getOrDefineLazyField[MemoryBufferWrapper](Code.newInstance[MemoryBufferWrapper](), (this, "bufferWrapper"))
```
tpoterba, line 33: This was a subtle one -- we have to name these separate things because otherwise they'd refer to the same field! The tuple `(this, "buffer")` was used in both.

catoverdrive: Is this a problem if we call this on multiple AggregatorState objects? Is that a pattern that will ever come up?

hail/src/main/scala/is/hail/expr/ir/IR.scala
```diff
@@ -430,6 +430,7 @@ final case class CombOp(i1: Int, i2: Int, aggSig: PhysicalAggSig) extends IR
 final case class ResultOp(startIdx: Int, aggSigs: IndexedSeq[PhysicalAggSig]) extends IR
 final case class CombOpValue(i: Int, value: IR, aggSig: PhysicalAggSig) extends IR
 final case class AggStateValue(i: Int, aggSig: AggStateSig) extends IR
+final case class InitFromSerializedValue(i: Int, value: IR, aggSig: PhysicalAggSig) extends IR
```
catoverdrive, line 433: I think this probably only needs the AggStateSig, since we don't need the op to serialize or deserialize.

hail/src/main/scala/is/hail/expr/ir/agg/AggregatorState.scala
tpoterba: No, since our compiled code is single-threaded, and these fields do not store persistent state, only a memoized object that could be expensive to construct many times. The problem with the clashing definition here is that the fields have different types.

catoverdrive: Ah, I see. The field doesn't actually get constructed from the data; it's a container object for the byte array.

tpoterba: yep, exactly.

PR #9025, opened 2020-06-29
[bug] Fix and test single row blockmatrix vector sum bug

annamiraotoole:
CHANGELOG: Fix bug where summing a BlockMatrix with a single row or column didn't work.


hail/python/test/hail/linalg/test_linalg.py
```diff
@@ -1076,3 +1076,29 @@ def test_sparse_transposition(self):
...
+
+        # Summing horizontally along a row vector to create a single value
+        d = row.sum(axis=1)
+        assert d.to_numpy().shape == (1,1)
+
+        # # Summing vertically along a row vector to make sure nothing changes
+        e = row.sum(axis=0)
+        assert e.to_numpy().shape == (1, 10)
+
+        # FAILS
```
johnc1231, line 1099: Remove this `FAILS` comment.

hail/python/hail/ir/blockmatrix_ir.py
```diff
@@ -400,6 +399,7 @@ def _serialize_list(xs):
 
 
 def _matrix_shape_to_tensor_shape(n_rows, n_cols):
+
```
johnc1231, line 402: Shouldn't have a newline here.

PR #9029, opened 2020-06-30
[query] Fix bug in linreg aggregator when n == 0

tpoterba:
CHANGELOG: Fix crash when using `hl.agg.linreg` with no aggregated data.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/agg/LinearRegressionAggregator.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/agg/LinearRegressionAggregator.scala
@@ -13,7 +13,8 @@ object LinearRegressionAggregator {
   val vector = PCanonicalArray(scalar, true)
   val stateType: PCanonicalTuple = PCanonicalTuple(true, vector, vector, PInt32(true))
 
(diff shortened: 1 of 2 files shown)


hail/src/main/scala/is/hail/expr/ir/agg/LinearRegressionAggregator.scala
```diff
@@ -13,7 +13,8 @@ object LinearRegressionAggregator {
   val vector = PCanonicalArray(scalar, true)
   val stateType: PCanonicalTuple = PCanonicalTuple(true, vector, vector, PInt32(true))
 
-  def resultType: PCanonicalStruct = PCanonicalStruct(required = true, "xty" -> vector, "beta" -> vector, "diag_inv" -> vector, "beta0" -> vector)
+  private val optVector = vector.setRequired(false)
```
johnc1231, line 16: Why do we need `optVector` vs just making the original `vector` on line 13 not required?

tpoterba: We want the state to have required vectors, I think.

PR #9030, opened 2020-06-30
[query] Refactor LoweringPipeline so that optimization is a pass

tpoterba:
I need this for TableAggregate work, but think this is a good change regardless.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/lowering/LoweringPipeline.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/lowering/LoweringPipeline.scala
@@ -5,9 +5,6 @@ import is.hail.utils._
 
 case class LoweringPipeline(lowerings: LoweringPass*) {
   assert(lowerings.nonEmpty)
-  lowerings.zip(lowerings.tail).foreach { case (l, r) =>
-    assert(l.after == r.before)
-  }
 
   final def apply(ctx: ExecuteContext, ir: BaseIR): BaseIR = {
     var x = ir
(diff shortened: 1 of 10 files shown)


hail/src/main/scala/is/hail/expr/ir/lowering/LoweringPipeline.scala
```diff
@@ -1,29 +1,24 @@
...
     lowerings.foreach { l =>
       try {
         x = l.apply(ctx, x)
-        if (optimize)
-          x = Optimize(x, noisy = true, context = s"${l.context}, post Lowering", ctx)
       } catch {
-        case e: HailException => throw e
         case e: Throwable =>
-          throw new RuntimeException(s"error while applying lowering '${ l.context }'", e)
+          log.error(s"error while applying lowering '${ l.context }'")
```
tpoterba, line 20: slight change with what's going on in error handling -- I think this is better, in that it doesn't wrap errors and confuse people, but still puts the info in the log.

patrick-schultz: This is an improvement for the user experience, but makes the dev experience worse. Right now, the wrapped stack trace (the one I actually care about) shows up in the IDE, with clickable line numbers. With the change, I would have to open the log, and navigate to the referenced lines manually.

Could we maybe control this behavior with an environment flag? Or is there a trick I don't know for working with stack traces in logs?

tpoterba: the wrapped error is the only error now. I basically never used the stack trace to this line, because it's control flow code and the same every time.

patrick-schultz: Yeah, just realized that. Never mind!

hail/src/main/scala/is/hail/expr/ir/lowering/LoweringPipeline.scala
```diff
@@ -32,11 +27,83 @@ case class LoweringPipeline(lowerings: IndexedSeq[LoweringPass]) {
...
+      LowerMatrixToTablePass,
+      OptimizePass("relationalLowerer, after LowerMatrixToTable"),
+      InterpretNonCompilablePass,
+      OptimizePass("relationalLowerer, after InterpretNonCompilable")
+    )
+  else
+    LoweringPipeline(
+      LowerMatrixToTablePass,
+      InterpretNonCompilablePass
+    )
```
patrick-schultz, line 42: An idea to simplify the pipeline definitions: you could add a method on `LoweringPipeline` to drop all optimization passes. Then you could just define the pass once, and return `if (optimize) pass else pass.noOptimization`. For a cross cutting concern like whether to optimize between every other pass, I like that better. But if you prefer this, I'm fine with it.

tpoterba: oh, definitely better, thanks for the suggestion. That's why I non-randomly assigned you 😉

tpoterba:
I removed the assertions about the chain of IRState properties. We could take the parameters in OptimizePass, but that seems a bit hacky/unnecessary. The IRStates are checked at runtime, so I'm confident we'll still be able to debug issues.

patrick-schultz:
Oh, I didn't realize we're running pre- and post-condition checks on every pass. I would think the main reason to require the post-condition of one pass to match the pre-condition of the next (besides general hygiene) is to only have to do the check once. Anyways, I agree this seems fine for now.

I think the root of the issue is that we're assuming that `Optimize` preserves all possible `IRState`s, which is a pretty bold claim. Eventually we should probably refactor `Optimize` to separate out rules that apply to different `IRState`s.

PR #9033, opened 2020-06-30
[query] add StreamMultiMerge and StreamZipJoin nodes

patrick-schultz:
This PR adds stream nodes `StreamMultiMerge` and `StreamZipJoin`, which will be used to implement `TableUnion` and `TableMultiWayZipJoin`. The two nodes were so similar, both in implementation and in interface, that I thought bundling them into one PR would actually make it *easier* to review, but I can split them up if you disagree.

The implementations of both nodes use tournament trees, a very simple data structure ideal for this problem. Think of it as a priority queue specialized to holding exactly `k` elements, so when you pop the top element, you must immediately replace it with a new value.

A tournament tree is just what it sounds like. It is a complete binary tree with `k` leaves. Conceptually, the leaves hold the current `k` elements (the heads of each of the `k` input streams), while each internal node records the result of the comparison between the "winners" of the two subtrees, where in this case the winner is the least element. Thus we can find the smallest of all `k` elements by looking at the root node.

If we remove the smallest element, and replace it with the next value from that stream, we change the value in the corresponding leaf node, then we just need to replay the comparisons at all internal nodes on the path to the root. Note that to replay a comparison, we only need to know what element *lost* at this node previously. It must have lost to the previous overall winner, the element we just replaced.

Using that observation, we only need to store the `k` current values in the `k` leaves, and in each of the `k-1` internal nodes we store the index of the element which lost the comparison at that node. That just leaves the overall winner, which we can store in a separate variable. Note that each element besides the overall winner loses exactly one "match", so the internal nodes store a permutation of the indices 0 to (k-1), minus the overall winner. This is a so-called "loser tree".

In the implementation, I store the `k` leaves in a `Array[Long]`, where each element is a pointer to the head of the corresponding stream, and I store the `k-1` internal nodes in a `Array[Int]`, in the usual breadth-first order, where each element is the index of the stream which lost the comparison at that node (had the larger value). I use an index of `-1` to represent an imaginary element smaller than all real elements, and similarly an index of `k` is larger than all real elements. The tournament tree begins filled with `-1`, and each stream is advanced once, as their values push out all the `-1`s. When a steam ends, it's leaf is given the value `k`, and once the overall winner is `k`, we know all streams have ended.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
@@ -84,20 +84,26 @@ object COption {
       some(value)
   }
 
-  def lift[A](opts: IndexedSeq[COption[A]]): COption[IndexedSeq[A]] = new COption[IndexedSeq[A]] {
-    def apply(none: Code[Ctrl], some: IndexedSeq[A] => Code[Ctrl])(implicit ctx: EmitStreamContext): Code[Ctrl] = {
-      val L = CodeLabel()
-      def nthOpt(i: Int, acc: IndexedSeq[A]): Code[Ctrl] =
-        if (i == 0)
-          opts(i)(Code(L, none), a => nthOpt(i+1, acc :+ a))
-        else if (i == opts.length - 1)
-          opts(i)(L.goto, a => some(acc :+ a))
-        else
-          opts(i)(L.goto, a => nthOpt(i+1, acc :+ a))
+  def lift[A](opts: IndexedSeq[COption[A]]): COption[IndexedSeq[A]] =
+    if (opts.length == 0)
+      COption.present(FastIndexedSeq())
+    else if (opts.length == 1)
+      opts.head.map(a => IndexedSeq(a))
+    else
+      new COption[IndexedSeq[A]] {
+        def apply(none: Code[Ctrl], some: IndexedSeq[A] => Code[Ctrl])(implicit ctx: EmitStreamContext): Code[Ctrl] = {
+          val L = CodeLabel()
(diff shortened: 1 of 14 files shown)


hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
```diff
@@ -801,7 +897,157 @@ object EmitStream {
...
+
+      val runMatch = CodeLabel()
+      val LpullChild = CodeLabel()
+      val LloopEnd = CodeLabel()
+      val LaddToResult = CodeLabel()
+      val LstartNewKey = CodeLabel()
+      val Leos = CodeLabel()
+      val Lpush = CodeLabel()
+
+      Code(Leos, eos)
```
chrisvittal, line 959: I don't understand what side effects this statement has that makes it work. It appears that we create a `Code` and then just throw it away. How does `eos` get emitted?

patrick-schultz: Yeah, this syntax is a bit hacky. As part of `CodeBuilder`ifying streams, I have plans to clean this up.

A label is really just an lir block. It starts empty (`Leos._start.first == null`). This line defines the contents of the block (by calling `Leos.end.append(lir.goto(eos.start))`, where `Leos.end == Leos._start` since it's empty), so it is no longer empty. But if `Leos` were never mentioned again, there would be no references to that block after this scope ends, and it would get GCed. But if we jump to this label from some other block, that block will contain a pointer to this one (in the control statement that jumps to the label), so there will be a path from the root block of the method we're building to `Leos._start`.

PR #9035, opened 2020-06-30
[batch] BatchPoolExecutor

danking:
I can break this up further if you want.

Big changes:

- change batch.py to support multi-line commands (use `{\n...\n}`)
- change batch.py and job.py to support per-job environment variables (and add tests to test_batch.py)
- add `partition` to hail top mirroring the implementation in Scala
- implement BatchPoolExecutor which attempts to faithfully implement the interface of concurrent.futures.Executor


danking:
bump. I didn't realize this hadn't merged already.

hail/python/test/hailtop/batch/test_batch_pool_executor.py
```diff
@@ -0,0 +1,145 @@
...
+            list(bpe.map(lambda _: sleep_forever(), range(5), timeout=2))
+        except asyncio.TimeoutError:
+            pass
+        else:
+            assert False
+
+
+def test_map_error_without_wait_no_error():
+    with BatchPoolExecutor(project='hail-vdc') as bpe:
+        bpe.map(lambda _: sleep_forever(), range(5), timeout=2)
```
jigold, line 88: Why doesn't this not throw an error?

hail/python/test/hailtop/batch/test_batch_pool_executor.py
```diff
@@ -0,0 +1,145 @@
...
+        else:
+            assert False
+
+
+def test_no_exception_when_shutdown():
+    def raise_value_error():
+        raise ValueError('dead')
+    with BatchPoolExecutor(project='hail-vdc') as bpe:
+        future = bpe.submit(raise_value_error)
+    try:
```
jigold, line 135: I don't see how this test is different than the one above and where the shutdown comes into play.

jigold:
You still have test failures.

hail/python/test/hailtop/batch/test_batch_pool_executor.py
danking: The documentation of Executor.map (the interface I'm implementing) states:
> The returned iterator raises a concurrent.futures.TimeoutError if __next__() is called and the result isn’t available after timeout seconds from the original call to Executor.map(). timeout can be an int or a float. If timeout is not specified or None, there is no limit to the wait time.

The returned iterator is discarded, no one calls next, converts it to a list, etc. So we never violate a timeout. I changed from sleep_forever to sleep for 10 seconds so as not to create run-forever jobs.

hail/python/test/hailtop/batch/test_batch_pool_executor.py
danking: I've renamed to `when_exiting_context`. Leaving the context just invokes the shutdown code.

I should not raise an error when I leave the context of the BatchPoolExecutor, even if `wait_on_exit=True`. The BatchPoolExecutor will wait for all tasks to complete (possibly with an exception) and then shut everything down. I won't receive that exception until I actually look at the future itself, as I do below. In a previous version of BPE, I had a bug where shutdown actually raised exceptions for failing tasks.

danking:
The tests will pass now. I had to disable output checking in the doctests because `Batch.run` prints batch ids and urls which would be annoying to maintain.

danking:
it's passing

hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -0,0 +1,506 @@
...
+        example for more detail. It is not possible to pass keyword arguments.
+
+        This function returns a generator which will produce each result in
+        order, only blocking if the result is not yet ready. You can convert the
+        generator to a list with :func:`.list`.
+
+        Examples
+        --------
+
+        Do nothing, but on the cloud:
```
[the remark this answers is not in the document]
jigold: Iterate over a range of numbers, but on the cloud:

hail/python/test/hailtop/batch/test_batch_pool_executor.py
jigold: I still see sleep forever is 1 hour.

hail/python/test/hailtop/batch/test_batch_pool_executor.py
```diff
@@ -0,0 +1,145 @@
...
+            list(bpe.map(lambda _: sleep_forever(), range(5), timeout=2))
+        except asyncio.TimeoutError:
+            pass
+        else:
+            assert False
+
+
+def test_map_error_without_wait_no_error():
+    with BatchPoolExecutor(project='hail-vdc', wait_on_exit=False) as bpe:
+        bpe.map(lambda _: time.sleep(10), range(5), timeout=2)
```
danking, line 88: @jigold I think you're looking at old changes. All the sleep forevers in the current commit should get cancelled.

PR #9041, opened 2020-07-01
[query] Extend BindingEnv to include relational bindings

tpoterba:
Use in ComputeUsesAndDefs.


hail/src/main/scala/is/hail/expr/ir/Binds.scala
```diff
@@ -72,12 +72,17 @@ object AggBindings {
     def base: Option[Iterable[(String, Type)]] = parent.agg.map(_ => FastIndexedSeq())
 
     x match {
-      case AggLet(name, value, _, false) => if (i == 1) wrapped(FastIndexedSeq(name -> value.typ)) else base
-      case AggExplode(a, name, _, false) => if (i == 1) wrapped(FastIndexedSeq(name -> a.typ.asInstanceOf[TIterable].elementType)) else base
-      case AggArrayPerElement(a, elementName, _, _, _, false) => if (i == 1) wrapped(FastIndexedSeq(elementName -> a.typ.asInstanceOf[TIterable].elementType)) else base
+      case AggLet(name, value, _, false) => if (i == 1) wrapped(FastIndexedSeq(name -> value.typ)) else None
```
catoverdrive, line 75: Can you clarify why this changed from base to None? Is it just because the `value` IR can't access the aggScope?

tpoterba: yes, exactly. My change to use AggBindings in ChildEnvWithoutBindings revealed a bunch of latent incorrect rules.

PR #9044, opened 2020-07-01
[query] Per Pipeline Static Filesystem

chrisvittal:
The interfaces are `newStaticField` and `newStaticMethod` on (Emit)ClassBuilder. The first
usage of this feature is to move the `FS` off of the class itself and onto a container class.

### DIFF
--- a/hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
+++ b/hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
@@ -79,7 +79,8 @@ class ModuleBuilder() {
 
   def newClass[C](name: String)(implicit cti: TypeInfo[C]): ClassBuilder[C] = {
     val c = new ClassBuilder[C](this, name)
-    c.addInterface(cti.iname)
+    if (cti != UnitInfo)
+      c.addInterface(cti.iname)
     classes += c
     c
   }
--- a/hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
+++ b/hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
@@ -154,6 +154,9 @@ trait WrappedClassBuilder[C] extends WrappedModuleBuilder {
     maybeGenericReturnTypeInfo: MaybeGenericTypeInfo[_]): MethodBuilder[C] =
     cb.newMethod(name, maybeGenericParameterTypeInfo, maybeGenericReturnTypeInfo)
 
+  def newStaticMethod(name: String, parameterTypeInfo: IndexedSeq[TypeInfo[_]], returnTypeInfo: TypeInfo[_]): MethodBuilder[C] =
+    cb.newStaticMethod(name, parameterTypeInfo, returnTypeInfo)
+
   def getOrGenMethod(
     baseName: String, key: Any, argsInfo: IndexedSeq[TypeInfo[_]], returnInfo: TypeInfo[_]
   )(body: MethodBuilder[C] => Unit): MethodBuilder[C] =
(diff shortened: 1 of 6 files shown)


chrisvittal:
This is the start of a series of changes that will culminate in the generation of static ordering methods between PTypes.

The next step here is to get reference genomes into static container classes as well. We need this change first because
reference genomes may use the filesystem to load a fasta.

chrisvittal:
cc @cseed @tpoterba

chrisvittal:
This still isn't perfect because it requires calling `addFS` from at least one instance of a class. This won't work for user added reference genomes, which may read a FASTA file from what I want to be a static method. Still working on a solution there, though it will probably be in the next PR.

cseed:
> The first usage of this feature is to move the FS off of the class itself and onto a container class.

Is the container class generated globally?  That's going to be a problem.  The service backend has multiple FSes for multiple, different users.  Keeping those FSes separate is **critical** for security.

The reference genomes are the same story.  Each service query will come with its own, distinct context of reference genomes, and they can't get mixed up.  In particular, a reference genome sequence may be sensitive data and can't be exposed to other users.  The reference genome global state will only be stored in the Python client.  Removing the global reference state from the Scala side is a pending Query service project.

chrisvittal:
My assumption has been that each fresh pipeline should get a new `EmitModuleBuilder` and `ExecuteContext`, if this is the case, I believe what I have done here should be alright.

chrisvittal:
Furthermore, we can't have static comparison functions without static reference genomes, so a solution will need to be found. This is a step in that direction.

cseed:
> My assumption has been that each fresh pipeline should get a new EmitModuleBuilder and ExecuteContext, if this is the case, I believe what I have done here should be alright.

OK, great.  I jumped the gun.  But something to think about in places we retry to share across invocations of Compile.

 > The first usage of this feature is to move the FS off of the class itself and onto a container class.

If the the FS field in the container class is static, how does it get serialized?  The point is the C returned by EmitClassBuilder.resultWithIndex captures and serializes the FS.

 > Furthermore, we can't have static comparison functions without static reference genomes, so a solution will need to be found. This is a step in that direction.

The RG can be serialized.  That's what EmitClassBuilder.getReferenceGenome does.  So you can put the reference genomes in static classes that decode them lazily, and then the comparison functions and main code can call into those static methods.  This doesn't work for FS because (currently) it can't be serialised as code.

chrisvittal:
There are code paths where ReferenceGenome's `codeSetup` invokes the filesystem. Those must be handled in a static context in order for this to work at all.

https://github.com/tpoterba/hail/blob/e5b3fb3d2fba971d0c226094ae0c7ac66f190bbe/hail/src/main/scala/is/hail/variant/ReferenceGenome.scala#L493-L512
```scala
    if (fastaReader != null) {
      rg = rg.invoke[String, FS, String, String, Int, Int, ReferenceGenome](
        "addSequenceFromReader",
        localTmpdir,
        cb.getFS,
        fastaReader.fastaFile,
        fastaReader.indexFile,
        fastaReader.blockSize,
        fastaReader.capacity)
    }


    for ((destRG, lo) <- liftoverMaps) {
      rg = rg.invoke[String, FS, String, String, ReferenceGenome](
        "addLiftoverFromFS",
        localTmpdir,
        cb.getFS,
        lo.chainFile,
        destRG)
    }
    rg
```

The static isn't final or anything, the setter is still currently called from the `FunctionWithFS` interface's `addFS` method, so the `resultWithIndex` closure is still what captures and saves the FS. Like I said, needs some work to make it work always.

chrisvittal:
I figured it out, `FS` is already searialazable, so we can just serialize the data into the class at compile time and deserialize it at load time.

PR #9046, opened 2020-07-02
[query] Move reference genomes onto static classes

chrisvittal:
One class per genome is generated per module.


tpoterba:
This is still stacked, right?

chrisvittal:
Yes, but either this or #9044 could go in.

PR #9048, opened 2020-07-02
[query] Add ndarray to all_types, test read/write for T/MT

tpoterba:


hail/python/test/hail/helpers.py
```diff
@@ -90,7 +90,8 @@ def create_all_values():
         c=hl.call(0, 1),
         mc=hl.null(hl.tcall),
         t=hl.tuple([hl.call(1, 2, phased=True), 'foo', hl.null(hl.tstr)]),
-        mt=hl.null(hl.ttuple(hl.tlocus('GRCh37'), hl.tbool))
+        mt=hl.null(hl.ttuple(hl.tlocus('GRCh37'), hl.tbool)),
+        nd=hl.nd.array(hl.range(0, 10)).reshape((2, 5))
```
johnc1231, line 94: Just as an FYI, the easier way to do this is `hl.nd.arange(10).reshape((2, 5))`.

tpoterba: oh, duh.

johnc1231:
Missed renaming an `ndarray_to_array`.

PR #9049, opened 2020-07-03
[query] Add forLoop to CodeBuilder

chrisvittal:
currently unused, may be useful in the future.

### DIFF
--- a/hail/src/main/scala/is/hail/asm4s/CodeBuilder.scala
+++ b/hail/src/main/scala/is/hail/asm4s/CodeBuilder.scala
@@ -99,6 +99,26 @@ trait CodeBuilderLike {
 
   def whileLoop(c: Code[Boolean], emitBody: => Unit): Unit = whileLoop(c, _ => emitBody)
 
+  def forLoop(setup: => Unit, cond: Code[Boolean], incr: => Unit, emitBody: (CodeLabel) => Unit): Unit = {
+    val Lstart = CodeLabel()
+    val Lbody = CodeLabel()
+    val Lafter = CodeLabel()
+    val Lincr = CodeLabel()
+
+    setup
+    define(Lstart)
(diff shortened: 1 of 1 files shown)


chrisvittal:
No need to merge this anytime soon. Just had the itch and wanted to write it up. May be useful in porting stream code to CodeBuilder eventually.

tpoterba:
This is very nice. There are three while loops in BgenRDDPartitions.scala that could easily be translated to for loops, so that it gets tested

chrisvittal:
I think that would require CodeBuildering bgen... Ok!

hail/src/main/scala/is/hail/asm4s/CodeBuilder.scala
```diff
@@ -99,6 +99,26 @@ trait CodeBuilderLike {
 
   def whileLoop(c: Code[Boolean], emitBody: => Unit): Unit = whileLoop(c, _ => emitBody)
 
+  def forLoop(setup: => Unit, cond: Code[Boolean], incr: => Unit, emitBody: (CodeLabel) => Unit): Unit = {
```
catoverdrive, line 102: Looks good! I kinda wish there were a way to make it obvious from the method signature what the CodeLabel argument was here---it looks like it's essentially so things like conditions have an "after" label to jump to?

chrisvittal: I wish I could make it more obvious too. The label argument is `continue`. There's currently no `break`, though it can be easily simulated with an extra label.

```scala
val Lbreak = CodeLabel()
cb.whileLoop(cond, {
     // do stuff
     cb.ifx(cond2, cb.goto(Lbreak))
     // do more stuff
})
cb.define(Lbreak)
```

PR #9061, opened 2020-07-08
[query] Expose _nd as nd

tpoterba:
Pycharm thought `hl.nd` didn't exist, and I'm pretty sure we could
have had issues on certain python installations without this change.


johnc1231:
Looks like there were some more `hl._nd`'s around in tests. Sorry about that, I can fix if you'd like.

PR #9063, opened 2020-07-08
[query] Add tree_matmul, matrix multiply in case of large inner dimension 

johnc1231:
This PR introduces `tree_matmul`, a method on `BlockMatrix` that allows for greater parallelism when multiplying two large matrices that result in a small matrix (i.e. the inner dimension is much larger than the outer dimensions). 

In order to do this, this PR takes a `split_on_inner` parameter. This parameter defines how many subdivisions to break the two matrices into. Corresponding subdivisions are multiplied, written to disk, then read back in and summed. 

For example, if you are multiplying a 4k by 500k matrix by its transpose, your answer will be a 4k by 4k block. With normal matrix multiply, you'd get no parallelism at all, since the result is only a single partition. If you instead use `tree_matmul` with `split_on_inner = 5`, hail will do the following:

1. Break up 4k by 500k matrix into five 4k by 100k matrices.
2. Break up 500k by 4k matrix into five 100k by 4k matrices. 
3. Multiply corresponding matrices together, creating five 4k by 4k matrices.
4. Write them all to disk, then read them back in.
5. Sum the 5 read in matrices together to get your result.

This PR also introduces a `write_block_matrices` function that writes out a list of `BlockMatrix` in parallel. This was necessary to write out all of the intermediate matrices in `tree_matmul`. 

@konradjk 
@danking

### DIFF
--- a/hail/python/hail/linalg/blockmatrix.py
+++ b/hail/python/hail/linalg/blockmatrix.py
@@ -564,11 +564,13 @@ class BlockMatrix(object):
 
     @property
     def _last_col_block_width(self):
-        return self.n_cols % self.block_size
+        remainder = self.n_cols % self.block_size
+        return remainder if remainder != 0 else self.block_size
 
     @property
     def _last_row_block_height(self):
-        return self.n_rows % self.block_size
+        remainder = self.n_rows % self.block_size
+        return remainder if remainder != 0 else self.block_size
 
     @typecheck_method(path=str,
                       overwrite=bool,
@@ -1435,7 +1437,7 @@ class BlockMatrix(object):
         start_col = start_bcol * self.block_size
         stop_col = (stop_bcol - 1) * self.block_size + (self._last_col_block_width if stop_bcol == self._n_block_cols else self.block_size)
 
-        #print(f"{block_row_range}, {block_col_range} -> {(start_row, stop_row)}, {(start_col, stop_col)}")
+        print(f"{block_row_range}, {block_col_range} -> {(start_row, stop_row)}, {(start_col, stop_col)}")
 
         return self[start_row:stop_row, start_col:stop_col]
 
@@ -1463,7 +1465,7 @@ class BlockMatrix(object):
             inner_ranges = list(zip(split_points[:-1], split_points[1:]))
             print(inner_ranges)
             blocks_to_multiply = [(self._select_blocks((0, self._n_block_rows), (start, stop)),
-                                self._select_blocks((start, stop), (0, self._n_block_cols))) for start, stop in inner_ranges]
+                                b._select_blocks((start, stop), (0, b._n_block_cols))) for start, stop in inner_ranges]
 
             intermediate_multiply_exprs = [b1 @ b2 for b1, b2 in blocks_to_multiply]
 
--- a/hail/python/hail/linalg/blockmatrix.py
+++ b/hail/python/hail/linalg/blockmatrix.py
@@ -1459,7 +1459,7 @@ class BlockMatrix(object):
 
         return BlockMatrix(BlockMatrixDot(self._bmir, b._bmir))
 
-    @typecheck_method(b=oneof(np.ndarray, block_matrix_type), split_on_inner=int, path_prefix=nullable(str))
+    @typecheck_method(b=oneof(np.ndarray, block_matrix_type), splits=int, path_prefix=nullable(str))
     def tree_matmul(self, b, *, splits, path_prefix=None):
         """Matrix multiplication in situations with large inner dimension.
 
--- a/hail/python/hail/linalg/blockmatrix.py
+++ b/hail/python/hail/linalg/blockmatrix.py
@@ -1441,8 +1441,8 @@ class BlockMatrix(object):
 
         return self[start_row:stop_row, start_col:stop_col]
 
-    @typecheck_method(b=oneof(np.ndarray, block_matrix_type), _split_on_inner=int)
-    def __matmul__(self, b, _split_on_inner=1):
+    @typecheck_method(b=oneof(np.ndarray, block_matrix_type))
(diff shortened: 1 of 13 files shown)


hail/python/test/hail/linalg/test_linalg.py
```diff
@@ -457,6 +457,36 @@ def test_matrix_ops(self):
...
+
+        self._assert_eq(m.tree_matmul(m.T, 2, new_temp_file()), nm @ nm.T)
+        self._assert_eq(m.tree_matmul(nm.T, 2, new_temp_file()), nm @ nm.T)
+        self._assert_eq(row.tree_matmul(row.T, 2, new_temp_file()), nrow @ nrow.T)
+        self._assert_eq(row.tree_matmul(nrow.T, 2, new_temp_file()), nrow @ nrow.T)
+
+        self._assert_eq(m.T.tree_matmul(m, 2, new_temp_file()), nm.T @ nm)
+        self._assert_eq(m.T.tree_matmul(nm, 2, new_temp_file()), nm.T @ nm)
+        self._assert_eq(row.T.tree_matmul(row, 2, new_temp_file()), nrow.T @ nrow)
+        self._assert_eq(row.T.tree_matmul(nrow, 2, new_temp_file()), nrow.T @ nrow)
```
danking, line 474: call me paranoid, but it would be nice to have at least one example of x @ y where x and y are both block matrices and neither is the transpose of the other.

hail/src/main/scala/is/hail/linalg/BlockMatrix.scala
```diff
@@ -259,15 +273,15 @@ object BlockMatrix {
...
-    customFilenames: Option[Array[String]]): Unit = {
+                           fs: FS,
+                           bms: IndexedSeq[BlockMatrix],
+                           prefix: String,
+                           overwrite: Boolean,
+                           delimiter: String,
+                           header: Option[String],
+                           addIndex: Boolean,
+                           compression: Option[String],
+                           customFilenames: Option[Array[String]]): Unit = {
```
danking, line 284: this is an odd formatting change? Maybe add a newline before the `)` to get the desired formatting from IntelliJ?

hail/python/hail/linalg/blockmatrix.py
```diff
@@ -1427,6 +1459,47 @@ def __matmul__(self, b):
 
         return BlockMatrix(BlockMatrixDot(self._bmir, b._bmir))
 
+    @typecheck_method(b=oneof(np.ndarray, block_matrix_type), split_on_inner=int, path_prefix=str)
+    def tree_matmul(self, b, split_on_inner, path_prefix):
+        """Matrix multiplication in situations with large inner dimension. This function splits a single matrix
```
danking, line 1464: We generally prefer a one sentence description on its own line, followed by a full description. Some python tools use this in tooltips

hail/python/hail/linalg/blockmatrix.py
```diff
@@ -1427,6 +1459,47 @@ def __matmul__(self, b):
 
         return BlockMatrix(BlockMatrixDot(self._bmir, b._bmir))
 
+    @typecheck_method(b=oneof(np.ndarray, block_matrix_type), split_on_inner=int, path_prefix=str)
+    def tree_matmul(self, b, split_on_inner, path_prefix):
```
danking, line 1463: I think adding a default argument for path_prefix of `None` and creating a temp file in that case will improve usability.

danking, line 1463: I think we should make either the last or the last two parameters keyword-only arguments, so: `(self, b, *, split_on_inner, path_prefix)`.

hail/src/main/scala/is/hail/linalg/BlockMatrix.scala
johnc1231: Thanks, this was an accident.

hail/python/test/hail/linalg/test_linalg.py
johnc1231: Heh, I just copied the current matmul tests plus that looping of different block sizes, but you're right, these are only transposes. I'll add another test.

hail/python/hail/linalg/blockmatrix.py
johnc1231: I'd be willing to make `path_prefix` a keyword only argument, but can you explain why you'd prefer that? I think `split_on_inner` I'll leave the way it is, since you always have to specify it. `path_prefix` with a default value of `None` as your above comment suggests makes the keyword only argument sound better.

johnc1231:
I addressed most of your comments, I'm not sure why you want the keyword only arguments though.

danking:
It's kind of a gut reaction for me. There's some conversation at [PEP 1302's rationale](https://www.python.org/dev/peps/pep-3102/#rationale). For me, everything is about backwards compatibility and the two cases are:
- changing from positional arguments without varargs to varargs and keywords is backwards-incompatible.
- changing a positional-without-default to positional-with-default values is backwards-incompatible unless the positional argument is the last one.

In this concrete example, I worry about two possible evolutions of this function.

1. Accept multiple other matrices:
```
def tree_matmul(self, *others, split_on_inner, path_prefix)
```
2. Allow the compiler/python code to choose a good value for split_on_inner
```
```

danking:
I might also call `split_on_inner` `splits` for brevity in the keyword case.

patrick-schultz:
@danking Should I review this too, or do you feel like you covered it?

danking:
@patrick-schultz I only looked at the Python interface, so I would much appreciate your thoughts on the Scala stuff.

johnc1231:
Tagging WIP because I have two tiny changes Dan asked for to make (argument name and keyword only)

PR #9069, opened 2020-07-09
[query] Lower TableMapRows with scans

tpoterba:


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -877,7 +877,7 @@ class Emit[C](
           cb += srvb.advance()
         }, "aggs_result")
 
-        sc.store(cb)
+//        sc.store(cb)
```
catoverdrive, line 880: Why do we need to take this out? Store has logic in it that releases region references for aggregator state fields.

tpoterba: ResultOP can be called multiple times. If it doesn't load, it can't store.

catoverdrive: Womp. Do we have a thing to clean up region references post-aggregation/can we guarantee that they're cleared if we take this out?

tpoterba: I'll investigate this...

tpoterba: Kinda think RunAgg should clean up after

PR #9071, opened 2020-07-10
[query] Reimplement Table._same as outer join on row idx

tpoterba:
Is it this simple?

### DIFF
--- a/hail/python/test/hail/table/test_grouped_table.py
+++ b/hail/python/test/hail/table/test_grouped_table.py
@@ -8,8 +8,7 @@ tearDownModule = stopTestHailContext
 
 
 class GroupedTableTests(unittest.TestCase):
-    def \
-            test_aggregate_by(self):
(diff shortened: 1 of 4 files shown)


patrick-schultz:
That looks like it should work (though some tests are failing), but rekeying by the global index might be more expensive than necessary. What about using `multi_way_zip_join`?

tpoterba:
OK, working now.

hail/python/test/hail/table/test_grouped_table.py
```diff
@@ -8,7 +8,8 @@
 
 
 class GroupedTableTests(unittest.TestCase):
-    def test_aggregate_by(self):
+    def \
+            test_aggregate_by(self):
```
patrick-schultz, line 12: What happened here?

tpoterba: oh, oopz

PR #9074, opened 2020-07-10
[batch] input file cache

jigold:
- I left in pvc_size for backwards compatibility. We can rip it out at some point.
- I'm not happy with how the batch_worker_image seems slower now with the gsutil addition. Not sure if there's a better solution here.

I tested a lot of this by hand on dev deploy. For example, making sure the flocks were right, the project quotas were right, the cache was working, the garbage collector, and the backwards compatibility with pvc_size was working. Let me know if you think there are other things I should test.

If this is too much, I can think about splitting the storage and the cache into two PRs. I just figured since I almost had it all done together to push on that.

### DIFF
--- a/batch/Dockerfile.worker
+++ b/batch/Dockerfile.worker
@@ -3,7 +3,8 @@ FROM python:3.7-slim-stretch
 RUN apt-get update && \
   apt-get -y install \
     curl \
-    gnupg && \
+    gnupg \
+    xfsprogs && \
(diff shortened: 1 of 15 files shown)


jigold:
Closing in favor of separate PRs while I figure out how to handle wild cards in path names. Storage quotas are in #9076

jigold:
Closing for now while I dev deploy.

PR #9076, opened 2020-07-11
[batch] add storage quotas to worker

jigold:


jigold:
This is working except I can't figure out how to delete entries from /etc/projects and /etc/projid when the job finishes. You can't copy or move a file to a bind-mounted file in Docker or you get a "resource or device busy" error. This will all work without deleting the entries, but then every time you try and add a new project, you get a directory not found error for previously deleted paths, but the operation will continue successfully ignoring the errors.

Any ideas on how to get around this? `sed -i` won't work as well as `mv` and `cp`.

jigold:
The only thing that might work is using the in_place option in this Python class: https://docs.python.org/3/library/fileinput.html

jigold:
After this is merged, we need to warn users that storage quotas actually are enforced.

jigold:
This is working now!!!! I'm happy with it :)

batch/batch/utils.py
```diff
@@ -118,6 +122,29 @@ def adjust_cores_for_memory_request(cores_in_mcpu, memory_in_bytes, worker_type)
...
+def worker_storage_per_core_bytes(worker_cores):
+    return (total_worker_storage() * 1024**3) // worker_cores
+
+
+def storage_bytes_to_cores_mcpu(storage_in_bytes, worker_cores):
+    return math.ceil((storage_in_bytes / worker_storage_per_core_bytes(worker_cores)) * 1000)
+
+
+def cores_mcpu_to_storage_bytes(cores_in_mcpu, worker_cores):
+    return int((cores_in_mcpu / 1000) * worker_storage_per_core_bytes(worker_cores))
```
danking, line 140: I suspect these two functions will work fine, but floating point arithmetic gives me the heebie jeebies. I prefer:

```
(cores_in_mcpu * worker_storage_per_core_bytes(worker_cores)) // 1000
```
and
```
def round_up_division(items, group_size):
    return (items + group_size - 1) // group_size
# ...
1000 * round_up_division(storage_in_bytes, worker_storage_per_core_bytes(worker_cores))
```

I think I just find integer arithmetic less surprising and thus prefer it when possible. Moreover, the values in question here will divide evenly. Consider that the smallest CPU fraction we permit is `250mcpu`. 250 = 5 * 5 * 5 * 2. The worker storage per core bytes has at least 26 factors of two (30 from 1024**3 minus 4 from 16 cores). As such, our values for storage should always be divisible by `1000 = 5 * 5 * 5 * 2 * 2 * 2`.

batch/batch/worker/flock.py
```diff
@@ -0,0 +1,54 @@
...
+        if filename:
+            fd = os.open(self.path, os.O_CREAT)
+            self.fds.append(fd)
+            fcntl.flock(fd, self.flock_flags)
+
+        return self
+
+    def __exit__(self, type, value, traceback):
+        for fd in reversed(self.fds):
+            fcntl.flock(fd, fcntl.LOCK_UN)
```
danking, line 39: we should close the file descriptors

batch/batch/worker/flock.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class Flock:
+    def __init__(self, path, nonblock=False):
+        self.path = os.path.abspath(path)
+        self.flock_flags = fcntl.LOCK_EX
+        self.fds = []
+
+        if nonblock:
+            self.flock_flags |= fcntl.LOCK_NB
```
danking, line 14: Unless you need this I think we should remove it, we don't seem to catch the associated exception anywhere.
```
In [4]: fcntl.flock(open('/tmp/foo'), fcntl.LOCK_EX | fcntl.LOCK_NB)                                                                                                                           
  File "<ipython-input-4-8375ce168aff>", line 1, in <module>
    fcntl.flock(open('/tmp/foo'), fcntl.LOCK_EX | fcntl.LOCK_NB)
BlockingIOError: [Errno 35] Resource temporarily unavailable
```

danking: We'd have to stick the blocking activity onto a thread pool as we do elsewhere. We could maybe use `LOCK_NB` to first check if we can lock the file without using a thread pool.

batch/batch/worker/worker.py
```diff
@@ -361,6 +365,18 @@ async def run(self, worker):
...
 
+            c = await docker_call_retry(MAX_DOCKER_OTHER_OPERATION_SECS, f'{self}')(self.container.show)
+
+            merged_overlay_path = c['GraphDriver']['Data']['MergedDir']
+            assert merged_overlay_path.endswith('/merged')
+            self.overlay_path = merged_overlay_path[:-7].replace(LOCAL_SSD_MOUNT, '/host')
+            os.makedirs(f'{self.overlay_path}/', exist_ok=True)
+
+            await check_shell(f'''
+python3 -m batch.worker.flock /xfsquota/projects -c "echo \"{self.job.project_id}:{self.overlay_path}\" >> /xfsquota/projects"
```
danking, line 376: Hmm. This feels a bit weird. You you are shelling out from python to python. Is the issue that the different async threads might race to update the projects file? It seems more readable to do something like:

```python3
with Flock('/xfsquota/projects'):  # maybe need to also use a python lock to block other async threads
    with open('/xfsquota/projects', 'a') as f:
        f.write(f'{self.job.project_id}:{self.overlay_path}\n')
```

batch/batch/worker/worker.py
```diff
@@ -361,6 +365,18 @@ async def run(self, worker):
                 self.container = await docker_call_retry(MAX_DOCKER_OTHER_OPERATION_SECS, f'{self}')(
                     create_container, config, name=f'batch-{self.job.batch_id}-job-{self.job.job_id}-{self.name}')
 
+            c = await docker_call_retry(MAX_DOCKER_OTHER_OPERATION_SECS, f'{self}')(self.container.show)
+
+            merged_overlay_path = c['GraphDriver']['Data']['MergedDir']
+            assert merged_overlay_path.endswith('/merged')
+            self.overlay_path = merged_overlay_path[:-7].replace(LOCAL_SSD_MOUNT, '/host')
+            os.makedirs(f'{self.overlay_path}/', exist_ok=True)
```
danking, line 373: can you remind me the purpose of this block?

batch/batch/worker/worker.py
```diff
@@ -610,15 +633,22 @@ def __init__(self, batch_id, user, gsa_key, job_spec, format_version):
...
 
         self.cpu_in_mcpu = cpu_in_mcpu
         self.memory_in_bytes = cores_mcpu_to_memory_bytes(self.cpu_in_mcpu, worker_config.instance_type)
+        self.storage_in_bytes = cores_mcpu_to_storage_bytes(self.cpu_in_mcpu, CORES)
 
         self.resources = worker_config.resources(self.cpu_in_mcpu, self.memory_in_bytes)
 
+        self.project_name = f'batch-{self.batch_id}-job-{self.job_id}'
+        self.project_id = Job.quota_project_id
+        Job.quota_project_id += 1
```
danking, line 650: Let's wrap this into `Job.get_next_project_id()`.

danking: or maybe `next_xfs_project_id`

danking: or xfsquota maybe?

hail/python/hailtop/batch/docs/service.rst
```diff
@@ -119,10 +119,10 @@ scheduled on this machine, then the cost per core hour is **$0.02774**.
 
 .. note::
 
-    The amount of CPU reserved for a job can be rounded up if the equivalent memory request
+    The amount of CPU reserved for a job can be rounded up if the equivalent memory and/or storage request
     requires a larger fraction of the worker. Currently, each 1 core requested
-    gets 3.75 GB of memory. Therefore, if a user requests 1 CPU and 7 GB of memory, the user
-    will get 2 cores for their job and will be billed for 2 cores.
+    gets 3.75 GB of memory and 21.875 GB of storage. Therefore, if a user requests 1 CPU and 7 GB of memory,
+    the user will get 2 cores for their job and will be billed for 2 cores.
```
danking, line 125: We should do a batch support announce when this happens.

batch/batch/front_end/front_end.py
```diff
@@ -589,24 +591,34 @@ async def create_jobs(request, userdata):
                     resources['cpu'] = BATCH_JOB_DEFAULT_CPU
                 if 'memory' not in resources:
                     resources['memory'] = BATCH_JOB_DEFAULT_MEMORY
+                if 'storage' not in resources:
+                    # FIXME: pvc_size is deprecated
+                    pvc_size = spec.get('pvc_size')
```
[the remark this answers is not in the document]
jigold: Announce that users should upgrade their hail version after the next release and then give a deadline for switching until I make a new PR that rips the pvc_size out.

batch/batch/worker/flock.py
jigold: I needed it for the garbage collector for the file cache. I'll remove it from here and figure out the correct thing then.

batch/batch/worker/worker.py
jigold: I'm worried that the Flock command isn't async. I think I need an async implementation... sigh.

batch/batch/worker/worker.py
jigold: I want to add all of the container files it writes to the quota tracking. So I get the location of the merged dir and then figure out the root overlay path from that.

batch/batch/utils.py
jigold: Where do you get that it's always divisible by 1000? i.e. where do the powers of 5 come in 2**26?

batch/batch/driver/instance_pool.py
```diff
@@ -327,17 +332,20 @@ async def create_instance(self, cores=None, max_idle_time_msecs=None):
...
     -v /batch:/batch \
     -v /logs:/logs \
     -v /gcsfuse:/gcsfuse:shared \
+    -v /xfsquota:/xfsquota \
+    --mount type=bind,source=/mnt/disks/$LOCAL_SSD_NAME,target=/host \
     -p 5000:5000 \
     --device /dev/fuse \
-    --cap-add SYS_ADMIN \
-    --security-opt apparmor:unconfined \
+    --privileged \
```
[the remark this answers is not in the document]
jigold: This works:
```
jigold@jg-file-cache:~$ sudo docker run --rm --mount type=bind,source=/mnt/test_xfs,target=/host --privileged test-xfs /bin/bash -c 'xfs_quota -x -c "report -h" /host'
Project quota on /host (/dev/loop3)
                        Blocks
Project ID   Used   Soft   Hard Warn/Grace
---------- ---------------------------------
#0             4K      0      0  00 [------]
#200            0      0      0  00 [------]
```

This doesn't work with just SYS_ADMIN and apparmor:unconfined
```
jigold@jg-file-cache:~$ sudo docker run --rm --mount type=bind,source=/mnt/test_xfs,target=/host --cap-add SYS_ADMIN --security-opt apparmor:unconfined test-xfs /bin/bash -c 'xfs_quota -x -c "report -h" /host'
xfs_quota: cannot setup path for mount /host: No such device or address
```

I tried giving it all capabilities with --cap-add ALL and that doesn't work either.

hail/python/hailtop/batch/docs/service.rst
jigold: Agreed.

batch/batch/utils.py
jigold: This round_up_division function is dangerous!!! It rounds up 0.25 to 1.

jigold:
This is good to go now (tests are passing and the xfsquota stuff looks right when I looked at a worker by hand).

batch/batch/utils.py
danking: I suppose that's a bad name. It's something like `how_many_groups_do_i_need`.

`cores_in_mcpu`, unless I'm mistaken, is always a multiple of 250, ergo it always has three factors of five.

Ah, yes this would calculate the cores necessary, but we want the mcpu necessary for that much storage, ergo we should push the `1000 *` into the numerator:

```
round_up_division(1000 * storage_in_bytes, worker_storage_per_core_bytes(worker_cores))
```

batch/batch/driver/instance_pool.py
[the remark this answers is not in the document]
danking: I suspect we need to also mount the corresponding device. There are two docker flags which seem related:
```
--device "/dev/sdc:/dev/sdc:rwm"
```
```
--device-cgroup-rule "a 1:3 rwm"
```

Hopefully the former is sufficient to give `xfs_quota` access. The latter requires us to figure out the device numbers.

jigold:
@danking Can you double check the last commit is fine?

danking:
haha, you are an Awk Queen! This looks great to me, I'm glad we were able to nail down the security.

jigold:
@danking I need to address the benchmark storage requests to not be 100 Gi.

danking:
I don't think we should hold up this PR for that. Benchmarks are run rarely and they should be setting their limits using PYSPARK_SUBMIT_ARGS anyway.

danking:
They are users of batch like anyone else and are subject to the same changes

PR #9078, opened 2020-07-13
[hailctl dataproc] Add some tests for hailctl dataproc modify

nawatts:
Follow on to #9066.

Still todo: tests for `--update-hail-version`. Those will require a mock for deploy metadata.

### DIFF
--- a/hail/python/hailtop/hailctl/dataproc/modify.py
+++ b/hail/python/hailtop/hailctl/dataproc/modify.py
@@ -1,8 +1,9 @@
 import os.path
 import sys
-import subprocess as sp
 import yaml
 import pkg_resources
+
+from . import gcloud
 
 
 def init_parser(parser):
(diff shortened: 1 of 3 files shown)


hail/python/hailtop/hailctl/dataproc/modify.py
```diff
@@ -39,32 +40,20 @@ def main(args, pass_through_args):
...
             print("Updating cluster '{}'...".format(args.name))
-            sp.check_call(cmd)
-
-    if args.zone:
-        zone = args.zone
-    else:
-        zone = sp.check_output(["gcloud", "config", "get-value", "compute/zone"], stderr=sp.DEVNULL).decode().strip()
-
-    if not zone:
-        raise RuntimeError("Could not determine compute zone. Use --zone argument to hailctl, or use `gcloud config set compute/zone <my-zone>` to set a default.")
```
nawatts, line 67: Zone is only required when modifying the Hail wheel on the cluster. With this check here, `hailctl dataproc modify some-cluster --num-workers=2` would unnecessarily error if no zone was provided or `compute/zone` was not set.

tpoterba: ah, didn't know that!

nawatts: It's the `gcloud compute` commands that take a zone argument. The `gcloud dataproc` commands for working with an existing cluster only use the region.

I recently noticed that `hailctl dataproc diagnose` handles this by using `gcloud dataproc clusters describe` to obtain the zone for a cluster.
https://github.com/hail-is/hail/blob/71801b763ea2653fc6a6fcec812c812f57fe9a7e/hail/python/hailtop/hailctl/dataproc/diagnose.py#L39-L51

PR #9081, opened 2020-07-14
[query] lower TableUnion and TableMultiWayZipJoin

patrick-schultz:


hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -764,6 +764,56 @@ object LowerTableIR {
...
+            repartitioned.flatMap(_.letBindings),
+            repartitioned.flatMap(_.broadcastVals),
+            repartitioned.head.globals,
+            newPartitioner,
+            zipIR(repartitioned.map(_.contexts), ArrayZipBehavior.AssumeSameLength) { ctxRefs =>
+              MakeTuple.ordered(ctxRefs)
+            },
+            ctxRef =>
+              StreamMultiMerge(repartitioned.indices.map(i => repartitioned(i).partition(GetTupleElement(ctxRef, i))), keyType.fieldNames)
+          )
```
tpoterba, line 783: Nice. Love how terse these lowering rules are with the right nodes.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -764,6 +764,56 @@ object LowerTableIR {
...
 
+        case x@TableUnion(children) =>
+          val lowered = children.map(lower)
+          val keyType = x.typ.keyType
+          val newPartitioner = RVDPartitioner.generate(keyType, lowered.flatMap(_.partitioner.rangeBounds))
+          val repartitioned = lowered.map(_.repartitionNoShuffle(newPartitioner))
+
+          TableStage(
+            repartitioned.flatMap(_.letBindings),
+            repartitioned.flatMap(_.broadcastVals),
```
tpoterba, line 775: I think the globals from the tail children will be here, but should get pruned away.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -764,6 +764,56 @@ object LowerTableIR {
...
+          val keyType = x.typ.keyType
+          val newPartitioner = RVDPartitioner.generate(keyType, lowered.flatMap(_.partitioner.rangeBounds))
+          val repartitioned = lowered.map(_.repartitionNoShuffle(newPartitioner))
+          val newGlobals = MakeStruct(FastSeq(
+            globalName -> MakeArray(lowered.map(_.globals), TArray(lowered.head.globalType))))
+          val globalsRef = Ref(genUID(), newGlobals.typ)
+
+          val keyRef = Ref(genUID(), keyType)
+          val valsRef = Ref(genUID(), TArray(children.head.typ.rowType))
+          val projectedVals = ToArray(mapIR(ToStream(valsRef)) { elt =>
```
tpoterba, line 796: we may want to rewrite the StreamZipJoin semantics to avoid this in the future, if it's a performance bottleneck.

patrick-schultz: Yeah, I forgot I had wanted to try making `curVals` be a stream, in which case I don't think this has any overhead vs. projecting out the keys in the StreamZipJoin emit.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
patrick-schultz: Don't they have to be here, because the globals could be referenced in the tail children's bodies?

tpoterba: Possible. I always forget the semantics of globals and broadcast vals when I haven't looked at it in the last day or two.

PR #9084, opened 2020-07-15
[query] Add ImputeType aggregator

tpoterba:
Will use this to implement type imputation for TextTableReader in python.


hail/python/hail/expr/aggregators/__init__.py
```diff
@@ -35,5 +35,6 @@
     'group_by',
     'downsample',
     'array_agg',
-    '_prev_nonnull'
+    '_prev_nonnull',
+    '_impute_type'
```
chrisvittal, line 39: Nit: Why do we not use trailing commas here? It makes diffs like these smaller.

tpoterba: yeah, I usually to add/leave them, just messed up here

PR #9086, opened 2020-07-15
[query] Rewrite import_table type imputation in Python

tpoterba:


hail/python/test/hail/table/test_table.py
```diff
@@ -1383,7 +1383,7 @@ def test_can_process_wide_tables():
     for w in widths:
         print(f'working on width {w}')
         path = resource(f'width_scale_tests/{w}.tsv')
-        ht = hl.import_table(path, impute=True)
+        ht = hl.import_table(path, impute=False)
```
tpoterba, line 1386: these tests still pass, but they now take much longer with impute=True because we're evaluating IR proportional to the width. The test is kinda bad and should be improved to actually do something O(fields) like aggregating every field. Then we should fix performance problems (main one seems to be generating GBs of log statements because we're logging the full row type O(fields) times) and reenable impute.

PR #9089, opened 2020-07-16
[query] Add TableMapPartitions node

tpoterba:
To come next:
* Expose in Python, add more tests of Python functionality

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Compile.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Compile.scala
@@ -143,7 +143,7 @@ object CompileIterator {
     def loadAddress(): Long
   }
 
-  private trait TMPStepFunction extends AsmFunction3RegionLongIteratorRegionBoolean with StepFunctionBase
+  private trait TMPStepFunction extends AsmFunction3RegionLongIteratorJLongBoolean with StepFunctionBase
 
   private abstract class LongIteratorWrapper extends Iterator[java.lang.Long] {
     def step(): Boolean
@@ -182,8 +182,7 @@ object CompileIterator {
     val er = EmitRegion.default(stepF)
     val emitter = new Emit(ctx, stepFECB)
 
-    var ir = body
-    ir = LoweringPipeline.compileLowerer(true)(ctx, ir).asInstanceOf[IR].noSharing
+    val ir = LoweringPipeline.compileLowerer(true)(ctx, body).asInstanceOf[IR].noSharing
     TypeCheck(ir)
     InferPType(ir, Env.empty[PType])
     val returnType = ir.pType.asInstanceOf[PStream].elementType.asInstanceOf[PStruct].setRequired(true)
@@ -193,6 +192,7 @@ object CompileIterator {
     val elementAddress = stepF.genFieldThisRef[Long]("elementAddr")
 
     val didSetup = stepF.genFieldThisRef[Boolean]("didSetup")
+    stepF.cb.emitInit(didSetup := false)
 
     implicit val ecc: EmitStreamContext = EmitStreamContext(stepF)
     var source: Source[EmitCode] = null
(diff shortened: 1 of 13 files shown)


hail/src/main/scala/is/hail/expr/ir/Compile.scala
```diff
@@ -133,3 +134,130 @@ object CompileWithAggregators {
...
+
+    val fb = EmitFunctionBuilder.apply[F](ctx, "stream", argTypeInfo.toFastIndexedSeq, CodeParamType(BooleanInfo))
+    val stepF = fb.apply_method
+    val stepFECB = stepF.ecb
+
+    val er = EmitRegion.default(stepF)
+    val emitter = new Emit(ctx, stepFECB)
+
+    var ir = body
+    ir = LoweringPipeline.compileLowerer(true)(ctx, ir).asInstanceOf[IR].noSharing
```
patrick-schultz, line 186: Why the var? Can't this just be
```
    val ir = LoweringPipeline.compileLowerer(true)(ctx, body).asInstanceOf[IR].noSharing
```
?

hail/src/main/scala/is/hail/expr/ir/Compile.scala
```diff
@@ -133,3 +134,130 @@ object CompileWithAggregators {
...
+    val didSetup = stepF.genFieldThisRef[Boolean]("didSetup")
+
+    implicit val ecc: EmitStreamContext = EmitStreamContext(stepF)
+    var source: Source[EmitCode] = null
+
+    val pullLabel = CodeLabel()
+    val eosField = stepF.genFieldThisRef[Boolean]("eos")
+    val eosLabel = CodeLabel()
+
+    optStream.addSetup(Code(elementAddress := 0L, eosField := false))
```
patrick-schultz, line 204: I think it's clearer if you just do these after `source.setup` below.

hail/src/main/scala/is/hail/expr/ir/Compile.scala
```diff
@@ -133,3 +134,130 @@ object CompileWithAggregators {
...
+    ir = LoweringPipeline.compileLowerer(true)(ctx, ir).asInstanceOf[IR].noSharing
+    TypeCheck(ir)
+    InferPType(ir, Env.empty[PType])
+    val returnType = ir.pType.asInstanceOf[PStream].elementType.asInstanceOf[PStruct].setRequired(true)
+
+    val optStream = EmitStream.emit(emitter, ir, stepF, er.region, Env.empty, None);
+
+    val elementAddress = stepF.genFieldThisRef[Long]("elementAddr")
+
+    val didSetup = stepF.genFieldThisRef[Boolean]("didSetup")
```
patrick-schultz, line 195: This is never initialized. It's probably better not to rely on Java's default initialization behavior. Add `didSetup := false` to the class initializer?

hail/src/main/scala/is/hail/expr/ir/Interpretable.scala
```diff
@@ -1,11 +1,12 @@
 package is.hail.expr.ir
 
-import is.hail.types.virtual.TNDArray
+import is.hail.types.virtual.{TNDArray, TStream}
 
 object Interpretable {
   def apply(ir: IR): Boolean = {
     !ir.typ.isInstanceOf[TNDArray] &&
-      (ir match {
+      !ir.typ.isInstanceOf[TStream] &&
```
patrick-schultz, line 8: Why this? There are interpret rules for most of the stream nodes.

hail/src/main/scala/is/hail/expr/ir/Compile.scala
tpoterba: good catch, this is an artifact of now-removed control flow.

hail/src/main/scala/is/hail/expr/ir/Interpretable.scala
tpoterba: not sure, actually. will revert.

cseed:
Curious, did you benchmark this change, in particular the local/field business?

tpoterba:
no difference: https://gist.github.com/tpoterba/2d3c72db1eba30e601e7a1876a675cd4

tpoterba:
if there's an effect, I expect we'll see it when we do whole-stage codegen, which we can't benchmark now.

PR #9091, opened 2020-07-16
[query] Remove zip join

tpoterba:


hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -1174,26 +1173,15 @@ case class TableJoin(left: TableIR, right: TableIR, joinType: String, joinKey: I
       }
     }
 
-    val joinedRVD = if (joinType == "zip") {
-      val leftRVD = leftTV.rvd
-      val rightRVD = rightTV.rvd
-      leftRVD.orderedZipJoin(
```
patrick-schultz, line 1180: I think you can get rid of the (ordered)zipJoin methods on RVD, KeyedRVD, and OrderedRVIterator too.

tpoterba: oh, good call!

PR #9092, opened 2020-07-16
[hailctl dataproc] Add tests for hailctl dataproc connect

nawatts:


hail/python/hailtop/hailctl/dataproc/connect.py
```diff
@@ -16,6 +18,26 @@ def init_parser(parser):
     parser.add_argument('--dry-run', action='store_true', help="Print gcloud dataproc command, but don't run it.")
 
 
+def get_chrome_path():
```
nawatts, line 21: Moved this into a function for easier mocking.

hail/python/hailtop/hailctl/dataproc/connect.py
```diff
@@ -16,6 +18,26 @@ def init_parser(parser):
...
+    system = platform.system()
+
+    if system == 'Darwin':
+        return '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome'
+
+    elif system == 'Linux':
```
tpoterba, line 27: linter doesn't like this, prefers `if system == 'Linux'` since the `if` above always returns from the function.

I'm not sure I personally agree with this style, but we decided to follow the Python preferred syntax™

nawatts: Should pylint be added to [hail/python/dev-requirements.txt](https://github.com/hail-is/hail/blob/master/hail/python/dev-requirements.txt)? That might help ensure developers are using it in their environments/editors. Right now, I only see it specified in [docker/requirements.txt](https://github.com/hail-is/hail/blob/1039d1681d0963f2366ce7f0f5e2f7d96a0d4c9d/docker/requirements.txt#L30)

tpoterba: oh, good question. Yeah, almost certainly. I think pretty much everything in docker/requirements.txt should be in dev-requirements.txt, but there may be a few unnecessary libs.

It's also possible to run the linting locally with `make -C ${HAIL_HOME}/hail/python check-hail`

tpoterba: (linting is SO slow though, wtf?)

nawatts: After trying `make -C ${HAIL_HOME}/hail/python check-hail`, it looks like `flake8` should go in dev-requirements.txt as well.

Pylint can be sped up a bit by running in parallel. Adding `-j 0` to pylint took `make -C ${HAIL_HOME}/hail/python check-hail` from ~90s to ~40s on my machine.
http://pylint.pycqa.org/en/latest/user_guide/run.html#parallel-execution

johnc1231:
@nawatts , if you made the fixes Tim wanted, you should dismiss his review so he knows to review it again.

tpoterba:
oh, crap, sorry this dropped off my radar. Been juggling a lot recently.

johnc1231:
Feel free to reassign to me if you want Tim.

johnc1231:
@nawatts also remove the WIP label when you're ok with the PR getting merged. That label will prevent merging.

nawatts:
@tpoterba no problem, no rush.

@johnc1231 I looked the other day, but GitHub didn't show an option to dismiss Tim's review. I think that may have been because I re-requested a review after I made the changes. Also, I don't have the ability to edit labels.

PR #9093, opened 2020-07-17
[query] ndarray inverse using LAPACK

akotlar:
to flesh out our linear algebra, and in service of regenie. eventually, this, qr, and other linalg functions should be moved to hl.linalg I think, to match scipy, numpy, dask, etc

we should also decide whether to break out LU factorization. Numpy, scipy, dask do this, so that's my preference for a followup pr, because people who are looking for a fleshed out linalg experience may want this.

also, we should have less than double precision support, but this matches QR.

### DIFF
--- a/hail/src/test/scala/is/hail/expr/ir/IRSuite.scala
+++ b/hail/src/test/scala/is/hail/expr/ir/IRSuite.scala
@@ -2244,6 +2244,14 @@ class IRSuite extends HailSuite {
     assertEvalsTo(makeNDArrayRef(matMulCube, IndexedSeq(0, 0, 0)), 30.0)
   }
 
+  @Test def testNDArrayInv() {
+    implicit val execStrats: Set[ExecStrategy] = ExecStrategy.compileOnly
+    val matrixColMajor = makeNDArray(FastSeq(1.5, 2.0, 4.0, 5.0), FastSeq(2, 2), False())
+    val inv = NDArrayInv(matrixColMajor)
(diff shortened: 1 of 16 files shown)


hail/src/test/scala/is/hail/expr/ir/IRSuite.scala
```diff
@@ -2244,6 +2244,14 @@ class IRSuite extends HailSuite {
     assertEvalsTo(makeNDArrayRef(matMulCube, IndexedSeq(0, 0, 0)), 30.0)
   }
 
+  @Test def testNDArrayInv() {
+    implicit val execStrats: Set[ExecStrategy] = ExecStrategy.compileOnly
```
akotlar, line 2248: It's not clear to me if this is an oversight, but most of our NDArray tests are not actually tested, because execStrats are Set(); the values specified in these assertEvalsTo are arbitrary. See TestNDArrayMatMul as a worrisome example of this.

tpoterba: oof, good catch! That's consistent with my memory of the history here. Will happily (and a bit nervously...) review the PR to reenable most tests.

These nodes are also tested pretty well in Python, so this isn't as horrifying as it may appear at first.

akotlar:
it was fun, thanks Tim!

hail/src/test/scala/is/hail/expr/ir/IRSuite.scala
akotlar: Yeah, I verified that matmul for instance works perfectly enabled.

johnc1231:
Very cool!

PR #9094, opened 2020-07-17
[query] enable ndarray tests, support row/column major in MakeNDArray

akotlar:
Enabling NDArray tests revealed a bug relative to the test suite: MakeNDArray doesn't actually consider row/column major. I added a simple fix for this, predicated on rowMajorIR being a literal, which seems like a reasonable choice.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -831,7 +831,10 @@ class Emit[C](
                 })
             }
 
-            PCode(pt, xP.construct(shapeBuilder, xP.makeRowMajorStridesBuilder(shapeCodeSeq, mb), requiredData, mb))
+            rowMajorIR match {
+              case True() => PCode(pt, xP.construct(shapeBuilder, xP.makeRowMajorStridesBuilder(shapeCodeSeq, mb), requiredData, mb))
+              case False() => PCode(pt, xP.construct(shapeBuilder, xP.makeColumnMajorStridesBuilder(shapeCodeSeq, mb), requiredData, mb))
+            }
           }
         }
       case NDArrayRef(nd, idxs) =>
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -806,34 +806,30 @@ class Emit[C](
         val dataPType = xP.data.pType
         val nDims = shapePType.size
 
-        assert(rowMajorIR == True() || rowMajorIR == False())
-        emitI(shapeIR).flatMap(cb) { case shapeTupleCode: PBaseStructCode =>
-          emitI(dataIR).map(cb) { case dataCode: PIndexableCode =>
-            val shapeTupleValue = shapeTupleCode.memoize(cb, "make_ndarray_shape")
-            val dataValue = dataCode.memoize(cb, "make_ndarray_data")
-            val dataPtr = dataValue.get.tcode[Long]
-            val requiredData = dataPType.checkedConvertFrom(mb, region, dataPtr, coerce[PArray](dataContainer), "NDArray cannot have missing data")
+        emitI(rowMajorIR).flatMap(cb) { case isRowMajorCode: PPrimitiveCode =>
+          emitI(shapeIR).flatMap(cb) { case shapeTupleCode: PBaseStructCode =>
+            emitI(dataIR).map(cb) { case dataCode: PIndexableCode =>
+              val shapeTupleValue = shapeTupleCode.memoize(cb, "make_ndarray_shape")
+              val dataValue = dataCode.memoize(cb, "make_ndarray_data")
+              val dataPtr = dataValue.get.tcode[Long]
+              val requiredData = dataPType.checkedConvertFrom(mb, region, dataPtr, coerce[PArray](dataContainer), "NDArray cannot have missing data")
 
-            (0 until nDims).foreach { index =>
-              cb.ifx(shapeTupleValue.isFieldMissing(index),
-                cb.append(Code._fatal[Unit](s"shape missing at index $index")))
-            }
+              (0 until nDims).foreach { index =>
+                cb.ifx(shapeTupleValue.isFieldMissing(index),
+                  cb.append(Code._fatal[Unit](s"shape missing at index $index")))
+              }
 
-            val shapeCodeSeq = (0 until nDims).map(shapeTupleValue[Long](_).get)
+              def shapeBuilder(srvb: StagedRegionValueBuilder): Code[Unit] = {
+                Code(
+                  srvb.start(),
+                  Code.foreach(0 until nDims) { index =>
+                    Code(
+                      srvb.addLong(shapeTupleValue(index)),
+                      srvb.advance())
+                  })
+              }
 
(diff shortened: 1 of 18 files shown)


hail/src/test/scala/is/hail/expr/ir/IRSuite.scala
```diff
@@ -2206,8 +2210,8 @@ class IRSuite extends HailSuite {
     assertEvalsTo(makeNDArrayRef(colVectorWithMatrix, FastIndexedSeq(1, 0)), 2.0)
   }
 
-  @Test def testNDArrayAgg() {
-    implicit val execStrats: Set[ExecStrategy] = Set()
+  @Test(enabled = false) def testNDArrayAgg() {
```
akotlar, line 2213: there is no NDArrayAgg emitter

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -806,7 +806,7 @@ class Emit[C](
         val dataPType = xP.data.pType
         val nDims = shapePType.size
 
-        // val rowMajort = emit(rowMajorIR) // XXX Unused?
+        assert(rowMajorIR == True() || rowMajorIR == False())
```
tpoterba, line 809: This is the wrong fix. We should just make the IR have a boolean parameter, instead of taking an IR child that must be constant.

akotlar: I think the proposal is to instead pass a boolean literal in place of the IR, so in Parser.scala parsing boolean_literal, instead of ir_value_expr(). 

Could you help me understand why reading literal boolean IR isn't appropriate?

tpoterba: I'm saying this:
```
final case class MakeNDArray(data: IR, shape: IR, rowMajor: IR) extends NDArrayIR
```

should be this:
```
final case class MakeNDArray(data: IR, shape: IR, rowMajor: Boolean) extends NDArrayIR
```

Having an IR child that must be a constant value introduces unnecessary complexity.

It's also probably not too hard to fix the implementation to properly branch on the row major IR value in the staged code, this would be even better I think.

akotlar: > Having an IR child that must be a constant value introduces unnecessary complexity.

Right, so my question is: why do we have constant value IR?

akotlar: > It's also probably not too hard to fix the implementation to properly branch on the row major IR value in the staged code, this would be even better I think.

Yes I could do this, if we want arbitrary boolean expressions for rowMajorIR. I didn't see a use for this, but I haven't thought about it as much as you. Could you help me understand the use case?

tpoterba: > Right, so my question is: why do we have constant value IR?

I think it would be really hard to write any programming/query language without constants. How would you represent `thing > 5` without a representation of the constant `5`?

> if we want arbitrary boolean expressions for rowMajorIR

John has the most context here, but it looks to me that while we don't really take advantage of the programmable row major / col major behavior, the structure of the node suggests that it is possible (though the emitter doesn't have these semantics and is therefore broken). Changing rowMajor to a Boolean rather than IR parameter would change the structure of the node to indicate that rowMajor is chosen at compile-time but not runtime. Fixing the emitter to properly support runtime branching is probably easier (a more local change) and brings the behavior back in line with what the structure suggests.

akotlar: > I think it would be really hard to write any programming/query language without constants. How would you represent `thing > 5` without a representation of the constant `5`?

A serialized 5. We wouldn't use a Five() IR node. Why do we use a True() IR node?

> Changing rowMajor to a Boolean rather than IR parameter would change the structure of the node to indicate that rowMajor is chosen at compile-time but not runtime.

How is "true" different than "True()"? We can enforce that rowMajorIR is True() or False() without using "true" and "false", by adding a single assert(), which seems like the simplest change (this could be added to the Parser branch).

tpoterba: > A serialized 5 that is parsed as an integer literal. We wouldn't use a Five() IR node. Why do we use a True() IR node?

Having a True() and False() node is similar to having a `Bool(x: Boolean)` node, but is slightly terser to pattern match on, like here: https://github.com/hail-is/hail/blob/6444d3fe7b6b0e36dbd1e25a4c88caa0e6e7dc61/hail/src/main/scala/is/hail/expr/ir/Simplify.scala#L156-L158

> How is "true" different than "True()"? We can enforce that rowMajorIR is True() or False() without using "true" and "false", by adding a single assert(), which seems like the simplest change (this could be added to the Parser branch).

I disagree that this is the simplest change. It might be a small diff, but it changes the implied semantics of the node based on the signature. It breaks from conventions with the rest of the IR (that compile-time flags are Boolean parameters on IR nodes, not constant-valued IRs), making the IR as a whole more complex.

I think you can fix the behavior to branch on the runtime value of rowMajor with a ~5 line diff, which seems like the best thing to do.

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
```diff
@@ -107,6 +111,15 @@ final case class PCanonicalNDArray(elementType: PType, nDims: Int, required: Boo
     )
   }
 
+  def makeStridesBuilder(sourceShape: PBaseStructValue, isRowMajor: Code[Boolean], mb: EmitMethodBuilder[_]): StagedRegionValueBuilder => Code[Unit] = { srvb =>
+    val shapeCodeSeq1 = (0 until nDims).map(sourceShape[Long](_).get)
+    val shapeCodeSeq2 = (0 until nDims).map(sourceShape[Long](_).get)
```
akotlar, line 116: Is there a less awkward way of using a Code in two branches? Memoization seems not right, since value is not actually consumed twice. Store is not branch aware.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
akotlar: > It breaks from conventions with the rest of the IR (that compile-time flags are Boolean parameters on IR nodes, not constant-valued IRs)

I don't understand the difference between a True() and a compile-time flag. This is a literal IR with exactly 1 possible value. You can call that a compile-time flag.

tpoterba: I agree that what you've done is changed the semantics of `rowMajor` to be a compile-time flag, but the problem is that it's now inconsistent with the way we represent compile-time flags in the rest of the IR.

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
tpoterba: can just make it a def

tpoterba: I also think this should just be defined inline in the makeNDArray emitter, we shouldn't need to change this file.

akotlar:
> I also think this should just be defined inline in the makeNDArray emitter, we shouldn't need to change this file

Good point that's simpler

PR #9095, opened 2020-07-17
[batch] cache input files

jigold:
Based on #9076. This PR caches input files at `/cache` on the worker. I do not support wildcard characters in filenames. I had to change the Flock implementation to only lock the directories and not a file if a file name is given. This is because we don't know if a user means a file or directory a priori. For example, `gs://jigold/test`. Is that a directory test or a file test? I figured more coarse-grained locking was fine.


jigold:
It looks like my cache change is passing tests now. I'd like for you to take a look before I confirm one last time that the cache is actually working by submitting jobs downloading a 512 MB file and making sure the timings of the non-first job is a couple of seconds. It looks like the tests got a bit slower. I'm not sure if that's because of the docker image having gsutil in it. I don't see how the extra copying infrastructure would make a huge difference.

danking:
looking

batch/batch/worker/copy.py
```diff
@@ -0,0 +1,83 @@
...
+    src_tail_dir = src_path.rstrip('/').split('/')[-1]
+    rsync_dst = f'{dst.rstrip("/")}/{src_tail_dir}/'
+
+    with Flock(cache_src):
+        os.makedirs(os.path.dirname(dst), exist_ok=True)
+        os.makedirs(os.path.dirname(cache_src), exist_ok=True)
+        check_call(f'cp -p -R --reflink {shq(cache_src)} {shq(dst)} || true')
+        check_call(f'''
+gsutil -q stat {shq(src)}
+if [ $? = 0 ]; then
```
danking, line 63: stat returns zero if there is a file whose name ends in `/`. There's an easy fix: if `src` ends in `/`, don't both calling stat and checking the output. Just `check_call(f'''mkdir -p ...)`

batch/batch/worker/flock.py
```diff
@@ -7,18 +7,21 @@
...
-    def __init__(self, path, pool=None):
+    def __init__(self, path, pool=None, nonblock=False):
         self.path = os.path.abspath(path)
         self.pool = pool
 
         self.flock_flags = fcntl.LOCK_EX
         self.fds = []
 
+        if nonblock:
+            self.flock_flags |= fcntl.LOCK_NB
```
danking, line 18: I think this is a hold-over from before my comments to the underlying PR.

batch/batch/worker/copy.py
```diff
@@ -0,0 +1,83 @@
...
+        raise NotImplementedError(f'glob wildcards are not allowed in file paths, got destination {dst}')
+
+    if not src.startswith('gs://'):
+        raise NotImplementedError(f'cannot copy from sources that are not GCS paths, got source {src}')
+
+    cache_src = f'{cache_path}/{user}{src[4:]}'
+    escaped_src = re.escape(os.path.basename(src))
+
+    if dst.startswith('/io'):
+        dst = f'/host{io_path}{dst[3:]}'
```
danking, line 48: When does the destination not start with /io?

batch/batch/worker/copy.py
```diff
@@ -0,0 +1,83 @@
...
+fi
+''')
+        check_call(f'rm -Rf {shq(cache_src)}')
+        check_call(f'cp -p -R --reflink {shq(dst)} {shq(cache_src)}')
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--user', type=str, required=True)
+    parser.add_argument('--io-path', type=str, required=True)
```
danking, line 77: I prefer the explicitness of `io-host-path`

batch/batch/worker/worker.py
```diff
@@ -660,8 +679,16 @@ def __init__(self, batch_id, user, gsa_key, job_spec, format_version):
         containers = {}
 
         if input_files:
+            input_mounts = [{
+                'Source': LOCAL_SSD_MOUNT,
+                'Target': '/host',
+                'Type': 'bind',
+                'BindOptions': {
+                    'Propogation': 'shared'  # FIXME: look up if needed
```
danking, line 687: The questions are: 
- should Docker expose devices mounted inside LOCAL_SSD_MOUNT to the container
- should Docker expose devices mounted inside `/host` to the host OS
- should either of the above happen recursively, e.g. I mount something at LOCAL_SSD_MOUNT/foo and then I mount another thing at LOCAL_SSD_MOUNT/foo/foo2.

batch/batch/worker/worker.py
```diff
@@ -856,12 +884,56 @@ def __str__(self):
...
+                    if await self.used_disk_space() < 0.7:
+                        break
+
+                    file_path = f'{root}/{file}'
+                    try:
+                        async with Flock(file_path, pool=worker.pool, nonblock=True):
+                            await self.remove(file_path)
+                    except BlockingIOError:
+                        log.exception(f'could not remove in-use file {file_path}')
+            self.cleanup.clear()
```
danking, line 927: use `retry_long_running` from hailtop.utils, see examples in batch front-end.

batch/test/test_dag.py
```diff
@@ -174,6 +174,24 @@ def test_no_parents_allowed_in_other_batches(client):
...
+                                          ('/io/data2', f'gs://{bucket_name}')])
+    tail = batch.create_job('ubuntu:18.04',
+                            command=['/bin/sh', '-c', 'cat /io/data1; cat /io/data2'],
+                            input_files=[(f'gs://{bucket_name}/data1', '/io/'),
+                                         (f'gs://{bucket_name}/data2', '/io/')],
+                            parents=[head])
+    batch.submit()
+    tail.wait()
+    assert head._get_exit_code(head.status(), 'main') == 0, head._status
+    assert tail.log()['main'] == 'head1\nhead2\n', tail.status()
```
danking, line 191: can we also include tail.log() in the assertion message? `, (tail.log()['main'], tail.status())`

batch/batch/worker/flock.py
jigold: I need this so that the garbage collector doesn't wait on a file to be downloaded if someone else has a lock on that file. I want it to just give up on that and try locking / deleting the next file.

batch/batch/worker/copy.py
```diff
@@ -0,0 +1,83 @@
...
+
+from .flock import Flock
+
+
+GCS_PATH_REGEX = re.compile('gs://(?P<bucket>[^/]+)/(?P<path>.+)')
+
+
+def check_call(cmd):
+    print(cmd)
+    return sp.check_call(cmd, shell=True, stderr=sp.STDOUT)
```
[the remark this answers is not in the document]
jigold: All of your commands in the tls code give lists to sp.check_call. Can I replace them with a script and f-strings?

jigold:
@danking I think I addressed most of your comments. Everything is passing. Can you look if you like the interface for `sync_check_call` etc.? If so, then can I translate your TLS commands in lists into a single string? Is there ever a reason to want to pass the list explicitly in this case?

Once you're happy with the changes, then I'll do another test round with dev deploy to make sure the cache is actually doing its thing and the garbage collection loop is working.

jigold:
Actually, I realize there's still a bit of garbage in there. I'll clean that up.

danking:
@jigold remove WIP when you're ready.

jigold:
Closing while I dev deploy.

PR #9105, opened 2020-07-20
[hail] NDArray eye

akotlar:
Identity matrix, needed for ridge regression in Hail. This is obviously suboptimal, it should be a special instance of a sparse ndarray that specifies 'k', but this matches the Numpy implementation and represents a good start.

### DIFF
--- a/hail/python/hail/nd/nd.py
+++ b/hail/python/hail/nd/nd.py
@@ -338,7 +338,8 @@ def eye(N, M=None, dtype=hl.tfloat64):
 
     See Also
     --------
-    diag : Gets the diagonal of a 2 dimensional NDArray
+    :func:`.identity`
+    :func:`.diagonal`
 
     Examples
     --------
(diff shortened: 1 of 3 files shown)


akotlar:
Good point John. I had omitted "identity" since we didn't have an identity function, but it's worth adding.

johnc1231:
I'm back, feel free to assign me to some of these ndarray things. (Also let me know issues you run into when adding stuff)

hail/python/hail/nd/nd.py
```diff
@@ -314,3 +314,84 @@ def concatenate(nds, axis=0):
...
+      An ndarray whose elements are equal to one on the main diagonal, zeroes elsewhere.
+
+    See Also
+    --------
+    :func:`.identity`
+    :func:`.diagonal`
+
+    Examples
+    --------
+    >>> hl.eval(hl.nd.eye(2, dtype=hl.tint32))
```
johnc1231, line 346: I think this should have one example where you pass `M`, otherwise I think we are good to merge.

PR #9106, opened 2020-07-20
[query][smm 1] scaffolding for stream memory management

patrick-schultz:
This PR introduces a second region `eltRegion` into the `emitStream` environment, and plumbs it through everywhere it is needed. One complication is that the `eltRegion` for a stream bound in the environment (e.g. in a `StreamMap` over a `Stream[Stream[A]]`) must be set by the consumer of the stream, so the `PCanonicalStreamCode` stored in the environment needs to be a function `Code[Region] => SizedStream`.

For the most part, this PR leaves the `eltRegion`s unused. In the new semantics, explained below, it is still correct for a producer to only use the `outerRegion`, and for a consumer to pass the same region for both `outerRegion` and `eltRegion`. This lets us migrate to the new memory management one node at a time. When I talk about correctness, I mean that: no pointer is used after the region it points into has been freed; and, every region which is created is eventually freed. When reasoning about correctness, I consider `r.clear()` equivalent to `r.invalidate(); r.getNewRegion()`.

As a demonstration of how this should work, this PR does convert `EmitStream.{toArray, write}` to use `eltRegion` correctly, which are used in the consumer nodes `ToArray`, `ArraySort`, `ToSet`, `ToDict`, and `GroupByKey`. Going forward, the plan is to convert the rest of the consumer and producer nodes, before converting the transformer nodes (which both consume and produce streams), as correctness can only be tested on pipelines in which all nodes have been converted.

### Semantics

A stream producer is passed two regions from its consumer: `outerRegion` and `eltRegion`. The node being emitted does not own either region, so may not free/clear them. The only thing a producer may do with either region is to put data in them, by writing directly to them or by giving/sharing ownership of a producer-owned region to them.

Consumers' contract:
* The lifetime of `outerRegion` must contain the lifetime of the producer stream, i.e. `outerRegion` must be valid before the producer's `setup0` is called, and must still be valid when the producer's `close0` is called. Thus a producer may use `outerRegion` to store state that persists between elements, or it may use a region it owns itself.
* When the producer's `pull` is called, `eltRegion` must be valid (it does not need to be valid for setup/close).

Producers' contract:
* When the consumer's `push` is called, the pushed value must be owned by either `outerRegion` or `eltRegion` (really `eltRegion`; using `outerRegion` is correct but unnecessarily extends the lifetime of the value). This can be accomplished by writing directly to the consumer's region, or by giving/sharing ownership of a producer-owned region to it. Thus the consumer may assume the pushed value is valid until it frees/clears `eltRegion`. If the consumer needs the value to live longer, it must deep-copy the value to another region.

These allow us to reason about (and, in principle, to prove) the correctness of each node independently.


tpoterba:
high level comment -- this is an infrastructure change that will hurt performance. We need to know by how much, and probably add benchmarks that target the specific cases where the added work will hurt most.

If things get noticeably slower, we're probably going to need to rework streams so that per-row stream code doesn't incur memory management overhead.

patrick-schultz:
We definitely need a mechanism to force a stream pipeline (or sub-pipeline) to put all allocations in a single region, and avoid any region management overhead. Then, for example, in table lowering we can set a flag on any single-row stream processing to use a single region, preserving the existing behavior.

I have some thoughts on how to do that. We can just pass an "allocator" to EmitStream, which is a Region factory, that stream nodes must use to create new regions. An allocator that creates new regions gives the "free between rows" behavior. To implement the "within one row" behavior, we can pass down an allocator that returns regions backed by a single fixed RegionMemory, with no-op freeing. Maybe that needs to be put in place before this can merge?

tpoterba:
>  can pass down an allocator that returns regions backed by a single fixed RegionMemory, with no-op freeing

Even this may be expensive when we're doing something like ToArray(StreamRange). Let's start by benchmarking and seeing where we're at with the current benchmarks?

patrick-schultz:
> Even this may be expensive

This would all be staged. The generated code would just directly allocate into the provided region, no overhead. But I agree we should benchmark first.

tpoterba:
I'm more worried about the no-op free, I think.

tpoterba:
If it's just decrementing a reference count, might be OK.

patrick-schultz:
> I'm more worried about the no-op free, I think.

Also staged, so no-op means no code generated.

patrick-schultz:
I've updated this to allow stream consumers to choose an allocation strategy independently of the producer: the producer is written to allocate temporary regions that it owns, as needed, but the consumer can override that behavior to actually put all temporary allocations in a single region.

This is implemented via the two abstract interfaces `StagedRegion` and `StagedOwnedRegion`, which have two implementations, one of which can create temporary regions, and one which cannot. The producer is written using the interface, and the consumer passes in one of the concrete implementations, determining the allocation strategy.

Currently, only the non-allocating instance is ever used, which generates the same code we did before, so this should have zero run-time effect, even if we continue to update more stream nodes to use smarter region management. We can figure out separately how to introduce the other strategy (use less memory, but with more allocator overhead) in a controlled way.

patrick-schultz:
Weird, I can't replicate the `testShuffleIR` failure any more.

tpoterba:
yeah, there's a non-deterministic failure, I think.

patrick-schultz:
The diff in `deforestNDArray` is terrible. I just changed
```
  // in Emit[C].emit()
  def emitDeforestedNDArray(ir: IR): EmitCode =
    deforestNDArray(ir, mb, region, env).emit(mb, coerce[PNDArray](ir.pType))
  
  // in Emit[C]
  def deforestNDArray(x: IR, mb: EmitMethodBuilder[C], region: Value[Region], env: E): NDArrayEmitter[C] = {
    def deforest(nd: IR): NDArrayEmitter[C] = deforestNDArray(nd, mb, region, env)
    x match { ... }
  }
  
  // in NDArrayEmitter[C]
  def emit(mb: EmitMethodBuilder[C], targetType: PNDArray): EmitCode
```
to
```
  // in Emit[C].emit()
  def emitDeforestedNDArray(ir: IR): EmitCode =
    deforestNDArray(ir, mb, region, env)
    
  // in Emit[C]
  def deforestNDArray(x0: IR, mb: EmitMethodBuilder[C], region: StagedRegion, env: E): EmitCode = {
    def deforest(x: IR): NDArrayEmitter[C] = {
      x match { ... }
    }
    deforest(x0).emit(mb, coerce[PNDArray](x0.pType), region.code)
  }
  
  // in NDArrayEmitter[C]
  def emit(mb: EmitMethodBuilder[C], targetType: PNDArray, region: Value[Region])
```
Before, `NDArrayEmitter.emit` was always building the NDArray in the region argument to `mb`, no matter what region was used in `deforestNDArray`. This passes the region into `NDArrayEmitter.emit`, and refactors things to better enforce that the same region is used in both parts.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -2415,393 +2438,395 @@ class Emit[C](
       PCode(pt, value))
   }
 
-  def deforestNDArray(x: IR, mb: EmitMethodBuilder[C], region: Value[Region], env: E): NDArrayEmitter[C] = {
+  def deforestNDArray(x0: IR, mb: EmitMethodBuilder[C], region: StagedRegion, env: E): EmitCode = {
     def emit(ir: IR, env: E = env): EmitCode =
       this.emitWithRegion(ir, mb, region, env, None)
 
-    def deforest(nd: IR): NDArrayEmitter[C] = deforestNDArray(nd, mb, region, env)
```
tpoterba, line 2422: why this change?

hail/src/main/scala/is/hail/expr/ir/StagedRegion.scala
```diff
@@ -0,0 +1,74 @@
...
+  def shareWithParent(): Code[Unit]
+
+  def copyToParent(mb: EmitMethodBuilder[_], value: PCode): PCode
+}
+
+object StagedRegion {
+  def apply(r: Value[Region], allowSubregions: Boolean = false): StagedRegion =
+    if (allowSubregions) new RealStagedRegion(r) else new DummyStagedRegion(r)
+
+  class RealStagedRegion(r: Value[Region]) extends StagedRegion {
```
tpoterba, line 32: Prefer not to define this in the object, the compiled class's name gets mangled and this has been annoying in the past.

hail/src/main/scala/is/hail/utils/richUtils/RichCodeRegion.scala
```diff
@@ -4,9 +4,11 @@ import is.hail.annotations.Region
 import is.hail.asm4s._
 
 class RichCodeRegion(val region: Code[Region]) extends AnyVal {
-  def allocate(alignment: Code[Long], n: Code[Long]): Code[Long] = {
+  def allocate(n: Code[Long]): Code[Long] =
```
tpoterba, line 7: Where is this used?

hail/src/main/scala/is/hail/expr/ir/Emit.scala
patrick-schultz: Does my last comment (https://github.com/hail-is/hail/pull/9106#issuecomment-665034873) answer this? If not I can go into more detail.

hail/src/main/scala/is/hail/expr/ir/StagedRegion.scala
patrick-schultz: Ah, sure.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
tpoterba: oh, sorry, I totally missed that!

hail/src/main/scala/is/hail/utils/richUtils/RichCodeRegion.scala
patrick-schultz: Looks like this was left from an earlier version. Will remove.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -1436,16 +1451,17 @@ class Emit[C](
 
             val codeB = emit(body, env = bodyenv)
             Code(xElt := elt,
-              tmpAcc := codeB.map(v => accType.copyFromPValue(mb, region, PCode(body.pType, codeB.v))),
+              tmpAcc := codeB.map(v => accType.copyFromPValue(mb,
+                                                              region.code, PCode(body.pType, codeB.v))),
```
tpoterba, line 1455: is this generated automatically from your IDE? I think our typical style settings would break params onto the next line + 1 indent, not at the same level of indentation.

patrick-schultz: Oops, yeah. I tried to fix them up after the refactoring, but missed this one.

hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
```diff
@@ -806,53 +806,84 @@ object EmitStream {
...
-            ss.stream.forEach(mb, { et =>
-              Code(
+            ss.stream(eltRegion).forEach(mb, { et =>
+              Code(FastSeq(
                 et.setup,
-                et.m.mux(srvb.setMissing(), srvb.addIRIntermediate(aTyp.elementType)(et.v)),
-                srvb.advance())
+                et.m.mux(
+                  srvb.setMissing(),
+                  srvb.addIRIntermediate(et.pv, deepCopy = true)),
```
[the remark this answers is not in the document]
tpoterba: er, I thought I actually left this comment on the earlier version, this is the ToArray thing you fixed.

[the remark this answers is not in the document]
tpoterba: ignore

PR #9108, opened 2020-07-20
[query] Lower TableFilterIntervals

tpoterba:


hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -514,6 +514,96 @@ object LowerTableIR {
...
+        case TableFilterIntervals(child, intervals, keep) =>
+          val loweredChild = lower(child)
+          val part = loweredChild.partitioner
+          val kt = child.typ.keyType
+          val ord = PartitionBoundOrdering(kt)
+          val iord = ord.intervalEndpointOrdering
+          val nPartitions = part.numPartitions
+
+
+          val filterPartitioner = new RVDPartitioner(kt, Interval.union(intervals.toArray, ord.intervalEndpointOrdering), 0)
```
patrick-schultz, line 526: I don't think you want `allowedOverlap = 0`. That's the strongest restriction, so if `intervals` is `[[{0, 1}, {2, 5}], [{2, 7}, {3, 0}]]`, this would throw an error because the key prefix `2` is split across partitions. I think you want no restriction here, which means you can just leave out `allowedOverlap` (the default is `kt.size`).

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -514,6 +514,96 @@ object LowerTableIR {
...
+          } else {
+            // keep = False
+            val (newRangeBounds, includedIndices, startAndEndInterval) = part.rangeBounds.zipWithIndex.flatMap { case (interval, i) =>
+              val lb = filterPartitioner.lowerBoundInterval(interval)
+              val ub = filterPartitioner.upperBoundInterval(interval)
+              if ((lb until ub).map(filterPartitioner.rangeBounds).exists { filterInterval =>
+                iord.compareNonnull(filterInterval.left, interval.left) <= 0 && iord.compareNonnull(filterInterval.right, interval.right) >= 0
+              })
+                None
+              else Some((interval, i, (filterPartitioner.lowerBoundInterval(interval).min(nPartitions), filterPartitioner.upperBoundInterval(interval).min(nPartitions))))
```
patrick-schultz, line 560: Already computed these as `lb, ub`.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -514,6 +514,96 @@ object LowerTableIR {
...
+                    key)
+                }
+            }
+            (newRangeBounds, includedIndices, startAndEndInterval, f)
+          }
+
+          val newPart = new RVDPartitioner(kt, newRangeBounds, part.allowedOverlap)
+
+          TableStage(
+            letBindings = loweredChild.letBindings ++ FastIndexedSeq((filterIntervalsRef.name, Literal(boundsType, filterIntervals))),
```
patrick-schultz, line 579: Not a big deal, but `filterIntervalsRef` doesn't need to be in `letBindings`, since it's not used in `contexts` or other `letBindings`. Can just move the definition to `broadcastVals`.

hail/src/main/scala/is/hail/rvd/RVDPartitioner.scala
```diff
@@ -14,7 +14,7 @@ class RVDPartitioner(
   // rangeBounds: Array[Interval[kType]]
   // rangeBounds is interval containing all keys within a partition
   val rangeBounds: Array[Interval],
-  allowedOverlap: Int
+  val allowedOverlap: Int
```
patrick-schultz, line 17: This really shouldn't be considered part of the data of a partitioner. It's just a constructor argument to request a stronger precondition enforced. The reason it shouldn't be part of the data is that a given `rangeBounds` satisfies many different `allowedOverlap`, because smaller values imply all larger values, with `allowedOverlap = kType.size` always satisfied. The `allowedOverlap` that was enforced during construction is irrelevant to the behavior of the partitioner.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -514,6 +514,96 @@ object LowerTableIR {
...
+                  acc && !invoke("partitionIntervalContains",
+                    TBoolean,
+                    elt,
+                    key)
+                }
+            }
+            (newRangeBounds, includedIndices, startAndEndInterval, f)
+          }
+
+          val newPart = new RVDPartitioner(kt, newRangeBounds, part.allowedOverlap)
```
patrick-schultz, line 576: As mentioned there, there's no need to know what `allowedOverlap` `part` was constructed with. Here you don't want to enforce any extra checks, so just leave off the `allowedOverlap` argument.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
tpoterba: Interval.union returns a sorted array of intervals that have this property, this should be OK

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
tpoterba: whoops

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
tpoterba: thanks, worth fixing.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
tpoterba: ah, I see. Thanks!

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
patrick-schultz: That doesn't sound right. In my example
```
[[{0, 1}, {2, 5}], [{2, 7}, {3, 0}]]
```
union wouldn't change anything, because the two intervals don't overlap, right? But the prefix `{2}` does overlap, so the partitioner doesn't satisfy `allowedOverlap = 0`. I think the result of union will always satisfy `allowedOverlap = kt.size - 1` (equivalently, `partitionKey = kt.fieldNames`), if you want to enforce that.

tpoterba: ah... I see. Thanks for explaining.

PR #9114, opened 2020-07-22
[query] Fix simplify rule for nested TableFilterIntervals

tpoterba:


johnc1231:
Would you mind explaining what was wrong / context?

hail/src/main/scala/is/hail/expr/ir/Simplify.scala
```diff
@@ -769,7 +770,7 @@ object Simplify {
     case TableFilterIntervals(TableAggregateByKey(child, expr), intervals, keep) =>
       TableAggregateByKey(TableFilterIntervals(child, intervals, keep), expr)
     case TableFilterIntervals(TableFilterIntervals(child, i1, keep1), i2, keep2) if keep1 == keep2 =>
-      val ord = child.typ.keyType.ordering.intervalEndpointOrdering
+      val ord = PartitionBoundOrdering(child.typ.keyType).intervalEndpointOrdering
```
tpoterba, line 773: Somewhat recently we refactored things so that the struct prefix ordering (where you can compare structs that may have different sizes) is separate from the IR struct ordering (assert same length). The former is now called PartitionBoundOrdering, and it's used with partition intervals (also filter_intervals). This is a bug because if you tried to filter a table with a length-2 key with length-1 intervals, you'd hit that assertion.

PR #9116, opened 2020-07-22
[query] ndarray hstack

akotlar:


danking:
A reminder for next time, due to weird GitHub behavior, if you have a one commit PR, then the commit's title and message become the commit in master. As reviewers, we should enforce that one commit PRs have meaningful titles and messages in said commit.

PR #9117, opened 2020-07-22
[query] ndarrays from Python tuples

akotlar:
Numpy both supports this and prominently uses it in documentation. We should support this for people translating Numpy code to Hail code. Note that I don't extend this feature to Hail TupleExpressions, because Hail tuple types are distinct from Hail array types (in that the latter has a single element_type, and the former has a collection of possible types). Python tuples are distinct from lists only in their immutability.

Also, we support tuples in concatenate/NDArrayConcat, which makes the constructor difference from Numpy extra confusing.

Example Numpy use: 
https://numpy.org/doc/stable/reference/generated/numpy.hstack.html?highlight=hstack#numpy.hstack
`a = np.array((1,2,3))` (this is the first example, which would fail if translated to hail)
https://numpy.org/doc/stable/reference/generated/numpy.array.html#numpy.array
`x = np.array([(1,2),(3,4)])`


hail/python/test/hail/linalg/test_linalg.py
```diff
@@ -1028,6 +1028,32 @@ def assert_same_columns_up_to_sign(a, b):
         s = x.svd(compute_uv=False, complexity_bound=0)
         assert np.all(s >= 0)
 
+    def test_ndarray(self):
```
johnc1231, line 1031: I'd suggest we put this in `test_nd.py` with the rest of the other ndarray tests

johnc1231:
@akotlar, I just realized now you've been putting these tests in `test_linalg`. We have a file called `test_nd` where all the ndarray tests are. We should probably be putting them there instead.

akotlar:
Cool, I'll move them over.

Issue #9118, opened 2020-07-22
Hail MatrixTable show shows the wrong column keys


danking:
Wrong output for `.show()`, note it the column header should be `10.x`
```
In [5]: mt = hl.utils.range_matrix_table(1,1) 
   ...: mt = mt.annotate_entries(x = 1) 
   ...: mt = mt.key_cols_by(col_idx = mt.col_idx + 10) 
   ...: mt.show()                                                                                                                                                                                                                             
+---------+-------+
| row_idx |   0.x |
+---------+-------+
|   int32 | int32 |
+---------+-------+
|       0 |     1 |
+---------+-------+
```
Compare to correct output for `cols().show()`:
```
In [6]: mt = hl.utils.range_matrix_table(1,1) 
   ...: mt = mt.annotate_entries(x = 1) 
   ...: mt = mt.key_cols_by(col_idx = mt.col_idx + 10) 
   ...: mt.cols().show()                                                                                                                                                                                                                      
2020-07-22 13:55:28 Hail: WARN: cols(): Resulting column table is sorted by 'col_key'.
    To preserve matrix table column order, first unkey columns with 'key_cols_by()'
+---------+
| col_idx |
+---------+
|   int32 |
+---------+
|      10 |
+---------+
```

tpoterba:
this appears to be something you did intentionally?

```python
        if len(col_key_type) == 1 and col_key_type[0] == hl.tstr:
            col_key_field_name = list(col_key_type)[0]
            cols = t.cols.collect()
            entries = {cols[0][i][col_key_field_name]: t.entries[i]
                       for i in range(0, displayed_n_cols)}
        else:
            entries = {str(i): t.entries[i] for i in range(0, displayed_n_cols)}
```

Key is an int, so it's printing the index.

We can fix by adding integers to the type match, maybe?

danking:
ohhhhhh. Hmm. Yeah, I think any primitive type should be fine, right? I just wanted to have good behavior in the case that there was exactly one column key. The index thing is pretty confusing actually... hmm

PR #9119, opened 2020-07-22
[hail] Add an example for full_outer_join_mt

danking:
I took this example directly from an interaction with Akhil on Zulip.


danking:
The SKIP OUTPUT CHECK is because of a bug: https://github.com/hail-is/hail/issues/9118

PR #9122, opened 2020-07-22
[notebook][hailtop][gear] move logging configuration out of gear

danking:
`gear` should *only* be used in services, *never* in developer or user facing
tools. This change fixes `scale-test.py` to not use `gear`.

I fixed some use of `\` to use `( ... )`.
I verified that I didn't add any new required packages to hail top.


danking:
bump

PR #9127, opened 2020-07-23
[docker] make pip fast

danking:
Old versions of pip could only download some packages in source form which
requires compiling them. New version of pip can download these packages in
binary form which requires no compilation. This *substantially* improves docker
build times when you have to run `pip` in any layer.

### DIFF
--- a/docker/Dockerfile.base
+++ b/docker/Dockerfile.base
@@ -39,7 +39,7 @@ ENV PATH "$PATH:$SPARK_HOME/sbin:$SPARK_HOME/bin"
 
 COPY docker/requirements.txt .
 # re: pip install --upgrade pip: https://github.com/grpc/grpc/issues/22815
-RUN python3 -m pip install --upgrade pip keyrings.alt && \
+RUN python3 -m pip install --upgrade pip && \
     python3 -m pip install --no-cache-dir --upgrade -r requirements.txt
 
 ENV PYTHONPATH "${PYTHONPATH:+${PYTHONPATH}:}$SPARK_HOME/python:$SPARK_HOME/python/lib/py4j-0.10.7-src.zip"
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.base
```diff
@@ -38,7 +38,9 @@ ENV SPARK_HOME /spark-2.4.0-bin-hadoop2.7
 ENV PATH "$PATH:$SPARK_HOME/sbin:$SPARK_HOME/bin"
 
 COPY docker/requirements.txt .
-RUN python3 -m pip install --no-cache-dir -U -r requirements.txt
+# re: pip install --upgrade pip: https://github.com/grpc/grpc/issues/22815
+RUN python3 -m pip install --upgrade pip keyrings.alt && \
+    python3 -m pip install --no-cache-dir --upgrade -r requirements.txt
```
jigold, line 43: I think there's also something in the Dockerfile.services-base file and the batch/Dockerfile.worker file.

docker/Dockerfile.base
```diff
@@ -38,7 +38,9 @@ ENV SPARK_HOME /spark-2.4.0-bin-hadoop2.7
 ENV PATH "$PATH:$SPARK_HOME/sbin:$SPARK_HOME/bin"
 
 COPY docker/requirements.txt .
-RUN python3 -m pip install --no-cache-dir -U -r requirements.txt
+# re: pip install --upgrade pip: https://github.com/grpc/grpc/issues/22815
+RUN python3 -m pip install --upgrade pip keyrings.alt && \
```
jigold, line 42: What is the keyrings.alt for?

docker/Dockerfile.base
danking: I added the pip upgrade to Dockerfile.worker. It doesn't need to be added to services-base because that image uses base. It will already have an up-to-date version of pip.

docker/Dockerfile.base
danking: dunno, removed.

PR #9129, opened 2020-07-23
[address] implement the address service

danking:
This is a name to IP address and port service. GKE exposes pod IPs onto our VDC
network. As such, regular Google Cloud VMs can access pods by IP. GKE cannot
expose our services as IPs on our VDC because the way services load balance
traffic is more complex than DNS can handle. We acknowledge and accept the
limitations of client-side load balancing. In particular, if there are not many
clients and clients re-use address-port-pairs traffic will likely be
unbalanced. This is not a problem for the planned Shuffle service because the
clients are intended to be numerous (consider all the workers in a Query or
Batch pipeline).

The big change is that deploy config now has an `addresses` function which will
return a list of address-port pairs. Deploy config also now has `address` which
randomly chooses one of the address-port pairs.

I have included a simple test.

Please review both code and overall design, considering how it fits in the wider system.


danking:
The AddressAndPort pylint failure will be fixed by https://github.com/hail-is/hail/pull/9126

address/Makefile
```diff
@@ -0,0 +1,33 @@
...
+PYTHON := PYTHONPATH=$(PYTHONPATH) python3
+
+.PHONY: check
+check:
+	$(PYTHON) -m flake8  --config ../setup.cfg address
+	$(PYTHON) -m pylint --rcfile ../pylintrc address --score=n
+
+.PHONY: build
+build:
+	# make -C ../docker build
```
jigold, line 16: delete?

address/Makefile
```diff
@@ -0,0 +1,33 @@
...
+.PHONY: push
+push: build
+	docker tag address $(ADDRESS_LATEST)
+	docker push $(ADDRESS_LATEST)
+	docker tag address $(ADDRESS_IMAGE)
+	docker push $(ADDRESS_IMAGE)
+
+.PHONY: deploy
+deploy: push
+	! [ -z $(NAMESPACE) ]  # set NAMESPACE
```
jigold, line 30: fix???

address/address/address.py
```diff
@@ -0,0 +1,155 @@
...
+from hailtop.tls import get_server_ssl_context
+from web_common import setup_aiohttp_jinja2, setup_common_static_routes, render_template
+import uvloop
+import sortedcontainers
+from collections import defaultdict
+
+
+NAMESPACE = os.environ['HAIL_DEFAULT_NAMESPACE']
+
+uvloop.install()
```
jigold, line 21: Did we decide uvloop is a good thing?

address/address/address.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class CacheEntry(Generic[T]):
+    def __init__(self, value: T):
+        self.value: T = value
+        self.expire_time: float = time.time() + VALID_DURATION_IN_SECONDS
+
+
+class AddressAndPort(NamedTuple):
+    address: str
```
jigold, line 76: Huh. That's an interesting format. I guess this has to do with inheriting from NamedTuple?

address/address/address.py
```diff
@@ -0,0 +1,155 @@
...
+                         for ip in endpoint.addresses]
+            self.entries[key] = CacheEntry(endpoints)
+            self.keys.add(key)
+            log.info(f'fetched new value for {key}: {endpoints}')
+            return endpoints
+
+    async def maybe_remove_one_old_entry(self):
+        if len(self.keys) > 0:
+            key = self.keys.pop()
+            if self.entries[key].expire_time < time.time():
```
jigold, line 119: Why wouldn't you remove all old entries?

address/address/address.py
```diff
@@ -0,0 +1,155 @@
...
+            if key in self.entries:
+                entry = self.entries[key]
+                if entry.expire_time >= time.time():
+                    return entry.value
+
+            asyncio.ensure_future(self.maybe_remove_one_old_entry())
+            k8s_endpoints = await self.k8s_client.read_namespaced_endpoints(key, NAMESPACE)
+            endpoints = [AddressAndPort(ip.ip, port.port)
+                         for endpoint in k8s_endpoints.subsets
+                         for port in endpoint.ports
```
jigold, line 109: Is this the cross product of port and ip? Or will every IP have the same list of ports?

address/deployment.yaml
```diff
@@ -0,0 +1,117 @@
...
+  selector:
+    matchLabels:
+      app: address
+  replicas: 3
+  template:
+    metadata:
+      labels:
+        app: address
+        hail.is/sha: "{{ code.sha }}"
+        grafanak8sapp: "true"
```
jigold, line 18: Do we need this?

address/setup.cfg
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length=120
```
jigold, line 2: New line at end of file.

jigold, line 2: Did we not have a default one in the main hail directory?

build.yaml
```diff
@@ -2007,3 +2009,73 @@ steps:
...
+    - default_ns
+    - deploy_router
+    - address_image
+    - create_certs
+ - kind: runImage
+   name: test_address
+   image:
+     valueFrom: service_base_image.image
+   script: |
+     python3 -m pytest --log-date-format="%Y-%m-%dT%H:%M:%S" --log-format="%(asctime)s %(levelname)s %(name)s %(filename)s:%(lineno)d:%(funcName)s %(message)s" --log-cli-level=INFO -s -vv --instafail /io/test/
```
jigold, line 2055: Do you want this to be instafail?

hail/python/hailtop/config/deploy_config.py
```diff
@@ -104,6 +110,25 @@ async def get_healthcheck(request):  # pylint: disable=unused-argument,unused-va
...
 
+    async def addresses(self, service: str) -> List[Tuple[str, int]]:
+        from ..auth import service_auth_headers  # pylint: disable=cyclic-import
+        namespace = self.service_ns(service)
+        headers = service_auth_headers(self, namespace)
+        async with ssl_client_session(
+                raise_for_status=True,
+                timeout=aiohttp.ClientTimeout(total=5),
+                headers=headers) as session:
+            async with await session.get(
```
jigold, line 121: Is `{namespace}/{service}` treated as the key by the cache? It looks like it should just be the name of the service from the code above.

hail/python/hailtop/config/deploy_config.py
```diff
@@ -104,6 +110,25 @@ async def get_healthcheck(request):  # pylint: disable=unused-argument,unused-va
...
+                    self.url('address', f'/api/{namespace}/{service}')) as resp:
+                dicts = await resp.json()
+                return [(d['address'], d['port']) for d in dicts]
+
+    async def address(self, service: str) -> Tuple[str, int]:
+        service_addresses = await self.addresses(service)
+        n = len(service_addresses)
+        assert n > 0
```
jigold, line 129: Should there be a nicer error message than this? It will throw a 500.

hail/src/main/scala/is/hail/services/DeployConfig.scala
```diff
@@ -92,4 +97,20 @@ class DeployConfig(
...
+    val addressBaseUrl = baseUrl("address")
+    val url = s"${addressBaseUrl}/api/${service}"
+    val addresses = requester.request(new HttpGet(url)).asInstanceOf[JArray].children.asInstanceOf[List[JObject]]
+    addresses.map(x => ((x \ "address").extract[String], (x \ "port").extract[Int]))
+  }
+
+  def address(service: String): (String, Int) = {
+    val serviceAddresses = addresses(service)
+    val n = serviceAddresses.length
+    assert(n > 0)
```
jigold, line 113: Same question about error message here.

hail/src/main/scala/is/hail/services/Requester.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  import Requester._
+
+  def request(req: HttpUriRequest, body: HttpEntity = null): JValue = {
+    log.info(s"request ${ req.getMethod } ${ req.getURI }")
+
+    if (body != null)
+      req.asInstanceOf[HttpEntityEnclosingRequest].setEntity(body)
+
+    tokens.addServiceAuthHeaders("batch", req)
```
jigold, line 53: Why is this `batch`?

address/address/address.py
danking: yep, it's a new feature

address/address/address.py
danking: cross product. it works the same as nested for loops

address/address/address.py
danking: I want to do O(1) work, not O(N) work.

address/deployment.yaml
danking: I copied this from batch. It should probably be removed. Let's remove all of them in one separate PR.

address/setup.cfg
danking: I copied this from batch

danking: Let's remove all the extraneous setup.cfg's as a separate PR

build.yaml
danking: Yes, instafail shows errors/failures as soon as the happen instead of at the end of all the tests. I copied this from batch.

https://github.com/pytest-dev/pytest-instafail

hail/python/hailtop/config/deploy_config.py
danking: good catch, this is a bug, fixed to just the name

hail/python/hailtop/config/deploy_config.py
danking: This won't usually be used inside a web server. It's for clients to discover an IP address for a container providing a hail service, such as `shuffler`.

I'm not really sure what should happen if we ask for an address and there are none available (for example, if all the containers happen to be down). We should probably retry waiting for the service to come alive. I'm inclined to leave it like this until we have more experience using address.

hail/src/main/scala/is/hail/services/Requester.scala
danking: good find, this is a bug

danking:
rebased on master

hail/python/hailtop/tls.py
```diff
@@ -61,6 +62,22 @@ def get_in_cluster_client_ssl_context() -> ssl.SSLContext:
...
+def get_in_cluster_no_hostname_checks_client_ssl_context() -> ssl.SSLContext:
+    global no_hostname_checks_client_ssl_context
+    if no_hostname_checks_client_ssl_context is None:
+        ssl_config = _get_ssl_config()
+        no_hostname_checks_client_ssl_context = ssl.create_default_context(
+            purpose=Purpose.SERVER_AUTH,
+            cafile=ssl_config['outgoing_trust'])
+        no_hostname_checks_client_ssl_context.load_cert_chain(
+            ssl_config['cert'],
+            keyfile=ssl_config['key'],
```
jigold, line 74: For my education, can you give me the high level view of why you needed to add this function for the address service only?

danking: Line 77 is the key line. It sets check_hostname to `False`.

In Figure 11-5 on page 137 of Rice's "Container Security", we see that after the SYN/ACK and HELLOs, the server sends its certificate to the client. The client found this server directly with an IP address or indirectly by looking up a domain name. Every certificate has a list of acceptable "names" for servers using this certificate. Those "names" can be IP addresses or domain names. If `check_hostname` is set to `True`, then the client will terminate the *TLS handshake* if the name the client used does not appear in the acceptable names list.

The address service is replacing DNS. Normally DNS is performed by the TLS library itself. We can't replace DNS directly. Instead, we resolve the k8s service name first, using `DeployConfig.address`. Then we pass an IP address to the TLS library. The TLS library looks for the IP address in the acceptable names. It will not find the name and thus terminate the handshake in failure.

We cannot place pod IP addresses inside the certificates because a pod could die and return with a new IP address. We would have to change certificates every time a pod died. That's not tractable.

The long term solution is to replace DNS directly. We could modify /etc/hosts on every machine. We could try and use Google CloudDNS directly. We could use a mixture of DNS and iptables. We could literally implement the DNS protocol and point all our VMs at our new DNS service.

danking: See test_address.py for an example of looking up the IP address first then making an HTTPS request using the IP address.

danking:
bump

hail/python/hailtop/tls.py
jigold: You got rid of the check_hostname to False argument. Is this still needed then?

PR #9139, opened 2020-07-23
[hail] split java tests into 5 pieces

danking:


tpoterba:
oh, sweet!

danking:
This more or less follows the same strategy as the python splitting. Instead of each job figuring out which split it owns, I create 5 separate testng.xml files in `build_hail`. Each job reads one of those files. testng.xml has a pretty simple XML format and you can explicitly list the classes of interest.

I noticed that hail java tests were up to 15 minutes which was really cramping my PR merging style. With this, some of the splits are a minute or two. I think we suffer a bit from programmatically generated tests since I'm splitting at the granularity of classes rather than methods, or, even better, generated methods.

danking:
The TableIRSuite and the OrderingSuite, in particular, are rather slow compared to other suites.

danking:
damn testTableJoin is crazy, that one method generates ~500 tests.

danking:
testTableJoin takes five minutes

PR #9143, opened 2020-07-23
[query] lower BlockMatrixFilter/BlockMatrixSlice

catoverdrive:


hail/src/test/scala/is/hail/expr/ir/BlockMatrixIRSuite.scala
```diff
@@ -119,6 +119,37 @@ class BlockMatrixIRSuite extends HailSuite {
     }
   }
 
+  @Test def testBlockMatrixFilter() {
```
tpoterba, line 122: are these actually using the relational strategy? 

Er, just confirmed, yes they are 👍

PR #9146, opened 2020-07-24
NDArray Slicing Clamp Bounds

johnc1231:
When slicing an ndarray, if someone asks for a slice bigger than the array, we need to clamp the edges of that slice. 

I will make a follow up PR that elevates `clamp` to a proper Hail function like `min` and `max`, as I think it's generally useful.

### DIFF
--- a/hail/python/hail/expr/expressions/typed_expressions.py
+++ b/hail/python/hail/expr/expressions/typed_expressions.py
@@ -3578,6 +3578,12 @@ class NDArrayExpression(Expression):
             raise ValueError(f'Must specify one index per dimension. '
                              f'Expected {self.ndim} dimensions but got {len(item)}')
 
+        def clamp(expr, min, max):
+            return (hl.case()
+                    .when(expr < min, min)
+                    .when(expr > max, max)
+                    .default(expr))
+
         n_sliced_dims = len([s for s in item if isinstance(s, slice)])
         if n_sliced_dims > 0:
             slices = []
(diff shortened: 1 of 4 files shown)


hail/src/main/scala/is/hail/asm4s/Code.scala
```diff
@@ -824,6 +824,12 @@ class CodeLong(val lhs: Code[Long]) extends AnyVal {
 
   def compare(rhs: Code[Long]): Code[Int] = Code(lhs, rhs, lir.insn2(LCMP))
 
+  def max(rhs: Code[Long]): Code[Long] =
```
johnc1231, line 827: These changes snuck in an earlier approach to fix this is in Scala before I settled on clamping in python. I can remove if you'd like

hail/python/hail/expr/expressions/typed_expressions.py
```diff
@@ -3578,15 +3578,23 @@ def __getitem__(self, item):
...
         if n_sliced_dims > 0:
             slices = []
             for i, s in enumerate(item):
                 if isinstance(s, slice):
                     step = hl.case().when(s.step != 0, s.step).or_error("Slice step cannot be zero") if s.step is not None else to_expr(1, tint64)
                     start = hl.cond(s.start >= 0, s.start, self.shape[i] + s.start) if s.start is not None else hl.cond(step >= 0, to_expr(0, tint64), self.shape[i] - 1)
+                    clamped_start = clamp(start, 0, self.shape[i] - 1)
                     stop = hl.cond(s.stop >= 0, s.stop, self.shape[i] + s.stop) if s.stop is not None else hl.cond(step >= 0, self.shape[i], to_expr(-1, tint64))
-                    slices.append(hl.tuple((start, stop, step)))
+                    clamped_stop = clamp(stop, -1, self.shape[i])
```
akotlar, line 3596: should t min here be -self.shape[I]?

johnc1231: I don't think so. The long `stop = hl.cond(....` line above should translate everything into positive numbers.

johnc1231: I only handle positive starts and stops in the emitter, this cleans them up so we match numpy semantics with indexing from the end using negatives

johnc1231:
Suddenly getting really surprising error here about low level codegen stuff.

hail/src/main/scala/is/hail/asm4s/Code.scala
akotlar: Let's remove this since not used, unneeded

akotlar:
I see that as well, isOpenEnded.

PR #9150, opened 2020-07-24
[query] Move slow/bad memory test to benchmarks, improve BM.WFEE pruning

tpoterba:
* Improves pruning of BlockMatrix.write_from_entry_expr to avoid large
allocation in the slow test.
* Delete the slow test.
* Add two benchmarks that should catch this


hail/python/hail/matrixtable.py
```diff
@@ -4102,7 +4102,7 @@ def summarize(self, *, rows=True, cols=True, entries=True, handler=None):
 
     def _write_block_matrix(self, path, overwrite, entry_field, block_size):
         mt = self
-        mt = mt.select_entries(entry_field).select_cols().select_globals()
+        mt = mt._select_all(entry_exprs={entry_field: mt[entry_field]})
```
johnc1231, line 4105: What large allocation was happening here, and why does this fix it?

tpoterba: This test Dan added on June 23 after the countPerPartition memory fix:
```python
def test_memory_issue_from_9009():
    mt = hl.utils.range_matrix_table(1000, 1, n_partitions=1)
    mt = mt.annotate_entries(x=hl.float(mt.row_idx * mt.col_idx))
    mt = mt.annotate_rows(big=hl.zeros(100_000_000))
    try:
        hl.linalg.BlockMatrix.write_from_entry_expr(mt.x, new_temp_file(), overwrite=True)
    except Exception:
        assert False
```

I expected this test to be instant (and not test memory usage), because `big` should be a dead field. We do pruning in Python with `select*` methods for many of the relational functions like MatrixWriteBlockMatrix, but we neglected to do a select_rows(), so `big` wasn't eliminated.

johnc1231: Ah that makes sense. So another fix would have just been to add a `.select_rows()` after the `select_globals()`, but you decided you might as well use `_select_all()` instead.

tpoterba: yeah, seemed simpler.

PR #9152, opened 2020-07-25
[query] ndarray slice bounds fixed

akotlar:
This PR supersedes #9146, sets ndarray slicing behavior to follow ndarray/python semantics when out of bounds: upper_bound = len_dim if step > 0 else len_dim - 1, lower_bound = 0 if step > 0 else -1 (since our -1 has the effect of a slice bound that is None)

The other PR doesn't clamp dimensions properly (min and max out of bounds for start and stop are identical), and missed a 2nd issue, which is that python changes clamping behavior when step is negative. I bet this polymorphism is the reason we don't allow non-1 steps for ArrayExpressions.
  - @johnc1231 could you still follow up on the byte code generation issue?

I've verified this works in the boundary cases.

### DIFF
--- a/hail/python/test/hail/expr/test_ndarrays.py
+++ b/hail/python/test/hail/expr/test_ndarrays.py
@@ -93,7 +93,11 @@ def test_ndarray_slice():
         (mat[::-1, :], np_mat[::-1, :]),
         (flat[4:1:-2], np_flat[4:1:-2]),
         (flat[0:0:1], np_flat[0:0:1]),
-        (flat[-4:-1:2], np_flat[-4:-1:2])
+        (flat[-4:-1:2], np_flat[-4:-1:2]),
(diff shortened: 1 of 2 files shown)


hail/python/test/hail/expr/test_ndarrays.py
```diff
@@ -93,7 +93,11 @@ def test_ndarray_slice():
         (mat[::-1, :], np_mat[::-1, :]),
         (flat[4:1:-2], np_flat[4:1:-2]),
         (flat[0:0:1], np_flat[0:0:1]),
-        (flat[-4:-1:2], np_flat[-4:-1:2])
+        (flat[-4:-1:2], np_flat[-4:-1:2]),
```
akotlar, line 96: I borrowed your tests

akotlar:
added some more tests, passes

PR #9161, opened 2020-07-27
CHANGELOG: Fix bug that prevented concatenating ndarrays that are fields of a table

johnc1231:
CHANGELOG: Fix bug that prevented concatenating ndarrays that are fields of a table

Two bugs:

1. We can't call `.typ` on `concat_ir`, since that kicks off the type inference process before this IR gets wrapped in a `TableIR` that provides the necessary context to infer the type.
2. We need to `_compute_type` on `self.nds` as a whole, not just each of its `args`.


akotlar:
Thanks John!

PR #9164, opened 2020-07-28
[query] Better error message when using `in` incorrectly

johnc1231:
Let me know if you think the error message is clear.

### DIFF
--- a/hail/python/hail/expr/expressions/base_expression.py
+++ b/hail/python/hail/expr/expressions/base_expression.py
@@ -642,6 +642,10 @@ class Expression(object):
     def __len__(self):
         raise TypeError("'Expression' objects have no static length: use 'hl.len' for the length of collections")
 
+    def __contains__(self, item):
+        class_name = type(self).__name__
+        raise TypeError(f"`{class_name} objects don't support the `in` operator.")
+
     def __hash__(self):
         return super(Expression, self).__hash__()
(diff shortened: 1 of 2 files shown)


tpoterba:
we should put this on top level expression. Someone could do something like `'NA12878' in mt.s` and that should have an error as well.

johnc1231:
@tpoterba, I thought about that, but it would break compatibility, since we currently allow:

```
"x" in hl.struct(x = 3, y = 5)
```

tpoterba:
oh, right, good point. though I think maybe StructExpression would override the Expression definition?

johnc1231:
I guess I could put it on top level and then override it for `StructExpression`

johnc1231:
Is there any `in` behavior besides the struct one? I don't know of any, but I don't often use things like `Interval` or `Call` ,etc.

tpoterba:
TupleExpression may override (with something wrong) too.

johnc1231:
Ok, so now there's a generic catch all error on `Expression`, and I override it on `StructExpression` to make sure that one works. `CollectionExpression` and `StringExpression` also get overrides to give appropriate examples.

johnc1231:
@konradjk @tpoterba I think this is ready to merge whenever.

johnc1231:
Bump

PR #9165, opened 2020-07-28
[query] make concatenate operations work with array expressions

akotlar:
CHANGELOG: make hl.nd.concatenate, hl.nd.vstack, and hl.nd.hstack accept ArrayExpression of NDArrayExpression, in addition to Python collections of NDArrayExpression

### DIFF
--- a/hail/python/hail/nd/nd.py
+++ b/hail/python/hail/nd/nd.py
@@ -11,8 +11,7 @@ from hail.expr.expressions import (
 from hail.expr.expressions.typed_expressions import NDArrayNumericExpression
 from hail.ir import NDArrayQR, NDArrayInv, NDArrayConcat
 
-tsequenceof_nd = oneof(sequenceof(expr_ndarray()), tupleof(expr_ndarray()),
-                       expr_array(expr_ndarray()))
+tsequenceof_nd = oneof(sequenceof(expr_ndarray()), expr_array(expr_ndarray()))
 shape_type = oneof(expr_int64, tupleof(expr_int64), expr_tuple())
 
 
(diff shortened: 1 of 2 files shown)


hail/python/hail/nd/nd.py
```diff
@@ -443,13 +442,15 @@ def vstack(arrs):
            [3],
            [4]], dtype=int32)
     """
-
-    if arrs[0].ndim == 1:
-        arrays = [x._broadcast(2) for x in arrs]
+    if isinstance(arrs, list):
```
johnc1231, line 445: I think we can avoid this. If we change the type check from `tsequence_of(tndarray)` to `expr_array(expr_ndarray)` (or something like that), then all python arrays passed into this method will automatically be converted into hail arrays, and we won't need any branching.

akotlar:
I can't make the proposed change, without fixing both the sequenceof checker, and the expr_tuple coercer. Both are subtly broken.

The former is broken because it changes the type of the income sequence. If you provide a tuple,  SequenceChecker will convert that to a list. That is undocumented behavior, and is therefore either a bug, or if that magic is desired, a documentation bug.

The latter is broken because it tries to call len on NDArrayCoercer.

hail/python/hail/nd/nd.py
```diff
@@ -314,8 +313,7 @@ def concatenate(nds, axis=0):
     """
     head_nd = nds[0]
     head_ndim = head_nd.ndim
-    for nd in nds:
-        assert(nd.ndim == head_ndim)
+    hl.case().when(hl.all(lambda a: a.ndim == head_ndim, nds), True).or_error("Mismatched ndim")
```
akotlar, line 316: Also, unfortunately hl.all does not work with tuples, which I think should be classified as bug @tpoterba

PR #9170, opened 2020-07-29
[query] Fix tree agg depth logic

tpoterba:


danking:
changelings!

tpoterba:
this probably won't really affect users

PR #9176, opened 2020-07-30
[batch] fix docker resource doc

akotlar:
Difficulty in containerizing code is a current difficulty stated by some ATGU members as blocker on using Batch.  

The instructions will not work as stated on master, because the image-name needs to include the fully qualified gcr.io path (gcr.io/project/). Also removes unnecessary `docker tag`, cutting instruction length. During `docker build` users can provide an arbitrary number of `-t <name>`, saves a step, and a lot of words.

In future iterations we may also want to think about changing the instructions to use DockerHub by default, because this requires less permission wrangling.

### DIFF
--- a/hail/python/hailtop/batch/docs/docker_resources.rst
+++ b/hail/python/hailtop/batch/docs/docker_resources.rst
@@ -71,11 +71,21 @@ For more information about Dockerfiles and directives that can be used see the f
 Building Images
 ---------------
 
-To create a Docker image, use `docker build <dir> -t <name> -f <dockerfile>`. <dir> specifies the context directory (typically `.`, for the current working directory).
-Use `-t` to name the image, and `-f` to specify the Dockerfile path.
-More information can be found `here: <https://docs.docker.com/engine/reference/commandline/build/>`__.
+To create a Docker image, use
 
 .. code-block:: sh
+
+    docker build -t gcr.io/<my-project>/<my-image>:<tag> -f Dockerfile .
+
+    * `<dir>` is the context directory, `.` means the current working directory,
+    * `-t <name>` specifies the image name, and
+    * `-f <dockerfile>` specifies the Dockerfile file.
+    * A more complete description may be found `here: <https://docs.docker.com/engine/reference/commandline/build/>`__.
+
+For example, we can build an image named gcr.io/<my-project>/<my-image> based on the Dockerfile named Dockerfile, using the current working directory as the context:
+
+.. code-block:: sh
+
     docker build -t gcr.io/<my-project>/<my-image>:<tag> -f Dockerfile .
 
 In this example we prepend the image name with `gcr.io/<my-project>/` so that it may be pushed to the Google Container Registry, in the next step.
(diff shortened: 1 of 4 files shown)


hail/python/hailtop/batch/docs/docker_resources.rst
```diff
@@ -71,29 +71,26 @@ For more information about Dockerfiles and directives that can be used see the f
 Building Images
 ---------------
 
-To create a Docker image, we use a series of commands to build the image from a Dockerfile by specifying
-the context directory (in this case the current directory `.`). The `-f` option
-specifies what Dockerfile to read from. The `-t` option is the name of the image.
-More in depth information can be found `here <https://docs.docker.com/engine/reference/commandline/build/>`__.
+To create a Docker image, use `docker build <dir> -t <name> -f <dockerfile>`. <dir> specifies the context directory (typically `.`, for the current working directory).
```
danking, line 74: In ReStructured Text, back ticks indicate "content" which is interpreted in a context dependent way. In Python docs, "content" is rendered italicized. AFAICT, we've preferred using shell code blocks when demonstrating code examples. I think we should stick to that style.

I think something like this reads well:
> To build a Docker image, use
> ```
> docker build <dir> -t <name> -f <dockerfile>
> ```
> - ``<dir>`` is the context directory, ``.`` means the current working directory,
> - ``-t <name>`` specifies the image name, and
> - ``-f <dockerfile>`` specifies the Dockerfile file.
> 
> For example, we can build an image named ``gcr.io/my-project/my-image`` based on the Dockerfile named ``Dockerfile``, using the current working directory as the context:
> ```
> docker build . -t gcr.io/my-project/my-image -f Dockerfile
> ```
>
> In this example ...

I think it's important we stay consistent about where we place the context directory.

hail/python/hailtop/batch/docs/docker_resources.rst
```diff
@@ -71,29 +71,26 @@ For more information about Dockerfiles and directives that can be used see the f
...
 Pushing Images
 --------------
 
-To use an image with Batch, you need to upload your image to a place where Batch can access it.
-You can store images inside the `Google Container Registry <https://cloud.google.com/container-registry/docs/>`__ in
-addition to Dockerhub. Below is an example of pushing the image to the Google Container Registry.
-It's good practice to specify a tag that is unique for your image. If you don't tag your image, the default is
-`latest`. For more information about pulling and pushing images for the Google Container Registry, see
+To use an image with Batch, you need to upload it somewhere Batch can read it: either the `Google Container Registry <https://cloud.google.com/container-registry/docs/>`__ or
+Dockerhub. Below is an example of pushing the image to the Google Container Registry. For more information about pulling and pushing images for the Google Container Registry, see
```
danking, line 89: I think the docker folks prefer the styling "Docker Hub," do you mind fixing that while we're editing this file anyway? Thanks!

akotlar: Confusing:

Screen Shot 2020-08-06 at 5 47 07 AM

akotlar:
bump

PR #9178, opened 2020-07-30
[query][smm 5] Memory management for StreamFold and StreamFold2

patrick-schultz:
~~Stacked on #9177~~

These are tricky because of the overlapping lifetimes of the accumulator values. The solution implemented here is to keep two regions. When running the fold body, one region is empty, and the other holds just the previous accumulator. Then we deep copy the new accumulator to the empty region, clear the other one, then swap the two regions before moving to the next iteration.

This required expanding the StagedRegion API a bit. First, it now strictly tracks the parent StagedRegion, even in the non-allocating case where the parent wraps the same run-time region. Second, in addition to copying a value to the parent, we can now copy a value to a sibling. It enforces at compile time that the two regions have the same parent. It follows that either both are actually separate regions, or both are just wrappers around the parent region. In the latter case, copying to a sibling is a no-op.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -334,12 +334,8 @@ case class EmitCode(setup: Code[Unit], m: Code[Boolean], pv: PCode) {
     IEmitCode(Lmissing, Lpresent, pv)
   }
 
-  def castTo(mb: EmitMethodBuilder[_], region: Value[Region], destType: PType): EmitCode = {
-    EmitCode(
-      setup,
-      m,
-      pv.castTo(mb, region, destType))
-  }
+  def castTo(mb: EmitMethodBuilder[_], region: Value[Region], destType: PType, deepCopy: Boolean = false): EmitCode =
+    EmitCode(setup, m, pv.castTo(mb, region, destType, deepCopy))
 
   def codeTuple(): IndexedSeq[Code[_]] = {
     val tc = pv.codeTuple()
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -844,43 +840,79 @@ class Emit[C](
...
-          stream.asStream.stream.getStream(region).forEachI(cb, { elt =>
+          stream.asStream.stream.getStream(eltRegion).forEachI(cb, { elt =>
+            // pre- and post-condition: 'accVars' contain current accumulators,
+            // all of whose heap memory is contained in 'eltRegion'. 'tmpRegion'
+            // is empty.
             cb.assign(xElt, elt)
-            (accVars, seq).zipped.foreach { (accVar, ir) =>
-              cb.assign(accVar, emitI(ir, env = seqEnv).map(cb)(_.castTo(mb, region.code, accVar.pt)))
+            (tmpAccVars, seq).zipped.foreach { (accVar, ir) =>
+              // FIXME: needs to be simultaneous assignment
```
tpoterba, line 902: what does this mean?

patrick-schultz: Whoops, that's old. The two stage assignment to `tmpAccVars` then to `accVars` addresses this, so all `seq` IRs are evaluated in an environment with the previous accumulator values. I'll delete.

tpoterba: ok, cool. Looked right to me.

PR #9181, opened 2020-07-30
[query][smm 3] force region to be passed to StagedRegionValueBuilder constructor

patrick-schultz:
Stacked on #9220

The region parameter to emit ostensibly means "construct the value in this region". But in many places `StagedRegionValueBuilder`s are constructed with the default region, which is the argument to the method being emitted into.

This PR fixes all the uses in the emitter to build values in the right region. Most of the remaining users of the constructor with the default region are in tests, so I thought it safest to remove those constructors, and require explicitly passing the region to build into. Otherwise it's too easy to create memory leaks.

### DIFF
--- a/hail/src/main/scala/is/hail/annotations/StagedRegionValueBuilder.scala
+++ b/hail/src/main/scala/is/hail/annotations/StagedRegionValueBuilder.scala
@@ -45,18 +45,6 @@ class StagedRegionValueBuilder private (val mb: EmitMethodBuilder[_], val typ: P
     this(mb, typ, parent.region, parent.currentOffset)
   }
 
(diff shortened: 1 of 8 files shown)


hail/src/main/scala/is/hail/annotations/StagedRegionValueBuilder.scala
```diff
@@ -45,18 +45,6 @@ class StagedRegionValueBuilder private (val mb: EmitMethodBuilder[_], val typ: P
     this(mb, typ, parent.region, parent.currentOffset)
   }
 
-  def this(fb: EmitFunctionBuilder[_], rowType: PType) = {
```
tpoterba, line 48: yay deletions!

PR #9184, opened 2020-07-30
[batch] cleanup content in worker.log

jigold:
I wasn't sure if we should print at least one Docker config or not. I'm worried about the test for out_of_memory where that prints out a bunch of the letter 'a' in the script.


jigold:
@cseed added these print statements.

I've never used the output in debugging in production except to quickly realize a job is one of a given user's when I was looking at worker performance. Early on, I might have checked it to make sure the Docker config resource options and volume mounts were correct. But I can add those back when debugging if needed.

PR #9189, opened 2020-07-31
[batch] Add fluentd support to worker

jigold:
This should wait for #9184 

This PR adds fluentd support for streaming the logs to Logs Viewer.

Things to Note:
-  I didn't have to setup the authorization at all, so I believe it's using the service account on the worker which would be the batch2 service account. I had to give that service account Logs Writer permissions. This is the same service account regardless of the namespace.
- All namespaces have output written.
- I added some labels of the instance id and namespace to help with searching. Let me know if you think we need something else.
- I used a filter to parse the JSON in the worker log to get the right timestamp of the record rather than the publication of the log message in the Logs Viewer

Example Stackdriver Queries:

```
resource.type="gce_instance"
labels.namespace="jigold"
logName="projects/hail-vdc/logs/worker.log"
```

```
resource.type="gce_instance"
labels."compute.googleapis.com/resource_name"="batch-worker-jigold-uyvo5"
logName="projects/hail-vdc/logs/worker.log"
```

I basically just followed this: https://cloud.google.com/logging/docs/agent

Will attempt to get dockerd logs in some other time.

### DIFF
--- a/batch/build-batch-worker-image-startup.sh
+++ b/batch/build-batch-worker-image-startup.sh
@@ -1,11 +1,16 @@
 #!/bin/bash
 
+curl -sSO https://dl.google.com/cloudagents/add-logging-agent-repo.sh
+bash add-logging-agent-repo.sh
+
 apt-get update
 
 apt-get install -y \
     apt-transport-https \
     ca-certificates \
     curl \
+    google-fluentd \
+    google-fluentd-catch-all-config-structured \
     jq \
     software-properties-common
 
@@ -34,5 +39,7 @@ docker-credential-gcr configure-docker
 
 docker pull ubuntu:18.04
 docker pull google/cloud-sdk:269.0.0-alpine
+
+service google-fluentd start
 
 shutdown -h now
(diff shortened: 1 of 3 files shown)


jigold:
For dockerd, we probably need to use this: https://github.com/fluent-plugin-systemd/fluent-plugin-systemd

batch/build-batch-worker-image-startup.sh
```diff
@@ -1,11 +1,16 @@
 #!/bin/bash
 
+curl -sSO https://dl.google.com/cloudagents/add-logging-agent-repo.sh
```
danking, line 3: I like to use the full flags when we can, and I think we should add `--fail`:

```
curl --silent --show-error --remote-name --fail \
     https://dl.google.com/cloudagents/add-logging-agent-repo.sh
```

```
       -f, --fail
              (HTTP) Fail silently (no output at all) on server errors. This is
              mostly done to better enable scripts etc to better deal with
              failed attempts. In normal cases when an HTTP server fails to
              deliver a document, it returns an HTML document stating so (which
              often also describes why and more). This flag will prevent curl
              from outputting that and return error 22.

              This method is not fail-safe and there are occasions where
              non-successful response codes will slip through, especially when
              authentication is involved (response codes 401 and 407).
```

danking: `--fail` with `--show-error` is a great combo that prints useful information but also ensures that we get an error code when something goes wrong. On that note, should this shell script have `set -ex`?

danking:
Great work!

PR #9191, opened 2020-07-31
[batch] add close method to Backend

akotlar:
CHANGELOG: Adds close, default implementation pass, to Backend as an abstract method.

This reduces the number of conditional statements needed by 1 when using variable backends. Variable backends (Local or Service) are useful when prototyping batches, or when the user will know in advance that a particular batch routine will work locally (since in future work it will be much easier to consume dockerized methods using Batch than anything else). Additionally, as provided, the only Batch tutorial, GWAS clumping will not work with  LocalBackend without this. Use case:

```python
    parser.add_argument('--local', required=False, action="store_true")
    if is_local:
        backend = hb.LocalBackend()
        run_opts = {}
    else:
        backend = hb.ServiceBackend()
        run_opts = {open: True, wait: True}

    # do a bunch of Batch stuff to 
    batch.run(**run_opts)
    backend.close()
```

In a similar vein, I'd like to allow LocalBackend to ignore unused run opts. Again, GWAS tutorial would not work with LocalBackend without this (or an opts dict as above).


akotlar:
I can add it if you'd like, thought we'd prefer to save keystrokes:

```python
In [3]: class Backend: 
   ...:     """ 
   ...:     Abstract class for backends. 
   ...:     """ 
   ...:  
   ...:     @abc.abstractmethod 
   ...:     def close(self): 
   ...:         """ 
   ...:         Close a Hail Batch backend. 
   ...:         """ 
   ...:         pass 
   ...:  
   ...:  
   ...: class LocalBackend(Backend):                                                                                                                                             

In [4]: n = LocalBackend()                                                                                                                                    

In [5]: n.close()         
```

This allows us to declare an interface and a void implementation in one.

danking:
Ah cool, I didn't know that's how Python abstract methods work!

danking:
Ahh, hmm. Pylint agrees with me that this isn't Kosher:
```
/hailtop/batch/backend.py:44:8: W0107: Unnecessary pass statement (unnecessary-pass)
/hailtop/batch/backend.py:47:0: W0223: Method 'close' is abstract in class 'Backend' but is not overridden (abstract-method)
```

We've generally heeded pylint's advice even if it clashes with our intuition.

danking:
Perhaps eliding the `@abc.abstractmethod` is sufficient?

akotlar:
yep, nice catch, didn't check pylint.

PR #9194, opened 2020-07-31
[batch] regenie

akotlar:
CHANGELOG: early implementation of regenie

I'd like to get this basic version in and iterate. It works on local.

Has a few todos and fixmes; I've seen Cotton, others use these, and they'll be gone in fairly short order, saves a bit of time over making a formal issue, for something that is clearly wip. 

This looks like a scary amount of lines, but almost all of the work is held in regenie-batch.py. Example files are included in contrib/regenie/regenie, which is a redacted copy of their repo, and which is sufficient, lighter-weight than the full.

When you have a chance, I'd like to discuss increasing the max capacity of the SSD (striping partitions). I want this as a fallback mechanism, when not enough memory can be allocated (lowmem). It will likely perform terribly with persistent storage. I'm happy to contribute that.


danking:
I added you to ServicesDev, let's start a thread there to discuss the SSD situation.

akotlar:
> What's the pressing reason for this to join the mainline?

Ben has wanted this for a while, and there's interest from the group.

> I'd prefer to stick to the usual process on services team where thorough code review of code and test-code is done for stuff under our purview (as things in hailtop/batch do).

Sure. Regarding testing, in my experience regenie (the C++ program) crashes when given improper inputs, rather than writing anything out, so there is a reduced testing need.

My plan for getting regenie in is roughly 4 part, of which this is #1. 
1) Get the most basic / canonical version of regenie up, on the local backend. This version should take all of the arguments demonstrated in the regenie tutorial / example, pass them to step 1 and step 2, and write an output. 
  * Here I'm mostly interested in making sure I'm using Batch Resource classes correctly.
2) Get the same working for the ServiceBackend.
3) Expand the service offering to run the paper's example (may require some tweaks to the input handling code), and test that we can run this at scale
4) Introduce a per-phenotype parallelism mode, which loses a bit of per-core efficiency in favor of greater scale out, benchmark that.
5) Handle all other combinations of inputs.

So, if at all possible, I would like to stay focused on #1, within the bounds of what you need to accomplish as leader of the services team.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+            add[name] = batch.read_input_group(**inf)
+            continue
+
+        add[name] = batch.read_input(val)
+
+    return add
+
+
+def prepare_jobs(batch, args: BatchArgs, step1_args: argparse.Namespace,
+                 step2_args: argparse.Namespace):
```
danking, line 135: Prefer either all on one line or each argument on its own line.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+        if r.extract or r.exclude:
+            raise Exception("--extract and --exclude only work with --step 1")
+
+    if r.lowmem and not r.lowmem_prefix:
+        raise Exception("When --lowmem provided, --lowmem-prefix required")
+
+    return r
+
+
+# TODO: check --phenoCol, --phenoColList, --remove
```
danking, line 96: let's nix the TODO here, and remove the three arguments from the argparse above so that users cannot accidentally specify them and then be surprised when they're not used.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+        if name not in input_file_args or val is None:
+            continue
+
+        if name == "bed":
+            prefix = step_args.bed
+            inf = {}
+            inf["bed"] = f"{prefix}.bed"
+            inf["bim"] = f'{prefix}.bim'
+            inf["fam"] = f'{prefix}.fam'
+            add[name] = batch.read_input_group(**inf)
```
danking, line 117: Let's just inline the kwargs here it reads easier:
```python
add[name] = batch.read_input_group(
    bed=f'{prefix}.bed',
    bim=f'{prefix}.bim',
    fam=f'{prefix}.fam')
```

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+            continue
+
+        if name == "pgen":
+            prefix = step_args.pgen
+            inf = {}
+            inf["pgen"] = f'{prefix}.pgen'
+            inf["pvar"] = f'{prefix}.pvar'
+            inf["psam"] = f'{prefix}.psam'
+            add[name] = batch.read_input_group(**inf)
```
danking, line 126: same kwargs thing here

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+def get_input(batch, step_args: argparse.Namespace):
+    add = {}
+    for name, val in vars(step_args).items():
+        if name in from_underscore:
+            name = from_underscore[name]
+
+        if name not in input_file_args or val is None:
+            continue
```
danking, line 109: instead of if..continue..if..continue, prefer:

```
if ...:
elif ...:
elif ...:
else:
```

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+        return f.readline().strip().split(" ")[2:]
+
+
+def get_input(batch, step_args: argparse.Namespace):
+    add = {}
+    for name, val in vars(step_args).items():
+        if name in from_underscore:
+            name = from_underscore[name]
+
+        if name not in input_file_args or val is None:
```
danking, line 108: Under what circumstance does an input file argument have no value? That seems like would always be a bug?

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+    return add
+
+
+def prepare_jobs(batch, args: BatchArgs, step1_args: argparse.Namespace,
+                 step2_args: argparse.Namespace):
+    j1 = batch.new_job(name='run-regenie')
+    j1.image('akotlar/regenie:latest')
+    j1.cpu(args.cores)
+    j1.memory(args.memory)
+    j1.storage(args.storage)
```
danking, line 140: These are the same for both jobs, so let's use `default_cpu`, `default_memory`, and `default_storage` parameters to the `Batch` constructor.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+
+        add[name] = batch.read_input(val)
+
+    return add
+
+
+def prepare_jobs(batch, args: BatchArgs, step1_args: argparse.Namespace,
+                 step2_args: argparse.Namespace):
+    j1 = batch.new_job(name='run-regenie')
+    j1.image('akotlar/regenie:latest')
```
danking, line 137: Hmm. We probably want this to be a script option and we'll want to build an image in `build.yaml` that we can use in CI tests to test `regenie-batch.py`. I also understand that we're still ironing out details here. We'll want the Hail deploy process to push out a copy of this image versioned by the hail pip version, e.g. `hailgenetics/regenie:0.2.53`.

I'm really inclined to have automated tests. How far are we from being able to run this on a tiny (like 10 variants and 10 samples, say) and verify the results? It sounds like you've been running tests already, are those hard to get integrated into an automated test as a part of this PR? We don't have to verify outputs for the first PR, we can just verify REGENIE runs without error.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+    j1.cpu(args.cores)
+    j1.memory(args.memory)
+    j1.storage(args.storage)
+
+    in_step1 = get_input(batch, step1_args)
+
+    phenos = get_phenos(step1_args)
+    nphenos = len(phenos)
+
+    s1pre = step1_args.out
```
danking, line 147: can we call this `step1_output_prefix`?

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+            cmd1 += f" --{name} {in_step1[name]}"
+            continue
+
+        if name == "out":
+            cmd1 += f" --{name} {j1[s1pre]}"
+            continue
+
+        if isinstance(val, bool):
+            cmd1 += f" --{name}"
+            continue
```
danking, line 182: same thing here about if-continue vs if-elif.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+
+    cmd1 = ""
+    for name, val in vars(step1_args).items():
+        if val is None or val is False:
+            continue
+
+        if name in from_underscore:
+            name = from_underscore[name]
+
+        # pred is not used in step 1 according to documentation
```
danking, line 168: Sounds like we should remove it from the argument parser for step1, yeah? I prefer errors to silently dropping things, when possible.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+        s1out[f"{s1pre}_{i}"] = f"{s1pre}_{i}.loco"
+
+    if step1_args.lowmem:
+        for i in range(1, nphenos + 1):
+            pfile = f"{step1_args.lowmem_prefix}_l0_Y{i}"
+            s1out[pfile] = pfile
+
+    j1.declare_resource_group(**{s1pre: s1out})
+
+    cmd1 = ""
```
danking, line 160: minor preference to create a list and then `' '.join(cmd1)`; seems like we're less likely to accidentally miss a leading space down below.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+
+        if name == "out":
+            cmd1 += f" --{name} {j1[s1pre]}"
+            continue
+
+        if isinstance(val, bool):
+            cmd1 += f" --{name}"
+            continue
+
+        cmd1 += f" --{name} \"{val}\""
```
danking, line 184: We should probably shlex.quote all the values of non-resource arguments for good measure

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+        for i in range(1, nphenos + 1):
+            pfile = f"{step1_args.lowmem_prefix}_l0_Y{i}"
+            s1out[pfile] = pfile
+
+    j1.declare_resource_group(**{s1pre: s1out})
+
+    cmd1 = ""
+    for name, val in vars(step1_args).items():
+        if val is None or val is False:
+            continue
```
danking, line 163: This seems like a source of accidental bugs as we modify and extend this file? How do we get None or False? If we don't expect this to happen, let's put an assert here: `assert val not in (None, False), val`

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+    in_step1 = get_input(batch, step1_args)
+
+    phenos = get_phenos(step1_args)
+    nphenos = len(phenos)
+
+    s1pre = step1_args.out
+    s1out = {"log": f"{s1pre}.log", "pred_list": f"{s1pre}_pred.list"}
+
+    for i in range(1, nphenos + 1):
+        s1out[f"{s1pre}_{i}"] = f"{s1pre}_{i}.loco"
```
danking, line 151: this doesn't seem quite right to me. see above comment.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
danking: Allowing the user to specify the output file prefix used internally in batch seems unnecessary and appears to add complexity to this implementation?

I guess if we already have it, we can leave it. In the controlled docker environment, I'm inclined to just pick a descriptive name, like `output`. That'll make all this batch resource management code a bit easier to read and removes some argument parsing logic. This all simplifies to:

```python
j1.declare_resource_group(
    output={
        'log': '{root}.log',
        'pred_list': '{root_pred.list}',
        **{f'pheno_{i+1}': '{{root}}_{i+1}.loco' for i in range(0, nphenos)},
        # ... and add lowmem and w/e is necessary
})
```

danking: n.b. the `{{root}}` thing becomes `{root}` after python string formatting, then batch further refines that to the right root for this resource group.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+        step1_args = read_args(args.step1)
+        step2_args = read_args(args.step2)
+
+    batch_args = BatchArgs(cores=args.cores, memory=args.memory, storage=args.storage)
+
+    batch = hb.Batch(backend=backend, name='regenie')
+
+    j1, j2, j2_out_key = prepare_jobs(batch, batch_args, step1_args, step2_args)
+
+    # FIXME: this will never write to an output directory
```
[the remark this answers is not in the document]
akotlar: It seems so. I cannot get it to print to a directory. Was planning on issuing a PR after some more investigation. The use seems idiomatic in batch, so shouldn't block this PR

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+
+    if r.step == 2:
+        if r.pred:
+            print("--pred provided for --step 2, but Batch will constrain the --pred output prefix to that of --step 1.")
+
+        if r.extract or r.exclude:
+            raise Exception("--extract and --exclude only work with --step 1")
+
+    if r.lowmem and not r.lowmem_prefix:
+        raise Exception("When --lowmem provided, --lowmem-prefix required")
```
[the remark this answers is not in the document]
akotlar: Yes, I want the interface to be nearly identical. Less of an interface for us to maintain

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
akotlar: I think this was a line length issue, will double check

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
akotlar: When argparse specifies it but it is not provided. False is what happens when a Boolean argument is not provided. These are distinct. Boolean/flag arguments should be False by default.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
akotlar: They won't be. This PR is to establish a working local version, a follow up PR will better handle resources.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
akotlar: As of this PR the "demo" version works and outputs the correct file. I'll push a test today (as well as fixes for most of your comments, thanks for those)

akotlar: The image should just go to hail genetics; I am not sure I have permissions to push there, will check when I get home. I could not push to hail-vdc or broad-ctsa projects in my first attempt, and didn't bother working out the permissions issues as I had something that worked better for the initial pr

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
akotlar: So I found this frustrating, but maybe I did the wrong thing. The arguments/interface for regenie takes an --out for step1 and --step2. If I want regenie to be able to use these (and why not, they're provided, I shouldn't have to do anything to fight batch on this). The natural thing is for the user to continue using the regenie arguments, and for us to use them as well for the final output.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
akotlar: Having to verify every valid regenie argument combination seems wrong. We're providing a way to run regenie, so I'd like to keep it as open as possible and let regenie complain. I agree with your earlier arguments, wrt lowmem.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
akotlar: I think I answered this above

hail/python/hailtop/batch/contrib/regenie/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+# License path: regenie/LICENSE
+# Thanks to Nathan Weeks for sharing this Dockerfile
+# minor edits were made
+# Filename: Dockerfile
+
+FROM ubuntu:16.04 AS builder
+
+WORKDIR /src
+
+ADD http://code.enkre.net/bgen/tarball/release/v1.1.7 v1.1.7.tgz
```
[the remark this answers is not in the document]
akotlar: Why do they recommend against it?

akotlar:
> Thanks for sharing this detailed plan!
> 
> So, I don't currently have any plan for, I suppose, "applications" that use Batch. Since we plan to support, maintain, and test the Hail Batch regenie implementation, I think it doesn't belong in a "contrib" directory. The hail python package has a `genetics` module for genetics-specific Hail Query functionality, let's mirror that structure. Let's move REGENIE and any non-Python dependencies of it into `hailtop/batch/genetics/regenie`.

Sure

> How is the Dockerfile meant to be used? As written it doesn't appear that it would work because there doesn't exist any regenie source code to COPY in.

It's meant to create a Regenie docker image that we could use. It's a copy of the regenie c++ repo's dockerfile, with the removal of the ENTRYPOINT /usr/local/bin/regenie, so that I could issue a command that included an executable, which is convenient to give me the ability to check that intermediate files are actually created (wc, ls) by batch, and because that seems more idiomatic for batch. I don't think there is a published regenie image, but docker hub is down so can't double check. 

You're right, I shouldn't have deleted the bulk of the repo, kept as is. Didn't want to deal with submodules.

> I've made some other in-line comments in the python file. It's not clear to me how all those other files are related to the python files and I'm a bit uncomfortable adding a whole directory with a LICENSE file, especially when not all the files in the directory fall under that license (e.g. the regenie py file) and moreover the license makes claims about things linking to BGEN, which none of our code here does.

The license is only contained within the folder with the licensed files. In a previous conversation with Nate/Cotton, if we use any open-source software, best to keep those files segregated, alongside their license.

> We have some BGEN files for testing in the Hail src/test/resources.

I already have an example provided. The example folder contains the config for that, and the regenie folder contains the example. We need an example that has a known result, and regenie's c++ repo conveniently provides that. This is what the regenie/regenie folder contains.

hail/python/hailtop/batch/contrib/regenie/Dockerfile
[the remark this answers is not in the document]
akotlar: Since we need to maintain the source files anyway, my plan is to avoid maintaining our own, very similar docker file, and just add the regenie repo as a submodule. We can build using their docker file, and pin to a given commit. There will be no need to check for docker file changes in between commit pins. Would this work for you?

Workflow for batch is something like

git clone --recursive https://github.com/hail-is/hail
# do this step only if we update the submodule
docker build hail/python/hailtop/batch/genetics/regenie/regenie -t gcr.io/hailgenetics/...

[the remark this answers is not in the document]
akotlar: Then there is no need to worry about Dockerfile code style; we're clearly using someone else's docker file (the docker file of the project we're serving in batch), style is not our concern.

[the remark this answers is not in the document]
akotlar: Ugh, unfortunately Batch doesn't seem to allow me to modify how docker run works, which means that any ENTRYPOINT completely fubars Batch. I think we should fix that (override entrypoint), would you accept a PR?

So we need to go back to maintaining a lightly modified docker file. The remote url issue appears fixed using DOCKER_BUILDKIT=1, is the recommended solution by docker folks, and has touted build performance advantages: https://github.com/moby/moby/issues/15717#issuecomment-495426529 , https://docs.docker.com/develop/develop-images/build_enhancements/. Since the solution you provided isn't recommended, would you mind use of buildkit?

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+
+def run_regenie(args):
+    is_local = True if args.local or args.demo else False
+
+    if is_local:
+        backend = hb.LocalBackend()
+        run_opts = {}
+    else:
+        backend = hb.ServiceBackend()
+        run_opts = {"open": True, "wait": True}
```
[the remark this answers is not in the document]
akotlar: removed for now; I'll issue a separate service backend PR once I get that working.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+        raise Exception(f"""More than one input file specified.
+                            Check that only one of --bed or --bgen or --pgen are specified for --step {r.step}
+                        """)
+
+    if r.step == 2:
+        if r.pred:
+            print("--pred provided for --step 2, but Batch will constrain the --pred output prefix to that of --step 1.")
+
+        if r.extract or r.exclude:
+            raise Exception("--extract and --exclude only work with --step 1")
```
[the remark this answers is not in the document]
akotlar: nice

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
```diff
@@ -0,0 +1,293 @@
...
+        r = parser.parse_args(shlex.split(path_or_str))
+
+    else:
+        print(f"Found {path_or_str}, reading")
+
+        with hopen(path_or_str, "r") as f:
+            t = shlex.split(f.read())
+            r = parser.parse_args(t)
+
+    ninputs = sum(i is not None for i in [r.bed, r.bgen, r.pgen])
```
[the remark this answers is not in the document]
akotlar: nice

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
akotlar: Ah, VS Code wasn't readying pylintrc, I see we are using 120 columns

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
akotlar: There is no need for an elif after a continue, it's like elif after return

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
akotlar: > Allowing the user to specify the output file prefix used internally in batch seems unnecessary and appears to add complexity to this implementation?

Regenie expects --out to be respected, so we need to specify this or clobber --out

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
akotlar: I don't think this is needed, shlex.split escapes each value. https://docs.python.org/3/library/shlex.html

hail/python/hailtop/batch/contrib/regenie/Dockerfile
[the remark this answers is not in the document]
danking: I'm very interested in an entry point fix. It's a longstanding bug.

hail/python/hailtop/batch/contrib/regenie/regenie-batch.py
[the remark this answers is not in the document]
danking: When you say print to a directory, do you mean you cannot get batch to copy the resource identified by `j2[j2_out_key]` out of the docker image and into a directory on the host called `args.out`? That should definitely work, I think we have tests for it. Can you describe a bit more the behavior that does happen and the behavior that you expect?

hail/python/hailtop/batch/genetics/regenie/regenie.py
```diff
@@ -0,0 +1,312 @@
...
+    j2.command(f"regenie {cmd2}")
+
+    return j1, j2, step2_output_prefix
+
+
+def run(args):
+    is_local = True if args.local or args.demo else False
+
+    if not is_local:
+        _error("Currently only support LocalBackend (--local)")
```
akotlar, line 266: Will follow up with a ServiceBackend PR that will expose service backend options and will take per-step memory/cpu/storage.

hail/python/test/hailtop/batch/test_regenie.py
```diff
@@ -0,0 +1,65 @@
...
+
+cwd = os.getcwd()
+rdir = "hailtop/batch/genetics/regenie"
+
+
+class LocalTests(unittest.TestCase):
+    def read(self, file):
+        with open(file, 'r') as f:
+            result = f.read().rstrip()
+        return result
```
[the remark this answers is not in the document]
danking: Is the rstrip really necessary? I feel like the batch version of regenie certainly shouldn't add or remove any newlines to the end of a file. I'd expect us to achieve byte-for-byte equivalence.

hail/python/test/hailtop/batch/test_regenie.py
```diff
@@ -0,0 +1,65 @@
...
+        assert(not os.path.isfile(out2))
+
+        os.remove(out_log)
+        os.remove(out1)
+
+        args = br.parse_input_args(["--local", "--step1", "example/step1.txt", "--step2", "example/step2-phenoColList.txt", "--out", out_prefix])
+        br.run(args)
+
+        assert(len(self.read(out2)) > 0)
+        assert(not os.path.isfile(out1))
```
danking, line 60: prefer no parentheses on asserts in Python, assert isn't a function and parentheses tends to encourage `assert(thing, message)` which is always true because `(thing,  message)` is a tuple and `assert some_tuple` always passes.

akotlar:
> A few more thoughts.
> 
> 
> 
> I'm apprehensive to make our build process depend on building some third-party software. I think we should assume a REGENIE binary/image exists. It's fine if we produce it by hand and upload it. We should ask the REGENIE folks to start publishing binaries or images.

As we discussed, I will ask them to produce it. In the meantime I can include the example folder from the regenie repo alone, as before.

hail/python/hailtop/batch/genetics/regenie/regenie.py
```diff
@@ -0,0 +1,310 @@
...
+            cmd1.append(f"--{name} {in_step1[name]}")
+        elif name == "out":
+            cmd1.append(f"--{name} {j1[step1_output_prefix]}")
+        elif isinstance(val, bool):
+            cmd1.append(f"--{name}")
+        elif name == "phenoCol":
+            for pheno in val:
+                cmd1.append(f"--{name} {pheno}")
+        else:
+            cmd1.append(f"--{name} {val}")
```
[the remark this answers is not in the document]
akotlar: > I just read your comments here: https://github.com/hail-is/hail/pull/9194/files/3719e7eef49811ec6149aa4dba0702b893ca89ca#r465142027.
> 
> 
> 
> Resource groups are for a set of files that have the same name but different file extensions. We don't have anything for a set of files that share a prefix but differ in idiosyncratic ways (e.g. end in "_3.loco").
> 
> 
> 
> Given my changes above, if you specify `--out gs://foo/bar/baz` then you'll get these files:
> 
> ```
> 
> gs://foo/bar/baz.log
> 
> gs://foo/bar/baz.pred_list
> 
> gs://foo/bar/baz.1_loco
> 
> gs://foo/bar/baz.2_loco
> 
> ...
> 
> ```
> 
> 
> 
> which, it seems violates your desire to preserve the file structure exactly.
> 
> 
> 
> I think in this circumstance, you'll have a better time using a folder rather than trying to use resource groups. For example, here I use a folder to hold two related files:
> 
> ```python3
> 
> In [17]: b = hb.Batch(backend=hb.ServiceBackend()) 
> 
>     ...: j = b.new_job() 
> 
>     ...: j.command(f'mkdir -p {j.out}') 
> 
>     ...: j.command(f'echo "Hello World" > {j.out}/hi') 
> 
>     ...: j.command(f'echo "Goodbye World" > {j.out}/bye') 
> 
>     ...: b.write_output(j.out, 'gs://danking/test123') 
> 
>     ...: b.run()                                                                                                                                                                                                                                     
> 
> ```
> 
> ```bash
> 
> # gsutil ls gs://danking/test123
> 
> gs://danking/test123/bye
> 
> gs://danking/test123/hi
> 
> ```
> 
> 
> 
> Something like this should work for regenie.
> 
> ```python3
> 
> j1.command(f'mkdir {j.output_dir}')
> 
> 
> 
> # ...
> 
>         elif name == "out":
> 
>             cmd1.append(f'--out {j.output_dir}/out')
> 
>             batch.write_output(j.output_dir, val)
> 
> ```
> 
> 
> 
> If I run this with `--out gs://danking/regenie-run123` I'll get:
> 
> ```
> 
> gs://danking/regenie-run123/out.log
> 
> gs://danking/regenie-run123/out_pred.list
> 
> gs://danking/regenie-run123/out_1.loco
> 
> ...
> 
> ```
> 
> Which seems fine?

> Resource groups are for a set of files that have the same name but different file extensions. We don't have anything for a set of files that share a prefix but differ in idiosyncratic ways (e.g. end in "_3.loco").

This doesn't make sense to me. It should be for a set of output files that are controlled by a single parameter. That is the case here: --out specifies the prefix of all our files. These files differ however by more than extension.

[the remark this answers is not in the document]
akotlar: > Which seems fine?

No, I want no out prefix. There is no way to tel regenie "don't output a file with a name", and batch always appends the identifier as an extension, which seems an unnecessary restriction.

hail/python/hailtop/batch/genetics/regenie/regenie.py
```diff
@@ -0,0 +1,310 @@
...
+        elif isinstance(val, bool):
+            cmd2.append(f"--{name}")
+        elif name == "phenoCol":
+            for pheno in val:
+                cmd2.append(f"--{name} {pheno}")
+        else:
+            cmd2.append(f"--{name} {val}")
+
+    if not step2_args.ignore_pred:
+        cmd2.append(f"--pred {j1[step1_output_prefix]['pred_list']}")
```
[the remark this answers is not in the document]
akotlar: If you read the documentation, this is exactly the case. There is no point specifying --press if you know --out in step 1.

hail/python/hailtop/batch/genetics/regenie/regenie.py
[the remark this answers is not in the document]
akotlar: Also the output you described is not correct. I need {out_prefix}_Y1.regenie, not. {out_prefix}.Y1.regenie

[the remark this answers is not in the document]
akotlar: > I still don't really grok the input munging stuff, but the following seems somewhat easier to follow and eliminates the need for `get_input` and the preprocessing of the output files:
> 
> ```python
> 
>  def prepare_jobs(batch, args: BatchArgs, step1_args: argparse.Namespace, step2_args: argparse.Namespace):
> 
>     j1 = batch.new_job(name='run-regenie')
> 
>     # ...
> 
>     phenos = get_phenos(step1_args)
> 
>     nphenos = len(phenos)
> 
> 
> 
>     cmd1 = []
> 
>     for name, val in vars(step1_args).items():
> 
>         if name in from_underscore:
> 
>             name = from_underscore[name]
> 
> 
> 
>         if val is None or val is False:
> 
>             pass
> 
>         elif name == "bed":
> 
>             bed_files = batch.read_input_group(bed=f"{val}.bed", bim=f"{val}.bim", fam=f"{val}.fam")
> 
>             cmd1.append(f'--bed {bed_files}')
> 
>         elif name == "pgen":
> 
>             pgen_files = batch.read_input_group(pgen=f"{val}.pgen", pvar=f"{val}.pvar", psam=f"{val}.psam")
> 
>             cmd1.append(f'--pgen {pgen_files}')
> 
>         elif name in input_file_args:
> 
>             input_file = batch.read_input(val)
> 
>             cmd1.append(f'--{name} {input_file}')
> 
>         elif name == "out":
> 
>             j1.declare_resource_group(
> 
>                 output={'log': '{root}.log',
> 
>                         'pred_list': '{root}_pred.list',
> 
>                         **{f'{i}_loco': f'{{root}}_{i}.loco' for i in range(i, nphenos + 1)}}),
> 
>             cmd1.append(f'--out {j1.output}')
> 
>             batch.write_output(j1.output, val)
> 
>         elif name == "lowmem_prefix":
> 
>             j1.declare_resource_group(
> 
>                 lowmem={f'l0_Y{i}': f'{{root}}_l0_Y{i}'})
> 
>             cmd1.append(f'--lowmem_prefix {j1.lowmem}')
> 
>             batch.write_output(j1.lowmem, val)
> 
>         elif isinstance(val, bool):
> 
>             cmd1.append(f"--{name}")
> 
>         elif name == "phenoCol":
> 
>             for pheno in val:
> 
>                 cmd1.append(f"--phenoCol {pheno}")
> 
>         else:
> 
>             if name == "step":
> 
>                 assert val == "1"
> 
>             cmd1.append(f"--{name} {val}")
> 
> ```
> 
> 
> 
> I think I don't really understand the meaning of `--out FOO` as an argument to batch regenie. Presumably `FOO` is some GCS path to which we should copy step1's output? That's what I implemented above. Same for `--lowmem-prefix FOO`. Prefix doesn't really sound right though? It's just the remote location at which to store the lowmem files.

Batch --out is a remote location, yes. Regenie step 1/2 --out FOO is where regenie expects to write files, locally. Same for lowmem prefix. It's not a remote location (unless you mean remote as in something remote that emulates locally attached block storage). 

Note that regenie can run a single step, and users may want to get intermediate files. We can avoid duplicating interfaces by having the regenie basename named in step1/2 and the output directory provided in batch args. I had previously called this --outdir, switched because batch write_output is broken.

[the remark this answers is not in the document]
akotlar: > I still don't really grok the input munging stuff, but the following seems somewhat easier to follow and eliminates the need for `get_input` and the preprocessing of the output files

I don't see the benefit here. The input function creates the necessary resources (every argument it accepts is one that specifies a potential input file), then those resources are used as part of a short procedure to generate the string command.

hail/python/test/hailtop/batch/test_regenie.py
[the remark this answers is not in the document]
akotlar: agreed

hail/python/hailtop/batch/genetics/regenie/regenie.py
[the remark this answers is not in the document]
danking: What do you mean by "if you know --out in step 1"? As written, with `--ignore-pred`, this batch will not copy any files from step 1 to step 2, in fact batch will run these two jobs in parallel. That seems wrong to me, but maybe I misunderstand how REGENIE's step 2 depends (or doesn't) on step 1.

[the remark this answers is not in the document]
akotlar: --pred is a file that is generated by step 1, prefixed with whatever you passed to step 1 as --out.

akotlar:
Now uses hailgenetics/regenie image.

akotlar:
Issue created for Regenie: https://github.com/rgcgithub/regenie/issues/23

akotlar:
Getting some weird CI errors: the hail module isn't found

akotlar:
It seems the biggest remaining issue is that CI doesn't have access to hail imports when testing hailtop. Could we address this? Do you want a PR?

akotlar:
Ok, updated the test_batch and test_hailtop_batch to use run_image, which is a superset of service_image, that should fix the import issue.

hail/python/hailtop/batch/genetics/regenie/regenie.py
```diff
@@ -0,0 +1,310 @@
...
+
+    cmd2 = f"--step 2 {' '.join(cmd2)}"
+
+    j2.command(f"regenie {cmd2}")
+
+    return j1, j2, step2_output_prefix
```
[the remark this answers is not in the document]
akotlar: The file names, I think, in your proposal won't be right. I want the output prefix that the user specifies to be respected, and have the final output able to be written to a folder with the name unmodified. This will take a small followup pr.

akotlar:
bump, would love to get this in today

akotlar:
Dan, I ended up revisiting your output suggestion, after realizing that I could eventually solve the output issue by changing how write_output concatenates identifiers with the prefix. This would be a smaller change than the original thought, which was to allow write_output to treat the output path as a directory and write the identifier unchanged. I still think this would be more what users would expect, but going this route is a bigger deviation so from a project management standpoint less desirable.

I think you will now be happy with the code.

build.yaml
```diff
@@ -1832,14 +1832,19 @@ steps:
  - kind: runImage
    name: test_hailtop_batch
    image:
-     valueFrom: service_base_image.image
+     valueFrom: hail_run_image.image
    script: |
      set -ex
+     cd /io
+     tar xvf wheel-container.tar
+     python3 -m pip install --no-dependencies hail-*-py3-none-any.whl
```
danking, line 1840: I do not want `hailtop.batch` to require a user to have the JVM installed (which `hl.hadoop_open` and `hl.hadoop_exists` require).

The tool we have available is `hailtop.google_storage.GCS`. GCS does not implement a file system API, but it can check existence and read `gs://` URLs.

hail/python/hailtop/batch/__init__.py
```diff
@@ -1,15 +1,17 @@
 import nest_asyncio  # type: ignore
 
-from .batch import Batch
+from .batch import Batch, Resource
 from .batch_pool_executor import BatchPoolExecutor
 from .backend import LocalBackend, ServiceBackend
 from .utils import BatchException
 
 __all__ = ['Batch',
+           'Resource',
```
danking, line 9: I prefer we do not directly expose `Resource` to our users. Moreover, the fact `Resource` is in the scope of `.batch` is an implementation detail. It properly resides `resource`.

Instead, let's reference it in the tests as `from hailtop.batch.resource import Resource`.

hail/python/test/hailtop/batch/test_regenie.py
```diff
@@ -0,0 +1,83 @@
...
+def read(file):
+    with open(file, 'r') as f:
+        return f.read()
+
+
+def assert_same_file(file1, file2):
+    assert read(file1) == read(file2)
+
+
+class LocalTests(unittest.TestCase):
```
danking, line 19: These tests do not enable continuous integration because they depend on docker. Regenie, due to its use of custom images, cannot use LocalBackend in CI.

Let's remove all the `skipIf`, change the name to BatchTests, and follow the style of test_batch.py's BatchTests (which really should be called ServiceTests).

danking:
Please also delete `.gitmodules`

build.yaml
akotlar: I don't understand. regenie.py needs to read both local and remote files.

hail/python/test/hailtop/batch/test_regenie.py
akotlar: This PR does not have a service backend version. These tests do not block CI, they are just skipped

build.yaml
akotlar: This also seems like a far afield problem given that we require users to install hail to get hailtop currently. https://hail.is/docs/batch/getting_started.html

akotlar:
I've realised that hailtop is not the place for contributed methods that use batch if we want to restrict hail imports. Think about how strange it is to exclude our own Batch cookbook example (gwas clumping) from a viable contributed module.

danking:
Sorry, I wasn't clear before. The Batch LD Clumping example does not require Hail to be installed on *the computer that submits the batch*. Hail is imported and used inside of the Batch task that performs GWAS. That task runs inside a Docker container that has Hail installed (its derived from `hailgenetics/hail`).

I'm hesitant to make the *submission* of a batch dependent on the Hail Query library. Particularly when we have relatively low-effort alternative approaches. I'm delighted any time I see batch tasks use Hail Query! Konrad's Pan UKB work also does this.

build.yaml
danking: By normal filesystem operations, I mean `basedir`, `dirname`, and friends. AFAICT, the only operations you need are "does XXX exist" and "give me XXX as a string". Upon a second look, I realize that GCS hides all its synchronous operations as underscore-prefixed-names. Luckily, it's pretty simple to use the synchronous google cloud storage library directly:

```python3
from google.cloud import storage
from google.cloud.storage.blob import Blob
client = storage.Client()
blob = Blob.from_string("gs://danking/a", client)
if blob.exists():
    step = blob.download_as_string()
```

akotlar:
Sounds good, thanks Dan!

akotlar:
> Sorry, I wasn't clear before. The Batch LD Clumping example does not require Hail Query (and, more importantly, a JVM) to be installed on *the computer that submits the batch*. Hail is imported and used inside of the Batch task that performs GWAS. That task runs inside a Docker container that has Hail installed (its derived from `hailgenetics/hail`).
> 
> 
> 
> I'm hesitant to make the *submission* of a batch dependent on the Hail Query library. Particularly when we have relatively low-effort alternative approaches. I'm delighted any time I see batch tasks use Hail Query! Konrad's Pan UKB work also does this.

Right I realize this, why I mentioned unit test (to make unit tests work without calling hail, we would need to wrap each function we want to test in a cli interface, and then containerise those functions...this seems not so friendly to contributors). A 2nd question: how would a contributor submit write hybrid hail query/batch code that called batch from within a hail query script? So hail transforms data, say generates pc's, writes table, and then issues some batch commands that use a pre-built image, as opposed to requiring them to containerise their hail cod3. This will be useful: containerisation is the stated biggest pain point of batch, from my interviews. And one of the biggest hail issues is shuttling data between hail and other processes, which has led people to ask for tighter integration, or more statistical tools (so that they don't need to go out to those other tools, like plink). Showing Batch working from within hail-query scripts would be useful because it will show a much easier integration, and those examples or contributed modules should live somewhere (and that seems not to be hailtop, because this requires batch to be called from code that is calling hail directly, rather than through an image).

danking:
I don't understand your first sentence.

In the case of a hybrid query and batch script, using Hail Query is totally kosher! What I don't want is dependence on a JVM and Hail Query just to check if a file exists in Google Cloud Storage for a script that is otherwise independent of Hail Query.

akotlar:
> I don't understand your first sentence.

What didn't you understand? I'd like to clarify if that would be helpful.

> In the case of a hybrid query and batch script, using Hail Query is totally kosher! What I don't want is dependence on a JVM and Hail Query just to check if a file exists in Google Cloud Storage for a script that is otherwise independent of Hail Query.

So are you ok with keeping the build.yml changes (which would allow hailtop to have something within in that called hail query without being called from batch as a docker image + command line arguments).

build.yaml
akotlar: I wrote a FS abstraction for Hail, and a concrete implementation GoogleCloudStorageFS last year. This does what I need in a 2 line change. To avoid transitive dependencies from requiring JVM, I moved two imports into the scope they're used in in FS.

akotlar: Re-reading your last comment, outside of this thread, I see you also don't want to import hail query in any way.

hail/python/hailtop/batch/genetics/regenie/regenie.py
```diff
@@ -0,0 +1,314 @@
...
+    if _is_local(spath):
+        with open(spath, "r") as f:
+            return f.read()
+
+    client = storage.Client()
+    blob = Blob.from_string(spath, client)
+    return blob.download_as_string()
+
+
+def _read_first_line(spath: str):
```
akotlar, line 40: This is a waste of code to maintain. GoogleCloudStorageFS is available to do exactly what I'm replicating here.

akotlar:
Should be ready to review

PR #9195, opened 2020-07-31
[hail] check hail using 3.6 and 3.7

danking:
CC: @tpoterba I've fixed up CI to check hail and hailtop under both python 3.6 and 3.7. There were a few minor fixes to hail.


tpoterba:
seems fine to me

danking:
bump @jigold

PR #9196, opened 2020-07-31
[dev-docs] Describe TLS

danking:
This is an adaptation of my comment on the TLS PR.

I moved the old `tls.md` to `tls-cookbook.md`. Git doesn't realize that.

Dania & @catoverdrive, y'all are probably the two folks most likely to benefit from tls.md, so I'd appreciate your comments on the readability of this document.


danking:
bump :)

PR #9198, opened 2020-07-31
[hailtop][batch] Use autodoc for parameter types

danking:
cc: @tpoterba

There's a Sphinx extension that uses Python type annotations to figure out
parameter types so we don't have to write it twice. I implement that here. I
wrestled for a while with the cyclical import structure until I realized that

```
from . import module
```

side steps a lot of the cyclical issues (pylint still dislikes it).

I also added `sphinx.ext.intersphinx` which makes every reference to a Python
standard library thing work correctly. Now when I write
`concurrent.futures.ThreadPoolExecutor` it links right to the Python docs
page. It's really quite nice.


danking:
bummppp

danking:
fixed conflicts

hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -84,31 +84,31 @@ class BatchPoolExecutor:
 
     Parameters
     ----------
-    name: :obj:`str`, optional
+    name:
```
jigold, line 87: I think you're missing the elimination of parameter types and return types in job.py.

hail/python/hailtop/batch/batch.py
```diff
@@ -309,36 +295,36 @@ def write_output(self, resource: Resource, dest: str) -> None:  # pylint: disabl
...
         Parameters
         ----------
-        resource: :class:`.ResourceFile` or :class:`.ResourceGroup`
+        resource:
             Resource to be written to a file.
-        dest: :obj:`str`
+        dest:
             Destination file path. For a single :class:`.ResourceFile`, this will
-            simply be `dest`. For a :class:`.ResourceGroup`, `dest` is the file
+            simply be `dest`. For a :class:`._resource.ResourceGroup`, `dest` is the file
```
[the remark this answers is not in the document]
danking: OK, fixed, that was a ham-fisted rename. Now its just `.ResourceGroup`

hail/python/hailtop/batch/batch.py
```diff
@@ -269,30 +259,26 @@ def read_input_group(self, **kwargs: str) -> ResourceGroup:
 
         The file extensions for each file are derived from the identifier.
         This is equivalent to `"{root}.identifier"` from
-        :meth:`.Job.declare_resource_group`. We are planning on adding flexibility
+        :meth:`.job.Job.declare_resource_group`. We are planning on adding flexibility
```
jigold, line 262: The ResourceFile.add_extension isn't getting picked up properly.

danking:
@jigold I added a commit with many changes. Sorry. There were several broken links. I fixed all of them and enabled `nitpicky` which forces Sphinx to throw an error on all broken references. I also removed an unnecessary `rm -rf` from the Makefile.

PR #9199, opened 2020-07-31
Java11 Scala2.12 Spark3.0.0 changes

k2sriram96:


hail/src/main/scala/is/hail/linalg/BlockMatrix.scala
```diff
@@ -117,7 +117,7 @@ object BlockMatrix {
       val iOffset = i * blockSize
       val jOffset = j * blockSize
 
-      HailContext.backend.broadcast(lm(iOffset until iOffset + blockNRows, jOffset until jOffset + blockNCols).copy)
+      HailContext.backend.broadcast(lm(iOffset until iOffset + blockNRows, jOffset until jOffset + blockNCols))
```
k2sriram96, line 120: Copy function not working as expected --> applies operation for first block in block matrix and returns the result of that for all blocks.

Could be related to this:
https://github.com/scalanlp/breeze/issues/772

johnc1231:
Closing this as will not merge. If you want to make a PR with some of the individual backwards compatible fixes you made, that would be welcome.

PR #9200, opened 2020-07-31
[address] retry address

danking:
This PR changes the `addresses` function on `DeployConfig` to retry all transient errors. In particular, if the address service is temporarily down (maybe its getting redeployed), this change allows the client to repeatedly retry until the address service comes back to life.

I also added some type annotations to `retry_transient_errors`. It takes a function that returns something we can `await` and then applies that function in a loop until it does not raise an error.

### DIFF
--- a/hail/python/hailtop/config/deploy_config.py
+++ b/hail/python/hailtop/config/deploy_config.py
@@ -5,7 +5,7 @@ import os
 import json
 import logging
 from aiohttp import web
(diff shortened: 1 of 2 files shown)


jigold:
@Dania-Abuhijleh I'll let you take a look, ask any questions you might have, and then approve / request changes.

danking:
bump 🙂

PR #9205, opened 2020-08-03
[query] Remove PType.copyFromPValue, replace with PCode.castTo

chrisvittal:
This is the first in a series of changes that will push PCode through more of
the PType construction interfaces. The intent is to reduce code duplication,
and have as few interfaces as possible where `Code[_]` or `Code[Long]`
represents a hail type.


chrisvittal:
bump

PR #9208, opened 2020-08-03
[query] Move typeToTypeInfo and defaultValue to ptypes package

chrisvittal:
Felt wrong for the ptypes to typeinfo and default values being in the ir package.


chrisvittal:
bump

PR #9209, opened 2020-08-03
NDArray Sum Aggregator

johnc1231:
The ability to sum over many ndarrays with an aggregator is necessary for several distributed matmul like operations. We'll probably want something more general in the future, but for now this should suffice. This PR:

- Adds `hl.agg.ndarray_sum`, a python aggregator that sums over ndarrays of the same shape
- The corresponding Scala classes, mainly `NDArraySumAggregator`
- `PNDArrayValue.sameShape`, a helper function that checks if two `PNDArray`s are the same shape. 

Future work:

- Add an optional initOp that specifies a shape in cases when we want the sum of no elements to be an ndarray of 0s. 

@danking

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/agg/NDArraySumAggregator.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/agg/NDArraySumAggregator.scala
@@ -18,7 +18,7 @@ class NDArraySumAggregator (ndTyp: PNDArray, knownShape: Option[IR]) extends Sta
 
   override def resultType: PType = ndTyp
 
-  override def initOpTypes: Seq[PType] = Array(ndTyp.shape.pType)
+  override def initOpTypes: Seq[PType] = Array[PType]()//Array(ndTyp.shape.pType)
 
   override def seqOpTypes: Seq[PType] = Array(ndTyp)
 
@@ -40,10 +40,10 @@ class NDArraySumAggregator (ndTyp: PNDArray, knownShape: Option[IR]) extends Sta
   override protected def _seqOp(cb: EmitCodeBuilder, state: State, seq: Array[EmitCode]): Unit = {
     // Check if the thing is initialized. If it is, then check shapes match. If they do, add ndarrays.
     val Array(nextNDCode) = seq
-    cb.ifx(isInitialized(state),
-      {
-        nextNDCode.toI(cb).consume[Unit](cb, {}, { case nextNDPCode: PNDArrayCode =>
-          val nextNDPValue = nextNDPCode.memoize(cb, "ndarray_sum_seqop_next")
+    nextNDCode.toI(cb).consume(cb, {}, {case nextNDPCode: PNDArrayCode =>
+      val nextNDPValue = nextNDPCode.memoize(cb, "ndarray_sum_seqop_next")
+      cb.ifx(isInitialized(state),
+        {
           val currentNDPValue = PCode(ndTyp, ndArrayPointer(state)).asNDArray.memoize(cb, "ndarray_sum_seqop_current")
 
           val sameShape = currentNDPValue.sameShape(nextNDPValue, cb.emb)
@@ -72,13 +72,14 @@ class NDArraySumAggregator (ndTyp: PNDArray, knownShape: Option[IR]) extends Sta
             columnMajorLoops,
             Code._fatal[Unit]("Can't sum ndarrays of different shapes.")
           ))
-        })
-      },
-      {
-        // TODO Is this really safe, does it copy the initial array?
-        ndTyp.constructAtAddress(cb.emb, stateType.fieldOffset(state.off, 1), state.region, nextNDCode.pt, nextNDCode.value[Long], true)
-      }
-    )
+        },
+        {
+          // TODO Is this really safe, does it copy the initial array?
+          ndTyp.constructAtAddress(cb.emb, stateType.fieldOffset(state.off, 1), state.region, nextNDCode.pt, nextNDPValue.get.tcode[Long], true)
+        }
+      )
+    })
+
   }
 
   override protected def _combOp(cb: EmitCodeBuilder, state: State, other: State): Unit = {
--- a/hail/src/main/scala/is/hail/expr/ir/agg/NDArraySumAggregator.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/agg/NDArraySumAggregator.scala
@@ -75,10 +75,10 @@ class NDArraySumAggregator (ndTyp: PNDArray, knownShape: Option[IR]) extends Sta
       )
     }
 
-    cb.ifx(sameShape,
+    cb.append(sameShape.mux(
       columnMajorLoops,
-      Code._fatal("Can't sum ndarrays of different shapes.")
-    )
+      Code._fatal[Unit]("Can't sum ndarrays of different shapes.")
+    ))
 
   }
 
--- a/hail/src/main/scala/is/hail/expr/ir/agg/NDArraySumAggregator.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/agg/NDArraySumAggregator.scala
@@ -2,17 +2,12 @@ package is.hail.expr.ir.agg
 
 import is.hail.annotations.{Region, StagedRegionValueBuilder}
 import is.hail.asm4s._
-import is.hail.expr.ir.functions.StringFunctions
-import is.hail.expr.ir.{EmitCode, EmitCodeBuilder, EmitRegion, coerce}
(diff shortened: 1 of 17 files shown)


johnc1231:
Sure. Any size preference? I could do the sum of 100 2ds with dims 4096 x 4096 ndarrays for block matrix similarity.

danking:
Just big enough to get consistent timing numbers (probs easiest to check this on your laptop for now), imo. Also very curious to hear how master compares to this PR.

johnc1231:
Doing 100 4096 x 4096s on my laptop took 1min20s. 

```
In [3]: %time ht.aggregate(hl.agg.ndarray_sum(ht.nd))
[Stage 1:====================================================>    (11 + 1) / 12]CPU times: user 2.74 s, sys: 554 ms, total: 3.29 s
Wall time: 1min 20s
Out[3]:
array([[100., 100., 100., ..., 100., 100., 100.],
       [100., 100., 100., ..., 100., 100., 100.],
       [100., 100., 100., ..., 100., 100., 100.],
       ...,
       [100., 100., 100., ..., 100., 100., 100.],
       [100., 100., 100., ..., 100., 100., 100.],
       [100., 100., 100., ..., 100., 100., 100.]])
```

johnc1231:
Your hacky version took 1min41 seconds on my laptop. Not as big a difference as I'd hoped.

johnc1231:
Ah, I bet I'm hitting the same problem I hit with the original `NDArrayEmitter`, in that my nested loops aren't getting jitted. I'll see if getting them in a separate method improves the situation.

johnc1231:
Ok, so Cotton's new thing means emitting separate methods by hand is not a thing we do anymore. But there are two factors hurting the benchmark. 

One is that the benchmark is hiding the fact that we are spending ~25 seconds serializing and de-serializing JSON for this ndarray. So the real comparison is more like 55 seconds vs 75 seconds, which is a roughly 25% speed improvement. 

The other is that `hl.nd.ones` is just an alias for `hl.nd.array(hl.range(shape_product)).map(lambda x: 1).reshape((n_rows, n_cols))`. This is going to create a bunch of row major data, copy it to column major in a pretty cache inefficient way during the reshape, then do the additions. So that's eating some of the time too. We should probably have a way for all the constant methods to not go through regular array. 

Anyway, 25% improvement + better interface is a win for now, we can revisit ways to make this faster in the future.

tpoterba:
you can definitely still generate methods by hand!

johnc1231:
I talked to Cotton about it and he said not to. But it's not clear how much of a difference that makes yet anyway. I think this version is pretty good and an improvement. Plus it'll add a benchmark which we can work on optimizing.

hail/src/main/scala/is/hail/types/physical/PNDArray.scala
```diff
@@ -54,7 +56,11 @@ abstract class PNDArray extends PType {
 abstract class PNDArrayValue extends PValue {
   def apply(indices: IndexedSeq[Value[Long]], mb: EmitMethodBuilder[_]): Value[_]
 
+  override def pt: PNDArray = ???
```
patrick-schultz, line 59: Why not leave abstract?

hail/src/main/scala/is/hail/expr/ir/agg/NDArraySumAggregator.scala
```diff
@@ -0,0 +1,125 @@
...
+class NDArraySumAggregator (ndTyp: PNDArray) extends StagedAggregator {
+  override type State = TypedRegionBackedAggState
+
+  // State needs:
+  // 1. Initialized or not
+  // 2. The ndarray itself.
+
+  override def resultType: PType = ndTyp
+
+  val stateType = PCanonicalTuple(true, PBooleanRequired, ndTyp)
```
patrick-schultz, line 19: Because the seqOp skips missing values, the ndarray field can never be missing if the state is initialized, right? Is there a reason not to just let a missing ndarray be the uninitialized state? Then the result is exactly the state.

If not, up to you if you want to change it. I think it's a bit cleaner, but not a big difference.

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
```diff
@@ -126,6 +126,10 @@ final case class PCanonicalNDArray(elementType: PType, nDims: Int, required: Boo
     ))
   }
 
+  def mutateElement(indices: IndexedSeq[Value[Long]], ndAddress: Value[Long], newElement: Code[_], mb: EmitMethodBuilder[_]): Code[Unit] = {
```
patrick-schultz, line 129: I think I'd expect this to be called `setElement`. What do you think?

hail/src/main/scala/is/hail/types/physical/PNDArray.scala
johnc1231: This lets me set the return type for things that override it so they don't have to be cast.

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
johnc1231: Done

johnc1231:
Addressed all comments, should be good to go. 

One question I did have is that I have to specify a region size in this aggregator, and I picked `TINY`. Does it matter? `TINY` does seem small compared to the kinds of operations that we'll do on dndarray and block matrix

johnc1231:
Added new test that checks summing mix of regular and transposed data to make sure that future changes respect striding.

patrick-schultz:
> Addressed all comments, should be good to go.
> 
> One question I did have is that I have to specify a region size in this aggregator, and I picked `TINY`. Does it matter? `TINY` does seem small compared to the kinds of operations that we'll do on dndarray and block matrix

That specifies the granularity of allocations. Each time the region needs more space than it has allocated, it allocates a new block of size determined by the size parameter. Unless it's trying to make a single allocation larger than the block size, in which case it allocates a block of exactly the desired size. The ideal thing here would be to make a single block with exactly the size needed for the state, but the current Region interface doesn't support that. The closest we can get is using the smallest block size, which is `TINIER`. Then the missing bit and pointer to the data would go in the initial tiny block, and the data would go there too if it's small enough, otherwise it would go in an allocation of exactly the right size.

hail/src/main/scala/is/hail/expr/ir/agg/NDArraySumAggregator.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  override def initOpTypes: Seq[PType] = Array[PType]()
+
+  override def seqOpTypes: Seq[PType] = Array(ndTyp)
+
+  val ndarrayFieldNumber = 0
+
+  def isInitialized(state: State): Code[Boolean] = {
+    stateType.isFieldDefined(state.off, ndarrayFieldNumber)
+  }
```
patrick-schultz, line 29: Now that you're just using the missing bit, you shouldn't need this or `ndArrayPointer`. Instead of
```
cb.ifx(isInitialized(state), {
  val currentNDPValue = PCode(ndTyp, ndArrayPointer(state)).asNDArray.memoize(cb, "ndarray_sum_seqop_current")
  ...
}, {
  // uninitialized case
})
```
you can simplify to
```
val statePV = new PCanonicalBaseStructSettable(stateType, state.off)
...
statePV.loadField(ndarrayFieldNumber).consume(cb, {
  // uninitialized case
}, { currentNDPCode =>
  val currentNDPValue = currentNDPCode.asNDArray.memoize(cb, "ndarray_sum_seqop_current")
  ...
})
```
A bit simpler and more idiomatic.

johnc1231:
Thanks for pushing me to clean that up with the `consume` infrastructure, much neater / shorter. 

However, I can't seem to write a working `_result` function in that style. I tried several different ways, all led to segfaults, not sure why. I ended up just inlining calls to `isFieldDefined` and `PBaseStruct.loadField` for that one, as I didn't want to spend more time digging into it.

patrick-schultz:
Weird. Did you do something like this?
```
val statePV = new PCanonicalBaseStructSettable(stateType, state.off)
statePV.loadField(ndarrayFieldNumber)
       .consume(cb, srvb.setMissing(), srvb.addIRIntermediate(_, deepCopy = true))
```
Anyways, happy to approve if you don't want to mess with it.

johnc1231:
Yup, I did that, and I also tried a version where I did `state.get` to get the emit code, then `toI.consume`ed that, and used that to get the pvalue I loaded the field of, and that also didn't work. I'm going to pick not to mess with it.

hail/src/main/scala/is/hail/types/physical/PNDArray.scala
akotlar: Tim and I previously handled this using 

```scala
private def ptND: PNDArray
override def pt: PNDArray = ptND
```

PR #9211, opened 2020-08-04
[Benchmark] submit button read JSON from google bucket

Dania-Abuhijleh:


benchmark-service/benchmark/benchmark.py
```diff
@@ -16,6 +20,65 @@
...
+        log.info('could not get blob: ' + message, exc_info=True)
+        raise web.HTTPBadRequest(text=message)
+
+    # with open('file-to-download-to') as file_obj:
+    #     storage_client.download_blob_to_file(file_path, file_obj)
+
+    # with open(file_path) as f:
+    #     pre_data = json.load(f)
+
+    x = re.findall('.*/+(.*)-(.*)-(.*)?\.json', file_path)  #was file_path
```
jigold, line 55: What I'd do in this case if there are multiple formats that can match is either

1. Make two regexes and then check the longer one first to see if that matches and if not check the shorter one. 

2. My preference here would be to use named capture groups and then use the parentheses to group as well. Something like this (you'll need to test this and fix it potentially): `'gs://hail-benchmarks/(?P<user>[^/]+)/)(?P<version>[^-]+)-)(?P<sha>[^-]+)-)(?P<tag>)?\.json'`

This is an example of how we usually do this. Make a compiled REGEX as a global variable.
```
DISK_TYPE_REGEX = re.compile('(projects/(?P<project>[^/]+)/)?zones/(?P<zone>[^/]+)/diskTypes/(?P<disk_type>.+)')
```

and then use it in code similar to this:

```
def parse_disk_type(name):
    match = DISK_TYPE_REGEX.fullmatch(name)
    return match.groupdict()
```

benchmark-service/benchmark/benchmark.py
```diff
@@ -33,15 +96,50 @@ async def greet_user(request: web.Request) -> web.Response:
...
     }
+    benchmarks = get_benchmarks(filepath)
     response = aiohttp_jinja2.render_template('index.html', request,
-                                              context=context)
+                                              context=benchmarks)
+    return response
+
+
+@router.post('/lookup')
+@web_authenticated_developers_only(redirect=False)
```
jigold, line 135: Here's you'd want to add the decorator `@check_csrf_token`. Take a look at what this does in gear.

benchmark-service/benchmark/templates/index.html
```diff
@@ -1,11 +1,56 @@
...
-Posts index
-{% endblock %}
+{% block title %} Benchmark {% endblock %}
 
 {% block content %}
 <h1>Benchmark System</h1>
+
+<form action="{{ base_path }}/dabuhijl/benchmark/lookup" method="post">
+    <input type="text" name="file"></input>
+    <input type="submit" value="lookup"></input>
```
jigold, line 9: So the way CSRF protection works as you have a hidden field here that expects there's a context field `csrf_token`.

```
<input type="hidden" name="_csrf" value="{{ csrf_token }}">
```

benchmark-service/benchmark/benchmark.py
```diff
@@ -33,15 +96,50 @@ async def greet_user(request: web.Request) -> web.Response:
...
+async def lookup(request, userdata):  # pylint: disable=unused-argument
+    data = await request.post()
+    file = data['file']
+    global filepath
+    filepath = file
+    benchmarks = get_benchmarks(file)
+    response = aiohttp_jinja2.render_template('index.html', request, context=benchmarks)
```
jigold, line 142: Instead of the aiohttp_jinja2 `render_template`, you'll want our custom `render_template` in `web_common/web_common/web_common.py`. This will automatically set both the session cookie csrf token and the matching one in the jinja context.

PR #9213, opened 2020-08-04
[hailtop] not every ClientConnectorError has an os_error

danking:


danking:
@catoverdrive Assigned you since it will be good to get you acclimated to the services team code. We have a function called `is_transient_error` which checks if an error is a known-to-be-transient error. An error is transient if perpetually retrying the error-raising-operation will eventually lead to success.

We consider most things transient, even 500 Internal Server Error, because we expect our systems work but sometimes fail under heavy load. Eventually load should drop.

danking:
bump

PR #9215, opened 2020-08-04
[Benchmark] read JSON file from google bucket 

Dania-Abuhijleh:


benchmark-service/benchmark/benchmark.py
```diff
@@ -2,19 +2,78 @@
...
+def parse_file_path(name):
+    match = FILE_PATH_REGEX.fullmatch(name)
+    return match.groupdict()
+
+
+def get_benchmarks(file_path):
+    # create storage client
+    storage_client = storage.Client()
+    # get bucket with name
+    bucket = storage_client.get_bucket('hail-benchmarks')
```
jigold, line 36: This is great, but let's make a couple of changes.

1. Let's create a new file called `utils.py`. In it, let's create a class that represents reading from google storage. An attribute of the class should be the client. Then we want a method that reads an arbitrary file from google storage and returns the data as a string. You'll probably want to write a parser that figures out the bucket and then the rest of the path from a file path.

benchmark-service/benchmark/benchmark.py
```diff
@@ -2,19 +2,78 @@
...
+        # convert to string
+        json_data = blob.download_as_string()
+        pre_data = json.loads(json_data)
+    except Exception:
+        message = 'could not find file'
+        log.info('could not get blob: ' + message, exc_info=True)
+        raise web.HTTPBadRequest(text=message)
+
+    # x = re.findall('.*/+(.*)-(.*)-(.*)?\.json', file_path)
+    # sha = x[0][1]
```
jigold, line 49: delete

benchmark-service/benchmark/benchmark.py
```diff
@@ -2,19 +2,78 @@
...
+
+    file_info = parse_file_path(file_path)
+    sha = file_info['sha']
+
+    data = list()
```
jigold, line 54: I think the canonical python is to do `data = []`

benchmark-service/benchmark/benchmark.py
```diff
@@ -2,19 +2,78 @@
...
+        if not d['failed']:
+            prod_of_means *= d['mean']
+            stats['f-stat'] = round(d['f-stat'], 6)
+            stats['mean'] = round(d['mean'], 6)
+            stats['median'] = round(d['median'], 6)
+            stats['p-value'] = round(d['p-value'], 6)
+            stats['stdev'] = round(d['stdev'], 6)
+            stats['times'] = d['times']
+        data.append(stats)
+    geometric_mean = prod_of_means ** (1.0 / len(pre_data['benchmarks']))
```
jigold, line 69: Let's move this equation to the utils.py file and make it a function that you can call.

benchmark-service/benchmark/benchmark.py
```diff
@@ -2,19 +2,78 @@
...
 router = web.RouteTableDef()
 logging.basicConfig(level=logging.DEBUG)
 deploy_config = get_deploy_config()
 log = logging.getLogger('benchmark')
 
+FILE_PATH_REGEX = re.compile(r'((?P<user>[^/]+)/)((?P<version>[^-]+)-)((?P<sha>[^-]+))(-(?P<tag>[^\.]+))?\.json')
+filepath = 'tpoterba/0.2.45-ac6815ee857c-master.json'
+
+
+def parse_file_path(name):
```
jigold, line 27: See the comment about moving these to `utils.py`.

benchmark-service/benchmark/benchmark.py
```diff
@@ -33,16 +92,43 @@ async def greet_user(request: web.Request) -> web.Response:
     return response
 
 
+@router.get('/name/{name}')
+@web_authenticated_developers_only(redirect=False)
+async def show_name(request: web.Request, userdata) -> web.Response:  # pylint: disable=unused-argument
+    benchmarks = get_benchmarks(filepath)
+    name_data = next((item for item in benchmarks['data'] if item['name'] == str(request.match_info['name'])), None)
+    fig = px.scatter(x=list(range(0, len(name_data['times']))), y=name_data['times'])
+    # [item for item in range(0, len(name_data['times']))]
```
jigold, line 101: delete

benchmark-service/benchmark/benchmark.py
```diff
@@ -33,16 +92,43 @@ async def greet_user(request: web.Request) -> web.Response:
...
+
+    plot = json.dumps(fig, cls=plotly.utils.PlotlyJSONEncoder)
+    context = {
+        'name': request.match_info.get('name', ''),
+        'plot': plot
+    }
+
+    # response = aiohttp_jinja2.render_template('user.html', request,
```
jigold, line 109: delete

benchmark-service/benchmark/benchmark.py
```diff
@@ -33,16 +92,43 @@ async def greet_user(request: web.Request) -> web.Response:
     return response
 
 
+@router.get('/name/{name}')
+@web_authenticated_developers_only(redirect=False)
+async def show_name(request: web.Request, userdata) -> web.Response:  # pylint: disable=unused-argument
+    benchmarks = get_benchmarks(filepath)
+    name_data = next((item for item in benchmarks['data'] if item['name'] == str(request.match_info['name'])), None)
```
jigold, line 99: Let's throw an exception if `name` is not in the data.

benchmark-service/benchmark/benchmark.py
```diff
@@ -33,16 +92,43 @@ async def greet_user(request: web.Request) -> web.Response:
...
+    return await render_template('benchmark', request, userdata, 'index.html', benchmarks_context)
+
+
+@router.post('/lookup')
+@check_csrf_token
+@web_authenticated_developers_only(redirect=False)
+async def lookup(request, userdata):  # pylint: disable=unused-argument
+    data = await request.post()
+    file = data['file']
+    global filepath
```
jigold, line 128: I don't think you need a global variable here.

benchmark-service/benchmark/templates/user.html
```diff
@@ -4,16 +4,68 @@
 {% endblock %}
 
 {% block content %}
-<h1>Hello, {{username}}</h1>
+<h1>Hello, {{name}}</h1>
```
jigold, line 7: Can you rename this file?

benchmark-service/benchmark/templates/user.html
```diff
@@ -4,16 +4,68 @@
...
+
+<script src="https://cdn.plot.ly/plotly-latest.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.6/d3.min.js"></script>
```
jigold, line 10: Does plotly require d3? I can't remember.

benchmark-service/benchmark/benchmark.py
```diff
@@ -33,16 +92,43 @@ async def greet_user(request: web.Request) -> web.Response:
     return response
 
 
+@router.get('/name/{name}')
+@web_authenticated_developers_only(redirect=False)
+async def show_name(request: web.Request, userdata) -> web.Response:  # pylint: disable=unused-argument
```
jigold, line 97: This is fine for now, but I think we'll need to pass a query parameter with the name of the file path to look at. Maybe that would be a good way to pass the benchmark name as well.

I think I like benchmark and benchmarks as the equivalent of a single benchmark test and benchmarks as a collection of benchmark tests.

benchmark-service/Dockerfile
```diff
@@ -4,9 +4,15 @@ COPY benchmark-service/benchmark /benchmark-service/benchmark/
...
 COPY benchmark-service/setup.py benchmark-service/MANIFEST.in  /benchmark-service/
 
+COPY benchmark-service/resources/0.2.45-ac6815ee857c-master.json /
+
 RUN pip3 install --no-cache-dir /benchmark-service && \
     rm -rf /benchmark-service
 
+RUN pip3 install plotly
+
+RUN pip3 install pandas
```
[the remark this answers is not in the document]
Dania-Abuhijleh: I don't think I use it either, but when I removed it I got an error like "plotly needs pandas" or something

benchmark-service/benchmark/utils.py
```diff
@@ -0,0 +1,36 @@
...
+        return text
+
+    def parse_file_path(self, name):
+        match = self.FILE_PATH_REGEX.fullmatch(name)
+        return match.groupdict()
+
+    def get_data_as_string(self, file_path):
+        file_info = self.parse_file_path(file_path)
+        bucket = self.storage_client.get_bucket(file_info['bucket'])
+        shorter_file_path = self.remove_prefix(file_path, 'gs://' + file_info['bucket'] + '/')
```
jigold, line 27: I think remove_prefix is a static method, so I'm not sure why this is able to work the way it is. I think you need to call it `ReadGoogleStorage.remove_prefix(...)`

benchmark-service/benchmark/utils.py
```diff
@@ -0,0 +1,36 @@
...
+
+        try:
+            # get bucket data as blob
+            blob = bucket.blob(shorter_file_path)
+            # convert to string
+            json_data = blob.download_as_string()
```
jigold, line 33: Can you name this `data` instead of `json_data`? It doesn't necessarily have to be json.

benchmark-service/benchmark/utils.py
```diff
@@ -0,0 +1,36 @@
...
+    FILE_PATH_REGEX = re.compile(r'gs://((?P<bucket>[^/]+)/)((?P<user>[^/]+)/)((?P<version>[^-]+)-)((?P<sha>[^-]+))(-(?P<tag>[^\.]+))?\.json')
+
+    @staticmethod
+    def remove_prefix(text, prefix):
+        if text.startswith(prefix):
+            return text[len(prefix):]
+        return text
+
+    def parse_file_path(self, name):
+        match = self.FILE_PATH_REGEX.fullmatch(name)
```
jigold, line 21: I think this function can return a tuple of (bucket, path) by parsing the groupdict. Then you don't need the remove_prefix function.

benchmark-service/resources/0.2.45-ac6815ee857c-master.json
```diff
@@ -0,0 +1 @@
...1.4180844830000012], "median": 1.2551967819999987, "mean": 1.3235203892000016, "stdev": 0.1997291787840194, "f-stat": 4.083676735620424, "p-value": 0.03557964330211377}, {"failed": false, "trials": [[1800.0], [1800.0], [1800.0], [1800.0], [1800.0]], "name": "table_range_join_1b_1b", "times": [1800.0, 1800.0, 1800.0, 1800.0, 1800.0], "median": 1800.0, "mean": 1800.0, "stdev": 0.0, "f-stat": NaN, "p-value": NaN}, {"failed": false, "trials": [[6.922640386000012, 6.84629538199988, 7.265700216999903], [6.947766942000001, 6.683704954999996, 6.879588456999997], [7.257506964000001, 6.643026243999998, 6.4678616900000065], [7.3989795199999975, 7.208221214999995, 6.54738759], [8.023792319000009, 7.301427963999998, 6.807952096999998]], "name": "table_range_means", "times": [6.922640386000012, 6.84629538199988, 7.265700216999903, 6.947766942000001, 6.683704954999996, 6.879588456999997, 7.257506964000001, 6.643026243999998, 6.4678616900000065, 7.3989795199999975, 7.208221214999995, 6.54738759, 8.023792319000009, 7.301427963999998, 6.807952096999998], "median": 6.922640386000012, "mean": 7.01345679613332, "stdev": 0.3894407669333821, "f-stat": 0.9921544078155394, "p-value": 0.45514157177711556}]}
```
jigold, line 1: Can you delete this file?

benchmark-service/benchmark/templates/name.html
```diff
@@ -0,0 +1,25 @@
...
+<div class="chart" id="scatterplot">
+    <script>
+        var graphs = {{plot | safe}};
+        Plotly.plot('scatterplot',graphs,{});
+    </script>
+</div>
+
+
+<hr>
+<p>
```
jigold, line 21: Can we delete the goose link now? Not sure you have this route anymore.

benchmark-service/benchmark/templates/name.html
```diff
@@ -0,0 +1,25 @@
+{% extends "base.html" %}
+{% block title %}
+Posts index
+{% endblock %}
+
+{% block content %}
+<h1>Hello, {{name}}</h1>
```
jigold, line 7: Let's just call this `<h1> {{name}} </h1>`

benchmark-service/benchmark/templates/name.html
```diff
@@ -0,0 +1,25 @@
+{% extends "base.html" %}
+{% block title %}
+Posts index
```
jigold, line 3: Let's rename this to `Benchmark Summary`

benchmark-service/benchmark/templates/index.html
```diff
@@ -1,11 +1,57 @@
-
 {% extends "base.html" %}
-{% block title %}
-Posts index
-{% endblock %}
+{% block title %} Benchmark {% endblock %}
```
jigold, line 2: Let's call this `Benchmarks`

benchmark-service/benchmark/templates/base.html
```diff
@@ -1,11 +1,13 @@
...
-    <meta charset='UTF-8'>
-    <title>{% block title %}
-        Index
-        {% endblock %}
-    </title>
+    <title>Hail | {% block title %}{% endblock %}</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1"></meta>
+    <link rel="shortcut icon" href="{{ base_path }}/common_static/hail_logo_sq.ico" type="image/x-icon" />
+    <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
```
jigold, line 8: Why did you have to add these? I think you can inherit from our existing jinja templates with this:

```
{% extends "layout.html" %}
```

An example is in `batch/batch/front_end/templates/batch.html`

benchmark-service/benchmark/utils.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def get_geometric_mean(prod_of_means, num_of_means):
+    return prod_of_means ** (1.0 / num_of_means)
+
+
+class ReadGoogleStorage:
+    storage_client = storage.Client()
+
+    FILE_PATH_REGEX = re.compile(r'gs://((?P<bucket>[^/]+)/)((?P<user>[^/]+)/)((?P<version>[^-]+)-)((?P<sha>[^-]+))(-(?P<tag>[^\.]+))?\.json')
```
jigold, line 12: I think this regex just needs to be `gs://((?P<bucket>[^/]+)/)(?P<path>*)`.

auth/auth/auth.py
```diff
@@ -111,7 +111,7 @@ async def callback(request):
 @routes.get('/user')
 @web_authenticated_users_only()
 async def user_page(request, userdata):
-    return await render_template('auth', request, userdata, 'user.html', {})
+    return await render_template('auth', request, userdata, 'name.html', {})
```
jigold, line 114: You shouldn't need to change this.

benchmark-service/benchmark/benchmark.py
```diff
@@ -2,19 +2,64 @@
...
+import plotly
+import plotly.express as px
 
 configure_logging()
 router = web.RouteTableDef()
 logging.basicConfig(level=logging.DEBUG)
 deploy_config = get_deploy_config()
 log = logging.getLogger('benchmark')
 
+FILE_PATH_REGEX = re.compile(r'gs://hail-benchmarks/((?P<user>[^/]+)/)((?P<version>[^-]+)-)((?P<sha>[^-]+))(-(?P<tag>[^\.]+))?\.json')
```
jigold, line 22: Let's call this `BENCHMARK_FILE_REGEX`

benchmark-service/benchmark/benchmark.py
```diff
@@ -2,19 +2,64 @@
...
+default_filepath = 'gs://hail-benchmarks/tpoterba/0.2.45-ac6815ee857c-master.json'
+
+
+def get_benchmarks(file_path):
+
```
jigold, line 27: delete extra new line.

benchmark-service/benchmark/benchmark.py
```diff
@@ -2,19 +2,64 @@
...
+    read_gs = ReadGoogleStorage()
+    file_info = read_gs.parse_file_path(file_path)
```
jigold, line 29: Can you move the parsing of the file path to get the sha further down so we don't mix the reading logic with the parsing?

benchmark-service/benchmark/benchmark.py
```diff
@@ -2,19 +2,64 @@
...
+    sha = file_info['sha']
+    try:
+        json_data = read_gs.get_data_as_string(file_path)
+        pre_data = json.loads(json_data)
+    except Exception:
+        message = 'could not find file'
```
jigold, line 35: Can you make the message like this?

`message = f'could not find file, {file_path}`

benchmark-service/benchmark/benchmark.py
```diff
@@ -28,21 +73,51 @@ async def greet_user(request: web.Request) -> web.Response:
         'username': request.match_info.get('username', ''),
         'current_date': 'July 10, 2020'
     }
-    response = aiohttp_jinja2.render_template('user.html', request,
+    response = aiohttp_jinja2.render_template('name.html', request,
```
jigold, line 76: Can we delete this route?

benchmark-service/benchmark/benchmark.py
```diff
@@ -28,21 +73,51 @@ async def greet_user(request: web.Request) -> web.Response:
...
+        message = 'could not find name'
+        log.info('name is of type NoneType: ' + message, exc_info=True)
+        raise web.HTTPBadRequest(text=message)
+
+    context = {
+        'name': request.match_info.get('name', ''),
+        'plot': plot
+    }
+
+    return await render_template('benchmark', request, userdata, 'name.html', context)
```
jigold, line 100: Is the issue you were running into that the template needed to be in the web_common directory?

benchmark-service/benchmark/benchmark.py
```diff
@@ -28,21 +73,51 @@ async def greet_user(request: web.Request) -> web.Response:
...
 
 
+@router.get('/name/{name}')
+@web_authenticated_developers_only(redirect=False)
+async def show_name(request: web.Request, userdata) -> web.Response:  # pylint: disable=unused-argument
+    benchmarks = get_benchmarks(default_filepath)
+    name_data = next((item for item in benchmarks['data'] if item['name'] == str(request.match_info['name'])),
+                     None)
+    try:
+        fig = px.scatter(x=list(range(0, len(name_data['times']))), y=name_data['times'])
```
jigold, line 88: I think for the x axis you want a function that takes `name_data['times']` and converts it to something that looks like this:

`[0, 0, 0, 1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 4]`

benchmark-service/benchmark/benchmark.py
```diff
@@ -28,21 +73,51 @@ async def greet_user(request: web.Request) -> web.Response:
...
-    response = aiohttp_jinja2.render_template('user.html', request,
+    response = aiohttp_jinja2.render_template('name.html', request,
                                               context=context)
     return response
 
 
+@router.get('/name/{name}')
+@web_authenticated_developers_only(redirect=False)
+async def show_name(request: web.Request, userdata) -> web.Response:  # pylint: disable=unused-argument
+    benchmarks = get_benchmarks(default_filepath)
```
jigold, line 84: I changed my mind. I think the goal should be to show the benchmarks for any arbitrary file here. @danking Do you have suggestions on how to pass the file name in this GET request?

benchmark-service/benchmark/benchmark.py
```diff
@@ -28,21 +73,51 @@ async def greet_user(request: web.Request) -> web.Response:
...
+    return await render_template('benchmark', request, userdata, 'index.html', benchmarks_context)
+
+
+@router.post('/lookup')
+@check_csrf_token
+@web_authenticated_developers_only(redirect=False)
+async def lookup(request, userdata):  # pylint: disable=unused-argument
+    data = await request.post()
+    file = data['file']
+    # global default_filepath
```
jigold, line 117: delete. I think what's causing confusion here is that we don't want our application to have a saved state. What happens if multiple people are looking at benchmarks at the same time? Instead, the information needed to render the response should come from the request in the form of POST fields or GET query parameters. The reason for the default file is just to have an initial starting point for the first time a user loads the page if they don't give a file path in the index page. But for this lookup request, we'll always get the file from the POST request.

benchmark-service/benchmark/benchmark.py
```diff
@@ -28,21 +73,51 @@ async def greet_user(request: web.Request) -> web.Response:
...
 @router.get('')
 @web_authenticated_developers_only(redirect=False)
 async def index(request: web.Request, userdata) -> Dict[str, Any]:  # pylint: disable=unused-argument
-    context = {
-        'current_date': 'July 10, 2020'
-    }
-    response = aiohttp_jinja2.render_template('index.html', request,
-                                              context=context)
-    return response
+    benchmarks_context = get_benchmarks(default_filepath)
```
jigold, line 107: I think you want something that checks whether a file path is given as a query parameter and if not, use the default file path.

benchmark-service/benchmark/benchmark.py
```diff
@@ -2,19 +2,64 @@
...
+            prod_of_means *= d['mean']
+            stats['f-stat'] = round(d['f-stat'], 6)
+            stats['mean'] = round(d['mean'], 6)
+            stats['median'] = round(d['median'], 6)
+            stats['p-value'] = round(d['p-value'], 6)
+            stats['stdev'] = round(d['stdev'], 6)
+            stats['times'] = d['times']
+        data.append(stats)
+    geometric_mean = get_geometric_mean(prod_of_means, len(pre_data['benchmarks']))
+        # prod_of_means ** (1.0 / len(pre_data['benchmarks']))
```
jigold, line 55: delete if not used.

scorecard/scorecard/scorecard.py
```diff
@@ -68,7 +68,7 @@ async def html_get_user(request, userdata):
         'user_data': user_data,
         'updated': updated,
     }
-    return await render_template('scorecard', request, userdata, 'user.html', page_context)
+    return await render_template('scorecard', request, userdata, 'name.html', page_context)
```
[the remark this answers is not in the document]
Dania-Abuhijleh: oops that was a mistake when refactoring

PR #9217, opened 2020-08-04
Checked Memory

johnc1231:
This is mostly Cotton's checked memory allocator code, plus an adjustment to the `Makefile` to allow it to be used instead of the regular `Memory.java`. I was trying to do a generic thing where we'd just automatically replace anything in `src/main` with its corresponding file in `src/debug`, but I was running into lots of issues where I wasn't running shell commands correctly, and got annoyed and just hard coded the one file. If we ever need two debug files I'll fix this up, or I can try some more tomorrow if hard coding the names like this is bad form. 

Anyway, this should allow anyone to run code with the checked memory allocator if they pass the `HAIL_DEBUG_MODE=1` argument to any of the `make` commands.


danking:
I think this should go to someone on compilers team. I'll fix the scorecard dice.

johnc1231:
Heh, I was not actually worried about the Java code after Cotton and I both looked at it. I figured you were most likely to actually carefully read makefile changes

johnc1231:
I'll unassign you and assign Chris when I finish testing it.

johnc1231:
Confirmed this is working. If I introduce a segfault to some code, then run `make install-editable HAIL_DEBUG_MODE=1`, then execute the python code that would cause a segfault, it throws a RuntimeError.

johnc1231:
@cseed  I will make a follow up PR that changes testing to make the checked version run. This at least allows developers to turn on this behavior when debugging locally.

hail/Makefile
```diff
@@ -41,13 +42,25 @@ SCALA_BUILD_INFO := src/main/resources/build-info.properties
...
+ifdef HAIL_DEBUG_MODE
+shadowJar: debugShadowJar
+endif
 shadowJar: $(SHADOW_JAR)
 
+.PHONY: debugShadowJar
+debugShadowJar: $(SHADOW_JAR)
+	mkdir -p build/debug/scala/is/hail/annotations/ && \
+	javac src/debug/scala/is/hail/annotations/Memory.java && \
+	mv src/debug/scala/is/hail/annotations/Memory.class build/debug/scala/is/hail/annotations && \
```
danking, line 60: the ampersands and `\` are unnecessary, make will fail if any step fails

danking: I think these three lines can be:
```
mkdir -p build/debug/scala/is/hail/annotations
javac -d build/debug src/debug/scala/is/hail/annotations/Memory.java
```

hail/Makefile
```diff
@@ -41,13 +42,25 @@ SCALA_BUILD_INFO := src/main/resources/build-info.properties
 
 SHADOW_JAR := build/libs/hail-all-spark.jar
 SHADOW_TEST_JAR := build/libs/hail-all-spark-test.jar
+DEBUG_CLASSES := build/debug
```
chrisvittal, line 45: Remove, it's unused.

hail/Makefile
```diff
@@ -41,13 +42,25 @@ SCALA_BUILD_INFO := src/main/resources/build-info.properties
...
 EGG := build/deploy/dist/hail-$(HAIL_PIP_VERSION)-py3.6.egg
 
 .PHONY: shadowJar
+ifdef HAIL_DEBUG_MODE
+shadowJar: debugShadowJar
+endif
 shadowJar: $(SHADOW_JAR)
 
+.PHONY: debugShadowJar
+debugShadowJar: $(SHADOW_JAR)
```
chrisvittal, line 57: I don't think we should be adding another phony target like this. The shadow jar rule should just add the debug classes.

hail/Makefile
```diff
@@ -41,13 +42,25 @@ SCALA_BUILD_INFO := src/main/resources/build-info.properties
...
+	mkdir -p build/debug/scala/is/hail/annotations/ && \
```
chrisvittal, line 58: The appropriate path would be `build/classes/scala/debug`

hail/Makefile
chrisvittal: What Dan said. Also prefix your `mkdir` commands with `@` to suppress echoing of the commands by `make`.

chrisvittal, line 60: Have a look at the `-d` option of `javac`. It does what you want here.

The syntax would be:
```
javac -d build/classes/debug/scala src/debug/scala/is/hail/annotations/Memory.java
```

johnc1231:
I applied your changes Chris, that does make it easier. I haven't had to do anything nontrivial in a Makefile, so I appreciate you taking the time to clean up / explain. Thanks!

hail/Makefile
```diff
@@ -22,14 +22,18 @@ $(eval $(call ENV_VAR,BRANCH))
...
 
 HAIL_PYTHON3 ?= python3
 PIP ?= $(HAIL_PYTHON3) -m pip
 
 # not perfect, not robust to simdpp changes, but probably fine
+BUILD_DEBUG_PREFIX := build/classes/scala/debug
 JAR_SOURCES := $(shell git ls-files src/main)
 JAR_SOURCES += build.gradle
 JAR_TEST_SOURCES := $(shell git ls-files src/test)
+JAR_DEBUG_SOURCES := src/debug/scala/is/hail/annotations/Memory.java
```
chrisvittal, line 35: For the record, I did this manually because I don't actually like the `git ls-files` shell out that I originally did to get this quick and correct. I don't like how `git ls-files` means that you can't build this from a source tarball.

PR #9219, opened 2020-08-05
[batch] ENTRYPOINT support

akotlar:
CHANGELOG: added default_shell, shell, and fix a bug related to using the wrong shell.


hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -88,6 +88,52 @@ def test_single_job(self):
...
+            j.entrypoint('echo foobar')
+            j.command(f'echo "{msg}" > {j.ofile}')
+
+            b.write_output(j.ofile, output_file.name)
+            b.run()
+            out = self.read(output_file.name)
+
+            assert self.read(output_file.name) == msg
+
+    def test_single_job_with_entrypoint(self):
```
akotlar, line 106: This test isn't working in CI (file not found), works fine on local. @danking as I don't have access to the cluster anymore, would you have any suggestions?

akotlar:
This is generating some 500 errors, having a hard time debugging without hail-vdc.

akotlar:
This should now be passing tests. I had to make one LocalBackend test conditional on docker, because we do not have docker in docker (DiD), and there were historically some security considerations that felt out of scope to fully grapple with in this PR (out of scope to create a DiD image and have test_hailtop_batch and test_batch_dcos use this DiD image.

akotlar:
Long term the right solution is to use python's docker engine library to manage docker calls. This will allow us to unify ServiceBackend and LocalBackend container configuration/create/start and allow us to re-enable these tests.

akotlar:
Long term, I think it would be better to use docker engine for LocalBackend, so that container config/create/start between ServiceBackend and LocalBackend could have a common interface.

If you're interested, after regenie is finished, I can PR.

cseed:
Just saw this.

 > adds job.entrypoint method

I'm going to argue against this.  The goal is not to make Batch a Docker wrapper, in fact, maybe that's an anti-goal.  The goal is to build on docker (and other container tools) to create a natural interface for the problems Batch is trying to solve.  The expectation is that the commands to be executed will be specified as part of the Job.  In particular, I think if you don't specify the command, the command set is empty and no commands get run.  It doesn't default to the docker command by default when no command is specified.

So I would solve this by just always invoking the image with `--entrypoint <shell>`.

 > Long term, I think it would be better to use docker engine for LocalBackend, so that container config/create/start between ServiceBackend and LocalBackend could have a common interface.

While code sharing often sounds nice, it also has the effect of entangling otherwise separate pieces of code.  That means they cannot evolve separately and makes development more difficult.  We have lots of plans to evolve the service backend that won't have analogues in the local backend, and I feel this proposal would make those changes harder and make the code more complex to handle two use cases.

akotlar:
I have a Dockerfile, that I cannot modify that has an ENTRYPOINT that is not /bin/bash or /bin/sh. How do I get this to run in batch without this PR? I don't understand the proposal.

I'm getting complaints from users saying that containerizing existing applications is a pain. What is the plan to help them?

cseed:
> I have an image built from Dockerfile that I cannot modify, and which has an ENTRYPOINT that is not /bin/bash or /bin/sh. How do I get this to run in batch without this PR? I don't understand the proposal.

The proposal is to have the backend run `docker run --entrypoint <shell> ...` unconditionally.  This involves no change to the user interface.  I believe this would implement the intended Batch behavior on any docker image.

akotlar:
Should be ready to reviews, addressed cotton's comment

danking:
I'm sorry, I was unclear when I suggested that we "fix the ENTRYPOINT bug". In my view, the bug is that we do not *unset* the entry point when we invoke Docker (locally and in the service backend). Locally this means including the argument `--entrypoint ""`. In the service we need to modify `worker.py` to set the `config`'s `Entrypoint` to, I think, the empty string. I do not want `entrypoint` in the job spec at all.

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -88,6 +89,21 @@ def test_single_job(self):
...
             assert self.read(output_file.name) == msg
 
+    @unittest.skipIf(not which("docker"), "docker command is missing")
+    def test_single_job_with_entrypoint(self):
+        with tempfile.NamedTemporaryFile('w') as output_file:
+            msg = 'hello world'
+
+            b = self.batch()
+            j = b.new_job()
+            j.image('hailgenetics/entrypoint-test')
```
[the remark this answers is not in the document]
danking: Could you also delete the `entrypoint-test` image?

batch/batch/front_end/validate.py
```diff
@@ -8,6 +8,7 @@
 #   'always_run': bool,
 #   'attributes': {str: str},
 #   'command': [str],
+#   'entrypoint': str or list[str]
```
cseed, line 11: My proposal is backend-only and shouldn't involve changing the Python client library or REST API.

Wait, is your problem the docker image has an `ENDPOINT` already defined, or that the docker image doesn't have `/bin/bash` installed?

I see `/bin/bash` is still hardcoded.  I thought this had been fixed, this is still a pending Asana item.  The solution here is that the `Batch` constructor should take a default shell option that can be overridden per-job.  This should be used by the backend, defaulting to `/bin/bash` if not specified.  This design overlaps somewhat with what you've proposed.

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -88,6 +89,21 @@ def test_single_job(self):
 
             assert self.read(output_file.name) == msg
 
+    @unittest.skipIf(not which("docker"), "docker command is missing")
```
danking, line 92: Let's not include this test for now. I know it leaves a testing hole (jobs with images in local mode), but that's a known testing hole. Safe handling of docker in docker will be enabled by short term goals we have around getting off of Docker. In the meantime, I'd rather not have unmaintained tests (I never run the tests locally).

batch/batch/front_end/validate.py
akotlar: ENTRYPOINT defined. Goal is to unset. Since you all don't have any interest in entrypoint being threaded through, I can just set Entrypoint="" in the worker

akotlar: The 2nd thing my PR allowed was handling non /bin/bash environments. I thought you said you don't want that, so removed ability to set it. That was effectively handled with this PR

hail/python/test/hailtop/batch/test_batch.py
akotlar: There is, apparently safe docker in docker now, without --priveleged. I haven't looked too far into it.

akotlar: Ok, we're going to have the exact same problem with regenie

hail/python/test/hailtop/batch/test_batch.py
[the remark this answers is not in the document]
akotlar: Sure. Yes, I hadn't done this because nginx is a much heavier image than alpine with bash, so using nginx would be slower. I also didn't find any docker images that we rely on that use entrypoints.

batch/batch/front_end/validate.py
cseed: I'm not asking for a use case, I'm asking for an interface.  Let me be more specific:

 - The `Batch` constructor takes a `shell: str = None` keyword argument,
 - `Batch.new_job` takes a `shell: str = None` keyword argument that overrides the Batch argument,
 - the job spec in the service backend has a new optional `shell: str` field,
 - the backends use shell, if given, instead of `/bin/bash` on a per-Job basis.

hail/python/test/hailtop/batch/test_batch.py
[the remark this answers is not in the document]
cseed: Personally I'd build the image during the build step instead of introducing an external dependency.  You have to build the image locally to test locally.  That logic normally goes in a Makefile.  A Makefile for batch tests that builds the entrypoint image seems reasonable.  That image should be built on ubuntu:18.04 like all (most) of our images.  That will always be lighter weight than an external image because it will (almost) always be cached.

batch/batch/front_end/validate.py
akotlar: >  I'm asking for an interface.

The interface: 

* Not needed: The Batch constructor changes
* Not needed: batch.new_job changes
* job.entrypoint(arg: Union(str, List[str]), job spec has a new optional entrypoint. 
* The backend uses entrypoint (I will need a small change to this PR to make it override /bin/bash in service backend)

cseed: I'm asking for the interface I wrote down.  I thought I said clearly: our interface is not meant to be a wrapper for Docker and endpoint is the wrong concept here.

akotlar: Why does the command I wrote need to have anything to do with Docker? Call the method whatever you want. The current backend is Docker, but that's an implementation detail.  I wrote what you wrote except pointed out that in this PR the first two things you wrote are not necessary.

akotlar: Also I don't know what "I'm asking for the interface I wrote down means" in the context of your earlier suggestion that there may be overlap with this implementation except a mapping.

cseed: Let be be as clear as I can: change this PR to remove entrypoint from the interface and implement this interface instead:

 > The Batch constructor takes a shell: str = None keyword argument,
 > Batch.new_job takes a shell: str = None keyword argument that overrides the Batch argument,
 > the job spec in the service backend has a new optional shell: str field,
 > the backends use shell, if given, instead of /bin/bash on a per-Job basis.

Also unset the entrypoint with `--entrypoint ""` in the local backend to fix your original issue.

akotlar: As an aside, your proposed interface is called "shell". I don't like that. Everyone who uses Batch will think about Docker. Docker has a primitive for overriding a command. Why not use something that has the same name if it does almost the same thing (yours is a subset of the Docker ENTRYPOINT interface). 

People have enough difficulty understanding how to containerize things. We're now going to add another term for the same thing. So TJ will try to containerize something in Docker. He sees that there is something in Docker called Entrypoint. He will realize that Job command has a bunch of magic behind the scenes that makes it impossible to supply the command that he wants to his image. Now we wants to override ENTRYPOINT. He'll search for that. Nothing. He has no idea what various "shell"s are. This is a computer science artifact. 

Better to name the thing what is expected. This has nothing to do with Batch having a docker backend, or a cgroup backend. You'll need some way to go from a docker image to executes code. My 2c.

akotlar: > Let be be as clear as I can: change this PR to remove entrypoint from the interface and implement this interface instead:
> 
> 
> 
>  > The Batch constructor takes a shell: str = None keyword argument,
> 
>  > Batch.new_job takes a shell: str = None keyword argument that overrides the Batch argument,
> 
>  > the job spec in the service backend has a new optional shell: str field,
> 
>  > the backends use shell, if given, instead of /bin/bash on a per-Job basis.
> 
> 
> 
> Also unset the entrypoint with `--entrypoint ""` in the local backend to fix your original issue.
> 
> 

You said this was per-job. Why make it per batch? You specify images per job, commands, they may have different entry point needs.

cseed: I'm telling you, we already decided not to do that.  `shell` isn't entrypoint by a different name.  `shell` is literally the shell used to interpret the shell script constructed from the job command.  What you are proposing will not work.  If you want to run plink, setting the shell or entrypoint to plink will not work: it does not take `-c <script>` as valid arguments.  We use this for several features: `verbose` adds `set +x` to the script, running `command` multiple times creates a shell script with `&&`, etc.  What you are proposing is incorrect in the context of the current design which I am trying to describe.

I understand you dislike the existing design.  The options I see are:
 - you just unset `entrypoint` for this PR and we sort out the interface question separately,
 - you implement the interface we've created to work with the existing design, or
 - you close this and make a larger proposal to replace our existing interface with one modelled on the docker interface and discuss.

cseed: > You said this was per-job. Why make it per batch? You specify images per job, commands, they may have different entry point needs.

No, there is a default image set on the batch.  The jobs can override that.  I'm asking for exactly the existing interface for the shell.  (It is not an endpoint.  See my other comment.)  In fact, that's true for many of the Job attributes:

```
class Batch:
    ...
    def __init__(self,
                 ...
                 default_image: Optional[str] = None,
                 default_memory: Optional[str] = None,
                 default_cpu: Optional[str] = None,
                 default_storage: Optional[str] = None,
                 default_timeout: Optional[Union[float, int]] = None):
```

akotlar: good point, thanks

akotlar: Let me know if current state of PR is what you're looking for. There's a way to do this without adding anything to job_spec (set job._shell instead of /bin/bash directly in command= in ServiceBackend, much like LocalBackend), but you specified it so I added it, assuming you have a plan that needs it.

hail/python/hailtop/batch/backend.py
```diff
@@ -470,6 +475,7 @@ def _cp(src, dst):
 
             j = bc_batch.create_job(image=job._image if job._image else default_image,
                                     command=['/bin/bash', '-c', cmd],
```
akotlar, line 477: could override here instead of adding to spec

batch/batch/front_end/validate.py
```diff
@@ -8,6 +8,7 @@
 #   'always_run': bool,
 #   'attributes': {str: str},
 #   'command': [str],
+#   'shell': str,
```
cseed, line 11: `shell` should be a property of the high-level batch client (hailtop.batch), not the REST API.

hail/python/hailtop/batch/backend.py
cseed: Yes, that is much simpler.

hail/python/hailtop/batch/batch.py
```diff
@@ -147,7 +150,14 @@ def new_job(self,
         if attributes is None:
             attributes = {}
 
-        j = Job(batch=self, name=name, attributes=attributes)
+        if shell is not None:
+            jshell = shell
+        elif self._default_shell is not None:
+            jshell = self._default_shell
+        else:
+            jshell = None
```
cseed, line 158: This code will go, but you could have written the simpler:

```
        if shell is None:
            shell = self._default_shell
        j = Job(batch=self, name=name, attributes=attributes, shell=shell)

```

hail/python/test/hailtop/batch/test_batch.py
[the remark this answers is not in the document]
akotlar: >You have to build the image locally to test locally

What does this mean? 

Trying to understand how batch/Makefile is used; I don't see where it's called by CI. I do see that I can use a buildImage step; I could just point that to the Dockerfiles I made for this PR (based on ubuntu:18.04, adds entrypoint and /bin/bash), which could live in batch/.

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -521,3 +542,24 @@ def test_envvar(self):
...
+
+            assert self.read(output_file.name) == msg
+
+    def test_single_job_with_nonsense_shell(self):
+        msg = 'hello world'
+
+        b = self.batch()
+        j = b.new_job(shell='bin/ajdsfoijasidojf')
+        j.command(f'echo "{msg}"')
+        assert b.run().status()['state'] != 'success'
```
cseed, line 565: You run this command twice in two different tests.  This is not the fastest, and we've consolidated multiple asserts in other batch tests.

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -521,3 +542,24 @@ def test_envvar(self):
         j.env('SOME_VARIABLE', '123abcdef')
         j.command('[ $SOME_VARIABLE = "123abcdef" ]')
         assert b.run().status()['state'] == 'success'
+
+    def test_single_job_with_shell(self):
+        with tempfile.NamedTemporaryFile('w') as output_file:
+            msg = 'hello world'
+
+            b = self.batch()
+            j = b.new_job(shell='bin/sh')
```
cseed, line 551: A leading slash seems safer.

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -521,3 +542,24 @@ def test_envvar(self):
...
+            j.command(f'echo "{msg}" > {j.ofile}')
+
+            b.write_output(j.ofile, output_file.name)
```
cseed, line 554: `output_file.name` is a local file, you can't do that.  Much easier to just echo and check the output.

hail/python/hailtop/batch/backend.py
```diff
@@ -199,6 +200,8 @@ def symlink_input_resource_group(r):
             resource_defs = [r._declare(tmpdir) for r in job._mentioned]
             env = [f'export {k}={v}' for k, v in job._env.items()]
 
+            shell = job._shell if job._shell else '/bin/bash'
```
akotlar, line 203: '/bin/bash' could be placed in Batch _default_shell instead.

hail/python/hailtop/batch/batch.py
```diff
@@ -90,7 +90,8 @@ def __init__(self,
                  default_memory: Optional[str] = None,
                  default_cpu: Optional[str] = None,
                  default_storage: Optional[str] = None,
-                 default_timeout: Optional[Union[float, int]] = None):
+                 default_timeout: Optional[Union[float, int]] = None,
+                 default_shell: str = None):
```
akotlar, line 94: default_shell: str = '/bin/bash'?

batch/batch/front_end/validate.py
cseed: One more thing, so you're aware: all information that's stored to disk must be versioned so the schema can evolve while future readers still know how to interpret it.  The config goes into the database, and therefore must be versioned and this change would require a version bump.

There is also a multi-step protocol for rolling out new versions: if you add a new field, you first have to commit support for it, deploy, verify all the workers have upgraded, and then you can enable the new feature.  Otherwise, an old worker could receive a config with `shell` that it doesn't handle.

akotlar: Thanks, I didn't realize we versioned job spec, though in retrospect this could have been "obvious" as we have versioned rest endpoints.

cseed: Well, the endpoints are all currently alpha so subject to change.  We'll change that when they are stable enough and there's enough user pressure.  We're probably close.

hail/python/hailtop/batch/batch.py
cseed: Either seems fine, but None has the advantage you can tell if the user specified the value or not.  It doesn't seem relevant here, but it can be.

hail/python/test/hailtop/batch/test_batch.py
akotlar: Consolidated as suggested, and left in to test the service backend handling of shell.

hail/python/hailtop/batch/batch.py
```diff
@@ -90,7 +90,8 @@ def __init__(self,
                  default_memory: Optional[str] = None,
                  default_cpu: Optional[str] = None,
                  default_storage: Optional[str] = None,
-                 default_timeout: Optional[Union[float, int]] = None):
+                 default_timeout: Optional[Union[float, int]] = None,
+                 default_shell: str = '/bin/bash'):
```
akotlar, line 94: if we specify default_shell here we can avoid specifying /bin/bash for every backend (although maybe we want different shell defaults?), and save a job._shell None check per backend, but as you said at the disadvantage of a small amount of visibility into user behavior (although I don't see where we log Batch() invocations)

hail/python/test/hailtop/batch/test_batch.py
cseed: Oops, sorry, I misread the diff, this was fine.

hail/python/test/hailtop/batch/test_batch.py
akotlar: What's the easiest way to read stdout?

hail/python/hailtop/batch/batch.py
cseed: Sleeping on this, the point is that if you know there is user input, you can add another default (e.g an environment variable).  Also, None matches the rest of the code, which is always a benefit.

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -521,3 +542,18 @@ def test_envvar(self):
         j.env('SOME_VARIABLE', '123abcdef')
         j.command('[ $SOME_VARIABLE = "123abcdef" ]')
         assert b.run().status()['state'] == 'success'
+
+    def test_single_job_with_shell(self):
+        msg = 'hello world'
+        b = self.batch()
+        j = b.new_job(shell='/bin/sh')
+        j.command(f'echo "{msg}" > {j.ofile}')
+        assert b.run().status()['state'] == 'success'
```
cseed, line 551: You can use log to get the output, just don't redirect it.  Ideal if the local and service tests match.

akotlar: I found a solution to reading the log, but I suspect not the easiest one. I searched documentation, nothing on logs besides something that would require redirecting echo to file handle, and since neither batch.Job nor batch.Batch have a get-log method, I went through _batch_client.

akotlar:
I can't undismiss. Missed the default change request

akotlar:
Done

hail/python/hailtop/batch/batch.py
danking, line 94: I'm pretty sure this won't satisfy mypy it needs to be `Optional[str]`.

cseed: This is correct.  Note, this behavior changed.  At one point, PEP-848 allowed omitting `Optional` but that is no longer the case.  From https://www.python.org/dev/peps/pep-0484/#union-types:

 > A past version of this PEP allowed type checkers to assume an optional type when the default value is None, as in this code:

> def handle_employee(e: Employee = None): ...

> This would have been treated as equivalent to:

> def handle_employee(e: Optional[Employee] = None) -> None: ...

> This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

hail/python/hailtop/batch/backend.py
```diff
@@ -469,7 +470,7 @@ def _cp(src, dst):
                 resources['storage'] = job._storage
 
             j = bc_batch.create_job(image=job._image if job._image else default_image,
-                                    command=['/bin/bash', '-c', cmd],
+                                    command=[job._shell if job._shell else '/bin/bash', '-c', cmd],
```
cseed, line 473: If you're concerned about duplicating bash, you can do something like `DEFAULT_SHELL = '/bin/bash'` in the Backend abstract base class.

hail/python/hailtop/batch/batch.py
```diff
@@ -112,12 +113,14 @@ def __init__(self,
...
         self._default_timeout = default_timeout
+        self._default_shell = default_shell
 
         self._backend = backend if backend else LocalBackend()
 
     def new_job(self,
                 name: Optional[str] = None,
-                attributes: Optional[Dict[str, str]] = None) -> Job:
+                attributes: Optional[Dict[str, str]] = None,
+                shell: str = None) -> Job:
```
cseed, line 123: Also need `Optional[str]` here.

hail/python/hailtop/batch/job.py
```diff
@@ -65,8 +65,10 @@ def _new_uid(cls):
     def __init__(self,
                  batch: batch.Batch,
                  name: Optional[str] = None,
-                 attributes: Optional[Dict[str, str]] = None):
+                 attributes: Optional[Dict[str, str]] = None,
+                 shell: str = None):
```
cseed, line 69: Same.

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -521,3 +540,18 @@ def test_envvar(self):
...
         j.command('[ $SOME_VARIABLE = "123abcdef" ]')
         assert b.run().status()['state'] == 'success'
+
+    def test_single_job_with_shell(self):
+        msg = 'hello world'
+        b = self.batch()
+        j = b.new_job(shell='/bin/sh')
+        j.command(f'echo "{msg}"')
+        r = b.run(wait=True)
+        res = self.backend._batch_client.get_job_log(r.id, b._jobs[0]._job_id)["main"].rstrip()
```
cseed, line 550: @jigold do we intend to let Batch users query the status and log of jobs?  I don't see it in the interface and obviously the code Alex had to write here is far from ideal.

hail/python/hailtop/batch/batch.py
akotlar: Thanks for this Cotton, nice catch Dan!

akotlar:
Besides the question to Jackie, I'd still like a bit of clarity on https://github.com/hail-is/hail/pull/9219#discussion_r469015358 (if you want me to add an entrypoint-override image and test)

hail/python/test/hailtop/batch/test_batch.py
cseed: OK, to keep this moving forward, while we're working on the log story with Jackie, I think we should just change this to match the other batch tests and check `assert b.run().status()['state'] == 'success'` and follow up with a more precise test when that's sorted.

hail/python/test/hailtop/batch/test_batch.py
cseed: I think you want the more specific `== 'failure'` here.  It would be incorrect if the state was, `cancelled`, say.

hail/python/test/hailtop/batch/test_batch.py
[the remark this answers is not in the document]
cseed: Sorry I missed this, thanks for the bump.  The Makefiles are not called by CI, they are parallel constructions, and are primarily in place to bootstrap the cluster.  The Makefiles (roughly) contain the steps needed to install everything up to CI on a fresh cluster so that CI is then able to do a full deploy.  They are also local versions of certain CI build steps.  So if you want to run a build step locally (like a test), I would say that's the natural place.  I don't know if there are existing examples for local tests.  If your step (a test step) depends on another step (like building an image), it would also have to go into the Makefile.

To put this in context, in the future, we might want:
 - A tool that can run some build.yaml steps locally.  That's complicated because some steps need secrets or access to K8s/GCP networking that are not available locally, so the design isn't clear here.  This could the current Makefiles.
 - Fully automate (and test!) spinning up or upgrading a new cluster (probably using Terraform).  This could replace the current Makefiles (or there'd be just a few Terraform make targets).

The Makefiles rot quickly because they are not tested automatically.

 > I can create a buildImage step to build this as a prerequisite to test_hailtop_batch. Is that what you meant?

Yes.  That handles the case of running these tests under CI.  I thought you had concerns about running the tests locally. That's what the Makefile comment was about.

hail/python/test/hailtop/batch/test_batch.py
jigold: Thanks for bringing this up! I think I can expose it on a job and batch object, but there isn't a good way to query the status and logs for a batch in a Python script where the object no longer exists. For example, doing queries after the fact a week later. I can put it on Asana to think this through more carefully. I think it would involve adding functionality to the ServiceBackend that takes either a Job or Batch object or the batch_id, job_id and returns the logs and status etc rather than having it as attributes on the batch or Job object directly.

cseed: > object no longer exists

You mean locally?  Yeah, I thought that might be why this wasn't there already.  Some options:

 - The LocalBackend can't support it, so do nothing.
 - Don't delete the docker images in the local backend (e.g. `docker run` without `--rm`) so they can be queried afterwards.  Maybe make it an option in LocalBackend.  I'm not sure how cleanup gets handled.  Maybe have a hailctl tool to query the docker containers (do they have labels?) to see which batches are there and let you delete them.
 - Capture the docker state in the local backend before deleting.  Doing this with the logs seems problematic.
 - Don't change the interface, but make private methods to get status and log on the service backend that we can use in the tests.
 - Have the log and status functions throw an support exception in the local backend.

Has anyone asked for a local batch feature like this?  An internal functions for the service backend seem like a no brainer.  Then we can think about how/whether we should support them on the local backend.

Separately, how do you know if a LocalBackend batch suceeds or fails?  It looks like LocalBackend._run doesn't return.  We should probably document more clearly what the Batch.run return value is.

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -88,6 +88,25 @@ def test_single_job(self):
...
+            b.write_output(j.ofile, output_file.name)
+            b.run()
+
+            assert self.read(output_file.name) == msg
+
+    def test_single_job_with_nonsense_shell(self):
+        b = self.batch()
+        j = b.new_job(shell='bin/ajdsfoijasidojf')
+        j.command(f'echo "hello"')
+        self.assertRaises(Exception, b.run())
```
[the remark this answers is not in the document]
akotlar: Good call. It takes a callable, should not have been invoked here. Can't use .status() here because this is a LocalBackend test and sp.check_call raises. It would be nice to have similar .status() available on LocalBackend.

hail/python/test/hailtop/batch/test_batch.py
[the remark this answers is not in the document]
akotlar: Thanks for the explanation. I've dismissed your last review after addressing the shel testing comments. If you prefer that I test entrypoint in this PR as well, I will follow up in the morning, updating creating Dockerfile.entrypoint-test (or use docker build with --build-arg), updating docker/Makefile, batch/Makefile, build.yaml, and adding new tests for local backend and service backend.

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -88,6 +88,26 @@ def test_single_job(self):
...
+            b.run()
+
+            assert self.read(output_file.name) == msg
+
+    def test_single_job_with_nonsense_shell(self):
+        b = self.batch()
+        j = b.new_job(shell='/bin/ajdsfoijasidojf')
+        j.image('google/cloud-sdk:237.0.0-alpine')
+        j.command(f'echo "hello"')
+        self.assertRaises(Exception, b.run)
```
cseed, line 109: This code seems fine, but I'm baffled by something.  Was this test passing when you had `b.run()`?  I explicitly checked and thought so.  If `b.run()` was throwing, that would make the test fail. So I assume `b.run()` wasn't throwing.  But now this is passing, and I just verified locally `b.run()` is throwing in this case.  So why wasn't the test case failing before?  I'm afraid it isn't being run or checked correctly.  Do you know what's happening here?

batch/batch/worker/worker.py
```diff
@@ -264,14 +264,16 @@ def container_config(self):
...
         config = {
             "AttachStdin": False,
             "AttachStdout": False,
             "AttachStderr": False,
             "Tty": False,
             'OpenStdin': False,
             'Cmd': self.spec['command'],
-            'Image': self.image
+            'Image': self.image,
+            'Entrypoint': '',
```
cseed, line 276: Nit: we don't use the final trailing comma.

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -88,6 +88,26 @@ def test_single_job(self):
...
+
+            b.write_output(j.ofile, output_file.name)
+            b.run()
+
+            assert self.read(output_file.name) == msg
+
+    def test_single_job_with_nonsense_shell(self):
+        b = self.batch()
+        j = b.new_job(shell='/bin/ajdsfoijasidojf')
+        j.image('google/cloud-sdk:237.0.0-alpine')
```
cseed, line 107: Why are you overriding the image?  If you need an image, you should use `ubuntu:18.04` or something based on it.  Since this is used (nearly) everywhere, it is the fastest because it is always cached.

Parameterizing this is on the TODO (either when we upgrade to 20.04 or the Google managed images).  When that happens, we'll search everywhere for `ubuntu:18.04` to do the replacement, and this would get missed.

akotlar: Thanks for the explanation, will change (I had it as the google image because that was the default_shell for the BatchTests).

> Why are you overriding the image? 
To trigger the branch of LocalBackend._run that previously specified `/bin/bash`. Without job._image specified, commands are executed directly in a system shell, using `sp.check_call(script, shell=True)`, which defaults to `/bin/sh` on POSIX systems (COMSPEC env variable on Windows). This can be overridden using the `executable` argument, but I didn't want to introduce complexity:

With job._image specified, the docker command is also executed using `sp.check_call(script, shell=True)`, so if we want to use `sp.check_call(script, shell=True, executable=job._shell or job._default_shell)` we'll need to decide whether we check_call's shell to be overridden when job._image is specified. Seemed to be a bit out of scope, but happy to address now, you all probably have a plan here.

hail/python/test/hailtop/batch/test_batch.py
akotlar: It was the missing `j.image(..)`! Without image specified we override nothing, because the command is directly invoked in the system shell (well, as elaborated below in response to the question on the use of j.image, I am not overriding sp.check_call's shell, which defaults to /bin/sh on POSIX system). I thought we wouldn't want to change this behavior, because prior to this PR we were not specifying a shell executable for jobs that did not specify`image`.

cseed: OK!  I'm glad I asked.  I think you should change the other case, too.  I think this is a mistake.  The goal is to run exactly the same command but locally instead of inside a container.  This is bad because you get totally different shell behavior inside and outside a container, and shell is ignored depending on the setting of an independent value (input).  Let's get two tests, image or not.

batch has many fewer users and is much more experimental than the Hail interface.  I don't view the interface is immutable.  @jigold are you OK with this?

hail/python/hailtop/batch/backend.py
```diff
@@ -117,8 +118,7 @@ def _run(self,
 
         tmpdir = self._get_scratch_dir()
 
-        lines = ['#!/bin/bash',
-                 'set -e' + 'x' if verbose else '',
+        lines = ['set -e' + 'x' if verbose else '',
```
akotlar, line 121: when `shell=True` the following occurs:

`Popen(['/bin/sh', '-c', args[0], args[1], ...])`. The shebang has no effect.

hail/python/hailtop/batch/backend.py
```diff
@@ -199,25 +199,30 @@ def symlink_input_resource_group(r):
...
                              f"{cpu} "
-                             f"{job._image} /bin/bash "
-                             f"-c {shq(joined_env + defs + cmd)}")
+                             f"{job._image} "
+                             f"{job_shell} -c {job_script}")
             else:
-                lines += env
-                lines += resource_defs
-                lines += job._command
+                lines.append(f"{job_shell} -c {job_script}")
```
cseed, line 225: Can you put the shq here, or call this quoted_job_script?  Quoting is so important, it should be clear it is happening at the point of use, otherwise this line looks wrong.

hail/python/hailtop/batch/backend.py
cseed: Yep

hail/python/test/hailtop/batch/test_batch.py
jigold: I'm okay with changes to the interface as long as you post the changes to the Zulip forum and a mailing list that we should create (does not exist yet).

cseed: And it should go in the change log.  Are CHANGELOG commit messages the right thing for the Batch change log, too?

jigold: Agreed. The change log behaves the exact same as for Hail Query. Whomever makes the release is supposed to sort out which change logs go to the correct location.

akotlar:
One more change: had to remove plink from batch.py docktest; we weren't previously catching that plink is not in PATH (not include in service_base_image). Alternative is to change the image, but the modified doesn't actually do anything with plink.

hail/python/hailtop/batch/batch.py
```diff
@@ -242,7 +248,6 @@ def read_input_group(self, **kwargs: str) -> ResourceGroup:
         ...                            bim="data/example.bim",
         ...                            fam="data/example.fam")
         >>> j = b.new_job()
-        >>> j.command(f"plink --bfile {bfile} --geno --make-bed --out {j.geno}")
```
danking, line 245: This removes a critical example of using `bfile` without any extension, just add `  # doctest: +SKIP` to the end of this line.

PR #9220, opened 2020-08-05
[query][smm 2] Use stream region management in tests

patrick-schultz:
~~Stacked on #9106.~~

Adds an enumeration to control the allocation strategy in emitted code. There are currently three options, `Default`, `OneRegion`, and `ManyRegions`. `OneRegion` replicates the current behavior, and is the default for now. `ManyRegions` always allocates new regions when given the choice. This PR makes Java tests use the `ManyRegions` strategy to catch more lifetime errors. `Default` is somewhere in the middle, and will become the default. It will use one region (at least a constant number of regions) per row of a table. `Default` currently behaves the same as `ManyRegions`. Future work will implement the per-row behavior.

### DIFF
--- a/hail/src/main/scala/is/hail/backend/spark/SparkBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/spark/SparkBackend.scala
@@ -257,13 +257,27 @@ class SparkBackend(
     case Right((pt, off)) => SafeRow(pt, off).get(0)
   }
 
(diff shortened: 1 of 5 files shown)


patrick-schultz:
Yeah, I was thinking about that. It could be a parameter on all `ExecStrategy`s, but that's a lot of cases to test.

PR #9221, opened 2020-08-05
[query] Add JAVA_HOME configuration to Makefile

chrisvittal:
Furthermore, update build.gradle to (finally) not print all the
Unsafe warnings.


tpoterba:
nice!

PR #9223, opened 2020-08-06
[query] NDArray Slice with Indices Needs Bounds Checks

### LINKED ISSUE #9144: ndarray: Bounds checking on slicing is not working
```python
>>> a = hl.nd.array( [ [1,2,3,4,5], [7,6,8,9, 10] ] ) 
>>> a = a.T 
>>> hl.eval(a[0:a.shape[0],0:5])   

array([[         1,          7,          0,          4,      32667],
       [         2,          6,          5,          0,          2],
       [         3,          8,          0, 2013954072,          0],
       [         4,          9,         20,      32667,          5],
       [         5,         10,          0, 2013954376,          0]],
      dtype=int32)
```

johnc1231:
CHANGELOG: Fixed issue where ndarrays being sliced and indexed into in one expression didn't have sufficient bounds checks

Fixes #9144 by checking if indices provided in a slicing expression that mixes slices and single indices are out of bounds. 

Alex, please tell me if you're able to get anymore errors from #9144 with this change. I wasn't able to, but that issue was not super well defined at first so it's unclear to me if this covers all of your problems. I retitled the issue to reflect my understanding of the problems. 

More detail:

In numpy, it is ok for the upper bound of a slice to go past the end of an array, but it is not ok for an indexing operation to do the same. For example:

```
n = np.array([[1,2,3,4], [1,2,3,4]])
n[0:1000, 0:1000]
```

is allowed, the bounds just get clamped. However, this is not allowed:

```
n[1000, 1000]
```

nor is:

```
n[1000, 0:1000]
```

Hail was not handling that last case with a mix of slices and indices correctly. Namely, it was not throwing an out bounds error on the first axis index.

### DIFF
--- a/hail/python/hail/expr/expressions/typed_expressions.py
+++ b/hail/python/hail/expr/expressions/typed_expressions.py
@@ -3617,7 +3617,7 @@ class NDArrayExpression(Expression):
                         stop = hl.cond(step > 0, dlen, to_expr(-1, tint64))
 
                     slices.append(hl.tuple((start, stop, step)))
-                else:  # Not a slice, just an int index
+                else:
                     adjusted_index = hl.if_else(s < 0, s + dlen, s)
                     checked_int = hl.case().when((adjusted_index < dlen) & (adjusted_index >= 0), adjusted_index).or_error(
                         hl.str("Index ") + hl.str(s) + hl.str(f" is out of bounds for axis {i} with size ") + hl.str(dlen)
--- a/hail/python/hail/expr/expressions/typed_expressions.py
+++ b/hail/python/hail/expr/expressions/typed_expressions.py
@@ -3583,8 +3583,8 @@ class NDArrayExpression(Expression):
         if n_sliced_dims > 0:
             slices = []
             for i, s in enumerate(item):
+                dlen = self.shape[i]
                 if isinstance(s, slice):
-                    dlen = self.shape[i]
 
                     if s.step is not None:
                         step = hl.case().when(s.step != 0, s.step) \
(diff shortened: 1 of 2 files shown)


akotlar:
Nice!

hail/python/hail/expr/expressions/typed_expressions.py
```diff
@@ -3617,8 +3617,12 @@ def __getitem__(self, item):
                         stop = hl.cond(step > 0, dlen, to_expr(-1, tint64))
 
                     slices.append(hl.tuple((start, stop, step)))
-                else:
-                    slices.append(s)
+                else:  # Not a slice, just an int index
```
akotlar, line 3620: We can avoid this comment; the typechecker already requires this to be either a slice or an int64

akotlar:
This also matches the corresponding Numpy error message almost precisely, which is nice. Difference is that "Index" is lowercase for Numpy. It's too bad that we can't keep the error message in python land, and the stack trace is utterly useless to a user. I'll make an issue for this

hail/python/hail/expr/expressions/typed_expressions.py
```diff
@@ -3617,8 +3617,12 @@ def __getitem__(self, item):
...
+                    adjusted_index = hl.if_else(s < 0, s + dlen, s)
+                    checked_int = hl.case().when((adjusted_index < dlen) & (adjusted_index >= 0), adjusted_index).or_error(
+                        hl.str("Index ") + hl.str(s) + hl.str(f" is out of bounds for axis {i} with size ") + hl.str(dlen)
```
akotlar, line 3623: This can be simplified to a single statement I think

```python
.or_error(f"Index {hl.str(s)} is out of bounds for axis {i} with size {hl.str(dlen)}
```

johnc1231: You can't do this. `f` strings get turned into python strings at python runtime, which means you get:

```
Index <StringExpression of type str> is out of bounds for axis 0 with size <StringExpression of type str>'
```

johnc1231: We'd need a special hail string formatting thing. Maybe that exists, I didn't look hard for it.

akotlar: yeah, cool, thanks for the explanation

johnc1231:
I agree with you about the stack traces though. If something expected happens, I don't want one.

akotlar:
needs bump, looks like random failure (shuffler was last test run)

PR #9224, opened 2020-08-06
[query/combiner] Add options to better partition imported GVCFs

chrisvittal:
[query/combiner] Add options to better partition imported GVCFs

Teach the VCF combiner three new options:
  * --genomes/-G to indicate that it is operating on genomes
  * --exomes/-E to indicate that it is operating on exomes
  * --import-interval-size to more finely control the partitioning of
    the reference genome on GVCF import.

These are specified as a required mutually exclusive group, exactly one
of them must be present.

This goes along with removing `default_exome_intervals` and adding
`calculate_even_genome_partitioning`. The new function takes a reference
genome and a target interval size and computes a partitioning of the
reference into intervals of length at most the target size. It does so
by looping over the 'regular' contigs of the reference (as such it only
supports GRCh37 and GRCh38 for now) and dividing each contig's length
by the target size and rounding up, obtaining the number of partitions
the contig will be subdivided into. Then dividing the length of the
contig by the number of partitions, getting the true partition size.
Then using the true partition size, generating the intervals. This
hopefully leads to few degenerate cases where some partitions at the end
of contigs are very small.

### DIFF
--- a/hail/python/hail/experimental/vcf_combiner/__main__.py
+++ b/hail/python/hail/experimental/vcf_combiner/__main__.py
@@ -14,6 +14,15 @@ def main():
     parser.add_argument('out_file', help='Path to final combiner output.')
     parser.add_argument('tmp_path', help='Path to folder for intermediate output '
                                          '(should be an object store path, if running on the cloud).')
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument('--genomes', '-G', dest='is_genomes', action='store_true',
+                       help='Indicates that the combiner is operating on genomes. '
+                            'Affects how the genome is partitioned on input.')
+    group.add_argument('--exomes', '-E', dest='is_genomes', action='store_false',
+                       help='Indicates that the combiner is operating on exomes. '
+                            'Affects how the genome is partitioned on input.')
+    group.add_argument('--import-interval-size', type=int,
+                       help='Interval size for partitioning the reference genome for GVCF import.')
     parser.add_argument('--log', help='Hail log path.')
     parser.add_argument('--header',
                         help='External header, must be readable by all executors. '
@@ -41,6 +50,8 @@ def main():
                  batch_size=args.batch_size,
                  branch_factor=args.branch_factor,
                  target_records=args.target_records,
+                 import_interval_size=args.import_interval_size,
+                 is_genomes=args.is_genomes,
                  overwrite=args.overwrite,
                  reference_genome=args.reference_genome,
                  key_by_locus_and_alleles=args.key_by_locus_and_alleles)
--- a/hail/python/hail/experimental/vcf_combiner/vcf_combiner.py
+++ b/hail/python/hail/experimental/vcf_combiner/vcf_combiner.py
@@ -338,31 +338,55 @@ def calculate_new_intervals(ht, n, reference_genome):
     return intervals
 
 
-@typecheck(reference_genome=reference_genome_type)
-def default_exome_intervals(reference_genome) -> List[hl.utils.Interval]:
-    """create a list of locus intervals suitable for importing and merging exome gvcfs. As exomes
-    are small. One partition per chromosome works well here.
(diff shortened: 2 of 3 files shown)


hail/python/hail/experimental/vcf_combiner/__main__.py
```diff
@@ -20,9 +20,19 @@ def main():
                              'WARNING: if this option is used, the sample names in the '
                              'GVCFs will be overridden by the names in sample map.',
                         required=False)
+    parser.add_argument('--genomes', dest='is_genomes', action='store_true',
```
tpoterba, line 23: should we require the user to add this flag? That seems like a good idea to avoid people unintentionally using exome intervals for genome data.

chrisvittal: The question to me is what we would like the defaults to be. I thought about having genomes be the default, but decided against it while making this change. I have no particular reason or justification for doing so.

The reason I added both `--genomes` and `--exomes` was to better allow users to indicate their intentions in scripts.

Thoughts on the default?

tpoterba: sorry, wasn't clear -- I'm proposing that we don't actually have a default, but instead require that one of the flags is passed explicitly by the user.

chrisvittal: For reference, here's what changed between the force pushes:
```diff
diff --git a/hail/python/hail/experimental/vcf_combiner/__main__.py b/hail/python/hail/experimental/vcf_combiner/__main__.py
index 9a5847372..91092add2 100644
--- a/hail/python/hail/experimental/vcf_combiner/__main__.py
+++ b/hail/python/hail/experimental/vcf_combiner/__main__.py
@@ -14,24 +14,25 @@ def main():
     parser.add_argument('out_file', help='Path to final combiner output.')
     parser.add_argument('tmp_path', help='Path to folder for intermediate output '
                                          '(should be an object store path, if running on the cloud).')
-    group = parser.add_mutually_exclusive_group(required=True)
-    group.add_argument('--genomes', '-G', dest='is_genomes', action='store_true',
-                       help='Indicates that the combiner is operating on genomes. '
-                            'Affects how the genome is partitioned on input.')
-    group.add_argument('--exomes', '-E', dest='is_genomes', action='store_false',
-                       help='Indicates that the combiner is operating on exomes. '
-                            'Affects how the genome is partitioned on input.')
-    group.add_argument('--import-interval-size', type=int,
-                       help='Interval size for partitioning the reference genome for GVCF import.')
     parser.add_argument('--log', help='Hail log path.')
     parser.add_argument('--header',
                         help='External header, must be readable by all executors. '
                              'WARNING: if this option is used, the sample names in the '
                              'GVCFs will be overridden by the names in sample map.',
                         required=False)
+    parser.add_argument('--genomes', dest='is_genomes', action='store_true',
+                        help='Indicates that the combiner is operating on genomes. '
+                             'Affects how the genome is partitioned on input. '
+                             'Ignored if --import-interval-size is used')
+    parser.add_argument('--exomes', dest='is_genomes', action='store_false',
+                        help='Indicates that the combiner is operating on exomes '
+                             '(This is the default). '
+                             'Affects how the genome is partitioned on input. '
+                             'Ignored if --import-interval-size is used')
     parser.add_argument('--branch-factor', type=int, default=CombinerConfig.default_branch_factor, help='Branch factor.')
     parser.add_argument('--batch-size', type=int, default=CombinerConfig.default_batch_size, help='Batch size.')
     parser.add_argument('--target-records', type=int, default=CombinerConfig.default_target_records, help='Target records per partition.')
+    parser.add_argument('--import-interval-size', type=int, help='Interval size for partitioning the reference genome for GVCF import.')
     parser.add_argument('--overwrite', help='overwrite the output path', action='store_true')
     parser.add_argument('--key-by-locus-and-alleles', help='Key by both locus and alleles in the final output.', action='store_true')
     parser.add_argument('--reference-genome', default='GRCh38', help='Reference genome.')
```

PR #9227, opened 2020-08-06
[hailctl] Add hailctl auth user

jigold:
@tpoterba I think this should be fine. It returns a formatted json string that you can use jq to parse.


tpoterba:
oops, thought I approved...

PR #9229, opened 2020-08-06
[query] Makefile: allow folder to be overriden

akotlar:
A commit from a few months ago removed the ability to specify a single file for testing.

### DIFF
--- a/hail/Makefile
+++ b/hail/Makefile
@@ -127,6 +127,7 @@ python/README.md: ../README.md
 $(PYTHON_JAR): shadowJar
 	cp -f $(SHADOW_JAR) $@
 
+PYTEST_ARGS = "test/hail"
 .PHONY: pytest
 pytest: $(PYTHON_VERSION_INFO) $(SHADOW_JAR) $(INIT_SCRIPTS)
 pytest: python/README.md $(PYTHON_JAR)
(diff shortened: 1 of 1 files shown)


hail/Makefile
```diff
@@ -127,6 +127,7 @@ python/README.md: ../README.md
 $(PYTHON_JAR): shadowJar
 	cp -f $(SHADOW_JAR) $@
 
+PYTEST_ARGS = test/hail
```
tpoterba, line 130: doesn't this overwrite whatever PYTEST_ARGS is?

I think we could do something like `PYTEST_TARGET ?= test/hail` and then use that above $(PYTEST_ARGS)

akotlar: This sets a default PYTEST_ARGS. If you pass PYTEST_ARGS to the make command, it will overwrite the default specified.

tpoterba: don't you need `?=` for that?

tpoterba: Also, this would break my setup, where I have `PYTEST_ARGS="--color=yes -vv"`. I'm happy to parameterize the target, we should leave extra args separate

akotlar: ah, would you like a TARGET = ?

tpoterba: see my first comment, yeah

akotlar: > don't you need ?= for that?

You don't for the simple case (make pytest PYTEST_ARGS=" "). The variable specified on the command line will take precedence. It looks like ?= allows you to specify the variable **before** you invoke the make command, which is I think a win, so I'll change to ?=

PR #9233, opened 2020-08-06
[batch] consistently name Docker Hub

akotlar:
In #9176 Dan and I settled on Docker Hub as the preferred name for that service. This PR makes that consistent across batch docs. I left `docker_resources.rst` alone to avoid conflicting with #9176.


tpoterba:
Alex, can you use the PR naming convention for the first line of your commit messages too? Github doesn't use the PR text as the squashed commit message for single-commit PRs, so the untagged/non-descriptive single commit message went into main history:

`we setlled on a name (#9233)`

akotlar:
Yep, will do.

PR #9237, opened 2020-08-07
[combiner] Force users to pass partitioning information to run_combiner

tpoterba:


hail/python/hail/experimental/vcf_combiner/vcf_combiner.py
```diff
@@ -512,13 +512,27 @@ def run_combiner(sample_paths: List[str],
...
-                 is_genomes: bool = False,
+                 use_genome_default_intervals: bool = False,
+                 use_exome_default_intervals: bool = False,
                  overwrite: bool = False,
                  reference_genome: str = 'default',
                  contig_recoding: Optional[Dict[str, str]] = None,
                  key_by_locus_and_alleles: bool = False):
     """Run the Hail VCF combiner, performing a hierarchical merge to create a combined sparse matrix table.
 
+    **Partitioning**
```
chrisvittal, line 523: Consider adding something about how the input partition and output partitions are unrelated unless number of gvcfs is less than the branch_factor.

chrisvittal: As a followup, but may be unnecessary, discuss how smaller and larger numbers of import partitions affect performance. Notably increased parallelism at the cost of re-parsing tabix indices every time and weird object storage performance properties where GCS 'warms up' and can seem to get a little faster overall on long reads.

These come from my observations that as parallelism decreased, core hours decreased, in many cases significantly.

hail/python/hail/experimental/vcf_combiner/vcf_combiner.py
```diff
@@ -568,10 +586,37 @@ def run_combiner(sample_paths: List[str],
...
+        info(f"Using {len(intervals)} user-supplied intervals as partitioning for GVCF import")
+    elif import_interval_size is not None:
+        intervals = calculate_even_genome_partitioning(reference_genome, import_interval_size)
+        info(f"Using {len(intervals)} intervals with user-supplied size"
+             f" {import_interval_size} as partitioning for GVCF import")
+    elif use_genome_default_intervals:
+        size = CombinerConfig.default_genome_interval_size
+        intervals = calculate_even_genome_partitioning(reference_genome, size)
+        info(f"Using {len(intervals)} intervals with default whole-genome size"
+             f" {import_interval_size} as partitioning for GVCF import")
```
chrisvittal, line 612:
```suggestion
             f" {size} as partitioning for GVCF import")
```

hail/python/hail/experimental/vcf_combiner/vcf_combiner.py
```diff
@@ -568,10 +586,37 @@ def run_combiner(sample_paths: List[str],
...
+    elif use_exome_default_intervals:
+        size = CombinerConfig.default_exome_interval_size
+        intervals = calculate_even_genome_partitioning(reference_genome, size)
+        info(f"Using {len(intervals)} intervals with default exome size"
+             f" {import_interval_size} as partitioning for GVCF import")
```
chrisvittal, line 617:
```suggestion
             f" {size} as partitioning for GVCF import")
```

tpoterba:
> I know it's outside the scope of these changes, but how about making all arguments to run_combiner after tmp_path keyword only?

Was thinking the same.

hail/python/hail/experimental/vcf_combiner/vcf_combiner.py
```diff
@@ -568,10 +586,37 @@ def run_combiner(sample_paths: List[str],
...
-    intervals = intervals or calculate_even_genome_partitioning(reference_genome, import_interval_size)
+    n_partition_args = (int(intervals is not None)
+                        + int(import_interval_size is not None)
+                        + int(use_genome_default_intervals)
+                        + int(use_exome_default_intervals))
+
+    if n_partition_args == 0:
+        raise ValueError("'run_combiner': require one argument from 'intervals', 'import_interval_size', "
+                         "'use_genome_default_intervals', or 'use_exome_default_intervals' to choose GVCF partitioning")
+    if n_partition_args > 0:
```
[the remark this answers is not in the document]
tpoterba: oops

hail/python/hail/experimental/vcf_combiner/vcf_combiner.py
tpoterba: This is hard, since I think we don't understand the relationship super well. I think we should communicate that people should generally use the default_$YOUR_DATA_TYPE_intervals

PR #9240, opened 2020-08-07
[query][smm 12] Memory management for k-way merge and zip-join

patrick-schultz:
Stacked on #9235


hail/src/main/scala/is/hail/expr/ir/StagedRegion.scala
```diff
@@ -53,13 +66,41 @@ object StagedRegion {
...
+      def store(rhs: Code[Region]): Code[Unit] = regionArray.update(i, rhs)
+    }
+
+    new StagedOwnedRegionArray {
+      def apply(i: Value[Int]): StagedOwnedRegion = new RealStagedOwnedRegion(get(i), self)
+
+      def allocateRegions(mb: EmitMethodBuilder[_], size: Int): Code[Unit] = {
+        val i = mb.newLocal[Int]("sora_alloc_i")
+        Code(
+          regionArray := Code.newArray(length),
```
tpoterba, line 92: is type inference filling in the type info here?

patrick-schultz: Yeah, `regionArray` is a `ThisFieldRef[Array[Region]]`, so `regionArray.:=` expects a `Code[Array[Region]]`.

PR #9241, opened 2020-08-07
[monitoring] Added monitoring service with billing UI

jigold:
- Added BigQuery client for aiogoogle
- Added monitoring service
- Added billing portion which queries BigQuery for the billing data in broad-ctsa

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -11,6 +11,7 @@ steps:
     - auth-oauth2-client-secret
     - scorecard-github-access-token
     - zulip-config
+    - billing-monitor-gsa-key
  - kind: deploy
    name: deploy_batch_sa
    namespace:
@@ -103,6 +104,8 @@ steps:
        to: /repo/batch/
      - from: /io/repo/notebook/sql
        to: /repo/notebook/
+     - from: /io/repo/monitoring/sql
+       to: /repo/monitoring/
      - from: /io/repo/hail/python/hailtop
        to: /repo/hailtop/
      - from: /io/repo/hail/python/test
@@ -450,6 +453,13 @@ steps:
    dockerFile: benchmark-service/Dockerfile
    contextPath: .
    publishAs: benchmark
+   dependsOn:
+     - service_base_image
+ - kind: buildImage
+   name: monitoring_image
+   dockerFile: monitoring/Dockerfile
+   contextPath: .
+   publishAs: monitoring
    dependsOn:
      - service_base_image
  - kind: buildImage
@@ -1330,6 +1340,78 @@ steps:
     - deploy_router
     - create_certs
     - deploy_auth
+    - create_accounts
+ - kind: runImage
+   name: check_monitoring
+   image:
+     valueFrom: monitoring_image.image
+   script: |
+     set -ex
+     python3 -m flake8 monitoring
+     python3 -m pylint --rcfile pylintrc monitoring
+   dependsOn:
+     - monitoring_image
+ - kind: runImage
+   name: delete_monitoring_tables
+   image:
+     valueFrom: service_base_image.image
+   script: |
+     set -ex
+     mysql --defaults-extra-file=/sql-config/sql-config.cnf < /io/sql/delete-monitoring-tables.sql
+   inputs:
+    - from: /repo/monitoring/sql
+      to: /io/
+   secrets:
+    - name: database-server-config
+      namespace:
+        valueFrom: default_ns.name
+      mountPath: /sql-config
+   runIfRequested: true
+   scopes:
+    - dev
+   dependsOn:
+    - default_ns
+    - service_base_image
+    - copy_files
+ - kind: createDatabase
+   name: monitoring_database
+   databaseName: monitoring
+   migrations:
+    - name: initial
+      script: /io/sql/initial.sql
+   inputs:
+    - from: /repo/monitoring/sql
+      to: /io/
+   namespace:
+     valueFrom: default_ns.name
+   shutdowns:
+    - kind: Deployment
+      namespace:
+        valueFrom: default_ns.name
+      name: monitoring
+   dependsOn:
+    - default_ns
+    - copy_files
+    - delete_monitoring_tables
+ - kind: deploy
+   name: deploy_monitoring
+   namespace:
+     valueFrom: default_ns.name
+   config: monitoring/deployment.yaml
+   wait:
+    - kind: Service
+      name: monitoring
+      for: alive
+   scopes:
+    - deploy
+    - dev
+   dependsOn:
+    - default_ns
+    - monitoring_image
+    - monitoring_database
+    - deploy_auth
+    - deploy_router
+    - create_certs
     - create_accounts
  - kind: runImage
    name: test_auth_copy_paste_login
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -431,6 +431,14 @@ async def run_if_changed(changed, f, *args, **kwargs):
             await changed.wait()
 
 
+async def run_if_changed_idempotent(changed, f, *args, **kwargs):
+    while True:
+        should_wait = await f(*args, **kwargs)
+        changed.clear()
+        if should_wait:
+            await changed.wait()
+
+
 class LoggingTimerStep:
     def __init__(self, timer, name):
         self.timer = timer
--- a/monitoring/deployment.yaml
+++ b/monitoring/deployment.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: apps/v1
 kind: Deployment
 metadata:
--- a/monitoring/deployment.yaml
+++ b/monitoring/deployment.yaml
@@ -1,0 +1,84 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: monitoring
+  labels:
+    app: monitoring
+    hail.is/sha: "{{ code.sha }}"
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: monitoring
+  template:
+    metadata:
+      labels:
+        app: monitoring
+        hail.is/sha: "{{ code.sha }}"
+    spec:
+      nodeSelector:
+        preemptible: "true"
+      tolerations:
+       - key: preemptible
+         value: "true"
+      containers:
+        - name: monitoring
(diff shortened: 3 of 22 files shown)


build.yaml
```diff
@@ -452,6 +455,13 @@ steps:
    publishAs: benchmark
    dependsOn:
      - service_base_image
+ - kind: buildImage
+   name: monitoring_image
+   dockerFile: monitoring/Dockerfile
+   contextPath: .
```
danking, line 461: This docker file only depends on the contents of `monitoring`. Let's be specific about that.

monitoring/deployment.yaml
```diff
@@ -0,0 +1,84 @@
+---
```
danking, line 1: this leading --- is unnecessary

monitoring/sql/estimated-current.sql
```diff
@@ -0,0 +1,16 @@
...
+  `mark` BIGINT
+) ENGINE = InnoDB;
+
+INSERT INTO `monitoring_billing_mark` (mark) VALUES (NULL);
+
+CREATE TABLE IF NOT EXISTS `monitoring_billing_data` (
+  `year` INT NOT NULL,
+  `month` INT NOT NULL,
+  `service_description` VARCHAR(100) NOT NULL,
+  `sku_description` VARCHAR(100) NOT NULL,
```
danking, line 11: I'm a bit uncomfortable that we're using "descriptions" rather than some, like, unique name. Can you comment on why descriptions are OK? One concrete concern I have is, what if google changes the description name, that will make it harder to link costs from before and after the description change.

monitoring/monitoring/monitoring.py
```diff
@@ -0,0 +1,219 @@
...
+    for record in records:
+        cost_by_sku_label.append(record)
+
+        cost_by_service[record['service_description']] += record['cost']
+
+        if record['service_description'] == 'Compute Engine':
+            compute_cost_breakdown[record['source']] += record['cost']
+
+    cost_by_service = sorted([{'service': k, 'cost': v} for k, v in cost_by_service.items()], key=lambda x: x['cost'], reverse=True)
+    compute_cost_breakdown = sorted([{'source': k, 'cost': v} for k, v in compute_cost_breakdown.items()], key=lambda x: x['cost'], reverse=True)
```
danking, line 52: These lines are a mouthful! Can we try newlines between the arguments? I think that might make it easier to see what's happening.

monitoring/monitoring/monitoring.py
```diff
@@ -0,0 +1,219 @@
...
+        return ([], [], [], time_period_query)
+
+    try:
+        time_period = datetime.datetime.strptime(time_period_query, date_format)
+        month = time_period.month
+        year = time_period.year
+        if time_period < datetime.datetime.strptime('07/2020', date_format):
+            return await parse_error(f"Invalid value for time_period '{time_period_query}'; time periods before 07/2020 are not supported.")
+    except ValueError:
+        return await parse_error(f"Invalid value for time_period '{time_period_query}'; must be in the format of MM/YYYY.")
```
danking: I think you intend to only catch ValueErrors coming from the first line of the try, right? I'd prefer we tighten the try-except to just wrap that line.

monitoring/monitoring/monitoring.py
```diff
@@ -0,0 +1,219 @@
...
+    compute_cost_breakdown = defaultdict(lambda: 0)
+    cost_by_sku_label = []
+
+    for record in records:
+        cost_by_sku_label.append(record)
+
+        cost_by_service[record['service_description']] += record['cost']
+
+        if record['service_description'] == 'Compute Engine':
+            compute_cost_breakdown[record['source']] += record['cost']
```
danking, line 49: should we assert `record['source']` is None if its not compute engine?

monitoring/monitoring/monitoring.py
```diff
@@ -0,0 +1,219 @@
...
+SELECT service.description as service_description, sku.description as sku_description, SUM(cost) as cost,
+CASE
+  WHEN EXISTS(SELECT 1 FROM UNNEST(labels) WHERE key = "namespace" and value = "default") THEN "batch-production"
+  WHEN EXISTS(SELECT 1 FROM UNNEST(labels) WHERE key = "namespace" and value LIKE '%pr-%') THEN "batch-test"
+  WHEN EXISTS(SELECT 1 FROM UNNEST(labels) WHERE key = "namespace") THEN "batch-dev"
+  WHEN EXISTS(SELECT 1 FROM UNNEST(labels) WHERE key = "role" and value LIKE 'vdc') THEN "k8s"
+  WHEN service.description = "Compute Engine" THEN "unknown"
+  ELSE NULL
+END AS source
+FROM `broad-ctsa.hail_billing.gcp_billing_export_v1_0055E5_9CA197_B9B894`
```
danking, line 127: this is just some random identifier associate with our billing project? And it gets updated continuously by the Google billing infrastructure?

monitoring/monitoring/monitoring.py
```diff
@@ -0,0 +1,219 @@
...
+WHERE DATE(_PARTITIONTIME) >= "{start_str}" AND DATE(_PARTITIONTIME) <= "{end_str}" AND project.name = "hail-vdc" AND invoice.month = "{invoice_month}"
```
danking, line 128: Can you tell me more about what _PARTITIONTIME is and why we need to use it here? Seems like invoice.month would be sufficient, but I assume not.

monitoring/monitoring/monitoring.py
```diff
@@ -0,0 +1,219 @@
...
+GROUP BY service_description, sku_description, source;
+'''
+
+        records = [x async for x in await bigquery_client.query(cmd)]
+
+        data = [(year, month, record['service_description'], record['sku_description'], record['source'], record['cost'])
+                for record in records]
```
danking, line 135: seems like we ought to do:
```
records_agen = await bigquery_client.query(cmd)
data = [...
        async for record in records_agen]
```

yeah? Or am I misunderstanding something here?

monitoring/monitoring/monitoring.py
```diff
@@ -0,0 +1,219 @@
...
+async def polling_loop(app):
+    db = app['db']
+
+    while True:
+        try:
+            now = datetime.datetime.now()
+            row = await db.select_and_fetchone('SELECT mark FROM monitoring_billing_mark;')
+            if not row['mark'] or now > datetime.datetime.fromtimestamp(row['mark'] / 1000) + datetime.timedelta(days=1):
+                app['query_billing_event'].set()
+                log.info('set event')
```
danking, line 175: maybe go with "query_billing_event set" or "query billing loop notified"

monitoring/monitoring/monitoring.py
```diff
@@ -0,0 +1,219 @@
...
+        except Exception:
+            log.exception('error while polling for billing mark', exc_info=True)
```
danking, line 177: I think you can ditch the `try-except` and use `retry_long_running` (but without `run_if_changed_idempotent`).

danking: e.g. `bump_loop` in batch's scheduler

danking:
Hey @catoverdrive! We decided this PR might be a good opportunity for you to dig in to the implementation of a brand new service.

On services team, our PRs are part-critique and part-education. I did a quick, initial review that you can scan over to see how we tend to do it.

monitoring/sql/estimated-current.sql
jigold: This is the schema of the google billing table. I could add the sku.id and service.id to the tables we store, but I think we want to display the descriptions.   https://cloud.google.com/billing/docs/how-to/export-data-bigquery-tables#data-schema

monitoring/monitoring/monitoring.py
jigold: exactly.

monitoring/monitoring/monitoring.py
jigold: I think it's not sufficient and we'll scan all the data. I did a test query plan for July and without the partition time it will scan 128 MB instead of 32 MB. Some examples have the partition time and others don't. https://cloud.google.com/billing/docs/how-to/bq-examples#query-by-project-by-invoice

https://cloud.google.com/bigquery/docs/querying-partitioned-tables#ingestion-time_partitioned_table_pseudo_columns

build.yaml
jigold: I don't understand this comment. We need functions in hailtop and gear in this Docker image.

monitoring/monitoring/monitoring.py
jigold: I added a check to make the source "Compute Engine" specific. Kubernetes Engine was getting labelled as 'k8s'.

jigold:
@catoverdrive Bump.

hail/python/hailtop/utils/utils.py
```diff
@@ -431,6 +431,14 @@ async def run_if_changed(changed, f, *args, **kwargs):
             await changed.wait()
 
 
+async def run_if_changed_idempotent(changed, f, *args, **kwargs):
```
catoverdrive, line 434: I spend a while trying to figure out this function, so just to make sure I understand correctly:

The difference here is that we clear the event flag after running the function instead of before. So the "idempotent" here refers to `f` being idempotent? As in, `f` might/will set the event flag, but running it twice in a row produces the same result as running it once, so we don't want to run it again if the event flag gets set from running `f` the first time?

jigold: I didn't want the behavior if the event is set in the middle of while we're querying BigQuery to have the query repeat itself because the clear happened before the function completed.

jigold:
I put the WIP tag on this. I don't have the energy to debug any failures today. Will merge it on Monday.

jigold:
There's something wrong with this PR. The database migration step says it's successful, but the new database is never actually created. I think this is the same thing I saw with dev deploy and attributed it to the wrong cause.

```
+------------------------------------+
| Database                           |
+------------------------------------+
...
| pr-9241-auth-zdyt4a4geys7          |
| pr-9241-batch-y0qvw1vpniad         |
| pr-9241-ci-nkuua31y7nxn            |
| pr-9241-test-instance-zxxeu6gotctw |
...
```

PR #9246, opened 2020-08-10
[letsencrypt] add benchmark to hail.is certificate domains

danking:


danking:
I will run lets encrypt now.

letsencrypt/Makefile
```diff
@@ -19,10 +19,7 @@ push: build
 	docker tag letsencrypt gcr.io/$(PROJECT)/letsencrypt
 	docker push gcr.io/$(PROJECT)/letsencrypt
 
-start-service: service.yaml
-	kubectl -n default apply -f service.yaml
-
-run: letsencrypt-pod.yaml service.yaml push
+run: letsencrypt-pod.yaml push
```
jigold, line 22: What happened to the service?

danking:
I restored the service.yaml. I think it was deleted when I created the TLS PR. My bad!

PR #9249, opened 2020-08-10
[ci] fix create_passwords for dev deploy

jigold:
I don't know how to test this works unless I can dev deploy to my own copy of CI that's running with the new changes.

The issue I was seeing is the database step runs fine, but the new tables weren't actually created.


jigold:
I agree. This needs more work!

jigold:
@danking Sorry. Can you reapprove? I think this will work and reverts to the original behavior. Once this merges, then I'll see if there's something else that needs to be fixed.

PR #9250, opened 2020-08-11
[hailctl dataproc] Replace deprecated gcloud arguments in hailctl dataproc start and modify

nawatts:
[gcloud 284.0.0](https://cloud.google.com/sdk/docs/release-notes#28400_2020-03-10) deprecated the `--num-preemptible-workers` and `--preemptible-worker-boot-disk-size` arguments to `gcloud dataproc clusters create`.

[gcloud 285.0.0](https://cloud.google.com/sdk/docs/release-notes#28500_2020-03-17) deprecated the `--num-preemptible-workers` argument to `gcloud dataproc clusters update`.

This replaces `--num-preemptible-workers` with `--num-secondary-workers` and `--preemptible-worker-boot-disk-size` with `--secondary-worker-boot-disk-size` in calls to `gcloud` from `hailctl dataproc start` and `hailctl dataproc modify`.

Since the new arguments were added in gcloud version 285.0.0 (released in March 2020), this also adds a requirement that gcloud be at least version 285.0.0. Alternatively, the arguments could be switched based on the gcloud version.

### DIFF
--- a/hail/python/hailtop/hailctl/dataproc/cli.py
+++ b/hail/python/hailtop/hailctl/dataproc/cli.py
@@ -5,11 +5,15 @@ import argparse
 from . import connect
 from . import describe
 from . import diagnose
+from . import gcloud
 from . import list_clusters
 from . import modify
 from . import start
 from . import stop
 from . import submit
+
+
+MINIMUM_REQUIRED_GCLOUD_VERSION = (285, 0, 0)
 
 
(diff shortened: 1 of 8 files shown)


nawatts:
Thanks. Didn't know what the protocol was and didn't want to be pushy.

danking:
We respond well to pushy 😉

PR #9255, opened 2020-08-11
[batch] sort attempts by start time

jigold:
I checked this works with dev deploy


batch/batch/front_end/front_end.py
```diff
@@ -1082,6 +1082,9 @@ async def _get_attempts(app, batch_id, job_id, user):
         raise web.HTTPNotFound()
     if len(attempts) == 1 and attempts[0]['attempt_id'] is None:
         return None
+
+    attempts.sort(key=lambda x: x['start_time'])
+
```
danking, line 1087: whitespace!

PR #9258, opened 2020-08-11
[notebook] notebooks are NOT preemptible

danking:


notebook/notebook/notebook.py
```diff
@@ -162,7 +162,7 @@ async def start_pod(k8s, service, userdata, notebook_token, jupyter_token):
         tolerations=[
             kube.client.V1Toleration(
                 key='preemptible',
-                value='true')
+                value='false')
```
cseed, line 165: As per our discussion on Zulip: this should not have a toleration, but it should select non-preemptible nodes.  CI's deployment is the canonical example here.  Just:

```
      nodeSelector:
        preemptible: "false"
```

PR #9259, opened 2020-08-12
[auth] teach auth to specify SSL for SQL secrets

danking:
In https://github.com/hail-is/hail/pull/9113, I forced the auth driver to use
the modern, TLS-required, SQL config format. I incorrectly forgot to specify the
TLS file paths. Luckily, when I tried to create a user account for Patrick
Cummings, instead of creating a broken secret, the auth driver
error'ed. Moreover, the clean up code was broken. As a result, Patrick's account
was stuck in `creating`.

This PR fixes both the clean up code issue (I set `self.namespace` in
`K8sSecretResource`) and specifies the TLS file paths (see driver.py near
line 217).

In addition, this PR attempts to avoid future problems with the sql
configuration by codifying the required components as a NamedTuple, `SQLConfig`. I also
co-located all the parsing and transformation logic between JSON, dicts, and CNF
in the `SQLConfig` class.

I traced back all the users of `create_secret_data_from_config` to ensure they
all now use SQLConfig. I added lots of type annotations, but those won't do
anything right now because we don't have mypy enabled for hailtop.auth.

---

There's a separate issue of us not getting notified that Patrick's account was
not being created due to an error. The relevant logs are linked below. I'm glad
we're starting work on better monitoring. Hopefully error logs like these will
trigger emails to services team.

https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.container_name%3D%22auth-driver%22;timeRange=2020-08-11T15:44:00.000Z%2F2020-08-11T23:55:00.000Z?project=hail-vdc&query=%0A

Moreover, the infinite retry of his account created tens of google service
accounts that were not cleaned up. I do not yet understand why the google
service account clean up code failed. The clean up code bug that I *do* fix in
this PR addresses the GSA secret and the tokens secret.

### DIFF
--- a/auth/auth/driver/driver.py
+++ b/auth/auth/driver/driver.py
@@ -9,7 +9,7 @@ import asyncio
 import aiohttp
 import kubernetes_asyncio as kube
 from hailtop.utils import time_msecs
-from hailtop.auth.sql_config import create_secret_data_from_config
+from hailtop.auth.sql_config import create_secret_data_from_config, SQLConfig
 from hailtop import aiogoogle
 from gear import create_session, Database
 
@@ -114,6 +114,7 @@ class K8sSecretResource:
                     for k, v in data.items()
                 }))
         self.name = name
+        self.namespace = namespace
 
     async def _delete(self, name, namespace):
         try:
@@ -202,7 +203,7 @@ GRANT ALL ON `{name}`.* TO '{name}'@'%';
 
     def secret_data(self):
         with open('/database-server-config/sql-config.json', 'r') as f:
-            server_config = json.loads(f.read())
+            server_config = SQLConfig.from_json(f.read())
         with open('/database-server-config/server-ca.pem', 'r') as f:
             server_ca = f.read()
         with open('/database-server-config/client-cert.pem', 'r') as f:
@@ -217,15 +218,17 @@ GRANT ALL ON `{name}`.* TO '{name}'@'%';
         assert self.name is not None
         assert self.password is not None
 
-        config = {
-            'host': server_config['host'],
-            'port': server_config['port'],
-            'user': self.name,
-            'password': self.password,
-            'instance': server_config['instance'],
-            'connection_name': server_config['connection_name'],
-            'db': self.name
-        }
+        config = SQLConfig(
+            host=server_config.host,
+            port=server_config.port,
+            user=self.name,
+            password=self.password,
+            instance=server_config.instance,
+            connection_name=server_config.connection_name,
+            db=self.name,
+            ssl_ca='/sql-config/server-ca.pem',
+            ssl_cert='/sql-config/client-cert.pem',
+            ssl_key='/sql-config/client-key.pem')
         return create_secret_data_from_config(
             config, server_ca, client_cert, client_key)
 
--- a/hail/python/hailtop/auth/sql_config.py
+++ b/hail/python/hailtop/auth/sql_config.py
@@ -1,22 +1,94 @@
+from typing import Dict, NamedTuple, Optional, Any
 import json
+import os
 
 
-def create_secret_data_from_config(config, server_ca, client_cert, client_key):
-    assert all(config.get(x) is not None
-               for x in ('host', 'user', 'password', 'db', 'ssl-ca', 'ssl-cert',
-                         'ssl-key', 'ssl-mode')), config.keys()
-    secret_data = dict()
-    secret_data['sql-config.json'] = json.dumps(config)
-    secret_data['sql-config.cnf'] = f'''[client]
-host={config["host"]}
-user={config["user"]}
-password="{config["password"]}"
-database={config["db"]}
-ssl-ca={config["ssl-ca"]}
-ssl-cert={config["ssl-cert"]}
-ssl-key={config["ssl-key"]}
-ssl-mode={config["ssl-mode"]}
+class SQLConfig(NamedTuple):
+    host: str
+    port: int
+    user: str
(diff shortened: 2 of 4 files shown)


auth/auth/driver/driver.py
```diff
@@ -114,6 +114,7 @@ async def create(self, name, namespace, data):
                     for k, v in data.items()
                 }))
         self.name = name
+        self.namespace = namespace
```
jigold, line 117: This interface seems confusing to me. Why would we initialize the name and namespace in the class if we're just going to use the one the user passes in `create`?

hail/python/hailtop/auth/sql_config.py
```diff
@@ -1,22 +1,94 @@
...
-ssl-cert={config["ssl-cert"]}
-ssl-key={config["ssl-key"]}
-ssl-mode={config["ssl-mode"]}
+class SQLConfig(NamedTuple):
+    host: str
+    port: int
+    user: str
+    password: str
+    instance: str
+    connection_name: str
```
jigold, line 12: Remind me -- is this new syntax going to force an upgrade of the Hail version required by users?

hail/python/hailtop/auth/sql_config.py
```diff
@@ -1,22 +1,94 @@
...
+    def from_json(s: str) -> 'SQLConfig':
+        return SQLConfig.from_dict(json.loads(s))
+
+    @staticmethod
+    def from_dict(d: Dict[str, Any]) -> 'SQLConfig':
+        for k in ('host', 'user', 'password', 'ssl-ca', 'ssl-cert', 'ssl-key', 'ssl-mode'):
+            assert k in d, f'{k} should be in {d}'
+            assert d[k] is not None, f'{k} should not be None in {d}'
+        return SQLConfig(d['host'],
+                         d['port'],
```
jigold, line 72: I prefer you put the name of the arguments here to avoid errors in the future.

hail/python/hailtop/auth/sql_config.py
```diff
@@ -1,22 +1,94 @@
...
+        if not os.path.isfile(self.ssl_ca):
+            raise ValueError(f'specified ssl-ca, {self.ssl_ca}, does not exist')
+
+    @staticmethod
+    def from_json(s: str) -> 'SQLConfig':
+        return SQLConfig.from_dict(json.loads(s))
+
+    @staticmethod
+    def from_dict(d: Dict[str, Any]) -> 'SQLConfig':
+        for k in ('host', 'user', 'password', 'ssl-ca', 'ssl-cert', 'ssl-key', 'ssl-mode'):
```
jigold, line 68: Should this check be part of `check`?

auth/auth/driver/driver.py
danking: Cotton wrote this, some I'm inferring his intentions, but:

K8sSecretResource can represent a Secret that exists or does not yet exist. The top of this method asserts both name and namespace are `None`. When we're creating a new secret we do this:
```python
        tokens_secret = K8sSecretResource(k8s_client)
        cleanup.append(tokens_secret.delete)
        await tokens_secret.create(tokens_secret_name, BATCH_PODS_NAMESPACE, {
```
So even if `create` works well enough to create a k8s resource, but somehow triggers an error afterwards, we'll still delete the secret.

If I was designing this, I'd probably create a factory method:
```python3
@staticmethod
def create(k8s_client, name, namespace, data):
    try:
        secret = K8sSecretResource(k8s_client, name, namespace)
        await secret._delete()
        await self.k8s_client.create_namespace_secret(...)
        return secret
    except:
        await secret.delete()
```

This makes `create` idempotent and ... failure-proof? Idempotence is about running `f` twice being equivalent to running `f` once. We want one more property: `f` failing ("running" zero times) is equivalent to `f` not being invoked. Moreover *every* instance of `K8sSecretResource` has a well defined name and namespace.

I can transform all the resources in this file to follow that style if you think it is an improvement. You and I (and, over time, Arcturus and Dania) own `auth`, so its future is up to us.

hail/python/hailtop/auth/sql_config.py
danking: Nah, it's in 3.6: https://docs.python.org/3.6/library/typing.html#typing.NamedTuple. It's not in 3.5: https://docs.python.org/3.5/library/typing.html#typing.NamedTuple.

The issue from before was my use of `re.Pattern` vs `typing.Pattern`.

But good looking out for version incompatibilities!

hail/python/hailtop/auth/sql_config.py
danking: I added these checks to `check`, but left them here too. `check` also verifies that the referenced files exist, which I don't want to require on `from_dict` because sometimes we mount the sql_config as `/database-server-config/` even though the `sql_config`'s ssl-ca says `/sql-config/...`. We only do that when we're mounting the database-server-config to use as a template for creating dev namespace database-server-configs.

hail/python/hailtop/auth/sql_config.py
danking: done.

auth/auth/driver/driver.py
jigold: Save it for a future PR.

danking:
bump

danking:
bump

jigold:
I put the WIP tag on it. Merge at your leisure in case there are problems.

PR #9261, opened 2020-08-12
[query] Fix multiplicity used in TableMapPartitions

tpoterba:


tpoterba:
cc @patrick-schultz

patrick-schultz:
Definitely hacky, but this looks right as far as it goes.

PR #9262, opened 2020-08-12
[hailctl] address pesky deprecation warnings

danking:
CHANGELOG: Teach `hailctl dataproc` to use new `gcloud` flag names which suppresses the warnings about `--num-secondary-workers`. `hailctl` now requires at least gcloud 284.0.0. Run `gcloud components update` to update.

This has been available since March, it seems high time to fix this and recommend people update gcloud.


nawatts:
Hey @danking. I've also got a PR open for this issue (#9250) which includes a check for the minimum required version of `gcloud`. Also, including the `--secondary-worker-type` flag increases the minimum `gcloud` version to 291.0.0 (released in May).

PR #9263, opened 2020-08-12
[hailctl] improve accessibility of max age and expiration time

danking:
CHANGELOG: Teach `hailctl dataproc start` about `--expiration-time`. Teach `hailctl dataproc modify` about `--no-max-idle`, `no-max-age`, `--max-age`, and `--expiration-time`.

These flags were always available as pass-throughs. This change adds them to the
help documentation and improves error messages in the case that an incompatible
set of arguments are provided together.


danking:
I rolled the scorecard dice for an assignee.

PR #9273, opened 2020-08-14
batch `make deploy` could push stale image

cseed:
Context: `batch/Makefile` is for manually building and deploying batch from your local machine.  This is rare, and is done when we are bootstrapping a new cluster, or something goes wrong and we need to manually intervene.    The service Makefiles have three main targets: `build`, `push` and `deploy`.  `build` builds docker images for the service locally.  `push` pushes them to the Google Container Repository.  `deploy` deploys the new image to Kubernetes.

The `build` target builds a new image, and BATCH_IMAGE uses docker to get the current image.  `:=` means evaluate once.  This means the JINJA_ENVIRONMENT had the image when `make` was invoked.  So if a new image was built, deploy would deploy the wrong image.

@danking @jigold I wonder if this might explain surprising behavior when you were working on the batch issue earlier today.

### DIFF
--- a/batch/Makefile
+++ b/batch/Makefile
@@ -46,13 +46,14 @@ push: build
 	docker tag batch-copy $(BATCH_COPY_IMAGE)
 	docker push $(BATCH_COPY_IMAGE)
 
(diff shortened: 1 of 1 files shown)


danking:
Ahah! And this was indeed my bad. I added `JINJA_ENVIRONMENT`.

danking:
A project I've been ruminating on for quite some time now is to implement a local-mode version of build.py. I think it might actually be quite easy. Then we can axe all the makefiles and have one, canonical definition of how to build hail.

PR #9278, opened 2020-08-14
[query] Add ArrayExpression.grouped

johnc1231:
This PR exposes `StreamGrouped` IR node in python as a `grouped(group_size)` method on `ArrayExpression`.


johnc1231:
Bump

PR #9282, opened 2020-08-17
[Benchmark] added indices and axis labels plot tags

Dania-Abuhijleh:
Screen Shot 2020-08-19 at 10 55 18 AM


benchmark-service/benchmark/benchmark.py
```diff
@@ -71,7 +72,11 @@ async def show_name(request: web.Request, userdata) -> web.Response:  # pylint:
                      None)
 
     try:
-        fig = px.scatter(x=enumerate_list_index(name_data['trials']), y=name_data['times'])
+        df = pd.DataFrame(dict(trial=enumerate_list_index(name_data['trials']),
```
jigold, line 75: The format we've had for something like this is to do the following:

```
try:
     d = {
         'trial': enumerate_list_index(name_data['trials']),
         'wall_time': name_data['times'],
         'index': ['{}'.format(i+1) for i in range(len(name_data['times']))]
     }
     df = pd.DataFrame(d)
```

benchmark-service/benchmark/benchmark.py
```diff
@@ -71,7 +72,11 @@ async def show_name(request: web.Request, userdata) -> web.Response:  # pylint:
...
+                               wall_time=name_data['times'],
+                               index=['{}'.format(i+1) for i in range(len(name_data['times']))]))
```
jigold, line 77: It's hard for me to tell from the picture. This is the within group index -- correct? I think you might need to modify `enumerate_list_index` to give both the index of the group, the wall time, and within the group index. Then everything should come from `name_data['trials']`.

PR #9283, opened 2020-08-17
[combiner] fix import_gvcfs call in run_combiner

chrisvittal:
`import_gvcfs` makes no assumptions about the number of samples in
a file, and so external sample names must be a list of lists.


chrisvittal:
ugh, ok

PR #9284, opened 2020-08-17
[Benchmark] add page to compare two benchmarks

Dania-Abuhijleh:
Screen Shot 2020-08-17 at 10 52 59 AM

Benchmarks with different tests:
Screen Shot 2020-08-19 at 9 55 34 AM


benchmark-service/benchmark/benchmark.py
```diff
@@ -57,6 +59,44 @@ def get_benchmarks(file_path):
     return benchmarks
 
 
+def get_comparisons(benchmarks1, benchmarks2, metric):
+    def get_metric(data):
+        if metric == 'median':
+            return data['median']
+        return min(data['times'])  # metric == 'best'
```
jigold, line 66: Our team code style is to add this here:

```
assert metric == 'best'
```

benchmark-service/benchmark/benchmark.py
```diff
@@ -57,6 +59,44 @@ def get_benchmarks(file_path):
...
+
+    comparisons = []
+    for d in benchmarks1['data']:
+        for item in benchmarks2['data']:
+            if item['name'] == d['name']:
+                comparisons.append((d['name'], get_metric(d), get_metric(item)))
```
jigold, line 72: Did you want an inner join here? I think it might be nice to show the outer join on metrics and have None for the comparisons if a benchmark isn't in a different benchmarks file.

benchmark-service/Dockerfile
```diff
@@ -11,6 +11,10 @@ RUN pip3 install plotly
 
 RUN pip3 install pandas
 
+RUN pip3 install scipy
+
```
jigold, line 15: It's possible these are already in the base image. Did you try them out?

benchmark-service/benchmark/benchmark.py
```diff
@@ -100,6 +140,29 @@ async def index(request, userdata):  # pylint: disable=unused-argument
...
 
 
+@router.get('/compare')
+@web_authenticated_developers_only(redirect=False)
+async def compare(request, userdata):  # pylint: disable=unused-argument
+    file1 = request.query.get('file1')
+    file2 = request.query.get('file2')
+    metric = request.query.get('metrics')
+    if file1 is None or file2 is None:
+        benchmarks_context1 = None
```
jigold, line 150: Should it be None, or should this raise an error, or should it only fill in the data for the files it does have?

benchmark-service/benchmark/benchmark.py
```diff
@@ -100,6 +140,29 @@ async def index(request, userdata):  # pylint: disable=unused-argument
...
+        benchmarks_context2 = None
+        comparisons = None
+    else:
+        benchmarks_context1 = get_benchmarks(file1)
```
jigold, line 154: What happens if you put a garbage file name in here? Do you get a 500 or an error message?

benchmark-service/Dockerfile
Dania-Abuhijleh: it got mad when I didn't have those lines, not sure if there's another way around it

benchmark-service/benchmark/benchmark.py
Dania-Abuhijleh: error message

benchmark-service/benchmark/benchmark.py
Dania-Abuhijleh: I'm thinking we can show the outer join and have a button that filters to show only the ones in common

benchmark-service/benchmark/benchmark.py
Dania-Abuhijleh: Right now if you don't enter a file name it doesn’t let you press compare saying "please fill out", but if you enter a file name that doesn't exist it gives a page with an error message saying "cannot find file <name>"

benchmark-service/benchmark/benchmark.py
jigold: That would be good if it's easy to implement. Otherwise, I think the outer join is sufficient.

benchmark-service/Dockerfile
jigold: oh, i was thinking of a different docker image. You're right, you need to install it here.

benchmark-service/benchmark/benchmark.py
```diff
@@ -57,6 +59,58 @@ def get_benchmarks(file_path):
...
 
+def get_comparisons(benchmarks1, benchmarks2, metric):
+    def get_metric(data):
+        if metric == 'median':
+            return data['median']
+        assert metric == 'best'
+        return min(data['times'])
+
+    comparisons = []
+    for d in benchmarks1['data']:
```
jigold, line 70: So this is one way to implement an outer join, but it's not very efficient because you are doing O(n^2) work. There's three ways to do this efficiently. One is sort the keys for each dictionary and then do a zip join. The second is to make a set of all the keys, sort them, and then try to get the key from each dictionary. The third way is to iterate over all the keys in dict1. Then iterate over the keys in dict2 and only output data if it isn't present in dict1.

I think in this case getting a set of all the possible names and then iterating through them is the best way to go.

benchmark-service/benchmark/benchmark.py
jigold: That sounds good.

benchmark-service/benchmark/benchmark.py
```diff
@@ -45,18 +47,74 @@ def get_benchmarks(file_path):
...
+    def get_metric(data):
+        if metric == 'median':
+            return data['median']
+        assert metric == 'best'
+        return min(data['times'])
+
+    set_of_names = set()
+    for name1 in benchmarks1['data'].keys():
+        set_of_names.add(name1)
+    for name2 in benchmarks2['data'].keys():
```
jigold, line 72: This is fine, but you could also do it this way which might be clearer.

```
d1_keys = set(benchmarks1['data'].keys())
d2_keys = set(benchmarks2['data'].keys())
set_of_names = d1_keys.union(d2_keys)
```

benchmark-service/benchmark/benchmark.py
```diff
@@ -45,18 +47,74 @@ def get_benchmarks(file_path):
...
+        if r1 is None:
+            comps.append((name, None, None, fmt_time(r2)))
+        elif r2 is None:
+            comps.append((name, None, fmt_time(r1), None))
+        else:
+            r = r1 / r2
+            ratios.append(r)
+            comps.append((name, fmt_diff(r), fmt_time(r1), fmt_time(r2)))
+    final_comps['comps'] = comps
+    final_comps['harmonic_mean'] = fmt_diff(hmean(ratios))
```
jigold, line 111: Can we do something here where if the ratios is empty, then we set these values to None? Then in the Jinja template, we only show the value if it's not None.

benchmark-service/benchmark/templates/compare.html
```diff
@@ -0,0 +1,74 @@
...
+            unknown
+            {% endif %}
+        </th>
+    </tr>
+    </thead>
+    <tbody>
+    {% for name, ratio, r1, r2 in comparisons['comps'] %}
+    <tr>
+        <td>{{ name }}</td>
+        <td>{{ ratio }}%</td>
```
jigold, line 60: Can we template this to be empty if the ratio is None? We could also do the same for r1 and r2, but I'll leave that decision up to you.

benchmark-service/benchmark/templates/compare.html
```diff
@@ -0,0 +1,74 @@
...
+        <td>{{ r1 }}</td>
+        <td>{{ r2 }}</td>
+    </tr>
+    {% endfor %}
+    </tbody>
+</table>
+
+<p> Harmonic Mean: {{ comparisons['harmonic_mean'] }}% </p>
```
jigold, line 68: Let's not include the percentage if the value is None. See comment above.

PR #9285, opened 2020-08-17
[batch] add worker local SSD and PD SSD options to globals

cseed:
Expose in front end
Currently unused

### DIFF
--- a/batch/batch/driver/instance_pool.py
+++ b/batch/batch/driver/instance_pool.py
@@ -106,6 +106,8 @@ class InstancePool:
         self.worker_type = None
         self.worker_cores = None
         self.worker_disk_size_gb = None
+        self.worker_local_ssd_data_disk = None
+        self.worker_pd_ssd_data_disk_size_gb = None
         self.standing_worker_cores = None
         self.max_instances = None
         self.pool_size = None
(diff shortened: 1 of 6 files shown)


cseed:
In addition to the main change:
 - I made all the global configuration options editable now that we have flexible billing,
 - We now check the worker cores and standing worker cores against the worker type to make sure they are valid (there is no highmem/highcpu-1).

jigold:
The reason we didn't expose the other parameters was what if we had 16 core jobs waiting to be scheduled and then we changed the worker pool size to 4 cores.

jigold:
It looks fine other than the comment I made above.

cseed:
> The reason we didn't expose the other parameters was what if we had 16 core jobs waiting to be scheduled and then we changed the worker pool size to 4 cores.

Yep.  Let's call that admin operator error and let's not do that.  The other reason was we had hardcoded the billing computation in the code, but that's fixed now.  But it is hardcoded in the documentation, so we still shouldn't really be changing any of these settings (I see this mainly for the second instance at this point).

Separately, we should decouple the billing from the details of the implementation so we get a bit more flexibility on the backend in the main instance, as we've discussed.

I'm OK with this.

PR #9286, opened 2020-08-17
[batch] rip out rest of copy image infrastructure

jigold:


batch/batch/worker/worker.py
```diff
@@ -702,7 +676,6 @@ def __init__(self, batch_id, user, gsa_key, job_spec, format_version):
         containers = {}
 
         if input_files:
-            # input_volume_mounts.append(f'{LOCAL_SSD_MOUNT}:/host')
```
danking, line 705: should we just delete this?

jigold: Not sure what you mean here. I did delete it.

PR #9287, opened 2020-08-17
[query] Refactor define_function to go through backend.

tpoterba:


tpoterba:
Up to 522 passed against local backend.

PR #9288, opened 2020-08-17
[batch] Configurable worker disk type

jigold:
Stacked on #9285

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -61,7 +61,7 @@ INSTANCE_ID = os.environ['INSTANCE_ID']
 PROJECT = os.environ['PROJECT']
 WORKER_CONFIG = json.loads(base64.b64decode(os.environ['WORKER_CONFIG']).decode())
 MAX_IDLE_TIME_MSECS = int(os.environ['MAX_IDLE_TIME_MSECS'])
-LOCAL_SSD_MOUNT = os.environ['LOCAL_SSD_MOUNT']
+WORKER_DATA_DISK_MOUNT = os.environ['WORKER_DATA_DISK_MOUNT']
 BATCH_COPY_IMAGE = os.environ['BATCH_COPY_IMAGE']
 
 log.info(f'CORES {CORES}')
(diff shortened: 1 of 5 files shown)


jigold:
We decided to ignore this for now.

jigold:
This is passing the tests. I didn't double check the billing numbers in the database look correct. Can do so if you think it's prudent.

batch/batch/worker_config.py
```diff
@@ -98,6 +98,11 @@ def __init__(self, config):
         self.cores = instance['cores']
         self.preemptible = instance['preemptible']
 
+        data_disk = [d for d in self.disks if not d['boot']][0]
+        assert data_disk
+        self.local_ssd_data_disk = (data_disk['type'] == 'local-ssd')
+        self.data_disk_size_gb = data_disk['size']
```
cseed, line 104: Can we change this to asserting there are two disks and the first one is the boot disk, and just take the second one?  This code needs to change if there are >2 disks.

batch/batch/worker/worker.py
```diff
@@ -677,21 +677,15 @@ def __init__(self, batch_id, user, gsa_key, job_spec, format_version):
 
         req_cpu_in_mcpu = parse_cpu_in_mcpu(job_spec['resources']['cpu'])
         req_memory_in_bytes = parse_memory_in_bytes(job_spec['resources']['memory'])
-        if 'storage' in job_spec['resources']:
-            req_storage_in_bytes = parse_memory_in_bytes(job_spec['resources']['storage'])
```
cseed, line 681: Just want to confirm this was due to Ben's bug and new job specs no longer have a `storage` key?

jigold: The opposite. All job_specs should have a "resources" dictionary with "storage" in it now.

jigold:
I addressed the comment. The only thing to note is the default storage is set to 10 Gi I believe and I'm subtracting 25 Gi from the persistent disk size when figuring out how many cores the job requires. @konradjk

batch/batch/worker/worker.py
cseed: Got it, thanks!

PR #9289, opened 2020-08-17
[query] Remove PBaseStructValue.apply

chrisvittal:
It's wildly unsafe. It's better to scope the unsafety in
`IEmitCode.handle` for `loadField`.

Also add `PNDArrayValue.shapes` to handle the previous use case of
`PBaseStructValue.apply`.


chrisvittal:
PR for CI for now.

johnc1231:
Thanks for fixing this Chris.

PR #9290, opened 2020-08-17
[batch] pull batch-copy before building for caching

cseed:


jigold:
I'll approve this, although I was going to rip all of this out in #9286.

PR #9292, opened 2020-08-17
[query] Use size-based LZ4 compression codec for wireCodec (shuffles)

tpoterba:
I'm having a lot of of trouble demonstrating that this is faster using the benchmark system. I can see a marked improvement in shuffle benchmarks using my laptop:
```
$ hb compare /tmp/lz4-before.json /tmp/lz4-after.json
shuffle_order_by_10m_int      70.4%     47.193     33.245
```

I also don't see any slowdown for the larger shuffles (which do get slower if we use one of the uncompressed codecs).

Running benchmarks on batch have produced results where the median varies from 1% slower to 6% faster.

### DIFF
--- a/hail/src/main/scala/is/hail/io/BufferSpecs.scala
+++ b/hail/src/main/scala/is/hail/io/BufferSpecs.scala
@@ -155,7 +155,7 @@ final case class LZ4FastBlockBufferSpec(blockSize: Int, child: BlockBufferSpec)
 
 final case class LZ4SizeBasedBlockBufferSpec(hc: Boolean, blockSize: Int, minCompressionSize: Int, child: BlockBufferSpec)
     extends BlockBufferSpec {
-  val lz4: LZ4 = if (hc) LZ4.hc else LZ4.fast
+  def lz4: LZ4 = if (hc) LZ4.hc else LZ4.fast
   def stagedlz4: Code[LZ4] = Code.invokeScalaObject0[LZ4](LZ4.getClass, "fast")
   def typeName = "LZ4SizeBasedBlockBufferSpec"
 
--- a/hail/src/main/scala/is/hail/io/BufferSpecs.scala
+++ b/hail/src/main/scala/is/hail/io/BufferSpecs.scala
@@ -24,7 +24,7 @@ object BufferSpec {
 
   val wireSpec: BufferSpec = LEB128BufferSpec(
     BlockingBufferSpec(32 * 1024,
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/io/BufferSpecs.scala
```diff
@@ -151,6 +153,23 @@ final case class LZ4FastBlockBufferSpec(blockSize: Int, child: BlockBufferSpec)
   def typeName = "LZ4FastBlockBufferSpec"
 }
 
+final case class LZ4SizeBasedFastBlockBufferSpec(blockSize: Int, minCompressionSize: Int, child: BlockBufferSpec)
```
cseed, line 156: Instead of making this LZ4SizeBasedFast..., can you put the hc vs fast in a parameter to the BufferSpec?  That will make it more flexible.

hail/src/main/scala/is/hail/io/BufferSpecs.scala
```diff
@@ -151,6 +153,23 @@ final case class LZ4FastBlockBufferSpec(blockSize: Int, child: BlockBufferSpec)
   def typeName = "LZ4FastBlockBufferSpec"
 }
 
+final case class LZ4SizeBasedFastBlockBufferSpec(hc: Boolean = false, blockSize: Int, minCompressionSize: Int, child: BlockBufferSpec)
```
tpoterba, line 156: is this what you meant?

tpoterba:
> Using the automatic Json deserializer kind of makes us write the wrong code here, because we really want one LZ4CodecSpec which takes the appropriate parameters, and then deserializer can construct it with the proper arguments based on the specific case being derserialized (a legacy format, or the current one with parameters).

Yeah, I agree. I think it would be a couple days of work to write ser/deser stuff that avoids the json4s case class extraction, but I'm not sure how high-priority that is right now since this stuff is somewhat stable.

hail/src/main/scala/is/hail/io/BufferSpecs.scala
```diff
@@ -151,6 +153,23 @@ final case class LZ4FastBlockBufferSpec(blockSize: Int, child: BlockBufferSpec)
   def typeName = "LZ4FastBlockBufferSpec"
 }
 
+final case class LZ4SizeBasedBlockBufferSpec(hc: Boolean, blockSize: Int, minCompressionSize: Int, child: BlockBufferSpec)
```
cseed, line 156: This is fine, but something like `compressorType: String` seems more forward extensible.

PR #9294, opened 2020-08-18
[hailctl] Python3 kernel also gets the spark monitor

danking:


danking:
I verified this by building locally and starting a cluster and checking both kernels.

PR #9295, opened 2020-08-18
[query] Add close() to BGZipInputStream

tpoterba:


chrisvittal:
This doesn't look like it can hurt, but why is this necessary? `SplitCompressionInputStream` extends `CompressionInputStream` which has a `close` method already.

tpoterba:
ah, yeah. This isn't the fix, then.

We must be messing this up elsewhere -- we definitely have a file handle leak.

PR #9296, opened 2020-08-18
[Benchmark] Enable benchmark to read from any benchmarks bucket

Dania-Abuhijleh:


benchmark-service/deployment.yaml
```diff
@@ -66,3 +66,6 @@ spec:
         - name: ssl-config-benchmark
           secret:
             secretName: ssl-config-benchmark
+        - name: benchmark-gsa-key
```
jigold, line 69: I think you need to mount the secret as well.

benchmark-service/benchmark/utils.py
```diff
@@ -27,7 +27,8 @@ def enumerate_list_index(list_of_lists):
 
 class ReadGoogleStorage:
     def __init__(self):
-        self.storage_client = storage.Client()
+        credentials = google.oauth2.service_account.Credentials.from_service_account_file('/benchmark-gsa-key/key.json')
```
jigold, line 30: Let's parameterize this. The class should take the credentials to use which can be None. So a good default is to do this:

```
def __init__(self, credentials=None):
    self.storage_client = storage.Client(credentials=credentials)
```

Then I think you'll need to make the ReadGoogleStorage class construction in your benchmark.py code a global variable and you'll read the credentials where you define the ReadGoogleStorage class.

jigold:
Can we close this in favor of #9359?

PR #9297, opened 2020-08-18
[batch]  fix LocalBackend.run() succeeding when intermediate command fails

akotlar:
CHANGELOG: fix LocalBackend.run() succeeding when intermediate command fails

Stacked on #9219 as that PR is essentially approved, and to avoid a merge conflict. The commit in this PR is https://github.com/hail-is/hail/pull/9297/commits/cbc3bbe7f14c01d44c89995a03375d983fc14f4f

Caused by associativity of the ternary conditional ('set -e' + 'x' is the operand `a` in `a if cond else b`)

Easy reproduction case on main:

```python
def test_single_job_with_mixed_shells(self):
        b = self.batch()
        j = b.new_job()
        j.command(f'echoddd "hello"')
        j2 = b.new_job()
        j2.command(f'echo "world"')

        self.assertRaises(Exception, b.run)
```


akotlar:
Rolled Dania.

tpoterba:
doesn't the ServiceBackend exit(0) even when the batch fails? That's my vague recollection of running benchmarks

akotlar:
Yes, it will exit(0), but upon any job failure `state` will be failure, and the STDOUT message will be "batch {bc_batch.id} complete: failure" instead of "Batch completed successfully" as in the LocalBackend case currently. We rely on exception handling to catch job errors for LocalBackend.run, and the current bug occurs because of an improperly formed ternary expression (we can tell this because `verbose` should not enable `-e`).

akotlar:
ready to review

PR #9298, opened 2020-08-18
[query] Revert TabixReader Improvements from 3302850eb9d93c30793dc231a2c778969c1422ba

chrisvittal:
…a2c778969c1422ba

The tabix line iterator works by building a list of virtual offsets from
the index that correspond to a requested interval. We recently
discovered an issue that would lead to a runtime exception if the
following conditions held:

* An offset pair ended exactly on a block boundary.
* The block boundary was exactly the start of a line.

In a blocked file with virtual offsets, there are two ways to point to
the start of every block, (previous block start offset, previous block size) or
(current block start offset, 0).

Because of the way curOff was calculated in TabixLineIterator, this would
lead to a situation where curOff was the (previous block start offset, previous block size)
value, causing the jump to next chunk comparision `!TbiOrd.less64(curOff, offsets(i)._2)`
to fail when it should succeed, causing an extra line to be read, which
then makes the assertion check in the next iteration.

Here we revert the improvements made earlier in
3302850eb9d93c30793dc231a2c778969c1422ba which is where this logic error
was introduced.

### DIFF
--- a/hail/src/main/scala/is/hail/io/tabix/TabixReader.scala
+++ b/hail/src/main/scala/is/hail/io/tabix/TabixReader.scala
@@ -1,14 +1,15 @@
 package is.hail.io.tabix
 
-import java.io.InputStream
-import java.nio.charset.StandardCharsets
-
-import htsjdk.samtools.util.FileExtensions
-import htsjdk.tribble.util.ParsingUtils
+import is.hail.backend.BroadcastValue
 import is.hail.io.compress.BGzipInputStream
 import is.hail.io.fs.FS
 import is.hail.utils._
-import is.hail.backend.BroadcastValue
+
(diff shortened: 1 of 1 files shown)


chrisvittal:
This is the easiest fix.

chrisvittal:
Confirmed fixed on our data.

tpoterba:
I don't think this is the right solution. We should fix the control flow instead -- a ~40% performance degradation is pretty horrific, even as a stop-gap.

chrisvittal:
It's not 40%, I measured later, it's more like 10%. I'm not sure what you're calling a control flow bug. If `TabixLineIterator.getVirtualOffset` returned the correct value, we wouldn't have any issue. The code's just plain wrong. The control is fine.

PR #9300, opened 2020-08-18
[site] Some homepage tweaks from Jon B

tpoterba:


site/www/index.xslt
```diff
@@ -80,17 +80,19 @@ show(p)
...
-                            While convenient, they don't capture the structure of genetic data, which has row (variant) and
-                            column
-                            (genotype) groups.
-                            To remedy this, Hail introduces a distributed, dataframe-like structure called
-                            <a href="/docs/0.2/overview/matrix_table.html?highlight=matrix%20table" target="_blank">MatrixTable</a>.
+                        <p>Modern data science is driven by numeric matrices (see <a href="https://numpy.org/">Numpy</a>) and
+                           tables (see <a href="https://www.r-project.org/about.html">R</a> and <a href="https://pandas.pydata.org">Pandas</a>).
+                           While sufficient for many tasks, none of these tools adequately capture
+                           the structure of genetic data. Genetic data combines multiple axes (variants and samples) like
+                           matrices, and structured entries (genotypes) like tables or dataframes.
```
jigold, line 92: I think this comma shouldn't be here.

site/www/index.xslt
```diff
@@ -170,7 +172,7 @@ show(p)
                 function reveal(e) {
                     cachedCode.style.display = "block";
                     cachedImg.style.display = "none";
-                    cachedText.textContent = "hide code";
+                    cachedText.textContent = "show plot";
```
jigold, line 175: I can't figure out why this change was made or where it's used.

johnc1231: Screen Shot 2020-08-18 at 5 00 36 PM
It's the text under the image of the code / plot.

site/www/index.xslt
tpoterba: I struggled hard with this comma. I know it's not proper grammar, but also it looked confusingly verbose without. Will remove.

site/www/index.xslt
jigold: Maybe I'm not looking in the right place. Can you send me the exact link to this? I don't see it on the homepage or on the input unification page clicking on Learn More.

jigold:
> Should there be a place that now says "show code"?

tpoterba:
ah, we answered that in the other comment thread. It already says "show code".

PR #9303, opened 2020-08-18
[query] Add `hl.nd.svd` for doing local Singular Value Decompositions

johnc1231:
This adds an `svd` method to `hl.nd`, allowing us to take singular value decomposition of local ndarrays. This is a generally useful operation, but my primary reason for adding it is to avoid the need to call numpy's SVD as a substep of Annamira's new randomized PCA method. The interface is designed to match numpy exactly. 

Note that this method uses the LAPACK method `dgesdd`, which is newer, faster version of `dgesvd`. As far as I can tell, there's never a reason to use `dgesvd`.

### DIFF
--- a/hail/python/hail/docs/nd/index.rst
+++ b/hail/python/hail/docs/nd/index.rst
@@ -11,7 +11,8 @@ nd
 
 Notes
 _____
-This is a recently added, experimental module. We would love to hear what use cases you have for this as we expand this functionality
+This is a recently added, experimental module. We would love to hear what use cases you have for this as we expand this functionality.
+As much as possible, we try to mimic the numpy array interface.
 
 .. autosummary::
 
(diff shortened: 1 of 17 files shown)


patrick-schultz:
> As far as I can tell, there's never a reason to use dgesvd.

It appears dgesvd uses less memory and is more accurate, particularly on very small singular values, but is potentially several times slower. I agree dgesdd is the right default, but it's possible (though I'm guessing unlikely) we'll eventually have reason to expose dgesvd as an option.

Here's a thread of Julia devs discussing which to use: https://groups.google.com/u/1/g/julia-dev/c/mmgO65i6-fA?pli=1

johnc1231:
That's fair. I didn't consider length of the work array. It does sound like `dgesdd` calls `dgesvd` for subproblems, would be nice if it just called out to that on any matrix of a certain small size. But yeah, sounds like something we maybe include in the future as an improvement after some experimenting if the user is doing small svds. For now though, `dgesdd` seems like the right choice.

johnc1231:
For what it's worth, numpy doesn't use `dgesvd` in any case.

PR #9307, opened 2020-08-19
[gear] fix db connection leak in transaction due to cancellation

cseed:
The old code in `Transaction` to exit the transaction and release the connection back to the pool looked like this:

```
    async def _aexit(self, exc_type, exc_val, exc_tb):
        try:
            if self.conn is not None:
                try:
                    if exc_type:
                        await self.conn.rollback()
                    else:
                        await self.conn.commit()
                finally:
                    self.conn = None
        finally:
            if self.conn_context_manager is not None:
                try:
                    await aexit(self.conn_context_manager, exc_type, exc_val, exc_tb)
                finally:
                    self.conn_context_manager = None
```

The problem was if the current coroutine was cancelled in the call to `aexit(self.conn_context_manager, ...)`, which ultimately calls aiomysql `Pool.release`, the release never happens.  This was happening when database calls in `@only_active_instances` were getting cancelled when the client timed out and terminated the request.  Roughly, the solution is to shield exiting the connection, and return the connection asynchronously in a background task (using `ensure_future`).

FYI @danking @jigold This is a subtle bug/pattern for managing resources that we should all be aware of.

### DIFF
--- a/gear/gear/database.py
+++ b/gear/gear/database.py
@@ -1,5 +1,6 @@
 from typing import Optional
 import os
+import asyncio
 import pymysql
 import aiomysql
 import logging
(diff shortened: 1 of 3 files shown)


cseed:
bump

cseed:
bump  bump

jigold:
I'm happy with this PR. @catoverdrive did you want to look at it? If not, please assign it to me or Dan.

PR #9308, opened 2020-08-19
[query] Add enumerate, deprecate zip_with_index

johnc1231:
`zip_with_index` seems based on Scala's `zipWithIndex`. Here I deprecate that in favor of using the more pythonic `enumerate`. To match the signature of the python function, I add a `start` argument and make our custom `index_first` argument keyword only.

### DIFF
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -3410,7 +3410,7 @@ def enumerate(a, start=0, *, index_first=True):
     ----------
     a : :class:`.ArrayExpression`
     start : :class:`.Int32Expression`
-        The number the first element of the array is labeled with.
+        The index value from which the counter is started, 0 by default.
     index_first: :obj:`bool`
         If ``True``, the index is the first value of the element tuples. If
         ``False``, the index is the second value.
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -513,7 +513,7 @@ def rbind(*exprs, _ctx=None):
 
     *args, f = exprs
     args = [expr_any.check(arg, 'rbind', f'argument {index}')
-            for index, arg in enumerate(args)]
+            for index, arg in builtins.enumerate(args)]
 
     return hl.bind(f, *args, _ctx=_ctx)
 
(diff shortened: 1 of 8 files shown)


hail/python/hail/expr/functions.py
```diff
@@ -3384,9 +3384,45 @@ def zip(*arrays, fill_missing: bool = False) -> ArrayExpression:
                           aggregations)
 
 
+@typecheck(a=expr_array(), start=expr_int32, index_first=bool)
```
akotlar, line 3387: does this need to be int32?

johnc1231: Yes, because currently our arrays store their length as an int32, so that's what `hl.range` expects. All the ndarray stuff stores the shape as an int64 because I imagine we will eventually want to blow past 2^31 as an array length limit.

hail/python/hail/expr/functions.py
```diff
@@ -3384,9 +3384,45 @@ def zip(*arrays, fill_missing: bool = False) -> ArrayExpression:
...
+    [(3, 'A'), (4, 'B'), (5, 'C')]
+
+    >>> hl.eval(hl.enumerate(['A', 'B', 'C'], index_first=False))
+    [('A', 0), ('B', 1), ('C', 2)]
+
+
+    Parameters
+    ----------
+    a : :class:`.ArrayExpression`
+    start : :class:`.Int32Expression`
```
akotlar, line 3407: I would say something like "the starting 'index'", index can be first or second.

johnc1231: So the reason I worded it the way I did was because I was afraid to give the implication that "index" meant "indexing into `a` (i.e., specifying start = 2 would mean to start at position 2 of `a`)".

akotlar: Let's make our language consistent with Python's `enumerate` method, or if you feel strongly that this isn't the right direction, make our language internally consistent. We use "index" throughout, and "label" does not indicate that the number included in the tuple is ordinal. 

Here is how Python's enumerate is described elsewhere:

"The __next__() method of the iterator returned by enumerate() returns a tuple containing a count (from start which defaults to 0) and the values obtained from iterating over iterable."
  - https://docs.python.org/3/library/functions.html#enumerate

"And there is more! enumerate also accepts an optional argument that allows us to specify the starting index of the counter."
  - https://book.pythontips.com/en/latest/enumerate.html 

"Start: the index value from which the counter is 
              to be started, by default it is 0 "
  - https://www.geeksforgeeks.org/enumerate-in-python/


A few of our uses of "index":
"Returns an array of (index, element) tuples."
"Array of (index, element) or (element, index) tuples."
"The number the first element of the array is labeled with.
     index_first: :obj:`bool` ...  If ``True``, the index is the first value of the element tuples"

johnc1231: You've convinced me. I'll go with language that includes "index", that seems to be the general understanding. I like "The index value from which the counter is started", I think that's clear.

PR #9309, opened 2020-08-19
[query] Change decompression order for BGZipInputStream

chrisvittal:
Instead of decompressing a block as soon as we reach the end of the
previous block, or during construction. We now wait, and if it is
necessary for that read, decompress the next block.

### DIFF
--- a/hail/src/main/java/is/hail/io/compress/BGzipInputStream.java
+++ b/hail/src/main/java/is/hail/io/compress/BGzipInputStream.java
@@ -179,7 +179,12 @@ public class BGzipInputStream extends SplitCompressionInputStream {
     }
 
     public long getVirtualOffset() {
-        return BlockCompressedFilePointerUtil.makeFilePointer(inputBufferInPos, outputBufferPos);
+        if (outputBufferPos == outputBufferSize) {
+            int bsize = bgzipHeader != null ? bgzipHeader.bsize : 0;
+            return BlockCompressedFilePointerUtil.makeFilePointer(inputBufferInPos + bsize, 0);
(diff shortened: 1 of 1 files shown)


chrisvittal:
Let's see what CI says.

hail/src/main/java/is/hail/io/compress/BGzipInputStream.java
```diff
@@ -184,21 +183,23 @@ public long getVirtualOffset() {
     }
 
     public int readBlock(byte[] b) throws IOException {
+        assert(outputBufferPos == 0);
+        decompressNextBlock();
+        outputBufferPos = outputBufferSize;
```
chrisvittal, line 188: this line is unnecessary, and probably redundant, will fix

hail/src/main/java/is/hail/io/compress/BGzipInputStream.java
```diff
@@ -184,21 +183,23 @@ public long getVirtualOffset() {
...
 
-        outputBufferPos = outputBufferSize;
-        decompressNextBlock();
-
         return blockSize;
     }
 
     public int read(byte[] b, int off, int len) throws IOException {
+        if (outputBufferPos == outputBufferSize)
+            decompressNextBlock();
```
chrisvittal, line 202: This should be below the length check.

chrisvittal:
Spun the wheel for this. This is probably needed for #9304, which we need for gnomAD.

chrisvittal:
This and #9304 have a mutual dependency (ugh). Close this in favor of the other.

PR #9312, opened 2020-08-20
[query][smm 13] Memory management for StreamScan and RunAggScan

patrick-schultz:
Stacked on #9240.

PR #9313, opened 2020-08-20
[query][smm 14] Memory management for TableMapPartitions

patrick-schultz:
~~Stacked on #9312.~~

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Compile.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Compile.scala
@@ -7,6 +7,7 @@ import is.hail.asm4s._
 import is.hail.expr.ir.Stream.Source
 import is.hail.expr.ir.agg.AggStateSig
 import is.hail.expr.ir.lowering.LoweringPipeline
+import is.hail.rvd.RVDContext
 import is.hail.types.physical.{PStream, PStruct, PType}
 import is.hail.types.virtual.Type
 import is.hail.utils._
@@ -144,7 +145,10 @@ object CompileIterator {
     def loadAddress(): Long
   }
 
-  private trait TMPStepFunction extends AsmFunction3RegionLongIteratorJLongBoolean with StepFunctionBase
+  private trait TMPStepFunction extends StepFunctionBase {
+    def apply(o: Object, a: Long, b: StreamArgType): Boolean
+    def setRegions(outerRegion: Region, eltRegion: Region): Unit
+  }
 
   private abstract class LongIteratorWrapper extends Iterator[java.lang.Long] {
     def step(): Boolean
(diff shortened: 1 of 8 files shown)


hail/src/main/scala/is/hail/expr/ir/Compile.scala
```diff
@@ -177,18 +181,24 @@ object CompileIterator {
...
+    val outerRegionField = fb.genFieldThisRef[Region]("outerRegion")
+    val eltRegionField = fb.genFieldThisRef[Region]("eltRegion")
+    val setF = fb.newEmitMethod("setRegions", FastIndexedSeq(CodeParamType(typeInfo[Region]), CodeParamType(typeInfo[Region])), CodeParamType(typeInfo[Unit]))
+    setF.emit(Code(outerRegionField := setF.getCodeParam[Region](1), eltRegionField := setF.getCodeParam[Region](2)))
+
     val stepF = fb.apply_method
     val stepFECB = stepF.ecb
 
-    val region = StagedRegion(EmitRegion.default(stepF).region, allowSubregions = false)
+    val eltRegion = StagedRegion(eltRegionField, allowSubregions = false)
```
tpoterba, line 192: we'll flip these to `true` when we solve the mixed memory management design, right?

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -1637,18 +1637,15 @@ class Emit[C](
               val updateMap = Map(fields: _*)
               val srvb = new StagedRegionValueBuilder(mb, x.pType, region.code)
 
-              val addFields = { (newMB: EmitMethodBuilder[C], t: PType, v: EmitCode) =>
-                Code(
-                  v.setup,
-                  v.m.mux(srvb.setMissing(), srvb.addIRIntermediate(t)(v.v)),
-                  srvb.advance())
-              }
+              def addFields(t: PType, v: EmitCode): Code[Unit] = Code(
```
tpoterba, line 1640: why these changes? Did something break?

hail/src/main/scala/is/hail/expr/ir/Compile.scala
patrick-schultz: I don't know yet what the final design will look like for setting and passing down the allocation strategy, but as things are now I think we'd want `outerRegion` to allow subregions, but not `eltRegion`.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
patrick-schultz: The `emitSelf.emit` call is bad, because it will use the Region argument of `mb` rather than the region passed to emit. This looked to me like some cruft left from the method splitter change, so I cleaned it up while I was fixing the bug. I don't remember anything breaking, I think I found this by looking at where the `emit` overload that uses the region argument is used.

tpoterba: ahhhh, got it.

PR #9316, opened 2020-08-20
[batch] weight zones by quota capacity

cseed:
Instead of using a fixed weighting, query the GCP /regions endpoint which includes the quota limit and usage.  Use this to weight zones for new instance requests.  Added a ComputeClient.list endpoint that should work with any GCE list endpoint that is paged.

I'm working on improving this logic further by tracking how often instance create requests fail and then weighting zones by `capacity * p(create instance will succeed)`.

### DIFF
--- a/batch/batch/driver/instance_pool.py
+++ b/batch/batch/driver/instance_pool.py
@@ -24,70 +24,6 @@ BATCH_WORKER_IMAGE = os.environ['HAIL_BATCH_WORKER_IMAGE']
 
 log.info(f'BATCH_WORKER_IMAGE {BATCH_WORKER_IMAGE}')
 
-REGIONS = [
-    {
-        'name': 'us-central1',
-        'zones': ('a', 'b', 'c', 'f'),
-        'preemptible_cpus': 100000,
-        'local_ssd_total_storage_gb': 100000
-    },
-    {
-        'name': 'us-east1',
-        'zones': ('b', 'c', 'd'),
-        'preemptible_cpus': 25000,
-        'local_ssd_total_storage_gb': 300000
-    },
-    {
-        'name': 'us-east4',
-        'zones': ('a', 'b', 'c'),
-        'preemptible_cpus': 25000,
-        'local_ssd_total_storage_gb': 100000
-    },
-    {
-        'name': 'us-west1',
-        'zones': ('a', 'b', 'c'),
-        'preemptible_cpus': 25000,
-        'local_ssd_total_storage_gb': 100000
-    },
-    {
-        'name': 'us-west2',
-        'zones': ('a', 'b', 'c'),
-        'preemptible_cpus': 25000,
-        'local_ssd_total_storage_gb': 50000
-    },
-    # no quota in us-west3
-    # could consider northamerica-northeast1 Montreal
-    {
-        'name': 'us-west4',
-        'zones': ('a', 'b', 'c'),
-        'preemptible_cpus': 10000,
-        'local_ssd_total_storage_gb': 20000
-    }
-]
-
-USCENTRAL1_CORES = None
-
-ZONES = []
-ZONE_WEIGHTS = []
-for region in REGIONS:
-    cores = region['preemptible_cpus']
-    ssd_in_cores = int(region['local_ssd_total_storage_gb'] / 375 * 16)
-    if ssd_in_cores < cores:
-        cores = ssd_in_cores
-
-    name = region['name']
-    if name == 'us-central1':
-        USCENTRAL1_CORES = cores
-
-    r_zones = region['zones']
-    weight = cores / len(r_zones)
-    for z in r_zones:
-        ZONES.append(f'{name}-{z}')
-        ZONE_WEIGHTS.append(weight)
-
-assert len(ZONES) == len(ZONE_WEIGHTS)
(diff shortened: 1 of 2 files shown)


hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -1,6 +1,45 @@
...
+        self._page = None
+        self._index = None
+
+    def __aiter__(self) -> 'PagedIterator':
+        return self
+
+    async def __anext__(self):
+        if self._page is None:
+            assert 'pageToken' not in self._request_params
+            print('first page')
```
jigold, line 23: delete.

hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -1,6 +1,45 @@
...
+        while True:
+            if 'items' in self._page and self._index < len(self._page['items']):
+                i = self._index
+                self._index += 1
+                return self._page['items'][i]
+
+            next_page_token = self._page.get('nextPageToken')
+            if next_page_token is not None:
+                self._request_params['pageToken'] = next_page_token
+                print('next page')
```
jigold, line 36: delete

batch/batch/driver/instance_pool.py
```diff
@@ -253,11 +246,11 @@ async def create_instance(self, cores=None, max_idle_time_msecs=None):
             if machine_name not in self.name_instance:
                 break
 
-        if self.live_total_cores_mcpu // 1000 < int(USCENTRAL1_CORES * 0.9):
+        if self.live_total_cores_mcpu // 1000 < 4_000:
             zones = ['us-central1-a', 'us-central1-b', 'us-central1-c', 'us-central1-f']
             zone = random.choice(zones)
         else:
-            zone = random.choices(ZONES, ZONE_WEIGHTS)[0]
+            zone = random.choices(self.zones, self.zone_weights)[0]
```
jigold, line 253: Did you test this code path works since you're only using it with a big cluster?

batch/batch/driver/instance_pool.py
```diff
@@ -129,6 +70,54 @@ def __init__(self, app, machine_name_prefix):
...
+        new_zones = []
+        new_zone_weights = []
+
+        async for r in await self.compute_client.list('/regions'):
+            name = r['name']
+            if name not in northamerica_regions:
+                continue
+
+            quota_remaining = {
+                q['metric']: q['limit'] - q['usage']
```
jigold, line 94: oh cool. That's nice they give you both the quota and the usage.

batch/batch/driver/instance_pool.py
```diff
@@ -129,6 +70,54 @@ def __init__(self, app, machine_name_prefix):
...
+                for q in r['quotas']
+            }
+
+            remaining = quota_remaining['PREEMPTIBLE_CPUS'] // self.worker_cores
+            if self.worker_local_ssd_data_disk:
+                remaining = min(remaining, quota_remaining['LOCAL_SSD_TOTAL_GB'] // 350)
```
jigold, line 100: This should be 375 instead of 350. The 350 number comes from the amount of space we reserve for jobs (25 GB goes for Docker images etc.)

batch/batch/driver/instance_pool.py
```diff
@@ -129,6 +70,54 @@ def __init__(self, app, machine_name_prefix):
...
+            else:
+                remaining = min(remaining, quota_remaining['SSD_TOTAL_GB'] // self.worker_pd_ssd_data_disk_size_gb)
+
+            weight = max(remaining // len(r['zones']), 1)
```
jigold, line 104: This is basically either 0 or 1? It looks like you don't weight by the amount remaining.

batch/batch/driver/instance_pool.py
```diff
@@ -110,6 +48,9 @@ def __init__(self, app, machine_name_prefix):
         self.max_instances = None
         self.pool_size = None
 
+        self.zones = ['us-central1-a', 'us-central1-b', 'us-central1-c', 'us-central1-f']
```
jigold, line 51: This is the defaults until the zones get updated in the first iteration of the update_zones_loop?

batch/batch/driver/instance_pool.py
cseed: Yeah.  I started with another API, the Service Usage API (https://cloud.google.com/service-usage/docs/manage-quota), but it only gives you limits.  Finding this was a major simplification.

batch/batch/driver/instance_pool.py
cseed: Yeah, exactly.  Hitting the /regions API is very slow (seconds), so I didn't want to do it synchronously in the startup code.  I'll add a comment to that effect.

batch/batch/driver/instance_pool.py
cseed: I lived deployed it to the large cluster, it is working admirably.

batch/batch/driver/instance_pool.py
cseed: That's a `max`.  I always make this mistake, too.  So if remaining = 300 and there are three zones, the weight is max(300 // 3, 1) = 100 per zone.

PR #9317, opened 2020-08-21
[batch] weight zones by capacity * p(success)

cseed:
- track last 10 instance.create operations per zone
 - weight probability by the success rate
 - fixed bug where mark = None wouldn't return any messages


batch/batch/driver/instance_pool.py
```diff
@@ -97,13 +156,13 @@ async def update_zones(self):
                 for q in r['quotas']
             }
 
-            remaining = quota_remaining['PREEMPTIBLE_CPUS'] // self.worker_cores
+            remaining = quota_remaining['PREEMPTIBLE_CPUS'] / self.worker_cores
```
cseed, line 159: I changed these because the `random.choices` can handle floats.  Seems more accurate.

PR #9319, opened 2020-08-21
Fix pk-sorted and unsorted LoweredTableCoercers

tpoterba:
Up to 569 tests passing.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/TableIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/TableIR.scala
@@ -173,51 +173,51 @@ object LoweredTableReader {
       Let("n", ApplyAggOp(FastIndexedSeq(), FastIndexedSeq(), count),
         AggLet("key", GetField(Ref("x", xType), "key"),
           MakeStruct(FastIndexedSeq(
-          "n" -> Ref("n", TInt64),
-          "minkey" ->
-            ArrayRef(
-              ApplyAggOp(
-                FastIndexedSeq(I32(1)),
-                FastIndexedSeq(Ref("key", keyType), Ref("key", keyType)),
-                minkey),
-              I32(0)),
-          "maxkey" ->
-            ArrayRef(
-              ApplyAggOp(
-                FastIndexedSeq(I32(1)),
-                FastIndexedSeq(Ref("key", keyType), Ref("key", keyType)),
-                maxkey),
-              I32(0)),
-          "ksorted" ->
-            ApplyComparisonOp(EQ(TInt64),
-              ApplyAggOp(
-                FastIndexedSeq(),
-                FastIndexedSeq(
-                  invoke("toInt64", TInt64,
-                    invoke("lor", TBoolean,
-                      IsNA(GetField(Ref("x", xType), "prevkey")),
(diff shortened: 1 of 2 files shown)


tpoterba:
reassigned to Patrick since John is OOO.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -350,22 +351,45 @@ object LoweredTableReader {
...
             body)
 
-          throw new LowererUnsupportedOperation("pksorted VCF import requires repartition")
+          pkPartitioned
+            .strictify()
+            .mapPartition(None) { part =>
+              flatMapIR(StreamGroupByKey(part, pkType.fieldNames)) { inner =>
+                ToStream(sortIR(inner) { case (l, r) => ApplyComparisonOp(LT(l.typ), l, r) })
+              }
+            }
```
patrick-schultz, line 369: This probably works, but it does break the invariant that a TableStage is sorted by its key. I think it would be better to move this to a method on TableStage, `localSort` or something, that takes a stage keyed by `pkType` and returns a stage keyed by `keyType`, so the invariant is never violated.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -332,14 +332,15 @@ object LoweredTableReader {
         }
       }
     } else if (pksorted) {
```
patrick-schultz, line 334: I know you didn't change it in this PR, but I'm confused why `pksorted` requires the max pk of one partition to *equal* the min pk of the next. Shouldn't that be LTEQ? Similarly for `ksorted`.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
tpoterba: it's easy to move the logic around so we preserve the invariant.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
tpoterba: good catch. This actually explains something I saw in the tests but didn't dig into -- some VCFs that ostensibly are sorted were throwing errors about unsorted not being supported!

tpoterba: I'll PR that separately.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -352,20 +353,43 @@ object LoweredTableReader {
 
           val partitioner = new RVDPartitioner(pkType,
             sortedPartData.map { partData =>
-              Interval(selectPK(partData.getAs[Row](0)), selectPK(partData.getAs[Row](1)), includesStart = true, includesEnd = true)
-            }, key.length)
+              Interval(selectPK(partData.getAs[Row](1)), selectPK(partData.getAs[Row](2)), includesStart = true, includesEnd = true)
+            }, pkType.size)
```
patrick-schultz, line 357: Doesn't really matter, but this is the default value of `allowedOverlap`; no need to write it explicitly, since it really means "do no extra checks".

PR #9320, opened 2020-08-21
[query] Framework for making compiler passes stack safe

patrick-schultz:
This PR adds `utils/StackSafe.scala`, which contains generic tools for writing stack safe code. To illustrate its use, I've converted `NormalizeNames` and `RewriteBottomUp` to be stack safe.

This approach optimizes for the minimal possible change to existing code to make it stack safe. I originally expected this to have mediocre performance, and designed this to have optimization opportunities--requiring more substantial rewrites--where we found it was necessary.

In a follow up PR, I converted the IR parser to be stack safe. In benchmarking that, I'm not able to see any performance penalty (if anything, the stack safe version looks slightly faster, which is probably just noise). So it's possible this will perform well enough as is, but we can keep an eye on it as we convert more passes.

The basic idea is to rewrite functions that can be called recursively (directly or indirectly through a path of mutually recursive functions) from `f: (...) => A` to `f: (...) => StackFrame[A]`. Where the former evaluates, executing all recursive calls, and then returns the `A` result, the later returns a description of the work to be done before making any recursive calls. The method `StackFrame[A].run(): A` executes that description in a non-recursive loop.

`StackFrame` is a monad, implementing `map` and `flatMap`, which allows the `for` syntactic sugar to be used. When a method makes several recursive calls, this can be significantly more readable.

The public api is small. There are the free functions
```scala
def done[A](result: A): StackFrame[A]
def call[A](body: => StackFrame[A]): StackFrame[A]
```
and the methods
```scala
abstract class StackFrame[A] {
  def flatMap[B](f: A => StackFrame[B]): StackFrame[B]
  def map[B](f: A => B): StackFrame[B] = flatMap(a => Done(f(a)))
  def run(): A
}
```
`done` is basically the return statement. `call` is very important: it wraps a recursive call in a thunk, so that returning a `StackFrame` doesn't require descending all the way to the leaves of the recursion. The basic rule is that any cycle of mutually recursive function calls must have at least one `call` on the cycle. Trying to keep it to just one `call` per cycle minimizes the number of closures that must be allocated.

As an example, `NormalizeNames.normalizeIR` now returns `StackFrame[BaseIR]`, and `def apply(ir: BaseIR): BaseIR` calls `normalizeIR(ir, ...).run()` to actually run the traversal. The case for `Let` in `normalizeIR` is rewritten from
```scala
case Let(name, value, body) =>
  val newName = gen()
  Let(newName, normalize(value), normalize(body, env.copy(eval = env.eval.bind(name, newName))))
```
to
```scala
case Let(name, value, body) =>
  val newName = gen()
  for {
    newValue <- normalize(value)
    newBody <- normalize(body, env.copy(eval = env.eval.bind(name, newName)))
  } yield Let(newName, newValue, newBody)
```
or without the sugar
```scala
case Let(name, value, body) =>
  val newName = gen()
  normalize(value).flatMap { newValue =>
    normalize(body, env.copy(eval = env.eval.bind(name, newName))).map { newBody =>
      Let(newName, newValue, newBody)
    }
  }
```
All recursive calls go through `normalize`, which has been rewritten to `call(normalizeIR(...))`. This guarantees all recursive cycles contain one `call`.

### DIFF
--- a/hail/src/main/scala/is/hail/utils/StackSafe.scala
+++ b/hail/src/main/scala/is/hail/utils/StackSafe.scala
@@ -1,0 +1,176 @@
+package is.hail.utils
+
+import scala.annotation.tailrec
+import scala.collection.generic.{CanBuild, CanBuildFrom}
+
+object StackSafe {
+
+  def done[A](result: A): StackFrame[A] = Done(result)
+
+  def call[A](body: => StackFrame[A]): StackFrame[A] =
+    Thunk(() => body)
+
+  object StackFrame {
+    @tailrec def run[A](frame: StackFrame[A]): A = frame match {
+      case Done(result) => result
+      case Thunk(thunk) => run(thunk())
+      case more: More[x, A] => run(more.advance())
+    }
+  }
+
+  abstract class StackFrame[A] {
+    def flatMap[B](f: A => StackFrame[B]): StackFrame[B] = this match {
+      case Done(result) => f(result)
+      case thunk: Thunk[A] =>
+        val cell = new ContCell[A, B](f)
+        new More[A, B] (thunk, cell, cell)
+      case more: More[x, A] =>
+        val cell = new ContCell[A, B](f)
+        more.append(cell, cell)
+    }
+
+    def map[B](f: A => B): StackFrame[B] = flatMap(a => Done(f(a)))
+
+    final def run(): A = StackFrame.run(this)
+  }
+
+  private class ContCell[A, B](val f: A => StackFrame[B], var next: ContCell[B, _] = null)
+
+  private final case class Done[A](result: A) extends StackFrame[A]
+
+  private final case class Thunk[A](force: () => StackFrame[A]) extends StackFrame[A]
+
+  private final class More[A, B](
+    // private vars don't create getters/setters
+    private var next: StackFrame[A],
+    private var contHead: ContCell[A, _],
+    private var contTail: ContCell[_, B]
+  ) extends StackFrame[B] {
+
+    @inline def advance(): StackFrame[B] = {
+      if (contHead == null) next.asInstanceOf[StackFrame[B]]
+      else next match {
+        case Done(result) =>
+          contHead match {
+            case head: ContCell[A, x] =>
+              val ret = this.asInstanceOf[More[x, B]]
+              ret.next = head.f(result)
+              ret.contHead = head.next
+              ret
+          }
+        case thunk: Thunk[A] =>
+          next = thunk.force()
+          this
+        case more2: More[_, A] =>
+          if (contHead != null)
+            more2.append(contHead, contTail)
+          else
+            more2.asInstanceOf[More[_, B]]
+      }
+    }
+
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/utils/StackSafe.scala
```diff
@@ -0,0 +1,176 @@
...
+      case more: More[x, A] => run(more.advance())
+    }
+  }
+
+  abstract class StackFrame[A] {
+    def flatMap[B](f: A => StackFrame[B]): StackFrame[B] = this match {
+      case Done(result) => f(result)
+      case thunk: Thunk[A] =>
+        val cell = new ContCell[A, B](f)
+        new More[A, B] (thunk, cell, cell)
```
cseed, line 26: Can you do:

```
      case thunk: Thunk[A] =>
        Thunk(() => f(run(thunk)))
```

here?  I can't quite puzzle out the motivation for More and ContCell.

cseed: Nevermind, I see, the thunk bodies get nested.

patrick-schultz: To minimize the number of `Thunk` closures that are allocated, we can sometimes make multiple `flatMap` calls without any `call` wrappers. E.g. in the parser, I have
```
  def ir_value_exprs(env: IRParserEnvironment)(it: TokenIterator): StackFrame[Array[IR]] = {
    punctuation(it, "(")
    for {
      irs <- ir_value_children(env)(it)
      _ = punctuation(it, ")")
    } yield irs
  }

  def ir_value_children(env: IRParserEnvironment)(it: TokenIterator): StackFrame[Array[IR]] =
    repUntil(it, ir_value_expr(env), PunctuationToken(")"))

  def ir_value_expr(env: IRParserEnvironment)(it: TokenIterator): StackFrame[IR] = {
    punctuation(it, "(")
    for {
      ir <- call(ir_value_expr_1(env)(it))
      _ = punctuation(it, ")")
    } yield ir
  }

  def ir_value_expr_1(env: IRParserEnvironment)(it: TokenIterator): StackFrame[IR] = {
    identifier(it) match {
      ...
    }
  }
```
where the only `call` is in `ir_value_expr`, because any recursive call to `ir_value_expr_1` must go through that. So we can end up with chains of flatMaps of the shape `sf.flatMap(cont1).flatMap(cont2)...`, and for efficiency I accumulate the continuations in a linked list `More(sf, cont1, contN)`.

hail/src/main/scala/is/hail/utils/StackSafe.scala
```diff
@@ -0,0 +1,176 @@
...
+import scala.collection.generic.{CanBuild, CanBuildFrom}
+
+object StackSafe {
+
+  def done[A](result: A): StackFrame[A] = Done(result)
+
+  def call[A](body: => StackFrame[A]): StackFrame[A] =
+    Thunk(() => body)
+
+  object StackFrame {
```
tpoterba, line 13: I generally don't like nesting objects, since it generates JVM bytecode that's a bit harder for the JVM to trace to a specific line. Can we move this outside?

PR #9324, opened 2020-08-21
[hail] create relatedness group

danking:
This groups IBD and pc_relate together. I will follow up with KING. I plan to
lift some of the documentation verbiage from PC-Relate up and to unify the
mathematical presentation for PC-Relate and KING.

I did not change the implementations, I just moved them into the relatedness
package. Both functions are still available at `hl.methods.XXX`.

The import changes are just me using `(` and `)` instead of line continuation.


danking:
bump

PR #9325, opened 2020-08-21
[hail] add hl.utils.deduplicate

danking:
We deduplicate a list of names in a few places and I was about
to do it again in DNDArray and in the BlockMatrix KING
implementation. This abstracts over all existing uses of which
I am aware. I will use it in subsequent KING and DNDArray PRs.

I copied the implementation from `misc.py`.


hail/python/hail/utils/deduplicate.py
```diff
@@ -0,0 +1,57 @@
...
+        max_attempts: Optional[int] = None,
+        already_used: Set[int] = None
+) -> Tuple[List[Tuple[str, str]], List[str]]:
+    """Deduplicate the strings in `ids`.
+
+    Example
+    -------
+
+    >>> deduplicate(["a", "a", "a"])
+    ([("a": "a_1"), ("a": "a_2")], ["a", "a_1", "a_2"])
```
konradjk, line 16: These should be commas not colons (here and below), right?

PR #9329, opened 2020-08-24
[batch/genetics] regenie service backend 

akotlar:
CHANGELOG: Enable batch.genetics.regenie to use the ServiceBackend.

Thought I would contribute this in case it would be useful and out of a sense of responsibility; may be easier for me to create the service backend implementation as I'm closest to the project. Feel free to close if not beneficial.

cc @tpoterba in case you were interested

### DIFF
--- a/hail/python/hailtop/batch/genetics/regenie/regenie.py
+++ b/hail/python/hailtop/batch/genetics/regenie/regenie.py
@@ -1,16 +1,16 @@
-from hailtop.batch import Batch, LocalBackend
+from hailtop.batch import Batch, LocalBackend, ServiceBackend
 from hailtop.batch.resource import Resource
 from collections import namedtuple
 import sys
 import shlex
-from argparse import Namespace, ArgumentParser
-from typing import Set
+from argparse import Namespace, ArgumentParser, SUPPRESS
+from typing import Set, Dict
 from os.path import splitext, basename, exists
 from google.cloud import storage
 from google.cloud.storage.blob import Blob
 
 
-BatchArgs = namedtuple("BatchArgs", ['cores', 'memory', 'storage'])
+BatchArgs = namedtuple("BatchArgs", ['threads', 'memory', 'storage'])
 input_file_args = ["bgen", "bed", "pgen", "sample", "keep", "extract", "exclude", "remove",
                    "phenoFile", "covarFile"]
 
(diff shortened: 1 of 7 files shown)


hail/python/test/hailtop/batch/test_regenie.py
```diff
@@ -72,11 +77,166 @@ def test_regenie_1pheno(self):
...
+    def setUpClass(cls):
+        os.system("git clone --depth 1 --branch v1.0.5.6 https://github.com/rgcgithub/regenie.git")
+        cls.step2_out_prefix = 'test_bin_out_firth'
+
+        cls.bucket_name = get_user_config().get('batch', 'bucket')
+
+        input_folder = 'batch-tests/resources/regenie-v1.0.5.6'
+        cls.gcs_input_dir = f'gs://{cls.bucket_name}/{input_folder}'
+
+        in_cluster_key_file = '/test-gsa-key/key.json'
```
danking, line 101: This is controlled by the `HAIL_GSA_KEY_FILE` environment variable, we should not hard code it here.

hail/python/hailtop/batch/genetics/regenie/regenie.py
```diff
@@ -256,60 +254,71 @@ def prepare_jobs(batch, args: BatchArgs, step1_args: Namespace, step2_args: Name
...
+        run_parser.add_argument('--delete_scratch_on_exit', required=False, action="store_true")
+        run_parser.add_argument('--wait', required=False, action="store_true")
+        run_parser.add_argument('--open', required=False, action="store_true")
+        run_parser.add_argument('--disable_progress_bar', required=False, action="store_true")
+        run_parser.add_argument('--callback', required=False)
+
+    backend_args = backend_parser.parse_known_args(input_args)[0]
+    run_args = run_parser.parse_known_args(input_args)[0]
+
+    return {"args": args, "backend_opts": vars(backend_args), "run_opts": vars(run_args)}
```
danking, line 319: Using a different Argument Parser instance does lead to elegant code; however, the fact that the help message does not documentation the mere existence of these other arguments seems to doom the approach. Moreover, since these arguments are not REGENIE-specific, I think we ought to include a help string.

danking:
I ham-fistedly hit approve rather than request changes, thus the two reviews.

akotlar:
Hold off for a bit please, parents=[] copies arguments, and that isn't the right approach.

PR #9330, opened 2020-08-24
[query] Fix import_plink with contig_recoding and skip_invalid_loci

tpoterba:


chrisvittal:
What's the bug?

chrisvittal:
Ok. I see it. In the future, try to put a little explainer as to why, like

```
[query] Fix import_plink with contig_recoding and skip_invalid_loci

The code was using contig rather than recodedContig, leading to
incorrect results.
```

tpoterba:
oh crap, sorry, didn't see your question. Will definitely try to improve commit message in the future.

PR #9332, opened 2020-08-24
[query] Stack safe parser

patrick-schultz:
~~Stacked on #9320~~

Use the framework introduced in #9320 to make the IR parser stack safe. This touches a lot of lines, but it is a completely mechanical refactoring.

I did some preliminary benchmarking by timing the parse of the IR in `test_ld_prune`. (I chose that test fairly arbitrarily, and we can probably find better test cases, or generate synthetic large IRs.) Running a loop parsing the same IR repeatedly, with 10 burn-in rounds, and 60 timed rounds, I got the following results:
* On main
  ```
  quartiles = [4.55816, 5.209579, 5.647135]
  avg = 5.332496950000001, std = 1.13761574944604
  ```
* Using `StackSafe`, without the optimization in `repUntil`
  ```
  quantiles = [4.610798, 5.09793, 7.159075]
  avg = 5.7519015000000016, std = 1.612397075594852
  ```
* Using `StackSafe, with the optimization which reuses the `cont` closure, instead of allocating a new one for each token.
  ```
  quantiles = [4.466849, 4.826873, 5.719238]
  avg = 5.2787357833333335, std = 1.272006325411254
  ```

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Parser.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Parser.scala
@@ -10,6 +10,7 @@ import is.hail.types.{MatrixType, TableType}
 import is.hail.expr.{JSONAnnotationImpex, Nat, ParserUtils}
 import is.hail.io.{AbstractTypedCodecSpec, BufferSpec}
 import is.hail.rvd.{AbstractRVDSpec, RVDType}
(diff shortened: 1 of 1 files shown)


tpoterba (approved):
sorry this took so long! Rebase conflict, otherwise looks good.

PR #9334, opened 2020-08-24
[query] Improve version check to include 3.8

tpoterba:


hail/python/hail/__init__.py
```diff
@@ -1,8 +1,8 @@
 import pkg_resources
 import sys
 
-if sys.version_info < (3, 6):
-    raise EnvironmentError('Hail requires Python 3.6, found {}.{}'.format(
+if sys.version_info < (3, 6) or sys.version_info > (3, 7):
```
chrisvittal, line 4: This isn't quite right on `3.7.8`, `sys.version_info > (3, 7)` is `True`. How about:
```suggestion
if not (sys.version_info >= (3, 6) and sys.version_info < (3, 8)):
```

PR #9335, opened 2020-08-24
[batch] actually check the mypy exit status

danking:
I am not really sure why $? is not set to 1 if these parentheses are absent. Nonetheless,
this fixes CI to actually complain about mypy failures.


danking:
bump

PR #9337, opened 2020-08-24
[internal-gateway] lower per-instance request rate to 45

cseed:
It seems that 83 is potentially too high now.  I saw batch-driver get overwhelmed with the current setting.  However, there was also a bug related leaking database connections which also contributed to the observed behavior.  Batch was able to make forward progress with 45, so at least this is safe and can be adjusted again next time we're running at scale.


jigold:
Are you going to redeploy this on the production hail-vdc cluster?

cseed:
> Are you going to redeploy this on the production hail-vdc cluster?

Yes, done.

PR #9338, opened 2020-08-24
[benchmark] use latest SHA

danking:
Make evaluates the entire recipe before executing it. Consider this:
```
(base) # cat Makefile; rm -rf file; make foo
foo:
	echo hello > file
	echo $(shell cat file)
cat: file: No such file or directory
echo hello > file
echo 

```
The file does not exist because Make runs cat before the recipe is executed.


tpoterba:
thanks!

PR #9339, opened 2020-08-24
[many] put shared make config in config.mk file

cseed:


config.mk
```diff
@@ -0,0 +1,6 @@
+DOCKER_ROOT_IMAGE := ubuntu:18.04
```
cseed, line 1: This is not (yet) used but we should push it through so it is easy to upgrade the root image.

cseed:
Oops, quite right.  I pulled this diff from the other install, I just fixed the pieces I deployed there.

I didn't intend this to be a common include for all Makefiles (should it be?), but just pull together the relevant defines used in services Makefiles.  I grep'ed to make sure the variables defined config.mk are now only defined there.

cseed:
edit: hardcoded PROJECT and added REGION.

PR #9344, opened 2020-08-25
[query] Use runtime instead of compile-time checks for requiredness cast

tpoterba:


hail/src/main/scala/is/hail/expr/ir/Pretty.scala
```diff
@@ -117,6 +117,13 @@ object Pretty {
 
       sb.append(prettyClass(ir) )
 
+      /**
```
catoverdrive, line 120: delete?

tpoterba: left this intentionally, it's quite useful for debugging.

PR #9348, opened 2020-08-25
[hail][docs] overhaul docs build Makefile

danking:
It takes one minute to build the docs *even if nothing has changed since the
last build*. There are a few things that lengthen the feedback cycle:

- We defeat Sphinx's input cache by deleting and re-copying over all the source
  files.
- We defeat Sphinx's output cache by `mv`ing the output to a new location.
- We check that Hail is installed (at a cost of two seconds) *every* time we
  build the docs. This isn't necessary, Sphinx prints a reasonable message
  ("cannot import ...") if Hail is not installed.
- We create a wheel file every time we build the docs at a cost of several
  seconds.
- We recreate the tutorials tar even if it has not changed.

Instead, I propose this PR:

- Do not copy the source files.
- Copy the output to the new location.
- Do not check hail is installed.
- Do not even install Hail.
- Use Make to check if the tutorial tar need be recreated.

Regarding not installing Hail: even install-editable takes two seconds. It is
the developer's responsibility to ensure the right version of Hail is
installed. When you check out a branch just run `make install-editable`
once. Then edit the docs to your heart's desire, never re-install Hail.

With this PR it takes ~3.5 seconds to rebuild the docs if nothing has
changed. We do work proportional to the number of changed files, not
proportional to all files. Sphinx itself takes 2-3 seconds, so we can't do much
better than this.

Dice came up Patrick, but I imagine @tpoterba has thoughts.

### DIFF
--- a/hail/.gitignore
+++ b/hail/.gitignore
@@ -5,6 +5,7 @@ python/hail.egg-info
 python/hail/backend/hail-all-spark.jar
 python/hail/hail_pip_version
 python/hail/hail_version
+python/hail/docs/change_log.rst
 python/hail/docs/_build/*
 python/hail/docs/_static/hail_version.js
 python/hailtop/hailctl/hail_version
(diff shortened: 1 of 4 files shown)


danking:
bump

PR #9349, opened 2020-08-25
[memory] add simple read-only cache

catoverdrive:
I think I still have to fix the pylint stuff (and add typechecking?) but otherwise should be complete.

I'll add the dev doc in a separate PR.

### DIFF
--- a/hail/python/hailtop/google_storage.py
+++ b/hail/python/hailtop/google_storage.py
@@ -39,7 +39,6 @@ class GCS:
                 key_file = os.environ['HAIL_GSA_KEY_FILE']
                 credentials = google.oauth2.service_account.Credentials.from_service_account_file(key_file)
 
-
         if project:
             self.gcs_client = google.cloud.storage.Client(
                 project=project, credentials=credentials)
@@ -60,6 +59,9 @@ class GCS:
 
     def shutdown(self, wait: bool = True):
         self.blocking_pool.shutdown(wait)
+
+    async def get_etag(self, uri: str):
+        return await retry_transient_errors(self._wrap_network_call(GCS._get_etag), self, uri)
 
     async def write_gs_file_from_string(self, uri: str, string: str, *args, **kwargs):
         return await retry_transient_errors(self._wrapped_write_gs_file_from_string,
@@ -117,6 +119,13 @@ class GCS:
                                            *args,
                                            **kwargs)
         return wrapped
+
+    def _get_etag(self, uri: str):
+        b = self._get_blob(uri)
+        if b.exists():
+            b.reload()
+            return b.etag
+        return None
 
     def _write_gs_file_from_string(self, uri: str, string: str, *args, **kwargs):
         b = self._get_blob(uri)
--- a/hail/python/hailtop/google_storage.py
+++ b/hail/python/hailtop/google_storage.py
@@ -33,11 +33,12 @@ class GCS:
         # project=None doesn't mean default, it means no project:
         # https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/client.py#L86
         if credentials is None:
-            if key is None and 'HAIL_GSA_KEY_FILE' in os.environ:
+            if key is not None:
+                credentials = google.oauth2.service_account.Credentials.from_service_account_info(key)
+            elif 'HAIL_GSA_KEY_FILE' in os.environ:
                 key_file = os.environ['HAIL_GSA_KEY_FILE']
                 credentials = google.oauth2.service_account.Credentials.from_service_account_file(key_file)
-            elif key is not None:
-                credentials = google.oauth2.service_account.Credentials.from_service_account_info(key)
+
 
         if project:
             self.gcs_client = google.cloud.storage.Client(
--- a/hail/python/hailtop/google_storage.py
+++ b/hail/python/hailtop/google_storage.py
@@ -122,10 +122,8 @@ class GCS:
 
     def _get_etag(self, uri: str):
         b = self._get_blob(uri)
-        if b.exists():
-            b.reload()
-            return b.etag
-        return None
+        b.reload()
+        return b.etag
 
     def _write_gs_file_from_string(self, uri: str, string: str, *args, **kwargs):
         b = self._get_blob(uri)
--- a/hail/python/hailtop/google_storage.py
+++ b/hail/python/hailtop/google_storage.py
@@ -27,13 +27,18 @@ class GCS:
                  blocking_pool: concurrent.futures.Executor,
                  *,
                  project: Optional[str] = None,
+                 key: Optional[str] = None,
                  credentials: Optional[google.oauth2.service_account.Credentials] = None):
         self.blocking_pool = blocking_pool
         # project=None doesn't mean default, it means no project:
         # https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/client.py#L86
-        if credentials is None and 'HAIL_GSA_KEY_FILE' in os.environ:
-            credentials = google.oauth2.service_account.Credentials.from_service_account_file(
-                os.environ['HAIL_GSA_KEY_FILE'])
+        if credentials is None:
+            if key is None and 'HAIL_GSA_KEY_FILE' in os.environ:
+                key_file = os.environ['HAIL_GSA_KEY_FILE']
+                credentials = google.oauth2.service_account.Credentials.from_service_account_file(key_file)
+            elif key is not None:
+                credentials = google.oauth2.service_account.Credentials.from_service_account_info(key)
+
         if project:
             self.gcs_client = google.cloud.storage.Client(
                 project=project, credentials=credentials)
--- a/memory/Makefile
+++ b/memory/Makefile
@@ -6,6 +6,12 @@ PYTHON := PYTHONPATH=$(PYTHONPATH) python3
 MEMORY_LATEST = gcr.io/$(PROJECT)/memory:latest
 MEMORY_IMAGE = gcr.io/$(PROJECT)/memory:$(shell docker images -q --no-trunc memory | sed -e 's,[^:]*:,,')
 
+ifeq ($(NAMESPACE), "default")
+BATCH_PODS_NS = batch-pods
+else
+BATCH_PODS_NS = $(NAMESPACE)
+endif
+
 .PHONY: check
 check:
 	$(PYTHON) -m flake8  --config ../setup.cfg memory
@@ -30,7 +36,7 @@ push: build
 deploy: push
 	! [ -z $(NAMESPACE) ]  # set NAMESPACE
 	kubectl -n $(NAMESPACE) apply -f service-account.yaml
-	python3 ../ci/jinja2_render.py '{"default_ns":{"name":"default"}}' service-account-batch-pods.yaml service-account-batch-pods.yaml.out
+	python3 ../ci/jinja2_render.py '{"default_ns":{"name":"$(NAMESPACE)"}}' service-account-batch-pods.yaml service-account-batch-pods.yaml.out
 	kubectl -n batch-pods apply -f service-account-batch-pods.yaml.out
-	python3 ../ci/jinja2_render.py '{"code":{"sha":"$(shell git rev-parse --short=12 HEAD)"},"deploy":true,"memory_image":{"image":"$(MEMORY_IMAGE)"},"batch_pods_ns":{"name":"batch-pods"},"global":{"project":"$(PROJECT)","domain":"$(DOMAIN)"}}' deployment.yaml deployment.yaml.out
+	python3 ../ci/jinja2_render.py '{"code":{"sha":"$(shell git rev-parse --short=12 HEAD)"},"deploy":true,"memory_image":{"image":"$(MEMORY_IMAGE)"},"batch_pods_ns":{"name":"$(BATCH_PODS_NS)"},"global":{"project":"$(PROJECT)","domain":"$(DOMAIN)"}}' deployment.yaml deployment.yaml.out
 	kubectl -n $(NAMESPACE) apply -f deployment.yaml.out
--- a/memory/Makefile
+++ b/memory/Makefile
@@ -37,6 +37,6 @@ deploy: push
 	! [ -z $(NAMESPACE) ]  # set NAMESPACE
 	kubectl -n $(NAMESPACE) apply -f service-account.yaml
 	python3 ../ci/jinja2_render.py '{"default_ns":{"name":"$(NAMESPACE)"}}' service-account-batch-pods.yaml service-account-batch-pods.yaml.out
-	kubectl -n batch-pods apply -f service-account-batch-pods.yaml.out
+	kubectl -n $(BATCH_PODS_NS) apply -f service-account-batch-pods.yaml.out
 	python3 ../ci/jinja2_render.py '{"code":{"sha":"$(shell git rev-parse --short=12 HEAD)"},"deploy":true,"memory_image":{"image":"$(MEMORY_IMAGE)"},"batch_pods_ns":{"name":"$(BATCH_PODS_NS)"},"global":{"project":"$(PROJECT)","domain":"$(DOMAIN)"}}' deployment.yaml deployment.yaml.out
 	kubectl -n $(NAMESPACE) apply -f deployment.yaml.out
--- a/memory/Makefile
+++ b/memory/Makefile
@@ -1,4 +1,4 @@
-PROJECT := $(shell gcloud config get-value project)
+include ../config.mk
 
 PYTHONPATH := $${PYTHONPATH:+$${PYTHONPATH}:}../hail/python:../gear
 PYTHON := PYTHONPATH=$(PYTHONPATH) python3
@@ -26,9 +26,11 @@ push: build
 	docker push $(MEMORY_IMAGE)
 
 .PHONY: deploy
+
 deploy: push
-	kubectl -n default apply -f service-account.yaml
+	! [ -z $(NAMESPACE) ]  # set NAMESPACE
+	kubectl -n $(NAMESPACE) apply -f service-account.yaml
 	python3 ../ci/jinja2_render.py '{"default_ns":{"name":"default"}}' service-account-batch-pods.yaml service-account-batch-pods.yaml.out
 	kubectl -n batch-pods apply -f service-account-batch-pods.yaml.out
 	python3 ../ci/jinja2_render.py '{"code":{"sha":"$(shell git rev-parse --short=12 HEAD)"},"deploy":true,"memory_image":{"image":"$(MEMORY_IMAGE)"},"batch_pods_ns":{"name":"batch-pods"},"global":{"project":"$(PROJECT)","domain":"$(DOMAIN)"}}' deployment.yaml deployment.yaml.out
-	kubectl -n default apply -f deployment.yaml.out
+	kubectl -n $(NAMESPACE) apply -f deployment.yaml.out
--- a/memory/Makefile
+++ b/memory/Makefile
@@ -1,0 +1,34 @@
+PROJECT := $(shell gcloud config get-value project)
+
+PYTHONPATH := $${PYTHONPATH:+$${PYTHONPATH}:}../hail/python:../gear
+PYTHON := PYTHONPATH=$(PYTHONPATH) python3
+
+MEMORY_LATEST = gcr.io/$(PROJECT)/memory:latest
+MEMORY_IMAGE = gcr.io/$(PROJECT)/memory:$(shell docker images -q --no-trunc memory | sed -e 's,[^:]*:,,')
+
+.PHONY: check
+check:
+	$(PYTHON) -m flake8  --config ../setup.cfg memory
+	$(PYTHON) -m pylint --rcfile ../pylintrc memory --score=n
+
+.PHONY: build
+build:
+	make -C ../docker build
+	-docker pull $(MEMORY_LATEST)
+	python3 ../ci/jinja2_render.py '{"service_base_image":{"image":"service-base"}}' Dockerfile Dockerfile.out
+	docker build -t memory -f Dockerfile.out --cache-from memory,$(MEMORY_LATEST),service-base ..
+
+.PHONY: push
+push: build
+	docker tag memory $(MEMORY_LATEST)
+	docker push $(MEMORY_LATEST)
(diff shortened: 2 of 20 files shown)


cseed:
Awesome!  Look forward to reading it over tomorrow!

memory/memory/memory.py
```diff
@@ -0,0 +1,100 @@
...
+
+socket = '/tmp/redis.sock'
+
+@routes.get('/healthcheck')
+async def healthcheck(request):  # pylint: disable=unused-argument
+    return web.Response()
+
+@routes.post('/getfile')
+@rest_authenticated_users_only
+async def getfile(request, userdata):
```
cseed, line 33: Few comments on the API:

 - I think it is more common to call these objects instead of files in this context.
 - APIs should be versioned.  Our API URL naming scheme is `/api/<version>/...`.  Work-in-progress APIs should be `v1alpha`.
 - This should be GET, not POST
 - GET requests should not have a semantically meaningful body.  That means the object should go into the URL or inside they query parameters.  For fixed objects, the URL seems more natural.  So it would be something like `GET /api/v1alpha/objects/<bucket>/<key>`.  See:

 > https://stackoverflow.com/questions/978061/http-get-with-request-body
 > Server semantics for GET, however, are restricted such that a body, if any, has no semantic meaning to the request.

memory/memory/memory.py
```diff
@@ -0,0 +1,100 @@
...
+    body = await request.json()
+    billing_project = body['billing_project']
```
cseed, line 35: Hmm.  This doesn't seem used now, let's remove it.  This is also tricky: I run pipelines under two billing accounts that use the same input file.  Do they share the cache?  Who pays for it?

memory/memory/memory.py
```diff
@@ -0,0 +1,100 @@
...
+    filename = body['filename']
+    userinfo = await get_or_add_user(request.app, userdata, billing_project)
+    body = await get_file_or_none(request.app, userdata['username'], userinfo, filename)
+    if body is None:
+        raise web.HTTPNotFound()
+    print(f'{filename} got result of length { len(body) }')
```
cseed, line 41: No space in around {}s in Python f-strings.

Remove or convert to logging message.

memory/memory/memory.py
```diff
@@ -0,0 +1,100 @@
...
+            app['worker_pool'].call_or_error(load_file, app['redis_pool'], app['file_queue'], filekey, userinfo['fs'], filepath)
+        except asyncio.QueueFull:
+            pass
+    return result
+
+async def load_file(redis, files, file_key, fs, filepath):
+    files.add(file_key)
+    data = await fs.read_binary_gs_file(filepath)
+    await redis.execute('SET', file_key, data)
+    files.remove(file_key)
```
cseed, line 75: This needs to be protected with a finally.

memory/memory/memory.py
```diff
@@ -0,0 +1,100 @@
...
+
+async def on_startup(app):
+    app['thread_pool'] = concurrent.futures.ThreadPoolExecutor()
+    app['worker_pool'] = AsyncWorkerPool(parallelism=4, queue_size=10)
+    app['file_queue'] = set()
```
cseed, line 80: This does not seem like queue to me.

memory/run.sh
```diff
@@ -0,0 +1,2 @@
+/redis-6.0.6/src/redis-server /redis/redis.conf &
```
cseed, line 1: Unused.

memory/memory/client.py
```diff
@@ -0,0 +1,79 @@
...
+        self.url = deploy_config.base_url('memory')
+
+        if session is None:
+            session = get_context_specific_ssl_client_session(
+                raise_for_status=True,
+                timeout=aiohttp.ClientTimeout(total=60))
+        self._session = session
+
+        if fs is None:
+            fs = GCS(blocking_pool=concurrent.futures.ThreadPoolExecutor(), project=billing_project)
```
cseed, line 29: I'm a little confused by this.  `billing_project` normally refers to a Hail billing project, not a Google project.

memory/memory/client.py
```diff
@@ -0,0 +1,79 @@
...
+        data = await self._get_file_if_exists(filename)
+        if data is not None:
+            return data
+        return await self._fs.read_binary_gs_file(filename)
+
+    async def close(self):
+        await self._session.close()
+        self._session = None
+
+class BlockingMemoryClient:
```
cseed, line 67: Do we need this?  I don't think so.  We only need synchronous/blocking clients if they are user-facing.  Internal code should all be async.  I think the blocking Batch client should probably go away, and the user-facing Batch itself, which is synchronous, should be written against the internal async client.  Let's use this model going forward.  So the only thing that might be synchronous is a user-facing FileSystem interface that would replace the hadoop_* methods.

memory/memory/client.py
```diff
@@ -0,0 +1,79 @@
...
+
+    async def _get_file_if_exists(self, filename):
+        request = {
+            'billing_project': self.billing_project,
+            'filename': filename}
+        try:
+            response = await request_retry_transient_errors(
+                self._session, 'POST',
+                self.url + '/getfile', json=request, headers=self._headers)
+            async with response:
```
cseed, line 49: More natural to write this as `async with await request_...(...) as request: ...`.

memory/memory/client.py
```diff
@@ -0,0 +1,79 @@
...
+                if response.status == 200:
```
cseed, line 50: The session raises if the response is not successful, so this is unnecessary.

memory/memory/client.py
```diff
@@ -0,0 +1,79 @@
...
+                    return await response.read()
+        except aiohttp.ClientResponseError as e:
+            if e.status == 404:
+                return None
+            raise e
+
+    async def read_file(self, filename):
+        data = await self._get_file_if_exists(filename)
```
cseed, line 58: I think this should hit Google and verify the object exists before checking the cache.  It should pass the etag to the cache, and that (+/- the object key/path) should be the lookup key.

cseed: Can we benchmark getting the etags/metadata?

memory/memory/client.py
```diff
@@ -0,0 +1,79 @@
...
+import concurrent
+
+from hailtop.auth import service_auth_headers
+from hailtop.config import get_deploy_config
+from hailtop.google_storage import GCS
+from hailtop.tls import get_context_specific_ssl_client_session
+from hailtop.utils import async_to_blocking, request_retry_transient_errors
+
+@asyncinit
+class AsyncMemoryClient:
```
cseed, line 12: Just so we are clear, I imagine this will go away and be replaced with a unified file system interface that has various caching options, yes?

memory/test/test_memory.py
```diff
@@ -0,0 +1,42 @@
...
+
+from hailtop.config import get_user_config
+from hailtop.google_storage import GCS
+
+
+class FileTests(unittest.TestCase):
+    def setUp(self):
+        bucket_name = get_user_config().get('batch', 'bucket')
+        token = uuid.uuid4()
+        self.test_path = f'gs://{ bucket_name }/memory-tests/{ token }'
```
cseed, line 14: f-string spacing again, and in various other places.

memory/test/test_memory.py
```diff
@@ -0,0 +1,42 @@
...
+
+        self.fs = GCS(concurrent.futures.ThreadPoolExecutor(), project='hail-vdc')
```
cseed, line 16: You need to pass the project in, it is a CI global parameter.  See batch deployment, for example.

I'm not totally sure why GCS needs a project.  Storage is globally scoped, and you only need a project for requestor pays.  Our aiogoogle client should fix this.

memory/test/test_memory.py
```diff
@@ -0,0 +1,42 @@
+import concurrent
+import unittest
+import uuid
+from memory.client import BlockingMemoryClient
+
+from hailtop.config import get_user_config
+from hailtop.google_storage import GCS
+
+
+class FileTests(unittest.TestCase):
```
cseed, line 10: We should have and additional test that verifies users are isolated.

cseed (dismissed):

memory/deployment.yaml
```diff
@@ -0,0 +1,71 @@
...
+      labels:
+        app: memory
+        hail.is/sha: "{{ code.sha }}"
+    spec:
+      serviceAccountName: memory
+{% if deploy %}
+      priorityClassName: production
+{% endif %}
+      containers:
+        - name: memory
```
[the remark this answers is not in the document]
catoverdrive: Not sure what you mean by this. If I have a redis instance running in a standalone instance, I'll need to use redis's auth stuff to ensure that only the memory container is capable of talking to the redis instance running in the other container within the k8s cluster. Seemed easier and more secure to run them in the same container, expose networking the same way we do for other services, and use a unix socket for communication.

If you mean that I should have some sort of container to manage load, etc. when we start sharding data, and that that should be the container called "memory" instead of the containers actually storing data, I can rename this container? It didn't seems super worth splitting them out right now since there's only one node.

memory/memory/memory.py
catoverdrive: Yeah. It represents the files currently in the read queue, but isn't itself the queue. I can rename to something more descriptive, like `files_in_progress`?

memory/memory/client.py
catoverdrive: I can move this into the test suite. I use it for the tests.

memory/memory/client.py
catoverdrive: Yeah! This was a simple client I wrote, mostly for testing reasons.

memory/deployment.yaml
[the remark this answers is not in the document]
cseed: Pod is the natural security boundary in k8s.  Pods are a collection of containers.  Notice, in the YAML, containers is a list, not an item.  Containers in a pod are designed to work together.  They share a network and IP.  You can give them a shared file system.  I'm not sure how much of the container security reading group you attended, but containers are a collection of techniques for managing isolation **and** sharing of resources in running processes.  So while the containers in a pod are containers, they are less isolated than containers in two different pods, say.  The general recommendation is that each process in a pod should have its own container.  To give them a shared filesystem for the socket, use an emptyDir volume:

https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

For more, see:

https://kubernetes.io/docs/concepts/workloads/pods/

"A Pod can encapsulate an application composed of multiple co-located containers that are tightly coupled and need to share resources."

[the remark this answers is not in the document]
catoverdrive: ah, cool! That also solves the "build redis from source" problem since there's a prebuilt redis container.

memory/test/test_memory.py
catoverdrive: i think this is going to be involved enough that I'd like to address it in a separate PR.

memory/memory/client.py
catoverdrive: So the way I ended up implementing this:
- client does an etag check for existence/version info, passes etag info to server
- server checks to see if file exists in cache and if the etag matches; if so, return file
- if file doesn't exist or etag does not match, add file to read queue
- when the server reads the file, it also gets the etag and stores both into the cache (to deal with the case where the file could be updated between when the client sends the request and the server reads the file)

Right now I'm getting the etag and the file data in separate calls since that was the easiest way to implement it on the current google storage interface; I'll unify the calls when we lift this to the new API, as discussed.

memory/memory/memory.py
```diff
@@ -0,0 +1,120 @@
...
+
+def make_redis_key(username, filepath):
+    return f'{ username }_{ filepath }'
+
+
+async def get_file_or_none(app, username, userinfo, filepath, etag):
+    filekey = make_redis_key(username, filepath)
+    fs = userinfo['fs']
+
+    cached_etag = await app['redis_pool'].execute('HGET', filekey, 'etag', encoding='ascii')
```
cseed, line 70: You could eliminate a call to redis if you put the etag in the redis key.  Your solution has the advantage it removes the old version from the cache when caching the new one, however, I'm not sure how big a deal that is since the old version will just expire out of the cache anyway.

However, this solution does have a potential race condition:

 - read the cached etag for filekey
 - asynchronously, cache a new version of filekey
 - check the etag and cached etag: they match!
 - read the new body which is now wrong

memory/memory/memory.py
```diff
@@ -0,0 +1,120 @@
...
+        data = await fs.read_binary_gs_file(filepath)
+        etag = await fs.get_etag(filepath)
+        await redis.execute('HMSET', file_key, 'etag', etag.encode('ascii'), 'body', data)
+    finally:
+        files.remove(file_key)
+
+
+async def on_startup(app):
+    app['thread_pool'] = concurrent.futures.ThreadPoolExecutor()
+    app['worker_pool'] = AsyncWorkerPool(parallelism=4, queue_size=10)
```
cseed, line 98: The parallelism can be much higher.  I'd say start with 100.

memory/test/test_memory.py
```diff
@@ -0,0 +1,58 @@
...
+    def _get_file_if_exists(self, filename):
+        return async_to_blocking(self._client._get_file_if_exists(filename))
+
+    def read_file(self, filename):
+        return async_to_blocking(self._client.read_file(filename))
+
+    def close(self):
+        return async_to_blocking(self._client.close())
+
+class Tests(unittest.TestCase):
```
cseed, line 26: Although I have mixed feelings about it, I'm a pytest convert.  It also makes async tests very easy, and has better support for fixtures (setup and teardown).  You don't need to change it in this PR, but we should switch afterwards.  It will also remove the need for a blocking client.  See my recent FS per for examples:

https://github.com/hail-is/hail/blob/208fbe2c20682590779d67dbf1c4dbf077cee0f2/hail/python/test/hailtop/test_aiogoogle.py

cseed (dismissed):
Looking good!

memory/Makefile
```diff
@@ -0,0 +1,42 @@
...
+	docker tag memory $(MEMORY_IMAGE)
+	docker push $(MEMORY_IMAGE)
+
+.PHONY: deploy
+
+deploy: push
+	! [ -z $(NAMESPACE) ]  # set NAMESPACE
+	kubectl -n $(NAMESPACE) apply -f service-account.yaml
+	python3 ../ci/jinja2_render.py '{"default_ns":{"name":"$(NAMESPACE)"}}' service-account-batch-pods.yaml service-account-batch-pods.yaml.out
+	kubectl -n batch-pods apply -f service-account-batch-pods.yaml.out
```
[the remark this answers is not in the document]
catoverdrive: Yep!

memory/Makefile
```diff
@@ -0,0 +1,42 @@
...
+PYTHON := PYTHONPATH=$(PYTHONPATH) python3
+
+MEMORY_LATEST = gcr.io/$(PROJECT)/memory:latest
+MEMORY_IMAGE = gcr.io/$(PROJECT)/memory:$(shell docker images -q --no-trunc memory | sed -e 's,[^:]*:,,')
+
+ifeq ($(NAMESPACE), "default")
+BATCH_PODS_NS = batch-pods
+else
+BATCH_PODS_NS = $(NAMESPACE)
+endif
```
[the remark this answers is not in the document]
catoverdrive: This is basically what Dan does for the shuffler---it errors in deploy if `NAMESPACE` isn't set.

I think he wants to move away from using makefiles for deploying stuff (which I think we already don't do for most stuff), so the only real use case I see for the deploy step in the Makefile is to deploy things for testing in my dev namespace without having to deploy all the dependencies, as I would need to do using dev deploy. I don't actually want to deploy to the default namespace unless I'm really actively definitely trying to replace the thing in production, which I don't think is going to be the common use case, if any; I certainly don't want to accidentally omit the namespace and deploy something that I was trying to test.

The actual thing I want to do here is to just rely on the user to have their kubeconfig set correctly, with the correct namespace, but I don't think anyone else uses that; thoughts?

memory/Makefile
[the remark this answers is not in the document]
catoverdrive: Yes, oops!

memory/memory/memory.py
```diff
@@ -0,0 +1,120 @@
...
+        except asyncio.QueueFull:
+            pass
+    return None
+
+
+async def load_file(redis, files, file_key, fs, filepath):
+    try:
+        files.add(file_key)
+        data = await fs.read_binary_gs_file(filepath)
+        etag = await fs.get_etag(filepath)
```
cseed, line 90: This is also a race condition but as you pointed out on my FS PR, this will be fixed by making the etag available on the read object stream.

catoverdrive: Yep! This is exactly why I wanted the etag to be available and I'll fix this when I port to the new python aiostorage stuff.

hail/python/hailtop/google_storage.py
```diff
@@ -112,6 +120,13 @@ async def wrapped(*args, **kwargs):
                                            **kwargs)
         return wrapped
 
+    def _get_etag(self, uri: str):
+        b = self._get_blob(uri)
+        if b.exists():
+            b.reload()
+            return b.etag
+        return None
```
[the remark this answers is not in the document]
catoverdrive: I fixed 2. Going to skip 1. since I didn't introduce `get_blob` in this PR and I'm using the non-wrapped/async version here, anyway and `get_etag` should be wrapped since it does make an API call.

memory/Makefile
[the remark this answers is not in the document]
cseed: OK, fair.  Since shuffler is already doing it, let's address this separately.

Yeah, we're all agreement about the Makefiles.  We should build a local tool that executes build.yaml.

 > the only real use case I see for the deploy step in the Makefile is to deploy things for testing in my dev namespace without having to deploy all the dependencies

Ah, so I had a rough plan to address this, but it never got executed.  There were a few variations I haven't decided among.  One issues is there are really two kinds of dependencies: coordination between external state (e.g. deploy of a service you depend on) and "value" dependencies that produce configuration data used by the step (e.g. build_image is used by deploy_service in the k8s yaml).  You either have to run the value dependencies, or take them from the current values deployed in default (I think that's always possible but I haven't checked).  So basically I wanted an option for `dev deploy` like `--no-deps` or `--assume a,b,c` that skipped a,b,c (and their dependencies) if the requested steps depend on them.  FYI @danking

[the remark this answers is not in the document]
cseed: > I don't actually want to deploy to the default namespace unless I'm really actively definitely trying to replace the thing in production, which I don't think is going to be the common use case, if any; I certainly don't want to accidentally omit the namespace and deploy something that I was trying to test.

 > The actual thing I want to do here is to just rely on the user to have their kubeconfig set correctly, with the correct namespace, but I don't think anyone else uses that; thoughts?

I agree, the Makefile here is slightly different than the others.  The others are used to bootstrap a fresh cluster (or manually patch a critical issue).  Nobody uses them for test deployment (I think default is hardwired in most of them).

I do set the kubectl namespace a lot.  I have this alias:

```
kcd () {
    k config set-context ... --namespace $1
}
```

This would be reasonable to use.  There is still some danger deploying to default.

memory/memory/memory.py
```diff
@@ -0,0 +1,120 @@
...
+            log.info(f"memory: Loading {filepath} to cache for user {username}")
+            app['worker_pool'].call_nowait(load_file, app['redis_pool'], app['files_in_progress'], filekey, fs, filepath)
+        except asyncio.QueueFull:
+            pass
+    return None
+
+
+async def load_file(redis, files, file_key, fs, filepath):
+    try:
+        files.add(file_key)
```
cseed, line 88: Ah, sorry I missed this:  there should be a test here.  You need to do nothing if you're already loading the file.

cseed (changes requested):
So close: found one logical issue.

memory/memory/memory.py
catoverdrive: I realized I hadn't implemented this as intended: the actual semantics I'd wanted here were to only add the file to the queue if it wasn't already in the queue, and not just if it wasn't being actively loaded. I moved the `files.add(file_key)` line to reflect that; I *think* that makes adding a check here redundant, but I've added one just in case.

cseed: Hmm, I still not sure about the current logic.  I don't see where the loading is skipped if it is queued.  I feel like you want something like, at the call site:

```
  if not file_key in files:
    files.add(file_key)
    call_nowait(load_file, ...)

  def load_file():
    try:
      # load that file
    finally:
      files.remove(file_key)
```

memory/memory/memory.py
```diff
@@ -0,0 +1,121 @@
...
+    filekey = make_redis_key(username, filepath)
+    fs = userinfo['fs']
+
+    cached_etag, result = await app['redis_pool'].execute('HMGET', filekey, 'etag', 'body')
+    if cached_etag is not None and cached_etag.decode('ascii') == etag:
+        log.info(f"memory: Retrieved file {filepath} for user {username} with etag'{etag}'")
+        return cached_etag.decode('ascii'), result
+
+    log.info(f"memory: Couldn't retrieve file {filepath} for user {username}: current version not in cache.")
+    if filekey not in app['files_in_progress']:
```
cseed, line 77: I missed this, this looks good.

memory/memory/memory.py
```diff
@@ -0,0 +1,121 @@
...
+            log.info(f"memory: Loading {filepath} to cache for user {username}")
+            app['worker_pool'].call_nowait(load_file, app['redis_pool'], app['files_in_progress'], filekey, fs, filepath)
+            files.add(file_key)
+        except asyncio.QueueFull:
+            pass
+    return None
+
+
+async def load_file(redis, files, file_key, fs, filepath):
+    if file_key in files:
```
cseed, line 88: Strike the last comment, I misunderstood.  Everything looks good except, shouldn't this always be true now?

cseed (changes requested):

memory/memory/memory.py
catoverdrive: Yep! I'm pretty sure it wasn't necessary, left it in just in case. I'll take it out.

cseed: > I'm pretty sure it wasn't necessary, left it in just in case.

It threw me off because I was trying to understand how the false case arised.  I was going to say you could change it to an assert, but I actually you already have: `set.remove` throws if the element is not present.  In general, I like to distinguish "remove (if present)" from "remove (assert present)" as a signal to the reader.

cseed (approved):

PR #9352, opened 2020-08-26
[hailctl dataproc] Allow specifying project for hailctl dataproc connect

nawatts:
Unlike many other `hailctl dataproc` commands, there is currently no way to specify a project for `hailctl dataproc connect`.


nawatts:
Scorecard returned @johnc1231, but it looks like I don't have permission to request a review.

PR #9353, opened 2020-08-26
[batch] block containers from using the metadata server

danking:
We're supposed to use DOCKER-USER https://docs.docker.com/network/iptables/.

I also added a test.

### DIFF
--- a/batch/test/test_batch.py
+++ b/batch/test/test_batch.py
@@ -557,3 +557,11 @@ echo $HAIL_BATCH_WORKER_IP
             assert e.status == 400
         else:
             assert False, f'should receive a 400 Bad Request {batch.id}'
+
+    def test_verify_no_access_to_metadata_server(self):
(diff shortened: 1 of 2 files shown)


danking:
reassigning for speedy review.

batch/test/test_batch.py
```diff
@@ -557,3 +557,11 @@ def test_duplicate_parents(self):
...
         else:
             assert False, f'should receive a 400 Bad Request {batch.id}'
+
+    def test_verify_no_access_to_metadata_server(self):
+        builder = self.client.create_batch()
+        j = builder.create_job('google/cloud-sdk', ['gcloud', 'auth', 'list'])
+        builder.submit()
+        status = j.wait()
+        assert status['state'] == 'Success', status
+        assert "No credentialed accounts." in j.log()['main']
```
jigold, line 567: Can you add the failure to print if the last assert fails?

PR #9346, opened 2020-08-25
[batch] add tokens to aggregated_batch_resources

jigold:
I tested this by adding the check_aggregated_resources loop which didn't throw any errors. I tested the cost by looking at the UI and submitting one job that cost $0.0004 and then two of that same job in a batch and it cost $0.0008. The database looked like this:

mysql> select * FROM aggregated_batch_resources where batch_id > 46;
+----------+--------------------------+-------------+-------+
| batch_id | resource                 | usage       | token |
+----------+--------------------------+-------------+-------+
|       47 | compute/n1-preemptible/1 |    25867000 |    11 |
|       47 | compute/n1-preemptible/1 |           0 |   115 |
|       47 | compute/n1-preemptible/1 |           0 |   132 |
|       47 | disk/local-ssd/1         |  9932928000 |    11 |
|       47 | disk/local-ssd/1         |           0 |    54 |
|       47 | disk/local-ssd/1         |           0 |   132 |
|       47 | disk/pd-ssd/1            |   529756160 |    11 |
|       47 | disk/pd-ssd/1            |           0 |   132 |
|       47 | disk/pd-ssd/1            |           0 |   186 |
|       47 | ip-fee/1024/1            |    26487808 |    11 |
|       47 | ip-fee/1024/1            |           0 |   132 |
|       47 | ip-fee/1024/1            |           0 |   188 |
|       47 | memory/n1-preemptible/1  |    99329280 |    11 |
|       47 | memory/n1-preemptible/1  |           0 |    26 |
|       47 | memory/n1-preemptible/1  |           0 |   132 |
|       47 | service-fee/1            |    25867000 |    11 |
|       47 | service-fee/1            |           0 |   110 |
|       47 | service-fee/1            |           0 |   132 |
|       48 | compute/n1-preemptible/1 |           0 |    31 |
|       48 | compute/n1-preemptible/1 |           0 |    76 |
|       48 | compute/n1-preemptible/1 |    27659000 |    94 |
|       48 | compute/n1-preemptible/1 |    26520000 |   122 |
|       48 | compute/n1-preemptible/1 |           0 |   156 |
|       48 | compute/n1-preemptible/1 |           0 |   168 |
|       48 | disk/local-ssd/1         | 10621056000 |    94 |
|       48 | disk/local-ssd/1         | 10183680000 |   122 |
|       48 | disk/local-ssd/1         |           0 |   125 |
|       48 | disk/local-ssd/1         |           0 |   154 |
[...]

PR #9354, opened 2020-08-27
[batch] Aggregated billing project resources infrastructure

jigold:
Stacked on #9346 

Changes:
- infrastructure needed for kill switch
- UI page
- Default value for the limit is None

Testing:
- In the database migration, there's two updates that populate the initial state of the aggregated_billing_resources_table. I tested this by hand using a database that hadn't been migrated previously, but this might be good to double check.
- I ran the `check_resource_aggregation` loop while running `test_batch` and made sure there were no errors.
- I tested the UI page editing the limits with negative values and gibberish by hand to make sure those failed. I also refreshed the page to make sure the values were in the database and the update worked.

So here's a PR where I convinced myself it was correct a couple of days ago, but the longer this sits, the less confident I'm going to be that there's not a mistake somewhere, especially if there are a lot of changes that need to be made to the code.

### DIFF
--- a/batch/batch/batch.py
+++ b/batch/batch/batch.py
@@ -7,7 +7,7 @@ import traceback
 
 from hailtop.utils import (
     time_msecs, sleep_and_backoff, is_transient_error,
-    time_msecs_str, humanize_timedelta_msecs, cost_str)
+    time_msecs_str, humanize_timedelta_msecs)
 from hailtop.tls import in_cluster_ssl_client_session
 
 from .globals import complete_states, tasks, STATUS_FORMAT_VERSION
@@ -73,7 +73,7 @@ def batch_record_to_dict(record):
     d['msec_mcpu'] = msec_mcpu
 
     cost = format_version.cost(record['msec_mcpu'], record['cost'])
-    d['cost'] = cost_str(cost)
+    d['cost'] = cost
 
     return d
 
@@ -229,7 +229,7 @@ def job_record_to_dict(record, name):
     result['msec_mcpu'] = msec_mcpu
 
     cost = format_version.cost(record['msec_mcpu'], record['cost'])
-    result['cost'] = cost_str(cost)
+    result['cost'] = cost
 
     return result
 
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1242,7 +1242,7 @@ async def ui_get_billing_limits(request, userdata):
     return await render_template('batch', request, userdata, 'billing_limits.html', page_context)
 
 
-def _parse_billing_limit(billing_project, limit):
+def _parse_billing_limit(limit):
     if limit == 'None' or limit is None:
         limit = None
     else:
@@ -1250,12 +1250,14 @@ def _parse_billing_limit(billing_project, limit):
             limit = float(limit)
             assert limit >= 0
         except Exception as e:
-            raise InvalidBillingLimitError(billing_project, limit) from e
+            raise InvalidBillingLimitError(limit) from e
     return limit
 
 
 async def _edit_billing_limit(db, billing_project, limit):
-    limit = _parse_billing_limit(billing_project, limit)
+    log.info(f'limit argument in edit billing limit {limit}')
+    limit = _parse_billing_limit(limit)
+    log.info(f'limit argument after parsing {limit}')
 
     @transaction(db)
     async def insert(tx):
@@ -1280,6 +1282,7 @@ UPDATE billing_projects SET `limit` = %s WHERE name = %s;
 ''',
             (limit, billing_project))
     await insert()  # pylint: disable=no-value-for-parameter
+    log.info(f'limit argument after inserting into the database {limit}')
 
 
 @routes.post('/api/v1alpha/billing_limits/{billing_project}/edit')
@@ -1290,7 +1293,9 @@ async def post_edit_billing_limits(request, userdata):  # pylint: disable=unused
     billing_project = request.match_info['billing_project']
     data = await request.json()
     limit = data['limit']
+    log.info(f'limit before editing {limit}')
     await _handle_api_error(_edit_billing_limit, db, billing_project, limit)
+    log.info(f'limit after editing {limit}')
     return web.json_response({'billing_project': billing_project, 'limit': limit})
 
 
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1249,7 +1249,7 @@ def _parse_billing_limit(limit, billing_project):
         try:
             limit = float(limit)
             assert limit >= 0
-        except Exception:
+        except Exception:  # pylint: disable=raise-missing-from
             raise InvalidBillingLimitError(limit, billing_project)
     return limit
 
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1255,9 +1255,7 @@ def _parse_billing_limit(limit):
 
 
 async def _edit_billing_limit(db, billing_project, limit):
-    log.info(f'limit argument in edit billing limit {limit}')
     limit = _parse_billing_limit(limit)
-    log.info(f'limit argument after parsing {limit}')
 
     @transaction(db)
     async def insert(tx):
@@ -1282,7 +1280,6 @@ UPDATE billing_projects SET `limit` = %s WHERE name = %s;
 ''',
             (limit, billing_project))
     await insert()  # pylint: disable=no-value-for-parameter
-    log.info(f'limit argument after inserting into the database {limit}')
 
 
 @routes.post('/api/v1alpha/billing_limits/{billing_project}/edit')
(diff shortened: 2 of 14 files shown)


batch/sql/estimated-current.txt
```diff
@@ -196,6 +201,15 @@ CREATE TABLE IF NOT EXISTS `batch_attributes` (
 ) ENGINE = InnoDB;
 CREATE INDEX batch_attributes_key_value ON `batch_attributes` (`key`, `value`(256));
 
+CREATE TABLE IF NOT EXISTS `aggregated_billing_project_resources` (
+  `billing_project` VARCHAR(100) NOT NULL,
+  `resource` VARCHAR(100) NOT NULL,
+  `usage` BIGINT NOT NULL DEFAULT 0,
+  PRIMARY KEY (`billing_project`, `resource`),
+  FOREIGN KEY (`billing_project`) REFERENCES billing_projects(name) ON DELETE CASCADEE,
```
danking, line 209: CASCADE

batch/sql/add-aggregated-billing-project-resources.sql
```diff
@@ -0,0 +1,127 @@
+CREATE INDEX `batches_billing_project_state` ON `batches` (`billing_project`, `state`);
```
danking, line 1: I don't see any joins on batches that use this index, am I missing something?

batch/sql/create-billing-projects-test-limits.sql
```diff
@@ -0,0 +1,11 @@
...
+VALUES ('test-zero-limit', 0);
+
+INSERT INTO `billing_projects` (`name`, `limit`)
+VALUES ('test-tiny-limit', 0.00001);
+
+INSERT INTO `billing_project_users` (`billing_project`, `user`)
+VALUES ('test-zero-limit', 'test');
+
+INSERT INTO `billing_project_users` (`billing_project`, `user`)
+VALUES ('test-tiny-limit', 'test');
```
danking, line 11: These should go in estimated current right?

danking (dismissed):
We can write some automated tests for this, right? We'd need the batch client to expose things like billing limits. Then we could have a test that creates two batches, gets their costs, and then the cost of their billing project and asserts that the billing project's cost is the sum of those batches.

batch/sql/add-aggregated-billing-project-resources.sql
jigold: It's for the kill switch. See part 2.

jigold:
Stacked on #9375 as well.

danking (changes requested):
Some questions

danking:
Re-assigning to cotton for next week while I am away.

batch/batch/batch.py
```diff
@@ -229,7 +229,7 @@ def job_record_to_dict(record, name):
     result['msec_mcpu'] = msec_mcpu
 
     cost = format_version.cost(record['msec_mcpu'], record['cost'])
-    result['cost'] = cost_str(cost)
+    result['cost'] = cost
```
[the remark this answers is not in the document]
jigold: For the tests, I needed the float rather than the string.

batch/batch/front_end/front_end.py
```diff
@@ -1160,6 +1167,71 @@ async def ui_get_job(request, userdata):
...
+  ON resources.resource = aggregated_billing_project_resources.resource
+GROUP BY name, msec_mcpu, `limit`;
+'''
+
+    def record_to_dict(record):
+        cost_msec_mcpu = cost_from_msec_mcpu(record['msec_mcpu'])
+        cost_resources = record['cost']
+        record['accrued_cost'] = coalesce(cost_msec_mcpu, 0) + coalesce(cost_resources, 0)
+        del record['msec_mcpu']
+        return record
```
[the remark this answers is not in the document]
jigold: Not sure where you mean it's duplicated. I removed the other function that was doing the exact same thing now.

jigold:
Let's get a final approval on the design doc before merging this.

batch/batch/front_end/front_end.py
```diff
@@ -1177,6 +1184,54 @@ async def ui_get_job(request, userdata):
...
+        if limit == 'None':
+            limit = None
+        else:
+            limit = float(limit)
+            assert limit >= 0
+
+        await db.execute_update('UPDATE billing_projects SET `limit` = %s WHERE name = %s;', (limit, billing_project))
+        set_message(session, f'Modified limit {limit} for billing project {billing_project}.', 'info')
+    except Exception:
+        set_message(session, f'Invalid value {limit} for limit for billing_project {billing_project}', 'error')
```
danking, line 1230: This will also hide any database error as an invalid value message. I prefer:
```
try:
    ...
except Exception:
    set_message(session, f'Invalid ...')
else:
    await db.execute_update(...)
    set_message(session, f'Modified ...')
```
Unless there's a database error that we expect to get and want to treat that as an invalid value?

batch/batch/front_end/templates/billing_limits.html
```diff
@@ -0,0 +1,36 @@
...
+        <tr>
+            <th>Billing Project</th>
+            <th>Accrued Cost</th>
+            <th>Limit</th>
+            <th></th>
+        </tr>
+        </thead>
+        <tbody>
+        {% for row in billing_limits %}
+        <form action="{{ base_path }}/billing_limits/{{ row['billing_project'] }}/edit" method="POST">
```
danking, line 17: I'm pretty sure this is not valid HTML. Forms shouldn't be children of tables. The fix seems easy here: move the form and the hidden input all inside the `td` that contains the limit

danking: Yeah, I tried this:
```
<!DOCTYPE html>
<html lang="en">
<head><title>test</title></head>
<body>
<table>
<tbody>
<form>
<input type="hidden" name="_csrf" value="123">
<tr>
  <td>
    <input type="text" required name="limit" value="1.0">
  </td>
</tr>
<tr>
  <td>
    <input type="text" required name="limit" value="2.0">
  </td>
</tr>
</input>
</form>
</tbody>
</table>
</body>
</html>
```

in the HTML validator: https://validator.w3.org/nu/#textarea

danking: This validated:

```
<!DOCTYPE html>
<html lang="en">
<head><title>test</title></head>
<body>
<table>
<tbody>
<tr>
  <td>
    <form>
    <input type="hidden" name="_csrf" value="123">
    <input type="text" required name="limit" value="1.0">
    </form>
  </td>
</tr>
<tr>
  <td>
    <form>
    <input type="hidden" name="_csrf" value="123">
    <input type="text" required name="limit" value="2.0">
    </form>
  </td>
</tr>
</tbody>
</table>
</body>
</html>
```

web_common/web_common/templates/header.html
```diff
@@ -22,8 +22,13 @@
 	<div class="batch-caret header-dropdown-menu-caret"></div>
 	{% if userdata['is_developer'] == 1 %}
 	  <a class="header-dropdown-menu-link" href="{{ batch_driver_base_url }}">Driver</a>
-          <a class="header-dropdown-menu-link" href="{{ batch_base_url }}/billing">Billing</a>
 	  <a class="header-dropdown-menu-link" href="{{ batch_base_url }}/billing_projects">Billing Projects</a>
+        {% endif %}
```
danking, line 26: I think this endif is matched to the is_developer one, so we should align it the same

web_common/web_common/templates/header.html
```diff
@@ -22,8 +22,13 @@
...
+
+        <a class="header-dropdown-menu-link" href="{{ batch_base_url}}/billing_limits">Billing Limits</a>	  	  
+
+	{% if userdata['is_developer'] == 1 %}
+          <a class="header-dropdown-menu-link" href="{{ batch_base_url }}/billing">Billing</a>
```
danking, line 31: I'm not sure why this was two spaces further in above. All the anchors should be at the same level.

danking (changes requested):
Just a few more cosmetic things to clean up!

jigold:
I need to rethink the test-tiny-limit and test-zero-limit. This is going to fail every time after the first merge on the deployment.

jigold:
Ok. The limits for the tests will need to be fixed with a REST API for editing a billing limit and a new build step that is `setup_test_batch`. I'll work on that now. But I think this can go in while I'm working on that, but #9355 should depend on the new PR.

jigold:
Nevermind. Added tests for editing the limits to this PR. Will figure out the CLI after @catoverdrive has had a chance to do what they wanted to do with that for billing projects.

jigold:
There's no way to fix this problem other than not run those particular tests on the production server. I think that's okay.

jigold:
New plan. @catoverdrive already figured out how to have multiple users with different clients to be able to test the dev only endpoints such as editing a billing limit. I'm going to wait for their PR to go in and then reevaluate (#9553).

danking (dismissed):
moving off my scorecard; dismiss when you're ready

jigold:
Closing while I dev deploy.

danking (approved):

PR #9358, opened 2020-08-27
[hail] use rename in dndarray

danking:
When I first dndarray, I did not know you could use rename for keys. This
change takes advantage of rename instead of using the `r_field` and `c_field`
indirection.


johnc1231:
Did you mean to assign this?

danking:
Yep.

PR #9359, opened 2020-08-27
[Benchmark] Enable benchmark to read from multiple buckets

Dania-Abuhijleh:


benchmark-service/benchmark/benchmark.py
```diff
@@ -23,7 +23,7 @@
 
 
 def get_benchmarks(file_path):
-    read_gs = ReadGoogleStorage()
+    read_gs = ReadGoogleStorage(file='/benchmark-gsa-key/broad-ctsa-7ac34ab517ba.json')
```
jigold, line 26: Let's modify the kubernetes secret so the secret file name is `key.json`

benchmark-service/benchmark/utils.py
```diff
@@ -26,8 +26,8 @@ def enumerate_list_index(list_of_lists):
 
 
 class ReadGoogleStorage:
-    def __init__(self):
-        self.storage_client = storage.Client()
+    def __init__(self, file=None):
```
jigold, line 29: Let's call this `service_account_key_file`

jigold:
I approved this, but I just realized that we should take off the IAM permissions for the Broad-ctsa project and just have the specific permissions on the bucket. Can you see if that still works? We probably want the same thing for hail-vdc as well where we just have the fine-grained permissions on those 2 buckets.

PR #9361, opened 2020-08-27
[Benchmark] Global state for gs reader

Dania-Abuhijleh:


jigold:
Cool. You also need to rebase on the latest version of main and fix the merge conflicts. Let me know if you don't know how to do that.

benchmark-service/benchmark/benchmark.py
```diff
@@ -180,30 +183,35 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
+               'comparisons': comparisons,
+               'benchmark_file_list': app['benchmark_file_list']}
     return await render_template('benchmark', request, userdata, 'compare.html', context)
 
 
-def init_app() -> web.Application:
+def on_startup(app):
+    app['gs_reader'] = ReadGoogleStorage(service_account_key_file='/benchmark-gsa-key/key.json')
+    files = app['gs_reader'].list_files()
+    app['benchmark_file_list'] = files
```
jigold, line 202: You don't need the list files here. We talked about this and decided we should just do the query on demand rather than caching the results.

benchmark-service/benchmark/utils.py
```diff
@@ -47,3 +47,15 @@ def get_data_as_string(self, file_path):
         except Exception:
             raise NameError()
         return data
+
+    def list_files_in_bucket(self, bucket_name):
```
jigold, line 51: let's call this list_files

benchmark-service/benchmark/utils.py
```diff
@@ -47,3 +47,15 @@ def get_data_as_string(self, file_path):
...
+        list_of_files = []
+        bucket = self.storage_client.get_bucket(bucket_name)
+        for blob in bucket.list_blobs():
+            list_of_files.append('gs://' + bucket_name + '/' + blob.name)
+        return list_of_files
+
+    def list_files(self):
```
jigold, line 58: Let's move this function out of the GoogleStorageReader class and it should be it's own utils function. We should have a list of valid benchmarks buckets that's a set of constants. You could either make a new file for that called `constants.py` or put it at the top of the `utils.py` file as `BENCHMARK_BUCKETS  = ['hail-benchmarks', 'hail-benchmarks-2']`. This function should call `list_files` for both of the buckets. So something like this:

```
def list_benchmark_files():
    list_of_files = []
    for bucket in BENCHMARK_BUCKETS:
       ...
```

benchmark-service/benchmark/utils.py
```diff
@@ -48,14 +57,11 @@ def get_data_as_string(self, file_path):
...
         bucket = self.storage_client.get_bucket(bucket_name)
         for blob in bucket.list_blobs():
             list_of_files.append('gs://' + bucket_name + '/' + blob.name)
         return list_of_files
 
-    def list_files(self):
-        list_of_files = []
-        list_of_files.extend(self.list_files_in_bucket('hail-benchmarks'))
-        return list_of_files
+
```
jigold, line 67: delete this extra line.

benchmark-service/benchmark/utils.py
```diff
@@ -31,6 +32,14 @@ def enumerate_list_of_trials(list_of_trials):
     return res_dict
 
 
+def list_benchmark_files(read_gs):
+    list_of_files = []
+    #list_of_files.extend(self.list_files_in_bucket('hail-benchmarks'))
```
jigold, line 37: delete.

PR #9371, opened 2020-08-28
[query] Switch _buildEncoder methods to use CodeBuilder infrastructure

chrisvittal:
This streamlines the logic significantly. No attempt has been made to have the encoder methods take
PCode/EmitCode directly, but that is the future direction here.

### DIFF
--- a/hail/src/main/scala/is/hail/types/encoded/ETransposedArrayOfStructs.scala
+++ b/hail/src/main/scala/is/hail/types/encoded/ETransposedArrayOfStructs.scala
@@ -2,7 +2,7 @@ package is.hail.types.encoded
 
 import is.hail.annotations.{Region, UnsafeUtils}
 import is.hail.asm4s._
-import is.hail.expr.ir.{EmitMethodBuilder, ParamType}
+import is.hail.expr.ir.{EmitCodeBuilder, EmitMethodBuilder, ParamType}
 import is.hail.types.{BaseStruct, BaseType}
 import is.hail.types.physical._
 import is.hail.types.virtual._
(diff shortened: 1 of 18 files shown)


chrisvittal:
Ok, hopefully ironed out the bugs.

chrisvittal:
@tpoterba The tests now all pass. I'd appreciate your input on 32965c8 , the last commit in this sequence. It shortens the names of methods, possibly truncating them more than we already do, but I had found that something was blowing memory related to large classes/constants and once the types are much larger than a few KB, they're not really that useful.

Thanks!

tpoterba:
yes, truncating more is a great idea.

chrisvittal:
This is ready for review or merging.

hail/src/main/scala/is/hail/types/encoded/ETransposedArrayOfStructs.scala
```diff
@@ -212,76 +212,55 @@ final case class ETransposedArrayOfStructs(
...
-
-      val encoders = fieldGroup.map { encodedField =>
-        val fidx = elementPStruct.fieldIdx(encodedField.name)
-        val pf = elementPStruct.fields(fidx)
-        val encodeField = encodedField.typ.buildEncoder(pf.typ, groupMB.ecb)
-        val elem = () => arrayPType.elementOffset(addr, len, j)
+      if (!ef.typ.required) {
+        cb.forLoop(cb.assign(j, 0), j < len, cb.assign(j, j + 1), {
+          array.loadElement(cb, j).consume(cb, { /* do nothing */ }, { pbsc =>
+            // FIXME may be bad if structs get memoized into their fields, otherwise
```
tpoterba, line 241: I think this should probably be fine, good to note though.

PR #9373, opened 2020-08-28
Add BlockMatrixFilterRowsRDD to decrease transposing

johnc1231:
The current `filter_rows` is implemented as a transpose. Nik and Konrad are running into driver memory issues when doing `tree_matmul`, and one theory of mine is that making lots of `BlockMatrixTransposeRDD`'s is contributing to this.  I'd like to make this change to avoid that in the future.


johnc1231:
Assigning @tpoterba since this is quite similar to #9374 that you reviewed. It's basically just a copy paste of the column oriented one, but adjusted to do rows instead.

PR #9375, opened 2020-08-28
[batch] Add testing for invariants

jigold:
- New job only runs after test_batch because I don't want to spin up everything when running this with dev deploy
- Only run in test and dev (not deploy)


batch/batch/driver/main.py
```diff
@@ -679,8 +692,12 @@ async def on_startup(app):
     await scheduler.async_init()
     app['scheduler'] = scheduler
 
-    # asyncio.ensure_future(check_incremental_loop(db))
-    # asyncio.ensure_future(check_resource_aggregation(db))
+    app['check_incremental'] = None
+    app['check_resource_aggregation'] = None
```
danking, line 696: I think it would be more clear if these were named something like `check_incremental_error` and `check_resource_aggregation_error`

build.yaml
```diff
@@ -2336,3 +2336,38 @@ steps:
...
+    - dev
+   dependsOn:
+    - create_deploy_config
+    - create_accounts
+    - default_ns
+    - batch_pods_ns
+    - copy_files
+    - batch_image
+    - deploy_batch
+    - test_batch
```
danking, line 2373: Probably should depend on test_hailtop_batch and the batch docs tests. We should update `delete_batch_instances` to depend on this job since this is the new sink of batch tests.

batch/test/test_invariants.py
```diff
@@ -0,0 +1,36 @@
...
+        async def wait_forever():
+            data = None
+            while data is None:
+                resp = await utils.request_retry_transient_errors(
+                    session, 'GET', url, headers=headers)
+                data = await resp.json()
+                await asyncio.sleep(5)
+            return data
+
+        data = await asyncio.wait_for(wait_forever(), timeout=30 * 60)
```
danking, line 34: the CI job has 5 minute timeout, so I don't feel like we need (or could ever hit) this timeout.

batch/test/test_invariants.py
```diff
@@ -0,0 +1,36 @@
...
+        assert data['check_incremental'] is None, str(data)
+        assert data['check_resource_aggregation'] is None, str(data)
```
danking, line 36: I don't think you need to `str` here:
```
In [1]: assert False, {'hello': 'jackie goldstein'}                                                                                                                                                                                                                                                                           
  File "<ipython-input-1-27688bb9f4ad>", line 1, in <module>
    assert False, {'hello': 'jackie goldstein'}
AssertionError: {'hello': 'jackie goldstein'}

```

batch/test/test_invariants.py
```diff
@@ -0,0 +1,36 @@
...
+            timeout=aiohttp.ClientTimeout(total=60)) as session:
+
+        async def wait_forever():
+            data = None
+            while data is None:
+                resp = await utils.request_retry_transient_errors(
+                    session, 'GET', url, headers=headers)
+                data = await resp.json()
+                await asyncio.sleep(5)
+            return data
```
danking, line 32: It looks like we always get a non-None response from the server, so would we ever go through this loop more than once?

build.yaml
jigold: I disagree. I want to be able to run this with dev deploy without spinning up all the infrastructure for ci and running test_hailtop_batch. The batch docs tests run everything locally I believe. The hailtop batch tests do run things on the server.

jigold: I'll make two jobs. One is runIfRequested only that depends on test_batch. The other is the one for test and deploy.

jigold: ok. I decided you're right and if I am dev deploying, then I'll comment out dependencies by hand.

build.yaml
```diff
@@ -2896,3 +2858,88 @@ steps:
...
+        valueFrom: batch_pods_ns.name
+      mountPath: /test-gsa-key
+   scopes:
+    - dev
+    - test
+   dependsOn:
+    - default_ns
+    - batch_pods_ns
+    - base_image
+    - test_batch_invariants
```
jigold, line 2933: What happens if you depend on a step that isn't in the deploy scope? Are you just ignored?

build.yaml
```diff
@@ -2896,3 +2858,88 @@ steps:
...
+    - name: test-dev-tokens
+      namespace:
+        valueFrom: batch_pods_ns.name
+      mountPath: /user-tokens
+    - name: ssl-config-batch-tests
+      namespace:
+        valueFrom: default_ns.name
+      mountPath: /ssl-config
+   scopes:
+    - test
```
jigold, line 2885: Actually, we probably could run this in deploy mode. The errors will always be None though.

build.yaml
danking: Yeah the `job` is `None` so `wrapped_job` returns the empty list, so it is effectively not a dependency for its dependents.

batch/deployment.yaml
```diff
@@ -60,6 +60,10 @@ spec:
          - name: HAIL_SHOULD_PROFILE
            value: "1"
 {% endif %}
+{% if scope == "test" %}
```
danking, line 63: Should we check this in dev deploys too?

build.yaml
danking: I guess I don't really understand why delete_batch_instances depends on the tests. I anticipated it to delete the instances *before* deploying batch. I tried using it like: `--steps delete_batch_instances,test_batch_...` and was surprised that the instances weren't deleted first.

danking (changes requested):
I think the scope for check should probably include dev. thoughts?

build.yaml
jigold: No. The point of delete_batch_instances is to delete instances that are hanging around after the tests have been run once the driver is killed. For example, a worker in an infinite loop.

danking:
We were worried about the speed impact of this, right? Let's revisit after the other PRs regarding speed land.

danking (changes requested):
see comment above

jigold:
I think this can go in now. Not sure what comment you are referring to.

danking (approved):

PR #9376, opened 2020-08-31
Run hail Java tests with checked memory allocator

johnc1231:
We want to eventually run all of our tests with the checked memory allocator, plus a few representative tests with the regular allocator. This PR switches the Java tests to the checked memory allocator, while the python tests remain on the normal allocator. I'll follow up to deal with the Python tests, but wanted to get a start in here to at least make sure some CI tests are exercising each allocator.


tpoterba (approved):
nice!

PR #9382, opened 2020-08-31
[graphics] add svgs and 32x32 icons to repository

catoverdrive:
I'm not sure where these should live, but I wanted to move them off my laptop and into a place where people can access them, and the repo seems as good of a place as any.

There are svgs for a bunch of icons in both blue and white, as well as high-ish resolution images of both versions of the logo and 32x32 icon pngs.

I haven't started integrating them into website stuff yet, but I figured that raw images should have a central-ish place to live anyways.

All of the images in the PR are as below:
all


danking:
I spot checked, and noticed that [`log-b.png`](https://github.com/hail-is/hail/blob/aa3cd5c4a64247a550e9fd61e79c0b7de7713559/graphics/32x32/log-b.png) looks like just blue?

catoverdrive:
Bah, fixed. There was some weird grouping going on in the svgs that was causing the PNG conversion to do something weird, so I simplified all the svgs.

danking (approved):

PR #9388, opened 2020-09-01
[query] encode allocation behavior in stream PTypes

patrick-schultz:
* Add a `separateRegions` flag to `PStream`.
* Add a `separateRegions` flag to the stream producer nodes `MakeStream`, `StreamRange`, and `ToStream`.
* Propogate `separateRegions` through all stream nodes in `InferPType`.

The intended semantics is that if a stream's type has the `separateRegions` flag set, its consumer must pass it a region which gets cleared every element. If the flag is not set, there is no requirement; depending on context, the consumer may put every element in the same region, but is also allowed to use separate regions for each element. For example, in a zip, the elements of the zipped stream are put in separate regions iff at least one child stream requires separate regions; in that case, all children will get emitted with separate regions, whether they requested it or not.

In this PR, the `separateRegions` flags are left unused. Eventually, stream consumers will inspect the flag on their child streams' types, and use that information to construct the appropriate `StagedRegion` to pass to `emitStream`. In implementing that, I did some refactoring of the `StagedRegion` design, which I separated out into a follow-up PR.

### DIFF
--- a/hail/python/hail/ir/ir.py
+++ b/hail/python/hail/ir/ir.py
@@ -602,16 +602,20 @@ class ArrayZeros(IR):
 
 
 class StreamRange(IR):
-    @typecheck_method(start=IR, stop=IR, step=IR)
-    def __init__(self, start, stop, step):
+    @typecheck_method(start=IR, stop=IR, step=IR, separate_regions=bool)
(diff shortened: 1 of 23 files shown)


patrick-schultz:
Sorry, forgot to mirror the changes in the python IR.

PR #9391, opened 2020-09-01
[internal-gateway] permit TLS connections to internal gateway

danking:
The plan for TLS workers is:
- internal-gateway supports HTTP and HTTPS
- workers speak server-auth-only TLS
- internal-gateway supports only HTTPS
- workers proffer client auth
- internal-gateway requires client auth

This is step 1.


danking:
It's already deployed, so, I guess that's how I tested it.

danking:
I might need to make more changes as I realize exactly what needs to be done

PR #9393, opened 2020-09-01
[Benchmark] add autocomplete

Dania-Abuhijleh:
Screen Shot 2020-09-10 at 1 59 44 PM

Screen Shot 2020-09-08 at 9 11 40 AM


benchmark-service/benchmark/benchmark.py
```diff
@@ -172,6 +175,7 @@ async def index(request, userdata):  # pylint: disable=unused-argument
     return await render_template('benchmark', request, userdata, 'index.html', context)
 
 
+
```
jigold, line 178: delete extra new line

benchmark-service/benchmark/templates/index.html
```diff
@@ -2,11 +2,15 @@
...
 
 {% block content %}
+<link rel="stylesheet" href="{{ base_path }}/static/css/autocomplete.css" />
+
 <h1>Benchmark System</h1>
 
-<form action="{{ base_path }}/" method="get">
-    <input type="text" name="file"></input>
-    <input type="submit" value="lookup"></input>
+<form autocomplete="off" action="{{ base_path }}/" method="get">
```
jigold, line 9: Can you confirm whether this makes a new GET request each time you type a character? We don't want that behavior.

benchmark-service/benchmark/static/css/autocomplete.css
```diff
@@ -0,0 +1,192 @@
...
+    &:focus {
+	outline: 0;
+    }
+    &:active,
+    &:hover {
+	outline: 0;
+        color: $devil-gray;
+	border-color: $devil-gray;
+    }
+}
```
jigold, line 192: Add new line

benchmark-service/benchmark/benchmark.py
```diff
@@ -158,8 +161,9 @@ async def show_name(request: web.Request, userdata) -> web.Response:  # pylint:
 
 @router.get('/')
 @router.get('')
-@web_authenticated_developers_only(redirect=False)
-async def index(request, userdata):  # pylint: disable=unused-argument
+#@web_authenticated_developers_only(redirect=False)
```
jigold, line 164: Can you uncomment this back?

benchmark-service/benchmark/templates/index.html
jigold: ???

jigold (changes requested):
Can you make the search results wider so it fits on one line?

jigold:
@danking Do you have anything else to add about the CSS?

benchmark-service/benchmark/templates/index.html
Dania-Abuhijleh: Wasn't sure if I was checking it correctly yesterday, but the browser shows GET requests being made only when you press the lookup button, not while typing

benchmark-service/benchmark/static/css/autocomplete.css
```diff
@@ -0,0 +1,189 @@
...
+    color:#7b7b7b;
+    transition:all .1s ease-in-out;
+    background-color:#fff;
+}
+.autoComplete_result::selection {
+    color:rgba(#fff,0);
+    background-color:rgba(#fff,0);
+}
+.autoComplete_result:last-child {
+    border-radius:0 0 1rem 1rem;
```
danking, line 97: I do like the rounded corners, but they feel aggressively big! What do you think of 4px instead?

Screen Shot 2020-09-03 at 12 02 30 PM

benchmark-service/benchmark/static/css/autocomplete.css
```diff
@@ -0,0 +1,189 @@
+#autoComplete_wrapper {
```
danking, line 1: You've probably noticed that the list has a little unnecessary space on the left hand side. This is because `ul` elements, by default, add a little bit of left padding. This creates space for the bullets. Since we're not using bullets (you have list-style:none on the list items (li's)), lets' nix the left padding with:
```
ul.file_list {
    padding-left: 0;
}
```

benchmark-service/benchmark/static/css/autocomplete.css
```diff
@@ -0,0 +1,189 @@
...
+    list-style:none;
+    text-align:left;
+    font-size:1.1rem;
+    color:#7b7b7b;
+    transition:all .1s ease-in-out;
+    background-color:#fff;
+}
+.autoComplete_result::selection {
+    color:rgba(#fff,0);
+    background-color:rgba(#fff,0);
```
danking, line 94: I don't think these are valid CSS rules. I believe `rgba` requires decimal numbers, not hexadecimal ones. If you want white, you can just write: `color: #fff`.

benchmark-service/benchmark/static/css/autocomplete.css
```diff
@@ -0,0 +1,189 @@
...
+    border-top:2px solid transparent;
+    border-bottom:2px solid transparent;
+}
+.autoComplete_result:focus {
+    outline:0;
+    background-color:rgba(192,192,192,0.3);
+    border-left:2px solid rgba(192,192,192,1);
+    border-right:2px solid rgba(192,192,192,1);
+    border-top:2px solid transparent;
+    border-bottom:2px solid transparent;
```
danking, line 113: I think I prefer no top or bottom border. If you remove those, then the height of the element is constant. If you keep them, the list shimmies up and down as you scroll through it. I find that kind of confusing.

benchmark-service/benchmark/static/css/autocomplete.css
```diff
@@ -0,0 +1,189 @@
...
+    width:16.5rem;
+    background-size:1.7rem;
+    background-position:left 1.05rem top .7rem;transition:all .3s ease;
+    -webkit-transition:all -webkit-transform .3s ease;
+}
+#autoComplete:focus {
+    padding:0 2rem 0 3.5rem;
+    color:rgba(192,192,192,1);
+    height:3rem;
+    width:16.5rem;
```
danking, line 60: I find the width rules kind of annoying, especially how it changes width. Let's remove width rules from #autoComplete:focus and #autoComplete:hover and change #autoComplete width to `100%`. 
Screen Shot 2020-09-03 at 12 17 27 PM

benchmark-service/benchmark/static/css/autocomplete.css
```diff
@@ -0,0 +1,189 @@
...
+}
+#autoComplete::selection {
+    color:transparent;
+}
+#autoComplete:focus::selection {
+    color:#fff;
+    background-color:rgba(192,192,192,1);
+}
+#autoComplete:hover {
+    padding:0 2rem 0 3.5rem;
```
danking, line 48: I find this padding a bit roomy! What do you think of just `padding-left: 0.5em;`?
Screen Shot 2020-09-03 at 12 18 49 PM

benchmark-service/benchmark/static/css/autocomplete.css
```diff
@@ -0,0 +1,189 @@
...
+    color:rgba(192,192,192,1);
```
danking, line 49: I find this light grey a bit hard to read, especially for the text that I'm actively typing. It's also a bit weird to have it change color to something lighter and less striking when I focus it (i.e. when I start typing). What do you think of removing this rule from the hover and focus rules?
Screen Shot 2020-09-03 at 12 20 15 PM

danking: I kind of prefer the results also being black, but I'm less committed to that idea.

danking: In particular black+bold for the highlighted bit is a bit much! So the grey maybe is the right choice.

benchmark-service/benchmark/static/css/autocomplete.css
```diff
@@ -0,0 +1,189 @@
...
+    background-position:center;
+}
+#autoComplete:hover {
+    border-width:1px;
+}
+#autoComplete:focus {
+    border-width:1px;
+    background-position:left 1.1rem top .8rem;
+}
+#button {
```
danking, line 171: I think for the visual consistency of the Hail brand, all buttons should look the same. I'm open to new button proposals, but let's do that to every service or no service. If you have an idea for an alternative button design, let's make a doc with some pictures and chat about it at design meeting!

danking (changes requested):
This is extremely awesome! I love the auto-complete. I left some comments about how we might streamline the CSS and also design critique on the visual aesthetics.

jigold:
Let's not ignore this PR while we're researching the next steps in the project.

jigold (dismissed):
Can you confirm whether you removed the SCSS and replaced it with CSS?

benchmark-service/benchmark/static/css/autocomplete.css
```diff
@@ -0,0 +1,127 @@
+ul {
```
jigold, line 1: Should this rule be specific to autoComplete? Is there a way to do that?

benchmark-service/benchmark/templates/index.html
```diff
@@ -2,11 +2,15 @@
...
+    <div class="autocomplete">
+        <input type="text" name="file" id="autoComplete"></input>
```
jigold, line 11: I see the code below sets the default that's listed as gs://..., but your image is empty for the lookup without any text in it. Is there a way to get gs://... to be the placeholder?

jigold (changes requested):
Almost there :)

benchmark-service/benchmark/templates/index.html
Dania-Abuhijleh: it is there but only when hovering or clicking on the box, I tried to get it to show up normally also but it's not working..

Dania-Abuhijleh: working now

jigold (approved):

PR #9394, opened 2020-09-02
[build.yaml] consolidate check build steps

cseed:
In particular, check hailtop, which appears to have been missing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ check-batch:
 
 .PHONY: check-benchmark
 check-benchmark:
-	make -C benchmark/python check
+	make -C benchmark-service check
 
 .PHONY: check-ci
 check-ci:
--- a/Makefile
+++ b/Makefile
@@ -5,13 +5,16 @@ default:
 	     build hail see: https://hail.is/docs/0.2/
 	exit 1
 
+.PHONY: check-all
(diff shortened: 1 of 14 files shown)


cseed:
There are now three check steps:
 - check_hail (in the sense of $HAIL_HOME/hail directory), which checks the hail and hailtop packages, using the base image, from the source via `make check-hail`
 - check_services, using the services image, except the benchmark service, which installs addition dependencies, from the source via `make -k check-services`, `-k` forces make to check all packages, even if one fails
 - check_benchmark_service, which checks benchmark as installed in that image

cseed:
FYI @danking There was also a Makefile bug: make creates make variables from the environment variables by default.  Also, when make runs a rule, it makes the make variables available in the environment.  That means PYTHONPATH in those Makefiles was getting set to some literal ${...} string, not the right value.  I solved this by not assigning PYTHONPATH in the Makefile.

https://www.gnu.org/software/make/manual/html_node/Environment.html

"Every environment variable that make sees when it starts up is transformed into a make variable with the same name and value. However, an explicit assignment in the makefile, or with a command argument, overrides the environment."

"When make runs a recipe, variables defined in the makefile are placed into the environment of each shell."

PR #9397, opened 2020-09-02
[hailtop/utils] check with mypy

cseed:
One fix.  Also, prefer global `--ignore-missing-imports` to local hints.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -931,6 +931,9 @@ steps:
 
      python3 -m flake8 --config setup.cfg $SITE_PACKAGES/hail $SITE_PACKAGES/hailtop || exit_status=$?
      python3 -m pylint --rcfile pylintrc hailtop || exit_status=$?
+     # debugging
+     pwd
+     echo $SITE_PACKAGES
      mypy --ignore-missing-imports $SITE_PACKAGES/hailtop/utils $SITE_PACKAGES/hailtop/batch $SITE_PACKAGES/hailtop/google_storage.py || exit_status=$?
 
      exit $exit_status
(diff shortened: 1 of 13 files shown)


danking (dismissed):
`mypy` will read `setup.cfg`, so lets add
```
[mypy]
ignore_missing_imports = True
```
there.

https://mypy.readthedocs.io/en/stable/config_file.html

cseed:
Reassigning since Dan is out.

build.yaml
```diff
@@ -931,7 +931,7 @@ steps:
 
      python3 -m flake8 --config setup.cfg $SITE_PACKAGES/hail $SITE_PACKAGES/hailtop || exit_status=$?
      python3 -m pylint --rcfile pylintrc hailtop || exit_status=$?
-     (cd $SITE_PACKAGES/hailtop && mypy batch google_storage.py) || exit_status=$?
+     (cd $SITE_PACKAGES/hailtop/ && mypy --config-file /setup.cfg utils batch google_storage.py) || exit_status=$?
```
jigold, line 934: Does this one need `--ignore-missing-imports`?

jigold (dismissed):

build.yaml
cseed: Oops.  No, I removed it from the other step.  We shouldn't specify the option on teh command line because we're setting it in the setup.cfg now.

jigold: Did you forget a commit? It looks like one step still has the global flag.

cseed: Fixed.

jigold (approved):

PR #7792, opened 2020-01-02
[hail] include Python stack trace in ArrayRef out-of-bounds error

tpoterba:
Fixes #2921

Example:

```
In [3]: mt = mt.annotate_entries(PL=[0])

In [4]: hl.split_multi_hts(mt)._force_count_rows()

... hundreds of lines of java trace ...

Hail version: 0.2.30-bae67f384161
Error summary: HailException: array index out of bounds: index=1, length=1
----------
Python traceback:
  File "<ipython-input-4-ba11ec1cd68e>", line 1, in <module>
    hl.split_multi_hts(mt)._force_count_rows()

  File "/Users/tpoterba/hail/hail/python/hail/methods/statgen.py", line 2246, in split_multi_hts
    (hl.range(0, 3).map(lambda i:

  File "/Users/tpoterba/hail/hail/python/hail/methods/statgen.py", line 2250, in <lambda>
    ).map(lambda j: split.PL[j]))))))

  File "/Users/tpoterba/hail/hail/python/hail/methods/statgen.py", line 2250, in <lambda>
    ).map(lambda j: split.PL[j]))))))
```

PR #9398, opened 2020-09-02
Better Python Errors

johnc1231:
This PR is based on discussion here: https://dev.hail.is/t/better-python-error-messages-idea/201/9 The intention is to create a system to give better error messages from python in a generic way. Tim's work in #7792 does a good job introducing behavior like this this specifically for `ArrayRef` nodes, but I want to add three things on top of that:

1. I don't want to have to do as much custom per IR node work
2. I don't want to send the entire python stack trace over py4j to scala for every node
3. I don't want the user to see a Java stack trace in this scenarios.

This first PR is a proof of concept that adds this behavior for the `Die` node, which will catch any errors generated by uses of `CaseBuilder.or_error`. Follow up PRs should change `ArrayRef` to work this way, as well as catch things like looking up a key in a dictionary but not finding it. In an ideal future, we'd bolt on some extra mechanism to give types to these errors, and we could throw a proper `IndexError` in the `ArrayRef` case or `KeyError` in the dictionary case. 

It feels a little bit messy right now, open to suggestions. I don't love using `-1` as the "no error" situation, but I thought it was probably easier than dealing with optionals between python and scala. 

To give an example of what it looks like, the error message for this script:

```
import hail as hl

ht = hl.utils.range_table(10)
ht = ht.annotate(foo = hl.nd.array([[1], [2], [3]]))
ht = ht.annotate(bar = ht.foo[0:4, 12])
ht.collect()
```

is

```
  File "better_error_test.py", line 6, in <module>
    ht.collect()
  File "<decorator-gen-1103>", line 2, in collect
  File "/Users/johnc/Code/hail/hail/python/hail/typecheck/check.py", line 614, in wrapper
    return __original_func(*args_, **kwargs_)
  File "/Users/johnc/Code/hail/hail/python/hail/table.py", line 1903, in collect
    return Env.backend().execute(e._ir)
  File "/Users/johnc/Code/hail/hail/python/hail/backend/spark_backend.py", line 325, in execute
    raise HailUserError(message_and_trace) from None
hail.utils.java.HailUserError: Error summary: HailException: Index 12 is out of bounds for axis 1 with size 1
------------
Hail stack trace:
  File "better_error_test.py", line 5, in <module>
    ht = ht.annotate(bar = ht.foo[0:4, 12])

  File "<decorator-gen-707>", line 2, in __getitem__

  File "/Users/johnc/Code/hail/hail/python/hail/expr/expressions/typed_expressions.py", line 3709, in __getitem__
    hl.str("Index ") + hl.str(s) + hl.str(f" is out of bounds for axis {i} with size ") + hl.str(dlen)

  File "<decorator-gen-1299>", line 2, in or_error

  File "/Users/johnc/Code/hail/hail/python/hail/expr/builders.py", line 311, in or_error
    die_ir.save_error_info()
```

(Note that ndarray bounds checking internally uses a case builder, which is why this example works).

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -1686,11 +1686,7 @@ class Emit[C](
         val ev = mb.getEmitParam(2 + i)
         assert(ev.pt == expectedPType)
         ev
-      case x@Die(m, typ, errorId) =>
-        log.info("errorId of Die was " + errorId + s" IR was ${x}")
-        if (errorId == -1) {
-          log.info(x.stackTrace)
-        }
+      case @Die(m, typ, errorId) =>
         val cm = emit(m)
         EmitCode(
           Code(
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -559,11 +559,11 @@ class Emit[C](
 
         cb.assign(ib, Code._null)
 
-      case Die(m, typ) =>
+      case Die(m, typ, errorId) =>
         val cm = emitI(m)
         val msg = cb.newLocal[String]("exmsg", "<exception message missing>")
         cm.consume(cb, {}, s => cb.assign(msg, s.asString.loadString()))
-        cb._throw(Code.newInstance[HailException, String](msg))
+        cb._throw(Code.newInstance[HailException, String, Int](msg, errorId))
 
       case x@WriteMetadata(annotations, writer) =>
         writer.writeMetadata(emitI(annotations), cb, region.code)
@@ -1686,15 +1686,15 @@ class Emit[C](
         val ev = mb.getEmitParam(2 + i)
         assert(ev.pt == expectedPType)
         ev
-      case Die(m, typ) =>
+      case Die(m, typ, errorId) =>
         val cm = emit(m)
         EmitCode(
           Code(
             cm.setup,
-            Code._throw[HailException, Unit](Code.newInstance[HailException, String](
+            Code._throw[HailException, Unit](Code.newInstance[HailException, String, Int](
               cm.m.mux[String](
                 "<exception message missing>",
-                coerce[String](StringFunctions.wrapArg(EmitRegion(mb, region.code), m.pType)(cm.v)))))),
+                coerce[String](StringFunctions.wrapArg(EmitRegion(mb, region.code), m.pType)(cm.v))), errorId))),
           true,
           pt.defaultValue)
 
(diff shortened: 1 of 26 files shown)


johnc1231:
One other question: Is it a breaking change to change the type of error we throw?

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -1685,15 +1685,15 @@ class Emit[C](
         val ev = mb.getEmitParam(2 + i)
         assert(ev.pt == expectedPType)
         ev
-      case Die(m, typ) =>
+      case @Die(m, typ, errorId) =>
```
patrick-schultz, line 1688: Extraneous @

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -1508,7 +1508,7 @@ case class TableMapRows(child: TableIR, newRow: IR) extends TableIR {
         FastIndexedSeq(classInfo[Region], LongInfo, LongInfo), LongInfo,
         Coalesce(FastIndexedSeq(
           extracted.postAggIR,
-            Die("Internal error: TableMapRows: row expression missing", extracted.postAggIR.typ))))
+            Die("Internal error: TableMapRows: row expression missing", extracted.postAggIR.typ, -1))))
```
patrick-schultz, line 1511: You added the constructor that sets the errorId to -1 implicitly; did you not want to use that here (and several other places)? That would keep the -1 magic number hard coded in fewer places.

patrick-schultz (changes requested):
This is great. Overall design looks good, just a few minor things.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
johnc1231: Ah, I originally added the numeric argument everywhere manually because I didn't want to miss anywhere where I had to copy the argument along. I can go back now and remove the extras.

johnc1231:
Thanks for the review, all addressed now.

patrick-schultz (approved):

hail/src/main/scala/is/hail/expr/ir/IR.scala
```diff
@@ -1,5 +1,6 @@
 package is.hail.expr.ir
 
+import com.google.api.gax.rpc.InvalidArgumentException
```
tpoterba, line 3: delete this?

tpoterba (dismissed):

hail/src/main/scala/is/hail/expr/ir/IR.scala
johnc1231: Oh, that was a mistake where I typed "Invalid" instead of "Illegal", must have autoimported.

tpoterba: yeah, I do that all the time!

PR #9400, opened 2020-09-02
[query] memory management for StreamJoinRightDistinct

patrick-schultz:
I apparently missed this during my first pass of conversions. The PR to follow this one refines the `StagedRegion` class hierarchy to better work with allocation behaviors encoded in `PStream`. I caught this while making that change, because the tighter API makes it much harder to "cheat" like stream joins were doing.


tpoterba (approved):
looks good.

PR #9401, opened 2020-09-02
[query] enable memory managenment in lowered IR

patrick-schultz:
Stacked on #9400 

This PR modifies stream consumers' emit cases to look up `separateRegions` on the child stream's type, using that to construct the appropriate `StagedRegion`, then passing that region into `emitStream`. It also removes `EmitAllocationStrategy`, as this is now encoded in the IR. We may at some point want something similar to choose what allocation behavior to generate during lowering, but that is left for future work.

Getting this to work well required some changes to the `StagedRegion` design.

First, when a stream node needs to create a fresh region which it owns, the new region is now created as a child of the `outerRegion`, not the consumer's `eltRegion` as was done before. This simplifies the semantics of the parent/child relationship: now the lifetime of a child region is always nested inside that of its parent.

The second change also has to do with the tree structure of `StagedRegion`s. Before, any `StagedRegion` could have children, which meant we could have arbitrarily deep trees of `StagedRegion`s, and also that every `StagedRegion` had to know whether children would be genuine sub-regions or just aliases of the parent.

Now, the `StagedRegion` hierarchy is split into `ParentStagedRegion` and `ChildStagedRegion`. A `ParentStagedRegion` can have children, while a `ChildStagedRegion` cannot. The main effect of this is that the `allowAllocations` bit is no longer hereditary. A `ChildStagedRegion` can be converted into a `ParentStagedRegion`, and hence another generation of children can be created, but now doing so requires explicitly specifying `allowAllocations` for the new generation. In emit, this happens at stream consumers, where the allocation strategy will be read off the PType of the stream.

The full hierarchy is now:
* `StagedRegion` - The root class is the type for passing around unowned regions. Besides writing into the region, the method `asRoot` allows converting to a `RootStagedRegion` by specifying `allowAllocations`
  * `ParentStagedRegion` - for passing around un-owned regions which allow creating child regions. Has a `allowSubregions` flag which encodes whether child regions are actually separate run-time regions.
  * `ChildStagedRegion` - un-owned regions which know their parent, allowing methods like `copyToParent` and `createSiblingRegion`
    * `OwnedStagedRegion` - owned regions are always children; adds methods like `free`, `clear`, `giveToParent`
      * The concrete classes `RealOwnedStagedRegion` and `DummyOwnedStagedRegion`. These encode whether the parent region is real or dummy. They are intended to be private to the `StagedRegion` implementation.

### Update

At first, I restricted methods like `copyTo` and `giveTo` to copy/move to either a parent or sibling region. That is easy to verify correctness, but turned out to be too restrictive. I had been assuming an invariant that the `elementRegion` of a stream is always a direct child of the `outerRegion`. But this invariant is violated by some nested stream nodes like `StreamFlatMap` and `StreamGroupBy`.

Take `StreamFlatMap`: the `outerRegion` of the inner stream should be the `eltRegion` of the outer stream, which is created and owned by the FlatMap code. But the `eltRegion` of the final flatMapped stream is passed in from the consumer, so is not a child of the inner stream's `outerRegion`.

It is also possible for the `eltRegion` to be a descendant, but not child, of the `outerRegion`.

To fix this, I had to relax the assertion on `copyTo`, `giveTo`, etc. Now the precondition is that the destination region is a subregion of the parent of the source, captured by the `<=` method: `dest <= this.parent`. Dynamically, `r1 <= r2` should mean that the lifetime of `r1` is nested inside that of `r2`. If `copyTo` actually copies, it is always correct, so consider the case where the source is a `DummyOwnedStagedRegion`, i.e. an alias of its parent. Then `copyTo` generates no code, and the "copied" data actually just points to the original data. For this to be valid for the lifetime of the destination region, we need exactly `dest <= this.parent`.

Finally, to handle the `StreamFlatMap` case, we need a way to register a subregion relationship between two regions after both have been constructed. This is implemented by the `asSubregionOf` method, which adds to a list of `otherAncestors`, which are regions other than the parent which we assume have lifetimes containing that of `this` region.

I also added an assertion to verify that for every stream emitted, `eltRegion <= outerRegion`.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
@@ -1075,7 +1075,7 @@ object EmitStream {
 
       Code(LstartNewKey,
         Code.forLoop(i := 0, i < k, i := i + 1, result(i) = 0L),
-        curKey := eltRegions(winner).copyToSibling(mb, winnerPc, destRegion, keyType),
+        curKey := eltRegions(winner).copyTo(mb, winnerPc, destRegion, keyType),
         LaddToResult.goto)
 
       Code(LaddToResult,
@@ -1907,8 +1907,8 @@ object EmitStream {
                 val accRegion = eltRegion.createSiblingRegion(mb)
 
                 Code(Lpush,
-                     xAccInAccR := xAccInEltR.map(StagedRegion.copy(mb, _, eltRegion, accRegion)),
-                     push(xAccInEltR))
+                  xAccInAccR := xAccInEltR.map(eltRegion.copyTo(mb, _, accRegion)),
+                  push(xAccInEltR))
 
                 val bodyEnv = env.bind(accName -> xAccInAccR, eltName -> xElt)
                 val body = emitIR(bodyIR, env = bodyEnv, region = eltRegion)
--- a/hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
@@ -370,7 +370,7 @@ object Stream {
         Code(Lunreachable, Code._fatal[Unit]("unreachable"))
 
         val innerEltRegion = eltRegion
-          .asRoot(innerStreamType.separateRegions)
+          .asParent(innerStreamType.separateRegions, "StreamGrouped inner")
           .createChildRegion(mb)
 
         // LinnerPush is never executed; childEltRegion is cleared every element.
@@ -872,7 +872,7 @@ object EmitStream {
     mb: EmitMethodBuilder[_],
     pcStream: PCanonicalStreamCode,
     ab: StagedArrayBuilder,
-    destRegion: RootStagedRegion
+    destRegion: ParentStagedRegion
   ): Code[Unit] = {
     _write(mb, pcStream.stream, ab, destRegion)
   }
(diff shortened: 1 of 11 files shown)


patrick-schultz:
Sorry, missed a problem before pushing. Working on it.

tpoterba:
needs rebase.

patrick-schultz:
This required some changes. I'll let the CI run to make sure I didn't miss any other issues before removing the WIP tag and updating the description.

patrick-schultz:
Okay, ready for review.

patrick-schultz:
Running python tests in local mode gets some violations of the `eltRegion <= outerRegion`. Adding the WIP label until I fix that.

tpoterba:
yeah, the local backend tests have way more coverage than the java.

patrick-schultz:
Okay, python tests in local mode now have the same number of failures as on main. I just needed to be more careful in preserving the information that determines the subregion relation

tpoterba:
great, thanks Patrick! I'll finish my pass first thing tomorrow.

tpoterba:
benchmarks: https://gist.github.com/tpoterba/7b681ff2a6e5cba8304a19f32f5208b5

hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
```diff
@@ -1999,8 +2035,11 @@ object EmitStream {
                     eltRegion,
                     compare)
                     .map(joinF)
+              }
 
-              SizedStream(leftSetup, newStream, if (joinType == "left") leftLen else None)
+              sized(leftSetup,
+                          newStream,
```
tpoterba, line 2041: weird spacing

tpoterba (dismissed):
sorry this took a while to get through, it looks great. Minor formatting issue

tpoterba (approved):

PR #9402, opened 2020-09-02
[hail] fix Hail docs

danking:
We need to include the :members: directive in newer Sphinx versions. I also elided
hidden and inherited members for now, though we can add those back if we like.


johnc1231 (approved):
Thanks Dan!

PR #9405, opened 2020-09-02
Reduce serialization in writeBlockMatrices

johnc1231:
Using `inCtx` in `cmapPartitionsWithContextAndIndex` was serializing the whole RDD. I'm just inlining it by specifying `Iterator.single`.


johnc1231:
Not a significant difference

PR #9406, opened 2020-09-02
[auth] missing ssl_mode

danking:
SQLConfig requires an `ssl_mode`, this prevents auth from creating Developer accounts.

I already deployed this.


danking:
Overriding review for hotfix.

danking:
protection restored

PR #9407, opened 2020-09-03
[query] Reduce serialization/broadcasting in RepartitionedOrderedRDD2

tpoterba:


johnc1231 (approved):

johnc1231:
@tpoterba Compilation error, I think maybe you can only mark things that are `val` or `var` transient

tpoterba:
oops, thanks!

PR #9410, opened 2020-09-03
[annotationdb] Specify region when creating annotation database instance

pwc2:
Updated db.py to allow user to specify region as shown below.

`db = hl.experimental.DB(region='us')`
`mt = db.annotate_rows_db(mt, 'gnomad_lof_metrics')`

Data will be accessed from the requester pays bucket in the region specified by the user, available regions are `'us'` and `'eu'`.

Modified the following to include region parameter:
  - `DB` class 
  - `Dataset.from_name_and_json()`

Added method `DatasetVersion.insert_region()` to replace `'{region}'` in `DatasetVersion.url` instance variable with the specified region.

### DIFF
--- a/hail/python/hail/experimental/annotation_db.json
+++ b/hail/python/hail/experimental/annotation_db.json
@@ -3,52 +3,52 @@
   "url": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4341060/",
   "key_properties": ["unique"],
   "versions": [
-    {"url":"gs://hail-datasets-us/datasets/1/DANN.GRCh37.ht",
+    {"url":"gs://hail-datasets-{region}/datasets/1/DANN.GRCh37.ht",
      "version": "GRCh37"},
-    {"url":"gs://hail-datasets-us/datasets/1/DANN.GRCh38.ht",
+    {"url":"gs://hail-datasets-{region}/datasets/1/DANN.GRCh38.ht",
      "version": "GRCh38"}]},
  "CADD":
  {"description": "CADD: an algorithm designed to annotate both coding and non-coding variants.",
   "url": "http://europepmc.org/abstract/med/11384848",
   "key_properties": ["unique"],
   "versions": [
-    {"url":"gs://hail-datasets-us/datasets/1/CADD.v1.4.GRCh37.ht",
+    {"url":"gs://hail-datasets-{region}/datasets/1/CADD.v1.4.GRCh37.ht",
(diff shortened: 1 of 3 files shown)


hail/python/hail/experimental/db.py
```diff
@@ -78,31 +87,39 @@ class DB:
...
-                with open(config_path) as f:
-                    config = json.load(f)
+                if self.region in DB._valid_regions:
+                    config_path = pkg_resources.resource_filename(__name__, "annotation_db.json")
+                    assert os.path.exists(config_path), f'{config_path} does not exist'
+                    with open(config_path) as f:
+                        config = json.load(f)
+                else:
+                    raise ValueError(f'Specify valid region parameter, received: region={self.region}. '
+                                     f'Valid regions are {DB._valid_regions}.')
```
danking, line 122: We should verify the region is correct even if you're using a custom-url-loaded or explicit config, right? 

Let's move this check up above all these if-statements. Something like:

```python3
if region not in DB._valid_regions:
    raise ValueError(...)
```

danking (changes requested):
Awesome! Excited for this.

hail/python/hail/experimental/db.py
pwc2: Sounds good, I just moved the check up and pushed the changes.

hail/python/hail/experimental/annotation_db.json
```diff
@@ -3,119 +3,119 @@
   "url": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4341060/",
```
danking, line 3: add hail-datasets-eu to hailctl/start.py

danking (changes requested):

hail/python/hail/experimental/annotation_db.json
pwc2: Done

hail/python/hailtop/hailctl/dataproc/start.py
```diff
@@ -134,7 +134,7 @@
     'australia-southeast1': 'aus-sydney'
 }
 
-ANNOTATION_DB_BUCKETS = ["hail-datasets-us"]
+ANNOTATION_DB_BUCKETS = ["hail-datasets-us", "hail-datasets-eu"]
```
danking, line 137: `gnomad-public-requester-pays` needs to be here too

danking (changes requested):

PR #9411, opened 2020-09-03
[datasets] Update datasets.py to use checked-in config file

pwc2:
Renamed and moved `datasets/annotation_db.json` config file to `hail/experimental/datasets.json` and modified urls in `dataset[path]` to use a region parameter to load datasets from bucket in the appropriate region.

Modified `load_datasets()` function to no longer use the `config_file` parameter, and to require user to specify `region` parameter. The checked-in `hail/experimental/datasets.json` file will now be used as the config file.


danking (approved):
This is great!

PR #9420, opened 2020-09-08
[hailctl] Add `--debug-mode` flag to `hailctl dataproc start`

tpoterba:
Currently the only debug feature is the heap dump flag, but probably best to control debug features we might add in the future as a unit.


johnc1231 (approved):
Great inclusion, thanks. I think another thing that could be helpful is to periodically do a `jmap -hist:live` to count / inspect sizes of objects. That's a little more complicated, but might be helpful to avoid having to download full sized heap dumps. I'm also not sure if we are going to get heap dumps if yarn kills a process for using too much memory.

tpoterba:
yeah, that's a good idea. I think we'll get a heap dump on driver OOM, but not worker.

PR #9423, opened 2020-09-08
[ci] raise resources for java services tests

patrick-schultz:
Increase memory and cpu for test_hail_services_java to match java query tests. This contains tests of the shuffler IR, which runs the hail compiler, so it seems it should have the same resource limits as the other java query tests. #9401 is getting an out of memory error in `testShuffleIR`.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2836,8 +2836,8 @@ steps:
    image:
      valueFrom: hail_run_image.image
    resources:
-     memory: "3.75G"
-     cpu: "1"
+     memory: "7.5G"
+     cpu: "2"
    script: |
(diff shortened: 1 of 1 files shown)


tpoterba (approved):

jigold:
Will this actually get scheduled? I thought the max on our 2 core nodes was 1.8 or so.

patrick-schultz:
I made the change on #9401 and it ran.

tpoterba:
these are the same resource requests as `test_hail_java_*`

jigold:
I was confused with pods in kubernetes. You're both right.

PR #9425, opened 2020-09-09
[query] Remove invalid boundary on RepartitionedOrderedRDD2

tpoterba:
Fixes TableMapPartitions(indexed read)

### DIFF
--- a/hail/src/main/scala/is/hail/sparkextras/RepartitionedOrderedRDD2.scala
+++ b/hail/src/main/scala/is/hail/sparkextras/RepartitionedOrderedRDD2.scala
@@ -36,7 +36,7 @@ object RepartitionedOrderedRDD2 {
 class RepartitionedOrderedRDD2 private (@transient val prev: RVD, @transient val newRangeBounds: IndexedSeq[Interval])
   extends RDD[ContextRDD.ElementType[Long]](prev.crdd.sparkContext, Nil) { // Nil since we implement getDependencies
 
-  val prevCRDD: ContextRDD[Long] = prev.boundary.crdd
+  val prevCRDD: ContextRDD[Long] = prev.crdd
   val typ: RVDType = prev.typ
   val kOrd: ExtendedOrdering = PartitionBoundOrdering(typ.kType.virtualType)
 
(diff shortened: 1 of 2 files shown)


johnc1231:
Something is wrong here, you have tons of failing python tests.

tpoterba:
yeah... not looking forward to this debugging.

tpoterba:
I think this was probably a "two wrongs *did* make a right" scenario

johnc1231:
@patrick-schultz is usually good at seeing when we are breaking region memory management rules. But yeah, this seems like it broke everything.

patrick-schultz:
Could you explain why you think the boundary is invalid?

patrick-schultz:
Hmm, I think boundary is just unsound in general. It doesn't ensure that the pointers returned by `next` live as long as the consumer region.

I'll look over the PR and see if I can find any problems.

tpoterba:
> Could you explain why you think the boundary is invalid?

Boundary can only be used by consumers who generate an iterator with a new context. Here we were just inserting a clear before each next(), no matter who was consuming the iterator. The particular pipeline that triggered this error was a TableMapPartitions with a ToArray(Ref rows) after an IR with a repartitionedOrderedRDD2

hail/src/main/scala/is/hail/sparkextras/RepartitionedOrderedRDD2.scala
```diff
@@ -55,20 +55,74 @@ class RepartitionedOrderedRDD2 private (@transient val prev: RVD, @transient val
...
-        }.takeWhile { ptr =>
-          ur.set(ctx.r, ptr)
-          pord.lteq(key, range.right)
+
+    Iterator.single { (outerCtx: RVDContext) =>
+      new Iterator[Long] {
+        private[this] val innerCtx = outerCtx.freshContext()
+        private[this] val outerRegion = outerCtx.region
+        private[this] val innerRegion = innerCtx.region
+        private[this] val parentIterator = ordPartition.parents.iterator.flatMap(p => prevCRDD.iterator(p, context).next().apply(innerCtx))
```
patrick-schultz, line 64: You changed from `flatMap` to `next`. I'm not sure we can assume a ContextRDD has exactly one element per partition.

hail/src/main/scala/is/hail/sparkextras/RepartitionedOrderedRDD2.scala
```diff
@@ -55,20 +55,74 @@ class RepartitionedOrderedRDD2 private (@transient val prev: RVD, @transient val
...
+        private[this] val ur = new UnsafeRow(typ.rowType)
+        private[this] val key = new SelectFieldsRow(ur, typ.kFieldIdx)
+
+        // drop left elements at iterator allocation to avoid extra control flow in hasNext()
+        dropLeft()
+
+        private[this] def dropLeft(): Unit = {
+          var eltLessThan = true
+          while (parentIterator.hasNext && eltLessThan) {
+            innerRegion.clear()
```
patrick-schultz, line 76: This looks fishy. You're clearing the producer region after calling `hasNext` but before calling `next`. The `parentIterator` may have produced the element when you called `hasNext`, then you clear the region, then `next` returns the dangling pointer.

One way to fix it:
```scala
@tailrec def dropLeft(): Unit = {
  if (parentIterator.hasNext) {
    pull()
    if (pord.lt(key, range.left) {
      innerRegion.clear()
      dropLeft()
    } else if (pord.gt(key, range.right) {
      end()
    }
  } else {
    end()
  }
}
```

patrick-schultz (dismissed):

hail/src/main/scala/is/hail/sparkextras/RepartitionedOrderedRDD2.scala
tpoterba: really? oh, I'll change that.

hail/src/main/scala/is/hail/sparkextras/RepartitionedOrderedRDD2.scala
tpoterba: ahhhhh, yeah, this has to be it. Thanks Patrick!

tpoterba: one of the failing tests is passing, I'll push it up! 🎉

johnc1231: Thanks for catching that Patrick.

johnc1231:
This looks good to me now.

johnc1231 (approved):

johnc1231:
Thanks for tracking this down Tim

PR #9426, opened 2020-09-09
[hailtop/aiogoogle] added StorageClient

cseed:
Summary of changes:
 - Added a (low level) aiogoogle.StorageClient for Google Storage.
 - Added an abstract AsyncFS interface.  There are currently 3 implementations: GoogleStorageAsyncFS, LocalAsyncFS, and RouterAsyncFS, which can route (based on the scheme) to other file systems.  I tried to copy the standard Python os file system interface (open, remove, mkdir, isfile, etc.)  I'm skeptical about pushing hard on this perspective, and I think we're better off finding a natural interface that that unifies local, gs, (hadoop) and s3.  Maybe a read-only HttpFS.  I think I want to split open() into open (for reading) and create (for writing).  LocalFS operations are run in a thread pool which controls the level of IO parallelism.
 - AsyncFS.open returns an AsyncStream.  This is an async abstraction for Python File-like objects.  It still amazes me Python doesn't have something built in for this.

Next step is a high-level interface for listing objects.  Eventually this and JVM FS should mirror each other.

### DIFF
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -1,0 +1,134 @@
+import abc
+import asyncio
+import aiohttp
+from hailtop.aiotools import AsyncStream
+from .base_client import BaseClient
+
+
+class InsertObjectStream(AsyncStream):
+    def __init__(self, it, request_task):
+        super().__init__()
+        self._it = it
+        self._request_task = request_task
+        self._value = None
+
+    def writable(self):
+        return True
+
+    async def write(self, b):
+        assert not self.closed
+        await self._it.feed(b)
+
+    async def _wait_closed(self):
+        await self._it.stop()
+        async with await self._request_task as resp:
+            self._value = await resp.json()
+
+
+class GetObjectStream(AsyncStream):
+    def __init__(self, resp):
+        super().__init__()
+        self._resp = resp
+        self._content = resp.content
+
+    def readable(self) -> bool:
+        return True
+
+    async def read(self, n: int =-1) -> bytes:
+        assert not self._closed
+        return await self._content.read(n)
+
+    async def _wait_closed(self) -> None:
+        self._content = None
+        self._resp.release()
+        self._resp = None
+
+
+class StorageClient(BaseClient):
+    def __init__(self, **kwargs):
+        super().__init__('https://storage.googleapis.com/storage/v1', **kwargs)
+
+    # docs:
+    # https://cloud.google.com/storage/docs/json_api/v1
+
+    async def insert_object(self, bucket: str, name: str, **kwargs) -> InsertObjectStream:
+        if 'params' in kwargs:
+            params = kwargs['params']
+        else:
+            params = {}
+            kwargs['params'] = params
+        assert 'name' not in params
+        params['name'] = name
+        assert 'uploadType' not in params
+        params['uploadType'] = 'media'
+
+        assert 'data' not in kwargs
+        it = FeedableAsyncIterable()
+        kwargs['data'] = aiohttp.AsyncIterablePayload(it)
+        request_task = asyncio.create_task(self._session.post(
+            f'https://storage.googleapis.com/upload/storage/v1/b/{bucket}/o',
+            **kwargs))
+        return InsertObjectStream(it, request_task)
+
+    async def get_object(self, bucket: str, name: str, **kwargs) -> GetObjectStream:
+        if 'params' in kwargs:
+            params = kwargs['params']
+        else:
+            params = {}
+            kwargs['params'] = params
+        assert 'alt' not in params
+        params['alt'] = 'media'
+
+        resp = await self._session.get(f'https://storage.googleapis.com/storage/v1/b/{bucket}/o/{name}', **kwargs)
+        return GetObjectStream(resp)
+
+    async def get_object_metadata(self, bucket: str, name: str, **kwargs):
+        params = kwargs.get('params')
+        assert not params or 'alt' not in params
+        return await self.get('/b/{bucket}/o/{name}', **kwargs)
+
+
+class GoogleStorageAsyncFS(AsyncFS):
+    def __init__(self, storage_client: StorageClient = None):
+        if not storage_client:
+            storage_client = StorageClient()
+        self._storage_client = storage_client
+
+    def schemes(self) -> List[str]:
+        return ['gs']
+
+    async def open(self, url: str, mode: str = 'r') -> AsyncStream:
+        if not all(c in 'rwxabt+' for c in mode):
+            raise ValueError(f"invalid mode: {mode}")
+        if 't' in mode and 'b' in mode:
+            raise ValueError(f"can't have text and binary mode at once: {mode}")
+        if 'b' not in mode:
+            raise ValueError(f"text mode not supported: {mode}")
+        if 'x' in mode:
+            raise ValueError(f"exclusive creation not supported: {mode}")
+        if 'a' in mode:
+            raise ValueError(f"append mode not supported: {mode}")
+        if '+' in mode:
+            raise ValueError(f"updating not supported: {mode}")
+        if ('r' in mode) + ('w' in mode) != 1:
+            raise ValueError(f"must have exactly one of read/write mode: {mode}")
+
+        parsed = urllib.parse.urlparse(url)
+        if parsed.scheme != 'gs':
+            raise ValueError(f"invalid scheme, expected gs: {scheme}")
+        bucket = parsed.netloc
+
+        name = parsed.path
+        if name:
+            assert name[0] == '/'
+            name = name[1:]
+
+        if 'r' in mode:
+            return self._storage_client.get_object(bucket, name)
+        else:
+            assert 'w' in mode
+            return self._storage_client.insert_object(bucket, name)
+
+    async def close(self):
+        await self._storage_client.close()
+        self._storage_client = None
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -2,20 +2,17 @@ from typing import Tuple, Any, Set, Optional
 import asyncio
 import urllib.parse
 import aiohttp
-from hailtop.aiotools import AsyncStream, AsyncFS, FeedableAsyncIterable
+from hailtop.aiotools import ReadableStream, WritableStream, AsyncFS, FeedableAsyncIterable
 from multidict import CIMultiDictProxy  # pylint: disable=unused-import
 from .base_client import BaseClient
 
 
-class InsertObjectStream(AsyncStream):
+class InsertObjectStream(WritableStream):
     def __init__(self, it, request_task):
         super().__init__()
         self._it = it
         self._request_task = request_task
         self._value = None
-
-    def writable(self):
-        return True
 
     async def write(self, b):
         assert not self.closed
@@ -26,17 +23,13 @@ class InsertObjectStream(AsyncStream):
         await self._it.stop()
         async with await self._request_task as resp:
             self._value = await resp.json()
-            print(self._value)
 
 
-class GetObjectStream(AsyncStream):
+class GetObjectStream(ReadableStream):
     def __init__(self, resp):
         super().__init__()
         self._resp = resp
         self._content = resp.content
-
-    def readable(self) -> bool:
-        return True
 
     async def read(self, n: int = -1) -> bytes:
         assert not self._closed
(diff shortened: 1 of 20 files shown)


cseed:
@jigold came up, but FYI @catoverdrive

cseed:
And I'd eventually like a (synchronous) version of this interface to replace the hadoop_* functions.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -0,0 +1,164 @@
...
+    async def remove(self, url: str) -> None:
+        path = self._get_path(url)
+        return os.remove(path)
+
+    async def rmtree(self, url: str) -> None:
+        path = self._get_path(url)
+        await blocking_to_async(self._thread_pool, shutil.rmtree, path)
+
+
+class RouterAsyncFS(AsyncFS):
```
jigold, line 103: Why is this called a RouterAsyncFS and takes a list of filesystems? This seems like a wrapper for all file systems and should just have all possible file systems as a class property and not take the list of filesystems as an argument.

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -0,0 +1,204 @@
...
+    @staticmethod
+    def _get_bucket_name(url: str) -> Tuple[str, str]:
+        parsed = urllib.parse.urlparse(url)
+        if parsed.scheme != 'gs':
+            raise ValueError(f"invalid scheme, expected gs: {parsed.scheme}")
+
+        name = parsed.path
+        if name:
+            assert name[0] == '/'
+            name = name[1:]
```
jigold, line 123: I didn't know you could do this. Very clever!

```
>>> import urllib.parse
>>> parsed = urllib.parse.urlparse(url)
>>> str(parsed)
"ParseResult(scheme='gs', netloc='hail-jigold', path='/a/b/c.txt', params='', query='', fragment='')"
```

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -0,0 +1,204 @@
...
+
+        bucket, name = self._get_bucket_name(url)
+
+        if 'r' in mode:
+            return await self._storage_client.get_object(bucket, name)
+
+        assert 'w' in mode
+        return await self._storage_client.insert_object(bucket, name)
+
+    async def mkdir(self, url: str) -> None:
```
jigold, line 151: This is where I got stuck when I tried to do this. I'd think that if you call `mkdir` and then followed by `isdir` that the directory would exist. But I didn't want a bunch of hanging files that ended in slashes. Like you said, unifying the local file system and the google one with object storage is not straightforward and it's not worth trying to unify them.

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -0,0 +1,204 @@
...
+        pass
+
+    async def isfile(self, url: str) -> bool:
+        try:
+            bucket, name = self._get_bucket_name(url)
```
jigold, line 156: This lets you have a file that ends in a slash. It might be worth putting as comments some of these assumptions for downstream users of this library.

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -0,0 +1,204 @@
...
+        if not name.endswith('/'):
+            name = f'{name}/'
+        params = {
+            'delimiter': '/',
+            'includeTrailingDelimiter': 'true',
+            'maxResults': 1,
+            'prefix': name
+        }
+        data = await self._storage_client._list_objects(bucket, params=params)
+        assert 'prefixes' not in data or data['prefixes'] is None
```
jigold, line 175: Is the prefixes in the case where you don't specify a delimiter?

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -0,0 +1,204 @@
...
+            name = f'{name}/'
+        params = {
+            'prefix': name
+        }
+        done = False
+        while not done:
+            done = True
+            data = await self._storage_client._list_objects(bucket, params=params)
+            items = data.get('items')
+            print(items)
```
jigold, line 195: delete?

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -0,0 +1,204 @@
...
+            if items:
+                for item in items:
+                    await self._storage_client.delete_object(bucket, item['name'])
+                    print('done: False')
```
jigold, line 199: delete?

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -0,0 +1,204 @@
...
+        bucket, name = self._get_bucket_name(url)
+        if not name.endswith('/'):
+            name = f'{name}/'
+        params = {
+            'prefix': name
+        }
+        done = False
+        while not done:
+            done = True
+            data = await self._storage_client._list_objects(bucket, params=params)
```
jigold, line 193: Is there a page that list_objects returns? I would have thought this should be an async generator or have some sort of params option that lists the maximum number of results. Otherwise, a while loop here doesn't make sense.

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -0,0 +1,204 @@
...
+    async def write(self, b):
+        assert not self.closed
+        await self._it.feed(b)
+        return len(b)
+
+    async def _wait_closed(self):
+        await self._it.stop()
+        async with await self._request_task as resp:
+            self._value = await resp.json()
+            print(self._value)
```
jigold, line 29: delete?

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -0,0 +1,204 @@
...
+        super().__init__()
+        self._resp = resp
+        self._content = resp.content
+
+    def readable(self) -> bool:
+        return True
+
+    async def read(self, n: int = -1) -> bytes:
+        assert not self._closed
+        return await self._content.read(n)
```
jigold, line 43: Is the response keeping track of where it is in the stream and you don't manually have to do anything? If so, that's nice.

hail/python/test/hailtop/test_aiogoogle.py
```diff
@@ -0,0 +1,140 @@
...
+
+@pytest.mark.asyncio
+async def test_get_object_headers():
+    bucket = os.environ['HAIL_TEST_BUCKET']
+    file = secrets.token_hex(16)
+
+    async with StorageClient() as client:
+        async with await client.insert_object(bucket, file) as f:
+            await f.write(b'foo')
+        async with await client.get_object(bucket, file) as f:
```
jigold, line 136: Should this not test `client.get_object_metadata` as well?

jigold (changes requested):

hail/python/hailtop/aiotools/fs.py
```diff
@@ -0,0 +1,164 @@
...
+
+    @staticmethod
+    def _get_path(url):
+        parsed = urllib.parse.urlparse(url)
+        if parsed.scheme and parsed.scheme != 'file':
+            raise ValueError(f"invalid scheme, expected file: {parsed.scheme}")
+        return parsed.path
+
+    async def open(self, url: str, mode: str = 'r') -> AsyncStream:
+        if 'b' not in mode:
```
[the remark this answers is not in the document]
cseed: I agree.  That's part of why I said "I'm skeptical about pushing hard on [copying the standard Python interface]".  Let me change this now.  I'm going to have:

```
  def open(self, url: str) -> AsyncStream:
    ...

  def create(self, url: str) -> AsyncStream
    ...
```

And the `mode` argument will go away.  If I do this, I can also split AsyncStream into `ReadableStream` and `WritableStream`.  That's better than the current design, because we won't support read-and-writable streams.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -0,0 +1,164 @@
...
+            raise ValueError(f"can't open: text mode not supported: {mode}")
+        return blocking_stream_to_async(self._thread_pool, cast(BinaryIO, open(self._get_path(url), mode)))
```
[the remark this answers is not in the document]
cseed: Yes, for mypy.  Based on the mode, `open()` can return a TextIO or a BinaryIO, but I declared blocking_stream_to_async to take a BinaryIO only, so we need to cast.  Alternatively, I think we could have written something like:

```
  f = open(...)
  assert isinstance(f, BianryIO)
  return blocking_stream_to_async(..., f)
```

hail/python/hailtop/aiotools/fs.py
```diff
@@ -0,0 +1,164 @@
...
+        await blocking_to_async(self._thread_pool, shutil.rmtree, path)
+
+
+class RouterAsyncFS(AsyncFS):
+    def __init__(self, default_scheme: Optional[str], filesystems: List[AsyncFS]):
+        scheme_fs = {}
+        schemes = set()
+        for fs in filesystems:
+            for scheme in fs.schemes():
+                scheme_fs[scheme] = fs
```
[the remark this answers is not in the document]
cseed: Oops, I got it backwards, I meant to search them in order.  Will fix.

Right now, none of our filesystems having overlapping schemes.

hail/python/hailtop/aiotools/fs.py
cseed: We will not always be able to construct all filesystems in all deployments.  For example, the Google Storage FS fails if it cannot find credentials.

hail/python/hailtop/aiogoogle/client/storage_client.py
cseed: > I'd think that if you call mkdir and then followed by isdir that the directory would exist.

Yep, there were a few choices here, and I think the simplest is to actually break the mkdir/isdir relationship.  You notice in my isdir test, I both call mkdir and create a file before calling isdir.  The other option was not to include isdir, and I'm very tempted to do this.  Directories are not a real concept in gs://, and it is maybe a mistake to try to emulate them.

The other option is to rename isdir/mkdir.  What it `isdir(p)` really means: are there are objects with `p/` as a prefix, and `mkdir(p)` means: you can now create an object with prefix `p/`.

I think once we get the interface sorted (including directory listing), we should document how we're thinking about it very carefully.

hail/python/hailtop/aiogoogle/client/storage_client.py
cseed: Yes.  I haven't implemented that yet, I was planning to do it as part of adding a high level list objects interface.  This works without the paging by deleting the objects on the first page (the API defaults to 1000 items), and then listing again after deleting those.  Since we deleted, when we list again, we'll get a new page of objects without using the next page token.

hail/python/hailtop/aiogoogle/client/storage_client.py
cseed: Yep.  So `content` is a AsyncStream-like interface that's part of aiohttp, StreamReader:

https://docs.aiohttp.org/en/stable/streams.html#aiohttp.StreamReader

The Python standard library has something similar which is returned when creating network connections, StreamReader:

https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader

Annoyingly, neither of those interfaces can be implemented by external code, hence I created AsyncStream.   GetObjectStream is just a simple AsyncStream wrapper around the aiohttp stream.

hail/python/hailtop/aiogoogle/client/storage_client.py
cseed: Since I think we're still working on the interface, can we leave downstream docs until it has settled?

 > I think once we get the interface sorted (including directory listing), we should document how we're thinking about it very carefully.

hail/python/hailtop/aiotools/fs.py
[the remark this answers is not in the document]
cseed: I tried this and it didn't work.  `open()` returns a io.BufferedWriter, but that isn't actually a subclass of `BinaryIO`.  `BinaryIO` appears to be an interface just for typing purposes.  `BinaryIO` is better, because it specifies that the read/write methods take/return bytes (instead of str, as you'd get for a TextIO).  Going to leave as is.

hail/python/hailtop/aiotools/fs.py
cseed: Also, I'm not a fan of totally global configuration.  The Hadoop design for example caused us lots of problem and should be avoided.

hail/python/hailtop/aiogoogle/client/storage_client.py
cseed: I think this code is wrong.  I need to test it on nested directories.

So I don't totally understand the Google API on this point, but list objects returns two things: the items you requested, and "prefixes".  See:

https://cloud.google.com/storage/docs/json_api/v1/objects/list

 > prefixes[] contains truncated object names for objects whose names contain delimiter after any prefix. Object names are truncated beyond the first applicable instance of the delimiter. If multiple objects have the same truncated name, duplicates are omitted.

It sounds like prefixes basically "subdirectories" of the requested prefix.  However, `maxResults` limits the combined items and prefixes:

 > Maximum combined number of entries in items[] and prefixes[] to return in a single page of responses. Because duplicate entries in prefixes[] are omitted, fewer total results may be returned than requested. The service uses this parameter or 1,000 items, whichever is smaller.

So I think I need to check prefixes and items here.

I will run some experiments and add some additional tests.

cseed: OK, yeah, when you specify a delimiter, prefixes are essentially the subdirectories, and items are the files.  I modified the code to look for a prefix or an item, and added a subdirectory test.

hail/python/test/hailtop/test_aiogoogle.py
cseed: Done

jigold (approved):

PR #9427, opened 2020-09-09
[query] Fix data race in FASTA reader

chrisvittal:
In readBlock, membership was checked in a LinkedHashMap, then if it was
found, the value was retrieved from the cache. This lead to a data race
where a value could be evicted between the check and retrival. The
solution is to grab the block value out of the map, and if it is not
null, return it, otherwise, grab the block string, put it in the map,
and return it.

Co-authored-by: Tim Poterba <tpoterba@broadinstitute.org>


tpoterba:
Did you run on dataproc? Could you replicate either the NPE or ClosedChannelException?

chrisvittal:
The NPE, no. ClosedChannel, again, no.

chrisvittal:
I did run it on Dataproc.

tpoterba:
huh, weird. So we maybe don't expect this to fix Julia's issue?

chrisvittal:
I'm worried it's some other weird IO bug, but this should still fix the data race.

patrick-schultz (approved):

PR #9428, opened 2020-09-09
Reduce copying when emitting NDArrays

johnc1231:
# Current situation

In several places, we call a method `emitNDArrayStandardStrides`, which has the effect of calling `emitDeforestedNDArray`, since that is known to always emit things in column major order. However, the downside of this is that we were doing unnecessary copies of the data, even when it was already in column major order, by constructing an `NDArrayEmitter` that just emitted an NDArray and then looked up values from it:

```
 case _ =>
          val ndt = emit(x)
          val ndAddress = mb.genFieldThisRef[Long]()
          val setup = (ndAddress := ndt.value[Long])
          val xP = x.pType.asInstanceOf[PNDArray]

          val shapeAddress = new Value[Long] {
            def get: Code[Long] = xP.shape.load(ndAddress)
          }
          val shapeTuple = new CodePTuple(xP.shape.pType, shapeAddress)

          val shapeArray = (0 until xP.shape.pType.nFields).map(i => shapeTuple.apply[Long](i))

          new NDArrayEmitter[C](nDims, shapeArray,
            xP.shape.pType, xP.elementType, setup, ndt.setup, ndt.m) {
            override def outputElement(elemMB: EmitMethodBuilder[C], idxVars: IndexedSeq[Value[Long]]): Code[_] =
              xP.loadElementToIRIntermediate(idxVars, ndAddress, elemMB)
          }
```

# New Situation

We now have `emitNDArrayColumnMajorStrides`, which calls `emit` on an ndarray, checks if the emitted thing is column major, and only does a copy if it needs to. This uses new `LinalgCodeUtils` methods `checkColumnMajor` and `createColumnMajorCode`. Everything else in `LinalgCodeUtils` was unused / old style and I removed them.


hail/src/main/scala/is/hail/linalg/LinalgCodeUtils.scala
```diff
@@ -1,69 +1,70 @@
...
-    Code.forLoop(rowIndex := 0L, rowIndex < nRows, rowIndex := rowIndex + 1L,
-      Code.forLoop(colIndex := 0L, colIndex < nCols, colIndex := colIndex + 1L,
-        Region.storePrimitive(elementPType, (targetFirstElementAddress + colMajorCoord * elementPType.byteSize))(currentElement)
-      )
-    )
+    val elementType = pndv.pt.elementType
+    val nDims = pndv.pt.nDims
+    cb.append(Code(
+      runningProduct := elementType.byteSize,
+      Code._println(pndv.value.asInstanceOf[Value[Long]].toS),
```
chrisvittal, line 20: should this print be here?

chrisvittal (changes requested):
Think you missed a print, otherwise, looks great!

hail/src/main/scala/is/hail/linalg/LinalgCodeUtils.scala
johnc1231: Nope, fixed. Thanks.

chrisvittal (approved):

PR #9429, opened 2020-09-09
[query] make StreamFlatMap use reducible control flow

patrick-schultz:
Going off a suspicion that the JVM jit won't compile methods containing irreducible control flow, I tried to fix StreamFlatMap to be reducible. A simple benchmark based on `testES2FlatMap` showed a 2x speedup, which seems to confirm the suspicion.

What caused the irreducibility was the following set of control flow paths (pretend the `Lpull` label was defined in the old version too):
* `Lpull -> LinnerPull`
* `Lpull -> LouterPull`
* `LinnerPull -> innerSource.eos -> LinnerEos -> LouterPull`
* `LouterPull -> outerSource.push -> LinnerPull`

The later two paths form a loop, and the first two make two entries into the loop - the basic irreducible control flow pattern. The fix redirects the last path to go to `Lpull` instead, which will happen to branch to `LinnerPull`.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
@@ -389,6 +389,7 @@ object Stream {
         val LouterPull = CodeLabel()
         var innerSource: Source[A] = null
         val LinnerPull = CodeLabel()
+        val Lpull = CodeLabel()
         val LinnerEos = CodeLabel()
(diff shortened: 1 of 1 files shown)


tpoterba (changes requested):
can you add a benchmark that does range_table.annotate() with the thing you tested locally, and run on your laptop before/after?

patrick-schultz:
Sure

PR #9430, opened 2020-09-09
Avoid calling a tail -> collect in blanczos pca

johnc1231:
Tried this locally on a `hl.balding_nichols_model(20, 6000, 50000)` dataset, overall time for 10 pcs with 2 iterations went from ~50 seconds to ~40 seconds. Surprised it made that much of a difference.


tpoterba (approved):
wow, nice!

PR #9433, opened 2020-09-10
[batch] Add implicit_dirs to gcsfuse

jigold:


jigold:
Let's try to get this in today so Ben W. can use it. Thanks!

jigold:
The issue was that he couldn't see directories and files beyond the top level of the bucket.

catoverdrive (approved):

PR #9435, opened 2020-09-10
[query] Work around concurrency issues in FASTAReader

chrisvittal:
Introduce FASTAReaderConfig to act as a kind of factory for FASTAReader,
while all FASTAReaders themselves are confined to ThreadLocals (except
in tests).

Furthermore, add a lock around the fasta file map to prevent more than
one fasta from being copied per jvm. The can be lock contention on the
map, but if there is large amounts of waiting for said lock, then it
usually means that a fasta is downloading and we definitely should be
waiting.

### DIFF
--- a/hail/src/main/scala/is/hail/variant/ReferenceGenome.scala
+++ b/hail/src/main/scala/is/hail/variant/ReferenceGenome.scala
@@ -7,21 +7,22 @@ import is.hail.HailContext
 import is.hail.asm4s.Code
 import is.hail.backend.BroadcastValue
 import is.hail.check.Gen
-import is.hail.types._
+import is.hail.expr.ir.{EmitClassBuilder, ExecuteContext, RelationalSpec}
 import is.hail.expr.{JSONExtractContig, JSONExtractIntervalLocus, JSONExtractReferenceGenome, Parser}
-import is.hail.io.reference.FASTAReader
+import is.hail.io.fs.FS
+import is.hail.io.reference.LiftOver
+import is.hail.io.reference.{FASTAReader, FASTAReaderConfig}
+import is.hail.types._
+import is.hail.types.virtual.{TInt64, TLocus, Type}
 import is.hail.utils._
-import org.json4s._
-import org.json4s.jackson.{JsonMethods, Serialization}
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable
 import scala.language.implicitConversions
-import is.hail.expr.ir.{EmitClassBuilder, ExecuteContext, RelationalSpec}
-import is.hail.types.virtual.{TInt64, TLocus, Type}
-import is.hail.io.reference.LiftOver
-import is.hail.io.fs.FS
 import org.apache.spark.TaskContext
+import org.json4s._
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/variant/ReferenceGenome.scala
```diff
@@ -343,33 +344,41 @@ case class ReferenceGenome(name: String, contigs: Array[String], lengths: Map[St
...
   }
 
   def addSequenceFromReader(tmpdir: String, fs: FS, fastaFile: String, indexFile: String, blockSize: Int, capacity: Int): ReferenceGenome = {
-    fastaReader = new FASTAReader(tmpdir, fs.broadcast, this, fastaFile, indexFile, blockSize, capacity)
+    fastaReaderCfg = FASTAReaderConfig(tmpdir, fs.broadcast, this, fastaFile, indexFile, blockSize, capacity)
     this
   }
 
-  def getSequence(contig: String, position: Int, before: Int = 0, after: Int = 0): String = {
+  private val realFastaReader: ThreadLocal[FASTAReader] = new ThreadLocal[FASTAReader]
```
tpoterba, line 355: should be transient and with a default value of `null`, I think?

tpoterba: will need 2 levels of null handling in the fastaReader method I think

tpoterba (dismissed):
This looks like the Right Solution™, thanks for pushing on this.

One comment on serialization

hail/src/main/scala/is/hail/variant/ReferenceGenome.scala
chrisvittal: I've tested this, and it works, in a cluster even. There's no need for the extra layer of transience/null checks.

tpoterba: have you tested with a pipeline that would initialize the fasta reader on the driver machine before generating a spark stage?

I'm not sure how to reason about what's serialized if the value is defined on the driver.

perhaps:

```
ht = ht.annotate_globals(foo = hl.locus('1', 100000).sequence_context())
ht = ht.annotate_globals(foo_glob = ht.foo, foo_row = ht.locus.sequence_context())
```

chrisvittal: Just did. Worked fine.

chrisvittal: Not sure how this is working though, since ThreadLocal is not serializable in java, but none of this seems to effect scala's serialization.

chrisvittal: I'm pretty sure that making this `@transient lazy val` is close to what we should do anyways, so I just did that.

chrisvittal:
gnomAD people have mentioned that they need the fixes soonish. Is there anything else we need to address here?

tpoterba (approved):

tpoterba:
oops, thought I'd hit approve this morning!

PR #9436, opened 2020-09-10
[query] Add retry to FASTAReader.getSequence

chrisvittal:
On certain occasions, the file inside the FASTA reader can be closed. It
is not yet understood why. Tests indicate that simply reopening the file
seems to resolve the issue when we run the appropriate exception.

This also removes the need for SerializableReferenceSequenceFile, and so
deletes it.


chrisvittal:
Will rebase as soon as stacked PR goes in.

tpoterba (approved):

PR #9385, opened 2020-08-31
[batch] add rest api for billing_projects

jigold:
- I use a right join in the query so that if there are no users, the record['users'] ends up being None rather than [None].
- I decided that an arbitrary user should be able to see all the other users for billing projects they are on.
- I decided that a developer can see all billing projects.

This is infrastructure that I figured was nice to have but also will make #9354 better so I can add proper tests.

PR #9437, opened 2020-09-10
[batch] add option for mounting user token inside batch job

catoverdrive:
I'd like to give the user the ability to authenticate to our services from within a batch job. The specific use case I need it for is for the query service to be able to cache things with the memory service, but it seems like it could be more broadly applicable. I'm unsure whether this is the correct way to do it.

This is currently not exposed in the python user interface (only in BatchClient), but I can pipe the option through in this PR if we want to.

### DIFF
--- a/batch/test/test_batch.py
+++ b/batch/test/test_batch.py
@@ -578,11 +578,12 @@ echo $HAIL_BATCH_WORKER_IP
         assert with_token_status['state'] == 'Success', with_token_status
 
         username = get_userinfo()['username']
-        auth_pattern = r'{' + ','.join(r'\s*"' + k + r'":\s*"(?P<' + k + '>[-@._\w\d]+)"' for k in ["username", "email", "gsa_email"]) + r'\s*}'
 
-        import re
-        m = re.search(auth_pattern, with_token.log()['main'])
-        assert m is not None and m.group("username") == username, (username, with_token.log()['main'])
+        try: 
+            job_userinfo = json.loads(with_token.log()['main'].strip())
+        except: 
+            job_userinfo = None
+        assert job_userinfo is not None and job_userinfo["username"] == username, (username, with_token.log()['main'])
 
         no_token_status = no_token.wait()
         assert no_token_status['state'] == 'Failed', no_token_status
--- a/batch/test/test_batch.py
+++ b/batch/test/test_batch.py
@@ -569,8 +569,7 @@ echo $HAIL_BATCH_WORKER_IP
 
     def test_user_authentication_within_job(self):
         batch = self.client.create_batch()
-        default_ns = os.environ['HAIL_DEFAULT_NAMESPACE']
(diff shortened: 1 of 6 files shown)


catoverdrive:
Ok, I finally figured out why the test wasn't picking up the tokens. This is going to be fairly useless to users unless/until we also bundle in a dev config, although it definitely should suffice for what I need to do. Thoughts on doing that? (either here or in a separate PR?) I imagine the config would just be like {location: gce, namespace: $default_ns} but I'm not sure if that would leads to issues?

jigold:
Maybe do it in a separate PR if your stuff is working without it. Otherwise, I don't think it hurts.

catoverdrive:
Sounds good! I'll PR it separately. Mostly a question of whether it's a safe/desirable thing to do, I think?

jigold:
I think it's safe. I was more worried about what the namespace should be and whether for developers it should have the service_namespace. I think the deploy config only tells you how to get the correct URL. There's nothing special about it.

catoverdrive:
Yeah, that's true. I've updated this PR to include the default gce deploy config for the default namespace, which I think is generally what you'd want for both testing and in production; I'm not sure under what circumstances you'd want separate service namespaces?

batch/test/test_batch.py
```diff
@@ -566,3 +566,18 @@ def test_verify_no_access_to_metadata_server(self):
         status = j.wait()
         assert status['state'] == 'Failed', status
         assert "Connection timed out" in j.log()['main'], (j.log()['main'], status)
+
+    def test_user_authentication_within_job(self):
+        batch = self.client.create_batch()
+        cmd = ['bash', '-c', f'hailctl batch list']
```
jigold, line 572: This is not the best test because it's going to list all possible batches that have been created ever (I think). I tried to add a CLI for billing projects in batch. That would be better to use. See #9385

The other option is to see if we can add paging to the CLI interface. But that seems like more work. I'm not sure why my PR keeps failing test_query.


Actually, there's a `-q` parameter. Maybe we can use that to select only running batches. That would probably work :)

jigold (changes requested):

batch/test/test_batch.py
catoverdrive: cool, I'll take a look!

I think that for most of the tests that get run e.g. inside PRs, we'll always have a more-or-less limited number of batches (~120?) but it would be definitely good to restrict this. I can also probably do `hailctl auth list` and check for the expected namespace, since the purpose is just to test that the auth is working correctly; do you think that would work?

jigold: Yes. That should work too. I'm worried about the deploy tests.

jigold: I think figuring out the `-q running` would be fine as well.

jigold (approved):

PR #9438, opened 2020-09-11
emitI NDArrayMatmul and NDArrayInv

johnc1231:
I'm trying to move all NDArray emission to emitI and update the style a bit, with the plan to eventually overhaul `NDArrayEmitter` to be more in line with the `PCode` interface. This PR updates `NDArrayMatmul` and `NDArrayInv` emit rules to be `emitI` rules, and cleans it up a bit. The diff isn't great, main changes were doing those `emit / flatmap` calls at the top to create `PCode`s, then using the corresponding `PValue`s everywhere I could instead of relying on `PType` methods. All the "setup" and missingness check code also went away since the `flatmap`s handle that stuff.


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -857,6 +857,199 @@ class Emit[C](
...
+                    region.code, (M * N).toI)
+                ),
+                outputPType.construct(
+                  outputPType.makeShapeBuilder(IndexedSeq(M, N)),
+                  outputPType.makeColumnMajorStridesBuilder(IndexedSeq(M, N), mb),
+                  answerPArrayAddress,
+                  mb,
+                  region.code))
+
+              EmitCode(Code._empty, false, PCode(pt, multiplyViaDGEMM)).toI(cb)
```
catoverdrive, line 940: I'm not super excited by the idea of mixing the two paradigms here. If you need to use the NDArrayEmitter here, it might be a good opportunity to introduce a mostly imperative version before you port this since it sounds like that's the direction you're heading.

catoverdrive (changes requested):

hail/src/main/scala/is/hail/expr/ir/Emit.scala
johnc1231: When you say two paradigms, do you mean that I am making an `EmitCode` still? Or just generally that I'm mostly still using giant `Code` blocks? I can clean up this top part you highlighted some more and not construct an `EmitCode`, but the below emitter parts would require some emitter updates. 

As for direction I'm heading, mostly I was trying to get all the NDArray stuff a little cleaner and into `emitI`, at which point I was planning to make a follow up PR that pulled all the NDArray emission stuff into something called `emitNDArray` so that I can make sure any final NDArray results that get emitted come back column major, even if it's possible for intermediate NDArrays not to be. I wanted to do that so that we can remove `strides` from `TNDArray`, where it certainly doesn't belong. I was also going to go off and try to create an imperative version of `NDArrayEmitter`.

catoverdrive: Yes, both. 

My suggestion would be to get the imperative NDArray emitter in before/simultaneous with this change. Afaict, that's not blocked on this change (although feel free to push back if it turns out I'm wrong!) but will make future lifts smoother and more complete (including this one!).

johnc1231:
Closing in favor of #9450

PR #9439, opened 2020-09-11
[batch] convert activity logs to audit logs

jigold:
I got an email saying the activity logs are no longer supported after September 30th. Here's the [migration instructions](https://cloud.google.com/compute/docs/logging/migrating-from-activity-logs-to-audit-logs#log_entry_field_mappings).

I figured out how to map the fields mostly by trial and error looking at the JSON for an event. The only thing that didn't map at all was the operationType. I hardcoded that as 'insert'.

There are different event_subtype names such as 'v1.compute.instances.insert' or 'beta.compute.instances.insert'. So I did what they suggested and looked for a partial match such as 'compute.instances.insert'.

I can send you the full JSON for the events if you want to double check anything.

I also double checked that the activity logs aren't used anywhere else in the repo, but it might be good for you to confirm that since you wrote a lot of this.

### DIFF
--- a/batch/batch/driver/instance_pool.py
+++ b/batch/batch/driver/instance_pool.py
@@ -678,13 +678,11 @@ gsutil -m cp dockerd.log gs://$WORKER_LOGS_BUCKET_NAME/batch/logs/$INSTANCE_ID/w
             log.warning(f'unknown event resource type {resource_type}')
             return
 
-        if 'last' in event['operation']:
-            event_type = 'GCE_OPERATION_DONE'
-        elif 'first' in event['operation']:
-            event_type = 'GCE_API_CALL'
+        operation_started = event['operation'].get('first', False)
+        if operation_started:
+            event_type = 'STARTED'
         else:
-            log.warning(f'unknown event_type {event["operation"]}')
-            event_type = None
+            event_type = 'COMPLETED'
 
         event_subtype = payload['methodName']
         resource = event['resource']
@@ -696,8 +694,8 @@ gsutil -m cp dockerd.log gs://$WORKER_LOGS_BUCKET_NAME/batch/logs/$INSTANCE_ID/w
             log.warning(f'event for unknown machine {name}')
             return
 
-        if 'compute.instances.insert' in event_subtype:
-            if event_type == 'GCE_OPERATION_DONE':
+        if event_subtype == 'v1.compute.instances.insert':
+            if event_type == 'COMPLETED':
                 severity = event['severity']
                 operation_type = 'insert'
                 success = (severity != 'ERROR')
@@ -708,14 +706,14 @@ gsutil -m cp dockerd.log gs://$WORKER_LOGS_BUCKET_NAME/batch/logs/$INSTANCE_ID/w
                 log.warning(f'event for unknown instance {name}')
                 return
 
-            if 'compute.instances.preempted' in event_subtype:
+            if event_subtype == 'v1.compute.instances.preempted':
                 log.info(f'event handler: handle preempt {instance}')
                 await self.handle_preempt_event(instance, timestamp)
-            elif 'compute.instances.delete' in event_subtype:
-                if event_type == 'GCE_OPERATION_DONE':
+            elif event_subtype == 'v1.compute.instances.delete':
+                if event_type == 'COMPLETED':
                     log.info(f'event handler: delete {instance} done')
                     await self.handle_delete_done_event(instance, timestamp)
-                elif event_type == 'GCE_API_CALL':
+                elif event_type == 'STARTED':
                     log.info(f'event handler: handle call delete {instance}')
                     await self.handle_call_delete_event(instance, timestamp)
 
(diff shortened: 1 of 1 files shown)


batch/batch/driver/instance_pool.py
```diff
@@ -657,49 +666,52 @@ async def handle_call_delete_event(self, instance, timestamp):
...
         if not name.startswith(self.machine_name_prefix):
             log.warning(f'event for unknown machine {name}')
             return
 
-        if event_subtype == 'compute.instances.insert':
+        if 'compute.instances.insert' in event_subtype:
             if event_type == 'GCE_OPERATION_DONE':
                 severity = event['severity']
-                operation_name = payload['operation']['name']
+                operation_type = 'insert'
```
jigold, line 702: @cseed Here's the hardcoded thing that didn't map between the old interface and the new one.

batch/batch/driver/instance_pool.py
```diff
@@ -657,49 +666,52 @@ async def handle_call_delete_event(self, instance, timestamp):
...
             if not instance:
                 log.warning(f'event for unknown instance {name}')
                 return
 
-            if event_subtype == 'compute.instances.preempted':
+            if 'compute.instances.preempted' in event_subtype:
                 log.info(f'event handler: handle preempt {instance}')
                 await self.handle_preempt_event(instance, timestamp)
-            elif event_subtype == 'compute.instances.delete':
+            elif 'compute.instances.delete' in event_subtype:
```
danking, line 714: All these are just prefixed with `v1.`, right? I'm inclined to write out out the exact match, seems cleaner, less likely to cause an issue later, etc.

batch/batch/driver/instance_pool.py
```diff
@@ -657,49 +666,52 @@ async def handle_call_delete_event(self, instance, timestamp):
...
 
         if not name.startswith(self.machine_name_prefix):
             log.warning(f'event for unknown machine {name}')
             return
 
-        if event_subtype == 'compute.instances.insert':
+        if 'compute.instances.insert' in event_subtype:
             if event_type == 'GCE_OPERATION_DONE':
                 severity = event['severity']
-                operation_name = payload['operation']['name']
```
danking, line 690: This is in protoPayload.response.operationType, I think.

batch/batch/driver/instance_pool.py
```diff
@@ -657,49 +666,52 @@ async def handle_call_delete_event(self, instance, timestamp):
...
             log.warning(f'unknown event resource type {resource_type}')
             return
 
-        event_type = payload['event_type']
-        event_subtype = payload['event_subtype']
-        resource = payload['resource']
-        name = resource['name']
+        if 'last' in event['operation']:
+            event_type = 'GCE_OPERATION_DONE'
+        elif 'first' in event['operation']:
```
danking, line 683: It looks like this is a boolean flag, seems better practice to do `event['operation'].get('first', False)`

batch/batch/driver/instance_pool.py
```diff
@@ -657,49 +666,52 @@ async def handle_call_delete_event(self, instance, timestamp):
...
+            event_type = 'GCE_API_CALL'
+        else:
+            log.warning(f'unknown event_type {event["operation"]}')
+            event_type = None
```
danking, line 687: We have the opportunity now to use more sensible names than these two. Maybe we can use `STARTED` and `COMPLETE` instead of GCE_API_CALL and GCE_OPERATION_DONE?

batch/batch/driver/instance_pool.py
danking: If this no longer varies, I'm inclined to simplify ZoneSuccessRate to not accept this argument at all.

batch/batch/driver/instance_pool.py
```diff
@@ -657,49 +666,52 @@ async def handle_call_delete_event(self, instance, timestamp):
...
+
+        event_subtype = payload['methodName']
+        resource = event['resource']
+        name = parse_resource_name(payload['resourceName'])['name']
```
danking, line 691: we can take this from `payload['request']['name']`, yeah?

danking (changes requested):
A few thoughts from looking at the logs

batch/batch/driver/instance_pool.py
jigold: I don't think so. It's only for certain entries.

```
wmecc-475:hail jigold$ jq -r '."protoPayload".request.name' ~/audit_log.txt
batch-worker-jigold-biyen
null
batch-worker-jigold-biyen
null

wmecc-475:hail jigold$ jq -r '."protoPayload".resourceName' ~/audit_log.txt
projects/hail-vdc/zones/us-central1-c/instances/batch-worker-jigold-biyen
projects/hail-vdc/zones/us-central1-c/instances/batch-worker-jigold-biyen
projects/hail-vdc/zones/us-central1-c/instances/batch-worker-jigold-biyen
projects/hail-vdc/zones/us-central1-c/instances/batch-worker-jigold-biyen
```

batch/batch/driver/instance_pool.py
jigold: @cseed -- can you please comment on this?

danking (dismissed):

jigold:
@danking bump. Seems fine to me

https://ci.hail.is/batches/106044/jobs/167

danking (approved):

PR #9440, opened 2020-09-11
emitI NDArrayQR

johnc1231:
After this and #9450 go in, all NDArray IR nodes will be moved to `emitI`, making it easier for me to update `NDArrayEmitter` to use `EmitCodeBuilder` and `PCode`. I cleaned up `NDArrayQR` a little bit in this PR, but all the low level LAPACK stuff means it's still kind of a mess.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -1049,13 +1049,13 @@ class Emit[C](
           val dataAddress = pndValue.pt.data.load(pndValue.tcode[Long])
 
           val tauPType = PCanonicalArray(PFloat64Required, true)
-          val tauAddress = mb.genFieldThisRef[Long]()
-          val workAddress = mb.genFieldThisRef[Long]()
-          val aAddressDGEQRF = mb.genFieldThisRef[Long]() // Should be column major
-          val rDataAddress = mb.genFieldThisRef[Long]()
-          val aNumElements = mb.genFieldThisRef[Long]()
+          val tauAddress = cb.newLocal[Long]("ndarray_qr_tauAddress")
+          val workAddress = cb.newLocal[Long]("ndarray_qr_workAddress")
+          val aAddressDGEQRF = cb.newLocal[Long]("ndarray_qr_aAddressDGEQRF") // Should be column major
+          val rDataAddress = cb.newLocal[Long]("ndarray_qr_rDataAddress")
+          val aNumElements = cb.newLocal[Long]("ndarray_qr_aNumElements")
 
-          val infoDGEQRFResult = mb.newLocal[Int]()
+          val infoDGEQRFResult = cb.newLocal[Int]("ndaray_qr_infoDGEQRFResult")
           val infoDGEQRFErrorTest = (extraErrorMsg: String) => (infoDGEQRFResult cne 0)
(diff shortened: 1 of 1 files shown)


johnc1231:
Bump

chrisvittal:
Thanks for the reminder. Looking now.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -993,6 +993,244 @@ class Emit[C](
...
+              pndValue.pt.data.pType.firstElementOffset(aAddressDGEQRF, aNumElements.toI),
+              LDA.toI,
+              tauPType.firstElementOffset(tauAddress, K.toI),
+              workAddress,
+              LWORK
+            ),
+
+            Code.invokeStatic1[Memory, Long, Unit]("free", workAddress.load()),
+            infoDGEQRFErrorTest("Failed to compute H and Tau.")
+          ))
```
chrisvittal, line 1071: You use `computeHAndTau` in both branches, this indicates that the setup could just exist on the CodeBuilder.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -993,6 +993,244 @@ class Emit[C](
...
+              val infoDORQRErrorTest = (extraErrorMsg: String) => (infoDORGQRResult cne 0)
+                .orEmpty(Code._fatal[Unit](const(s"LAPACK error DORGQR. $extraErrorMsg Error code = ").concat(infoDORGQRResult.toS)))
+
+              val qCondition = mb.genFieldThisRef[Boolean]()
+              val numColsToUse = mb.genFieldThisRef[Long]()
+              val aAddressDORGQR = mb.genFieldThisRef[Long]()
+
+              val qNumElements = mb.genFieldThisRef[Long]()
+
+              val computeCompleteOrReduced = Code(Code(FastIndexedSeq(
```
chrisvittal, line 1170: This is another long stretch of code that could largely be placed on the CodeBuilder.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -993,6 +993,244 @@ class Emit[C](
...
+              val qStridesBuilder = qPType.makeColumnMajorStridesBuilder(qShapeArray, mb)
+
+              val rNDArrayAddress = mb.genFieldThisRef[Long]()
+              val qDataAddress = mb.genFieldThisRef[Long]()
+
+              val infoDORGQRResult = mb.genFieldThisRef[Int]()
+              val infoDORQRErrorTest = (extraErrorMsg: String) => (infoDORGQRResult cne 0)
+                .orEmpty(Code._fatal[Unit](const(s"LAPACK error DORGQR. $extraErrorMsg Error code = ").concat(infoDORGQRResult.toS)))
+
+              val qCondition = mb.genFieldThisRef[Boolean]()
```
chrisvittal, line 1164: In new code, with method splitting, there should be no reason to generate fields.

chrisvittal (changes requested):
This code doesn't feel very imperative. If we're going to port it, we should also probably do things like using PCode/PValue methods and updating some of the fields we generate into locals (since method splitting is in).

johnc1231:
On the first pass, I had just moved things to CodeBuilder when I thought it improved readability. After your comments, I just moved everything.

chrisvittal (approved):
LGTM. Great work!

PR #9441, opened 2020-09-11
[batch] Change number test worker pool size

jigold:
The tests relying on Batch are getting slower because it takes a long time to download and build Docker images and we're putting more load on Batch. This will increase parallelism and reduce test failures due to timeouts.

### DIFF
--- a/batch/sql/change_test_worker_pool_size.py
+++ b/batch/sql/change_test_worker_pool_size.py
@@ -1,0 +1,26 @@
+import os
+import asyncio
+from gear import Database
+
+
+async def main():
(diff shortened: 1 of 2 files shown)


jigold:
Arcturus -- I'm assigning this to you, but please don't take off the WIP tag as merging this will cause Batch to shutdown (database migration). If we're not prepared for it, then it could cause an extended outage.

catoverdrive:
This basically limits the number of nodes in the worker pool for a test/dev deployment of batch to (default 3, max 4), right? This looks fine to me; should I approve this and let you take off the WIP tag when it's good to merge?

danking (approved):

PR #9442, opened 2020-09-11
[Benchmark] add autocomplete to compare

Dania-Abuhijleh:
Screen Shot 2020-09-11 at 4 06 56 PM
Screen Shot 2020-09-11 at 4 07 04 PM


benchmark-service/benchmark/benchmark.py
```diff
@@ -45,10 +45,12 @@ def get_benchmarks(app, file_path):
         stats['failed'] = d['failed']
         if not d['failed']:
             prod_of_means *= d['mean']
-            stats['f-stat'] = round(d['f-stat'], 6)
+            if d.get('f-stat') is not None:
```
jigold, line 48: Let's write a wrapper function for all of these values in utils.py in case more of them are not defined. Something like this:

```python
def round_if_defined(x):
    if x is not None:
        return round(x, 6)
    return None
```

```python
stats['p-value'] = round_if_defined(d.get('p-value'))
```

jigold (changes requested):
@danking Do you have suggestions to make the CSS / javascript not as duplicated between the lookup.html and compare.html files?

benchmark-service/benchmark/benchmark.py
```diff
@@ -41,17 +41,17 @@ def get_benchmarks(app, file_path):
     prod_of_means = 1
     for d in pre_data['benchmarks']:
         stats = dict()
-        stats['name'] = d['name']
-        stats['failed'] = d['failed']
+        stats['name'] = d.get('name')
+        stats['failed'] = d.get('failed')
         if not d['failed']:
-            prod_of_means *= d['mean']
+            prod_of_means *= d.get('mean', 1)
```
jigold, line 47: Let's move the benchmark fixes to another branch.

benchmark-service/benchmark/templates/compare.html
```diff
@@ -2,26 +2,31 @@
...
+<link rel="stylesheet" href="{{ base_path }}/static/css/autocomplete.css" />
 
 <h1>Benchmark System</h1>
 <div class="flex-col-align-left">
-    <form method="get" action="{{ base_path }}/compare">
+    <form autocomplete="off" method="get" action="{{ base_path }}/compare">
         <label for="file1">File 1:</label>
-        <input style="vertical-align:text-bottom;" name="file1" size=30 type="text" required
+        <div class="autocomplete">
+        <input style="vertical-align:text-bottom;" name="file1" class="autoComplete" size=30 type="text" required
```
jigold, line 12: You're missing the id field here.

benchmark-service/benchmark/static/js/autocomplete.js
```diff
@@ -0,0 +1,95 @@
+
+    new autoComplete({
```
jigold, line 2: What I was thinking for the deduplication is whether we could create a function that took some parameters that returned a new AutoComplete object/function. I'm not sure if that's possible.

@danking Can you please advise?

benchmark-service/benchmark/benchmark.py
```diff
@@ -67,9 +67,11 @@ def get_benchmarks(app, file_path):
 def get_comparisons(benchmarks1, benchmarks2, metric):
     def get_metric(data):
         if metric == 'median':
-            return data['median']
+            return data.get('median')
         assert metric == 'best'
-        return min(data['times'])
+        if data.get('times') is None:
```
jigold, line 72: I think this is clearer if it's

```
times = data.get('times')
if times:
    return min(times)
return None

**Review by jigold — CHANGES_REQUESTED** [review]

File: benchmark-service/benchmark/static/js/autocomplete.js [hunk 6]
```diff
@@ -0,0 +1,142 @@
+function auto_complete(place_holder, str_selector) {
```
danking, line 1: `str_selector` gives me flashbacks to [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation). Let's call it `input_selector` which is less about argument types and more about its purpose. I think in this context its clear it should be a string representation of a [CSS Selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors).

benchmark-service/benchmark/static/js/autocomplete.js
```diff
@@ -0,0 +1,142 @@
...
+                       if (a.match < b.match) return -1;
+                       if (a.match > b.match) return 1;
+                       return 0;
+                   },
+                   placeHolder: place_holder,     // Place Holder text                 | (Optional)
+                   selector: str_selector,           // Input field selector              | (Optional)
+                   threshold: 2,                        // Min. Chars length to start Engine | (Optional)
+                   debounce: 300,                       // Post duration for engine to start | (Optional)
+                   searchEngine: "strict",              // Search Engine type/mode           | (Optional)
+                   resultsList: {                       // Rendered results list object      | (Optional)
```
danking, line 17: Let's remove the comments here. Generally, Hail team avoids in-line comments like this because they get out of date quickly. That's because we don't maintain them, of course, but the effort of maintaining them doesn't seem worth it. If we need to understand these parameters, we can always check the [API reference for autocomplete.js](https://tarekraafat.github.io/autoComplete.js/#/?id=api-configuration).

benchmark-service/benchmark/static/js/autocomplete.js
```diff
@@ -0,0 +1,142 @@
...
+                       render: true,
+                       /* if set to false, add an eventListener to the selector for event type
+                          "autoComplete" to handle the result */
```
danking, line 20: and this comment too.

benchmark-service/benchmark/static/js/autocomplete.js
```diff
@@ -0,0 +1,142 @@
...
+                       result.innerHTML = "No Results";
+                       document.querySelector(".autoComplete_list").appendChild(result);
+                   },
+                   onSelection: feedback => {
+                   feedback.event.preventDefault();
+                   document.querySelector(str_selector).value = feedback.selection.value;
+             }
+           });
+}
+//    new autoComplete({
```
danking: and let's delete these once everything seems to be working correctly.

benchmark-service/benchmark/templates/compare.html
```diff
@@ -86,4 +91,14 @@ <h1>Benchmark System</h1>
...
 {% endif %}
 
+<script src="https://cdn.jsdelivr.net/npm/@tarekraafat/autocomplete.js@7.2.0/dist/js/autoComplete.min.js"></script>
+<script>
+    var files = {{ benchmark_file_list|tojson }};
+</script>
+<script src="{{ base_path }}/static/js/autocomplete.js"></script>
+<script>
+    auto_complete("gs://...", ".autoComplete");
+    auto_complete("gs://...", "#autoComplete2");
```
danking, line 101: On lines 100 and 101, we intend to refer to exactly one element in each case. When we want to refer to exactly one element, we usually use identifier selectors rather than class selectors because "identifiers" (kind of definitionally) refer to exactly one HTML element. In contrast, class selectors usually refer to many ("a class of") HTML elements.

Let's change line 100 to refer to use `#autoComplete1`

benchmark-service/benchmark/static/js/autocomplete.js
danking: And, I think this is a great resolution of the duplication, good work!

danking (changes requested):
This is looking pretty good, just a few minor comments!

jigold (approved):

PR #9443, opened 2020-09-14
[query] Make IEmitCode generic

patrick-schultz:
Nothing in `IEmitCode` depends on the value being a `PCode`. This PR makes `IEmitCode` a type alias, `IEmitCode = IEmitCodeGen[PCode]`, where `IEmitCodeGen` is a generic version. This supports any pattern where we generate missing and present branches, and the value can be anything to be used while generating the present branch, e.g. `Code`, `PCode`, `Array[PCode]`, `Stream`, or even another optional value (an `EmitCode` or `CodeBuilder => IEmitCode`.


johnc1231 (approved):
Nice, thanks Patrick.

PR #9444, opened 2020-09-14
NDArrayRef Better Error Message

johnc1231:
Extending the idea of better error messages from #9398  to include `NDArrayRef`.

### DIFF
--- a/hail/python/hail/ir/ir.py
+++ b/hail/python/hail/ir/ir.py
@@ -734,9 +734,13 @@ class NDArrayRef(IR):
         super().__init__(nd, *idxs)
         self.nd = nd
         self.idxs = idxs
+        self.save_error_info()
 
     def copy(self, *args):
         return NDArrayRef(args[0], args[1:])
+
+    def head_str(self):
+        return str(self._error_id)
 
     def _compute_type(self, env, agg_env):
         self.nd._compute_type(env, agg_env)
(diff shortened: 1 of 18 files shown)


hail/python/hail/ir/ir.py
```diff
@@ -729,14 +729,21 @@ def renderable_bindings(self, i, default_value=None):
...
-    def __init__(self, nd, idxs):
+    @typecheck_method(nd=IR, idxs=sequenceof(IR), error_id=nullable(int), stack_trace=nullable(str))
+    def __init__(self, nd, idxs, error_id=None, stack_trace=None):
         super().__init__(nd, *idxs)
         self.nd = nd
         self.idxs = idxs
+        self._error_id = error_id
+        self._stack_trace = stack_trace
+        if error_id is None or stack_trace is None:
+            self.save_error_info()
```
johnc1231, line 740: I'm not happy about having to change the constructor in Python like this. Ideally, we'd just let `self.save_error_info` in the constructor do all the work. However, when we copy IR around, we need to copy the error id and stack trace too. Open to suggestions about better ways to do this.

catoverdrive (approved):
Looks good! Couple of things leading to failing tests---you'll need to add the error id to ir.Pretty in Scala and one of the python tests is still looking for a FatalError.

I think putting these in the constructor is fine. I wonder if you could condense the logic if you intend to extend it to many more IRs by having it extend a `PotentiallyErroringIR` or something, but this also seems fine as-is.

PR #9447, opened 2020-09-15
[query] fix IEmitCode.consume

patrick-schultz:
`IEmitCode.consume`, with signature `consume[T](cb: EmitCodeBuilder, ifMissing: => Unit, ifPresent: (A) => T): T`, is wrong: it exports a value that is only defined in the present branch to be used after the two branches have joined.

The only use of this manually guarded the exported value by a bit remembering which branch was taken, but this is better handled by memoizing the `IEmitCode` in an `EmitValue`. Almost every other use of `consume` could be more simply expressed using `handle`, and moreover almost every use of `handle` could be more simply expressed using a method `get` which simply throws an exception if the missing branch is taken.

For those few cases where one really does want to consume an `IEmitCode` to a `Code` or `PCode`, with definitions given for each branch, I added `consumeCode` and `consumePCode`.


chrisvittal (approved):
Nice work! This is definitely a better API than what we had.

PR #9449, opened 2020-09-15
[query] Print file size on Table/MT write

tpoterba:
Next, I want to put file sizes into the metadata.json

### DIFF
--- a/hail/python/hail/experimental/vcf_combiner/vcf_combiner.py
+++ b/hail/python/hail/experimental/vcf_combiner/vcf_combiner.py
@@ -321,7 +321,7 @@ def calculate_new_intervals(ht, n, reference_genome):
     ht = ht.annotate(y=ht.x + 1)
     ht = ht.filter((ht.x // n != ht.y // n) | (ht.x == (n_rows - 1)))
     ht = ht.select()
-    ht = ht.annoatte(start=hl.or_else(
+    ht = ht.annotate(start=hl.or_else(
         hl.scan._prev_nonnull(hl.locus_from_global_position(ht.locus.global_position() + 1,
                                                             reference_genome=reference_genome)),
         hl.locus_from_global_position(0, reference_genome=reference_genome)))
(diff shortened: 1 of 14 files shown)


hail/src/main/scala/is/hail/utils/package.scala
```diff
@@ -200,18 +200,32 @@ package object utils extends Logging
...
+    } else if (absds < gib) {
+      div = mib
+      suffix = "MiB"
+    } else if (absds < tib) {
+      div = gib
+      suffix = "GiB"
+    } else {
+      div = tib
+      suffix = "TiB"
+    }
```
chrisvittal, line 226: Why not
```scala
val (div, suffix) = if (...)
  1 -> "B"
else if (...)
  kib -> "KiB"
// etc.
```

chrisvittal (changes requested):
Minor comments. Seems fine otherwise.

Intention with the metadata is to add another array that is the partition sizes in bytes?

hail/python/hail/experimental/vcf_combiner/vcf_combiner.py
```diff
@@ -321,7 +321,7 @@ def calculate_new_intervals(ht, n, reference_genome):
     ht = ht.annotate(y=ht.x + 1)
     ht = ht.filter((ht.x // n != ht.y // n) | (ht.x == (n_rows - 1)))
     ht = ht.select()
-    ht = ht.annotate(start=hl.or_else(
+    ht = ht.annoatte(start=hl.or_else(
```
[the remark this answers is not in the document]
tpoterba: huh, wonder how I even made this change!

hail/src/main/scala/is/hail/utils/package.scala
tpoterba: wanted to avoid the allocation, but this is probably cleaner.

tpoterba:
good comments, thanks! Addressed.

chrisvittal (approved):

PR #9450, opened 2020-09-15
NDArrayMatMul and NDArrayInv, emitI and NDArrayEmitter2

johnc1231:
This PR is a redo of #9438 that attempts to address Arcturus's comment about `NDArrayMatMul` using the old `NDArrayEmitter` interface. I created `NDArrayEmiter2` and moved `NDArrayMatMul` over to using that. I don't want this PR to get too much bigger, so after this one goes in I can make a second one that kills the original `NDArrayEmitter` and gets everything over to this new version.


johnc1231:
This got a little bit messy and is not easy to rebase. Closing in favor of two PRs, one each for `NDArrayInv` and `NDArrayMatMul`.

PR #9451, opened 2020-09-15
[aiogoogle, aiotools] added statfile, listfiles

cseed:
Current plan going forward for this:
 - I'm getting pretty happy with the interface (but maybe want to rename some things).  I think I will document what I have next.
 - Get rid of legacy hailtop.GCS.  This means we can get rid of all the client libraries except the oauth2 flows in auth.
 - The add a glob operation.
 - Add a copy/sync operation.

I might also add an S3AsyncFS.  I could have used that recently.

### DIFF
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -1,4 +1,5 @@
-from typing import Tuple, Any, Set, Optional, Mapping, Dict, AsyncIterator
+import os
+from typing import Tuple, Any, Set, Optional, Mapping, Dict, AsyncIterator, cast
 import asyncio
 import urllib.parse
 import aiohttp
@@ -114,10 +115,10 @@ class StorageClient(BaseClient):
             f'https://storage.googleapis.com/storage/v1/b/{bucket}/o/{urllib.parse.quote(name, safe="")}', **kwargs)
         return GetObjectStream(resp)
 
-    async def get_object_metadata(self, bucket: str, name: str, **kwargs) -> Any:
+    async def get_object_metadata(self, bucket: str, name: str, **kwargs) -> Dict[str, str]:
         params = kwargs.get('params')
         assert not params or 'alt' not in params
-        return await self.get(f'/b/{bucket}/o/{urllib.parse.quote(name, safe="")}', **kwargs)
+        return cast(Dict[str, str], await self.get(f'/b/{bucket}/o/{urllib.parse.quote(name, safe="")}', **kwargs))
 
     async def delete_object(self, bucket: str, name: str, **kwargs) -> None:
         await self.delete(f'/b/{bucket}/o/{urllib.parse.quote(name, safe="")}', **kwargs)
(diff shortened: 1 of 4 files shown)


hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -91,10 +122,42 @@ async def get_object_metadata(self, bucket: str, name: str, **kwargs) -> Any:
...
+
+    def is_file(self) -> bool:
+        return self._is_file
+
+    def is_dir(self) -> bool:
+        return not self._is_file
+
+    def status(self) -> FileStatus:
+        if self._status is None:
+            self._status = GetObjectFileStatus(self._items)
```
jigold, line 158: I think you're missing a return here.

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -128,6 +191,58 @@ async def create(self, url: str) -> WritableStream:
...
+
+            items = page.get('items')
+            if items:
+                for item in page['items']:
+                    yield GoogleStorageFileListEntry(f'gs://{bucket}/{item["name"]}', True, item)
+
+    def listfiles(self, url: str, recursive: bool = False) -> AsyncIterator[FileListEntry]:
+        bucket, name = self._get_bucket_name(url)
+        if not name.endswith('/'):
+            name = f'{name}/'
```
jigold, line 240: So this method assumes we're always passing a directory.

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -91,10 +122,42 @@ async def get_object_metadata(self, bucket: str, name: str, **kwargs) -> Any:
...
 
+class GetObjectFileStatus(FileStatus):
+    def __init__(self, items):
+        self._items = items
+
+    async def size(self) -> int:
+        return int(self._items['size'])
+
+    async def __getitem__(self, key: str) -> Any:
+        return self._items[key]
```
jigold, line 140: self._items can be None in the case of a directory.

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -128,6 +191,58 @@ async def create(self, url: str) -> WritableStream:
...
+                for prefix in prefixes:
+                    url = f'gs://{bucket}/{prefix}'
+                    if url.endswith('/'):
+                        url = url[:-1]
+                    yield GoogleStorageFileListEntry(url, False, None)
+
+            items = page.get('items')
+            if items:
+                for item in page['items']:
+                    yield GoogleStorageFileListEntry(f'gs://{bucket}/{item["name"]}', True, item)
```
jigold, line 235: This will allow an object that ends in a slash. Maybe we just ignore it as if it didn't exist here?

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -128,6 +191,58 @@ async def create(self, url: str) -> WritableStream:
...
+            'delimiter': '/',
+            'includeTrailingDelimiter': 'true'
+        }
+        async for page in await self._storage_client.list_objects(bucket, params=params):
+            prefixes = page.get('prefixes')
+            if prefixes:
+                for prefix in prefixes:
+                    url = f'gs://{bucket}/{prefix}'
+                    if url.endswith('/'):
+                        url = url[:-1]
```
jigold, line 229: So is the contract here that every GoogleStorageFileListEntry must not end in a '/'

hail/python/test/hailtop/test_aiogoogle.py
```diff
@@ -171,3 +171,60 @@ async def test_get_object_headers():
...
+@pytest.mark.asyncio
+async def test_statfile_directory(filesystem):
+    fs, base = filesystem
+
+    await fs.mkdir(f'{base}/dir')
+    await fs.touch(f'{base}/dir/foo')
+
+    with pytest.raises(FileNotFoundError):
+        # statfile raises FileNotFound on directories
+        await fs.statfile(f'{base}/dir')
```
jigold, line 193: Can you also add a check for `{base}/dir/`?

hail/python/test/hailtop/test_aiogoogle.py
```diff
@@ -171,3 +171,60 @@ async def test_get_object_headers():
...
+    await fs.touch(foobar)
+    await fs.mkdir(f'{base}/foo')
+    await fs.touch(a)
+    await fs.mkdir(f'{base}/foo/b')
+    await fs.touch(c)
+
+    async def listfiles(dir, recursive):
+        return {(entry.url(), entry.is_file()) async for entry in fs.listfiles(dir, recursive)}
+
+    assert await listfiles(f'{base}/foo', recursive=True) == {(a, True), (c, True)}
```
jigold, line 229: Can you also add tests for `{base}/foo/`?

jigold (changes requested):

hail/python/hailtop/aiotools/fs.py
```diff
@@ -12,6 +13,40 @@
...
+    @abc.abstractmethod
+    async def size(self) -> int:
+        pass
+
+    @abc.abstractmethod
+    async def __getitem__(self, key: str) -> Any:
+        pass
+
+
+class FileListEntry(abc.ABC):
```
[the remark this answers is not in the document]
jigold: I think FileListEntry is fine. I'm not sure what I was thinking earlier.

cseed (commented):
OK, I think I addressed all your comments.  Two points I clarified:

1. FileStatus only refers to files.  It can come from two places: statfile, which throws not found if the path doesn't exist and is a real file, or from FileListEntry.entry, which throws ValueError if the entry is not a file.

2. Paths which are directories always have a trailing slash.  If a function (e.g. listfiles) can only take a slash, the slash is optional (like on the traditional UNIIX interface).  In the current interface, you always know if a path is referring to directory or a file, so all incoming trailing slashes are optional.  Returned paths that are directories will always have a trailing slash.

I'm not totally convinced by this interface.  Let's see what happens when we get to implementing copy/sync method.

jigold (approved):

PR #9456, opened 2020-09-16
[query] Add tests for local backend

tpoterba:


build.yaml
```diff
@@ -1139,6 +1139,216 @@ steps:
      - batch_pods_ns
      - hail_run_image
      - build_hail
+ - kind: runImage
```
tpoterba, line 1142: changes to build.yaml involve copying the python test tasks and adding 
```
export HAIL_QUERY_BACKEND=local
```

hail/python/test/hail/helpers.py
```diff
@@ -127,6 +128,19 @@ def wrapper(func, *args, **kwargs):
...
+
+    @decorator
+    def wrapper(func, *args, **kwargs):
+        if isinstance(hl.utils.java.Env.backend(), LocalBackend):
+            with pytest.raises(BaseException):
+                return func(*args, **kwargs)
+        else:
+            return func(*args, **kwargs)
+
+    return wrapper
```
patrick-schultz, line 142: It looks like there's an official pytest decorator to mark a test as expected to fail, which the test reports treat specially. https://docs.pytest.org/en/latest/skipping.html#xfail-mark-test-functions-as-expected-to-fail

Using that, I think you could write this
```
fails_local_backend = pytest.mark.fail(isinstance(hl.utils.java.Env.backend(), LocalBackend), reason="doesn't yet work on local backend", strict=True)
```

build.yaml
```diff
@@ -1139,6 +1139,216 @@ steps:
...
+     tar xzf test.tar.gz
+     tar xzf resources.tar.gz
+     tar xzf data.tar.gz
+     tar xvf wheel-container.tar
+     python3 -m pip install --no-dependencies hail-*-py3-none-any.whl
+     export HAIL_TEST_RESOURCES_DIR=./resources
+     export HAIL_DOCTEST_DATA_DIR=./data
+     export HAIL_TEST_BUCKET=hail-test-dmk9z
+     export GOOGLE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
+     export PYSPARK_SUBMIT_ARGS="--driver-memory 6g pyspark-shell"
```
patrick-schultz, line 1161: I'm guessing some of these aren't needed in local mode. Is it worth deleting them to keep things tidy?

patrick-schultz (changes requested):
Nice! Just a couple comments.

build.yaml
tpoterba: oh, I can delete the pyspark one

hail/python/test/hail/helpers.py
patrick-schultz: Thoughts on this? It seems nice to have the pytest report list separate counts for “expected to fail” and “unexpectedly passing”.

tpoterba: oh, nice! I couldn't find this when I was poking around for something better.

tpoterba: using xfail now, way better!

build.yaml
```diff
@@ -1242,7 +1236,6 @@ steps:
      export HAIL_DOCTEST_DATA_DIR=./data
      export HAIL_TEST_BUCKET=hail-test-dmk9z
      export GOOGLE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
-     export PYSPARK_SUBMIT_ARGS="--driver-memory 6g pyspark-shell"
```
patrick-schultz: I'm assuming it's a mistake that only `test_hail_python_local_backend_2` still has HAIL_TEST_BUCKET and GOOGLE_APPLICATION_CREDENTIALS?

patrick-schultz (dismissed):

patrick-schultz (approved):

Issue #9188, opened 2020-07-31
StringExpression should have join function


cseed:
`<string>.join(<str-collection>)` should work like in Python.

edit: fixed join argument to be string collection.

konradjk:
You mean CollectionExpression yeah? I agree but one minor complication is that python’s is `delimiter.join(list)` not `list.join(delimiter)` and that’s a bit more awkward in hail (since you then have to use a hail string rather than pythons)

cseed:
Sorry, I wrote the example code backwards.  It should match Python: it should be a method on StringExpression that takes a collection of strings.

johnc1231:
So my worry is that everyone is going to try to do:

```
" ".join(my_hail_array)
```

and be very confused when it doesn't work. It's not often they have to coerce a python value into a hail expression explicitly.

johnc1231:
I guess that's no worse than the current situation though, where they probably also try and do that.

tpoterba:
This doesn't belong in a github issue. I'll move to Asana.

PR #9459, opened 2020-09-17
Add join method to StringExpression

nawatts:
Requested in #9188.

This is similar to [hl.delimit](https://github.com/hail-is/hail/blob/6b5d7a5cde246c5caf53b97c5b17d7e7c6440fec/hail/python/hail/expr/functions.py#L4293-L4326). However, unlike hl.delimit, this throws a TypeError if a collection with a non-string element type is provided. This matches the behavior of Python's [str.join](https://docs.python.org/3/library/stdtypes.html#str.join).

### DIFF
--- a/hail/python/hail/expr/expressions/typed_expressions.py
+++ b/hail/python/hail/expr/expressions/typed_expressions.py
@@ -5,7 +5,7 @@ from .base_expression import Expression, ExpressionException, to_expr, \
     unify_all, unify_types
 from .expression_typecheck import coercer_from_dtype, \
     expr_any, expr_array, expr_set, expr_bool, expr_numeric, expr_int32, \
-    expr_int64, expr_str, expr_dict, expr_interval, expr_tuple
+    expr_int64, expr_str, expr_dict, expr_interval, expr_tuple, expr_oneof
 from hail.expr.types import HailType, tint32, tint64, tfloat32, \
     tfloat64, tbool, tcall, tset, tarray, tstruct, tdict, ttuple, tstr, \
     tndarray, tlocus, tinterval, is_numeric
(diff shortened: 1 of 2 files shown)


hail/python/hail/expr/expressions/typed_expressions.py
```diff
@@ -2808,6 +2808,35 @@ def reverse(self):
...
+        ----------
+        collection : :class:`.ArrayExpression` or :class:`.SetExpression`
+            Collection.
+
+        Returns
+        -------
+        :class:`.StringExpression`
+            Joined string expression.
+        """
+        if collection.dtype.element_type != tstr:
```
tpoterba, line 2835: should we convert elements to string automatically if of another type? Would let you do `hl.str(',').join([1,2,3])`

hail/python/hail/expr/expressions/typed_expressions.py
```diff
@@ -2808,6 +2808,35 @@ def reverse(self):
...
+            raise TypeError(f"Expected str collection, {collection.dtype.element_type} found")
+
+        return collection._method("mkString", tstr, self)
```
tpoterba, line 2838: prefer to just call `hl.delimit`, since then we can change the implementation in once place.

hail/python/hail/expr/expressions/typed_expressions.py
tpoterba: not necessary though, up to you.

tpoterba (changes requested):

tpoterba:
thanks Nick!

hail/python/hail/expr/expressions/typed_expressions.py
nawatts: I started out having this call `hl.delimit`, but added the element type check to make it behave more like Python's [string.join](https://docs.python.org/3/library/stdtypes.html#str.join), which raises a TypeError if passed an iterable with non-string elements.

tpoterba: I think checking first and then calling delimit() would replicate Python functionality?

johnc1231: Regardless of whether we do the element type check or not, we should still call `hl.delimit` afterwards as opposed to `mkString`, to minimize how often we leak java implementation details.

nawatts: After thinking about it some more, I like the type check as a guard rail. Automatic type conversion would be useful for things like joining an array of numbers. On the other hand, it would also let you accidentally join an array of structs when it's more likely you wanted to join from one of the struct's fields.

tpoterba (approved):

PR #9462, opened 2020-09-17
[batch] do not explicitly set cpu in BPE, which improves test time

danking:
The `BatchPoolExecutor` assumes, incorrectly, that the default number of CPUs for
a batch job is 1. This is not true in test and development environment.

I avoid explicitly setting the value if it is `None`. This choice preserves the
default value in this environment. I set the thread limit to 1 if the CPU is `None`.


jigold (approved):

danking:
I think fixing this will mask an issue where a 1 CPU job blocks the scheduling of smaller jobs. I'm gonna debug that first, then remove WIP.

PR #9464, opened 2020-09-17
[batch] Fix deadlock with exceeded shares

jigold:
I noticed that jobs in test deployments were deadlocking because we weren't spinning up extra instances (compared to the production version of Batch). Although each job could fit on an open instance, its allocated share is still less than the core request for that job. This PR aims to increase the probability in which we ignore an exceed shares error the more we have these errors such that at a certain point the rate will be 100% and we'll be able to continue scheduling.


danking (approved):
LGTM

PR #9469, opened 2020-09-17
linear_regression_rows_nd using TableMapPartitions

johnc1231:
In this PR, I rewrite `linear_regression_rows_nd` to use `_map_partitions` instead of `_group_within_partitions`. By doing this, I've eliminated the need to do a `key_by` at the end of `linear_regression_rows_nd`. I also think this makes the code clearer. 

This PR also makes a few seemingly random changes that are actually bug fixes:

1. When emitting `Apply` nodes, we were grabbing the `Code[Region]` from the first argument to the `MethodBuilder`. However, the assumption that the first argument will always be a `Region` seems to no longer be true. As such, we just construct a `CodeParam` from the `StagedRegion` we have available. 

2. In the NDArrayEmitter, I want to make sure I call the local `emit` method that passes off to `emitWithRegion`, for the same reason as 1: (Can't trust first argument to be a `Region`)

3. In `EmitStream`, I need to use `memoizeField` instead of `memoize`, because regular `memoize` saves to a `LocalRef`, and that will get reset to 0 when `next` is called on a stream. Lesson: don't trust locals for things that must live between elements of a stream.

I feel like you have a better idea of how the Stream stuff gets emitted than I do Patrick. I'm curious if what I wrote in `process_block` could be written in a way that would lead to better code getting emitted, as I still need to figure out how to squeeze more performance out of this.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -2564,7 +2564,7 @@ class Emit[C](
 
               val bodyEnv = env.bind(lName, lElemRef)
                 .bind(rName, rElemRef)
-              val bodyt = dEmit(body, bodyEnv)//emitSelf.emit(body, elemMB, bodyEnv, None)
+              val bodyt = dEmit(body, bodyEnv)
 
               val lIdxVars2 = NDArrayEmitter.zeroBroadcastedDims2(elemMB, idxVars, nDims, leftChildEmitter.outputShape)
               val rIdxVars2 = NDArrayEmitter.zeroBroadcastedDims2(elemMB, idxVars, nDims, rightChildEmitter.outputShape)
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -2016,7 +2016,7 @@ class Emit[C](
         val vars = args.map { a => coerce[Any](mb.newLocal()(typeToTypeInfo(a.pType))) }
         val ins = vars.zip(codeArgs.map(_.v)).map { case (l, i) => l := i }
         val value = Code(Code(ins), meth.invokeCode[Any](
-          ((mb.getCodeParam[Region](1): Param) +: vars.map(_.get: Param)): _*))
+          (CodeParam(region.code) +: vars.map(_.get: Param)): _*))
         strict(pt, value, codeArgs: _*)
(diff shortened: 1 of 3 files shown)


johnc1231:
`linear_regression_rows_nd` benchmark on my laptop went from 125 seconds to 62 seconds. So this made a 2x difference. The regular linear regression benchmark is still like 25 seconds though, have to figure out next steps to close that gap.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -2513,6 +2513,8 @@ class Emit[C](
     def emit(ir: IR, env: E = env): EmitCode =
       this.emitWithRegion(ir, mb, region, env, None)
 
+    def dEmit(ir: IR, env: E = env): EmitCode = emit(ir, env)
```
patrick-schultz, line 2516: What's this? It's just an alias of `emit`.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -2555,12 +2558,13 @@ class Emit[C](
...
           new NDArrayEmitter[C](lP.shape.pType.size, shapeArray, lP.shape.pType, body.pType, setupShape, setupMissing, leftChildEmitter.missing || rightChildEmitter.missing) {
             override def outputElement(elemMB: EmitMethodBuilder[C], idxVars: IndexedSeq[Value[Long]]): Code[_] = {
+              assert(elemMB == mb)
               val lElemRef = elemMB.newPresentEmitField(lName, lP.elementType)
               val rElemRef = elemMB.newPresentEmitField(rName, rP.elementType)
 
               val bodyEnv = env.bind(lName, lElemRef)
                 .bind(rName, rElemRef)
-              val bodyt = emitSelf.emit(body, elemMB, bodyEnv, None)
+              val bodyt = dEmit(body, bodyEnv)//emitSelf.emit(body, elemMB, bodyEnv, None)
```
patrick-schultz, line 2567: Stray comment

patrick-schultz (dismissed):
Very cool! Just a couple things to tidy up.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
johnc1231: There's an `emit` method on `NDArrayEmitter`, so I needed a different name to use within `outputElement`.

patrick-schultz (approved):

PR #9471, opened 2020-09-17
Use NDArray Version of linear_regression_rows when not on SparkBackend

johnc1231:
SparkBackend is the only one that can actually use the Scala/Breeze version of `linear_regression_rows`, right? Otherwise, we need to use the new ndarray version.

### DIFF
--- a/hail/python/hail/methods/statgen.py
+++ b/hail/python/hail/methods/statgen.py
@@ -308,7 +308,7 @@ def linear_regression_rows(y, x, covariates, block_size=16, pass_through=()) -> 
     -------
     :class:`.Table`
     """
-    if isinstance(Env.backend(), SparkBackend):
+    if not isinstance(Env.backend(), SparkBackend):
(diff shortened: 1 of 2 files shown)


hail/python/hail/methods/statgen.py
```diff
@@ -307,6 +308,9 @@ def linear_regression_rows(y, x, covariates, block_size=16, pass_through=()) ->
     -------
     :class:`.Table`
     """
+    if isinstance(Env.backend(), SparkBackend):
```
tpoterba, line 311: shouldn't this be `if not`?

tpoterba (changes requested):

hail/python/hail/methods/statgen.py
johnc1231: Yep, duh. Thanks. That would have been not so good.

tpoterba (approved):

PR #9472, opened 2020-09-17
[batch] use a slim image for BPE

danking:
The old BPE images (which still exist) are based on the full python image. The
slim image reduces the size by about 750MB. This should substantially improve
image pull on the n1-standards. I have seen image pull for this image take 136
seconds before in a test job.

### DIFF
--- a/hail/python/hailtop/batch/batch_pool_executor.py
+++ b/hail/python/hailtop/batch/batch_pool_executor.py
@@ -142,7 +142,7 @@ class BatchPoolExecutor:
         if image is None:
             if version.major != 3 or version.minor not in (6, 7, 8):
                 raise ValueError(
-                    'You must specify an image if you are using a Python version other than 3.6, 3.7, or 3.8 (you are using {version})')
+                    f'You must specify an image if you are using a Python version other than 3.6, 3.7, or 3.8 (you are using {version})')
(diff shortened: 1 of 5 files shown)


hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -139,7 +139,7 @@ def __init__(self, *,
         self.finished_future_count = 0
         self._shutdown = False
         version = sys.version_info
-        self.image = image or f'hailgenetics/python-dill:{version.major}.{version.minor}'
+        self.image = image or f'hailgenetics/python-dill:{version.major}.{version.minor}-slim'
```
jigold, line 142: Can we add the slim image Docker files to the repository? Should we also assert here what the available version numbers are?

jigold (changes requested):

hail/python/hailtop/batch/batch_pool_executor.py
danking: None of these docker files are in the repository. I'll add them to docker, but its been a long outstanding task to have a solution for building and deploying docker images to docker hub; though that's now in question given the docker hub changes.

danking: I just build them locally and push them.

jigold (approved):

PR #9473, opened 2020-09-17
[batch] use the gcr repo for the cloud sdk

danking:
Also update all uses of cloud sdk to same version.

Seems good to do anyway. I hope the newer version of gcloud includes a faster gsutil. I also hope that using gcr will improve the pull time a bit.

### DIFF
--- a/batch/build-batch-worker-image-startup.sh
+++ b/batch/build-batch-worker-image-startup.sh
@@ -30,7 +30,7 @@ rm -rf /var/lib/apt/lists/*
 [ -f /etc/docker/daemon.json ] || echo "{}" > /etc/docker/daemon.json
 
 VERSION=1.5.0
-OS=linux 
+OS=linux
 ARCH=amd64
 
 curl -fsSL "https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v${VERSION}/docker-credential-gcr_${OS}_${ARCH}-${VERSION}.tar.gz" \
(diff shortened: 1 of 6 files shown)


batch/build-batch-worker-image-startup.sh
```diff
@@ -40,7 +40,7 @@ curl -fsSL "https://github.com/GoogleCloudPlatform/docker-credential-gcr/release
 docker-credential-gcr configure-docker
 
 docker pull ubuntu:18.04
-docker pull google/cloud-sdk:269.0.0-alpine
+docker pull gcr.io/google.com/cloudsdktool/cloud-sdk:310.0.0-alpine
```
jigold, line 43: Let's make a new disk image and update the instance creation code.

jigold (changes requested):

batch/build-batch-worker-image-startup.sh
danking: OK, good, I also want to fold in some stuff from isntance_pool.py in here (like the daemon.json) to try and speed up worker start up.

I have to work on hail team and ATGU meeting prep, but will return to this next week.

danking:
I moved a couple things over into the image itself rather than on startup. Saves us a restart of docker and a restart of fluentd. I think that'll save ~20 seconds of worker startup.

jigold (changes requested):
Something failed with the docker rules:

```
________________ Test.test_verify_no_access_to_metadata_server _________________

self = <test.test_batch.Test testMethod=test_verify_no_access_to_metadata_server>

    def test_verify_no_access_to_metadata_server(self):
        builder = self.client.create_batch()
        j = builder.create_job(os.environ['HAIL_CURL_IMAGE'],
                               ['curl', '-fsSL', 'metadata.google.internal', '--max-time', '10'])
        builder.submit()
        status = j.wait()
>       assert status['state'] == 'Failed', status
E       AssertionError: {'batch_id': 69, 'cost': '$0.0000', 'duration': 8664, 'exit_code': 0, ...}
E       assert 'Success' == 'Failed'
E         - Success
E         + Failed
```

danking:
OK, this should be good.

jigold (approved):

PR #9474, opened 2020-09-18
Add first and last methods to ArrayExpression

nawatts:
Add `first` method to ArrayExpression to replace `head`. Add `last` method as the counterpart to `first`.

### DIFF
--- a/hail/python/hail/expr/expressions/typed_expressions.py
+++ b/hail/python/hail/expr/expressions/typed_expressions.py
@@ -546,7 +546,7 @@ class ArrayExpression(CollectionExpression):
         >>> hl.eval(names.filter(lambda x: x.startswith('D')).first())
         None
         """
-        return hl.or_missing(hl.len(self) > 0, self[0])
+        return hl.rbind(self, lambda x: hl.or_missing(hl.len(x) > 0, x[0]))
 
     def last(self):
         """Returns the last element of the array, or missing if empty.
@@ -565,7 +565,7 @@ class ArrayExpression(CollectionExpression):
         >>> hl.eval(names.filter(lambda x: x.startswith('D')).last())
         None
         """
-        return hl.rbind(hl.len(self), lambda n: hl.or_missing(n > 0, self[n - 1]))
+        return hl.rbind(self, hl.len(self), lambda x, n: hl.or_missing(n > 0, x[n - 1]))
 
     @typecheck_method(x=oneof(func_spec(1, expr_any), expr_any))
     def index(self, x):
--- a/hail/python/hail/expr/expressions/typed_expressions.py
+++ b/hail/python/hail/expr/expressions/typed_expressions.py
@@ -546,6 +546,7 @@ class ArrayExpression(CollectionExpression):
         >>> hl.eval(names.filter(lambda x: x.startswith('D')).first())
         None
         """
+        # FIXME: this should generate short-circuiting IR when that is possible
         return hl.rbind(self, lambda x: hl.or_missing(hl.len(x) > 0, x[0]))
 
     def last(self):
--- a/hail/python/hail/expr/expressions/typed_expressions.py
+++ b/hail/python/hail/expr/expressions/typed_expressions.py
@@ -529,7 +529,26 @@ class ArrayExpression(CollectionExpression):
         # FIXME: this should generate short-circuiting IR when that is possible
         return hl.rbind(self, lambda x: hl.case().when(x.length() > 0, x[0]).or_missing())
 
-    def tail(self):
+    def first(self):
+        """Returns the first element of the array, or missing if empty.
+
+        Returns
+        -------
+        :class:`.Expression`
+            Element.
+
+        Examples
+        --------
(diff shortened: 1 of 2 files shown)


hail/python/hail/expr/expressions/typed_expressions.py
```diff
@@ -529,6 +529,25 @@ def head(self):
...
+        Examples
+        --------
+        >>> hl.eval(names.head())
+        'Charlie'
+
+        If the array has no elements, then the result is missing:
+        >>> hl.eval(names.filter(lambda x: x.startswith('D')).tail())
+        None
+        """
+        return hl.rbind(self, hl.len(self), lambda x, n: hl.or_missing(n > 0, x[n - 1]))
```
nawatts, line 549: Blindly following the implementation of `head` with `hl.rbind(self, ...`. Not sure why that is necessary.

johnc1231:
I guess we already have `head` on arrays, but I think these would be better called `first` and `last`, given that Table.head takes an argument for how many elements to grab

nawatts:
Good point @johnc1231, it would be nice to have the same interface for `head`/`tail` on tables and arrays. Instead of adding `tail`, should we add `first`/`last` and deprecate `head`?

johnc1231:
We just had a chat about this in Dev: https://hail.zulipchat.com/#narrow/stream/123011-Hail-Dev/topic/head.2Ftail.20vs.20first.2Flast/near/210522900

To summarize, I think `head` shouldn't exist, neither should `tail`. Konrad likes the idea of having something that does what current `head` does, and I think a good name for that is `first`. `last` would correspond to that. I'm somewhat surprised people want this behavior often enough that `hl.or_else(hl.len(x) > 0, x[0])` or `hl.or_else(hl.len(x) > 0, x[-1])` is insufficient.

johnc1231:
We have a couple of deprecated things now, I think I'm going to go add a deprecation decorator so that people get warnings.

nawatts:
Sounds good. Updated this PR to add `first` and `last` methods.

nawatts:
Also, if this addition is approved, could someone comment on whether the `hl.rbind(self, ...` used in `head` is necesssary.
https://github.com/hail-is/hail/blob/e6ec5a2952b8a2145a758e6bf30f0b7d3bcad7be/hail/python/hail/expr/expressions/typed_expressions.py#L529-L530

tpoterba:
> Also, if this addition is approved, could someone comment on whether the hl.rbind(self, ... used in head is necessary?

Should leave it for now, I think. Our common subexpression elimination pass would insert a binding, but it's not a bad idea to insert one explicitly in library code.

The comment in the head() implementation refers to the fact that in many cases, we don't need to bind the entire array (suppose it's `range(1000000).map(lambda x: something_expensive(x))` if we're just going to use the first element.

nawatts:
Restored the binding. Not sure what to do about this:

> The comment in the head() implementation refers to the fact that in many cases, we don't need to bind the entire array (suppose it's range(1000000).map(lambda x: something_expensive(x)) if we're just going to use the first element.

tpoterba:
yeah, we don't have a way to avoid constructing the full array right now

hail/python/hail/expr/expressions/typed_expressions.py
```diff
@@ -529,6 +529,44 @@ def head(self):
...
+        Examples
+        --------
+        >>> hl.eval(names.first())
+        'Alice'
+
+        If the array has no elements, then the result is missing:
+        >>> hl.eval(names.filter(lambda x: x.startswith('D')).first())
+        None
+        """
+        return hl.rbind(self, lambda x: hl.or_missing(hl.len(x) > 0, x[0]))
```
johnc1231, line 549: I'd copy the `FIXME` comment down here, since it's still something we should address. I'd also then change `head` to just call `first`. That the logic only exists once, and any tests that may use `head` will now be tests for `first`.

johnc1231 (changes requested):
You need to add a test to ensure `last` works. You'll find a method called `test_array_head` in `test_expr.py`. I'd change that to `test_array_first`, call `first` instead. Then I'd add `test_array_last` and make sure there's a test for empty and nonempty arrays. Thanks

hail/python/hail/expr/expressions/typed_expressions.py
nawatts: Done (fa35a4a3920c205c1c78fc46fb696c570c3baf36) and done (15bcc78bd1a3abc62e395b8d3aba7849a85a0951)

nawatts:
> You need to add a test to ensure `last` works. You'll find a method called `test_array_head` in `test_expr.py`. I'd change that to `test_array_first`, call `first` instead. Then I'd add `test_array_last` and make sure there's a test for empty and nonempty arrays. Thanks

Tests were included in e2fcfb61a5df5ab497269d32c12a5d5fb652ddd4. Are there more checks that should be added to those?

johnc1231:
You're good on tests, I was looking at the wrong thing.

johnc1231 (approved):

PR #9475, opened 2020-09-18
[query] fix performance bug in lir.SimplifyControl

patrick-schultz:
This was a good one :)

I haven’t measured the performance difference, but clearly the bug defeated the purpose of using a union-find data structure in the first place, which was to reduce the complexity of `unify` from quadratic to linear.

While I was here, I made a separate simplification. Now that the sets are being unioned as intended, each set contains exactly one block that doesn’t start with a `GotoX`, and that block is the final target of all blocks in the set. That observation allows a simplification when computing the `rootFinalTarget` map.

### DIFF
--- a/hail/src/main/scala/is/hail/lir/SimplifyControl.scala
+++ b/hail/src/main/scala/is/hail/lir/SimplifyControl.scala
@@ -107,7 +107,7 @@ class SimplifyControl(m: Method) {
     for (b <- blocks) {
       if (b.first != null &&
         b.first.isInstanceOf[GotoX]) {
-        u.sameSet(
+        u.union(
           blocks.index(b),
           blocks.index(b.first.asInstanceOf[GotoX].L))
       }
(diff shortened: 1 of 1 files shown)


hail/src/main/scala/is/hail/lir/SimplifyControl.scala
```diff
@@ -107,18 +107,16 @@ class SimplifyControl(m: Method) {
...
+        u.union(
           blocks.index(b),
           blocks.index(b.first.asInstanceOf[GotoX].L))
       }
     }
 
     val rootFinalTarget = mutable.Map[Int, Block]()
     blocks.indices.foreach { i =>
-      val r = u.find(i)
-      if (r == i) {
```
patrick-schultz, line 119: For context without needing to go look at `UnionFind`: `sameSet(Int, Int): Boolean` tests if two elements are in the same partition, while `union(Int, Int): Unit` actually combines the two partitions. So this first loop was doing nothing, and the `(r == i)` test at the bottom was always true.

tpoterba (approved):

PR #9480, opened 2020-09-18
NDArrayMatMul EmitI

johnc1231:
I broke up #9450. This PR moves just `NDArrayMatMul` to `emitI`, and does so by introducing a new, simpler `NDArrayEmitter` called `NDArrayEmitter2`. In a subsequent PR, I'll delete the old `NDArrayEmitter` entirely and move everything over to new version.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -880,6 +880,133 @@ class Emit[C](
           }
         }
 
+      case NDArrayMatMul(lChild, rChild) =>
+        emitNDArrayColumnMajorStrides(lChild).flatMap(cb) { case leftPCode: PNDArrayCode =>
+          emitNDArrayColumnMajorStrides(rChild).flatMap(cb) { case rightPCode: PNDArrayCode =>
+            val lPType = leftPCode.pt
+            val rPType = rightPCode.pt
+
+            val leftPVal = leftPCode.memoize(cb, "left_ndarray_matmul")
+            val rightPVal = rightPCode.memoize(cb, "right_ndarray_matmul")
+
+            val lShape = leftPVal.shapes()
+            val rShape = rightPVal.shapes()
+
+            val unifiedShape = NDArrayEmitter.matmulShape(cb, lShape, rShape)
+
+            val leftBroadcastMask = if (lPType.nDims > 2) NDArrayEmitter.broadcastMask(lShape) else IndexedSeq[Value[Long]]()
+            val rightBroadcastMask = if (rPType.nDims > 2) NDArrayEmitter.broadcastMask(rShape) else IndexedSeq[Value[Long]]()
+
+            val outputPType = PCanonicalNDArray(lPType.elementType, TNDArray.matMulNDims(lPType.nDims, rPType.nDims), true)
+
+            if ((lPType.elementType.isInstanceOf[PFloat64] || lPType.elementType.isInstanceOf[PFloat32]) && lPType.nDims == 2 && rPType.nDims == 2) {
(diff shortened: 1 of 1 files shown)


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -880,6 +880,133 @@ class Emit[C](
...
+                }
+              )
+              val res = outputPType.construct(
+                outputPType.makeShapeBuilder(IndexedSeq(M, N)),
+                outputPType.makeColumnMajorStridesBuilder(IndexedSeq(M, N), mb),
+                answerPArrayAddress,
+                mb,
+                region.code)
+
+              EmitCode(Code._empty, false, PCode(pt, res)).toI(cb)
```
catoverdrive, line 964: I think this can be `IEmitCode.present(cb, PCode(pt, res))` to avoid the intermediate `EmitCode` object

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -3074,3 +3053,68 @@ abstract class NDArrayEmitter[C](
...
+  def emit(cb: EmitCodeBuilder, targetType: PNDArray, region: Value[Region]): IEmitCode = {
+    outputShape.map(cb){ shapeArray: IndexedSeq[Value[Long]] =>
+      val dataSrvb = new StagedRegionValueBuilder(cb.emb, targetType.data.pType, region)
+      val dataAddress: Code[Long] =
+        Code(
+          dataSrvb.start(targetType.numElements(shapeArray, cb.emb).toI),
+          emitLoops(cb, shapeArray, targetType.elementType, dataSrvb),
+          dataSrvb.end())
+
+      def shapeBuilder(srvb: StagedRegionValueBuilder): Code[Unit] = {
```
catoverdrive, line 3072: `IEmitCode` things shouldn't contain things of type `Code[Unit]` --- they should be functions of type `Unit` that add the instructions directly to the CodeBuilder.

catoverdrive (changes requested):

hail/src/main/scala/is/hail/expr/ir/Emit.scala
johnc1231: I'm not sure what you want me to change here. These `StagedRegionValueBuilder => Code[Unit]` functions are constructed this way to pass to `StagedRegionValueBuilder.addBaseStruct`, which expects a function of that type. Updating how SRVBs work seems potentially out of scope for this PR.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
johnc1231: Agree that's better, just didn't know the function to do this. Thanks

johnc1231:
Addressed or responded with a question.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
catoverdrive: Ah, I see. That comment was both for this and `emitLoops`, which I think can get changed to add code to `cb` directly. I agree, updating SRVB seems out of scope for this PR. But if I'm not mistaken, one of the things we were talking about doing was phasing out SRVB in favor of constructors on the PType, so having a non-SRVB-based constructor signature on PNDArray would be nice (not necessary for this PR).

catoverdrive (dismissed):

johnc1231:
I changed `emitLoops` to directly add to `cb`. I'll handle the SRVB thing in a subsequent PR.

johnc1231:
Bump, think this is good to go.

catoverdrive (approved):

PR #9482, opened 2020-09-18
Deprecate ArrayExpression's head method

nawatts:
Follow up to #9474.


johnc1231 (dismissed):

johnc1231:
This looks fine, but I need to confirm we don't use `head` in any other code before approving

nawatts:
Admittedly, I did not do a thorough search, but this was the only `git grep '\.head('` result that looked like ArrayExpression head instead of Table/MatrixTable head.

johnc1231 (approved):
I agree this is the only usage, thanks!

PR #9483, opened 2020-09-18
[query] Encapsulate TabixLineIterator buffering logic

chrisvittal:
Split it out into BGZipLineReader.


chrisvittal:
This isn't quite right, the new reader doesn't return `null` at end of file properly, will fix.

chrisvittal:
What needed to get fixed for this is in #9485 , the whole reason why this PR exists is #9485. Closing in favor of it.

PR #9484, opened 2020-09-19
[query] add ServiceCacheableFS and use in code/context caching

catoverdrive:


danking:
What's the status of this PR? It looks like its awaiting review but also needs conflicts resolved?

danking (changes requested):
@catoverdrive Can you review my rebase to ensure I didn't make any mistakes? If it looks good to you, let's get this in.

danking (approved):

danking:
@catoverdrive bump, this is gonna supercharge the effectiveness of query service

danking:
@catoverdrive bump, let's get this in this week

danking:
I'm planning to remove flags in the service because they have bad idempotency properties. Is there a reason to not make the cached FS the default one?

danking (changes requested):
xx

PR #9485, opened 2020-09-19
[query] Quick and dirty tabix creation/writing for export_vcf

chrisvittal:
Adds the long requested feature for tabix creation support to export_vcf.

This represents an 80% solution due to the requirement that the file already exist in order to write the index. Leans entirely on htsjdk for index creation.

There is a limitation where index writing is only supported for the export type of `CONCATENATED`, but this should be easy to add to.

CHANGELOG : Teach tabix writing to `export_vcf`, experimental.

### DIFF
--- a/hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
+++ b/hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
@@ -224,7 +224,7 @@ object ExportVCF {
     getAttributes(k1, attributes).flatMap(_.get(k2))
 
   def apply(ctx: ExecuteContext, mv: MatrixValue, path: String, append: Option[String],
-    exportType: String, metadata: Option[VCFMetadata]) {
+    exportType: String, metadata: Option[VCFMetadata], tabix: Boolean = false) {
     val fs = ctx.fs
 
     mv.typ.requireColKeyString()
@@ -460,6 +460,18 @@ object ExportVCF {
         sb.result()
       }
     }.writeTable(ctx, path, Some(header), exportType = exportType)
+
+    if (tabix) {
+      exportType match {
+        case ExportType.CONCATENATED =>
+          info(s"Writing tabix index for $path")
+          TabixVCF(fs.broadcast, path)
+        case ExportType.PARALLEL_SEPARATE_HEADER
+          | ExportType.PARALLEL_COMPOSABLE
+          | ExportType.PARALLEL_HEADER_IN_SHARD =>
+            throw new UnsupportedOperationException("tabix on parallel output is not supported")
+      }
+    }
   }
 }
 
--- a/hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
+++ b/hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
@@ -469,15 +469,13 @@ object ExportVCF {
         case ExportType.PARALLEL_SEPARATE_HEADER | ExportType.PARALLEL_HEADER_IN_SHARD =>
           val localFsBc = ctx.fsBc
           val files = fs.glob(path + "/part-*").map(_.getPath.getBytes)
-          info(s"Writing tabix index for $path")
+          info(s"Writing tabix index for ${ files.length } in $path")
           ctx.backend.parallelizeAndComputeWithIndex(ctx.backendContext, files) { (pathBytes, _) =>
             TabixVCF(localFsBc, new String(pathBytes))
-            null
+            Array.empty
           }
         case ExportType.PARALLEL_COMPOSABLE =>
           warn("Writing tabix index for `parallel=composable` is not supported. No index will be written.")
-        case _ =>
-          fatal(s"Unknown export type: $exportType")
       }
     }
   }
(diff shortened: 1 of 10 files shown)


chrisvittal:
Some thoughts:

- This should probably be the default
- Export table is spark dependent. It shouldn't be.
- For the `CONCATENATED` case, we're probably not going to do better than this.
- For the parallel cases, even if we don't handle indexing while the files are being written (which would be the best option), It would be nice to have a `parallelize` kind of functionality that doesn't rely on spark.
    - This is less important though since we still use RDD for `export_table` and `export_vcf`.

chrisvittal:
After sleeping on how I want the interface to be in general, I think this is ready for review. Thanks for your patience.

hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
```diff
@@ -453,5 +460,52 @@ object ExportVCF {
...
+          val localFsBc = ctx.fsBc
+          val files = fs.glob(path + "/part-*").map(_.getPath.getBytes)
+          info(s"Writing tabix index for $path")
+          ctx.backend.parallelizeAndComputeWithIndex(ctx.backendContext, files) { (pathBytes, _) =>
+            TabixVCF(localFsBc, new String(pathBytes))
+            null
+          }
+        case ExportType.PARALLEL_COMPOSABLE =>
+          warn("Writing tabix index for `parallel=composable` is not supported. No index will be written.")
+        case _ =>
```
tpoterba, line 479: probably unnecessary -- not a user error, match error should be fine.

hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
```diff
@@ -453,5 +460,52 @@ object ExportVCF {
...
+        case ExportType.PARALLEL_SEPARATE_HEADER | ExportType.PARALLEL_HEADER_IN_SHARD =>
+          val localFsBc = ctx.fsBc
+          val files = fs.glob(path + "/part-*").map(_.getPath.getBytes)
+          info(s"Writing tabix index for $path")
+          ctx.backend.parallelizeAndComputeWithIndex(ctx.backendContext, files) { (pathBytes, _) =>
+            TabixVCF(localFsBc, new String(pathBytes))
+            null
+          }
+        case ExportType.PARALLEL_COMPOSABLE =>
+          warn("Writing tabix index for `parallel=composable` is not supported. No index will be written.")
```
tpoterba, line 478: I think it's OK to throw an exception here.

hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
```diff
@@ -453,5 +460,52 @@ object ExportVCF {
...
+        case ExportType.CONCATENATED =>
+          info(s"Writing tabix index for $path")
+          TabixVCF(ctx.fsBc, path)
+        case ExportType.PARALLEL_SEPARATE_HEADER | ExportType.PARALLEL_HEADER_IN_SHARD =>
+          val localFsBc = ctx.fsBc
+          val files = fs.glob(path + "/part-*").map(_.getPath.getBytes)
+          info(s"Writing tabix index for $path")
+          ctx.backend.parallelizeAndComputeWithIndex(ctx.backendContext, files) { (pathBytes, _) =>
+            TabixVCF(localFsBc, new String(pathBytes))
+            null
```
tpoterba, line 475: how does this work? A quick read of the parallelizeAndComputeWithIndex indicates this may be problematic

tpoterba: Maybe Array.empty[Byte] instead?

hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
```diff
@@ -453,5 +460,52 @@ object ExportVCF {
...
+
+    if (tabix) {
+      exportType match {
+        case ExportType.CONCATENATED =>
+          info(s"Writing tabix index for $path")
+          TabixVCF(ctx.fsBc, path)
+        case ExportType.PARALLEL_SEPARATE_HEADER | ExportType.PARALLEL_HEADER_IN_SHARD =>
+          val localFsBc = ctx.fsBc
+          val files = fs.glob(path + "/part-*").map(_.getPath.getBytes)
+          info(s"Writing tabix index for $path")
```
tpoterba, line 472: can we make this more descriptive:

```
          info(s"Writing tabix index for ${files.length} files in $path")

```

tpoterba (changes requested):
looks great. Few comments on the control flow.

hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
chrisvittal: Yeah, nonnull would probably be better here.

hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
chrisvittal: Sure!

hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
chrisvittal: I think it's also OK to not. NACK.

hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
chrisvittal: Especially since if we get here, the mode has already been validated, removed.

tpoterba (approved):
nice job, Chris!

PR #9486, opened 2020-09-21
[query/lir] assert postcondition of SimplifyControl

patrick-schultz:
… and make it true.

Randomly assigned. John, I know you haven't looked at lir much; happy to answer any questions.

### DIFF
--- a/hail/src/main/scala/is/hail/lir/SimplifyControl.scala
+++ b/hail/src/main/scala/is/hail/lir/SimplifyControl.scala
@@ -73,6 +73,14 @@ class SimplifyControl(m: Method) {
 
           // if there is one parent, it is L
           q += L
+
+          // if L is now a trivial jump, revisit parents to let them jump over L
+          if (L.first eq L.last) {
+            if (L eq m.entry)
+              m.setEntry(finalTarget(L))
+            else
+              L.uses.foreach(q += _._1.parent)
+          }
         }
 
       case x: GotoX =>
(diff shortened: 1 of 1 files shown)


johnc1231:
I mostly understand. Can you clarify what the `_entry` of a `Block` is?

patrick-schultz:
That's the `entry` of a `Method` (a `Block` has a `first` and `last` `StmtX`). It is the entry block of the method. The blocks of a method are precisely those that are reachable from the entry block by following control edges.

`entry` is showing up in the statement of invariants because it's a bit of a special case: for most blocks `b`, if there are no control edges tageting `b` (if `b.uses` is empty), then `b` is unreachable, and hence is dead code. But there is no control edge from a method to its entry block, so the entry block is always live, even if it is targeted by no control edges.

patrick-schultz:
So the line `case g: GotoX => g.L.uses.size > 1 || (g.L eq m.entry)` is saying that the target of a goto should be the target of at least one other control edge (otherwise the block ending in the goto and the target block should have been combined), *unless* the target is the entry block, in which case there's the unaccounted for implicit control edge from the method to the entry block, so the goto can very well be the only real control edge leading in.

johnc1231 (approved):
Thanks for explaining, that makes sense to me. Knowing that the `entry` has no `uses` clarifies all of the checks about it.

This PR is also mostly setting new invariants / asserts, so if all our tests are passing I feel pretty good about correctness.

PR #9487, opened 2020-09-21
[query] Add EncodedLiteral node to avoid java serialization

johnc1231:
Introduces `EncodedLiteral`, a literal that is stored as an `EType` encoded array of bytes. 

I did this with the intention of making PCA faster, but so far this seems to take roughly the same amount of time. Currently this is only used by `_blanczos_pca`, since that's the only code using `IR._persist`

### DIFF
--- a/hail/src/main/scala/is/hail/backend/spark/SparkBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/spark/SparkBackend.scala
@@ -346,15 +346,6 @@ class SparkBackend(
     }
   }
 
-//  def executeLiteral(ir: IR): IR = {
-//    val t = ir.typ
-//    assert(t.isRealizable)
-//    val (value, timings) = execute(ir, optimize = true)
-//    timings.finish()
-//    timings.logInfo()
-//    Literal.coerce(t, value)
-//  }
-
   def executeLiteral(ir: IR): IR = {
     val t = ir.typ
     assert(t.isRealizable)
@@ -363,7 +354,7 @@ class SparkBackend(
       log.info(s"starting execution of query $queryID} of initial size ${ IRSize(ir) }")
       val (retVal, timer) = _execute(ctx, ir, true)
       val literalIR = retVal match {
-        case Left(x) => Literal(TVoid, x)
+        case Left(x) => throw new HailException("Can't create literal")
         case Right((pt, addr)) => GetFieldByIdx(EncodedLiteral.hailValueToByteArray(pt, addr, ctx), 0)
       }
 
--- a/hail/src/main/scala/is/hail/backend/spark/SparkBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/spark/SparkBackend.scala
@@ -346,13 +346,33 @@ class SparkBackend(
     }
   }
 
+//  def executeLiteral(ir: IR): IR = {
+//    val t = ir.typ
+//    assert(t.isRealizable)
+//    val (value, timings) = execute(ir, optimize = true)
+//    timings.finish()
+//    timings.logInfo()
+//    Literal.coerce(t, value)
+//  }
+
   def executeLiteral(ir: IR): IR = {
     val t = ir.typ
     assert(t.isRealizable)
-    val (value, timings) = execute(ir, optimize = true)
-    timings.finish()
-    timings.logInfo()
-    Literal.coerce(t, value)
+    val (literalIR, timer) = withExecuteContext() { ctx =>
+      val queryID = Backend.nextID()
+      log.info(s"starting execution of query $queryID} of initial size ${ IRSize(ir) }")
+      val (retVal, timer) = _execute(ctx, ir, true)
+      val literalIR = retVal match {
+        case Left(x) => Literal(TVoid, x)
+        case Right((pt, addr)) => GetFieldByIdx(EncodedLiteral.hailValueToByteArray(pt, addr, ctx), 0)
+      }
+
+      log.info(s"finished execution of query $queryID")
+      (literalIR, timer)
+    }
+    timer.finish()
+    timer.logInfo()
+    literalIR
   }
 
   def executeJSON(ir: IR): String = {
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -695,11 +695,6 @@ class Emit[C](
       case x@Literal(t, v) =>
         presentPC(mb.addLiteral(v, x.pType))
       case x@EncodedLiteral(codec, value) =>
-//        val storedBytes = mb.addLiteral(value.ba, PCanonicalBinaryRequired).asInstanceOf[PBinaryValue]
-//        val stagedIs = Code.newInstance[ByteArrayInputStream, Array[Byte]](storedBytes.loadBytes())
-//        val ib = cb.newLocal[InputBuffer]("encoded_literal_ib")
-//        cb.assign(ib, codec.buildCodeInputBuffer(stagedIs))
-//        val pc = codec.buildEmitDecoder(x.typ, cb.emb.ecb)(region.code, ib)
         presentPC(mb.addEncodedLiteral(x))
       case True() =>
         presentC(const(true))
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -695,6 +695,7 @@ class Emit[C](
       case x@Literal(t, v) =>
         presentPC(mb.addLiteral(v, x.pType))
       case x@EncodedLiteral(codec, value) =>
+        assert(x.pType == codec.decodedPType())
         presentPC(mb.addEncodedLiteral(x))
       case True() =>
         presentC(const(true))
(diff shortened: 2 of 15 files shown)


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -694,6 +694,13 @@ class Emit[C](
             Class.forName("is.hail.expr.ir.package$"), "uuid4"))))
       case x@Literal(t, v) =>
         presentPC(mb.addLiteral(v, x.pType))
+      case x@EncodedLiteral(codec, value) =>
```
tpoterba, line 697: clever, but we can't do this. This means that every usage of the literal will decode the encoded value -- think if about if we have:

```
StreamMap
  StreamRange idx 0 100 1
  ArrayRef 
    Ref idx
    EncodedLiteral(some array)
```

The literal will be decoded for every element.

I think the right solution is to extend the underlying infrastructure that encodeLiterals() calls to allow for an Array[Array[Byte]] that are decoded separately into separate fields on the class. This means that this line:

https://github.com/hail-is/hail/blob/f48f6fc56f8febcf6d1b87aa9ab89c43f67487d5/hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala#L318

Is going to have to decode an Array[Array[Byte]] into multiple fields, and so on.

hail/src/main/scala/is/hail/expr/ir/IR.scala
```diff
@@ -91,6 +91,21 @@ final case class Literal(_typ: Type, value: Annotation) extends IR {
   // require(SafeRow.isSafe(value))
 }
 
+object EncodedLiteral {
+
+  def hailValueToByteArray(pt: PType, addr: Long, ctx: ExecuteContext): IR = {
```
tpoterba, line 96: nice.

hail/src/main/scala/is/hail/expr/ir/InferPType.scala
```diff
@@ -120,7 +120,7 @@ object InferPType {
     }
     node._pType = node match {
       case x if x.typ == TVoid => PVoid
-      case _: I32 | _: I64 | _: F32 | _: F64 | _: Str | _: UUID4 | _: Literal | _: True | _: False
+      case _: I32 | _: I64 | _: F32 | _: F64 | _: Str | _: UUID4 | _: Literal | _: EncodedLiteral | _: True | _: False
```
tpoterba, line 123: encodedLiteral should have its own rule that calls cidec.decodedPType(), even if the two will usually be the same. Should add an assertion in the emitter that the decodedPType() is equal to the ptype of the node.

hail/src/main/scala/is/hail/backend/spark/SparkBackend.scala
```diff
@@ -346,13 +346,33 @@ class SparkBackend(
...
-    val (value, timings) = execute(ir, optimize = true)
-    timings.finish()
-    timings.logInfo()
-    Literal.coerce(t, value)
+    val (literalIR, timer) = withExecuteContext() { ctx =>
+      val queryID = Backend.nextID()
+      log.info(s"starting execution of query $queryID} of initial size ${ IRSize(ir) }")
+      val (retVal, timer) = _execute(ctx, ir, true)
+      val literalIR = retVal match {
+        case Left(x) => Literal(TVoid, x)
```
tpoterba, line 366: this is invalid. Can't have a void typed literal.

tpoterba: The assertion above (t.isRealizable) guarantees this code is dead, too.

tpoterba (changes requested):
nice turnaround! a few comments to address and then we can test this out.

hail/src/main/scala/is/hail/expr/ir/IR.scala
```diff
@@ -91,6 +91,21 @@ final case class Literal(_typ: Type, value: Annotation) extends IR {
...
+    val etype = EType.defaultFromPType(pt)
+    val codec = TypedCodecSpec(etype, pt.virtualType, BufferSpec.defaultUncompressed)
+    val bytes = codec.encode(ctx, pt, addr)
+    EncodedLiteral(codec, bytes)
+  }
+}
+
+final case class EncodedLiteral(codec: TypedCodecSpec, value: Array[Byte]) extends IR {
+  !CanEmit(codec.encodedVirtualType)
```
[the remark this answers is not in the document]
tpoterba: Also, need to define equals() / hashCode() so that the value is appropriately handled. Add a test for this as well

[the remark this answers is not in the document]
johnc1231: I thought when you do a `case class` you get a good definition of `hashCode` and `equals`?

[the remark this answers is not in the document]
johnc1231: I just copied that assertion from regular `Literal`. I think the idea is just that we shouldn't be constructing a literal of a type that has its own special IR node for constants, like `Str` or `Int32`, etc.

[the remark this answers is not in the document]
tpoterba: yeah, but it's just based on the case class fields, and the equals() / hashCode() for Array is based on reference, not value. I actually think the better solution is to make a `class WrappedByteArray(b: Array[Byte])` which defines these, and use that as the EncodedLiteral argument.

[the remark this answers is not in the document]
tpoterba: > I just copied that assertion from regular Literal. I think the idea is just that we shouldn't be constructing a literal of a type that has its own special IR node for constants, like Str or Int32, etc.

Ah, right. Missing the `require()` though :D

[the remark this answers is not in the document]
johnc1231: Ah, remembered how the case class one worked, but forgot that default Java array equality is just references. I'll do what you suggested with `WrappedByteArray`.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
johnc1231: I don't quite understand yet. We are talking about decoding literals to separate fields, but then you talked about `encodeLiterals`. It seems like that method encodes all the literals currently stored in the `literalsMap`. I have an `[Array[Byte]]` that lives outside of that on the IR node. The current emit rule adds the `Array[Byte]` into that map, but encoding that would be weird, as it's already encoded.

johnc1231: Oh! Is encoding a `PCanonicalBinary` a no-op, since it's already just bytes?

tpoterba: I think I often don't communicate my design ideas effectively, and I'm trying to improve! Here's another shot:

encodeLiterals() takes the `literalsMap`, wraps it in a tuple, encodes it to bytes, broadcasts the bytes, and decodes the bytes on the worker.

The function we want to call to emit EncodedLiteral just should do the last two: broadcast the bytes, decode the bytes on the worker. This means I think we need to extend/abstract the machinery used here so that in addition to tracking a `literalsMap`, we also track encoded literals: `mutable.Map[mutable.Map[EncodedLiteral, PSettable]`. `encodeLiterals()` should return an `Array[Array[Byte]]` (say the 0th element is the current return value built literalsMap) with the 1-Nth elements being arrays taken directly from the EncodedLiterals. The code emitted into `mb2` in encodeLiterals() should also decode into the PSettables inside the encoded literals map.

Is this any clearer? Still not describing it super well.

johnc1231: Totally clear, makes sense. Thanks!

johnc1231:
I believe I've addressed all of your comments now. The test failure is some spurious batch thing. I'll rerun when it's approved.

tpoterba (changes requested):
Sorry for the slow review. Just needs a few things:

1. Test for node == node (wrapped byte array equality)
2. Need a rule in IRParser and Pretty. I think this node shouldn't be parsable right now (we may not need every node to have string serializability, especially ones used internally in the compiler) but at least the type should be printed in Pretty.

johnc1231:
Ok, added test and rules. Back to you.

hail/src/main/scala/is/hail/expr/ir/Requiredness.scala
```diff
@@ -445,6 +445,8 @@ class Requiredness(val usesAndDefs: UsesAndDefs, ctx: ExecuteContext) {
 
       case _: NA => requiredness.union(false)
       case Literal(t, a) => requiredness.unionLiteral(a)
+        //TODO: Ask if I did this one right, not sure.
```
tpoterba, line 448: remove

hail/src/main/scala/is/hail/expr/ir/InferPType.scala
```diff
@@ -120,12 +120,14 @@ object InferPType {
...
```
tpoterba, line 123: I think this needs to be removed

tpoterba (dismissed):
minor touches and ready to go

hail/src/main/scala/is/hail/expr/ir/InferPType.scala
johnc1231: Yes, good catch.

tpoterba (approved):

johnc1231:
Should be all set

PR #9488, opened 2020-09-21
[scorecard] Added Asana integration

jigold:
- Use the scorecard tag or design-docs tag to have Asana tasks show up in scorecard
- Easy to add different tags to track
- Currently using my personal access token. Not sure if that's ideal, but seemed fine for now.

### DIFF
--- a/scorecard/scorecard/scorecard.py
+++ b/scorecard/scorecard/scorecard.py
@@ -17,10 +17,8 @@ from gear import setup_aiohttp_session, web_maybe_authenticated_user
 from web_common import setup_aiohttp_jinja2, setup_common_static_routes, render_template
 
 from types import TracebackType
-from typing import Any, Optional, TypeVar, Type
+from typing import Any, Optional, Type
 
-
-ClientType = TypeVar('ClientType', bound='AsanaClient')
 
 log = logging.getLogger('scorecard')
 
@@ -93,7 +91,7 @@ class AsanaClient:
             await self._session.close()
             self._session = None
 
-    async def __aenter__(self: ClientType) -> ClientType:
+    async def __aenter__(self: 'AsanaClient') -> 'AsanaClient':
         return self
 
     async def __aexit__(self,
--- a/scorecard/scorecard/scorecard.py
+++ b/scorecard/scorecard/scorecard.py
@@ -61,7 +61,7 @@ timestamp = None
 class AsanaClient:
     _session: Optional[ClientSession]
 
-    def __init__(self, *, token: Optional[str] = None, **kwargs):
+    def __init__(self, *, token: Optional[str] = None):
         self._base_url = 'https://app.asana.com/api/1.0'
 
         if token is None:
@@ -334,7 +334,7 @@ def get_issue_data(repo_name, issue):
 
 async def get_asana_data(asana_client):
     result = defaultdict(lambda: [])
-    for tag_name, tag_gid in asana_tags.items():
+    for _, tag_gid in asana_tags.items():
         tag_assignees = await get_tag_assignees(asana_client, tag_gid)
         for assignee, data in tag_assignees.items():
             result[assignee].extend(data)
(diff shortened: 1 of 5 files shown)


scorecard/scorecard/scorecard.py
```diff
@@ -28,12 +36,97 @@
...
+
+asana_to_github_user = {
+    'Jackie Goldstein': 'jigold',
+    'Cotton Seed': 'cseed',
+    'Daniel King': 'danking',
+    'John Compitello': 'johnc1231',
+    'Tim Poterba': 'tpoterba',
+    'Arcturus Wang': 'catoverdrive',
+    'Dania Abuhijleh': 'Dania-Abuhijleh',
+    'Patrick Schultz': 'patrick-schultz'
```
jigold, line 52: I know I'm missing some people. Will add.

scorecard/scorecard/scorecard.py
```diff
@@ -28,12 +36,97 @@
...
 timestamp = None
 
 
+class AsanaClient:
+    _session: Optional[ClientSession]
+
+    def __init__(self, *, session: Optional[ClientSession] = None, **kwargs):
+        self._base_url = 'https://app.asana.com/api/1.0'
+        if session is None:
+            session = ClientSession(**kwargs)
```
jigold, line 67: This is based on the aiogoogle client. We should probably deduplicate.

scorecard/scorecard/scorecard.py
```diff
@@ -28,12 +36,97 @@
...
+
+async def get_tag_assignees(client, tag_gid):
+    assignees = defaultdict(lambda: [])
+
+    async def get_task_data(task_gid):
+        data = await client.get(f'/tasks/{task_gid}')
+        return data['data']
+
+    result = await client.get(f'/tags/{tag_gid}/tasks')
+    task_data = await asyncio.gather(*[get_task_data(task['gid'])
```
jigold, line 108: We have a rate limit of API calls with Asana of either 1500 or 150 per minute depending on the plan we have. I think the gather is okay for now. Otherwise, we might need to rate limit this.

scorecard/scorecard/scorecard.py
```diff
@@ -28,12 +36,97 @@
...
+        async with await self._session.delete(
+                f'{self._base_url}{path}', **kwargs) as resp:
+            return await resp.json()
+
+    async def close(self) -> None:
+        if self._session is not None:
+            await self._session.close()
+            self._session = None
+
+    async def __aenter__(self: ClientType) -> ClientType:
```
danking, line 90: Usually `self` doesn't have a type, right? I think the return type can be `'AsanaClient'`, no?

scorecard/scorecard/scorecard.py
```diff
@@ -278,13 +397,25 @@ async def on_startup(app):
     gh_client = gidgethub.aiohttp.GitHubAPI(session, 'scorecard', oauth_token=token)
     app['gh_client'] = gh_client
 
-    await update_data(gh_client)
-    asyncio.ensure_future(poll(gh_client))
+    asana_token_file = os.environ.get('ASANA_TOKEN_PATH', '/asana/token.txt')
+    with open(asana_token_file, 'r') as f:
+        token = f.read().strip()
+    asana_session = ClientSession(headers={'Authorization': f'Bearer {token}'})
+    asana_client = AsanaClient(session=asana_session)
```
danking, line 404: I think the interface makes a bit more sense for the AsanaClient to take a `token` argument and if the token argument is `None`, it looks up the token as you do here. This will keep all the Asana-related stuff in one place, the AsanaClient. If we ever want to use Asana elsewhere, we can easily move that class.

danking (dismissed):
Just a few things

scorecard/scorecard/scorecard.py
```diff
@@ -28,12 +36,97 @@
...
+        gid = task['gid']
+        name = task['name']
+        assignee = task['assignee']
+        if assignee:
+            assignee_name = asana_to_github_user.get(assignee['name'])
+        else:
+            assignee_name = None
+        permalink_url = task['permalink_url']
+
+        if assignee_name and not task['completed']:
```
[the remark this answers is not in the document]
jigold: This is a good idea. but let's get the basic thing working now.

scorecard/scorecard/scorecard.py
jigold: This is based on the aiogoogle client Cotton wrote.

danking (approved):

PR #9489, opened 2020-09-21
[query/lir] findBlocks asserts well-formed blocks

patrick-schultz:
This PR asserts a stronger precondition for `Method.findBlocks`: all reachable blocks must be non-empty, must be terminated by a control expression, and all targets of the control expression must be non-null.


patrick-schultz:
Folding into #9491

PR #9490, opened 2020-09-21
[query/lir] remove dead code on the fly

patrick-schultz:
Stacked on #9489 

This PR enforces a stronger invariant on `lir.Block`: a block can never contain dead code, i.e. code that follows a control statement. Prior to this, dead code could be added to a block, only to be removed later. But it's just as easy to have `Block.append` check if a statement to be added would be dead, and just not add it in the first place. And having the stronger invariant always hold allows for some simplifications in the code, in addition to a simpler mental model.


patrick-schultz:
Folding into #9491

PR #9491, opened 2020-09-21
[query/lir] simplify Classx.asBytes

patrick-schultz:
~~Stacked on #9490~~

With the stronger invariants enforced by the previous PRs, some of the work done in `asBytes` is redundant.

### DIFF
--- a/hail/src/main/scala/is/hail/lir/X.scala
+++ b/hail/src/main/scala/is/hail/lir/X.scala
@@ -50,37 +50,7 @@ class Classx[C](val name: String, val superName: String) {
     classes += this
 
     for (m <- methods) {
-      val blocks = m.findBlocks()
-
-      val locals = m.findLocals(blocks)
-      for (l <- locals) {
-        // FIXME parameters, too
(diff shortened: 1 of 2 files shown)


hail/src/main/scala/is/hail/lir/X.scala
```diff
@@ -400,17 +382,20 @@ class Block {
 
   def append(x: StmtX): Unit = {
     assert(x.parent == null)
+    x.parent = this
     if (last == null) {
       first = x
       last = x
-    } else {
```
chrisvittal, line 406: Does a check like this need to be added to prepend as well?

chrisvittal (commented):
Just one question, otherwise, I found this straightforward.

hail/src/main/scala/is/hail/lir/X.scala
patrick-schultz: Oh, you're right, if a `ControlX` is prepended we need to drop all of the existing statements, though I'd be surprised if we ever do that. Will fix.

chrisvittal (approved):

PR #9493, opened 2020-09-23
[hailctl] Fix config file name

chrisvittal:
It was previously called config.yaml, when it was actually a python
configparser config file, which is a subset of [ini syntax].

[ini syntax]: https://docs.python.org/3/library/configparser.html

Check to see if the .yaml file exists and the .ini file does not exist.
If so, silently rename the config file to the new name.

### DIFF
--- a/hail/python/hailtop/config/user_config.py
+++ b/hail/python/hailtop/config/user_config.py
@@ -1,12 +1,14 @@
 import os
 import configparser
 
+from pathlib import Path
+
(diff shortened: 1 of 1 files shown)


chrisvittal:
Tested locally. Renames `config.yaml` to `config.ini`  and then otherwise operates as normal for set/get/unset.

danking (approved):
Thanks!

PR #9495, opened 2020-09-23
[hail] fix hl.import_plink docs

danking:
CHANGELOG: Fix hl.import_plink docs to properly report the type of `is_case` and `quant_pheno`.

I verified the correct types by checking `LoadPlink.scala`.


johnc1231 (approved):
Both obviously wrong, thanks for fixing.

PR #9503, opened 2020-09-24
[query] Fix NDArrays of Non-numerics: Part 1

johnc1231:
CHANGELOG: Fixed bug where making NDArrays of non-numeric types would fail. Non-numeric ndarrays still cannot be collected to python though. 

NDArrays of non numeric types are broken, have been for a while. No one seems to use them for that currently, so it hasn't been an issue, but I suspect with `dndarray` or BlockedMatrixTable experiments it's going to be desirable. 

This PR starts to address that problem by doing the following:

1. `checkedConvertFrom`, which only supported primitive arrays, is replaced with the more flexible `copyFromType`. As this was the only use of `checkedConvertFrom`, I removed it altogether. 

2. Add tests that show that it's now possible to make an ndarray of non-numeric types, so long as the only things that get returned in python are numbers.

The remaining problems all involve conversions to numpy. If you never convert to numpy, things should be fine:

1. I need to get strides out of the Java ndarray representation. Strides make no sense for non-numeric objects after converting from Java to Python. We say the size of a required tuple of 3 int32's is 12 bytes, but that's not going to be the size of the python object

2. Strings are tricky too, since the numpy string dtype comes with a max length, so we'll have to do a pass over the strings to figure out how large the largest one is before converting.

### DIFF
--- a/hail/python/hail/expr/types.py
+++ b/hail/python/hail/expr/types.py
@@ -655,6 +655,7 @@ class tndarray(HailType):
 
     def _convert_from_json(self, x):
         if is_numeric(self._element_type):
+            np_type = self.element_type.to_numpy()
             return np.ndarray(shape=x['shape'], buffer=np.array(x['data'], dtype=np_type), strides=x['strides'], dtype=np_type)
         else:
             raise TypeError("Hail cannot currently return ndarrays of non-numeric or boolean type.")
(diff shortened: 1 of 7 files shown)


patrick-schultz:
Did you confirm that the behaviors being tested in the deleted `checkedConvertFrom` tests are tested elsewhere for `copyFromType`? Otherwise, looks good.

patrick-schultz (dismissed):
Just awaiting answer to previous question.

patrick-schultz (approved):

PR #9505, opened 2020-09-25
[Benchmark] Submit

Dania-Abuhijleh:


benchmark-service/benchmark/submit.py
```diff
@@ -0,0 +1,267 @@
...
+                f'hail-bench combine -o {combine.ofile} ' + ' '.join(all_output[i:i + combine_branch_factor]))
+            new_output.append(combine.ofile)
+            i += combine_branch_factor
+            job_i += 1
+
+        phase_i += 1
+        all_output = new_output
+
+    combine = b.new_job('final_combine_output')
+    combine.command(f'mv {build_hail.wheel} hail-*-py3-none-any.whl')
```
jigold, line 223: Same thing about the `*` in the file name

jigold (dismissed):

benchmark-service/benchmark/submit.py
```diff
@@ -0,0 +1,257 @@
...
+
+    build_hail = b.new_job('build_hail_wheel')
+    build_hail.command(f'''
+ set -ex
+ { hail_code.checkout_script() }
+ cd hail
+ time ./gradlew --version
+ time make jars python-version-info wheel
+ time (cd python && zip -r hail.zip hail hailtop)
+ time tar czf test.tar.gz -C python test
```
jigold, line 118: I don't think you don't need test.tar.gz, resources.tar.gz, and data.tar.gz.

benchmark-service/benchmark/submit.py
```diff
@@ -0,0 +1,257 @@
...
+ time tar czf resources.tar.gz -C src/test resources
+ time tar czf data.tar.gz -C python/hail/docs data
+ (cd build/deploy/dist/ && tar -cvf wheel-container.tar hail-*-py3-none-any.whl)
+ cp build/deploy/dist/hail-*-py3-none-any.whl {build_hail.wheel}
+ cp build/libs/hail-all-spark.jar {build_hail.jars}
```
jigold, line 123: I'm not sure if you need the jar file. I think you only reference the wheel file?

benchmark-service/benchmark/submit.py
```diff
@@ -0,0 +1,257 @@
...
+ export HAIL_BENCHMARK_VERSION=$HAIL_VERSION
+ cd benchmark/python/ && python3 setup.py -q bdist_wheel
+ python3 -m pip -q install dist/benchmark_hail-$HAIL_VERSION-py3-none-any.whl
+ cp dist/benchmark_hail-$HAIL_VERSION-py3-none-any.whl {build_benchmark.wheel}
+''')
+
+    resource_jobs = {}
+    for r in all_resources:
+        j = b.new_job(f'create_resource_{r.name()}').cpu(4)
+        j.command(f'mv {build_hail.wheel} hail-*-py3-none-any.whl')
```
jigold, line 141: Again, let's not use an asterisk here.

benchmark-service/benchmark/submit.py
```diff
@@ -0,0 +1,257 @@
...
+                f'hail-bench combine -o {combine.ofile} ' + ' '.join(all_output[i:i + combine_branch_factor]))
+            new_output.append(combine.ofile)
+            i += combine_branch_factor
+            job_i += 1
+
+        phase_i += 1
+        all_output = new_output
+
+    combine = b.new_job('final_combine_output')
+    combine.command(f'mv {build_hail.wheel} hail--py3-none-any.whl')
```
jigold, line 210: I don't think you need the double dash here

jigold (dismissed):
Looking good!

benchmark-service/benchmark/submit.py
Dania-Abuhijleh: when I remove the double dash or replace it with one dash I get errors saying it's not a valid file name

jigold: Interesting. The name has to be a certain format. I think you can just do this `hail-0.2-py3-none-any.whl`. The version doesn't really matter since this is only being used locally.

benchmark-service/benchmark/submit.py
```diff
@@ -0,0 +1,252 @@
...
+                             'hail_code': str(hail_code),
+                             'benchmark_code': str(benchmark_code)})
+
+    build_hail = b.new_job('build_hail_wheel')
+    build_hail.command(f'''
+ set -ex
+ { hail_code.checkout_script() }
+ cd hail
+ time ./gradlew --version
+ time make jars python-version-info wheel
```
jigold, line 118: Sorry. Nit picky thing. We just need `make wheel`

benchmark-service/benchmark/submit.py
```diff
@@ -0,0 +1,252 @@
...
+ export HAIL_VERSION=$(cat hail/python/hail/hail_pip_version)
+ export HAIL_BENCHMARK_VERSION=$HAIL_VERSION
+ cd benchmark/python/ && python3 setup.py -q bdist_wheel
+ python3 -m pip -q install dist/benchmark_hail-$HAIL_VERSION-py3-none-any.whl
+ cp dist/benchmark_hail-$HAIL_VERSION-py3-none-any.whl {build_benchmark.wheel}
+''')
+    resource_jobs = {}
+    for r in all_resources:
+        j = b.new_job(f'create_resource_{r.name()}').cpu(4)
+        j.command(f'mv {build_hail.wheel} hail--py3-none-any.whl')
```
jigold, line 138: disregard my previous comment. This is fine.

jigold (dismissed):

jigold (approved):

PR #9506, opened 2020-09-25
[query] add hailctl function to set global flags

catoverdrive:
I wanted a quick way to set global feature flags on the service backend. Also took the opportunity to move query tests into their own file.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1986,7 +1986,7 @@ steps:
       namespace:
         valueFrom: default_ns.name
       mountPath: /deploy-config
-    - name: test-tokens
+    - name: test-dev-tokens
       namespace:
         valueFrom: batch_pods_ns.name
       mountPath: /user-tokens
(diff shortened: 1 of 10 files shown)


build.yaml
```diff
@@ -1977,15 +1979,7 @@ steps:
...
-     cat > test.py <<EOF
-     import hail as hl
-     t = hl.utils.range_table(50, 3)
-     t = t.filter((t.idx % 3 == 0) | ((t.idx / 7) % 3 == 0))
-     n = t.count()
-     print(f'n {n}')
-     assert n == 17
-     EOF
-     HAIL_BILLING_PROJECT=test HAIL_DONT_RETRY_500=1 HAIL_QUERY_BACKEND=service python3 test.py
+     HAIL_BILLING_PROJECT=test HAIL_DONT_RETRY_500=1 HAIL_QUERY_BACKEND=service python3 -m pytest --log-cli-level=INFO -s -vv --instafail --durations=50 /io/test/
```
danking, line 1982: I think you forgot to check in the test file

catoverdrive: >.<

danking (approved):
LGTM

PR #9510, opened 2020-09-26
 [hailctl batch client] Add ability to cache batch status.

chrisvittal:
This will avoid extra requests especially since most batches statuses
will not change (most batches are finished).

The new function last_known_status() will make an HTTP request if and
only if the saved status is None.

### DIFF
--- a/hail/python/hailtop/batch_client/aioclient.py
+++ b/hail/python/hailtop/batch_client/aioclient.py
@@ -309,11 +309,12 @@ class SubmittedJob:
 
 
 class Batch:
-    def __init__(self, client, id, attributes, n_jobs):
+    def __init__(self, client, id, attributes, n_jobs, last_known_status=None):
         self._client = client
         self.id = id
         self.attributes = attributes
         self.n_jobs = n_jobs
+        self._last_known_status = last_known_status
 
     async def cancel(self):
         await self._client._patch(f'/api/v1alpha/batches/{self.id}/cancel')
@@ -336,7 +337,13 @@ class Batch:
 
     async def status(self):
         resp = await self._client._get(f'/api/v1alpha/batches/{self.id}')
-        return await resp.json()
+        self._last_known_status = await resp.json()
+        return self._last_known_status
+
+    async def last_known_status(self):
+        if self._last_known_status is None:
+            return await self.status()  # updates _last_known_status
+        return self._last_known_status
 
     async def wait(self, *, disable_progress_bar=TQDM_DEFAULT_DISABLE):
         i = 0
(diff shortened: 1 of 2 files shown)


hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -334,9 +335,11 @@ async def jobs(self, q=None):
             if last_job_id is None:
                 break
 
-    async def status(self):
-        resp = await self._client._get(f'/api/v1alpha/batches/{self.id}')
-        return await resp.json()
+    async def status(self, refresh=False):
```
jigold, line 338: You probably need to add these to hailtop/batch_client/client.py as well.

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -608,7 +611,8 @@ async def list_batches(self, q=None):
             body = await resp.json()
 
             for batch in body['batches']:
-                yield Batch(self, batch['id'], attributes=batch.get('attributes'), n_jobs=int(batch['n_jobs']))
+                yield Batch(self, batch['id'], attributes=batch.get('attributes'),
```
jigold, line 614: I'm not sure if `batch` here is the same as the `status` we actually want cached.

jigold (changes requested):

hail/python/hailtop/batch_client/aioclient.py
chrisvittal: You're right, I need to at least add the caching initialization to `client.Batch.__init__`. Will do.

hail/python/hailtop/batch_client/aioclient.py
chrisvittal: From my reading of it, they are.

The result of `hailctl curl default batch /api/v1alpha/batches | jq '.batches[0]'` and `hailctl curl default batch /api/v1alpha/batches/$ID` are the same (`$ID` is the id of the first batch returned by the `batches` endpoint).

The backend uses the same select and join expressions on the code paths for these endpoints. Furthermore, they are both processed through `batch_record_to_dict`.

hail/python/hailtop/batch_client/aioclient.py
chrisvittal: Hold on. Do I though? The client `Batch` class is only ever constructed by `_from_async_batch` and it seems like we don't want consumers of this API actually calling `client.Batch.__init__` or `aioclient.Batch.__init__`.

chrisvittal: You meant the `refresh` parameter, right. Done now.

jigold (approved):

chrisvittal:
After discussion with @danking, I redesigned this. The PR was failing due to timeouts since we weren't refreshing statuses in `wait` causing everything to loop forever.

I split the API up into "always hit the endpoint", and "if you use this function, hit the endpoint at most one time". I think it's more explicit.

We'll want to audit uses of `status()` to ensure that we're using the cached one if possible in several circumstances especially in `hailctl batch` itself, but that can be a future change.

jigold (approved):

PR #9511, opened 2020-09-28
[query/lir] Don’t use asm.tree in lir.Emit

patrick-schultz:
This is a simple refactoring of `lir.Emit` to directly use the core visitor based interface of ASM, rather than the higher level `tree` interface. This should have a small performance benefit, as we aren't building the in-memory tree representation only to immediately walk it with a visitor. But I also find this version of `Emit` slightly cleaner.

For reference, you can find the documentation for ASM 5.1 [here](https://javadoc.io/doc/org.ow2.asm/asm/5.1/index.html).

### DIFF
--- a/hail/src/main/scala/is/hail/lir/Emit.scala
+++ b/hail/src/main/scala/is/hail/lir/Emit.scala
@@ -1,71 +1,38 @@
 package is.hail.lir
 
-import java.io.{PrintWriter, StringWriter}
+import java.io.PrintWriter
 
 import is.hail.utils._
-import org.objectweb.asm.{ClassReader, ClassWriter}
+import org.objectweb.asm.{ClassReader, ClassVisitor, ClassWriter, Label}
 import org.objectweb.asm.Opcodes._
-import org.objectweb.asm.tree._
 import org.objectweb.asm.util.{CheckClassAdapter, Textifier, TraceClassVisitor}
 
-import scala.collection.JavaConverters._
 import scala.collection.mutable
 
 object Emit {
-  def asBytes(cn: ClassNode, print: Option[PrintWriter]): Array[Byte] = {
-    val bytes = try {
-      for (method <- cn.methods.asScala) {
-        val count = method.instructions.size
-        log.info(s"instruction count: $count: ${ cn.name }.${ method.name }")
-        if (count > 8000)
-          log.warn(s"big method: $count: ${ cn.name }.${ method.name }")
-      }
-
-      val cw = new ClassWriter(ClassWriter.COMPUTE_MAXS + ClassWriter.COMPUTE_FRAMES)
-      cn.accept(cw)
-      val b = cw.toByteArray
-      //       This next line should always be commented out!
-//      CheckClassAdapter.verify(new ClassReader(b), false, new PrintWriter(System.err))
-      b
-    } catch {
-      case e: Exception =>
-        val trace = new TraceClassVisitor(new PrintWriter(System.err))
-        val check = new CheckClassAdapter(trace)
-        cn.accept(check)
-        throw e
-    }
-    print.foreach { pw =>
-      val cr = new ClassReader(bytes)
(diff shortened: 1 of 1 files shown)


hail/src/main/scala/is/hail/lir/Emit.scala
```diff
@@ -1,95 +1,52 @@
...
+import org.objectweb.asm.{ClassReader, ClassVisitor, ClassWriter, Label}
 import org.objectweb.asm.Opcodes._
-import org.objectweb.asm.tree._
 import org.objectweb.asm.util.{CheckClassAdapter, Textifier, TraceClassVisitor}
 
-import scala.collection.JavaConverters._
 import scala.collection.mutable
 
 object Emit {
-  def asBytes(cn: ClassNode, print: Option[PrintWriter]): Array[Byte] = {
```
patrick-schultz, line 15: `asBytes` has been folded into `apply`

hail/src/main/scala/is/hail/lir/Emit.scala
```diff
@@ -127,10 +84,9 @@ object Emit {
             maxStack = depth + 2
 
           setLineNumber(x.lineNumber)
-
-          mn.instructions.add(
-            new MethodInsnNode(INVOKESPECIAL,
-              x.ctor.owner, x.ctor.name, x.ctor.desc, x.ctor.isInterface))
+          mv.visitMethodInsn(
+            INVOKESPECIAL, x.ctor.owner, x.ctor.name, x.ctor.desc, x.ctor.isInterface)
+          instructionCount += 3
```
johnc1231, line 89: Is this something that should have been there before, or is this new way of doing things changing the instruction count? Is `INVOKE_SPECIAL` 3 instructions?

hail/src/main/scala/is/hail/lir/Emit.scala
```diff
@@ -146,99 +102,132 @@ object Emit {
...
+  }
+
+  def apply(c: Classx[_], print: Option[PrintWriter]): Array[Byte] = {
+    val bytes = try {
+      val cw = new ClassWriter(ClassWriter.COMPUTE_MAXS + ClassWriter.COMPUTE_FRAMES)
+
+      emitClass(c, cw, logMethodSizes = true)
+
+      val b = cw.toByteArray
+      // This next line should always be commented out!
```
johnc1231, line 216: Is it for debugging? Maybe just say why I would want to uncomment it?

johnc1231 (changes requested):
Agree this is clearer. Left 2 questions.

hail/src/main/scala/is/hail/lir/Emit.scala
patrick-schultz: I just added `instructionCount`. Before, the `asm.tree` interface built an in-memory representation of the class file, and in particular it we could interrogate the length of the instruction list of methods. But now we're writing to the class file as we traverse the lir, so we have to keep our own instruction count.

Most lir nodes generate one bytecode instruction, which is handled below the big `match` below. `lir.NewInstanceX` requires 3 bytecode instructions: a `NEW` to allocate a new object, `DUP` to duplicate it, leaving it on the top of the operand stack twice, then `INVOKESPECIAL` to call the constructor on it (after emitting all the arguments to the constructor), which consumes the top one, leaving the bottom one as the result of the `NewInstanceX`.

hail/src/main/scala/is/hail/lir/Emit.scala
patrick-schultz: This has been here a long time; I just moved it from `asBytes`. But I can add a bit of explanation.

hail/src/main/scala/is/hail/lir/Emit.scala
patrick-schultz: I forgot to mention, `instructionCount` is only used for logging.

johnc1231 (approved):

PR #9512, opened 2020-09-28
[batch] change worker to default to the private network

danking:
At this point, every job that needs the private network specifies
it (https://github.com/hail-is/hail/pull/9380). This change moves jobs with
unspecified `network` to the private network.


jigold (approved):

danking:
this is getting stuck on the `six` issues.

PR #9514, opened 2020-09-28
[hail] Wrap AbstractRVDSpec.read failure to log the path

chrisvittal:
The exception message will now have the metadata path to more easily
assist in diagnosing issues.


tpoterba (approved):
nice

PR #9515, opened 2020-09-28
[batch] expose ABC Resource to users

danking:
CHANGELOG: The `Resource` class is now also available at hailtop.batch.Resource.


hail/python/hailtop/batch/__init__.py
```diff
@@ -4,14 +4,17 @@
 from .batch_pool_executor import BatchPoolExecutor
 from .backend import LocalBackend, ServiceBackend, Backend
 from .utils import BatchException
+from .resource import Resource, ResourceGroup
```
jigold, line 7: Should ResourceFile be here as well?

jigold (changes requested):

jigold (approved):

PR #9521, opened 2020-09-29
[query/lir] Print emitted lir, embed line refs in bytecode

patrick-schultz:
This PR is a step towards a general picture for generating debugging information in bytecode. The general picture is to write a sequence of files, each corresponding to a certain point in the compile pipeline, where each line in each file includes a line number pointing to the line in the previous file from which this line was derived. (We may eventually want richer source information than just a single line number, like a range or list of ranges.) The top of each file has the file name of the previous printout, which is the target of all line numbers in the file. We could in the future also print a list of transformations that were applied to get from the previous printed state to this one.

The idea to implement this picture is simple. Each IR node stores a line number in a mutable variable. When we want to generate a printed checkpoint, we walk the IR, printing a representation of each node, including the stored line number, and then overwriting the node's line number with the current line count of the file being written to. Some work will be required to preserve this source information in all IR transformations.

This PR implements this idea in lir only. If the `HAIL_WRITE_IR_FILES` environment variable is set, it is hardcoded to print the lir after the first `SimplifyControl` (because before that is very hard to read), and after method splitting right before emitting bytecode. It also prints out the class files themselves. Longer term we'll want to be able to control which points in the compilation get printed.

### DIFF
--- a/hail/src/main/scala/is/hail/HailContext.scala
+++ b/hail/src/main/scala/is/hail/HailContext.scala
@@ -637,7 +637,8 @@ object HailFeatureFlags {
     ("use_column_encoding", ("HAIL_DEV_USE_COLUMN_ENCODING" -> null)),
     ("use_spicy_ptypes", ("HAIL_USE_SPICY_PTYPES" -> null)),
     ("log_service_timing", ("HAIL_DEV_LOG_SERVICE_TIMING" -> null)),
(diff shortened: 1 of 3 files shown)


tpoterba (approved):
nice start. Looking forward to seeing the visibility this gets us!

PR #9523, opened 2020-09-29
[batch] Update flock strategy

chrisvittal:
flock(2) does not prevent other processes from doing anything to the
locked file unless they also gate usages on an flock call. Update Flock
to open only one descriptor, the directory that is the parent of `path`,
creating said directory if it doesn't exist.

The semantics have changed slightly. Previously if path was an already
extant directory, it would be opened and that descriptor would be
flocked. Now, the parent is always locked.

### DIFF
--- a/batch/batch/worker/flock.py
+++ b/batch/batch/worker/flock.py
@@ -3,40 +3,29 @@ import os
 import argparse
 import subprocess as sp
 
+from pathlib import Path
 from hailtop.utils import blocking_to_async
 
 
 class Flock:
     def __init__(self, path, pool=None, nonblock=False):
-        self.path = os.path.abspath(path)
+        self.path = Path(path).resolve()
+        self.lock_path = self.path.parent
         self.pool = pool
-
         self.flock_flags = fcntl.LOCK_EX
-        self.fds = []
-
         if nonblock:
             self.flock_flags |= fcntl.LOCK_NB
-
-        if os.path.isdir(self.path):
-            self.path = self.path.rstrip('/') + '/'
+        self.fd = -1
 
     def __enter__(self):
-        dirname, _ = os.path.split(self.path)
-
-        components = dirname.split('/')
-        for i in range(2, len(components) + 1):
-            path = '/'.join(components[:i])
-            os.makedirs(path, exist_ok=True)
-            fd = os.open(path, os.O_RDONLY)
-            self.fds.append(fd)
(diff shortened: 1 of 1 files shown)


batch/batch/worker/flock.py
```diff
@@ -3,40 +3,29 @@
 import argparse
 import subprocess as sp
 
+from pathlib import Path
```
chrisvittal, line 6: The path manipulations are simpler in this version than they were in my first draft. I'm not sure this is needed, and `os.path` may be fine.

jigold:
I can't remember why I did it the other way, but your way seems simpler.

jigold (dismissed):

jigold:
Oh. I remember why now. When we were using rsync I was worried someone would simultaneous try to do this:

cp gs://bucket/a/b/
cp gs://bucket/a/

Then they would clobber each other.

jigold:
But if we're copying by file, then maybe it's fine to use your strategy. What do you think?

chrisvittal:
> Oh. I remember why now. When we were using rsync I was worried someone would simultaneous try to do this:
> 
> cp gs://bucket/a/b/
> cp gs://bucket/a/
> 
> Then they would clobber each other.

Can you flesh this out, I'm not sure I understand the sequence of commands here.

jigold:
Let's say we have the following structure

gs://bucket/a/b/foo.txt
gs://bucket/a/b/bar.txt
gs://bucket/a/baz.txt


In the old design with rsync, we could be doing in parallel the following:

```
gsutil rsync gs://bucket/a
```

and

```
gsutil rsync gs://bucket/a/b/
```

I was worried we would delete files in the rsync in the middle of one job doing the copying. Basically some kind of race condition.

However, with the new design of the cache, this might not be a problem where we're not using gsutil rsync but instead writing our own version of rsync. Cotton's fleshing out the design of that.

chrisvittal:
I believe that this patch and the original version both prevent this. They would both lock the parent directory of `path` and thus would prevent concurrent copying.

Also, it appears that as long as we don't use the `-d` option, and the filenames are unique, `gsutil rsync` kinda already does what we want.

I also feel like the orignal approach was way too aggressive, it seems like it was serializing _all_ locked filesystem operations in the entire filesystem subtree since it would wait for the lock on every parent other than `/` of the requested file.

jigold (approved):

PR #9525, opened 2020-09-30
Update all make in makefiles to use $(MAKE)

chrisvittal:
As stated in the GNU make manual, "Recursive make commands should always
use the variable MAKE, not the explicit command name ‘make’".

https://www.gnu.org/software/make/manual/make.html#MAKE-Variable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,16 @@
 .DEFAULT_GOAL := default
 
 default:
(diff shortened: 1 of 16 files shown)


chrisvittal:
bump

patrick-schultz (approved):

PR #9526, opened 2020-09-30
[git][services, mostly] add *.out files to gitignore

catoverdrive:
We generate these a lot as intermediates when we manually build service images. I think it should be pretty safe to ignore all of these, since I don't know of anything we need to check in with that extension.


danking (approved):
LGTM

PR #9527, opened 2020-09-30
[batch] Add combiner functions

jigold:
CHANGELOG: Added concatenate and plink_merge functions that use tree aggregation when merging.

I tested the PLINK function locally since we don't have tests that use PLINK (can't remember exactly why).


danking:
I believe we removed PLINK because we didn't want to depend on it. It was historically hard to get a stable URL to a fixed version of the PLINK binary.

hail/python/hailtop/batch/__init__.py
```diff
@@ -4,14 +4,18 @@
...
+from .functions import _combine, concatenate, plink_merge
 
 __all__ = ['Batch',
            'LocalBackend',
            'ServiceBackend',
            'Backend',
            'BatchException',
            'BatchPoolExecutor',
-           'genetics'
+           '_combine',
```
danking, line 15: I'm inclined to not expose this at the top-level if its undocumented. Users can always access it as `hailtop.batch.functions._combine`.

hail/python/hailtop/batch/__init__.py
```diff
@@ -4,14 +4,18 @@
 from .batch_pool_executor import BatchPoolExecutor
 from .backend import LocalBackend, ServiceBackend, Backend
 from .utils import BatchException
+from .functions import _combine, concatenate, plink_merge
```
danking, line 7: I think I prefer your original intuition to make these `utils`. Functions doesn't feel like a very informative name to our users. It wouldn't matter too much except that we use the module name as the section name in the Docs.

Utilities seems good. Combinators is good, but math-y. Tools?

hail/python/hailtop/batch/functions.py
```diff
@@ -0,0 +1,111 @@
...
+from .resource import ResourceGroup, ResourceFile
+
+
+def concatenate(b: Batch, files: List[ResourceFile], branching_factor: int = 100):
+    """
+    Concatenate files using tree aggregation.
+
+    Notes
+    -----
+    Runs the `cat` command aggregating files using the `branching_factor`.
```
danking, line 17: I think this sentence is just an extended description, I don't think it needs to go under the Notes heading. We also usually put the Notes section after Examples.

hail/python/hailtop/batch/functions.py
```diff
@@ -0,0 +1,111 @@
...
+
+    >>> b = Batch()
+    >>> j1 = b.new_job()
+    >>> j1.command(f'touch {j1.ofile}')
+    >>> j2 = b.new_job()
+    >>> j2.command(f'touch {j2.ofile}')
+    >>> j3 = b.new_job()
+    >>> j3.command(f'touch {j3.ofile}')
+    >>> files = [j1.ofile, j2.ofile, j3.ofile]
+    >>> ofile = concatenate(b, files, branching_factor=2)
```
danking, line 31: What do you think of taking the batch from one of the jobs? Maybe also asserting all the jobs have the same batch? That seems like a nicer interface. That would also allow us to make the interface:
```python3
ofile = concatenate(j1.ofile, j2.ofile, j3.ofile, brancing_factor=2)
# or
ofile = concatenate(*files, branching_factor=2)
```

hail/python/hailtop/batch/functions.py
```diff
@@ -0,0 +1,111 @@
...
+    Concatenated output file.
+    """
+
+    def _concatenate(b, name, xs):
+        j = b.new_job(name=name)
+        j.command(f'cat {" ".join(xs)} > {j.ofile}')
+        return j.ofile
+
+    if len(files) == 0:
+        raise BatchException(f'Must have at least one file to concatenate.')
```
danking, line 54: Hmm. So, I understand the desire to warn the user. Concatenating only one file also seems warning worthy?

In the spirit of flexible, generic tools, it seems like `len(files) == 0` should be `j.comand(f'touch {j.ofile}')`, what do you think of that? Maybe I'm pushing things too far towards a software-engineering-mindset.

hail/python/hailtop/batch/functions.py
```diff
@@ -0,0 +1,111 @@
...
+        if len(xs) == 1:
+            return xs[0]
+        j = b.new_job(name=name)
+        for f in xs[1:]:
+            j.command(f'echo "{f.bed} {f.bim} {f.fam}" >> merge_list')
+        j.command(f'plink --bfile {xs[0]} --merge-list merge_list --out {j.ofile}')
+        return j.ofile
+
+    if len(bfiles) == 0:
+        raise BatchException(f'Must have at least one binary PLINK file to merge.')
```
danking, line 95: This, on the other hand, seems right because a PLINK dataset with no columns is kind of non-sensical and maybe impossible to create?

hail/python/hailtop/batch/utils.py
```diff
@@ -1,5 +1,6 @@
 import subprocess as sp
 
+
```
danking, line 3: remove

hail/python/hailtop/batch/functions.py
```diff
@@ -0,0 +1,111 @@
...
+    if len(files) == 0:
+        raise BatchException(f'Must have at least one file to concatenate.')
+
+    if not all([isinstance(f, ResourceFile) for f in files]):
+        raise BatchException(f'Invalid input file(s) - all inputs must be resource files.')
+
+    return _combine(_concatenate, b, 'concatenate', files, branching_factor=branching_factor)
+
+
+def plink_merge(b: Batch, bfiles: List[ResourceGroup], branching_factor: int = 100):
```
danking, line 62: let's also specify the output type with `) -> ...:`

hail/python/hailtop/batch/functions.py
```diff
@@ -0,0 +1,111 @@
...
+
+from typing import List
+
+from ..utils.utils import grouped
+from .utils import BatchException
+from .batch import Batch
+from .resource import ResourceGroup, ResourceFile
+
+
+def concatenate(b: Batch, files: List[ResourceFile], branching_factor: int = 100):
```
danking, line 11: Let's specify the output type

hail/python/hailtop/batch/functions.py
```diff
@@ -0,0 +1,111 @@
...
+
+    return _combine(_plink_merge, b, 'plink-merge', bfiles, branching_factor=branching_factor)
+
+
+def _combine(combop, b, name, xs, branching_factor=100):
+    n_levels = math.log(len(xs), branching_factor)
+    level = 0
+    while level < n_levels:
+        grouped_xs = grouped(branching_factor, xs)
+        xs = [combop(b, f'{name}-{level}-{i}', xs) for i, xs in enumerate(grouped_xs)]
```
danking, line 108: Let's specify zero prefixes so things look visually nice in a table:

```python3
# maybe put this in hailtop.utils.utils
def digits_needed(i: int):
    assert i >= 0
    if (i < 10)
      1
    else
      1 + digits_needed(I // 10)

...
    level_digits = digits_needed(n_levels)
    branch_digits = digits_needed(brancing_factor)
    while level < n_levels:
        ...
        xs = [combop(b, f'{name}-{level:0{level_digits}}-{i:0{branch_digits}}'...
```

danking (changes requested):
This is extremely awesome! I love this so much.

hail/python/hailtop/batch/functions.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+
+    def _plink_merge(b, name, xs):
+        assert xs
+        if len(xs) == 1:
+            return xs[0]
+        j = b.new_job(name=name)
+        for f in xs[1:]:
+            j.command(f'echo "{f.bed} {f.bim} {f.fam}" >> merge_list')
+        j.command(f'plink --bfile {xs[0]} --merge-list merge_list --out {j.ofile}')
```
danking, line 91: The user should probably be allowed to specify the plink image, right? Maybe an extra optional argument to plink_merge?

hail/python/hailtop/batch/functions.py
jigold: I don't like this. It should just be whatever the cat command accepts as valid inputs.

hail/python/hailtop/batch/functions.py
jigold: I can't really do this right now. I need to add an attribute to the InputResourceFile class to know what batch it came from. But I can't get the cyclical imports to work for the typing of this. I really think in some cases the typing is helpful, but it's an absolute nightmare trying to add new things.

danking (approved):

PR #9528, opened 2020-09-30
[hailctl dev] Teach dev deploy better handling of steps argument

chrisvittal:
Allow --steps/-s to be supplied multiple times and with multiple
arguments. It still handles comma separated lists to preserve the old
behavior.

### DIFF
--- a/hail/python/hailtop/hailctl/dev/deploy/cli.py
+++ b/hail/python/hailtop/hailctl/dev/deploy/cli.py
@@ -11,8 +11,8 @@ from hailtop.tls import get_context_specific_ssl_client_session
 def init_parser(parser):
     parser.add_argument("--branch", "-b", type=str,
                         help="Fully-qualified branch, e.g., hail-is/hail:feature.", required=True)
-    parser.add_argument("--steps", "-s", type=str,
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/hailctl/dev/deploy/cli.py
```diff
@@ -11,8 +11,8 @@
 def init_parser(parser):
     parser.add_argument("--branch", "-b", type=str,
                         help="Fully-qualified branch, e.g., hail-is/hail:feature.", required=True)
-    parser.add_argument("--steps", "-s", type=str,
-                        help="Comma-separated list of steps to run.", required=True)
+    parser.add_argument("--steps", "-s", nargs='+', action='append',
```
catoverdrive, line 14: why not use `'extend'` here and omit the double-unpacking below?

catoverdrive (changes requested):

hail/python/hailtop/hailctl/dev/deploy/cli.py
catoverdrive: wait, never mind. I don't think extend is supported in python 3.7.

catoverdrive (approved):

PR #9530, opened 2020-10-01
Revert "[query] add hailctl function to set global flags (#9506)"

danking:
This reverts commit 06b5a9ad7e10a564fc8fc848826beb3c445425db.

9506 modified a migration file which is not permitted. It passes the PR tests because the PR test environment doesn't use an extant database schema.

We should really test that the migration works for an extant database ... 

We could at least check that none of the SQL files are modified in the current commit.


danking:
Force merging to un-break batch.

PR #9531, opened 2020-10-01
[batch] Update some APIs to return more computer friendly datatypes

chrisvittal:
Don't format cost, timestamps, duration until they are about to be displayed in the web UI.


johnc1231:
Conflicts, and was never assigned to anyone

chrisvittal:
I won't be picking this up any time soon closing. Thanks for keeping track of this john.

PR #9537, opened 2020-10-01
[query] Expand Python version compatibility with spark 3

chrisvittal:
Spark 3 resolved the issues with library changes python 3.8, so we can
now use python 3.8 as long as pyspark is version 3 or greater.

I also took this opertunity to make every link in or welcome doc message
a canonical https link that will be recognized as such by many parsers
such as sufficiently advanced terminal emulators.

### DIFF
--- a/hail/python/hail/__init__.py
+++ b/hail/python/hail/__init__.py
@@ -1,9 +1,13 @@
 import pkg_resources
 import sys
+import pyspark
 
-if not (3, 6) < sys.version_info < (3, 8):
-    raise EnvironmentError('Hail requires Python 3.6 or 3.7, found {}.{}'.format(
+if sys.version_info < (3, 6):
+    raise EnvironmentError('Hail requires Python 3.6 or later, found {}.{}'.format(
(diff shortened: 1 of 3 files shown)


hail/python/hail/__init__.py
```diff
@@ -1,9 +1,13 @@
 import pkg_resources
 import sys
+import pyspark
```
tpoterba, line 3: I don't want to import pyspark if we don't have to. Only the Spark backend should import pyspark -- this check should go there, and not in code run at import-time.

hail/python/hail/__init__.py
```diff
@@ -16,10 +20,10 @@
  /_/ /_/\_,_/_/_/
 ===================
 
-For API documentation, visit the website: www.hail.is
+For API documentation, visit the website: https://www.hail.is
```
tpoterba, line 23: nice changes.

tpoterba (changes requested):

tpoterba (approved):

PR #9432, opened 2020-09-09
Remove TableGroupWithinPartitions IR Node

johnc1231:
It's entirely expressible in python now.

PR #9538, opened 2020-10-01
[query] Remove TableGroupWithinPartitions IR, Attempt 2

johnc1231:
There was a PruneDeadFields bug the first time I tried this, and I only just now went and fixed it. I can't reopen #9432 though because the branch was force pushed since I closed it. 

Anyway, like it says in #9432, we no longer need `TableGroupWithinPartitions`, as we have the ability to write it explicitly in Python now using a mix of `TableMapPartitions` and `StreamGrouped`.


hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
```diff
@@ -415,7 +415,7 @@ object PruneDeadFields {
         val reqRowsType = TStream(requestedType.rowType)
         val bodyDep = memoizeValueIR(body, reqRowsType, memo)
         val depGlobalType = unifySeq(child.typ.globalType,
-          bodyDep.eval.lookupOption(gName).map(_.result()).getOrElse(Array()))
+          bodyDep.eval.lookupOption(gName).map(_.result()).getOrElse(Array()) :+ requestedType.globalType)
```
johnc1231, line 418: Also worth looking here. This was a bug I uncovered while doing this. We need to consider the requested type when doing the `PruneDeadFields` pass for `TableMapPartitions`.

tpoterba: ahh, good

hail/python/hail/table.py
```diff
@@ -3493,7 +3493,13 @@ def multi_way_zip_join(tables, data_field_name, global_field_name) -> 'Table':
             [t._tir for t in tables], data_field_name, global_field_name))
 
     def _group_within_partitions(self, name, n):
-        return Table(ir.TableGroupWithinPartitions(self._tir, name, n))
+        def grouping_func(part):
+            groups = part.grouped(n)
+            key_names = list(self.key)
+            return groups.map(lambda group:
+                              hl.struct(**{key_name: group[0][key_name] for key_name in key_names}).annotate(**{name: group}))
```
patrick-schultz, line 3500: This is probably better as `group[0].select(*key_names, **{name: group})` (I think that works)

patrick-schultz (changes requested):

hail/python/hail/table.py
johnc1231: Nice, that is better.

patrick-schultz (approved):

PR #9541, opened 2020-10-01
[query] port firstMatchIn to IEmitCode

chrisvittal:
This function used `defaultValue`, which is going away as it requires `typeToTypeInfo` (which is also going away)

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/functions/Functions.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/functions/Functions.scala
@@ -421,7 +421,7 @@ abstract class RegistryFunctions {
     calculateReturnPType: (Type, Seq[PType]) => PType,
     typeParameters: Array[Type] = Array.empty
   )(
-    impl: (EmitCodeBuilder, Value[Region], PType, Array[IEmitCode]) => IEmitCode
+    impl: (EmitCodeBuilder, Value[Region], PType, Array[() => IEmitCode]) => IEmitCode
   ) {
     IRFunctionRegistry.addJVMFunction(
       new UnseededMissingnessAwareJVMFunction(name, typeParameters, valueParameterTypes, returnType, calculateReturnPType) {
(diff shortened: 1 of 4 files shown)


hail/src/main/scala/is/hail/expr/ir/functions/StringFunctions.scala
```diff
@@ -212,38 +212,34 @@ object StringFunctions extends RegistryFunctions {
...
+          val out = cb.newLocal[IndexedSeq[String]]("out",
+            Code.invokeScalaObject2[String, String, IndexedSeq[String]](
+              thisClass, "firstMatchIn", s, r))
+          IEmitCode(cb, out.isNull, {
+            val len = cb.newLocal[Int]("len", out.invoke[Int]("size"))
+            val srvb: StagedRegionValueBuilder = new StagedRegionValueBuilder(cb.emb, rt, region)
+            val elt = cb.newLocal[String]("elt")
+            val value = Code(
+                srvb.start(len),
+                Code.whileLoop(srvb.arrayIdx < len,
```
tpoterba, line 232: oh man, I can't wait until we have a stack string ptype.

tpoterba (approved):

hail/src/main/scala/is/hail/expr/ir/functions/Functions.scala
```diff
@@ -430,15 +430,12 @@ abstract class RegistryFunctions {
           r: Value[Region],
           rpt: PType,
           typeParameters: Seq[Type],
-          args: IEmitCode*
-        ): IEmitCode = {
-          assert(unify(typeParameters, args.map(_.pt.virtualType), rpt.virtualType))
```
chrisvittal, line 435: I removed this assertion. I _think_ it's okay without it.

tpoterba:
ahhhh, yes, good catch.

tpoterba:
there is a long road of subtle bugs in the codebuilder-ing road ahead.

tpoterba (approved):

PR #9543, opened 2020-10-01
[batch] Teach BatchPoolExecutor how to handle container errors

danking:
CHANGELOG: BatchPoolExecutor now raises an informative error message for a variety of "system" errors, such as missing container images.

If the main container fails for reasons beyond BatchPoolExecutor's control, such
as a missing container image, we previously did not report these errors. In
fact, we encountered errors when trying to load the output file that cannot
exist if the main container errors.

Smaller included changes:
- directly use the asynchronous, low-level client instead of the synchronous,
  low-level client
- introduce an `async_cancel` now that we have access to the async client.

### DIFF
--- a/hail/python/hailtop/batch/batch_pool_executor.py
+++ b/hail/python/hailtop/batch/batch_pool_executor.py
@@ -453,7 +453,7 @@ class BatchPoolFuture:
         self.fetch_lock = asyncio.Lock()
         executor._add_future(self)
 
-    async def cancel(self):
+    def cancel(self):
         """Cancel this job if it has not yet been cancelled.
 
         ``True`` is returned if the job is cancelled. ``False`` is returned if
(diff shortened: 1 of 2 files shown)


danking:
Hey, @chrisvittal, I assigned to you for curiosity's sake. If you don't want to review it on Friday, just send it to Jackie.

hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -438,23 +442,33 @@ class BatchPoolFuture:
...
+        self.job = job
         self.output_gcs = output_gcs
         self.value: Any = NO_VALUE
         self._exception: Optional[BaseException] = None
         self.fetch_lock = asyncio.Lock()
         executor._add_future(self)
 
-    def cancel(self):
-        """Cancels this job if it has not yet been cancelled.
+    async def cancel(self):
```
chrisvittal, line 456: should still be sync.

chrisvittal (changes requested):
Interesting! Seems that there's one logic error here.

hail/python/hailtop/batch/batch_pool_executor.py
danking: Indeed

chrisvittal (approved):

PR #9544, opened 2020-10-01
[batch][auth][ci][monitoring][notebook] check for sql mutation

danking:
SQL migrations are not permitted to be modified. Unfortunately, our tests
previously did not verify this at all. Indeed, a PR merged which modified a SQL
file. This PR caused main to fail a deploy.

This change verifies that no SQL migration is mutated in the source SHA relative
to the target SHA. One can also use it locally by running `make
check-services` from the root. Unfortunately, it does not work properly when run
on the main branch because there is no obvious point of comparison.

I considered comparing against the previous commit, but that might cause
failures if we have to manually fix something in batch. As such, I prefer a
non-deploy only test.

### DIFF
--- a/check-sql.sh
+++ b/check-sql.sh
@@ -9,7 +9,7 @@ target_treeish=${HAIL_TARGET_SHA:-$(git merge-base main HEAD)}
 modified_sql_file_list=$(mktemp)
 
 git diff --name-status $target_treeish sql \
-    | grep -Ev '^A|^M\t[^/]+/sql/estimated-current.txt' \
+    | grep -Ev '^A|^M\t[^/]+/sql/(estimated-current.sql|delete-[^ ]+-tables.sql)' \
            > $modified_sql_file_list
 
 if [ "$(cat $modified_sql_file_list | wc -l)" -ne 0 ]
--- a/check-sql.sh
+++ b/check-sql.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# Verify the the working tree modifices no sql files relative to the main
+# branch. This will always pass on a deploy because the working tree is an
+# unmodified copy of the main branch.
+
+target_treeish=${HAIL_TARGET_SHA:-$(git merge-base main HEAD)}
+
+modified_sql_file_list=$(mktemp)
+
+git diff --name-status $target_treeish sql \
+    | grep -Ev '^A|^M\t[^/]+/sql/estimated-current.txt' \
+           > $modified_sql_file_list
+
+if [ "$(cat $modified_sql_file_list | wc -l)" -ne 0 ]
+then
+    cat $modified_sql_file_list
+    echo 'At least one migration file was modified. See above.'
+    exit 1
+fi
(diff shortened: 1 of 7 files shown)


check-sql.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Verify the the working tree modifices no sql files relative to the main
```
jigold, line 3: typo in modifies.

check-sql.sh
```diff
@@ -0,0 +1,20 @@
...
+# branch. This will always pass on a deploy because the working tree is an
+# unmodified copy of the main branch.
+
+target_treeish=${HAIL_TARGET_SHA:-$(git merge-base main HEAD)}
+
+modified_sql_file_list=$(mktemp)
+
+git diff --name-status $target_treeish sql \
+    | grep -Ev '^A|^M\t[^/]+/sql/estimated-current.txt' \
```
jigold, line 12: This also needs to ensure we can modify the delete-*-tables.sql file.

jigold, line 12: This didn't work for me:

```
wmecc-475:hail jigold$ git diff --name-status $target_treeish monitoring/sql | grep -Ev '^A|^M\t[^/]+/sql/estimated-current.txt'
M	monitoring/sql/estimated-current.sql
M	monitoring/sql/initial.sql
```

check-sql.sh
```diff
@@ -0,0 +1,20 @@
...
+           > $modified_sql_file_list
+
+if [ "$(cat $modified_sql_file_list | wc -l)" -ne 0 ]
```
jigold, line 15: This didn't work for me locally. Might still be right?

```
wmecc-475:hail jigold$ $(cat $modified_sql_file_list | wc -l)
-bash: 2: command not found
wmecc-475:hail jigold$ cat $modified_sql_file_list | wc -l
       2
```

jigold (dismissed):

check-sql.sh
danking: You're getting a command not found because bash resolves
```
 $(cat $modified_sql_file_list | wc -l)
```
To the string
```
        2
```
Which it then tries to execute, but there is no program named "2".

check-sql.sh
danking: Why is initial.sql modified?

danking: I renamed everything to `.sql` and updated this PR, once https://github.com/hail-is/hail/pull/9560 merges this will work.

danking:
stacked on https://github.com/hail-is/hail/pull/9560

check-sql.sh
jigold: I was doing some tests to make sure it worked.

jigold: Please address this!

jigold (changes requested):
See comment about delete_tables

jigold:
create_accounts isn't idempotent
https://ci.hail.is/batches/130708/jobs/24

jigold (approved):

danking:
huh this should have merged..

PR #9546, opened 2020-10-02
[annotationdb] update annotation db docs page for deploy

pwc2:
Changes to make sure that only the annotation datasets are visible on the docs page, now that the `datasets.json` config file contains all available datasets.

Overview:

- In `datasets.json`, moved "key_properties" inside an "annotation_db" field, like `"annotation_db":  {"key_properties":  []}`, so that only the datasets with the "annotation_db" key are shown in the annotation DB docs page. Removed "key_properties" from non-annotation datasets.

- Minor reformatting changes to docs page, added a reference genome column to the HTML table.

- Updated deploy script to reflect the filename change from `annotation_db.json` to `datasets.json`.

- Modified checks for keys in dicts from `assert key in doc, doc` to `assert key in doc` in `DatasetVersion.from_json()` and `Dataset.from_name_and_json()`. Since the `doc` that is passed to these methods from the checked in JSON file is just a dict like `doc = {"annotation_db": {"key_properties": [...]}, "description": ..., "url":  ..., "versions": [...]}` this seems to work fine. Let me know if `key in doc, doc` form was used for other reasons I've overlooked.

### DIFF
--- a/hail/python/hail/experimental/db.py
+++ b/hail/python/hail/experimental/db.py
@@ -144,7 +144,13 @@ class Dataset:
         """
         assert 'description' in doc, doc
         assert 'url' in doc, doc
-        assert 'key_properties' in doc, doc
+        if 'annotation_db' in doc:
+            assert 'key_properties' in doc['annotation_db']
+            key_props = set(doc['annotation_db']['key_properties'])
+        else:
+            assert 'key_properties' in doc
+            key_props = set(doc['key_properties'])
+        assert 'annotation_db' in doc, doc
         assert 'versions' in doc, doc
         versions = [DatasetVersion.from_json(x) for x in doc['versions']]
         if not custom_config:
@@ -153,7 +159,7 @@ class Dataset:
             return Dataset(name,
                            doc['description'],
                            doc['url'],
-                           set(doc['key_properties']),
+                           key_props,
                            versions)
 
     def __init__(self,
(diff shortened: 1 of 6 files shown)


hail/python/hail/experimental/db.py
```diff
@@ -33,8 +33,8 @@ def from_json(doc: dict) -> 'DatasetVersion':
         -------
         :class:`.DatasetVersion`
         """
-        assert 'url' in doc, doc
-        assert 'version' in doc, doc
+        assert 'url' in doc
+        assert 'version' in doc
```
danking, line 37: In Python, the `, ...` provides more detail. For example:
```
In [1]: key = 'B'                                                                                                                                                                                                                                                               

In [2]: doc = {'A': 1}                                                                                                                                                                                                                                                          

In [3]: assert key in doc                                                                                                                                                                                                                                                       
  File "<ipython-input-3-c6a5ebd6a2ac>", line 1, in <module>
    assert key in doc
AssertionError


In [4]: assert key in doc, doc                                                                                                                                                                                                                                                  
  File "<ipython-input-4-28f2c139973f>", line 1, in <module>
    assert key in doc, doc
AssertionError: {'A': 1}

```

Let's keep them

danking (changes requested):
Looks great!

hail/python/hail/experimental/db.py
pwc2: ah gotcha, just put them back in

danking (approved):
Looks good!

In the future, just dismiss my review, a PR assigned to me without a review from me will show up in my scorecard: https://scorecard.hail.is.

PR #9547, opened 2020-10-02
[Benchmark] GitHub polling loop

Dania-Abuhijleh:


benchmark-service/deployment.yaml
```diff
@@ -32,6 +32,8 @@ spec:
             value: /deploy-config/deploy-config.json
           - name: HAIL_SHA
             value: "{{ code.sha }}"
+          - name: HAIL_CI_OAUTH_TOKEN
+            value: /secrets/oauth-token/oauth-token
```
jigold, line 36: I don't think you need the second `oauth-token`. Is that what CI did?

jigold: Do you need this? I think the default path should be the one for the deployment rather than the local one. Then you can get rid of this.

jigold (changes requested):

benchmark-service/benchmark/benchmark.py
```diff
@@ -200,8 +210,37 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
     return await render_template('benchmark', request, userdata, 'compare.html', context)
 
 
+async def query_github(app):
+    github_client = app['github_client']
+    request_string = f'/repos/hail-is/hail/commits?sha={main_commit_sha}&since={START_POINT}'
```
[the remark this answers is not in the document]
Dania-Abuhijleh: Maybe it's a misleading name, I think it's the branch you want to get the commits from

benchmark-service/deployment.yaml
Dania-Abuhijleh: don't think I need it but then what should I replace this with
with open(os.environ.get('HAIL_CI_OAUTH_TOKEN', 'oauth-token/oauth-token'), 'r') as f:

benchmark-service/benchmark/benchmark.py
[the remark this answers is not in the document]
jigold: I think this can be 'main'. https://developer.github.com/v3/repos/commits/#list-commits

benchmark-service/deployment.yaml
jigold: I believe you can replace the default path with `/secrets/oauth-token/oauth-token`. I looked at CI and it looks like they kept the environment variable. Up to you I guess.

benchmark-service/benchmark/benchmark.py
```diff
@@ -200,8 +207,37 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
     return await render_template('benchmark', request, userdata, 'compare.html', context)
 
 
+async def query_github(app):
+    global START_POINT
+    github_client = app['github_client']
+    request_string = f'/repos/hail-is/hail/commits?since={START_POINT}'
```
jigold, line 213: So was the decision to not include the sha because it doesn't work?

benchmark-service/benchmark/benchmark.py
```diff
@@ -200,8 +207,37 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
+        START_POINT = commit.get('commit').get('author').get('date')
+        log.info(f'start point is now {START_POINT}')
+    log.info('got new commits')
+
+
+async def github_polling_loop(app):
+    while True:
+        await query_github(app)
+        log.info('successfully queried github')
+        await asyncio.sleep(60)
```
jigold, line 230: Let's change this to 600 for now. That's 10 minutes. I don't want to blow out our API request limits to Github.

We'll merge this in and then we can reevaluate later once we've seen the logs in the production instance of the service.

docker/requirements.txt
```diff
@@ -51,3 +51,4 @@ uvloop==0.14.0
 Werkzeug==0.15.4
 wheel>=0.31.0
 zulip==0.6.3
+six>1.14.0
```
jigold, line 54: I think other PRs are running into this problem. Where is six used? Why was this change necessary?

jigold (changes requested):

docker/requirements.txt
Dania-Abuhijleh: looks like it's used in google cloud, I think the problem was because pip changed something 
`ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.
We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

pytest-astropy 0.8.0 requires pytest-cov>=2.0, which is not installed.
pytest-astropy 0.8.0 requires pytest-filter-subpackage>=0.1, which is not installed.
astroid 2.3.3 requires typed-ast<1.5,>=1.4.0; implementation_name == "cpython" and python_version < "3.8", which is not installed.`

benchmark-service/benchmark/benchmark.py
Dania-Abuhijleh: I think a specific sha was working but sha=main didn't work

docker/requirements.txt
danking: Nice hunting, this is a good fix. I'm working on a more thorough fix but it's been caught up in lots of silly details.

benchmark-service/benchmark/benchmark.py
```diff
@@ -200,8 +207,37 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
+    github_client = app['github_client']
+    request_string = f'/repos/hail-is/hail/commits?since={START_POINT}'
+
+    data = await github_client.getitem(request_string)
+    new_commits = []
+    for commit in data:
+        sha = commit.get('sha')
+        new_commits.append(commit)
+        log.info(f'commit {sha}')
+        START_POINT = commit.get('commit').get('author').get('date')
```
danking, line 221: get will return `None` if the key doesn't exist and then you'll get an error on the subsequent call to `.get`. That error won't include which key caused the error. I think you're bette roff with:
```
commit['commit']['author'].get('date')
```
If you don't expect `date` to be missing, you can use `[]` for that too.

danking (changes requested):
Lookin' awesome, just a couple comments!

danking (approved):
ship it like fedex

PR #9552, opened 2020-10-02
NDArrayInv ported to EmitI

johnc1231:
This ports `NDArrayInv` to `EmitI`. 

With this, all NDArray emit rules are ported to `EmitI`, though the `NDArrayEmitter` still uses `EmitCode` internally. Will address that in a future PR.


chrisvittal (approved):
Great work!

Now we just need construct to return a PCode. :stuck_out_tongue_winking_eye:

PR #9554, opened 2020-10-02
[tls] Simplify TLS use and enable TLS for address

danking:
This PR attempts to simplify the use of TLS and HTTP(S) in Hail. The big changes
are in `hail/python/hailtop`. In particular I removed several functions with
confusingly overlapping functionality in `tls.py`. Instead, we now have three
functions:

- `internal_server_ssl_context`
- `internal_client_ssl_context`
- `external_client_ssl_context`

The client context is configured to seek certificates from its peers. Both
internal contexts load the Hail certificate chain specified in the Hail SSL
Config. The external client context does not load the hail certificate chain.

I intend all Hail's HTTP(S) requests to use `httpx.py` (so named to not conflict
with modules named `http`). Again, I have simplified the landscape. We now have
two functions:

- `httpx.client_session`: The constructor for all asynchronous, HTTPS client
  sessions.
- `httpx.blocking_client_session`: The constructor for all synchronous, HTTPS
   client sessions.

Both sessions have the exact same configuration parameters. The API is exactly
the same except the blocking client session replaces asynchronous methods with
synchronous ones.

Both sessions accept the `aiohttp.ClientSession` constructor parameters. They
support one new parameter and modify the behavior of one old parameter.
- `retry_transient`: when set to `True` this parameter will retry all transient
  errors in all requests made by this session. This defaults to `True`.
- `raise_for_status`: this parameter now defaults to `True` and includes the
  response body text in the error message. Both
Both parameters may be overridden on a per-request basis.

- `httpx.ResponseManager` and `httpx.ClientSession` work together to enable
  `retry_transient` and `raise_for_status`. Aiohttp has this unusual structure where
  all the request methods are synchronous but they return an object that is both
  awaitable and an async context manager. I mirror their structure exactly. The
  `httpx.ResponseManager` is both awaitable and an async context manager. Its
  `response_coroutine` field is a coroutine that includes the retry and
  raise-for-status logic.

- The `HailResolver` overrides domain name resolution to first consult the Hail
  `address` service. `address` is effectively a domain name server. It watches
  kubernetes services and publishes the pod IPs. It supports two name styles:
  `service` and `service.namespace`. The former uses the deploy config to
  determine in which namespace to find the given service. Currently, the
  client-side library only looks up IPs for `shuffler` and `address`.

- `BlockingClientSession` and `BlockingContextManager` wrap the
  aforementioned `httpx` classes. `BlockingClientResponse` wraps an
  `aiohttp.ClientResponse`

---

Examples of correct usage:

A blocking HTTPS request:

```python3
with httpx.blocking_client_session() as session:
    with session.post(url, json=config, headers=headers) as resp:
        assert resp.status == 200
        print(resp.text())
```

An asynchronous HTTPS request to auth:
```python3
async with httpx.client_session() as session:
    async with session.get(
            deploy_config.url('auth', '/api/v1alpha/userinfo'),
            headers=headers) as resp:
        assert resp.status == 200
        print(await resp.json())
```

A blocking HTTPS session with a large default timeout:

```python3
httpx.blocking_client_session(
    headers=service_auth_headers(deploy_config, 'query'),
    timeout=aiohttp.ClientTimeout(total=600))
```

cc: @catoverdrive @Dania-Abuhijleh

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -21,7 +21,7 @@ from hailtop.batch_client.parse import (parse_cpu_in_mcpu, parse_memory_in_bytes
                                         parse_storage_in_bytes)
 from hailtop.config import get_deploy_config
 from hailtop.tls import internal_server_ssl_context
-from hailtop.httpx import client_session
+from hailtop import httpx
 from hailtop.hail_logging import AccessLogger
 from gear import (Database, setup_aiohttp_session,
                   rest_authenticated_users_only, web_authenticated_users_only,
@@ -965,7 +965,7 @@ WHERE user = %s AND id = %s AND NOT deleted;
                 reason=f'wrong number of jobs: expected {expected_n_jobs}, actual {actual_n_jobs}')
         raise
 
-    async with client_session(raise_for_status=True) as session:
+    async with httpx.client_session(raise_for_status=True) as session:
         await request_retry_transient_errors(
             session, 'PATCH',
             deploy_config.url('batch-driver', f'/api/v1alpha/batches/{user}/{batch_id}/close'),
@@ -1505,7 +1505,7 @@ async def index(request, userdata):  # pylint: disable=unused-argument
 
 
 async def cancel_batch_loop_body(app):
-    async with client_session(raise_for_status=True) as session:
+    async with httpx.client_session(raise_for_status=True) as session:
         await request_retry_transient_errors(
             session, 'POST',
             deploy_config.url('batch-driver', '/api/v1alpha/batches/cancel'),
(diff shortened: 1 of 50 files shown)


batch/batch/front_end/front_end.py
```diff
@@ -964,8 +965,7 @@ async def close_batch(request, userdata):
                 reason=f'wrong number of jobs: expected {expected_n_jobs}, actual {actual_n_jobs}')
         raise
 
-    async with in_cluster_ssl_client_session(
-            raise_for_status=True, timeout=aiohttp.ClientTimeout(total=60)) as session:
```
danking, line 968: This seemed like too generous a timeout, so I'm reverting it to the default of 5s.

danking:
needs a rethink

PR #9556, opened 2020-10-03
Use ?= in make variables for spark and scala versions

chrisvittal:
The semantics of setting a variable in make are kinda weird, example:

```make
FOO := foo
BAR ?= bar
.PHONY: echo
echo:
	echo $(FOO) $(BAR)
```

```sh
$ FOO=baz make
echo foo bar
foo bar
$ make FOO=baz
echo baz bar
baz bar
$ BAR=baz make
echo foo baz
foo baz
```

This will fix an issue where ci wasn't actually building with spark 3.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -627,7 +627,7 @@ steps:
      {{ code.checkout_script }}
      cd hail
      time retry ./gradlew --version
-     export SPARK_VERSION="3.0.0"
+     export SPARK_VERSION="3.0.1" SCALA_VERSION="2.12.12"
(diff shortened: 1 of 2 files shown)


tpoterba (approved):
good catch, Chris!

tpoterba:
looks like this linting rule was necessary for the compilation: https://github.com/hail-is/hail/pull/9551

chrisvittal:
Will fix. But then will also probably get fixing all our unused stuff too.

PR #9557, opened 2020-10-03
[hailctl batch] Teach batch list verbosity, limits, and formatting

chrisvittal:
Summary: hailctl batch list --help
```
usage: hailctl batch list [-h] [--query QUERY] [--limit LIMIT] [--all] [--before BEFORE] [--full] [--no-header] [-o O]

List batches

optional arguments:
  -h, --help            show this help message and exit
  --query QUERY, -q QUERY
                        see docs at https://batch.hail.is/batches
  --limit LIMIT, -l LIMIT
                        number of batches to return (default 50)
  --all, -a             list all batches (overrides --limit)
  --before BEFORE       start listing before supplied id
  --full                when output is tabular, print more information
  --no-header           do not print a table header
  -o O                  specify output format (json, yaml, csv, tsv, or any tabulate format)
```

Details:
* Default listing to a limit of 50 records, once batch statuses are
cached from `list_batches`, this should result in 1 http request for the
default behavior of this tool.
* Teach --limit option to cap the number of records returned
* Teach --all to override --limit
* Teach --before to pass a last_batch_id query parameter to list_batches
* Teach --full to print all status information
* Teach --no-header to enable not printing a header for tabular output
* Teach -o {format} to change the output format the following are supported:
    - json: always full json output, like hitting the list enpoint manually
    - yaml: like json, but yaml!
    - csv/tsv: simple comma/tab separated output for machine processing
    - any python-tabulate output format, listed here:
        https://github.com/astanin/python-tabulate#table-format

The only default that has been changed is the listing limit.

### DIFF
--- a/hail/python/hailtop/hailctl/batch/list_batches.py
+++ b/hail/python/hailtop/hailctl/batch/list_batches.py
@@ -1,12 +1,57 @@
+import sys
+import csv
 import tabulate
+
+from .batch_cli_utils import make_formatter
 
 
 def init_parser(parser):
     parser.add_argument('--query', '-q', type=str, help="see docs at https://batch.hail.is/batches")
+    parser.add_argument('--limit', '-l', type=int, default=50,
+                        help='number of batches to return (default 50)')
+    parser.add_argument('--all', '-a', action='store_true',
+                        help='list all batches (overrides --limit)')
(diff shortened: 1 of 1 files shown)


danking (approved):
Love it

PR #9558, opened 2020-10-04
[query] Clean up build.gradle

chrisvittal:
- Remove all the deprecated tasks
- Re-add -Xlint:-unused, but only for scala 2.12


tpoterba (approved):
nice!

PR #9564, opened 2020-10-06
[ci] Remove batches endpoint

jigold:


jigold:
Whoops. Just read the rest of the zulip thread. Will fix.

PR #9567, opened 2020-10-06
[benchmark] fix missed HAIL_VERSION rename in makefile

patrick-schultz:
#9500 didn't update the `hail/hail` makefile


danking (approved):
So that's how people normally run benchmark.

patrick-schultz:
I only use install-benchmark

PR #9570, opened 2020-10-07
[query] group_by aggregator needs to check axes in grouping expression

johnc1231:
Previously, when we did `hl.agg.group_by(group_expr, aggregation_expr)`, we were only tracking the indices picked up from the `aggregation_expr`. This led to us throwing bad error messages.


johnc1231:
Lots of tests failing, this isn't quite right yet.

johnc1231:
This is incorrect, should fix at some point, but not on priority list right now.

PR #9571, opened 2020-10-07
[batch] Get rid of double copying in io

jigold:
This PR fixes the double copying of input and output files. This was due to missing parentheses in the shell command for copying.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -552,7 +552,7 @@ def copy_command(src, dst, requester_pays_project=None):
 
     cp = f'retry gsutil {requester_pays_project} -m cp -R {shq(src)} {shq(dst)}'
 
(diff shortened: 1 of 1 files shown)


johnc1231 (approved):

danking:
&& binds tighter than ||, I don't see how the parentheses make any difference? (Also, we should use curlies, which are for precedence as opposed to parens which trigger sub-shells to start and can be confusing wrt environment variables)

jigold:
Well, I look forward to your PR fixing this. I tested the fix with dev deploy before and after my change and it fixed it.

danking:
Ah, you're totally right, [and and or have equal precedence](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html)

PR #9574, opened 2020-10-07
[query] use a 10 minute timeout

danking:
If query has to pull its image fresh each job takes ~2 minutes. I also fixed some
whitespace issues in the tests and a little bug in `retry_response_returning_functions`.

### DIFF
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -410,16 +410,16 @@ def retry_response_returning_functions(fun, *args, **kwargs):
         errors += 1
         if errors % 10 == 0:
             log.warning(f'encountered {errors} bad status codes, most recent '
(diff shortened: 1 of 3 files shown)


hail/python/hailtop/utils/utils.py
```diff
@@ -410,16 +410,16 @@ def retry_response_returning_functions(fun, *args, **kwargs):
         errors += 1
         if errors % 10 == 0:
             log.warning(f'encountered {errors} bad status codes, most recent '
-                        f'one was {response.status_code}', exc_info=True)
+                        f'one was {response.status_code}')
```
danking, line 413: there's no exception here so python prints:
```
NoneType: None
```
removing exc_info removes this message.

tpoterba (approved):

johnc1231 (approved):

PR #9576, opened 2020-10-07
[batch] change worker to default to the public network

danking:
At this point, every job that needs the private network specifies it (#9380). This change moves jobs with unspecified `network` to the public network.

Stacked on #9594.


danking:
needs more thought.

PR #9577, opened 2020-10-07
[benchmark] add profiler options to hail-bench

patrick-schultz:
Add async-profiler support to hail-bench.

Add options:
* `--profile` (`-p`). Can be `-p cpu` or `-p alloc`; `-p` alone defaults to `cpu`.
* `--prof-fmt` (`-f`). Can be `-f html`, `-f flame`, or `-f jfr`. If not specified, defaults to `html`. `html` writes a collapsible tree as you would see in a profiler gui. `flame` writes an svg format flame graph. `jfr` writes a Java Flight Recorder format file which can be read by Java Mission Control or the integrated IntelliJ profiler.

Uses [async-profiler](https://github.com/jvm-profiling-tools/async-profiler), which must be installed at the path `$ASYNC_PROFILER_HOME`. It attaches to the spark JVM using the JVMTI argument interface. The JVMTI async-profiler arguments are described [here](https://github.com/jvm-profiling-tools/async-profiler/blob/v1.8.1/src/arguments.cpp#L49).

### DIFF
--- a/benchmark/python/benchmark_hail/run/cli.py
+++ b/benchmark/python/benchmark_hail/run/cli.py
@@ -48,6 +48,14 @@ def main(args_):
     parser.add_argument('--dry-run',
                         action='store_true',
                         help='Print benchmarks to execute, but do not run.')
+    parser.add_argument('--profile', '-p',
+                        choices=['cpu', 'alloc'],
+                        nargs='?', const='cpu',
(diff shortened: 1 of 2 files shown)


tpoterba (commented):
awesome! can't wait to try it out

tpoterba (approved):

PR #9578, opened 2020-10-07
[query] Python Version of Chained Linreg

johnc1231:
`_linear_regression_rows_nd` now supports chained linear regression correctly. Also added support for pass through arguments. This should now support everything that regular linear regression supports, but all written in Python. This allows a few more tests to run on local backend. 

I'd be particularly interested in whether you think there's a way to write this that would generate more efficient IR. It seems hacky to have to map over the range of rows and index into it, but I didn't see a better way.


johnc1231:
Hmm, so I fixed one pruning bug by inserting the `TableKeyBy` in `TableMapPartitions` case of `PruneDeadFields`. Now it seems there's some other pruning bug with a missing key, but it only happens on `LocalBackend`. Looking into it, but shouldn't require any changes to the python implementation.

johnc1231:
Ok, second pruning bug fixed. This commit illustrates the change, probably worth looking at in isolation: https://github.com/hail-is/hail/pull/9578/commits/f4ce01d59a56bef7496490145e66d274f51139e6

Essentially, the problem was that we were rebuilding an `InsertFields(child, newFields, _)` node with a requested type of empty struct. The old rebuilding strategy handled this by recursively rebuilding the `child` node, then filtering `newFields` to include only those in the requested type. However, if the `child` node can't be filtered (say it's a `Ref` of a certain type), and one of the fields being inserted is overwriting a field in the child, you need to make sure you do the overwriting so the typing works out. Happy to explain more if this is unclear.

I also added an optimization to just drop the whole `InsertFields` if the requested type is the empty struct, not sure if that's good pruning manners though. 

CC @tpoterba

hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
```diff
@@ -1844,19 +1846,24 @@ object PruneDeadFields {
...
-        InsertFields(rebuiltChild,
-          fields.flatMap { case (f, fir) =>
-            if (depFields.contains(f))
-              Some(f -> rebuildIR(fir, env, memo))
-            else {
-              log.info(s"Prune: InsertFields: eliminating field '$f'")
-              None
-            }
-          }, fieldOrder.map(fds => fds.filter(f => depFields.contains(f) || preservedChildFields.contains(f))))
+        if (depFields.isEmpty) {
```
tpoterba, line 1854: in general I don't like these kind of branches -- the general case should handle empty structs. I think the core bug is that `if (depFields.contains(f))` should be `if (depFields.contains(f)) || preservedChildFields.contains(f))`, right?

johnc1231: Correct, the `if (depFields.isEmpty)` is unnecessary. It was intended to further prune the particular case I was in, which was:

```
InsertFields(Ref(g, TStruct("a": TArray)), Seq(("a", ....Some linear algebra.....))
```

with a requested type of empty. This bug fix means you have to  keep the linear algebra around so that `a` gets replaced by something of the right type, but since the requested type was empty it seemed silly to do the linear algebra, so I added that `if (depFields.isEmpty)` case. I guess we can maybe just assume some higher up pruning will remove this whole `InsertFields`, so there's no need to try and special case the empty case here?

tpoterba: actually, this is also incorrect -- InsertFields has the same missingness has `old`, but `MakeStruct` is always defined.

I think the way to fix this is probably with a rewrite that inserts a SelectFields on top of `rebuiltChild`.

johnc1231: Ah yeah, that's a better idea, plus it removes the need to do `if (depFields.contains(f)) || preservedChildFields.contains(f))`. Thanks, I'll try this.

tpoterba: Yeah, so I think you want to insert a select fields if there are any fields that are in both the `rebuiltChild` and `insertedField`, and NOT in `depFields`, right?

johnc1231: Yeah, I think that's right. Currently have it working with always inserting a `SelectFields`,  am modifying to do this check

johnc1231: Seems to work now, the check was like you said:

```
val insertOverwritesUnrequestedButPreservedField = fields.exists{ case (fieldName, _) =>
      preservedChildFields.contains(fieldName) && !depFields.contains(fieldName)
}
```

johnc1231:
https://github.com/hail-is/hail/pull/9578/commits/c14503d921c90bbd8d9379b23db5a9e43ab71bf7 Now fixes the `InsertFields` pruning issue by inserting a `SelectFields` in the situation where an inserted field is going to overwrite a field in the child IR that is not part of the requested type.

johnc1231:
Something about that must still be invalid, as I'm  now failing `TypeCheck` on a seemingly unrelated test as a result of pruning

johnc1231:
The only thing I can think of so far is that somehow doing this messes up the `fieldOrder`? Getting an error in a `MakeArray` TypeCheck that suggests I'm maybe grabbing the wrong field of a struct in a `GetField`. Specifically:

```
GetField(Ref(global,struct{__cols: array<struct{col_idx: int32}>, __iruid_493: array<struct{}>}),__iruid_493)
```

is an argument to `MakeArray` where the array element type should be `array<struct{col_idx: int32}>`.

tpoterba:
ah!!!

johnc1231:
I'm not actually sure what the right way to fix this is. Should the memoization pass ensure that the requested column type for left and right children of `MatrixUnionCols` are consistent, or is it just a problem to handle when rebuilding?

tpoterba:
Doing this on memoization isn't enough, since the rebuild can produce any type bounded by the current type and the requested type. I think we want to do what we do in `If`, and upcast to the requested type if the types differ.

johnc1231:
I'll have to figure out tomorrow. It would seem I haven't gotten the right upcasting yet, as something about the keys is getting messed up. The never ending PR continues.

johnc1231:
I'm going to close this and break it up once I see how tests shake out. I think it's fixed now, but half of this PR has become pruner fixes that probably ought to be their own PR.

hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
```diff
@@ -1313,7 +1348,11 @@ object PruneDeadFields {
         )
       case SelectFields(old, fields) =>
         val sType = requestedType.asInstanceOf[TStruct]
-        memoizeValueIR(old, TStruct(fields.flatMap(f => sType.fieldOption(f).map(f -> _.typ)): _*), memo)
+        val oldReqType = TStruct(old.typ.asInstanceOf[TStruct]
+          .fieldNames
+          .flatMap(fn => sType.fieldOption(fn).map(fd => (fd.name, fd.typ))): _*)
+
+        memoizeValueIR(old, oldReqType, memo)
```
johnc1231, line 1355: Fixed by #9626

hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
```diff
@@ -1437,7 +1478,12 @@ object PruneDeadFields {
         val body2 = rebuildIR(body, BindingEnv(Env(
           gName -> child2.typ.globalType,
           pName -> TStream(child2.typ.rowType))), memo)
-        TableMapPartitions(child2, gName, pName, body2)
+        val body2ElementType = body2.typ.asInstanceOf[TStream].elementType.asInstanceOf[TStruct]
+        val child2Keyed = if (child2.typ.key.exists(k => !body2ElementType.hasField(k)))
+          TableKeyBy(child2, child2.typ.key.takeWhile(body2ElementType.hasField))
+        else
+          child2
+        TableMapPartitions(child2Keyed, gName, pName, body2)
```
johnc1231, line 1486: Fixed separately in #9627

hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
```diff
@@ -1584,6 +1630,35 @@ object PruneDeadFields {
...
+          )
+        } else {
+          MatrixUnionCols(
+            upcast(left2, requestedType, upcastRows=false, upcastGlobals = false),
+            upcast(right2, requestedType, upcastRows=false, upcastGlobals = false),
+            joinType
+          )
+        }
+
+
```
johnc1231, line 1661: Fixed and tested in #9629

hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
```diff
@@ -1845,20 +1922,47 @@ object PruneDeadFields {
...
+          preservedChildFields.contains(fieldName) && !depFields.contains(fieldName)
+        }
+
+        val wrappedChild = if (insertOverwritesUnrequestedButPreservedField) {
+          val selectedChildFields = preservedChildFields.filter(s => depFields.contains(s))
+          SelectFields(rebuiltChild, rebuiltChild.typ.asInstanceOf[TStruct].fieldNames.filter(selectedChildFields.contains(_)))
+        } else {
+          rebuiltChild
+        }
+
```
johnc1231, line 1941: Handled by #9633

johnc1231:
Closing, will make a new PR with just the linreg stuff now that the pruning PRs are starting to go in.

PR #9579, opened 2020-10-07
[batch] add close billing project functionality

catoverdrive:
@jigold this is more-or-less as we talked about yesterday. I added a check within batch creation so it can only open a batch against an open billing project, and when we close a billing project we check that there all of the batches associated with it are completed (can't close a billing project mid-batch).

I don't have automated tests for this yet (planning on exposing the API in the billing project test PR) but I dev-deployed it and clicked the buttons and it does the right thing.


batch/batch/front_end/front_end.py
```diff
@@ -1327,7 +1336,8 @@ async def ui_get_billing_projects(request, userdata):
     db = request.app['db']
     billing_projects = await _query_billing_projects(db)
     page_context = {
-        'billing_projects': billing_projects
+        'billing_projects': [dict(p, size=len(p['users'])) for p in billing_projects if not p['closed']],
```
jigold, line 1339: What does the `size=` do?

batch/batch/front_end/front_end.py
```diff
@@ -810,7 +812,11 @@ async def create_batch(request, userdata):
     async def insert(tx):
         rows = tx.execute_and_fetchall(
             '''
-SELECT * FROM billing_project_users
+SELECT *, (
```
jigold, line 815: Is the reason for doing the subquery like this so you don't have to do a group by?

batch/batch/front_end/front_end.py
```diff
@@ -1489,6 +1499,89 @@ async def insert(tx):
...
     return web.HTTPFound(deploy_config.external_url('batch', '/billing_projects'))
 
+async def _close_billing_project(db, billing_project):
+    @transaction(db)
+    async def close_project(tx):
+        row = await tx.execute_and_fetchone(
+            '''
+SELECT name, closed,
+  (SELECT id FROM batches WHERE 
+    billing_project = billing_projects.name AND
```
jigold, line 1509: I think we need a new data structure for this that tracks the number of open batches, running batches, etc. per billing project. This will get updated with a trigger. Otherwise, maybe an index on the batches table with billing_project and state might do it, but I'm worried that will be slow. Not sure.

batch/batch/front_end/front_end.py
```diff
@@ -1489,6 +1499,89 @@ async def insert(tx):
...
+        if not row:
+            raise KeyError(billing_project)
+        assert row['name'] == billing_project
+        if row['closed']:
+            raise ValueError(f'Billing project {billing_project} is already closed.')
+        if row['batch'] is not None:
+            raise RuntimeError(f'Billing project {billing_project} has open or running batches.')
+
+        await tx.execute_update(
+            'UPDATE billing_projects SET closed = TRUE WHERE name = %s;',
```
jigold, line 1523: I think you can say `closed = 1`. That's what we do in other places for booleans.

batch/batch/front_end/front_end.py
```diff
@@ -1489,6 +1499,89 @@ async def insert(tx):
...
+async def post_close_billing_projects(request, userdata):  # pylint: disable=unused-argument
+    db = request.app['db']
+    billing_project = request.match_info['billing_project']
+
+    session = await aiohttp_session.get_session(request)
+    try:
+        await _close_billing_project(db, billing_project)
+        set_message(session, f'Closed billing project {billing_project}.', 'info')
+    except KeyError as e:
+        set_message(session, f'Billing project {str(e)} does not exist.', 'error')
```
jigold, line 1541: This setup is clever!

batch/batch/front_end/front_end.py
```diff
@@ -1489,6 +1499,89 @@ async def insert(tx):
     set_message(session, f'Added billing project {billing_project}.', 'info')
     return web.HTTPFound(deploy_config.external_url('batch', '/billing_projects'))
 
+async def _close_billing_project(db, billing_project):
+    @transaction(db)
+    async def close_project(tx):
```
jigold, line 1504: Where are the REST API endpoints? Are those coming later?

batch/batch/front_end/front_end.py
```diff
@@ -1489,6 +1499,89 @@ async def insert(tx):
...
+            'SELECT name, closed FROM billing_projects WHERE name = %s FOR UPDATE;',
+            (billing_project,))
+        if not row:
+            raise KeyError(billing_project)
+        assert row['name'] == billing_project
+        if not row['closed']:
+            raise ValueError(f'Billing project {billing_project} is already open.')
+
+        await tx.execute_update(
+            'UPDATE billing_projects SET closed = FALSE WHERE name = %s;',
```
jigold, line 1562: `closed = 0`

batch/batch/front_end/front_end.py
```diff
@@ -1489,6 +1499,89 @@ async def insert(tx):
...
+@check_csrf_token
+@web_authenticated_developers_only(redirect=False)
+async def post_reopen_billing_projects(request, userdata):  # pylint: disable=unused-argument
+    db = request.app['db']
+    billing_project = request.match_info['billing_project']
+
+    session = await aiohttp_session.get_session(request)
+    try:
+        await _reopen_billing_project(db, billing_project)
+        set_message(session, f'Re-opened billing project {billing_project}.', 'info')
```
jigold, line 1578: I think Dan wanted me in a different PR to put this not in the try except and put it in `else`

batch/sql/add-closed-flag-billing-projects.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE billing_projects ADD COLUMN closed BOOLEAN NOT NULL DEFAULT FALSE;
```
jigold, line 1: Same thing. Look at what we do for other boolean default values.

batch/batch/front_end/templates/billing_projects.html
```diff
@@ -3,46 +3,50 @@
...
     </thead>
-    <tbody>
       {% for billing_project in billing_projects %}
+    <tbody>
       <tr>
-        <td>{{ billing_project['billing_project'] }}</td>
-        <td></td>
-        <td></td>
-      </tr>
+        <td rowspan="{{ billing_project.size + 1 }}">
```
jigold, line 17: Can you add a picture of what this looks like? My CSS/HTML knowledge is pretty limited.

jigold (changes requested):

batch/batch/front_end/front_end.py
jigold: Ah. Reread your comment.

batch/batch/front_end/front_end.py
catoverdrive: The size is purely for page formatting reasons. I adjusted the billing project display page so that the project name is a multi-row cell that spans all of the user rows (so `rowspan=len(users) + 1`); jinja didn't like when I tried to compute this within the template.

batch/batch/front_end/front_end.py
catoverdrive: I did it like this because I need to lock the billing_projects table to check that there's not the exclusive lock from the `_close_billing_project` transaction; not sure if there's a better way to do it.

batch/batch/front_end/front_end.py
catoverdrive: Good call.

batch/batch/front_end/templates/billing_projects.html
catoverdrive: Screen Shot 2020-10-07 at 4 53 07 PM

batch/batch/front_end/front_end.py
catoverdrive: Hmm, good call. Do we need to do something similar for generating billing reports? Can I use what we're doing there?

I don't think we need this query to be super fast, since it should be rarely called and only on billing projects without any new activity, but definitely querying the entire `batches` table seems like a bad plan as it grows.

catoverdrive: I could also rewrite this as `SELECT id FROM batches WHERE time_completed IS NULL and billing_project = billing_projects.name`, right? That should take advantage of the index on `time_completed` while giving the same result?

batch/batch/front_end/templates/billing_projects.html
catoverdrive: (the part that I forgot to illustrate here is what happens if there are users in a closed billing project; I just display them on the same line as a comma-separated list.)

catoverdrive: Update: I'm tweaking this to combine the header cells for the "User" and <blank> (add/remove button) columns like I did for "Billing Project" and also to move the "Add" button in the last row under the "Close" button for the billing projects. I'll post a screenshot once it's done dev deploying.

catoverdrive: Screen Shot 2020-10-08 at 12 40 26 PM

catoverdrive:
ok, I've updated this PR so that billing projects can be "open", "closed", or "deleted". Allowed flow is `open` <--> `closed` --> `deleted`. Closed billing projects will still show up in the UI and billing reports; deleted projects show up in neither, but batches belonging to deleted billing projects will still show up in the batches table unless they're deleted.

There's no endpoint here for deleting closed projects; I'll add an API endpoint in the billing project API PR so that I can run the tests properly there, but currently don't intend to add a button to the UI.

batch/batch/front_end/templates/billing_projects.html
jigold: I'm not sure we need the list of users.

batch/batch/front_end/front_end.py
```diff
@@ -1434,18 +1466,23 @@ async def post_billing_projects_add_user(request, userdata):  # pylint: disable=
...
   ON billing_projects.name = t.billing_project
-WHERE billing_projects.name = %s;
+WHERE billing_projects.name = %s AND NOT billing_projects.closed;
 ''',
             (billing_project, user, billing_project))
         if row is None:
             set_message(session, f'No such billing project {billing_project}.', 'error')
             raise web.HTTPFound(deploy_config.external_url('batch', '/billing_projects'))
 
+        if row['status'] in {'closed', 'deleted'}:
```
jigold, line 1482: I think the filter already filters out closed billing projects. I don't think you need that if you're checking it here.

batch/batch/front_end/front_end.py
```diff
@@ -1285,7 +1310,7 @@ async def ui_get_billing(request, userdata):
 async def _query_billing_projects(db, user=None, billing_project=None):
     args = []
 
-    where_conditions = []
+    where_conditions = ["billing_projects.status != 'deleted'"]
```
jigold, line 1313: I'm not sure, but you might have to put status in backticks. I can't remember if it's a special word.

jigold, line 1313: Eventually, will you have a test that ensures the rest endpoint get_billing_project fails if the project is deleted?

batch/batch/front_end/front_end.py
jigold: @danking Can you comment on this query?

batch/batch/front_end/templates/billing_projects.html
```diff
@@ -3,57 +3,88 @@
...
-        <th>Billing Project</th>
-        <th>User</th>
-        <th></th>
+        <th colspan="2">Billing Project</th>
+        <th colspan="2">User</th>
       </tr>
     </thead>
-    <tbody>
       {% for billing_project in billing_projects %}
+    <tbody>
```
jigold, line 14: Why is the tbody under the for loop now?

jigold (changes requested):
This looks great!

batch/batch/front_end/front_end.py
```diff
@@ -87,6 +92,16 @@ async def get_healthcheck(request):  # pylint: disable=W0613
     return web.Response()
 
 
+async def _handle_ui_error(session, f, *args, **kwargs):
```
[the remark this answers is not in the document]
catoverdrive: yeah! I'll add that in the billing project api PR.

batch/batch/front_end/templates/billing_projects.html
catoverdrive: I'll take it out.

batch/batch/front_end/templates/billing_projects.html
catoverdrive: Each billing project is now wrapped in its own `<tbody>` tag.

The big UI change I made here were to treat the billing project (and the "close" button) as a multi-row cell that spans the height of its users, instead of having it as a distinct row in front of the users. I prefer this because the visual grouping is cleaner, but I can switch it back if you'd prefer.

What this meant was that there was a weird color thing going on with the billing project cell depending on whether the billing project had an even or odd number of users (since they're colored according to the color of the top row, since they technically "live" there). 

I solved this by changing the color pattern to be alternate for every other *billing project*, but each user still needed to have its own row. This meant wrapping each billing project in a `<tbody>` tag and defining a new type of table that has the `:nth-of-type(even)` style definition on the `tbody` tab instead of the `tr` tag.

This means that there's no color differentiation between users in a row, but the table is short enough horizontally that I think I prefer this to the other way (visually differentiated users but not billing projects). The on-hover background change still applies to row, so it's still pretty obvious which user you're trying to remove.

batch/batch/front_end/front_end.py
```diff
@@ -34,6 +34,9 @@
                      cost_from_msec_mcpu, adjust_cores_for_packability, coalesce,
                      adjust_cores_for_storage_request, total_worker_storage_gib)
 from ..batch import batch_record_to_dict, job_record_to_dict
+from ..exceptions import (BatchUserError, NonExistentBillingProjectError,
+                          DeletedBillingProjectError,
+                          UncloseableBillingProjectError)
```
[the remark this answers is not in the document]
catoverdrive: I decided to take it out, but forgot to remove the import >.<

batch/batch/front_end/front_end.py
catoverdrive: yeah, good call. I actually forgot to rewrite the query from the old schema, so I fixed it and it now filters for "deleted" but not closed so it 

(I think I'll plan to treat deleted projects the same as non-existent projects everywhere except billing project creation, where you'll get an explicit error about creating a project with the same name as a deleted billing project. This means that deleted billing projects "don't exist" for basically all purposes.)

batch/batch/front_end/front_end.py
catoverdrive: Good call. I'll add that in the other PR.

batch/batch/front_end/front_end.py
```diff
@@ -1301,8 +1325,9 @@ async def _query_billing_projects(db, user=None, billing_project=None):
         where_condition = ''
 
     sql = f'''
-SELECT billing_projects.name as billing_project, users
-FROM (
+SELECT billing_projects.name as billing_project,
+billing_projects.`status` as `status`,
+users FROM (
```
danking, line 1330: the syntax here is a bit funky, maybe:
```sql
SELECT billing_projects.name as billing_project,
       billing_projects.`status` as `status`,
       users
FROM (...
```

batch/batch/front_end/front_end.py
```diff
@@ -1301,8 +1325,9 @@ async def _query_billing_projects(db, user=None, billing_project=None):
...
   SELECT billing_project, JSON_ARRAYAGG(`user`) as users
   FROM billing_project_users
   GROUP BY billing_project
```
danking, line 1333: Hmm, (not your problem, but) I think this query can also be written without a sub-query:
```sql
SELECT billing_projects.name as billing_project,
       billing_projects.`status` as `status`,
       IF (COUNT(`user`) = 0, JSON_ARRAY(), JSON_ARRAYAGG(`user`)) as users
FROM billing_projects
LEFT JOIN billing_project_users
ON billing_projects.name = billing_project_users.billing_project
GROUP BY billing_projects.name, `status`
{where_condition}
```
```
mysql> explain SELECT billing_projects.name as billing_project, users
    -> FROM (
    ->   SELECT billing_project, JSON_ARRAYAGG(`user`) as users
    ->   FROM billing_project_users
    ->   GROUP BY billing_project
    -> ) AS t
    -> RIGHT JOIN billing_projects
    ->   ON t.billing_project = billing_projects.name
    -> ;
+----+-------------+-----------------------+------------+-------+---------------+------------------------+---------+------+------+----------+----------------------------------------------------+
| id | select_type | table                 | partitions | type  | possible_keys | key                    | key_len | ref  | rows | filtered | Extra                                              |
+----+-------------+-----------------------+------------+-------+---------------+------------------------+---------+------+------+----------+----------------------------------------------------+
|  1 | PRIMARY     | billing_projects      | NULL       | index | NULL          | billing_project_status | 1       | NULL |   25 |   100.00 | Using index                                        |
|  1 | PRIMARY     | <derived2>            | NULL       | ALL   | NULL          | NULL                   | NULL    | NULL |    6 |   100.00 | Using where; Using join buffer (Block Nested Loop) |
|  2 | DERIVED     | billing_project_users | NULL       | index | PRIMARY       | PRIMARY                | 604     | NULL |    6 |   100.00 | Using index; Using filesort                        |
+----+-------------+-----------------------+------------+-------+---------------+------------------------+---------+------+------+----------+----------------------------------------------------+
3 rows in set, 1 warning (0.00 sec)

mysql> explain SELECT billing_projects.name as billing_project,
    ->        billing_projects.`status` as `status`,
    ->        IF (COUNT(`user`) = 0, JSON_ARRAY(), JSON_ARRAYAGG(`user`)) as users
    -> FROM billing_projects
    -> LEFT JOIN billing_project_users
    -> ON billing_projects.name = billing_project_users.billing_project
    -> GROUP BY billing_projects.name, `status`
    -> ;
+----+-------------+-----------------------+------------+-------+------------------------+------------------------+---------+----------------------------+------+----------+-----------------------------+
| id | select_type | table                 | partitions | type  | possible_keys          | key                    | key_len | ref                        | rows | filtered | Extra                       |
+----+-------------+-----------------------+------------+-------+------------------------+------------------------+---------+----------------------------+------+----------+-----------------------------+
|  1 | SIMPLE      | billing_projects      | NULL       | index | billing_project_status | billing_project_status | 1       | NULL                       |   25 |   100.00 | Using index; Using filesort |
|  1 | SIMPLE      | billing_project_users | NULL       | ref   | PRIMARY                | PRIMARY                | 302     | wang.billing_projects.name |    1 |   100.00 | Using index                 |
+----+-------------+-----------------------+------------+-------+------------------------+------------------------+---------+----------------------------+------+----------+-----------------------------+
2 rows in set, 1 warning (0.01 sec)
```

The `IF` kind of sucks, but I think that's the only way to deal with the left joins.

danking: I'll PR this separately.

batch/batch/front_end/front_end.py
```diff
@@ -1418,46 +1449,53 @@ async def delete(tx):
...
+async def _add_user_to_billing_project(db, billing_project, user):
     @transaction(db)
     async def insert(tx):
         row = await tx.execute_and_fetchone(
             '''
-SELECT billing_projects.name as billing_project, user
-FROM billing_projects
+SELECT billing_projects.name as billing_project,
+billing_projects.`status` as `status`,
+user FROM billing_projects
```
danking, line 1459: sql syntax again here

batch/batch/front_end/front_end.py
```diff
@@ -1418,46 +1449,53 @@ async def delete(tx):
...
 LEFT JOIN (SELECT * FROM billing_project_users
-    WHERE billing_project = %s AND user = %s FOR UPDATE) AS t
-  ON billing_projects.name = t.billing_project
-WHERE billing_projects.name = %s;
-''',
+WHERE billing_project = %s AND user = %s FOR UPDATE) AS t
```
danking, line 1461: Hmm. Not relevant to your change, but I think we ought to lock `billing_projects` as well. It'd be extra confusing if you're able to race to insert a user into a billing project with deleting that billing project.

batch/batch/front_end/front_end.py
```diff
@@ -1418,46 +1449,53 @@ async def delete(tx):
...
+ON billing_projects.name = t.billing_project
+WHERE billing_projects.name = %s AND billing_projects.`status` != 'deleted';
+        ''',
```
danking, line 1464: I think I'd actually do this query as:

```sql
SELECT billing_projects.name as billing_project,
       billing_projects.`status` as `status`,
       user
FROM billing_projects
INNER JOIN billing_project_users
ON billing_projects.name = billing_project_users.billing_project
WHERE billing_projects.name = %s
  AND billing_projects.`status` != 'deleted'
  AND billing_projects.user = %s
FOR UPDATE
```

batch/batch/front_end/front_end.py
```diff
@@ -1418,46 +1449,53 @@ async def delete(tx):
...
-            set_message(session, f'User {user} is already member of billing project {billing_project}.', 'info')
-            raise web.HTTPFound(deploy_config.external_url('batch', '/billing_projects'))
+            raise BatchUserError(f'User {user} is already member of billing project {billing_project}.', 'info')
 
         await tx.execute_insertone(
             '''
 INSERT INTO billing_project_users(billing_project, user)
 VALUES (%s, %s);
-''',
+        ''',
```
danking, line 1479: nix this, or move the `'''` to the end of the VALUES line.

batch/batch/front_end/front_end.py
```diff
@@ -1496,6 +1534,81 @@ async def insert(tx):
...
+    @transaction(db)
+    async def close_project(tx):
+        row = await tx.execute_and_fetchone(
+            '''
+SELECT name, `status`,
+  (SELECT 1 FROM batches WHERE
+    time_completed IS NULL AND
+    billing_project = billing_projects.name
+    LIMIT 1) AS batch
+FROM billing_projects WHERE name = %s and `status` != 'deleted' FOR UPDATE;
```
danking, line 1547:
```sql
select name, `status`, batches.id as batch_id
from billing_projects
left join batches
on billing_projects.name = batches.billing_project
where name = %s and `status` != 'deleted' and batches.time_completed is NULL
limit 1
FOR UPDATE
```
I suppose you might be intentionally not locking the batches table? I guess this is safe because time_completed should only transition from NULL to not NULL and if it happens to transition from NULL to not NULL while you're executing this transaction, that doesn't change the validity of this transaction.

batch/batch/front_end/templates/billing_projects.html
danking: This is really cool! I love this UI innovation.

danking (changes requested):
This is a lot of awesome stuff. Just a few comments, mostly about SQL.

batch/batch/front_end/front_end.py
catoverdrive: Thanks for the explanation! I'll rewrite it as an inner join.

batch/batch/front_end/front_end.py
catoverdrive: Hm… I don't need to lock the `billing_projects` for update, though---seems perfectly valid to e.g. simultaneously add a bunch of users to the same billing project.

catoverdrive: Can I write this join with slightly more granularity on the locks?

batch/batch/front_end/front_end.py
catoverdrive: Yeah, I'll lock this.

batch/batch/front_end/front_end.py
catoverdrive: Yeah, that's true. I wasn't intentionally not locking it; it would still be good to lock it to prevent batch creation during this transaction, in fact, but again, I think I'd like to say "shared lock for this table and exclusive lock for this one, please" inside the join.

danking (changes requested):

batch/batch/front_end/front_end.py
catoverdrive: I want to preserve the behavior where we pop up the different messages when a user is added through the UI (although I'll do this for the API version)---currently neither are errors, but one will tell the user "X has been added to billing project Y" vs "X is already a member of billing project Y".

Also, as discussed, I'm keeping the subquery structure here since it was already there (but putting a shared lock on the billing projects table), making the update to the batches table query as you suggested below, and we'll revisit this in a separate PR.

danking (approved):

PR #9581, opened 2020-10-08
[batch] Add infrastructure for not running batch tests in production

jigold:
This is for @catoverdrive and I to be able to manipulate billing projects and their limits in tests. But we don't want to litter the default namespace with a bunch of bogus projects. We also need to reset the accrued costs to 0 in order for the billing limit tests to work.

@danking -- FYI


catoverdrive (approved):

catoverdrive:
@jigold not sure why this test is being marked as "failed"---maybe because there are currently no tests in that filter?

PR #9584, opened 2020-10-08
[batch] refactor test_batch.py to use pytest instead of unittest

jigold:
I did this because I think we want to add more flexibility and use more pytest features like how you wrote the two different clients and I want to add pyest marks.


catoverdrive (approved):
Drives me totally nuts that the diff doesn't realize that most of these changes are just "unindent and use two lines of separation instead of one", but this looks good.

PR #9585, opened 2020-10-08
[hailtop-auth] allow tokens to be loaded from user-specified file

catoverdrive:
I need to be able to load tokens for multiple users for #9553 and would like to pipe it through the auth utilities in a reasonable way so I can use them. Let me know if this looks ok or if there's a better way to do it?

### DIFF
--- a/hail/python/hailtop/auth/auth.py
+++ b/hail/python/hailtop/auth/auth.py
@@ -24,7 +24,7 @@ def get_userinfo(deploy_config=None):
     return async_to_blocking(async_get_userinfo(deploy_config))
 
 
-def namespace_auth_headers(deploy_config, ns, authorize_target=True, _token_file=None):
+def namespace_auth_headers(deploy_config, ns, authorize_target=True, *, _token_file=None):
     tokens = get_tokens(_token_file)
     headers = {}
     if authorize_target:
@@ -34,9 +34,9 @@ def namespace_auth_headers(deploy_config, ns, authorize_target=True, _token_file
     return headers
 
 
-def service_auth_headers(deploy_config, service, authorize_target=True, _token_file=None):
+def service_auth_headers(deploy_config, service, authorize_target=True, *, _token_file=None):
     ns = deploy_config.service_ns(service)
-    return namespace_auth_headers(deploy_config, ns, authorize_target, _token_file)
+    return namespace_auth_headers(deploy_config, ns, authorize_target, _token_file=_token_file)
 
 
 def copy_paste_login(copy_paste_token, namespace=None):
--- a/hail/python/hailtop/auth/auth.py
+++ b/hail/python/hailtop/auth/auth.py
@@ -24,8 +24,8 @@ def get_userinfo(deploy_config=None):
     return async_to_blocking(async_get_userinfo(deploy_config))
 
 
-def namespace_auth_headers(deploy_config, ns, authorize_target=True, *, _token_file=None):
-    tokens = get_tokens(_token_file)
+def namespace_auth_headers(deploy_config, ns, authorize_target=True, *, token_file=None):
+    tokens = get_tokens(token_file)
     headers = {}
     if authorize_target:
         headers['Authorization'] = f'Bearer {tokens.namespace_token_or_error(ns)}'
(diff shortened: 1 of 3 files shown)


hail/python/hailtop/auth/auth.py
```diff
@@ -34,9 +34,9 @@ def namespace_auth_headers(deploy_config, ns, authorize_target=True):
     return headers
 
 
-def service_auth_headers(deploy_config, service, authorize_target=True):
+def service_auth_headers(deploy_config, service, authorize_target=True, _token_file=None):
     ns = deploy_config.service_ns(service)
-    return namespace_auth_headers(deploy_config, ns, authorize_target)
+    return namespace_auth_headers(deploy_config, ns, authorize_target, _token_file)
```
jigold, line 39: I think it's a good idea to make these kwargs like `_token_file=_token_file`. We've run into problems in the past where the arguments get out of order for the kwargs ones.

hail/python/hailtop/auth/tokens.py
```diff
@@ -64,12 +63,15 @@ def write(self):
...
 
 
-def get_tokens():
+def get_tokens(file=None):
     global tokens
 
-    if not tokens:
-        tokens = Tokens()
-    return tokens
+    if file is None:
```
jigold, line 72: I had it like this because I don't think tokens is a dictionary of file:{tokens}. 

```
class Tokens(collections.abc.MutableMapping):
    def __init__(self, tokens_file=None):
        if tokens_file is None:
            tokens_file = self.get_tokens_file()
        if os.path.isfile(tokens_file):
            with open(tokens_file, 'r') as f:
                self._tokens = json.load(f)
            log.info(f'tokens loaded from {tokens_file}')
        else:
            log.info(f'tokens file not found: {tokens_file}')
            self._tokens = {}

def get_tokens(tokens_file=None):
    global tokens

    if not tokens:
        tokens = Tokens(tokens_file=tokens_file)
    return tokens
```

jigold (changes requested):

hail/python/hailtop/auth/tokens.py
catoverdrive: It's not currently! But I need to be able to call `get_tokens` for multiple different token files in the same python environment, so I changed it to be a dictionary instead of a single token. The use case is something like 
```
dev_tokens=get_tokens(dev_file)
tokens = get_tokens(normal_file)
```

which doesn't return two separate tokens if I'm only caching the first file.

jigold (dismissed):

hail/python/hailtop/auth/auth.py
```diff
@@ -24,8 +24,8 @@ def get_userinfo(deploy_config=None):
     return async_to_blocking(async_get_userinfo(deploy_config))
 
 
-def namespace_auth_headers(deploy_config, ns, authorize_target=True):
-    tokens = get_tokens()
+def namespace_auth_headers(deploy_config, ns, authorize_target=True, *, _token_file=None):
```
danking, line 27: All this stuff is fully internal, I'm inclined to not use the underscore prefix. I read `_` in services code function parameters "I don't think this through fully; be careful".

hail/python/hailtop/auth/tokens.py
```diff
@@ -17,8 +17,7 @@ def get_tokens_file():
             return os.path.expanduser('~/.hail/tokens.json')
         return '/user-tokens/tokens.json'
 
-    def __init__(self):
-        tokens_file = self.get_tokens_file()
+    def __init__(self, tokens_file):
         if os.path.isfile(tokens_file):
             with open(tokens_file, 'r') as f:
                 self._tokens = json.load(f)
```
danking, line 23: If you specify a tokens file and it doesn't exist, that feels like it should be an error.

danking: What do you think of this?
```python3
@staticmethod
def default_tokens():
    default_tokens_file = self.get_tokens_file()
    if os.path.isfile(default_tokens_file):
        return Tokens.tokens_from_file(default_tokens_file)
    return Tokens({})

@staticmethod
def tokens_from_file(fname):
    with open(fname, 'r') as f:
        return Tokens(json.load(f))

def __init__(self, tokens):
    self._tokens = tokens
```

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -552,7 +552,7 @@ async def submit(self,
 @asyncinit
 class BatchClient:
     async def __init__(self, billing_project, deploy_config=None, session=None,
-                       headers=None, _token=None):
+                       headers=None, _token=None, _token_file=None):
```
danking, line 555: same thoughts here. I also don't think _token is used. We can remove it if so.

danking (changes requested):
This kind of clashes with the deploy config.

hail/python/hailtop/batch_client/aioclient.py
catoverdrive: `_token` is used in one of the tests.

hail/python/hailtop/auth/tokens.py
catoverdrive: Yeah, that makes sense. I'll change it.

danking (approved):

PR #9586, opened 2020-10-08
[query] Add lowered execution path for `hl.read_matrix_table`

tpoterba:
11 new tests passing.

Can briefly walk you through the change in 1:1.


johnc1231 (approved):
:+1:

PR #9591, opened 2020-10-09
[query] RVD to table stage infrastructure

tpoterba:
Currently tested with TableLiteral on the Spark backend.

### DIFF
--- a/hail/src/main/scala/is/hail/io/CodecSpec.scala
+++ b/hail/src/main/scala/is/hail/io/CodecSpec.scala
@@ -24,6 +24,8 @@ trait AbstractTypedCodecSpec extends Spec {
 
   def decodedPType(requestedType: Type): PType
 
+  def decodedPType(): PType = encodedType.decodedPType(encodedVirtualType)
+
   def buildDecoder(ctx: ExecuteContext, requestedType: Type): (PType, (InputStream) => Decoder)
 
   def encode(ctx: ExecuteContext, t: PType, offset: Long): Array[Byte] = {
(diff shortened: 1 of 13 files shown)


hail/src/main/scala/is/hail/rvd/RVD.scala
```diff
@@ -1534,3 +1534,13 @@ object RVD {
     fileData
   }
 }
+
+case class BroadcastRVD(@transient val backend: SparkBackend, @transient val rvd: RVD) extends Serializable {
+  private[this] val crdd = rvd.crdd
+  private[this] val bcPartitions = backend.broadcast(rvd.crdd.partitions)
```
tpoterba, line 1540: longer-term, we should serialize the partition to a TBinary inside the context and use that.

hail/src/main/scala/is/hail/io/CodecSpec.scala
```diff
@@ -24,6 +24,8 @@ trait AbstractTypedCodecSpec extends Spec {
 
   def decodedPType(requestedType: Type): PType
 
+  def decodedPType(): PType = encodedType.decodedPType(encodedVirtualType)
```
tpoterba, line 27: moved stuff up from the concrete TypedCodecSpec implementation so it's on the abstract interface, which TableLiteral uses.

hail/src/main/scala/is/hail/rvd/RVD.scala
```diff
@@ -1534,3 +1534,13 @@ object RVD {
     fileData
   }
 }
+
+case class BroadcastRVD(@transient val backend: SparkBackend, @transient val rvd: RVD) extends Serializable {
```
patrick-schultz, line 1538: It looks to me like these don't need to be `val`s at all (and this doesn't need to be a case class), which should avoid the need for `transient`. In general, I agree with Cotton that we make too many things case classes.

hail/src/main/scala/is/hail/types/physical/PPrimitive.scala
```diff
@@ -82,4 +82,6 @@ class PPrimitiveCode(val pt: PType, val code: Code[_]) extends PCode {
   def memoize(cb: EmitCodeBuilder, name: String): PValue = defaultMemoizeImpl(cb, name)
 
   def memoizeField(cb: EmitCodeBuilder, name: String): PValue = defaultMemoizeFieldImpl(cb, name)
+
+  def primCode[T]: Code[T] = coerce[T](code)
```
patrick-schultz, line 86: I assume you're not using `PCode.tcode` because that will go away but this will stay? Probably a good idea to copy the implementation of `tcode`, which checks that the `T` agrees with the ptype.

patrick-schultz (changes requested):
Looks good, just a couple requests.

hail/src/main/scala/is/hail/rvd/RVD.scala
tpoterba: fair.

hail/src/main/scala/is/hail/types/physical/PPrimitive.scala
tpoterba: correct. I'll add that.

patrick-schultz (approved):

PR #9593, opened 2020-10-09
[docker] Introduce Hail Ubuntu Image and hail-pip-install

danking:
We've had a number of pip issues recently, particularly incompatibilities with the `six` package. The central issue is that pip will install incompatible package versions with only a warning message. This primarily happens when a set of packages is installed (say in a parent image) and then a new set of packages are installed later. Pip attempts to satisfy all requirements of the packages in an installed set, but it does not consider the versions of packages already installed.

In this PR, I pervasively change every use of `pip` to be followed by the use of `pip check`. Pip check errors if there are incompatible package versions installed. Rather than change every place that we use `pip install`, I created an ubuntu image with a script, `hail-pip-install`, and pip configuration which makes it easy to do the right thing. Now, we should always install python packages like this:

```
hail-pip-install package1 package2 package3
```

The script will ensure that all packages are compatible with existing packages. Moreover, it will retry downloads five times.

I did not make every Dockerfile depend on the new hail-ubuntu-image because they are legacy Dockerfile that I didn't want to edit. These include the benchmark Dockerfile and some notebook workers.

PySpark presented a challenge because we did not install it via pip. As a result, packages, such as Hail, which depend on PySpark failed the `pip check`. I modified all uses of PySpark to instead use the standard pip-installed version of PySpark. In particular, take a look at how query and shuffler changed.

cc: @Dania-Abuhijleh @catoverdrive

### DIFF
--- a/auth/Dockerfile
+++ b/auth/Dockerfile
@@ -1,12 +1,11 @@
 FROM {{ service_base_image.image }}
 
-RUN pip3 install --upgrade \
+RUN hail-pip-install \
       google-auth-oauthlib==0.4.1 \
-      google-auth==1.20.0
+      google-auth==1.21.1
 
 COPY auth/setup.py auth/MANIFEST.in /auth/
 COPY auth/auth /auth/auth/
-RUN pip3 install --no-cache-dir /auth && \
-  rm -rf /auth
+RUN hail-pip-install /auth && rm -rf /auth
 
 EXPOSE 5000
--- a/batch/Dockerfile.worker
+++ b/batch/Dockerfile.worker
@@ -17,22 +17,21 @@ RUN apt-get update && \
     gcsfuse && \
   rm -rf /var/lib/apt/lists/*
 
+COPY docker/hail-ubuntu/pip.conf /root/.config/pip/pip.conf
+COPY docker/hail-ubuntu/hail-pip-install /bin/hail-pip-install
+RUN python3 -m pip install --upgrade --no-cache-dir pip
+
 COPY docker/requirements.txt .
-# re: pip install --upgrade pip: https://github.com/grpc/grpc/issues/22815
-RUN python3 -m pip install --upgrade pip && \
-    python3 -m pip install --no-cache-dir -U -r requirements.txt
+RUN hail-pip-install -r requirements.txt
 
 COPY hail/python/setup-hailtop.py /hailtop/setup.py
 COPY hail/python/hailtop /hailtop/hailtop/
-RUN python3 -m pip install --no-deps --no-cache-dir /hailtop \
-  && rm -rf /hailtop
+RUN hail-pip-install --no-deps /hailtop && rm -rf /hailtop
 
 COPY gear/setup.py /gear/setup.py
 COPY gear/gear /gear/gear/
-RUN python3 -m pip install --no-deps --no-cache-dir /gear \
-  && rm -rf /gear
+RUN hail-pip-install --no-deps /gear && rm -rf /gear
 
 COPY batch/setup.py batch/MANIFEST.in /batch/
 COPY batch/batch /batch/batch/
-RUN pip3 install --no-cache-dir /batch && \
-  rm -rf /batch
+RUN hail-pip-install --no-deps /batch && rm -rf /batch
--- a/batch/Makefile
+++ b/batch/Makefile
@@ -24,7 +24,7 @@ build:
 	-docker pull gcr.io/$(PROJECT)/python:3.7-slim-stretch
 	-docker pull $(BATCH_WORKER_LATEST)
 	python3 ../ci/jinja2_render.py '{"global":{"project":"$(PROJECT)"}}' Dockerfile.worker Dockerfile.worker.out
-	docker build -t batch-worker -f Dockerfile.worker --cache-from batch-worker,$(BATCH_WORKER_LATEST),gcr.io/$(PROJECT)/python:3.7-slim-stretch ..
+	docker build -t batch-worker -f Dockerfile.worker.out --cache-from batch-worker,$(BATCH_WORKER_LATEST),gcr.io/$(PROJECT)/python:3.7-slim-stretch ..
 
 .PHONY: push
 push: build
--- a/blog/Dockerfile.nginx
+++ b/blog/Dockerfile.nginx
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM {{ hail_ubuntu_image.image }}
 
 RUN apt-get update -y && \
     apt-get install -y nginx && \
--- a/build.yaml
+++ b/build.yaml
@@ -1,4 +1,9 @@
 steps:
+ - kind: buildImage
+   name: hail_ubuntu_image
+   dockerFile: docker/hail-ubuntu/Dockerfile
+   contextPath: docker/hail-ubuntu
+   publishAs: hail-ubuntu
  - kind: createNamespace
    name: default_ns
    namespaceName: default
@@ -19,6 +24,8 @@ steps:
    dockerFile: echo/Dockerfile
    contextPath: echo
    publishAs: echo
+   dependsOn:
+     - hail_ubuntu_image
  - kind: deploy
    name: deploy_echo
    namespace:
@@ -89,6 +96,8 @@ steps:
    dockerFile: docker/Dockerfile.base
    contextPath: .
    publishAs: base
+   dependsOn:
+    - hail_ubuntu_image
  - kind: buildImage
    name: service_base_image
    dockerFile: docker/Dockerfile.service-base
@@ -534,7 +543,7 @@ steps:
    contextPath: router
    publishAs: router
    dependsOn:
-     - base_image
+     - hail_ubuntu_image
  - kind: buildImage
    name: batch_image
    dockerFile: batch/Dockerfile
@@ -968,6 +977,7 @@ steps:
    publishAs: hail-base
    dependsOn:
     - build_hail
+    - hail_ubuntu_image
    inputs:
     - from: /wheel-container.tar
       to: /wheel-container.tar
@@ -978,6 +988,7 @@ steps:
    publishAs: hail-base
    dependsOn:
     - build_hail
+    - hail_ubuntu_image
    inputs:
     - from: /wheel-container.tar
       to: /wheel-container.tar
@@ -2193,14 +2204,18 @@ steps:
    name: netcat_ubuntu_image
    dockerFile:
      inline: |
-       FROM gcr.io/{{ global.project }}/ubuntu:18.04
+       FROM {{ hail_ubuntu_image.image }}
        RUN apt-get update && apt-get install netcat -y
+   dependsOn:
+    - hail_ubuntu_image
  - kind: buildImage
    name: curl_image
    dockerFile:
      inline: |
-       FROM gcr.io/{{ global.project }}/ubuntu:18.04
+       FROM {{ hail_ubuntu_image.image }}
        RUN apt-get update && apt-get install curl -y
+   dependsOn:
+    - hail_ubuntu_image
  - kind: runImage
    name: test_memory
    image:
@@ -2681,7 +2696,7 @@ steps:
       namespace:
         valueFrom: default_ns.name
       mountPath: /ssl-config
-   timeout: 2400
+   timeout: 5400
    scopes:
     - test
     - dev
@@ -3111,6 +3126,8 @@ steps:
    dockerFile: blog/Dockerfile.nginx
    contextPath: blog
    publishAs: blog_nginx
+   dependsOn:
+    - hail_ubuntu_image
  - kind: deploy
    name: deploy_blog
    namespace:
@@ -3242,6 +3259,7 @@ steps:
    publishAs: hail-public
    dependsOn:
     - copy_files
+    - hail_ubuntu_image
    inputs:
      - from: /hail_pip_version
        to: /hail_pip_version
--- a/ci/test/test_ci.py
+++ b/ci/test/test_ci.py
@@ -39,5 +39,5 @@ async def test_deploy():
             log.info(f'returning {deploy_status} {failure_information}')
             return deploy_state, failure_information
 
-        deploy_state, failure_information = await asyncio.wait_for(wait_forever(), timeout=30 * 60)
+        deploy_state, failure_information = await wait_forever()
         assert deploy_state == 'success', str(failure_information)
--- a/docker/Dockerfile.base
+++ b/docker/Dockerfile.base
@@ -1,4 +1,4 @@
-FROM gcr.io/{{ global.project }}/ubuntu:18.04
+FROM {{ hail_ubuntu_image.image }}
 
 ENV LANG C.UTF-8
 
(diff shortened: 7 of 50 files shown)


batch/Dockerfile.worker
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7-slim-stretch
+FROM {{ hail_ubuntu_image.image }}
```
jigold, line 1: Can you compare the sizes of both images? I'm worried it will be substantially bigger and will affect worker startup time.

site/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM debian:9.5
 
-RUN apt-get update -y && \
+RUN apt-get update && \
```
jigold, line 3: What is the reason for this change?

shuffler/Dockerfile
```diff
@@ -1,7 +1,3 @@
 FROM {{ service_base_image.image }}
 
 COPY hail.jar /
-
-EXPOSE 443
```
jigold, line 5: Do you not need the port anymore?

query/query/query.py
```diff
@@ -189,10 +190,19 @@ async def on_startup(app):
...
+                                         capture_output=True,
+                                         check=False)
+        if find_spark_home.returncode != 0:
+            raise ValueError(f'''SPARK_HOME is not set and find_spark_home.py returned non-zero exit code:
+STDOUT:
+{find_spark_home.stdout}
+STDERR:
+{find_spark_home.stderr}''')
+        spark_home = find_spark_home.stdout.decode().strip()
+    port = launch_gateway(die_on_exit=True, classpath=f'{spark_home}/jars/*:/hail.jar')
```
jigold, line 205: Why do you not need the jarpath here?

docker/hail-ubuntu/pip.conf
```diff
@@ -0,0 +1,6 @@
+[global]
+timeout = 5
+use-feature = 2020-resolver
+upgrade = true
+no-cache-dir = true
+retries = 5
```
jigold, line 6: new line at end.

docker/hail-ubuntu/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:bionic-20200921
```
jigold, line 1: Should we check whether there is a newer image available?

docker/hail-ubuntu/hail-pip-install
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -ex
+
+python3 -m pip install "$@"
```
jigold, line 5: Is this guaranteed to only take the arguments in the statement we are working with. i.e.

```
hail-pip-install a b c && rm f
```

will only work give hail-pip-install the arguments a, b, and c

docker/Makefile
```diff
@@ -15,11 +18,19 @@ HAIL_PUBLIC_IMAGE = gcr.io/$(PROJECT)/hail-public:$(shell docker images -q --no-
 GENETICS_PUBLIC_LATEST = gcr.io/$(PROJECT)/genetics-public:latest
 GENETICS_PUBLIC_IMAGE = gcr.io/$(PROJECT)/genetics-public:$(shell docker images -q --no-trunc genetics-public:latest | sed -e 's,[^:]*:,,')
 
+.PHONY: hail-ubuntu
+hail-ubuntu: hail-ubuntu-stmp
+
+hail-ubuntu-stmp: hail-ubuntu/Dockerfile hail-ubuntu/hail_pip_install hail-ubuntu/pip.conf
```
jigold, line 24: Can you explain to me why there is this hail-ubuntu-stmp instead of just hail-ubuntu?

docker/Dockerfile.base
```diff
@@ -26,23 +23,15 @@ RUN /bin/sh -c 'curl https://sdk.cloud.google.com | bash' && \
     /google-cloud-sdk/bin/gcloud -q components install beta kubectl
 ENV PATH $PATH:/google-cloud-sdk/bin
 
-RUN wget -nv -O spark-2.4.0-bin-hadoop2.7.tgz https://archive.apache.org/dist/spark/spark-2.4.0/spark-2.4.0-bin-hadoop2.7.tgz && \
```
jigold, line 29: Why was this removed?

docker/Dockerfile.base
```diff
@@ -26,23 +23,15 @@ RUN /bin/sh -c 'curl https://sdk.cloud.google.com | bash' && \
...
-
 COPY docker/requirements.txt .
-# re: pip install --upgrade pip: https://github.com/grpc/grpc/issues/22815
-RUN python3 -m pip install --upgrade pip && \
-    python3 -m pip install --no-cache-dir --upgrade -r requirements.txt
+RUN hail-pip-install -r requirements.txt
 
-ENV PYTHONPATH "${PYTHONPATH:+${PYTHONPATH}:}$SPARK_HOME/python:$SPARK_HOME/python/lib/py4j-0.10.7-src.zip"
+ENV SPARK_HOME /usr/local/lib/python3.7/dist-packages/pyspark
+ENV PATH "$PATH:$SPARK_HOME/sbin:$SPARK_HOME/bin"
```
jigold, line 30: Can you explain why this is simplified from what was there previously?

build.yaml
```diff
@@ -2549,7 +2564,7 @@ steps:
       namespace:
         valueFrom: default_ns.name
       mountPath: /ssl-config
-   timeout: 2400
+   timeout: 5400
```
jigold, line 2567: This seems crazy to me that we can have jobs that need 1.5 hours! Can you explain this to me again?

blog/Dockerfile.nginx
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM {{ hail_ubuntu_image.image }}
```
jigold, line 1: Does this image have nginx in it? Just trying to think of all the components that are not tested.

batch/Makefile
```diff
@@ -23,7 +23,8 @@ build:
 	docker build -t batch -f Dockerfile.out --cache-from batch,$(BATCH_LATEST),service-base ..
 	-docker pull python:3.7-slim-stretch
 	-docker pull $(BATCH_WORKER_LATEST)
-	docker build -t batch-worker -f Dockerfile.worker --cache-from batch-worker,$(BATCH_WORKER_LATEST),python:3.7-slim-stretch ..
+	python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"hail-ubuntu"}}' Dockerfile.worker Dockerfile.worker.out
+	docker build -t batch-worker -f Dockerfile.worker.out --cache-from batch-worker,$(BATCH_WORKER_LATEST),python:3.7-slim-stretch ..
```
jigold, line 27: This shouldn't cache from python:3.7... Instead it should be hail-ubuntu.

jigold (changes requested):
> PySpark presented a challenge because we did not install it via pip. As a result, packages, such as Hail, which depend on PySpark failed the pip check. I modified all uses of PySpark to instead use the standard pip-installed version of PySpark. In particular, take a look at how query and shuffler changed.

I'm a little hesitant to okay these. Can you get someone else to look specifically at those changes?

blog/Dockerfile.nginx
danking: The hail_ubuntu_image does not have nginx. The blog image does have nginx (its installed on line 4 of this file).

build.yaml
danking: I just picked a big number. It took more than 2400 seconds to rebuild every image from scratch. The base image, for example, took 18 minutes. Normally it takes 9 minutes. The context is that the CI-under-test has to use the batch-under-test which makes build jobs use 0.250 CPU. The machines are also tiny, so their bandwidth is lower.

docker/Dockerfile.base
danking: Because I install pyspark using pip instead. Downloading spark manually and not pip-installing it means that all our packages that depend on pyspark will fail `pip check` because pyspark isn't actually pip installed.

docker/Dockerfile.base
danking: This particular line is unchanged from the old line 38. The Spark home line refers to the pip-installed pyspark. The GCS connector (old lines 33-36) is now installed on lines 33-36.

Line 27 replaces the old lines dealing with Spark because pyspark is listed in the requirements.txt now.

docker/Makefile
danking: Cotton chose to create two targets for each image. `image-name` and `image-name-stmp`. I don't know why he didn't use a file named `image-name` to hold the timestamps. Instead, he chose to make `image-name` a PHONY target which just references `image-name-stmp`. The later is a file whose timestamp is updated when an image build is successfully run. File timestamps are used by Make to decide if it needs to re-run a rule.

docker/hail-ubuntu/Dockerfile
danking: It's still the latest version: https://hub.docker.com/_/ubuntu?tab=description.

Do you mean an automated check? That would be good, though I don't want to fail random PR builds if docker publishes a new ubuntu image. As we've discussed before, the real solution is a bot that automatically checks for new versions and makes PRs that update the version. For now, we'll have to periodically check for new versions and update our dependencies.

docker/hail-ubuntu/hail-pip-install
danking: Yes, in Bash, [&&, ||, and ;](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html) are some of the operators which terminate a command. This gets parsed by bash as:
```
COMMAND1 && COMMAND2
```
where `COMMAND1` is  `hail-pip-install a b c`. `$@` refers to the arguments to the currently executing script, so in this case `a`, `b`, and `c`

docker/hail-ubuntu/Dockerfile
jigold: No, I meant while we are updating this, is there a release from October.

docker/hail-ubuntu/hail-pip-install
jigold: That's what I thought, but I just wanted to make sure.

query/query/query.py
danking: PySpark and, consequently, Py4J are now installed in the proper location so py4j has no trouble finding its own jar.

shuffler/Dockerfile
danking: This never had any effect, kubernetes ignores it.

site/Dockerfile
danking: `-y` has no effect on `apt-get update`. It has always been unnecessary. I noticed it while auditing all our uses of pip and apt-get.

batch/Dockerfile.worker
danking: The Hail Ubuntu adds about 500MB. The worker image was 300MB previously. I switched it back to the python image.

batch/Dockerfile.worker
```diff
@@ -18,21 +18,16 @@ RUN apt-get update && \
   rm -rf /var/lib/apt/lists/*
 
 COPY docker/requirements.txt .
-# re: pip install --upgrade pip: https://github.com/grpc/grpc/issues/22815
-RUN python3 -m pip install --upgrade pip && \
-    python3 -m pip install --no-cache-dir -U -r requirements.txt
+RUN hail-pip-install -r requirements.txt
```
jigold, line 25: These are all going to fail now I think.

build.yaml
```diff
@@ -520,6 +527,8 @@ steps:
    dockerFile: batch/Dockerfile.worker
    contextPath: .
    publishAs: batch-worker
+   dependsOn:
+    - hail_ubuntu_image
```
jigold: This can be deleted now.

batch/Makefile
```diff
@@ -23,7 +23,8 @@ build:
 	docker build -t batch -f Dockerfile.out --cache-from batch,$(BATCH_LATEST),service-base ..
 	-docker pull python:3.7-slim-stretch
 	-docker pull $(BATCH_WORKER_LATEST)
-	docker build -t batch-worker -f Dockerfile.worker --cache-from batch-worker,$(BATCH_WORKER_LATEST),python:3.7-slim-stretch ..
+	python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"hail-ubuntu"}}' Dockerfile.worker Dockerfile.worker.out
```
jigold, line 26: You can remove this Jinja template command now.

docker/hail-ubuntu/pip.conf
jigold: Looks like this still needs to be fixed.

docker/Makefile
jigold: I remember now. He was checking the permissions because if the file permissions were wrong, then it made the builds slower.

jigold (changes requested):

batch/Dockerfile.worker
danking: I copied the script in above

build.yaml
danking: I think you commented on an old sha, this is gone

danking:
I removed jinja and restored the build command. I can't see your last comment on docker/Makefile, GitHub won't load it for me, can you repost it here?

batch/Makefile
```diff
@@ -23,7 +23,7 @@ build:
 	docker build -t batch -f Dockerfile.out --cache-from batch,$(BATCH_LATEST),service-base ..
 	-docker pull python:3.7-slim-stretch
 	-docker pull $(BATCH_WORKER_LATEST)
-	docker build -t batch-worker -f Dockerfile.worker --cache-from batch-worker,$(BATCH_WORKER_LATEST),python:3.7-slim-stretch ..
+	docker build -t batch-worker -f Dockerfile.worker.out --cache-from batch-worker,$(BATCH_WORKER_LATEST),python:3.7-slim-stretch ..
```
jigold, line 26: I think this should be Dockerfile.worker

jigold (changes requested):

batch/Makefile
danking: good catch

auth/Dockerfile
```diff
@@ -1,12 +1,11 @@
 FROM {{ service_base_image.image }}
 
-RUN pip3 install --upgrade \
+RUN hail-pip-install \
       google-auth-oauthlib==0.4.1 \
-      google-auth==1.20.0
+      google-auth==1.21.1
```
danking, line 5: this fixes a latent version incompatibility we had.

ci/test/test_ci.py
```diff
@@ -39,5 +39,5 @@ async def wait_forever():
             log.info(f'returning {deploy_status} {failure_information}')
             return deploy_state, failure_information
 
-        deploy_state, failure_information = await asyncio.wait_for(wait_forever(), timeout=30 * 60)
+        deploy_state, failure_information = await wait_forever()
```
danking, line 42: the test job itself already has a timeout, so this is just a confusing redundancy

docker/requirements.txt
```diff
@@ -34,7 +34,7 @@ prometheus_async==19.2.0
 prometheus_client==0.7.1
 pyjwt==1.7.1
 pylint==2.6.0
-PyMySQL==0.9.3
+PyMySQL==0.9.2
```
danking, line 37: another latent version incompatibility

docker/requirements.txt
```diff
@@ -55,4 +55,3 @@ wheel>=0.31.0
 # yarl 1.6.0 broke query string parsing in aiohttp 3.6.[012] https://github.com/aio-libs/aiohttp/issues/4972#issuecomment-700290809
 yarl<1.6.0
 zulip==0.6.3
-six>1.14.0
```
danking, line 58: this is no longer needed because the new pip install feature will always upgrade to compatible versions if necessary and fail if incompatible versions exist.

danking:
@jigold ready for a second look

jigold (approved):

PR #9594, opened 2020-10-09
[docker] Teach Hail Ubuntu Image how to resiliently install apt packages

danking:
This PR is stacked on #9593. The key files to look at are in `docker/hail-ubuntu`. I introduced `hail-apt-get-install` which packages up the `apt-get update` and the removal of temporary files. I also set the number of package-download retries to 5.

### DIFF
--- a/letsencrypt/Dockerfile
+++ b/letsencrypt/Dockerfile
@@ -1,15 +1,10 @@
 FROM {{ hail_ubuntu_image.image }}
 
-RUN apt-get update -y
-
 # get add-apt-repository
-RUN apt-get update -y && \
-    apt-get install -y nginx software-properties-common && \
+RUN hail-apt-get-install nginx software-properties-common && \
     bash -c 'DEBIAN_FRONTEND=noninteractive apt-get install -y tzdata' && \
     add-apt-repository -y ppa:certbot/certbot && \
-    apt-get update -y && \
-    apt-get install -y python-certbot-nginx wget && \
-    rm -rf /var/lib/apt/lists/*
+    hail-apt-get-install python-certbot-nginx wget
 
 RUN wget -O /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.11.3/bin/linux/amd64/kubectl && \
     chmod +x /usr/local/bin/kubectl
--- a/site/Dockerfile
+++ b/site/Dockerfile
@@ -1,8 +1,10 @@
-FROM gcr.io/{{ global.project }}/debian:9.5
+FROM {{ hail_ubuntu_image.image }}
 
-RUN apt-get update && \
-    apt-get install -y nginx curl python-minimal && \
-    rm -rf /var/lib/apt/lists/*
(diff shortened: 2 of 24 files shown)


jigold:
I took a quick look and it looks fine, but I want to wait until the first goes in as it's a lot of lines.

jigold:
This needs a rebase. Ping me when it's ready.

danking:
@jigold ready; I also fixed the recently added `echo` image.

site/Dockerfile
```diff
@@ -1,8 +1,10 @@
...
+FROM {{ hail_ubuntu_image.image }}
 
-RUN apt-get update && \
-    apt-get install -y nginx curl python-minimal && \
-    rm -rf /var/lib/apt/lists/*
+RUN hail-apt-get-install \
+  nginx \
+  curl \
+  python-minimal \
+  libnginx-mod-http-subs-filter
```
jigold, line 7: Where did this one come from?

letsencrypt/Dockerfile
```diff
@@ -1,15 +1,10 @@
 FROM {{ hail_ubuntu_image.image }}
 
-RUN apt-get update -y
-
 # get add-apt-repository
-RUN apt-get update -y && \
-    apt-get install -y nginx software-properties-common && \
+RUN hail-apt-get-install nginx software-properties-common && \
     bash -c 'DEBIAN_FRONTEND=noninteractive apt-get install -y tzdata' && \
```
jigold, line 5: Should we have not converted this image to be based on Ubuntu? Or was it always on ubuntu/base_image. Looking at the DEBIAN_FRONTEND variable

jigold (changes requested):

site/Dockerfile
danking: Ubuntu nginx does not include this nginx module which we use. Debian nginx included it by default.

letsencrypt/Dockerfile
danking: Ubuntu is built on top of Debian, so it leaks into ubuntu things. Here, we're telling apt-get, a Debian package manager that Ubuntu adopted, to be non-interactive.

This was originally using the ubuntu image.

jigold (approved):

jigold:
Approved. Is there any reason to not merge. The Stacked PR label is still on.

danking:
whoops, nope

tpoterba:
commit went in as "wip" :(

tpoterba:
we can add a check in CI to look at commit messages and reject if the only difference between the branch and main is a single commit that's fewer than X chars or somethign

danking:
That's a great suggestion

danking:
Actually, I think the right fix is for CI to grab the PR message and set it as the merge message

tpoterba:
is that something the GitHub API supports? If so, that definitely sounds like the right thing!

danking:
yeah, "merge_message" https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#merge-a-branch

PR #9595, opened 2020-10-10
[query] Add PCodeParam and PCodeParamType

chrisvittal:
Remove implicit conversion from ptype to EmitParamType, adding asParam
and asEmitParam to PType for explicit conversion.

Currently PCodeParams are implemented pretty poorly as we should really
be creating the proper PValue from the parameters. This will require
either expanding nCodes for PCode or automatically memoizing the value
when getPCodeParam is called.

This is left for a future PR.


tpoterba (approved):
nice!

PR #9596, opened 2020-10-13
[hail] minor fixes

cseed:
- Expose init_local.
 - Fix formatting of some error messages (stray }).
 - Fix index paths, they don't have a "parts" component, have ".idx" suffix.  This showed up as an issue interopreating between Spark and local modes.

FYI @tpoterba rather than just testing them independently, it might be worthwhile to have write/read interop tests between the various backends.  Spark to local is partially tested by the pre-existing (matrix)tables tests, but not the other way.

### DIFF
--- a/hail/python/hail/__init__.py
+++ b/hail/python/hail/__init__.py
@@ -48,7 +48,7 @@ from hail.utils import (Struct, Interval, hadoop_copy, hadoop_open, hadoop_ls,  
                         hadoop_stat, hadoop_exists, hadoop_is_file,
                         hadoop_is_dir, copy_log)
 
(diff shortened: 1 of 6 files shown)


tpoterba:
> Fix index paths, they don't have a "parts" component, have ".idx" suffix. This showed up as an issue interopreating between Spark and local modes.

This is why I was so confused implementing the indexed reads.

patrick-schultz (approved):

cseed:
It is my intention to eventually document it when when local mode is feature complete and reliable.  I'm not sure what the standard process is for this.  I'd be OK to rename it when that happens.

PR #9597, opened 2020-10-13
[hail][site] XSLT is dead. Long live Jinja2.

danking:
High level take-aways:

- Hail docs now have syntax highlighting (we just needed to import pygments.css).
- Search works again.
- There are now only two root HTML templates: `site/templates/base.html` and
  `web_common/web_common/templates/layout.html`. I cannot unify these further
  because our services and our main websites actually differ significantly.
- The search/nav bar is now present in the HTML, no JS nonsense to
  asynchronously load it into place after HTML rendering.
- Site now has a `make watch` which watches for changes and automatically
  re-renders the HTML.
- Site now has a few make rules that facilitate experimenting with how the docs
  are displayed within the context of the current development version of site's
  CSS & HTML.
- XSLT is now only used by the C++ tests.

Smaller things:

- Removed bootstrap dependencies. Did we ever actually use these?
- Removed "clipboard.js" dependency. Also not clear from where this came.
- Removed use of the `subtitle` tag, which isn't actually an HTML tag?

Future work:

- Simplify our CSS. It's not possible to logically reason about our CSS. And it
  interacts in bad ways with the latent RTD themes. I want a unified Hail visual
  theme.
- Clean up the search-related JavaScript in nav-bottom.html and
  search.html. These both seem too complicated to just make search work.

---

The thrust of this PR is to restructure Hail's website and documentation to
entirely rely on Jinja2 templates. Previously, we used a mix of Jinja2, XSLT,
and in-browser JavaScript DOM manipulation to piece together a web page.

Now, all of Hail's non-service website derives from
`site/templates/base.html`. It is a Jinja2 template with four blocks: title,
meta_description, head, and content. It ensures that:
- Hail's CSS is loaded,
- the Hail icon is set,
- the fonts are loaded,
- the source code highlighter is loaded (prism.js, only used outside the docs)
- the nav bar is present and configured.
The nav bar is somewhat complicated. There are two pieces. `nav-top.html`
contains the nav bar HTML elements. `nav-bottom.html` contains the JavaScript
code that hooks the search bar up to Algolia and sets the active page in the
navigation. I believe JavaScript which modifies the HTML DOM is traditionally
placed at the bottom of the `body` tag so that it is executed *after* the HTML
DOM is mostly rendered. That's why the navigation/search bar is split across two
files. I also load the `prism.js` source code highlighter at the end of the
body.

All of the non-docs pages are defined by html files in `pages`. Each one of
these is a Jinja2 template which derives from the base template. `make render`
converts every template in `pages` into a real HTML file in `www`. Check out
404.html for a simple example.

Once I had the site in working order, I turned my eyes to the docs. I converted
`docs/_templates/layout.html`, the base template for our docs, into a template
which derives from `site/templates/base.html`. That ensures everyone is using
the same CSS, the same navigation/search bar, same icon set, etc. Convincing
Sphinx to work like this was actually really easy because Sphinx already uses
Jinja2 templates! I just added site's templates folder to the Sphinx
`templates_path`.

I eliminated a few conditionals that are only relevant if your docs are also
rendered on RTD's server, which ours are not.

Finally, in order to experiment quickly with this, I changed site's Makefile to
add some rules that keep the automatically render the site and docs in response
to file edits.


danking:
So, you can look at https://internal.hail.is/dking/site/index.html, but the docs are pulled from the 0.2 production docs, not from the docs from that page. We should fix this for PRs, but you'll have to trust me or build the docs locally. The doc search now works.

johnc1231:
Failing to build, template not found `base.html`

site/pages/references.html
```diff
@@ -0,0 +1,99 @@
...
+  </div>
+  <div class="about">
+    <div class="content">
+      <p>If you use Hail for published work, please cite the software. You can get a citation for the version of Hail you installed by executing:</p>
+      <pre><code class="language-python">import hail as hl
+print(hl.citation())</code></pre>
+      <p>Or you could include the following line in your bibliography:</p>
+      <pre><code>Hail Team. Hail 0.2. https://github.com/hail-is/hail</code></pre>
+      <p>Otherwise, we welcome you to add additional examples by <a href="https://github.com/hail-is/hail/edit/main/site/pages/references.html">editing this page directly</a>, after which we will review the pull request to confirm the addition is valid. Please adhere to the existing formatting conventions.</p>
+      <p><em>Last updated on March 30th, 2020</em></p>
```
johnc1231, line 25: I know you didn't write this, but this can't be right, since some of these papers are after March 30th 2020

site/pages/references.html
```diff
@@ -0,0 +1,99 @@
+{% extends "base.html" %}
+{% block title %} 404 Not Found {% endblock %}
+{% block meta_description %} 404 Not Found {% endblock %}
+{% block content %}
```
johnc1231, line 4: Wrong title for this page.

site/pages/cheatsheets.html
```diff
@@ -0,0 +1,28 @@
+{% extends "base.html" %}
+{% block title %} Cheatsheets {% endblock %}
+{% block meta_description %} Cheatsheets {% endblock %}
```
johnc1231, line 3: I've since learned that Cheatsheet is not one word, Should probably be Cheat Sheets. Idk if you care about this, I don't think it matters much but while I was reviewing figured I'd mention it.

site/pages/tutorial.html
```diff
@@ -0,0 +1,163 @@
+{% extends "base.html" %}
+{% block title %} 404 Not Found {% endblock %}
+{% block meta_description %} 404 Not Found {% endblock %}
+{% block content %}
```
johnc1231, line 4: Wrong title for this page.

johnc1231 (changes requested):
Two wrong titles, two nits you can ignore for now if you'd rather, and `make docs` is failing on CI.

danking:
Sending to WIP while I iron out an CORS issue

danking:
OK, I addressed your comments and also fixed the ball and line animation (which I originally broke). We had a file called "vanta.js" which was not the VantaJS library. I've restored that file in a `script` tag in `index.html`. I believe everything now works. The docs are still using the latest *main* docs rather than the docs from this branch. We'll have to fix that another day.

site/pages/search.html
```diff
@@ -0,0 +1,337 @@
...
+            if (routeState.page !== 1) {
+              queryParameters.page = routeState.page;
+            }
+
+            const queryString = qsModule.stringify(queryParameters, {
+              addQueryPrefix: true,
+              arrayFormat: 'repeat'
+            });
+
+            return `/search.html${queryString}`;
```
chrisvittal, line 199: Noticed this while I was looking at the dev deployment. If we want to unbreak search in dev deployment, this should be a releative URL path, the leading slash is causing the path to be replaced, not joined to.

johnc1231 (dismissed):
Still failing `make docs`

site/pages/search.html
danking: I'll fix but I think search is deeply broken in dev anyway because all the links point back to hail.is.

johnc1231 (changes requested):
Still failing, can't find a path.

johnc1231 (approved):

PR #9602, opened 2020-10-14
[query] Support Spark shuffles in lowered IRs with SparkBackend

tpoterba:


tpoterba:
Stacked on #9591

tpoterba:
Dice came up @johnc1231. I'll write a paragraph here describing this change before you review.

tpoterba:
OK, here's the high level design:

This diff touches a lot of files, but it is conceptually a small
list of connected changes:

1. TableStageToRVD. This is infrastructure that lets us take a
   TableStage and generate a RVD and BroadcastRow (globals). This
   is used in...
2. SparkBackend shuffle lowering. This now uses TableStageToRVD to
   create an RVD, uses RVD methods to shuffle, then RVDToTableStage
   to produce a TableStage that reads the shuffled RDD.
3. TableStageDependencies. As I mentioned in our design meeting this
   past week, the RDD we generate in the SparkBackend's parallelizeAndComputeWithIndex
   needs to depend on the RDDs consumed by RVDToTableStage. In order to
   support this, I have added a notion of a `TableStageDependency` to TableStage,
   CollectDistributedArray, and BackendUtils.collectDArray.

hail/src/main/scala/is/hail/expr/ir/lowering/RVDToTableStage.scala
```diff
@@ -1,15 +1,92 @@
...
+    val encodedBcVals = sparkContext.broadcast(bcValsSpec.encodeValue(ctx, bcValsPType, gbPType.loadField(gbAddr, 1)))
+    val (decodedBcValsPType: PStruct, makeBcDec) = bcValsSpec.buildDecoder(ctx, bcValsPType.virtualType)
+
+    val contextsPType = gbPType.fieldType("contexts").asInstanceOf[PArray]
+    val contextPType = contextsPType.elementType
+    val contextSpec = TypedCodecSpec(contextPType, BufferSpec.wireSpec)
+    val contextsAddr = gbPType.loadField(gbAddr, 2)
+    val nContexts = contextsPType.loadLength(contextsAddr)
+
+    val (decodedContextPType: PStruct, makeContextDec) = contextSpec.buildDecoder(ctx, contextPType.virtualType)
```
johnc1231, line 59: Somehow `buildDecoder` is actually returning a `PArray`, and a test is failing. Maybe assert `contextPType` is a `PStruct` and figure out what is going wrong? I thought it was just a `contextPType` vs `contextsPType` mixup (the names are maybe too similar for easy reading) but after reading I don't see any instances of that.

johnc1231 (changes requested):

hail/src/main/scala/is/hail/expr/ir/lowering/RVDToTableStage.scala
tpoterba: oh, thanks for the guidance!

tpoterba: ah, I see what's going on -- contexts don't actually need to be structs! So I'll just have to wrap in a struct so the interpreted code (called from staged code) works out.

johnc1231: Makes sense 👍

tpoterba:
ok, passing!

johnc1231 (approved):

PR #9603, opened 2020-10-15
[query] Fix memory management for partition readers

tpoterba:
Use eltRegion instead of partitionRegion.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/TableIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/TableIR.scala
@@ -431,9 +431,10 @@ object TableNativeReader {
     val filterIntervals = params.options.map(_.filterIntervals).getOrElse(false)
 
     if (filterIntervals && !spec.indexed)
-      fatal("""`intervals` specified on an unindexed table.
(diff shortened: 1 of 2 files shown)


hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -531,19 +534,18 @@ case class PartitionNativeReader(spec: AbstractTypedCodecSpec) extends AbstractN
       val decRes = mb.newEmitLocal(PInt64Optional)
       val hasNext = mb.newLocal[Boolean]("pnr_hasNext")
       val next = mb.newLocal[Long]("pnr_next")
-      val stream = Stream.unfold[Code[Long]](
+
+      SizedStream.unsized(eltRegion => Stream.unfold[Code[Long]](
```
tpoterba, line 538: These diffs look weird, but what I did was move the `val stream` definition inside the `unsized` call and use `eltRegion` instead of `region`.

patrick-schultz (approved):
Simple change. Sorry for the delay.

PR #9604, opened 2020-10-15
[query] 'EmitCode tuple' to Actual EmitCode in BTree using aggregators

chrisvittal:


catoverdrive (approved):
Sweet! This is starting to look so much cleaner.

chrisvittal:
Unfortunately, I can't figure out the verification errors.

tpoterba:
I can take a look tomorrow - can you point me to the test that's failing?

catoverdrive:
@chrisvittal  I can't figure out how to comment on unedited code, but here:

https://github.com/hail-is/hail/pull/9604/files#diff-689808a42e9fe9329e347edede9bc12419e29a1634afc5d8c899a41c9d550659R211

you probably want to switch to passing in an EmitCode as well. Not sure if that's the source of the code verification errors, but I think the order I was originally passing in the tuples (m, v) is swapped from how emitCodeParams represent them (v, m), so that might help.

catoverdrive:
argh, that link doesn't work but I was trying to link to the `getOrElseInitialize` method on the staged btree class, which still passes in `m` and `v` separately, I think

chrisvittal:
Thanks! Tim and I were able to figure it out a little earlier.

PR #9611, opened 2020-10-20
[hail] remove legacy lowering

cseed:
now identical to the non-legacy version


tpoterba (changes requested):
You'll have to remove the IRState stuff. 

The `legacy` lowerers are used for interpreting relational IRs to TableValue and BlockMatrix. We therefore can't produce a compilable IR that `relationalLowerer` expects to produce.

cseed:
I see, thanks for clarifying.

PR #9612, opened 2020-10-20
[query] Fix `TableForceCount` lowering.

tpoterba:
The `Consume` was getting optimized away in (StreamLen(StreamMap ...))


johnc1231 (approved):
Agree this fixes the problem. Though I wonder if we should have some sort of check that `Consume` isn't removed by `Simplify` rules, since I'm pretty sure we only use it for timing / testing purposes, and I wouldn't be that surprised if we optimized it away again in the future.

PR #9614, opened 2020-10-20
[Benchmark] Submit dummy batch test

Dania-Abuhijleh:


jigold:
Can you rebase this and then I'll look at it this afternoon?

benchmark-service/benchmark/benchmark.py
```diff
@@ -214,26 +217,34 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
-async def query_github(app):
-    global START_POINT
-    github_client = app['github_client']
-    request_string = f'/repos/hail-is/hail/commits?since={START_POINT}'
+@router.post('/api/v1alpha/benchmark/create_benchmark')
+async def submit(request):  # pylint: disable=unused-argument
+    app = request.app
+    batch_client = app['batch_client']
+    body = await request.json()
+    commit = body['commit']
```
jigold, line 225: I don't think this should take the full commit. Rather it should take the sha and then we should look up the full status with the gh_client.

benchmark-service/benchmark/benchmark.py
```diff
@@ -214,26 +217,34 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
-    for commit in data:
-        sha = commit.get('sha')
-        new_commits.append(commit)
-        log.info(f'commit {sha}')
-        START_POINT = commit['commit']['author'].get('date')
-        log.info(f'start point is now {START_POINT}')
-    log.info('got new commits')
+@router.get('/api/v1alpha/benchmark/batches/{batch_id}')
+async def batch_status(request):  # pylint: disable=unused-argument
+    app = request.app
```
jigold, line 234: I'm not sure what your reasoning is for this interface. I think it should take the sha to look up and then query for batches that match that sha. Then return the status for the last batch that matches the sha (I think that's the correct ordering) or returns None if it doesn't exist.

jigold (changes requested):
I know there are a lot of comments here. Perhaps you and @danking discussed the code already and this is what you decided on. If my comments are confusing or you don't agree with them, then maybe it would be best to setup a meeting with the three of us.

jigold:
So I reread the title of your PR. This is fine if you just submit the test batch. However, I think you should come up with a plan to eventually get to the full implementation in stages. Whatever works best for you.

jigold:
Sorry, fine meaning just having the submit routes and the test. Not fine with the polling of GitHub with the start time stuff.

PR #9618, opened 2020-10-20
[query] Guard against deoptimization with TableCollect(TableOrderBy)

tpoterba:


johnc1231 (approved):

tpoterba:
ugh, test failure in local backend. Will debug in a bit.

johnc1231:
@tpoterba What's going on with this?

tpoterba:
could be worse:
image

tpoterba:
I did debug this, though. The failing test (which succeeds in Spark, but fails in local backend) collects a table and asserts that the result is equal to a list of expected rows. The failure is caused by an ordering issue - the rows are the same, but the order is slightly different between the Spark backend (which produces the expected output) and the local backend.

However, I think that actually *both* orders are valid under Hail's guarantees. I'll bring this to our next team meeting for group discussion.

PR #9622, opened 2020-10-21
[services] pervasively gracefully shutdown

danking:
`asyncio.ensure_future` creates background tasks. When a service is shutdown by
kubernetes, these background tasks keep the service up for the full 30 second
grace-period. This impedes development and deployment velocity.

This change eliminates all uses of `asyncio.ensure_future` which ignore the
returned future. Instead, all these futures are managed by a
`aiotools.BackgroundTaskManager`. This is a new class which holds a (weak)
reference to all of its created tasks and provides `shutdown` to cancel all the
tasks. Canceling the tasks ensures a speedy shutdown.

The use of a weak reference is critical to enable the garbage collector to clean
up tasks that have completed successfully. In particular, the batch worker
creates short-lived, background tasks.

cc: @catoverdrive @Dania-Abuhijleh

### DIFF
--- a/batch/batch/driver/instance_pool.py
+++ b/batch/batch/driver/instance_pool.py
@@ -191,8 +191,8 @@ FROM globals;
         self.task_manager.ensure_future(self.control_loop())
         self.task_manager.ensure_future(self.instance_monitoring_loop())
         self.task_manager.ensure_future(retry_long_running(
-                'update_zones_loop',
-                self.update_zones_loop))
+            'update_zones_loop',
+            self.update_zones_loop))
 
     def shutdown(self):
         self.task_manager.shutdown()
--- a/batch/batch/driver/instance_pool.py
+++ b/batch/batch/driver/instance_pool.py
@@ -11,6 +11,7 @@ import base64
 import dateutil.parser
 import sortedcontainers
 import aiohttp
+from hailtop import aiotools
 from hailtop.utils import (
     retry_long_running, time_msecs, secret_alnum_string)
 
@@ -112,6 +113,8 @@ class InstancePool:
 
         self.name_instance = {}
 
+        self.task_manager = aiotools.BackgroundTaskManager()
+
     async def update_zones(self):
         northamerica_regions = {
             # 'northamerica-northeast1',
@@ -184,13 +187,15 @@ FROM globals;
             instance = Instance.from_record(self.app, record)
             self.add_instance(instance)
 
-        asyncio.ensure_future(self.event_loop())
-        asyncio.ensure_future(self.control_loop())
-        asyncio.ensure_future(self.instance_monitoring_loop())
+        self.task_manager.ensure_future(self.event_loop())
+        self.task_manager.ensure_future(self.control_loop())
+        self.task_manager.ensure_future(self.instance_monitoring_loop())
+        self.task_manager.ensure_future(retry_long_running(
+                'update_zones_loop',
+                self.update_zones_loop))
 
-        asyncio.ensure_future(retry_long_running(
-            'update_zones_loop',
-            self.update_zones_loop))
+    def shutdown(self):
+        self.task_manager.shutdown()
 
     def config(self):
         return {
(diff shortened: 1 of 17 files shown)


batch/batch/driver/instance_pool.py
```diff
@@ -112,6 +113,8 @@ def __init__(self, app, machine_name_prefix):
 
         self.name_instance = {}
 
+        self.task_manager = aiotools.BackgroundTaskManager()
```
jigold, line 116: Should this be the same TaskManager as the one from the app?

batch/batch/driver/scheduler.py
```diff
@@ -43,21 +44,28 @@ def __init__(self, app):
         self.inst_pool = app['inst_pool']
         self.async_worker_pool = AsyncWorkerPool(parallelism=100, queue_size=100)
         self.exceeded_shares_counter = ExceededSharesCounter()
+        self.task_manager = aiotools.BackgroundTaskManager()
```
jigold, line 47: Same question about whether this should come from the app?

batch/batch/worker/worker.py
```diff
@@ -889,13 +898,18 @@ def __init__(self, path, pool):
         self.path = path
         self.cleanup_event = asyncio.Event()
         self.pool = pool
+        self.task_manager = aiotools.BackgroundTaskManager()
```
jigold, line 901: Same question.

ci/ci/ci.py
```diff
@@ -334,7 +335,7 @@ async def update_all():
         for wb in watched_branches:
             await wb.update(request.app)
 
-    await asyncio.ensure_future(update_all())
+    request.app['task_manager'].ensure_future(update_all())
```
jigold, line 338: was this just broken with the await before?

hail/python/hailtop/aiotools/tasks.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self):
+        self.tasks: weakref.WeakSet = weakref.WeakSet()
+
+    def ensure_future(self, coroutine):
+        self.tasks.add(asyncio.ensure_future(coroutine))
+
+    def shutdown(self):
+        for task in self.tasks:
+            try:
+                if task:
```
jigold, line 19: Should this check whether the task is done rather than exists?

hail/python/hailtop/utils/utils.py
```diff
@@ -200,6 +205,14 @@ async def call(self, f, *args, **kwargs):
     def call_nowait(self, f, *args, **kwargs):
         self._queue.put_nowait((f, args, kwargs))
 
+    def shutdown(self):
+        for worker in self.workers:
+            if worker:
```
jigold, line 210: Same question about checking whether the worker is done as above

jigold (changes requested):

ci/ci/ci.py
danking: `await` ing the `Task` returned by `ensure_future` has the same effect as `await update_all()`. So, broken in the sense that it hung while running the update.

batch/batch/driver/instance_pool.py
danking: hmm. I'm not sure how much it matters. Seems a little better to have InstancePool be responsible for cleaning up its own tasks?

Since everything gets shutdown together when the app shuts down, it's mostly a matter of what code we find most easy to understand.

batch/batch/driver/scheduler.py
danking: In this case, the scheduler needs a shutdown method anyway (it should have been shutting down the pool), so it seems best to keep all the scheduler's shutdown logic in one place.

hail/python/hailtop/aiotools/tasks.py
danking: Oh whoops. This is always True because tasks are always true. No, it shouldn't check if the task is done, shutdown is only called when the app needs to go away *now*. Anything that needs to avoid getting cancelled can use `asyncio.shield`. Those tasks will continue to keep the app running.

hail/python/hailtop/utils/utils.py
danking: whoops again, same deal.

jigold (approved):

PR #9624, opened 2020-10-21
[query] Move new field methods to EmitClassBuilder

chrisvittal:


chrisvittal:
Merged as part of #9604

PR #9625, opened 2020-10-21
[query] TableAggregateByKey Memoization Fix

johnc1231:
When you have `TableAggregateByKey(child, expr)`, you cannot assume that the keys of the result are present in `expr`, so the memoization rule has to be changed to request only the subset of fields in the requested type that are present in `expr`. 

One of the bugs found during #9578

### DIFF
--- a/hail/src/main/scala/is/hail/types/virtual/TStruct.scala
+++ b/hail/src/main/scala/is/hail/types/virtual/TStruct.scala
@@ -76,17 +76,7 @@ final case class TStruct(fields: IndexedSeq[Field]) extends TBaseStruct {
 
   def hasField(name: String): Boolean = fieldIdx.contains(name)
 
-  def field(name: String): Field = {
-    try {
-      fields(fieldIdx(name))
-    } catch {
-      case e: Throwable => {
-        println(s"FIELDS WERE ${fields}")
-        throw e
-      }
-    }
-
-  }
+  def field(name: String): Field = fields(fieldIdx(name))
 
   def fieldType(name: String): Type = types(fieldIdx(name))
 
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/types/virtual/TStruct.scala
```diff
@@ -76,7 +76,17 @@ final case class TStruct(fields: IndexedSeq[Field]) extends TBaseStruct {
 
   def hasField(name: String): Boolean = fieldIdx.contains(name)
 
-  def field(name: String): Field = fields(fieldIdx(name))
+  def field(name: String): Field = {
+    try {
```
patrick-schultz, line 80: Did you leave this intentionally?

patrick-schultz (changes requested):
Amazing we didn't hit this earlier!

hail/src/main/scala/is/hail/types/virtual/TStruct.scala
johnc1231: No, thanks, I'll remove.

johnc1231:
I'm also surprised we didn't hit earlier. In doing my linear regression debugging, I had added assertions in every call to memoize and rebuild that we are maintaining the expected supertype relationships, which is how I found most of these. I originally thought all of the problems I found were uncovered by IR changes from #9633, but I now suspect that actually we just have/had cases in `PruneDeadFields` where we break the supertype rules temporarily, but it gets cancelled out by some higher IR and doesn't end up mattering. Would be nice if when debugging / testing we could run those asserts to make sure these things don't slip through in the future, as finding all of these bugs was a big headache.

patrick-schultz (approved):

PR #9627, opened 2020-10-21
[query] TableMapPartitions Rebuild Fix

johnc1231:
`TableMapPartitions` needs to insert a TableKeyBy if any element of its child's key is not in the rebuilt body. This is the same thing TableMapRows does.

I can figure out a PruneSuite test for this if you want, but the incoming chained linear regression PR will test this behavior as well, I'm just trying to break things up.


tpoterba:
I've added a note to add a test for this later. Thanks John!

tpoterba (approved):

PR #9631, opened 2020-10-22
[query] allow unrealizable ref in extract

patrick-schultz:
This assertion was getting triggered by the ref `inner` in a perfectly correct IR of the form `StreamMap(inner, StreamGrouped(...), ...)`. There's no way to avoid the unrealizable ref when mapping over a nested stream, and I don't see why `extract` depends on this assertion.


tpoterba:
Can we add a test that triggers it? I think Nick's example from Zulip should do it.

tpoterba (changes requested):
test

tpoterba (approved):
thanks!

PR #9634, opened 2020-10-22
Python Chained Linear Regression, Attempt 2

johnc1231:
Replacing #9578 

_linear_regression_rows_nd now supports chained linear regression correctly. Also added support for pass through arguments. This should now support everything that regular linear regression supports, but all written in Python. This allows a few more tests to run on local backend.

I'd be particularly interested in whether you think there's a way to write this that would generate more efficient IR. It seems hacky to have to map over the range of rows and index into it, but I didn't see a better way.

### DIFF
--- a/hail/python/hail/methods/statgen.py
+++ b/hail/python/hail/methods/statgen.py
@@ -463,21 +463,23 @@ def _linear_regression_rows_nd(y, x, covariates, block_size=16, pass_through=())
 
     cov_nds = [make_one_cov_matrix(ys_and_covs_to_keep) for ys_and_covs_to_keep in list_of_ys_and_covs_to_keep]
 
-    y_nds = [make_one_y_matrix(ys_and_covs_to_keep, one_y_field_name_set)
-             for ys_and_covs_to_keep, one_y_field_name_set in zip(list_of_ys_and_covs_to_keep, y_field_names)]
+    y_nds = hl.array([make_one_y_matrix(ys_and_covs_to_keep, one_y_field_name_set)
+             for ys_and_covs_to_keep, one_y_field_name_set in zip(list_of_ys_and_covs_to_keep, y_field_names)])
 
     ht = ht.annotate_globals(kept_samples=list_of_indices_to_keep,
-                             __y_nds=y_nds,
                              __cov_nds=cov_nds)
     k = builtins.len(covariates)
     ns = ht.index_globals().kept_samples.map(lambda one_sample_set: hl.len(one_sample_set))
-    ht = ht.annotate_globals(ds=ns.map(lambda n: n - k - 1))
     cov_Qts = hl.if_else(k > 0,
                          ht.__cov_nds.map(lambda one_cov_nd: hl.nd.qr(one_cov_nd)[0].T),
                          ns.map(lambda n: hl.nd.zeros((0, n))))
-    ht = ht.annotate_globals(__cov_Qts=cov_Qts)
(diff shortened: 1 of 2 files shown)


johnc1231:
@patrick-schultz, you may have tuned out the long thread about all the issues with the first Python Chained Linear Regression and pruning, but this one is now clear of all of that and is purely a Python implementation of  Scala's `LinearRegressionChained`: https://github.com/hail-is/hail/blob/main/hail/src/main/scala/is/hail/methods/LinearRegression.scala#L175

hail/python/hail/methods/statgen.py
```diff
@@ -440,49 +441,86 @@ def select_array_indices(hl_array, indices):
...
-    ht = ht.annotate_globals(__Qty=cov_Qt @ ht.__y_nd)
-    ht = ht.annotate_globals(__yyp=dot_rows_with_themselves(ht.__y_nd.T) - dot_rows_with_themselves(ht.__Qty.T))
+    ns = ht.index_globals().kept_samples.map(lambda one_sample_set: hl.len(one_sample_set))
+    ht = ht.annotate_globals(ds=ns.map(lambda n: n - k - 1))
+    cov_Qts = hl.if_else(k > 0,
+                         ht.__cov_nds.map(lambda one_cov_nd: hl.nd.qr(one_cov_nd)[0].T),
+                         ns.map(lambda n: hl.nd.zeros((0, n))))
+    ht = ht.annotate_globals(__cov_Qts=cov_Qts)
+    ht = ht.annotate_globals(__Qtys=hl.range(num_y_lists).map(lambda i: ht.__cov_Qts[i] @ ht.__y_nds[i]))
+    ht = ht.annotate_globals(__yyps=hl.range(num_y_lists).map(lambda i: dot_rows_with_themselves(ht.__y_nds[i].T) - dot_rows_with_themselves(ht.__Qtys[i].T)))
```
patrick-schultz, line 480: I think this chain of `annotate_globals` can be cleaned up, making use of the python CSE. Something like
```
cov_Qts = hl.if_else(k > 0,
                     cov_nds.map(lambda one_cov_nd: hl.nd.qr(one_cov_nd)[0].T),
                     ns.map(lambda n: hl.nd.zeros((0, n))))
Qtys = hl.range(num_y_lists).map(lambda i: cov_Qts[i] @ y_nds[i])
ht = ht.annotate_globals(
  kept_samples=list_of_indices_to_keep,
  __y_nds=y_nds,
  __cov_nds=cov_nds,
  ds=ns.map(lambda n: n - k - 1),
  __cov_Qts=cov_Qts,
  __Qtys=Qtys,
  __yyps=hl.range(num_y_lists).map(lambda i: dot_rows_with_themselves(y_nds[i].T) - dot_rows_with_themselves(Qtys[i].T))
```
(possibly dropping some of those from the `annotate_globals` if they aren't actually needed again.

patrick-schultz (changes requested):
Sorry about that John. Looks good. I only see one thing that can be cleaned up a bit.

> It seems hacky to have to map over the range of rows and index into it, but I didn't see a better way.

Yeah, I don't see any better way either.

hail/python/hail/methods/statgen.py
johnc1231: I did this and was able to delete `__cov_nds`, so that was nice. `kept_samples` currently needs to stay where it is due to what I believe is a bug where we are losing the `source` of an expression's `_indices`. I have yet to discover why.

patrick-schultz (approved):

johnc1231:
This PR seems to have made the non-chained case substantially slower somehow (the benchmark is running 3x slower). Either need to figure out what's wrong with the code we are generating and/or split out different chained and unchained methods.

PR #9636, opened 2020-10-23
[query] use websockets for query service

catoverdrive:
Implements existing query service endpoints using websockets instead of long-running http requests. I open a new socket for each request instead of attempting to hold one open for each client to send all its requests; not sure if one is preferable to the other, but this one seemed easier to handle and more in line with what we were doing before.

~~I haven't put any sort of heartbeat on either end for now for simplicity; the `blocking_to_async` wrapper around the jvm execution interferes with the server's ability to send/receive pings and pongs, and I'm not currently handling retries for timeouts anyways; would love suggestions on how to make this more robust.~~

Since nginx's default behavior is to close the websocket after 60s of non-activity (which seems pretty reasonable), I have a 30s heartbeat on the server-side websocket connection. This meant rewriting the flow to split a task off to execute the blocking JVM function and keeping the websocket task open to handle the heartbeat. Currently, we rely on the client to close the connection once the jvm task is completed and the result response is received; if the connection is closed/something errors for some other reason, we check to see if the task is completed and cancel it if it's not. The client side still doesn't poll the server for existence, but if the socket is unexpectedly closed we'll retry the request.


catoverdrive:
For some reason this is timing out after a minute in my dev deploy even though I've removed the heartbeat and it's working on a local server I have running. Need to investigate further.

hail/python/hail/backend/service_backend.py
```diff
@@ -8,14 +10,43 @@
...
 from .backend import Backend
 from ..hail_logging import PythonOnlyLogger
 
 
+class ServiceSocket:
+    def __init__(self, *, deploy_config=None):
+        if not deploy_config:
+            deploy_config = get_deploy_config()
+        self.url = deploy_config.base_url('query')
+        self.session = aiohttp.ClientSession(headers=service_auth_headers(deploy_config, 'query'))
```
danking, line 25: add a close method to ServiceSocket which closes this session. Call close in backend's stop.

hail/python/hail/backend/service_backend.py
```diff
@@ -8,14 +10,43 @@
...
+            raise FatalError(f'Error raised while waiting for response from server: {resp}.')
+        assert resp.type == aiohttp.WSMsgType.TEXT
+        return resp.data
+
+    async def async_request(self, endpoint, **data):
+        async with self.session.ws_connect(f'{self.url}/api/v1alpha/{endpoint}') as socket:
+            await socket.send_str(json.dumps(data))
+            resp = self.handle_response(await socket.receive())
+            if resp != endpoint:
+                raise FatalError(f'Error from server: {json.loads(resp)["error"]}')
```
danking, line 40: What function does this step of the protocol serve? IIUC, the protocol is:
```
CLIENT: ws://.../endpoint data
SERVER: endpoint
SERVER: data2
```

hail/python/hail/backend/service_backend.py
```diff
@@ -8,14 +10,43 @@
...
-from hailtop.utils import (retry_response_returning_functions,
-                           external_requests_client_session)
+from hailtop.utils import async_to_blocking
 from hail.ir.renderer import CSERenderer
 
 from .backend import Backend
 from ..hail_logging import PythonOnlyLogger
 
 
+class ServiceSocket:
```
danking, line 20: As written, this assumes you can hold a web socket open for a long time. In general, Hail services will get restarted. I think `async_request` needs to be re-written to reconnect and retry the whole request if it encounters a `CLOSED`.

Longer term, we probably want operation UUIDs so that if the socket hangs up but the service is still alive, we can reconnect and wait on the same operation rather than starting a second one.

query/query/query.py
```diff
@@ -209,6 +202,7 @@ async def on_startup(app):
     app['jhc'] = jhc
 
     app['users'] = set()
+    app['sockets'] = set()
```
danking, line 205: what is this for?

danking (dismissed):

hail/python/hail/backend/service_backend.py
catoverdrive: whoops, yes. I'll probably just change this to `assert resp == endpoint`, I think I got it mixed up with the next response.

hail/python/hail/backend/service_backend.py
catoverdrive: Sounds good. I'll fix it to retry on `CLOSED` as soon as I figure out why the socket is timing out after 60 seconds even without the heartbeat.

query/query/query.py
catoverdrive: This tracks all of the currently-open websockets, with the idea that we'd close them down nicely if the server shuts down (see https://docs.aiohttp.org/en/stable/web_advanced.html#graceful-shutdown). I'd forgotten to implement the on_shutdown hook, though.

catoverdrive:
@danking I ended up rewriting this a bit to make it work with the nginx timeout (instead of getting rid of the timeout, since having a heartbeat seems like a pretty reasonable thing); updated the PR description to match.

hail/python/hail/backend/service_backend.py
danking: I still think we're sending an unnecessary message. Why does query respond with the endpoint? The client knows the endpoint (it requested it).

query/query/query.py
```diff
@@ -52,107 +66,97 @@ async def healthcheck(request):  # pylint: disable=unused-argument
...
+        return ws
+    finally:
+        if not ws.closed:
+            await ws.close()
+            log.info('Websocket was not closed. Closing.')
+        if task is not None and not task.done():
+            task.cancel()
+            log.info('Task has been cancelled due to websocket closure.')
+        log.info('websocket connection closed')
+        app['sockets'].remove(ws)
```
danking, line 123: I don't think we need this. handle_ws_response is only invoked by route handlers. Route handlers are cancelled when the server thread receives an interrupt. We can't leak web sockets because of the try-finally.

query/query/query.py
```diff
@@ -52,107 +66,97 @@ async def healthcheck(request):  # pylint: disable=unused-argument
...
+
+        log.info(f"{cmd_str}: {body}")
+        await ws.send_str(cmd_str)
+        task = asyncio.ensure_future(send_ws_response(app['thread_pool'], ws, f, jbackend, userdata, body))
+        r = await ws.receive()
+        log.info('Received websocket message. Expected CLOSE, got {r}')
+        assert r.type == aiohttp.WSMsgType.CLOSE
+        return ws
+    finally:
+        if not ws.closed:
```
danking, line 116: You're allowed to double close a web socket.

query/query/query.py
```diff
@@ -52,107 +66,97 @@ async def healthcheck(request):  # pylint: disable=unused-argument
...
+            await ws.close()
+            log.info('Websocket was not closed. Closing.')
+        if task is not None and not task.done():
```
danking, line 119: you can cancel a done or cancelled task and that's OK

query/query/query.py
```diff
@@ -52,107 +66,97 @@ async def healthcheck(request):  # pylint: disable=unused-argument
...
+        log.info(f'websocket prepared: {ws}')
+        body = await ws.receive_json()
+
+        log.info(f"{cmd_str}: {body}")
+        await ws.send_str(cmd_str)
+        task = asyncio.ensure_future(send_ws_response(app['thread_pool'], ws, f, jbackend, userdata, body))
+        r = await ws.receive()
+        log.info('Received websocket message. Expected CLOSE, got {r}')
+        assert r.type == aiohttp.WSMsgType.CLOSE
+        return ws
```
danking, line 114: I'm not convinced you need all this. If the client breaks the HTTP/TCP connection, `handle_ws_response` will be cancelled by aiohttp because aiohttp cancels all request handlers in response to broken connections.

This seems sufficient to me:
```python3
async def handle_ws_response(request, userdata, cmd_str, f):
    log.info('connecting websocket')
    app = request.app
    thread_pool = app['thread_pool']
    jbackend = app['jbackend']
    await add_user(app, userdata)
    ws = web.WebSocketResponse(heartbeat=5):
    await ws.prepare(request)  # prepare should come before close
    try:
        body = await ws.receive_json()
        log.info(f"{cmd_str}: {body}")
        jresp = await blocking_to_async(app['thread_pool'], f,
                                        jbackend=jbackend,
                                        userdata=userdata,
                                        body))
        status = jresp.status()
        value = jresp.value()
        log.info(f'response status {status} value {value}')
        if status in (400, 500):
            await ws.send_json({'error': value})
        else:
            assert status == 200, status
            await ws.send_json({'status': status, 'result': value})
        return ws
    finally:
        await ws.close()
```

If a client closes the web socket without also closing the connection, we'll keep working. That seems fine. If the client wanted to DDoS us, they could just leave the socket open with little added cost.

query/query/query.py
```diff
@@ -29,6 +31,18 @@ def java_to_web_response(jresp):
...
+async def send_ws_response(thread_pool, ws, f, *args, **kwargs):
+    jresp = await blocking_to_async(thread_pool, f, *args, **kwargs)
+    status = jresp.status()
+    value = jresp.value()
+    log.info(f'response status {status} value {value}')
+    if status in (400, 500):
+        await ws.send_json({'error': value})
+    else:
+        assert status == 200, status
+        await ws.send_json({'status': status, 'result': value})
```
danking, line 43: I would prefer a consistent wire format. Let's go with `{'status': int, 'result': optional[any]}` and result may be elided instead of setting it to `null`.

danking: In particular, I don't like having `'error'` be a distinct field name.

hail/python/hail/backend/service_backend.py
```diff
@@ -8,14 +10,46 @@
...
+        assert resp.type == aiohttp.WSMsgType.TEXT
+        return resp.data
+
+    async def async_request(self, endpoint, **data):
+        async with self.session.ws_connect(f'{self.url}/api/v1alpha/{endpoint}') as socket:
+            await socket.send_str(json.dumps(data))
+            resp = self.handle_response(await socket.receive())
+            assert resp == endpoint
+            result = json.loads(self.handle_response(await socket.receive()))
+            if result['status'] != 200:
```
danking, line 45: see my note below, but I think this will error when `'status'` is missing from the struct (in the error case).

danking (changes requested):

query/query/query.py
```diff
@@ -52,107 +66,97 @@ async def healthcheck(request):  # pylint: disable=unused-argument
...
-    log.info(f'table type: {code}')
+
     await add_user(app, userdata)
-    jresp = await blocking_to_async(thread_pool, blocking_table_type, jbackend, userdata['username'], code)
-    return java_to_web_response(jresp)
+    log.info('connecting websocket')
+    ws = web.WebSocketResponse(heartbeat=30, max_msg_size=0)
+    task = None
+    try:
+        await ws.prepare(request)
```
danking, line 98: prepare needs to happen outside the `try`, if anything goes wrong before `_post_start` is called in `prepare`, the `_writer` will be `None` and `close` will raise an exception.

query/query/query.py
```diff
@@ -52,107 +62,92 @@ async def healthcheck(request):  # pylint: disable=unused-argument
...
+
     await add_user(app, userdata)
-    jresp = await blocking_to_async(thread_pool, blocking_table_type, jbackend, userdata['username'], code)
-    return java_to_web_response(jresp)
+    log.info(f'{endpoint}: connecting websocket')
+    ws = web.WebSocketResponse(heartbeat=30, max_msg_size=0)
+    task = None
+    try:
+        await ws.prepare(request)
+        app['sockets'].add(ws)
```
danking, line 99: I still assert this is unnecessary. This web socket is always owned by a coroutine owned by aiohttp. Aiohttp will cancel the handler if the server is shutting down or the request is terminated. When this coroutine receives a TimeoutError, it will hit the finally block and kill the web socket. The conversation in "Graceful Shutdown" applies to web sockets owned by tasks that aren't connected to aiohttp by an ownership chain.

danking (dismissed):

query/query/query.py
catoverdrive: I did take the rest of it out, but somehow missed this line…

danking (approved):

PR #9637, opened 2020-10-23
[query] Let lowered TableHead short-circuit

patrick-schultz:
Optimize lowered TableHead to not scan a partition past the requested number of rows.

Currently TableHead performs a loop, calculating the lengths of the first 4 partitions, then the first 16, etc. If we fix it to not count a partition multiple times, instead starting each loop at the first uncounted partition, we can further optimize to stop scanning a partition after the number of rows still needed (requested number minus sum of partitions counted so far).

Note that there is no similar optimization to TableTail. We must compute the full length of each partition, to compute how many rows to drop.


tpoterba (approved):

johnc1231:
Oof, method verification error on one of the lowering tests

patrick-schultz:
Oh wow, thanks for the heads up, I haven't been watching github.

PR #9638, opened 2020-10-23
[Benchmark] update commits

Dania-Abuhijleh:


benchmark-service/benchmark/benchmark.py
```diff
@@ -215,27 +217,59 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
-        log.info(f'commit {sha}')
-        START_POINT = commit['commit']['author'].get('date')
-        log.info(f'start point is now {START_POINT}')
+    formatted_new_commits = []
+    for gh_commit in gh_data:
+
+        sha = gh_commit.get('sha')
+
+        batches = [b async for b in batch_client.list_batches(q=f'sha={sha} running')]
+        try:
```
jigold, line 235: This is a better way to write this:

```
batches = [b async for b in batch_client.list_batches(q=f'sha={sha} running')]
if len(batches) == 0:
    batch_status = None
else:
    batch = batches[-1]
    batch_status = await batch.status()
```
```

File: benchmark-service/benchmark/benchmark.py [hunk 2]
```diff
@@ -215,27 +217,59 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
+            batch = batches[-1]
+            batch_status = await batch.status()
+        except Exception:  # pylint: disable=broad-except
+            batch_status = None
+
+        file_path = f'gs://{BENCHMARK_TEST_BUCKET_NAME}/benchmark-test/{sha}'
```
jigold, line 241: Let's move the check for the file path to be first.

benchmark-service/benchmark/benchmark.py
```diff
@@ -215,27 +217,59 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
-        log.info(f'start point is now {START_POINT}')
+    formatted_new_commits = []
+    for gh_commit in gh_data:
+
+        sha = gh_commit.get('sha')
+
+        batches = [b async for b in batch_client.list_batches(q=f'sha={sha} running')]
+        try:
+            batch = batches[-1]
+            batch_status = await batch.status()
```
jigold, line 237: Actually, I don't think you use the batch_status here at all.

benchmark-service/benchmark/benchmark.py
```diff
@@ -215,27 +217,59 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
+        file_path = f'gs://{BENCHMARK_TEST_BUCKET_NAME}/benchmark-test/{sha}'
+        has_results_file = gs_reader.file_exists(file_path)
+
+        if not batches and not has_results_file:
+            new_commits.append(gh_commit)
+
     log.info('got new commits')
+    for gh_commit in new_commits:
+        sha = gh_commit.get('sha')
+        batch_id = await submit_batch(sha, batch_client)
```
jigold, line 250: Can you flip the order of batch_client and sha here? Our services code usually has it this way.

benchmark-service/benchmark/benchmark.py
```diff
@@ -215,27 +217,59 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
+        batch = batch_client.get_batch(batch_id)
+        batch_status = await batch.last_known_status()
```
jigold, line 252: What is the batch_status used for?

benchmark-service/benchmark/benchmark.py
```diff
@@ -215,27 +217,59 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
+            batch_status = None
+
+        file_path = f'gs://{BENCHMARK_TEST_BUCKET_NAME}/benchmark-test/{sha}'
+        has_results_file = gs_reader.file_exists(file_path)
+
+        if not batches and not has_results_file:
+            new_commits.append(gh_commit)
+
     log.info('got new commits')
+    for gh_commit in new_commits:
```
jigold, line 248: I don't think you need two separate for loops here. Is there a way to combine the two?

jigold (changes requested):
This is looking a lot better!

benchmark-service/benchmark/utils.py
```diff
@@ -68,3 +82,10 @@ def list_files(self, bucket_name):
         for blob in bucket.list_blobs():
             list_of_files.append('gs://' + bucket_name + '/' + blob.name)
         return list_of_files
+
+    def file_exists(self, file_path):
+        file_info = parse_file_path(FILE_PATH_REGEX, file_path)
+        bucket_name = file_info['bucket']
+        bucket = self.storage_client.bucket(bucket_name)
+        stats = storage.Blob(bucket=bucket, name=file_path).exists(self.storage_client)
```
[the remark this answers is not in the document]
Dania-Abuhijleh: I just checked the documentation and it says its optional so I'll remove it, that way it uses the client stored on the blob's bucket

benchmark-service/benchmark/benchmark.py
Dania-Abuhijleh: I forgot why I was trying to get the status of the last batch

benchmark-service/benchmark/benchmark.py
Dania-Abuhijleh: yeah you're right

benchmark-service/benchmark/benchmark.py
```diff
@@ -245,6 +271,7 @@ async def on_startup(app):
     app['github_client'] = gidgethub.aiohttp.GitHubAPI(aiohttp.ClientSession(),
                                                        'hail-is/hail',
                                                        oauth_token=oauth_token)
+    app['batch_client'] = await bc.BatchClient(billing_project='test')
```
jigold, line 274: This is fine for now, but ultimately we want this to be benchmark. Remind me before we start running the full submit function. Right now, this requires a Batch database migration.

benchmark-service/benchmark/utils.py
```diff
@@ -45,6 +49,16 @@ def list_benchmark_files(read_gs):
     return list_of_files
 
 
+async def submit_batch(batch_client, sha):
+    batch = batch_client.create_batch()
+    job = batch.create_job(image='ubuntu:18.04',
+                           command=['/bin/bash', '-c', 'touch /io/test; sleep 300'],
```
jigold, line 55: Can you make the resources on this be like this:

```
resources = {'cpu': 0.25}
```

We don't want to use entire cores for 5 minutes when we only have 3 cores available in the testing scenario.

benchmark-service/benchmark/utils.py
```diff
@@ -45,6 +49,16 @@ def list_benchmark_files(read_gs):
     return list_of_files
 
 
+async def submit_batch(batch_client, sha):
```
jigold, line 52: Can you name this `submit_test_batch`?

benchmark-service/benchmark/utils.py
```diff
@@ -68,3 +82,10 @@ def list_files(self, bucket_name):
         for blob in bucket.list_blobs():
             list_of_files.append('gs://' + bucket_name + '/' + blob.name)
         return list_of_files
+
+    def file_exists(self, file_path):
+        file_info = parse_file_path(FILE_PATH_REGEX, file_path)
+        bucket_name = file_info['bucket']
+        bucket = self.storage_client.bucket(bucket_name)
+        stats = storage.Blob(bucket=bucket, name=file_path).exists()
+        return stats
```
jigold, line 91: Can you rename stats to `exists`?

benchmark-service/deployment.yaml
```diff
@@ -34,6 +34,17 @@ spec:
...
+{% if deploy %}
+          - name: HAIL_BENCHMARK_BUCKET_NAME
+            value: hail-test
+{% else %}
+          - name: HAIL_BENCHMARK_BUCKET_NAME
+            value: hail-test-dmk9z
+{% endif %}
+{% if deploy %}
+          - name: START_POINT
+            value: "2020-10-01T00:00:00Z"
```
jigold, line 46: Let's set this to November 1st now.

jigold (dismissed):
This is looking really great! Just some small things.

benchmark-service/benchmark/benchmark.py
```diff
@@ -215,27 +217,51 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
+
+        batches = [b async for b in batch_client.list_batches(q=f'sha={sha} running')]
+
+        if not batches and not has_results_file:
+            new_commits.append(gh_commit)
+            batch_id = await submit_test_batch(batch_client, sha)
+            batch = await batch_client.get_batch(batch_id)
+            batch_status = await batch.last_known_status()
+            log.info(f'submitted a batch {batch_id} for commit {sha}')
+            commit = {
```
jigold, line 244: So something that is confusing to me is when benchmark restarts, won't you not have old commits in the `formatted_new_commits` and you're overwriting your old commits? Where is `new_commits` used? Maybe you want to break out this loop to something like this:

```
formatted_commits = []

for gh_commit in gh_data:
    sha = gh_commit.get('sha')
    file_path = f'gs://{HAIL_BENCHMARK_BUCKET_NAME}/benchmark-test/{sha}'
    has_results_file = gs_reader.file_exists(file_path)
    batches = [b async for b in batch_client.list_batches(q=f'sha={sha} running')]

    if not batches and not has_results_file:
         batch_id = await submit_test_batch(batch_client, sha)
         batch = await batch_client.get_batch(batch_id)
    else:
         batch = batches[:-1]

     commit = {

     }

     formatted_commits.append(commit)

benchmark_data = {
    'commits': formatted_commits
}

File: benchmark-service/benchmark/benchmark.py (hunk 8 above)
Comment (jigold) [reply benchmark-service/benchmark/benchmark.py:274]: I'll work on a PR now that handles this.

**Review by jigold — CHANGES_REQUESTED** [review]
Sorry. One last thing.

File: benchmark-service/benchmark/utils.py [hunk 14]
```diff
@@ -45,6 +49,17 @@ def list_benchmark_files(read_gs):
     return list_of_files
 
 
+async def submit_test_batch(batch_client, sha):
+    batch = batch_client.create_batch()
+    job = batch.create_job(image='ubuntu:18.04',
+                           command=['/bin/bash', '-c', 'touch /io/test; sleep 180'],
```
jigold, line 55: Alright. I'm so sorry for changing my mind. But I'm worried that 180 is way too long. Let's do 5. The problem is if there's 10 commits and they're each sleeping that long and we only have 3 cores to share amongst all the test_batch_* and test_ci, we're going to run into timeouts.

benchmark-service/benchmark/benchmark.py
jigold: See #9668

build.yaml
```diff
@@ -2042,6 +2042,19 @@ steps:
     - batch_database
     - deploy_auth
     - create_certs
+ - kind: deploy
```
jigold, line 2045: Was this duplicated now? I don't see the deletion of the previous code on the left.

jigold (approved):

jigold (approved):

PR #9639, opened 2020-10-23
[batch] clarify egress charges for input files

jigold:


hail/python/hailtop/batch/docs/service.rst
```diff
@@ -231,3 +231,12 @@ button next to the row for that batch. You can also delete a batch with the "Del
...
     Deleting a batch only removes it from the UI. You will still be billed for a deleted batch.
+
+
+Important Notes
+---------------
+
+.. warning::
+
+    To avoid expensive egress charges, input files should be in buckets that are multi-regional
+    in the United States / North America.
```
cseed, line 242: Awesome, thanks!

So I think we should remove "North America".  The multi-region is called "US": https://cloud.google.com/storage/docs/locations#location-mr.  I thought it included the Montreal region, but now I can't verify that and I'm not sure.  (Either way, we're not using it.)

I'd also specifically state that's because Batch stripes instances across multiple us- regions.

cseed (changes requested):

jigold:
Reassigning to @catoverdrive

cseed (approved):
Oops, sorry, I thought I had approved this!

PR #9643, opened 2020-10-26
[web_common] fix header when not logged in

jigold:


web_common/web_common/templates/header.html
```diff
@@ -25,8 +25,10 @@
 	  <a class="header-dropdown-menu-link" href="{{ batch_base_url }}/billing_projects">Billing Projects</a>
         {% endif %}
 
+	{% if userdata['username'] %}
         <a class="header-dropdown-menu-link" href="{{ batch_base_url}}/billing_limits">Billing Limits</a>
-
+        {% endif %}
```
danking, line 30: can you fix the formatting here? There might be tabs that need to be removed. Cotton's editor was misconfigured when he first wrote this stuff.

danking (changes requested):

danking (approved):

PR #9646, opened 2020-10-27
[query] Add stypes

tpoterba:


hail/src/main/scala/is/hail/types/physical/stypes/SStruct.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  def isFieldMissing(fieldIdx: Int): Code[Boolean] = {
+    pt.isFieldMissing(a, fieldIdx)
+  }
+}
+
+class SBaseStructPointerCode(val st: SBaseStructPointer, val a: Code[Long]) extends PBaseStructCode {
+  val pt: PBaseStruct = st.pType
+
+  def code: Code[_] = a
```
johnc1231, line 64: I think you can just rename `a` to `code` in the constructor, will keep you from forgetting the type is `Code[Long]` and remove need for `code` method.

hail/src/main/scala/is/hail/types/physical/PCanonicalString.scala
```diff
@@ -16,16 +18,8 @@ class PCanonicalString(val required: Boolean) extends PString {
   override def byteSize: Long = 8
 
   lazy val binaryFundamentalType: PCanonicalBinary = PCanonicalBinary(required)
```
johnc1231, line 20: I don't think there's any benefit to the indirection. here, seems like we should just delete `binaryFundamentalType` and let `fundamentalType = PCanonicalBinary(required)`

johnc1231 (changes requested):
This looks good! I left a few comments, am almost done reading through things. The only design thing I feel like is weird is the constant `null`s in `loadFrom` for the region. I think passing nulls is almost always a bad idea, and any future users of the method are not going to anticipate that it's ok to pass a `null` in less they go looking for examples of the method call. I could live with `assert` statements in the implementations of `loadFrom` for the branches that are going to require `Region` not be `null`, just so that we don't get random NPE's that we have to track down / figure out which object was null.

hail/src/main/scala/is/hail/annotations/ValueVisitor.scala
```diff
@@ -19,6 +21,12 @@ trait ValueVisitor {
 
   def visitBinary(b: Array[Byte]): Unit
 
+  def visitLocus(l: Locus); Unit
```
johnc1231, line 24: I'm pretty sure this should be a colon, not a semicolon? Not sure what the presence of a semicolon here is doing.

hail/src/main/scala/is/hail/annotations/Region.scala
```diff
@@ -355,7 +355,13 @@ object Region {
...
         }
         v.leaveTuple()
-      case t: ComplexPType =>
+      case t: PCanonicalInterval =>
+        visit(t.representation, off, v)
+      case t: PCanonicalLocus =>
+        visit(t.representation, off, v)
+      case t: PCanonicalCall =>
+        visit(t.representation, off, v)
+      case t: PCanonicalNDArray =>
```
johnc1231, line 364: Should these not use the `visitInterval`, `visitLocus`, etc. methods you added to `PrettyVisitor` in `ValueVisitor.scala`?

johnc1231 (commented):
Some comments about visitors

hail/src/main/scala/is/hail/types/physical/PCanonicalBaseStruct.scala
```diff
@@ -224,54 +166,45 @@ abstract class PCanonicalBaseStruct(val types: Array[PType]) extends PBaseStruct
...
+  def storeAtAddress(cb: EmitCodeBuilder, addr: Code[Long], region: Value[Region], value: PCode, deepCopy: Boolean): Unit = {
+    value.st match {
+      case SBaseStructPointer(t) if t.equalModuloRequired(this) =>
+        val pcs = value.asBaseStruct.memoize(cb, "pcbasestruct_store_src").asInstanceOf[SBaseStructPointerSettable]
+        val addrVar = cb.newLocal[Long]("addr", addr)
+        cb += Region.copyFrom(pcs.a, addrVar, byteSize)
+        if (deepCopy)
+          deepPointerCopy(cb, region, addrVar)
+      case _ =>
+        val addrVar = cb.newLocal[Long]("addr", addr)
```
johnc1231, line 194: `"addr"` is not a sufficiently unique name, let's use something more searchable / inline with `pcbasestruct_store_src`.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -446,7 +443,9 @@ class Emit[C](
 
   val methods: mutable.Map[(String, Seq[Type], Seq[PType], PType), EmitMethodBuilder[C]] = mutable.Map()
 
-  import Emit.E
+  import is.hail.utils._
+
+  import scala.language.{existentials, postfixOps}
```
johnc1231, line 448: Why here, not at top?

hail/src/main/scala/is/hail/types/physical/PType.scala
```diff
@@ -454,18 +449,21 @@ abstract class PType extends Serializable with Requiredness {
...
+  def getPointerTo(cb: EmitCodeBuilder, addr: Code[Long]): PCode
 
-  def constructAtAddress(addr: Long, region: Region, srcPType: PType, srcAddress: Long, deepCopy: Boolean): Unit
+  // stores a stack value as a region value of this type
+  def store(cb: EmitCodeBuilder, region: Value[Region], value: PCode, deepCopy: Boolean): Code[Long]
+
+  // stores a stack value inside pre-allocated memory of this type (in a nested structure, for instance).
+  def storeAtAddress(cb: EmitCodeBuilder, addr: Code[Long], region: Value[Region], value: PCode, deepCopy: Boolean): Unit
+
+  def unstagedStoreAtAddress(addr: Long, region: Region, srcPType: PType, srcAddress: Long, deepCopy: Boolean): Unit
```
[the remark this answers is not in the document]
tpoterba: The overloading has been driving me crazy. I can change this back if you want, but I think I'd prefer to go through and prefix all of the unstaged methods with `unstagedX` after this goes in.

hail/src/main/scala/is/hail/types/physical/stypes/SFloat32.scala
```diff
@@ -0,0 +1,77 @@
+package is.hail.types.physical.stypes
+
+import is.hail.annotations.{CodeOrdering, Region}
+import is.hail.asm4s.{Code, IntInfo, FloatInfo, Settable, SettableBuilder, TypeInfo, Value}
+import is.hail.expr.ir.{EmitCodeBuilder, EmitMethodBuilder, SortOrder}
+import is.hail.types.physical.{PCanonicalCall, PCode, PFloat32, PSettable, PType, PValue}
+import is.hail.utils.FastIndexedSeq
+
+case class SFloat32(required: Boolean) extends SType {
+  override def pType: PFloat32  = PFloat32(required)
```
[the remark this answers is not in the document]
tpoterba: these are implemented with singleton objects, so no allocation.

hail/src/main/scala/is/hail/types/physical/stypes/SBoolean.scala
```diff
@@ -0,0 +1,70 @@
...
+    val s = new SBooleanSettable(required, sb.newSettable[Boolean]("sboolean_memoize"))
+    s.store(cb, this)
+    s
+  }
+
+  def memoize(cb: EmitCodeBuilder, name: String): SBooleanSettable = memoizeWithBuilder(cb, name, cb.localBuilder)
+
+  def memoizeField(cb: EmitCodeBuilder, name: String): SBooleanSettable = memoizeWithBuilder(cb, name, cb.fieldBuilder)
+
+  def boolValue(cb: EmitCodeBuilder): Code[Boolean] = code
```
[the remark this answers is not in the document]
tpoterba: great point. Will fix all.

hail/src/main/scala/is/hail/types/physical/stypes/SFloat32.scala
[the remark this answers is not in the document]
johnc1231: Ah, neat :+1:

hail/src/test/scala/is/hail/types/physical/PSubsetStructSuite.scala
```diff
@@ -47,15 +47,4 @@ class PSubsetStructSuite extends PhysicalTestUtils {
...
-  @Test def testConstruction(): Unit = {
-    val ps1 = PCanonicalStruct("a" -> PCanonicalArray(PInt32(true)), "b" -> PInt64())
-    val ps2 = PCanonicalStruct("a" -> PCanonicalArray(PInt32(true)), "b" -> PInt64())
-
-    val srcType = PSubsetStruct(ps1, "b")
-    val destType = PSubsetStruct(ps2, "b")
-    val srcValue = Annotation(IndexedSeq(1,5,7,2,31415926), 31415926535897L)
-    val dstValue = Annotation(31415926535897L)
-    copyTestExecutor(srcType, destType, srcValue, deepCopy = false, interpret = true, expectedValue = dstValue)
-  }
```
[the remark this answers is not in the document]
tpoterba: These are testing the ptype functionality, which is something that doesn't really make sense. We should add back tests for the stypes and type coercion in that world, but that needs some new infrastructure we don't have yet.

hail/src/main/scala/is/hail/types/physical/stypes/SStruct.scala
tpoterba: `code` is gonna go away, so I'm not super concerned about this. Losing the type on the `Code[_]` isn't an issue, since clients who call `.code` on a pcode won't see a SStructPointerSettable, they'll see a SStructCode or SStructValue

hail/src/main/scala/is/hail/types/physical/PCanonicalString.scala
tpoterba: the PString interface has a `binaryFundamentalType: PBinary` abstract val that this needs to implement. Agree that this needs to change, but hard to do right now. We'll fix this as we get rid of fundamental types.

hail/src/main/scala/is/hail/annotations/ValueVisitor.scala
tpoterba: huh, interesting that compiles!

tpoterba: good catch.

hail/src/main/scala/is/hail/types/physical/PCanonicalBaseStruct.scala
tpoterba: 👍

hail/src/main/scala/is/hail/annotations/Region.scala
tpoterba: oops, meant to remove that stuff from ValueVisitor. We need new methods on the ptypes that load the java rep from the region rep, so this is a workaround for now.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
tpoterba: oh, huh. not sure.

hail/src/main/scala/is/hail/types/physical/PType.scala
```diff
@@ -454,18 +449,21 @@ abstract class PType extends Serializable with Requiredness {
     _copyFromAddress(region, srcPType, srcAddress, deepCopy)
   }
 
-  def constructAtAddress(mb: EmitMethodBuilder[_], addr: Code[Long], region: Value[Region], srcPType: PType, srcAddress: Code[Long], deepCopy: Boolean): Code[Unit]
-  def constructAtAddressFromValue(mb: EmitMethodBuilder[_], addr: Code[Long], region: Value[Region], srcPType: PType, src: Code[_], deepCopy: Boolean): Code[Unit]
-    = constructAtAddress(mb, addr, region, srcPType, asm4s.coerce[Long](src), deepCopy)
+  // return a PCode that can cheaply operate on the region representation. Generally a pointer type, but not necessarily (e.g. primitives).
+  def getPointerTo(cb: EmitCodeBuilder, addr: Code[Long]): PCode
```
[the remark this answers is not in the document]
tpoterba: renamed to "loadCheapPCode"

tpoterba:
oops, still need to fix the `null` loadFrom stuff.

tpoterba:
OK, I have a proposal that I'd prefer to break off into a followup PR that will solve the null issue. I want to add two methods to SType:

```scala
def codeFromCodeTuple(tup: IndexedSeq[Code[_]]): PCode

def settableFromSettableTuple(tup: IndexedSeq[Settable[_]]): PSettable
```

The concrete implementations of these methods will check that the size and types of the codes/settables match the expected tuples (a single value of type Long for SBaseStructPointer, a single value of type Int for SInt32, etc) and construct the appropriate code/settable.

As I'm writing this, though, I'm realizing I can also directly construct the appropriate PCodes= in the current PR as a temporary solution until we have the above.

tpoterba:
ok, done.

hail/src/main/scala/is/hail/types/physical/PType.scala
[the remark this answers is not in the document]
johnc1231: No, that's fine, I just thought being consistent would be good. Either we prefix the staged ones with "staged" or the unstaged ones with "unstaged". This is fine.

johnc1231 (approved):

PR #9648, opened 2020-10-27
[query/pcode] Remove IRInterval from IntervalFunctions

chrisvittal:
It was basically a proto PIntervalValue

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/functions/IntervalFunctions.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/functions/IntervalFunctions.scala
@@ -123,28 +123,26 @@ object IntervalFunctions extends RegistryFunctions {
           val compare = cb.emb.getCodeOrdering(int1.pt.pointType, int2.pt.pointType, CodeOrdering.Compare())
 
           def isAboveOnNonempty(cb: EmitCodeBuilder, lhs: PIntervalValue, rhs: PIntervalValue): Code[Boolean] = {
-            val cmp = r.mb.newLocal[Int]()
             val start = EmitCode.fromI(cb.emb)(lhs.loadStart(_))
             val end = EmitCode.fromI(cb.emb)(rhs.loadEnd(_))
             cb += start.setup
             cb += end.setup
-            cb.assign(cmp, compare(start.m -> start.v, end.m -> end.v))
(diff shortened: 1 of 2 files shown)


hail/src/main/scala/is/hail/expr/ir/functions/IntervalFunctions.scala
```diff
@@ -82,54 +82,71 @@ object IntervalFunctions extends RegistryFunctions {
...
-        Code(
-          interval1.storeToLocal,
-          interval2.storeToLocal,
-          !(interval1.isEmpty || interval2.isEmpty ||
-            interval1.isBelowOnNonempty(interval2) ||
-            interval1.isAboveOnNonempty(interval2))
-        )
+          interval1.isEmpty(cb) || interval2.isEmpty(cb) ||
+            isBelowOnNonempty(cb, interval1, interval2) ||
+            isAboveOnNonempty(cb, interval1, interval2)
```
johnc1231, line 147: I think this is going to result in generating the setup code for the start and end of both intervals twice. I'm not sure if that's a bad thing, but wanted to flag it and ask.

johnc1231 (changes requested):

hail/src/main/scala/is/hail/expr/ir/functions/IntervalFunctions.scala
tpoterba: it's memoized on line 121, this should be fine I think

johnc1231: So I see that it's memoized, but then it's converted into an `EmitCode` and the `setup` is added to the code builder in both methods. Is the expectation that the `setup`s are empty (or at least tiny) for something that is coming from a memoized value?

chrisvittal: 1. EmitCode.fromI currently always produces an empty setup.
2. This should be no worse than the old code in that it also does a double load, once for the `isEmpty`, and again for the `isAbove/isBelow` calls. An extra memoize to turn these loads into variable loads could be done, but I feel it's largely unnecessary.

johnc1231: Just wanted to make sure I understood. Reasoning makes sense

johnc1231:
Compilation errors

PR #9650, opened 2020-10-27
[query] Remove 'EmitCode' tuple from StagedArrayBuilder

chrisvittal:


hail/src/main/scala/is/hail/expr/ir/agg/TakeAggregator.scala
```diff
@@ -75,14 +75,15 @@ class TakeRVAS(val eltType: PType, val resultType: PArray, val kb: EmitClassBuil
...
   def combine(other: TakeRVAS): Code[Unit] = {
     val j = kb.genFieldThisRef[Int]()
-    val (eltJMissing, eltJ) = other.builder.loadElement(j)
+    val elt = other.builder.loadElement(j)
 
     Code(
       j := const(0),
       Code.whileLoop((builder.size < maxSize) & (j < other.builder.size),
-        eltJMissing.mux(
+        elt.setup,
```
chrisvittal, line 83: I know there's technically no setup here, but I don't want to hunt down weird bugs should there be setup in the future.

tpoterba (approved):

PR #9652, opened 2020-10-28
[query] New generic pretty-printer implementation

patrick-schultz:
This PR introduces a generic pretty-printing package, which for now is only used in `ir.Pretty`. The primary motivation was to separate format specification from rendering, to simplify the formatting code, and make it easier to add multiple formatting options, including forms with line number references.

Some nice properties of the new pretty-printer:
* Generic. Should be able to be used for all pretty-printing in hail scala code. This simplifies the codebase by making client pretty-printers easy to understand and modify, without getting bogged down in low-level details.
* Composable. Formatting specifications are trivially combinable, without needing to manually track context like the current indentation level, max line length, etc.
* Stack safe. Uses constant stack space.
* Uses constant heap space. Only keeps in memory text which might print in the current line, if it fits. (The pretty printer writes to a `java.io.Writer`, and I'm ignoring any heap space used by the writer.)
* Lazy. If the number of lines to print is capped, doesn't scan more of the tree than is needed to print those lines.
* Produces more readable output, printing nodes on a single line where possible. As we work to increase visibility into the compiler, I think this will be very helpful.

A formatted document is represented by the `Doc` type. This defines a `render` method, which takes three parameters to control the output:
* `width`: the maximum length of a line, including indentation
* `ribbonLength`: the maximum length of a line, not including indentation (too many characters on a line is hard to read, regardless of indentation)
* `maxLines`: the maximum number of lines to print

There are only a few `Doc` constructors, which suffice to define all methods in the richer api contained in the `prettyPrint` package object.
* `Text(t: String)`
* `Line(ifFlat: String)`
* `Indent(i: Int, body: Doc)`
* `Concat(it: Iterable[Doc])`
* `Group(body: Doc)`

Ignoring `Group` for the moment, the first four constructors define formatted documents with only one possible layout, regardless of `width` or `ribbonLength`. `Text` simply prints the string `t`; `Line` prints a newline, followed by the current level of indentation (`ifFlat` is explained when we discuss `Group`); `Indent` increases the indentation of all `Line`s contained in `body` by `i`; and `Concat` simply prints all documents in `it` sequentially.

`Group` is the sole source of alternatives which `render` must choose between. `Group(body)` can be rendered in one of two ways:
* replace all `Line`s contained in `body` (including in nested `Group`s) by their `ifFlat` alternative (almost always either " " or ""), or, if that would cause the line to exceed `width`
* print `body` normally, as described in the previous paragraph, allowing nested `Group`s to print either flat or normally.

This pretty-printer DSL has become fairly standard, with some common enhancements that I don't think we need. It was first described in [A prettier printer](https://homepages.inf.ed.ac.uk/wadler/papers/prettier/prettier.pdf) by Wadler (though my implementation is completely different).

This achieves stack safety by `Concat` taking an `Iterable`, so each contained `Doc` can be produced on demand. `render` pulls from these iterators, keeping in memory only things that might print to the current line, but where the format hasn't been decided yet. As soon as the formatting of a group is decided, as much of its body as possible is written to the `java.io.Writer`.

A very quick and dirty performance comparison had the new pretty printer about 20% slower. That's paying for both the stack safety and the added smarts. And I think there's still room for optimization if it becomes necessary.

Here is a snippet of the IR generated by `test_ld_score_regression`, first on master, then this PR:
```
                    (InsertFields
                      (SelectFields (SNP A1 A2 N Z)
                        (Ref row))
                      None
                      (chi_squared
                        (Apply pow () Float64
                          (GetField Z
                            (Ref row))
                          (ApplyIR toFloat64 () Float64
                            (I32 2))))
                      (n
                        (GetField N
                          (Ref row)))
                      (ld_score
                        (GetField L2
                          (GetField __uid_3
                            (Ref row))))
                      (locus
                        (Apply Locus () Locus(GRCh37)
                          (GetField CHR
                            (GetField __uid_4
                              (Ref row)))
                          (GetField BP
                            (GetField __uid_5
                              (Ref row)))))
                      (alleles
                        (MakeArray Array[String]
                          (GetField A2
                            (Ref row))
                          (GetField A1
                            (Ref row))))
                      (phenotype
                        (Str "50_irnt")))))
                (InsertFields
                  (SelectFields (locus alleles chi_squared n ld_score phenotype)
                    (SelectFields (SNP A1 A2 N Z chi_squared n ld_score locus alleles phenotype)
                      (Ref row)))
                  None))
```
```
                    (InsertFields
                      (SelectFields (SNP A1 A2 N Z) (Ref row))
                      None
                      (chi_squared
                        (Apply pow () Float64
                          (GetField Z (Ref row))
                          (ApplyIR toFloat64 () Float64 (I32 2))))
                      (n (GetField N (Ref row)))
                      (ld_score
                        (GetField L2 (GetField __uid_3 (Ref row))))
                      (locus
                        (Apply Locus () Locus(GRCh37)
                          (GetField CHR (GetField __uid_4 (Ref row)))
                          (GetField BP (GetField __uid_5 (Ref row)))))
                      (alleles
                        (MakeArray Array[String]
                          (GetField A2 (Ref row))
                          (GetField A1 (Ref row))))
                      (phenotype (Str "50_irnt")))))
                (InsertFields
                  (SelectFields 
                    (locus alleles chi_squared n ld_score phenotype)
                    (SelectFields 
                      (SNP A1 A2 N Z chi_squared n ld_score locus
                        alleles phenotype)
                      (Ref row)))
                  None))
```


hail/src/main/scala/is/hail/expr/ir/Pretty.scala
```diff
@@ -33,455 +52,329 @@ object Pretty {
...
+    case NDArrayAgg(_, axes) => single(prettyInts(axes, elideLiterals))
+    case NDArrayRef(_, _, errorId) => single(s"$errorId")
+    case ArraySort(_, l, r, _) => FastSeq(prettyIdentifier(l), prettyIdentifier(r))
+    case ApplyIR(function, typeArgs, _) => FastSeq(prettyIdentifier(function), prettyTypes(typeArgs), ir.typ.parsableString())
+    case Apply(function, typeArgs, _, t) => FastSeq(prettyIdentifier(function), prettyTypes(typeArgs), t.parsableString())
+    case ApplySeeded(function, _, seed, t) => FastSeq(prettyIdentifier(function), seed.toString, t.parsableString())
+    case ApplySpecial(function, typeArgs, _, t) => FastSeq(prettyIdentifier(function), prettyTypes(typeArgs), t.parsableString())
+    case SelectFields(_, fields) => single(fillList(fields.view.map(f => text(prettyIdentifier(f)))))
+    case LowerBoundOnOrderedCollection(_, _, onKey) => single(prettyBooleanLiteral(onKey))
+    case In(i, typ) => single(s"$typ $i")
```
tpoterba, line 200: this is wrong, yeah?

hail/src/main/scala/is/hail/utils/prettyPrint/PrettyPrintWriter.scala
```diff
@@ -0,0 +1,187 @@
...
+            kont.push(UnindentK(i))
+            currentNode = body
+          case Concat(bodyIt) =>
+            kont.push(ConcatK(bodyIt.iterator))
+            advance()
+        }
+      }
+      closeGroups()
+    } catch {
+      case _: MaxLinesExceeded =>
```
tpoterba, line 155: why don't we want to throw here?

hail/src/main/scala/is/hail/utils/prettyPrint/PrettyPrintWriter.scala
```diff
@@ -0,0 +1,187 @@
...
+  }
+}
+
+private[prettyPrint] case class Text(t: String) extends Doc
+private[prettyPrint] case class Line(ifFlat: String) extends Doc
+private[prettyPrint] case class Group(body: Doc) extends Doc
+private[prettyPrint] case class Indent(i: Int, body: Doc) extends Doc
+private[prettyPrint] case class Concat(it: Iterable[Doc]) extends Doc
+
+private[prettyPrint] abstract class ScanedNode
```
tpoterba, line 177: typo: scanned

tpoterba (dismissed):
a few small questions to address

hail/src/main/scala/is/hail/expr/ir/Pretty.scala
patrick-schultz: Whoops, good catch

hail/src/main/scala/is/hail/utils/prettyPrint/PrettyPrintWriter.scala
patrick-schultz: I'm using the exception just as a control flow mechanism, it's not an error. This is to implement a `maxLines` option, which will keep formatting and printing lines until it hits `maxLines`, then stops without needing to traverse the rest of the input tree.

patrick-schultz: So basically, it lets any of the helper functions break out of the outer while loop.

patrick-schultz: `MaxLinesExceeded` should really be private to `prettyPrint`. Hopefully that makes the intent clearer. Adding a comment in the empty handler too.

tpoterba: ah, got it!

tpoterba (approved):

PR #9653, opened 2020-10-28
[batch] Add pod disruption budget for batch-driver

jigold:
Not sure if this will help make GKE move the pod if the node needs to be repaired rather than waiting for the node to repair.

### DIFF
--- a/batch/deployment.yaml
+++ b/batch/deployment.yaml
@@ -124,6 +124,16 @@ spec:
            optional: false
            secretName: ssl-config-batch-driver
 ---
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: batch-driver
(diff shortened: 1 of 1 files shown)


batch/deployment.yaml
```diff
@@ -124,6 +124,16 @@ spec:
...
 ---
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: batch-driver
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      app: batch
```
danking, line 135: this needs to be batch-driver

danking (changes requested):
I think the real fix is to make batch-driver non-preemptible. I'm not sure this would help since there's only ever 1 driver anyway.

PR #9657, opened 2020-10-29
Throw a good error message if ASYNC_PROFILER_HOME isn't set

johnc1231:
Without this change, it just sticks `"None"` in the url it generates, then says it couldn't find the `.so` it's looking for.


patrick-schultz (approved):
Thanks, John

PR #9658, opened 2020-10-29
[batch] Self service create accounts for Broad users

jigold:
- Signup page with web socket and spinner while waiting for account to create
- Upon account creation, a billing project named `{username}-trial` is created with $10 limit and a user `{username}`
- When deleting an account, the billing project is reopened if it's closed, then remove the user, and finally close the billing project. This behavior might be debatable. We may not need to remove the user from the billing project. I think it's better to be safe in case the billing project is reopened.

Auth-driver is implicitly dependent on the batch front end, but this dependency isn't stated in build.yaml. An event queue is a future solution.

I can get rid of my personal email being whitelisted, but I think it will be useful if we need to debug later and for a possible demo on Monday (although I'll probably just use some screenshots).

If you want to test it in your namespace, make sure to comment out all the create and delete steps that are not related to billing projects.


auth/auth/auth.py
```diff
@@ -53,13 +76,153 @@ async def get_index(request):  # pylint: disable=unused-argument
...
 async def login(request, userdata):
+    db = request.app['db']
+
     next = request.query.get('next', deploy_config.external_url('notebook', ''))
+    session = await aiohttp_session.get_session(request)
+
     if userdata:
         return aiohttp.web.HTTPFound(next)
 
+    if 'pending' in session:
```
danking, line 196: is this if block exactly the same as the one above? Can we coalesce?

auth/auth/auth.py
```diff
@@ -53,13 +76,153 @@ async def get_index(request):  # pylint: disable=unused-argument
...
+        user = await user_from_email(db, email)
+
+        if user is None:
+            cleanup_session(session)
+            return aiohttp.web.HTTPFound(next)
+
+        if user['state'] == 'deleting' or user['state'] == 'deleted':
+            cleanup_session(session)
+            set_message(session, f'Account for {user["username"]} is deleted.', 'error')
+            return aiohttp.web.HTTPFound(next)
```
danking, line 99: I feel like this shouldn't go to `next`, the user has no account, they can't possibly do anything. Shouldn't we bounce back to notebook.hail.is?

danking: Or maybe an error page specifically about your account being deleted?

auth/auth/auth.py
```diff
@@ -53,13 +76,153 @@ async def get_index(request):  # pylint: disable=unused-argument
...
+
+    return aiohttp.web.HTTPFound(authorization_url)
+
+
+async def _wait_websocket(request, email):
+    app = request.app
+    db = app['db']
+
+    ws = web.WebSocketResponse()
+    await ws.prepare(request)
```
danking, line 136: You need to add a `try:` after the prepare and a `finally` at the end of this method which closes the web socket.

auth/auth/auth.py
```diff
@@ -53,13 +76,153 @@ async def get_index(request):  # pylint: disable=unused-argument
...
+
+    user = await user_from_email(db, email)
+    if not user:
+        return web.HTTPNotFound()
```
danking, line 140: You should do this before you create a web socket. I'm not even sure what happens if you try to send an HTTP response over a connection that has been converted to a web socket.

auth/auth/auth.py
```diff
@@ -53,13 +76,153 @@ async def get_index(request):  # pylint: disable=unused-argument
...
+
+    # forward authorization
+    headers = {}
```
danking, line 143: this variable is unused?

auth/auth/auth.py
```diff
@@ -53,13 +76,153 @@ async def get_index(request):  # pylint: disable=unused-argument
...
+    if 'Authorization' in request.headers:
+        headers['Authorization'] = request.headers['Authorization']
+    if 'X-Hail-Internal-Authorization' in request.headers:
+        headers['X-Hail-Internal-Authorization'] = request.headers['X-Hail-Internal-Authorization']
+
+    cookies = {}
```
danking, line 149: this variable is unused?

auth/auth/auth.py
```diff
@@ -53,13 +76,153 @@ async def get_index(request):  # pylint: disable=unused-argument
...
+                log.info(f"user {user['username']} status changed: {user['state']} => {updated_user['state']}")
+                break
+        except Exception:  # pylint: disable=broad-except
+            log.exception(f"/signup/wait: error while updating status for user {user['username']}")
+        await asyncio.sleep(1)
+        count += 1
+
+    ready = updated_user and updated_user['state'] == 'active'
+
+    # 0/1 ready
```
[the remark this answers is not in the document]
danking: oh 0 means not ready and 1 means ready? Can we be more explicit about that in the comment or just not have the comment?

auth/auth/auth.py
```diff
@@ -53,13 +76,153 @@ async def get_index(request):  # pylint: disable=unused-argument
...
+        try:
+            updated_user = await user_from_email(db, email)
+            changed = user['state'] != updated_user['state']
+            if changed:
+                log.info(f"user {user['username']} status changed: {user['state']} => {updated_user['state']}")
+                break
+        except Exception:  # pylint: disable=broad-except
+            log.exception(f"/signup/wait: error while updating status for user {user['username']}")
+        await asyncio.sleep(1)
+        count += 1
```
danking, line 166: Just to be clear, this fails if the user isn't created in 10s, that's the intended behavior? I think we should log if we loop 10 times without seeing the status change.

auth/auth/auth.py
```diff
@@ -53,13 +76,153 @@ async def get_index(request):  # pylint: disable=unused-argument
...
+        cookies['session'] = request.cookies['session']
+    if 'sesh' in request.cookies:
+        cookies['sesh'] = request.cookies['sesh']
+
+    count = 0
+    while count < 10:
+        try:
+            updated_user = await user_from_email(db, email)
+            changed = user['state'] != updated_user['state']
+            if changed:
```
danking, line 160: This seems a bit odd? shouldn't we just loop until `user['state'] in {'active', 'deleting', 'deleted'}`?

auth/auth/auth.py
```diff
@@ -92,18 +256,48 @@ async def callback(request):
...
-                 "SELECT * FROM users WHERE email = %s AND state = 'active';", email)]
 
-    if len(users) != 1:
+    user = await user_from_email(db, email)
+
+    caller = session['caller']
+
+    if caller == 'login':
+        if user is None:
+            set_message(session, f'Account does not exist for email {email}', 'error')
```
danking, line 266: This case should go to notebook.hail.is or an error page, not to `next`.

danking (changes requested):
Looking good, just a few comments mostly about getting all the possible flows right.

auth/auth/driver/driver.py
```diff
@@ -288,11 +289,74 @@ async def delete(self):
         self.name = None
 
 
+class BillingProjectResource:
```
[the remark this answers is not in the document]
jigold: The cleanup at the end in `_delete` closes the billing project if there's an error. This is run if the state is 'deleting' or if the 'creating' step fails. If it fails in 'creating', we want to be able to try again.

auth/auth/auth.py
jigold: copy paste mess-up

auth/auth/auth.py
jigold: It's not failing per se. It just causes the signup page to reload

```
  let protocol = location.protocol.replace("http", "ws")
  let sock = new WebSocket(protocol + location.host + "{{ base_path }}/signup/wait");
  sock.onmessage = function (event) {
      window.location.reload()
  }
```

auth/auth/templates/signup.html
```diff
@@ -0,0 +1,35 @@
+{% extends "layout.html" %}
+{% block title %}Signup{% endblock %}
+{% block content %}
+<div class="vcentered">
+  {% if state == 'active' %}
```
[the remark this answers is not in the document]
jigold: I just took out the active block instead. I didn't quite follow your example.

jigold:
Closing while I debug.

jigold:
I completely refactored this. You'll probably want to review the auth.py code from scratch.

auth/auth/auth.py
```diff
@@ -53,12 +76,91 @@ async def get_index(request):  # pylint: disable=unused-argument
...
+        count = 0
+        while count < 10:
+            try:
+                user = await user_from_email(db, email)
+                assert user
+                if user['state'] != 'creating':
+                    log.info(f"user {user['username']} is no longer creating")
+                    break
+            except Exception:  # pylint: disable=broad-except
+                log.exception(f"/signup/wait: error while updating status for user {user['username']}")
```
danking, line 146:
```suggestion
                log.exception(f"/creating/wait: error while updating status for user {user['username']}")
```

danking (dismissed):
still looking, but this should get fixed.

danking (changes requested):
OK, this looks much better. I think the flows are mostly correct now.

I think we can simplify the oauth2callback to make it easier to follow and less duplicative.

auth/auth/auth.py
```diff
@@ -53,12 +76,91 @@ async def get_index(request):  # pylint: disable=unused-argument
...
+            'email': user['email']
+        }
+
+        if user['state'] == 'deleting' or user['state'] == 'deleted':
+            cleanup_session(session)
+            return await render_template('auth', request, userdata, 'account-error.html', page_context)
+
+        if user['state'] == 'active':
+            session_id = await create_session(db, user['id'])
+            cleanup_session(session)
```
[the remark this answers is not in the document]
jigold: It doesn't happen in the creating case, but I'll just explicitly add the pending and email back to the session in that case.

jigold:
Closing while I double check everything is still good.

jigold:
Thanks for the good review!

danking (approved):

PR #9660, opened 2020-10-30
[docker] Infrastructure for moving third party images to gcr

jigold:
`make -C docker/third-party copy`

### DIFF
--- a/docker/third-party/copy_images.sh
+++ b/docker/third-party/copy_images.sh
@@ -6,7 +6,7 @@ images=$(cat images.txt)
 
 if [ -z "${PROJECT}" ]
 then
-    echo ERROR: GCR_PROJECT must be set before running this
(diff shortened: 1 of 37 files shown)


jigold:
Maybe this is a better solution?

https://cloud.google.com/container-registry/docs/pulling-cached-images

docker/third-party/copy_images.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -ex
+
+images=$(cat images.txt)
+
+if [ -z "${PROJECT}" ]
+then
+    echo ERROR: GCR_PROJECT must be set before running this
```
danking, line 9: no GCR_

danking (changes requested):
Looks fine, one error.

danking (approved):

PR #9661, opened 2020-10-30
[batch] Add warnings for non-gcr images

jigold:
I'm not sure why we had the default be ubuntu:latest but I think 18.04 is better.

### DIFF
--- a/hail/python/hailtop/batch/__init__.py
+++ b/hail/python/hailtop/batch/__init__.py
@@ -1,4 +1,5 @@
 import nest_asyncio
+import warnings
 
 from .batch import Batch
 from .batch_pool_executor import BatchPoolExecutor
@@ -22,3 +23,6 @@ __all__ = ['Batch',
            ]
 
 nest_asyncio.apply()
+
+warnings.filterwarnings('once', append=True)
+del warnings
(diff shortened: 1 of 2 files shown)


konradjk:
When does this get triggered?

jigold:
Not sure I understand what you mean. When the image name doesn't start with gcr.io in the backend code for the service. You can't check it until you call b.run()

konradjk:
Yeah I meant what point in the pipeline. One could check on `j.image` (or `hb.Batch(..., default_image=...)`), but maybe `b.run` is best - is it only triggered once per batch, or once per offending job, or once per offending unique image?

jigold:
It should be triggered once per offending unique image. I ran the same job three times each for ubuntu:18.04 and ubuntu:16.04 for a total of 6 jobs.

```
wmecc-475:hail jigold$ python3 batch/test_image_warnings.py
/Users/jigold/hail/hail/python/hailtop/batch/backend.py:478: UserWarning: Using an image ubuntu:18.04 not in GCR. Jobs may fail due to Docker Hub rate limits.
  UserWarning)
/Users/jigold/hail/hail/python/hailtop/batch/backend.py:478: UserWarning: Using an image ubuntu:16.04 not in GCR. Jobs may fail due to Docker Hub rate limits.
  UserWarning)
```

konradjk (approved):

PR #9662, opened 2020-10-30
[batch] Make driver nonpreemptible in production

jigold:
I think this is how to make it nonpreemptible in production only.

### DIFF
--- a/batch/deployment.yaml
+++ b/batch/deployment.yaml
@@ -22,12 +22,13 @@ spec:
       serviceAccountName: batch
 {% if deploy %}
       priorityClassName: production
-{% endif %}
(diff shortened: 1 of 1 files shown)


batch/deployment.yaml
```diff
@@ -22,12 +22,13 @@ spec:
       serviceAccountName: batch
 {% if deploy %}
       priorityClassName: production
-{% endif %}
```
danking, line 25: add:
```
      nodeSelector:
        preemptible: "false"
```
Like ci & ghost

danking (changes requested):

danking (approved):

PR #9663, opened 2020-10-30
[datasets][annotationdb] clean up docs, add doc page for DB class

pwc2:
Made changes to clean up/add to the documentation for the datasets API and annotation DB.

Changes to documentation:

- Moved raw html in `annotation_database_ui.rst` into `hail/python/hail/docs/_static/annotationdb/annotationdb.html`.
- Added html table to bottom of datasets doc page to show all available datasets, similar to what is currently on annotation DB doc page. Added relevant files to `hail/python/hail/docs/_static/datasets`.
- Moved schemas currently on datasets doc page to their own page. Datasets doc page links to this new page.
- Moved some minor styling from html files to `annotationdb.css`, and cleaned up formatting of html.
- Added doc page for the `DB` class, referenced on the Python API doc page for the experimental module. Only exposes the `available_datasets` attribute and `annotate_rows_db` method, as I didn't think most users need to see any internal methods or the `Dataset` and `DatasetVersion` classes/methods.

Most of diff is just formatting changes to `db.py` and `datasets.py` for consistency/readability that did not change functionality. Also added docstrings to methods in `db.py` that were missing them.

Added a check in `DB` constructor to make sure the cloud and region combination is valid. To prevent an empty annotation DB instance from being created if user specifies `db = hl.experimental.DB(region='eu', cloud='aws')`, since we don't have an EU bucket on AWS.


tpoterba:
This is awesome.

danking (approved):
This is a huge improvement and I love it!

I pushed [a commit](https://github.com/hail-is/hail/pull/9663/commits/3ae2d5fea8169734dfa4eeec68d642a38105ceeb) to your PR which cleaned up some visual issues I noticed. In particular, I switched to using an auto-generated hail.experimental.DB.rst file. I'm not sure why, but our auto generator was set to only trigger on certain files. I switched it to auto generate class RST files for *every* class reference it finds.

I also fixed a latent formatting in sparse_split_multi.

PR #9666, opened 2020-11-02
Linear Regression ND Performance Fix

johnc1231:
#9634 Introduced a large performance regression in the `linear_regression_rows_nd` benchmark (making it about 4x slower). This PR fixes that by doing two things:

1. Move all the global into one single `annotate_globals` expression, so that CSE can work properly. This required fixing a bug in some ndarray expressions that were not correctly tracking their source tables. To make sure I was only referencing the global versions of this computed things, rather accidentally recomputing, I wrapped the global setup in a function to scope the variables. This improvement was minor, didn't hit the real root of the problem.

2. Much more significantly, and not 100% clear why: `process_y_group` is now a function that returns a python dictionary, instead of a hail struct. I can guess that the allocation required by making a struct was wasteful, but it seems crazy that it was "make the benchmark 4x slower" amounts of wasteful. 

While this is not user facing yet, would be good to get this in before an eventual 0.2.60 release if we want to avoid benchmarks regressing between versions.

### DIFF
--- a/hail/python/hail/methods/statgen.py
+++ b/hail/python/hail/methods/statgen.py
@@ -498,10 +498,10 @@ def _linear_regression_rows_nd(y, x, covariates, block_size=16, pass_through=())
             xxpRec = (dot_rows_with_themselves(X.T) - dot_rows_with_themselves(Qtx.T)).map(lambda entry: 1 / entry)
             b = xyp * xxpRec
             se = ((1.0 / ht.ds[idx]) * (ht.__yyps[idx].reshape((-1, 1)) @ xxpRec.reshape((1, -1)) - (b * b))).map(lambda entry: hl.sqrt(entry))
-            t = b / se
-            p = t.map(lambda entry: 2 * hl.expr.functions.pT(-hl.abs(entry), ht.ds[idx], True, False))
+            # t = b / se
+            # p = t.map(lambda entry: 2 * hl.expr.functions.pT(-hl.abs(entry), ht.ds[idx], True, False))
             return hl.struct(n=hl.range(rows_in_block).map(lambda i: n), sum_x=sum_x._data_array(), y_transpose_x=ytx.T._data_array(), beta=b.T._data_array(),
-                             standard_error=se.T._data_array(), t_stat=t.T._data_array(), p_value=p.T._data_array())
+                             standard_error=se.T._data_array()) # t_stat=t.T._data_array(), p_value=p.T._data_array())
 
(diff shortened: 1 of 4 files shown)


johnc1231:
So after chatting last week, I think the conclusion we came to is that `per_y_list` was probably being recomputed for each row of output, which is why changing it to a python list comprehension helped. I tried changing things to avoid that by adding an `rbind` to the old version:

```
def build_row(per_y_list, row_idx):
            # For every field we care about, map across all y's, getting the row_idxth one from each.
            idxth_keys = {field_name: block[field_name][row_idx] for field_name in key_field_names}
            computed_row_field_names = ['n', 'sum_x', 'y_transpose_x', 'beta', 'standard_error', 't_stat', 'p_value']
            computed_row_fields = {
                #field_name: [one_y[field_name][row_idx] for one_y in per_y_list] for field_name in computed_row_field_names
                field_name: per_y_list.map(lambda one_y: one_y[field_name][row_idx]) for field_name in computed_row_field_names
            }
            pass_through_rows = {
                field_name: block[field_name][row_idx] for field_name in row_field_names
            }

            if not is_chained:
                computed_row_fields = {key: value[0] for key, value in computed_row_fields.items()}

            return hl.struct(**{**idxth_keys, **computed_row_fields, **pass_through_rows})

        new_rows = hl.rbind(per_y_list, lambda l_per_y_list: hl.range(rows_in_block).map(lambda inner_i: build_row(l_per_y_list, inner_i)))
```

and had no luck, still incredibly slow. Binding has actually somehow made it slower than the current `main` branch, which doesn't make sense to me.

johnc1231:
I tried incrementally computing one additional field at a time in `process_y_group`, turns out everything was fine until I calculated `p_value`. To fix, I needed to let bind both `t` and `ht.ds[idx]`. I still don't really understand why though. The `ht.ds[idx]` change is just a code motion change that makes a huge difference.

hail/python/hail/methods/statgen.py
```diff
@@ -530,7 +538,7 @@ def process_partition(part):
     res = ht._map_partitions(process_partition)
 
     if not y_is_list:
-        fields = ['y_transpose_x', 'beta', 'standard_error', 't_stat', 'p_value']
+        fields = ['y_transpose_x', 'beta' 'standard_error', 't_stat', 'p_value']
```
patrick-schultz, line 541: Typo?

patrick-schultz (dismissed):

hail/python/hail/methods/statgen.py
johnc1231: Agh, I hate this about python. Fixing.

patrick-schultz (approved):

PR #9667, opened 2020-11-03
[build] time jar modification

tpoterba:


tpoterba:
er, can't time like this in make

tpoterba:
why did this pass? um...

danking:
The shadow jar file was newer than its dependencies so Make assumed it was good to go. The Make-Way TM would be to make the output of gradle an intermediate step which is then copied to a new file by the jar modification.

danking (changes requested):
xxx

danking (approved):

tpoterba:
I never fixed this

PR #9670, opened 2020-11-03
[benchmark] fix file path by adding .json ext

Dania-Abuhijleh:


jigold (dismissed):

benchmark-service/benchmark/benchmark.py
```diff
@@ -229,7 +229,7 @@ async def update_commits(app):
 
     for gh_commit in gh_data:
         sha = gh_commit.get('sha')
```
jigold, line 231: Let's add one here for logging what commit we're looking at.

jigold: I can't put a comment higher, but let's add one with the length of `gh_data`

jigold (changes requested):

benchmark-service/benchmark/benchmark.py
jigold: Let's also log START_POINT. It's possible it's only looking at today.

jigold: oh. maybe the date is exclusive and not inclusive. If that's the case, then your code is correct and we just need to keep an eye on why it's not finding the results file

jigold (approved):

PR #9673, opened 2020-11-04
[Benchmark] fix file path

Dania-Abuhijleh:
fix file path in file_exists function

### DIFF
--- a/benchmark-service/benchmark/utils.py
+++ b/benchmark-service/benchmark/utils.py
@@ -67,10 +67,10 @@ class ReadGoogleStorage:
     def get_data_as_string(self, file_path):
         file_info = parse_file_path(FILE_PATH_REGEX, file_path)
         bucket = self.storage_client.get_bucket(file_info['bucket'])
-        shorter_file_path = file_info['path']
+        path = file_info['path']
         try:
             # get bucket data as blob
-            blob = bucket.blob(shorter_file_path)
+            blob = bucket.blob(path)
             # convert to string
             data = blob.download_as_string()
         except Exception as e:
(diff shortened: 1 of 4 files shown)


benchmark-service/deployment.yaml
```diff
@@ -37,13 +37,11 @@ spec:
 {% if deploy %}
           - name: HAIL_BENCHMARK_BUCKET_NAME
             value: hail-test
+          - name: START_POINT
```
jigold, line 40: I don't understand why this change would fix the behavior we were seeing. Can you explain?

benchmark-service/benchmark/utils.py
```diff
@@ -88,6 +88,7 @@ def file_exists(self, file_path):
         file_info = parse_file_path(FILE_PATH_REGEX, file_path)
         bucket_name = file_info['bucket']
         bucket = self.storage_client.bucket(bucket_name)
-        exists = storage.Blob(bucket=bucket, name=file_path).exists()
-        log.info(f'file in bucket {bucket_name} exists? {exists}')
+        shorter_file_path = file_info['path']
```
jigold, line 91: This is just `path`. There's two components: the bucket and the path.

jigold (changes requested):

benchmark-service/deployment.yaml
Dania-Abuhijleh: It wouldn't, I just thought it was repetitive to have {% if deploy %} twice

benchmark-service/benchmark/utils.py
Dania-Abuhijleh: I thought I got the bucket on line 89, is there something else?

jigold: You did. I'm saying I don't like the name `shorter_file_path`. Just name it `path`

benchmark-service/deployment.yaml
jigold: Ok. There's still a problem with START_POINT. The deployment logs says this is the START_POINT:

```
2020-11-03 22:03:59,472 start point is 2020-11-02T21:38:53Z
```

benchmark-service/benchmark/utils.py
Dania-Abuhijleh: ahhh

jigold (approved):

PR #9675, opened 2020-11-04
Region Pool Per Task instead of Per Thread

johnc1231:
This PR moves us from having thread local `RegionPool`s, to "task local" `RegionPool`s, based on Spark tasks. This should make it easier for us to track peak memory usage across a pipeline.

### DIFF
--- a/hail/src/main/scala/is/hail/annotations/Region.scala
+++ b/hail/src/main/scala/is/hail/annotations/Region.scala
@@ -265,9 +265,6 @@ object Region {
     case t: PBaseStruct => (addr, v) => Region.copyFrom(coerce[Long](v), addr, t.byteSize)
   }
 
-  def stagedCreate(blockSize: Size): Code[Region] =
-    Code.invokeScalaObject2[Int, RegionPool, Region](Region.getClass, "apply", asm4s.const(blockSize), Code._null)
-
   def stagedCreate(blockSize: Size, pool: Code[RegionPool]): Code[Region] =
     Code.invokeScalaObject2[Int, RegionPool, Region](Region.getClass, "apply", asm4s.const(blockSize), pool)
 
--- a/hail/src/main/scala/is/hail/annotations/Region.scala
+++ b/hail/src/main/scala/is/hail/annotations/Region.scala
@@ -268,7 +268,7 @@ object Region {
   def stagedCreate(blockSize: Size, pool: Code[RegionPool]): Code[Region] =
     Code.invokeScalaObject2[Int, RegionPool, Region](Region.getClass, "apply", asm4s.const(blockSize), pool)
 
-  def apply(blockSize: Region.Size = Region.REGULAR, pool: RegionPool = null): Region = {
+  def apply(blockSize: Region.Size = Region.REGULAR, pool: RegionPool): Region = {
     (if (pool == null) {
       val htc = HailTaskContext.get()
       if (htc == null) {
--- a/hail/src/main/scala/is/hail/annotations/Region.scala
+++ b/hail/src/main/scala/is/hail/annotations/Region.scala
@@ -268,6 +268,9 @@ object Region {
   def stagedCreate(blockSize: Size): Code[Region] =
     Code.invokeScalaObject2[Int, RegionPool, Region](Region.getClass, "apply", asm4s.const(blockSize), Code._null)
 
+  def stagedCreate(blockSize: Size, pool: Code[RegionPool]): Code[Region] =
+    Code.invokeScalaObject2[Int, RegionPool, Region](Region.getClass, "apply", asm4s.const(blockSize), pool)
+
   def apply(blockSize: Region.Size = Region.REGULAR, pool: RegionPool = null): Region = {
     (if (pool == null) RegionPool.get else pool)
       .getRegion(blockSize)
@@ -448,6 +451,10 @@ final class Region protected[annotations](var blockSize: Region.Size, var pool: 
 
   def prettyBits(): String = {
     "FIXME: implement prettyBits on Region"
+  }
+
+  def getPool(): RegionPool = {
+    pool
   }
 }
 
--- a/hail/src/main/scala/is/hail/annotations/Region.scala
+++ b/hail/src/main/scala/is/hail/annotations/Region.scala
@@ -15,8 +15,6 @@ object Region {
   val SIZES: Array[Long] = Array(64 * 1024, 8 * 1024, 1024, 256)
   val BLOCK_THRESHOLD: Long = 4 * 1024
 
-  def scoped[T](f: Region => T): T = using(Region())(f)
-
   def loadInt(addr: Long): Int = Memory.loadInt(addr)
 
   def loadLong(addr: Long): Long = Memory.loadLong(addr)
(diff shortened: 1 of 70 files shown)


hail/src/main/scala/is/hail/backend/local/LocalBackend.scala
```diff
@@ -68,7 +78,10 @@ class LocalBackend(
   def parallelizeAndComputeWithIndex(backendContext: BackendContext, collection: Array[Array[Byte]],
     dependency: Option[TableStageDependency] = None)(f: (Array[Byte], Int) => Array[Byte]): Array[Array[Byte]] = {
     collection.zipWithIndex.map { case (c, i) =>
-      f(c, i)
+      HailTaskContext.setTaskContext(new LocalTaskContext())
+      val bytes = f(c, i)
+      HailTaskContext.unset()
+      bytes
```
johnc1231, line 84: @tpoterba I think this PR is almost done. I'm not sure if this is the right way to handle the `HailTaskContext` setting and unsetting on the `LocalBackend`. Does `LocalBackend` even need a TaskContext analogue to hold a `RegionPool`? Or should we just be doing like `Backend.get().getRegionPool()` and the local one just holds onto one whereas the Spark one checks the task context?

tpoterba: > Does LocalBackend even need a TaskContext analogue to hold a RegionPool?

I think using similar patterns in all the backends will make it much easier for us to think about what the code is doing. I can certainly imagine that we'll want to use the local backend to interrogate per-task memory usage, for instance, with which the task context will help.

hail/src/main/scala/is/hail/annotations/Region.scala
```diff
@@ -269,16 +265,17 @@ object Region {
...
-    Code.invokeScalaObject2[Int, RegionPool, Region](Region.getClass, "apply", asm4s.const(blockSize), Code._null)
+  def stagedCreate(blockSize: Size, pool: Code[RegionPool]): Code[Region] =
+    Code.invokeScalaObject2[Int, RegionPool, Region](Region.getClass, "apply", asm4s.const(blockSize), pool)
 
   def apply(blockSize: Region.Size = Region.REGULAR, pool: RegionPool = null): Region = {
-    (if (pool == null) RegionPool.get else pool)
-      .getRegion(blockSize)
-  }
-
-  def makeNamed(blockSize: Region.Size = Region.REGULAR, pool: RegionPool = null): Region = {
```
johnc1231, line 280: makeNamed was a clone of `apply` that should never have been commited, was part of me debugging the region management overhaul from earlier this year.

hail/src/main/scala/is/hail/backend/local/LocalBackend.scala
```diff
@@ -28,6 +28,14 @@ import scala.reflect.ClassTag
 
 class LocalBroadcastValue[T](val value: T) extends BroadcastValue[T] with Serializable
 
+class LocalTaskContext(val partitionId: Int) extends HailTaskContext {
+  override type BackendType = LocalBackend
+
+  override def stageId(): Int = ???
+
+  override def attemptNumber(): Int = ???
```
johnc1231, line 36: Not sure what the implementation of these is outside of Spark

hail/src/main/scala/is/hail/annotations/Region.scala
```diff
@@ -269,16 +265,17 @@ object Region {
...
-    (if (pool == null) RegionPool.get else pool)
+    (if (pool == null) {
+      val htc = HailTaskContext.get()
+      if (htc == null) {
+        throw new IllegalStateException(s"RegionPool requested but HailTaskContext was null. On worker = ${TaskContext.get() != null}")
+      }
+      htc.getRegionPool()
+    } else pool)
```
johnc1231, line 278: We maybe just shouldn't have a default `Region` with null pool constructor anymore. Seems better to just explicitly make someone grab a pool. Otherwise too easy to do the wrong thing.

johnc1231:
Seeing sporadic failures, random each run of ci, looking into it.

hail/src/main/scala/is/hail/annotations/Region.scala
tpoterba: I agree.

hail/src/main/scala/is/hail/backend/local/LocalBackend.scala
tpoterba: they can be 0 and 0, due to our design of the local backend

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
```diff
@@ -529,6 +532,17 @@ class EmitClassBuilder[C](
     _backendField
   }
 
+  def pool(): Value[RegionPool] = {
+    if (_poolField == null) {
+      cb.addInterface(typeInfo[FunctionWithPool].iname)
```
tpoterba, line 537: I think you don't need the complexity of the additional interface here. FunctionWithPartitionRegion should add the pool field, I think.

hail/src/main/scala/is/hail/rvd/RVD.scala
```diff
@@ -46,6 +46,8 @@ class RVD(
 
   require(typ.kType.virtualType isIsomorphicTo partitioner.kType)
 
+  val myOrigin = Thread.currentThread().getStackTrace.mkString("\n")
```
tpoterba, line 49: debug info

tpoterba (changes requested):
initial comments, I intend to do another look-over this afternoon, but this looks pretty straightforward.

hail/src/main/scala/is/hail/backend/HailTaskContext.scala
```diff
@@ -1,11 +1,33 @@
...
+
 object HailTaskContext {
   def get(): HailTaskContext = taskContext.get
 
-  private[this] val taskContext: ThreadLocal[HailTaskContext] = new ThreadLocal[HailTaskContext]
-  protected[backend] def setTaskContext(tc: HailTaskContext): Unit = taskContext.set(tc)
-  protected[backend] def unset(): Unit = taskContext.remove()
+  private[this] val taskContext: ThreadLocal[HailTaskContext] = new ThreadLocal[HailTaskContext]() {
+    override def initialValue(): HailTaskContext = {
+      val backend = HailContext.get.backend
```
johnc1231, line 14: It occurs to me that this is going to be a problem. Can't decide how to construct based on the backend, because you can't ask for the `HailContext` on a worker. Not sure what the best way to do this.

tpoterba: oh, I think just assume it's spark backend. This can just be the lines 16-18 I think -- the other backends will always be setting the task context in parallelizeAndComputeWithIndex

johnc1231: How does this work in the `ServiceBackend`? I'm not sure where the task context will be set.

johnc1231:
Tests are passing now. I am going to mark WIP because I want to run benchmarks on Monday before this merges. You can look whenever though. 

Some things to look at:

1. Does `ServiceTaskContext` have a notion of stages or retries, or should those also always be 0? 
2. Some of the logic with combOps getting region pools is a little different, now depends on where combOp is running 

For the most part though, PR is just refactoring to pass `RegionPool` everywhere

johnc1231:
Here are the benchmarks: https://gist.github.com/johnc1231/3e576bf2e8a39cb73785af0faa451976

@tpoterba  should be good for review.

tpoterba (dismissed):

tpoterba (approved):

PR #9565, opened 2020-10-06
[docker] retry apt download at most 3 times

danking:

PR #9681, opened 2020-11-05
[echo] implement a TCP echo server

danking:
I will test this with a forthcoming PR which adds a TCP gateway.

### DIFF
--- a/echo/Dockerfile
+++ b/echo/Dockerfile
@@ -1,0 +1,3 @@
+FROM gcr.io/{{ global.project }}/ubuntu:18.04
+RUN apt-get update && apt-get install -y socat
+CMD ["socat", "TCP4-LISTEN:5000,fork", "EXEC:cat"]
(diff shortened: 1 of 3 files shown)


echo/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM gcr.io/{{ global.project }}/ubuntu:18.04
```
jigold, line 1: Depending on whether #9565 goes in first, we'll need to fix this.

jigold (approved):

jigold:
Does this need to have the service with the router like all the other services?

jigold (changes requested):
see comment

danking:
No, it's not an HTTP service, it's plain old TCP, so `router` (which is an HTTP router) won't work here. The TCP gateway / router will come next.

jigold (approved):

PR #9682, opened 2020-11-05
[auth][gear] functions to work with session ids

danking:
We now have two functions which convert back and forth between the two
representations of session ids: `bytearray` and `str`.

### DIFF
--- a/gear/gear/auth_utils.py
+++ b/gear/gear/auth_utils.py
@@ -1,5 +1,6 @@
 import secrets
 import base64
+from hailtop.auth session_id_encode_to_str
 
 
 async def insert_user(db, spec):
@@ -15,7 +16,7 @@ VALUES ({', '.join([f'%({k})s' for k in spec.keys()])})
 
 # 2592000s = 30d
 async def create_session(db, user_id, max_age_secs=2592000):
-    session_id = base64.urlsafe_b64encode(secrets.token_bytes(32)).decode('ascii')
+    session_id = session_id_encode_to_str(secrets.token_bytes(32))
     await db.just_execute('INSERT INTO sessions (session_id, user_id, max_age_secs) VALUES (%s, %s, %s);',
                           (session_id, user_id, max_age_secs))
     return session_id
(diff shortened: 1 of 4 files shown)


gear/gear/auth_utils.py
```diff
@@ -1,5 +1,6 @@
 import secrets
 import base64
+from hailtop.auth session_id_encode_to_str
```
Dania-Abuhijleh, line 3: incorrect syntax?

Dania-Abuhijleh (changes requested):

gear/gear/auth_utils.py
danking: whoops

hail/python/hailtop/auth/tokens.py
```diff
@@ -8,6 +9,14 @@
 log = logging.getLogger('gear')
 
 
+def session_id_encode_to_str(session_id_bytes):
+    return base64.urlsafe_b64encode(session_id_bytes).decode('ascii')
+
+
+def session_id_decode_from_str(session_id_str):
+    return base64.urlsafe_b64decode(session_id_str.encode('ascii'))
```
Dania-Abuhijleh, line 17: Should this be base64.urlsafe_b64decode(session_id_str).encode('ascii')?

Dania-Abuhijleh (changes requested):

hail/python/hailtop/auth/tokens.py
danking: I added types that should make this a bit more clear. The input `session_id_str` is a Python `str`. I "encode" that string as a Python `bytearray` using the ASCII encoding. Then I send that into the base64 decoder which converts a bytearray of base64 encoded ASCII into a byte array of regular old ASCII.


Step-wise:
`encode_to_str`
- bytes (`bytearray`)
- base64-encoded bytes (`bytearray`)
- base64-encoded ASCII string (`str`)

`decode_from_str`
- base64-encoded ASCII string (`str`)
- base64-encoded bytes (`bytearray`)
- bytes (`bytearray`)

It's maybe more clear with intermediate variables:
```python3
    session_id_base64_bytes = session_id_str.encode('ascii')
    session_id_just_bytes = base64.urlsafe_b64decode(session_id)
    return session_id_just_bytes
```

Dania-Abuhijleh (approved):

PR #9683, opened 2020-11-05
[hail] fix bugs in InputBuffers and RichInputStream

danking:
The documentation for [`InputStream.read(byte[], int, int)`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read-byte:A-int-int-) explicitly notes:
> Reads up to len bytes of data from the input stream into an array of
> bytes. An attempt is made to read as many as len bytes, but a smaller
> number may be read.

Therefore, our `InputBuffer` methods must not assume `read` returns the requested number of bytes. Moreover, `readFully` should not assume that a read of zero bytes indicates end of stream. End of stream is only indicated by a `-1`.


danking:
These were uncovered while using the TCP Proxy to enter the cluster. Forthcoming Shuffler tests will add tests that would catch this behavior.

chrisvittal (approved):

PR #9689, opened 2020-11-06
[Benchmark] test update_commits

Dania-Abuhijleh:
tested 3 cases, when a commit has a results file, doesn't have a results file but has a running batch, or no file and no running batch.


benchmark-service/benchmark/test/test_update_commits.py
```diff
@@ -0,0 +1,42 @@
...
+    headers = service_auth_headers(deploy_config, 'benchmark')
+    create_benchmark_url = deploy_config.url('benchmark', '/api/v1alpha/benchmark/update_commit')
+
+    async with get_context_specific_ssl_client_session(
+            raise_for_status=True,
+            timeout=aiohttp.ClientTimeout(total=60)) as session:
+
+        async def wait_forever():
+            commit_info = None
+            while commit_info is None:
```
jigold, line 31: Once you have the three operations done, then you'll need to change this a bit.

jigold (dismissed):
I don't think your test is sufficient, but it's a good start. You're testing nearly identical code to the polling loop, but not the actual code in the loop.

I think you want update_commits to call the code inside update_commit. Update_commit should return the commit that is appended to the data structure.

Then you can have 3 new methods: POST, GET, and DELETE.

DELETE will remove the file from google storage and remove the commit from benchmark_data['commits']
GET will get the status from benchmark_data['commits'] if it exists. This is how you'll know if the batch is running or completed
POST will call update_commit with the new SHA.

If this doesn't make sense, then we can discuss over Zoom.

jigold:
Oh. This isn't actually running at all. I think you need to add a new test_benchmark step in build.yaml etc.

benchmark-service/benchmark/benchmark.py
```diff
@@ -272,6 +275,81 @@ async def update_commits(app):
...
+    global benchmark_data
+    app = request.app
+    gs_reader = app['gs_reader']
+    sha = str(request.match_info['sha'])
+    file_path = f'{BENCHMARK_RESULTS_PATH}/{sha}.json'
+    if gs_reader.file_exists(file_path):
+        gs_reader.delete_file(file_path)
+    commit = next((item for item in benchmark_data['commits'] if item['sha'] == sha), None)
+    if commit is not None:
+        benchmark_data['commits'].remove(commit)
```
jigold, line 342: Can you not just do something like this?

```
if sha in benchmark_data['commits']:
    del benchmark_data['commits][sha]
```

benchmark-service/benchmark/benchmark.py
```diff
@@ -272,6 +275,81 @@ async def update_commits(app):
...
+
+    if has_results_file:
+        assert complete_batch_statuses, batch_statuses
+        log.info(f'commit {sha} has a results file')
+        status = complete_batch_statuses[-1]
+        case = 'has_results_file'
+    elif running_batch_statuses:
+        status = running_batch_statuses[-1]
+        log.info(f'batch already exists for commit {sha}')
+        case = 'running_batch'
```
jigold, line 302: I don't think you need the case here.

benchmark-service/benchmark/test/test_update_commits.py
```diff
@@ -0,0 +1,40 @@
...
+        async def wait_forever():
+            commit = None
+            while commit is None:
+                resp = await utils.request_retry_transient_errors(
+                    session, 'POST', f'{create_benchmark_url}', headers=headers, json={'sha': sha})
+                resp_text = await resp.text()
+                commit = json.loads(resp_text)
+            return commit
+
+        commit = await wait_forever()
```
jigold, line 38: What I was expecting was a delete, then a get asserting the commit status is None, followed by a POST to update the commit where it says the batch was submitted. Then wait forever asserting the batch is running until the GET returns a commit status that is complete / file exists.

jigold (changes requested):
There might be other fixes in the future, but this is getting there.

benchmark-service/benchmark/benchmark.py
```diff
@@ -223,47 +226,19 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
-        commit = {
-            'sha': sha,
-            'title': gh_commit['commit']['message'],
-            'author': gh_commit['commit']['author']['name'],
-            'date': gh_commit['commit']['author']['date'],
-            'status': status
-        }
-        formatted_new_commits.append(commit)
+        commit = await update_commit(app, sha)
+        formatted_new_commits.update({sha: commit})
```
jigold, line 241: You can just do this:

```
formatted_new_commits[sha] = commit
```

benchmark-service/benchmark/benchmark.py
```diff
@@ -272,6 +247,93 @@ async def update_commits(app):
...
+    global benchmark_data
+    sha = str(request.match_info['sha'])
+
+    app = request.app
+    batch_client = app['batch_client']
+    batch_statuses = [b._last_known_status async for b in batch_client.list_batches(q=f'sha={sha}')]
+
+    try:
+        status = batch_statuses[-1]
+        log.info('got commit status')
```
jigold, line 300: I was thinking that this should return the same thing as the full commit and not just the status. I'm also not sure you need to wrap the status in `{'status': status}` and could just return the commit directly.

```
commit = {
         'sha': sha,
         'title': gh_commit['commit']['message'],
         'author': gh_commit['commit']['author']['name'],
         'date': gh_commit['commit']['author']['date'],
         'status': status
     }
```

To do this, you might need to split `update_commit` into two functions one of which is `get_commit`. This should look like this:

```python3
async def get_commit(app, sha):  # pylint: disable=unused-argument
     github_client = app['github_client']
     batch_client = app['batch_client']
     gs_reader = app['gs_reader']

      file_path = f'{BENCHMARK_RESULTS_PATH}/{sha}.json'

      request_string = f'/repos/hail-is/hail/commits/{sha}'

      gh_commit = await github_client.getitem(request_string)

    has_results_file = gs_reader.file_exists(file_path)
     batch_statuses = [b._last_known_status async for b in batch_client.list_batches(q=f'sha={sha}')]
     complete_batch_statuses = [bs for bs in batch_statuses if bs['complete']]
     running_batch_statuses = [bs for bs in batch_statuses if not bs['complete']]

      if has_results_file:
         assert complete_batch_statuses, batch_statuses
         log.info(f'commit {sha} has a results file')
         status = complete_batch_statuses[-1]
     elif running_batch_statuses:
         status = running_batch_statuses[-1]
         log.info(f'batch already exists for commit {sha}')
     else:
         status = None
         log.info(f'no batches or results file exists for {sha}')

      commit = {
         'sha': sha,
         'title': gh_commit['commit']['message'],
         'author': gh_commit['commit']['author']['name'],
         'date': gh_commit['commit']['author']['date'],
         'status': status
     }
     return commit
```

```python3
async def update_commit(app, sha):  # pylint: disable=unused-argument
     commit = await get_commit(app, sha)
     if commit['status'] is not None:
           return commit

     batch_client = app['batch_client']

     batch_id = await submit_test_batch(batch_client, sha)
     batch = await batch_client.get_batch(batch_id)
     commit['status'] = batch._last_known_status
     log.info(f'submitted a batch {batch_id} for commit {sha}')

     return commit
```

benchmark-service/benchmark/benchmark.py
```diff
@@ -272,6 +247,93 @@ async def update_commits(app):
...
+async def update_commit(app, sha):  # pylint: disable=unused-argument
+    github_client = app['github_client']
+    batch_client = app['batch_client']
+    gs_reader = app['gs_reader']
+
+    file_path = f'{BENCHMARK_RESULTS_PATH}/{sha}.json'
+
+    request_string = f'/repos/hail-is/hail/commits/{sha}'
+
+    gh_commit = await github_client.getitem(request_string)
```
jigold, line 259: What happens if the commit doesn't exist? In a future PR, can we fix this? I have some ideas on how this should be done.

benchmark-service/test/test_update_commits.py
```diff
@@ -0,0 +1,57 @@
...
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger(__name__)
+
+sha = 'd626f793ad700c45a878d192652a0378818bbd8b'
+
+
+async def test_update_commits():
+    deploy_config = get_deploy_config()
+    headers = service_auth_headers(deploy_config, 'benchmark')
+    create_benchmark_url = deploy_config.url('benchmark', f'/api/v1alpha/benchmark/commit/{sha}')
```
jigold, line 23: let's rename this to `commit_benchmark_url`

benchmark-service/test/test_update_commits.py
```diff
@@ -0,0 +1,57 @@
...
+            raise_for_status=True,
+            timeout=aiohttp.ClientTimeout(total=60)) as session:
+
+        await utils.request_retry_transient_errors(
+            session, 'DELETE', f'{create_benchmark_url}', headers=headers, json={'sha': sha})
+
+        resp_status = await utils.request_retry_transient_errors(
+            session, 'GET', f'{create_benchmark_url}', headers=headers, json={'sha': sha})
+        resp_status_text = await resp_status.text()
+        status = json.loads(resp_status_text)
```
jigold, line 35: Let's change status here to commit.

benchmark-service/test/test_update_commits.py
```diff
@@ -0,0 +1,57 @@
...
+        assert status['status'] is None, status
+
+        resp_commit = await utils.request_retry_transient_errors(
+            session, 'POST', f'{create_benchmark_url}', headers=headers, json={'sha': sha})
+        resp_commit_text = await resp_commit.text()
```
jigold, line 40: You can just use `await resp_commit.json()`. Then you don't need to use json.loads below. This applies to other places in this function.

benchmark-service/test/test_update_commits.py
```diff
@@ -0,0 +1,57 @@
...
+            while commit_status is None or not commit_status['status']['complete']:
+                resp = await utils.request_retry_transient_errors(
+                    session, 'GET', f'{create_benchmark_url}', headers=headers, json={'sha': sha})
+                resp_text = await resp.text()
+                commit_status = json.loads(resp_text)
+                await asyncio.sleep(5)
+                print(commit_status['status'])
+            return commit_status
+
+        # await wait_forever()
```
jigold, line 54: delete comments.

build.yaml
```diff
@@ -436,7 +436,7 @@ steps:
      fi
      # created ci
      # create test
-     N=$(kubectl -n {{ default_ns.name }} get secret --ignore-not-found=true --no-headers test-tokens | wc -l | tr -d '[:space:]')
+     N=$(kubectl -n {{ batch_pods_ns.name }} get secret --ignore-not-found=true --no-headers test-tokens | wc -l | tr -d '[:space:]')
```
jigold, line 439: Can you get rid of these two changes with `batch_pods_ns.name`. It won't block your PR and I wanted Dan to look at it first in my PR.

docker/Makefile
```diff
@@ -21,7 +21,7 @@ GENETICS_PUBLIC_IMAGE = gcr.io/$(PROJECT)/genetics-public:$(shell docker images
 .PHONY: hail-ubuntu
 hail-ubuntu: hail-ubuntu-stmp
 
-hail-ubuntu-stmp: hail-ubuntu/Dockerfile hail-ubuntu/hail_pip_install hail-ubuntu/pip.conf hail-ubuntu/hail-apt-get-install
+hail-ubuntu-stmp: hail-ubuntu/Dockerfile hail-ubuntu/hail-pip-install hail-ubuntu/pip.conf hail-ubuntu/hail-apt-get-install
```
jigold, line 24: I don't think this should be part of this PR. Should already be in main.

benchmark-service/benchmark/benchmark.py
```diff
@@ -34,7 +34,10 @@
 
 BENCHMARK_ROOT = os.path.dirname(os.path.abspath(__file__))
 
-benchmark_data = None
+benchmark_data = {
+    'commits': {}
+}
+shas = None
```
jigold, line 40: I don't think this is used at all.

benchmark-service/benchmark/benchmark.py
```diff
@@ -223,47 +226,19 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
-
-        commit = {
-            'sha': sha,
-            'title': gh_commit['commit']['message'],
-            'author': gh_commit['commit']['author']['name'],
-            'date': gh_commit['commit']['author']['date'],
-            'status': status
-        }
-        formatted_new_commits.append(commit)
+        commit = await update_commit(app, sha)
```
jigold, line 240: I'm debating whether update_commit should update `benchmark_data['commits']` directly. I don't have a strong preference, but I think it might make the code nicer.

jigold (dismissed):
Here's the high level comments. Great job getting this all working!

benchmark-service/benchmark/benchmark.py
```diff
@@ -223,53 +225,118 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
+
+
+@router.get('/api/v1alpha/benchmark/commit/{sha}')
+async def get_status(request):  # pylint: disable=unused-argument
+    sha = str(request.match_info['sha'])
+    app = request.app
+    commit = await get_commit(app, sha)
+    return web.json_response({'commit': commit})
+
+    # batch_client = app['batch_client']
```
jigold, line 302: Delete

benchmark-service/benchmark/benchmark.py
```diff
@@ -223,53 +225,118 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
+    return web.Response()
+
+
+@router.post('/api/v1alpha/benchmark/commit/{sha}')
+async def call_update_commit(request):  # pylint: disable=unused-argument
+    body = await request.json()
+    sha = body['sha']
+    log.info('call_update_commit')
+    commit = await update_commit(request.app, sha)
+    return web.json_response({'commit': commit})
```
jigold, line 339: Why does this need to be wrapped like this? Why not just `web.json_response(commit)`?

benchmark-service/test/test_update_commits.py
```diff
@@ -0,0 +1,52 @@
...
+            session, 'GET', f'{commit_benchmark_url}', headers=headers, json={'sha': sha})
+        commit = await resp_status.json()
+        assert commit['commit']['status'] is None, commit
+
+        resp_commit = await utils.request_retry_transient_errors(
+            session, 'POST', f'{commit_benchmark_url}', headers=headers, json={'sha': sha})
+        commit = await resp_commit.json()
+
+        async def wait_forever():
+            commit_status = None
```
jigold, line 42: Now this part is confusing. I don't think you need `commit_status = None`. Rename commit_status to commit.

docker/Makefile
```diff
@@ -21,7 +21,7 @@ GENETICS_PUBLIC_IMAGE = gcr.io/$(PROJECT)/genetics-public:$(shell docker images
 .PHONY: hail-ubuntu
 hail-ubuntu: hail-ubuntu-stmp
 
-hail-ubuntu-stmp: hail-ubuntu/Dockerfile hail-ubuntu/hail-pip-install hail-ubuntu/pip.conf hail-ubuntu/hail-apt-get-install
+hail-ubuntu-stmp: hail-ubuntu/Dockerfile hail-ubuntu/hail_pip_install hail-ubuntu/pip.conf hail-ubuntu/hail-apt-get-install
```
jigold, line 24: You need to rebase your work. This change shouldn't be here at all.

jigold (changes requested):

docker/Makefile
Dania-Abuhijleh: I'm rebasing after pulling from upstream main but nothing is changing

jigold (changes requested):
almost done.

jigold:
You're going to need to rebase on main again. Cotton changed some things tonight especially with create_accounts.

build.yaml
```diff
@@ -2065,6 +2077,35 @@ steps:
...
+       namespace:
+         valueFrom: default_ns.name
+       mountPath: /ssl-config
+   timeout: 300
+   dependsOn:
+     - test_benchmark_image
+     - create_deploy_config
+     - create_accounts
+     - default_ns
+     - batch_pods_ns
```
jigold, line 2106: Delete batch_pods_ns. It's no longer used.

build.yaml
```diff
@@ -2065,6 +2077,35 @@ steps:
...
+     set -ex
+     python3 -m pytest --log-cli-level=INFO -s -vv --instafail --durations=50 /test/
+   secrets:
+     - name: gce-deploy-config
+       namespace:
+         valueFrom: default_ns.name
+       mountPath: /deploy-config
+     - name: test-dev-tokens
+       namespace:
+         valueFrom: batch_pods_ns.name
```
jigold, line 2094: This should be default_ns now.

build.yaml
```diff
@@ -416,6 +416,7 @@ steps:
      if [[ $N = 0 ]]; then
          # if benchmark does not exist
          rm -f /user-tokens/tokens.json
+         kubectl -n {{ default_ns.name }} get -o json secret test-gsa-key | jq '{apiVersion, kind, type, data, metadata: {name: "benchmark-gsa-key"}}' | kubectl -n {{ batch_pods_ns.name }} apply -f -
```
jigold, line 419: We don't need this line anymore. Cotton got rid of batch_pods.

benchmark-service/test/test_update_commits.py
```diff
@@ -0,0 +1,55 @@
...
+async def test_update_commits():
+    deploy_config = get_deploy_config()
+    headers = service_auth_headers(deploy_config, 'benchmark')
+    commit_benchmark_url = deploy_config.url('benchmark', f'/api/v1alpha/benchmark/commit/{sha}')
+
+    async with get_context_specific_ssl_client_session(
+            raise_for_status=True,
+            timeout=aiohttp.ClientTimeout(total=60)) as session:
+
+        global commit
```
jigold, line 30: Let's define commit = None here instead of above. Then you don't need this to be global here.

benchmark-service/benchmark/benchmark.py
```diff
@@ -223,53 +225,108 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
+@router.delete('/api/v1alpha/benchmark/commit/{sha}')
+async def delete_commit(request):  # pylint: disable=unused-argument
+    global benchmark_data
+    app = request.app
+    gs_reader = app['gs_reader']
+    batch_client = app['batch_client']
+    sha = str(request.match_info['sha'])
+    file_path = f'{BENCHMARK_RESULTS_PATH}/{sha}.json'
+    if gs_reader.file_exists(file_path):
+        gs_reader.delete_file(file_path)
```
jigold, line 312: nit. Can we add some new lines between the if statements and the async block.

benchmark-service/benchmark/benchmark.py
```diff
@@ -223,53 +225,108 @@ async def compare(request, userdata):  # pylint: disable=unused-argument
...
-    formatted_new_commits = []
 
     for gh_commit in gh_data:
         sha = gh_commit.get('sha')
         log.info(f'for commit {sha}')
-        file_path = f'{BENCHMARK_RESULTS_PATH}/{sha}.json'
-        has_results_file = gs_reader.file_exists(file_path)
 
-        batch_statuses = [b._last_known_status async for b in batch_client.list_batches(q=f'sha={sha}')]
+        await update_commit(app, sha)
```
jigold, line 238: Let's get rid of the extra new line above this line.

jigold (changes requested):
Some minor things. Once this goes in, I want you to watch for it to be redeployed by following the deploy batch at ci.hail.is. Once it's redeployed check the default namespace benchmark logs to make sure it's still working as intended.

jigold:
You might need to rebase to incorporate cotton's changes.

benchmark-service/test/test_update_commits.py
Dania-Abuhijleh: but then it says commit is referenced before defining in the while loop

Dania-Abuhijleh: ah unless i use 'nonlocal'

jigold (dismissed):

jigold (approved):

PR #9693, opened 2020-11-10
[datasets] add gnomad public datasets

pwc2:
Updated the paths in `datasets.json` for `gnomad_exome_sites` and `gnomad_genome_sites` to use the `gs://gcp-public-data--gnomad` bucket. 

Added the Hail Table for version 3.1 of genome sites to `datasets.json`, as well as the gnomad HGDP + 1KG callset MatrixTable.

### DIFF
--- a/hail/python/hail/docs/datasets/schemas/gnomad_hgdp_1kg_callset.rst
+++ b/hail/python/hail/docs/datasets/schemas/gnomad_hgdp_1kg_callset.rst
@@ -1,0 +1,651 @@
+.. _gnomad_hgdp_1kg_callset:
+
+gnomad_hgdp_1kg_callset
+=======================
+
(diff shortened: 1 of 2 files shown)


konradjk:
Are there tests for the existence of these files? (I can see how that would be counterproductive if tests fail because the files move, but on the other hand, it's information you may want to know - maybe an "optional" test?)

pwc2:
> Are there tests for the existence of these files? (I can see how that would be counterproductive if tests fail because the files move, but on the other hand, it's information you may want to know - maybe an "optional" test?)

Not currently, that may be something worth looking into.

danking (approved):

danking:
Yeah, we really need an alerting system.

PR #9708, opened 2020-11-17
[query Start undoing value misuse related to ndarrays

tpoterba:
I detect no performance difference on blanczos running the benchmark
locally. This pattern appears a lot in the NDArrayEmitter, though,
so we should fix it everywhere and see what happens!


johnc1231 (approved):
Ha, I was about to assign the same PR to you, beat me by a minute. I also detected no change from benchmarks

PR #9711, opened 2020-11-17
[amundsen] Add front end service

jigold:
This works in the sense that I didn't get a 404 or 502 contacting the server in my namespace. I think we need to redeploy lets encrypt -- can't remember.


jigold:
cc: @pwc2

danking (approved):
Thank you so much Jackie@

PR #9713, opened 2020-11-17
Add infra/ using Terraform.  Remove vdc/.

cseed:
Right now this spins up K8s, a database, some of the networking stuff,
and creates a default/global-config secret that includes the
information stored in $HAIL/config.mk.  This isn't used yet and will
probably change.


infra/main.tf
```diff
@@ -0,0 +1,190 @@
...
+    ip_configuration {
+      ipv4_enabled    = false
+      private_network = google_compute_network.internal.id
+      require_ssl = true
+    }
+  }
+}
+
+# FIXME rename this to gateway
+resource "google_compute_address" "gateway" {
```
lgruen, line 151: Should this be using `google_compute_global_address`?

infra/README.md
```diff
@@ -0,0 +1,31 @@
+This is a work in progress to use Terraform to manage our cloud
+infrastructure.
+
+Instructions:
+
+- You will need a GCP project.  Create a service account for Terraform
```
lgruen, line 6: Might be worth mentioning the required APIs that should be enabled, otherwise it'll take a few Terraform runs. E.g. https://console.developers.google.com/marketplace/details/google-cloud-platform/compute-engine, https://console.developers.google.com/apis/library/cloudresourcemanager.googleapis.com.

infra/README.md
```diff
@@ -0,0 +1,31 @@
...
+  with Editor role, and create a service account key and place it in
```
lgruen, line 7: The service account will also need the Service Networking Admin role, otherwise the google_service_networking_connection.private_vpc_connection creation will fail.

infra/main.tf
```diff
@@ -0,0 +1,190 @@
...
+      disable-legacy-endpoints = "true"
+    }
+
+    oauth_scopes = [
+      "https://www.googleapis.com/auth/cloud-platform"
+    ]
+  }
+}
+
+resource "google_compute_global_address" "db_ip_address" {
```
lgruen, line 112: Should this be using `google_compute_address`?

infra/README.md
cseed: Did you have to enable servicenetworking?  I had a resource for that:

```
resource "google_project_service" "service_networking" {
  service = "servicenetworking.googleapis.com"
}
```

and I'm wondering if it didn't work.

lgruen: Yes, here's the error message I got:

```
google_project_service.service_networking: Creating...
google_compute_global_address.gateway: Creation complete after 15s [id=projects/<redacted>/global/addresses/gateway]
kubernetes_secret.global_config: Modifying... [id=default/global-config]
kubernetes_secret.global_config: Modifications complete after 1s [id=default/global-config]
google_project_service.service_networking: Still creating... [10s elapsed]
google_project_service.service_networking: Still creating... [20s elapsed]
google_project_service.service_networking: Creation complete after 28s [id=<redacted>/servicenetworking.googleapis.com]

Error: googleapi: Error 403: Service Networking API has not been used in project <redacted> before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/servicenetworking.googleapis.com/overview?project=<redacted> then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry., accessNotConfigured
```

After enabling the API, I still got a 403, but that was because of missing permissions for the service account:

```
google_service_networking_connection.private_vpc_connection: Creating...

Error: googleapi: Error 403: The caller does not have permission, forbidden
```

cseed: Ah, looks like it needs an explicit dependency.  That's annoying and race condition-y.  I'll just put the gcloud enable in the README.  Thanks!

lgruen: https://github.com/terraform-google-modules/terraform-google-project-factory/tree/master/modules/project_services#project-api-activation might be worth a try.

infra/main.tf
```diff
@@ -0,0 +1,190 @@
...
+  location = var.gcp_zone
+  network = google_compute_network.internal.name
+
+  # We can't create a cluster with no node pool defined, but we want to only use
+  # separately managed node pools. So we create the smallest possible default
+  # node pool and immediately delete it.
+  remove_default_node_pool = true
+  initial_node_count       = 1
+
+  master_auth {
```
danking, line 54: this block appears optional, is it necessary? 

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster

infra/main.tf
```diff
@@ -0,0 +1,190 @@
...
+    ]
+  }
+}
+
+resource "google_compute_global_address" "db_ip_address" {
+  name          = "db-ip-address"
+  purpose       = "VPC_PEERING"
+  address_type  = "INTERNAL"
+  prefix_length = 16
+  network = google_compute_network.internal.id
```
danking, line 117: nit: can we either align all `=`s or not align all `=`s?

infra/main.tf
```diff
@@ -0,0 +1,190 @@
...
+      "https://www.googleapis.com/auth/cloud-platform"
+    ]
+  }
+}
+
+resource "google_compute_global_address" "db_ip_address" {
+  name          = "db-ip-address"
+  purpose       = "VPC_PEERING"
+  address_type  = "INTERNAL"
+  prefix_length = 16
```
danking, line 116: I don't understand this bit, isn't the db IP address a single address?

Is this telling us something about the network?

infra/main.tf
```diff
@@ -0,0 +1,190 @@
...
+      ipv4_enabled    = false
+      private_network = google_compute_network.internal.id
+      require_ssl = true
+    }
+  }
+}
+
+# FIXME rename this to gateway
+resource "google_compute_address" "gateway" {
+  name = "gateway"
```
danking, line 152: seems odd to be explicit about region in internal_gateway but implicit here

danking (changes requested):
Leo also has a couple outstanding comments, can you mark those resolved if you think you've addressed them?

infra/main.tf
```diff
@@ -0,0 +1,508 @@
...
+  }
+}
+
+resource "kubernetes_namespace" "ukbb_rg" {
+  metadata {
+    name = "ukbb-rg"
+  }
+}
+
+resource "kubernetes_service" "ukbb_rb_browser" {
```
lgruen, line 313: I think this service might need a `depends_on = [kubernetes_namespace.ukbb_rg]` to avoid a race.

infra/main.tf
```diff
@@ -0,0 +1,508 @@
...
+      protocol = "TCP"
+      target_port = 80
+    }
+    selector = {
+      app = "ukbb-rg-browser"
+    }
+  }
+}
+
+resource "kubernetes_service" "ukbb_rb_static" {
```
lgruen, line 333: Ditto.

infra/README.md
```diff
@@ -0,0 +1,113 @@
+This is a work in progress to use Terraform to manage our cloud
+infrastructure.
+
+Instructions:
+
+- You will need a GCP project.  We assume `gcloud` is configured to
+  point at your project.
+
```
lgruen, line 8:
```suggestion
gcloud config set project <project-id>
gcloud config set compute/zone <compute-zone>
```

infra/README.md
```diff
@@ -0,0 +1,113 @@
...
+   gcp_zone = "<gcp-zone>"
+   domain = "<domain>"
+   ```
+
+- Run `terraform init`.
+
+- Run `terraform apply -var-file="$HOME/.hail/global.tfvars"`.
+  Terraform has created a GKE cluster named `vdc`.  We assume
+  `kubectl` is configured to point at this cluster.
+
```
lgruen, line 41:
```suggestion
gcloud container clusters get-credentials vdc
```

infra/README.md
```diff
@@ -0,0 +1,113 @@
...
+- Go to the Google Cloud console, VPC networks > internal > Private
```
lgruen, line 42: I think you might be able to express this in Terraform using `google_compute_network_peering` (https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_network_peering#export_custom_routes), but I haven't looked into it in detail.

infra/README.md
```diff
@@ -0,0 +1,113 @@
...
+  service connection > Private connections to services, and enable
+  Export custom routes to both connections.
+
+You can now install Hail:
+
+- Run `$HAIL/docker/third-party/copy_images.sh`.  This copies some
```
lgruen, line 48: This step also requires setting the `PROJECT` environment variable beforehand and running `gcloud auth configure-docker gcr.io`.

I felt this was slightly faster running this from a VM, so maybe it would be good to combine this step with the VM setup further below.

infra/README.md
```diff
@@ -0,0 +1,113 @@
...
+  base images from Dockerhub (which now has rate limits) to GCR.
+
+- Generate TLS certificates.  See ../dev-docs/tls-cookbook.md.
+
+- Update $HAIL/config.mk with your infrastructure settings.  You can
+  get settings from the default/global-config secret.
+
```
lgruen, line 55:
```suggestion
kubectl get secret global-config -o json | jq '.data | map_values(@base64d)'
```

infra/README.md
```diff
@@ -0,0 +1,113 @@
...
+- Run `kubectl -n default apply -f $HAIL/ci/bootstrap.yaml`.
```
lgruen, line 56: Currently, `KUBERNETES_SERVER_URL` is hardcoded in ci/bootstrap.py, so probably should also be updated?

infra/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+- Build the CI utils image.  Run `make push-ci-utils` in $HAIL/ci.
```
lgruen, line 58: This failed without running `make service-base` in $HAIL/docker first.

infra/main.tf
lgruen: Understanding this a bit better now: Services (loadBalancerIP) can only use regional IPs. To use a global IP, an Ingress would be required. But there's probably not much benefit in using an Ingress for global load balancing at this point.

infra/README.md
cseed: I looked into this.  It just expands into a set of "google_project_service" resources like I had originally, but doesn't appear to address the dependency issue.

infra/README.md
cseed: I fixed this by making the terraform service account Owner instead of Editor.

infra/main.tf
cseed: So I couldn't get that to work.  The documentation wasn't super explicit about this, but as far as I can tell, the Could SQL database wants to be on the global rather than regional network.

infra/main.tf
cseed: I think it is necessary.  From the linked page:

 > If this block is provided and both username and password are empty, basic authentication will be disabled. This block also contains several computed attributes, documented below. If this block is not provided, GKE will generate a password for you with the username admin.

I don't want basic auth.

infra/main.tf
cseed: I changed it to not align.  Aligned is way to tedious to edit, but all the Terrafrom examples in the docs are aligned.

infra/main.tf
cseed: It is a block of IP addressed used by Google managed services (like CouldSQL).  I changed the name to "google-managed-services-internal".

infra/main.tf
cseed: I now set the region here.  Currently all our IP addresses are regional.

infra/main.tf
cseed: Thanks!  I modified these entries to refer to the namespace resource to get the name, which should introduce the proper dependency.

infra/README.md
cseed: Yeah, that might work.  I tried it but couldn't get it to work (probably user error).  The docs are written to peer two user-created networks, but we want to peer our internal network with the Google managed Cloud SQL network.  I imagine we'll want to make a pass over all of this to make it more automated.  Let's revisit this then.

infra/README.md
cseed: > This failed without running make service-base in $HAIL/docker first.

There should be dependencies for this, I don't see  what's wrong.  push-ci-utils depends on build-ci-utils, and:

```
build-ci-utils:
	$(MAKE) -C ../docker build
	...
```

and in docker/Makefile:

```
build: base-stmp service-base hail-ubuntu
```

> When running this on a Ubuntu VM, I also encountered an odd issue in that Makefile, when stat is called: [ "$(shell bash stat-permissions.sh Dockerfile.base)" = "644" ]. The mode of those files was 664 for me, so I had to run a chmod g-w to get the build working.

Yeah, this is a check for docker caching.  Docker includes files permissions in the layer hash, so unless the permissions agree everywhere, you get painful docker rebuilds.  I added a not about seeing the umask on the VM.  I added a step to set the umask on the VM.

infra/README.md
cseed: I know grab it from the env.  Thanks!

infra/main.tf
cseed: Yeah, agree, I don't think there is benefit until we're running globally distributed infra.  Let's revisit when we're approaching world domination, haha.

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+- Generate TLS certificates.  See ../dev-docs/tls-cookbook.md.
+
+- Run `kubectl -n default apply -f $HAIL/ci/bootstrap.yaml`.
+
+- Build the CI utils image.  Run `make push-ci-utils` in $HAIL/ci.
+
+- Deploy the bootstrap gateway.  Run `make deploy` in
+  $HAIL/bootstrap-gateway.
+
+- Create Let's Encrypt certs. Run `make run` in $HAIL/letsencrypt.
```
lgruen, line 97: This step required increasing `server_names_hash_bucket_size` for letsencrypt's nginx, similar to the gateway.

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+  images from Dockerhub (which now has rate limits) to GCR.
+
+- Generate TLS certificates.  See ../dev-docs/tls-cookbook.md.
+
+- Run `kubectl -n default apply -f $HAIL/ci/bootstrap.yaml`.
+
+- Build the CI utils image.  Run `make push-ci-utils` in $HAIL/ci.
+
+- Deploy the bootstrap gateway.  Run `make deploy` in
+  $HAIL/bootstrap-gateway.
```
lgruen, line 95: This gateway backends will crash-loop, as the router service hasn't been deployed yet:
```
nginx: [emerg] host not found in upstream "router" in /etc/nginx/conf.d/gateway.conf:73
```

That's not a big deal, as the bootstrapping later will take care of it, but maybe worth mentioning.

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+- Create Let's Encrypt certs. Run `make run` in $HAIL/letsencrypt.
+
+- Deploy the gateway.  Run `make deploy` in $HAIL/gateway.
```
lgruen, line 99: This gateway backends will crash-loop, as the router service hasn't been deployed yet:
```
nginx: [emerg] host not found in upstream "router" in /etc/nginx/conf.d/gateway.conf:73
```
That's not a big deal, as the bootstrapping later will take care of it, but maybe worth mentioning.

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+- Deploy the internal-gateway.  Run `make deploy` in $HAIL/internal-gateway.
+
+- Go to the Google Cloud console, API & Services, Credentials.
+  Configure the consent screen.  Add the scope:
```
lgruen, line 104:
```suggestion
  Configure the consent screen for internal users.  Add the scope:
```

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+  https://www.googleapis.com/auth/userinfo.email.  Create an OAuth
```
lgruen, line 105:
```suggestion
  https://www.googleapis.com/auth/userinfo.email.  Back in Credentials, create an OAuth
```

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+  ```
+
+- Install some dependencies on the VM:
+
+  ```
+  sudo apt update
+  sudo apt install -y docker.io python3-pip
+  sudo usermod -a -G docker $USER
+  gcloud -q auth configure-docker
+  gcloud container clusters get-credentials --zone us-central1-a vdc
```
lgruen, line 77:
```suggestion
  gcloud container clusters get-credentials --zone <gcp-zone> vdc
```

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+  git clone https://github.com/cseed/hail.git
+  python3 -m pip install -r $HOME/hail/docker/requirements.txt
+  ```
+
+  You will have to log out/in for usermod to take effect.
+
+- Run `PROJECT=<gc-project-id>
```
lgruen, line 84:
```suggestion
- Run `PROJECT=<gcp-project-id>
```

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+  HAIL_SHA=$(git rev-parse HEAD) \
+  HAIL_CI_UTILS_IMAGE=gcr.io/<gcp-project>/ci-utils:latest \
+  HAIL_CI_BUCKET_NAME=dummy \
+  KUBERNETES_SERVER_URL='<k8s-server-url>' \
+  HAIL_DEFAULT_NAMESPACE='default' \
+  HAIL_DOMAIN=<domain> \
+  HAIL_GCP_ZONE=<gcp-zone> \
+  GCP_PROJECT=<gcp-project> \
+  PYTHONPATH=$HOME/hail/ci:$HOME/hail/batch:$HOME/hail/hail/python \
+  python3 $HAIL/ci/bootstrap.py
```
lgruen, line 127:
```suggestion
  cd $HOME/hail
  export HAIL_SHA=$(git rev-parse HEAD) \
  export HAIL_CI_UTILS_IMAGE=gcr.io/<gcp-project-id>/ci-utils:latest \
  export HAIL_CI_BUCKET_NAME=dummy \
  export KUBERNETES_SERVER_URL='<k8s-server-url>' \
  export HAIL_DEFAULT_NAMESPACE='default' \
  export HAIL_DOMAIN=<domain> \
  export HAIL_GCP_ZONE=<gcp-zone> \
  export HAIL_GCP_PROJECT=<gcp-project-id> \
  export PYTHONPATH=$HOME/hail/ci:$HOME/hail/batch:$HOME/hail/hail/python \
  python3 ci/bootstrap.py
```

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+  ```
+
+- You may want to add a suitable ssh forward rule to the internal network.
```
lgruen, line 130: This should probably precede the previous point, as otherwise one won't be able to ssh into the VM to run the bootstrap command.

How about this?

Add an "allow-ssh" network tag to your VM and then create the following firewall rule to allow connections to your instance in the internal network:
```
gcloud compute firewall-rules create internal-allow-ssh --network internal --action=allow --target-tags=allow-ssh --rules=tcp:22
```

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+- Go to the Google Cloud console, VPC networks > internal > Private
+  service connection > Private connections to services, and enable
+  Export custom routes to both connections.
+
+You can now install Hail:
+
+- Create a VM on the internal network, 100GB, Ubuntu 20.04 TLS, allow
+  full access to all Cloud APIs.  10GB will run out of space.  We
+  assume the rest of the commands are run on the VM.
```
lgruen, line 57: This pretty much worked for everything except the $HAIL/letsencrypt step, due to a permissions error. I ended up running that outside the VM, but it probably wouldn't be too difficult to fix:

```
Error from server (Forbidden): error when creating "service.yaml": roles.rbac.authorization.k8s.io is forbidden: User "<redacted>" cannot create resource "roles" in API group "rbac.authorization.k8s.io" in the namespace "default": requires one of ["container.roles.create"] permission(s).
Error from server (Forbidden): error when creating "service.yaml": rolebindings.rbac.authorization.k8s.io is forbidden: User "<redacted>" cannot create resource "rolebindings" in API group "rbac.authorization.k8s.io" in the namespace "default": requires one of ["container.roleBindings.create"] permission(s).
make: *** [Makefile:30: run] Error 1
```

infra/README.md
lgruen, line 127: In order for create_accounts/main to succeed, I had to add a "ci-gsa-key" secret:
https://github.com/populationgenomics/hail/commit/46bbd7e29e7cf092a2d7566b9592df1b63690157

There's still an error at the very end of bootstrap.py, but it's probably just a harmless race?
```
ERROR:asyncio:Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f8895f01340>
ERROR:asyncio:Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x7f8895ecefa0>, 3555.901845417)]']
connector: <aiohttp.connector.TCPConnector object at 0x7f8895f016d0>
```

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+  HAIL_CI_UTILS_IMAGE=gcr.io/<gcp-project>/ci-utils:latest \
+  HAIL_CI_BUCKET_NAME=dummy \
+  KUBERNETES_SERVER_URL='<k8s-server-url>' \
+  HAIL_DEFAULT_NAMESPACE='default' \
+  HAIL_DOMAIN=<domain> \
+  HAIL_GCP_ZONE=<gcp-zone> \
+  GCP_PROJECT=<gcp-project> \
+  PYTHONPATH=$HOME/hail/ci:$HOME/hail/batch:$HOME/hail/hail/python \
+  python3 $HAIL/ci/bootstrap.py
+  ```
```
lgruen, line 128: The gateway kept redirecting to hail.is, as it's hardcoded in hail/python/hailtop/config/deploy_config.py (and a few other places):
https://github.com/populationgenomics/hail/commit/b1c3c3fa59b37bcffb0820e50008cd5c270c323f

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+- Deploy the gateway.  Run `make deploy` in $HAIL/gateway.
+
+- Deploy the internal-gateway.  Run `make deploy` in $HAIL/internal-gateway.
+
+- Go to the Google Cloud console, API & Services, Credentials.
+  Configure the consent screen.  Add the scope:
+  https://www.googleapis.com/auth/userinfo.email.  Create an OAuth
+  client ID.  Authorize the redirect URI:
+  https://auth.<domain>/oauth2callback.  Download the client secret
```
lgruen, line 107: For `hailctl auth login` to work, http://127.0.0.1/oauth2callback needs to be added as a redirect URI too.

infra/README.md
```diff
@@ -0,0 +1,130 @@
...
+- Enable the the GCP services needed by Hail:
+
+   ```
+   gcloud services enable \
+       compute.googleapis.com \
+       cloudresourcemanager.googleapis.com \
+       servicenetworking.googleapis.com \
+       sqladmin.googleapis.com \
+       container.googleapis.com \
+       serviceusage.googleapis.com
```
lgruen, line 26: cloudprofiler.googleapis.com might be good, too.

infra/main.tf
```diff
@@ -0,0 +1,595 @@
...
+  member = "serviceAccount:${google_service_account.batch_agent.email}"
+}
+
+resource "google_compute_firewall" "vdc_to_batch_worker" {
+  name    = "vdc-to-batch-worker"
+  network = google_compute_network.internal.name
+
+  source_ranges = [google_container_cluster.vdc.cluster_ipv4_cidr]
+
+  target_tags = ["batch2-worker"]
```
lgruen, line 580: Where is this network tag set?

Should this rule apply to the default network?

infra/README.md
lgruen: This works when using e.g. the terraform service account for running the VM.

Also it seems good to use at least 32 GB of RAM for the VM (otherwise compiling the query service didn't work for me).

infra/README.md
```diff
@@ -0,0 +1,180 @@
...
+- Create `$HOME/.hail/global.tfvars` that looks like:
+
+   ```
+   gsuite_organization = "<gsuite-organization>"
+   # batch_gcp_regions is a JSON array of string, the names of the gcp
+   # regions to schedule over in Batch.
+   batch_gcp_regions = "<batch-gcp-regions>"
+   gcp_project = "<gcp-project-id>"
+   # gcp_location is the bucket location that spans the regions you're
+   # going to schedule across in Batch.  If you are running on one
```
lgruen, line 45: > If you are running on one region, it can just be that region.

Unfortunately that doesn't quite work in the Australian setting. `location = "australia-southeast1"` and `storage_class = "MULTI_REGIONAL"` yields:
> Error: googleapi: Error 400: The combination of locationConstraint and storageClass you provided is not supported for your project, invalid

Unfortunately there's only a US, EU, and ASIA setting for multi-region buckets, so that unfortunately would be pretty expensive from Australia.

infra/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+- Run `terraform apply -var-file="$HOME/.hail/global.tfvars"`.
+  Terraform has created a GKE cluster named `vdc`.  Configure
+  `kubectl` to point at the vdc cluster:
+
+  ```
+  gcloud container clusters get-credentials --zone <gcp-zone> vdc
+  ```
+
+- Go to the Google Cloud console, VPC networks > internal > Private
```
lgruen, line 63:
```suggestion
- Go to the Google Cloud console, VPC networks > default > Private
```

infra/README.md
```diff
@@ -0,0 +1,180 @@
...
+  get settings from the default/global-config secret:
+
+  ```
+  kubectl -n default get secret global-config -o json | jq '.data | map_values(@base64d)'
+  ```
+
+- In `$HAIL/docker/third-party` run:
+
+  ```
+  PROJECT=<gc-project-id> ./copy_images.sh
```
lgruen, line 107:
```suggestion
  PROJECT=<gcp-project-id> ./copy_images.sh
```

infra/README.md
```diff
@@ -0,0 +1,180 @@
...
+    KUBERNETES_SERVER_URL='<k8s-server-url>' \
+    HAIL_DEFAULT_NAMESPACE='default' \
+    HAIL_DOMAIN=<domain> \
+    HAIL_GCP_ZONE=<gcp-zone> \
+    HAIL_GCP_PROJECT=<gcp-project> \
+    PYTHONPATH=$HOME/hail/ci:$HOME/hail/batch:$HOME/hail/hail/python \
+    python3 $HAIL/ci/bootstrap.py --extra-code-config '{"username":"<username>","email":"<email>"}' hail-is/hail:main $(git rev-parse HEAD) create_initial_user
+  ```
+
+  Additional users can be added by the intial user by going to auth.hail.is/users.
```
lgruen, line 180:
```suggestion
  Additional users can be added by the initial user by going to auth.<domain>/users.
```

infra/README.md
cseed: Great, I now recommend a n2-standard-8, and to run the VM as a the terraform service account.

infra/README.md
cseed: Thanks!  Yeah, we've run into this kind of issue before.  The k8s services for the Hail services are all in the router deployment, which also runs nginx.  I moved the router service from the router deployment to the gateway deployment.

infra/README.md
cseed: OK, I added advice about ssh to the instructions for spinning up the VM.  We have custom firewall rules for ssh that aren't going to generalize and I think different users will have different security postures, so OK with this being hand managed for now.

infra/README.md
cseed: OK, I renamed gcp_location => batch_logs_bucket_location and added a batch_logs_bucket_storage_class tfvar for this.

bootstrap-gateway/deployment.yaml
```diff
@@ -0,0 +1,71 @@
...
+      priorityClassName: infrastructure
+      nodeSelector:
+        preemptible: "true"
+      tolerations:
+       - key: preemptible
+         value: "true"
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - labelSelector:
```
danking, line 28: I think we should use a consistent spacing for yaml dashes. build.yaml seems to do one space. You have two spaces here and a couple other spots and also zero spaces (under `values`).

ci/bootstrap.py
```diff
@@ -0,0 +1,336 @@
...
+            if j._input_files:
+                copy_script = 'set -ex\n'
+                for src, dest in j._input_files:
+                    assert src.startswith(prefix), (prefix, src)
+                    src = f'/shared{src[len(prefix):]}'
+                    if dest.endswith('/'):
+                        copy_script = copy_script + f'mkdir -p {dest}\n'
+                    else:
+                        copy_script = copy_script + f'mkdir -p {os.path.dirname(dest)}\n'
+                    copy_script = copy_script + f'cp -a {src} {dest}\n'
```
danking, line 122: use `+=`

ci/bootstrap.py
```diff
@@ -0,0 +1,336 @@
...
+
+            print(f'{j._index}: {job_name}: running...')
+
+            if j._parents:
+                for p in j._parents:
+                    assert p._done
+                    if not p._succeeded:
+                        print(f'{j._index}: {job_name}: SKIPPED: parent {p._index} failed')
+                        j._done = True
+                        j._failed = True
```
danking, line 106: It looks like we still try to copy inputs if this happens. Shouldn't we give up on this job at this point?

danking, line 106: You don't use `_failed` anywhere, should this be `_succeeded = False`?

danking: And I think `_done` is just `_succeeded is not None`. It seems like we should have:

```
    def __init(self, ..):
        ...
        self._succeeded = None

    @property
    def done(self):
        return self._succeeded is not None
```

ci/bootstrap.py
```diff
@@ -0,0 +1,336 @@
...
+    cid = outerr[0].decode('ascii').strip()
+
+    outerr = await check_shell_output(f'docker wait {cid}')
+
+    exit_code = int(outerr[0].decode('ascii').strip())
+    return cid, exit_code == 0
+
+
+class LocalBatchBuilder:
+    def __init__(self, attributes, callback):
```
danking, line 65: New services team code needs types on the easy stuff at least. `attributes` is a `Dict[str, str]`, etc.

ci/bootstrap.py
```diff
@@ -0,0 +1,336 @@
...
+{clone_or_fetch_script(self.branch_url())}
+
+git checkout {shq(self._sha)}
+'''
+
+
+async def main():
+    await kube.config.load_kube_config()
+
+    parser = argparse.ArgumentParser(description='Create initial Hail as a service account.')
```
danking, line 290: This does much more than create a HaaS account. I think this is "Execute build.yaml without an extant CI"

ci/bootstrap.py
```diff
@@ -0,0 +1,336 @@
...
+class Branch(Code):
+    def __init__(self, owner, repo, branch, sha, extra_config):
+        self._owner = owner
+        self._repo = repo
+        self._branch = branch
+        self._sha = sha
+        self._extra_config = extra_config
+
+    def short_str(self):
+        return f'br-{self._owner}-{self._repo}-{self._branch}'
```
danking, line 260: This isn't used and it's not the format we usually expect of a branch (e.g. owner/repo:branch), can we remove it?

ci/bootstrap.py
```diff
@@ -0,0 +1,336 @@
...
+
+    def repo_dir(self):
+        return '.'
+
+    def branch_url(self):
```
danking, line 265: this is the repo url

ci/bootstrap.py
```diff
@@ -0,0 +1,336 @@
...
+    assert len(repo_pieces) == 2, f'{repo_pieces} {branch_pieces[0]}'
+    owner = repo_pieces[0]
+    repo_name = repo_pieces[1]
+
+    branch_name = branch_pieces[1]
+
+    if args.extra_code_config is not None:
+        extra_code_config = json.loads(args.extra_code_config)
+    else:
+        extra_code_config = {}
```
danking, line 316: let's set a `default='{}'` on the `add_argument` for extra-code-config.

ci/bootstrap.yaml
```diff
@@ -0,0 +1,71 @@
...
+globalDefault: false
+description: "For production services like apiserver, batch, etc."
+---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: user
+value: 500000
+globalDefault: true
+description: "For user activities like user batch jobs."
```
danking, line 23: User jobs are run outside k8s. Is this for pods in PR and dev namespaces?

ci/bootstrap.yaml
```diff
@@ -0,0 +1,71 @@
...
+  name: production
+value: 900000
+globalDefault: false
+description: "For production services like apiserver, batch, etc."
+---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: user
+value: 500000
```
danking, line 21: The relative value of these classes doesn't matter, right? I feel like this sets a bad precedent of seemingly arbitrary choices. Should we maybe do factors of two? So, 1M, 500k, 250k, 0?

ci/bootstrap_create_accounts.py
```diff
@@ -0,0 +1,81 @@
...
+        namespace_name = DEFAULT_NAMESPACE
+    else:
+        namespace_name = None
+
+    return await db.execute_insertone(
+        '''
+INSERT INTO users (state, username, email, is_developer, is_service_account, gsa_email, gsa_key_secret_name, namespace_name)
+VALUES (%s, %s, %s, %s, %s, %s, %s, %s);
+''',
+        ('creating', username, email, is_developer, is_service_account, gsa_email, gsa_key_secret_name, namespace_name))
```
danking, line 42: This function races between attempts of this job. I think we ought to do an `insert ... on duplicate key ignore` followed by a select to get the user's id. `create_user` looks easy to make idempotent. We can race to create a session, it doesn't matter which one ends up in the secret as long as it's a valid one. All the k8s secret creates need to be idempotent create.

ci/create_initial_account.py
```diff
@@ -0,0 +1,38 @@
...
+        if row['state'] == 'active':
+            return None
+        return row['id']
+
+    return await db.execute_insertone(
+        '''
+INSERT INTO users (state, username, email, is_developer, is_service_account)
+VALUES (%s, %s, %s, %s, %s);
+''',
+        ('creating', username, email, 1, 0))
```
danking, line 19: same thing here about insert ignore

config.mk
```diff
@@ -1,9 +1,9 @@
...
+PROJECT := hail-vdc-staging
 DOCKER_ROOT_IMAGE := gcr.io/$(PROJECT)/ubuntu:18.04
-DOMAIN := hail.is
-INTERNAL_IP := 10.128.0.57
-IP := 35.188.91.25
-KUBERNETES_SERVER_URL := https://104.198.230.143
+DOMAIN := staging.hail.is
+INTERNAL_IP := 10.128.0.2
+IP := 35.224.188.20
+KUBERNETES_SERVER_URL := https://34.123.185.11
```
danking, line 6: Revert these changes, right?

danking (changes requested):
Before I approve, I want to run through your readme on an empty project to convince myself we've sufficiently documented the process.

infra/main.tf
```diff
@@ -0,0 +1,694 @@
...
+
+  master_auth {
+    username = ""
+    password = ""
+
+    client_certificate_config {
+      issue_client_certificate = false
+    }
+  }
+}
```
lgruen, line 66: What do you think about using a release channel for the GKE cluster?

```suggestion
  release_channel {
    channel = "STABLE"
  }
}
```

ci/bootstrap.yaml
cseed: FYI, this code isn't new, it came from vdc/ stuff from a long time ago.  So I'd probably rather not change it all things being equal.  K8s priority classes are 32-bit, and user classes should be in the range 0-1B.  Numbers greater than 1B are reserved for system processes that shouldn't be evicted/preempted.

Yes, the numbers are relatively.  All the examples I've seen use big numbers like this, so that's what I did.  I think the point is to leave space if you want to insert other priorities (so don't make them 1, 2, 3, etc.)

ci/bootstrap.yaml
cseed: Again, this is old cold pulled from vdc/.  I'll update the comment.  It is for everything that isn't specifically infrastructure or production: that includes tests, notebooks, etc.

gateway/deployment.yaml
```diff
@@ -83,3 +83,18 @@ spec:
   selector:
     matchLabels:
       app: gateway
+---
```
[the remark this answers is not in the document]
cseed: I'm not sure what bootstrap-gateway has to do with it.  That's a standalone gateway to get Let's Encrypt before any of the other infra is spun up.

So this is the usual nginx DNS problem.  The gateway deploy will fail at this point without this.  Maybe that's OK?  It should recover (eventually) once the bootstrap deploy happens.  But it is nice to check each step is working after running it, and this is needed for that.  Happy to do it whichever way you prefer.

bootstrap-gateway/deployment.yaml
cseed: Agreed.  When I'm paying attention, I always try to use one.

ci/bootstrap.py
cseed: It isn't?  From ci/github.py:

```
class UnwatchedBranch(Code):
    ...
    def short_str(self):
        return f'br-{self.branch.repo.owner}-{self.branch.repo.name}-{self.branch.name}'
```

ci/bootstrap_create_accounts.py
cseed: I handled this slightly differently.  I already check the user exists above, but you're right there is still a race condition (two racing inserts) so I wrapped this in a transaction.

cseed:
OK, I think I addressed the comments.  I will do a dry run through to make everything is working.  I will test this again from scratch, but I think it's more important to get this in and focus on building automated tests for it rather than spend time hand testing.

danking (approved):

cseed:
Thanks, @danking!  Looks like I broke auth deploy, investigating.

PR #9717, opened 2020-11-18
[hail] add services-jvm-test to Makefile and gradle

danking:
Without this it is a huge pain to run the services tests locally.


tpoterba (approved):
yay!

PR #9718, opened 2020-11-18
remove batch-pods

cseed:
This removes batch-pods and redirects all secret references to the default namespace.  Before this goes in, I will make sure all batch-pods secrets that don't exist in default are copied there.


jigold:
I'll review tomorrow unless you need it tonight. I'm putting the WIP label on it regardless.

jigold:
Also, I'm so glad we're doing this! I always found this incredibly confusing!

cseed:
I copied all the secrets from batch-pods into default that (1) didn't already exist (by name) in default, and (2) weren't k8s service account tokens (which are batch-pods specific).

I also fixed the remaining test failures so this should be ready to go.

build.yaml
```diff
@@ -41,56 +42,20 @@ steps:
    config: batch/service-account.yaml
    dependsOn:
     - default_ns
- - kind: createNamespace
-   name: batch_pods_ns
-   namespaceName: batch-pods
-   public: false
-   secrets:
-    - gcr-push-service-account-key
```
jigold, line 49: Do we not use the push service account key anymore?

batch/batch/front_end/front_end.py
```diff
@@ -540,12 +540,9 @@ def check_service_account_permissions(user, sa):
         return
     if user == 'ci':
         if sa['name'] in ('ci-agent', 'admin'):
-            if DEFAULT_NAMESPACE == 'default':  # real-ci needs access to all namespaces
-                return
-            if sa['namespace'] == BATCH_PODS_NAMESPACE:
-                return
-    if user == 'test':
-        if sa['name'] == 'test-batch-sa' and sa['namespace'] == BATCH_PODS_NAMESPACE:
+            return
```
jigold, line 543: I think the new rule should be this:

```
if user == 'ci':
    if sa['name'] in ('ci-agent', 'admin'):
        if DEFAULT_NAMESPACE == 'default' or sa['namespace'] == DEFAULT_NAMESPACE:
              return
```

The rationale was that ci needs to be able to access all namespaces, but if it's a test version of ci, it should only have access to its own namespace.

i.e. ci in the default namespace can use a service account in `test-pr-XXX-default` or in `default`, but the ci in the namespace `test-pr-XXX-default` should only be able to access `test-pr-XXX-default` and not `default`.

build.yaml
```diff
@@ -41,56 +42,20 @@ steps:
...
-   name: deploy_batch_sa_batch_pods
-   namespace:
-     valueFrom: batch_pods_ns.name
-   config: batch/service-account-batch-pods.yaml
-   dependsOn:
-    - default_ns
-    - batch_pods_ns
-    - deploy_batch_sa
- - kind: deploy
-   name: deploy_default_admin_admin
```
jigold, line 65: I don't remember what this was used for.

ci/ci-agent-default.yaml
```diff
@@ -1,24 +0,0 @@
-kind: Role
```
jigold, line 1: I don't see the new file ci/ci-agent.yaml

ci/default-admin-admin.yaml
```diff
@@ -1,12 +0,0 @@
-kind: RoleBinding
-apiVersion: rbac.authorization.k8s.io/v1
-metadata:
-  name: default-admin-admin
-subjects:
- - kind: ServiceAccount
```
jigold, line 6: Do we not need an admin service account anymore? If so, then we can change the batch rule.

ci/test/resources/build.yaml
```diff
@@ -149,7 +145,7 @@ steps:
    secrets:
     - name: test-tokens
       namespace:
-        valueFrom: batch_pods_ns.name
+        valueFrom: default_ns.name
```
jigold, line 148: Where does this test-tokens secret come from? I guess this test depends on create_accounts because test_ci depends on that.

jigold (changes requested):

ci/test/resources/build.yaml
cseed: This is the test build.yaml.  test-tokens is created by the outer build.yaml.

ci/default-admin-admin.yaml
cseed: I think this is dead.  I can't find where the account is used or created.  I'll update the batch rule.

cseed:
I hope this breaks so I can use up some of that downtime budget!  Totally worth it to get rid of batch-pods.

batch/batch/front_end/front_end.py
```diff
@@ -539,8 +539,9 @@ def check_service_account_permissions(user, sa):
     if sa is None:
         return
     if user == 'ci':
-        if sa['name'] in ('ci-agent', 'admin'):
-            return
+        if sa['name'] == 'ci-agent':
+            if DEFAULT_NAMESPACE == 'default' or sa['namespace'] == DEFAULT_NAMESPACE:
```
jigold, line 543: You're getting a pylint failure now.

```
batch/front_end/front_end.py:543:15: R1714: Consider merging these comparisons with "in" to "DEFAULT_NAMESPACE in ('default', sa['namespace'])" (consider-using-in)
```

jigold (approved):
I think this looks good other than the pylint error. If you have to make substantial changes, dismiss the review and I'll relook.

batch/batch/front_end/front_end.py
cseed: Thanks, fixed!

PR #9723, opened 2020-11-19
[auth] fix handling of empty email for service accounts

danking:
Empty text boxes are (always?) transmitted to the server as an empty string.
This change teaches auth to interpret both the empty string or a missing email attribute
equivalently. Moreover, service accounts with the empty string email are converted to
NULL emails.


danking:
I noticed this because the `atgu` had an empty string email. I fixed it already.

PR #9727, opened 2020-11-19
SVD tests should be robust to sign ambiguity

johnc1231:
If hail and numpy use two different lapack implementations, might not get the same sign on the singular vectors. This PR switches our vector comparison tests to check the absolute values of the matrices, plus make sure they multiply back together properly.

### DIFF
--- a/hail/python/test/hail/expr/test_ndarrays.py
+++ b/hail/python/test/hail/expr/test_ndarrays.py
@@ -789,7 +789,7 @@ def test_svd():
             np.testing.assert_array_almost_equal(hs, ns)
 
             # U is orthonormal
-            uut = hu @ hu.T
+            uut = hu.T @ hu
             np.testing.assert_array_almost_equal(uut, np.identity(uut.shape[0]))
 
             # V is orthonormal
--- a/hail/python/test/hail/expr/test_ndarrays.py
+++ b/hail/python/test/hail/expr/test_ndarrays.py
@@ -781,18 +781,25 @@ def test_svd():
         k = min(np_array.shape)
         rank = np.linalg.matrix_rank(np_array)
 
-        # SVD might give you reflected singular vectors.
-        def vectors_equal_magnitude(nd1, nd2):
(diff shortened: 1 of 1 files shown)


patrick-schultz:
Actually, comparing singular vectors is not a robust test, even accounting for sign. Suppose `A` has two equal (or nearly equal) singular values. Then there is a 2-dimensional subspace of vectors, all of which are equally good singular vectors for that singular value. If the singular values are sufficiently separated, then comparing singular vectors should be safe, but I don't think it's necessary; the other checks should force that.

I think we only need to check (all approximate comparisons),
* we got the right singular values, by comparing with numpy (unless we constructed a test matrix with known singular values)
* the singular vectors are orthonormal (i.e. `Ut U = Id` and `Vt V = Id`)
* the factorization `A = U Sigma Vt`

Then it follows that for each right singular vector `V_i`, `A v_i = sigma_i u_i` holds approximately, so `v_i` is a good singular vector, i.e. it really does capture `sigma_i` variance, and we checked that `sigma_i` is close to the true singular value.

johnc1231:
I've made the changes you suggested patrick, but now we have a failing test. I'm getting a situation where Vt is clearly orthonormal, and U@S@Vt multiplies back to the input matrix, but U@U.t is not particularly close to the identity matrix. It's not clear why this is. The test is:

```
np_rank_2_wide_rectangle = np.arange(12).reshape((4, 3))
rank_2_wide_rectangle = hl.nd.array(np_rank_2_wide_rectangle)

.........

assert_evals_to_same_svd(rank_2_wide_rectangle, np_rank_2_wide_rectangle, full_matrices=False)
```

(The 4th test in the `test_svd` method)

patrick-schultz:
`U` should be a tall-skinny matrix, say m x n, with orthonormal columns, right? Then `U.t @ U` should be (close to) the n x n identity. `U @ U.t` is an m x m matrix with rank n, so it can't be the identity.

johnc1231:
Duh, thank you. Fixed.

patrick-schultz (approved):

PR #9733, opened 2020-11-23
[query] Slightly better Error Message from realized_relationship_matrix

johnc1231:
I just added a slightly more informative error message. Do you want something more than this?

### DIFF
--- a/hail/python/hail/methods/statgen.py
+++ b/hail/python/hail/methods/statgen.py
@@ -1997,7 +1997,8 @@ def realized_relationship_matrix(call_expr) -> BlockMatrix:
         bm = BlockMatrix.from_entry_expr(normalized_gt)
         return (bm.T @ bm) / (bm.n_rows / bm.n_cols)
     except FatalError as fe:
-        raise FatalError("Could not convert MatrixTable to BlockMatrix. It's possible all variants were dropped by variance filter.") from fe
+        raise FatalError("Could not convert MatrixTable to BlockMatrix. It's possible all variants were dropped by variance filter.\n"
+                         "Check that the input MatrixTable has at least two samples in it:  mt.count_cols().") from fe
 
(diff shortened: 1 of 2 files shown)


hail/python/hail/methods/statgen.py
```diff
@@ -1989,13 +1989,15 @@ def realized_relationship_matrix(call_expr) -> BlockMatrix:
...
-    bm = BlockMatrix.from_entry_expr(normalized_gt)
-
-    return (bm.T @ bm) / (bm.n_rows / bm.n_cols)
+    normalized_gt = hl.or_else((fmt.__gt - fmt.__mean_gt) / fmt.__centered_length, 0.0)
 
+    try:
+        bm = BlockMatrix.from_entry_expr(normalized_gt)
+        return (bm.T @ bm) / (bm.n_rows / bm.n_cols)
+    except FatalError as fe:
+        raise FatalError("Could not convert MatrixTable to BlockMatrix. It's possible all variants were dropped by variance filter.") from fe
```
danking, line 2000: In other user-focused error messages we've provided suggested fixes. Is it appropriate to do that here?

```
FatalError: Could not convert MatrixTable to BlockMatrix. It's possible all variants were dropped by variance filter.
Check that the input MatrixTable has at least two samples in it:  mt.count_cols().
Check that the input MatrixTable has at least one non-zero-variance variant: ...
```
Hmm. It'd be nice if the variance computation above was also in a user accessible method so we could just include it there. Maybe that's a step too far.

danking (changes requested):

hail/python/hail/methods/statgen.py
johnc1231: I'll do the check at least two samples thing. I don't know how to tell them to calculate the variance. Also, this error should virtually never happen.

danking (approved):

PR #9737, opened 2020-11-24
[batch] automatically set GOOGLE_APPLICATION_CREDENTIALS

danking:
gcloud uses this variable to find the tokens file. We should set it so
that users do not need to configure anything. I also fixed some
missing arguments in the non-async get_userinfo.

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -683,11 +683,6 @@ WHERE user = %s AND id = %s AND NOT deleted;
                     if user != 'ci':
                         raise web.HTTPBadRequest(reason=f'unauthorized secret {(secret["namespace"], secret["name"])}')
 
-                env = spec.get('env')
-                if not env:
-                    env = []
-                    spec['env'] = env
-
                 spec['secrets'] = secrets
                 secrets.append({
                     'namespace': DEFAULT_NAMESPACE,
@@ -695,7 +690,12 @@ WHERE user = %s AND id = %s AND NOT deleted;
                     'mount_path': '/gsa-key',
                     'mount_in_copy': True
                 })
-                if all('GOOGLE_APPLICATION_CREDENTIALS' != envvar['name'] for envvar in spec.env):
+
+                env = spec.get('env')
+                if not env:
+                    env = []
+                    spec['env'] = env
+                if all(envvar['name'] != 'GOOGLE_APPLICATION_CREDENTIALS' for envvar in spec.env):
                     spec.env.append(
                         {'name': 'GOOGLE_APPLICATION_CREDENTIALS', 'value': '/gsa-key/key.json'})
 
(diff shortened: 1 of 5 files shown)


batch/test/test_batch.py
```diff
@@ -598,6 +598,16 @@ def test_verify_no_access_to_metadata_server(client):
...
 
 
+def test_can_use_hailctl_auth(client):
+    builder = client.create_batch()
+    j = builder.create_job(os.environ['CI_UTILS_IMAGE'],
+                           'gcloud', 'auth', 'list')
+    status = j.wait()
+    assert status['state'] == 'Success', f'{j.log(), status}'
+    log = j.log()
+    assert '*       test-665@hail-vdc.iam.gserviceaccount.com' in log, f'{j.log(), status}'
```
jigold, line 608: Will this be true for tests running in the default namespace? I can't remember how many service accounts we have going around.

batch/batch/front_end/front_end.py
```diff
@@ -683,13 +683,22 @@ async def create_jobs(request, userdata):
...
+                    spec['env'] = env
+
                 spec['secrets'] = secrets
                 secrets.append({
                     'namespace': DEFAULT_NAMESPACE,
                     'name': userdata['gsa_key_secret_name'],
                     'mount_path': '/gsa-key',
                     'mount_in_copy': True
                 })
+                if all('GOOGLE_APPLICATION_CREDENTIALS' != envvar['name'] for envvar in spec.env):
```
jigold, line 698: Why did you break up the env definition with the secrets?

jigold (dismissed):

batch/batch/front_end/front_end.py
danking: I guess I thought of this block as part of the secrets and moving the env down here would break up the secrets definition. I'll switch.

batch/test/test_batch.py
danking: The test-gsa-key is one of the secrets copied from default by the `createNamespace` that creates `default`, so it should all be the same account.
```
(base) # k get secret test-gsa-key -o json  | jq '.data["key.json"]' -r | base64 -D | jq '.client_email'
"test-665@hail-vdc.iam.gserviceaccount.com"

```

danking:
bump

build.yaml
```diff
@@ -2244,9 +2244,11 @@ steps:
      export HAIL_CURL_IMAGE={{ curl_image.image }}
      export HAIL_DEFAULT_NAMESPACE={{ default_ns.name }}
      export HAIL_NETCAT_UBUNTU_IMAGE={{ netcat_ubuntu_image.image }}
+     export HAIL_HAIL_BASE_IMAGE={{ hail_base_image.image }}
      export DOCKER_ROOT_IMAGE="gcr.io/{{ global.project }}/ubuntu:18.04"
      export HAIL_TEST_TOKEN_FILE=/user-tokens/tokens.json
      export HAIL_TEST_DEV_TOKEN_FILE=/dev-tokens/tokens.json
+     export HAIL_TOKEN="{{ token }}"
```
jigold, line 2251: Should we make sure this token shouldn't be used in other places in the tests? I think I just generated random tokens inside the tests.

jigold (changes requested):

build.yaml
danking: I modified the billing project creation. The other ones don't modify global state so it doesn't seem important.

jigold (approved):

PR #9743, opened 2020-11-29
[httpx] add Hail httpx library

danking:
I simplified the TLS context methods into three cases: internal server, external
client session, internal client.

I added a blocking shim around aiohttp so that all our HTTP goes through
aiohttp. I added automatic retrying directly into the httpx library.


danking:
bump

PR #9744, opened 2020-11-29
[batch] change worker to default to the private network

danking:
At this point, every job that needs the private network specifies
it (https://github.com/hail-is/hail/pull/9380). This change moves jobs with
unspecified `network` to the private network.

I removed the gce-deploy-config because jobs should not, by default, assume they
are on the private network. The GCE Deploy Config instructs you to enter GKE
using the internal-gateway, which is on our private network.


danking:
needs default-auth to work correclty

PR #9749, opened 2020-11-30
[query] Add or_error method to SwitchBuilder

nawatts:
CaseBuilder has an `or_error` method to throw an error if no `when` conditions are true. Currently, SwitchBuilder does not have an equivalent method: it only supports returning a default value or missing. The option to throw an error on an unhandled value can be useful for making sure that all possible values for an enum expression have been accounted for in switch cases.

### DIFF
--- a/hail/python/hail/expr/builders.py
+++ b/hail/python/hail/expr/builders.py
@@ -83,7 +83,7 @@ class SwitchBuilder(ConditionalBuilder):
     @typecheck_method(value=expr_any, then=expr_any)
     def when(self, value, then) -> 'SwitchBuilder':
         """Add a value test. If the `base` expression is equal to `value`, then
-         returns `then`.
+        returns `then`.
 
         Warning
         -------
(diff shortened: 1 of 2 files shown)


hail/python/hail/expr/builders.py
```diff
@@ -83,7 +83,7 @@ def f(base):
     @typecheck_method(value=expr_any, then=expr_any)
     def when(self, value, then) -> 'SwitchBuilder':
         """Add a value test. If the `base` expression is equal to `value`, then
-         returns `then`.
+        returns `then`.
```
nawatts, line 86: Unrelated formatting fix for [SwitchBuilder.when docs](https://hail.is/docs/0.2/functions/hail.expr.builders.SwitchBuilder.html#hail.expr.builders.SwitchBuilder.when).

johnc1231 (approved):
Thanks, good addition. The one thing I'd say is just that when we make a user facing change, we try to make one of the commits look like:

CHANGELOG: Added `or_error` method to `SwitchBuilder`

This ends up being helpful when I have to go through and generate the change log for a new version release.

nawatts:
> The one thing I'd say is just that when we make a user facing change, we try to make one of the commits look like:
>
> CHANGELOG: Added or_error method to SwitchBuilder
>
> This ends up being helpful when I have to go through and generate the change log for a new version release.

Good to know. It would be nice to document conventions like this in the [docs for software developers](https://hail.is/docs/0.2/getting_started_developing.html#contributing) and/or a [contributing file](https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/setting-guidelines-for-repository-contributors).

johnc1231:
This is a good point, we should look into doing that. Thanks for the suggestion.

nawatts:
Added the "CHANGELOG" note to contribution guidelines in #9752.

PR #9753, opened 2020-11-30
[batch] use marketplace.gcr.io Ubuntu image to avoid Docker Hub rate limits warning

lgruen:


hail/python/hailtop/batch/backend.py
```diff
@@ -359,7 +359,7 @@ def _run(self,
         remote_tmpdir = f'gs://{self._bucket_name}/batch/{token}'
         local_tmpdir = f'/io/batch/{token}'
 
-        default_image = 'ubuntu:18.04'
+        default_image = 'marketplace.gcr.io/google/ubuntu1804:latest'
```
cseed, line 362: I feel like the existence of `default_image` is a mistake.  We definitely don't want to fix it to ubuntu:18.04 forever, and upgrading it here will break existing pipelines.  I think we shouldn't have a default image, and users should always specify it.  And then we should recommend a default image (which will basically be latest Ubuntu TLS or something).

Not directly related, but we're about to switching the root Hail services image from Ubuntu 18 to 20, so these will diverge.

jigold: I think we should make a list of breaking changes we want to make and try to do them at the same time. Mainly I'm thinking of the private / shared interface changes that will be breaking.

cseed: What's the private / shared interface changes?  I maybe missed this.

jigold: The difference between submitting a job private instance job or on a shared pool. The memory, cpu, and storage interfaces will also change.

cseed: Ah, yes.  Agreed!

jigold:
@lgruen Can this change wait until January when we make breaking interface changes?

lgruen:
> @lgruen Can this change wait until January when we make breaking interface changes?

Yes, no problem, especially if you want to get rid of `default_image` altogether. I'll close this PR for now.

PR #9757, opened 2020-11-30
[batch] limit docker context for faster builds

danking:
We never used any other directory anyway.

### DIFF
--- a/batch/Dockerfile
+++ b/batch/Dockerfile
@@ -1,7 +1,7 @@
 FROM {{ service_base_image.image }}
 
-COPY batch/setup.py batch/MANIFEST.in /batch/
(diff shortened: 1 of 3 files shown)


jigold (changes requested):
Why does changing the directory make such a huge difference here?

danking:
There's significantly fewer things in `/batch` than in our entire repository:
```
(base) # du -sh .
295M	.
(base) # du -sh batch
700K	batch

```

At least on Mac, you have to "send the docker context" to the Docker daemon, and that process seems to have a bandwidth of ~5MB/s, so 295MB takes an unpleasantly long amount of time.

jigold (approved):

PR #9759, opened 2020-11-30
[sync] alternative to ksync

danking:
I got fed up with ksync not working the way I wanted and wrote my own
version. Assuming `repo/devbin` is on your path, then:

```
sync.sh batch dking
```

will keep the web_common, gear, hailtop, and batch repositories all up to date
in k8s. It just copies all the files over from your local machine, so its not
the fastest thing in the world. It takes maybe 2 seconds to get back to a
working container.

It also assumes there's one container per-pod.

It is completely disabled in the default namespace (I remove it in
deployment.yaml if we are in deploy mode).


jigold (changes requested):
I'm having a bit of a hard time with the bash code.

danking:
OK, I reimplemented the sync-er in Python. This works well enough though it would benefit from something that waited for changes to settle down and did one copy-restart.  Currently, you can queue up a bunch of changes and it sometimes take as long as 5 seconds for the whole system to settle down enough that you can refresh and get the new page.

danking:
Actually, it's just the restart that you want to wait on. The copying could effectively be constantly happening. Anyway, it works really well as is, though there's more to do.

danking:
I can't debug this without https://github.com/hail-is/hail/pull/9781, so I'll wait for that.

docker/hail-ubuntu/controller.sh
```diff
@@ -0,0 +1,39 @@
...
+trap term SIGTERM
+
+
+$COMMAND &
+PID=$!
+restarted=yes
+
+while [ $restarted == "yes" ]
+do
+    restarted=no
```
jigold, line 34: Can you describe in words how this all works together? I understand the traps and the code in `restart`. I just don't see how this loop works with `restarted`.

devbin/sync.py
```diff
@@ -0,0 +1,125 @@
...
+import argparse
+import asyncio
+import kubernetes_asyncio as kube
+import logging
+import sys
+import signal
+
+
+configure_logging()
+log = logging.getLogger('scorecard')
```
jigold, line 19: wrong logger.

devbin/sync.py
```diff
@@ -0,0 +1,125 @@
...
+        action='append',
+        type=str,
+        help='An app label to watch.')
+    parser.add_argument(
+        '--namespace',
+        required=False,
+        type=str,
+        help='The namespace in which to watch.')
+    parser.add_argument(
+        '--path',
```
jigold, line 93: I'm just thinking whether there is any reason why we don't need more specificity. I guess we can add it later if needed.

```
app1: [(local1, remote1)]
app2: [(local1, remote1), (local2, remote2)]
```

devbin/krsync.sh
```diff
@@ -0,0 +1,22 @@
...
+if [ -z "$KRSYNC_STARTED" ]; then
+    export KRSYNC_STARTED=true
+    exec rsync --blocking-io --rsh "$0" $@
+fi
+
+# Running as --rsh
+namespace=''
+pod=$1
+shift
+
```
jigold, line 13: Can you explain how this is structured? When would KRSYNC_STARTED ever be set? Also, I thought the argument to `--rsh` was a shell. Finally, how is this block below used and when?

ci/ci/build.py
```diff
@@ -743,6 +743,8 @@ def build(self, batch, code, scope):
 kubectl -n {self.namespace} rollout status --timeout=1h deployment {name} && \
   kubectl -n {self.namespace} wait --timeout=1h --for=condition=available deployment {name}
 EC=$?
+kubectl -n {self.namespace} get deployment -l app={name} -o yaml
+kubectl -n {self.namespace} get pods -l app={name} -o yaml
```
jigold, line 747: I think this already merged.

jigold (changes requested):

devbin/krsync.sh
danking: The first time this file is invoked, `KRSYNC_STARTED` is not set so we enter the if statement at line 4 and call `rsync`, telling rsync that it should use *this script* as `ssh`.

Rsync then calls *back into* us (sync.py -> krsync.sh -> rsync -> krsync.sh). If rsync sees `/from pod:/to`, it calls us with the arguments `pod ...`.  If rsync sees `/from pod@namespace:/to`, it calls us with the arguments `-l pod namespace`. We optimistically assume the first form on line 11. then we shift away the first argument. If the first argument was actually `-l`, then the if statement on line 15 will succeed. In that case, the second (now first) and third (now second) arguments are the pod and namespace. Finally we call kubectl exec, which is effectively a kubectl version of ssh.

So there's like three levels of masquerading! This script, at first, masquerades as `rsync`. Then rsync calls this script, who is now masquerading as `ssh`. Then this script calls `kubectl exec` who also masquerades as `ssh`.

jigold (approved):

PR #9760, opened 2020-12-01
[query] fix default_ns setting for query service deployment

lgruen:


jigold:
I think this change is no longer needed.

lgruen:
> I think this change is no longer needed.

Are you sure? https://github.com/hail-is/hail/blob/main/query/deployment.yaml#L36 still references `default_ns.name`.

lgruen:
Closing this PR as the change is also included in https://github.com/hail-is/hail/pull/9713/files#diff-047888015f6a4348a0f65bedff31969a101531c8037eec89336f542c39973a9c.

PR #9762, opened 2020-12-01
[hailbuntu] Move to 20.04

danking:


danking:
Fighting with docker and ubuntu packages is my least favorite pastime. I think this works.

Dania-Abuhijleh (approved):

PR #9765, opened 2020-12-01
[hailtop/batch] Add logs messages to tests

jigold:
I didn't test the getting the logs code.

### DIFF
--- a/docker/third-party/images.txt
+++ b/docker/third-party/images.txt
@@ -5,12 +5,6 @@ amundsendev/amundsen-search:2.4.0
 debian:9.5
 ghost:3.0-alpine
 google/cloud-sdk:305.0.0-slim
-hailgenetics/python-dill:3.6
-hailgenetics/python-dill:3.6-slim
-hailgenetics/python-dill:3.7
-hailgenetics/python-dill:3.7-slim
-hailgenetics/python-dill:3.8
-hailgenetics/python-dill:3.8-slim
 jupyter/scipy-notebook:c094bb7219f9
 jupyter/scipy-notebook
 python:3.6
--- a/docker/third-party/images.txt
+++ b/docker/third-party/images.txt
@@ -5,6 +5,12 @@ amundsendev/amundsen-search:2.4.0
 debian:9.5
 ghost:3.0-alpine
 google/cloud-sdk:305.0.0-slim
+hailgenetics/python-dill:3.6
+hailgenetics/python-dill:3.6-slim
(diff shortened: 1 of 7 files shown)


docker/third-party/images.txt
```diff
@@ -5,6 +5,12 @@ amundsendev/amundsen-search:2.4.0
 debian:9.5
 ghost:3.0-alpine
 google/cloud-sdk:305.0.0-slim
+hailgenetics/python-dill:3.6
+hailgenetics/python-dill:3.6-slim
+hailgenetics/python-dill:3.7
+hailgenetics/python-dill:3.7-slim
+hailgenetics/python-dill:3.8
+hailgenetics/python-dill:3.8-slim
```
danking, line 13: this seems a bit odd. These images are defined in docker/python-dill. Maybe we should just add a `gcr.io/..` alias there?

hail/python/test/hailtop/batch/utils.py
```diff
@@ -0,0 +1,10 @@
+from hailtop.batch_client.client import Job
+
+
+def job_logs(batch):
+    logs = {}
+    jobs = batch.jobs()
+    for j_status in jobs:
+        j = Job(batch, j_status['job_id'])
+        logs[j.id] = j.log()
+    return logs
```
danking, line 10: Pytest will truncate the output unless it is a string. Either convert to a string here with `str(logs)` or convert in each assert: `str(job_logs(res))`.

hail/python/test/hailtop/batch/test_batch_pool_executor.py
```diff
@@ -1,30 +1,33 @@
...
 import time
+import os
 from hailtop.batch import BatchPoolExecutor
 
+PYTHON_DILL_IMAGE = os.environ['PYTHON_DILL_IMAGE']
+
 
 def test_simple_map():
-    with BatchPoolExecutor(project='hail-vdc') as bpe:
+    with BatchPoolExecutor(project='hail-vdc', image= PYTHON_DILL_IMAGE) as bpe:
```
danking, line 10: no spaces after kwargs

danking (approved):

danking (changes requested):
xxx

docker/third-party/images.txt
jigold: Can you expand on this idea more?

danking (approved):

PR #9768, opened 2020-12-02
[query] Fix column major test to avoid extra copying

johnc1231:
There were two bugs when checking if an ndarray is column major, causing it to always return false. This method is used to decide whether we need to copy the ndarray into a column major form. By fixing this bug, we now don't have to copy both arrays before every matrix multiply.

Benchmark: linear_regression_rows_nd

Before fix timings:

[50.794169104999995, 51.562208821, 60.402329871999996]

After fix timings: 

[38.504665161, 39.919831891, 37.882298633999994]


tpoterba:
🎉 🎉 🎉 🎉 🎉 🎉 🎉

tpoterba (approved):

PR #9769, opened 2020-12-02
[batch] Refactor zone quotas code into a zone monitor

jigold:
This PR is the first towards the goal of supporting arbitrary machine types.

I tested both code paths where we create zones from a static list versus if there's over 4000 cores, creating it with east and west regions as well.

I thought about putting the regions and init_zones in the database, but figured that could be a follow up PR. I wasn't sure what the best way to do that was.

### DIFF
--- a/batch/batch/driver/main.py
+++ b/batch/batch/driver/main.py
@@ -32,6 +32,7 @@ from ..batch_configuration import (REFRESH_INTERVAL_IN_SECONDS,
                                    HAIL_SHOULD_CHECK_INVARIANTS, WORKER_LOGS_BUCKET_NAME, PROJECT)
 from ..globals import HTTP_CLIENT_MAX_SIZE
 
+from .zone_monitor import ZoneMonitor
 from .instance_pool import InstancePool
 from .scheduler import Scheduler
 from .k8s_cache import K8sCache
@@ -745,6 +746,10 @@ SELECT worker_type, worker_cores, worker_disk_size_gb,
     log_store = LogStore(BATCH_BUCKET_NAME, WORKER_LOGS_BUCKET_NAME, instance_id, pool, credentials=credentials)
     app['log_store'] = log_store
 
+    zone_monitor = ZoneMonitor(app)
+    app['zone_monitor'] = zone_monitor
+    await zone_monitor.async_init()
+
     inst_pool = InstancePool(app, machine_name_prefix)
     app['inst_pool'] = inst_pool
     await inst_pool.async_init()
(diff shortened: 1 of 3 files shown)


batch/batch/driver/zone_monitor.py
```diff
@@ -0,0 +1,132 @@
...
+
+    async def async_init(self):
+        self.task_manager.ensure_future(retry_long_running(
+            'update_zones_loop',
+            self.update_region_quotas_loop))
+
+    def shutdown(self):
+        self.task_manager.shutdown()
+
+    def zone_weights(self, worker_cores, worker_local_ssd_data_disk, worker_pd_ssd_data_disk_size_gb):
```
jigold, line 91: I refactored this code from what Cotton had as the cores, local ssd, etc are not static in the case of multiple pools and are no longer global variables.

jigold:
FYI: @lgruen

batch/batch/driver/main.py
```diff
@@ -745,6 +746,10 @@ async def on_startup(app):
     log_store = LogStore(BATCH_BUCKET_NAME, WORKER_LOGS_BUCKET_NAME, instance_id, pool, credentials=credentials)
     app['log_store'] = log_store
 
+    zone_monitor = ZoneMonitor(app)
```
lgruen, line 749: Could we also pass `regions` and `init_zones` here, maybe after adding them to `batch_configuration`?

jigold: Exactly, I just wasn't sure the best way to do this. Should it be in the database as a globals field in text that's JSON or should it be passed as part of the deployment.yaml as an environment variable that's also JSON? Perhaps, there's a better way than these options.

lgruen: Yet another option might be to move all the environment variables from batch/batch/batch_configuration.py into a single JSON K8s secret, similar to hail/python/hailtop/config/deploy_config.py. But that would be a larger refactoring effort.

batch/batch/driver/zone_monitor.py
```diff
@@ -0,0 +1,132 @@
...
+        if self.region_info is None:
+            return self.init_zone_weights
+
+        _zone_weights = []
+        for _, r in self.region_info.items():
```
danking, line 96: use .values() instead.

batch/batch/driver/main.py
danking: all the config is currently in the db, I'm inclined to stick with that unless we rethink everything. deployment env var seems natural. I don't think we have any strong case currently to change regions or initial zones at run-time.

danking (changes requested):
looks good overall

jigold:
I propose saving the configuration question to a different PR.

danking (approved):

PR #9770, opened 2020-12-02
[query] Propogate line numbers through IR->Code->LIR

patrick-schultz:
`BaseIR` currently has a `lineNumber` field, which by default is 0 (the "no line number" sentinel). This PR introduces infrastructure to carry that line number information through the emitter, and anything that generates code, such that each generated lir instruction gets a line number from whichever IR node was responsible for its generation (if any). This PR does nothing to set meaningful line numbers in IR, so for now this adds no new functionality.

Every `Code` method that generates lir instructions now takes an implicit `LineNumber` argument. Then implicit `LineNumber` arguments are added wherever needed. The chain ends either by passing in a `lineNumber` from an IR, or by passing `LineNumber.none` (which is just 0), when generating code in a context where there is no IR node to attribute.


tpoterba:
I know I said I thought this was a reasonable approach a while ago, but I’ve been thinking hard about this change since last week, and I think I want us to explore a larger set of designs before committing to this strategy. The approach in this PR doubles down on the functional Code[T] structure, which is something we’re trying to move away from with CodeBuilder. I think if I could choose an interface for injecting line numbers from IR in emit it would look something like:

```scala
class Emit[C](
  val ctx: ExecuteContext,
  val cb: EmitClassBuilder[C]) { emitSelf =>

  val methods: mutable.Map[(String, Seq[Type], Seq[PType], PType), EmitMethodBuilder[C]] = mutable.Map()

  private[ir] def emitVoid(cb: EmitCodeBuilder, ir: IR, mb: EmitMethodBuilder[C], region: StagedRegion, env: E, container: Option[AggContainer], loopEnv: Option[Env[LoopRef]]): Unit = {
    cb.startLine(ir.lineNumber)
    ... implementaiton
    cb.endLine(ir.lineNumber)
```

How could we make something like this work? Can we get away without every Code[T] knowing the source line? The JVM represents line numbers as an array of (line start bytecode index, line bytecode length) tuples, and I think it will be possible to produce this more easily.

I think part of my concern is that I’m not entirely sold by the need to have a whole stack of IR printouts and associated line numbers — right now, the option to get debug information by LIR line number or IR (fully lowered, compile-ready) seems plenty sufficient.

Part of my pushback is that I'm hesitant to use Scala implicits pervasively without a careful cost/benefit consideration — they’ve been a source of confusion and frustration in the past, and the intentional paucity of implicits in our current codebase reflects that.

patrick-schultz:
> The approach in this PR doubles down on the functional Code[T] structure

I don't see anything in the design that prevents us from moving away from `Code[T]`, but it does have to support it for now.

> I think if I could choose an interface for injecting line numbers from IR in emit it would look something like:

This is also the interface I would like to see for CodeBuilder. And you're right, making that change would allow methods taking a `CodeBuilder` to not need a line number argument. I agree that's better. I may have gotten a bit of tunnel vision in the middle of the giant mechanical refactoring :) I will make this change.

> I think part of my concern is that I’m not entirely sold by the need to have a whole stack of IR printouts and associated line numbers — right now, the option to get debug information by LIR line number or IR (fully lowered, compile-ready) seems plenty sufficient.

I think most of this PR is necessary for debug information with the fully lowered IR line numbers. It doesn't do anything to propagate line numbers through IR lowerings, which is what would be needed to support line numbers at earlier compiler stages.

> Part of my pushback is that I'm hesitant to use Scala implicits pervasively without a careful cost/benefit consideration

My main reason for that approach was to manage the number of changes required in this PR. We could follow up on this with making line number arguments explicit in manageable chunks. But personally this seems like the ideal use case for implicit arguments. And as `Code` goes away, the number of places with implicit line number arguments should go down significantly.

johnc1231:
What's going on with this PR? It's accumulating a lot of conflicts.

patrick-schultz:
We ended up deciding this should wait for the Code->CodeBuilder transition to finish.

PR #9772, opened 2020-12-02
[batch] Split instance_pool into instance_pool and instance_monitor

jigold:
Stacked on #9769 

I tried to make the code changes as small as possible and this is just a refactoring. I split the current instance_pool into the instance_monitor and the instance_pool. The main difference is the instance pool and instance monitor are recording two exact copies of the instances by state and live total cores etc which are linked via `adjust_for_*_instance`. Eventually with multiple pools, these numbers won't be identical and one will be for all instances and the other will be pool specific.

### DIFF
--- a/batch/batch/driver/instance_monitor.py
+++ b/batch/batch/driver/instance_monitor.py
@@ -1,0 +1,291 @@
+import re
+import asyncio
+import aiohttp
+import logging
+import collections
+import datetime
+import sortedcontainers
+import dateutil.parser
+
+from hailtop import aiotools
+from hailtop.utils import time_msecs, secret_alnum_string
+
+from .instance import Instance
+from ..batch_configuration import PROJECT
+
+log = logging.getLogger('instance_monitor')
+
+
+def parse_resource_name(resource_name):
+    match = RESOURCE_NAME_REGEX.fullmatch(resource_name)
+    assert match
+    return match.groupdict()
+
+
+RESOURCE_NAME_REGEX = re.compile('projects/(?P<project>[^/]+)/zones/(?P<zone>[^/]+)/instances/(?P<name>.+)')
+
+
+class InstanceMonitor:
+    def __init__(self, app, machine_name_prefix):
+        self.app = app
+        self.db = app['db']
+        self.compute_client = app['compute_client']
+        self.logging_client = app['logging_client']
+        self.log_store = app['log_store']
+        self.zone_monitor = app['zone_monitor']
+
+        self.machine_name_prefix = machine_name_prefix
+
+        self.instances_by_last_updated = sortedcontainers.SortedSet(
+            key=lambda instance: instance.last_updated)
+
+        self.name_instance = {}
+
+        self.n_instances_by_state = {
+            'pending': 0,
+            'active': 0,
+            'inactive': 0,
+            'deleted': 0
+        }
+
+        # pending and active
+        self.live_free_cores_mcpu = 0
+        self.live_total_cores_mcpu = 0
+
+        self.task_manager = aiotools.BackgroundTaskManager()
+
+    async def async_init(self):
+        log.info('initializing instance monitor')
+
+        async for record in self.db.select_and_fetchall(
+                'SELECT * FROM instances WHERE removed = 0;'):
+            instance = Instance.from_record(self.app, record)
+            self.add_instance(instance)
+
+        log.info('finished initializing instance monitor')
+
+    async def run(self):
+        self.task_manager.ensure_future(self.event_loop())
+        self.task_manager.ensure_future(self.instance_monitoring_loop())
+
+    def shutdown(self):
+        self.task_manager.shutdown()
+
+    def unique_machine_name(self):
+        while True:
+            # 36 ** 5 = ~60M
+            suffix = secret_alnum_string(5, case='lower')
+            machine_name = f'{self.machine_name_prefix}{suffix}'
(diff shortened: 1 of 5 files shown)


batch/batch/driver/instance_monitor.py
```diff
@@ -0,0 +1,291 @@
...
+        assert instance in self.instances_by_last_updated
+
+        self.instances_by_last_updated.remove(instance)
+        self.n_instances_by_state[instance.state] -= 1
+
+        if instance.state in ('pending', 'active'):
+            self.live_free_cores_mcpu -= max(0, instance.free_cores_mcpu)
+            self.live_total_cores_mcpu -= instance.cores_mcpu
+
+        instance.inst_pool.adjust_for_remove_instance(instance)
```
jigold, line 93: This line is new.

batch/batch/driver/instance_monitor.py
```diff
@@ -0,0 +1,291 @@
...
+        assert instance not in self.instances_by_last_updated
+
+        self.instances_by_last_updated.add(instance)
+        self.n_instances_by_state[instance.state] += 1
+
+        if instance.state in ('pending', 'active'):
+            self.live_free_cores_mcpu += max(0, instance.free_cores_mcpu)
+            self.live_total_cores_mcpu += instance.cores_mcpu
+
+        instance.inst_pool.adjust_for_add_instance(instance)
```
jigold, line 114: This line is new

batch/batch/driver/instance_monitor.py
```diff
@@ -0,0 +1,291 @@
...
+        log.info('initializing instance monitor')
+
+        async for record in self.db.select_and_fetchall(
+                'SELECT * FROM instances WHERE removed = 0;'):
+            instance = Instance.from_record(self.app, record)
+            self.add_instance(instance)
+
+        log.info('finished initializing instance monitor')
+
+    async def run(self):
```
jigold, line 67: I had to split async_init into async_init and run because of circular dependencies between the instance monitor and the instance pool.

danking: I think the InstancePool (of which there are eventually many) should be a member of the InstanceMonitor.


We'll have an easier time working with this system if information flows consistently in one direction. I particularly find it confusing that instance pool's create_instance calls up to instance monitor who then calls back down to instance pool's adjust_for_add_instance.

I'm inclined for events to flow down:
```
... -> InstanceMonitor -> InstancePool
```
And information to flow up:
```
InstancePool -> InstanceMonitor -> ...
```

In particular, if we assign each instance pool a unique name prefix like `001` etc., then name generation can be handled entirely in the pool. When an event comes in, the name prefix can be used to dispatch the event to the correct instance pool for handling.

The instance monitor's live_free_cores_mcpu etc. can be calculated by summing the constituent pools on demand (there shouldn't be more than 10 or so pools per monitor anyway so this isn't a huge cost).

I guess as I pull the thread on this, I don't understand the essence of InstanceMonitor. Clearly we need someone who dispatches events to all the pools. We might call that job "instance event dispatcher".

There's also this monitoring loop, which feels rather self-contained and not related to anything else. It's just a sorted set and a loop that does something like:
```python3
while True:
    if self.instances_by_updated:
        instance = self.instances_by_updated[0]
        if instance.too_old():
             instance.inst_pool.check_on(instance)  # or maybe instance.check_in()
```

What do you think of all this?

danking (changes requested):

jigold:
I agree with the confusion. However, the reason for this structure is because the JobPrivate Instances can't be in a "pool". So we decided when we discussed this two weeks ago that the monitor needed to be centralized and do the tasks we defined: monitor instances and handle events.

danking:
The doc wasn't very clear about where information lived. I had imagined that the pools and the JobPrivateInstanceManager (JPIM) owned the information. Nit: the doc doesn't say the instance monitor monitors instances, it just monitors and handles *events*.

Let me be explicit: I think the doc is wrong about the monitor doing health checking because that requires it to track the instances, which I just said should be owned by the pools and the JPIM. That didn't occur to me when we were writing the doc, my apologies.

I still think the monitor should:
- route events to the right pool or to the JPIM, and
- aggregate summaries up for the web UI.

---

Let me try to be more specific in my critique:

I think of the system as three layers: the top most is the driver, the middle layer is the monitor, and the bottom layer is the pool or JobPrivateInstanceManager (JPIM). I don't want control flow to go down, up, and back down again. If that happens, then we can't reason about our system as separate layers, we necessarily have to think about the middle and bottom layer together.

Very specifically, this flow worries me: (instance pool) create_instance -> (instance monitor) add_instance -> adjust_for_add_instance -> (instance pool) adjust_for_add_instance. We move from low to mid *back to low*. I want information to flow in one direction: either its downward information or its upward information.

---

I'm guessing you're also concerned about code organization / code duplication. I'm not that worried about this. The JPIM and the Pool are similar things and we might inevitably produce some duplication. That's OK with me. To be honest, I think a few stand-alone functions that both of them use will eliminate any code duplication.

Both pools and the JPIM will have a `name_instances` and `instance_by_last_updated`. If the duplication gets hard to manage, we might pack that up into another class like InstanceCollection.

I realize this means we have several monitoring loops. I'm not very worried about that. I think it's fine and it helps simplify the architecture. It avoids entangling the monitor with the pools and the JPIM.

jigold:
Closing in favor of different design.

PR #9773, opened 2020-12-03
[query] Clarify matrix table distinct methods

johnc1231:
Clarified that matrix table distinct methods keep one value per key after a user asked if it deleted all rows that had a key duplicated somewhere or all but one of them.

### DIFF
--- a/hail/python/hail/matrixtable.py
+++ b/hail/python/hail/matrixtable.py
@@ -3948,7 +3948,7 @@ class MatrixTable(ExprContainer):
         return MatrixTable(ir.MatrixRename(self._mir, global_map, col_map, row_map, entry_map))
 
     def distinct_by_row(self) -> 'MatrixTable':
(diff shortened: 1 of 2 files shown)


hail/python/hail/matrixtable.py
```diff
@@ -3948,7 +3948,7 @@ def rename(self, fields: Dict[str, str]) -> 'MatrixTable':
         return MatrixTable(ir.MatrixRename(self._mir, global_map, col_map, row_map, entry_map))
 
     def distinct_by_row(self) -> 'MatrixTable':
-        """Remove rows with a duplicate row key.
+        """Remove rows with a duplicate row key, keeping only one row for each unique key.
```
patrick-schultz, line 3951: Maybe replace "only" with "exactly"?

patrick-schultz (changes requested):

patrick-schultz (approved):
Thanks John!

PR #9774, opened 2020-12-03
[batch] Refactor scheduler code for multiple instance pools

jigold:
Stacked on #9772 

This code accounts for a scheduler per instance pool even though right now there is only one instance pool. In the future when we add job private instance scheduling, then we'll need to rethink how this code is structured. For now, it should be fine.


jigold:
Also, I didn't want N bump loops all bumping the global events for scheduling and cancel events. I also have all schedulers share the same async worker pool.

jigold:
Closing for now.

PR #9775, opened 2020-12-03
[query/combiner] Fix MatrixMultiWrite to avoid metadata races

tpoterba:


chrisvittal (approved):

chrisvittal:
What are scala's defaults for `.par`? Number of cores?

tpoterba:
Yes. from scala.concurrent.ExecutionContext:

```
object ExecutionContext {
  /**
   * The explicit global `ExecutionContext`. Invoke `global` when you want to provide the global
   * `ExecutionContext` explicitly.
   *
   * The default `ExecutionContext` implementation is backed by a work-stealing thread pool.
   * It can be configured via the following [[scala.sys.SystemProperties]]:
   *
   * `scala.concurrent.context.minThreads` = defaults to "1"
   * `scala.concurrent.context.numThreads` = defaults to "x1" (i.e. the current number of available processors * 1)
   * `scala.concurrent.context.maxThreads` = defaults to "x1" (i.e. the current number of available processors * 1)
   * `scala.concurrent.context.maxExtraThreads` = defaults to "256"
   *
   * The pool size of threads is then `numThreads` bounded by `minThreads` on the lower end and `maxThreads` on the high end.
   *
   * The `maxExtraThreads` is the maximum number of extra threads to have at any given time to evade deadlock,
   * see [[scala.concurrent.BlockContext]].
   *
   * @return the global `ExecutionContext`
   */
  def global: ExecutionContextExecutor = Implicits.global.asInstanceOf[ExecutionContextExecutor]


```

PR #9778, opened 2020-12-03
[batch] remove unused worker logs bucket code

cseed:
Unused since worker logs go into GCP logging now


cseed:
And I'll remove the bucket after this goes in (it should be unused, but to be safe)

jigold (dismissed):

jigold:
I think you need to rip out more things in InstancePool.create

jigold (approved):

PR #9779, opened 2020-12-03
[batch] Add SQL infrastructure to support multiple pools

jigold:
Stacked on #9774 

This PR leaves the system in a state that has multiple pools in the SQL code, but the instance pool is hard coded as standard and there's only one of them. The global variables of local ssd, standing worker cores, etc. are still in globals and not per pool yet. This is because I need to write the code for the instance pool manager which will be in subsequent PRs so that we can configure each pool individually.

The goal of this PR is to make sure the incremental data structures are correct for supporting multiple pools.


jigold:
Closing for now.

PR #9781, opened 2020-12-03
[ci] better ci information on deploy failure

danking:
Both Cotton and I have PRs failing due to pod failures that are impossible to debug without this.


jigold (dismissed):
Do we need an example stateful set in the test_ci build.yaml?

jigold (approved):

PR #9786, opened 2020-12-04
[global] add batch regions to global-config

cseed:
... and use in batch instead  of hard-coded regions.  There a new global-config key batch_gpc_regions which is a JSON list of region names to use.


batch/batch/driver/instance_pool.py
```diff
@@ -207,10 +207,15 @@ async def create_instance(self, cores=None, max_idle_time_msecs=None):
                 break
 
         if self.live_total_cores_mcpu // 1000 < 4_000:
+            if not self.zone_monitor.init_zones:
+                return
+
             zone = random.choice(self.zone_monitor.init_zones)
```
danking, line 213: This is the only place init_zones is used and it doesn't do us any good any more because we won't have a list of zones on start up. Why even have init_zones any more?

danking (changes requested):

batch/batch/driver/instance_pool.py
cseed: init_zones is badly named.  It is the zones in the region where the k8s cluster/batch driver are running.  This logic is intended to avoid egress between other regions when the cluster is small.

There is actually a problem with this, inter-zone egress is the same as inter-region egress in the US.  So in fact this zone should just be the zone that the cluster is on.  I will make this change (and rename things to be a bit clearer).

cseed: OK, init_zones is gone.  We now deploy (up to 1K cores) into GCP_ZONE, and then stripe.  init_zones beyond GCP_ZONE wasn't actually helping because there is still inter-zone egress chargers within the same region.

danking (approved):

danking:
@jigold FYI, this changes how zones are handled, you probably want to rebase your pools PR

PR #9787, opened 2020-12-04
[batch] maybe fix cp dirs

danking:


jigold (changes requested):

batch/batch/worker/worker.py
```diff
@@ -542,16 +542,29 @@ def copy_command(src, dst, requester_pays_project=None):
...
         requester_pays_project = ''
 
+    def gsutil_cp_r(src, dst, * recursive):
+        flags = '-R' if recursive else ''
+        return f'retry gsutil {requester_pays_project} -m cp {flags} {shq(src)} {shq(dst)}'
+
     if not dst.startswith('gs://'):
-        mkdirs_file = f'mkdir -p {shq(os.path.dirname(dst))} && '
-        mkdirs_dir = f'mkdir -p {shq(dst)} && '
+        target_directory = os.path.dirname(dst)
```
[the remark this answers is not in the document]
jigold: This is specifically for the source. I'm worried about `gs://hail-jigold/`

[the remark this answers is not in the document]
jigold: Maybe not, actually. Forget this comment.

batch/batch/worker/worker.py
```diff
@@ -542,16 +542,29 @@ def copy_command(src, dst, requester_pays_project=None):
...
         requester_pays_project = ''
 
+    def gsutil_cp_r(src, dst, *, recursive):
+        flags = '-R' if recursive else ''
+        return f'retry gsutil {requester_pays_project} -m cp {flags} {shq(src)} {shq(dst)}'
+
     if not dst.startswith('gs://'):
-        mkdirs_file = f'mkdir -p {shq(os.path.dirname(dst))} && '
-        mkdirs_dir = f'mkdir -p {shq(dst)} && '
+        target_directory = os.path.dirname(dst)
```
jigold, line 550: Maybe this needs to be `os.path.dirname(dst.rstrip('/') + '/')`?

batch/batch/worker/worker.py
```diff
@@ -542,16 +542,29 @@ def copy_command(src, dst, requester_pays_project=None):
...
+        target_basename = os.path.basename(dst)
```
jigold, line 551: Same here. We are assuming that dst is a directory.

jigold (dismissed):

batch/batch/worker/worker.py
```diff
@@ -537,21 +537,35 @@ async def add_gcsfuse_bucket(mount_path, bucket, key_file, read_only):
 
 
 def copy_command(src, dst, requester_pays_project=None):
+    src = src.rstrip('/')  # ensure that the source_basename is always non-empty
```
danking, line 540: @jigold this was the bug. I believe it will work now.

jigold (approved):

PR #9788, opened 2020-12-04
[batch] eliminate non-determinism in test

danking:
stderr and stdout non-deterministically interleave, breaking the string match in `test_can_use_google_credentials`


jigold (approved):
It looks like test_query is failing. I tried looking at the batch logs and didn't see anything in the driver logs.

PR #9791, opened 2020-12-04
Add domain to the deploy config.

cseed:
Modify hailctl dev config to let you set the domain.  Note, this is an
interface change since I changed `hailctl dev config` to act like
gcloud/kubectl `... set property value`.

The hardest part of this was getting a doubly-nested subcommand to work in argprase.  The existing code is wack, but I will change everything to work like hailctl dev/dev config does below.

### DIFF
--- a/hail/python/hailtop/hailctl/dev/config/set_property.py
+++ b/hail/python/hailtop/hailctl/dev/config/set_property.py
@@ -7,7 +7,7 @@ from hailtop.config import get_deploy_config
 def init_parser(parser):
     parser.add_argument("property", type=str,
                         help="Property to set.",
-                        choices=['location', 'default', 'domain'])
+                        choices=['location', 'default_namespace', 'domain'])
     parser.add_argument("value", type=str,
                         help="Value to set property to.")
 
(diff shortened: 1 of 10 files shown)


hail/python/hailtop/hailctl/dev/config/set_property.py
```diff
@@ -0,0 +1,27 @@
...
+                        help="Value to set property to.")
+
+
+def main(args):
+    deploy_config = get_deploy_config()
+    config = deploy_config.get_config()
+
+    p = args.property
+    if p == 'default':
+        p = 'default_namespace'
```
danking, line 21: I don't like this, it's a confusing name and this is extra complexity. I'm pretty sure all of us have canned scripts / history lines for `hailctl dev config` anyway, let's go with `hailctl dev config set default_namespace ...`.

danking (changes requested):
One comment and this:

While verifying this I found a bad configuration in web_common:
```
../../../../web_common/web_common/templates/header.html:37:        <a class="header-dropdown-menu-link" href="https://hail.is/docs/batch/">Docs</a>
```
Can you fix that while you're in here?

hail/python/hailtop/hailctl/dev/config/set_property.py
cseed: Agreed, fixed.

cseed:
The web_common fix isn't totally trivial.  I will PR that separately.

danking (approved):

PR #9802, opened 2020-12-08
[web_common] fix hardcoded Batch docs URL

cseed:
The slight subtly was that the internal URL (internal.hail.is/namespace/service) requires a service, so you can't link to "hail.is".  I used www instead.

### DIFF
--- a/hail/python/hailtop/config/deploy_config.py
+++ b/hail/python/hailtop/config/deploy_config.py
@@ -101,6 +101,8 @@ class DeployConfig:
     def external_url(self, service, path, base_scheme='http'):
         ns = self.service_ns(service)
         if ns == 'default':
+            if service == 'www':
+                return f'{base_scheme}s://{self._domain}{path}'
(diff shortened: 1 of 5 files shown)


hail/python/hailtop/config/deploy_config.py
```diff
@@ -90,6 +90,8 @@ def auth_session_cookie_name(self):
     def external_url(self, service, path, base_scheme='http'):
         ns = self.service_ns(service)
         if ns == 'default':
+            if service == 'site' or service == 'www':
```
danking, line 93: Pylint is gonna suggest `in`.

Any reason to have the alias? I prefer the one name.

danking (changes requested):

hail/python/hailtop/config/deploy_config.py
cseed: OK, I use www everywhere, but left the site deployment and service named site. The router nginx config is the one that maps www to the site service.

danking (approved):

PR #9804, opened 2020-12-08
[query] Update pandas to 1.1.4

johnc1231:
Let's see if we can update to newest pandas.

### DIFF
--- a/hail/python/requirements.txt
+++ b/hail/python/requirements.txt
@@ -10,7 +10,7 @@ humanize==1.0.0
 hurry.filesize==0.9
 nest_asyncio
 numpy<2
-pandas>=1.1.0,<1.2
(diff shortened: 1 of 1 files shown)


johnc1231:
For future upgraders, there's something about pandas 1.1.5 that causes `pylint` to fail. It seems like underlying ast processing library `astroid` enters infinite recursive loop when pandas 1.1.5 is installed.

chrisvittal:
Has the bug been reported upstream to pandas, pylint, or asteroid? If no, we should.

chrisvittal (approved):

johnc1231:
I haven't yet Chris, but `astroid` also hasn't released in ~4 months, so I still have to build and check if it's fixed yet. Planning on doing so when I have some time.

PR #9806, opened 2020-12-08
[dev-docs] information about creating a developer account

danking:
As far as I can tell, there is no way to completely automate the creation of
a developer namespace. I document why and how to manually do so here.

cc: @jigold


danking:
bump @cseed

danking:
bump

danking:
@jigold I realized if I want reviews I shouldn't send it to cotton 😉

jigold (approved):

PR #9810, opened 2020-12-09
[LSM] Add Bloom Filter

Dania-Abuhijleh:


lsm/main.cpp
```diff
@@ -6,30 +6,169 @@
...
+        istrm.read(reinterpret_cast<char *>(&v), sizeof v);
+        istrm.read(reinterpret_cast<char *>(&d), sizeof d);
+        keys.push_back(k);
+      }
+    } else {
+      std::cerr << "could not open " << filename << "\n";
+      exit(3);
+    }
+    return keys;
+  }
```
danking, line 57: Let's avoid writing the file and then reading it back to get the keys. Change `write_to_file` to return a `File`. Inside `write_to_file`, create a bloom filter and `insert_key` each key before you write it to the file. Then change the constructor of `File` to take a `BloomFilter` in addition to the `filename`.

I think you don't need `get_keys` after this change.

lsm/main.cpp
```diff
@@ -6,30 +6,169 @@
...
+  void put(int32_t k, int32_t v, char deleted = 0) {
+    if (m.size() >= 4) {
+      std::string filename;
+      if (files.empty()) {
+        filename = "0";
+      } else {
+        filename = std::to_string(std::stoi(files.back().filename) + 1);
+      }
+      write_to_file(filename);
+      files.push_back(File(filename));
```
danking, line 81: with the change above this would be:

```
files.push_back(write_to_file(filename));
```

lsm/main.cpp
```diff
@@ -6,30 +6,169 @@
...
-  void put(int32_t k, int32_t v) {
-    m.insert(std::make_pair(k,v));
+  void put(int32_t k, int32_t v, char deleted = 0) {
+    if (m.size() >= 4) {
+      std::string filename;
+      if (files.empty()) {
+        filename = "0";
+      } else {
+        filename = std::to_string(std::stoi(files.back().filename) + 1);
+      }
```
danking, line 79: instead:
```
std::string filename = std::to_string(files.size())
```

lsm/main.cpp
```diff
@@ -6,30 +6,169 @@
...
+      write_to_file(filename);
+      files.push_back(File(filename));
+      m.clear();
+      m.insert_or_assign(k,maybe_value(v, deleted));
+    } else {
+      m.insert_or_assign(k,maybe_value(v, deleted));
```
danking, line 85: remove the `else` branch and move the `m.insert_or_assign(...)` outside of the if.

lsm/main.cpp
```diff
@@ -6,30 +6,169 @@
...
+    }
   }
   std::optional<int32_t> get(int32_t k) {
     auto it = m.find(k);
-    if (it != m.end()) {
-        return it->second;
+
+    if (it != m.end() && !it->second.is_deleted) {
+      return it->second.v;
+    } else {
```
danking, line 93: I think if you add this test you'll notice there is a bug here:
```
p 1 3
p 10 1
p 11 1
p 12 1
d 1
g 1
```

This should print an empty line because there's the key 1 has been deleted.

lsm/main.cpp
```diff
@@ -6,30 +6,169 @@
...
+      for (auto file : files) {
+        if (file.bloomFilter.contains_key(k) && k >= file.min && k <= file.max) {
+          std::map<int32_t, maybe_value> file_map = read_from_file(file.filename);
+          auto it_m = file_map.find(k);
+          if (it_m != m.end() && !it_m->second.is_deleted) {
+            return it_m->second.v;
+          }
```
danking, line 100: You iterate from oldest file to newest file, so I think this will print 3. It should print an empty line.
```
p 1 3
p 10 1
p 11 1
p 12 1
d 1
p 13 1
p 14 1
p 15 1
p 16 1
g 1
```

lsm/main.cpp
```diff
@@ -6,30 +6,169 @@
...
+
+    for (auto file : files) {
+      if (r >= file.min && l <= file.max) {
+        std::map<int32_t, maybe_value> file_map = read_from_file(file.filename);
+        auto it_ml = file_map.lower_bound(l);
+        auto it_mu = file_map.lower_bound(r);
+        for (auto it=it_ml; it!=it_mu; ++it) {
+          if (!it->second.is_deleted) {
+            res_map.insert_or_assign(it->first, it->second.v);
+          }
```
danking, line 118: Similar to the get case, I think this will reveal a bug:

```
p 1 3
p 10 1
p 11 1
p 12 1
d 1
p 13 1
p 14 1
p 15 1
p 16 1
r 0 9
```

lsm/main.cpp
```diff
@@ -6,30 +6,169 @@
...
+        istrm.read(reinterpret_cast<char *>(&d), sizeof d);
+        new_m.insert_or_assign(k, maybe_value(v,d));
+      }
+    } else {
+      std::cerr << "could not open " << filename << "\n";
+      exit(3);
+    }
+    return new_m;
+  }
+  int dump_map() {
```
danking, line 167: Great addition! I love this. Make it `void` with no `return` though.

lsm/test.sh
```diff
@@ -64,3 +64,83 @@ cat > build/expected <<EOF
...
+r 1 6
+EOF
+cat > build/expected <<EOF
+1:9 2:8 3:7
+2
+5
+6
+7
+1:9 2:99 3:7
+1:9 2:99 3:7 4:6 5:5
```
danking, line 143: I love all the new tests!

danking (changes requested):

lsm/main.cpp
```diff
@@ -6,30 +6,169 @@
 #include <vector>
 #include <fstream>
```
danking, line 7: add `#include <bitset>`

danking (dismissed):

lsm/main.cpp
```diff
@@ -5,31 +5,158 @@
...
+          }
+        }
+      }
+    }
+
     auto it_l = m.lower_bound(l);
     auto it_u = m.lower_bound(r);
     for (auto it=it_l; it!=it_u; ++it) {
-        res.push_back(std::make_pair(it->first, it->second));
+      res_map.insert_or_assign(it->first, it->second.v);
```
danking, line 109: There can be deletes in the in-memory buffer as well:
```
p 1 3
p 10 1
p 11 1
p 12 1
d 1
r 0 9
```
should be an empty line

danking: In fact, I think this test will fail:
```
p 1 3
d 1
r 0 9
```
should be an empty line

lsm/main.cpp
```diff
@@ -5,31 +5,158 @@
...
+
   void del(int32_t k) {
-    m.erase(k);
+    put(k ,0, 1);
+  }
+  File write_to_file(std::string filename) {
+    std::ofstream ostrm(filename, std::ios::binary);
+    BloomFilter bloomFilter;
+    int min = std::numeric_limits<int>::max();
+    int max = std::numeric_limits<int>::lowest();
```
danking, line 128: Good use of the the standard library!

danking (changes requested):

danking (approved):

PR #9816, opened 2020-12-10
[hail] include the diagonal in PCRelate

danking:
CHANGELOG: `hl.pc_relate` now includes values on the diagonal of kinship, IBD-0, IBD-1, and IBD-2


tpoterba:
how would this work with maximal_independent_set?

danking:
Ugh. I added a keyword parameter that defaults to the old behavior.

tpoterba:
Yeah, annoying. We should probably be boxing this up so that most users never actually use maximal_independent_set but just `prune_related_samples(mt, ...)` and we can do what we want in there.

tpoterba:
Hail is 1-2 levels of abstraction too low for most users

danking:
One of us should do one boring genetics project with someone and just publish whatever we write during that process.

danking:
bump @chrisvittal

chrisvittal (approved):

olavurmortensen:
Awesome, thanks guys!

PR #9817, opened 2020-12-11
[utils] added url_basename and url_join

cseed:
This is part of the copy tool changes.

### DIFF
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -535,12 +535,12 @@ class LoggingTimer:
             log.info(f'{self.description} timing {self.timing}')
 
 
-def url_basename(url: str):
+def url_basename(url: str) -> str:
     """Return the basename of the path of the URL `url`."""
     return os.path.basename(urllib.parse.urlparse(url).path)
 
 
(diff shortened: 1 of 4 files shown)


hail/python/hailtop/utils/utils.py
```diff
@@ -532,3 +533,12 @@ async def __aexit__(self, exc_type, exc, tb):
...
             self.timing['total'] = total
             log.info(f'{self.description} timing {self.timing}')
+
+
+def url_basename(url):
+    return os.path.basename(urllib.parse.urlparse(url).path)
+
+
+def url_join(url, path):
+    parsed = urllib.parse.urlparse(url)
```
jigold, line 543: Why did you choose not to just use `urlib.parse.urljoin()`?

jigold (changes requested):

hail/python/hailtop/utils/utils.py
cseed: urljoin takes a pair of URLs, url_join takes a url and a path.  I updated the docs to clarify.

jigold (approved):

PR #9818, opened 2020-12-11
[aiotools, aiogoogle] AsyncFS.listfiles raises FileNotFoundError if the directory doesn't exist

cseed:
Existence of directories is filesystem dependent, and they exist on Google if there is an object with that directory name a prefix.

This required a little refactoring that changed fs.listfiles to return a coroutine that gives an iterator, rather than returning an iterator directly.  This is because, if you use the `async def foo ... yield ...` syntax, it is not possible to write any code that will run between calling the async genreator function `foo` and the first call to `__anext__`.

### DIFF
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -256,7 +256,7 @@ class GoogleStorageAsyncFS(AsyncFS):
         try:
             n = await it.__anext__()
         except StopAsyncIteration:
-            raise FileNotFoundError(url)
+            raise FileNotFoundError(url)  # pylint: disable=raise-missing-from
 
         async def cons(n, it):
             yield n
@@ -291,7 +291,7 @@ class GoogleStorageAsyncFS(AsyncFS):
             prefixes = page.get('prefixes')
             items = page.get('items')
             return prefixes or items
-        raise UnreachableError()
+        assert False  # unreachable
 
     async def remove(self, url: str) -> None:
         bucket, name = self._get_bucket_name(url)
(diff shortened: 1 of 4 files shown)


hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -242,13 +242,31 @@ async def _listfiles_flat(self, bucket: str, name: str) -> AsyncIterator[FileLis
...
         if recursive:
-            return self._listfiles_recursive(bucket, name)
-        return self._listfiles_flat(bucket, name)
+            it = self._listfiles_recursive(bucket, name)
+        else:
+            it = self._listfiles_flat(bucket, name)
+
+        it = it.__aiter__()
+        try:
+            n = await it.__anext__()
```
jigold, line 257: I thought `n` was a number. Can you change the variable name?

jigold (changes requested):

hail/python/hailtop/aiogoogle/client/storage_client.py
cseed: Done.

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -254,12 +254,12 @@ async def listfiles(self, url: str, recursive: bool = False) -> AsyncIterator[Fi
...
         it = it.__aiter__()
         try:
-            n = await it.__anext__()
+            first_entry = await it.__anext__()
         except StopAsyncIteration:
             raise FileNotFoundError(url)  # pylint: disable=raise-missing-from
 
         async def cons(n, it):
-            yield n
+            yield first_entry
```
jigold, line 262: This isn't quite right. The cons takes an argument that should be first_entry. Can you rename `n` to `head` and it's invoked `cons(first_entry, it)`?

jigold (changes requested):

hail/python/hailtop/aiogoogle/client/storage_client.py
cseed: Sorry about that, fixed.

jigold (approved):

PR #9819, opened 2020-12-11
[hail] relax pandas requirement

danking:
Pandas only releases [breaking changes in major versions](https://pandas.pydata.org/docs/development/policies.html). It seems safe
to be flexible on patch version. Just today I had an issue where I ran `pip install pandas` to
upgrade from an old pandas version and I landed on 1.1.5.

### DIFF
--- a/hail/python/dev-requirements.txt
+++ b/hail/python/dev-requirements.txt
@@ -1,6 +1,6 @@
 flake8==3.7.8
 mypy==0.780
-pylint==2.5.3
+pylint==2.6.0
 pytest==4.6.3
 pytest-html==1.20.0
 pytest-xdist==1.28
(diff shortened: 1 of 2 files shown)


danking:
cc: @johnc1231

tpoterba:
John had issues with 1.1.5. He's OOO today, but this was intentional.

johnc1231:
You'll note that `check_hail_37` is failing because of an infinite recursion bug. I believe it's pylint's fault through its `astroid` depenency, though I'm not sure what the change in pandas 1.1.5 is that causes this. Mentioned here: https://github.com/hail-is/hail/pull/9804

danking:
Remarkable. `hailtop` doesn't even import pandas and yet it triggers this error. I can run pylint on `hail` and it produces many errors, but does not crash.

johnc1231:
Yeah I'm stumped by this, was going to maybe look into this this week while listening to Broad retreat talks as a low priority thing. It's quite odd

PR #9824, opened 2020-12-15
[query] Clean up NDArrayEmitter / Start cleaning up PNDArray

johnc1231:
This PR updates a lot of NDArray code that was using the old `Code[T]` and emit triplet interface in favor of using code builders and `IEmitCode`. This is going to make it easier to update the `PNDArray` interface to not use `Code[Long]` everywhere, among other things. 

Before this PR, there existed `NDArrayEmitter`, which did the old thing, and `NDArrayEmitter2`, which was a prototype of a new emitter. 

After this PR, a tweaked version of `NDArrayEmitter2` has become the new `NDArrayEmitter`. `outputElement` now returns a `PCode`, and all the missingness problems are handled by carrying a `IEmitCodeGen[NDArrayEmitter]` around throughout the deforesting process, meaning the `NDArrayEmitter` no longer needs internal state about missingness. 

I think the diff for `Emit.scala` is going to be pretty confusing. I'd at least opt for a side by side view instead of the intermingled one, as I've mostly implemented the same logic, just on top of our new code builder primitives. 

All tests pass, but marking WIP since I'm noticing some slow down in the slice test that I want to look into. 

This PR also adds a `get` function to `EmitValue` that gets the underlying `PValue` and moves two functions off of the `PNDArray` interface into `LinalgCodeUtils`.


hail/src/main/scala/is/hail/asm4s/CodeBuilder.scala
```diff
@@ -33,14 +33,14 @@ trait CodeBuilderLike {
 
   def isOpenEnded: Boolean
 
-  // def code: Code[Unit] // debugging only
+   def code: Code[Unit] // debugging only
```
tpoterba, line 36: debug stuff here

johnc1231: Beat me to it by a second, it's gone now.

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
```diff
@@ -1032,6 +1045,16 @@ class EmitMethodBuilder[C](
             mb.getArg(codeIndex + i)(t).get
           }))
       }
+
+      override def get(cb: EmitCodeBuilder): PValue = {
```
tpoterba, line 1049: This seems wrong

hail/src/main/scala/is/hail/types/physical/PContainer.scala
```diff
@@ -89,7 +89,9 @@ abstract class PContainer extends PIterable {
   def nextElementAddress(currentOffset: Code[Long]): Code[Long]
 }
 
-abstract class PIndexableValue extends PValue with SIndexableValue
+abstract class PIndexableValue extends PValue with SIndexableValue {
+  def pt: PContainer
```
tpoterba, line 93: Where is this used? We should be moving away from using ptypes on PCodes/PValues in favor of STypes or methods on the codes/values themselves

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
johnc1231: Agreed, clearly wrong. I didn't expect anyone to look at this as soon as I posted haha, I usually go through the diff myself while the tests are running and clean up anything I forgot about. Clearly I didn't finish implementing `get` on the different `EmitValue`s.

hail/src/main/scala/is/hail/types/physical/PContainer.scala
johnc1231: https://github.com/hail-is/hail/pull/9824/files#diff-5fcfe4833387843730db9b875d97b36742722849a88b4685797bed3b19a66ef5R2819

I just did this so I could call `elementType` on something I knew to be a `PArray` so I could call `nDims` on `PNDArray` type contained in the array.

johnc1231:
Unassigning Patrick until I clean up the `EmitValue` stuff Tim pointed out.

hail/src/main/scala/is/hail/types/physical/PContainer.scala
tpoterba: cool, thanks. I realized today that stypes are missing the virtual types. If SNDArray had `vt: TNDArray` you could get the ndims right?

johnc1231: If we put `elementType` on `PIndexableCode` that would remove the need for this.

tpoterba: awesome. I can make that PR tomorrow, does that work?

johnc1231: I definitely want to be able to get the number of dimensions easily if I'm holding an `SNDArray`. Whether that's by going through the virtual type or the `SNDArray` just has its own `nDims` field on it, definitely should be an option. 

Separately though, I want to be able to make a decision based on the element type of a regular array. I could have just done:

```
coerce[PArray](myArrayPValue.pt).elementType
```

but I just made this change to avoid the cast. Any way you would like me to get the number of dimensions of the ndarray type which is contained in this array is fine with me.

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
johnc1231: Fixed, I think.

hail/src/main/scala/is/hail/linalg/LinalgCodeUtils.scala
```diff
@@ -68,4 +68,40 @@ object LinalgCodeUtils {
...
+      Code.foreach(shapeArray.zip(newIndices)) { case (shapeElement, newIndex) =>
+        Code(
+          elementsInProcessedDimensions := elementsInProcessedDimensions / shapeElement,
+          newIndex := workRemaining / elementsInProcessedDimensions,
+          workRemaining := workRemaining % elementsInProcessedDimensions
+        )
+      }
+    )
+    (createShape, newIndices)
+  }
```
johnc1231, line 106: Copy and pasted from `PCanonicalNDArray`

johnc1231:
Reassigning to Patrick now. Addressed Tim's comment and things are working, though I am investigating perceived slowness in slicing.

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
```diff
@@ -322,6 +322,16 @@ class EmitClassBuilder[C](
           cb.assign(ms, false)
           cb.assign(vs, value)
         })
+
+    override def get(cb: EmitCodeBuilder): PValue = {
+      if (_pt.required) {
+        vs
+      } else {
+        // TODO: I don't think this is right. Seems to get an open ended issue instead of a fatal error.
+        cb.ifx(ms, cb._fatal(s"Can't convert missing ${_pt} to PValue"))
```
johnc1231, line 331: At some point I got an open ended error from this code, but after some changes I think I stopped hitting this code path. It may be that the open ended error I briefly saw from here was a result of some other issue somewhere else, and that it was resolved. Just wanted to flag this with a TODO to see if you thought it looked right.

hail/src/main/scala/is/hail/types/physical/PContainer.scala
johnc1231: I can remove this change and do what I want without any other changes, but I still think it would be helpful in future to be able to get element type of arrays.

johnc1231:
Bump @patrick-schultz

patrick-schultz (dismissed):
Looks good, just a few minor nits. Have you resolved the performance issue?

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -283,6 +318,10 @@ case class IEmitCodeGen[+A](Lmissing: CodeLabel, Lpresent: CodeLabel, value: A)
   def get(cb: EmitCodeBuilder, errorMsg: String = "expected non-missing"): A =
     handle(cb, cb._fatal(errorMsg))
 
+  def getNeverMissing(cb: EmitCodeBuilder): A = {
+    handle(cb, cb._fatal(s"Internal compiler error, unexpected value was missing."))
+  }
```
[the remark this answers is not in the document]
johnc1231: No, this is dumb. I don't know why I didn't implement it in terms of `get`, and the only reason I added it in the first place is because I wanted to stop typing out error messages. But `get` has a default message which handles this. I'll remove.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -242,6 +268,15 @@ object IEmitCodeGen {
     def memoize(cb: EmitCodeBuilder, name: String): EmitValue =
       cb.memoize(iec, name)
   }
+
+  def apply[B](cb: EmitCodeBuilder, m: Code[Boolean], value: => B): IEmitCodeGen[B] = {
```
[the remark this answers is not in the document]
johnc1231: I just kept the one in `IEmitCode`. I think I just didn't realize that it was also general.

johnc1231:
This should be all set now I think

patrick-schultz (approved):

PR #9825, opened 2020-12-15
[query]: transition CodeOrdering to PCode/EmitCode/CodeBuilder

chrisvittal:
This is not this interface's final form, but it is an important
transitionary step in moving the entire code generation codebase away
from Code[T].

The CodeOrdering interface now (as before) has 12 methods, but before
the function signatures looked like:

    def compare(lhs: (Code[Boolean], Code[T]), rhs: (Code[Boolean], Code[T])) -> Code[Int]

They now look like:

    def compare(cb: EmitCodeBuilder, lhs: EmitCode, rhs: EmitCode) -> Code[Int]

There have been a lot of miscellaneous changes needed to get this to
work. There's some stuff that needs to be cleaned up, notably in Emit,
it's not always the case that the type of the IR/Aggregator is the same
as the type of the EmitCode that emit produces. I've tried to add
assertions where possible, but if there is an error, then a rather
cryptic 'Cannot pop value off of empty stack' exception is thrown,
generally indicating that an ordering expected an optional value, but
a required emitcode was provided.


tpoterba:
Woohoo!

chrisvittal:
I think this builds. Gotta make it pass tests.

chrisvittal:
Don't think all the tests pass yet, but its single digit python tests failing now.

hail/src/main/scala/is/hail/expr/ir/agg/GroupedAggregator.scala
```diff
@@ -16,19 +16,22 @@ class GroupedBTreeKey(kt: PType, kb: EmitClassBuilder[_], region: Value[Region],
     "regionIdx" -> PInt32(true),
     "container" -> states.storageType)
   val compType: PType = kt
+  private val kcomp = kb.getCodeOrdering(kt, CodeOrdering.Compare(), ignoreMissingness = false)
```
tpoterba, line 19: this is incorrect, I think.

tpoterba (changes requested):

hail/src/main/scala/is/hail/annotations/CodeOrdering.scala
```diff
@@ -43,22 +44,10 @@ object CodeOrdering {
...
-    def setup(i: Int)(x: Value[Long], y: Value[Long]): Code[Unit] = {
-      val tf1 = t1.types(i)
-      val tf2 = t2.types(i)
-      Code(
-        m1 := t1.isFieldMissing(x, i),
-        m2 := t2.isFieldMissing(y, i),
-        v1s(i).storeAny(m1.mux(defaultValue(tf1), Region.loadIRIntermediate(tf1)(t1.fieldOffset(x, i)))),
-        v2s(i).storeAny(m2.mux(defaultValue(tf2), Region.loadIRIntermediate(tf2)(t2.fieldOffset(y, i)))))
+
+    private val setup: (EmitCodeBuilder, PCode, PCode) => (PBaseStructValue, PBaseStructValue) = {
```
tpoterba, line 48: this'll probably look a bit simpler as a `def`

hail/src/main/scala/is/hail/annotations/CodeOrdering.scala
```diff
@@ -68,294 +57,518 @@ object CodeOrdering {
         if (sortOrders == null) Ascending else sortOrders(i),
         op)
 
-    override def compareNonnull(x: Code[Long], y: Code[Long]): Code[Int] = {
-      val cmp = mb.newLocal[Int]()
+    override def compareNonnull(cb: EmitCodeBuilder, x: PCode, y: PCode): Code[Int] = {
```
tpoterba, line 60: why do we need to implement all the orderings? Can't we implement them all in terms of `compareNonNull`?

hail/src/main/scala/is/hail/annotations/CodeOrdering.scala
```diff
@@ -68,294 +57,518 @@ object CodeOrdering {
...
+      val lhsv = lhs.asIndexable.memoize(cb, "container_ord_lhs")
+      val rhsv = rhs.asIndexable.memoize(cb, "container_ord_rhs")
+      lhsv -> rhsv
     }
 
-    override def compareNonnull(x: Code[Long], y: Code[Long]): Code[Int] = {
-      val mbcmp = mb.getCodeOrdering(t1.elementType, t2.elementType, CodeOrdering.Compare())
-      val cmp = mb.newLocal[Int]()
+    private[this] def loop(cb: EmitCodeBuilder, lhs: PIndexableValue, rhs: PIndexableValue)(
+      f: (EmitCode, EmitCode) => Unit
```
tpoterba, line 199: Here we take `cb` as a parameter to `loop`, and `f` doesn't take a `cb` -- it appends to another cb, implicitly assuming they're the same. I think it's much safer to always have functions like `f` here take a cb, so it's easier to locally interpret the code generated.

hail/src/main/scala/is/hail/annotations/CodeOrdering.scala
```diff
@@ -368,80 +581,133 @@ object CodeOrdering {
...
 
-  def gtNonnull(x: Code[T], y: Code[T]): Code[Boolean]
+  def gteqNonnull(cb: EmitCodeBuilder, x: PCode, y: PCode): Code[Boolean]
 
-  def gteqNonnull(x: Code[T], y: Code[T]): Code[Boolean]
+  def equivNonnull(cb: EmitCodeBuilder, x: PCode, y: PCode): Code[Boolean]
 
-  def equivNonnull(x: Code[T], y: Code[T]): Code[Boolean]
+  def compare(cb: EmitCodeBuilder, x: EmitCode, y: EmitCode, missingEqual: Boolean = true): Code[Int] = {
+    cb += x.setup
```
tpoterba, line 597: This is old-style code -- just use `x.toI` and `y.toI`

tpoterba (changes requested):

hail/src/main/scala/is/hail/annotations/CodeOrdering.scala
chrisvittal: Yes. It is correct to do so, but if the comparison is significantly more expensive than `lt` or `eq` then we end up doing extra work.

tpoterba (approved):
:shipit:

PR #9826, opened 2020-12-16
[k8s] set most resource requests to 100mCPU

danking:
I've already changed ukbb-rg.

### DIFF
--- a/internal-gateway/deployment.yaml
+++ b/internal-gateway/deployment.yaml
@@ -37,8 +37,8 @@ spec:
          image: "{{ internal_gateway_image.image }}"
          resources:
            requests:
-             memory: 375Mi
-             cpu: 100m
+             memory: "250M"
+             cpu: "200m"
            limits:
              memory: 3750Mi
              cpu: "1"
--- a/internal-gateway/deployment.yaml
+++ b/internal-gateway/deployment.yaml
@@ -37,8 +37,11 @@ spec:
          image: "{{ internal_gateway_image.image }}"
          resources:
            requests:
-             memory: "250M"
-             cpu: "200m"
+             memory: 375Mi
+             cpu: 100m
+           limits:
+             memory: 3750Mi
+             cpu: "1"
          ports:
           - containerPort: 80
          volumeMounts:
--- a/shuffler/deployment.yaml
+++ b/shuffler/deployment.yaml
@@ -48,10 +48,10 @@ spec:
              readOnly: true
           resources:
             requests:
-              memory: "3.75G"
-              cpu: "800m"
+              memory: 375Mi
+              cpu: 100m
             limits:
-              memory: "3.75G"
+              memory: 3750Mi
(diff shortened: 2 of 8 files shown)


danking:
I also already applied the internal-gateway changes.

internal-gateway/deployment.yaml
```diff
@@ -37,8 +37,11 @@ spec:
          image: "{{ internal_gateway_image.image }}"
          resources:
            requests:
-             memory: "250M"
-             cpu: "200m"
+             memory: 375Mi
+             cpu: 100m
+           limits:
+             memory: 3750Mi
+             cpu: "1"
```
cseed, line 44: OK, this is slightly complicated and can also impact the router rate limiting.  internal-gateway has (1) request rate limits, and (2) is fixed at 3 replicas (no horizontal scaling) to limit the max incoming batch-driver request rate from workers.  I think 200m was tuned as the observed CPU when we're at the max rate limit.  So, again, this will reduce our max scheduling rate.  Might be safer to leave.

shuffler/deployment.yaml
```diff
@@ -48,10 +48,10 @@ spec:
              readOnly: true
           resources:
             requests:
-              memory: "3.75G"
-              cpu: "800m"
+              memory: 375Mi
```
cseed, line 51: This seems tight.  Shuffler is currently at 343Mi.

ukbb-rg/deployment.yaml
```diff
@@ -37,10 +37,10 @@ spec:
          imagePullPolicy: Always
          resources:
            requests:
-             memory: "3.75G"
-             cpu: "800m"
+             memory: 375Mi
```
cseed, line 40: Too low?  Currently:

```
NAME                CPU(cores)   MEMORY(bytes)   
ukbb-rg-browser-0   1m           1657Mi          
ukbb-rg-static-0    0m           2454Mi          
```

cseed (changes requested):

internal-gateway/deployment.yaml
danking: ok

shuffler/deployment.yaml
danking: It seems we should set the request to the idle number. If a node is overloaded, my understanding, is that k8s will kill off and reschedule pods exceeding their requests. Ideally this moves the shuffler to a less loaded node.

If we see high turn over for shuffler, I'll adjust the requests then.

ukbb-rg/deployment.yaml
danking: heh. I'm trying to have a set memory:cpu ratio to simplify how I think about resources. I just set the requests to 3750Mi / 1cpu. ukbb-rg is an unfortunate memory hog, I think we just accept that. It only runs once in default, so not so bad.

danking:
bump

PR #9828, opened 2020-12-16
[k8s] reduce batch request & reduce namespaced batch,auth,router cpu …

danking:
…request

### DIFF
--- a/auth/deployment.yaml
+++ b/auth/deployment.yaml
@@ -81,10 +81,15 @@ spec:
                 key: gsuite_organization
          resources:
            requests:
-             memory: "250M"
-             cpu: "100m"
+{% if deploy %}
+             memory: 375Mi
+             cpu: 100m
+{% else %}
+             memory: 37Mi
+             cpu: 10m
+{% endif %}
            limits:
-             memory: "1G"
+             memory: 3750Mi
              cpu: "1"
          volumeMounts:
           - name: deploy-config
@@ -191,10 +196,15 @@ spec:
                 key: gsuite_organization
          resources:
            requests:
-             memory: "250M"
-             cpu: "100m"
+{% if deploy %}
+             memory: 375Mi
+             cpu: 100m
+{% else %}
+             memory: 37Mi
+             cpu: 10m
+{% endif %}
            limits:
-             memory: "1G"
+             memory: 3750Mi
              cpu: "1"
          volumeMounts:
           - name: deploy-config
--- a/auth/deployment.yaml
+++ b/auth/deployment.yaml
@@ -85,8 +85,8 @@ spec:
              memory: 375Mi
              cpu: 100m
 {% else %}
-             memory: 37Mi
-             cpu: 10m
+             memory: 75Mi
+             cpu: 20m
 {% endif %}
            limits:
              memory: 3750Mi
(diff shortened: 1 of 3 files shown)


auth/deployment.yaml
```diff
@@ -81,10 +81,15 @@ spec:
                 key: gsuite_organization
          resources:
            requests:
-             memory: "250M"
-             cpu: "100m"
+{% if deploy %}
+             memory: 375Mi
+             cpu: 100m
+{% else %}
+             memory: 37Mi
```
cseed, line 88: Looking at `k top pods`, the current batch-driver usage is 51Mi (but auth is 37Mi).  Can we leave in a little slack?  Maybe make it 100Mi for auth-driver and 50Mi for auth?

batch/deployment.yaml
```diff
@@ -42,9 +42,15 @@ spec:
...
-            memory: "1G"
-            cpu: "1"
+{% if deploy %}
+           requests:
+             memory: 375Mi
+             cpu: 100m
+{% else %}
+           requests:
+             memory: 37Mi
+             cpu: 10m
```
cseed, line 52: Should you add limit here?  Can we boost the memory?  batch-driver is currently running at 86Mi and it will definitely need more memory when there are many instances, since the instance data structures are stored in memory.

Also, batch-driver is bursty and definitely uses more than 10m when it is running.  Again, my question about requests stands.  If batch-driver gets packed into a 10mCPU slot because of this and it is trying to schedule a massive batch, will it peg at 10m and nothing happens?  batch-driver isn't horizontally scaled.  That seems bad.  It seems like batch-driver request also limits or scheduling rate.

batch/deployment.yaml
```diff
@@ -236,12 +242,17 @@ spec:
...
-            cpu: "100m"
-          limits:
-            memory: "1G"
-            cpu: "1"
+           requests:
+{% if deploy %}
+             memory: 375Mi
+             cpu: 100m
+{% else %}
+             memory: 37Mi
```
cseed, line 250: Again, 37Mi is too small for some currently running batch services.

cseed (changes requested):

auth/deployment.yaml
danking: Sure. I set batch, batch-driver, auth, and auth-driver, in non-deploy-mode, to 20m / 75Mi. (I'm aiming for a set cpu:mem ratio so as to simplify how I think about resources)

batch/deployment.yaml
danking: No, that's not how requests work. The scheduler only sees requests, so batch-driver might get packed in with lots of other pods. batch-driver is allowed to exceed its request (eating into the requests of other pods) *as long as the node has resources to spare*. If, however, the node has no spare resources (say, it is full and everyone is using 100% of its request, or it is not full, but a few people are exceeding their requests significantly), then it will kill & reschedule pods that are exceeding their requests.

"peg" behavior is only created by a resource limit.

Now, what I'm not sure is how the autoscaler relates to this. For this system to work, we need the auto-scaler to select the number of nodes based on *actual* use not *request* or *limit*.

I'm inclined to take this aggressive step and then adjust based on our experience. I think it's clear that even after these changes we'll have *lots* of unused CPU (because most pods will use below their request).

danking: I'll add a limit though.

danking:
bump

danking:
bump

PR #9830, opened 2020-12-16
[k8s] set admin pod request to 1m

danking:


admin-pod/admin-pod.yaml
```diff
@@ -40,6 +40,13 @@ spec:
           - name: ssl-config-admin-pod
             mountPath: /ssl-config
             readOnly: true
+         resources:
+           requests:
+             memory: 3Mi
```
cseed, line 45: I feel like we should increase this, very unlikely stuff we run on admin pod will fit in 3Mi, and it would be annoying to get killed/restarted if you're, say, debugging a production issue using admin-pod.

I'm also slightly confused about the requests behavior.  Say you request 3Mi, get scheduled on a node with 10Mi free, and run a process that requires 20Mi.  What happens?  I think you get killed, but can then be rescheduled in the same slot that was too small.  Is that right?  That seems bad for the admin pod!  I say we make it 1G.

cseed (changes requested):

danking:
bump

PR #9831, opened 2020-12-16
[lsm] Merge files in levels T=2

Dania-Abuhijleh:


lsm/lsm.cpp
```diff
@@ -1,26 +1,93 @@
...
+#include <limits>
+#include <bitset>
+#include "MurmurHash3.h"
 
 void BloomFilter::insert_key(int32_t k) {
-  auto key_hash = std::hash<int32_t>{}(k);
-  auto last_d = key_hash % 10;
-  bset[last_d] = 1;
+    uint32_t seed = 1;
+    uint64_t key_hash;
```
danking, line 13: murmuhash3_x86_32 accepts a pointer to a 32-bit integer as its `out`, not a 64-bit integer.

lsm/lsm.cpp
```diff
@@ -1,26 +1,93 @@
...
+    auto last_d = key_hash % 10;
+    bset[last_d] = 1;
 }
-
 char BloomFilter::contains_key(int32_t k) {
-  auto key_hash = std::hash<int32_t>{}(k);
-  auto last_d = key_hash % 10;
-  return bset[last_d];
+    uint32_t seed = 1;
+    uint64_t key_hash;
```
danking, line 20: same here

lsm/lsm.cpp
```diff
@@ -1,26 +1,93 @@
...
+    MurmurHash3_x86_32(&k, sizeof k, seed, &key_hash);
+    auto last_d = key_hash % 10;
+    return bset[last_d];
+}
+int Level::get_size() {
```
danking, line 25: just call this `size`; hail team generally avoids the get/set terminology

lsm/lsm.cpp
```diff
@@ -1,26 +1,93 @@
...
+  }
+}
+File Level::merge(File older_f, File newer_f, std::string merged_filename) {
+  std::map<int32_t, maybe_value> m;
+  read_to_map(older_f, m);
+  read_to_map(newer_f, m);
+  return write_to_file(m, merged_filename);
 }
 
+void LSM::add_to_level(File f, int l_index) {
```
danking, line 68: I think if you make `l_index` have type `size_t`, everything will work without the cast.

lsm/lsm.cpp
```diff
@@ -1,26 +1,93 @@
...
+  if ((size_t)l_index >= levels.size()) {
+    Level l = Level();
+    l.add_file(f);
+    levels.push_back(l);
+  } else if (levels[l_index].get_size() + 1 >= levels[l_index].max_size) {
+    File merged_f = levels[l_index].merge(levels[l_index].files.back(), f, "foo" );
+    add_to_level(merged_f, l_index + 1);
+    levels[l_index].files.pop_back();
```
danking, line 76: this block only really works when max_size == 2, I think. If max_size == 4, for example, you'd want to merge `f` with all three files in `.files` not just the last one. Given this limitation, I'm inclined to add an assert to the beginning of this else if block:
```cpp
assert(levels[l_index].max_size == 2)
```

lsm/lsm.cpp
```diff
@@ -1,26 +1,93 @@
...
+  }
+}
 void LSM::put(int32_t k, int32_t v, char deleted) {
   if (m.size() >= 4) {
     std::string filename = directory / std::to_string(files.size());
-    files.push_back(write_to_file(filename));
+    File f = write_to_file(filename);
+    files.push_back(f);
+    add_to_level(f, 0);
     m.clear();
```
danking, line 87: Let's add an assertion here that loops through the levels and verifies the number of files in each level is less than the maximum number of files.

lsm/lsm.cpp
```diff
@@ -83,14 +149,11 @@ std::vector<std::pair<int32_t, int32_t>> LSM::range(int32_t l, int32_t r) {
   for (auto it=rit_l; it!=rit_u; ++it) {
     res.push_back(std::make_pair(it->first, it->second));
   }
-
   return res;
 }
-
 void LSM::del(int32_t k) {
-  put(k, 0, 1);
+  put(k ,0, 1);
```
danking, line 155: Prefer the old code format

lsm/lsm.cpp
```diff
@@ -1,26 +1,93 @@
...
+  } else {
+    levels[l_index].add_file(f);
+  }
+}
 void LSM::put(int32_t k, int32_t v, char deleted) {
   if (m.size() >= 4) {
     std::string filename = directory / std::to_string(files.size());
-    files.push_back(write_to_file(filename));
+    File f = write_to_file(filename);
+    files.push_back(f);
```
danking, line 85: Do you still need LSM.files? Each file is tracked by the level now. You can have the Level class manage file names. If each level knows its index in `levels` then it could return new file names as: `directory / std::to_string(my_level_index) / std::to_string(files.size())`

danking: You can also use this in place of `"foo"` above when you need a merged filename.

It's OK to re-use filenames within a given level. As soon as you call `levels[l_index].files.pop_back` that file is meaningless to you.

lsm/lsm.cpp
danking, line 76: Let's delete the file as well, take a look at [`std::filesystem`](https://en.cppreference.com/w/cpp/header/filesystem).

lsm/test/integration-tests.sh
```diff
@@ -246,3 +246,153 @@ cat > build/expected <<EOF
...
+310
+150
+
+88
+
+
+1:310 2:22 3:33 4:44 5:150 6:66 7:77 9:99
+EOF
+diff build/out build/expected || failure
+success
```
danking, line 398: Let's add more test cases where a `delete` ends up in a file.

lsm/test/simple.cpp
```diff
@@ -9,3 +9,53 @@ TEST_CASE("Simple LSM test", "") {
...
+  m_newer.insert_or_assign(5, maybe_value(13, 0));
+  m_newer.insert_or_assign(98, maybe_value(107, 0));
+  File f_older = l.write_to_file(m_older, "older_file");
+  File f_newer = l.write_to_file(m_newer, "newer_file");
+  File f_merged = l.merge(f_older, f_newer, "merged_file");
+  std::map<int32_t, maybe_value> m_merged;
+  l.read_to_map(f_merged, m_merged);
+
+  for (auto const&x : m_merged) {
+    std::cout << x.first << "\n";
```
danking, line 34: Instead of printing the keys, very `m_merged.get()` returns what you expect on several (all?) inputs.

lsm/test/simple.cpp
```diff
@@ -9,3 +9,53 @@ TEST_CASE("Simple LSM test", "") {
...
+  m_newer.insert_or_assign(10, maybe_value(13, 0));
+  m_newer.insert_or_assign(4, maybe_value(0, 1));
+  File f_older = l.write_to_file(m_older, "older_file");
+  File f_newer = l.write_to_file(m_newer, "newer_file");
+  File f_merged = l.merge(f_older, f_newer, "merged_file");
+  std::map<int32_t, maybe_value> m_merged;
+  l.read_to_map(f_merged, m_merged);
+
+  for (auto const&x : m_merged) {
+    std::cout << x.first << ":" << x.second.v << "\n";
```
danking, line 59: same here

lsm/test/simple.cpp
```diff
@@ -9,3 +9,53 @@ TEST_CASE("Simple LSM test", "") {
...
+  }
+}
```
danking, line 61: add a test where thee two files have overlapping keys with both deletes and puts.

danking (changes requested):

lsm/MurmurHash3.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  ((uint32_t*)out)[0] = h1;
+  ((uint32_t*)out)[1] = h2;
+  ((uint32_t*)out)[2] = h3;
+  ((uint32_t*)out)[3] = h4;
+}
+
+//-----------------------------------------------------------------------------
+
+void MurmurHash3_x64_128 ( const void * key, const int len,
```
danking, line 255: just delete this method since we're not using it.

lsm/MurmurHash3.cpp
```diff
@@ -0,0 +1,334 @@
...
+    case  7: k2 ^= tail[ 6] << 16;
+    case  6: k2 ^= tail[ 5] << 8;
+    case  5: k2 ^= tail[ 4] << 0;
+      k2 *= c2; k2  = ROTL32(k2,16); k2 *= c3; h2 ^= k2;
+
+    case  4: k1 ^= tail[ 3] << 24;
+    case  3: k1 ^= tail[ 2] << 16;
+    case  2: k1 ^= tail[ 1] << 8;
+    case  1: k1 ^= tail[ 0] << 0;
+      k1 *= c1; k1  = ROTL32(k1,15); k1 *= c2; h1 ^= k1;
```
danking, line 228: For each one of these `case`s, modify them so that the last line before the next `case` is `[[fallthrough]];`, e.g.:
```cpp
    case  3: k1 ^= tail[ 2] << 16; [[fallthrough]];
    case  2: k1 ^= tail[ 1] << 8; [[fallthrough]];
    case  1: k1 ^= tail[ 0] << 0;
      k1 *= c1; k1  = ROTL32(k1,15); k1 *= c2; h1 ^= k1;
```
You don't need it on the last one, I believe.

PR #9832, opened 2020-12-16
[batch] Add instance collections

jigold:
This PR introduces a couple of new concepts:
- InstanceCollection which is a generic class that keeps track of instances in its collection
- Pool which is a shared, growable pool of instances with a control loop and fair share scheduler and is a subclass of InstanceCollection
- InstanceCollectionManager which keeps track of all instance collections.

The cancel code remains the same and operates globally and is now in canceller.py.

The GCE event monitor is in gce.py.

The original instance_pool.py code has been split amongst zone_monitor.py, pool.py, gce.py, create_instance.py, and instance_collection.py

The scheduler code is now in PoolScheduler in pool.py

The SQL code has vectorized user_resources by instance_collection as well as batch_cancellable_resources and batches_staging. There are also two new tables: inst_colls and pools. Each job and instance must belong to an instance collection noted by the field `inst_coll`. The job_update trigger had to be updated to insert into user_resources to the correct pool. The cancel_batch and close_batch functions changed to vectorize by instance collection. I deleted the global `ready_cores` table.

The front end code does not change except looking for a `worker_type` field in the resources field of the job spec (default if undefined is standard). I added a PoolSelector class which is overkill for now, but will be used in the future for more complicated scenarios.

There was an issue with our existing code for converting between memory in bytes to memory in MB in the worker_config.py code for the `resources()` function. For the highcpu case, it is impossible for the memory in bytes to be divisible by 1024**2. The utils.py code now rounds up bytes using math.ceil.

The hailtop.batch library adds a `worker_type` method on Job. I didn't change the interface significantly at this time as I think this is fine for now. More significant changes will come when we change how cpu and memory and storage are interpreted by the worker.

I added 3 new tests: we spin up a highmem and highcpu machine in test_batch.py. In the hailtop.batch tests, I check the `worker_type` interface.

The driver UI page has changed such that there's a list of pools with links to a pools.html page, which gives information on the instances in the pool, the configuration updates, and the current user resources for that pool.

Things to double check:
- SQL code especially cancel_batch where I had to use temporary variables and recompute_incremental


jigold:
One thing I wasn't sure about was whether there should be a global async worker pool or whether each scheduler and the canceller can have their own.

batch/batch/driver/main.py
```diff
@@ -496,31 +534,33 @@ async def check(tx):
...
+        for u_ic in user_inst_colls:
             for f in fields:
-                v = user_get(user_resources, u, f)
-                computed_v = user_get(user_resources, u, f)
+                v = user_inst_coll_get(user_inst_coll_resources, u_ic, f)
+                computed_v = user_inst_coll_get(user_inst_coll_resources, u_ic, f)
 
                 if v != computed_v:
-                    log.error(f'user_resources corrupt: user_resources[{u}][{f}] {v} != computed_user_resources[{u}][{f}] {computed_v}')
+                    log.error(f'user_inst_coll_resources corrupt: user_inst_coll_resources[{u_ic}][{f}] {v} != computed_user_inst_coll_resources[{u_ic}][{f}] {computed_v}')
```
danking, line 563: I find this python code kind of hard to follow. What do you think of just using the database to do this? It also saves a round trip and transferring unnecessary data.

We can avoid the `COALESCE` because there are no NULLs in these field (they're all `NOT NULL`). We don't need `IF` because booleans can be used as `1` (TRUE) and `0` (FALSE).
```python
async def check_incremental_loop(app, db):
    async def check():
        users_with_broken_resources = db.execute_and_fetchall('''
SELECT
  t.*,
  u.*
FROM
(
  SELECT user,
    SUM(state = 'Ready' AND runnable) AS actual_n_ready_jobs,
    SUM(cores_mcpu * (state = 'Ready' AND runnable)) AS actual_ready_cores_mcpu,
    SUM(state = 'Running' AND (NOT cancelled)) AS actual_n_running_jobs,
    SUM(cores_mcpu * (state = 'Running' AND (NOT cancelled))) AS actual_running_cores_mcpu,
    SUM(state = 'Ready' AND cancelled) AS actual_n_cancelled_ready_jobs,
    SUM(state = 'Running' AND cancelled) AS actual_n_cancelled_running_jobs
  FROM
  (
    SELECT batches.user, jobs.state, jobs.cores_mcpu,
      (jobs.always_run OR NOT (jobs.cancelled OR batches.cancelled)) AS runnable,
      (NOT jobs.always_run AND (jobs.cancelled OR batches.cancelled)) AS cancelled
    FROM batches
    INNER JOIN jobs ON batches.id = jobs.batch_id
    WHERE batches.`state` = 'running'
  ) as v
  group by user
) as t
INNER JOIN
(
  SELECT user, inst_coll,
    SUM(n_ready_jobs) as expected_n_ready_jobs,
    SUM(ready_cores_mcpu) as expected_ready_cores_mcpu,
    SUM(n_running_jobs) as expected_n_running_jobs,
    SUM(running_cores_mcpu) as expected_running_cores_mcpu,
    SUM(n_cancelled_ready_jobs) as expected_n_cancelled_ready_jobs,
    SUM(n_cancelled_running_jobs) as expected_n_cancelled_running_jobs
  from user_inst_coll_resources
  group by user, inst_coll
) as u
on t.user = u.user
WHERE actual_n_ready_jobs != expected_n_ready_jobs
   OR actual_ready_cores_mcpu != expected_ready_cores_mcpu
   OR actual_n_running_jobs != expected_n_running_jobs
   OR actual_running_cores_mcpu != expected_running_cores_mcpu
   OR actual_n_cancelled_ready_jobs != expected_n_cancelled_ready_jobs
   OR actual_n_cancelled_running_jobs != expected_n_cancelled_running_jobs
;
''')
        async for record in users_with_broken_resources:
            log.error(f'user_inst_coll_resources corrupt: {record}')

    while True:
        try:
            await check()  # pylint: disable=no-value-for-parameter
        except Exception as e:
            app['check_incremental_error'] = serialization.exception_to_dict(e)
            log.exception('while checking incremental')
        await asyncio.sleep(10)
```

batch/batch/driver/gce.py
```diff
@@ -0,0 +1,147 @@
...
+                    await self.handle_event(event)
+
+                if new_mark is not None:
+                    await self.db.execute_update(
+                        'UPDATE `gevents_mark` SET mark = %s;',
+                        (new_mark,))
+            except asyncio.CancelledError:  # pylint: disable=try-except-raise
+                raise
+            except Exception:
+                log.exception('in event loop')
```
danking, line 146: why prefer explicit try and except to using `retry_long_running`?

batch/batch/driver/instance_collection_manager.py
```diff
@@ -0,0 +1,89 @@
...
+        self.app = app
+        self.db = app['db']
+        self.zone_monitor = app['zone_monitor']
+        self.machine_name_prefix = machine_name_prefix
+        self.inst_coll_regex = re.compile(f'{self.machine_name_prefix}(?P<inst_coll>[^-]+)-.*')
+
+        self.name_inst_coll = {}
+        self.name_pool = {}
+
+        self.task_manager = aiotools.BackgroundTaskManager()
```
danking, line 23: this isn't used

batch/batch/driver/instance_collection_manager.py
```diff
@@ -0,0 +1,89 @@
...
+            inst_coll = Pool(self.app, inst_coll_name, self.machine_name_prefix)
+            self.name_pool[inst_coll_name] = inst_coll
+            self.name_inst_coll[inst_coll_name] = inst_coll
+
+            await inst_coll.async_init()
+
+        log.info('finished initializing instance collections')
+
+    def shutdown(self):
+        self.task_manager.shutdown()
```
danking, line 44: All of the pools have `shutdown`s that need to be called here.

batch/batch/driver/instance_collection.py
```diff
@@ -0,0 +1,163 @@
...
+                    since_last_updated = time_msecs() - instance.last_updated
+                    if since_last_updated > 60 * 1000:
+                        log.info(f'checking on {instance}, last updated {since_last_updated / 1000}s ago')
+                        await self.check_on_instance(instance)
+            except asyncio.CancelledError:  # pylint: disable=try-except-raise
+                raise
+            except Exception:
+                log.exception('in monitor instances loop')
+
+            await asyncio.sleep(1)
```
danking, line 163: why not use `retry_long_running`?

danking (changes requested):
My plan is to roughly go from top to bottom starting with driver/main.py which I focused on in this pass.

batch/batch/driver/canceller.py
```diff
@@ -0,0 +1,198 @@
...
+        total = sum(user_n_cancelled_running_jobs.values())
+        if not total:
+            should_wait = True
+            return should_wait
+        user_share = {
+            user: max(int(300 * user_n_jobs / total + 0.5), 20)
+            for user, user_n_jobs in user_n_cancelled_running_jobs.items()
+        }
+
+        async def user_cancelled_running_jobs(user, remaining):
```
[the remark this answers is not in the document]
danking: This also avoids the `remaining` stuff in the same way.

danking (approved):
OK, I made some more progress, I'm done with canceller.py.

batch/batch/driver/gce.py
```diff
@@ -0,0 +1,147 @@
...
+    async def handle_delete_done_event(self, instance, timestamp):
+        await instance.inst_coll.remove_instance(instance, 'deleted', timestamp)
+
+    async def handle_call_delete_event(self, instance, timestamp):
+        await instance.mark_deleted('deleted', timestamp)
+
+    async def handle_event(self, event):
+        payload = event.get('protoPayload')
+        if payload is None:
+            log.warning('event has no payload')
```
danking, line 53: I think whenever something goes wrong handling the event, we should include the event dict in the message. I remember debugging some issue with workers a while ago and it was quite frustrating that the bad event wasn't included in the error message.

batch/batch/driver/gce.py
```diff
@@ -0,0 +1,147 @@
...
+                await self.handle_preempt_event(instance, timestamp)
+            elif event_subtype == 'v1.compute.instances.delete':
+                if event_type == 'COMPLETED':
+                    log.info(f'event handler: delete {instance} done')
+                    await self.handle_delete_done_event(instance, timestamp)
+                elif event_type == 'STARTED':
+                    log.info(f'event handler: handle call delete {instance}')
+                    await self.handle_call_delete_event(instance, timestamp)
+
+    async def event_loop(self):
```
danking, line 108: I notice that we have this pattern of
```python
async def VERB_CLAUSE_loop(...):
    while True:
        ....
        await asyncio.sleep(TIME)
```
For example: scheduling cancel bump loop, monitor billing limits loop, the worker file cache clean up, the worker disk space monitor, and update region quotas.

I feel like this would be more clear with a new utility function. Perhaps `periodically_call(f, *, period, name)`. I think it's implementation is:
```python
async def periodically_call(period, f, *args, **kwargs):
    async def loop():
        while True:
            await f(*args, **kwargs)
            await asyncio.sleep(period)
    await retry_long_running(f.__name__, loop)
```

This separates the periodicity from the code. We'll see things like:
```
    self.background_tasks.ensure_future(periodically_call(10, self.process_event))
```

A little background: I'm searching for some unifying themes here. I don't like that we have to go hunting across the entire batch codebase to see what periodic loops we have an how often they run. I don't know the full answer yet, but I think this moves in the right direction.

danking: It also reduces the volume of indentation in the meaningful code, which I like.

batch/batch/driver/gce.py
```diff
@@ -0,0 +1,147 @@
...
+                        (mark,))
+
+                filter = f'''
+logName="projects/{PROJECT}/logs/cloudaudit.googleapis.com%2Factivity" AND
+resource.type=gce_instance AND
+protoPayload.resourceName:"{self.machine_name_prefix}" AND
+timestamp >= "{mark}"
+'''
+
+                new_mark = None
```
danking, line 127: We don't need `new_mark` and `mark`, I think it's more clear to just use `mark` throughout. That way `mark` always the point in time to which we have processed.

Notably, we update `new_mark` *before* we handle an event, so if handling an event triggers an exception, we'll ignore that event. I suppose that's the right thing though!

danking (changes requested):
OK, I'm good with gce.py

batch/batch/driver/pool.py
```diff
@@ -0,0 +1,469 @@
...
+        self.task_manager.ensure_future(self.control_loop())
+
+        async def scheduler_relay():
+            self.scheduler_state_changed.set()
+            return True
+
+        self.task_manager.ensure_future(retry_long_running(
+            'scheduler_state_changed_relay_loop',
+            run_if_changed, self.global_scheduler_state_changed,
+            scheduler_relay))
```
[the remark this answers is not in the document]
danking: In fact, I don't think we need `should_wait`. We can just set the event.

batch/batch/driver/instance_collection.py
```diff
@@ -0,0 +1,163 @@
...
+
+log = logging.getLogger('inst_collection')
+
+
+class InstanceCollection:
+    def __init__(self, app, name, machine_name_prefix):
+        self.app = app
+        self.db = app['db']
+        self.compute_client = app['compute_client']
+        self.zone_monitor = app['zone_monitor']
```
danking, line 17: let's put type annotations on anything we take out of `app`.

batch/batch/driver/instance_collection.py
```diff
@@ -0,0 +1,163 @@
...
+
+from hailtop.utils import time_msecs, secret_alnum_string
+from hailtop import aiotools
+
+log = logging.getLogger('inst_collection')
+
+
+class InstanceCollection:
+    def __init__(self, app, name, machine_name_prefix):
+        self.app = app
```
danking, line 14: not used

batch/batch/driver/pool.py
```diff
@@ -0,0 +1,469 @@
...
+
+    def adjust_for_remove_instance(self, instance):
+        super().adjust_for_remove_instance(instance)
+        if instance in self.healthy_instances_by_free_cores:
+            self.healthy_instances_by_free_cores.remove(instance)
+
+    def adjust_for_add_instance(self, instance):
+        super().adjust_for_add_instance(instance)
+        if (instance.state == 'active'
+                and instance.failed_request_count <= 1):
```
danking, line 143: weird indentation here. Maybe but `and` on the previous line?

batch/batch/driver/pool.py
```diff
@@ -0,0 +1,469 @@
...
+                              worker_pd_ssd_data_disk_size_gb=self.worker_pd_ssd_data_disk_size_gb,
+                              worker_disk_size_gb=self.worker_disk_size_gb)
+
+    async def control_loop(self):
+        log.info(f'starting control loop for {self}')
+        while True:
+            try:
+                ready_cores = await self.db.select_and_fetchone(
+                    '''
+SELECT CAST(COALESCE(SUM(ready_cores_mcpu), 0) AS SIGNED) AS ready_cores_mcpu
```
danking, line 179: We should not need the cast or coalesce. ready_cores_mcpu is a not null.

batch/batch/driver/pool.py
```diff
@@ -0,0 +1,469 @@
...
+
+                n_live_instances = self.n_instances_by_state['pending'] + self.n_instances_by_state['active']
+                if (ENABLE_STANDING_WORKER
+                        and self.standing_worker
+                        and n_live_instances == 0
+                        and self.max_instances > 0):
+                    await self.create_instance(cores=self.standing_worker_cores,
+                                               max_idle_time_msecs=STANDING_WORKER_MAX_IDLE_TIME_MSECS)
+            except asyncio.CancelledError:  # pylint: disable=try-except-raise
+                raise
```
danking, line 223: we do this pretty much everywhere we have manually written these loops. I think we should also do this in `retry_long_running`.

batch/batch/driver/pool.py
```diff
@@ -0,0 +1,469 @@
...
+        allocating_users_by_total_cores = sortedcontainers.SortedSet(
+            key=lambda user: user_total_cores_mcpu[user])
+
+        records = self.db.execute_and_fetchall(
+            '''
+SELECT user,
+  CAST(COALESCE(SUM(n_ready_jobs), 0) AS SIGNED) AS n_ready_jobs,
+  CAST(COALESCE(SUM(ready_cores_mcpu), 0) AS SIGNED) AS ready_cores_mcpu,
+  CAST(COALESCE(SUM(n_running_jobs), 0) AS SIGNED) AS n_running_jobs,
+  CAST(COALESCE(SUM(running_cores_mcpu), 0) AS SIGNED) AS running_cores_mcpu
```
danking, line 288: I'm pretty sure these are all NOT NULL.

batch/batch/driver/pool.py
```diff
@@ -0,0 +1,469 @@
...
+                    record['format_version'] = batch['format_version']
+                    yield record
+                if not batch['cancelled']:
+                    async for record in self.db.select_and_fetchall(
+                            '''
+SELECT job_id, spec, cores_mcpu
+FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_cancelled)
+WHERE batch_id = %s AND state = 'Ready' AND always_run = 0 AND inst_coll = %s AND cancelled = 0
+LIMIT %s;
+''',
```
danking, line 394: I think these can be one query as well

```
    select jobs.job_id, jobs.batch_id, userdata, user, format_version
      from batches
inner join jobs on jobs.batch_id = batches.batch_id
     where batches.user = %s
       and batches.`state` = 'running'
       and jobs.state = 'ready'
       and (always_run or not jobs.cancelled)
       and inst_coll = %s
     limit %s;

```

danking (changes requested):
Making progress through pool and instance collection manager.

batch/batch/driver/canceller.py
```diff
@@ -0,0 +1,198 @@
...
+''',
+                    (user,),
+                    timer_description=f'in cancel_cancelled_ready_jobs: get {user} running batches'):
+                if batch['cancelled']:
+                    async for record in self.db.select_and_fetchall(
+                            '''
+SELECT jobs.job_id
+FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_cancelled)
+WHERE batch_id = %s AND state = 'Ready' AND always_run = 0
+LIMIT %s;
```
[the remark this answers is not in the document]
jigold: Cotton had to add this last year because the query wasn't using the correct index. I'm not sure what would have changed that this is no longer the case.

batch/batch/driver/canceller.py
```diff
@@ -0,0 +1,198 @@
...
+                            timer_description=f'in cancel_cancelled_ready_jobs: get {user} batch {batch["id"]} ready cancelled jobs (2)'):
+                        record['batch_id'] = batch['id']
+                        yield record
+
+        waitable_pool = WaitableSharedPool(self.async_worker_pool)
+
+        should_wait = True
+        for user, share in user_share.items():
+            remaining = Box(share)
+            async for record in user_cancelled_ready_jobs(user, remaining):
```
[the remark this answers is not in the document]
jigold: I'm still worried about performance and why Cotton didn't do this to begin with. I tried your new query and this was the plan:

```
mysql> EXPLAIN SELECT batch_id, job_id FROM batches INNER JOIN jobs ON batches.id = jobs.batch_id WHERE batches.user = 'jigold' AND batches.state = 'running' AND jobs.state = 'Ready' AND jobs.always_run = FALSE AND (jobs.cancelled OR batches.cancelled) LIMIT 5;
+----+-------------+---------+------------+------+-----------------------------------------------------------------------------------------------------+------------------------------------------+---------+-------------------------------+------+----------+--------------------------+
| id | select_type | table   | partitions | type | possible_keys                                                                                       | key                                      | key_len | ref                           | rows | filtered | Extra                    |
+----+-------------+---------+------------+------+-----------------------------------------------------------------------------------------------------+------------------------------------------+---------+-------------------------------+------+----------+--------------------------+
|  1 | SIMPLE      | batches | NULL       | ref  | PRIMARY,batches_user_state_cancelled                                                                | batches_user_state_cancelled             | 424     | const,const                   |    1 |   100.00 | Using index              |
|  1 | SIMPLE      | jobs    | NULL       | ref  | PRIMARY,jobs_batch_id_state_always_run_cancelled,jobs_batch_id_state_always_run_inst_coll_cancelled | jobs_batch_id_state_always_run_cancelled | 131     | jigold.batches.id,const,const |   24 |   100.00 | Using where; Using index |
+----+-------------+---------+------------+------+-----------------------------------------------------------------------------------------------------+------------------------------------------+---------+-------------------------------+------+----------+--------------------------+
2 rows in set, 1 warning (0.00 sec)
```

batch/batch/driver/canceller.py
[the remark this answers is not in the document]
jigold: I checked the query plan:

```
+----+-------------+----------+------------+------+-----------------------------------------------------------------------------------------------------+------------------------------------------+---------+--------------------------------------+------+----------+--------------------------+
| id | select_type | table    | partitions | type | possible_keys                                                                                       | key                                      | key_len | ref                                  | rows | filtered | Extra                    |
+----+-------------+----------+------------+------+-----------------------------------------------------------------------------------------------------+------------------------------------------+---------+--------------------------------------+------+----------+--------------------------+
|  1 | SIMPLE      | batches  | NULL       | ref  | PRIMARY,batches_user_state_cancelled                                                                | batches_user_state_cancelled             | 424     | const,const                          |    1 |   100.00 | Using index              |
|  1 | SIMPLE      | jobs     | NULL       | ref  | PRIMARY,jobs_batch_id_state_always_run_cancelled,jobs_batch_id_state_always_run_inst_coll_cancelled | jobs_batch_id_state_always_run_cancelled | 130     | jigold.batches.id,const              |   31 |     1.00 | Using where; Using index |
|  1 | SIMPLE      | attempts | NULL       | ref  | PRIMARY                                                                                             | PRIMARY                                  | 12      | jigold.batches.id,jigold.jobs.job_id |    1 |   100.00 | NULL                     |
+----+-------------+----------+------------+------+-----------------------------------------------------------------------------------------------------+------------------------------------------+---------+--------------------------------------+------+----------+--------------------------+
3 rows in set, 1 warning (0.00 sec)
```

batch/batch/driver/pool.py
jigold: I think that indentation is correct, but I moved the and to the line above.

batch/batch/driver/pool.py
jigold: I disagree and am going to strongly push back here! If there are no rows that match, then SUM() will return NULL. The SUM() doesn't return the right type which is why we had to CAST it. We can try looking for the original PRs where these were added.

batch/batch/driver/pool.py
jigold: See comment above.

batch/batch/driver/pool.py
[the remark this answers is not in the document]
jigold: I implemented the Notice. The run_if_changed code won't work with Python 3.7. I'm going to leave the refactoring run_if_changed to a separate project.

batch/batch/driver/canceller.py
```diff
@@ -0,0 +1,198 @@
...
+
+    async def cancel_cancelled_ready_jobs_loop_body(self):
+        records = self.db.select_and_fetchall(
+            '''
+SELECT user, n_cancelled_ready_jobs
+FROM (SELECT user,
+    CAST(COALESCE(SUM(n_cancelled_ready_jobs), 0) AS SIGNED) AS n_cancelled_ready_jobs
+  FROM user_inst_coll_resources
+  GROUP BY user) AS t
+WHERE n_cancelled_ready_jobs > 0;
```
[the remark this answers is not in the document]
jigold: See previous comment on why this is necessary.

batch/batch/driver/pool.py
jigold: Here's the query plan:

```
mysql> EXPLAIN SELECT jobs.batch_id, jobs.job_id, userdata, user, format_version,
    ->   spec, cores_mcpu
    -> FROM batches
    -> INNER JOIN jobs ON batches.id = jobs.batch_id
    -> WHERE batches.user = 'jigold'
    ->   AND batches.state = 'running'
    ->   AND jobs.state = 'Ready'
    ->   AND (jobs.always_run OR NOT jobs.cancelled)
    ->   AND inst_coll = 'standard'
    -> LIMIT 4;
+----+-------------+---------+------------+------+-----------------------------------------------------------------------------------------------------+------------------------------+---------+-------------------+------+----------+-------------+
| id | select_type | table   | partitions | type | possible_keys                                                                                       | key                          | key_len | ref               | rows | filtered | Extra       |
+----+-------------+---------+------------+------+-----------------------------------------------------------------------------------------------------+------------------------------+---------+-------------------+------+----------+-------------+
|  1 | SIMPLE      | batches | NULL       | ref  | PRIMARY,batches_user_state_cancelled                                                                | batches_user_state_cancelled | 424     | const,const       |    1 |   100.00 | NULL        |
|  1 | SIMPLE      | jobs    | NULL       | ref  | PRIMARY,jobs_batch_id_state_always_run_cancelled,jobs_batch_id_state_always_run_inst_coll_cancelled | PRIMARY                      | 8       | jigold.batches.id |   73 |     0.91 | Using where |
+----+-------------+---------+------------+------+-----------------------------------------------------------------------------------------------------+------------------------------+---------+-------------------+------+----------+-------------+
2 rows in set, 1 warning (0.00 sec)
```

batch/batch/driver/pool.py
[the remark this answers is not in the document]
danking: Ah, I should have wrote `asyncio.wait(`

batch/batch/driver/pool.py
danking: Ugh, this grinds my gears.

batch/batch/driver/pool.py
danking: You're indeed right about the behavior above, which sends me into fits of rage, but here the `GROUP BY` ensures there are no empty sets of rows so these values always have a well-defined value.
```
mysql> drop temporary table if exists foo  ; create temporary table foo (user varchar(100) not null, n_ready_jobs int(11) not null , primary key(user));
Query OK, 0 rows affected, 1 warning (0.00 sec)

Query OK, 0 rows affected (0.00 sec)

mysql> select sum(n_ready_jobs) from foo;
+-------------------+
| sum(n_ready_jobs) |
+-------------------+
|              NULL |
+-------------------+
1 row in set (0.01 sec)

mysql> select sum(n_ready_jobs) from foo group by user;
Empty set (0.01 sec)
```

danking: But I guess we can leave it if the uniformity seems good. I'm in such a fit of rage right now.

jigold:
Stacked on #9844

batch/batch/driver/instance_collection.py
```diff
@@ -0,0 +1,163 @@
...
+
+        if instance.state in ('pending', 'active'):
+            self.live_free_cores_mcpu -= max(0, instance.free_cores_mcpu)
+            self.live_total_cores_mcpu -= instance.cores_mcpu
+
+    async def remove_instance(self, instance, reason, timestamp=None):
+        await instance.deactivate(reason, timestamp)
+
+        await self.db.just_execute(
+            'UPDATE instances SET removed = 1 WHERE name = %s;', (instance.name,))
```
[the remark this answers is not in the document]
jigold: This should be part of a plan to do cleanup. Not in this PR.

batch/batch/driver/pool.py
jigold: pylint complains if I change this.

jigold:
Can you respond to my concerns in the initial commit message? Specifically some math questions when computing resources and the AsyncWorkerPool usage and the temp variable in the SQL code.

batch/batch/driver/canceller.py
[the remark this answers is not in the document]
danking: That looks to me like as good a plan as you could have. It doesn't even hit the main tables, in each case an index has all the information we need.

jigold:
I didn't change the run_if_changed because I didn't understand what you were saying. We can do that later.

jigold:
I think I understand the run_if_changed now. I'll add that.

jigold:
ah. It was one or the other of Notice or changing run_if_changed to accept multiple events. I like the notice.

batch/batch/utils.py
```diff
@@ -69,15 +69,15 @@ def worker_memory_per_core_gb(worker_type):
 
 def worker_memory_per_core_bytes(worker_type):
     m = worker_memory_per_core_gb(worker_type)
-    return int(m * 1024**3)
+    return math.ceil(m * 1024**3)
```
danking, line 72: The smallest billable unit of memory we use in our system is MiB, so it doesn't make sense to talk about finer gradations of memory. For example, the quantity in the resources table is an integer, so we can't possibly bill fractions of an MiB.

We should define all memory in terms of this. We avoid ceilings and floating point math generally. We only need to convert to Bytes for Docker when we create a container specification.
```python3
def worker_memory_per_core_mib(worker_type):
    if worker_type == 'standard':
        return 3840
    elif worker_type == 'highmem':
        return 6656
    assert worker_type == 'highcpu', worker_type
    return 921
```

danking (changes requested):

batch/batch/driver/canceller.py
```diff
@@ -0,0 +1,171 @@
...
+log = logging.getLogger('canceller')
+
+
+class Canceller:
+    def __init__(self, app):
+        self.app = app
+        self.cancel_ready_state_changed: asyncio.Event = app['cancel_ready_state_changed']
+        self.cancel_running_state_changed: asyncio.Event = app['cancel_running_state_changed']
+        self.db: Database = app['db']
+        self.async_worker_pool = AsyncWorkerPool(parallelism=100, queue_size=100)
```
danking, line 20: This means we allow as many as 200 parallel threads of control because there are two distinct pools each with 100 workers. I don't know how many parallel threads is OK, but this is definitely a change from the old code. Seems prudent to keep it the same by sharing a worker pool between canceller and scheduler.

danking (changes requested):

danking:
SQL seems fine to me

batch/batch/utils.py
jigold: Why did you decide on 921 instead of 922?

jigold:
@danking Can you fix this? I don't quite know what the issue is other than a double import from the same file.

```
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.7/dist-packages/batch/front_end/__main__.py", line 5, in <module>
    from .front_end import run  # noqa: E402 pylint: disable=wrong-import-position
  File "/usr/local/lib/python3.7/dist-packages/batch/front_end/front_end.py", line 43, in <module>
    from ..batch import batch_record_to_dict, job_record_to_dict, cancel_batch_in_db
  File "/usr/local/lib/python3.7/dist-packages/batch/batch.py", line 22, in <module>
    from .driver.instance_collection_manager import InstanceCollectionManager  # pylint: disable=cyclic-import
  File "/usr/local/lib/python3.7/dist-packages/batch/driver/instance_collection_manager.py", line 7, in <module>
    from .pool import Pool
  File "/usr/local/lib/python3.7/dist-packages/batch/driver/pool.py", line 14, in <module>
    from ..batch import schedule_job
ImportError: cannot import name 'schedule_job' from 'batch.batch' (/usr/local/lib/python3.7/dist-packages/batch/batch.py)
```

danking:
This is caused by the circularity in the import chain. Imports need to be a directed acyclic graph. The simplest fix seems to  be to define `schedule_job` in `pool.py`, the only place it is used.

jigold:
It won't just be there in pool.py when we have a job private instance collection. I'd prefer to keep that function in batch.py. I think we just take out the typing or we import the module in the function where it's used.

batch/batch/batch.py
```diff
@@ -256,261 +98,21 @@ def job_record_to_dict(record, name):
...
-        except Exception:
-            await instance.incr_failed_request_count()
-            raise
-
-        log.info(f'schedule job {id} on {instance}: called create job')
-
-        rv = await db.execute_and_fetchone(
-            '''
-CALL schedule_job(%s, %s, %s, %s);
+async def cancel_batch_in_db(db, batch_id, user):
```
jigold, line 101: This is just a reorganization of the functions in this file. There were no changes from the code on the left side of the diff.

batch/sql/add-inst-colls.sql
```diff
@@ -0,0 +1,433 @@
+CREATE TABLE IF NOT EXISTS `inst_colls` (
+  `name` VARCHAR(255) NOT NULL,
+  `pool` BOOLEAN NOT NULL,
+  `boot_disk_size_gb` BIGINT NOT NULL,
+  `max_instances` BIGINT NOT NULL,
```
jigold, line 5: I changed the schema of this because all inst_colls need a boot disk and a size.

batch/batch/batch.py
```diff
@@ -80,153 +69,6 @@ def _time_msecs_str(t):
...
-        else:
-            log.warning(f'mark_complete for job {id} from unknown {instance}')
-
-    await add_attempt_resources(db, batch_id, job_id, attempt_id, resources)
-
-    if rv['rc'] != 0:
-        log.info(f'mark_job_complete returned {rv} for job {id}')
-        return
-
-    old_state = rv['old_state']
```
jigold, line 196: All of these functions moved to driver/job.py

batch/batch/driver/main.py
```diff
@@ -393,140 +404,162 @@ def validate_int(name, value, predicate, description):
...
-                'Updated batch configuration.',
+                f'Updated configuration for {pool}.',
                 'info')
 
-    return web.HTTPFound(deploy_config.external_url('batch-driver', '/'))
+    return web.HTTPFound(deploy_config.external_url('batch-driver', pool_url_path))
 
 
-@routes.get('/user_resources')
+@routes.get('/inst_coll/{inst_coll}')
```
jigold, line 442: I changed this route name. For now, it's always a pool, but may not be in the future. In future PRs, I'll make this function render the correct page depending on the type of the Instance collection.

batch/batch/driver/main.py
```diff
@@ -393,140 +404,162 @@ def validate_int(name, value, predicate, description):
...
   CAST(COALESCE(SUM(ready_cores_mcpu), 0) AS SIGNED) AS ready_cores_mcpu,
   CAST(COALESCE(SUM(n_running_jobs), 0) AS SIGNED) AS n_running_jobs,
-  CAST(COALESCE(SUM(running_cores_mcpu), 0) AS SIGNED) AS running_cores_mcpu,
-  CAST(COALESCE(SUM(n_cancelled_ready_jobs), 0) AS SIGNED) AS n_cancelled_ready_jobs,
-  CAST(COALESCE(SUM(n_cancelled_running_jobs), 0) AS SIGNED) AS n_cancelled_running_jobs
-FROM user_resources
+  CAST(COALESCE(SUM(running_cores_mcpu), 0) AS SIGNED) AS running_cores_mcpu
+FROM user_inst_coll_resources
 GROUP BY user
-LOCK IN SHARE MODE;
```
jigold, line 480: I'm not sure if we need the LOCK IN SHARE MODE here anymore.

batch/sql/estimated-current.sql
```diff
@@ -940,8 +955,7 @@ BEGIN
   INTO cur_job_state, cur_cores_mcpu, cur_job_cancel
   FROM jobs
   INNER JOIN batches ON batches.id = jobs.batch_id
-  WHERE batch_id = in_batch_id AND batches.`state` = 'running'
-    AND job_id = in_job_id
+  WHERE batch_id = in_batch_id AND job_id = in_job_id
```
jigold, line 958: I changed this to fix a bug where MJS happened after a batch was cancelled.

danking (changes requested):

batch/batch/worker/worker.py
```diff
@@ -165,10 +166,12 @@ async def start_container(container):
     try:
         return await container.start()
     except DockerError as e:
+        log.exception('while starting container', exc_info=True)
```
danking, line 169: exc_info=True is the default for `.exception`, we need not specify it here.

danking (changes requested):

batch/batch/utils.py
```diff
@@ -56,20 +56,20 @@ def cost_from_msec_mcpu(msec_mcpu):
...
     if worker_type == 'standard':
-        m = 3.75
+        m = 3840
     elif worker_type == 'highmem':
-        m = 6.5
+        m = 6656
     else:
         assert worker_type == 'highcpu', worker_type
-        m = 0.9
+        m = 920  # this number must be divisible by 4. I rounded down to the nearest MiB
```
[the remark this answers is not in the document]
danking: 922 seems to do the trick, but its not divisible by 4. I don't understand the divisibility by 4 requirement, where does that come from?
```
In [1]: 0.9 * 1024 * 1024 * 1024                                                                                                                                                                                                                                                                                       
Out[1]: 966367641.6

In [2]: 921 * 1024 * 1024                                                                                                                                                                                                                                                                                              
Out[2]: 965738496

In [3]: 922 * 1024 * 1024                                                                                                                                                                                                                                                                                              
Out[3]: 966787072
```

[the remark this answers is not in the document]
jigold: We allow 0.25 cpu jobs.

danking (approved):

PR #9835, opened 2020-12-17
[hail] Add two new forms to any all

danking:
CHANGELOG: `hl.any` and `hl.all` now also support a single collection argument and a varargs of Boolean expressions

Any and all currently take two arguments: a function and a collection. This
differs from the Python `any` and `all` which take only a collection and check
for truthy values.

We rectify this situation by supporting three forms:

1. ``hl.all(boolean, ...)``, are all arguments ``True``?
2. ``hl.all(collection)``, are all elements of the collection ``True``?
3. ``hl.all(function, collection)``, does ``function`` return ``True`` for
    all values in this collection?

### DIFF
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -1,3 +1,4 @@
+import operator
 import builtins
 import functools
 from typing import Union, Optional, Any, Callable, Iterable, TypeVar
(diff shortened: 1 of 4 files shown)


danking:
example of docs rendering on my laptop:

Screen Shot 2020-12-17 at 1 47 15 PM

tpoterba (approved):
good change, especially the checker refactor!

PR #9836, opened 2020-12-17
[hailctl] use argparse subparsers throughout

cseed:
for better or worse.

I believe the behavior is the same and the code and user ergonomics seem better.  The one known issue I haven't figured out to fix is the usage output on unknown flags:

```
$ hailctl --fleep auth login
usage: hailctl [-h] {version,dataproc,auth,dev,batch,curl,config} ...
hailctl: error: unrecognized arguments: --fleep

$ hailctl auth --fleep login
usage: hailctl [-h] {version,dataproc,auth,dev,batch,curl,config} ...
hailctl: error: unrecognized arguments: --fleep
```

This implies argparse should allow hailctl flags to appear after auth, but that isn't the case.  For example:

```
$ hailctl dataproc connect --beta foo nb
usage: hailctl [-h] {version,dataproc,auth,dev,batch,curl,config} ...
hailctl: error: unrecognized arguments: --beta
```

I will keep investigating a fix for this.


cseed:
So this is partially my fault.  I'm using parse_known_args, so we don't know where unknown args are placed, and therefore can't give the right usage (even if I can intervene on the usage being printed by argparse).  I really only want to allow unknown args in the relevant subcommands, but don't know how to write that argument given the strange behavior I'm seeing with `nargs='*'` and `nargs=argparse.REMAINDER`. 

Hmm.

cseed:
Closing this for further consideration.  Apologies for the distraction.

PR #9842, opened 2020-12-19
[hailctl] rewrite argument parsing

cseed:
This PR rewrites the hailctl command line argument parsing code.  While the interface remains largely the same, a few changes were made to how options are handled.

We first introduce a bit of terminology.  In a shell command invocation like `$ cmd a -o c`, `a`, `-o` and `c` are called parameters.    `a` and `c` which do not start with dashes, are called arguments.  `-o`, which starts with a dash, is an option.

  - For dataproc commands taking extra gcloud parameters, all parameters after a double-dash (--) are passed to gcloud.
  - The actual rule is slightly more complicated, but I think the above rule is the right take away.  In detail, extra parameters are passed to gcloud.  Unknown options (starting with a dash) before `--` are reported as an error.  So arguments (not options) before `--` and all parameters after are passed to gcloud. 
  - Short options don't need a `=` when specifying a value.  It is now `-p2`, not `-p=2`.
  - While I was making breaking changes, I changed `dataproc submit` `--gcloud_configuration` to `--gcloud-configuration`.  I am happy to undo this one.
 - Group arguments must go before the next command.  Write `hailctl dataproc --beta start ...` not `hailctl dataproc start --beta ...`, which is an error since `start` has no option `--beta`.

This PR rewrites argument parsing to use click instead of argparse: https://click.palletsprojects.com/en/7.x/.  Things you need to know about click:
 - A group is a group of commands or subgroups, like `hailctl dataproc`, `hailctl batch`, etc.  Groups are defined like this:
   
   ```
   @hailctl.group(
       help="Manage the Hail Batch service.")
   def batch():
       pass
   ```
 - A command in a group is defined like this:

   ```
   @batch.command(
       help="Get a particular batch's info.")
   @click.argument('batch_id', type=int)
   @click.option('--output-format', '-o',
                 type=click.Choice(['yaml', 'json']),
                 default='yaml', show_default=True,
                 help="Specify output format",)
   def get(batch_id, output_format):
       ...
    ```

   The command decorator replaces the function with one that takes a `List[str]` of command line parameters, parses them, and calls the original function.  The option options are pretty self-explanatory.

 - To access an argument to a group (like `dataproc --beta`) in a (sub)command, use the decorator `click.pass_context` to pass the click context which allows you to access parent group parameters.  `dataproc start` is an example:

   ```
   @dataproc.command(
       help="Start a Dataproc cluster configured for Hail.")
   @click.argument('cluster_name')
   ...
   @click.pass_context
   def start(ctx, cluster_name, ...):
       beta = ctx.parent.params['beta']
   ```

The help output for a group looks like this:

```
$ hailctl dataproc --help
Usage: hailctl dataproc [OPTIONS] COMMAND [ARGS]...

  Manage and monitor Hail deployments.

Options:
  --beta  Force use of `beta` in gcloud commands
  --help  Show this message and exit.

Commands:
  connect   Connect to a running Dataproc cluster
  describe  Gather information about a Hail (Table or MatrixTable) file...
  diagnose  Diagnose problems in a Dataproc cluster.
  list      List Dataproc clusters.
  modify
  start     Start a Dataproc cluster configured for Hail.
  stop      Shut down a Dataproc cluster.
  submit    Submit a Python script to a running Dataproc cluster.
```

The help output for a command looks like:

```
$ hailctl batch get --help
Usage: hailctl batch get [OPTIONS] BATCH_ID

  Get a particular batch's info.

Options:
  -o, --output-format [yaml|json]
                                  Specify output format  [default: yaml]
  --help                          Show this message and exit.
```

I also made `BatchClient` a context manager and made the default limit unbounded in `BatchClient.list_batches`.

I have marked this WIP until we are happy with the interface changes and how we're going to communicate them to the users.

CHANGELOG:  Rewrote the hailctl argument parsing code, which made some incompatible changes to the way hailctl handles arguments.  For more details, see [the URL of this PR].

Spice level: medium.


nawatts:
Hopefully you don't mind this unrequested review, I noticed this PR and had some thoughts.

As a developer, I like the new way of defining available parameters and the greater clarity between group and command options. As a user, I like the greater flexibility in providing parameters.

A few concerns though:
- For hailctl arguments that are also gcloud arguments (for example, `--project` to `hailctl dataproc start`), what happens if a user provides them in both places (for example, `hailctl dataproc start --project=project-a cluster-name -- --project=project-b`)? One nice attribute of the current parsing method is that it does not allow this, since in most cases the hailctl argument shadows the gcloud argument of the same name.
- It looks like this creates some inconsistency in how the same argument must be provided to different `hailctl dataproc` commands. For example, `--project` can be directly provided to `hailctl dataproc start`, but it would have to go after the `--` for `hailctl dataproc list` or `hailctl dataproc modify`. That seems likely to be surprising/annoying for users. This could be solved by moving such flags (`--project`, `--region`, and `--configuration` are the ones that immediately come to mind) to the `hailctl dataproc` group level or by adding definitions for those arguments to all `hailctl dataproc` commands.
- On the subject of `--`, a current issue with `hailctl dataproc submit` is that it does not support `--` for specifying parameters to the submitted script like `gcloud dataproc jobs submit` does. Thus, for example, you cannot currently submit a script that has a `--files` argument because `--files` will be interpreted as an argument to `hailctl dataproc submit` instead of the submitted script. It would be nice to support that behavior in `hailctl dataproc submit`. However, with this parsing approach, supporting script arguments like that might conflict with accepting pass through arguments to `gcloud dataproc jobs submit` such as `--async`, `--bucket`, etc.

And more minor:
- For `hailctl dataproc start` especially, it could seem pretty arbitrary to a user which arguments go before `--` vs after. For example, `--num-worker-local-ssds` is a `hailctl dataproc start` argument, but `--num-secondary-worker-local-ssds` is not. This could cause some confusion/annoyance. This could potentially be reduced by minimizing the number of `gcloud dataproc` arguments that are defined as `hailctl dataproc` arguments. For example, `--num-worker-local-ssds` is just passed through to `gcloud`, so there's no real need for it to be a `hailctl dataproc start` argument. On the other hand, it is nice to have some of those `gcloud` arguments show up in `hailctl dataproc start --help`.
- While making breaking changes, it would be nice if the `--configuration`/`--gcloud-configuration` argument was consistent across `hailctl dataproc` commands. For example, currently `hailctl dataproc start` takes `--configuration` but `hailctl dataproc submit` takes `--gcloud-configuration`. It would also be nice to standardize on kebab case for all arguments. There are some other arguments that use underscores (ex. `--init_timeout` to `hailctl dataproc start`).

cseed:
Hi @nawatts, your comments are very welcome, and I appreciate your perspective as a hailctl user.

OK, mulling over your comments, I think I can address collectively by removing all hailctl options that pass through to gcloud.  This removes the question of providing them twice, makes all the commands consistent.

I think this also addresses the issue `hailctl dataproc submit` not supporting `--`, because you can specify it twice: once to break out of hailctl options, and once to break out of gcloud options to specify options the script being submitted:  `hailctl dataproc submit --halictl-option -- --gcloud-options -- --script-options and-parameters`.

What do you think?

 > it would be nice if the --configuration/--gcloud-configuration argument was consistent across hailctl dataproc commands
 > It would also be nice to standardize on kebab case for all arguments.

Agree on both accounts, will fix.

Thanks again!

nawatts:
> I think I can address collectively by removing all hailctl options that pass through to gcloud. This removes the question of providing them twice, makes all the commands consistent.

Do you mean remove all options that are simple pass throughs (such as `--num-worker-local-ssds`) or all options that are also gcloud options (such as `--project`)? The latter could be difficult, since there are some gcloud options that hailctl also needs to read, like `hailctl dataproc start` using `--project` to set requester pays configuration, extending `--initialization-actions` with notebook/VEP init scripts, setting a higher default disk size when `--vep` is specified, etc. or `hailctl dataproc submit` automatically zipping `--py-files`.

> I think this also addresses the issue hailctl dataproc submit not supporting --, because you can specify it twice: once to break out of hailctl options, and once to break out of gcloud options to specify options the script being submitted: hailctl dataproc submit --halictl-option -- --gcloud-options -- --script-options and-parameters.

Nice. That would solve the problem, though the syntax seems a little unusual. I would guess submitting script arguments is more common than using gcloud options here, so it would be nice for the gcloud arguments group to be optional, so that `hailctl dataproc submit cluster -- --script-options` would work.

cseed:
> all options that are also gcloud options (such as --project)? [That] could be difficult

Yes, this is what I was thinking.  `hailctl` could parse (as much as is needed) the `gcloud` options to find options (like `--project`) and modify others (like `--initialization-actions`).  The latter is somewhat surprising since one expects everything after the `--` to pass through unchanged.

OK, summarizing our options so far:

 - hailctl has no options that are also gcloud options.  gcloud options go after the `--`, and get modified as needed by hailctl (with a message).
 - hailctl has no gcloud options that are simply pass through.  gcloud options that are needed by hailctl commands are hailctl options (like `--project`).  When a gcloud option is needed by some hailctl command, all hailctl commands take that option (when it makes sense), even if in some cases that makes them simply pass through.  This fixes the inconsistency issues, but the user still needs to keep track of which gcloud options needs to be passed to hailctl and which are passed to gcloud directly.  If you specify an option twice, once to hailctl and once to gcloud, we invoke gcloud with the option duplicated.

Pros and cons:
 - The first option has the most consistent interface.
 - The first option modifies options after the --, which is surprising.
 - The first option involves replication (some of) the gcloud option parsing semantics, which is annoying.
 - The second option requires the user to know which gcloud options need to be passed to hailctl instead (but globally, not per-command).
 - With the second option, if we want to warn (or error) on duplicate options, we're back to duplicating the gcloud option parsing logic.

I think I'm coming around to the second option.

 > so that hailctl dataproc submit cluster -- --script-options would work.

I see, so if there is only one `--` it refers to script options, and if there are two, the first one corresponds to gcloud options?  I think that should be doable.

johnc1231:
I'm a little worried about losing the self documenting nature of `hailctl start`. Right now, I just do `hailctl start -h` to figure out what the options are. If we move to a model where most options are things you specify after the `--` and get passed through to `gcloud`, will we still have good documentation when you `-h`, or will a user have to go look at the `gcloud dataproc start` docs? The later seems unfortunate.

cseed:
@johnc1231 Thanks for chiming in!  This is a tricky design problem, I appreciate the concerns.

I think we should be able to document the gcloud options commonly used by hailctl users in the hailctl help output even if those aren't options that we parse.  So let's say, whatever solution we include, whether we parse them or not, the current set of options continue to be documented in the hailctl help output.  Does that address your concern?

johnc1231:
Yes, that would address my concern. Thank you.

nawatts:
I also prefer the second option in https://github.com/hail-is/hail/pull/9842#issuecomment-758128554. The more I think about this, the more problematic the notion of having an opaque list of "arguments to pass through to gcloud" seems. For example, `hailctl dataproc start` may run multiple gcloud commands: one to start the cluster and another to apply tags to the master node. In that case, we'd want to pass through extra args to the cluster start command, but not the apply tags command. Or `hailctl dataproc modify`, where we might want to accept extra args for `gcloud dataproc clusters update`. Those args shouldn't be passed through to the `gcloud compute ssh` commands, but options like `--project` or `--configuration` should.

It seems like pass through arguments are likely going to have to be handled on a case by case basis for each hailctl command. That would probably make any approach that required parsing those pass through options more cumbersome to use.

nawatts:
It's unfortunate that gcloud will happily let you specify the same argument twice. For example, `gcloud compute instances list --project project-a --project project-b` lists instances in project B. So there's the opportunity for a user to override a gcloud option set by hailctl by setting the same option in the pass through arguments.

cseed:
> The more I think about this, the more problematic the notion of having an opaque list of "arguments to pass through to gcloud" seems

I agree.  Here's another option: don't have one/global way to specify extra gcloud command arguments.  Only `halictl dataproc submit` will support `--`, and that's for the script arguments.  Then hailctl commands can have an argument for specifying additional arguments to specific gcloud subinvocations, which allows us to have multiple.  Something like `hailctl dataproc start --extra-gcloud-start-args=... --extra-gcloud-tag-args=... ...` and those arguments will (1) be more specific about the gcloud subcommand being run, and (2) be specific to hailctl dataproc commands.

And then otherwise we got with the second option in https://github.com/hail-is/hail/pull/9842#issuecomment-758128554.

nawatts:
> Something like hailctl dataproc start --extra-gcloud-start-args=... --extra-gcloud-tag-args=... ... and those arguments will (1) be more specific about the gcloud subcommand being run, and (2) be specific to hailctl dataproc commands.

I think this would be my preference. I guess there's no way we're going to be able to totally guard against someone specifying pass through arguments that conflict with something in hailctl args. But I think this way presents as more obviously an escape hatch for doing something that's not well-supported and thus less surprising if it breaks something.

cseed:
Awesome!  Great discussion @nawatts thanks for the help thinking through the relevant constraints.  I'd like @tpoterba to weigh in, and if he's OK with it I'll make the changes.

tpoterba:
I'll look tomorrow. Apologies for my dereliction of PR duty, and thank you Nick!

cseed:
For @tpoterba's benefit (and to get things clear), here's the current proposal:
 - Move extra gcloud arguments to `--extra-gcloud-<description>-args="--arg1 ... --argN"` where there is one such argument for each invocation of gcloud inside a hailctl command.  gcloud args no longer go at the end.
 -  Only `hailctl dataproc submit` supports `--` which is used to separate submit arguments from the script arguments,
 -  Remove all gcloud arguments that are pass through in all commands, but mention them in the command help so users don't need to look at the gcloud help for commonly used options.
 - gcloud options that are needed by some hailctl command should be consistently available among all hailctl commands (where appropriate, and where in some cases they may simply be pass-through).
 -  The other consistency changes @nawatts highlighted.

(I wouldn't be surprised if I other issues come up when I make the changes, but this is a start.)

cseed:
OK, I've made most of the changes and I'd appreciate some feedback before I finalize the PR.

Notable changes:
 - The `hailctl dataproc` subcommand now has `--beta`, `--configuration=`, `--dry-run`, `--project=` and `--zone=`.  These apply to all commands.  There is a `GcloudRunner` object that takes these options, is set to the click context user `obj` field, and is used by all hailctl dataproc commands to invoke gcloud.  Note, not all dataproc subcommands invoke gcloud, but the current design doesn't differentiate.  Note, with `click`, the subcommand options must go on the subcommand, so `hailctl dataproc stop --dry-run` is an error.
 - hailctl no longer takes `--region` (for gcloud dataproc commands).  I compute region in `GcloudRunner` by checking dataproc/region or falling back to determining the region from the zone.  I error if the region and zone are incompatible (gcloud would also do this).
 - I stripped all gcloud pass through args from `hailctl dataproc modify`.  There aren't any left.  Invoking `modify` now looks like:

   ```
   hailctl dataproc modify my-cluster \
     --extra-glcoud-update-args='---num-workers=2 --num-secondary-workers=100'
   ```

   The `extra` in the option name sounds a little weird since they are the only options (and the command isn't run if they aren't specified), but I'm leaving it for consistency for now.  I moved the help text from the removed options into the help for the modify command itself.  The output of `modify --help` is included below.
 - I plan to leave the `--async` option to stop, although it is pass through.
 - Then there is `--files` for submit.  This is passed through, but `--py-files` is needed (it is not passed through, but modified).  Do I leave `--files`?  I'm currently inclined to.
 - Finally, I need to strip out the pass through arguments for start like I did with update.

```
$ hailctl dataproc modify --help
Usage: hailctl dataproc modify [OPTIONS] CLUSTER_NAME

  Modify an existing Dataproc cluster.

  'hailctl dataproc modify' works by calling 'gcloud dataproc clusters
  update' and then updating the Hail version if '--update-hail-version' or '
  --wheel' is specified.  You can pass arguments to the 'update' command
  with the option '--extra-gcloud-update-args'.

  The following 'gcloud dataproc clusters update' options may be useful:

    --num-workers=NUM_WORKERS: New number of worker machines, minimum 2.

    --num-secondary-workers=NUM_SECONDARY_WORKERS: New number of secondary
    (preemptible) worker machines.

    --graceful-decommission-timeout=GRACEFUL_DECOMMISSION_TIMEOUT: Graceful
    decommissioning allows removing nodes from the cluster without
    interrupting jobs in progress.  Timeout specifies how long to wait for
    jobs in progress to finish before forcefully removing nodes (and
    potentially interrupting jobs).  Timeout defaults to 0 if not set (for
    forceful decommission), and the maximum allowed timeout is 1 day.

    At most one of the following may be set:

      --expiration-time=EXPIRATION_TIME: The time when cluster will be auto-
      deleted.

      --max-age=MAX_AGE: The lifespan of the cluster before it is auto-
      deleted, such as '60m' or '1d'.

      --no-max-age: Cancel the cluster auto-deletion by maximum cluster age,
      as configured by max-age or --expiration-time flags.

    At most one of the following may be set:

        --max-idle=MAX_IDLE: The duration before cluster is auto-deleted
        after last job finished, such as '60m' or '1d'.

        --no-max-idle: Cancel the cluster auto-deletion by cluster idle
        duration (configured by --max-idle flag).

    See 'gcloud dataproc clusters update --help' for more information.

Options:
  --update-hail-version           Update the version of hail running on
                                  cluster to match the currently installed
                                  version.

  --wheel TEXT                    New Hail installation.
  --extra-gcloud-update-args TEXT
                                  Extra arguments to pass to 'gcloud dataproc
                                  clusters update'.  The 'update' command is
                                  only run if this option is specified.

  --help                          Show this message and exit.
```

nawatts:
> The hailctl dataproc subcommand now has --beta, --configuration=, --dry-run, --project= and --zone=. These apply to all commands. There is a GcloudRunner object that takes these options, is set to the click context user obj field, and is used by all hailctl dataproc commands to invoke gcloud. Note, not all dataproc subcommands invoke gcloud, but the current design doesn't differentiate. Note, with click, the subcommand options must go on the subcommand, so hailctl dataproc stop --dry-run is an error.

Nice. It's great that these are handled at the `hailctl dataproc` level instead of having to remember to account for them in every `hailctl dataproc` subcommand. That's going to resolve a lot of inconsistencies.

A nitpick though... is there a better name for the click context attribute than "obj"?

> hailctl no longer takes --region (for gcloud dataproc commands). I compute region in GcloudRunner by checking dataproc/region or falling back to determining the region from the zone. I error if the region and zone are incompatible (gcloud would also do this).

If consistency with `gcloud dataproc` is desired, I think the opposite (determining zone from cluster region) would be preferable. `gcloud dataproc` commands take a `--region` argument. [`--zone` is an optional argument for `gcloud dataproc clusters create`](https://cloud.google.com/sdk/gcloud/reference/dataproc/clusters/create#--zone).

When a cluster's zone is needed to run `gcloud compute` commands, it can be determined using `gcloud dataproc clusters describe <cluster> --format json`. `hailctl dataproc diagnose` currently does this.

I believe the only reason that we currently require a zone be provided either in gcloud configuration or on the command line is to maintain backwards compatibility. `cloudtools` and earlier versions of `hailctl` had a default value for the `--zone` option of `hailctl dataproc start` (I think it was `us-central1-b`).

> I stripped all gcloud pass through args from hailctl dataproc modify. There aren't any left. Invoking modify now looks like:
> 
> ```
> hailctl dataproc modify my-cluster \
>   --extra-glcoud-update-args='---num-workers=2 --num-secondary-workers=100'
> ```
>
> The extra in the option name sounds a little weird since they are the only options (and the command isn't run if they aren't specified), but I'm leaving it for consistency for now. I moved the help text from the removed options into the help for the modify command itself. The output of modify --help is included below.

I have mixed feelings on this one. On the one hand, `--extra-gcloud-update-args` sounds like it is extra arguments for a `gcloud update` command, which isn't a thing. On the other hand, `--extra-gcloud-dataproc-clusters-update-args` is an awfully long argument name.

> I plan to leave the --async option to stop, although it is pass through.

> Then there is --files for submit. This is passed through, but --py-files is needed (it is not passed through, but modified). Do I leave --files? I'm currently inclined to.

Agreed. I support having the most frequently used parameters as `hailctl` parameters, even if they are only simple pass throughs.

My original comment about minimizing the number of simple pass through parameters was mainly directed toward `hailctl dataproc start`, which has several options than can be specified separately for master node, worker nodes, and secondary worker nodes. I wanted to avoid cases where, for example, `--worker-boot-disk-size` was a `hailctl` option, but `--secondary-worker-boot-disk-size` had to be specified after a `--` or with `--extra-gcloud-start-args`.

nawatts:
While we're making breaking changes to the `hailctl` interface... there is a beta feature to start/stop Dataproc clusters. Once this is released, it could create some confusion that `hailctl dataproc start` runs `gcloud dataproc clusters create` instead of `gcloud dataproc clusters start`. Likewise for `hailctl dataproc stop` and `gcloud dataproc clusters delete`. Should we rename `hailctl dataproc` start/stop to create/delete?

https://cloud.google.com/dataproc/docs/guides/dataproc-start-stop

tpoterba:
Hey Nick, wanted to loop you in on an offline discussion I had with Cotton about this. First, thank you for picking up review responsibilities! I'll just do a brief review focusing on interaction of this change with intended directions for hailctl. Here are the conclusions from our discussion:

1. This is a breaking change to the hailctl interface. We're OK with that.
2. Although we are OK making breaking changes, we should get Grace's team on board and update their scripts/repos before merging/releasing. For that reason this will sit for a few weeks until their current urgent analysis push is done.

cseed:
> Should we rename hailctl dataproc start/stop to create/delete?

I think so.  Also, modify to update.

danking:
Closing until someone has bandwidth to push this forward.

PR #9843, opened 2020-12-22
[query] Allow bit shifts by zero bits

nawatts:
Currently, attempting to bit shift by zero bits throws an error.

```
$ hl.eval(hl.bit_rshift(4, 0))
HailUserError: Error summary: HailException: cannot shift by a negative value: 4 >> 0
```


johnc1231 (approved):
Good fix and good tests, thank you. 

Feel free to randomly assign PRs like this to Tim, Chris, Patrick, or me. There's a little random PR review picker here: https://scorecard.hail.is/ for compilers team at the bottom of the page. That's what we normally use to figure out who to assign to, though you're usually making small python fixes so just throwing it to any of us should be fine.

PR #9844, opened 2020-12-22
[batch] make standing worker configurable in the UI

danking:
This particularly important for rapid iteration in dev deploys. I can enable the standing worker
while I am working. I disable it when I head off to a series of meetings and then re-enable it
when I return.

@jigold I think this will conflict with your worker pool change because I replace "1 means boolean true" with a checkbox.


batch/batch/driver/instance_pool.py
```diff
@@ -85,18 +87,19 @@ async def async_init(self):
...
-  standing_worker_cores, max_instances, pool_size
+  standing_worker_cores, max_instances, pool_size, enable_standing_worker
 FROM globals;
 ''')
 
         self.worker_type = row['worker_type']
         self.worker_cores = row['worker_cores']
         self.worker_disk_size_gb = row['worker_disk_size_gb']
-        self.worker_local_ssd_data_disk = row['worker_local_ssd_data_disk']
+        self.worker_local_ssd_data_disk = bool(row['worker_local_ssd_data_disk'])
```
jigold, line 97: Why do you need to add the `bool` here? Shouldn't the values coming from pymysql be booleans already?

jigold (dismissed):

jigold:
I added the WIP tag so you could merge at your leisure once you answer my question.

batch/batch/driver/instance_pool.py
danking: MySQL represents booleans as TINYINT(1), afaik. They come into Python as integers:
```python3
>>> x = l.run_until_complete(db.execute_and_fetchone('select TRUE as value'))
>>> x
{'value': 1}

```

jigold (approved):

jigold:
Merge at your leisure.

PR #9845, opened 2020-12-22
[services] stop treating all logs as error logs

danking:
Google considers any message written to stderr as an error.

### DIFF
--- a/hail/python/hailtop/hail_logging.py
+++ b/hail/python/hailtop/hail_logging.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 from pythonjsonlogger import jsonlogger
 from aiohttp.abc import AbstractAccessLogger
 import datetime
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/hail_logging.py
```diff
@@ -15,7 +16,7 @@ def add_fields(self, log_record, record, message_dict):
 def configure_logging():
     fmt = CustomJsonFormatter('(levelname) (asctime) (filename) (funcNameAndLine) (message)')
 
-    stream_handler = logging.StreamHandler()
+    stream_handler = logging.StreamHandler(stream=sys.stdout)
```
jigold, line 19: will this cause all log messages to go to stdout including errors? Also, how do we make sure we can still see the logs with kubectl logs?

jigold (changes requested):

hail/python/hailtop/hail_logging.py
danking: Yes, all log messages will go to stdout. Kubectl shows stdout and stderr so all our logs will still be visible in kubectl logs.

jigold (approved):

PR #9848, opened 2020-12-22
[query] shuffle in the service

danking:
This PR enables shuffling in the service. It is stacked on several other PRs, so look only at the
most recent commit.

Some highlights:
- Open the public network back up. We should probably make query jobs special so that they can
  access the internal network. To do that, batch would need to accept a "acting on behalf of" user
  account: Query submits the job using its account "acting on behalf of" the user. Batch allows
  query to use the private network, but for all other purposes, the job is owned by the user.

- Allow public access to some the `gcr.io/hail-vdc/query` Docker image.

- Automatically rewrite uses of `hailgenetics/` Docker images to their `gcr.io` equivalents.

- Move `deploy_address` above `deploy_query` so that query can depend on address (necessary for
  shufles).

- Fix logging configuration. Services team wants all logs all the time to go to stdout.

- Implement lowerDistributedSort using the shuffler.

- Allow shuffle ids to be encoded so they can be used in `Literal`.
Unified Split

### DIFF
--- a/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
@@ -3,21 +3,27 @@ package is.hail.backend.service
 import java.io._
 
 import is.hail.HailContext
-import is.hail.annotations.{Region, UnsafeRow}
+import is.hail.annotations.{Annotation, Region, SafeRow, UnsafeRow}
 import is.hail.asm4s._
 import is.hail.backend.{Backend, BackendContext, BroadcastValue, HailTaskContext}
 import is.hail.expr.JSONAnnotationImpex
 import is.hail.expr.ir.lowering.{DArrayLowering, LowerDistributedSort, LoweringPipeline, TableStage, TableStageDependency}
-import is.hail.expr.ir.{Compile, ExecuteContext, IR, IRParser, MakeTuple, SortField}
+import is.hail.expr.ir.{Compile, ExecuteContext, IR, IRParser, Literal, MakeArray, MakeTuple, ShuffleRead, ShuffleWrite, SortField, ToStream}
 import is.hail.io.fs.GoogleStorageFS
 import is.hail.linalg.BlockMatrix
+import is.hail.rvd.RVDPartitioner
+import is.hail.services.{DeployConfig, Tokens}
 import is.hail.services.batch_client.BatchClient
-import is.hail.types.BlockMatrixType
+import is.hail.services.shuffler.ShuffleClient
+import is.hail.types.{BlockMatrixType, TypeWithRequiredness}
+import is.hail.types.encoded.{EBaseStruct, EType}
 import is.hail.types.physical.{PBaseStruct, PType}
+import is.hail.types.virtual.{TArray, TInt64, TInterval, TShuffle, TStruct, Type}
 import is.hail.utils._
 import is.hail.variant.ReferenceGenome
 import org.apache.commons.io.IOUtils
 import org.apache.log4j.LogManager
+import org.apache.spark.sql.Row
 import org.json4s.JsonAST._
 import org.json4s.jackson.JsonMethods
 import org.json4s.{DefaultFormats, Formats}
@@ -83,7 +89,10 @@ class ServiceBackendContext(
   @transient val sessionID: String,
   val billingProject: String,
   val bucket: String
-) extends BackendContext
+) extends BackendContext {
+  def tokens(): Tokens =
+    new Tokens(Map((DeployConfig.get.defaultNamespace, sessionID)))
+}
 
 object ServiceBackend {
   lazy val log = LogManager.getLogger("is.hail.backend.service.ServiceBackend")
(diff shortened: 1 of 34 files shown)


catoverdrive (changes requested):
Going to finish reviewing in a moment, but here's a comment in the meantime.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -363,8 +376,45 @@ class ServiceBackend() extends Backend {
...
+    shuffleClient.start()
+    try {
+      val uuid = shuffleClient.uuid
+
+      val successfulPartitionIds = execute(
+        ctx,
+        stage.mapCollect(relationalLetsAbove)(
+          ShuffleWrite(Literal(shuffleType, uuid), _)))
+
+      val partitionBoundsPointers = shuffleClient.partitionBounds(region, stage.numPartitions) // FIXME(danking): number of partitions should be configurable?
```
catoverdrive, line 397: I wonder where the right place to handle this is. It feels like figuring out how many partitions you actually want might happen in a downstream optimization pass so that any repartitions/joins/etc can be pushed up into the read. Are there cases where you'd want different/configurable behavior by default? Should we be computing the number of new partitions based on the number of rows written to the shuffler?

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -363,8 +376,45 @@ class ServiceBackend() extends Backend {
...
+
+      TableStage(
+        globals = Literal(TStruct(), Row()),
+        partitioner = partitioner,
+        TableStageDependency.none,
+        contexts = ToStream(MakeArray(
+          partitionIntervals.map(interval => Literal(TInterval(keyType), interval)),
+          TArray(TInterval(keyType)))),
+        interval => ShuffleRead(Literal(shuffleType, uuid), interval))
+    } finally {
```
catoverdrive, line 415: So the `finally` block closes the shuffleClient, but we'll still be able to read the result of this shuffle by creating a new shuffle client with the correct uuid, right?

Do we ever clean up the results of a shuffle once it's passed out of scope? I think it's pretty safe (for now) to delete a shuffle once we've exited the nested `execute` call on the IR that created it, since we can't reference it otherwise.

catoverdrive (commented):
Two more notes for the shuffle lowering bit.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: These are good questions, but they're questions for @tpoterba . I'm not sure we do anything that resembles SQL "query planning".

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: Right, this closes the socket. ShuffleClient.stop frees the shuffle.

This is another question for @tpoterba: when is the `ShuffleRead` on line 414 no longer usable? That ShuffleRead is contained inside a `TableStage`. The `TableStage` is returned by lowerDistributedSort. Is there any lifetime management I can piggyback on?

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
tpoterba: this is fine for now. We'll change this as we build in more sensible partition handling

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
tpoterba: You should call `shuffleClient.close()` when the ExecuteContext is closed, I think.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
catoverdrive: Sorry, I was kind of thinking out loud here but my actual comment that I forgot to state was: I don't think you need this FIXME here. Seems like you're doing the right thing at this point, which is just picking a hopefully-reasonable number of partitions here and downstream transformations can repartition as necessary.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -362,9 +375,58 @@ class ServiceBackend() extends Backend {
...
+    ctx: ExecuteContext,
+    stage: TableStage,
+    sortFields: IndexedSeq[SortField],
+    relationalLetsAbove: Map[String, IR],
+    tableTypeRequiredness: RTable
+  ): TableStage = {
+    val region = ctx.r
+    val rowType = stage.rowType
+    val keyType = rowType.typeAfterSelectNames(sortFields.map(_.field))
+    val rowEType = EType.fromTypeAndAnalysis(rowType, tableTypeRequiredness.rowType).asInstanceOf[EBaseStruct] // FIXME(danking): Probably not kosher
```
catoverdrive, line 388: remove FIXMEs?

catoverdrive (approved):

PR #9850, opened 2021-01-04
[query] Remove code generating folds from AppendOnlyBTree

chrisvittal:
This is a series of small refactorings in AppendOnlyBTree that removes
the folds in code generation, replacing a pattern that I find difficult
to follow with a more imperative CodeBuilder style that is necessary for
future changes to proceed as the folding style does not work well with
functions that need code builders.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/agg/AppendOnlyBTree.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/agg/AppendOnlyBTree.scala
@@ -100,16 +100,6 @@ class AppendOnlyBTree(kb: EmitClassBuilder[_], key: BTreeKey, region: Value[Regi
           (!isLeaf(srcNode)).orEmpty(setChild(destNode, destIdx, loadChild(srcNode, srcIdx))))
       }
 
-    val shiftAndInsert = Array.range(1, maxElements)
-      .foldLeft(makeUninitialized(0)) { (cont, destIdx) =>
-        Code(
-          hasKey(node, destIdx - 1).orEmpty(
-            copyFrom(node, destIdx, node, destIdx - 1)),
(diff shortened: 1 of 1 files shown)


hail/src/main/scala/is/hail/expr/ir/agg/AppendOnlyBTree.scala
```diff
@@ -178,31 +178,27 @@ class AppendOnlyBTree(kb: EmitClassBuilder[_], key: BTreeKey, region: Value[Regi
...
-    val cmp = get.newLocal[Int]()
-    val keyV = get.newLocal[Long]()
-
-    def insertOrGetAt(i: Int) = EmitCodeBuilder.scopedVoid(get) { cb =>
-      cb.ifx(isLeaf(node), {
-        cb.assign(keyV, cb.invokeCode(insert, node, const(i), k, const(0L)))
-        cb.assign(cmp, 0)
-      }, {
-        cb.assign(node, loadChild(node, i - 1))
+    get.emitWithBuilder { cb =>
```
tpoterba, line 181: voidWithBuilder

tpoterba (changes requested):

chrisvittal:
I am unsure why emitWithBuilder did not work in the first pass, but I've figured this out and we can use it as a pattern to transition other codegen here away from a fold with continuation pattern into this imperative one.

hail/src/main/scala/is/hail/expr/ir/agg/AppendOnlyBTree.scala
chrisvittal: This was an error, `btree_get` returns a value. Fixed now.

tpoterba (approved):
looks good.

PR #9851, opened 2021-01-04
[query] AppendOnlyBTree DeepCopy to codeBuilder

chrisvittal:

PR #9852, opened 2021-01-04
[query] remove a fold from AppendOnlyBTree insert

chrisvittal:


chrisvittal:
same as #9851, will rollup if it passes tests

chrisvittal:
Tests passed rolling up

PR #9853, opened 2021-01-04
[query] remove fold from AppendOnlyBTree insert shiftAndInsert

chrisvittal:


chrisvittal:
This is now all folded into #9850

PR #9854, opened 2021-01-05
[query] NDArray Java Representation

johnc1231:
Part of a refactoring effort. I need to break the model that a `TNDArray` is represented by an underlying `TStruct` and `TArray`. This adds an `UnsafeNDArray` Java representation, rather than the old model which was just to use`UnsafeRow` like it was a struct. 

cc @tpoterba


johnc1231:
Closing temporarily to make a few more changes.

johnc1231:
I added the `forall` method and methods to look up individual elements without loading whole thing into memory.

johnc1231:
Bump

chrisvittal (approved):

PR #9857, opened 2021-01-07
[batch] refactor job schema to understand different job types

catoverdrive:
Pulled out `command`, `image` and `mount_docker_socket` as docker-job-specific keys that don't apply to non-docker jobs. Also pushed handling of deprecated keys into schema validation.


batch/batch/front_end/validate.py
```diff
@@ -66,13 +81,69 @@ def __init__(self, reason):
...
 
+def handle_deprecated_args(i, job):
+    if 'pvc_size' in job:
+        if 'storage' in resources:
+            raise ValidationError(f"jobs[{i}].resources.storage is already defined, but "
+                                  f"deprecated key 'pvc_size'.")
+        deprecated_msg = "[pvc_size key is DEPRECATED. Use resources.storage]"
+        pvc_size = job['pvc_size']
+        if not isinstance(pvc_size, str):
+            raise ValidationError(f'{deprecated_msg} jobs[{i}].pvc_size not str.')
```
catoverdrive, line 100: Double validating e.g. here (here, and subsequent validation pass on `resources.storage`) to return less confusing error messages to old clients.
Once we can send messages like "this is deprecated" to the client, I'd like to remove validation logic in the `handle_deprecated_args` pass.

batch/batch/front_end/validate.py
```diff
@@ -363,6 +406,46 @@ def validate_job(i, job):
...
+    'docker': {'command', 'image', 'mount_docker_socket', 'type'}
+}
+
+
+def validate_process(i, process):
+    if 'type' not in process:
+        raise ValidationError(f'no required key type in jobs[{i}].process')
+    type = process['type']
+    if type not in PROCESS_TYPES:
+        raise ValidationError(f'jobs[{i}].process.type must be in set {PROCESS_TYPES}.')
```
danking, line 420: add `.keys()`  to list just the keys in the error message.

danking (changes requested):

danking (approved):

PR #9858, opened 2021-01-07
[sync.py] avoid infinite retry loop

danking:
If a pod is unreachable for any reason, we previously retried forever. However,
a pods are ephemeral; we cannot assume they will return. Instead, if we fail to
contact a pod, we remove it from the pods list and log the error. If the pod
really does exist, the monitor_pods loop will attempt to initialize it again.

### DIFF
--- a/devbin/sync.py
+++ b/devbin/sync.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-from typing import List, Tuple
+from typing import List, Tuple, Set
 from hailtop.aiotools import BackgroundTaskManager
 from contextlib import closing
 from hailtop.utils import check_shell, CalledProcessError
@@ -21,17 +21,6 @@ log = logging.getLogger('sync.py')
 RSYNC_ARGS = "-av --progress --stats --exclude='*.log' --exclude='.mypy_cache' --exclude='__pycache__' --exclude='*~'"
 
 
-async def retry(f, *args, **kwargs):
-    i = 0
-    while True:
-        try:
-            return await f(*args, **kwargs)
-        except CalledProcessError:
-            i += 1
-            if i % 10 == 0:
-                log.info(f'{i} errors encountered', exc_info=True)
-
-
 DEVBIN = os.path.abspath(os.path.dirname(__file__))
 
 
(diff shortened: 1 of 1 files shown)


devbin/sync.py
```diff
@@ -41,18 +30,27 @@ def __init__(self, paths: List[Tuple[str, str]]):
         self.paths = paths
 
     async def sync_and_restart_pod(self, pod, namespace):
-        await asyncio.gather(*[
-            retry(check_shell, f'{DEVBIN}/krsync.sh {RSYNC_ARGS} {local} {pod}@{namespace}:{remote}')
-            for local, remote in self.paths])
-        await retry(check_shell, f'kubectl exec {pod} --namespace {namespace} -- kill -2 1')
+        try:
+            await asyncio.gather(*[
+                retry(check_shell, f'{DEVBIN}/krsync.sh {RSYNC_ARGS} {local} {pod}@{namespace}:{remote}')
```
jigold, line 35: I'm confused how this can still work when you got rid of the retry function above. I think you want to remove the retry here?

devbin/sync.py
```diff
@@ -41,18 +30,27 @@ def __init__(self, paths: List[Tuple[str, str]]):
...
-        await asyncio.gather(*[
-            retry(check_shell, f'{DEVBIN}/krsync.sh {RSYNC_ARGS} {local} {pod}@{namespace}:{remote}')
-            for local, remote in self.paths])
-        await retry(check_shell, f'kubectl exec {pod} --namespace {namespace} -- kill -2 1')
-        self.pods_list.add((pod, namespace))
+        try:
+            await asyncio.gather(*[
+                retry(check_shell, f'{DEVBIN}/krsync.sh {RSYNC_ARGS} {local} {pod}@{namespace}:{remote}')
+                for local, remote in self.paths])
+            await retry(check_shell, f'kubectl exec {pod} --namespace {namespace} -- kill -2 1')
```
jigold, line 49: Same thing about retry here.

jigold (dismissed):

devbin/sync.py
danking: oh, yep, that's a bad rebase

jigold (approved):

PR #9861, opened 2021-01-07
[batch] rewrite schema validation so that schema description is working validation function

catoverdrive:
I thought about porting some of the typechecker stuff from hail, but they don't serve quite the same purpose, and the functions themselves are pretty simple. This does move the schema definitions to the bottom of the file rather than at the top, but this way we have a schema description that includes requiredness and is guaranteed to be the thing we're actually validating on.

### DIFF
--- a/batch/batch/front_end/validate.py
+++ b/batch/batch/front_end/validate.py
@@ -65,7 +65,7 @@ job_validator = keyed({
         required('namespace'): k8s_str,
         required('name'): k8s_str
     }),
-    'timeout': numeric(**{"x > 0", lambda x: x > 0})
+    'timeout': numeric(**{"x > 0": lambda x: x > 0})
 })
 
 batch_validator = keyed({
--- a/batch/batch/front_end/validate.py
+++ b/batch/batch/front_end/validate.py
@@ -57,7 +57,7 @@ job_validator = keyed({
         'storage': regex(MEMORY_REGEXPAT, MEMORY_REGEX)
     }),
     'secrets': listof(keyed({
-        required('namespace': k8s_str,
+        required('namespace'): k8s_str,
         required('name'): k8s_str,
         required('mount_path'): str_type
     })),
--- a/batch/batch/front_end/validate.py
+++ b/batch/batch/front_end/validate.py
@@ -1,9 +1,8 @@
-import re
-
 from hailtop.batch_client.parse import (MEMORY_REGEX, MEMORY_REGEXPAT,
                                         CPU_REGEX, CPU_REGEXPAT)
 
-from hailtop.utils.validate import *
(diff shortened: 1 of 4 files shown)


batch/batch/front_end/validate.py
```diff
@@ -3,76 +3,138 @@
...
+    def __getitem__(self, key):
+        return self.checkers[key]
+
+    def validate(self, name, obj):
+        super().validate(name, obj)
+        for k in obj:
+            if k not in self.checkers:
+                raise ValidationError(f'unknown key in {name}: {k}')
+        for k, checker in self.checkers.items():
+            if isinstance(checker, required):
```
danking, line 47: The syntax looks a bit weird:
```python3
keyed({
    'foo': required(int),
    'bar': nullabe(int),
    'baz': int
})
```

What do you think of this instead:
```python3
keyed({
    required('foo'): int,
    'bar': nullabe(int),
    'baz': int
})
```
That puts the required ness notion on the right part (the key).

It also simplifies `validate` a bit:
```python3
    def validate(self, name, obj):
         super().validate(name, obj)
         for k in obj:
             if k not in self.checkers:
                 raise ValidationError(f'unknown key in {name}: {k}')
         for k, checker in self.checkers.items():
             if isinstance(k, required) and k not in obj:
                 raise ValidationError(f'{name} missing required key {k}.')
             checker.validate(f"{name}.{key}", obj[k])

```

batch/batch/front_end/validate.py
```diff
@@ -3,76 +3,138 @@
...
+        self.checkers = keyed_checkers
+
+    def __getitem__(self, key):
+        return self.checkers[key]
+
+    def validate(self, name, obj):
+        super().validate(name, obj)
+        for k in obj:
+            if k not in self.checkers:
+                raise ValidationError(f'unknown key in {name}: {k}')
```
danking, line 45: Mild preference to announce all missing keys:
```python3
unknown_keys = obj.keys() - self.checkers.keys()
if unknown_keys:
    raise ValidationError(f'unknown keys in {name}: {", ".join(unknown_keys)}')
```

danking (changes requested):
neat, a few changes. You might consider moving all the validator classes to `hailtop.validate`. That keeps front_end/validate.py short.

catoverdrive:
Oooh, yes. hailtop seems like a good place for it because then we can use it more generally for json schemas, I'll do that.

batch/batch/front_end/validate.py
catoverdrive: yeah, that looks good! I'll change

danking:
needs rebase

danking (approved):

PR #9862, opened 2021-01-07
[tls] clean up TLS

danking:
`tls.py` has many different functions with long names. This change reduces it to three functions:

- internal_server_ssl_context
- internal_client_ssl_context
- external_client_ssl_context

I also added `httpx.py` which contains the HTTPS-related functions that `tls.py` previously
contained. I also simplified the HTTPS-related functions to just:

- client_session
- blocking_client_session

I determine internal vs. external using the deploy config.

---

An [`ssl.SSLContext`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext) defines how a
network library (such as `aiohttp`) should perform SSL/TLS. Let's look at an example:

```python3
server_ssl_context = ssl.create_default_context(
    purpose=Purpose.CLIENT_AUTH,
    cafile='/incoming.cacerts')
server_ssl_context.load_cert_chain(ssl_config['cert'],
                                   keyfile=ssl_config['key'],
                                   password=None)
server_ssl_context.verify_mode = ssl.CERT_OPTIONAL
server_ssl_context.check_hostname = False
```

The first function call states that we are a *server* performing *client
authentication* (`Purpose.CLIENT_AUTH`). We also state that anyone who sends requests to us will be
identified by a certificate that is trusted by our certificate database: `/incoming.cacerts` (which
is a file).

`load_cert_chain` states where to find the certificates and secret key that prove who we are. The
certificate and secret key together are like a property title that proves someone owns a house. The
`password=None` means that our secret key has no password. Some keys are themselves locked by a
password.

`verify_mode` means what do we expect our clients to have. `CERT_OPTIONAL` means anonymous clients
are OK. This is how servers normally operate (https://google.com does not care who you are).

`check_hostname` means should we verify that the client certificate matches the client's
hostname. Since we allow anonymous clients, this must be `False`.

---

`test-address.py` is a gross hack. It will disappear in subsequent PRs. For now, I pushed all the
gross hack stuff into that file alone.

### DIFF
--- a/address/address/address.py
+++ b/address/address/address.py
@@ -9,7 +9,7 @@ import kubernetes_asyncio as kube
 from gear import (setup_aiohttp_session, web_authenticated_developers_only,
                   rest_authenticated_users_only)
 from hailtop.config import get_deploy_config
-from hailtop.tls import get_in_cluster_server_ssl_context
+from hailtop.tls import internal_server_ssl_context
 from hailtop.hail_logging import AccessLogger
 from hailtop import aiotools
 from web_common import setup_aiohttp_jinja2, setup_common_static_routes, render_template
@@ -161,4 +161,4 @@ def run():
                 host='0.0.0.0',
                 port=5000,
                 access_log_class=AccessLogger,
-                ssl_context=get_in_cluster_server_ssl_context())
+                ssl_context=internal_server_ssl_context())
--- a/address/test/test_address.py
+++ b/address/test/test_address.py
@@ -1,8 +1,9 @@
+import ssl
 import pytest
 import aiohttp
 from hailtop.auth import service_auth_headers
 from hailtop.config import get_deploy_config
-from hailtop.tls import in_cluster_no_hostname_checks_ssl_client_connection
+from hailtop.tls import _get_ssl_config
 from hailtop.utils import request_retry_transient_errors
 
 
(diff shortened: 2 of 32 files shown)


jigold (changes requested):
This looks fine to me. I'm just not sure why there were so many different functions before for the ssl_context and now you're able to have 3. Were they just not used anywhere?

Also, you have an error somewhere because create_billing_projects failed

danking:
One of them was unused. Two of them (one SSLContext and one session) was only for address, so I in-lined it there. Address needs to change anyway.  All the `context_specific..` garbage was over complicated. We should always get the SSLContext or HTTP Session that works in your environment (k8s, gce, or external).

danking:
The test is fixed, we don't use encryption to talk to the internal gateway. I had to fix https.client_session to handle the gce case.

jigold (dismissed):

danking:
I disabled hostname checks for address, since address communicates directly by IP address, hostname checks don't work (there's no hostname)

jigold:
Why did you close this? Should I move on to the next PR?

danking:
Yeah, I reorganized it into several smaller PRs

jigold:
I'm confused. I was ready to approve this pending the test passing. Now there's stacked PRs that include these changes. What PR should I look at?

danking:
ah you're right, I got turned around this was the new broken up PR.

jigold:
Not sure why the BatchPoolExecutor test failed, but approving.

jigold (approved):

PR #9863, opened 2021-01-07
[tls2] implement blocking client session in terms of aiohttp

danking:
This change replaces uses of `requests.Session` with uses of
`BlockingClientSession`. The latter is a class introduced by this
PR. BlockingClientSession is a shim around
[`aiohttp.ClientSession`](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L167). I
also needed these new classes:

- `BlockingClientResponse`, representing a response to an HTTP request. Wraps `aiohttp.ClientResponse`.

- `BlockingClientWebSocketResponse`, representing a web socket session. Wraps
  `aiohttp.ClientWebSocketResponse`

- `BlockingClientresponseContextManager` and
  `BlockingClientWebSocketResponseContextManager` which wrap the context manager
  returned by `aiohttp_client_session.post(...)` or
  `....ws_connect(...)`. They wrap the `__aenter__` and `__aexit__` calls and
  also wrap the value returned by `__aenter__`.

### DIFF
--- a/hail/python/hailtop/httpx.py
+++ b/hail/python/hailtop/httpx.py
@@ -53,7 +53,7 @@ class BlockingClientResponse:
         return self.client_response.history()
 
     def __repr__(self) -> str:
-        return f'BlcokingClientRepsonse({repr(self.client_response)})'
+        return f'BlockingClientRepsonse({repr(self.client_response)})'
 
     @property
     def status(self) -> int:
--- a/hail/python/hailtop/httpx.py
+++ b/hail/python/hailtop/httpx.py
@@ -1,9 +1,9 @@
-import requests
+from typing import Any, Tuple, Optional, Type, TypeVar, Generic, Callable
+from types import TracebackType
 import aiohttp
-from requests.adapters import HTTPAdapter
-from urllib3.poolmanager import PoolManager
 
-from .tls import internal_client_ssl_context, external_client_ssl_context, _get_ssl_config
+from .utils import async_to_blocking
+from .tls import internal_client_ssl_context, external_client_ssl_context
 from .config.deploy_config import get_deploy_config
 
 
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/httpx.py
```diff
@@ -24,34 +24,241 @@ def client_session(*args, **kwargs) -> aiohttp.ClientSession:
...
+            encoding=encoding, loads=loads, content_type=content_type))
+
+    def __del__(self):
+        self.client_response.__del__()
+
+    def history(self) -> Tuple[aiohttp.ClientResponse, ...]:
+        return self.client_response.history()
+
+    def __repr__(self) -> str:
+        return f'BlcokingClientRepsonse({repr(self.client_response)})'
```
jigold, line 56: typo.

hail/python/hailtop/httpx.py
```diff
@@ -24,34 +24,241 @@ def client_session(*args, **kwargs) -> aiohttp.ClientSession:
...
+    def __exit__(self,
+                 exc_type: Optional[Type[BaseException]],
+                 exc: Optional[BaseException],
+                 tb: Optional[TracebackType]) -> None:
+        async_to_blocking(self.context_manager.__aexit__(exc_type, exc, tb))
+
+
+class BlockingClientResponseContextManager(AsyncToBlockingContextManager):
+    def __init__(self, context_manager):
+        super().__init__(context_manager, BlockingClientResponse)
```
jigold, line 163: BlockingClientResponse isn't Callable[[T], U]. How does this work?

jigold (changes requested):

hail/python/hailtop/httpx.py
danking: It is a Callable, the constructor takes exactly one argument. You can verify this with `mypy` and the the magic `reveal_type` command:
```
(base) # cat foo.py
from hailtop.httpx import BlockingClientResponse

reveal_type(BlockingClientResponse)
(base) # mypy foo.py
foo.py:3: note: Revealed type is 'def (client_response: aiohttp.client_reqrep.ClientResponse) -> hailtop.httpx.BlockingClientResponse'
```

jigold (approved):

PR #9864, opened 2021-01-08
[tls] default to raise_for_status=True

danking:
We almost always want to raise an error on a bad status.


batch/batch/batch.py
```diff
@@ -305,7 +307,7 @@ async def unschedule_job(app, record):
             break
         try:
             async with aiohttp.ClientSession(
-                    raise_for_status=True, timeout=aiohttp.ClientTimeout(total=60)) as session:
+                    timeout=aiohttp.ClientTimeout(total=60)) as session:
```
jigold, line 310: Why can you delete the raise_for_status here and the change directly below. They're not `client_session`

jigold (dismissed):

jigold (approved):

jigold:
The stacked pr tag is still on. Can you remove it? I just want to make sure this is the next one to go in.

PR #9866, opened 2021-01-08
[http] set project-wide default http timeout to 5 seconds

danking:


build.yaml
```diff
@@ -142,7 +142,7 @@ steps:
      - from: /io/repo/git_version
        to: /git_version
      - from: /io/repo/address/test
-       to: /repo/address/test
+       to: /repo/address/
```
jigold, line 145: Was this not used before?

jigold (changes requested):

build.yaml
danking: I don't know how the tests were working before. They had an extra layer of directories.

jigold (approved):

PR #9869, opened 2021-01-08
wip

danking:


batch/batch/driver/instance.py
```diff
@@ -39,121 +43,92 @@ async def create(app, name, activation_token, worker_cores_mcpu, zone):
...
+        self.instance_pool.instances_by_last_updated.add(self)
 
-        if rv['rc'] == 1:
-            log.info(f'{self} with in-memory state {self._state} could not be marked deleted; {rv}')
-            assert rv['cur_state'] == 'deleted'
+    @property
+    def failed_request_count(self):
+        return self._failed_request_count
 
-        self.instance_pool.adjust_for_remove_instance(self)
```
jigold, line 118: The reason why there's always an adjust for remove instance and an add instance is because you can't change the sort key of a sorted collection while the object is in that collection

batch/batch/driver/instance_pool.py
```diff
@@ -152,46 +166,100 @@ async def configure(
...
+        self.n_instances_by_state[instance.state] += 1
+        self.instances_by_last_updated.add(instance)
+        if instance.state in ('pending', 'active'):
+            self.live_free_cores_mcpu += max(0, instance.free_cores_mcpu)
+            self.live_total_cores_mcpu += instance.cores_mcpu
+        if instance.is_healthy():
+            self.healthy_instances_by_free_cores.add(instance)
 
-        self.n_instances_by_state[instance.state] -= 1
+    # async def create_instance should be here but it is huge currently
```
jigold, line 183: You can't do this here. You call `add` every time you add an instance such as on startup.

PR #9871, opened 2021-01-10
[batch] fix job page

danking:
We updated the job spec but forgot to update the display of jobs.


danking:
Verified it works. On Monday, I'll think about how to avoid this kind of bug.

PR #9873, opened 2021-01-11
[docker] specify jedi version explicitly

jigold:


danking (approved):

danking (changes requested):
I don't think this is necessary actually. Jupyter-client fixed their dependency in 6.11.1

PR #9874, opened 2021-01-11
[query] NDArray Reshape Simplification

johnc1231:
I think there are a lot of situations like this, like `hl.nd.zeros(shape)`, which is implemented as:

`hl.nd.array(hl.range(total_num_elements)).map(lambda x: 0).reshape(shape)`. No reason to actually generate code for reshaping in these situations.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Simplify.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Simplify.scala
@@ -237,7 +237,9 @@ object Simplify {
     case ToStream(Let(name, value, ToArray(x)), _) if x.typ.isInstanceOf[TStream] =>
       Let(name, value, x)
 
+    //TODO: A better system would push all reshapes down to the MakeNDArray calls if there aren't any shape dependent nodes in between.
     case NDArrayReshape(MakeNDArray(data, _, rowMajor), reshapedShape) => MakeNDArray(data, reshapedShape, rowMajor)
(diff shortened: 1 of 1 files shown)


hail/src/main/scala/is/hail/expr/ir/Simplify.scala
```diff
@@ -237,6 +237,10 @@ object Simplify {
     case ToStream(Let(name, value, ToArray(x)), _) if x.typ.isInstanceOf[TStream] =>
       Let(name, value, x)
 
+    //TODO: A better system would push all reshapes down to the MakeNDArray calls if there aren't any shape dependent nodes in between.
+    case NDArrayReshape(MakeNDArray(data, _, rowMajor), reshapedShape) => MakeNDArray(data, reshapedShape, rowMajor)
+    case NDArrayReshape(NDArrayMap(MakeNDArray(data, _, rowMajor), vName, body), newShape) => NDArrayMap(MakeNDArray(data, newShape, rowMajor), vName, body)
```
patrick-schultz, line 242: You could make this closer to the goal in your TODO by making this rule just swap the reshape with the map, and let the previous rule do the rest.

patrick-schultz (changes requested):
Ah, nice!

patrick-schultz (approved):

johnc1231:
Ooof, this introduces a bug if the shape contains a -1, since `NDArrayReshape` is what handles that.

Issue #9872, opened 2021-01-11
Artifact Registry


vladsaveliev:
Hi,

I was wondering if you had thoughts on supporting Artifact Registry as an image source for Hail Batch? [Artifact Registry](https://cloud.google.com/artifact-registry/) is meant to replace Container Registry in the GCloud universe, and the container tags are formatted using a different domain name: `<region>-docker.pkg.dev/<project>/<repository>/<image>`.

Vlad

PR #9875, opened 2021-01-12
[batch] pass credentials for both GCR and Artifact Registry images

lgruen:
The Artifact Registry is useful to avoid network egress costs for regions like Australia, where GCR doesn't provide local hosting.

Partially addresses #9872.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -341,7 +341,7 @@ class Container:
     async def run(self, worker):
         try:
             async with self.step('pulling'):
-                if self.image.startswith('gcr.io/'):
+                if self.image.startswith('gcr.io/') or 'docker.pkg.dev/' in self.image:
                     key = base64.b64decode(
                         self.job.gsa_key['key.json']).decode()
                     auth = {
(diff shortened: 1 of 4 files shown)


batch/batch/worker/worker.py
```diff
@@ -341,7 +341,7 @@ async def get_container_status(self):
     async def run(self, worker):
         try:
             async with self.step('pulling'):
-                if self.image.startswith('gcr.io/'):
+                if self.image.startswith('gcr.io/') or 'docker.pkg.dev/' in self.image:
```
danking, line 344: hmm, we really want this to be the host portion of the image not the whole image.

Can we do instead:
```
image_parts = image.split('/')
image_hostname = ''
if len(image_parts) > 1:
    image_hostname = image_parts[0]
if image_hostname == 'gcr.io' or image_hostname.endswith('docker.pkg.dev'):
```

danking (changes requested):

batch/batch/worker/worker.py
lgruen: Good point! Moved this to a small utility function.

danking (approved):

danking:
Thanks for the improvement!

lgruen:
Thanks for the review! Could you please also merge the PR, as I don't have the necessary permissions?

danking:
Bah, sorry, CI does all merges and there's a bug introduced in a python dependency of ours. I'll get that fixed.

danking:
Ah, the fix was just merged, CI should test and merge this within 30 minutes.

lgruen:
Awesome, thanks a lot for debugging!

PR #9879, opened 2021-01-12
[query] Add subset method to SBaseStructCode

chrisvittal:
The subset method is used to construct SSubsetStructCode

### DIFF
--- a/hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SStruct.scala
+++ b/hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SStruct.scala
@@ -2,8 +2,9 @@ package is.hail.types.physical.stypes.interfaces
 
 import is.hail.asm4s.Code
 import is.hail.expr.ir.{EmitCodeBuilder, IEmitSCode}
-import is.hail.types.physical.{PBaseStruct, PBaseStructValue, PSettable}
+import is.hail.types.physical.PBaseStruct
 import is.hail.types.physical.stypes.{SCode, SSettable, SType, SValue}
+import is.hail.types.physical.stypes.concrete.{SSubsetStruct, SSubsetStructCode}
 
 trait SStruct extends SType {
   override def fromCodes(codes: IndexedSeq[Code[_]]): SBaseStructCode
@@ -23,9 +24,15 @@ trait SBaseStructValue extends SValue {
   def loadField(cb: EmitCodeBuilder, fieldName: String): IEmitSCode = loadField(cb, pt.fieldIdx(fieldName))
 }
 
-trait SBaseStructCode extends SCode {
+trait SBaseStructCode extends SCode { self =>
+  def st: SStruct
+
   def memoize(cb: EmitCodeBuilder, name: String): SBaseStructValue
 
   def memoizeField(cb: EmitCodeBuilder, name: String): SBaseStructValue
-}
 
+  def subset(fieldNames: String*): SSubsetStructCode = {
+    val st = SSubsetStruct(self.st, fieldNames.toIndexedSeq)
+    new SSubsetStructCode(st, self.asPCode.asBaseStruct) // FIXME, should be sufficient to just use self here
+  }
+}


chrisvittal:
This is completely unused, but will need to be used in the ordering changes.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SStruct.scala
```diff
@@ -23,9 +24,15 @@ trait SBaseStructValue extends SValue {
   def loadField(cb: EmitCodeBuilder, fieldName: String): IEmitSCode = loadField(cb, pt.fieldIdx(fieldName))
 }
 
-trait SBaseStructCode extends SCode {
+trait SBaseStructCode extends SCode { self =>
+  def st: SStruct
```
tpoterba, line 28: this isn't valid -- these can represent tuples too.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SStruct.scala
```diff
@@ -23,9 +24,15 @@ trait SBaseStructValue extends SValue {
...
+
   def memoize(cb: EmitCodeBuilder, name: String): SBaseStructValue
 
   def memoizeField(cb: EmitCodeBuilder, name: String): SBaseStructValue
-}
 
+  def subset(fieldNames: String*): SSubsetStructCode = {
+    val st = SSubsetStruct(self.st, fieldNames.toIndexedSeq)
```
tpoterba, line 35: match on `st` and fail if it's a tuple, maybe?

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SStruct.scala
```diff
@@ -23,9 +24,15 @@ trait SBaseStructValue extends SValue {
...
+    new SSubsetStructCode(st, self.asPCode.asBaseStruct) // FIXME, should be sufficient to just use self here
```
tpoterba, line 36: yep, OK for now.

tpoterba (changes requested):
Nice! minor comment about struct/tuple clashes.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SStruct.scala
chrisvittal: There's no issue here. There is no STuple, the code need only be a base struct.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SStruct.scala
chrisvittal: There is no STuple.

tpoterba: right, I mean match on the virtual type

tpoterba: Oops, haven't PRed that yet. You can add a `def virtualType: Type = pType.virtualType` to the `sType` trait.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SStruct.scala
tpoterba: ah, right. Carry on.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SStruct.scala
chrisvittal: I don't see the purpose of such a check.

tpoterba: OK, I think you're right that we don't need this right now. I'll keep an eye out for usages that might motivate this check though.

tpoterba (approved):

PR #9880, opened 2021-01-12
[sync.py] use add not append for sets

danking:
Also add more logging

### DIFF
--- a/devbin/sync.py
+++ b/devbin/sync.py
@@ -26,11 +26,11 @@ DEVBIN = os.path.abspath(os.path.dirname(__file__))
 
 class Sync:
     def __init__(self, paths: List[Tuple[str, str]]):
-        self.pods_list: Set[Tuple[str, str]] = set()
+        self.pods: Set[Tuple[str, str]] = set()
         self.paths = paths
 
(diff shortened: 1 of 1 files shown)


devbin/sync.py
```diff
@@ -42,6 +43,7 @@ async def sync_and_restart_pod(self, pod, namespace):
         log.info(f'reloaded {pod}@{namespace}')
 
     async def initialize_pod(self, pod, namespace):
+        log.info(f'initilizing {pod}@{namespace}')
```
jigold, line 46: typo in initializing.

devbin/sync.py
```diff
@@ -50,7 +52,7 @@ async def initialize_pod(self, pod, namespace):
         except CalledProcessError:
             log.warning(f'could not initialize {namespace}/{pod}', exc_info=True)
             return
-        self.pods_list.append((pod, namespace))
+        self.pods_list.add((pod, namespace))
```
jigold, line 55: Would it be better to name this pods rather than pods_list?

jigold (changes requested):

jigold (approved):

PR #9882, opened 2021-01-13
[hail] remove dev requirement due to jupyter changes

danking:
Jupyter now requires tornado>=6. We added the tornado<6 requirement to prevent previous issues in
which the notebook package did not correctly specify its dependency on tornado<6.


danking:
bump

tpoterba (approved):

PR #9883, opened 2021-01-14
[query] Clean up PNDArray / Improve loadElement performance

johnc1231:
This PR:

- Pushes code builders through PNDArray interface instead of method builders
- Starts switching away from `PNDArray.data.load` to using methods like `dataPArrayPointer` and `dataFirstElementPointer`. All `dataPArrayPointer` calls will go away when ndarrays are no longer backed by `PArray`. 
- Speeds up repeated calls to `loadElement` on `SNDArrayPointerSettable`, which speeds up linear regression nd benchmark ~12%. Now we are approximately 60% slower than breeze linear regression.'
- Removes `CodePTuple`, since all instances of its use are removed now and it predated the current `PCode` system

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -847,7 +847,7 @@ class Emit[C](
                 }
 
                 cb.ifx(isRowMajor, {
-                  cb += xP.makeRowMajorStridesBuilder(shapeCodeSeq1, cb)(srvb)
+                  cb += xP.makeRowMajorStridesStruct(shapeCodeSeq1, cb)(srvb)
                 }, {
                   cb += xP.makeColumnMajorStridesBuilder(shapeCodeSeq1, cb)(srvb)
                 })
@@ -1317,10 +1317,10 @@ class Emit[C](
 
             val hShapeArray = FastIndexedSeq[Value[Long]](N, M)
             val hShapeBuilder = hPType.makeShapeBuilder(hShapeArray)
-            val hStridesBuilder = hPType.makeRowMajorStridesBuilder(hShapeArray, cb)
+            val hStridesBuilder = hPType.makeRowMajorStridesStruct(hShapeArray, cb)
 
             val tauShapeBuilder = tauPType.makeShapeBuilder(FastIndexedSeq(K))
-            val tauStridesBuilder = tauPType.makeRowMajorStridesBuilder(FastIndexedSeq(K), cb)
+            val tauStridesBuilder = tauPType.makeRowMajorStridesStruct(FastIndexedSeq(K), cb)
 
             val h = hPType.construct(hShapeBuilder, hStridesBuilder, aAddressDGEQRF, cb, region.code)
             val tau = tauPType.construct(tauShapeBuilder, tauStridesBuilder, tauAddress, cb, region.code)
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -992,8 +992,6 @@ class Emit[C](
                       (lStackVars :+ n :+ k, rStackVars :+ k :+ m)
                   }
 
-                  val lElem = leftPVal.loadElement(lIndices, cb)
-                  val rElem = rightPVal.loadElement(rIndices, cb)
                   val kLen = cb.newField[Long]("ndarray_matmul_kLen")
 
                   def multiply(l: PCode, r: PCode): Code[_] = {
@@ -1011,9 +1009,11 @@ class Emit[C](
 
                   cb.assign(kLen, lShape(lPType.nDims - 1))
                   cb.assign(element, numericElementType.zero)
-                  cb.forLoop(cb.assign(k, 0L), k < kLen, cb.assign(k, k + 1L),
+                  cb.forLoop(cb.assign(k, 0L), k < kLen, cb.assign(k, k + 1L), {
+                    val lElem = leftPVal.loadElement(lIndices, cb)
+                    val rElem = rightPVal.loadElement(rIndices, cb)
                     cb.assign(element, numericElementType.add(multiply(lElem.asPCode, rElem.asPCode), element))
-                  )
+                  })
 
                   PCode(outputPType.elementType, element)
                 }
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -883,7 +883,7 @@ class Emit[C](
                   srvb.addLong(if (index < childPType.nDims) childStrides(index) else 0L),
                   srvb.advance())
               })
-          }, childPType.dataArrayPointer(pndVal.tcode[Long]), cb, region.code)
+          }, childPType.dataPArrayPointer(pndVal.tcode[Long]), cb, region.code)
         }
 
       case NDArrayRef(nd, idxs, errorId) =>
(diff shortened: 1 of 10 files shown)


hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
```diff
@@ -45,16 +45,23 @@ object SNDArrayPointerSettable {
 
 class SNDArrayPointerSettable(val st: SNDArrayPointer, val a: Settable[Long]) extends PNDArrayValue with PSettable {
   val pt: PNDArray = st.pType
+  private var shape: IndexedSeq[Value[Long]] = null
+  private var strides: IndexedSeq[Value[Long]] = null
+  private var dataFirstElement: Settable[Long] = null
```
johnc1231, line 50: Is this an ok thing to do Tim? You avoid the nulls when saving the length of arrays by passing in an uninitialized variable. I could do that too, but honestly I'd rather get an NPE at hail compile time than be confused about provenance of a 0 pointer at run time.

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
```diff
@@ -84,35 +84,31 @@ final case class PCanonicalNDArray(elementType: PType, nDims: Int, required: Boo
     ))
   }
 
-  def makeColumnMajorStridesBuilder(sourceShapeArray: IndexedSeq[Value[Long]], mb: EmitMethodBuilder[_]): StagedRegionValueBuilder => Code[Unit] = { srvb =>
-    val runningProduct = mb.newLocal[Long]()
-    val tempShapeStorage = mb.newLocal[Long]()
+  def makeColumnMajorStridesBuilder(sourceShapeArray: IndexedSeq[Value[Long]], cb: EmitCodeBuilder): StagedRegionValueBuilder => Code[Unit] = { srvb =>
```
tpoterba, line 87: This function takes a codebuilder but returns a function to Code[Unit]. I think this is invalid. It might work because all you're doing with the cb is making a local variable (and cb.emb is the same MB as srvb.mb probably), but we shouldn't open ourselves up to misuse like this.

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
```diff
@@ -84,35 +84,31 @@ final case class PCanonicalNDArray(elementType: PType, nDims: Int, required: Boo
...
         )
       }
     )
   }
 
-  def makeRowMajorStridesBuilder(sourceShapeArray: IndexedSeq[Value[Long]], mb: EmitMethodBuilder[_]): StagedRegionValueBuilder => Code[Unit] = { srvb =>
-    val runningProduct = mb.newLocal[Long]()
-    val tempShapeStorage = mb.newLocal[Long]()
-    val computedStrides = (0 until nDims).map(_ => mb.genFieldThisRef[Long]())
+  def makeRowMajorStridesBuilder(sourceShapeArray: IndexedSeq[Value[Long]], cb: EmitCodeBuilder): StagedRegionValueBuilder => Code[Unit] = { srvb =>
```
tpoterba, line 102: same

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
```diff
@@ -151,16 +146,30 @@ final case class PCanonicalNDArray(elementType: PType, nDims: Int, required: Boo
...
+      bytesAway + data.pType.firstElementOffset(dataStore, data.pType.loadLength(dataStore))
     ))
   }
 
-  def setElement(indices: IndexedSeq[Value[Long]], ndAddress: Value[Long], newElement: Code[_], mb: EmitMethodBuilder[_]): Code[Unit] = {
-    Region.storeIRIntermediate(this.elementType)(getElementAddress(indices, ndAddress, mb), newElement)
+  private def getElementAddressFromDataPointerAndStrides(indices: IndexedSeq[Value[Long]], dataFirstElementPointer: Value[Long], strides: IndexedSeq[Value[Long]], cb: EmitCodeBuilder): Code[Long] = {
+    val bytesAway = cb.newLocal[Long]("nd_get_element_address_bytes_away")
+
+    coerce[Long](Code(
```
tpoterba, line 156: append this stuff to the CB instead of using the `Code(...)` constructor?

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
tpoterba: the design you've got here (create the variables in `store`) slightly breaks the mental model of settables, because they should be thought of as a single variable or set of variables which you can assign to multiple times.

I think it's possible to write down an example of code generation where this actually breaks the interface of settables, because you end up with loads referencing the wrong (old) variables.

tpoterba (changes requested):

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
tpoterba: yes, here:

```

val nd = new NDArrayPointerSettable(...)

cb.assign(nd, someInitialValue)
cb.whileLoop(someBoolean(nd.load(), {
  
  // do some work
  cb.assign(nd, someNewValue)
})
```

The `nd.load()` in the condition of the while loop references different variables than the ones stored to on the last line. We have to use the same vars.

tpoterba: I think I see what you're trying to do, though -- you can just set a `private var storedAtLeastOnce: Boolean = false`, set that to true in `store`, and assert it's true in `load`.

johnc1231: Resolved this by changing to match the pattern of `SIndexablePointerSettable`. Decided it was less complicated to do the same thing in both places.

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
johnc1231: This was a lot less trivial to fix then I thought. There were several places calling `loadElement` (which calls this method as a helper) in a way that assumed it was returning a lazily evaluated code and wouldn't have any side effects. The changes to `NDArrayMatMul` in the emitter and to `EColumnMajorNDArray` both were necessary to avoid this.

johnc1231:
@tpoterba , should be good now.

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
tpoterba: oof, yup, that's a risk.

tpoterba: A pattern we used a lot was emitting a lazily-inserted Code[T] at the start of a block, when we could have generated it in a loop or something. This is painful to fix.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -1009,8 +992,6 @@ class Emit[C](
                       (lStackVars :+ n :+ k, rStackVars :+ k :+ m)
                   }
 
-                  val lElem = leftPVal.loadElement(lIndices, cb)
-                  val rElem = rightPVal.loadElement(rIndices, cb)
```
tpoterba, line 1013: yeah like here :)

johnc1231: Yeah exactly. I think I found all the relevant ones to this change though, seems to pass all tests now.

tpoterba (approved):

PR #9886, opened 2021-01-15
[query] Compute covariate missingness once in linear_reg_rows_nd

tpoterba:


johnc1231:
I see, it wasn't doing redundant work, just generating redundant IR by regenerating the IR to load covariates per set of phenotypes. This would only affect chained linear regression, since that's the only time there's more than one `one_y_field_name_set in y_field_names`.

hail/python/hail/methods/statgen.py
```diff
@@ -456,8 +456,13 @@ def make_one_y_matrix(ys_and_covs_to_keep, one_y_field_name_set):
         return hl.nd.array(ys_and_covs_to_keep.map(lambda struct: array_from_struct(struct, one_y_field_name_set)))
 
     def setup_globals(ht):
-        list_of_ys_and_covs_to_keep_with_indices = \
-            [hl.enumerate(ht[sample_field_name]).filter(lambda struct_with_index: all_defined(struct_with_index[1], one_y_field_name_set + cov_field_names)) for one_y_field_name_set in y_field_names]
+        all_covs_defined = ht[sample_field_name].map(lambda sample_struct: all_defined(sample_struct, cov_field_names))
```
johnc1231, line 459: I believe tests are failing because `cov_field_names` can be empty, so if you pass that to`all_defined` you hit that case where you'd need to do `hl.empty_array`.  This isn't a problem in the old version since there has to be at least one cov or phenotype.

johnc1231 (dismissed):

hail/python/hail/methods/statgen.py
johnc1231: If you'd like I can just fix and push

johnc1231 (approved):

PR #9892, opened 2021-01-19
[query] Unstaged Store Java Object

johnc1231:
I am trying to move towards removing the `fundamentalType` interface from `PType`s. As a first pass, this PR will remove RVB's use of `fundamentalType` by delegating the `addAnnotation`'s implementation to a new method, `unstagedStoreJavaObjectAtAddress`.

WIP until I see tests are passing and clean up comments.


tpoterba:
nice!

hail/src/main/scala/is/hail/annotations/RegionValueBuilder.scala
```diff
@@ -436,12 +441,14 @@ class RegionValueBuilder(var region: Region) {
...
-            case r: Row =>
-              addRow(t, r)
-          }
+//          a match {
+//            case ur: UnsafeRow if currentType() == ur.t =>
+//              addUnsafeRow(ur.t, ur)
+//            case r: Row =>
+//              addRow(t, r)
+//          }
+          currentType().unstagedStoreJavaObjectAtAddress(currentOffset(), a, region)
```
tpoterba, line 450: I think this should be the only implementation of addAnnotation!

johnc1231: I agree, it will be! I have a `return` on line 417 right now, I just haven't removed the old code yet.

tpoterba: oh, great!

johnc1231:
This should be done now @tpoterba. RVB now no longer expects fundamental types, and we've removed many places where RVBs were being used in favor of just calling the new unstagedAdd methods directly.

johnc1231:
Agh, the unsafeRow and UnsafeIndexedSeq optimizations must be wrong, getting out of bounds memory errors.

hail/src/main/scala/is/hail/annotations/RegionValueBuilder.scala
```diff
@@ -347,25 +328,13 @@ class RegionValueBuilder(var region: Region) {
...
-
   def selectRegionValue(fromT: PStruct, fromFieldIdx: Array[Int], fromRV: RegionValue) {
     selectRegionValue(fromT, fromFieldIdx, fromRV.region, fromRV.offset)
   }
 
   def selectRegionValue(fromT: PStruct, fromFieldIdx: Array[Int], region: Region, offset: Long) {
-    val t = fromT.typeAfterSelect(fromFieldIdx).fundamentalType
-    assert(currentType().setRequired(true) == t.setRequired(true))
+    val t = fromT.typeAfterSelect(fromFieldIdx)
+    assert(currentType().setRequired(true) == t.setRequired(true), s"${currentType()} != ${t}")
```
tpoterba, line 337: probably don't want the allocating assert here.

tpoterba: (may not matter since typeAfterSelect is allocating...)

tpoterba (approved):

PR #9897, opened 2021-01-21
[hail] extra assert in EType.fromTypeAndAnalysis

danking:


danking:
I'm not sure this is kosher, but I think it is.

danking:
I'm not sure this is kosher, but I think it is.

tpoterba (dismissed):

tpoterba:
er, need to fix size

danking:
I added a size to RStruct.

tpoterba (approved):

PR #9899, opened 2021-01-21
[gear][hailtop][web_common] enable mypy for shared modules

danking:
Without this empty file, mypy ignores the type annotations present in these modules. Adding
this file enables downstream modules (like CI and batch) to check the type annotations
provided by hailtop, gear, and web_common.

See more information here: https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages


daniel-goldstein (approved):
LGTM

PR #9901, opened 2021-01-22
[auth] usernames can't contain dots, otherwise the GSA creation fails

lgruen:


danking:
@jigold I think this is OK, but you were in this more recently than me.

jigold (approved):

PR #9904, opened 2021-01-22
[hail] add jar-only hail build job

danking:
Building the JAR only takes 5 minutes. Query and Shuffler don't need the other stuff, so it's better for services team if we can skip all that. I don't really care much about the added 5 compute minutes. It's worth the increase in developer velocity.


johnc1231 (approved):
Good solution, thanks :+1:

danking:
This is already paying dividends!

PR #9905, opened 2021-01-22
[ci] time most commands in the image build step

danking:
@daniel-goldstein , this is a relatively tiny change to CI, but I think it's a chance to take a peek at one part of it. Send me your questions!


daniel-goldstein (approved):
I think it'll take some hacking on it for me to really grock it, but this was a nice intro, thanks! In general I'm not so clear on what we use service-accounts vs other (user?) accounts for, so an overview on accounts I should be aware of would be useful.

PR #9906, opened 2021-01-22
[hail-ubuntu] retry pip install

danking:
I do not know why the retries setting in pip.conf did not catch https://ci.hail.is/batches/167314/jobs/27, but more retries never hurt anyone.

Another CI-related PR. This one changes the base image of everything else: hail-ubuntu. It's an ubuntu image with two scripts that make pip and apt more resilient. Take a look at docker/hail-ubuntu.


daniel-goldstein:
> I do not know why the retries setting in pip.conf did not catch https://ci.hail.is/batches/167314/jobs/27, but more retries never hurt anyone.

Is the whole pip process crashing because of the exception? Perhaps a different kind of crash that pip can't recover from and retry

> Another CI-related PR. This one changes the base image of everything else: hail-ubuntu. It's an ubuntu image with two scripts that make pip and apt more resilient. Take a look at docker/hail-ubuntu.

Should we do this for `apt` too?

danking:
I think this is just pip not honoring retries everywhere. The crash appears to occur while downloading a dependency.

Sure, I can add a retry to apt as well, though apt seems to be more careful about retrying on its own.

danking:
@daniel-goldstein, if you leave a review, https://scorecard.hail.is will tell me that I need to take action on this PR. It will also remove the PR from your scorecard page.

daniel-goldstein (dismissed):
I'm fine leaving `apt` to do its thing if it's not posing problems at the moment, but fine leaving the rest as is (except you seem to be having a permissions issue with `retry`)

daniel-goldstein:
I wasn't very clear initially, I think if `apt` hasn't shown reason to be retried then we can leave it without a retry. If it ever seriously acts up there's one less thing in the way.

danking:
OK, I reverted the retry on apt-get install, but kept the refactoring that makes apt-get update and pip use the same retry script.

daniel-goldstein (dismissed):

daniel-goldstein:
Was this supposed to go on this branch?

danking:
something is wrong and I need more debugging info; I should've dismissed your review

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -345,7 +345,7 @@ async def jobs(self, q=None):
                 break
 
     async def get_job_log(self, job_id: int):
-        return self._client.get_job_log(self.id, job_id)
+        return await self._client.get_job_log(self.id, job_id)
```
danking, line 348: noticed this while debugging.

daniel-goldstein: Nice catch. Could you add a return type annotation? Hopefully that can mitigate these bugs

daniel-goldstein (changes requested):
Looks good, just one ask. Could you also explain what was causing the problems up until now and what resolved them? This doesn't look much different to me on its face

danking:
At some point, we should think about how to improve the discoverability and machine-verifiability of our APIs. Currently the tightest type of job log is rather complex. If the performance is OK, I think we should move towards classes that define the request and response types of each call.

---

The main difference is `hail-pip-install` having `retry`. If pip exits with a non-zero exit code, we'll just rerun the command exactly, at most four more times. This mitigates missing retry logic in `pip` itself. For example, [this job](https://ci.hail.is/batches/167314/jobs/27) failed because pip encountered a connection reset while downloading a file. Ideally, pip would simply retry the download. Since we don't control the pip source code, I use a retry that treats all of pip as a black box.

There's definitely a failure mode: if you specify a package that doesn't exist, pip will error five times in a row and take ~30 seconds before the retry logic gives up. I'm OK with this because pip should basically never fail for legitimate reasons.

daniel-goldstein (approved):

danking:
https://github.com/PyCQA/pylint/issues/4131

Python packaging is such a garbage fire. I think this PR will finally succeed.

PR #9907, opened 2021-01-24
[batch] mount Hail tokens in user job containers

lgruen:
This is necessary for submitting a "nested batch" from the user's job.

### DIFF
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -489,7 +489,8 @@ class ServiceBackend(Backend):
                                     timeout=job._timeout,
                                     gcsfuse=job._gcsfuse if len(job._gcsfuse) > 0 else None,
                                     env=env_vars,
-                                    requester_pays_project=batch.requester_pays_project)
+                                    requester_pays_project=batch.requester_pays_project,
+                                    mount_tokens=True)
 
             n_jobs_submitted += 1
 


jigold:
@danking What do you think about this change? I think it's fine, but if it's not fine, then we need to lock down this feature as anyone can just set it to True in the batch_client.

danking:
Can we make it a job and/or batch level configuration? The user obviously can do whatever they like with their tokens. However, since most users don't need them, I prefer our default to be to not expose them. I think we just need a new method and attribute on `Job` and `Batch` that mirrors, say, `image` or `memory`.

lgruen:
Somewhat related to this is the deployment configuration. That's a little easier to handle on the user's side by including it in the job's image directly, but it would be pretty convenient if it could be mounted automatically as well (as `/deploy-config/deploy-config.json`, with `"location": "gce"` and the `domain` set accordingly).

Should that be another method on `Job` / `Batch` or should all of this functionality be hidden behind something more abstract to enable nested batches (that's really the main use case, I suppose).

danking:
I think we should just always mount the deploy config. That might trigger some latent bugs where we don't do quite the right thing, but we can fix those.

jigold:
Did we decide the change is okay as is?

lgruen:
If we always mount a deploy config (which sounds like a great idea), I think it would be nice to also always mount the user tokens too (similar to how the GSA key is always mounted). That way, nested batches "just work".

But happy to make it a configurable option, if you prefer that.

danking:
OK, let's always mount it. People download and run software on their laptops which can read their gcloud and hail credentials awnyway!

jigold (approved):

lgruen:
> I think we should just always mount the deploy config. That might trigger some latent bugs where we don't do quite the right thing, but we can fix those.

@danking Just realized you already took care of that in https://github.com/hail-is/hail/pull/9848/files#diff-21972f0c9ab321a85ffb4de98353c86fdea4992a5eb0d48c3489b977513b1da0! Thanks!

PR #9909, opened 2021-01-25
[query] Add densify aggregator

tpoterba:
The changes in this pull request reduce the runtime of
read / densify / force_count_rows by approximately 30%.

Before: 17m42s
After: 12m32s


chrisvittal (changes requested):
lint issues

chrisvittal (approved):

PR #9910, opened 2021-01-25
[logging] Add severity field to jsonlogger

daniel-goldstein:
GCP Logging attempts to infer the severity of a log entry and defaults to labelling everything written to `stdout` as `INFO` and everything written to `stderr` as `ERROR`. There are some [LogEntry fields](https://cloud.google.com/logging/docs/structured-logging) that can be overwritten by fields in our JSON output, including `severity`. Until now we had been logging the severity level as `levelname`, which is the expectation of `jsonlogger`, but this means GCP's levels and ours do not necessarily match.  This adds another `severity` field to the logs so GCP can pick up the level.

This should get rid of a swath of non-error log messages written to `stderr` (only JSON though) and lets other levels like `WARNING` get marked as such.

GCP does quite literally extract the severity field though, so it does not appear in the `jsonPayload`. I've kept the `levelname` in then as a sanity check but am also happy to try to remove it.

### DIFF
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -17,7 +17,7 @@ log = logging.getLogger('ci')
 
 pretty_print_log = "jq -Rr '. as $raw | try \
 (fromjson | if .hail_log == 1 then \
-    ([.levelname, .asctime, .filename, .funcNameAndLine, .message, .exc_info] | @tsv) \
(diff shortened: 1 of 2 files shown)


danking (approved):
I'm inclined to remove the level name once we're all satisfied things are working correctly

PR #9911, opened 2021-01-25
[router-resolver] add access logger to router resolver

daniel-goldstein:
Use our `AccessLogger` in `router-resolver` too.


danking (approved):
perfecto

PR #9912, opened 2021-01-25
[query] Add println to codebuilder

johnc1231:
This just adds a utility debugging method, instead of having to do `cb.append(Code._println(...))`. I find it helpful.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/EmitCodeBuilder.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/EmitCodeBuilder.scala
@@ -151,5 +151,5 @@ class EmitCodeBuilder(val emb: EmitMethodBuilder[_], var code: Code[Unit]) exten
   }
 
   // for debugging
-  def println(cString: Code[String]) = this += Code._println(cString)
+  def println(cString: Code[String]*) = this += Code._printlns(cString:_*)
 }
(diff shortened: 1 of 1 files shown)


hail/src/main/scala/is/hail/expr/ir/EmitCodeBuilder.scala
```diff
@@ -149,4 +149,7 @@ class EmitCodeBuilder(val emb: EmitMethodBuilder[_], var code: Code[Unit]) exten
   def strValue(r: Value[Region], t: PType, code: Code[_]): Code[String] = {
     StringFunctions.boxArg(EmitRegion(emb, r), t)(code).invoke[String]("toString")
   }
+
+  // for debugging
+  def println(cString: Code[String]) = this += Code._println(cString)
```
tpoterba, line 154: maybe have this take varargs of Code[String] and do `Code._printlns(strings:_*)`, I use that much more than the single arg (you get concat for free)

tpoterba (changes requested):

hail/src/main/scala/is/hail/expr/ir/EmitCodeBuilder.scala
johnc1231: Ah, didn't even know about that one, good idea.

tpoterba (approved):

PR #9913, opened 2021-01-25
[hailctl] add hailctl cp command

cseed:
This is stacked on https://github.com/hail-is/hail/pull/9842 which rewrites hailctl argument handling.  Probably not worth reviewing until that is in.


jigold:
Should we close this PR?

danking:
Sure. We have other priorities right now

PR #9914, opened 2021-01-25
[query-service] implement load_references_from_dataset

danking:
This fixes a bug faced by Konrad preventing him from reading any tables. There were two issues here:

1. `load_references_from_dataset` was unimplemented
2. `GoogleStorageFS.listStatus` included the directory itself in the listing which caused
   unexpected, wrong behavior in `ReferenceGenome.readReferences`. I fixed this by filtering out the
   directory itself from `listStatus`.

### DIFF
--- a/hail/python/hail/backend/service_backend.py
+++ b/hail/python/hail/backend/service_backend.py
@@ -140,7 +140,10 @@ class ServiceBackend(Backend):
         return self.socket.request('references/get', name=name)
 
     def load_references_from_dataset(self, path):
-        raise NotImplementedError("ServiceBackend does not support 'load_references_from_dataset'")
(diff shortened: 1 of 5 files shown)


catoverdrive (approved):

catoverdrive (dismissed):
Just realized that you should be able to read this from the query server directly using the filesystem in the user context. The tableType and matrixTableType functions should be doing this as well---e.g. parse_matrix_ir needs to be able to read metadata files in order to parse the IR. I'd prefer to handle this similarly to those endpoints as much as possible. The relevant filesystem should be in the scala User class.

catoverdrive (approved):

danking:
OK had my filter backwards. I always seem to read `filter(X)` as "remove things that satisfy X"

PR #9917, opened 2021-01-25
[hail] make TableNativeZippedReader work

danking:
`filterSet` requires the argument be a subset of the fields of the given struct. That is not
how `splitRequestedTypes` works. Instead, we explicitly compute the fields needed from each side
and pass those to `filterSet`.


tpoterba (approved):
looks good, I'll figure out what test should be catching this tomorrow.

danking:
There was a test marked as should fail: `test_matrix_table.py::test_read_write_all_types`, but it now passes! woo!

tpoterba:
oh, nice!

PR #9918, opened 2021-01-25
[fs] add AsyncFS.multi_part_create interface

cseed:
implement in LocalAsyncFS with test.

AsynFS needs to support multi-part writes in order to parallelize transfers of large files.  Next PR will implement this for gs:// efficiently using a temporary file and the compose operation.

### DIFF
--- a/hail/python/test/hailtop/test_aiogoogle.py
+++ b/hail/python/test/hailtop/test_aiogoogle.py
@@ -2,14 +2,16 @@ import os
 import secrets
 import shutil
 from concurrent.futures import ThreadPoolExecutor
+import asyncio
 import pytest
+from hailtop.utils import secret_alnum_string
 from hailtop.aiotools import LocalAsyncFS, RouterAsyncFS
 from hailtop.aiogoogle import StorageClient, GoogleStorageAsyncFS
 
 
 @pytest.fixture(params=['file', 'gs', 'router/file', 'router/gs'])
 async def filesystem(request):
-    token = secrets.token_hex(16)
+    token = secret_alnum_string()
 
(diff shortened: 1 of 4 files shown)


hail/python/test/hailtop/test_aiogoogle.py
```diff
@@ -248,3 +263,24 @@ async def listfiles(dir, recursive):
...
+async def test_multi_part_create(local_filesystem):
+    fs, base = local_filesystem
+
+    num_parts = 3
+    parts_data = [secrets.token_bytes(8192) for _ in range(num_parts)]
+
+    path = f'{base}a'
+    async with await fs.multi_part_create(path, num_parts) as c:
+        async def create_part(i):
+            async with await c.create_part(i, i * 8192) as f:
```
jigold, line 277: It seems like the num and start are directly dependent. I thought about just taking num (`i`) here instead of the start. Is there a reason you did it this way?

hail/python/test/hailtop/test_aiogoogle.py
```diff
@@ -248,3 +263,24 @@ async def listfiles(dir, recursive):
...
+                await f.write(parts_data[i])
+
+        await asyncio.gather(*[
```
jigold, line 280: Can you randomize the order you write the parts or do it in reverse order just to make sure it's working?

jigold (dismissed):

hail/python/test/hailtop/test_aiogoogle.py
cseed: Done!

jigold (approved):

PR #9920, opened 2021-01-26
[query] Add assertion that emitted and inferred ptype match

tpoterba:


tpoterba:
oh dear

tpoterba:
OK, I have to think what I want to do here. The problem is here:

https://github.com/hail-is/hail/blob/19dac15f9acf50f68274c0c3bd658cde7d16e917/hail/src/main/scala/is/hail/types/physical/stypes/concrete/SBaseStructPointer.scala#L58-L62

(array has the same problem)

tpoterba:
The requiredness of the loaded code is the requiredness of the field type. However, if the struct itself is optional, the EmitCode needs to be optional too, regardless of the requiredness of the field.

tpoterba:
I think what I want is for STypes to not track requiredness at the top level at all. We can't quite do this yet, though.

johnc1231:
Will approve when benchmarks finish

johnc1231 (approved):

PR #9922, opened 2021-01-26
[fs] add some parallelism to copy

cseed:
This includes two changes:

First, add parallelism (async, single-core) to copy by parallelizing
over transfers, sources and files in directory sources.  This doesn't
get multi-core or intra-file parallelism yet.

copy() returns a CopyReport now.  It summarizes each transfer, source
and records counts of files copied, exceptions, etc.  To preserve the
old behavior, there is a flag that causes the first exception in the
report to be raised (if there is one) (this is the default).

I will remark, this makes copy() very parallel and it will immediately
and do in parallel all the transfers, files and discovered files.

Separate PRs will add process parallelism and intra-file parallelism which on multi-part upload support: https://github.com/hail-is/hail/pull/9918

### DIFF
--- a/hail/python/hailtop/aiotools/fs.py
+++ b/hail/python/hailtop/aiotools/fs.py
@@ -8,7 +8,7 @@ import shutil
 import asyncio
 from concurrent.futures import ThreadPoolExecutor
 import urllib.parse
-from hailtop.utils import blocking_to_async, url_basename, url_join
+from hailtop.utils import blocking_to_async, url_basename, url_join, AsyncWorkerPool, WaitableSharedPool
 from .stream import ReadableStream, WritableStream, blocking_readable_stream_to_async, blocking_writable_stream_to_async
 
 AsyncFSType = TypeVar('AsyncFSType', bound='AsyncFS')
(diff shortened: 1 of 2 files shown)


hail/python/hailtop/aiotools/fs.py
```diff
@@ -458,45 +553,53 @@ async def _dest_type(self, transfer):
...
+            try:
+                src = transfer.src
+                if isinstance(src, str):
+                    await self.copy_source(worker_pool, transfer, transfer_report._source_report, src, dest_type_task)
+                else:
+                    if transfer.treat_dest_as == Transfer.TARGET_FILE:
+                        raise NotADirectoryError(transfer.dest)
+
+                    async with WaitableSharedPool(worker_pool) as pool:
+                        for r, s in zip(transfer_report._source_report, src):
```
jigold, line 574: Did you intend to continue trying to copy all files even if one fails? Opposite behavior of gsutil?

hail/python/hailtop/aiotools/fs.py
```diff
@@ -281,6 +282,80 @@ def __init__(self, src: Union[str, List[str]], dest: str, *, treat_dest_as: str
...
+    def __init__(self, transfer: Transfer):
+        self._transfer = transfer
+        if isinstance(transfer.src, str):
+            self._source_report = SourceReport(transfer.src)
+        else:
+            self._source_report = [SourceReport(s) for s in transfer.src]
+        self._exception: Optional[Exception] = None
+
+    def set_exception(self, exception: Exception):
+        assert not self._exception
```
jigold, line 324: I assume in the future these will store the number of bytes transferred and the time it took?

hail/python/hailtop/aiotools/fs.py
```diff
@@ -583,6 +686,14 @@ async def close(self) -> None:
...
             await fs.close()
 
-    async def copy(self, transfer: Union[Transfer, List[Transfer]]):
-        copier = Copier(self)
-        await copier.copy(transfer)
+    async def copy(self, transfer: Union[Transfer, List[Transfer]], raise_first_exception=True):
+        worker_pool = AsyncWorkerPool(50)
+        try:
+            copier = Copier(self)
+            copy_report = CopyReport(transfer)
```
jigold, line 693: Did you test the copy report by hand?

jigold (changes requested):

hail/python/hailtop/aiotools/fs.py
cseed: Yes.  This is designed for maximum parallelism.  So by the time one source fails, all the other sources in the transfer have potentially (and likely) already started.  I see two alternatives:
 - Copy things serially.  Obviously we don't like that, we're designing for performance.
 - Cancel the partial copies when one fails.  I don't like that much either, and would rather one fail and one succeed than one fail and one get stopped part way through.

hail/python/hailtop/aiotools/fs.py
cseed: Yep!

hail/python/hailtop/aiotools/fs.py
cseed: Not yet.  It is currently untested, except that the first exception is the one we expected.

jigold (approved):

PR #9925, opened 2021-01-26
[query] Change EmitCode implementation

patrick-schultz:
This replaces the EmitCode implementation by a start label together with an IEmitCode. I propose that EmitCode will still be needed occasionally in a CodeBuilder world, and that this is a form it could take.

For compatibility with existing EmitCode users, this defines `setup` and `m`, but those should be removed after the transition to CodeBuilder/IEmitCode is complete.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -345,19 +345,16 @@ case class IEmitCodeGen[+A](Lmissing: CodeLabel, Lpresent: CodeLabel, value: A) 
 object EmitCode {
   def apply(setup: Code[Unit], m: Code[Boolean], pv: PCode): EmitCode = {
     val start = CodeLabel()
-    val body = new CodeBoolean(Code(start, setup, m)).toCCode
+    val body = new CodeBoolean(Code(start, m)).toCCode
     val iec = IEmitCode(new CodeLabel(body.Ltrue), new CodeLabel(body.Lfalse), pv)
-    new EmitCode(start, iec)
+    new EmitCode(setup, start, iec)
   }
 
   def unapply(ec: EmitCode): Option[(Code[Unit], Code[Boolean], PCode)] =
     Some((ec.setup, ec.m, ec.pv))
 
-  def apply(setup: Code[Unit], ec: EmitCode): EmitCode = {
-    val start = CodeLabel()
-    Code(start, setup, ec.start.goto)
-    new EmitCode(start, ec.iec)
-  }
+  def apply(setup: Code[Unit], ec: EmitCode): EmitCode =
+    new EmitCode(Code(setup, ec.setup), ec.start, ec.iec)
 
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -378,27 +384,31 @@ object EmitCode {
...
-      new EmitCode(Code._empty, coerce[Boolean](ct.last), pt.fromCodeTuple(ct.init))
+      EmitCode(Code._empty, coerce[Boolean](ct.last), pt.fromCodeTuple(ct.init))
   }
 }
 
-case class EmitCode(setup: Code[Unit], m: Code[Boolean], pv: PCode) {
+class EmitCode(private val start: CodeLabel, private val iec: IEmitCode) {
+  def pv: PCode = iec.value
+
+  def setup: Code[Unit] = Code._empty
```
tpoterba, line 396: I suspect this may break tests. Do all users of EmitCode always check missingness?

tpoterba (changes requested):
This is a great change! Waiting for tests to pass before approval.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
patrick-schultz: Ah, you're probably right. Hmm...

patrick-schultz:
Tests are passing, but I'd like to see if I can go back to not having setup on EmitCode, now that we found the btree problem.

patrick-schultz:
Nice, it works now! I flipped through uses of `EmitCode.setup`, looking for places that might cause problems when `m` was more than a load of a boolean. I only found a couple, and fixed them. Also note that `EmitCode.fromI` also makes an empty `setup`, with all the work done in `m`, so we were already assuming `m` can do work.

tpoterba:
excellent! can you just back out the change to AppendOnlyBTree?

hail/src/main/scala/is/hail/expr/ir/agg/AppendOnlyBTree.scala
```diff
@@ -234,9 +234,12 @@ class AppendOnlyBTree(kb: EmitClassBuilder[_], val key: BTreeKey, region: Value[
 
   private def getF(cb: EmitCodeBuilder, root: Code[Long], kc: EmitCode): Code[Long] = {
     val get = kb.genEmitMethod("btree_get", FastIndexedSeq[ParamType](typeInfo[Long], kc.pv.st.pType.asEmitParam), typeInfo[Long])
-    val node = get.getCodeParam[Long](1)
     val k = get.getEmitParam(2)
     get.emitWithBuilder { cb =>
+      // FIXME load bearing conversion of arg to local ???
+//      val node = get.getCodeParam[Long](1)
+      val node = cb.newLocal("node", get.getCodeParam[Long](1))
```
tpoterba, line 241: remove

tpoterba (changes requested):

tpoterba (approved):

PR #9926, opened 2021-01-26
[query] CodeBuilder-ify part of EmitStream

patrick-schultz:
Refactor `_emitStream` to a `EmitCodeBuilder -> IEmitCode` signature.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
@@ -1927,6 +1927,7 @@ object EmitStream {
                     COption.fromIEmitCode(mb) { cb =>
                       emitStream(
                         innerIR,
+                        cb = cb,
                         outerRegion = innerStreamOuterRegion,
                         env = env.bind(name -> new EmitUnrealizableValue(outerEltType, elt)))
                     }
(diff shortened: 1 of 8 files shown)


tpoterba:
oh, this isn't stacked on the EmitCode one? Sorry!

tpoterba:
will look now

tpoterba:
It's hard to go through every line to look for bugs, but the structure looks great. You've got a couple test failures (code cast errors)

hail/src/main/scala/is/hail/expr/ir/EmitStream.scala
```diff
@@ -2117,105 +2165,95 @@ object EmitStream {
...
+                Code._null,
+                mb.ecb.getPType(keyPType)))
+
+              val startt = keyRange.loadStart(cb)
+                  .get(cb, "shuffle expects defined endpoints")
+                  .asLong
+                  .tcode[Long]
+              val endt = keyRange.loadEnd(cb)
+                  .get(cb, "shuffle expects defined endpoints")
+                  .asLong
```
tpoterba, line 2203: these casts will fail. keyRange isn't a long, it's a struct, I think.

ShuffleClient really has the wrong interfaces and we should fix those, but for now remove the `asLong` and it should work as it did before.

patrick-schultz: Ah, right. I added that to get it to compile, because the following `.tcode[Long]` isn't defined on `SCode`. Not clear to me the right way to fix it.

tpoterba: ah! You can cast to pcode with `asPCode`

tpoterba: (pcode is an scode represented by a single code, and so has tcode)

patrick-schultz: Perfect, thanks!

patrick-schultz:
Tests are passing. Should be ready for review.

tpoterba (approved):
Awesome, nice work!

PR #9928, opened 2021-01-26
[router,gateway,site,blog,internal-gateway] Change nginx logs to JSON

daniel-goldstein:
- Basic JSON format for NGINX access logs
- `message`, `remote_address`, `request_duration`, `response_status` and `x_real_ip` should match the Python Access logger so we should be able to query these fields all at once
- Unfortunately the NGINX `error_log` does not allow the same custom formatter, but we can create multiple, [custom access logs](https://www.nginx.com/blog/diagnostic-logging-nginx-javascript-module/#proxy_next_upstream) based on arbitrary failure criteria if there's a particular class of error logs we want to get in the future.

### DIFF
--- a/site/nginx.conf
+++ b/site/nginx.conf
@@ -27,7 +27,7 @@ http {
    '"request_duration":$request_time,'
    '"remote_address":"$remote_addr",'
    '"x_real_ip":"$http_x_real_ip",'
-   '"request_start_time":"[$time_local]",'
+   '"request_start_time":"$time_local",'
    '"body_bytes_sent":"$body_bytes_sent",'
    '"http_referer":"$http_referer",'
    '"http_user_agent":"$http_user_agent"'
--- a/site/nginx.conf
+++ b/site/nginx.conf
@@ -21,10 +21,19 @@ http {
   ssl_protocols TLSv1 TLSv1.1 TLSv1.2; # Dropping SSLv3, ref: POODLE
   ssl_prefer_server_ciphers on;
 
-  log_format combined_real_ip '$http_x_real_ip - $remote_addr - $remote_user [$time_local] '
-                              '$scheme "$request" $status $body_bytes_sent '
-                              '"$http_referer" "$http_user_agent"';
-  access_log /var/log/nginx/access.log combined_real_ip;
+  log_format json-log escape=json '{'
+   '"message":"$scheme $request done in ${request_time}s: $status",'
+   '"response_status":"$status",'
+   '"request_duration":"$request_time",'
+   '"remote_address":"$remote_addr",'
+   '"x_real_ip":"$http_x_real_ip",'
+   '"request_start_time":"[$time_local]",'
+   '"body_bytes_sent":"$body_bytes_sent",'
+   '"http_referer":"$http_referer",'
+   '"http_user_agent":"$http_user_agent"'
+ '}';
+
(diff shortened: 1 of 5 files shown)


site/nginx.conf
```diff
@@ -21,10 +21,19 @@ http {
...
-                              '$scheme "$request" $status $body_bytes_sent '
-                              '"$http_referer" "$http_user_agent"';
-  access_log /var/log/nginx/access.log combined_real_ip;
+  log_format json-log escape=json '{'
+   '"message":"$scheme $request done in ${request_time}s: $status",'
+   '"response_status":$status,'
+   '"request_duration":$request_time,'
+   '"remote_address":"$remote_addr",'
+   '"x_real_ip":"$http_x_real_ip",'
+   '"request_start_time":"[$time_local]",'
```
danking, line 30: I don't think we want to bracket the time in the JSON format.

site/nginx.conf
```diff
@@ -21,10 +21,19 @@ http {
   ssl_protocols TLSv1 TLSv1.1 TLSv1.2; # Dropping SSLv3, ref: POODLE
   ssl_prefer_server_ciphers on;
 
-  log_format combined_real_ip '$http_x_real_ip - $remote_addr - $remote_user [$time_local] '
-                              '$scheme "$request" $status $body_bytes_sent '
-                              '"$http_referer" "$http_user_agent"';
-  access_log /var/log/nginx/access.log combined_real_ip;
+  log_format json-log escape=json '{'
+   '"message":"$scheme $request done in ${request_time}s: $status",'
```
danking, line 25: let's move the status into a field. I want queries for status to be fast, I assume google indexes each field

danking: Honestly I'm not sure we need a message? Maybe just stick the `request` alone in there or put that in a field?

danking (changes requested):

site/nginx.conf
daniel-goldstein: Google extracts the `message` field as the info for the unexpanded log in the UI, so this is what you'd see while scrolling through the UI without expanding log entry. The status and request_time already have their own fields. `$request` is of the form "GET /path/to/thing HTTP/1.0" so I'm fine making that everything in `message`, perhaps also with the status

site/nginx.conf
daniel-goldstein: Any reason it is [here](https://github.com/hail-is/hail/blob/1215c5152e60825c5f7e1ab700cb6ba31788ecc2/hail/python/hailtop/hail_logging.py#L33)

site/nginx.conf
daniel-goldstein: For context, this matches the `AccessLogger` [message](https://github.com/hail-is/hail/blob/1215c5152e60825c5f7e1ab700cb6ba31788ecc2/hail/python/hailtop/hail_logging.py#L34)

site/nginx.conf
danking: I think I copied that from the internet.

danking (approved):

danking:
There's some intermittent issue with test_benchmark. I retried your PR so hopefully it passes now. At some point we should dig into test_benchmark and understand why it's flaking

site/nginx.conf
daniel-goldstein: Ok I'll update the jsonlogger to match

PR #9929, opened 2021-01-26
[query] Add new constructors for PBaseStruct/PCanonicalArray

tpoterba:
Use these to convert a few locus functions from Code => PCode

### DIFF
--- a/hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
+++ b/hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
@@ -161,7 +161,7 @@ final case class PCanonicalArray(elementType: PType, required: Boolean = false) 
       firstElementOffset(aoff, loadLength(aoff)) + i.toL * const(elementByteSize)
     }
 
-  private def elementOffsetFromFirst(addr: Code[Long], i: Code[Int]): Code[Long] = addr + i.toL * const(elementByteSize)
+  private def elementOffsetFromFirst(firstElementAddr: Code[Long], i: Code[Int]): Code[Long] = firstElementAddr + i.toL * const(elementByteSize)
 
   def nextElementAddress(currentOffset: Long) =
     currentOffset + elementByteSize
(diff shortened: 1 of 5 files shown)


hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
```diff
@@ -459,4 +461,48 @@ final case class PCanonicalArray(elementType: PType, required: Boolean = false)
...
+    new SIndexablePointerCode(SIndexablePointer(this), addr)
+  }
+
+  // unsafe StagedArrayBuilder-like interface that gives caller control over adding elements and finishing
+  def constructFromFunctions(cb: EmitCodeBuilder, region: Value[Region], length: Value[Int], deepCopy: Boolean):
+  (((EmitCodeBuilder, Value[Int], IEmitCode) => Unit, (EmitCodeBuilder => SIndexablePointerCode))) = {
+
+    val addr = cb.newLocal[Long]("pcarray_construct1_addr", allocate(region, length))
+    cb += stagedInitialize(addr, length, setMissing = false)
+    val i = cb.newLocal[Int]("pcarray_construct1_i", 0)
```
johnc1231, line 493: Should the name here be `"pcarray_construct2_i"`? I assumed the 1 was to distinguish between these two constructors.

hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
```diff
@@ -161,6 +161,8 @@ final case class PCanonicalArray(elementType: PType, required: Boolean = false)
       firstElementOffset(aoff, loadLength(aoff)) + i.toL * const(elementByteSize)
     }
 
+  private def elementOffsetFromFirst(addr: Code[Long], i: Code[Int]): Code[Long] = addr + i.toL * const(elementByteSize)
```
johnc1231, line 164: Maybe just call the argument `firstElementAddr` or something? This is the only function that takes an address to the first element that I'm aware of. Also, the name sort of sounds like it might be the number of bytes between the first element and element at position `i`, but that's maybe just part of using the word offset everywhere at this point.

johnc1231 (approved):

hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
tpoterba: oh, yes.

hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
tpoterba: good point.

johnc1231:
I marked it WIP, you can fix those names and then just remove that WIP

PR #9930, opened 2021-01-26
[image-fetcher] only redirect errors to stderr and remove old Make rules

daniel-goldstein:
- Using `-x` prints every line of the script to an error log entry. The main thing we're concerned about here is if it can fetch from notebook, so capture that output in `image-fetch-output.log` and cat that out to `stderr` only if it failed so it shows up as errors in the logs.

- Removed old templating from before the `jinja2` times

cc @cseed


danking (approved):
v nice

PR #9931, opened 2021-01-26
[format] black formatting

daniel-goldstein:
@danking Here's what it looks like running black on `ci.py` and `address.py` with only customizations being a line length of 120 (I personally prefer shorter but that's what we have currently) and no single quote -> double quote changes (`--skip-string-normalization`).

It is _super_ opinionated, so you have to like the defaults or be willing to roll with it on some calls. It took a lot of [discourse](https://github.com/psf/black/issues/118) just to allow single-quoting strings, but a commit hook to auto-format service code seems useful.

### DIFF
--- a/hail/python/dev-requirements.txt
+++ b/hail/python/dev-requirements.txt
@@ -1,6 +1,8 @@
 flake8==3.7.8
 mypy==0.780
 pylint==2.6.0
+pre-commit==2.9.2
+black==20.8b1
 pytest==4.6.3
 pytest-html==1.20.0
 pytest-xdist==1.28
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,15 @@
+---
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: https://github.com/psf/black
+    rev: 20.8b1
+    hooks:
+      - id: black
+        language_version: python3
+        files: 'ci/.*'
+        args: ['--line-length=120', '--skip-string-normalization']
--- a/ci/Makefile
+++ b/ci/Makefile
@@ -13,6 +13,7 @@ PYTHON := PYTHONPATH=$${PYTHONPATH:+$${PYTHONPATH}:}$(EXTRA_PYTHONPATH) python3
 check:
 	$(PYTHON) -m flake8 ci
 	$(PYTHON) -m pylint --rcfile ../pylintrc ci --score=n
+	$(PYTHON) -m black ci --check --line-length=120 --skip-string-normalization
 	../check-sql.sh
 
 .PHONY: build-ci-utils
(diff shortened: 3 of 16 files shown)


danking:
I have no qualms with this.

I do hate double quotes, but I guess he's right that black could just reformat my single quotes into double quotes.

danking:
Looks like I can just enable this with elpy in Emacs. What else would we do? Set a git hook?

danking:
We can probably turn flake8 off. I don't know what good it provides over pylint and mypy.

daniel-goldstein:
Ya I'll see if `pylint` and `mypy` give good enough coverage, and whether they fight with `black` over anything. Something useful from that thread was they point out how a big formatting commit ruins git blame unless you account for it.

I think just a pre-commit hook is good, anyone who wants to set it up in their editor can do so fairly easily.

vladsaveliev:
Hi,

I'm just curious if you tried black on a code that uses Hail query. As far as I see, PEP8 conflicts with the code style adopted in the Hail docs, e.g. black would remove spaces in named function arguments:

```
- mt = mt.annotate_entries(GT = lgt_to_gt(mt.LGT, mt.LA))
+ mt = mt.annotate_entries(GT=lgt_to_gt(mt.LGT, mt.LA))
```

On the other hand, query can be seen as a DSL on top of Python, so the same guidelines probably don't need to be applied to it. Wondering if you had thoughts about lining the query code? We will be writing a lot of that in the nearest future in the Centre for Population Genomics, and would love to set up some style checks, or even automate that with a tool like black or prettier.

And on black - are you considering automating code refactoring with black as part of the CI? Or you wanted to just do checks, alongside with pylint?

Vlad

tpoterba:
> On the other hand, query can be seen as a DSL on top of Python, so the same guidelines probably don't need to be applied to it

You've hit exactly on the center of a longstanding discussion on the team with no definitive resolution 😄

daniel-goldstein:
Great question! I have not thought to run black on query and agree with your point about it maybe not being a good match. The intention here was just for formatting internal python services. Even so I don't love quite how restrictive black is, but it does seem like the most popular formatter out there now.

danking:
FWIW, when I write hail code, I usually write it in PEP8 style because my editor complains. I've grown accustomed to it. I am in favor of blackening everything for consistency.

Re: CI, I think we'll ban (non-hail-query) code that doesn't conform to Black and developers are then incentivized to add a git-hook or editor modification that runs black.

daniel-goldstein:
I'm going to change this branch to instead add `pre-commit` with `black` as one of its commit hooks. Developers can choose to `pre-commit install` which will set up useful hooks and run `black` over any modified files for directories that opt into it (which I'll start as just CI).

hail/python/dev-requirements.txt
```diff
@@ -1,6 +1,7 @@
 flake8==3.7.8
 mypy==0.780
 pylint==2.6.0
+pre-commit==2.9.2
```
danking, line 4: should `black` be included here?

danking (changes requested):

danking (approved):
Disregard, I understand now.

danking (changes requested):
Can we also enforce this with a build.yaml step that verifies that `ci/` is blackened?

daniel-goldstein:
There's a `--check` flag, I added it to the existing ci `check`. Had to add `black` to the dev dependencies for that.

danking (approved):

danking:
Remove draft whenever you're ready. Looks good to me.

PR #9934, opened 2021-01-27
[copy] parallelize over chunks

cseed:
This PR introduces async parallelism for copy.  Summary of changes:
 - add compose to aiogoogle StorageClient and use it to implement (hierarchical) multi-part write for the google AsyncFS
 - add AsyncFS.open_from, to open a file starting at a given offset,
 - PART_SIZE is the size of parts to transfer and the combine with a multi-part write when copying large files.  PART_SIZE is currently 128MB.
 - BUFFER_SIZE is the size we request when reading in the copy.  Change this to 256KB (from 8KB).
 - Copy files in parts and use multi-part writes when copying large files.
 - Parallelize rmtree (it now takes a semaphore for controlling parallelism)
 - Add new primitives for parallelization, bounded_gather2 and OnlineBoundedGather2.

bounded_gather2 was an interesting design problem.  I would describe it like this:

Say you have serial asyncio code with I/O (which will block the serial algorithm when the I/O is pending) which you'd like to parallelize.  You can't do this generically, so in addition I imagine there are parallelization fork/join points where you want to run a bunch of subcomputations in parallel, wait for them all to complete, and resume where you left off.

The code begins by creating a asyncio.Semaphore that globally bounds the parallelism of the code, and acquiring the semaphore before executing the code (since the entrypoint into the code accounts for one thread of control).  For example:

```
sema = asyncio.Semaphore(50)
async with sema:
    await copy(sema, ...)
```

Then, to run a set of operations in parallel, subject to the global parallelism bound, use bounded_gather2:

```
  await bounded_gather2(sema, *aws)
```

The naive implementation of bounded_gather2 doesn't work: bounded_gather2 cannot spawn a task for each awaitable and have it try to acquiare the semaphore, because this can lead to deadlock: if 50 tasks launch and wait for children, but none of those children can run because the parents have all the threads of control, the algorithm will deadlock.

The key is to make sure at least one child can always be running in a bounded_gather2.  But the caller of bounded_gather2 had a reserved thread of execution and it is blocking, so that thread of execution should be transferred to the children while the bounded_gather2 is blocked.  This is what bounded_gather2 does.

There is also an "online" version of bounded_gather2 which lets you schedule an unbounded number of children (potentially generated asynchronously).  OnlineBoundedGather2 is used in parallelizing file transfers generated by directory listings, for example, which are enumerated via an async generator, and are potentially very large.

I will try to replace bounded_gather and the async worker pool with this mechanism in a future PR.

The parameters will likely need additional tuning.  I have done some rough timing, and already this is beating gsutil:

- Transfer 10GB spread over 40K files (times in ms):

  {'upload': 95803,
   'download': 55240,
   'compare': 54117,
   'clean file': 632,
   'clean gs': 117263,
   'total': 323061}

  vs the gsutil transfer:

  real	11m14.153s
  user	14m28.789s
  sys	1m29.090s

  and gsutil cleanup (removing 40K files in gs://):

  real	10m12.236s
  user	3m33.382s
  sys	0m55.450s

 - Transfer 10GB in one file takes ~20s (up) and ~30s (down) for copy vs ~1m for gsutil.

I'm still working on the benchmark harness and will post more complete comparisons in a future PR.

### DIFF
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -9,7 +9,7 @@ import asyncio
 import urllib.parse
 import aiohttp
 from hailtop.utils import (
-    secret_alnum_string, bounded_gather2,
+    secret_alnum_string, bounded_gather2, OnlineBoundedGather2,
     TransientError, retry_transient_errors,
     AsyncWorkerPool, WaitableSharedPool, secret_alnum_string)
 from hailtop.aiotools import (
@@ -525,7 +525,7 @@ class GoogleStorageMultiPartCreate(MultiPartCreate):
                 [self._part_name(i) for i in range(self._num_parts)],
                 self._dest_name)
         finally:
-            await self._fs.rmtree(f'gs://{self._bucket}/{self._dest_dirname}_')
+            await self._fs.rmtree(self._sema, f'gs://{self._bucket}/{self._dest_dirname}_')
 
 
 class GoogleStorageAsyncFS(AsyncFS):
@@ -694,19 +694,20 @@ class GoogleStorageAsyncFS(AsyncFS):
         async for page in await self._storage_client.list_objects(bucket, params=params):
             prefixes = page.get('prefixes')
             items = page.get('items')
-            return prefixes or items
+            return bool(prefixes or items)
         assert False  # unreachable
 
     async def remove(self, url: str) -> None:
         bucket, name = self._get_bucket_name(url)
         await self._storage_client.delete_object(bucket, name)
 
-    async def rmtree(self, url: str) -> None:
-        try:
-            async for entry in await self.listfiles(url, recursive=True):
-                await self.remove(await entry.url())
-        except FileNotFoundError:
-            pass
+    async def rmtree(self, sema: asyncio.Semaphore, url: str) -> None:
+        async with OnlineBoundedGather2(sema) as pool:
+            try:
+                async for entry in await self.listfiles(url, recursive=True):
+                    await pool.call(self.remove, await entry.url())
+            except FileNotFoundError:
+                pass
 
     async def close(self) -> None:
         await self._storage_client.close()
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -528,7 +528,7 @@ class GoogleStorageMultiPartCreate(MultiPartCreate):
 
                     for n in chunk_names:
                         cleanup_tasks.append(
-                            await pool.call(self._fs._remove_doesnt_exist_ok(n)))
+                            await pool.call(self._fs._remove_doesnt_exist_ok, f'gs://{self._bucket}/{n}'))
 
                 await tree_compose(
                     [self._part_name(i) for i in range(self._num_parts)],
@@ -722,6 +722,9 @@ class GoogleStorageAsyncFS(AsyncFS):
             await self._storage_client.delete_object(bucket, name)
         except FileNotFoundError:
             pass
+        except aiohttp.ClientResponseError as e:
+            if e.status != 404:
+                raise
 
     async def rmtree(self, sema: asyncio.Semaphore, url: str) -> None:
         async with OnlineBoundedGather2(sema) as pool:
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -712,15 +712,19 @@ class GoogleStorageAsyncFS(AsyncFS):
         await self._storage_client.delete_object(bucket, name)
 
     async def _remove_doesnt_exist_ok(self, url: str) -> None:
-        bucket, name = self._get_bucket_name(url)
         try:
+            bucket, name = self._get_bucket_name(url)
             await self._storage_client.delete_object(bucket, name)
         except FileNotFoundError:
             pass
 
     async def rmtree(self, sema: asyncio.Semaphore, url: str) -> None:
         async with OnlineBoundedGather2(sema) as pool:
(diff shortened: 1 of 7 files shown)


jigold:
I did a first pass through this PR. I tried to figure out the SubSemaphore and OnlineBoundedGather2 code, but I think it would be better to have you explain it to me when you have time. I think my lack of familiarity with tasks is making this hard for me as well.

cseed:
Yeah, of course! I have 30m at 3:30 or 4:30 or any time before 3:30pm tomorrow.

jigold:
Either is fine with me -- just ping me when you are ready.

cseed:
OK, I think this is ready for review again.  I rebased, was able to simplify the online bounded gather a bit, and parallelized the GCS rmtree more, so that cleanup can happen in parallel with the merge operations as much as possible.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -737,22 +814,15 @@ async def remove(self, url: str) -> None:
...
-        try:
-            copier = Copier(self)
-            copy_report = CopyReport(transfer)
-            await copier.copy(worker_pool, copy_report, transfer)
-            if raise_first_exception:
-                copy_report.raise_first_exception()
-            return copy_report
-        finally:
-            worker_pool.shutdown()
+    async def copy(self, sema: asyncio.Semaphore, transfer: Union[Transfer, List[Transfer]], return_exceptions=False):
```
jigold, line 825: Should we make the semaphore optional here and make the default 50? Same question about rmtree above. Then you wouldn't need to change all the test interfaces.

hail/python/test/hailtop/test_aiogoogle.py
```diff
@@ -283,24 +319,25 @@ async def test_multi_part_create(local_filesystem, permutation):
...
-    expected = b''.join(part_data)
-    async with await fs.open(path) as f:
-        actual = await f.read()
-    assert expected == actual
+    sema = asyncio.Semaphore(50)
+    async with sema:
+        async with await fs.multi_part_create(sema, path, len(part_data)) as c:
+            async def create_part(i):
+                async with await c.create_part(i, part_start[i]) as f:
+                    await f.write(part_data[i])
```
jigold, line 329: It's not quite clear to me that these tests cover the case where we have > 32 partitions to make sure the GoogleStorageMultiPartCreate is working correctly with the tree compose and the removal of files.

hail/python/hailtop/utils/utils.py
```diff
@@ -258,6 +259,220 @@ async def __aexit__(self,
...
+    try:
+        return await asyncio.gather(*tasks)
+    finally:
+        _, exc, _ = sys.exc_info()
+        if exc is not None:
+            tasks = []
+            for task in tasks:
+                if not task.done():
+                    task.cancel()
+                    tasks.append(task)
```
jigold, line 465: I don't think this is correct. I don't think you can append to a list you are currently iterating over. I tried it on a small test example and it caused an infinite loop.

jigold (changes requested):

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -389,6 +392,19 @@ async def delete_object(self, bucket: str, name: str, **kwargs) -> None:
...
         return PageIterator(self, f'/b/{bucket}/o', kwargs)
 
+    async def compose(self, bucket: str, names: List[str], destination: str, **kwargs) -> None:
+        n = len(names)
+        if n == 0:
+            raise ValueError('no components in compose')
+        if n > 32:
+            raise ValueError(f'too many components in compose, maximum of 32: {n}')
+        assert 'json' not in kwargs
+        assert 'body' not in kwargs
```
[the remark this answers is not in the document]
cseed: body and json both set the body of the request, just in different formats.  If json is given, it is serialized to create the body.  aiohttp doesn't allow both of them to be set.  I could let aiohttp generate the error, but I thought that it would be cleaner to check it here since they don't set json, we do.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -540,26 +595,29 @@ async def copy_as_dir(self, worker_pool: AsyncWorkerPool, source_report: SourceR
...
+
+            relsrcfile = srcfile[len(src):]
+            assert not relsrcfile.startswith('/')
 
-                relsrcfile = srcfile[len(src):]
-                assert not relsrcfile.startswith('/')
+            await self._copy_file_multi_part(sema, source_report, srcfile, await srcentry.status(), url_join(full_dest, relsrcfile), return_exceptions)
 
-                await pool.call(self._copy_file, source_report, srcfile, url_join(full_dest, relsrcfile))
+        await bounded_gather2(sema, *[
```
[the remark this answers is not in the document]
cseed: So I believe this is correct because if return_exceptions=True, then _copy_file_multi_part (via copy_source) will never through because they will be caught and saved in the report, which means there will never be anything to cancel.  If return_exceptions=False, then we do want to cancel.

hail/python/hailtop/aiotools/fs.py
cseed: So I thought about this, but I'd like to hold off until I understand the right design better.  I think defaulting to create a new semaphore is dangerous, because if you leave it off, you accidentally increase the amount of parallelism being used.  I'd rather have the right/safe interface for the normal code and let the tests be a bit more complicated.

hail/python/hailtop/utils/utils.py
cseed: Yes, this was totally wrong. Fixed.

hail/python/test/hailtop/test_aiogoogle.py
cseed: Agreed, added a new test: test_multi_part_create_many

jigold (approved):

PR #9935, opened 2021-01-27
 [query-service] add JVM jobs to worker

catoverdrive:
Currently, this just spins up a new JVM instance; I need to integrate this with a different branch with a standing JVM server that listens on a TCP socket for incoming jobs.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -548,7 +548,7 @@ class JVMProcess:
             *self.flags,
             *self.java_args,
             stdout=asyncio.subprocess.PIPE,
-            stderr=asyncio.subprocess.STDERR)
+            stderr=asyncio.subprocess.PIPE)
         out, err = await self.proc.communicate()
 
         finish_time = time_msecs()
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -574,7 +574,7 @@ class JVMProcess:
             self.log)
 
         if self.proc.returncode == 0:
-            self.state = 'success'
+            self.state = 'succeeded'
         else:
             self.state = 'error'
 
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -534,7 +534,7 @@ class JVMProcess:
         for var in main_spec.get('env', []):
             self.env[var['name']] = var['value']
 
-        self.flags = ['-classpath', self.classpath, f'-Xmx:{self.heap_size}', f'-Xss:{self.stack_size}']
+        self.flags = ['-classpath', self.classpath, f'-Xmx{self.heap_size}', f'-Xss{self.stack_size}']
         self.java_args = main_spec['command']
 
         self.proc = None
@@ -542,7 +542,7 @@ class JVMProcess:
         self.state = 'pending'
         self.log = ''
 
-    async def run(self):
+    async def run(self, worker):
         log.info(f'running {self}')
         self.timing['start_time'] = time_msecs()
         self.proc = await asyncio.create_subprocess_exec(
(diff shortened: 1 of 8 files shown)


batch/batch/worker/worker.py
```diff
@@ -529,6 +530,49 @@ def __str__(self):
...
+        self.command.extend(main_spec['command'])
+        self.env = {e['name']: e['value'] for e in main_spec['env']}
+
+        self.timing = {}
+
+    async def run(self):
+        log.info(f'running {self}')
+        self.timing['start_time'] = time_msecs()
+        # run jvm job here
+        await check_shell_output(' '.join(self.command))
```
danking, line 551: We can avoid the shell:
```python3
proc = await asyncio.create_subprocess_exec(
    *self.command,
    stdout=asyncio.subprocess.PIPE,
    stderr=asyncio.subprocess.PIPE)
out, err = await proc.communicate()
# ...
return proc.returncode, out, err
```

danking: actually we probably want to redirect err to out so we can upload the whole thing to GCS as one file:
```
stderr=asyncio.subprocess.STDOUT
```

danking: This also gives you a `proc` that you can `kill()`

batch/batch/worker/worker.py
```diff
@@ -652,6 +696,16 @@ def gcsfuse_path(self, bucket):
...
 
+    @classmethod
+    def create(cls, batch_id, user, gsa_key, job_spec, format_version, task_manager):
+        type = job_spec['process']['type']
+        if type == 'docker':
+            return DockerJob(batch_id, user, gsa_key, job_spec, format_version, task_manager)
+        elif type == 'jvm':
+            return JVMJob(batch_id, user, gsa_key, job_spec, format_version, task_manager)
+        else:
+            raise Exception(f"unsupported job type {type}")
```
danking, line 707: Let's stick to assert's for impossible states. Also, in services code we stick to pylint style, so no-else-return.

```python3
if type == 'docker':
    return DockerJob(...)
assert type == 'jvm'
return JVMJob(...)
```

danking (changes requested):

batch/batch/worker/worker.py
```diff
@@ -528,6 +529,81 @@ def __str__(self):
...
+
+        await worker.log_store.write_log_file(
+            self.job.format_version, self.job.batch_id,
+            self.job.job_id, self.job.attempt_id, 'main',
+            self.log)
+
+        if self.proc.returncode == 0:
+            self.state = 'succeeded'
+        else:
+            self.state = 'error'
```
danking, line 586: non-zero exit code is `'failed'` whereas something like python can't open a subprocess is `'error'`.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -44,11 +44,15 @@ object Worker {
     if (args.length != 2) {
       throw new IllegalArgumentException(s"expected two arguments, not: ${ args.length }")
     }
-
     val root = args(0)
     val i = args(1).toInt
+    log.info(s"running job $i at root $root")
```
danking, line 49: let's log scratch dir too.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -188,14 +191,14 @@ class ServiceBackend() extends Backend {
...
-              JString(s"java -cp $$SPARK_HOME/jars/*:/hail.jar is.hail.backend.service.Worker $root $i"))),
-            "type" -> JString("docker")),
-          "mount_tokens" -> JBool(true))
+              JString("is.hail.backend.service.Worker"),
+              JString(root),
+              JString(s"$i"))),
+            "type" -> JString("jvm")),
+          "mount_tokens" -> JBool(true),
+          "resources" -> JObject(
+            "cpu" -> JString("1")))
```
danking, line 201: let's remove this before merge

danking (dismissed):

danking (approved):

danking:
@catoverdrive I did a bad, pip.conf needs to come from `docker/hail-ubuntu/pip.conf`, you also need the download-gcs-connector.py file I added to that branch

PR #9936, opened 2021-01-27
[query] Deprecate `hl.null` in favor of `hl.missing`

johnc1231:
CHANGELOG: Deprecated `hl.null` in favor of `hl.missing`.

Dan suggested that `hl.null` and `hl.is_missing` didn't match well. Now it's `hl.missing` and `hl.is_missing`, which makes more sense.


danking (approved):
Awesome! Thank you JohN!

PR #9941, opened 2021-01-27
[batch] Make billing project operations idempotent

jigold:
FYI @danking 

I wasn't sure if removing a previously deleted user should be an error. I think it should be, but I left it as idempotent for now.


danking (approved):
Awesome awesome awesome! Thank you so much Jackie!

PR #9942, opened 2021-01-27
[batch] maximal debugging information when a job fails

danking:
I'm seeing a PR failure that I can't debug further without the status information.


hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -345,6 +345,9 @@ async def jobs(self, q=None):
             if last_job_id is None:
                 break
 
+    async def get_job_log(self, job_id: int):
```
jigold, line 348: Eventually, we should clean up this entire API.

jigold (approved):

PR #9943, opened 2021-01-27
[site] add latest release of docs to site build instead of live polling

daniel-goldstein:
The site container polls regularly for new releases of hail and pulls the docs for the latest release. But we deploy site regularly enough that we can just include the latest release docs in the site build. Eventually, a deploy to PyPi should trigger a deploy of the site but this gets us part of the way there and simplifies things considerably.


build.yaml
```diff
@@ -1584,6 +1584,17 @@ steps:
            --verbose \
            site/www/* \
            /io/www
+
+     # Grab the latest release of the docs
+     LATEST_02_SHA=$(gsutil cat gs://hail-common/builds/0.2/latest-hash/cloudtools-5-spark-2.4.0.txt 2>/dev/null)
+     gsutil cat gs://hail-common/builds/0.2/docs/hail-0.2-docs-$LATEST_02_SHA.tar.gz |
```
jigold, line 1590: I think you need `${LATEST_02_SHA}`

build.yaml
```diff
@@ -1584,6 +1584,17 @@ steps:
...
+         tar zxvf - -C /io/www --strip-components=1
+
+     mkdir /io/www/docs/0.1
+     LATEST_01_SHA=$(gsutil cat gs://hail-common/builds/0.1/latest-hash-spark-2.0.2.txt 2>/dev/null)
+     gsutil cat gs://hail-common/builds/0.1/docs/hail-0.1-docs-$LATEST_01_SHA.tar.gz |
```
jigold, line 1595: The old polling scripts changed the file permissions. Do we need that?

build.yaml
```diff
@@ -1584,6 +1584,17 @@ steps:
            --verbose \
            site/www/* \
            /io/www
+
+     # Grab the latest release of the docs
+     LATEST_02_SHA=$(gsutil cat gs://hail-common/builds/0.2/latest-hash/cloudtools-5-spark-2.4.0.txt 2>/dev/null)
```
jigold, line 1589: Should we ensure it's not the empty string?

build.yaml
```diff
@@ -1584,6 +1584,17 @@ steps:
...
+     gsutil cat gs://hail-common/builds/0.2/docs/hail-0.2-docs-$LATEST_02_SHA.tar.gz |
+         tar zxvf - -C /io/www --strip-components=1
```
jigold, line 1591: You're not handling the batch docs I don't think.

jigold, line 1591: Should this be /docs/0.2 like 0.1 below?

jigold (dismissed):

build.yaml
daniel-goldstein: I don't think so because we update the permissions on the entire site directory (which includes the docs at that point) in the Dockerfile. Permissions had to be changed again in the polling loops because that was done after the image is already running.

build.yaml
daniel-goldstein: Sure. What would be the reaction to it being empty? Fail in this script?

build.yaml
daniel-goldstein: Huh so I thought I was handling them correctly but agree it is not at all obvious, and took some playing with. Here's the layout:
- the `0.2` tarball expands into the `docs` directory which contains `0.2` and `batch`
- since `0.1` doesn't have batch, the tarball is structured so that it expands just to a `0.1` directory, so I expand that into the `www/docs` directory
Screen Shot 2021-01-27 at 4 52 30 PM

Even so I am getting a 404 trying to looks at the batch docs on my dev deploy... I'll look into that

daniel-goldstein: Aha. So I suspect this is the problem:
https://github.com/hail-is/hail/blob/b125f2b46bf10c79bc7b13648976a0fb08249926/site/hail.nginx.conf.in#L36-L40
We don't create namespace prefix directories in `www`. The only reason this doesn't fail in prod is that `prefix` is "". I checked and dev deploy'ing main has this same problem.

I could remove the `@prefix@` or just remove the rule entirely and use the same redirects as we use for the `0.2` docs. Is there any reason to keep references to `pipeline` around?

daniel-goldstein: addressed above

daniel-goldstein: Confirmed that the batch docs load correctly after removing the `@prefix@` in the nginx config. The above question remains on whether we need the pipeline mentions anymore but up to you whether it's worth talking about in this PR.

build.yaml
jigold: I think so.

build.yaml
jigold: keep the pipeline mentions for now.

jigold:
It looks fine to me, but I'd prefer if @danking double checked the nginx stuff.

build.yaml
```diff
@@ -1584,6 +1584,25 @@ steps:
...
            /io/www
+
+     # Grab the latest release of the docs
+     LATEST_02_SHA=$(gsutil cat gs://hail-common/builds/0.2/latest-hash/cloudtools-5-spark-2.4.0.txt 2>/dev/null)
+     if [[ -z "$LATEST_02_SHA" ]]; then
+       echo "Failed to retrieve latest sha for hail docs"
+       exit 1
+     fi
+     gsutil cat gs://hail-common/builds/0.2/docs/hail-0.2-docs-${LATEST_02_SHA}.tar.gz |
+         tar zxvf - -C /io/www --strip-components=1
```
danking, line 1595: This will fail on the commit that creates a new hail PyPI version. This file is updated by `deploy`, which runs long after the site is deployed. The docs are uploaded by `deploy_hail` (confusing name, my bad).

Hmm. Maybe just make the site deployment depend on `deploy`. I think that makes everything work. latest-hash will be up to date. The docs will be published.

site/Dockerfile
```diff
@@ -5,10 +5,6 @@ RUN hail-apt-get-install \
   curl \
   libnginx-mod-http-subs-filter
 
-RUN /bin/sh -c 'curl https://sdk.cloud.google.com | bash' && \
-    mv /root/google-cloud-sdk /
-ENV PATH $PATH:/google-cloud-sdk/bin
-
```
danking, line 11: this brings joy; you just saved this image like 1.2GB of data.

build.yaml
danking: Yeah this looks totally broken, I don't understand why I included `@prefix@` in the try_files. Nice catch y'all.

danking (changes requested):

build.yaml
daniel-goldstein: Does `deploy` actually deploy to PyPi? Wouldn't we not want to do this every time we push out a change to the site? Or do you mean depend on `deploy_hail`?

danking: `deploy` pushes to PyPI, if there is a new PyPI version. Otherwise it does nothing.

build.yaml
```diff
@@ -1602,6 +1621,7 @@ steps:
      - base_image
      - build_site_www
      - hail_ubuntu_image
+     - deploy_hail
```
danking, line 1624: gotta be deploy, otherwise the latest-hash will not be up-to-date if this is a build of a commit that changes the PyPI version.

danking (changes requested):

daniel-goldstein:
bump

danking (approved):

PR #9944, opened 2021-01-28
[address] Close all existing connections on cleanup

daniel-goldstein:
This ended up being a great lesson in `asyncio`!

`kubernetes_asyncio` has a deeply hidden `asyncio.ClientSession` that doesn't get properly closed when we restart services with `SIGINT`. Turns out the `__del__` method on the `RESTClient` deep inside the library that holds this client session sets up a future with asyncio to close the session, but I suspect that this object is getting deleted _after_ the event loop closes. As a result we get a bunch of garbage in the logs that
- the event loop is already closed when something is trying to happen
- a `ClientSession.close` was never properly awaited

I initially tried to explicitly close the client session but still dealt with the problem that the `k8s_client` was trying to interact with the even loop after it closed. Explicitly deleting the `k8s_client` on cleanup and then awaiting any remaining futures (so the client session `close`) appears to fix the problem as I get no more errors on shutdown, but let me know if this isn't kosher. If this is alright I'll see this through to the rest of the services.

### DIFF
--- a/address/address/address.py
+++ b/address/address/address.py
@@ -143,7 +143,7 @@ async def on_cleanup(app):
         blocking_pool = app['blocking_pool']
         blocking_pool.shutdown()
         del app['cache'].k8s_client
-        await asyncio.get_event_loop().shutdown_asyncgens()
+        await asyncio.gather(*asyncio.all_tasks())
     finally:
         app['cache'].shutdown()
 
--- a/address/address/address.py
+++ b/address/address/address.py
@@ -143,7 +143,7 @@ async def on_cleanup(app):
         blocking_pool = app['blocking_pool']
         blocking_pool.shutdown()
         del app['cache'].k8s_client
-        await asyncio.gather(*asyncio.all_tasks())
+        await asyncio.gather(*(t for t in asyncio.all_tasks() if t is not asyncio.current_task()))
     finally:
         app['cache'].shutdown()
 
--- a/address/address/address.py
+++ b/address/address/address.py
@@ -142,6 +142,8 @@ async def on_cleanup(app):
     try:
         blocking_pool = app['blocking_pool']
         blocking_pool.shutdown()
+        del app['cache'].k8s_client
+        await asyncio.get_event_loop().shutdown_asyncgens()
     finally:
         app['cache'].shutdown()
 
--- a/build.yaml
+++ b/build.yaml
@@ -2198,6 +2198,7 @@ steps:
       for: alive
    dependsOn:
     - default_ns
+    - deploy_address_sa
     - deploy_router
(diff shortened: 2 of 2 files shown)


danking:
Hmm. But close isn't an async generator, so I think you're getting lucky in a race here. I think `await asyncio.gather(*asyncio.all_tasks())` would actually wait for everything. This will hang if we have tasks that were not properly stopped.

daniel-goldstein:
That definitely makes more sense. Thanks!

> This will hang if we have tasks that were not properly stopped.

Do you see this as problematic? I figure a pod would get terminated if it took too long to shut down gracefully and it could more obviously reveal the underlying problem.

danking:
I suppose I'm a bit confused about what happens if you await the completion of yourself. We should probably filter out the `asyncio.current_task()`.

I agree with the spirit of your sentence. It is problematic because somewhere *else* we're not properly stopping an infinite loop. Ideally we'll get to a place where we try to kill a pod and if it doesn't terminate in, say, 5 seconds, we fail the CI tests.

danking:
I believe k8s forcible terminates a pod 30s after it sends SIGTERM.

danking:
Yeah, we have to filter current_task out:
```
In [8]: async def silly(): 
   ...:     await asyncio.current_task() 
   ...:  
   ...: asyncio.get_event_loop().run_until_complete(silly())                                                                                                                                                                                                                                 
  File "<ipython-input-8-6930b633a59d>", line 4, in <module>
    asyncio.get_event_loop().run_until_complete(silly())
  File "/Users/dking/miniconda3/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
  File "<ipython-input-8-6930b633a59d>", line 2, in silly
    await asyncio.current_task()
RuntimeError: Task cannot await on itself: <Task pending coro=<silly() running at <ipython-input-8-6930b633a59d>:2> cb=[_run_until_complete_cb() at /Users/dking/miniconda3/lib/python3.7/asyncio/base_events.py:157]>

```

daniel-goldstein:
Ah ok. I was seeing some `kill -9` that are now gone when excluding the `current_task`.

> It is problematic because somewhere else we're not properly stopping an infinite loop. Ideally we'll get to a place where we try to kill a pod and if it doesn't terminate in, say, 5 seconds, we fail the CI tests.

Ya it might not deliver great clarity on where the error is coming from but at least we can actively kill and log an error if it's unable to do the "right" thing within say 5 seconds. That and I'm unsure how else to solve this problem

danking (approved):
Great detective work. If address passes the tests, I think we should roll this out to all other services using a k8s_client.

PR #9951, opened 2021-01-28
[query] Refactor struct code constructor to take codes, not values

tpoterba:
This makes the interaction with other infrastructure a bit simpler.

### DIFF
--- a/hail/src/main/scala/is/hail/types/physical/PCanonicalBaseStruct.scala
+++ b/hail/src/main/scala/is/hail/types/physical/PCanonicalBaseStruct.scala
@@ -2,7 +2,7 @@ package is.hail.types.physical
 
 import is.hail.annotations.{Region, UnsafeUtils}
 import is.hail.asm4s._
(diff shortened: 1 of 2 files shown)


hail/src/main/scala/is/hail/types/physical/PCanonicalBaseStruct.scala
```diff
@@ -2,7 +2,7 @@ package is.hail.types.physical
 
 import is.hail.annotations.{Region, UnsafeUtils}
 import is.hail.asm4s._
-import is.hail.expr.ir.{EmitCodeBuilder, EmitValue}
+import is.hail.expr.ir.{EmitCode, EmitCodeBuilder, EmitValue}
```
chrisvittal, line 5: EmitValue Unused

chrisvittal (commented):
One nit, great otherwise.

chrisvittal (dismissed):
see previous review

(sorry)

chrisvittal (approved):

PR #9952, opened 2021-01-29
[hail] make variant_qc, sample_qc, and impute_sex into aggregators

danking:


johnc1231:
Going to need to add them to the aggregators RST as well so they get documented.

hail/python/hail/methods/qc.py
```diff
@@ -116,59 +198,34 @@ def has_field_of_type(name, dtype):
         gq_dp_exprs['gq_stats'] = hl.agg.stats(mt.GQ).select('mean', 'stdev', 'min', 'max')
 
     if not has_field_of_type('GT', hl.tcall):
-        raise ValueError("'sample_qc': expect an entry field 'GT' of type 'call'")
+        raise ValueError("'variant_qc': expect an entry field 'GT' of type 'call'")
 
     bound_exprs['n_called'] = hl.agg.count_where(hl.is_defined(mt['GT']))
     bound_exprs['n_not_called'] = hl.agg.count_where(hl.is_missing(mt['GT']))
+    n_cols_ref = hl.expr.construct_expr(hl.ir.Ref('n_cols'), hl.tint32,
```
tpoterba, line 205: The definition of "call rate" prevents us from doing this 😢 

`n_cols_ref` is a reference to the number of columns in the MT. If you use `variant_qc_aggregator` here inside a group_by or filter, you'll be dividing by the total number of cols each time, not the filtered or grouped denominator.

tpoterba (changes requested):

danking:
We need a filter entries solution before any of this is in the realm of reasonable

PR #9953, opened 2021-01-29
Add note about avoiding duplicate data to export_elasticsearch docs

nawatts:
Add information from https://discuss.hail.is/t/export-elasticsearch-function-documentation-for-updating-behavior/1899 to `export_elasticsearch` docs. This happens fairly often with preemptible Dataproc workers and has caused us some confusion in the past with the gnomAD browser. Without some understanding of how Hail works, it's not immediately clear why this happens.


johnc1231 (approved):
Thanks for adding this!

PR #9954, opened 2021-01-29
[batch] Give users in billing project access to batches, jobs

jigold:
There might be an opportunity to also push NOT DELETED into the decorator as well. I chose the decorator route because it's cleaner and I expect a more complicated query for checking permissions in the future.

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -113,13 +113,7 @@ def rest_authenticated_developers_or_auth_only(fun):
     return wrapped
 
 
-async def _batch_billing_project_user(request, userdata):
-    app = request.app
-    db: Database = app['db']
-
-    batch_id = int(request.match_info['batch_id'])
-    user = userdata['username']
-
+async def _user_can_access(db: Database, batch_id: int, user: str):
     record = await db.select_and_fetchone(
         '''
 SELECT id
@@ -136,10 +130,13 @@ def rest_billing_project_users_only(fun):
     @rest_authenticated_users_only
     @wraps(fun)
     async def wrapped(request, userdata, *args, **kwargs):
-        permitted_user = await _batch_billing_project_user(request, userdata)
+        db = request.app['db']
+        batch_id = int(request.match_info['batch_id'])
+        user = userdata['username']
+        permitted_user = await _user_can_access(db, batch_id, user)
         if not permitted_user:
-            raise web.HTTPUnauthorized()
-        return await fun(request, userdata, *args, **kwargs)
+            raise web.HTTPNotFound()
+        return await fun(request, userdata, batch_id, *args, **kwargs)
     return wrapped
 
 
@@ -148,10 +145,13 @@ def web_billing_project_users_only(redirect=True):
         @web_authenticated_users_only(redirect)
         @wraps(fun)
         async def wrapped(request, userdata, *args, **kwargs):
-            permitted_user = await _batch_billing_project_user(request, userdata)
+            db = request.app['db']
+            batch_id = int(request.match_info['batch_id'])
+            user = userdata['username']
+            permitted_user = await _user_can_access(db, batch_id, user)
             if not permitted_user:
-                raise web.HTTPUnauthorized()
-            return await fun(request, userdata, *args, **kwargs)
+                raise web.HTTPNotFound()
+            return await fun(request, userdata, batch_id, *args, **kwargs)
         return wrapped
     return wrap
 
@@ -293,9 +293,7 @@ LIMIT 50;
 @routes.get('/api/v1alpha/batches/{batch_id}/jobs')
 @prom_async_time(REQUEST_TIME_GET_JOBS)
 @rest_billing_project_users_only
-async def get_jobs(request, userdata):  # pylint: disable=unused-argument
-    batch_id = int(request.match_info['batch_id'])
-
+async def get_jobs(request, userdata, batch_id):  # pylint: disable=unused-argument
     db = request.app['db']
     record = await db.select_and_fetchone(
         '''
@@ -469,8 +467,7 @@ async def _get_full_job_status(app, record):
 @routes.get('/api/v1alpha/batches/{batch_id}/jobs/{job_id}/log')
 @prom_async_time(REQUEST_TIME_GET_JOB_LOG)
 @rest_billing_project_users_only
-async def get_job_log(request, userdata):  # pylint: disable=unused-argument
-    batch_id = int(request.match_info['batch_id'])
+async def get_job_log(request, userdata, batch_id):  # pylint: disable=unused-argument
     job_id = int(request.match_info['job_id'])
     job_log = await _get_job_log(request.app, batch_id, job_id)
     return web.json_response(job_log)
@@ -1060,16 +1057,14 @@ WHERE id = %s AND NOT deleted;
 @routes.get('/api/v1alpha/batches/{batch_id}')
 @prom_async_time(REQUEST_TIME_POST_GET_BATCH)
 @rest_billing_project_users_only
-async def get_batch(request, userdata):  # pylint: disable=unused-argument
-    batch_id = int(request.match_info['batch_id'])
+async def get_batch(request, userdata, batch_id):  # pylint: disable=unused-argument
     return web.json_response(await _get_batch(request.app, batch_id))
 
 
 @routes.patch('/api/v1alpha/batches/{batch_id}/cancel')
 @prom_async_time(REQUEST_TIME_PATCH_CANCEL_BATCH)
 @rest_billing_project_users_only
-async def cancel_batch(request, userdata):  # pylint: disable=unused-argument
-    batch_id = int(request.match_info['batch_id'])
+async def cancel_batch(request, userdata, batch_id):  # pylint: disable=unused-argument
     await _handle_api_error(_cancel_batch, request.app, batch_id)
     return web.Response()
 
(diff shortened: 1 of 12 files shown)


jigold:
I can't figure out why I'm getting an error in one test. But I also am not sure what to do with the `/batches` endpoint. I want the UI default to only show you your batches with the default query string 'user:jigold`. However, what should the REST endpoint be? All batches in all billing projects you have access for?

batch/batch/front_end/front_end.py
```diff
@@ -115,6 +115,49 @@ async def wrapped(request, userdata, *args, **kwargs):
...
+    return record is not None
+
+
+def rest_billing_project_users_only(fun):
+    @rest_authenticated_users_only
+    @wraps(fun)
+    async def wrapped(request, userdata, *args, **kwargs):
+        permitted_user = await _batch_billing_project_user(request, userdata)
+        if not permitted_user:
+            raise web.HTTPUnauthorized()
```
danking, line 143: 401 Unauthorized doesn't seem right. 401 [suggests that the client has not sent the proper authentication information](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401).

The question is then should we return 403 or 404. I think our policy is to always return 404.

batch/batch/front_end/front_end.py
```diff
@@ -250,17 +293,16 @@ async def _query_batch_jobs(request, batch_id):
 
 @routes.get('/api/v1alpha/batches/{batch_id}/jobs')
 @prom_async_time(REQUEST_TIME_GET_JOBS)
-@rest_authenticated_users_only
-async def get_jobs(request, userdata):
+@rest_billing_project_users_only
+async def get_jobs(request, userdata):  # pylint: disable=unused-argument
     batch_id = int(request.match_info['batch_id'])
```
danking, line 298: The fas that `rest_billing_project_users_only` does not pass the batch_id to get_jobs means that a refactoring can easily break our protection mechanisms. Let's make both `_billing_project_users_only` pass the approved batch_id as an argument to the wrapped method.

batch/batch/front_end/front_end.py
```diff
@@ -534,10 +575,9 @@ async def _query_batches(request, user):
 
 @routes.get('/api/v1alpha/batches')
 @prom_async_time(REQUEST_TIME_GET_BATCHES)
-@rest_authenticated_users_only
-async def get_batches(request, userdata):
-    user = userdata['username']
-    batches, last_batch_id = await _query_batches(request, user)
+@rest_billing_project_users_only
```
danking, line 578: This request doesn't have a batch_id, right?

batch/batch/front_end/front_end.py
```diff
@@ -427,20 +469,19 @@ async def _get_full_job_status(app, record):
...
     return web.json_response(job_log)
 
 
-async def _query_batches(request, user):
+async def _query_batches(request):
     db = request.app['db']
 
-    where_conditions = ['user = %s', 'NOT deleted']
-    where_args = [user]
+    where_conditions = ['NOT deleted']
```
danking, line 483: This method appears to now have no restrictions at all?

batch/batch/front_end/front_end.py
```diff
@@ -1112,10 +1144,9 @@ async def ui_delete_batch(request, userdata):
 
 @routes.get('/batches', name='batches')
 @prom_async_time(REQUEST_TIME_GET_BATCHES_UI)
-@web_authenticated_users_only()
+@web_billing_project_users_only()
 async def ui_batches(request, userdata):
-    user = userdata['username']
-    batches, last_batch_id = await _query_batches(request, user)
+    batches, last_batch_id = await _query_batches(request)
```
danking, line 1149: this now appears to return all batches for all billing projects

danking (changes requested):

batch/batch/front_end/front_end.py
jigold: Can you give an example?

batch/batch/front_end/front_end.py
jigold: No. It's broken. See the comment I made above about making the default query string `f'user:{user}'`

batch/batch/front_end/front_end.py
jigold: Again, see comment above. We have to do something different for query_batches or queries that don't have a specific batch idea like query_billing (which I already implemented previously).

batch/batch/front_end/front_end.py
jigold: See comment above.

batch/test/test_accounts.py
```diff
@@ -398,3 +398,101 @@ async def test_billing_limit_tiny(make_client, dev_client, new_billing_project):
...
+
+    try:
+        await dev_client._patch(f'/api/v1alpha/batches/{b.id}/cancel')
+    except aiohttp.ClientResponseError as e:
+        assert e.status == 401, e
+    else:
+        assert False, 'expected error'
+
+    try:
+        await dev_client._delete(f'/api/v1alpha/batches/{b.id}/delete')
```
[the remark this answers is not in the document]
jigold: That doesn't make sense. It should error out on line 443. I just added delete and cancel as methods on batch_client. At some point we should rewrite the whole thing.

batch/batch/front_end/front_end.py
danking: Like this:
```
 async def get_jobs(request, userdata, batch_id):  # pylint: disable=unused-argument
```
```
async def user_can_access(db: Database, batch_id: int, username: str):
     return None is not await db.select_and_fetchone(
         '''
 SELECT id
 FROM batches
 LEFT JOIN billing_project_users ON batches.billing_project = billing_project_users.billing_project
 WHERE id = %s AND billing_project_users.`user` = %s;
 ''',
         (batch_id, username))

def rest_billing_project_users_only(fun):
     @rest_authenticated_users_only
     @wraps(fun)
     async def wrapped(request, userdata, *args, **kwargs):
         batch_id = request['batch_id']
         if not await user_can_access(request['db'], batch_id, userdata['username']):
             raise web.HTTPNotFound()
        return await fun(request, userdata, batch_id, *args, **kwargs)
     return wrapped
```

batch/test/test_accounts.py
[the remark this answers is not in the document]
danking: ah, right, so we can just construct a batch object then, right?
```
Batch(dev_client, b.id, b.attributes, b.n_jobs)
```
I just mean that we should use the API we've built. That forces us to make sure the API works well.

batch/test/test_accounts.py
```diff
@@ -398,3 +398,101 @@ async def test_billing_limit_tiny(make_client, dev_client, new_billing_project):
     batch = await batch.submit()
     batch = await batch.wait()
     assert batch['state'] == 'cancelled', batch
+
+
+async def test_all_billing_project_users_operations(make_client, dev_client, new_billing_project):
```
danking, line 403: and this is something like `test_user_can_access_batch_made_by_other_user_in_shared_billing_project`, yeah?

batch/batch/front_end/front_end.py
```diff
@@ -471,6 +509,18 @@ async def _query_batches(request, user):
 ((batches.id) IN
  (SELECT batch_id FROM batch_attributes
   WHERE `key` = %s))
+'''
+            args = [k]
+        elif t.startswith('user:'):
```
danking, line 514: Can you add these new options to the documentation in `batches.html`?

batch/batch/front_end/front_end.py
```diff
@@ -448,7 +486,7 @@ async def _query_batches(request, user):
         where_conditions.append('(id < %s)')
         where_args.append(last_batch_id)
 
-    q = request.query.get('q', '')
+    q = request.query.get('q', f'user:{user}')
```
danking, line 489: This needs to be consistent with the UI. Currently the UI will show no search term. We should probably explicitly update the `request.query` term in both `get_batches` and `ui_batches`. That ensures that when we regenerate the page in `ui_batches`, we'll use the right query term.

danking (dismissed):

ci/Makefile
```diff
@@ -13,7 +13,7 @@ PYTHON := PYTHONPATH=$${PYTHONPATH:+$${PYTHONPATH}:}$(EXTRA_PYTHONPATH) python3
 check:
 	$(PYTHON) -m flake8 ci
 	$(PYTHON) -m pylint --rcfile ../pylintrc ci --score=n
-	$(PYTHON) -m black ci --check --line-length=120 --skip-string-normalization
+	$(PYTHON) -m black ci --check --diff --line-length=120 --skip-string-normalization
```
danking, line 16: what's this?

danking (dismissed):

ci/Makefile
jigold: I was getting errors in this check. `--diff` actually gives you the difference on what your original files looked like and what the rewrite looks like. I think this is fine to keep in so someone else who modifies CI doesn't have to look for this option if this check fails.

batch/batch/front_end/templates/batches.html
```diff
@@ -25,6 +25,8 @@ <h1>Batches</h1>
 	  <ul>
 	    <li>k=v - batches with an attribute with key k and value v</li>
 	    <li>has:k - batches that have an attribute with key k</li>
```
daniel-goldstein, line 27: This line confuses me now. Should this be `col:v` and not be attribute?

batch/batch/front_end/templates/batches.html
```diff
@@ -25,6 +25,8 @@ <h1>Batches</h1>
...
+			<li>user:u - batches that were created by user u</li>
+			<li>billing_project:bp - batches that were created in billing project bp</li>
```
danking, line 29: change these tabs to spaces and I think we're good to go.

danking (changes requested):

danking (approved):

PR #9955, opened 2021-01-29
[datasets] add gnomAD datasets

pwc2:
To make existing gnomAD datasets (already in HailTable/MatrixTable form) available via `load_dataset`:
  - For v2 now includes multinucleotide variants (MNVs), proportion expressed across transcript (pext), and linkage disequilibrium (LD)
  - For v3 now includes mitochondrial DNA (mtDNA)

The `load_dataset` function now can open a `BlockMatrix` as well, so the gnomAD v2 LD matrices are available through the datasets API in addition to the tables.

Also added try/except blocks to `load_datasets`, to handle errors. For example the `gnomad_annotation_pext` dataset has a url of `gs://gcp-public-data--gnomad/papers/2019-tx-annotation/pre_computed/all.possible.snvs.tx_annotated.021520.ht`, but this is actually a `MatrixTable`.  So now when trying to open a `HailTable`, it will try to open as a `MatrixTable` if an error is encountered (and vice versa).

### DIFF
--- a/hail/python/hail/docs/datasets/schemas/clinvar_gene_summary.rst
+++ b/hail/python/hail/docs/datasets/schemas/clinvar_gene_summary.rst
@@ -5,7 +5,7 @@ clinvar_gene_summary
 
 *  **Versions:** 2019-07
 *  **Reference genome builds:** None
-*  **Type:** :class:`.Table`
+*  **Type:** :class:`hail.Table`
 
 Schema (2019-07, None)
 ~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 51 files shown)


danking (approved):
Awesome work; thank you!

gtiao:
Next time, let us know when you come across issues with our data -- we want to hear about bugs, and in this case it would have been an easy fix on our end!

danking:
@gtiao, great!

We haven't released yet, so we can still fix. It seems the right fix is to re-generate that file as a Table rather than a MatrixTable because it doesn't have any entry data. @pwc2 could you furnish a list of all the misnamings?

pwc2:
Sounds good, thanks @gtiao! I just double checked and it looks like the only dataset with issues is the annotation-level pext for all possible SNVs Hail table at `gs://gcp-public-data--gnomad/papers/2019-tx-annotation/pre_computed/all.possible.snvs.tx_annotated.021520.ht`. As Dan said, regenerating this file as a Table should be all that is needed to fix the issue.

gtiao:
Julia has regenerated the file with the correct extension, and it is available at `"gs://gnomad-public/papers/2019-tx-annotation/pre_computed/all.possible.snvs.tx_annotated.021520.ht`

PR #9958, opened 2021-01-29
[query] Finish purging SRVB from aggregators

tpoterba:


hail/src/main/scala/is/hail/expr/ir/agg/TakeByAggregator.scala
```diff
@@ -522,39 +521,33 @@ class TakeByRVAS(val valueVType: VirtualTypeWithReq, val keyVType: VirtualTypeWi
...
+        r.load().allocate(4L, ab.size.toL * 4L))
+
+      val i = cb.newLocal[Int]("i", 0)
+
+      cb.whileLoop(i < ab.size, {
+        cb += Region.storeInt(indicesToSort + i.toL * 4L, i)
+        cb.assign(i, i + 1)
+      })
+
+      def indexOffset(idx: Code[Int]): Code[Long] = indicesToSort + idx.toL * 4L
```
cseed, line 537: Move this up, you use this expression above and you can use this function instead.

johnc1231 (changes requested):
Just Cotton's comment

johnc1231 (approved):

PR #9960, opened 2021-01-30
[query] Refactor NDArray constructor interfaces

tpoterba:
* construct doesn't take structs, takes IndexedSeq[Value[Long]]
* Implement the nested loops in one place
* Delete some methods from type interface, insert PCanonical casts


hail/src/main/scala/is/hail/linalg/LinalgCodeUtils.scala
```diff
@@ -32,41 +33,19 @@ object LinalgCodeUtils {
...
-        dimVar := 0L,
-        Code.whileLoop(dimVar < shape(dimIdx),
-          innerLoops,
-          dimVar := dimVar + 1L
-        )
-      )
+    val (addElem, finish) = dataType.constructFromFunctions(cb, region, dataLength, deepCopy = false)
+    val idx = cb.newLocal[Int]("nda_create_column_major_idx", 0)
+    SNDArray.forEachIndex(cb, shape, "nda_create_column_major") { case (cb, idxVars) =>
+      addElem(cb, idx, IEmitCode.present(cb, pndv.loadElement(idxVars, cb).asPCode))
```
johnc1231, line 44: You're getting assertion errors from here. I think it's not ok to `IEmitCode.present` multiple times inside of an `IEmitCode.map`, which is what you're indirectly doing here within the NDArrayEmitter

johnc1231 (changes requested):

hail/src/main/scala/is/hail/linalg/LinalgCodeUtils.scala
tpoterba: debugged that yesterday -- it's actually that IEmitCode.present is totally busted. I pulled out the fix in #9962

tpoterba: (assigned to you)

tpoterba: I'm now getting failures about mismatched shapes in BLockMatrixDot examples (and others), I must some incorrect shape calculations.

johnc1231:
Also some import conflicts that have to be resolved

johnc1231:
Addressed your initial problem and merge conflicts, but looks like there's more that's going wrong. Seems like it's mostly data comparison failures with reasonableish looking data, probably a striding issue.

johnc1231:
The problem is your loops are backwards. `SType.foreach` goes in row major order, and `NDArrayEmitter` expects to emit column major stuff.

johnc1231:
Ok, I switched the implementation of `SNDArray.forEach` to the column major style we'd been using in the NDArrayEmitter previously, and now I think everything should pass.

johnc1231 (approved):

johnc1231:
Anything that uses LAPACK seems broken. My guess is that this means `emitNDArrayColumnMajorStrides` is busted.

johnc1231:
Definitely something funny happening here:

```
def emitNDArrayColumnMajorStrides(ir: IR): IEmitCode = {
      emitI(ir).map(cb){case pNDCode: PNDArrayCode =>
        val pNDValue = pNDCode.memoize(cb, "ndarray_column_major_check")
        val isColumnMajor = LinalgCodeUtils.checkColumnMajor(pNDValue, cb)
        val pAnswer = cb.emb.newPField("ndarray_output_column_major", pNDValue.pt)
        cb.ifx(false, {cb.assign(pAnswer, pNDValue)},
        {
          cb.println("Shoud be making a copy")
          cb.assign(pAnswer, LinalgCodeUtils.createColumnMajorCode(pNDValue, cb, region.code))
        })
        val ndarrayColMajor = pAnswer.asNDArray.memoize(cb, "temp")
        cb.println("col major output")
        cb.println(pNDValue.strides(cb).map(_.toS.concat(" ")):_*)
        cb.println("col major output over")
        pAnswer.get
      }
    }
```

The prints are coming out with row major striding, even though prints internal to `LinalgCodeUtils` look correct.

johnc1231:
I believe everything works now, the problem was that `dataAddress` function was returning address of first element of data array, but we were acting like it was returning a pointer to the length of the array. Changed the name and the assumptions, and now it all works.

tpoterba:
thank you John!

PR #9964, opened 2021-02-01
[k8s] Close k8s client on cleanup

daniel-goldstein:
A follow-up on #9944 to make sure that client sessions used by the k8s client are closed on cleanup across all services that use it, along with a couple other client connections that need to get closed on shutdown.

### DIFF
--- a/scorecard/scorecard/scorecard.py
+++ b/scorecard/scorecard/scorecard.py
@@ -418,9 +418,11 @@ async def on_startup(app):
 async def on_shutdown(app):
     try:
         await app['asana_client'].close()
-        await app['gh_session'].close()
     finally:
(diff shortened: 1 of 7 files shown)


scorecard/scorecard/scorecard.py
```diff
@@ -418,6 +418,7 @@ async def on_startup(app):
 async def on_shutdown(app):
     try:
         await app['asana_client'].close()
+        await app['gh_session'].close()
```
danking, line 421: we've been nesting these try-finallies to ensure everything gets closed, let's be consistent, though I admit these deeply nested try-finallies seem odd to me.

danking (changes requested):

danking (approved):

PR #9968, opened 2021-02-01
[query] no SRVB in MakeTuple, InsertFields, MakeArray

patrick-schultz:


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -842,6 +847,36 @@ class Emit[C](
...
+
+              codeOld.map(cb) { oldPC =>
+                val oldPV = oldPC.asBaseStruct.memoize(cb, "insert_fields_old")
+
+                val itemsEC = x.pType.fields.map { f =>
+                  updateMap.get(f.name) match {
+                    case Some(vir) =>
+                      EmitCode.fromI(mb)(emitInNewBuilder(_, vir))
+                    case None =>
+                      EmitCode.fromI(mb)(oldPV.loadField(_, f.index).asInstanceOf[IEmitCode])
```
tpoterba, line 867: this cast may fail. I think we should make it `.typecast[PCode]`

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -842,6 +847,36 @@ class Emit[C](
...
+                  }
+                }
+
+                x.pType.asInstanceOf[PCanonicalBaseStruct]
+                 .constructFromFields(cb, region.code, itemsEC, deepCopy = false)
+                 .asPCode
+              }
+            case _ =>
```
tpoterba, line 875: this seems like dead code -- isn't the old ptype always a struct?

tpoterba (dismissed):

hail/src/main/scala/is/hail/expr/ir/Emit.scala
patrick-schultz: I figured this was to handle InsertFields on a tuple. But you're right, we don't allow that in type checking/inference. I'll delete.

tpoterba (approved):

PR #9971, opened 2021-02-02
[letsencrypt] fix letsencrypt

danking:
Installing certbot is hard so I stopped doing that. Instead, I use the cerbot docker image.

I also eliminated `sed` use in the letsencrypt directory.

We now maintain the options-ssl-nginx.conf file ourselves. I copied the settings from certbot
GitHub. They're straightforward, as a part of regular package versioning maintenance we should also
reconsider our cipher suites and TLS versions.

We now have a `make run DRY_RUN=true` option which can be run repeatedly without affecting the default certs.


daniel-goldstein (approved):
LGTM, I'll get started on a regular job for this and keep in mind to look back at the ciphers and TLS versions

PR #9972, opened 2021-02-02
[batch] Add job private instances

jigold:
**User-facing Changes:**
- Added preemptible and machine_type to resources configuration in addition to existing worker_type (will get removed at some point in the future).
- machine_type and preemptible are hidden options in hailtop.batch (I put them in for testing purposes)
- Can only specify one of worker_type and machine_type
- Preemptible is only valid for machine_type
- If you specify the machine_type, your storage gets rounded up to the nearest Gi with 10 Gi being the minimum (this is because we use a persistent-SSD as the worker data disk and the min for this is 10 Gi.

**Billing Changes:**
- The list of possible machine types we currently support is in the globals -- we only support the n1 family. I insert new resource rates into the Resources SQL table to account for the new nonpreemptible resources.
- If you have a job private instance, you are billed for that instance at the time of the instance's creation (not when it's activated). This is done by modifying the trigger for the attempts table to take the minimum attempt start time rather than the maximum. I'm not sure why we needed the maximum to begin with as the attempt start time was always the same. This is probably a place to double check before merging.

**Job State Changes:**
- We now support a new state "Creating" which represents an instance has been spun up for a job, but it has not been activated yet
- The SQL user resources table and cancellable resources table needed to be changed to add the n_creating_jobs, n_cancellable_creating_jobs, and the n_cancelled_creating_jobs
- Added a new SQL function `mark_job_creating` that is only called by the job private instance creator.
- A lot of the existing SQL functions had to change to account for the fact that an instance in the pending state can have job-specific operations done such as mark_job_complete if the job is cancelled. In addition, the "Creating" state is like "Running" for some operations in that an attempt has been created and actions are happening on behalf of the user.

**Driver Changes:**
- New cancel_creating_jobs event
- Two separate methods to get the pools or job private UI pages and two separate configuration methods. One each for pool and job-private.

**JobPrivateInstanceCollection:**
- Has two new loops: an instance creation loop and a scheduling loop
- The instance creation loop does a fair share calculation that is almost identical to the pool one except the resource being allocated is n_ready_jobs compared to total_jobs rather than ready_cores_mcpu. 
- The instance creation loop needs to extract the machine_type, storage_gib, and preemptible from the spec without hitting GCS. Therefore, it is stored in the "spec" field in the database which required changing the batch format version a bit.
- We avoid double scheduling by requiring that there are no live instances assigned to attempts for that job before creating an instance.
- We mark a job as creating after creating the instance for the new attempt
- The number of instances that can be created is similar to the pool control loop. The total number of instances we can create is fed to the fair share allocator.
- I added an asyncio.wait(15) at the end of the instance creation loop body to make sure we didn't run past our GCE limits.
- The scheduling loop iterates over all attempts with active instances in order of time of activation (no user fair share here -- FIFO)
- There is no possibility of double scheduling because there must only be one active instance per job based on the create instances loop

**Canceller:**
- There's a new canceller loop that looks for jobs that need to be cancelled in the creating state. It marks these jobs as complete "cancelled" in the database and then calls GCE to delete the instance.

**Mark Job Complete:**
- I modified this function to kill a job private instance if the job is marked as complete and the instance is active

**Worker:**
- I added a kill function
- Note: I did not change how storage is computed. For job private instances, it's possible to be billed for 10Gi but only get 5 Gi if you requested 5 Gi in the XFS quotas. I decided that thinking through the storage here can be delayed until the storage PR since no one is going to be using this functionality yet.

**Testing:**
- I added three new job private instance tests: preemptible, non preemptible, preemptible with cancellation in the creating state
- I also added the creating state to the check_incremental background test.
- I ran a chaos script and made sure everything was working with regards to cancellation.
- I looked at the UI pages in my dev deploy and made sure they were working

**Other:**
- The free cores for an instance is set to 0 at the time of instance creation. I wanted this behavior because we're billing for that time and the job has been allocated to that instance. So the free cores should be 0. It might be confusing. I also think of free cores as a resource we're wasting where as in this case, we're not wasting resources.


jigold:
Also, charging for activation time won't show up in the durations for the job, but will in the billing time for the attempts. Maybe this behavior is okay?

jigold:
I thought about the problem of the UI a bit and I think this is where input from others might be good.

jigold:
This is ready to look at. i tried to track down the one failure in test_benchmark, but couldn't do so.

batch/batch/worker/worker.py
```diff
@@ -1086,13 +1097,30 @@ async def run(self):
...
-                log.info('cleaned up app runner')
+            await self.cleanup()
+            self.shutdown()
+
+    async def kill(self, request):
+        body = await request.json()
+
+        expected_name = body['name']
+        if expected_name != NAME:
+            return web.HTTPForbidden()
```
danking, line 1108: If we're concerned enough that we could have the wrong IP for an instance in this case, shouldn't we verify the name on every request?

batch/batch/worker/worker.py
```diff
@@ -1086,13 +1097,30 @@ async def run(self):
...
+            async with client_session(
+                    raise_for_status=True, timeout=aiohttp.ClientTimeout(total=60)) as session:
+                # Don't retry.  If it doesn't go through, the driver
+                # monitoring loops will recover.  If the driver is
+                # gone (e.g. testing a PR), this would go into an
+                # infinite loop and the instance won't be deleted.
+                await session.post(
+                    deploy_config.url('batch-driver', '/api/v1alpha/instances/deactivate'),
+                    headers=self.headers)
+            log.info('deactivated')
```
danking, line 1120: this seems like a clean easy to abstract function

batch/batch/worker/worker.py
```diff
@@ -1086,13 +1097,30 @@ async def run(self):
...
+        finally:
+            await self.cleanup()
+            self.shutdown()
```
danking, line 1123: this feels a bit funky. The coroutine executing `run` will stay open until the idle time gets long. That while loop in `run` should have a way to indicate you want to stop. Then you don't need to duplicate the deactivate and shutdown code at all.

danking: Something like replacing the sleep with:
```
try:
    await asyncio.wait_for(stop_event.wait(), 15)
    break
except asyncio.TimeoutError:
    idle_duration = time_msecs() - self.last_updated
```

batch/batch/worker/worker.py
```diff
@@ -562,7 +562,7 @@ async def add_gcsfuse_bucket(mount_path, bucket, key_file, read_only):
         except asyncio.CancelledError:  # pylint: disable=try-except-raise
             raise
 
-    delay = await sleep_and_backoff(delay)
+        delay = await sleep_and_backoff(delay)
```
danking, line 565: put this in a standalone PR and lets get it in ASAP

batch/batch/driver/instance_collection_manager.py
```diff
@@ -16,10 +17,11 @@ def __init__(self, app, machine_name_prefix):
         self.db: Database = app['db']
         self.zone_monitor: ZoneMonitor = app['zone_monitor']
         self.machine_name_prefix = machine_name_prefix
-        self.inst_coll_regex = re.compile(f'{self.machine_name_prefix}(?P<inst_coll>[^-]+)-.*')
+        self.inst_coll_regex = re.compile(f'{self.machine_name_prefix}(?P<inst_coll>.*)-.*')
 
         self.name_inst_coll = {}
         self.name_pool = {}
+        self.name_job_private = {}
```
danking, line 24: let's add type annotations to these dictionaries

batch/batch/driver/instance.py
```diff
@@ -102,6 +107,31 @@ async def deactivate(self, reason, timestamp=None):
...
+            except Exception as e:
+                if (isinstance(e, aiohttp.ClientResponseError)
+                        and e.status == 403):  # pylint: disable=no-member
+                    log.info(f'cannot kill {self} -- does not exist at {self.ip_address}')
+                    break
+                if is_transient_error(e):
+                    pass
+                else:
+                    raise
+            delay = await sleep_and_backoff(delay)
```
danking, line 133: We can use use retry_transient_errors here:
```python3
async def make_request():
    if self._state in ('inactive', 'deleted'):
        return
    async with aiohttp.ClientSession(raise_for_status=True, timeout=aiohttp.ClientTimeout(total=30)) as session:
        url = f'http://...'
        try:
            await session.post(url, json=body)
        except ClientResponseError as err:
            if err.status == 403:
                 return
            raise

await retry_transient_errors(make_request)
```

batch/batch/driver/instance_collection_manager.py
```diff
@@ -16,10 +17,11 @@ def __init__(self, app, machine_name_prefix):
         self.db: Database = app['db']
         self.zone_monitor: ZoneMonitor = app['zone_monitor']
         self.machine_name_prefix = machine_name_prefix
-        self.inst_coll_regex = re.compile(f'{self.machine_name_prefix}(?P<inst_coll>[^-]+)-.*')
+        self.inst_coll_regex = re.compile(f'{self.machine_name_prefix}(?P<inst_coll>.*)-.*')
```
danking, line 20: inst_coll can't be empty, right? Should this be .+?

batch/batch/driver/job.py
```diff
@@ -155,6 +162,30 @@ async def mark_job_started(app, batch_id, job_id, attempt_id, instance, start_ti
...
+
+    try:
+        rv = await db.execute_and_fetchone(
+            '''
+CALL mark_job_creating(%s, %s, %s, %s, %s);
+''',
+            (batch_id, job_id, attempt_id, instance.name, start_time))
+    except Exception:
+        log.exception(f'error while marking job {id} creating on {instance}')
+        raise
```
danking, line 180: This function, `mark_job_creating`, is called once. At that call-site, the exception is caught and logged as `info`. Is this exception actually exceptional?

batch/batch/driver/main.py
```diff
@@ -440,43 +442,144 @@ def validate_int(name, value, predicate, description):
...
+    def validate_int(name, value, predicate, description):
+        try:
+            i = int(value)
+        except ValueError as e:
+            set_message(session,
+                        f'{name} invalid: {value}.  Must be an integer.',
+                        'error')
+            raise web.HTTPFound(deploy_config.external_url('batch-driver', url_path)) from e
+        return validate(name, i, predicate, description)
+
```
danking, line 481: these validators can be lifted out if we add a session argument.

batch/batch/driver/templates/index.html
```diff
@@ -37,6 +39,34 @@ <h1>Pools</h1>
     </tbody>
   </table>
 
+  <h2>Job Private</h2>
```
danking, line 42: Pools is a plural noun so I think we should call this `Job Private Instances`

batch/batch/driver/main.py
```diff
@@ -440,43 +442,144 @@ def validate_int(name, value, predicate, description):
     return web.HTTPFound(deploy_config.external_url('batch-driver', pool_url_path))
 
 
-@routes.get('/inst_coll/{inst_coll}')
+@routes.post('/config-update/job-private/{job_private}')
```
danking, line 445: Why there are multiple job private instance collections? I was imagining exactly one job private instance collection which managed all the job private instances.

batch/batch/front_end/inst_coll_selector.py
```diff
@@ -0,0 +1,82 @@
+import re
```
danking, line 1: I do not think this file and these classes enhance my reasoning capabilities about this code. PoolConfig seems like a stripped down Pool. PoolSelector seems very small? I thought all this would be managed by the InstanceCollectionManager who already has a list of pools and knows about the JPIM. It has all the information necessary to select a pool for a job.

batch/batch/driver/instance_collection_manager.py
```diff
@@ -16,10 +17,11 @@ def __init__(self, app, machine_name_prefix):
...
 
         self.name_inst_coll = {}
         self.name_pool = {}
+        self.name_job_private = {}
 
     async def async_init(self):
         inst_coll_records = self.db.execute_and_fetchall('''
```
danking, line 27: I feel like we should just get all the necessary information in one DB select? We can `LEFT JOIN` onto the pools table to get the necessary information (or NULL in the JPIM case) and pass it all into the Pool.

danking (changes requested):
The overall structure seems fine, though I'm not sure what's up with multiple JPIMs

jigold:
For the multiple JPICs, I was thinking that some day there could be a JPIC per cloud. I can try and convert it back so it's clear there is one instance of the class if that is clearer for now.

batch/batch/front_end/inst_coll_selector.py
jigold: That code currently creates instances and starts the control loops which I didn't want to do for obvious reasons on the front end. I can try and refactor this again and see if I can solve it a different way.

batch/batch/driver/templates/index.html
jigold: That doesn't make sense. It would have to be Job Private Instance Collection.

batch/batch/driver/instance_collection_manager.py
jigold: It used to be empty before when we did the pools migration. I'll have to double check with where this is used to make sure we can switch the regex.

danking:
I think it's clearer with one instance of the JPIC

batch/batch/driver/instance_collection_manager.py
jigold: Let's hold off on this for now. I'm just worried about the case where in the future we want to do something with instances that date pre-pool change.

jigold:
Closing this while I debug with dev deploy.

batch/batch/worker/worker.py
jigold: I'll leave this for a separate PR.

jigold:
I just realized there might be an issue with how storage is handled on the worker.

jigold:
ok. I'm happy with this.

danking (approved):

PR #9973, opened 2021-02-02
[vep] Rework VEP with more visibility

tpoterba:
1. Don't ignore parse errors by default. These shouldn't be happening
   with modern VEP versions anyway (they were caused by genes named NaN
   or something in older versions)
2. Don't silently drop variants that don't have a match in VEP. All keys
   coming in will have a row coming out with a VEP annotation (maybe NA)
   and a partition/block index.

### DIFF
--- a/hail/python/hail/methods/qc.py
+++ b/hail/python/hail/methods/qc.py
@@ -482,8 +482,9 @@ def concordance(left, right, *, _localize_global_statistics=True) -> Tuple[List[
            config=nullable(str),
            block_size=int,
            name=str,
-           csq=bool)
-def vep(dataset: Union[Table, MatrixTable], config=None, block_size=1000, name='vep', csq=False):
+           csq=bool,
+           tolerate_parse_error=bool)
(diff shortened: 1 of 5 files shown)


chrisvittal (approved):

johnc1231:
Not compiling

hail/python/hail/methods/qc.py
```diff
@@ -482,8 +482,9 @@ def n_discordant(counter):
            config=nullable(str),
            block_size=int,
            name=str,
-           csq=bool)
-def vep(dataset: Union[Table, MatrixTable], config=None, block_size=1000, name='vep', csq=False):
+           csq=bool,
+           tolerate_parse_error=bool)
+def vep(dataset: Union[Table, MatrixTable], config=None, block_size=1000, name='vep', csq=False, tolerate_parse_error=False):
```
chrisvittal, line 487: Missed this first go, keyword only for the new parameter?

chrisvittal (changes requested):

chrisvittal (approved):

PR #9974, opened 2021-02-02
[query] Add typechecking rule to detect badly-placed void arguments

tpoterba:


patrick-schultz:
It's a little concerning that this is necessary. Are there typecheck cases that don't make assertions about some of their children?

tpoterba:
> It's a little concerning that this is necessary. Are there typecheck cases that don't make assertions about some of their children?

This check would have to go in a LOT of places otherwise. Reading down from the top: CastRename, NA, IsNA, Coalesce, AggLet, TailLoop, MakeArray, ...

patrick-schultz:
Yeah, you're right. What's concerning is that I think of the typecheck rule for an IR as the one place to look up exactly what the valid uses are. But the weirdness of TVoid does seem worth special handling.

patrick-schultz (approved):

PR #9975, opened 2021-02-02
[batch] wait for iptables chain before initializing

danking:
Without this change, we race against dockerd. If the chain does not exist
before we add the rules, then containers will be able to access the metadata server
other containers, etc.


danking:
cc: @jigold

daniel-goldstein (approved):

PR #9977, opened 2021-02-03
[query] Refactor registerIEmitCode to use EmitCode instead of thunks

tpoterba:
() => IEmitCode is now a worse version of EmitCode!

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -243,7 +243,7 @@ object IEmitCode {
     IEmitCodeGen(Lmissing, Lpresent, pc)
   }
 
-  def strictMapEmitCodes(cb: EmitCodeBuilder, seq: IndexedSeq[EmitCode])(f: IndexedSeq[PCode] => PCode): IEmitCode = {
+  def multiMapEmitCodes(cb: EmitCodeBuilder, seq: IndexedSeq[EmitCode])(f: IndexedSeq[PCode] => PCode): IEmitCode = {
     val Lmissing = CodeLabel()
     val Lpresent = CodeLabel()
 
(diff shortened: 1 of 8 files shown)


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -243,8 +243,24 @@ object IEmitCode {
     IEmitCodeGen(Lmissing, Lpresent, pc)
   }
 
-  def flatten[A, B](seq: IndexedSeq[() => IEmitCodeGen[A]], cb: EmitCodeBuilder)(f: IndexedSeq[A] => B): IEmitCodeGen[B] =
-    sequence(seq, { (i: () => IEmitCodeGen[A]) => i() }, cb)(f)
+  def strictMapEmitCodes(cb: EmitCodeBuilder, seq: IndexedSeq[EmitCode])(f: IndexedSeq[PCode] => PCode): IEmitCode = {
```
patrick-schultz, line 246: This can still be handled by `sequence`:
```
sequence(seq, ec => ec.toI(cb), cb)(f)
```

And nitpicking about the name: I don't think the "strict" is necessary. I think `mapEmitCodes` is fine, or even something like `multiMap` or `mapN`. (I do agree flatten was a terrible name. Sorry!)

patrick-schultz (changes requested):
Very nice

hail/src/main/scala/is/hail/expr/ir/Emit.scala
tpoterba: yeah, I like `multiMap` better. I didn't use the sequence implementation because I want to get rid of sequence -- it's much harder to understand with generic types, and also dangerous in that the `seq` thunks have side effects that might append to the wrong builder if used incorrectly.

Turns out, every instance of `sequence` is actually a `multiMapEmitCodes`!

patrick-schultz (approved):

PR #9980, opened 2021-02-03
[batch] eliminate expected exceptional log statement

danking:
If there is a true issue, we raise the exception which is caught and printed by
docker_call_retry, or, if that re-raises, it is stored as an error and serialized
back to the driver in Container.run

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -113,20 +113,18 @@ def docker_call_retry(timeout, name):
             except DockerError as e:
                 # 408 request timeout, 503 service unavailable
                 if e.status == 408 or e.status == 503:
-                    log.exception(f'in docker call to {f.__name__} for {name}, retrying', stack_info=True)
+                    log.warning(f'in docker call to {f.__name__} for {name}, retrying', stack_info=True, exc_info=True)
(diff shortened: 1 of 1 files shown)


jigold (changes requested):
I added all the log.exception messages in other operations below. We should get rid of those as well.

batch/batch/worker/worker.py
```diff
@@ -113,20 +113,18 @@ async def wrapper(f, *args, **kwargs):
...
                 # DockerError(500, 'Get https://gcr.io/v2/: dial tcp: lookup gcr.io: Temporary failure in name resolution')
                 elif e.status == 500 and ("request canceled while waiting for connection" in e.message
                                           or re.match("error creating overlay mount.*device or resource busy", e.message)
                                           or "Temporary failure in name resolution" in e.message):
-                    log.exception(f'in docker call to {f.__name__} for {name}, retrying', stack_info=True)
+                    log.warning(f'in docker call to {f.__name__} for {name}, retrying', stack_info=True, exc_info=True)
                 else:
                     raise
-            except asyncio.CancelledError:  # pylint: disable=try-except-raise
-                raise
```
danking, line 127: this isn't a subtype of any of the other caught errors, so there's no need to catch and raise

danking:
OK, I eliminated a bunch of `log.exception` that are either retried or re-raised. I completely eliminated some of them and in other cases I downgraded them to warning.

jigold (approved):

PR #9981, opened 2021-02-03
[query] Rework PCArray.constructFromFunctions to track index internally

tpoterba:
Simpler and more ergonomic!


tpoterba:
(also safer, it was totally possible to inadvertently segfault with the previous version, this one guards against that)

johnc1231 (approved):

johnc1231:
Conflicts to resolve @tpoterba

tpoterba:
oh, right, thanks!

PR #9987, opened 2021-02-04
[batch] fix billing projects tests

danking:
These billing projects tests did not correctly handle multiple attempts. Instead,
we use one billing project prefix for the entire batch (based on the token) and
each attempt gets a unique prefix. After all batch tests are complete, a job runs
which deletes all billing projects for the current batch.

### DIFF
--- a/batch/test/test_accounts.py
+++ b/batch/test/test_accounts.py
@@ -41,11 +41,11 @@ async def delete_all_test_billing_projects():
     bc = BatchClient(None, token_file=os.environ['HAIL_TEST_DEV_TOKEN_FILE'])
     try:
         for project in await bc.list_billing_projects():
-            if project.startswith(billing_project_prefix):
+            if project['name'].startswith(billing_project_prefix):
                 try:
-                    await bc.close_billing_project(project)
+                    await bc.close_billing_project(project['name'])
                 finally:
-                    await bc.delete_billing_project(project)
+                    await bc.delete_billing_project(project['name'])
     finally:
         await bc.close()
 
--- a/batch/test/test_accounts.py
+++ b/batch/test/test_accounts.py
@@ -41,11 +41,11 @@ async def delete_all_test_billing_projects():
     bc = BatchClient(None, token_file=os.environ['HAIL_TEST_DEV_TOKEN_FILE'])
     try:
         for project in await bc.list_billing_projects():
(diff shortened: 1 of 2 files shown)


build.yaml
```diff
@@ -2607,6 +2607,45 @@ steps:
...
+        valueFrom: default_ns.name
+      mountPath: /dev-tokens
+    - name: ssl-config-batch-tests
+      namespace:
+        valueFrom: default_ns.name
+      mountPath: /ssl-config
+   dependsOn:
+    - create_deploy_config
+    - create_accounts
+    - default_ns
```
jigold, line 2640: Should this be always_run?

batch/test/test_accounts.py
```diff
@@ -31,12 +32,27 @@ async def dev_client():
     await bc.close()
 
 
+billing_project_prefix = f'__testproject_{os.environ["HAIL_TOKEN"]}'
+
+async def delete_all_test_billing_projects():
+    bc = BatchClient(None, token_file=os.environ['HAIL_TEST_DEV_TOKEN_FILE'])
+    try:
+        for project in bc.list_billing_projects():
```
jigold, line 40: Are you missing an await here?

jigold (dismissed):

batch/test/test_accounts.py
```diff
@@ -31,12 +32,31 @@ async def dev_client():
...
+def get_billing_project_prefix():
+    return f'__testproject_{os.environ["HAIL_TOKEN"]}'
+
+
+async def delete_all_test_billing_projects():
+    billing_project_prefix = get_billing_project_prefix()
+    bc = BatchClient(None, token_file=os.environ['HAIL_TEST_DEV_TOKEN_FILE'])
+    try:
+        for project in await bc.list_billing_projects():
+            if project.startswith(billing_project_prefix):
```
jigold, line 44:
```
  File "<string>", line 4, in <module>
  File "/usr/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
  File "/io/test/test_accounts.py", line 44, in delete_all_test_billing_projects
    if project.startswith(billing_project_prefix):
AttributeError: 'dict' object has no attribute 'startswith'
```

jigold (changes requested):

jigold (approved):

PR #9989, opened 2021-02-04
[query] Use `loadFromNested` rather than matching on fundamental type

johnc1231:
This replaces any instances I know about where we were using `match` to decide whether we had to dereference a pointer. This makes it easier for me to add reference counted ndarrays, and it makes it easier for us to nix fundamental type altogether. 

I had to add `unstagedLoadFromNested` as well


tpoterba (approved):
Nice!

johnc1231:
Should i run benchmarks before merging this? Changing the way things are copied around seems potentially prone to performance regression.

johnc1231:
(Though I don't think I'm actually changing the implementations, I think I'm just dispatching everything through methods instead of matches)

tpoterba:
yeah, I have a very low prior on this changing performance. This doesn't change the staged code generation at all anyway, and most of what we care about is staged.

PR #9990, opened 2021-02-04
[ci] add source branch to PR batch attributes

daniel-goldstein:
Now shows the source branch in the batch and PR pages along with the other attributes.


danking (approved):
Why did we wait so long to do this. Love it.

PR #9991, opened 2021-02-04
[image-fetcher,site] clean up normal bash output from error logs

daniel-goldstein:
the gcloud command write to `stderr` which gets interpreted as errors in the logs and causes [a lot of noise](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aseverity%3DERROR%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.container_name%3D%22image-fetcher%22;timeRange=PT3H?project=hail-vdc&folder=true&organizationId=548622027621&query=%0A). 

Added `gcr.io` because without it `configure-docker` would configure all the registries it can find, which is unnecessary. Also removed `-x` from the site script since each line of that goes to error logs and none of it is helpful. If errors occur they will show up in some other form.

### DIFF
--- a/image-fetcher/fetch-image.sh
+++ b/image-fetcher/fetch-image.sh
@@ -1,10 +1,16 @@
 #!/bin/sh
 set -e
 
-gcloud -q auth activate-service-account \
-  --key-file=/secrets/gcr-pull.json
+gcloud_auth() {
+    gcloud -q auth activate-service-account --key-file=/secrets/gcr-pull.json
+    gcloud -q auth configure-docker gcr.io
+}
 
-gcloud -q auth configure-docker
+if ! gcloud_auth > gcloud-auth.log 2>&1; then
+    1>&2 cat gcloud-auth.log
+else
+    cat gcloud-auth.log
+fi
 
 DEFAULT_NAMESPACE=$(jq -r '.default_namespace' < /deploy-config/deploy-config.json)
 case $DEFAULT_NAMESPACE in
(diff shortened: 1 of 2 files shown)


image-fetcher/fetch-image.sh
```diff
@@ -1,10 +1,16 @@
...
 
-gcloud -q auth activate-service-account \
-  --key-file=/secrets/gcr-pull.json
+gcloud_auth() {
+    gcloud -q auth activate-service-account --key-file=/secrets/gcr-pull.json
+    gcloud -q auth configure-docker gcr.io
+}
 
-gcloud -q auth configure-docker
+if ! gcloud_auth > gcloud-auth.log 2>&1; then
```
jigold, line 9: My bash knowledge isn't as good as yours. Do you mind explaining this to me?

jigold (dismissed):

image-fetcher/fetch-image.sh
daniel-goldstein: `2>&1` redirects everything that `gcloud_auth` would send to its stderr (2) to its stdout (1), so that line means "send everything `gcloud_auth` would output to that log file". Then if anything fails, we can capture any and all output from `gcloud` and log it as an error.

I am realizing though that this is slightly incorrect. I need to include `&&` between the two gcloud commands as right now if the first one fails and the second passes `gcloud_auth` will pass. Will push a fix.

jigold (approved):

PR #9992, opened 2021-02-04
[query] Translate BGEN import to codebuilder and new constructors

tpoterba:
BGEN benchmarks before/after:
```
$ hb compare  0.2.61-ffa3169a657f-bgen-cb-before.json 0.2.61-126a8c8d5147-bgen-cb.json
    import_bgen_force_count_all      98.0%    144.232    141.289
       import_bgen_filter_count      97.9%    142.950    139.965
import_bgen_force_count_just_gp      96.1%    146.766    141.000
         import_bgen_info_score      95.0%    197.271    187.353
----------------------
Harmonic mean: 96.7%
Geometric mean: 96.7%
Arithmetic mean: 96.7%
Median:  97.0%
```

I think this is within noise of no change.

### DIFF
--- a/hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
+++ b/hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
@@ -484,6 +484,36 @@ final case class PCanonicalArray(elementType: PType, required: Boolean = false) 
   }
 
   // unsafe StagedArrayBuilder-like interface that gives caller control over adding elements and finishing
+  def constructFromNextAddress(cb: EmitCodeBuilder, region: Value[Region], length: Value[Int], deepCopy: Boolean):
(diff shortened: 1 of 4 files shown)


hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
```diff
@@ -483,6 +483,36 @@ final case class PCanonicalArray(elementType: PType, required: Boolean = false)
     new SIndexablePointerCode(SIndexablePointer(this), addr)
   }
 
+  // unsafe StagedArrayBuilder-like interface that gives caller control over adding elements and finishing
+  def constructFromNextAddress(cb: EmitCodeBuilder, region: Value[Region], length: Value[Int], deepCopy: Boolean):
```
tpoterba, line 487: this can go away when we have StackStruct.

chrisvittal (approved):

PR #9993, opened 2021-02-04
[ci] link to batch from job page

daniel-goldstein:


daniel-goldstein:
bump

jigold:
I think this is a temporary fix as we want Batch to directly return these pages. @CDiaz96 and I will start thinking about how this should be done in Batch more generally tomorrow.

daniel-goldstein:
Ok, happy to close if you're working on a solution.

jigold:
I think your solution is better right now.

ci/ci/templates/job.html
```diff
@@ -3,6 +3,9 @@
 {% block content %}
     <h1>Batch {{ batch_id }} Job {{ job_id }}</h1>
 
+    <form method="GET" action="{{ base_path }}/batches/{{ batch_id }}">
+      <button>Back to batch</button>
+    </form>
```
jigold, line 8: Sorry to nitpick. Can we get another new line after the form?

jigold (changes requested):

jigold (approved):

PR #9996, opened 2021-02-05
[gateway] redirect to auth on 401 in internal

daniel-goldstein:
Previously, unauthenticated developers would get a 401 nginx page when hitting an internal url. This now sends them through the auth flow.

In the spirit of deploying before merge, you can try this out!


danking (approved):
Every day it's a gettin' better 🎶

PR #9997, opened 2021-02-05
[query] No SRVB in GroupByKey

patrick-schultz:


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -960,6 +968,119 @@ class Emit[C](
...
+                {
+                  cb.assign(groupSize, groupSize + 1)
+                }, {
+                  cb += groupSizes.add(groupSize)
+                  cb.assign(groupSize, 1)
+                })
+              cb.assign(eltIdx, eltIdx + 1)
+            })
+            cb += groupSizes.add(groupSize)
+            cb.println(groupSizes.size.toS)
```
tpoterba, line 1049: println

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -960,6 +968,119 @@ class Emit[C](
...
+        val outerSize = mb.newLocal[Int]("groupByKey_outerSize")
+        val groupSize = mb.newLocal[Int]("groupByKey_groupSize")
+
+        val outerRegion = region.asParent(collectionTyp.separateRegions, "GroupByKey")
+        emitStream(collection, outerRegion).map(cb) { stream =>
+          cb += EmitStream.write(mb, stream.asStream, sortedElts, outerRegion)
+          cb += sorter.sort(sortF)
+          cb += sorter.pruneMissing
+          cb.ifx(sortedElts.size.ceq(0), {
+            cb.assign(xOut, arrayTyp.constructEmpty(cb, region.code))
```
tpoterba, line 1032: can we take this out? We shouldn't need to special-case 0 here.

hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
```diff
@@ -444,6 +444,12 @@ final case class PCanonicalArray(elementType: PType, required: Boolean = false)
   private def deepRenameArray(t: TArray): PArray =
     PCanonicalArray(this.elementType.deepRename(t.elementType), this.required)
 
+  def constructEmpty(cb: EmitCodeBuilder, region: Value[Region]): SIndexablePointerCode = {
```
tpoterba, line 447: I don't think there needs to be a separate function for this (and I want to remove the usage in groupByKey above)

tpoterba (changes requested):

hail/src/main/scala/is/hail/expr/ir/Emit.scala
patrick-schultz: I don't think the other branch handles the empty case as is, but I'm sure it's possible to tweak it so that it does. I will try.

tpoterba: if, not can just call constructFromFunctions with 0 and finish

tpoterba (approved):

PR #9999, opened 2021-02-05
[query] Add independent inference pass for stream memory management

tpoterba:
Currently, this is done in parallel with the existing memory management
tracking (in ptypes), and we assert that the two produce the same
result. This can give us confidence that removing InferPType will not
change semantics.


tpoterba:
bump

patrick-schultz:
Sorry, for some reason I was thinking this was held up on something. Reviewing now.

tpoterba:
no worries, it wasn't blocking anything.

patrick-schultz (approved):

PR #10001, opened 2021-02-05
[query] Reference Counting NDArrays

johnc1231:
This PR introduces reference counted ndarrays. 

One consequence of this is that like arrays, ndarrays will now store a pointer inside of tuples / structs that points to their location. Many of the changes in this PR are just to support that change. 

Unlike arrays, ndarrays will live outside of regions, allowing them to be moved between regions without copies by editing the reference count associated with them. This will allow us to  spend significantly less time copying in operations like those generated by `hl.loop`.  

When we allocate an ndarray, we store a 16 byte header in front of the ndarray. The first 8 bytes is a long storing the reference count. The second 8 bytes is the total number of bytes (minus the header) stored in this ndarray.


tpoterba:
woah, this is a small PR for this change! Nice!

johnc1231:
Right now this passes all tests, but there are some `???`s in there that just never get triggered. Going to try and write tests that hit them before coming up with an implementation.

johnc1231:
Basically done, I just have to go through and rewrite the LAPACK calls to use the new interface (so far I've only done so with NDArrayInverse). If you have any thoughts on the actual reference counting  @tpoterba, that can be reviewed.

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
```diff
@@ -191,43 +326,66 @@ final case class PCanonicalNDArray(elementType: PType, nDims: Int, required: Boo
...
+
+        val currentOffset = cb.newLocal[Long]("pcanonical_ndarray_store_offset", targetDataFirstElementAddr)
+        SNDArray.forEachIndex(cb, shape, "PCanonicalNDArray_store") { (cb, currentIndices) =>
+          val oldElement = oldND.loadElement(currentIndices, cb)
+          elementType.storeAtAddress(cb, currentOffset, region, oldElement, true)
+          cb.assign(currentOffset, currentOffset + elementType.byteSize)
+        }
+
+        finish(cb).a
+      case _ =>
```
tpoterba, line 364: you can just leave this case out, and when we add something out we'll get a better error (match error with the thing that failed to match)

hail/src/main/scala/is/hail/types/physical/PCanonicalStruct.scala
```diff
@@ -99,6 +99,10 @@ final case class PCanonicalStruct(fields: IndexedSeq[PField], required: Boolean
   def loadField(offset: Code[Long], fieldName: String): Code[Long] =
     loadField(offset, fieldIdx(fieldName))
 
+  def loadField(offset: Long, fieldName: String): Long = {
```
tpoterba, line 102: don't add this -- it's pretty expensive to do a string lookup to do pointer arithmetic.

tpoterba (changes requested):
A couple minor things. Looks great otherwise.

hail/src/main/scala/is/hail/annotations/RegionPool.scala
```diff
@@ -136,6 +136,12 @@ final class RegionPool private(strictMemoryCheck: Boolean, threadName: String, t
   def scopedSmallRegion[T](f: Region => T): T = using(Region(Region.SMALL, pool=this))(f)
   def scopedTinyRegion[T](f: Region => T): T = using(Region(Region.TINY, pool=this))(f)
 
+  def allocateNDArray(nBytes: Long): Long = {
```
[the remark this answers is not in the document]
johnc1231: Yep, good point

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -1212,18 +1213,18 @@ class Emit[C](
...
           cb.assign(An, (M * N).toI)
-          cb.assign(Aaddr, ndPT.dataType.allocate(region.code, An))
-          cb.append(ndPT.dataType.stagedInitialize(Aaddr, An))
-          cb.append(Region.copyFrom(dataFirstAddress,
-            ndPT.dataType.firstElementOffset(Aaddr, An), An.toL * 8L))
 
           cb.assign(IPIVaddr, IPIVptype.allocate(region.code, N.toI))
           cb.append(IPIVptype.stagedInitialize(IPIVaddr, N.toI))
 
+          val (aAadrFirstElement, finish) = ndPT.constructDataFunction(shapeArray, stridesArray, cb, region.code)
```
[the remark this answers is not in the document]
johnc1231: I actually intended this to be "a addr first element", like the matrix is named a (that's the lapack argument name)

johnc1231:
Ok, so this is all working now. The one thing that's not ideal is that we are losing copy free transposes right now, as I don't know how to have two ndarrays share the same data and keep the reference counts correct. In some sense I want the ndarray's data to be reference counted, not the stride / shape bits which are cheap to copy.

johnc1231:
Benchmarks on this are not good, still investigating.

johnc1231:
Fixing MakeNDArray to not double allocate cut my laptop benchmark time by more than 50%, it's only a bit slower than main now (~5-10%), so we should probably try and get it in. Feel free to review again when you get a chance.

tpoterba:
wow!

hail/src/main/scala/is/hail/types/physical/PCanonicalStruct.scala
tpoterba, line 102: can we still get rid of this?

tpoterba (dismissed):

hail/src/main/scala/is/hail/types/physical/PCanonicalStruct.scala
johnc1231: Oh sorry i thought I did, will do now

hail/src/test/scala/is/hail/types/physical/PNDArraySuite.scala
```diff
@@ -20,4 +20,116 @@ class PNDArraySuite extends PhysicalTestUtils {
...
+
+        // Region 2 gets an ndarray at ndaddress2, plus a reference to the one at ndarray 1.
+        val (_, snd2Finisher) = nd.constructDataFunction(shapeSeq, shapeSeq, cb, codeRegion2)
+        val snd2 = snd2Finisher(cb).memoize(cb, "snd2")
+        cb.assign(r2PointerToNDAddress1, codeRegion2.allocate(8L, 8L))
+
+        nd.storeAtAddress(cb, r2PointerToNDAddress1, codeRegion2, snd1, true)
+
+        // Return the address of the 1st one
+        snd1.tcode[Long]
```
tpoterba, line 54: no tcode please! use a `store` method or something

tpoterba (changes requested):

hail/src/test/scala/is/hail/types/physical/PNDArraySuite.scala
johnc1231: Fixed!

tpoterba (approved):
:shipit:

PR #10002, opened 2021-02-05
[query] Fix egregious VEP memory leak

tpoterba:


tpoterba:
nevermind, this is fine.

PR #10009, opened 2021-02-08
[query] Begin simplifying EmitCode constructors.

tpoterba:
This is the first step toward top-level requiredness being an attribute
of IEmitCode rather than PType.

* Take cb/mb when necessary.
* Remove unnecessary constructors

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/EmitCodeBuilder.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/EmitCodeBuilder.scala
@@ -139,11 +139,11 @@ class EmitCodeBuilder(val emb: EmitMethodBuilder[_], var code: Code[Unit]) exten
     _invoke[T](callee, args: _*)
   }
 
-  def invokeEmit(callee: EmitMethodBuilder[_], args: Param*): EmitCode = {
+  def invokeEmit(cb: EmitCodeBuilder, callee: EmitMethodBuilder[_], args: Param*): IEmitCode = {
     val pt = callee.emitReturnType.asInstanceOf[EmitParamType].pt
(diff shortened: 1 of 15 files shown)


hail/src/main/scala/is/hail/expr/ir/EmitCodeBuilder.scala
```diff
@@ -139,11 +139,11 @@ class EmitCodeBuilder(val emb: EmitMethodBuilder[_], var code: Code[Unit]) exten
...
   }
 
-  def invokeEmit(callee: EmitMethodBuilder[_], args: Param*): EmitCode = {
+  def invokeEmit(cb: EmitCodeBuilder, callee: EmitMethodBuilder[_], args: Param*): IEmitCode = {
     val pt = callee.emitReturnType.asInstanceOf[EmitParamType].pt
     val r = newLocal("invokeEmit_r")(pt.codeReturnType())
-    EmitCode(r := _invoke(callee, args: _*),
-      EmitCode.fromCodeTuple(pt, Code.loadTuple(callee.modb, EmitCode.codeTupleTypes(pt), r)))
+    cb.assignAny(r, _invoke(callee, args: _*))
+    IEmitCode.fromCodeTuple(cb, pt, Code.loadTuple(callee.modb, EmitCode.codeTupleTypes(pt), r))
```
chrisvittal, line 146: Just use `this`.

chrisvittal (changes requested):

hail/src/main/scala/is/hail/expr/ir/EmitCodeBuilder.scala
tpoterba: oops, thanks!

chrisvittal (approved):

PR #10013, opened 2021-02-08
[grafana] Add grafana service for metrics monitoring

daniel-goldstein:
Deploying grafana in our GKE cluster gives us instant and easy access to the stackdriver backend with the same querying capabilities of our current front-end, but without the clutter and insanely slow load times. See [here](https://internal.hail.is/dgoldste/grafana/d/TVkleyLMk/detailed-service-resource-utilization?orgId=1) for some example dashboards I set up to look at resources across our services (credentials are the default admin/admin).

This alleviates the immediate pain of using the console (for metrics only, not logging), but my longer aim is that getting more regular use out of our metrics can reveal deeper pain points of our monitoring stack and if/where we need to eat up more responsibility from google monitoring.

This is a StatefulSet, so configuration through the UI will persist and is done manually. If we find that our dashboards are stable and boilerplate enough, I'd like to move to a code-based dashboard configuration.

Sadly, `check-yaml` does not appreciate our jinja templating in yaml, so I've removed it for now.

cc: @cseed

### DIFF
--- a/grafana/deployment.yaml
+++ b/grafana/deployment.yaml
@@ -27,7 +27,7 @@ spec:
             name: grafana-config
       containers:
        - name: grafana
-         image: grafana/grafana:latest
+         image: grafana/grafana:7.3.7
          volumeMounts:
           - mountPath: /var/lib/grafana
             name: grafana-storage
--- a/grafana/deployment.yaml
+++ b/grafana/deployment.yaml
@@ -1,0 +1,68 @@
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: grafana
+  labels:
+    app: grafana
+spec:
+  selector:
+    matchLabels:
+      app: grafana
+  serviceName: "grafana"
+  replicas: 1
+  template:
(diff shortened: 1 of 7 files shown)


danking:
this is very cool

grafana/deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+        fsGroup: 472
+        runAsNonRoot: true
+        runAsUser: 472
+      volumes:
+        - name: grafana-configmap-volume
+          configMap:
+            name: grafana-config
+      containers:
+       - name: grafana
+         image: grafana/grafana:latest
```
danking, line 30: Pick and pin a version, latest version seems fine to me. We prefer pinned version to latest because it gives us reproducibility (modulo people overwriting docker tags). A separate SRE thing we need to address is systematically updating versions.

danking (changes requested):

daniel-goldstein:
bump

danking (approved):

PR #10014, opened 2021-02-08
[batch] Fixed next page button

CDiaz96:
-When there were exactly 50 batches it would show a next page button 
that displayed an empty page.

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -525,8 +525,8 @@ LIMIT 51;
                in db.select_and_fetchall(sql, sql_args)]
 
     if len(batches) == 51:
-        last_batch_id = batches[-2]['id']
-        batches = batches[:50]
+        batches.pop()
+        last_batch_id = batches[-1]['id']
     else:
         last_batch_id = None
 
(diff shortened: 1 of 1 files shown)


batch/batch/front_end/front_end.py
```diff
@@ -516,16 +516,17 @@ async def _query_batches(request, user):
...
 
     batches = [batch_record_to_dict(batch)
                async for batch
                in db.select_and_fetchall(sql, sql_args)]
 
-    if len(batches) == 50:
-        last_batch_id = batches[-1]['id']
+    if len(batches) == 51:
+        last_batch_id = batches[-2]['id']
+        batches = batches[:50]
```
jigold, line 529: This is fine, but I just wanted to let you know that you can drop the last element in Python like this. Your call on what you like better. With the last option, you can write what you're doing a little more clearly:

```
last_batch_id = batches[-2]['id']
batches = batches[:-1:]
```

or

```
batches.pop()
last_batch_id = batches[-1]['id']
```

jigold (changes requested):
See comment.

batch/batch/front_end/front_end.py
jigold: I think I prefer the last option for its clarity.

jigold (approved):

PR #10015, opened 2021-02-08
[k8s] Reduce resource requests

daniel-goldstein:
This is still quite generous, based on activity from the past week in Grafana, but the goal is to get the utilization to rise to "interesting" levels. Since this doesn't touch limits, I'm tempted to just set requests to the minimum and correct if necessary, but am open to opinions. I also didn't touch `memory` or `query` since those requests seemed more intentional.


danking (approved):
Begin the experiment!

PR #10016, opened 2021-02-08
[ci] add batch timing to ci batch page

daniel-goldstein:


ci/ci/templates/batch.html
```diff
@@ -7,8 +7,14 @@ <h1>Batch {{ batch['id'] }}</h1>
       {% for name, value in batch['attributes'].items() %}
       <div>{{ name }}: {{ value }}</div>
       {% endfor %}
+      <br>
+      {% for field in ['time_created', 'time_completed', 'duration'] %}
+      {% if field in batch %}
+      <div>{{ field }}: {{ batch[field] }}</div>
```
jigold, line 13: I think this isn't quite right. I think we should have a separate block that isn't part of the attributes. Maybe give a header for attributes and then a header for the next section with the timings. Maybe call it Timings or Metadata?

If you want to combine them, then I think you need to fix the overall `{% if 'attributes' in batch %}` to not be there because we always want time_created, etc. to show up regardless of whether there are attributes.

jigold (dismissed):

ci/ci/templates/batch.html
daniel-goldstein: Good point, I've moved them under a new `Timing` heading. I still used the `attributes` css class, because it seemed like it's used in a lot of places just to get indentation, but it is perhaps a little weird.

ci/ci/templates/batch.html
```diff
@@ -3,18 +3,21 @@
...
     {% if 'attributes' in batch %}
+    <h2>Attributes</h2>
     <div class="attributes">
       {% for name, value in batch['attributes'].items() %}
       <div>{{ name }}: {{ value }}</div>
       {% endfor %}
-      <br>
+    </div>
+    {% endif %}
+    <h2>Timing</h2>
```
jigold, line 13: Sorry to be picky. I think the Timing block should come before the Attributes because it's not optional.

jigold (dismissed):

ci/ci/templates/batch.html
daniel-goldstein: No problem, I have no preference so I'll flip them.

jigold (approved):

PR #10017, opened 2021-02-09
[auth] bump PyPI versions of google auth libs

danking:
There is a new version of google-cloud-core which is incompatible with google-auth 1.12.1. Rather than
pin to an old version of google-cloud-core, I bump us to the latest of both auth libraries.


danking:
example failure: https://ci.hail.is/batches/179008/jobs/30

daniel-goldstein (approved):

PR #10018, opened 2021-02-09
[query] Convert TextMatrixReader to CB and new PType constructors

tpoterba:


hail/src/main/scala/is/hail/io/TextMatrixReader.scala
```diff
@@ -438,225 +437,233 @@ class CompiledLineParser(
...
+        cb.ifx(line.invoke[Int, String, Int, Int, Boolean]("regionMatches",
+          pos, missingValue, 0, missingValue.size),
+          {
+            cb.assign(pos, end)
+            cb.goto(Lfinished)
+          },
+          cb.goto(Ldefined)
+        ),
+        cb.goto(Ldefined)),
+      cb.goto(Ldefined))
```
chrisvittal, line 532: Why is the Ldefined lable needed? doesn't control just fall through?

hail/src/main/scala/is/hail/io/TextMatrixReader.scala
```diff
@@ -438,225 +437,233 @@ class CompiledLineParser(
...
+        cb.ifx(line.invoke[Int, String, Int, Int, Boolean]("regionMatches",
+          pos, missingValue, 0, missingValue.size),
+          {
+            cb.assign(pos, end)
+            cb.goto(Lmissing)
+          },
+          cb.goto(Ldefined1)
+        ),
+        cb.goto(Ldefined1)),
+      cb.goto(Ldefined1))
```
chrisvittal, line 504: Again, doesn't control just fall through if any of the ifs are false?

chrisvittal (changes requested):

hail/src/main/scala/is/hail/io/TextMatrixReader.scala
tpoterba: much nicer, thanks!

chrisvittal (approved):

PR #10019, opened 2021-02-09
[query] Push cb/ptype constructors through LocusFunctions

tpoterba:


tpoterba:
These changes include a performance regression - instead of allocating memory once and filling in each member of the nested value (srvb) we are using struct constructors that allocate e.g. the locus an extra time and copy it into the container.

I do not think it's worth creating constructors right now that prevent this regression -- the region value construction is much slower than the java calls here, and the right solution is coming down teh pike -- constructing containers using SStackStruct emit codes will have exactly the semantics we want.

hail/src/main/scala/is/hail/expr/ir/functions/LocusFunctions.scala
```diff
@@ -342,132 +246,205 @@ object LocusFunctions extends RegistryFunctions {
...
+              locusClass, "parseInterval",
+              locusStr.asString.loadString(),
+              rgCode(cb.emb, plocus.rg),
+              invalidMissing.asBoolean.boolCode(cb)))
+
+          cb.ifx(interval.isNull, cb.goto(Lmissing))
+
+          val intervalCode = emitLocusInterval(cb, r, interval, rt)
+          cb.goto(Ldefined)
+          IEmitCode(Lmissing, Ldefined, intervalCode)
```
patrick-schultz, line 295: A slightly simpler way to do this:
```
IEmitCode(cb, interval.isNull, interval)
  .map(cb)(emitLocusInterval(cb, r, _, rt))
```

There's several other places you could use the same pattern, but up to you which style you find clearer.

patrick-schultz (changes requested):
Looks good, just one stylistic suggestion.

hail/src/main/scala/is/hail/expr/ir/functions/LocusFunctions.scala
tpoterba: I want to push toward removing the generic type on IEmitCode in favor of SCode as soon as we can. I guess we could do something like:
```
IEmitCode(cb, interval.isNull, {
... stuff here ... })
```

Which would do the same. I'm probably not going to go back and change it though, if it's OK with you

patrick-schultz: That's fine, I'll approve.

patrick-schultz (approved):

PR #10020, opened 2021-02-09
[dev-docs] Hail For New Engineers

danking:
This doc introduces the entire Hail project to new software engineers. This is a response to
Carolin's and Daniel's critiques of our on-boarding documentation.

I think there's a lot more to say, but I think it's time for some feedback before I continue.

The other things I want to document:
- How we use git? Including practical examples.
- What does genetic data look like?
- What is a GWAS?
- What is PCA as applied to genetics?
- A Tour of Hail Query (in the spirit of Soustroup's a Tour of C++).
- A Bibliography of Hail Team's recommended books.

### DIFF
--- a/dev-docs/hail-for-new-engineers.md
+++ b/dev-docs/hail-for-new-engineers.md
@@ -80,9 +80,9 @@ aggregates, the resulting aggregated value is converted to a Python value and re
 ### Hail Batch
 
 Hail Batch is a system for executing arbitrary Linux programs. Each invocation of a program is
-called a "job". Zero or more jobs compose a Batch. Moreover, jobs may depend on the files written by
-other jobs in the same Batch. The job dependencies are allowed to form any [directed, acyclic graph
-(DAG)](https://en.wikipedia.org/wiki/Directed_acyclic_graph).
+called a "job". Zero or more jobs comprise a Batch. Moreover, jobs may depend on the files written
+by other jobs in the same Batch. The job dependencies are allowed to form any [directed, acyclic
+graph (DAG)](https://en.wikipedia.org/wiki/Directed_acyclic_graph).
 
 Users create batches and jobs using a Python library: `hailtop.batch`.
 
--- a/dev-docs/hail-for-new-engineers.md
+++ b/dev-docs/hail-for-new-engineers.md
@@ -25,7 +25,7 @@ In March of 2017, Hail team began work on a compiler.
 In October of 2018, the Hail Python interface was modified, in a backwards-incompatible way. This
 new Python interface was named "Hail 0.2". The old Python interface was retroactively named "Hail
 0.1". Hail team began deploying a Python package named `hail` to [PyPI](https://pypi.org). The Hail
-python package version compiles with [Semantic Versioning](https://semver.org).
+python package version complies with [Semantic Versioning](https://semver.org).
 
 Meanwhile, in September of 2018, Hail team began work on a system called "Batch". Batch runs
 programs in parallel on a cluster of virtual machines. Also in September, Hail team began work on a
@@ -34,9 +34,9 @@ request (PR) into the `main` branch of [`hail-is/hail`](https://github.com/hail-
 automatically merges into main any pull request that both passes the tests and has at least one
 "approving" review and no "changes requested" reviews. CI uses Hail Batch to run the tests.
 
-Around this time, the Hail team split organized itself into two sub-teams: "compilers" team and
-"services" team. The compilers team is responsible for the Hail Python library, the compiler, and
-the associated runtime. The compilers team code is entirely contained in the `hail` directory of the
+Around this time, the Hail team organized itself into two sub-teams: "compilers" team and "services"
+team. The compilers team is responsible for the Hail Python library, the compiler, and the
+associated runtime. The compilers team code is entirely contained in the `hail` directory of the
 `hail-is/hail` repository. The services team is responsible for Batch, CI, and the software
 infrastructure that supports those services. Each service has its own directory in the hail
 repository. More information about the structure of the repository can be found in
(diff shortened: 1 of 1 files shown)


dev-docs/hail-for-new-engineers.md
```diff
@@ -0,0 +1,189 @@
...
+In November of 2016, the Hail command-line interface was eliminated and a Python interface was
+introduced. During this time, Hail was not versioned. Users had to build and use Hail directly from
+the source code repository.
+
+In March of 2017, Hail team began work on a compiler.
+
+In October of 2018, the Hail Python interface was modified, in a backwards-incompatible way. This
+new Python interface was named "Hail 0.2". The old Python interface was retroactively named "Hail
+0.1". Hail team began deploying a Python package named `hail` to [PyPI](https://pypi.org). The Hail
+python package version compiles with [Semantic Versioning](https://semver.org).
```
tpoterba, line 28: complies

dev-docs/hail-for-new-engineers.md
```diff
@@ -0,0 +1,189 @@
...
+
+Meanwhile, in September of 2018, Hail team began work on a system called "Batch". Batch runs
+programs in parallel on a cluster of virtual machines. Also in September, Hail team began work on a
+system called "CI". CI is an acronym for Continuous Integration. CI runs the tests for every pull
+request (PR) into the `main` branch of [`hail-is/hail`](https://github.com/hail-is/hail). CI
+automatically merges into main any pull request that both passes the tests and has at least one
+"approving" review and no "changes requested" reviews. CI uses Hail Batch to run the tests.
+
+Around this time, the Hail team split organized itself into two sub-teams: "compilers" team and
```
tpoterba, line 37: "split organized" => one or the other?

dev-docs/hail-for-new-engineers.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Hail Query
+
+Hail Query is a Python library for the analysis of large datasets. In Hail Query, a dataset is
+represented as Table or a Matrix Table.
+
+Hail Tables are similar to SQL tables, Pandas Dataframes, Excel spreadsheets, and CSV files.
+
+Hail Matrix Tables do not have analogues in most other systems. Perhaps the only analogue is a
+[SciDB](https://dbdb.io/db/scidb) multi-dimensional array. A Hail Matrix Table can represent dense,
```
johnc1231, line 69: Is SciDB the same as TileDB? I think they have something analogous.

johnc1231:
I like this!

Responding to your bullet points:

- I don't really think we need git examples? We don't do anything crazy or abnormal with it, seems likely to just clutter things. 
- I think "A Tour of Hail Query" would be a nice thing to put on the website instead of just in the git repo.
- Everything else I agree with.

daniel-goldstein:
This is very nice! I didn't know some of things about the 0.1 days. I agree with John's comments, though I think it's worth making explicit whether a given Tour of Hail Query is for Scientists or Software Engineers. I think you're proposing the latter? Also, a stylistic nitpick but I find bullet-delimited sentences quite difficult to read, and think it might be clearer to drop the ands and commas at the end of the bullets.

daniel-goldstein:
> What is a GWAS?

This is probably one of the most important things you could add, targeted at a swe audience. I think explaining a couple other big ticket items that span Query's capabilities (like PC Relate) would be very useful as well. I would recommend "What is a GWAS? _and why is Hail pretty good at it_?"

dev-docs/hail-for-new-engineers.md
danking: The history is confusing and not entirely clear to me. I believe Stonebreaker (of SciDB fame) worked with Papadopoulos at one point. I believe TileDB is at the least inspired by SciDB. TileDB, unlike SciDB, is a file format with a library to read and write it. There is also GenomicsDB which is built on top of or forks TileDB. I'll add some info.

danking:
I intend the Tour of Hail Query to assume no genetics knowledge. Indeed, it probably won't ever mention genetics.

I removed the sentence-as-bulleted list. I like them, but I'm not the reader here 😉.

I intend for each of these documents to be separate files, and people with git experience can skip the git doc. I agree that we use git in one of the typical ways. However, I've learned that Atom's GitHub support doesn't consider the possibility of PRs from one remote to another. We also have some conventions around git messages and stacked PRs that I think are worth getting in writing.

danking:
I forgot to say: I actually agree with John that a tour of hail query would be nice to include in the website. Hail is way more broadly useful than we admit. It's worth starting to champion its use outside of genetics.

johnc1231:
Even if it takes a while for users outside of genetics to care, a document like that would be helpful for new users who are more like data engineers than geneticists

dev-docs/hail-for-new-engineers.md
```diff
@@ -0,0 +1,189 @@
...
+accessible at https://internal.hail.is/DEVELOPER_USERNAME/SERVICE/ . Unlike the default namespace,
+every other namespace has exactly one database (containing all tables from each service's database).
+
+The default namespace includes three services that are hand-deployed and do not exist in other
+namespaces: gateway, internal-gateway, and router-resolver. All traffic from the Internet enters the
+cluster through the gateway which routes traffic to the appropriate namespace and service. All
+traffic from the Batch workers enters through the internal-gateway. The router-resolver assists the
+gateway in routing.
+
+## Genetics, Briefly
```
CDiaz96, line 150: I think the brief overview of genetics should be at the top of the document. It should transition from big picture to more specifics, especially as someone joining the team who might not have any previous knowledge about genetics or hail. Then there should be a transition between the overview of genetics and the role that hail plays in everything.

dev-docs/hail-for-new-engineers.md
```diff
@@ -0,0 +1,189 @@
...
+In March of 2017, Hail team began work on a compiler.
+
+In October of 2018, the Hail Python interface was modified, in a backwards-incompatible way. This
+new Python interface was named "Hail 0.2". The old Python interface was retroactively named "Hail
+0.1". Hail team began deploying a Python package named `hail` to [PyPI](https://pypi.org). The Hail
+python package version compiles with [Semantic Versioning](https://semver.org).
+
+Meanwhile, in September of 2018, Hail team began work on a system called "Batch". Batch runs
+programs in parallel on a cluster of virtual machines. Also in September, Hail team began work on a
+system called "CI". CI is an acronym for Continuous Integration. CI runs the tests for every pull
```
CDiaz96, line 32: "CI (continuous integration)" rather than "CI is an acronym for Continuous Integration."

dev-docs/hail-for-new-engineers.md
```diff
@@ -0,0 +1,189 @@
...
+repository. More information about the structure of the repository can be found in
+[`hail-overview.md`](hail-overview.md).
+
+Beginning in early 2020, beta users were given access to Hail Batch.
+
+In April of 2020, the Hail team began referring to the Hail python library as "Hail Query" and
+created a system called "the Query Service". Inside the Hail Query Service, Hail Query uses Hail
+Batch to run an analysis across many computer cores instead of using Apache Spark.
+
+## Project Goals
```
CDiaz96, line 51: I think this block should be up top integrated with the hail history of why it was originally created / its purpose.

danking:
@CDiaz96 @daniel-goldstein @tpoterba @johnc1231  I've updated the document in response to y'all's helpful critique. Please review and approve if you're individually OK with it! Once I have four OKs, I'll remove the WIP tag.

tpoterba (approved):

daniel-goldstein (approved):

johnc1231 (approved):

PR #10026, opened 2021-02-09
[cpp] work in progress

cseed:
I think this is a good stopping point to push a version.  A basic skeleton of the compiler infrastructure is in place, and I can jit compile and execute expressions involving na, isna, arguments and (scalar) literals, and can return the result to C++ and print it out.

For the same reason we like async in services, so we can overlap IO with compute and other IO, we want the code generated by the compiler to be asynchronous.  I'm going to work on some async/runtime issues next before returning to the code generator.


libhail/docs/aggregators.txt
```diff
@@ -0,0 +1,174 @@
...
+  class ResultOp : public IR {
+    SeqOp(IR *aggstate);
+  };
+
+I think of RunAgg as creating (and then cleaning up) a scope of
+aggstates, and managing the aggregation process by running the body
+and result IR.  The signatures are the types of the aggstates bound in
+the scope.
+
+This perspective can be generalized to TableAggregate, allowing us to
```
tpoterba, line 141: This is a point I want to discuss further. ApplyAggOp has some nice features -- it makes dead code elimination dead simple with respect to aggregators. It's much harder to optimize void-typed IRs, so I want to make sure we have a plan to retain the control we have now.

tpoterba (approved):

PR #10027, opened 2021-02-10
[docs] internal

danking:
I don't think we should merge this. This PR got a bit out of hand and then died when I changed focus to the query service.

There are two distinct changes that I had hoped to unify: nicely rendered dev-docs *and* generated docs for gear, web_common, & hailtop. I got stuck after getting each one rendering OK but not integrated with one another. I also never got to dynamic rendering of the header (i.e. logged in users see batch > batches, etc.).

The first thing you should check out are the rendered library docs:
```
(cd docs && make html && (cd build/html && python3 -m http.server))
# now navigate to http://localhost:8000/
```
You'll notice the Hail CSS is missing a bunch of styles to make functions render nicely. Take a look at the generated HTML. Sphinx includes a few style tags that we should probably define. I also fixed a few docs issues. There are many more broken references to fix.

Screen Shot 2021-02-09 at 11 07 34 PM
Screen Shot 2021-02-09 at 11 07 50 PM

Now you should check out the rendered dev docs:
```
(cd site && make render && cd docs && python3 -m http.server)
```
Screen Shot 2021-02-09 at 11 11 07 PM
Screen Shot 2021-02-09 at 11 11 16 PM

---

What's left to do?

1. Make the header dynamic (i.e. logged-in users see their name, etc.)?
2. Move the dev-docs and the python library docs into one location.
3. Finish modifying `site` so that it hosts two servers: `hail.is` and `docs.hail.is`. `docs.hail.is` displays some landing page from which we can navigate to dev-docs or python library docs.


danking:
I think I also accidentally deleted a bunch of stuff in `site` during the rebase.

cseed:
Thanks, @danking!  I'm going to take this over and will revise/resubmit.

PR #10032, opened 2021-02-10
[query] Document deprecation of ArrayExpression.head

nawatts:
Noticed that ArrayExpression.head was not documented as deprecated when it was deprecated in #9482.

This also fixes a rendering error with one of its examples.

https://hail.is/docs/0.2/hail.expr.ArrayExpression.html#hail.expr.ArrayExpression.head


nawatts:
Scorecard returned @catoverdrive

tpoterba (approved):

PR #10036, opened 2021-02-11
[batch] Fix bug in schedule_job

jigold:
We had a bug yesterday where the return value for `delta_cores_mcpu` when scheduling a job was None instead of an integer. This messed up the instance collection data structures for that instance so we couldn't remove the instance and we couldn't handle the deactivate or delete events. Also, the instance was stuck with -1 free cores.

I think until we figure out why this happened, this is a perfectly good solution. If the job was actually scheduled correctly, then MJS will happen and the free cores will be correct.

### DIFF
--- a/batch/batch/driver/job.py
+++ b/batch/batch/driver/job.py
@@ -27,6 +27,10 @@ if TYPE_CHECKING:
 log = logging.getLogger('job')
 
 
+class BadSQLReturnValue(Exception):
+    pass
+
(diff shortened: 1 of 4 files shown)


danking (dismissed):
See comments in Zulip. I think we need to modify `schedule_job` to properly handle the case where the `select into` finds no rows.

danking (approved):

PR #10037, opened 2021-02-11
[ci] add user page with PRs and dev deploys

daniel-goldstein:
- Makes a user page at `/me` showing your PRs and your most recent (10) dev deploys. I added a `dev_deploy: 1` attribute to dev deploy batches to identify them, as seemed consistent with `test` and `deploy` attributes, though it might make more sense to change these to `scope` in the future? Either way this will only select dev deploys going forward but that seems like a non-issue. Also needed a hail username to gh username mapping, which is annoying, but deploys are hail username and PRs are github username and I didn't see an existing mapping between the two.

- Extracts the pr, job, and newly added dev deploy tables into macros for reuse across pages. As a result removed the duplicate code for the job table between PR and batch pages.


daniel-goldstein:
@danking bump. I also probably should not have snuck in this last commit and instead done a fresh PR so that Vlad and team didn't have to spend time debugging. Lmk if I should remove it.

ci/ci/ci.py
```diff
@@ -207,6 +208,28 @@ async def get_job(request, userdata):
...
 
 
+@routes.get('/me')
+@web_authenticated_developers_only()
+async def get_user(request, userdata):
+    username = userdata['username']
+    if username in dev_to_gh_user:
+        gh_username = dev_to_gh_user[username]
+    else:
+        return web.Response(text="Please submit a PR to CI with your github username to see your user page!")
```
danking, line 218: displaying an empty list of dev deploys seems simpler. We should include the GitHub username in the page context, that way folks know if they're not in the mapping or if the mapping is wrong.

ci/ci/ci.py
```diff
@@ -41,51 +42,51 @@
     for (index, [bss, deployable]) in enumerate(json.loads(os.environ.get('HAIL_WATCHED_BRANCHES', '[]')))
 ]
 
+dev_to_gh_user = {'dgoldste': 'daniel-goldstein'}
```
danking, line 45: Add information to AUTHORIZED_USERS or drive AUTHORIZED_USERS from some new structure that maps between these.

danking (changes requested):

ci/ci/ci.py
daniel-goldstein: As in, display a message on the UI above the empty table that they're not in the mapping?

danking: I was thinking, the page would just have an empty `<table>`, their username is listed at the top and their `gh_username` is listed, but, perhaps, blank? We could also use `set_message` from `web_common` to signal an error condition to the user.

ci/ci/constants.py
```diff
@@ -1,23 +1,32 @@
...
+    def __init__(self, gh_username: str, hail_username: Optional[str]):
+        self.gh_username = gh_username
+        self.hail_username = hail_username
+
+
+AUTHORIZED_USERS = [
+    User('danking', None),
+    User('cseed', None),
+    User('konradjk', None),
+    User('jigold', None),
```
danking, line 17: she is jigold

ci/ci/constants.py
```diff
@@ -1,23 +1,32 @@
...
-}
+
+class User:
+    def __init__(self, gh_username: str, hail_username: Optional[str]):
+        self.gh_username = gh_username
+        self.hail_username = hail_username
+
+
+AUTHORIZED_USERS = [
+    User('danking', None),
```
danking, line 14: I'm dking

ci/ci/constants.py
```diff
@@ -1,23 +1,32 @@
...
+    User('cseed', None),
+    User('konradjk', None),
```
danking, line 16: konradk

ci/ci/constants.py
```diff
@@ -1,23 +1,32 @@
...
+    User('lfrancioli', None),
+    User('tpoterba', 'tpoterba'),
+    User('chrisvittal', None),
+    User('catoverdrive', None),
+    User('johnc1231', 'johnc'),
+    User('nawatts', None),
+    User('mkveerapen', None),
+    User('Dania-Abuhijleh', None),
+    User('bw2', None),
+    User('pwc2', None),
```
danking, line 28: pcumming

ci/ci/constants.py
```diff
@@ -1,23 +1,32 @@
...
+    User('lgruen', None),
+    User('CDiaz96', None),
```
danking, line 30: carolin

ci/ci/constants.py
```diff
@@ -1,23 +1,32 @@
...
+AUTHORIZED_USERS = [
+    User('danking', None),
+    User('cseed', None),
+    User('konradjk', None),
+    User('jigold', None),
+    User('patrick-schultz', None),
+    User('lfrancioli', None),
+    User('tpoterba', 'tpoterba'),
+    User('chrisvittal', None),
+    User('catoverdrive', None),
```
danking, line 22: wang

ci/ci/constants.py
```diff
@@ -1,23 +1,32 @@
...
+        self.gh_username = gh_username
+        self.hail_username = hail_username
+
+
+AUTHORIZED_USERS = [
+    User('danking', None),
+    User('cseed', None),
+    User('konradjk', None),
+    User('jigold', None),
+    User('patrick-schultz', None),
```
danking, line 18: pschultz

ci/ci/constants.py
```diff
@@ -1,23 +1,32 @@
...
+    User('lfrancioli', None),
+    User('tpoterba', 'tpoterba'),
+    User('chrisvittal', None),
```
danking, line 21: I'm not sure, ask him

danking (dismissed):

danking (approved):

PR #10038, opened 2021-02-11
[query] Fix 0-partition matrix table writes.

tpoterba:
CHANGELOG: Fix crashes when writing/reading matrix tables with 0 partitions.

### DIFF
--- a/hail/src/main/scala/is/hail/rvd/AbstractRVDSpec.scala
+++ b/hail/src/main/scala/is/hail/rvd/AbstractRVDSpec.scala
@@ -124,7 +124,13 @@ object AbstractRVDSpec {
     val extendedNewPartitioner = newPartitioner.map(_.extendKey(partitioner.kType))
     val tmpPartitioner = extendedNewPartitioner match {
       case Some(np) => np.intersect(partitioner)
-      case None => partitioner
+      case None =>
+        val iOrd = partitioner.kord.intervalEndpointOrdering
+        val includedIndices = (0 until partitioner.numPartitions).filter { i =>
+          val rb = partitioner.rangeBounds(i)
+          !rb.isDisjointFrom(iOrd, rb)
+        }.toArray
+        partitioner.copy(rangeBounds = includedIndices.map(partitioner.rangeBounds))
     }
 
     val (indexSpecLeft, indexSpecRight) = (specLeft, specRight) match {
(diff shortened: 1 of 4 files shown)


hail/src/main/scala/is/hail/rvd/AbstractRVDSpec.scala
```diff
@@ -189,16 +189,26 @@ object AbstractRVDSpec {
...
+        val tmpPart = np.intersect(partitioner)
+        assert(specLeft.key.nonEmpty)
+        val p = tmpPart.rangeBounds.map { b => specLeft.partFiles(partitioner.lowerBoundInterval(b)) }
+        (p, tmpPart)
+      case None =>
+        val partsBuilder = new BoxedArrayBuilder[String]()
+        val rbBuilder = new BoxedArrayBuilder[Interval]()
+        partitioner.rangeBounds.foreach { b =>
+          val idx = partitioner.lowerBoundInterval(b)
+          if (idx < partitioner.numPartitions) {
```
patrick-schultz, line 203: This is confusing me. I don't see how this can ever be false: `idx` must be at most the index of `b`. If so, the None case looks equivalent to before. What was the intention here?

patrick-schultz (changes requested):

hail/src/main/scala/is/hail/rvd/AbstractRVDSpec.scala
tpoterba: This is not always true -- the test case I added here writes a table with a single partition whose bound is: `([0]-[0])`

`lowerBoundInterval` returns 1 in this case.

patrick-schultz: Ah, I see. Thanks. I was assuming non-degenerate intervals.

hail/src/main/scala/is/hail/rvd/AbstractRVDSpec.scala
```diff
@@ -189,16 +189,23 @@ object AbstractRVDSpec {
...
+        (p, tmpPart)
+      case None =>
+        // need to remove partitions with degenerate intervals
+        // these partitions are necessarily empty
+        val iOrd = partitioner.kord.intervalEndpointOrdering
+        val includedIndices = (0 until partitioner.numPartitions).filter { i =>
+          val rb = partitioner.rangeBounds(i)
+          rb.isDisjointFrom(iOrd, rb)
+        }.toArray
+        (includedIndices.map(specLeft.partFiles), partitioner.copy(rangeBounds = includedIndices.map(partitioner.rangeBounds)))
```
patrick-schultz, line 206: I actually think the None case could just be `(specLeft.partFiles, partitioner)`, but I'm okay with dropping degenerate partitions too.

patrick-schultz (approved):

hail/src/main/scala/is/hail/rvd/AbstractRVDSpec.scala
tpoterba: Oh! Yes, this is better. The root problem is that the lower bound is not meaningful, but this is fine.

johnc1231:
Conflicts.

johnc1231:
Fixed conflicts

tpoterba:
I think there's still a bug. Will get to it at some point...

PR #10042, opened 2021-02-11
[query] Removing fundamental types on strings

johnc1231:
Stop using fundamental type on PString where possible. We just will have a special  `binaryRepresentation` or whatever on strings to get the underlying binary representation as long as it exists. Also, pull out some other fundamentals I suspect are not needed


johnc1231:
cc @tpoterba , who asked for this.

chrisvittal (approved):

Issue #9895, opened 2021-01-20
hl.hadoop_ls is more than 20x slower than "gsutil ls -l" when a lot of files


bw2:
I have a gs:// directory with ~50k files. 

I can run 
```
In [13]: time x = !gsutil ls gs://bucket/some-dir/*.bam
CPU times: user 25.5 ms, sys: 20.1 ms, total: 45.6 ms
Wall time: 26.3 s
```

but running
```
In [14]: time y = hl.hadoop_ls("gs://bucket/some-dir/*.bam")
```
was still not done after more than 10 minutes.

bw2:
In case this allows for much faster run times, it would be nice to have the option to retrieve just the file paths via `hl.hadoop_ls("gs://bucket/some-dir/*.bam")`, and not the file sizes, mod times, etc.

tpoterba:
I think we can make this way faster by multithreading it.

tpoterba:
implementation of interest: 
https://github.com/hail-is/hail/blob/2225385866fdfd4f894c317e4c704ccb79bed194/hail/src/main/scala/is/hail/io/fs/HadoopFS.scala#L101-L108

tpoterba:
My guess is the slow bit is blocking on every file in the serial map in 106

tpoterba:
Nick's fix makes this much better. Cotton will add an async gcfs at some point.

PR #10043, opened 2021-02-11
[query] Parallelize HadoopFS.listStatus

nawatts:
This reduced the time to run `hl.hadoop_ls` on a glob that matches 1000 files in a GCS bucket from ~55s to ~10s. Still not as good as `gsutil ls -l`, which takes ~3s.

Related to #9895

### DIFF
--- a/hail/src/main/scala/is/hail/io/fs/HadoopFS.scala
+++ b/hail/src/main/scala/is/hail/io/fs/HadoopFS.scala
@@ -105,9 +105,10 @@ class HadoopFS(val conf: SerializableHadoopConfiguration) extends FS {
     if (statuses == null) {
       throw new FileNotFoundException(filename)
     } else {
-      statuses.map(_.getPath)
+      statuses.par.map(_.getPath)
         .flatMap(fs.listStatus(_))
         .map(new HadoopFileStatus(_))
+        .toArray
     }
   }
 


cseed:
I have a implemented a highly concurrent Python asyncio filesystem that supports GCS and the local file system (and soon S3).  It is my intention to ultimately replace the hadoop_* functions with this.  The new thing feels pretty fast: copy benchmarks 2-5x faster than gsutil for example, esp. when working with lots of files.  Some remarks:
 - It is designed to do the minimal number of system calls/API calls per operation so there is serial loops like this anywhere in the code.
 - Our short term goal is to use this for the input/output steps in Batch.
 - It doesn't support Hadoop (and I'm not super exciting about maintaining that).
 - Some things will be much faster because no round trip the JVM. 
 - The interface is fully async, so we'll need to build some wrappers if you want a synchronous interface.  The async interface will get you concurrency within operations (copy, rmtree), the sync interface only gets you currency within operations.
 - The list files operation doesn't support  globbing yet.
 - There are no docs yet.
 - Compared to Hadoop/POSIX, the interface is slightly lower level but it was designed to map well onto the filesystems we want to support.  There is no `stat`, for example, but is statfile (which requires the input to be a file) and listfiles (which requires it to be a directory), although we could build that.
 - I'd say the code is beta and not quite completely solid but getting close.

Here is the AsyncFS interface: https://github.com/hail-is/hail/blob/main/hail/python/hailtop/aiotools/fs.py#L70

Here is an example creating a router filesystem that supports GCS and the local file system: https://github.com/hail-is/hail/blob/main/hail/python/test/hailtop/test_aiogoogle.py#L17

I'd be happy to chat more about what would make this attractive for you guys to switch to.

nawatts:
> I'd be happy to chat more about what would make this attractive for you guys to switch to.

Tagging @bw2 since he filed the original issue (#9895) here. I suspect it was related to [gnomad-readviz](https://github.com/broadinstitute/gnomad-readviz).

nawatts:
> The interface is fully async, so we'll need to build some wrappers if you want a synchronous interface. The async interface will get you concurrency within operations (copy, rmtree), the sync interface only gets you currency within operations.

+1 for this. Synchronous wrappers would make replacing existing uses of `hl.hadoop_*` much easier.

cseed:
OK, great.  I'm working on this as we speak, so I'll prototype a draft sync interface and run it by you all.

nawatts:
Is there interest in this change to speed up `hl.hadoop_ls` in the meantime between now and the switch to the new asyncio filesystem? Or should this PR be closed?

bw2:
I would appreciate using this par version until the async one is merged.

tpoterba:
I'm fine with this.

tpoterba:
@cseed ?

johnc1231:
@tpoterba  @cseed What's going on with this?

cseed:
Sorry, I missed this. I'm fine with this.

tpoterba (approved):

PR #10045, opened 2021-02-12
[hail] super impute

danking:
CHANGELOG: `hl.Table.parallelize` is much more flexible and now successfully imports most Hail-compatible data.

I really wanted to load the hail-is/hail pull requests into Hail. I did not want to specify
the types of all 271 fields. I souped up Hail's `impute_type`:

- If an empty array, set, dict or `None` appears at any nesting level, but a "peer" is non-empty and
  non-missing, we accept the peer's type.
- We take the union of two struct types as long as they agree on their intersection.
- If we discover a dict that cannot be imputed as a Hail dict, we try to impute it as a struct.

If you like this change, I'll add tests. Note: I had to change `HailType` to include `None`.

### DIFF
--- a/hail/python/hail/expr/expressions/base_expression.py
+++ b/hail/python/hail/expr/expressions/base_expression.py
@@ -124,9 +124,23 @@ class ExpressionWarning(Warning):
         super(ExpressionWarning, self).__init__(msg)
 
 
-def impute_type(x):
+def impute_type(x, partial_type=None):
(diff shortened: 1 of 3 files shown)


tpoterba:
>  [hail] super impute #10045 

I was on board from when I read the title.

tpoterba:
I'll looks closer tomorrow but this seems great!

danking:
closing until I find time to add tests

PR #10049, opened 2021-02-12
[devbin] add devbin/functions.sh

danking:
I use these functions to monitor the k8s cluster. These are useful in the interim while we
move towards more robust monitoring solutions. To make these accessible modify your ~/.bashrc
or ~/.zshrc to have this line:

    source /path/to/hail-repository/devbin/functions.sh

cc: services-team: @jigold @CDiaz96 @catoverdrive @cseed

### DIFF
--- a/devbin/functions.sh
+++ b/devbin/functions.sh
@@ -162,10 +162,10 @@ download-secret() {
 }
 
 upload-secret() {
-    # upload-secret secret-name namespace
+    # upload-secret
     #
     # Upload a secret that has been previously downloaded using download-secret. If you intend to
-    # upload to a different namespace, ensure you've also modified contents.json.
+    # upload to a different namespace, ensure you've also modified secret.json.
 	  name=$(jq -r '.metadata.name' secret.json)
 	  namespace=$(jq -r '.metadata.namespace' secret.json)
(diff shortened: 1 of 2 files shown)


daniel-goldstein (changes requested):
These are awesome. Could you add that instruction to the dev docs under the Env / Tooling > Services?

devbin/functions.sh
```diff
@@ -0,0 +1,173 @@
...
+	  kubectl get secret $name --namespace $namespace -o json > secret.json
+	  mkdir contents
+	  for field in $(jq -r  '.data | keys[]' secret.json)
+	  do
+		    jq -r '.data["'$field'"]' secret.json | base64 -D > contents/$field
+	  done
+}
+
+upload-secret() {
+    # upload-secret secret-name namespace
```
daniel-goldstein, line 165: I don't see any function arguments used here?

devbin/functions.sh
```diff
@@ -0,0 +1,173 @@
...
+    #
+    # Upload a secret that has been previously downloaded using download-secret. If you intend to
+    # upload to a different namespace, ensure you've also modified contents.json.
```
daniel-goldstein, line 168: This should be `secret.json`

daniel-goldstein (approved):

PR #10054, opened 2021-02-16
[query] Default to Spark 3

johnc1231:
This PR:

- Changes default hail Spark version to Spark 3
- Changes default hail Scala version to 2.12
- ~Changes the `build_hail_spark3` rule to `build_hail_spark2`. I don't know how long we will care if we still build under Spark 2.~ Scratch that, I deleted this. We don't support spark 2 anymore.  
- Changes the dataproc image used in `hailctl dataproc start` to one that uses Spark 3.1.1.

### DIFF
--- a/batch/Dockerfile.worker
+++ b/batch/Dockerfile.worker
@@ -10,7 +10,7 @@ RUN hail-apt-get-install \
 COPY docker/hail-ubuntu/pip.conf /root/.config/pip/pip.conf
 COPY docker/hail-ubuntu/hail-pip-install /bin/hail-pip-install
 COPY docker/requirements.txt .
-RUN hail-pip-install -r requirements.txt pyspark==2.4.0
+RUN hail-pip-install -r requirements.txt pyspark==3.1.1
 
 ENV SPARK_HOME /usr/local/lib/python3.7/site-packages/pyspark
 ENV PATH "$PATH:$SPARK_HOME/sbin:$SPARK_HOME/bin"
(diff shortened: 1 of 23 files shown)


johnc1231:
PCA test failing, this is because https://github.com/apache/spark/blob/v3.0.1/mllib/src/main/scala/org/apache/spark/mllib/linalg/distributed/RowMatrix.scala#L795 is wrong prior to 3.1.1. `MAX_RESULT_SIZE` of 0 is supposed to be interpreted as "unlimited".

tpoterba:
oh dear.

johnc1231:
Actually also fixed in 3.0.2, so will test with that for now. Trying to get all tests passing in anticipation of Google's dataproc image update and Spark 3.1.1 release.

johnc1231:
Spark 3.1.1 is out, dataproc image should be updated from the release candidate dependency within a week or so I think. The only remaining issue I think is a weird one, a particular blockmatrix test is failing because json4s can't find a constructor for `BlockMatrixSparsity` objects. Looking into it.

johnc1231:
Closing in favor of a new approach that relocates json4s

johnc1231:
Reopening, this is the right approach. Spark's version of json4s has a bug that forced me to rename the `apply` method on `BlockMatrixSparsity` that was curried. It seems to not support curried methods.

johnc1231:
This is now just blocked on me testing that we have an elasticsearch export story.

johnc1231:
@tpoterba Added the warnings / checks you wanted to build.gradle.

tpoterba (approved):

tpoterba:
ship it! Awesome work!

PR #10056, opened 2021-02-16
[site][website] dynamic website

danking:
Website is an aiohttp application which templates docs pages and normal pages. This opens the path
towards unifying the visual appearance of hail.is, the docs, and the services.

I simplified documentation generation at the cost of building the docs twice per commit. A new step,
`make_pip_versioned_docs` builds the pip version of the docs without testing them. `make_docs`
continues to work as it did before. The website uses the docs from `make_pip_versioned_docs`. The
GCS docs location is now completely unused.

Website has four key folders:

- `website/website/pages/`: Jinja2 templated HTML pages. Served at `/`.

- `website/website/docs/`: Hail & Batch docs pages, all HTML pages are templated with Jinja2.
   Served at `/docs`

- `website/website/templates/`: Jinaj2 templates that are used in pages or in docs.

- `website/website/static/`: Non-templated files. Served at `/static`.

The website can be developed locally in or outside of Docker:
```
make -C website run
```
or
```
make -C website rundocker
```

---

I had to rename site to website due to a Python package conflict. I also deleted two unused css
files. I also removed PLINK from hail_run_image because it was slowing down my iteration speed
and was a long-term FIXME anyway.


hail/Makefile
```diff
@@ -297,6 +297,7 @@ batch-docs:
 	      BUILDDIR=$(HAIL_DIR)/build/docs/batch \
 	      html
 	rm -rf build/www/docs/batch
+	mkdir -p build/www/docs
```
danking, line 300: I found this bug. If you run `make -C hail batch-docs` on a fresh repo it will fail because none of these directories exist.

danking:
@cseed You can take this for a spin at https://internal.hail.is/dking/website/ . It's a bit sluggish. the end-to-end timing at gateway for the `api.html` (the API reference) hovers around 125ms. About 35ms is website rendering. Router adds anywhere from 5ms to 20ms. It seems like hitting the router-resolver and whatever else `gateway` does makes up the rest. [Service logs here](https://console.cloud.google.com/logs/query;query=labels.%22k8s-pod%2Fapp%22%3D%22gateway%22%20OR%20labels.%22k8s-pod%2Fapp%22%3D%22router%22%20OR%20labels.%22k8s-pod%2Fapp%22%3D%22website%22%0AjsonPayload.message:%22%2Fdking%2Fwebsite%2Fdocs%2F0.2%2Fapi.html%22;summaryFields=labels%252F%2522k8s-pod%252Fapp%2522:false:32:beginning?project=hail-vdc&query=%0A). I think in production it will be fine because there's a fair bit less indirection.

website/website/website.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+routes.static('/docs/0.2/_static', f'{MODULE_DIR}/docs/0.2/_static')
+routes.static('/docs/batch/_static', f'{MODULE_DIR}/docs/batch/_static')
+
+
+@routes.get('/docs/{tail:.*\.html}')
+@web_maybe_authenticated_user
+async def serve_docs(request, userdata):
+    return await render_template('website', request, userdata, request.match_info['tail'], dict())
```
cseed, line 41: This has a bad smell, I'm uncomfortable feeding arbitrary input (the URL) into jinja2.  Is it doing validation?  Can we be sure we're only pulling files from files from docs/ here. Letting a request pull an arbitrary template is probably not OK.

cseed (changes requested):

website/website/website.py
```diff
@@ -0,0 +1,80 @@
...
+routes.static('/docs', f'{MODULE_DIR}/docs')  # serve non-HTML
+
+
+def run(local_mode):
+    app = web.Application()
+
+    if local_mode:
+        assert deploy_config.location() == 'external', deploy_config.location()
+        aiohttp_session.setup(app, aiohttp_session.cookie_storage.EncryptedCookieStorage(
+            b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
```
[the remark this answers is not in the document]
danking: It looks like I can't deli the setup_aiohttp_session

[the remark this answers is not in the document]
danking: I added an error log to this effect.

danking (commented):
OK, I think this should satisfy your concerns.

1. The entire website python package directory is set to read-only.
2. I list all the static docs pages on start-up and only allow those paths to be used (this avoids `../` trickery).
3. I only template html files

I did not change the dynamic listing of `pages/`. I think that would be annoying to developers and I don't think it adds much security.

danking:
@cseed bump

website/website/website.py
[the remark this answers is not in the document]
cseed: > This is for the "local mode"

Ah, got it, I had misread, seems fine.

website/website/website.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def run(local_mode):
+    app = web.Application()
+
+    if local_mode:
+        log.error('running in local mode with bogus cookie storage key')
+        aiohttp_session.setup(app, aiohttp_session.cookie_storage.EncryptedCookieStorage(
+            b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
+            cookie_name=deploy_config.auth_session_cookie_name(),
```
cseed, line 101: This will overwrite the production cookie?   Do you want a different name here?

website/website/website.py
```diff
@@ -0,0 +1,120 @@
...
+for path in STATIC_DOCS_PATHS:
+    routes.static('/docs/' + path, DOCS_PATH + path)
+    FQ_STATIC_DOCS_PATHS.add(DOCS_PATH + path)
+
+
+docs_pages = set(
+    x[0][len(DOCS_PATH):] + '/' + y
+    for x in os.walk(DOCS_PATH)
+    if x[0] not in FQ_STATIC_DOCS_PATHS
+    for y in x[2])
```
cseed, line 66: This reads OK, but honestly, it still seems brittle.  There's no test for the `x[0] not in FQ_STATIC_DOCS_PATHS` condition being right, for example.  I feel like the main problem here is that we're mixing static and templated files  in docs/, but not elsewhere.  Is there a reason we have to do that?

cseed (changes requested):
One concrete change and I'll  approve, but a few further points of discussion.

website/website/website.py
danking: Sphinx is generating the structure. I'd have to walk all the HTML and replace URLs that reference `_static`. If we had our own doc generator this wouldn't be an issue.

cseed (approved):

PR #10057, opened 2021-02-16
[query] Redesign ETypes to better support PCodes

tpoterba:


tpoterba:
This is a spicy meatball. I'll write up some text about the changes as I debug, but wanted to get something up.

tpoterba:
One of the changes I made here was deleting ETransposedArrayOFStructs and EPackedIntArray - they're dead code, and it's not worth the effort to bring them along as we change things incrementally. We'll bring them back as soon as we have SStackStruct.

chrisvittal (approved):

PR #10061, opened 2021-02-17
[query] Improve code generation of If/Coalesce

tpoterba:


tpoterba:
https://gist.github.com/tpoterba/203a24cad9082bfe82e6119bae574f43

tpoterba:
```
linear_regression_rows_nd      15.7%    386.143     60.770
```

johnc1231:
So this made a big difference across the board. This is why we need benchmarks nightly, would have caught this.

johnc1231 (approved):

tpoterba:
yep, absolutely.

PR #10064, opened 2021-02-18
[query] Remove PType.fundamentalType

chrisvittal:
Fundamental types seem like a good idea, but using them in code generation at any stage but point manipulation of values can constrain our ability to use other representations for a given type.

### DIFF
--- a/hail/src/main/scala/is/hail/annotations/Region.scala
+++ b/hail/src/main/scala/is/hail/annotations/Region.scala
@@ -218,20 +218,22 @@ object Region {
     false
   }
 
-  def loadPrimitive(typ: PType): Code[Long] => Code[_] = typ.fundamentalType match {
+  def loadPrimitive(typ: PType): Code[Long] => Code[_] = typ match {
     case _: PBoolean => loadBoolean
     case _: PInt32 => loadInt
     case _: PInt64 => loadLong
     case _: PFloat32 => loadFloat
     case _: PFloat64 => loadDouble
+    case _: PCanonicalCall => loadInt
   }
 
-  def storePrimitive(typ: PType, dest: Code[Long]): Code[_] => Code[Unit] = typ.fundamentalType match {
+  def storePrimitive(typ: PType, dest: Code[Long]): Code[_] => Code[Unit] = typ match {
     case _: PBoolean => v => storeBoolean(dest, coerce[Boolean](v))
     case _: PInt32 => v => storeInt(dest, coerce[Int](v))
     case _: PInt64 => v => storeLong(dest, coerce[Long](v))
     case _: PFloat32 => v => storeFloat(dest, coerce[Float](v))
     case _: PFloat64 => v => storeDouble(dest, coerce[Double](v))
+    case _: PCanonicalCall => v => storeInt(dest, coerce[Int](v))
   }
 
   def stagedCreate(blockSize: Size, pool: Code[RegionPool]): Code[Region] =
--- a/hail/src/main/scala/is/hail/annotations/Region.scala
+++ b/hail/src/main/scala/is/hail/annotations/Region.scala
@@ -224,7 +224,6 @@ object Region {
     case _: PInt64 => loadLong
     case _: PFloat32 => loadFloat
     case _: PFloat64 => loadDouble
-    case _: PCanonicalCall => loadInt
   }
 
   def storePrimitive(typ: PType, dest: Code[Long]): Code[_] => Code[Unit] = typ match {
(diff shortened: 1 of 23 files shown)


hail/src/main/scala/is/hail/annotations/Region.scala
```diff
@@ -218,20 +218,22 @@ object Region {
...
     case _: PBoolean => loadBoolean
     case _: PInt32 => loadInt
     case _: PInt64 => loadLong
     case _: PFloat32 => loadFloat
     case _: PFloat64 => loadDouble
+    case _: PCanonicalCall => loadInt
   }
 
-  def storePrimitive(typ: PType, dest: Code[Long]): Code[_] => Code[Unit] = typ.fundamentalType match {
+  def storePrimitive(typ: PType, dest: Code[Long]): Code[_] => Code[Unit] = typ match {
```
chrisvittal, line 230: It's not essential to this PR, but load/storePrimitive are only used in AggregatorState and should be removed.

tpoterba: agreed. These need to go.

hail/src/main/scala/is/hail/utils/richUtils/RichCodeInputBuffer.scala
```diff
@@ -87,11 +87,12 @@ class RichCodeInputBuffer(
...
   }
 
-  def readPrimitive(typ: PType): Code[_] = typ.fundamentalType match {
+  def readPrimitive(typ: PType): Code[_] = typ match {
     case _: PBoolean => readBoolean()
     case _: PInt32 => readInt()
     case _: PInt64 => readLong()
     case _: PFloat32 => readFloat()
     case _: PFloat64 => readDouble()
+    case _: PCanonicalCall => readInt()
```
tpoterba, line 96: don't even add this case. Call isn't a primitive, I want to fail if we hit this.

hail/src/main/scala/is/hail/utils/richUtils/RichCodeOutputBuffer.scala
```diff
@@ -56,11 +56,12 @@ class RichCodeOutputBuffer(
...
     ob.invoke[String, Unit]("writeUTF", s)
 
-  def writePrimitive(typ: PType): Code[_] => Code[Unit] = typ.fundamentalType match {
+  def writePrimitive(typ: PType): Code[_] => Code[Unit] = typ match {
     case _: PBoolean => v => writeBoolean(coerce[Boolean](v))
     case _: PInt32 => v => writeInt(coerce[Int](v))
     case _: PInt64 => v => writeLong(coerce[Long](v))
     case _: PFloat32 => v => writeFloat(coerce[Float](v))
     case _: PFloat64 => v => writeDouble(coerce[Double](v))
+    case _: PCanonicalCall => v => writeInt(coerce[Int](v))
```
tpoterba, line 65: same

hail/src/main/scala/is/hail/annotations/Region.scala
```diff
@@ -218,20 +218,22 @@ object Region {
...
     case _: PBoolean => v => storeBoolean(dest, coerce[Boolean](v))
     case _: PInt32 => v => storeInt(dest, coerce[Int](v))
     case _: PInt64 => v => storeLong(dest, coerce[Long](v))
     case _: PFloat32 => v => storeFloat(dest, coerce[Float](v))
     case _: PFloat64 => v => storeDouble(dest, coerce[Double](v))
+    case _: PCanonicalCall => v => storeInt(dest, coerce[Int](v))
```
tpoterba, line 236: don't even add this case. I want to fail here instead.

hail/src/main/scala/is/hail/annotations/Region.scala
```diff
@@ -218,20 +218,22 @@ object Region {
...
   }
 
-  def loadPrimitive(typ: PType): Code[Long] => Code[_] = typ.fundamentalType match {
+  def loadPrimitive(typ: PType): Code[Long] => Code[_] = typ match {
     case _: PBoolean => loadBoolean
     case _: PInt32 => loadInt
     case _: PInt64 => loadLong
     case _: PFloat32 => loadFloat
     case _: PFloat64 => loadDouble
+    case _: PCanonicalCall => loadInt
```
tpoterba, line 227: same

tpoterba (changes requested):
🎉 

One small comment about not adding cases for Call

tpoterba (approved):

tpoterba:
🍾 🍾 🍾 🍾 🍾

PR #10069, opened 2021-02-19
[batch] Fix provisioning bugs

jigold:
The issue is we start billing for instances as soon as they're created with the API. However, if an instance is stuck in provisioning and never activates, we never update the start billing time to account for the lack of resource. This PR uses the `lastStartTimestamp` in the [Google REST API](https://cloud.google.com/compute/docs/reference/rest/v1/instances/get). This value is in RFC3339 format. I think this is the same format the timestamp in the activity logs, so I copied how we parse that value. If we delete the instance due to activation timeout, then we set the attempt start time to NULL so it's not billed.

I couldn't find good documentation on this, but it seems like the `lastStartTimestamp` approximates what we care about for the purposes of checking for stuck workers. I checked it on an instance that was provisioning and the value was missing. Once the instance was in starting, the value was about 10 seconds after the `creationTimestamp`. 

QUESTION: This does raise a question on whether we should be using the `lastStartTimestamp` when billing users if the difference is around 10 seconds. That will be a harder change, but is probably doable. We can't access the `lastStartTimestamp` through the metadata on the worker which would have been the easiest solution. We can get the compute client on the worker and access the `lastStartTimestamp` that way and set the job start time to the instance start time. I'd need to change the database for the attempts trigger to account for this. For the scenario where a job private job is cancelled while creating the instance, we would either need to make the additional API call or we just leave the time we created the instance.

Thoughts?

### DIFF
--- a/batch/batch/driver/instance_collection.py
+++ b/batch/batch/driver/instance_collection.py
@@ -135,13 +135,7 @@ class InstanceCollection:
         # PROVISIONING, STAGING, RUNNING, STOPPING, TERMINATED
         gce_state = spec['status']
 
-        last_start_timestamp = spec.get('lastStartTimestamp')
-        if last_start_timestamp:
-            last_start_time_msecs = dateutil.parser.isoparse(last_start_timestamp).timestamp() * 1000
-        else:
-            last_start_time_msecs = None
-
-        log.info(f'{instance} gce_state {gce_state} last_start_timestamp {last_start_timestamp}')
+        log.info(f'{instance} gce_state {gce_state}')
 
         if (gce_state == 'PROVISIONING'
                 and instance.state == 'pending'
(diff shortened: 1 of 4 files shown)


danking:
ISO8601 is roughly a superset of RFC3339, so yes, that dateutil library function is the right one.

danking (changes requested):
I'm disinclined to switch to something called "last start timestamp" for billing purposes because that sounds like something that can change arbitrarily (it's simply the "last" start timestamp, it's not some canonical time like "creationTimestamp").

Why is it more useful to use last start timestamp (which seems to possibly not exist) than time since creation time stamp for checking for stuck workers?

jigold:
> Why is it more useful to use last start timestamp (which seems to possibly not exist) than time since creation time stamp for checking for stuck workers?

Because the problem was if provisioning took > 5 minutes, the instant the machine was at staging or running, we'd kill the instance because it had been more than 5 minutes. This is what happened when google was taking forever to provision workers and then we'd immediately kill the ones we got.

batch/batch/driver/instance_collection.py
```diff
@@ -133,17 +134,30 @@ async def check_on_instance(self, instance):
...
+
+        last_start_timestamp = spec.get('lastStartTimestamp')
+        if last_start_timestamp:
+            last_start_time_msecs = dateutil.parser.isoparse(last_start_timestamp).timestamp() * 1000
+        else:
+            last_start_time_msecs = None
+
+        log.info(f'{instance} gce_state {gce_state} last_start_timestamp {last_start_timestamp}')
+
+        if (gce_state == 'PROVISIONING'
```
jigold, line 146: Do we want to kill any instance that is provisioning for more than 5 minutes?

jigold:
Addressed, but question about provisioning.

batch/batch/driver/instance_collection.py
```diff
@@ -133,18 +134,28 @@ async def check_on_instance(self, instance):
 
         # PROVISIONING, STAGING, RUNNING, STOPPING, TERMINATED
         gce_state = spec['status']
+
         log.info(f'{instance} gce_state {gce_state}')
 
+        if (gce_state == 'PROVISIONING'
+                and instance.state == 'pending'
+                and time_msecs() - instance.time_created > 5 * 60 * 1000):
+            log.info(f'{instance} did not provision within 5m after creation, deleting')
```
danking, line 143: I think this should be an error, it will cause us to respond to high rates of long-provisioning instances.

batch/batch/driver/instance_collection.py
danking: Yes, AFAICT, we pay for provisioning time.

danking (dismissed):

danking (approved):

jigold:
I had to rebase. This has a migration. Take the WIP tag off at your own leisure.

danking:
merge merge merge!

PR #10070, opened 2021-02-19
[batch] Add more info to UI pages

jigold:
@danking Feel free to reassign to Carolin if you're happy with the overall concept. We were trying to check off a task for February on the ToDo list.

cc: @CDiaz96 

Screen Shot 2021-02-19 at 11 25 29 AM

Screen Shot 2021-02-19 at 11 25 38 AM

Screen Shot 2021-02-19 at 11 25 52 AM

### DIFF
--- a/batch/batch/batch.py
+++ b/batch/batch/batch.py
@@ -41,6 +41,7 @@ def batch_record_to_dict(record):
 
     d = {
         'id': record['id'],
+        'user': record['user'],
         'billing_project': record['billing_project'],
         'token': record['token'],
         'state': state,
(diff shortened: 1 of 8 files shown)


danking:
Looks good but I think we need to stop making links the size of a single character. Let's make the entire `Batch 1` header clickable and also the whole `Batch ID: 1` clickable.

danking:
And the entire ID cell should be clickable as well, not just the number `1`.

jigold:
> Let's make the entire Batch 1 header clickable

Why would you want to do that? You're already on that page?

> And the entire ID cell should be clickable as well, not just the number 1.

Can we make this a separate PR that maybe Carolin can do? I think it requires CSS or something more complicated than that.

jigold:
Bump. @CDiaz96 Let me know if you want to walk through this again.

CDiaz96 (approved):

CDiaz96 (approved):

CDiaz96 (approved):

PR #10071, opened 2021-02-19
[batch] Cancel orphaned attempts

jigold:
This solves a problem we saw where an attempt was scheduled on the instance, but when entering the schedule event into the database it failed. In the meantime, the job is cancelled before the database call can be retried. Now the job is presumed cancelled even though there are still attempts running. I added a loop to find attempts that are orphaned and unschedule them.

Please double check the SQL query here that I got all the conditions correct!

### DIFF
--- a/batch/batch/driver/job_private.py
+++ b/batch/batch/driver/job_private.py
@@ -126,8 +126,6 @@ LIMIT 300;
             await waitable_pool.call(
                 schedule_with_error_handling, self.app, record, id, instance)
 
-            n_scheduled += 1
-
         await waitable_pool.wait()
 
         log.info(f'scheduled {n_scheduled} jobs for {self}')
--- a/batch/batch/driver/canceller.py
+++ b/batch/batch/driver/canceller.py
@@ -2,7 +2,7 @@ import logging
 import asyncio
 
 from hailtop.utils import (WaitableSharedPool, retry_long_running, run_if_changed,
-                           AsyncWorkerPool, time_msecs)
+                           AsyncWorkerPool, time_msecs, periodically_call)
 from hailtop import aiotools, aiogoogle
 from gear import Database
 
@@ -36,6 +36,8 @@ class Canceller:
         self.task_manager.ensure_future(retry_long_running(
             'cancel_cancelled_running_jobs_loop',
             run_if_changed, self.cancel_running_state_changed, self.cancel_cancelled_running_jobs_loop_body))
+
+        self.task_manager.ensure_future(periodically_call(60, self.cancel_orphaned_attempts_loop_body))
 
     def shutdown(self):
         try:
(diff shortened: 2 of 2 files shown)


batch/batch/driver/job_private.py
```diff
@@ -126,8 +126,6 @@ async def schedule_with_error_handling(app, record, id, instance):
             await waitable_pool.call(
                 schedule_with_error_handling, self.app, record, id, instance)
 
-            n_scheduled += 1
```
jigold, line 129: This was a duplicate of the increment from above.

danking:
I don't understand how an "orphaned attempt" differs from an attempt of a "cancelled running job". Can you explain the difference? In particular it seems like we should get *all* attempts in `user_cancelled_running_jobs`.

danking (dismissed):
xxx

jigold:
We run this query to get fair share values per user:

```
SELECT user, CAST(COALESCE(SUM(n_cancelled_running_jobs), 0) AS SIGNED) AS n_cancelled_running_jobs
FROM user_inst_coll_resources
GROUP BY user
HAVING n_cancelled_running_jobs > 0;
```

This query will return 0 even though there could be attempts still running. Plus these queries only look at running batches.

```
           async for batch in self.db.select_and_fetchall(
                    '''
SELECT id
FROM batches
WHERE user = %s AND `state` = 'running' AND cancelled = 1;
''',
                    (user,),
                    timer_description=f'in cancel_cancelled_running_jobs: get {user} cancelled batches'):
                async for record in self.db.select_and_fetchall(
                        '''
SELECT jobs.job_id, attempts.attempt_id, attempts.instance_name
FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_cancelled)
STRAIGHT_JOIN attempts
  ON attempts.batch_id = jobs.batch_id AND attempts.job_id = jobs.job_id
WHERE jobs.batch_id = %s AND state = 'Running' AND always_run = 0 AND cancelled = 0
LIMIT %s;
''',
                        (batch['id'], remaining.value),
                        timer_description=f'in cancel_cancelled_running_jobs: get {user} batch {batch["id"]} running cancelled jobs'):
                    record['batch_id'] = batch['id']
                    yield record
``` 

I'll think about whether I can combine these queries over the attempts. However, it seemed clearer to me to look for two different things as the other queries are optimized by looking at the batch and job state where the orphaned query doesn't care about fair share or the batch / job state..

danking (approved):
We, rather pervasively, focus on jobs. I think we need to focus on attempts more often. An unrelated example: I can't look at the logs from a preempted attempt because we don't make attempts first-class members of the UI.

Let's get this fix in, but let's also keep in mind what the system would look like if attempts were treated as first-class members. I think structuring the system in three distinct layers: batches, jobs, and attempts, might actually simplify some of the logic.

PR #10072, opened 2021-02-19
[query] get tests working

danking:
- [ ] (@tpoterba)                        caf1e1e673 add fails_service_backend
- [ ] (@tpoterba, @cseed)                a979dfba58 [hail] introduce and use mktemp and mktempd
- [ ] (@tpoterba)                        dcf026b01c [hail] make is.hail.expr.ir.functions threadsafe
- [ ] (@tpoterba)                        807f38c20e [hail] fix use of row requiredness in lowerDistributedSort
- [ ] (@catoverdrive)                    12df8eb456 [query-service] handle void-typed IRs in query-service
- [ ] (@catoverdrive)                    03357ee83d [query-service] make user cache thread-safe
- [ ] (@tpoterba)                        6c6734bc71 [query-service] bugfix: preserve globals through a shuffle
- [ ] (@catoverdrive)                    a3d2572ce7 [shuffler] log ShuffleCodecSpec anytime it is created
- [ ] (@daniel-goldstein)                8949dfec3c [scala-lsm] bugfix: least key may equal greatest key
- [ ] (@daniel-goldstein)                6067bd8e51 [services] discovered new transient error
- [ ] (@daniel-goldstein)                c8356d30bb [shuffler] more assertions in ShuffleClient
- [ ] (@daniel-goldstein)                9991da90f0 [shuffler] bugfix: shuffler needs a HailContext to decode loci
- [ ] (@daniel-goldstein)                bc0140ab6f [query-service] move hail.jar earlier in Dockerfile
- [ ] (@daniel-goldstein)                f96c28174d [query-service] permit pod scaling and remove cpu limit
- [ ] (@catoverdrive)                    6ae26339fe [query-service] simplify socket handling
- [ ] (@jigold)                          f3db30e23f [batch] teach JVMJob where to find the hail configuration files
- [ ] (@daniel-goldstein)                b5c6d85554 [query-service] switch to services team approved logging
- [ ] (@tpoterba)                        35a306c066 [query-service] query workers need a hail context
- [ ] (@daniel-goldstein, @catoverdrive) 051c89b8e7 [query-service] use a UNIX Domain Socket for Py-Scala communication
- [ ] (@daniel-goldstein, @catoverdrive) ad9ea73d7a [query-service] run tests against query service


danking:
The SHA's are changing a bit quicker than I want to update them. I'll update when it stabilizes, but the titles are still correct.

build.yaml
```diff
@@ -750,6 +750,26 @@ steps:
...
+   name: upload_test_resources_to_gcs
+   image:
+     valueFrom: base_image.image
+   script: |
+     rm -rf repo
+     mkdir repo
+     cd repo
+     {{ code.checkout_script }}
+     gcloud -q auth activate-service-account --key-file=/test-gsa-key/key.json
+     gsutil -m cp -r hail/src/test/resources/* gs://7-day/{{ token }}/test/resources/
```
cseed, line 763: What's this for?

build.yaml
```diff
@@ -761,16 +781,11 @@ steps:
      set -ex
      cd /io
      mkdir -p src/test
-     tar xzf resources.tar.gz -C src/test
      tar xzf splits.tar.gz
-     gcloud -q auth activate-service-account --key-file=/test-gsa-key/key.json
-     gsutil -m cp -r src/test/resources/fs gs://hail-test-dmk9z/{{ token }}/test/resources/
      export HAIL_TEST_SKIP_R=1
-     export HAIL_GS_FS_TEST_RESOURCES=gs://hail-test-dmk9z/{{ token }}/test/resources/fs
+     export HAIL_GS_FS_TEST_RESOURCES=gs://7-day/{{ token }}/test/resources/fs
```
cseed, line 786: So we generate the test bucket in Terraform and this will break that.

hail/python/hail/context.py
```diff
@@ -381,7 +394,55 @@ def tmp_dir():
     return Env.hc()._tmpdir
 
 
-def current_backend():
+def mktemp(*, suffix: str = '') -> DeletingFile:
+    """A temporary file that is deleted when it exits this context.
```
cseed, line 398: "this context" is confusing, because the context (manager) is returned.  "it exists"  What's "it"?  How about "Returns a context manager which creates a temporary file name which is cleaned up when the context exits."?

query/query/query.py
```diff
@@ -68,60 +46,84 @@ async def healthcheck(request):  # pylint: disable=unused-argument
...
         body = await ws.receive_json()
-        log.info(f'{endpoint}: {body}')
-        task = asyncio.ensure_future(send_ws_response(app['thread_pool'], endpoint, ws, f, jbackend, userdata, body))
-        r = await ws.receive()
-        log.info(f'{endpoint}: Received websocket message. Expected CLOSE, got {r}')
-        return ws
+        tasks = [asyncio.ensure_future(sender()),
+                 asyncio.ensure_future(receiver())]
+        done, _ = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+        assert len(done) == 1
```
cseed, line 115: They both could have completed?

query/deployment.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   selector:
     matchLabels:
       app: query
-  replicas: 1
+  replicas: 3
```
cseed, line 12: The connection the service is a websocket, right?  What happens if you reconnect and get connected to another instance?

cseed (changes requested):
Some general comments I noticed when scanning the whole thing.  I'd do a separate review for my assigned bits.

build.yaml
danking: The Hail Python tests rely on text files, among other things, in `src/test/resources`.

query/deployment.yaml
danking: The client does not support reconnection currently. If you lose your connection while waiting for a job, everything falls apart, currently. Resiliency to connection lost mid-job is on my list to address, but I haven't thought it through currently.

In the meantime, this lets me run the tests quickly because our tests seem to bottlenecked by compiler CPU use. I can saturate >10 query instances at >100% cpu use with a client using 96 processes to run all the Hail Python tests.

danking:
I'm not sure how best to address comments. I addressed all your miscellaneous comments, @cseed, in one new commit. For comments on a specific commit, I think I'll modify that commit directly.

danking:
Closing while I resolve the source of this new intermittent segfault.

PR #10073, opened 2021-02-19
[tls] use SHA256 instead of default SHA1 for internal certs

daniel-goldstein:


danking (dismissed):
Great! Can you confirm/deny if we can also use sha256 in the tis-cookbook.md examples?

daniel-goldstein:
Thankfully the same digest flags can be passed to `req`. Tested manually with `openssl x509 -noout -text` to see that the correct Signature Algorithm was chosen.

danking (approved):

PR #10079, opened 2021-02-21
[hail] fix \d warnings

danking:


johnc1231:
Can you explain why you need the double backslashes here? I've seen the warnings but didn't know why they were happening

danking:
So, the [Python docs on string literals](https://docs.python.org/3/reference/lexical_analysis.html#strings) say:
> The backslash (`\`) character is used to escape characters that otherwise have a special meaning, such as newline, backslash itself, or the quote character.

So, for example, a string literal containing a single newline character is:
```
"\n"
```
or:
```
'\n'
```

Most characters do not participate in an escape sequence. `\d`, for example, is not a valid escape sequence. Instead of erring, Python chooses to signal a warning and pretend that you actually wrote `\\d`. If you want a backslash character followed by another character you can either write any of these:
```
'\\x'
"\\x"
r'\x'
r'\\x'
```
`r"..."` indicates a "raw" string which does not interpret escape sequences.

johnc1231 (approved):
Makes sense, thanks

PR #10085, opened 2021-02-22
[auth] Add endpoint to get current server hail version

illusional:
Based on conversation in Zulip: https://hail.zulipchat.com/#narrow/stream/123000-general/topic/Server.20hail.20version.20through.20API/near/226462327

Add version endpoint to auth API.

As suggested, I've added the the `make python-version-info` to generate the `hail_version`, but this requires the git history within the docker, as the short commit hash is added to the end.

I've just cherry-picked the [merge commit](https://github.com/populationgenomics/hail/commit/700a0cf9d6e05827feca6bdd9c455e2b261e72db) from our populationgenomics fork.

### DIFF
--- a/auth/auth/auth.py
+++ b/auth/auth/auth.py
@@ -8,6 +8,7 @@ import uvloop
 import google.auth.transport.requests
 import google.oauth2.id_token
 import google_auth_oauthlib.flow
+from hailtop.hailctl import version
 from hailtop.config import get_deploy_config
 from hailtop.tls import internal_server_ssl_context
 from hailtop.hail_logging import AccessLogger
@@ -556,6 +557,14 @@ WHERE users.state = 'active' AND (sessions.session_id = %s) AND (ISNULL(sessions
     user = users[0]
 
     return web.json_response(user)
+
+
+@routes.get('/api/v1alpha/version')
+async def rest_get_version(request):  # pylint: disable=W0613
+    try:
+        return web.Response(text=version())
+    except Exception as e:
+        return web.json_response({"error": str(e)})
 
 
 async def on_startup(app):
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,7 +9,6 @@ hail/.bloop/
 hail/.gradle/
 hail/.idea/
 hail/.pytest_cache/
-.git/
 hail/.ensime.cache.d/
 hail/.ensime_cache.d/
 hail/.ensime_cache/
--- a/docker/Dockerfile.service-base
+++ b/docker/Dockerfile.service-base
@@ -7,6 +7,11 @@ RUN hail-pip-install -r service-base-requirements.txt
 
 COPY hail/python/setup-hailtop.py /hailtop/setup.py
 COPY hail/python/hailtop /hailtop/hailtop/
+COPY hail/Makefile hail/env_var.mk .git/ hailtop/
+RUN mkdir -p hailtop/python/hail hailtop/python/hailtop/hailctl hailtop/python/hail/docs/_static && \
+    (cd hailtop && echo $(pwd) && make python-version-info) && \
+    cp hailtop/python/hail/hail_*version hailtop/hailtop/hailctl && \
+    rm -rf hailtop/Makefile hailtop/env_var.mk .git/
 RUN hail-pip-install /hailtop && rm -rf /hailtop
 
 COPY gear/setup.py /gear/setup.py


auth/auth/auth.py
```diff
@@ -558,6 +559,14 @@ async def userinfo(request):
     return web.json_response(user)
 
 
+@routes.get('/api/v1alpha/version')
+async def rest_get_version(request):  # pylint: disable=W0613
```
danking, line 563: Why do you want to know the version of hailtop used by the auth service? I thought you wanted to add this to the query service.

danking (changes requested):

auth/auth/auth.py
illusional: Thanks @danking, I had initially thought the two were linked, and auth might just be a better place for it. But I see that putting this on the query service is much closer to what we want. I've moved the version endpoint to the query endpoint and tested this works on our local instance.

Thanks!

illusional:
Hi @danking, I see the initial CI result failed, but I'm unable to login and see what the failure is. I signed in to google with my popgen account, and get 504 Gateway Time-out on the `auth.hail.is/oauth2callback`, I imagine because I don't have an account there.

danking:
Hey @illusional ! Sorry for the massive latency.

OK, so, big apologies are in order, I totally lead you astray by mentioning the makefile. The Makefile *is* the source of hail version truth, but invoking the makefile inside an image build step feels wrong to me. Each step creates a layer which inflates the image sizes. Hail's images are already too big!

I took your commits and added one of my own that snags the version from the `copy_files` build step. Because I wanted `service_base_image` to depend on `copy_files`, I had to move the whole step after `copy_files`. This is a limitation in build.yaml: a step must appear *after* steps on which it depends.

I also had to move `check_services` for the same reason: it depends on `service_base_image`.

File dependencies in build.yaml work like this:
1. For runImage steps, you can only copy out-of or copy into `/io` (the reasoning is a bit complicated and somewhat historical).
2. For buildImage steps, you can copy out-of or copy into `/`
3. the `to` of an `output` specifies a file path in a "filesystem" that another step can access if it `dependsOn` the outputting step
4. the `from` of an `input` specifies a file path in the aforementioned "filesystem"; the filesystem contains all `outputs` from steps in the inputting step's `dependsOn` clause

We also have a `docker/Makefile` which is an emergency manual build system. I update that so that `hail_version` appears in the root of the docker context. The `service-base` uses the entire repository as its docker context, so I place hail_version at the root of the repository.

I moved the `version` function from `hailtop.hailctl` into `hailtop`. It seems broadly useful and isn't specific to hailctl in anyway.

Your concern about loading from pkg_resources repeated seems well-founded, so I went ahead and loaded the hail_version at package import time. This seems likely to ensure we learn about a missing hail_version file as early as possible (presumably at service start-time). This also means all hailtop installs need a hail_version file. I only found two other places that use hailtop. One of them was a completely unused Dockerfile. I deleted that (`Dockerfile.hailtop`). The other was Dockerfile.ci-test, which I updated to copy the hail_version file just like service-base.

https://github.com/hail-is/hail/compare/main...danking:add-version-endpoint

Do you want to cherry-pick that change onto your add-version-endpoint branch?

One more change request: can we remove the try-catch? I don't expect any errors in that call and I tend to avoid revealing anything about errors to our users. Aiohttp will log the error and the stack trace if you let it rise all the way.

Sorry for all the complication! Our build system is the second service we built and is clearly showing its age.

danking:
I promise I'll check tomorrow AM ET and approve if everything looks good.

danking (changes requested):
xxxx

illusional:
Hi @danking, sorry this took me a little to test. I think there's a problem with the latest changes, in my dev-deploy, it failed on the '`create_certs` and `create_accounts`, with  the error:

```
FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/lib/python3.7/dist-packages/hailtop/hail_version'
```

danking:
Hmmmm. OK, I think we're getting tripped up by MANIFEST.in and/or setup.py.

This bit in setup.py:

```
    package_data={
        'hail': ['hail_pip_version',
                 'hail_version',
                 'experimental/datasets.json'],
        'hail.backend': ['hail-all-spark.jar'],
        'hailtop.hailctl': ['hail_version', 'deploy.yaml']},
```

Should be
```
    package_data={
        'hail': ['hail_pip_version',
                 'hail_version',
                 'experimental/datasets.json'],
        'hail.backend': ['hail-all-spark.jar'],
        'hailtop': ['hail_version'],
        'hailtop.hailctl': ['deploy.yaml']},
```

A similar change needs to be made to `setup-hailtop.py`

And let's add a MANIFEST.in file in the same directory with this contents:
```
include hail/hail_pip_version
include hail/hail_version
include hail/experimental/datasets.json
include hail/backend/hail-all-spark.jar
include hailtop/hail_version
include hailtop/hailctl/deploy.yaml
```

And you'll need to copy that into Dockerfile.service-base:
```
COPY hail/python/MANIFEST.in /hailtop/MANIFEST.in
```

My bad on all this, I didn't fully check the Dockerfiles!

illusional:
Hi @danking, I've applied your changes: (dc005a9a1e746f2f752e934b42c5dcf076acbd10), and been having a look at this the past few days, and this still isn't working -  it's still failing on `create_certs` and `create_account` as it's unable to find the hail_version:

```
python3 create_certs.py <dev-namespace> config.yaml /ssl-config-hail-root/hail-root-key.pem /ssl-config-hail-root/hail-root-cert.pem
  File "create_certs.py", line 8, in <module>
    from hailtop.utils import sync_check_shell
  File "/usr/local/lib/python3.7/dist-packages/hailtop/__init__.py", line 4, in <module>
    _VERSION = pkg_resources.resource_string(__name__, 'hail_version').decode().strip()
  File "/usr/local/lib/python3.7/dist-packages/pkg_resources/__init__.py", line 1143, in resource_string
    self, resource_name
  File "/usr/local/lib/python3.7/dist-packages/pkg_resources/__init__.py", line 1388, in get_resource_string
    return self._get(self._fn(self.module_path, resource_name))
  File "/usr/local/lib/python3.7/dist-packages/pkg_resources/__init__.py", line 1611, in _get
    with open(path, 'rb') as stream:
FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/lib/python3.7/dist-packages/hailtop/hail_version'
```

Pulling the service-base image, I can see the hail_version is not there. I wonder if it's being copied in during the build.yaml correctly, I'm not quite sure how to test the actual  build.yaml context process, to see if the `hail_version` is being copied into the docker correctly, I tried seeing if the relative path was an issue but it still failed:

```diff
Index: docker/Makefile
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/docker/Makefile b/docker/Makefile
--- a/docker/Makefile	(revision 6c9840a43b53f23f80ce44a0801fac71e940d814)
+++ b/docker/Makefile	(revision 286c13f05b62a1a11051664471c736931a19fbd1)
@@ -49,7 +49,7 @@
 service-base: base-stmp
 	-docker pull $(SERVICE_BASE_LATEST)
 	$(MAKE) -C ../hail python/hailtop/hail_version
-	cp ../hail/python/hailtop/hail_version ../hail_version
+	cp ../hail/python/hailtop/hail_version /hail_version
 	python3 ../ci/jinja2_render.py '{"base_image":{"image":"base"}}' Dockerfile.service-base Dockerfile.service-base.out
 	[ "$(shell bash stat-permissions.sh Dockerfile.service-base.out)" = "644" ]
 	[ "$(shell bash stat-permissions.sh service-base-requirements.txt)" = "644" ]
```

illusional:
Hi @danking, this should be working now! I realises that hailtop isn't installed through `setup.py`, but `setup-hailtop.py`. The setup also requires `include_package_data=True`, so I've added that into both.

I've added one more line into the `Dockerfile.service-base` to check whether the version was actually installed correctly, as other services use this, and better to fail early.

This is working on my local dev-deploy (that uses CPG infrastructure though).

Happy to hopefully have this off my plate!

illusional:
Hey @danking, there were changes in your larger PR that changed the behaviour of the query service, I've migrated these changes to the `Dockerfile.service-java-run-base`. This is ready for review again.

docker/Dockerfile.service-java-run-base
```diff
@@ -24,7 +24,10 @@ RUN hail-pip-install -r service-base-requirements.txt
 
 COPY hail/python/setup-hailtop.py /hailtop/setup.py
 COPY hail/python/hailtop /hailtop/hailtop/
+COPY /hail_version /hailtop/hailtop/hail_version
+COPY hail/python/MANIFEST.in /hailtop/MANIFEST.in
 RUN hail-pip-install /hailtop && rm -rf /hailtop
+RUN echo 'from hailtop import version; print(version());' | python3
```
daniel-goldstein, line 30: Could you remove these echos? I get failing quickly but I'm hesitant to start adding validation like this into the Dockerfiles, as it's not exactly clear what should or should not be validated and it adds unused layers, so could get pretty messy quickly.

hail/python/hailtop/hailctl/__main__.py
```diff
@@ -81,7 +82,7 @@ def fmt_version(version):
 
 
 def print_version():
-    print(hailctl.version())
+    print(version())
```
daniel-goldstein, line 86: Can just use `print(version())` in place.

hail/python/setup-hailtop.py
```diff
@@ -17,8 +17,9 @@
     package_dir={
         'hailtop': 'hailtop'},
     package_data={
-        "hailtop": ["py.typed"],
-        'hailtop.hailctl': ['hail_version', 'deploy.yaml']},
+        "hailtop": ["py.typed", "hail_version"],
+        'hailtop.hailctl': ['hail_version', 'deploy.yaml']
```
daniel-goldstein, line 21: Why does `hailctl` need `hail_version` if it's getting it through `hailtop`? I'm not super well-versed in the wild west of python package management.

daniel-goldstein (changes requested):
Looks great! Just had a couple nits and a question. Nothing big so I'll authorize this SHA to make sure it passes on our infrastructure

docker/Dockerfile.service-java-run-base
illusional: Yep, removed in 7f487ad

hail/python/setup-hailtop.py
illusional: I'm not too sure, this code is already in main: 
https://github.com/hail-is/hail/blob/022d02f2d4322f26103d1242454279ad398e6032/hail/python/setup-hailtop.py#L21

I just formatted the bracket and added extra lines per Dan's suggestion: https://github.com/hail-is/hail/pull/10085#issuecomment-793064791

hail/python/hailtop/hailctl/__main__.py
illusional: Sorry, I'm not understanding your comment, the `print_version` method was already here and updating `hailctl.version` -> `version` seems like the right change?

hail/python/setup-hailtop.py
daniel-goldstein: Ok we'll leave it in for now then

hail/python/hailtop/hailctl/__main__.py
daniel-goldstein: Ya my comment wasn't very clear, just meant that having a separate function `print_version` seemed a bit unnecessary as opposed to writing `print(version())` later, but I didn't catch that you were just changing the definition. We can just leave this too

daniel-goldstein (approved):

hail/python/hailtop/hailctl/__main__.py
```diff
@@ -6,6 +6,7 @@
 import time
 
 from hailtop import hailctl
```
daniel-goldstein, line 8: failing the linter since this import is now unused

daniel-goldstein (changes requested):

hail/python/hailtop/hailctl/__main__.py
illusional: Fixed in: a94d96c

daniel-goldstein (approved):

batch/Dockerfile.worker
```diff
@@ -34,6 +34,7 @@ RUN python3 -m pip install --upgrade --no-cache-dir pip
 
 COPY hail/python/setup-hailtop.py /hailtop/setup.py
 COPY hail/python/hailtop /hailtop/hailtop/
+COPY /hail_version /hailtop/hailtop/hail_version
```
daniel-goldstein, line 37: Missed this earlier. Don't you need to copy the MANIFEST.in here?

daniel-goldstein (commented):
Failure in `test_ci` with the `hail_version` missing. I think you need to add the same COPYs to `ci/Dockerfile.test`

batch/Dockerfile.worker
illusional: Yes, thank you! I've addressed both these in f6d3580.

danking:
I think we need this fix:
```
commit b8633dab43ee56a0448e758dae0bb3c67d1f3eb3 (HEAD -> add-version-endpoint)
Author: Daniel King <dking@broadinstitute.org>
Date:   Mon Mar 22 17:47:22 2021 -0400

    fix

diff --git a/build.yaml b/build.yaml
index 7a100adec8..256ca99c91 100644
--- a/build.yaml
+++ b/build.yaml
@@ -86,7 +86,7 @@ steps:
      mkdir repo
      cd repo
      {{ code.checkout_script }}
-     make -C hail python/hail/hail_version python/hail/hail_pip_version
+     make -C hail python-version-info
      git rev-parse HEAD > git_version
    outputs:
      - from: /io/repo/auth/sql
```

illusional:
Thanks @danking! I've applied those changes, so hopefully smooth sailing.

illusional:
@danking i can’t see CI, but I see a Red Cross which looks bad.

danking:
It's always something isn't it. Digging into it now.

danking:
The security issues with sharing CI pages is a bit frustrating for external committers like yourself. I've picked up this PR here: https://github.com/hail-is/hail/pull/10220 and I'll get it to tests passing. We'll keep working towards making the CI pages open to the public. Once that PR is done, I'll have you push the commits to this branch and get it merged.

illusional:
Awesome, thanks Dan!

danking:
OK, if you merge https://github.com/danking/hail/tree/add-version-endpoint (`add-version-endpoint` in my repo, `github.com/danking/hail`) this should pass the tests and we can get it merged!

illusional:
Thanks @danking! I had trouble merging from your branch, but I've applied the diff and resolved conflicts with main in this branch. Hopefully this is it!!

illusional:
@danking, I still see red cross for CI

PR #10090, opened 2021-02-22
[batch] Flexible storage configuration

jigold:
This is a newer version of #9598.

We decided to give users min(5Gi/core, 5 Gi) in `/` with mounting external storage at `/io` if they need more storage. All storage requests can be 0, 0 < storage < 10 will be rounded up to 10 Gi, or 10+Gi rounded up to the nearest integer.

I added a loop to remove orphaned disks in gce.py.

I changed how the resources appear in the spec. Now there's `req_cpu`, `req_storage`, `req_memory` which stores what the user specified. Then we also have `cores_mcpu`, `memory_bytes`, and `storage_gib` which are the actual resources allocated. I think this will be simpler and more understandable. Resources are computed in the front end now and the worker just uses the values from the front end (no more doing conversions on both the worker and front end). I kept backwards compatibility on the worker for now which can get deleted once there are no more jobs with batch format version < 6. I bumped the instance version to 16 so we know which workers have the new storage functionality. 

I tested this by submitting 4 jobs on my 1 core test instance with 150Gi requests. I then looked at the worker logs to make sure the disks were created correctly and the value of the semaphore was correct. I also tested 0 Gi and 5 Gi by hand to make sure the resource fulfilled was 0Gi and 10Gi respectively. Lastly, I checked the billing to make sure we charged for the fraction of the SSD used as well as the cost of adding an extra persistent SSD for that job. I also looked at the disks in the GCE console to make sure they wear torn down correctly.

Although there isn't a migration, we should make sure there are no non-ci jobs running so that we don't over allocate the storage available. Also, once this is merged, we should send an email to all users to let them know the cores must be a power of 2 now and about the storage now being mounted at '/io`. I put the WIP tag on so I can do this when I'm ready to.


jigold:
This is passing except for test_query

batch/batch/driver/zone_monitor.py
```diff
@@ -117,6 +118,10 @@ async def update_region_quotas(self):
             for name in BATCH_GCP_REGIONS
         }
 
+        self.zones = [url_basename(z)
+                      for r in self.region_info.values()
+                      for z in r['zones']]
+
         log.info('updated region quotas')
```
danking, line 125: For a follow up PR: let's fix this terminology, this is no longer about quotas, it's about tracking the zones present in each region.

batch/batch/driver/main.py
```diff
@@ -425,6 +425,15 @@ async def pool_config_update(request, userdata):  # pylint: disable=unused-argum
...
         raise web.HTTPFound(deploy_config.external_url('batch-driver', pool_url_path))
 
+    if not worker_local_ssd_data_disk:
+        unreserved_disk_storage_gb = unreserved_worker_data_disk_size_gib(worker_local_ssd_data_disk, worker_pd_ssd_data_disk_size_gb, worker_cores)
+        if unreserved_disk_storage_gb < 0:
+            min_disk_storage = worker_pd_ssd_data_disk_size_gb - unreserved_disk_storage_gb
+            set_message(session,
+                        f'PD SSD must be at least {min_disk_storage} GB',
+                        'error')
+            raise web.HTTPFound(deploy_config.external_url('batch-driver', '/'))
```
danking, line 435: should this redirect to the `pool_url_path`?

batch/batch/disk.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class Disk:
+    @staticmethod
+    async def delete_disk(compute_client, name, zone):
+        log.info(f'deleting disk {name}')
+        resp = None
+        delay = 0.2
+        delete_uuid = str(uuid.uuid4())
+        while resp is None or resp['status'] != 'DONE':  # pylint: disable=unsubscriptable-object
```
danking, line 16: a bit tighter:
```
status = None
delay = 0.2
delete_uuid = str(uuid.uuid4())
while status != 'DONE':
    resp = await compute_client.delete(...)
    status = resp['status']
    ...
```

batch/batch/disk.py
```diff
@@ -0,0 +1,120 @@
...
+            resp = await compute_client.delete(
+                f'/zones/{zone}/disks/{name}',
+                params={'requestId': delete_uuid})
+            delay = await sleep_and_backoff(delay)
```
danking, line 20: This adds an unnecessary sleep after a successful delete, maybe this instead:

```
delay = 0.2
delete_uuid = str(uuid.uuid4())
while True:
    resp = await compute_client.delete(
        f'/zones/{zone}/disks/{name}',
        params={'requestId': delete_uuid})
    if resp['status'] == 'DONE':
        break
    delay = await sleep_and_backoff(delay)
```

batch/batch/disk.py
```diff
@@ -0,0 +1,120 @@
...
+        await self._attach()
+        await self._format()
+
+    async def delete(self):
+        await self._detach()
+        await self._delete()
+
+    async def _format(self):
+        code = f'''
+mkfs.ext4 -m 0 -E lazy_itable_init=0,lazy_journal_init=0,discard {self.disk_path}
```
danking, line 55: For my own education, can you share what these flags mean and why you chose these settings?

batch/batch/disk.py
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        resp = None
+        delay = 0.2
+        create_uuid = str(uuid.uuid4())
+        while resp is None or resp['status'] != 'DONE':  # pylint: disable=unsubscriptable-object
+            resp = await self.compute_client.post(f'/zones/{self.zone}/disks',
+                                                  json=config,
+                                                  params={'requestId': create_uuid})
+            delay = await sleep_and_backoff(delay)
```
danking, line 81: same issue here with respect to a sleep after success.

batch/batch/disk.py
```diff
@@ -0,0 +1,120 @@
...
+
+        resp = None
+        delay = 0.2
+        attach_uuid = str(uuid.uuid4())
+        while resp is None or resp['status'] != 'DONE':  # pylint: disable=unsubscriptable-object
+            resp = await self.compute_client.post(
+                f'/zones/{self.zone}/instances/{self.instance_name}/attachDisk',
+                json=config,
+                params={'requestId': attach_uuid})
+            delay = await sleep_and_backoff(delay)
```
danking, line 100: I think we should enrich the BaseClient with methods that handle this kind of operation. I'm not quite sure what the interface should be, but I think we should nip this repetitive code in the bud.

danking (changes requested):
Awesome work! Just a few comments

batch/test/test_batch.py
```diff
@@ -155,16 +161,38 @@ def test_out_of_memory(client):
...
+    builder.submit()
+    status = j.wait()
+    assert status['state'] == 'Success', str(status)
+
+
+def test_attached_disk(client):
+    builder = client.create_batch()
+    resources = {'cpu': '0.25', 'memory': '10M', 'storage': '400Gi'}
+    j = builder.create_job('ubuntu:18.04',
+                           ['/bin/sh', '-c', 'fallocate -l 350GiB /io/foo'],
```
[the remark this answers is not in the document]
jigold: I was worried that there's overhead for the disk and you can't actually allocate exactly 400Gi. I made it 395Gi.

batch/batch/worker/worker.py
```diff
@@ -931,18 +985,22 @@ async def run(self, worker):
...
+
                 async with Flock('/xfsquota/projid', pool=worker.pool):
                     with open('/xfsquota/projid', 'a') as f:
                         f.write(f'{self.project_name}:{self.project_id}\n')
 
-                async with Flock('/xfsquota/projects', pool=worker.pool):
-                    with open('/xfsquota/projects', 'a') as f:
-                        f.write(f'{self.project_id}:{self.scratch}\n')
+                if not self.disk:
+                    async with Flock('/xfsquota/projects', pool=worker.pool):
```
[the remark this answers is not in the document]
jigold: I switch all the commands to be synchronous. That should fix the need for Flock.

batch/batch/worker/worker.py
```diff
@@ -931,18 +985,22 @@ async def run(self, worker):
...
+                        with open('/xfsquota/projects', 'a') as f:
+                            f.write(f'{self.project_id}:{self.scratch}\n')
```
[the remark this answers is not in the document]
jigold: Good catch!

batch/batch/disk.py
jigold: Good catch!

batch/batch/disk.py
jigold: They're from the GCE disk tutorial on how to attach and format a new disk. I didn't actually look into if these parameters are optimal.

hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -1,5 +1,26 @@
...
+    request_uuid = str(uuid.uuid4())
+    if 'requestId' not in params:
+        params['requestId'] = request_uuid
+
+    delay = 0.2
+    while True:
+        resp = await request_f(path, params=params, **kwargs)
+        if resp['status'] == 'DONE':
+            return resp
+        delay = await sleep_and_backoff(delay)
```
danking, line 23: love it, very nice.

hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -47,6 +68,19 @@ def __init__(self, project, **kwargs):
...
+        return await request_with_wait_for_done(self.post, path, params, **kwargs)
+
+    async def attach_disk(self, path: str, *, params: Mapping[str, Any] = None, **kwargs):
+        return await request_with_wait_for_done(self.post, path, params, **kwargs)
+
+    async def detach_disk(self, path: str, *, params: Mapping[str, Any] = None, **kwargs):
+        return await request_with_wait_for_done(self.post, path, params, **kwargs)
+
+    async def delete_disk(self, path: str, *, params: Mapping[str, Any] = None, **kwargs):
+        return await request_with_wait_for_done(self.delete, path, params, **kwargs)
```
danking, line 86: I like this a lot more!

hail/python/hailtop/batch/docs/service.rst
```diff
@@ -125,30 +125,44 @@ machines with 10 GB of persistent SSD boot disk and 375 GB of local SSD. The cos
...
+    the worker. Currently, each 1 core requested gets 3.75 Gi of memory for standard worker types. Therefore, if
+    a user requests 1 CPU and 7 Gi of memory, the user will get 2 cores for their job and will be billed for 2 cores.
 
 .. note::
 
-    The amount of CPU reserved for a job is rounded up to powers of two with a minimum of 0.25 cores.
-    For example, a job requesting 5 cores will be rounded up to 8 cores.
+    The default storage per job is 5 Gi per core requested up to a maximum of 5 Gi. If you need more storage than this,
+    you can request more storage explicitly with the :meth:`.Job.storage` method. The minimum storage request is 10 GB
+    which can be incremented in units of 1 GB maxing out at 64 TB.
```
danking, line 165: Add a note about fractional-core jobs

hail/python/hailtop/batch/job.py
```diff
@@ -348,6 +348,17 @@ def storage(self, storage: Union[str, int]) -> 'Job':
         *G*, *Gi*, *T*, *Ti*, *P*, and *Pi*. Omitting a suffix means
         the value is in bytes.
 
+        For the :class:`.ServiceBackend`, by default every job gets 5 Gi
+        per core requested for the root file system `/` with a maximum of 5 Gi.
```
danking, line 352: I think this is less confusing as: "jobs requesting one or more cores receive 5 GiB of storage for the root file system `/`. Jobs requesting a fraction of a core receive the same fraction of 5 GiB of storage."

danking (dismissed):
I think we're good, just some thoughts about docs

jigold:
This is passing except for the service python tests

batch/batch/driver/create_instance.py
```diff
@@ -249,21 +262,26 @@ async def create_instance(app, zone, machine_name, machine_type, activation_toke
...
 -v /logs:/logs \
 -v /gcsfuse:/gcsfuse:shared \
 -v /xfsquota:/xfsquota \
 --mount type=bind,source=/mnt/disks/$WORKER_DATA_DISK_NAME,target=/host \
+--mount type=bind,source=/dev,target=/dev,bind-propagation=rshared \
 -p 5000:5000 \
 --device /dev/fuse \
 --device $XFS_DEVICE \
+--device /dev \
+--privileged \
```
danking, line 284: Are we sure privileged is still necessary?

danking (changes requested):
I'm roughly OK with this, but a question about privileged

batch/batch/driver/create_instance.py
jigold: I'll try again, but I think so. There must be other things privileged does that are not capabilities. I think it might be something with the namespace.

jigold:
Can you double check the code in `worker.py`. The rebase for that was pretty terrible. I have the WIP tag on -- please keep that on for now.

danking (changes requested):
Remind me why we changed all these shell calls to be synchronous?

jigold:
I thought we changed them because they didn't actually need a flock if the code runs synchronously and none of these operations takes a long time. I can switch them back if you think it will make a difference. I think the sed commands are definitely short. I'm not sure how long the xfs quota commands take.

danking (dismissed):
sorry for back-and-forth, let's leave the Flock stuff unchanged. I'm not sure I fully understand the implications of the change yet

danking:
erm, by which I mean, can we keep the code the way it is in main for now and revisit flock separately?

danking (approved):

PR #10091, opened 2021-02-22
[query] Reorganize code orderings

tpoterba:


tpoterba:
Sorry to drop another big one on you, Chris, but you have the context for this. 

The big diff doesn't reflect the size of the conceptual change here. This PR does as follows:
* Breaks up and moves CodeOrdering to its own package under `ir`. 
* Changes all CodeOrdering factories to take STypes instead of PTypes
* Removes some factory methods, like `pType.codeOrdering`.
* There are now two places to construct orderings: `EmitClassBuilder.getOrderingFunction` (which takes stypes and an `Op`), and `CodeOrdering.makeOrdering`.
* I slightly changed the method layout to memoize within each ordering. We check the types on each method call.
* There are only a couple meaningful changes in the typed ordering generators. I changed CallOrdering and StringOrdering a bit to more clearly describe what they're doing.

tpoterba:
Bump. This is blocking other SCode work.

chrisvittal (approved):
I've said my piece about how too much of the code is under the `expr.ir` package. I'll reorganize that (eventually). Otherwise, this is good.

PR #10092, opened 2021-02-22
[all] fix many incorrect uses of sed -i

danking:
If you do not include an empty argument to `sed` `-i`, it will use *whatever* the next argument is as the
file extension. As such, we have lots of `html-e` files in the docs. This fixes it by properly providing
an empty extension to `-i`.


danking:
Ah, so this is a place where BSD differs from linux. In linux you need `-i` and in BSD you need `-i ''`. I just fixed the makefile since that's where we might be on OS X.

tpoterba (approved):

PR #10094, opened 2021-02-22
[query] Remove Region.{load,store}Primitive

chrisvittal:


chrisvittal:
Let's see how CI likes it.

hail/src/main/scala/is/hail/expr/ir/agg/AggregatorState.scala
```diff
@@ -203,41 +196,38 @@ class PrimitiveRVAState(val vtypes: Array[VirtualTypeWithReq], val kb: EmitClass
...
-            cb += storageType.setFieldPresent(dest, i)
-            cb += Region.storePrimitive(t, storageType.fieldOffset(dest, i))(v)
-          })
-    })
+    // everything is primitives, so this is fine! :)
+    val r = new Value[Region] {
+      def get: Code[Region] = Code._null[Region]
+    }
+    val dest = cb.newLocal("prim_rvastate_store_dest", destc)
+    storageType.storeAtAddressFromFields(cb, dest, r, fields.map(_.load), false)
```
tpoterba, line 204: can you just pass a `null` `Code[Region]` instead of a wrapped null?

tpoterba (changes requested):
one drive by comment before full review. These changes are a breath of fresh air!

hail/src/main/scala/is/hail/expr/ir/agg/AggregatorState.scala
chrisvittal: We'll try it.

chrisvittal:
Scala tests pass. Gotta unbreak the python tests.

hail/src/main/scala/is/hail/expr/ir/agg/AggregatorState.scala
chrisvittal: seems to work

tpoterba:
nice! can you run benchmarks?

tpoterba (approved):
🎉 🎉 🎉

PR #10097, opened 2021-02-23
[batch] Compute optimal worker type based on cpu / memory configuration

jigold:
Stacked on #10090. This PR implements our design where you can specify the worker type you want to run on with the strings 'lowmem', 'standard', and 'highmem'. If you specify the memory in bytes, then we compute the optimal worker type for you and round the cores if needed.


jigold:
I put the WIP tag on this so we can control merging.

hail/python/hailtop/utils/validate/validate.py
```diff
@@ -137,6 +140,23 @@ def validate(self, name: str, obj):
...
+    def validate(self, name: str, obj):
+        excs = []
+        for checker in self.checkers:
+            try:
+                return checker.validate(name, obj)
+            except ValidationError as e:
+                excs.append(e)
+        if excs:
+            reasons = ' or '.join([e.reason for e in excs])
+            log.info(reasons)
```
danking, line 156: none of the other validators log, so I'm inclined not to log here either

hail/python/hailtop/utils/validate/validate.py
```diff
@@ -178,3 +198,7 @@ def numeric(**conditions: Callable[[Any], Any]):
 
 def switch(key: str, checkers: Dict[str, Dict[Key, Validator]]):
     return SwitchValidator(key, checkers)
+
+
+def anyof(*checkers: Validator):
```
danking, line 203: nice addition; I wonder if we can think of words that are more obviously different from one another than "oneof" and "anyof". Maybe "isoneof" and "anyof"? The "is" makes me think of python's `is`.

batch/batch/worker_config.py
```diff
@@ -83,6 +87,49 @@ def from_instance_config(instance_config, job_private=False):
...
+            'size': size,
+            'image': None
+        })
+
+        config = {
+            'version': WORKER_CONFIG_VERSION,
+            'instance': {
+                'project': None,
+                'zone': None,
+                'family': 'n1',  # FIXME: need to figure out how to handle variable family types
```
danking, line 122: This is just because we don't yet support any family other than n1, right?

batch/batch/inst_coll_config.py
```diff
@@ -74,6 +81,13 @@ def convert_requests_to_resources(self, cores_mcpu, memory_bytes, storage_bytes)
 
         return None
 
+    def cost_per_hour(self, resource_rates, cores_mcpu, memory_bytes, storage_gib):
+        cost_per_hour = self.worker_config.cost_per_hour(resource_rates,
+                                                         cores_mcpu,
+                                                         memory_bytes,
+                                                         storage_gib)
+        return cost_per_hour
```
danking, line 89: let's skip the variable assignment and just return

danking (commented):
looks copacetic to me; I'll approve once the underlying one merges.

danking (approved):
good with me

PR #10100, opened 2021-02-24
[query-service] get tests working

danking:
One more time, with feeling! (was: #10072)

- [ ] (@tpoterba)                        caf1e1e673 add fails_service_backend
- [ ] (@tpoterba, @cseed)                62dbe7d62e [hail] introduce and use mktemp and mktempd
- [ ] (@tpoterba)                        206707ca9c [hail] make is.hail.expr.ir.functions threadsafe
- [ ] (@tpoterba)                        b4150fffa0 [hail] fix use of row requiredness in lowerDistributedSort
- [ ] (@catoverdrive)                    d2937a9535 [query-service] handle void-typed IRs in query-service
- [ ] (@catoverdrive)                    ebe6684548 [query-service] make user cache thread-safe
- [ ] (@tpoterba)                        8b59b88516 [query-service] bugfix: preserve globals through a shuffle
- [ ] (@catoverdrive)                    0e391e40d3 [shuffler] log ShuffleCodecSpec anytime it is created
- [ ] (@daniel-goldstein)                a7350e7158 [scala-lsm] bugfix: least key may equal greatest key
- [ ] (@daniel-goldstein)                dc71bcda46 [services] discovered new transient error
- [ ] (@daniel-goldstein)                0543d01782 [shuffler] more assertions in ShuffleClient
- [ ] (@daniel-goldstein)                c3d51770bb [shuffler] bugfix: shuffler needs a HailContext to decode loci
- [ ] (@daniel-goldstein)                dda1154152 [query-service] move hail.jar earlier in Dockerfile
- [ ] (@daniel-goldstein)                7fd96cefde [query-service] permit pod scaling and remove cpu limit
- [ ] (@catoverdrive)                    71e996acd1 [query-service] simplify socket handling
- [ ] (@jigold)                          395f97f424 [batch] teach JVMJob where to find the hail configuration files
- [ ] (@daniel-goldstein)                ed0e3cb6f7 [query-service] switch to services team approved logging
- [ ] (@tpoterba)                        44d1c8b456 [query-service] query workers need a hail context
- [ ] (@daniel-goldstein, @catoverdrive) ceec7208a0 [query-service] use a UNIX Domain Socket for Py-Scala communication
- [ ] (@daniel-goldstein, @catoverdrive) 21d2810d72 [query-service] run tests against query service
- [ ] (@jigold)                          644e709680 [query-service] aiohttp.ClientSession must be created in async code
- [ ] (@cseed)                           fd72cf5995 [query-service][hail][build.yaml] address miscellaneous comments from cotton
- [ ] (@daniel-goldstein)                13ab77da83 [query-service] use two containers sharing an empty volume
- [ ] (@daniel-goldstein)                96d00aea97 [query-service] in the service, log everything to stdout
- [ ] (trivial)                          6fd10a4bf0 [query-service] do not catch CancelledError
- [ ] (trivial)                          deaa3ffc6e [query-service] slightly more useful error message when socket dies
- [ ] (@tpoterba)                        2c9abcdfc4 [query] cleanup execute contexts
- [ ] (@tpoterba)                        0a00cbde33 [shuffler] if we have an ExecuteContext, use it


cseed (dismissed):
See my comments on the tempfile commit.

build.yaml
```diff
@@ -781,6 +806,7 @@ steps:
       mountPath: /test-gsa-key
    dependsOn:
     - default_ns
+    - upload_test_resources_to_gcs
```
cseed, line 809: The other reason I prefer this to a global token is that getting the token requires you to write down the necessary dependency.

config.mk
```diff
@@ -1,6 +1,6 @@
-PROJECT := hail-vdc-staging
+PROJECT := hail-vdc
 DOCKER_ROOT_IMAGE := gcr.io/$(PROJECT)/ubuntu:18.04
-DOMAIN := staging.hail.is
+DOMAIN := hail.is
 INTERNAL_IP := 10.128.0.2
 IP := 35.224.188.20
 KUBERNETES_SERVER_URL := https://34.123.185.11
```
cseed, line 6: See my other comment, all these settings are wrong.

hail/python/hail/backend/service_backend.py
```diff
@@ -1,32 +1,43 @@
...
-        self.session = aiohttp.ClientSession(headers=service_auth_headers(deploy_config, 'query'))
+        self._session: Optional[aiohttp.ClientSession] = None
+
+    async def session(self) -> aiohttp.ClientSession:
+        if self._session is None:
+            self._session = aiohttp.ClientSession(
+                headers=service_auth_headers(self.deploy_config, 'query'))
+        return self._session
 
     def close(self):
```
cseed, line 37: So I think the right way to do this is to make `ServiceSocket` an (async) context manager.  Then you can construct the session in `__aenter__` and clean it up in `__aexit__`.

hail/python/hail/backend/service_backend.py
```diff
@@ -37,12 +48,22 @@ def handle_response(self, resp):
         return resp.data
 
     async def async_request(self, endpoint, **data):
-        async with self.session.ws_connect(f'{self.url}/api/v1alpha/{endpoint}') as socket:
+        data['token'] = secret_alnum_string(n=32)
```
cseed, line 51: Why 32?

hail/python/hail/context.py
```diff
@@ -381,7 +392,106 @@ def tmp_dir():
...
+        self.name = name
+
+    def __enter__(self):
+        return self.name
+
+    def __exit__(self, type, value, traceback):
+        return self.fs.remove(self.name)
+
+
+def TemporaryFilename(*,
```
cseed, line 407: Why not make this a class which is a context manager?

hail/src/main/scala/is/hail/annotations/RegionPool.scala
```diff
@@ -138,7 +138,14 @@ final class RegionPool private(strictMemoryCheck: Boolean, threadName: String, t
 
   override def finalize(): Unit = close()
 
+  private[this] var closed: Boolean = false
+
   def close(): Unit = {
+    if (closed) {
+      return
+    }
```
cseed, line 146: Curly braces unnecessary.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -139,25 +163,18 @@ class User(
...
-  def addUser(username: String, key: String): Unit = {
-    assert(!users.contains(username))
-    users += username -> new User(username, "/tmp", new GoogleStorageFS(key))
-  }
-
-  def removeUser(username: String): Unit = {
-    assert(users.contains(username))
-    users -= username
+  def addUser(username: String, key: String): Unit = synchronized {
+    users.put(username, new User(username, "/tmp", new GoogleStorageFS(key)))
```
cseed, line 172: You're overwriting a previous user if it exists?

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -472,13 +470,308 @@ class ServiceBackend() extends Backend {
...
+
+class ServiceBackendSocketAPI(backend: ServiceBackend, socket: Socket) extends Thread {
+  import ServiceBackendSocketAPI._
+
+  private[this] val LOAD_REFERENCES_FROM_DATASET = 1
+  private[this] val VALUE_TYPE = 2
+  private[this] val TABLE_TYPE = 3
+  private[this] val MATRIX_TABLE_TYPE = 4
+  private[this] val BLOCK_MATRIX_TYPE = 5
+  private[this] val REFERENCE_GENOME = 6
```
cseed, line 496: I'd be fine for the moment if the service only worked with builtin references.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -472,13 +470,308 @@ class ServiceBackend() extends Backend {
...
+  }
+
+  override def run(): Unit = {
+    try {
+      eventLoop()
+    } catch {
+      case t: Throwable =>
+        log.info("ServiceBackendSocketAPI caught exception", t)
+    } finally {
+      socket.close()
```
cseed, line 741: You didn't want to retry here?

query/deployment.yaml
```diff
@@ -83,3 +132,32 @@ spec:
          secret:
            optional: false
            secretName: ssl-config-query
+{% if not deploy %}
```
cseed, line 135: I'm fine if you just make this unconditional.  Nobody is going to be using it seriously and I feel like it is going to be changing quickly.

query/query/query.py
```diff
@@ -1,65 +1,44 @@
...
 from hailtop.hail_logging import AccessLogger
 from gear import setup_aiohttp_session, rest_authenticated_users_only, rest_authenticated_developers_only
 
+from .sockets import ServiceBackendSocketSession, ServiceBackendJavaConnector
+
 uvloop.install()
 
 DEFAULT_NAMESPACE = os.environ['HAIL_DEFAULT_NAMESPACE']
-log = logging.getLogger('batch')
+log = logging.getLogger('query.query')
```
cseed, line 23: I feel like we should just use `__name__` here.

cseed (changes requested):

batch/sql/increase-test-and-dev-pool-sizes.py
```diff
@@ -0,0 +1,31 @@
+import os
+import asyncio
+from gear import Database
+
+
+async def main():
```
cseed, line 6: I wonder if this is better as a build.yaml step instead of a migration. Then we can just change it (again) if we want different settings, instead of having to apply another whole migration.

hail/src/main/scala/is/hail/types/Box.scala
```diff
@@ -1,19 +1,25 @@
 package is.hail.types
 
-final case class Box[T](var b: Option[T] = None, matchCond: (T, T) => Boolean = { (a: T, b: T) => a == b }) {
-  def unify(t: T): Boolean = b match {
+import java.util.function._
+
+final case class Box[T](
+  b: ThreadLocal[Option[T]] = ThreadLocal.withInitial(
+    new Supplier[Option[T]] { def get = None }),
```
cseed, line 7: Ugh, this is real gross.  I guess the alternative is for `unify` to take and populate a `mutable.Map[TVariable, Type]`.  I guess I prefer that, but this should be fine as a first step.

query/deployment.yaml
```diff
@@ -60,19 +60,68 @@ spec:
...
+           - name: HAIL_DEPLOY_CONFIG_FILE
+             value: /deploy-config/deploy-config.json
+           - name: HAIL_DEFAULT_NAMESPACE
+             value: "{{ default_ns.name }}"
+           - name: HAIL_SHA
+             value: "{{ code.sha }}"
+           - name: HAIL_QUERY_WORKER_IMAGE
+             value: {{ query_image.image }}
+{% if deploy %}
+           - name: HAIL_QUERY_BUCKET
```
cseed, line 98: I don't think this is used anymore?

query/query/sockets.py
```diff
@@ -0,0 +1,265 @@
...
+            time.sleep(1)
+
+    def connect(self) -> 'ServiceBackendSocketSession':
+        sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        sync_retry_transient_errors(sock.connect, self.fname)
+        return ServiceBackendSocketSession(
+            ServiceBackendSocketAPI(sock))
+
+
+class ServiceBackendSocketSession:
```
cseed, line 26: I feel like you don't need the connector or the session, just the API (which I'd call a connection?)  It can connect to the socket on connect, and make existence of the socket file a transient error?

cseed (changes requested):

hail/python/hail/backend/service_backend.py
danking: I removed it

hail/python/hail/backend/service_backend.py
danking: I agree this class is kind of a mess, but I'm not sure a context manager helps? We're trying to share one Session for the entire lifetime of the service backend (many hail queries). If I change to a CM, I'd just __aenter__ it on line 78 / 99 in `class ServiceBackend`.

hail/src/main/scala/is/hail/types/Box.scala
danking: I'll have you take this up with the compiler's team. 📞  @tpoterba 

(Tim: I managed to avoid making the entire function registry thread local but just making the Box's `b` thread local, but Cotton is still disgusted. 😉 )

hail/python/hail/context.py
cseed: Yeah, I guess I don't see why that code can't be in the `__init__` of the wrapper.  Well, doesn't really matter, your choice.

query/deployment.yaml
danking: Fixed in 8f8d7ddb89

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
cseed: Ah, right, got it.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
cseed: Ah, OK, I see now.

query/query/sockets.py
danking: I don't grok your comment about transient errors, but I made the rest of the changes in b81b8fae32

danking:
@cseed erm, I rebased so I screwed up all the SHAs that I commented, sorry. The SHAs still have the relevant changes

danking:
@daniel-goldstein I believe I addressed all of your comments.

danking:
@tpoterba @catoverdrive @jigold buuump

batch/sql/increase-test-and-dev-pool-sizes.py
cseed: > I could've taken a day to write and test pool APIs and create a proper PR for that

I just meant directly modify the database in a build.yaml step by mounting the database credentials.  You don't need an API or anything.

cseed (approved):

danking:
WIP because it has a migration and I gotta step out

PR #10101, opened 2021-02-24
[batch] Expose idempotency token

jigold:


danking (approved):
Awesome thank you Jackie! cc: @tpoterba, you can use the token parameter to `create_batch` to ensure that the batch is idempotently created.

PR #10103, opened 2021-02-24
[libhail] Mac compilation fix and small array example

johnc1231:
This PR:

- Adds `format1` calls for `long long` and `unsigned long long`. 
- Fixes bug with storing doubles
- Fixes bug where all arrays, tuples, and strings were believed missing by default


libhail/src/hail/main.cpp
```diff
@@ -73,5 +75,20 @@ main() {
...
 
+  {
+    print("Array testing:");
+    auto region = std::make_shared<ArenaAllocator>(heap);
+    auto varray = cast<VArray>(tc.get_vtype(tc.tarray(tc.tfloat64)));
+
+    int array_length = 8;
+    auto my_array = Value::make_array(varray, region, array_length);
+    print("array_length = ", my_array.get_size());
+    for(int i = 0; i < array_length; i++) {
```
cseed, line 86: Space after for (in general after keywords)

cseed, line 86: ++i is better than i++

cseed (dismissed):

johnc1231:
Only weird thing now is that the array of doubles is printing like it's an array of ints, truncating the numbers.

cseed:
That's because this is wrong:

```
void
format1(FormatStream &s, double v) {
  write_with_iostream<unsigned long>(s, v);
}
```

johnc1231:
Ah, good catch. Should be good to merge now, that's fixed too.

cseed (approved):
Nice!

PR #10105, opened 2021-02-24
[query] frozendict

johnc1231:
CHANGELOG: Evaluating hail expressions in python will now return `frozendict`, an immutable dictionary type. Sets will return python's `frozenset`.

This is necessary because hail supports hashable dicts, but python does not. Same with sets.

### DIFF
--- a/hail/python/hail/utils/frozendict.py
+++ b/hail/python/hail/utils/frozendict.py
@@ -11,6 +11,10 @@ class frozendict(Mapping):
 
     >>> my_frozen_dict = hl.utils.frozendict({1:2, 7:5})
 
+    To get a normal python dictionary with the same elements from a `frozendict`:
+
+    >>> dict(frozendict({'a': 1, 'b': 2}))
+
     Note
     ----
     This object refers to the Python value returned by taking or collecting
@@ -35,3 +39,6 @@ class frozendict(Mapping):
 
     def __iter__(self):
         return iter(self.d)
+
+    def __repr__(self): 
+        return f'frozendict({self.d!r})'  
--- a/hail/python/hail/utils/frozendict.py
+++ b/hail/python/hail/utils/frozendict.py
@@ -4,6 +4,7 @@ from collections.abc import Mapping
 T = TypeVar("T")
 U = TypeVar("U")
 
+
 class frozendict(Mapping):
     """
     An object representing an immutable dictionary.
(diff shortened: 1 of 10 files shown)


tpoterba (changes requested):
can you add a test to ensure we can correctly impute the type of these when we do `hl.literal()` or something?

hail/python/hail/utils/frozendict.py
```diff
@@ -0,0 +1,37 @@
+from typing import TypeVar, Dict
+from collections.abc import Mapping
+
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+class frozendict(Mapping):
```
danking, line 8: I think Mapping can take type arguments `Mapping[T, U]`.

danking (changes requested):

hail/python/hail/utils/frozendict.py
```diff
@@ -0,0 +1,37 @@
...
+        return self.d[k]
+
+    def __hash__(self) -> int:
+        return hash(frozenset(self.items()))
+
+    def __len__(self) -> int:
+        return len(self.d)
+
+    def __iter__(self):
+        return iter(self.d)
```
danking, line 37: Can you add a repr:
```
def __repr__(self): 
    return f'frozendict({self.d!r})'                                                                                 
```

hail/python/hail/utils/frozendict.py
```diff
@@ -0,0 +1,37 @@
...
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+class frozendict(Mapping):
+    """
+    An object representing an immutable dictionary.
+
+    >>> my_frozen_dict = hl.utils.frozendict({1:2, 7:5})
+
```
danking, line 13: Can you give an example of converting a frozendict back to a normal dict?

```
dict(frozendict({'a': 1, 'b': 2}))
```

danking (changes requested):

hail/python/hail/utils/frozendict.py
johnc1231: `typing.Dict` supports this, but `typing.Mapping` doesn't seem to, unless I'm misreading. I'm new to the python type annotations stuff.

johnc1231:
@danking  @tpoterba  Done everything you both asked for.

tpoterba (approved):

PR #10107, opened 2021-02-25
Support Artifact Registry

vladsaveliev:
Support [Artifact Registry](https://cloud.google.com/artifact-registry) to store images.

It's useful for our deployment in Australia, as GCR does not support our continent. With AR, we would save on network egress.

The PR adds a `DOCKER_PREFIX` variable, which is passed along in the code together with `GCP_PROJECT` and others. To switch to AR, one would need to modify `DOCKER_PREFIX` in `config.mk`:

```
REGION := us-central1
DOCKER_PREFIX := gcr.io/$(PROJECT)
```

```
REGION := australia-southeast1
DOCKER_PREFIX := $(REGION)-docker.pkg.dev/$(PROJECT)/hail
```

Also, when making an initial deployment with Terraform, there is an extra variable `use_artifact_registry = false` that controls whether to use AR or GCR.


ci/ci/build.py
```diff
@@ -338,7 +341,7 @@ def build(self, batch, code, scope):
 
 time gcloud -q auth activate-service-account \
   --key-file=/secrets/gcr-push-service-account-key/gcr-push-service-account-key.json
-time gcloud -q auth configure-docker
+time gcloud -q auth configure-docker {docker_registry}
```
daniel-goldstein, line 344: yes!!

dev-docs/development_process.md
```diff
@@ -60,6 +60,8 @@ gcloud auth login
 gcloud config set project hail-vdc
 gcloud container clusters get-credentials vdc --zone=us-central1-a
 gcloud auth configure-docker
+# If you are using the Artifact Registry:
+# gcloud auth configure-docker $REGION-docker.pkg.dev
```
daniel-goldstein, line 64: Could these possibly be condensed into a phony make target that uses variables in config.mk?

ci/test/resources/build.yaml
```diff
@@ -7,7 +7,7 @@ steps:
    name: inline_image
    dockerFile:
      inline: |
-       FROM gcr.io/{{ global.project }}/ubuntu:18.04
+       FROM {{ global.region }}-docker.pkg.dev/{{ global.project }}/hail/ubuntu:18.04
```
daniel-goldstein, line 10: how come this differs from everything else with `docker_prefix`?

benchmark/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM gcr.io/{{ global.project }}/ubuntu:18.04
+FROM {{ global.docker_prefix }}/ubuntu:18.04
```
daniel-goldstein, line 1: Could this use `DOCKER_ROOT_IMAGE`?

batch/build-batch-worker-image-startup.sh
```diff
@@ -39,10 +39,12 @@ curl -fsSL "https://github.com/GoogleCloudPlatform/docker-credential-gcr/release
...
 export HOME=/root
-docker-credential-gcr configure-docker
 
-GCP_PROJECT=$(curl -s -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/project/project-id")
-docker pull gcr.io/$GCP_PROJECT/ubuntu:18.04
+DOCKER_PREFIX=$(curl -s -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/attributes/docker_prefix")
+
+docker-credential-gcr configure-docker --include-artifact-registry
+
+docker pull ${DOCKER_PREFIX}/ubuntu:18.04
```
daniel-goldstein, line 47: Could this use `DOCKER_ROOT_IMAGE`?

daniel-goldstein (changes requested):
This is amazing, thanks @vladsaveliev! I've done a light pass with some simple questions but will leave a deeper review to @cseed specifically for the infra/terraform changes.

infra/main.tf
```diff
@@ -19,6 +19,19 @@ variable "batch_logs_bucket_storage_class" {}
 variable "gcp_region" {}
 variable "gcp_zone" {}
 variable "domain" {}
+variable "use_artifact_registry" { 
+  type = bool
+  description = "pull the ubuntu image from Artifact Registry. Otherwise, GCR"
```
cseed, line 24: I feel like, rather than have this additional configuration, we should also switch to the artifact registry.  I'm happy for this to go in as is, but we should consider that @daniel-goldstein.

infra/README.md
```diff
@@ -179,7 +185,8 @@ You can now install Hail:
 
   ```
   cd $HAIL
-  export HAIL_CI_UTILS_IMAGE=gcr.io/<gcp-project>/ci-utils:latest
+  export HAIL_DOCKER_PREFIX=gcr.io/<gcp-project>
```
cseed, line 188: Do you want to put another comment here (and below) about using the AR?

cseed (changes requested):
The Terraform changes look good to me.  Otherwise, I noticed a couple of small things.

ci/test/resources/build.yaml
vladsaveliev: My bad!

benchmark/Dockerfile
vladsaveliev: Changed to DOCKER_ROOT_IMAGE, plus in a few other places where ubuntu:18.04 is used and it made sense to me.

batch/build-batch-worker-image-startup.sh
vladsaveliev: Actually `build-batch-worker-image-startup.sh` doesn't have access to DOCKER_PREFIX or DOCKER_ROOT_IMAGE. I switched to applying `jinja2_render.py` to insert them into the script.

dev-docs/development_process.md
vladsaveliev: Something like this?
https://github.com/hail-is/hail/pull/10107/commits/a3c892f21bf84a16ae72f99c40dac17445a650de

Though it required to add `include ../config.mk` into `hail/Makefile` - not sure if this will break any assumption?

Also not sure if it makes sense to parameterize the cluster name (`vdc`). We didn't override it for us, but curious about the general case.

infra/README.md
vladsaveliev: Added!

daniel-goldstein (approved):

hail/Makefile
```diff
@@ -372,3 +374,11 @@ update-hail-repl: NAMESPACE ?= default
 update-hail-repl: wheel
 	kubectl -n $(NAMESPACE) cp $(WHEEL) $$(kubectl get pods -n $(NAMESPACE) -l app=hail-repl | tail -n +2 | awk '{print $$1}'):.
 	kubectl -n $(NAMESPACE) exec -it $$(kubectl get pods -n $(NAMESPACE) -l app=hail-repl | tail -n +2 | awk '{print $$1}') -- pip3 install -U hail-$(HAIL_PIP_VERSION)-py3-none-any.whl
+
+.PHONY: configure-gcloud
+configure-gcloud:
+	gcloud config set project $(PROJECT)
+	gcloud container clusters get-credentials vdc --zone=$(ZONE)
+	docker_prefix=$(DOCKER_PREFIX)
+	docker_registry=${DOCKER_PREFIX%%/*}
```
cseed, line 383: These should be upper case?  In either case, the definition and use of DOCKER_PREFIX should agree.

cseed (changes requested):
Found one minor thing.  Otherwise, looks good.

hail/Makefile
vladsaveliev: Changed to the uppercase!

cseed (approved):

vladsaveliev:
Can't see why the tests are failing :( Perhaps it's about time we set up ci-test for our deployment.

daniel-goldstein:
hm so the CI output is reading that `'dict object' has no attribute 'docker_root_image'` when filling in jinja templates with `{{ global.docker_root_image }}`, though it looks like you've set it correctly in `input_config` afaik

cseed:
So the problem is probably that CI is running on master, and it creates the global config in the PR namespace and it won't have docker_root_image.

We'll also need to update our production global-config since we're not applying Terraform updates to our cluster.  You should be able to do this now.

vladsaveliev:
Right, we had to manually re-deploying CI for our setup as well.

Pinging @daniel-goldstein just in case :)

daniel-goldstein:
Hi @vladsaveliev! Sorry for the delay on this but I think we're just about done. If you merge my changes from #10347 this should pass! (sorry for walking back the gcloud make target but I'm going to move that into the dev functions in a separate pr)

batch/batch/worker/worker.py
```diff
@@ -260,9 +262,9 @@ def __init__(self, job, name, spec):
             image_ref.tag = 'latest'
 
         if image_ref.name() in HAIL_GENETICS_IMAGES:
-            image_ref.domain = 'gcr.io'
-            image_ref.path = f'{PROJECT}/{image_ref.name()}'
-
+            image_ref.domain = DOCKER_PREFIX.split('/')[0]
+            image_ref.path = '/'.join(DOCKER_PREFIX.split('/')[1:] + [image_ref.path])
+
```
daniel-goldstein, line 267: extra white space is failing style checks

daniel-goldstein (changes requested):

batch/batch/worker/worker.py
vladsaveliev: Argh, an artefact of resolving conflicts with the GitHub UI :) Fixed.

daniel-goldstein (approved):

daniel-goldstein:
Busy day today. I've fixed conflicts on #10347

daniel-goldstein:
Closed in favor of #10347

PR #10110, opened 2021-02-25
[ci] bring over useful scorecard features to ci user page

daniel-goldstein:
CI user page now contains assigned PRs and random team member selection. These seem to me like the only features of scorecard people use and potentially eliminates the need for scorecard altogether. The CSS is a little crude but I'm hoping it's temporary, also happy to take suggestions.

### DIFF
--- a/ci/ci/ci.py
+++ b/ci/ci/ci.py
@@ -22,12 +22,13 @@ from gear import (
     check_csrf_token,
     create_database_pool,
 )
+import random
 from typing import Dict, Any, Optional
 from web_common import setup_aiohttp_jinja2, setup_common_static_routes, render_template, set_message
 
 from .environment import BUCKET
 from .github import Repo, FQBranch, WatchedBranch, UnwatchedBranch, MergeFailureBatch, PR
-from .constants import AUTHORIZED_USERS
+from .constants import AUTHORIZED_USERS, TEAMS
 
 with open(os.environ.get('HAIL_CI_OAUTH_TOKEN', 'oauth-token/oauth-token'), 'r') as f:
     oauth_token = f.read().strip()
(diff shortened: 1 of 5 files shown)


ci/ci/constants.py
```diff
@@ -1,32 +1,34 @@
...
+from typing import Optional, List
 
 GITHUB_CLONE_URL = 'https://github.com/'
 GITHUB_STATUS_CONTEXT = 'ci-test'
+TEAMS = ['Services', 'Compiler']
 
 
 class User:
-    def __init__(self, gh_username: str, hail_username: Optional[str]):
+    def __init__(self, gh_username: str, hail_username: Optional[str] = None, teams: List[str] = []):
```
danking, line 9: In situations like this, we tend to stick a `*` before `hail_username` to force the subsequent arguments to be keyword-only. I suppose in this case you'll never have a teams and not a hail_username, so this is probably the right choice actually.

ci/ci/ci.py
```diff
@@ -207,28 +209,38 @@ async def get_job(request, userdata):
...
+    team_random_member = {
+        team: random.choice([user for user in AUTHORIZED_USERS if team in user.teams]).gh_username for team in TEAMS
+    }
+
     page_context = {
         'username': username,
         'gh_username': gh_username,
-        'watched_branches': wbs,
+        'pr_wbs': pr_wbs,
+        'review_wbs': review_wbs,
```
danking, line 242: So this is great, but scorecard does something different. Scorecard tells you things that you need to take action on. That means:
- your PRs that are failing tests
- your PRs that have changes requested reviews
- PRs assigned to you that you have no extant review (e.g. it got dismissed, you haven't reviewed yet)

danking (approved):
Love it, big improvement. See my comment about, I guess, an "action items" list.

ci/ci/ci.py
daniel-goldstein: Ya this does take a different approach, but I think the information is still there. For any PR, it shows the build state and review state, so you can see which PRs are failing tests and which PRs have review pending / changes requested / approved for your PRs as well as those assigned to you.

daniel-goldstein: Or do you mean that it would be nice to eliminate PRs which don't require action?

PR #10114, opened 2021-02-26
[query-service] remove flaky test_query

danking:
My query service PR will replace these tests with a complete test suite anyway. No one relies on
query being correct. Let us stop interrupting PRs with flaky tests.


jigold (approved):

johnc1231:
@danking Erroring in CI

PR #10116, opened 2021-02-26
[query] Method-wrap NDArray constructor code

tpoterba:
```
linear_regression_rows_nd      35.8%    274.506     98.311
```


johnc1231 (approved):

johnc1231:
Great change. Still a little slower on my laptop than it was before reference counting (48 seconds vs 39 seconds), but way better than what's in main.

PR #10117, opened 2021-02-26
[k8s] set consistent resource requests for better node packing

daniel-goldstein:
Our GKE nodes come with 2 CPU and 7.5 GB each, but not all of that is allocatable to our pods. In reality, somewhere between 5.7-5.9GB ([GCP Docs](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture) say 5.7, GKE console says 5.9) of memory and 1.9CPU are available for us to use. Some of our Big services request 1 CPU and 3.75GB, but then we can never fit two big pods on one node. This is an attempt to standardize our requests so their easier to reason about while hopefully getting better packing.  

| resource | Big | Medium | Small |
| --- | --- | --- | --- |
| CPU        | 600m   | 100m | 20m |
| Memory  | 2G | 200M | 20M |

The intentions here are:
- always be able to comfortably get 2 Big pods on a node
- medium pods shouldn't have to force new nodes to spin up just because there's a Big pod there already
- small pods take up minimal resources
- there's ample room for small pods (which are mostly on HPA) to scale up considering most nodes shouldn't be at their medium pod capacity

The ratios don't match exactly, because I didn't want to assign CPU lower than 20m to prevent HPA thrashing we saw with auth and see a bit now with router. Setting it to 20m should hopefully convince k8s that idle small apps don't need to be scaled up under normal fluctuation.

## Big
- query
- batch-driver
- shuffler
- memory

## Medium
- grafana
- ukbb-browser
- ukbb-static
- blog
- ci
- internal-gateway

## Small
- amundsen
- router
- gateway
- site
- batch
- address
- atgu
- router-resolver
- ci/test statefulset & deployment
- auth-driver
- echo
- benchmark
- image-fetcher

### Fun surprises I found along the way
- CI test statefulsets and deployment are getting .5GB and .5 CPU each
- We run a lot of image fetchers because a daemon set gets added per PR namespace

EDIT: It seems that discrepancy between GCP docs and GKE console is the console counts kube-system pods in "Allocatable Memory", and it really does take 2GB to run the Kubernetes Engine 🙃

### DIFF
--- a/address/deployment.yaml
+++ b/address/deployment.yaml
@@ -50,11 +50,11 @@ spec:
          - containerPort: 5000
         resources:
           requests:
-            memory: "5M"
-            cpu: "5m"
+            cpu: "20m"
+            memory: "20M"
           limits:
-            memory: "1G"
             cpu: "1"
+            memory: "1G"
         env:
          - name: HAIL_DOMAIN
            value: "{{ global.domain }}"
--- a/amundsen/deployment.yaml
+++ b/amundsen/deployment.yaml
@@ -34,8 +34,11 @@ spec:
         imagePullPolicy: Always
         resources:
           requests:
-            memory: "1G"
+            cpu: "20m"
+            memory: "20M"
+          limits:
             cpu: "1"
+            memory: "1G"
         ports:
           - containerPort: 5000
         env:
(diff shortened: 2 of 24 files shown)


daniel-goldstein:
Another option we discussed is demoting pods in dev namespaces off the Big tier, but I don't want to do that just yet since I want to keep dev and prod environments identical if possible.

danking (approved):

PR #10118, opened 2021-02-26
[batch] Add context managers for BatchClient, Backend

jigold:


daniel-goldstein (approved):

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -737,3 +737,9 @@ async def edit_billing_limit(self, project, limit):
     async def close(self):
         await self._session.close()
         self._session = None
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self):
```
daniel-goldstein, line 744: Does this need the same arguments as `__exit__`? CI failed here I think

PR #10119, opened 2021-02-26
[query] Use error code for ArrayRef

tpoterba:


tpoterba:
I probably missed some, let's see what the tests find.

tpoterba:
ah, crap, this requires more thought. indexArray is the function that python calls, we basically never generate ArrayRef.

johnc1231:
Could we just change the Python to generate `ArrayRef`? Not clear to me why we use `indexArray` instead.

tpoterba:
indexArray does python-style lookups, e.g. `[-1]` is the last element.

tpoterba:
I think what we want is an error code on apply nodes! There are lots of functions that can throw errors.

PR #10130, opened 2021-03-02
[ci] More splits for test_java

tpoterba:


danking (changes requested):
I think this is probably the wrong move. I'm not sure how to break up tests that use DataProvider like that.

PR #10131, opened 2021-03-02
[batch] use copy in batch instead of gsutil

cseed:


build.yaml
```diff
@@ -113,43 +113,43 @@ steps:
...
      - from: /io/repo/git_version
        to: /git_version
      - from: /io/repo/address/test
-       to: /repo/address/
+       to: /repo/address/test
      - from: /io/repo/lsm
        to: /repo/lsm
      - from: /io/repo/ci/bootstrap_create_accounts.py
-       to: /repo/ci/
+       to: /repo/ci/bootstrap_create_accounts.py
```
jigold, line 150: Just to make sure. You have to be explicit here because batch always uses DEST_IS_TARGET.

cseed: Correct!

jigold:
This is going to be really challenging to deploy! Can you bump the instance version? The issue is the new CI build will probably get run on old batch workers that have the old copying mechanism.

cseed:
I think it will actually just work out of the box.  In particular, the updated file paths that specify the full destination also work with gsutil.  The build.yaml changes are being run by the production batch, so we already know those are working with gsutil.

I'm running some benchmarks vs gsutil now and will have numbers in a while.  If that all looks good, I vote to merge this.

jigold (changes requested):
Can you bump the instance version so we know which ones have the new copying?

cseed:
`batch.copy` now prints a summary that looks like t his:

```
$ python3 -m batch.copy null '[{"from":"/home/cotton/hail/batch/foo","to":"gs://hail-cseed/foo"}]'
Transfer summary:
  Sources: 1
  Files: 3
  Bytes: 969.9 kB
  Average transfer rate: 1.4 MB/s
Sources:
  /home/cotton/hail/batch/foo: 3 files, 969.9 kB
```

hail/python/hailtop/aiotools/fs.py
```diff
@@ -407,6 +413,40 @@ def set_exception(self, exception: Exception):
...
+    def summarize(self):
+        source_reports = []
+        def add_source_reports(transfer_report):
+            if isinstance(transfer_report._source_report, SourceReport):
+                source_reports.append(transfer_report._source_report)
+            else:
+                source_reports.extend(transfer_report._source_report)
+
+        if isinstance(self._transfer_report, Transfer):
+            total_transfers = 1
```
jigold, line 430: You don't use total_transfers at all.

cseed: Fixed.

jigold (approved):

PR #10136, opened 2021-03-03
[batch] add more logs and do not wait for asyncgens

danking:
I think there is some unresolved issue with asyncgen shutdown that is keeping
workers alive. This is not an issue in worker because worker calls sys.exit
which forcibly stops execution. cc: @daniel-goldstein @jigold.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1336,8 +1336,6 @@ class Worker:
                 await app_runner.cleanup()
                 log.info('cleaned up app runner')
 
-            self.shutdown()
-
     async def kill_1(self, request):  # pylint: disable=unused-argument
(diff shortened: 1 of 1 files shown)


danking:
Also: we called shutdown twice

batch/batch/worker/worker.py
```diff
@@ -1483,12 +1481,14 @@ async def async_main():
     finally:
         try:
             worker.shutdown()
+            log.info(f'worker shutdown')
```
jigold, line 1484: You have to get rid of the fstrings in all your log statements.

jigold (changes requested):

jigold (approved):

jigold (approved):

PR #10138, opened 2021-03-03
[batch] split instance activation into two calls

jigold:


batch/batch/driver/main.py
```diff
@@ -216,14 +225,17 @@ async def activate_instance_1(request, instance):
...
-        'token': token,
-        'key': key
+        'token': token
     })
 
 
+@routes.get('/api/v1alpha/instances/gsa_key')
+@activating_instances_only
+async def get_gsa_key(request, instance):  # pylint: disable=unused-argument
+    return await asyncio.shield(get_gsa_key_1(instance))
```
danking, line 236: Hmm. Why shield this one? It's not doing any non-interruptible work, AFAICT.

danking (approved):
Approving because this is totally the right fix, but I left a comment we can address in a subsequent PR, if merited.

PR #10139, opened 2021-03-03
[grafana][auth] add nginx sidecar in front of grafana for tls and auth

daniel-goldstein:
Right now the Grafana is exposed as a k8s service speaking http with only the grafana auth. This puts an nginx sidecar in front of Grafana to bring TLS all the way through the the Grafana pod and perform dev authentication. This required adding an api endpoint to auth that can verify a connection based on the session and not an Authorization header. Other services like `router-resolver` have gotten away with not having this since they construct the Authorization header in python before hitting the `userinfo` endpoint, but this seems like a straightforward addition that will make it easier for internal authentication such as this case.

This does another deviant thing which is using a `runImage` step to template the nginx config instead of templating inside the container at container start time (like router and site currently do). It is a little janky, because there are essentially two jinja passes, one in CI to render the shell script for the job, and then the jinja line in the job itself to render the nginx config. But this looks to be the most straightforward way I could figure out without adding another `build.py` Step type and even in that case it would have to be some sort of no-op job.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -506,6 +506,41 @@ steps:
    publishAs: batch
    dependsOn:
      - service_base_image
+ - kind: runImage
+   name: render_grafana_nginx_conf
+   image:
+     valueFrom: service_base_image.image
+   script: |
+     set -ex
+     cd /io
+     rm -rf repo
+     mkdir repo
+     cd repo
+     {{ code.checkout_script }}
+     cd grafana
+     if [[ {{ deploy }} = True ]]; then
+      DEPLOY=true
+     else
+      DEPLOY=false
+     fi
+     python3 ../ci/jinja2_render.py '{"deploy": '${DEPLOY}', "default_ns": {"name": "{{ default_ns.name }}"}}' nginx.conf nginx.conf.out
+   outputs:
+     - from: /io/repo/grafana/nginx.conf.out
+       to: /nginx.conf.out
+   dependsOn:
+    - default_ns
+    - service_base_image
+ - kind: buildImage
+   name: grafana_nginx_image
+   dockerFile: grafana/Dockerfile.nginx
+   contextPath: grafana
+   publishAs: grafana
+   inputs:
+     - from: /nginx.conf.out
+       to: /nginx.conf.out
+   dependsOn:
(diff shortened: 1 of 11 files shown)


build.yaml
```diff
@@ -508,6 +508,36 @@ steps:
...
+     valueFrom: service_base_image.image
+   script: |
+     set -ex
+     cd /io
+     rm -rf repo
+     mkdir repo
+     cd repo
+     {{ code.checkout_script }}
+     cd grafana
+     python3 ../ci/jinja2_render.py '{"deploy": false, "default_ns": {"name": "{{ default_ns.name }}"}}' nginx.conf nginx.conf.out
```
daniel-goldstein, line 523: @danking Here's what I think a jinja templating step would consist of. I feel like this is also not quite as straightforward as I would have liked because there's two template rendering steps happening, first the bash script is rendered by CI and then the actual job does rendering of the conf file. `false` is not correct here but I was having trouble with `{{ deploy }}` since CI's rendering would fill in `False` instead of `false`, so `jinja2_render.py` threw a JSON decoding error. Any thoughts?

daniel-goldstein: Could maybe do `"default_ns": {{ default_ns }}` but idk if it renders as a json object. Either way that doesn't solve the deploy awkwardness

build.yaml
```diff
@@ -506,6 +506,41 @@ steps:
...
+     mkdir repo
+     cd repo
+     {{ code.checkout_script }}
+     cd grafana
+     if [[ {{ deploy }} = True ]]; then
+      DEPLOY=true
+     else
+      DEPLOY=false
+     fi
+     python3 ../ci/jinja2_render.py '{"deploy": '${DEPLOY}', "default_ns": {"name": "{{ default_ns.name }}"}}' nginx.conf nginx.conf.out
```
danking, line 526: I think the Jinja-way would be:
```
{% if deploy %}
DEPLOY=true
{% else %}
DEPLOY=false
{% endif %}
```

templating combined with bash conditions gives me the heebie-jeebies.

danking (dismissed):
Small nit, looks great though!

danking (approved):

PR #10141, opened 2021-03-04
Revert "[tls] async create_certs"

daniel-goldstein:
Reverts hail-is/hail#9738. Concurrently creating certs doesn't work when OpenSSL assigns consecutive serial numbers.


danking (dismissed):
remove image-fetcher

danking (approved):

PR #10142, opened 2021-03-05
[ci] add authorize sha and action items table to user page

daniel-goldstein:
This kind of does a lot of small things:

- Adds a table at the top of the page that is all the PRs that require the user's attention, meaning: a personally authored PR where the build has failed or there are changes requested, or a PR for review that the user has not yet reviewed.

- Under PRs that require a review, a user now sees those where they are listed under `Assignees` *as well as* under `Reviewers`. We have been using assignees to assign reviews, but outside collaborators are unable to do so or dismiss reviews. This is because we don't really align with GitHub's semantics for these labels. Per GitHub, Assignees are organization members responsible for the supervision of a PR, whether the author or the final merge sign-off. Meanwhile, reviewers are what they suggest. If an outside contributor submits a PR, they can request a review from organization members in the `Reviewers` box, and then re-request review once they are ready for the PR to be looked at again. We should probably just abandon assignees given how we currently use them, but I just wanted to first get all of the information onto CI so nothing falls through the cracks.

- Adds authorize SHA input

- Fixes some small cosmetic bugs like a repeated "running" statement in batch tables and enables '/' focus on the jobs table in a batch.

### DIFF
--- a/ci/ci/ci.py
+++ b/ci/ci/ci.py
@@ -213,6 +213,22 @@ def filter_wbs(wbs, pred):
     return [{**wb, 'prs': [pr for pr in wb['prs'] if pred(pr)]} for wb in wbs]
 
 
+def is_pr_author(gh_username, pr_config):
+    return gh_username == pr_config['author']
+
+
+def is_pr_reviewer(gh_username, pr_config):
+    return gh_username in pr_config['assignees']
+
+
+def pr_requires_action(gh_username, pr_config):
+    build_state = pr_config['build_state']
+    review_state = pr_config['review_state']
+    return (
+        is_pr_author(gh_username, pr_config) and (build_state == 'failure' or review_state == 'changes_requested')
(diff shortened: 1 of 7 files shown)


ci/ci/ci.py
```diff
@@ -213,20 +213,39 @@ def filter_wbs(wbs, pred):
...
+def is_pr_reviewer(gh_username, pr_config):
+    return gh_username in pr_config['assignees']
+
+
+def pr_requires_action(gh_username, pr_config):
+    build_state = pr_config['build_state']
+    review_state = pr_config['review_state']
+    return (
+        is_pr_author(gh_username, pr_config) and (build_state == 'failure' or review_state == 'changes_requested')
+    ) or (is_pr_reviewer(gh_username, pr_config) and review_state == 'pending')
```
daniel-goldstein, line 229: this would look prettier in lisp..

danking: I do quite like function notation for and and or.

danking (approved):
Love it.

PR #10150, opened 2021-03-06
[copy] fix overly agressive cleanup

cseed:
Multi-part uploads put their scratch data in `<dest-dir>/_/<token>`, however, they deleted `<dest-dir>/_`.  Therefore, if two multi-part uploads were happening to the same destination directory, they would potentially delete the other's uploads causing a 404.  I will hand merge this because it will fail the build.


cseed:
I will follow up with tests to make sure I test this case.

PR #10153, opened 2021-03-08
[query-service] maybe fix event loop not initialized

danking:
The event loop is supposed to be initialized in the main thread. Sometimes
our tests get placed in the non-main thread (always a thread named Dummy-1).
Hopefully the session-scoped fixture is run in the main thread.


danking:
Example failure: https://ci.hail.is/batches/202357/jobs/104

tpoterba (approved):

PR #10158, opened 2021-03-08
[query] apply nest-asyncio as early as possible

danking:


danking:
Confirmed by opening a notebook and running a query against the service.

johnc1231 (approved):
As long as it passes python tests this is fine with me.

PR #10161, opened 2021-03-09
[hailctl] add a `--service-account` parameter to `hailctl dataproc start`

lgruen:
This is useful for falling back to Dataproc clusters in Batch pipelines, until Query is feature-complete.


danking:
@tpoterba looks fine to me, but I probably shouldn't be approving hailctl things.

PR #10165, opened 2021-03-09
[prometheus] add prometheus to track SLIs

daniel-goldstein:
This adds a prometheus statefulset to track metrics like API request latency and uptime. It scrapes pods on a 15s interval and collects prometheus metrics from any container in a pod with `grafanak8sapp` label that exposes an https endpoint `/metrics`.
The batch front end was already exposing prometheus metrics, but I changed it up slightly. For any http endpoint there should be a single metric, `http_request_latency`. Prometheus adds app and namespace metrics so seeing latencies for batch in particular is just a filter applied to this single metric. You can track latency of an endpoint by adding the `@monitor_endpoint` decorator defined in `metrics.py`, which tracks latency as well as number of requests and status code per request, available in the `http_request_count` metric. I also added monitoring to all CI endpoints.

This also includes an `up` metric for tracking uptime at the same 15s granularity.

I'm not convinced prometheus will suit our finer-grained needs surrounding batch, but it should do well enough in the meantime for our more traditional SLIs and allows to focus on one problem at a time.

### DIFF
--- a/prometheus/prometheus.yaml
+++ b/prometheus/prometheus.yaml
@@ -1,0 +1,215 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: prometheus
+  namespace: {{ default_ns.name }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: prometheus
+  namespace: {{ default_ns.name }}
+rules:
+ - apiGroups: [""]
+   resources:
+    - nodes
+    - nodes/proxy
+    - services
+    - endpoints
+    - pods
+   verbs: ["get", "list", "watch"]
+ - apiGroups:
+    - extensions
+   resources:
+    - ingresses
+   verbs: ["get", "list", "watch"]
+ - nonResourceURLs: ["/metrics"]
+   verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: prometheus
+roleRef:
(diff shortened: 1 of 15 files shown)


prometheus/prometheus.yaml
```diff
@@ -0,0 +1,215 @@
...
+data:
+  prometheus.yml: |
+    global:
+      scrape_interval: 15s
+    scrape_configs:
+     - job_name: "kubernetes-kubelet"
+       scheme: https
+       tls_config:
+         ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+         insecure_skip_verify: true
```
danking, line 57: What does this line mean?

danking (changes requested):
awesome! small comments

prometheus/prometheus.yaml
daniel-goldstein: it tells prometheus to skip the verification of the target's cert, without which prometheus rejects the scrape with `cannot validate certificate for <POD_IP> because it doesn't contain any IP SANs`. I haven't looked much into what it would take to remedy that yet

daniel-goldstein: I'm not read up on how our DNS works outside of kubernetes services and what's baked into our certs. This directly scrapes pods, not services, so I don't see how we can verify the name

daniel-goldstein: it is possible to scrape services as well / instead of pods

danking: Ah, this is because k8s certs don't have the pod IPs in them, OK, this seems fine.

danking (approved):

PR #10173, opened 2021-03-10
[website] fix website

danking:
I build old versions of the docs and use them in new websites. This does not
work for versions of the docs before I introduced the new system. In particular
versions 0.2.63 and before generate old-style docs.


danking:
Also, the tutorials need to be templated like other docs.

daniel-goldstein (approved):

danking:
Force merging because we keep failing in service backend or in those damn copy tests.

PR #10175, opened 2021-03-11
[query] fix loading of reference genomes when GCS object versions are used

lgruen:
See https://hail.zulipchat.com/#narrow/stream/123010-Hail-0.2E2.20support/topic/ServiceBackend.3A.20read_matrix_table.20from.20gs/near/229819329 for context.


danking (changes requested):
I think this is not the right fix. See: https://github.com/hail-is/hail/pull/10184

PR #10179, opened 2021-03-11
[gateway] move ukbb routing into gateway

daniel-goldstein:
One more step of stripping down router. ukbb lives in its own namespace and we never run dev versions of it so it's pretty much a drop-in.


daniel-goldstein:
this is, uh, already deployed.

danking (approved):

PR #10182, opened 2021-03-11
[query] Fix filter intervals (keep=False) memory leak

tpoterba:
EDIT from John for future readers: The issue here is that in general we only expect consumers to free memory after they're done processing a row. So when doing any sort of filtering operation, we want to make our own region that we clear after each row is filtered out, otherwise the garbage rows will continue to accumulate until we finally find a row to keep. `filterWithContext` demonstrates the correct way to do this, and so switching to use that method fixes the problem.


johnc1231 (approved):
God this stuff is so sneaky. The only right thing to do is to try and use as few functions as possible so it's hard to mess up (just send every filtering function through `filterWithContext`)

PR #10184, opened 2021-03-11
[query-service] teach query service to read MTs and Ts created by Spark

danking:
Hail-on-Spark uses HadoopFS which emulates directories by creating size-zero files with
the name `gs://bucket/dirname/`. Note: the object name literally ends in a slash. Such files
should not be included in `listStatus` (they should always be empty anyway). Unfortunately,
my fix in https://github.com/hail-is/hail/pull/9914 was wrong because `GoogleStorageFileStatus` removes
the trailing slash. This prevented the path from matching `path`, which always ends in a `/`.

Sorry @jigold, you're now on the reviewers list for the Query Service with Arcturus's departure.


danking:
See also: https://hail.zulipchat.com/#narrow/stream/123010-Hail-0.2E2.20support/topic/ServiceBackend.3A.20read_matrix_table.20from.20gs

jigold (dismissed):

danking:
@jigold should be working now

jigold (approved):

PR #10186, opened 2021-03-11
[datasets] add pan-ukb datasets

pwc2:
Make available [pan-ukb datasets](https://pan.ukbb.broadinstitute.org/docs/hail-format) via datasets API.

Includes:
  - Summary statistics MatrixTable, meta-analysis MatrixTable (both on GCS and S3)
  - LD score Table, variant index table, and LD BlockMatrix for each population (AFR, AMR, CSA, EAS, EUR, MID) (S3 only)

### DIFF
--- a/hail/python/hail/experimental/datasets.json
+++ b/hail/python/hail/experimental/datasets.json
@@ -1583,7 +1583,7 @@
     ]
   },
   "panukb_ld_block_matrix_AFR": {
-    "description": "Pan-UKB: linkage disequilibrium (LD) matrix Hail BlockMatrix for African ancestry population.",
+    "description": "Pan-UKB: linkage disequilibrium (LD) matrix Hail BlockMatrix for African ancestry population. To determine which row/column corresponds to which variant, see the associated variant indices Hail Table: panukb_ld_variant_indices_AFR.",
     "url": "https://pan.ukbb.broadinstitute.org/docs/technical-overview",
     "versions": [
       {
@@ -1598,7 +1598,7 @@
     ]
   },
   "panukb_ld_block_matrix_AMR": {
-    "description": "Pan-UKB: linkage disequilibrium (LD) matrix Hail BlockMatrix for Admixed American ancestry population.",
+    "description": "Pan-UKB: linkage disequilibrium (LD) matrix Hail BlockMatrix for Admixed American ancestry population. To determine which row/column corresponds to which variant, see the associated variant indices Hail Table: panukb_ld_variant_indices_AMR.",
     "url": "https://pan.ukbb.broadinstitute.org/docs/technical-overview",
     "versions": [
       {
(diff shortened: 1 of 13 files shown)


hail/python/hail/experimental/datasets.json
```diff
@@ -1581,5 +1581,311 @@
...
+        "version": "0.2"
+      }
+    ]
+  },
+  "panukb_ld_block_matrix_EUR": {
+    "description": "Pan-UKB: linkage disequilibrium (LD) matrix Hail BlockMatrix for European ancestry population.",
+    "url": "https://pan.ukbb.broadinstitute.org/docs/technical-overview",
+    "versions": [
+      {
+        "reference_genome": "GRCh37",
```
danking, line 1650: Block Matrices don't have a reference genome

danking (changes requested):
Awesome! I'm very excited to have these in the Datasets API / Annotation DB.

The Block Matrices shouldn't have reference genomes attached. @konradjk, is there a version we should associate with these? Patrick, where did 0.2 come from?

konradjk:
Just checked the metadata.json.gz of one of the variant index files and looks like they were written with `0.2.58-3f304aae6ce2`

danking:
@konradjk ah, I meant does Pan-UKB itself have a release version? I guess we should just call this v1?

pwc2:
I went with version 0.2 based on the entry in the [changelog](https://pan.ukbb.broadinstitute.org/docs/changelog):

```
Version 0.2

Added LD scores and matrices (released Oct 29, 2020).
```

danking:
Ah, perfect. I'm a bit worried that the URLs aren't themselves versioned. Konrad, will subsequent releases use the same URL?

hail/python/hail/experimental/datasets.json
pwc2: In the description for the Block Matrices I can add something like "see associated Hail Table x". And you want that to refer to the associated variant index HT, correct?

konradjk:
Hmm, great question. At this time we have no plans for subsequent versions, so we haven't broached that topic. I think 0.2 is good based on the changelog, and maybe we can add a version number if we do change versions. We can't really change these as they've been out there and people are using them.

hail/python/hail/experimental/datasets.json
konradjk: Yep, that would be awesome!

danking (approved):

danking:
Awesome, thanks Konrad, and thanks Patrick, this is great. We'll keep the reference genome with the BM.

pwc2:
Okay cool. In a day or two I should have a PR ready that adds the GIANT exome array summary statistics tables as well.

hail/python/hail/experimental/datasets.json
pwc2: Sounds good! I just added that reference to the BM descriptions.

PR #10187, opened 2021-03-11
[query] NDArray Sum

johnc1231:
Finally correcting this glaring omission, this PR adds ability to sum ndarray along an axis. This is generally useful, but is also necessary for lowering BlockMatrix.sum.

### DIFF
--- a/hail/python/hail/expr/expressions/typed_expressions.py
+++ b/hail/python/hail/expr/expressions/typed_expressions.py
@@ -4099,7 +4099,6 @@ class NDArrayNumericExpression(NDArrayExpression):
 
         return res if result_ndim > 0 else res[()]
 
-
     @typecheck_method(axis=nullable(oneof(int, tupleof(int))))
     def sum(self, axis=None):
         if axis is None:
--- a/hail/python/hail/expr/expressions/typed_expressions.py
+++ b/hail/python/hail/expr/expressions/typed_expressions.py
@@ -4101,6 +4101,17 @@ class NDArrayNumericExpression(NDArrayExpression):
 
     @typecheck_method(axis=nullable(oneof(int, tupleof(int))))
     def sum(self, axis=None):
+        """Sum out one or more axes of an ndarray.
+
+        Parameters
+        ----------
+        axis : :class:`int` :class:`tuple`
+            The axis or axes to sum out.
+
+        Returns
+        -------
+        :class:`.NDArrayNumericExpression`
+        """
         if axis is None:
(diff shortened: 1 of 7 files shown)


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -2799,6 +2800,33 @@ class Emit[C](
...
+            val childDims = child.typ.asInstanceOf[TNDArray].nDims
+            val axesToKeep = (0 until childDims).filter(axis => !axesToSumOut.contains(axis))
+            val newOutputShape = axesToKeep.map(idx => childEmitter.outputShape(idx))
+            val newOutputShapeComplement = axesToSumOut.map(idx => childEmitter.outputShape(idx))
+
+            new NDArrayEmitter(newOutputShape) {
+              override def outputElement(cb: EmitCodeBuilder, idxVars: IndexedSeq[Value[Long]]): PCode = {
+                val numericElementType = coerce[PNumeric](child.pType.asInstanceOf[PNDArray].elementType)
+                val runningSum = coerce[Any](cb.newLocal("running_sum")(typeToTypeInfo(numericElementType)))
+                cb.assign(runningSum, numericElementType.zero)
```
johnc1231, line 2814: @tpoterba, is this hacky type info stuff still the right way to handle situations like this?

hail/python/hail/expr/expressions/typed_expressions.py
```diff
@@ -4099,6 +4099,20 @@ def __matmul__(self, other):
...
+    def sum(self, axis=None):
+        if axis is None:
+            axis = tuple(range(self.ndim))
+
+        axis = wrap_to_tuple(axis)
+        res_ir = ir.NDArrayAgg(self._ir, axis)
+
+        num_axes_deleted = len(set(axis))
+        if num_axes_deleted > self.ndim:
+            raise ValueError(f"axes list was of length {num_axes_deleted} but ndarray dim is only {self.ndim}")
```
patrick-schultz, line 4112: I'm confused what this is guarding against. Should it just check that `0 <= i < self.ndim` for all `i in axis`?

patrick-schultz (changes requested):

hail/python/hail/expr/expressions/typed_expressions.py
johnc1231: You're right, this just checked that you didn't specify more dimensions than the ndarray had. Replaced with better checks.

patrick-schultz (approved):

PR #10188, opened 2021-03-12
Overwrite ssl-config-hail-root secret for dev

vladsaveliev:
Dev certificates expire in 30 days, and rerunning `kubectl create secret` doesn't update the secret if it already exists. So adding a `kubectl delete secret` line to make sure the new secret will be added.

Though I'm not familiar with the `test` scope and not sure if it will break something for it?

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -170,7 +170,7 @@ steps:
          kubectl delete secret -n {{ default_ns.name }} ssl-config-hail-root
      done     
      openssl req -new -x509 -subj /CN=hail-root -nodes -newkey rsa:4096 -keyout hail-root-key.pem -out hail-root-cert.pem
-     until kubectl get secret -n {{ default_ns.name }} ssl-config-hail-root
+     until kubectl get secret -n {{ default_ns.name }} ssl-config-hail-root 2>/dev/null
      do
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
Hi @vladsaveliev, sorry for the delay on this. I'm going to leave this for when @danking comes back next week because I think there's some nuance around our dev cert expiration that I'm not 100% on. In the meantime, I looked and saw you try to delete then try to create repeatedly. Could you instead use a dry-run then apply trick like [this](https://stackoverflow.com/questions/45879498/how-can-i-update-a-secret-on-kubernetes-when-it-is-generated-from-a-file)? We do this for some other keys in `build.yaml`.

danking:
Hmm. This means every dev deploy will generate a new root key. I'm worried about the derived keys and trust lists. After this runs, any service which was not dev deployed needs to know to reload the trust list and start using the new key.

For example, if you dev deploy batch, then separately dev deploy query, the new query will get cert errors when talking to batch, I think.

I will give some thought this week to the right long-term certificate strategy.

danking:
I'm going to close this PR because its so old. I've bumped the priority of certificates in my todo list. Sincere apologies @vladsaveliev , just too many things to do these days :|.

PR #10189, opened 2021-03-13
[query] call init_service from init based on HAIL_QUERY_BACKEND value

lgruen:
The idea is to allow the execution on both `SparkBackend` and `ServiceBackend` without code changes, simply switching to the Query service by setting the environment variables `HAIL_QUERY_BACKEND`, `HAIL_BILLING_PROJECT`, and `HAIL_BUCKET`.


tpoterba (approved):
looks good. I'll also plug local backend support in here at some point too.

PR #10190, opened 2021-03-15
[website] dont jinja render any of the batch docs when serving

daniel-goldstein:
the batch tutorial contained some `{{` that made for an invalid jinja template, which caused a 500. The batch docs in general don't seem to need any additional rendering at runtime so this treats all the batch docs as static files.


jigold:
Is it possible to check this with dev deploy?

daniel-goldstein:
I dev deployed it [here](internal.hail.is/dgoldste/website/docs/batch/tutorial.html)

jigold (approved):

PR #10192, opened 2021-03-15
[ci] make test hello speak https

daniel-goldstein:
This is one piece of a multi-part saga to use tls for any and all communication between pods. This basically just required adding the `create_certs` step to the ci test `build.yaml` so we can create a ssl config for `hello`.

### DIFF
--- a/ci/test/resources/config.yaml
+++ b/ci/test/resources/config.yaml
@@ -1,0 +1,4 @@
+principals:
+- name: hello
+  domain: hello
+  kind: json
--- a/ci/test/resources/hello.py
+++ b/ci/test/resources/hello.py
@@ -15,12 +15,12 @@ SHA = os.environ['HAIL_SHA']
 
 
 @routes.get('/healthcheck')
-async def get_healthcheck(request):  # pylint: disable=W0613
+async def get_healthcheck(request):  # pylint: disable=unused-argument
     return web.Response()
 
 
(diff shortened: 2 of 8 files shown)


ci/test/resources/hello.py
```diff
@@ -19,10 +20,12 @@ async def get_healthcheck(request):  # pylint: disable=W0613
 
 
 @routes.get('/sha')
-async def get_sha(request):
+async def get_sha(request):  # pylint: disable=W0613
```
jigold, line 23: can we put the name of the error rather than the code. I think you want `# pylint: disable=unused-argument`

ci/test/resources/config.yaml
```diff
@@ -0,0 +1,4 @@
+principals:
```
jigold, line 1: What does this file do?

jigold (changes requested):

ci/test/resources/config.yaml
daniel-goldstein: This tells `create_certs.py` which applications to make certificates for and what the format of the kubernetes secret should be. JSON is used for all of the `aiohttp` services and `nginx` is for nginx proxies. This mirrors `tls/config.yaml`, I just didn't want to throw hello in there with the real services.

jigold (approved):

PR #10195, opened 2021-03-15
[query] BlockMatrix map

johnc1231:
This adds an arbitrary mapping operation to `BlockMatrix`. We already had the IR for this and supported it in the lowerer. We don't support this in the unlowered `execute` infrastructure, so I had to add a check to prevent that from happening.

I'm not sure how to document the fact that this operation is unavailable in Spark. I don't think we've had to do that yet. 

I'm also curious if you think `map_dense` and `map_sparse` are good names, or if you think we ought to just have one `map` with a boolean. I do worry what `sparse` is a funny name for this, since when I think of a sparse matrix I think of something closer to Spark's `CoordinateMatrix` (very few, random entries with no structure), as opposed to the block sparsity we have.

This PR also changes the implementation of adding/subtracting/multiplying/dividing a constant to BlockMatrix, making it use `BlockMatrixMap` instead of `BlockMatrixMap2`

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/BlockMatrixIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/BlockMatrixIR.scala
@@ -270,6 +270,17 @@ case class BlockMatrixMap(child: BlockMatrixIR, eltName: String, f: IR, needsDen
   override protected[ir] def execute(ctx: ExecuteContext): BlockMatrix = {
     val prev = child.execute(ctx)
 
+    val functionArgs = f match {
+      case ApplyUnaryPrimOp(_, arg1) => IndexedSeq(arg1)
+      case Apply(_, _, args, _) => args
+      case ApplyBinaryPrimOp(_, l, r) => IndexedSeq(l, r)
+    }
+
+    assert(functionArgs.forall(ir => IsConstant(ir) || ir.isInstanceOf[Ref]),
+      "Spark backend without lowering does not support general mapping over " +
+        "BlockMatrix entries. Use predefined functions like `BlockMatrix.abs`.")
+
+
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/expr/ir/BlockMatrixIR.scala
```diff
@@ -270,6 +270,17 @@ case class BlockMatrixMap(child: BlockMatrixIR, eltName: String, f: IR, needsDen
   override protected[ir] def execute(ctx: ExecuteContext): BlockMatrix = {
     val prev = child.execute(ctx)
 
+    val functionArgs = f match {
+      case ApplyUnaryPrimOp(_, arg1) => IndexedSeq(arg1)
+      case Apply(_, _, args, _) => args
+      case ApplyBinaryPrimOp(_, l, r) => IndexedSeq(l, r)
+    }
+
+    assert(functionArgs.forall(ir => IsConstant(ir) || ir.isInstanceOf[Ref]),
```
tpoterba, line 279: can this replace the match error below?

tpoterba (dismissed):

hail/src/main/scala/is/hail/expr/ir/BlockMatrixIR.scala
johnc1231: The reason this is necessary is because if you do something like `bm.map(lambda x: hl.abs(x + 1))` with the match, it will just apply the outermost function (the `abs`), not transitively apply the series of applies. But the other match is necessary because we don't support every possible apply function on BlockMatrix, just ones we've special cased.

tpoterba (approved):

PR #10199, opened 2021-03-17
[query] fix tmp_dir default in init(), which doesn't work for the service backend

lgruen:


hail/python/hail/context.py
```diff
@@ -129,7 +129,7 @@ def stop(self):
            _optimizer_iterations=nullable(int))
```
johnc1231, line 129: A few lines up, you have to change `tmp_dir=str` to `tmp_dir=nullable(str)`

johnc1231 (changes requested):

hail/python/hail/context.py
lgruen: Thanks for catching that! Fixed.

I've also realized that `_get_tmpdir` already handles `None`, so this becomes even simpler.

PR #10200, opened 2021-03-18
[cpp] add support for tuple

cseed:
And lots of other things along the way.  Summary of changes:
 - Add RawArenaAllocator
 - Add runtime to call arena allocate from generated code
 - Support MakeTuple, GetTupleElement
 - Intern STypes, add STypeContext
 - Add EmitValue.cast_to, SType.{load_from_address, construct_from_value, construct_at_address_from_value}
 - Add SType.get_memory_size, only works for types for which is_heap_stype<T>() is true
 - Use phi nodes to implement EmitValue.as_data()
 - isa now calls T::is_instance_tag rather than checking tag directly

### DIFF
--- a/libhail/CMakeLists.txt
+++ b/libhail/CMakeLists.txt
@@ -28,7 +28,9 @@ add_executable(main
   src/hail/query/backend/compile.cpp
   src/hail/query/backend/jit.cpp
   src/hail/query/backend/stype.cpp
-  src/hail/query/backend/svalue.cpp)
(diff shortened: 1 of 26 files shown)


johnc1231 (changes requested):
Not done reviewing, but noticed this is missing `runtime.cpp` and `runtime.hpp`, think you just forgot to add `git add` them.

johnc1231 (approved):

PR #10202, opened 2021-03-18
[query] Rework Emit.scala to use Value[Region] instead of StagedRegion

tpoterba:


johnc1231:
Failing tests.

tpoterba:
Closing in favor of a much spicier redesign.

PR #10204, opened 2021-03-19
[notebook] Move notebook and workshop nginx to sidecar

daniel-goldstein:
This moves the nginx proxy configuration out of router and into a sidecar in the notebook/workshop pod. This extends TLS termination from router to the notebook pod and consolidates the notebook routing logic. I didn't run a scale test but this doesn't change any functionality, and I tested in dev that I could log in to a workshop, start and open a notebook.

### DIFF
--- a/notebook/deployment.yaml
+++ b/notebook/deployment.yaml
@@ -66,8 +66,8 @@ spec:
          image: "{{ notebook_image.image }}"
          resources:
            requests:
-             memory: 1Mi
-             cpu: 1m
+             memory: 20M
+             cpu: 20m
            limits:
              memory: 200Mi
              cpu: "1"
--- a/notebook/deployment.yaml
+++ b/notebook/deployment.yaml
@@ -42,6 +42,11 @@ spec:
         app: notebook
         hail.is/sha: "{{ code.sha }}"
     spec:
+      hostAliases:
+       - ip: "127.0.0.1"
+         hostnames:
+         - "notebook.local"
+         - "workshop.local"
       serviceAccountName: notebook
 {% if deploy %}
       priorityClassName: production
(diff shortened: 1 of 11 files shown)


notebook/deployment.yaml
```diff
@@ -66,8 +66,8 @@ spec:
          image: "{{ notebook_image.image }}"
          resources:
            requests:
-             memory: 1Mi
-             cpu: 1m
+             memory: 20M
+             cpu: 20m
```
danking, line 70: shouldn't this have a second container?

danking (dismissed):

daniel-goldstein:
I think this is ready for another look but I put a WIP label on it because I don't want to merge it today while there's still workshop things happening. After it goes in I'll run another scale test.

The nginx config is mostly just lifted over from router, with a couple small changes. I wasn't able to just proxy to localhost because notebook couldn't figure out which subdomain the request was going to and everything would 404. Adding a Host header fixed the 404, but messed with the redirection to notebook pods so instead I added `workshop.local` and `notebook.local` to `/etc/hosts` on the pod, which I kind of like, which did the trick.

I needed two ssl configs, one for nginx and one for aiohttp to use internally, and ended up changing the `ssl-config-notebook` secret to `nginx` and creating a second `ssl-config-notebook-python`, but let me know if this seems off.

danking (dismissed):
I'm basically OK with this, but it appears that the notebook python code doesn't use the SSL secret at all anymore? Should we delete it? Or should nginx speak TLS over the loopback interface to notebook-python?

daniel-goldstein:
Notebook still does auth on incoming requests, which uses the ssl config to talk to `auth`. Potentially a little confusing with a proxy in front of it too, but that felt out of scope for this PR.

danking (approved):

PR #10207, opened 2021-03-19
[gateway] Cut out router and router-resolver from gateway internal routing

daniel-goldstein:
This PR only impacts `internal.hail.is`.

The current internal flow is:
1. gateway receives request for `internal.hail.is/$namespace/$service`
2. `auth_request` to router-resolver for auth & router ip, and in router-resolver
  a. request to auth that the user is a signed in developer
  b. retrieve router ip from k8s
3. `proxy_pass` to router IP returned from router-resolver in $namespace which proxies to $service in $namespace

This changes the flow to:
1. gateway receives request for `internal.hail.is/$namespace/$service`
2. request to auth that the user is a signed in developer
3. `proxy_pass` to $service in $namespace

This does impose the restriction that there is an equivalence between subdomains and k8s Services. For the most part we enforce this just by convention, with the exception of `www` -> `website`. `router` currently does this mapping and is the only capability that this version drops. However, `www` in internal is already broken, I think because `website` does not consider `www` as potentially part of its base path, so that might need to be considered on its own. Aside from the `base_path` problem, I think it would make sense to create alias k8s Services like we do with notebook/workshop.

### DIFF
--- a/gateway/gateway.nginx.conf
+++ b/gateway/gateway.nginx.conf
@@ -35,7 +35,7 @@ server {
         auth_request /auth;
 
         resolver kube-dns.kube-system.svc.cluster.local;
-        proxy_pass https://router.$namespace.svc.cluster.local;
+        proxy_pass https://$service.$namespace.svc.cluster.local;
 
         proxy_set_header Host $service.internal;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
I've been letting this cook all day with no unforeseen problems so unless someone redeploys gateway it is the current working version.

danking (approved):
This is really, really awesome.

PR #10210, opened 2021-03-22
[metrics] add metrics collection to API endpoints across services

daniel-goldstein:
Added endpoint monitoring to auth, query, memory, atgu, workshop and website, as well as removed an unused label on address.


danking (changes requested):
Do memory, auth, and query need `grafanak8sapp: "true"`?

daniel-goldstein:
`grafanak8sapp` is the label we're using to let prometheus know it should scrape metrics from a given pod

PR #10211, opened 2021-03-23
[query] Include info field key in parse error

tpoterba:


johnc1231 (approved):

hail/src/main/scala/is/hail/io/vcf/LoadVCF.scala
```diff
@@ -955,8 +955,14 @@ final class VCFLine(val line: String, arrayElementsRequired: Boolean,
...
         } else
           rvb.addBoolean(true)
-      } else
-        parseAddInfoField(rvb, c.infoFieldTypes(idx))
+      } else {
+        try {
+          parseAddInfoField(rvb, c.infoFieldTypes(idx))
+        } catch {
+          case e: VCFParseError => parseError(s"error while parsing info field '$key': ${ e.msg }")
+          case e => throw e
```
johnc1231, line 963: Delete this line. Don't catch something just to throw it.

johnc1231 (changes requested):

hail/src/main/scala/is/hail/io/vcf/LoadVCF.scala
johnc1231: This is also the cause of your compilation error. Scala views doing this as a mistake.

tpoterba:
addressed, sorry I lost track of this...

johnc1231 (approved):

PR #10212, opened 2021-03-23
[internal-gateway][router-resolver] Move internal gateway off of router-resolver and router

daniel-goldstein:
This is a follow-up on #10207, which circumvents `router-resolver` and `router` and uses kube dns to route requests passing through internal gateway directly to destination namespace and service. Previously, `internal-gateway` was only using `router-resolver` to discover the scheme of the `router` it needed to forward the request to. Since we are using `https` in front of all of our services now, and are ultimately retiring router, this is no longer necessary.

This is also the last use of `router-resolver` in the whole codebase, so we can remove the service entirely! (second commit)

This version of `internal-gateway` has been running since yesterday evening and there is currently no `router-resolver` deployment in default!


internal-gateway/internal-gateway.nginx.conf
```diff
@@ -49,22 +34,20 @@ server {
...
         proxy_set_header Connection $connection_upgrade;
-
-        # we do not verify namepsaced routers
-        # include /ssl-config/ssl-config-proxy.conf;
     }
 }
 
 server {
-    server_name hail;
+    server_name ~^(?<service>[\w-]+)\.hail$;
```
danking, line 41: I'm not sure how to read this regex. Isn't `[\w-]` the character class for whitespace and dashes? I'd think we'd just want `^(?<service>.+)\.hail$`.

internal-gateway/internal-gateway.nginx.conf
```diff
@@ -114,22 +87,20 @@ server {
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection $connection_upgrade;
-
-        # we do not verify namepsaced routers
-        # include /ssl-config/ssl-config-proxy.conf;
```
danking, line 119: I'm inclined to leave this as a note to ourselves that we don't actually check the certs of the service

danking (dismissed):

internal-gateway/internal-gateway.nginx.conf
daniel-goldstein: It's the word character so anything alphanumeric and underscores. We could probably do `.+` but I thought why not restrict some wonky characters and extra periods.

danking (approved):

PR #10215, opened 2021-03-23
[query] Add experimental support for resumable matrix writes

tpoterba:
This feature is not widely supported (only on MatrixTable & SparkBackend)
and is not well-tested in CI (we only test that matrix writes run and return
the correct result with the _checkpoint_file parameter, not the performance
semantics).

I've played around with this code on my laptop and the performance semantics
are what I expect. Logging messages provide some transparency too:

```

In [2]: mt = hl.utils.range_matrix_table(100000, 10, 500)

In [3]: mt.write('/tmp/mt_temp4.mt', _checkpoint_file='/tmp/mt_checkpoint')
	2021-03-23 14:50:06 Hail: INFO: creating new checkpoint at /tmp/mt_checkpoint
	^C---------------------------------------------------------------------------00]
	KeyboardInterrupt                         Traceback (most recent call last)
	<snip>
	KeyboardInterrupt:

In [4]: mt.write('/tmp/mt_temp4.mt', _checkpoint_file='/tmp/mt_checkpoint')
	2021-03-23 14:50:14 Hail: INFO: resuming matrix write from /tmp/mt_checkpoint with 192/500 partitions written
	^C---------------------------------------------------------------------------00]
	KeyboardInterrupt                         Traceback (most recent call last)
	<snip>
	KeyboardInterrupt:

In [5]: mt.write('/tmp/mt_temp4.mt', _checkpoint_file='/tmp/mt_checkpoint')
	2021-03-23 14:50:22 Hail: INFO: resuming matrix write from /tmp/mt_checkpoint with 300/500 partitions written
	^C---------------------------------------------------------------------------00]
	KeyboardInterrupt                         Traceback (most recent call last)
	<snip>
	KeyboardInterrupt:

In [6]: mt.write('/tmp/mt_temp4.mt', _checkpoint_file='/tmp/mt_checkpoint')
	2021-03-23 14:50:29 Hail: INFO: resuming matrix write from /tmp/mt_checkpoint with 372/500 partitions written
	2021-03-23 14:50:36 Hail: INFO: wrote matrix table with 100000 rows and 10 columns in 500 partitions to /tmp/mt_temp4.mt
	    Total size: 391.55 KiB
	    * Rows/entries: 391.51 KiB
	    * Columns: 31.00 B
	    * Globals: 11.00 B
	    * Smallest partition: 200 rows (505.00 B)
	    * Largest partition:  200 rows (835.00 B)

In [7]: mt_readback = hl.read_matrix_table('/tmp/mt_temp4.mt')
```

I think it would be cool to make this an automatic default feature. The version, a digest of the IR, and the global random seed
seem necessary, but we would really also need some digest of all file dependencies to ensure we're really resuming the same
computation, and that's hard.

I also don't have a plan for supporting this in the lowered execution path at the moment.


patrick-schultz (approved):

ptn24:
@tpoterba , would _checkpoint_file be compatible with any overwrite option value, or would it be required to pass a particular value? E.g.,

```
mt.write(url, overwrite=False, _checkpoint_file=checkpoint_file_url, ...)
```

versus

```
mt.write(url, overwrite=True, _checkpoint_file=checkpoint_file_url, ...)
```

tpoterba:
As written here, this feature is incompatible with `overwrite=True`. The checkpoint file makes it possible to recover progress from a partially written file at `url`, so I'm not sure what overwriting would mean in this context.

Is there a particular use case you have in mind? This was built for a specific application, and I am eager to hear others.

ptn24:
Thank you for the quick response, @tpoterba . I understand, I think we are on the same page. I was mainly curious what would happen if overwrite was True. Whether it would start from scratch or ignore overwrite, or something else

That being said, a possible use case re overwrite could be to actually remove partial files re tasks being recovered. Re a single task, if it would fail, then it would be possible for there to exist a partial task file. Usually if a task fails, then Spark would retry it. And if the task's next attempt passes, then there would exist a full task file plus a partial task file. So, if Hail would recover the task, then I imagine there would also exist a full task file plus a partial task file. I have not seen partial task files affecting downstream processing yet, but I am generally wary of "duplicate" files. Just a thought

PR #10216, opened 2021-03-24
[batch] add more error information when missing header

danking:


danking:
overriding review and status due to batch being in a bad state.

PR #10217, opened 2021-03-24
[k8s] never delete a secret

danking:
For some reason we delete secrets when re-creating them. This creates a race condition which Batch
appears to encounter whenever it is under heavy load during a deploy.

@cseed, do you recall why you chose to delete and then use apply in ci/create_database.py?

I am using the command sequence suggested here: https://stackoverflow.com/questions/45879498/how-can-i-update-a-secret-on-kubernetes-when-it-is-generated-from-a-file.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -289,14 +289,18 @@ steps:
      cat > deploy-config.json <<EOF
      {"location":"k8s","default_namespace":"{{ default_ns.name }}","domain":"{{ global.domain }}"}
      EOF
-     kubectl -n {{ default_ns.name }} delete --ignore-not-found secret deploy-config
-     kubectl -n {{ default_ns.name }} create secret generic deploy-config --from-file=./deploy-config.json
+     kubectl -n {{ default_ns.name }} create secret generic deploy-config \
+             --from-file=./deploy-config.json \
+             --save-config --dry-run=client -o yaml \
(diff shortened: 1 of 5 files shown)


danking:
bump @cseed

cseed:
> @cseed, do you recall why you chose to delete and then use apply in ci/create_database.py?

I think no reason.  I wasn't aware of the race condition and didn't have an alternative for applying changes at hand.  This looks good to me.  I ran a few examples but I couldn't actually tell how `--save-config` changes the output.

danking:
It adds the last applied configuration annotation which is used to know if a subsequent configuration is different.

```
(base) # kubectl -n default create secret generic foo \                
              --save-config --dry-run=client -o yaml
apiVersion: v1
kind: Secret
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"kind":"Secret","apiVersion":"v1","metadata":{"name":"foo","creationTimestamp":null}}
  creationTimestamp: null
  name: foo
  namespace: default
(base) # kubectl -n default create secret generic foo \
              --dry-run=client -o yaml              
apiVersion: v1
kind: Secret
metadata:
  creationTimestamp: null
  name: foo
  namespace: default

```

daniel-goldstein (approved):

PR #10218, opened 2021-03-24
[copy] Fix copying file with name ending in slash

jigold:
- Also handle case with GCS files ending in a slash

### DIFF
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -425,7 +425,7 @@ class GetObjectFileStatus(FileStatus):
 
 class GoogleStorageFileListEntry(FileListEntry):
     def __init__(self, url: str, items: Optional[Dict[str, Any]], status=None):
-        # assert url.endswith('/') == (items is None), f'{url} {items}'
+        assert url.endswith('/') == (items is None), f'{url} {items}'
         self._url = url
         self._items = items
         self._status: Optional[GetObjectFileStatus] = status
@@ -646,15 +646,12 @@ class GoogleStorageAsyncFS(AsyncFS):
             if items is not None:
                 for item in page['items']:
                     url = f'gs://{bucket}/{item["name"]}'
-                    print(f'url={url}')
                     status = None
                     if not force_all and url.endswith('/') and item is not None:
                         status = GetObjectFileStatus(item)
                         size = await status.size()
                         if size != 0:
-                            print(f'non-empty {url} with size {size}')
                             raise FileAndDirectoryError(url)
-                        print(f'skipping {url} with size {size}')
                         continue
                     yield GoogleStorageFileListEntry(url, item, status)
 
@@ -671,6 +668,10 @@ class GoogleStorageAsyncFS(AsyncFS):
                 for prefix in prefixes:
                     assert prefix.endswith('/')
                     url = f'gs://{bucket}/{prefix}'
+                    if not force_all and await self.isfile(url):
+                        stat = await self.statfile(url)
+                        if await stat.size() != 0:
+                            raise FileAndDirectoryError(url)
                     yield GoogleStorageFileListEntry(url, None)
 
             items = page.get('items')
@@ -682,9 +683,7 @@ class GoogleStorageAsyncFS(AsyncFS):
                         status = GetObjectFileStatus(item)
                         size = await status.size()
                         if size != 0:
-                            print(f'non-empty {url} with size {size}')
                             raise FileAndDirectoryError(url)
-                        print(f'skipping {url} with size {size}')
                         continue
                     yield GoogleStorageFileListEntry(url, item, status)
 
@@ -730,7 +729,7 @@ class GoogleStorageAsyncFS(AsyncFS):
 
     async def isdir(self, url: str) -> bool:
         bucket, name = self._get_bucket_name(url)
-        assert not name or name.endswith('/')
+        assert not name or name.endswith('/'), name
         params = {
             'prefix': name,
             'delimiter': '/',
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -424,10 +424,10 @@ class GetObjectFileStatus(FileStatus):
 
 
 class GoogleStorageFileListEntry(FileListEntry):
-    def __init__(self, url: str, items: Optional[Dict[str, Any]], status: Optional[GetObjectFileStatus] = None):
+    def __init__(self, url: str, items: Optional[Dict[str, Any]]):
         self._url = url
         self._items = items
-        self._status = status
+        self._status: Optional[GetObjectFileStatus] = None
 
     def name(self) -> str:
         parsed = urllib.parse.urlparse(self._url)
@@ -632,23 +632,6 @@ class GoogleStorageAsyncFS(AsyncFS):
                 raise FileNotFoundError(url) from e
             raise
 
-    async def _listfiles_all(self, url: str) -> AsyncIterator[FileListEntry]:
-        bucket, name = self._get_bucket_name(url)
-        if name and not name.endswith('/'):
-            name = f'{name}/'
-        params = {
-            'prefix': name
-        }
-        async for page in await self._storage_client.list_objects(bucket, params=params):
-            prefixes = page.get('prefixes')
-            assert not prefixes
-
-            items = page.get('items')
-            if items is not None:
-                for item in page['items']:
-                    url = f'gs://{bucket}/{item["name"]}'
-                    yield GoogleStorageFileListEntry(url, item)
-
     async def _listfiles_recursive(self, bucket: str, name: str) -> AsyncIterator[FileListEntry]:
         assert not name or name.endswith('/')
         params = {
@@ -661,16 +644,7 @@ class GoogleStorageAsyncFS(AsyncFS):
             items = page.get('items')
             if items is not None:
                 for item in page['items']:
-                    url = f'gs://{bucket}/{item["name"]}'
-                    status = None
-                    if url.endswith('/'):
-                        status = GetObjectFileStatus(item)
-                        size = await status.size()
-                        if size != 0:
-                            raise FileAndDirectoryError(url)
-                        yield GoogleStorageFileListEntry(url, None)
-                    else:
-                        yield GoogleStorageFileListEntry(url, item, status)
+                    yield GoogleStorageFileListEntry(f'gs://{bucket}/{item["name"]}', item)
 
     async def _listfiles_flat(self, bucket: str, name: str) -> AsyncIterator[FileListEntry]:
         assert not name or name.endswith('/')
@@ -684,29 +658,12 @@ class GoogleStorageAsyncFS(AsyncFS):
             if prefixes:
                 for prefix in prefixes:
                     assert prefix.endswith('/')
-                    url = f'gs://{bucket}/{prefix}'
-                    try:
-                        stat = await self.statfile(url)
-                    except FileNotFoundError:
-                        pass
-                    else:
-                        if await stat.size() != 0:
-                            raise FileAndDirectoryError(url)
-                    yield GoogleStorageFileListEntry(url, None)
+                    yield GoogleStorageFileListEntry(f'gs://{bucket}/{prefix}', None)
 
             items = page.get('items')
             if items:
                 for item in page['items']:
-                    url = f'gs://{bucket}/{item["name"]}'
-                    status = None
-                    if url.endswith('/'):
-                        status = GetObjectFileStatus(item)
-                        size = await status.size()
-                        if size != 0:
-                            raise FileAndDirectoryError(url)
-                        yield GoogleStorageFileListEntry(url, None)
-                    else:
-                        yield GoogleStorageFileListEntry(url, item, status)
+                    yield GoogleStorageFileListEntry(f'gs://{bucket}/{item["name"]}', item)
 
     async def listfiles(self, url: str, recursive: bool = False) -> AsyncIterator[FileListEntry]:
         bucket, name = self._get_bucket_name(url)
@@ -724,13 +681,22 @@ class GoogleStorageAsyncFS(AsyncFS):
         except StopAsyncIteration:
             raise FileNotFoundError(url)  # pylint: disable=raise-missing-from
 
+        async def is_file(entry):
+            url = await entry.url()
+            if url.endswith('/'):
+                stat = await entry.status()
+                if await stat.size() != 0:
+                    raise FileAndDirectoryError(url)
+                return False
+            return True
+
         async def cons(first_entry, it):
-            if await first_entry.is_file():
+            if await is_file(first_entry):
                 yield first_entry
             try:
                 while True:
                     next_entry = await it.__anext__()
-                    if await next_entry.is_file():
+                    if await is_file(next_entry):
                         yield next_entry
             except StopAsyncIteration:
                 pass
@@ -782,7 +748,10 @@ class GoogleStorageAsyncFS(AsyncFS):
 
     async def rmtree(self, sema: asyncio.Semaphore, url: str) -> None:
         async with OnlineBoundedGather2(sema) as pool:
-            it = self._listfiles_all(url)
+            bucket, name = self._get_bucket_name(url)
+            if name and not name.endswith('/'):
+                name = f'{name}/'
+            it = self._listfiles_recursive(bucket, name)
             async for entry in it:
                 await pool.call(self._remove_doesnt_exist_ok, await entry.url())
 
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -681,26 +681,22 @@ class GoogleStorageAsyncFS(AsyncFS):
         except StopAsyncIteration:
             raise FileNotFoundError(url)  # pylint: disable=raise-missing-from
 
-        async def is_file(entry):
-            if await entry.is_dir():
-                return False
-
+        async def should_yield(entry):
             url = await entry.url()
-            if url.endswith('/'):
+            if url.endswith('/') and await entry.is_file():
                 stat = await entry.status()
                 if await stat.size() != 0:
                     raise FileAndDirectoryError(url)
                 return False
-
             return True
 
         async def cons(first_entry, it):
-            if await is_file(first_entry):
+            if await should_yield(first_entry):
                 yield first_entry
             try:
                 while True:
                     next_entry = await it.__anext__()
-                    if await is_file(next_entry):
+                    if await should_yield(next_entry):
                         yield next_entry
             except StopAsyncIteration:
                 pass
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -363,6 +363,7 @@ class StorageClient(BaseClient):
         # https://cloud.google.com/storage/docs/performing-resumable-uploads
         assert upload_type == 'resumable'
         chunk_size = kwargs.get('bufsize', 256 * 1024)
+
         resp = await self._session.post(
             f'https://storage.googleapis.com/upload/storage/v1/b/{bucket}/o',
             **kwargs)
@@ -424,7 +425,7 @@ class GetObjectFileStatus(FileStatus):
 
 class GoogleStorageFileListEntry(FileListEntry):
     def __init__(self, url: str, items: Optional[Dict[str, Any]], status=None):
-        assert url.endswith('/') == (items is None), f'{url} {items}'
+        # assert url.endswith('/') == (items is None), f'{url} {items}'
         self._url = url
         self._items = items
         self._status: Optional[GetObjectFileStatus] = status
@@ -632,7 +633,7 @@ class GoogleStorageAsyncFS(AsyncFS):
                 raise FileNotFoundError(url) from e
             raise
 
-    async def _listfiles_recursive(self, bucket: str, name: str) -> AsyncIterator[FileListEntry]:
+    async def _listfiles_recursive(self, bucket: str, name: str, force_all: bool) -> AsyncIterator[FileListEntry]:
         assert not name or name.endswith('/')
         params = {
             'prefix': name
@@ -645,16 +646,19 @@ class GoogleStorageAsyncFS(AsyncFS):
             if items is not None:
                 for item in page['items']:
                     url = f'gs://{bucket}/{item["name"]}'
+                    print(f'url={url}')
                     status = None
-                    if url.endswith('/') and item is not None:
+                    if not force_all and url.endswith('/') and item is not None:
                         status = GetObjectFileStatus(item)
                         size = await status.size()
                         if size != 0:
+                            print(f'non-empty {url} with size {size}')
                             raise FileAndDirectoryError(url)
+                        print(f'skipping {url} with size {size}')
                         continue
                     yield GoogleStorageFileListEntry(url, item, status)
 
-    async def _listfiles_flat(self, bucket: str, name: str) -> AsyncIterator[FileListEntry]:
+    async def _listfiles_flat(self, bucket: str, name: str, force_all: bool) -> AsyncIterator[FileListEntry]:
         assert not name or name.endswith('/')
         params = {
             'prefix': name,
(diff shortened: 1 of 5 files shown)


hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -652,7 +652,14 @@ async def _listfiles_recursive(self, bucket: str, name: str) -> AsyncIterator[Fi
             items = page.get('items')
             if items is not None:
                 for item in page['items']:
-                    yield GoogleStorageFileListEntry(f'gs://{bucket}/{item["name"]}', item)
+                    url = f'gs://{bucket}/{item["name"]}'
+                    if url.endswith('/') and item is not None:
```
cseed, line 656: `item` cannot be None: we subscript it in the line above.

I'd write this like this:

```
  entry = GoogleStorageFileListEntry(url, item)
  if url.endswith('/'):
    status = await entry.status()
    size = await entry.size()
    if size != 0:
      raise FileAndDirectoryError(url)
  yield entry
```

The entry caches the status, so this way, if the user gets the status, it reuses the status object you create.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -634,6 +639,10 @@ async def copy_as_dir(self, sema: asyncio.Semaphore, source_report: SourceReport
             self.src_is_dir = False
             await self.release_barrier()
             return
+        except:
+            self.encountered_error = True
+            await self.release_barrier()
+            raise
```
cseed, line 645: I think I'd write this like this:

```
    async def copy_as_dir(self, sema: asyncio.Semaphore, source_report: SourceReport, return_exceptions: bool):
        try:
	    src = self.src
	    if not src.endswith('/'):
		src = src + '/'
	    try:
		srcentries = await self.router_fs.listfiles(src, recursive=True)
	    except (NotADirectoryError, FileNotFoundError):
		self.src_is_dir = False
		return
	    self.src_is_dir = True
        finally:
            await self.release_barrier()

        await self.barrier.wait()

        if self.src_is_file is not None and self.src_is_file:
            raise FileAndDirectoryError(self.src)

        ...
```

This makes the release of the barrier clearer.  Then we can remove release_barrier_and_wait.  We need the `self.src_is_file is not None` in the case `copy_as_file` got an error.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -673,11 +682,13 @@ async def copy(self, sema: asyncio.Semaphore, source_report: SourceReport, retur
                 return_exceptions=True)
 
             assert self.pending == 0
-            assert (self.src_is_file is None) == self.src.endswith('/')
-            assert self.src_is_dir is not None
 
-            if (self.src_is_file is False or self.src.endswith('/')) and not self.src_is_dir:
-                raise FileNotFoundError(self.src)
+            if not self.encountered_error:
```
cseed, line 686: Then you can put the check for errors above these checks, and get rid of encountered_error.

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -416,3 +417,22 @@ async def test_copy_src_parts(copy_test_context):
...
+async def test_file_and_directory_error_with_slash(router_filesystem):
+    sema, fs, bases = router_filesystem
+
+    src_base = await fresh_dir(fs, bases, 'gs')
+    dest_base = await fresh_dir(fs, bases, 'gs')
+
+    await fs.create(f'{src_base}empty/')
+
+    async with await fs.create(f'{src_base}not-empty/') as f:
+        await f.write('foo'.encode('utf-8'))
```
cseed, line 432: Or just `b'foo'`.

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -416,3 +417,22 @@ async def test_copy_src_parts(copy_test_context):
...
+
+    with pytest.raises(FileNotFoundError):
+        await fs.copy(sema, Transfer(f'{src_base}empty/', dest_base.rstrip('/')))
```
cseed, line 435: Might also be nice to check this works as expected when the empty file isn't at the toplevel.  Like, have src_base/a/{foo, empty/} and copy src_base/a and make sure foo copies but not empty/.

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -416,3 +417,22 @@ async def test_copy_src_parts(copy_test_context):
...
+
+    with pytest.raises(FileAndDirectoryError):
+        await fs.copy(sema, Transfer(f'{src_base}not-empty/', dest_base.rstrip('/')))
```
cseed, line 438: Same, not at top level.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -608,6 +609,10 @@ async def copy_as_file(self,
             self.src_is_file = False
             await self.release_barrier()
             return
+        except:
```
cseed, line 612: Can I ask, what unhandled error did we see here?  Just so I can keep this case in mind in the future.

cseed (changes requested):

hail/python/hailtop/aiotools/fs.py
jigold: Lindo's service account didn't have permissions to access the files. So he got a google 403 exception. But the error message was `assert pending == 0`

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -416,3 +417,34 @@ async def test_copy_src_parts(copy_test_context):
...
+    await fs.create(f'{src_base}a/empty/')
+    await fs.create(f'{src_base}a/foo')
+
+    async with await fs.create(f'{src_base}b/not-empty/') as f:
+        await f.write(b'not-empty')
+
+    async with await fs.create(f'{src_base}b/bar') as f:
+        await f.write(b'bar')
+
+    await fs.copy(sema, Transfer(f'{src_base}a/foo', dest_base.rstrip('/')))
```
cseed, line 438: What's this testing?

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -416,3 +417,34 @@ async def test_copy_src_parts(copy_test_context):
...
+
+@pytest.mark.asyncio
+async def test_file_and_directory_error_with_slash(router_filesystem):
+    sema, fs, bases = router_filesystem
+
+    src_base = await fresh_dir(fs, bases, 'gs')
+    dest_base = await fresh_dir(fs, bases, 'gs')
+
+    await fs.create(f'{src_base}a/empty/')
+    await fs.create(f'{src_base}a/foo')
```
cseed, line 430: Sorry, my previous comments weren't clear.  Let me try again:

I'm imagining you have:

 - {src_base}/empty/ (which is an empty file) and {src_base}/empty/foo
 - {src_base}/non-empty/ (which is a non-empty file) and {src_base}/non-empty/bar
 - {src_base}/empty-only/  (which is empty, but empty-only has no other files in it)

Then copy empty, and verify that the empty directory file _wasn't_ copied, and foo was.

Then verify copying from `non-empty/` gets a FileAndDirectoryError.

Then copy empty-only/.  I think this should work (it is a directory).  If you're copying to the local filesystem, it should create the target directory (I suspect we're not doing that now).

```
$ mkdir scratch; cd scratch
$ mkdir a
$ cp -a a b
$ find .
.
./b
./a
```

cseed (changes requested):
Made comment.

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -416,3 +417,40 @@ async def test_copy_src_parts(copy_test_context):
...
+    await fs.copy(sema, Transfer(f'{src_base}not-empty/bar', dest_base.rstrip('/')))
+    await expect_file(fs, f'{dest_base}not-empty/bar', 'bar')
+
+    with pytest.raises(FileAndDirectoryError):
+        await fs.copy(sema, Transfer(f'{src_base}not-empty/', dest_base.rstrip('/')))
+
+    # Test empty file that ends in a slash
+    await fs.create(f'{src_base}empty-only/')
+
+    with pytest.raises(FileNotFoundError):
```
cseed, line 455: I think this shouldn't be FileNotFound.  {src_base}empty_only/ exists and is a directory.  This should work, see my cp example in another comment.

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -416,3 +417,40 @@ async def test_copy_src_parts(copy_test_context):
...
+
+
+@pytest.mark.asyncio
+async def test_file_and_directory_error_with_slash(router_filesystem):
+    sema, fs, bases = router_filesystem
+
+    src_base = await fresh_dir(fs, bases, 'gs')
+    dest_base = await fresh_dir(fs, bases, 'gs')
+
+    # test empty file ending in slash and directory
```
cseed, line 429: OCD nit: this has a lower case t [t]est, while the comments below have upper case Ts.

cseed (changes requested):

hail/python/test/hailtop/aiotools/test_copy.py
jigold: This is not working. The case when copying `{src_base}/empty/foo` results in FileNotFound as a Google Exception.

jigold: Wait, sorry the Google Exception happens in the test's assert, not in the copy code.

jigold:
I'm going to split this into two PRs and I'll figure out why the file ending in a slash code isn't working when I get back.

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -416,3 +417,38 @@ async def test_copy_src_parts(copy_test_context):
...
+
+    await fs.copy(sema, Transfer(f'{src_base}not-empty/bar', dest_base.rstrip('/')))
+    await expect_file(fs, f'{dest_base}not-empty/bar', 'bar')
+
+    with pytest.raises(FileAndDirectoryError):
+        await fs.copy(sema, Transfer(f'{src_base}not-empty/', dest_base.rstrip('/')))
+
+    # Test empty file that ends in a slash
+    await fs.create(f'{src_base}empty-only/')
+    await fs.copy(sema, Transfer(f'{src_base}empty-only/', dest_base.rstrip('/')))
```
cseed, line 454: You should do this with a dest that doesn't exist, and verify it gets created.

cseed (changes requested):

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -416,3 +417,38 @@ async def test_copy_src_parts(copy_test_context):
...
+
+    # Test empty file ending in slash and directory
+    await fs.create(f'{src_base}empty/')
+    async with await fs.create(f'{src_base}empty/foo') as f:
+        await f.write(b'foo')
+
+    await fs.copy(sema, Transfer(f'{src_base}empty/', dest_base.rstrip('/')))
+    await expect_file(fs, f'{dest_base}empty/foo', 'foo')
+    with pytest.raises(AssertionError):
+        await expect_file(fs, f'{dest_base}empty/', '')
```
cseed, line 437: Don't expect_file something that doesn't exist.  Just do `assert not fs.exists(file)`.

hail/python/test/hailtop/aiotools/test_copy.py
jigold: It already doesn't exist. It's `empty-only` which is different than `empty`

jigold:
I don't think these tests will pass. I'll address the issues when I come back from vacation. Feel free to reopen if you want.

hail/python/test/hailtop/aiotools/test_copy.py
cseed: The code creates dest_base, see copy_test_context:

```
    # make sure dest_base exists
    async with await fs.create(f'{dest_base}keep'):
        pass
```

jigold:
Do I need to do something on my end in order to test copying from my computer rather than through CI? I keep getting a 400 regardless of whether I use my bucket `hail-jigold` or the test bucket `hail-test-dmk9z`. The only information I can get is "Bad Request" which I think is from this:

> The request cannot be completed based on your current Cloud Storage settings. For example, you cannot lock a retention policy if the requested bucket doesn't have a retention policy, and you cannot set ACLs if the requested bucket has Bucket Policy Only enabled.

I don't think this is the real error as we know copying works fine with the Hail test bucket.

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -423,11 +424,10 @@ async def __getitem__(self, key: str) -> str:
 
 
 class GoogleStorageFileListEntry(FileListEntry):
-    def __init__(self, url: str, items: Optional[Dict[str, Any]]):
-        assert url.endswith('/') == (items is None), f'{url} {items}'
+    def __init__(self, url: str, items: Optional[Dict[str, Any]], status=None):
```
cseed, line 427: Type hint on `status`

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -644,9 +644,17 @@ async def _listfiles_recursive(self, bucket: str, name: str) -> AsyncIterator[Fi
             items = page.get('items')
             if items is not None:
                 for item in page['items']:
-                    yield GoogleStorageFileListEntry(f'gs://{bucket}/{item["name"]}', item)
-
-    async def _listfiles_flat(self, bucket: str, name: str) -> AsyncIterator[FileListEntry]:
+                    url = f'gs://{bucket}/{item["name"]}'
+                    status = None
+                    if not force_all and url.endswith('/') and item is not None:
```
cseed, line 649: all is a bit general when you're just controlling empty directory files.  What about `skip_empty_directory_files` (or `include_empty_directory_files` if you want the positive sense, although personally I find the sense that doesn't require a `not` to be more readable.)

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -659,22 +667,34 @@ async def _listfiles_flat(self, bucket: str, name: str) -> AsyncIterator[FileLis
                 for prefix in prefixes:
                     assert prefix.endswith('/')
                     url = f'gs://{bucket}/{prefix}'
+                    if not force_all and await self.isfile(url):
```
cseed, line 670: I think this is wrong.  We should return directories that contain only empty files.

hail/python/hailtop/aiogoogle/client/storage_client.py
cseed: Also, I don't think `item is not None` is needed here.  Item shouldn't be able to be None (let's add an assertion if we're unsure).

hail/python/hailtop/aiotools/fs.py
```diff
@@ -632,11 +632,18 @@ async def copy_as_dir(self, sema: asyncio.Semaphore, source_report: SourceReport
...
+            srcentries = None
+
             try:
                 srcentries = await self.router_fs.listfiles(src, recursive=True)
-            except (NotADirectoryError, FileNotFoundError):
-                self.src_is_dir = False
-                return
+            except (NotADirectoryError, FileNotFoundError) as e:
+                if isinstance(e, FileNotFoundError):
+                    if await self.router_fs.isdir(src):
```
cseed, line 642: If possible, I really don't want to have to do additional API requests after getting the listings.  So we should make the listfiles interface operate in a way that lets us do the right thing here without more calls.

In addition to your questions, another one is how listfiles behaves in the presence of empty directory files.  I think listfiles shouldn't throw NotADirectoryError/FileNotFoundError if the directory only has an empty directory file.

cseed (changes requested):

hail/python/hailtop/aiogoogle/client/storage_client.py
jigold: See my comment above. For `rmtree` where `force_all` is True, then we do want to return non-empty directory/file files.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -665,9 +672,12 @@ async def copy_source(srcentry):
 
             await self._copy_file_multi_part(sema, source_report, srcfile, await srcentry.status(), url_join(full_dest, relsrcfile), return_exceptions)
 
-        await bounded_gather2(sema, *[
-            copy_source(srcentry)
-            async for srcentry in srcentries], cancel_on_error=True)
+        if srcentries is None:
```
jigold, line 675: I'm uneasy with the empty directory creation. Even if we handle this use case where a top level directory is an empty file/directory, there's still downstream empty file/directory being missed completely and we won't make those empty directories on the local file system where mkdir is not a noop.

For example,

path/foo/ is the only match

versus.

path/bar/foo/ and the query is path/bar/.

hail/python/hailtop/aiogoogle/client/storage_client.py
cseed: The reason I don't like `all` is that it isn't clear what files are excluded when you set `include_all_files=False`.

cseed (changes requested):
Getting close, some additional suggestions to simplify the code.

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -416,3 +417,135 @@ async def test_copy_src_parts(copy_test_context):
...
+    await fs.listfiles(f'{src_base}empty-only/')
+    await fs.listfiles(f'{src_base}empty-only/', recursive=True)
+
+    for transfer_type in (Transfer.DEST_IS_TARGET, Transfer.DEST_DIR, Transfer.INFER_DEST):
+        dest_base = await fresh_dir(fs, bases, 'gs')
+        await fs.copy(sema, Transfer(f'{src_base}empty-only/', dest_base.rstrip('/'), treat_dest_as=transfer_type))
+
+        # We ignore empty directories when copying
+        with pytest.raises(FileNotFoundError):
+            await fs.listfiles(f'{dest_base}empty-only/')
```
cseed, line 516: I think this one is wrong now?

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -416,3 +417,135 @@ async def test_copy_src_parts(copy_test_context):
...
+
+    src_base = await fresh_dir(fs, bases, 'gs')
+
+    await write_file(fs, f'{src_base}empty/', '')
+    await write_file(fs, f'{src_base}empty/foo', b'foo')
+
+    await fs.listfiles(f'{src_base}')
+    await fs.listfiles(f'{src_base}', recursive=True)
+    await fs.listfiles(f'{src_base}empty/')
+    await fs.listfiles(f'{src_base}empty/', recursive=True)
```
cseed, line 439: `listfiles` returns an async generator, and you won't test the full behavior unless you drain the generator.

cseed (dismissed):
Code is looking good!  I know the back and forth is tedious, but I think the chance turned out nice in the end.

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -682,11 +681,27 @@ async def listfiles(self, url: str, recursive: bool = False) -> AsyncIterator[Fi
         except StopAsyncIteration:
             raise FileNotFoundError(url)  # pylint: disable=raise-missing-from
 
+        async def is_file(entry):
+            if await entry.is_dir():
+                return False
```
cseed, line 686: So I think `is_file` is a bad name.  This should be something like `skip` or `should_yield`?  Because you want to return directories here.  You only want to skip empty-file-directories.

jigold: Why would I want to return directories? We're listing files (or at least that's what I thought it was). 

I also broke listfiles in other tests:

```
        async def listfiles(dir, recursive):
            return {(await entry.url(), await entry.is_file()) async for entry in await fs.listfiles(dir, recursive)}
    
        assert await listfiles(f'{base}foo/', recursive=True) == {(a, True), (c, True)}
>       assert await listfiles(f'{base}foo/', recursive=False) == {(a, True), (b, False)}
E       AssertionError: assert {('gs://hail-test-dmk9z/tmp/QPTOQ0djdUWM3WYNHTbKew/foo/a', True)} == {('gs://hail-test-dmk9z/tmp/QPTOQ0djdUWM3WYNHTbKew/foo/a', True),\n ('gs://hail-test-dmk9z/tmp/QPTOQ0djdUWM3WYNHTbKew/foo/b/', False)}
E         Extra items in the right set:
E         ('gs://hail-test-dmk9z/tmp/QPTOQ0djdUWM3WYNHTbKew/foo/b/', False)
E         Full diff:
E           {
E            ('gs://hail-test-dmk9z/tmp/QPTOQ0djdUWM3WYNHTbKew/foo/a',
E             True),
E         -  ('gs://hail-test-dmk9z/tmp/QPTOQ0djdUWM3WYNHTbKew/foo/b/',
E         -   False),
E           }
```

```
        assert await listfiles(f'{base}foo/', recursive=True) == {(a, True), (c, True)}
>       assert await listfiles(f'{base}foo/', recursive=False) == {(a, True), (b, False)}
E       AssertionError: assert {('gs://hail-test-dmk9z/tmp/fOxVwBRMRHj37hvn8H20i1/foo/a', True)} == {('gs://hail-test-dmk9z/tmp/fOxVwBRMRHj37hvn8H20i1/foo/a', True),\n ('gs://hail-test-dmk9z/tmp/fOxVwBRMRHj37hvn8H20i1/foo/b/', False)}
E         Extra items in the right set:
E         ('gs://hail-test-dmk9z/tmp/fOxVwBRMRHj37hvn8H20i1/foo/b/', False)
E         Full diff:
E           {
E            ('gs://hail-test-dmk9z/tmp/fOxVwBRMRHj37hvn8H20i1/foo/a',
E             True),
E         -  ('gs://hail-test-dmk9z/tmp/fOxVwBRMRHj37hvn8H20i1/foo/b/',
E         -   False),
E           }

```

jigold: Maybe these tests need to be fixed rather than my code being wrong.

cseed: Hmm, no, the code is definitely wrong.  `listfiles(..., recursive=False)` should definitely be returning **directories**.

cseed: Let me clarify a bit:

`listfiles(..., recursive=True)` automatically descends into directories and only returns files.

`listfiles(..., recursive=False)` acts like listing directories on UNIX, and only returns things directly inside a directory, whether it be files or more directories.

Google Storage listing API supports both of these modes directly.  UNIX directory listing only implements `recursive=False`, and we iterate recursively through directories explicit in the LocalAsyncFS backend.

We don't make much use of `recursive=False`, but I think users will want it when we expose this to them (replacing hl.hadoop_ls, etc.)

Does that help?

cseed: Obviously, I need to document the AsyncFS API ASAP.

cseed: And maybe rename `listfiles` to `listdir`?  We're listing the contents of a directory.

jigold: I think the equivalent operations of what we're doing is listfiles_flat is the equivalent of scandir in Python. listfiles_recursive doesn't have an equivalent operation in unix or python. I looked at `ls -R` but that lists both file and directory entries. `listfiles` makes sense to me for `listfiles_recursive`. `listdir` makes sense to me for `listfiles_flat`. Can we have both operations be top-level things rather than the `recursive=True` flag?

cseed: > Can we have both operations be top-level things rather than the recursive=True flag?

Yeah, definitely.  I'll put it on the list of things to cleanup and refine when I start documenting.

jigold:
bump

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -424,7 +425,6 @@ async def __getitem__(self, key: str) -> str:
 
 class GoogleStorageFileListEntry(FileListEntry):
     def __init__(self, url: str, items: Optional[Dict[str, Any]]):
-        assert url.endswith('/') == (items is None), f'{url} {items}'
```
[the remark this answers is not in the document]
cseed: Why did you remove this?  This assertion is still good, right?

cseed (dismissed):

hail/python/hailtop/aiogoogle/client/storage_client.py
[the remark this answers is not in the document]
jigold: No. It's basically representing a directory if items is None. We now have directory entries in our list, but we filter them out in `listfiles`. This is the central issue that needed to be fixed (user would hit this assertion error when they had both a file and a directory).

https://github.com/hail-is/hail/blob/f408e2734483befad8aaba9bb4ef2df5f017a161/hail/python/hailtop/aiogoogle/client/storage_client.py#L662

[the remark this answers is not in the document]
cseed: Ah, got it, thanks for clarifying!

cseed (approved):

PR #10221, opened 2021-03-24
[batch] Add more logging and fix site activation in worker

jigold:


danking (approved):
I want to talk more about the retry_long_running change, but I pushed a reversion of that and will approve the rest.

PR #10222, opened 2021-03-24
[website] make www the website service and base path

daniel-goldstein:
website is the only service that doesn't follow the subdomain == base_path convention, nor does the subdomain match the k8s service, which is website. This enforces that by adding a `www` k8s service and changes the base_path in `website.py`.

### DIFF
--- a/tls/create_certs.py
+++ b/tls/create_certs.py
@@ -28,19 +28,17 @@ def echo_check_call(cmd):
 
 def create_key_and_cert(p):
     name = p['name']
-    domain = p['domain']
+    domains = p['domains']
     unmanaged = p.get('unmanaged', False)
     if unmanaged and namespace != 'default':
-        return
+        return {}
     key_file = f'{name}-key.pem'
     csr_file = f'{name}-csr.csr'
     cert_file = f'{name}-cert.pem'
     key_store_file = f'{name}-key-store.p12'
-    names = [
-        domain,
-        f'{domain}.{namespace}',
-        f'{domain}.{namespace}.svc.cluster.local'
-    ]
+    names = [domain_name
+             for domain in domains
+             for domain_name in [domain, f'{domain}.{namespace}', f'{domain}.{namespace}.svc.cluster.local']]
 
     echo_check_call([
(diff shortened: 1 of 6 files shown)


daniel-goldstein:
Ok there's another change in here now. `tls/config.yaml` now allows multiple domains listed for a single cert, which are added as alternate names. The first domain listed is used as the Common Name (CN), though that should hopefully not matter.

tls/create_certs.py
```diff
@@ -28,19 +28,17 @@ def echo_check_call(cmd):
 
 def create_key_and_cert(p):
     name = p['name']
-    domain = p['domain']
+    domains = p['domains']
     unmanaged = p.get('unmanaged', False)
     if unmanaged and namespace != 'default':
-        return
+        return {}
```
danking, line 34: this should probably return None or something? an empty dict feels a bit odd to me.

website/deployment.yaml
```diff
@@ -110,3 +110,18 @@ spec:
   selector:
     matchLabels:
       app: website
+---
```
danking, line 113: I'm not sure where we are on the router elimination, but shouldn't we be modifying router's deployment.yaml?

danking (changes requested):

website/deployment.yaml
daniel-goldstein: This is all a bit annoyingly interweaved. Right now router's `deployment.yaml` declares the `website` service, and routes traffic to `www` to `https://website/`. If this PR goes in now, this `www` service is entirely unused in default, but gateway will be able to pick it up for `internal.hail.is/<namespace>/www`. When I delete router, I just won't copy over the `website` service from router's deployment.yaml to this one. Does that make sense?

tls/create_certs.py
daniel-goldstein: I did this because the only use of `create_key_and_certs` is a dict spread, and it seemed the most uniform in terms of return type, but don't have a problem changing it if you think we should.

danking (approved):

PR #10226, opened 2021-03-25
[copy] Fix assertion error on exceptions

jigold:
Split from #10218


jigold:
Tests are passing now.

cseed (approved):
Sorry for the delay on this.  Solid change!

PR #10227, opened 2021-03-25
[query] Fixed bug where hl.nd.qr didn't support 0 elements

johnc1231:
I'm not sure this is the right fix.  The segfault was coming from dgeqrf itself, and it was happening only when `N` is 0 (`M` being 0 was fine). The real issue is that when `N` is 0, LAPACK computes `LWORK` to be 0, meaning we don't actually allocate any memory for the `WORK` array and we segfault. To contrast, if I do a workspace size query for shape `(0, 10)`, I get `320` for `LWORK`. This also seems wrong though, as you shouldn't really need any work space for an empty matrix? 

Anyway, the segfault preventing fix was just to make sure LWORK was at least 1. I was hoping you (Patrick) could use Julia interface to play with the LAPACK functions and see what behavior you notice, and if it's consistent with the above.

### DIFF
--- a/hail/python/test/hail/expr/test_ndarrays.py
+++ b/hail/python/test/hail/expr/test_ndarrays.py
@@ -743,6 +743,16 @@ def test_ndarray_qr():
 
     assert_same_qr(identity4, np_identity4)
 
+    np_size_zero_n = np.zeros((10, 0))
+    size_zero_n = hl.nd.zeros((10, 0))
+
+    assert_same_qr(size_zero_n, np_size_zero_n)
+
+    np_size_zero_m = np.zeros((0, 10))
+    size_zero_m = hl.nd.zeros((0, 10))
+
+    assert_same_qr(size_zero_m, np_size_zero_m)
+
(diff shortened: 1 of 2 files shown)


patrick-schultz:
LAPACK/MKL documentation says `WORK` must be an array of length `max(1, LWORK)`. I think the "official" way to fix this is to leave `LWORK` as whatever the workspace query returns, and allocate the `WORK` arrray of size `max(1, LWORK)`.

patrick-schultz:
That's weird about the M=0 case. The LAPACK version (source [here](http://www.netlib.org/lapack/explore-html/dd/d9a/group__double_g_ecomputational_ga3766ea903391b5cf9008132f7440ec7b.html#ga3766ea903391b5cf9008132f7440ec7b)) returns `LWORK=0` if either M or N is 0. But it doesn't seem worth complicating things over.

johnc1231:
Ah ok, I'll just change the allocation and leave `LWORK` alone.

patrick-schultz (approved):

PR #10228, opened 2021-03-25
[website] add fontawesome!

danking:
I also changed the highlight color for the GitHub icon because it was too dark to clearly see.

The username and password for our Font Awesome account are in the usual place.

I also had a bit too much fun using "GIPHY CAPTURE", Giphy's Mac video capture app to create this gif of my change:
highlighticon


CDiaz96 (approved):
Yay! I am glad you were able to do this. Would love to hear more about how it was done. Doesn't seem like you had to change many files.

PR #10229, opened 2021-03-25
[query] New stream design

tpoterba:
stacked on #10208


tpoterba:
Benchmarks: https://gist.github.com/tpoterba/7b41c0e8c48ea7b1dec03dee8bce1c72

It looks like this change made Hail slower, but I'm pretty sure this is noise.

These:
```
                                        join_p1000_p10     130.8%    132.532    173.404
                                       union_p1000_p10     129.7%    107.998    140.027
                                       union_p10_p1000     128.6%    116.013    149.201
```                                       

Don't have any stream IRs!

tpoterba:
I also compared `variant_and_sample_qc_nested_with_filters_2` (33% worse on batch) between the two branches on my laptop, and could not detect a difference.

I do think the make_ndarray range speedup is real -- there are a few benchmarks that indicate improvement that all are heavily dependent on the performance of the `StreamRange` implementation, which I think slightly improved.

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
```diff
@@ -1055,6 +1052,24 @@ class EmitMethodBuilder[C](
       res.code
     })
   }
+
+  def implementLabel(label: CodeLabel)(f: EmitCodeBuilder => Unit): Unit = {
+    EmitCodeBuilder.scopedVoid(this) { cb =>
+      cb.define(label)
+      f(cb)
+//      assert(!cb.isOpenEnded)
```
cseed, line 1060: FYI

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
```diff
@@ -1055,6 +1052,24 @@ class EmitMethodBuilder[C](
...
+    }
+  }
+
+  def defineHangingLabel(f: EmitCodeBuilder => Unit): CodeLabel = {
+    val label = CodeLabel()
+    EmitCodeBuilder.scopedVoid(this) { cb =>
+      cb.define(label)
+      f(cb)
+//      assert(!cb.isOpenEnded)
```
cseed, line 1069: FYI

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
```diff
@@ -1055,6 +1052,24 @@ class EmitMethodBuilder[C](
...
+
+  def implementLabel(label: CodeLabel)(f: EmitCodeBuilder => Unit): Unit = {
+    EmitCodeBuilder.scopedVoid(this) { cb =>
+      cb.define(label)
+      f(cb)
+//      assert(!cb.isOpenEnded)
+    }
+  }
+
+  def defineHangingLabel(f: EmitCodeBuilder => Unit): CodeLabel = {
```
cseed, line 1064: Consider `defineAndImplementLabel`?

hail/src/main/scala/is/hail/expr/ir/EmitCodeBuilder.scala
```diff
@@ -82,24 +84,48 @@ class EmitCodeBuilder(val emb: EmitMethodBuilder[_], var code: Code[Unit]) exten
...
   def memoizeField[T](ec: EmitCode, name: String): EmitValue = {
+    require(ec.pt.isRealizable)
     val l = emb.newEmitField(name, ec.pt)
     l.store(this, ec)
     l
   }
 
+  private[this] def defineNestedUnusedLabels(x: EmitCode): Unit = {
+    x.pv match {
+      case SStreamCode(_, producer) => StreamProducer.markUnused(producer, emb)
```
cseed, line 109: I kind of don't like this but I'm not sure if I can improve it.  What happens if, instead, consumers who are expecting a stream or a realizable value switch on the type there and explicitly do the markUnused.

 > defineNestedUnusedLabels
 > markUnused

Should `markUnused` be called `defineUnused`?

hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
```diff
@@ -240,19 +236,21 @@ case class SplitPartitionNativeWriter(
...
-        cb.assign(result, pResultType.allocate(region.code))
+        cb.assign(result, pResultType.allocate(region))
         if (hasIndex)
           indexWriter.close(cb)
         cb += ob1.flush()
         cb += ob2.flush()
         cb += os1.invoke[Unit]("close")
         cb += os2.invoke[Unit]("close")
-        filenameType.storeAtAddress(cb, pResultType.fieldOffset(result, "filePath"),region.code, pctx, false)
+        filenameType.storeAtAddress(cb, pResultType.fieldOffset(result, "filePath"),region, pctx, false)
```
cseed, line 253: space before region

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -0,0 +1,2259 @@
...
+  def initialize(cb: EmitCodeBuilder): Unit
+
+  /**
+    * Stream element region, into which the `element` is emitted. The assignment, clearing,
+    * and freeing of the element region is the responsibility of the stream consumer.
+    */
+  val elementRegion: Settable[Region]
+
+  /**
+    * This boolean parameter indicates whether the producer's elements should be emitted into
```
cseed, line 66: > emitted

Allocated?  Emitting is putting code into a code builder at compile time.

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -0,0 +1,2259 @@
...
+    * separate regions (by clearing when elements leave a consumer's scope). This parameter
+    * propagates bottom-up from producers like [[ReadPartition]] and [[StreamRange]], but
+    * it is the responsibility of consumers to implement the right memory management semantics
+    * based on this flag.
+    */
+  val separateRegions: Boolean
```
cseed, line 72: This is minor, but I'm interpreting this to mean "the elements of this stream cannot safely be stored in memory at once".  Using separate regions is one solution (as is using one region and clearing it).  Rather than naming this after the (a) implementation strategy, I think a descriptive property of the stream is clearer.  What about `largeStream: Boolean`?

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -0,0 +1,2259 @@
...
+    cb.goto(this.LproduceElement)
+
+    cb.define(this.LendOfStream)
+    this.close(cb)
+  }
+
+  // only valid if `perElement` does not retain pointers into the element region after it returns (or adds region references)
+  final def memoryManagedConsume(outerRegion: Value[Region], cb: EmitCodeBuilder, setup: EmitCodeBuilder => Unit = _ => ())(perElement: EmitCodeBuilder => Unit): Unit = {
+
+    if (separateRegions)
```
cseed, line 125: If separateRegions is false, can you just fall back to `unmanagedConsume`?

cseed (changes requested):
Some minor initial things.

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
tpoterba: These are assertions I would *like* to add here, but they fail in many of the usages in stream code. I think patterns like:
```
cb.ifx(b, 
 cb.goto(L1),
 cb.goto(L2))
  ```
leave the codebuilder open ended. I'll add a comment that these are aspirational assertions that should be reenabled when we fix the codebuilder state.

File: hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala (hunk 7 above)
Comment (tpoterba) [reply hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala:72]: yeah, how about "requiresMemoryManagementPerElement"?

File: hail/src/main/scala/is/hail/expr/ir/Emit.scala [hunk 9]
```diff
@@ -2189,48 +2156,37 @@ class Emit[C](
...
-          val x = mb.newEmitField(name, valueType)
-          EmitCodeBuilder.scopedEmitCode(mb) { cb =>
-            x.store(cb, emit(value))
-            val bodyenv = env.bind(name, x)
-            emit(body, env = bodyenv)
-          }
-      }
+      case Let(name, value, body) =>
+        EmitCode.fromI(mb) { cb =>
+          val xVal = if (value.typ.isInstanceOf[TStream]) emitStream(value, region) else emit(value)
```
cseed, line 2179: Am I wrong in thinking emitStream can now be folded into emit?

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -0,0 +1,2259 @@
...
+
+          override def close(cb: EmitCodeBuilder): Unit = {}
+        }
+        IEmitCode.missing(cb, SStreamCode(st, producer))
+
+      case Ref(name, _typ) =>
+        assert(_typ.isInstanceOf[TStream])
+        env.lookup(name).toI(cb)
+          .map(cb) { case (stream: SStreamCode) =>
+            val childProducer = stream.producer
```
cseed, line 188: Why can't you just return the child producer?

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -0,0 +1,2259 @@
...
+      case In(n, _) =>
+        // this, Code[Region], ...
+        val param = mb.getEmitParam(2 + n, outerRegion)
+        param.st match {
+          case _: SStream =>
+          case t => throw new RuntimeException(s"parameter ${ 2 + n } is not a stream! t=$t, params=${ mb.emitParamTypes }")
+        }
+        param.load.toI(cb)
+
+
```
cseed, line 224: Extra space

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -0,0 +1,2259 @@
...
+        val (fixupUnusedStreamLabels, xValue) = cb.memoizeMaybeUnrealizableField(
+          EmitCode.fromI(mb)(cb => emit(value, cb)), s"let_$name")
+        val bodyStream = produce(body, cb, env = env.bind(name, xValue))
+        fixupUnusedStreamLabels()
+        bodyStream
+
+      case In(n, _) =>
+        // this, Code[Region], ...
+        val param = mb.getEmitParam(2 + n, outerRegion)
+        param.st match {
```
cseed, line 218: `if (!param.st.isInstanceOf[SStream) ...` clearer?

cseed (dismissed):

hail/src/main/scala/is/hail/expr/ir/Compile.scala
```diff
@@ -200,63 +200,76 @@ object CompileIterator {
...
+          cb.assign(producer.elementRegion, Region.stagedCreate(Region.REGULAR, outerRegion.getPool()))
+        else
+          cb.assign(producer.elementRegion, outerRegion)
+
+        producer.initialize(cb)
+        cb.assign(didSetup, true)
+        cb.assign(eosField, false)
+      })
+
+      cb.ifx(eosField, {
```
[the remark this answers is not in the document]
tpoterba: I think we do sometimes call hasNext more than once on our iterators.

hail/src/main/scala/is/hail/expr/ir/streams/StreamUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+    val xLen = mb.newLocal[Int]("sta_len")
+    val aTyp = PCanonicalArray(stream.element.st.canonicalPType(), true)
+    stream.length match {
+      case None =>
+        val vab = new StagedArrayBuilder(stream.element.st.canonicalPType(), mb, 0)
+        writeToArrayBuilder(cb, stream, vab, destRegion)
+        cb.assign(xLen, vab.size)
+
+        aTyp.constructFromElements(cb, destRegion, xLen, deepCopy = false) { (cb, i) =>
+          IEmitCode(cb, vab.isMissing(i), PCode(aTyp.elementType, vab(i)))
```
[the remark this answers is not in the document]
tpoterba: Yes. We can fix this with an stype!

hail/src/main/scala/is/hail/expr/ir/streams/StreamUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+        case _ =>
+      }
+    }
+
+    traverse(root, 1)
+    uses min 2
+  }
+
+  def isIterationLinear(ir: IR, refName: String): Boolean =
+    multiplicity(ir, refName) <= 1
```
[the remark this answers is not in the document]
tpoterba: I'm going to leave the multiplicity stuff for a separate PR, OK?

hail/src/main/scala/is/hail/expr/ir/Emit.scala
tpoterba: Yeah, it can. I'd like to save this for a smaller PR on top of this if possible, though.

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
tpoterba: This has to do with the unused label fixup. Consider:
```
StreamMap
  <some child stream<stream<T>>
  foo
  Ref foo
```

If we just return the child producer, then our label cleanup routine defines its labels, and the actual consumer (the node on top of the streamMap) tries to define the labels again.

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
tpoterba: yeah.

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
tpoterba: good call, done.

tpoterba:
I think I addressed all your comments.

hail/src/main/scala/is/hail/expr/ir/streams/StreamUtils.scala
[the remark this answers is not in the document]
cseed: Ah, I see the problem.  This is a really an issue with the design of the array representation.  In the C++ backend, we made array a compound 3-part SType with a length, missing bits pointer and elements pointer.  That way, the representation used in array builder can actually be the same as the array representation.

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
cseed: Doesn't the IR generating the stream define unused labels if there are no uses?  There is nobody else to do it, because if there are no references, the generator of the stream is the only place anything can happen.

 > If we just return the child producer, then our label cleanup routine defines its labels

I don't understand this.  Who's running the child clean routine in this example?  I won't hold up the PR for this discussion but I feel like there's something I'm missing.

cseed (approved):

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
tpoterba: > Doesn't the IR generating the stream define unused labels if there are no uses?

Stream producers leave labels for their consumers to define. If we're emitting a StreamRange, for instance, there's no way for that node to know that its stream is unused and fix it up.

Instead, stream consumers must define all labels. We define a consumer as any IR with a [nested] stream child:
```
SomeIR
  <nested stream IR>
  ...
```

In this example, when we emit `SomeIR` (either as a realizable IEmitCode or as a stream producer IEmitCode), all labels, including nested, of `<nested stream IR>, must be defined. In the case of a `StreamMap` where we map a nested stream and the `StreamMap` body does not consume the inner stream (`(StreamMap <some_nested_stream> foo I32(1))` for instance), it is the responsibility of the `StreamMap` emitter to define the labels of the nested stream.

The subtlety comes in defining "does not consume the inner stream". I want a definition where any reference to the nested stream in `StreamMap` does consume the stream -- Including `(StreamMap some_nested "foo" (Ref "foo"))`. This makes `Ref` a consumer of a stream bound in the environment, which we implement by creating a new producer shimming the bound one.

PR #10232, opened 2021-03-26
[query] Support Spark 2.4.x and 3.0.x

johnc1231:
`json4s` was the main problem with supporting Spark 2.4.x and 3.1.1 simultaneously. I think if I remove our use of `typeHintFieldName`, we might be able to use both versions of `json4s` interchangeably, and thus use both Sparks. 

As it turns out, AWS EMR only supports Spark 3.0.1 one right now, which also uses the older version of `json4s`. It seems therefore like supporting this is going to be a necessity. 

Right now, just PRing to see if my changes pass current tests with Spark 3.


hail/src/main/scala/is/hail/expr/ir/BlockMatrixIR.scala
```diff
@@ -92,16 +92,20 @@ case class BlockMatrixRead(reader: BlockMatrixReader) extends BlockMatrixIR {
...
-      typeHintFieldName = "name")
+      List(classOf[BlockMatrixNativeReader], classOf[BlockMatrixBinaryReader], classOf[BlockMatrixPersistReader]))
   }
 
   def fromJValue(ctx: ExecuteContext, jv: JValue): BlockMatrixReader = {
-    (jv \ "name").extract[String] match {
-      case "BlockMatrixNativeReader" => BlockMatrixNativeReader.fromJValue(ctx.fs, jv)
-      case "BlockMatrixPersistReader" => BlockMatrixPersistReader.fromJValue(ctx.backendContext, jv)
-      case _ => jv.extract[BlockMatrixReader]
+    try {
```
johnc1231, line 99: Will remove, just for debugging.

PR #10233, opened 2021-03-26
[query] Loop memory management

johnc1231:
This PR changes the way loop memory management works.  

Previously, memory used while in the loop was allocated in one region, with no cleanup happening between loops. A long running loop that did a lot of allocation could therefore easily cause hail to run out of memory. This PR fixes that by creating two regions for each loop. Every iteration, we emit everything in region 1, copy the state that is necessary to region 2, then clear region 1. Then we swap the regions and repeat this until the loop terminates.

### DIFF
--- a/hail/python/test/hail/experimental/test_experimental.py
+++ b/hail/python/test/hail/experimental/test_experimental.py
@@ -439,3 +439,10 @@ class Tests(unittest.TestCase):
                 hl.if_else(my_nd[0, 0] == 1000, my_nd, f(my_nd + 1)),
                 hl.tndarray(hl.tint32, 2), hl.nd.zeros((20, 10), hl.tfloat64))
 
+    def test_loop_memory(self):
+        result = hl.experimental.loop(
+            lambda f, my_nd:
+            hl.if_else(my_nd[0, 0] == 1000, my_nd, f(my_nd + 1)),
+            hl.tndarray(hl.tfloat64, 2), hl.nd.zeros((40_000, 40_000), hl.tfloat64))
+
+        np_result = hl.eval(result)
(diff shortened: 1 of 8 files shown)


johnc1231:
I'm curious how you would test this. I've manually tested and confirmed it works with the currently commented out test. That test is kind of slow and heavy though, and it's hard to write a good version if I don't know how much memory someone has available. Do you think it's worth adding a new CI job to run certain tests in constrained memory environment to verify they stay below a prescribed limit?

hail/python/test/hail/experimental/test_experimental.py
```diff
@@ -439,3 +439,10 @@ def test_loop_errors(self):
                 hl.if_else(my_nd[0, 0] == 1000, my_nd, f(my_nd + 1)),
                 hl.tndarray(hl.tint32, 2), hl.nd.zeros((20, 10), hl.tfloat64))
 
+    # def test_loop_memory(self):
```
tpoterba, line 442: commented because slow, right? Shall we pytest.mark.skip instead?

tpoterba (changes requested):
test comment, otherwise looks great.

Can we also make sure we have a test that passes around a struct with strings in it or something that would segfault if it's used after free?

hail/python/test/hail/experimental/test_experimental.py
johnc1231: Commented because slow. Note my question above: Any good way to add a test that checks that it doesn't blow up memory that isn't going to be slow?

johnc1231:
I added a Scala test like we discussed, but it's not as thorough as I want it to be. `totalAllocatedBytes` checking just checks that we cleaned up all the intermediate garbage before returning. It doesn't ensure that we never used more than a certain amount of memory.

tpoterba:
Let's expose this. The peak memory usage on a region pool is between `allocationEchoThreshold / 2` and `allocationEchoThreshold`

johnc1231:
Done

tpoterba (approved):

PR #10235, opened 2021-03-27
[datasets] add GIANT 2018 exome array summary statistics tables

pwc2:
Adds the available [GIANT 2018 Exome Array Summary Statistics](https://portals.broadinstitute.org/collaboration/giant/index.php/GIANT_consortium_data_files#2018_Exome_Array_Summary_Statistics) datasets for WHR, BMI, and height as Hail Tables.

For reproducibility, I added the notebook I used to generate the tables and schemas. The datasets were small in this case, and I ended up doing things locally on my machine. It didn't seem to make sense to try to redo things to fit into the older extract/load workflow once everything had already been generated.


danking (changes requested):
These should all be annotation_db as well, right? AFAIK, they're locus-keyed tables right?

pwc2:
Ah, yes that's right, I'll add that to the config.

pwc2:
Should be good now.

danking (approved):

PR #10237, opened 2021-03-29
[ci] auto-assign member from team using #assign comment in PR body

daniel-goldstein:
Outside collaborators have virtually no powers on the right-hand side of the PR page w.r.t. reviewers/assignees/labels, so this is my best shot at letting them get someone assigned on their PRs. If a PR does not have any assignees or reviewers, but has #assign services or #assign compiler in the PR body, CI will randomly select a collaborator from those teams and assign them (including both will assign one person from each). Outside collaborators can re-request reviews, so I think this solves the PR review problem.

### DIFF
--- a/ci/ci/github.py
+++ b/ci/ci/github.py
@@ -7,11 +7,12 @@ import concurrent.futures
 import aiohttp
 import gidgethub
 import zulip
+import random
 
 from hailtop.config import get_deploy_config
 from hailtop.batch_client.aioclient import Batch
 from hailtop.utils import check_shell, check_shell_output, RETRY_FUNCTION_SCRIPT
-from .constants import GITHUB_CLONE_URL, AUTHORIZED_USERS, GITHUB_STATUS_CONTEXT
+from .constants import GITHUB_CLONE_URL, AUTHORIZED_USERS, GITHUB_STATUS_CONTEXT, SERVICES_TEAM, COMPILER_TEAM
 from .build import BuildConfiguration, Code
 from .globals import is_test_deployment
 
@@ -25,6 +26,10 @@ deploy_config = get_deploy_config()
 CALLBACK_URL = deploy_config.url('ci', '/api/v1alpha/batch_callback')
 
 zulip_client = zulip.Client(config_file="/zulip-config/.zuliprc")
+
+
+def select_random_teammate(team):
+    return random.choice([user for user in AUTHORIZED_USERS if team in user.teams])
 
 
 class Repo:
@@ -128,6 +133,9 @@ class MergeFailureBatch:
         self.attributes = attributes
 
 
+ASSIGN_SERVICES = '#assign services'
+ASSIGN_COMPILER = '#assign compiler'
+
 HIGH_PRIORITY = 'prio:high'
 STACKED_PR = 'stacked PR'
 WIP = 'WIP'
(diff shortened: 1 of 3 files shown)


ci/ci/github.py
```diff
@@ -318,6 +331,23 @@ async def post_github_status(self, gh_client, gh_status):
...
+                assignees.add(select_random_teammate(SERVICES_TEAM).gh_username)
+            if ASSIGN_COMPILER in self.body:
+                assignees.add(select_random_teammate(COMPILER_TEAM).gh_username)
+            data = {'assignees': list(assignees)}
+            try:
+                await gh_client.post(
+                    f'/repos/{self.target_branch.branch.repo.short_str()}/issues/{self.number}/assignees', data=data
+                )
+            except gidgethub.HTTPException:
+                log.info(f'{self.short_str()}: post assignees to github failed due to exception: {data}', exc_info=True)
```
danking, line 347: this is really an error log, right?

danking (changes requested):

ci/ci/github.py
daniel-goldstein: Ya there were actually a bunch of this littered around github.py. I updated the exceptional logs to `exception`.

ci/ci/github.py
```diff
@@ -314,10 +327,31 @@ async def post_github_status(self, gh_client, gh_status):
                 f'/repos/{self.target_branch.branch.repo.short_str()}/statuses/{self.source_sha}', data=data
             )
         except gidgethub.HTTPException:
-            log.info(f'{self.short_str()}: notify github of build state failed due to exception: {data}', exc_info=True)
+            log.exception(
+                f'{self.short_str()}: notify github of build state failed due to exception: {data}', exc_info=True
```
danking, line 331: exc_info=True is unnecessary when we use `log.exception`.

ci/ci/github.py
```diff
@@ -314,10 +327,31 @@ async def post_github_status(self, gh_client, gh_status):
...
+            if ASSIGN_COMPILER in self.body:
+                assignees.add(select_random_teammate(COMPILER_TEAM).gh_username)
+            data = {'assignees': list(assignees)}
+            try:
+                await gh_client.post(
+                    f'/repos/{self.target_branch.branch.repo.short_str()}/issues/{self.number}/assignees', data=data
+                )
+            except gidgethub.HTTPException:
+                log.exception(
+                    f'{self.short_str()}: post assignees to github failed due to exception: {data}', exc_info=True
```
danking, line 350: same here

ci/ci/github.py
```diff
@@ -459,7 +493,7 @@ async def _update_batch(self, batch_client, dbpool):
             try:
                 s = await b.status()
             except Exception:
-                log.info(f'failed to get the status for batch {b.id}', exc_info=True)
+                log.exception(f'failed to get the status for batch {b.id}', exc_info=True)
```
danking, line 496: and here and at the other ones below.

danking (changes requested):

ci/ci/github.py
```diff
@@ -518,7 +548,7 @@ async def merge(self, gh):
             )
             return True
         except (gidgethub.HTTPException, aiohttp.client_exceptions.ClientResponseError):
-            log.info(f'merge {self.target_branch.branch.short_str()} {self.number} failed', exc_info=True)
+            log.exception(f'merge {self.target_branch.branch.short_str()} {self.number} failed')
```
danking, line 551: This isn't exceptional, AFAIK. A user sending us an unmergable branch is their fault, not ours.

danking (changes requested):
One more thing

danking (approved):

PR #10239, opened 2021-03-29
[batch] fix table row templating

daniel-goldstein:
We were able to realign the users back to a vertical stack and increased the rowspan for each billing project to compensate for the `tr` of the billing project name.

HTML tables are reorganized somewhere along the way to the browser to not have nested rows. The template previously had 1 `tr` per billing project, which contained 3 `td`s, and then the user `td` had nested `tr`s. Inspecting on the dev console showed that each `tr` was pulled to the top level of the corresponding `tbody`, meaning that in total the number of `tr`s per billing project is
- 1 for the name and close button +
- `billing_project.size` user rows +
- 1 for the add button

So we changed the organization of the `tbody`s to not have nested `tr`s and adjusted the `rowspan` accordingly.

@CDiaz96

### DIFF
--- a/batch/batch/front_end/templates/billing_projects.html
+++ b/batch/batch/front_end/templates/billing_projects.html
@@ -13,16 +13,18 @@
     {% for billing_project in billing_projects %}
     <tbody>
       <tr>
-        <td rowspan="{{ billing_project.size + 1 }}">
+        <td rowspan="{{ billing_project.size + 2 }}">
           {{ billing_project['billing_project'] }}
         </td>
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
I accidentally introduced this in #10172 while removing a templating error that inadvertently masked this problem.

danking:
Excellent work! Looks much better to me in a dev deploy:


Screen Shot 2021-03-29 at 3 06 16 PM
Screen Shot 2021-03-29 at 3 06 14 PM

danking (approved):

PR #10246, opened 2021-03-29
[terraform] fixes

danking:
@cseed, I want to add a new bucket for Hail Query to use for its own purposes. I tried to add it to
Terraform. I ran into some issues which I've fixed here.

Additionally, when I run `terraform plan -var-file="$HOME/.hail/global.tfvars`, it claims I need to
create all 82 resources. Do you know what is wrong? I thought terraform would recognize our, for
example, our MySQL database.

### DIFF
--- a/infra/README.md
+++ b/infra/README.md
@@ -14,6 +14,21 @@ Instructions:
   service account name `terraform`.  Create a JSON service account key
   and place it in `$HOME/.hail/terraform_sa_key.json`.
 
+  ```
+  gcloud iam service-accounts create terraform --display-name="Terraform Account"
+  gcloud projects add-iam-policy-binding hail-vdc --member='serviceAccount:terraform@<project-id>.iam.gserviceaccount.com' --role='roles/owner'
+  gcloud iam service-accounts keys create $HOME/.hail/terraform_sa_key.json  --iam-account=terraform@<project-id>.iam.gserviceaccount.com
+  ```
+
+  When you are finished using terraform, we recommend deleting both the service account and the key
+  file.
+
+  ```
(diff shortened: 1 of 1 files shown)


infra/README.md
```diff
@@ -14,6 +14,21 @@ Instructions:
...
   and place it in `$HOME/.hail/terraform_sa_key.json`.
 
+  ```
+  gcloud iam service-accounts create terraform --display-name="Terraform Account"
+  gcloud projects add-iam-policy-binding hail-vdc --member='serviceAccount:terraform@<project-id>.iam.gserviceaccount.com' --role='roles/owner'
+  gcloud iam service-accounts keys create $HOME/.hail/terraform_sa_key.json  --iam-account=terraform@<project-id>.iam.gserviceaccount.com
+  ```
+
+  When you are finished using terraform, we recommend deleting both the service account and the key
+  file.
```
cseed, line 24: I'm not sure I recommend that.  This wasn't intended to be 1-shot, but we eventually need to get things set up so changes to the terraform rules get applied automatically, so you'll need to keep around the terraform service account and Terraform state (which is created by `terraform apply`)

cseed (changes requested):

cseed:
> Additionally, when I run terraform plan -var-file="$HOME/.hail/global.tfvars, it claims I need to create all 82 resources. Do you know what is wrong? I thought terraform would recognize our, for example, our MySQL database.

Might be easier to explain the Terraform model in person.

danking:
I removed the objectionable verbiage.

danking:
bump

danking:
bump @cseed

cseed (approved):
Oops, sorry!

PR #10247, opened 2021-03-30
[gateway, router] Route directly to k8s services from gateway

daniel-goldstein:
This is the final stop for now on our nginx tour. This builds on #10207, which altered `internal.hail.is` use k8s dns instead of router/router-resolver to proxy directly to services, and does so in default now as well. Unlike #10207, however, traffic coming in through `hail.is` is not necessarily authenticated, and we do not expose just any k8s service that happens to exist. Instead, I've altered `letsencrypt/domains.txt` to now be `letsencrypt/subdomains.txt` and templated the gateway config to generate explicit server blocks for each subdomain in `subdomains.txt`. This enforces that you cannot expose a service unless it is also listed in the `letsencrypt` directory (the dev must still remember to regenerate the certs). Now, the process for exposing a service is:
- Add a subdomain to `subdomains.txt`
- Make a k8s Service with the name of the subdomain that points to new app
- Regenerate certs and redeploy gateway

Also added a default server block that returns a 444 (no response) for invalid subdomains. Unfortunately this still presents to the user that the cert is invalid, since *.hail.is is registered in dns (I think?) and browsers will verify certificates before anything else, but users won't be able to click through and land at the website like they could before.


daniel-goldstein:
bump @danking

danking (approved):
This is so good. Nearly a 1:3 ratio of added to deleted lines.

PR #10251, opened 2021-03-30
[vcf-combiner] Add support for haploid GT calls

chrisvittal:


tpoterba (dismissed):
add a small test as we talked about?

chrisvittal:
I've validated correctness for haploid calls manually, and confirmed that results for diploid calls don't change.

tpoterba (approved):

PR #10252, opened 2021-03-30
[batch] shutdown cleanly when cancelled or idle

daniel-goldstein:


batch/batch/worker/worker.py
```diff
@@ -960,12 +963,15 @@ async def run(self, worker):
...
                     self.state = input.state
+            except asyncio.CancelledError:
+                raise
             except Exception as e:
                 if not user_error(e):
                     log.exception(f'while running {self}')
 
                 self.state = 'error'
                 self.error = traceback.format_exc()
+            # TODO Put into function call and call from except Exception and after try/except block
```
danking, line 974: let's do this for this PR

danking (changes requested):
I'm good with this, but inclined to wait for Friday morning.

danking (approved):

PR #10254, opened 2021-03-31
[query] Support older versions of Spark

johnc1231:
Support Spark 3.0.1 and 2.4.5 still by bundling and relocating json4s and jackson library that we need


tpoterba (approved):
Thanks for slogging through this!

PR #10260, opened 2021-04-01
[batch/copy] fix for sporadic copy failures (1/2)

cseed:
This is a temporary fix for the sporadic copy failures.  The problem is that this code cancels a task managed by the online bounded pool, and the pool treats that cancellation as an exception that it propagates up.

I need to think through the details of the bounded gather with respect to cancellation, and that's going to take a few days.  We could put this back when that's done, but honestly, it doesn't seem like an important optimization (given how rarely this failure comes up), so I'll probably just leave it out.


danking (approved):
awesome

PR #10264, opened 2021-04-01
[query] pca simplification/optimization

patrick-schultz:
This flips the pca algorithm, so that we're locally collecting a Krylov space basis on the column side (presumably the smaller side). This leads to a few simplifications, and allows for an optimization in the `compute_loadings=False` case. Once we have a complete TSQR implementation, we can optimize the `compute_loadings=True` case as well. With that, there should be no memory limitation to the number of rows (no local value is O(rows)).

### DIFF
--- a/hail/python/hail/methods/pca.py
+++ b/hail/python/hail/methods/pca.py
@@ -315,49 +315,44 @@ def _blanczos_pca(entry_expr, k=10, compute_loadings=False, q_iterations=2, over
     # Generate random matrix G
     G = hl.nd.zeros((n, L)).map(lambda n: hl.rand_norm(0, 1))
 
-    def hailBlanczos(A, G, k, q):
+    def hailBlanczos(A, G, k, q, compute_U=False):
 
-        h_list = []
         G_i = hl.nd.qr(G)[0]
+        g_list = [G_i]
 
         for j in range(0, q):
-            info(f"blanczos_pca: Beginning iteration {j + 1}/{q+1}")
-            temp = A.annotate(H_i=A.ndarray @ G_i)
-            temp = temp.annotate(G_i_intermediate=temp.ndarray.T @ temp.H_i)
-            result = temp.aggregate(hl.struct(Hi_chunks=hl.agg.collect(temp.H_i),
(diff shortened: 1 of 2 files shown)


patrick-schultz:
This PR broke `test_blanczos_against_hail`, but the difference between the two was only slightly above the threshold. This is because, for a fixed value of `q`, the new version effectively does half an iteration less than before. For example, for `q=2`, the old version computed a basis for `[AG, AA'AG, (AA')^2AG]`, while the new version computes a basis for `[G, A'AG, (A'A)^2G]`. Incrementing `q` by one in the test fixed it.

hail/python/hail/methods/pca.py
```diff
@@ -315,64 +315,59 @@ def _blanczos_pca(entry_expr, k=10, compute_loadings=False, q_iterations=2, over
...
-            localized_H_i = hl.nd.vstack(result.Hi_chunks)
-            h_list.append(localized_H_i)
-            G_i = hl.nd.qr(result.G_i)[0]
-
-        info(f"blanczos_pca: Beginning iteration {q+ 1}/{q+1}")
-        temp = A.annotate(H_i=A.ndarray @ G_i)
-        result = temp.aggregate(hl.agg.collect(temp.H_i), _localize=False)._persist()
+            info(f"blanczos_pca: Beginning iteration {j+1}/{q}")
+            G_i = A.aggregate(hl.agg.ndarray_sum(A.ndarray.T @ (A.ndarray @ G_i)), _localize=False)
+            G_i = hl.nd.qr(G_i)[0]._persist()
```
johnc1231, line 326: Why persist individual `G_i`s? Seems like you use them each exactly once.

johnc1231 (changes requested):

hail/python/hail/methods/pca.py
patrick-schultz: We save them in `g_list` and use them in the next iteration.

johnc1231: Duh, you're right. Approved.

johnc1231 (approved):

PR #10265, opened 2021-04-01
[query-service] move Worker to Worker.scala

danking:
@daniel-goldstein, you and I are both modifying this class. I want to relocate it for
ease of editing/discovery. I hope we can merge this before either of our PRs lands, so
the rebase is easier.


daniel-goldstein:
Can't imagine how many times already I've tried to fuzzy find `Worker.scala`

daniel-goldstein (approved):

danking:
Yeah. I think most folks use IntelliJ and search for classes rather than files.

PR #10276, opened 2021-04-05
[ci] allow dev deploy to forcibly exclude build steps

daniel-goldstein:
Potential footgun but specifically limited to dev deploys. This allows you to explicitly exclude build steps when dev deploying. For example, `hailctl dev deploy -b … -s deploy_query --excluded_steps deploy_batch` would make sure that `deploy_batch` and any jobs only in the `deploy_batch` subtree are not run.

Thoughts?

### DIFF
--- a/hail/python/hailtop/hailctl/dev/deploy/cli.py
+++ b/hail/python/hailtop/hailctl/dev/deploy/cli.py
@@ -6,6 +6,7 @@ import sys
 from hailtop.config import get_deploy_config
 from hailtop.auth import service_auth_headers
 from hailtop.httpx import client_session
+from hailtop.utils import unpack_comma_delimited_inputs
 
 
 def init_parser(parser):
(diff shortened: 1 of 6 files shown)


hail/python/hailtop/hailctl/dev/deploy/cli.py
```diff
@@ -60,8 +63,11 @@ async def submit(args):
     steps = [s.strip() for steps in args.steps  # let's unpack this shall we?
              for step in steps
              for s in step.split(',') if s.strip()]
+    excluded_steps = [s.strip() for steps in args.excluded_steps
```
jigold, line 66: Should we have a function in utils to do this?

jigold (dismissed):

jigold (dismissed):
See comment.

hail/python/hailtop/utils/utils.py
```diff
@@ -34,6 +34,13 @@
 }"""
 
 
+def unpack_comma_delimited_inputs(inputs):
+    return [s.strip()
+            for steps in inputs
```
jigold, line 39: I know this is a direct copy from what was there, but it sure seems complicated instead of this:

```python3
return [s.strip() for s in inputs.split(',') if s.strip()]
```

jigold (approved):

PR #10277, opened 2021-04-05
Fix GitHub icon in top bar of docs

nawatts:
Currently, the GitHub icon in the top bar of the Hail docs looks like this:

Screen Shot 2021-04-05 at 11 19 36 AM

I don't know how to build the full website and the navigation bar doesn't show up in a local build of the docs, but changing `fab` to `fa` in the Chrome element inspector appears to fix it.

Screen Shot 2021-04-05 at 11 28 08 AM

Not sure how this works on other pages of the site but not on the docs.

### DIFF
--- a/website/website/templates/nav-top.html
+++ b/website/website/templates/nav-top.html
@@ -30,7 +30,7 @@
         </li>
         <li class="nav-item" style='margin-top:-2px'>
           <a href="https://github.com/hail-is/hail" class='img-link' target='_blank'>
-            <i class="fab fa-github fa-2x"></i>
+            <i class="fa fa-github fa-2x"></i>
           </a>
         </li>
       </ul>


nawatts:
@danking (related to #10228)

johnc1231:
What browser are you using? It looks right to me currently

nawatts:
Chrome 89.

nawatts:
Looking at fonts in the network panel of Chrome dev tools:

On https://hail.is/, I see https://ka-f.fontawesome.com/releases/v5.15.3/webfonts/free-fa-brands-400.woff2, which looks like it matches the font family specified by the `fab` class ("Font Awesome 5 Brands").

On https://hail.is/docs/0.2/index.html however, the only Font Awesome font I see is https://hail.is/docs/0.2/_static/fonts/fontawesome-webfont.woff2?v=4.7.0.

johnc1231:
Ah, I missed the last sentence about it being right everywhere except the docs. I was looking on main page. Agree that this is a problem, will let Dan look

nawatts:
Seems like the current docs don't include the changes to `website/website/templates/dynamic-base.html` made in #10228. I don't see `<script src="https://kit.fontawesome.com/7cdc07e2ca.js" crossorigin="anonymous"></script>` in view-source:https://hail.is/docs/0.2/index.html.

danking:
Thanks for the report and fix @nawatts ! You should be able to test changes with `make -C website run` from the root of the repo.

Let me take a closer look at the docs.

danking:
Yeah, so, this is my bad. The dynamic-base is used at doc-generation-time. Changes to that are not reflected until the docs are regenerated (i.e. a new PyPI version of Hail is released). The short-term fix is to release Hail. The long-term fix is to not make changes to dynamic-base.

PR #10279, opened 2021-04-05

danking:
The high-level problem:

The high-level solution:

Hail Batch Workers can load the JAR for a given Hail version on-demand. A fresh class loader for
each Hail version allows the classes to co-exist in the same JVM. We cache jars on the local
filesystem.

---

`javac` compiles Java files to JVM Bytecode. JVM Bytecode is normally stored in `class` files. A JAR
file is, essentially, a TAR file of a directory of class files.

`java` needs to find the `class` file that defines any Class. A `ClassLoader` defines:

1. (`findClass`) How to *find* the definition of a Class known to the current `ClassLoader`.

2. (`findResource`) How to *find* an arbitrary file known to the current `ClassLoader`.

3. (`loadClass` and `getResource`) The order in which to find a class in a set of
  `ClassLoader`s (e.g. if two `ClassLoader`s know about the same Class, which one should load the
  class?)

Every `ClassLoader` has a `parent` `ClassLoader`. The default implementation of `loadClass` and
`getResource` prefers loading classes from its parent ClassLoader before anything else. We invert
the loading order to allow multiple definitions of the same Class in the same JVM. In particular,
each instance of `LoadSelfFirstURLClassLoader` prefers to use its own definition of a Class. Each
`LoadSelfFirstURLClassLoader` instance knows about one version of the Hail JAR.

The remaining subtle issue is how to load resources. For example, `HailBuildInfo` needs to load the
build info resource file. To do so, you need an instance of a `ClassLoader` that can find the
file you want. Often times, you use `this.getClass().getClassLoader()`, which is the class loader
used to load the current class. Hail does not do this. I believe we do not do this because of issues
with how TestNG loads classes. :sigh: As a result, I also modify the worker Thread's
ContextClassLoader for the duration of the execution of an alternative version of Hail.

---

---

Small changes and fixes:

  key.
  gs://hail-query`.
- `hail-ubuntu-stmp` was missing from `docker/Makefile`'s `clean` rule
- Use a dummy `WorkerBackend` when we're on the worker. The worker isn't allowed to call these
  methods anyway, that would amount to Hail Queries inside of Hail Queries, madness!
- New transient error in Java.

### DIFF
--- a/hail/python/hail/__init__.py
+++ b/hail/python/hail/__init__.py
@@ -1,12 +1,22 @@
+import pkg_resources
+import sys
+import asyncio
 import nest_asyncio
-nest_asyncio.apply()
-
-import pkg_resources  # noqa: E402
-import sys  # noqa: E402
 
 if sys.version_info < (3, 6):
     raise EnvironmentError('Hail requires Python 3.6 or later, found {}.{}'.format(
         sys.version_info.major, sys.version_info.minor))
+
+if sys.version_info[:2] == (3, 6):
+    if asyncio._get_running_loop() is not None:
+        nest_asyncio.apply()
+else:
+    try:
+        asyncio.get_running_loop()
+        nest_asyncio.apply()
+    except RuntimeError as err:
+        assert 'no running event loop' in err.args[0]
+
 
 __pip_version__ = pkg_resources.resource_string(__name__, 'hail_pip_version').decode().strip()
 del pkg_resources
--- a/hail/src/main/scala/is/hail/backend/service/Worker.scala
+++ b/hail/src/main/scala/is/hail/backend/service/Worker.scala
@@ -138,6 +138,7 @@ object Worker {
         val localJarFile = new File(localJarLocation)
         if (!localJarFile.exists()) {
           log.info(s"$revision not in the disk cache")
+          localJarFile.getParentFile().mkdirs()
           downloadFile(workerGCSFS, remoteJarLocation, localJarLocation)
           log.info(s"$revision added to the disk cache at $localJarLocation")
         }
--- a/hail/src/main/scala/is/hail/backend/service/Worker.scala
+++ b/hail/src/main/scala/is/hail/backend/service/Worker.scala
@@ -1,17 +1,41 @@
 package is.hail.backend.service
 
 import java.io._
-import java.nio.charset.Charset
+import java.net._
+import java.nio.charset.StandardCharsets
+import java.util.concurrent._
 
+
+
+import is.hail.HAIL_REVISION
 import is.hail.HailContext
-import is.hail.backend._
-import is.hail.io.fs._
+import is.hail.annotations._
+import is.hail.asm4s._
+import is.hail.backend.{Backend, BackendContext, BroadcastValue, HailTaskContext}
+import is.hail.expr.JSONAnnotationImpex
+import is.hail.expr.ir.lowering.{DArrayLowering, LoweringPipeline, TableStage, TableStageDependency}
+import is.hail.expr.ir.{Compile, ExecuteContext, IR, IRParser, Literal, MakeArray, MakeTuple, ShuffleRead, ShuffleWrite, SortField, ToStream}
+import is.hail.io.fs.GoogleStorageFS
+import is.hail.linalg.BlockMatrix
+import is.hail.rvd.RVDPartitioner
 import is.hail.services._
+import is.hail.services.batch_client.BatchClient
+import is.hail.services.shuffler.ShuffleClient
+import is.hail.types._
+import is.hail.types.encoded._
+import is.hail.types.physical._
+import is.hail.types.virtual._
 import is.hail.utils._
+import is.hail.variant.ReferenceGenome
 import org.apache.commons.io.IOUtils
 import org.apache.log4j.Logger
+import org.json4s.JsonAST._
+import org.json4s.jackson.JsonMethods
+import org.json4s.{DefaultFormats, Formats}
+import org.newsclub.net.unix.{AFUNIXServerSocket, AFUNIXSocketAddress}
 
-import scala.collection.mutable
+import scala.annotation.switch
+import scala.reflect.ClassTag
 
 class ServiceTaskContext(val partitionId: Int) extends HailTaskContext {
   override type BackendType = ServiceBackend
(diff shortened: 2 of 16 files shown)


infra/main.tf
```diff
@@ -672,6 +703,17 @@ resource "google_storage_bucket" "batch_logs" {
...
 
+resource "random_id" "hail_query_bucket_name_suffix" {
+  byte_length = 2
+}
+
+resource "google_storage_bucket" "hail_query" {
+  name = "hail-query-${random_id.hail_query_bucket_name_suffix.hex}"
+  location = var.hail_query_bucket_location
+  force_destroy = true
+  storage_class = var.hail_query_bucket_storage_class
```
lgruen, line 714: Thanks very much! I've applied these Terraform changes in our deployment.

Can we use uniform bucket permissions? This seems to be the current GCP recommendation.

```suggestion
  storage_class = var.batch_logs_bucket_storage_class
  uniform_bucket_level_access = true
}

resource "random_id" "hail_query_bucket_name_suffix" {
  byte_length = 2
}

resource "google_storage_bucket" "hail_query" {
  name = "hail-query-${random_id.hail_query_bucket_name_suffix.hex}"
  location = var.hail_query_bucket_location
  force_destroy = true
  storage_class = var.hail_query_bucket_storage_class
  uniform_bucket_level_access = true
```

danking: Certainly!

hail/src/main/scala/is/hail/backend/service/Worker.scala
```diff
@@ -44,21 +69,114 @@ class WorkerTimer() {
...
     val timer = new WorkerTimer()
+    timer.start("main")
 
-    var scratchDir = System.getenv("HAIL_WORKER_SCRATCH_DIR")
-    if (scratchDir == null)
-      scratchDir = ""
+    log.info(s"is.hail.backend.service.Worker $myRevision")
+    if (args.length < 2) {
+      throw new IllegalArgumentException(s"expected at least two arguments, not: ${ args.length }")
+    }
```
daniel-goldstein, line 161: I feel like you might as well check for 4 arguments here instead of checking once for 2 and then for 4.

hail/src/main/scala/is/hail/backend/service/Worker.scala
```diff
@@ -44,21 +69,114 @@ class WorkerTimer() {
...
+  }
+
+  private[this] val STRING_ARRAY_CLASS = Class.forName("[Ljava.lang.String;")
+  private[this] def loadMainFromLocalJar(revision: String, localJarURL: URL): Array[String] => Unit = {
+    val classLoader = new LoadSelfFirstURLClassLoader(Array(localJarURL))
+    val reflectedMain = withThreadContextClassLoader(classLoader) {
+      val workerClass = classLoader.loadClass("is.hail.backend.service.Worker")
+
+      val actualRevision = workerClass.getMethod("myRevision").invoke(null).asInstanceOf[String]
```
daniel-goldstein, line 102: Should this not be `getValue`? Or are all values thunks? I don't know how Scala reflection works.

hail/src/main/scala/is/hail/backend/service/Worker.scala
```diff
@@ -44,21 +69,114 @@ class WorkerTimer() {
...
+    log.info(s"is.hail.backend.service.Worker $myRevision")
+    if (args.length < 2) {
+      throw new IllegalArgumentException(s"expected at least two arguments, not: ${ args.length }")
+    }
 
-    log.info(s"running job $i at root $root wih scratch directory '$scratchDir'")
+    val revision = args(0)
+    val remoteJarLocation = args(1)
+    if (revision != myRevision) {
+      timer.start("classLoading")
```
daniel-goldstein, line 166: Perhaps change to Worker.main class loading? As far as I understand there still might be other classes loaded at runtime of `mainMethod(args)`?

hail/src/main/scala/is/hail/backend/service/Worker.scala
```diff
@@ -44,21 +69,114 @@ class WorkerTimer() {
...
+        revisions.put(revision, mainMethod)
+        mainMethod
     }
-    val root = args(0)
-    val i = args(1).toInt
+  }
+
+  def main(args: Array[String]): Unit = {
     val timer = new WorkerTimer()
+    timer.start("main")
```
daniel-goldstein, line 156: Will this start two `main` timings when an alternate jar is loaded? I think this span is a bit confusing

daniel-goldstein (dismissed):
For the most part looks good, albeit a bit mind-bendy. Just some small comments.

hail/src/main/scala/is/hail/backend/service/Worker.scala
danking: I considered this, but I want to allow us to change the number of arguments without breaking backwards compatibility.

hail/src/main/scala/is/hail/backend/service/Worker.scala
danking: So, this is probably a bit not kosher. I'm using Java reflection, not Scala reflection. I spent maybe an hour fiddling with Scala reflection before I gave up. Scala compiles non `private[this]` fields as nullary methods, so I call that method. I use `null` as the `this` object because I know this is a static method (it's a val of an object).

I could try again with Scala reflection, that'd probably be more safe.

hail/src/main/scala/is/hail/backend/service/Worker.scala
danking: Good point, I changed to `"JAR fetching and Worker class loading"`

hail/src/main/scala/is/hail/backend/service/Worker.scala
danking: Yeah this is a bit confusing. Suppose we have the Worker who is first called `A` and the worker we reflect into `B`. `A`'s timer is basically abandoned. I never call end, so I never print out any information about that timing. `B`'s main method creates a new timer that knows nothing about the other timer (in fact, it probably can't refer to anything in `A` since it believes it is the one, true is.hail.backend.service.Worker).

I suppose a fix would be to break main like this:

```
  def main(args: Array[String]): Unit = {
    val timer = new WorkerTimer()
    timer.start("main")

    log.info(s"is.hail.backend.service.Worker $myRevision")
    if (args.length < 2) {
      throw new IllegalArgumentException(s"expected at least two arguments, not: ${ args.length }")
    }

    val revision = args(0)
    val remoteJarLocation = args(1)
    if (revision != myRevision) {
      timer.start("JAR fetching and Worker class loading")
      log.info(s"received job for different revision: $revision; I am $myRevision")
      val runMethod = runForRevision(revision, remoteJarLocation)
      timer.end("JAR fetching and Worker class loading")
      runMethod(args)
      timer.end("main")
    } else {
      run(args)
      timer.end("main")
    }
  }

  def run(args: Array[String]): Unit = {
    /* only Hail execution stuff, no class loading */
  }
```

hail/python/hail/__init__.py
```diff
@@ -1,13 +1,23 @@
...
 
+if sys.version_info[:2] == (3, 6):
+    if asyncio._get_running_loop() is not None:
+        nest_asyncio.apply()
+else:
+    try:
+        asyncio.get_running_loop()
+        nest_asyncio.apply()
+    except RuntimeError as err:
+        assert 'no running event loop' in err.args[0]
```
danking, line 18: I had to add this nonsense to avoid some pytest errors I saw in the service tests.

I think pytest imports python modules from a secondary thread but executes them from the main thread. If `nest_asyncio` is called when a loop isn't running, it tries to create a loop. Unfortunately, `asyncio` forbids loops outside the main thread. Therefore, we get an error. Luckily, pytest does not need nest_asyncio because it doesn't start an asyncio event loop.

Regarding the Python 3.6 underscore method: 3.6 doesn't have an officially supported way to do this. That's why they added `get_running_loop` in 3.7.

cc: @tpoterba

hail/src/main/scala/is/hail/backend/service/Worker.scala
daniel-goldstein: I think this is _a lot_ easier to follow

daniel-goldstein (changes requested):
Just some residual cleanup from the main -> run change.

danking:
gotta fix some things

PR #10280, opened 2021-04-06
[ci] Fix excluded steps

jigold:


daniel-goldstein:
Is this broken for you on the newest version of main? I encountered this and found that my `hailctl` was old and wasn't pushing an empty list for the excluded steps. Not sure if that's the best design instead of accepting optional arguments.

daniel-goldstein:
I guess that makes this change backwards incompatible, so for that reason we can make this optional

jigold:
bump?

daniel-goldstein:
Closing because this is due to not picking up changes in local hailctl.

PR #10283, opened 2021-04-06
[batch] Expose job private and nonpreemptible machines

jigold:
Stacked on #10097

- Expose `machine_type` and `preemptible` in hailtop.batch
- Sets default storage when specifying the machine type to 100Gi
- Selects the cheapest machine type automatically for the user when they want a nonpreemptible worker

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -83,8 +83,14 @@ routes = web.RouteTableDef()
 deploy_config = get_deploy_config()
 
 BATCH_JOB_DEFAULT_CPU = os.environ.get('HAIL_BATCH_JOB_DEFAULT_CPU', '1')
+<<<<<<< HEAD
 BATCH_JOB_DEFAULT_MEMORY = os.environ.get('HAIL_BATCH_JOB_DEFAULT_MEMORY', 'standard')
 BATCH_JOB_DEFAULT_STORAGE = os.environ.get('HAIL_BATCH_JOB_DEFAULT_STORAGE', '0Gi')
+=======
+BATCH_JOB_DEFAULT_MEMORY = os.environ.get('HAIL_BATCH_JOB_DEFAULT_MEMORY', '3.75Gi')
+BATCH_JOB_POOL_DEFAULT_STORAGE = os.environ.get('HAIL_BATCH_JOB_POOL_DEFAULT_STORAGE', '0Gi')
+BATCH_JOB_PRIVATE_DEFAULT_STORAGE = os.environ.get('HAIL_BATCH_JOB_PRIVATE_DEFAULT_STORAGE', '100Gi')
+>>>>>>> [batch] Expose job private and nonpreemptible machines
 BATCH_JOB_DEFAULT_PREEMPTIBLE = True
 
 
@@ -719,7 +725,10 @@ WHERE user = %s AND id = %s AND NOT deleted;
                     req_memory_bytes = None
 
                 if 'storage' not in resources:
-                    resources['storage'] = BATCH_JOB_DEFAULT_STORAGE
+                    if machine_type is None:
+                        resources['storage'] = BATCH_JOB_POOL_DEFAULT_STORAGE
+                    else:
+                        resources['storage'] = BATCH_JOB_PRIVATE_DEFAULT_STORAGE
                 resources['req_storage'] = resources['storage']
                 del resources['storage']
                 req_storage_bytes = parse_storage_in_bytes(resources['req_storage'])
@@ -733,12 +742,9 @@ WHERE user = %s AND id = %s AND NOT deleted;
 
                 inst_coll_configs: InstanceCollectionConfigs = app['inst_coll_configs']
 
-                result, exc = inst_coll_configs.select_inst_coll(
+                result = inst_coll_configs.select_inst_coll(
                     machine_type, preemptible, worker_type, req_cores_mcpu, req_memory_bytes, req_storage_bytes
                 )
-
-                if exc:
-                    raise web.HTTPBadRequest(reason=exc.message)
 
                 if result is None:
                     raise web.HTTPBadRequest(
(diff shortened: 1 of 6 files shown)


batch/batch/inst_coll_config.py
```diff
@@ -108,21 +136,78 @@ async def async_init(self):
...
+                    worker_type=worker_type,
+                    cores_mcpu=req_cores_mcpu,
+                    memory_bytes=req_memory_bytes,
+                    storage_bytes=req_storage_bytes)
+            else:
+                result = self.select_pool_from_cost(
+                    cores_mcpu=req_cores_mcpu,
+                    memory_bytes=req_memory_bytes,
+                    storage_bytes=req_storage_bytes)
+            if result is not None:
```
jigold, line 195: The implication of this line not being `return result` only is that jobs that require more than 16 cores will have private instances spun up for them. Do we want this behavior????

danking:
I'll review as soon as https://github.com/hail-is/hail/pull/10097 lands

danking (dismissed):
Dismiss when https://github.com/hail-is/hail/pull/10097 is done

batch/batch/front_end/front_end.py
```diff
@@ -733,13 +742,10 @@ async def create_jobs(request, userdata):
 
                 inst_coll_configs: InstanceCollectionConfigs = app['inst_coll_configs']
 
-                result, exc = inst_coll_configs.select_inst_coll(
+                result = inst_coll_configs.select_inst_coll(
```
danking, line 745: So, I stared at this code for a bit and then at the InstanceCollectionManager and I found it a bit hard to follow. Some concrete criticisms:
- I dislike repeated conditionals, e.g. we branch on `machine_type is None` twice.
- There's too many different, actively-used variables and fields for a single concept. Consider `cpu`:  the field `resources['cpu']`, the variable `req_cores_mcpu`, and the field `resources['req_cpu']`. Memory also has many forms, but it doesn't parallel all of them, for example whereas cpu does not have a variable `req_cpu`, memory does have `req_memory`.
- Inconsistent or non-analogue use of helper functions. For example, storage and cpu have "parse_X_in_UNITS" functions but for memory we have an `if` and two different functions.
- We don't always check for `None`, for example `parse_memory_in_bytes` can return `None`.
- Relatedly, we should add type annotations to protect us from the aforementioned error.
- I found it really hard to follow `select_inst_coll`. I think this is because it takes many arguments and it does not clearly distinguish between the two case: pool and private. I think that confusion comes from `front_end.py` which also does not make a clear distinction between pool and private. For me, the repeated conditionals were a symptom that we were not clearly distinguishing between the cases.
- I find implicit representations of distinct cases confusing. For example, the user specifies one of four cases for memory: `highmem`, `lowmem`, `standard`, or an explicit amount. We blur that into an explicit amount, an optionally requested number of cores, and an optional machine type. This creates two problems:
  - front_end.py is now both doing simple string parsing *and* machine selection (so as to convert a memory ratio to an explicit memory request)
  - InstanceCollectionConfigs has to recover this distinction from the memory amount, cores, and machine type. I believe this is how you ended up with `from_cost` and `from_machine_type`, which I found confusing.

In typical Dan-fashion, in order to understand what I found confusing, I had to manipulate the code myself. This is what I produced: https://github.com/jigold/hail/compare/expose-private-worker...danking:expose-private-worker. What do you think of incorporating those changes?

batch/batch/utils.py
```diff
@@ -87,6 +82,26 @@ def cost_from_msec_mcpu(msec_mcpu):
...
+            cost_per_core_hour = 0.0125
+    else:
+        if worker_type == 'standard':
+            cost_per_core_hour = 0.04749975
+        elif worker_type == 'highcpu':
+            cost_per_core_hour = 0.0354243
+        else:
+            assert worker_type == 'highmem'
+            cost_per_core_hour = 0.0591515
+    return cost_per_core_hour
```
danking, line 102: For functions like this, I find it less confusing when we don't use elif and else and just directly return when we find the right condition. That avoids the unnecessary variable and keeps all the conditions aligned (nb: `if ` is three characters but `elif ` and `else ` are 5).

batch/batch/utils.py
```diff
@@ -194,6 +209,36 @@ def is_valid_cores_mcpu(cores_mcpu: int):
     return quarter_cores & (quarter_cores - 1) == 0
 
 
+def resources_to_machine_type(cores_mcpu, memory_bytes, preemptible):
```
danking, line 212: I don't think we should automatically convert pool jobs into private jobs, so I think this function is unnecessary.

hail/python/hailtop/batch/docs/service.rst
```diff
@@ -175,6 +175,13 @@ is standard. If a job is scheduled on this machine, then the cost per core hour
     total cost for that job will be 7 minutes.
 
 
+Batch also supports private instances that are created for a single job that specify
+:meth:`.Job.machine_type`, set :meth:`.Job.preemptible` to False, or request 32 or
+64 cores based on the combined cpu and memory requests. A job is billed for the entire
+time the machine is running including the activation time for the instance. The costs
+are the same as above except the cost for the Local SSD is omitted.
```
danking, line 182: All our pools are preemptible. If I read this correctly, private instances must be non-preemptible. Why provide two configurable variables? It seems like `preemptible` is unnecessary and `machine_type` fully describes the request. Perhaps machine_type should be instead `def private_machine(machine_type: int)`?

danking (changes requested):

jigold:
I think this code needed to be more complicated than you desired because you're not handling the case where someone can request something like this:

```python3
j = b.new_job()
j.memory('standard')
j.cpu(1)
j.preemptible(False)
```

Your solution would require that any non preemptible job would have to be like this:

```python3
j = b.new_job()
j.machine_type('n1-standard-1', preemptible=False)
```

I was trying to make things easier for the user, especially once we live in a world where we have amazon, azure, etc. But maybe your interface is fine and we drop the preemptible method and put it on machine_type only. In our design discussions, we had decided that we'd support the former interface and select the correct / cheapest machine type for the user automatically.

I think we should have another interface discussion, especially since we might be supporting other clouds, before we finalize this PR.

jigold:
This is going to have to change drastically to support Azure.

PR #10287, opened 2021-04-07
[dataproc] Add start/stop/start cluster test

tpoterba:


hail/python/cluster-tests/cluster-start-stop.py
```diff
@@ -0,0 +1,10 @@
+import hail as hl
+
+# turn it on
+hl.init()
+
+# turn it off
+hl.stop()
+
+# turn ir back on
```
johnc1231, line 9: Silly, but `it`, not `ir`.

tpoterba: oops, fixed

johnc1231 (approved):

PR #10291, opened 2021-04-07
 [datasets] add GTEx eQTL tissue specific tables

pwc2:
Adds the 49 [GTEx](https://gtexportal.org/home/datasets) eQTL tissue-specific all SNP gene association datasets as Hail Tables.

Sourced from `gs://gtex-resources/GTEx_Analysis_v8_QTLs/GTEx_Analysis_v8_eQTL_all_associations`. 

Files were .gz so had to be rewritten as .bgz before creating the Hail Tables.


hail/python/hail/experimental/db.py
```diff
@@ -312,7 +312,7 @@ class DB:
     >>> db = hl.experimental.DB(region='us', cloud='gcp')
     """
 
-    _valid_key_properties = {'gene', 'unique'}
+    _valid_key_properties = {'gene', 'unique', 'non_unique'}
```
danking, line 315: I think we shouldn't do this. If `'unique'` is not in the properties, that implies non_unique. If `'unique'` is missing from the key properties, we annotate a list of matching results.

danking: Put another way: what does it mean if neither unique or non_unique are in the valid key properties.

danking (changes requested):

hail/python/hail/experimental/db.py
pwc2: Gotcha, yeah I just wasn't sure how explicit to be in the json with non-unique keys. Should be good now.

danking (approved):

PR #10294, opened 2021-04-08
[batch,query,memory] add signal handler to dump task stacktraces

daniel-goldstein:
Right now the query service is taking a long time to terminate and it seems it's waiting on a single task. We saw a similar thing happen with a zombie worker but found it difficult to attach a debugger into the process. This way if we ssh into the pod and send a `SIGUSR1` we can hopefully get some good information out.

### DIFF
--- a/hail/python/hailtop/utils/__init__.py
+++ b/hail/python/hailtop/utils/__init__.py
@@ -9,8 +9,8 @@ from .utils import (
     retry_response_returning_functions, first_extant_file, secret_alnum_string,
     flatten, partition, cost_str, external_requests_client_session, url_basename,
     url_join, is_google_registry_image, url_scheme, Notice, periodically_call,
-    dump_all_stacktraces,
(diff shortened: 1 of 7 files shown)


jigold (approved):

hail/python/hailtop/utils/__init__.py
```diff
@@ -9,6 +9,7 @@
     retry_response_returning_functions, first_extant_file, secret_alnum_string,
     flatten, partition, cost_str, external_requests_client_session, url_basename,
     url_join, is_google_registry_image, url_scheme, Notice, periodically_call,
+    dump_all_stacktraces,
```
jigold, line 12: Should you add the `dump_all_stacktraces` in `__all__`?

jigold (changes requested):

hail/python/hailtop/utils/__init__.py
daniel-goldstein: Whoops, good catch

jigold (approved):

PR #10295, opened 2021-04-08
[ci] add link back to PR in batch view

daniel-goldstein:
Batches that have a `pr` attribute now link back to the corresponding PR page. Also fixed a latent bug where we were only loading batches for a PR based on the PR number, so if we had multiple watched branches you could display batches for PRs of the same number across branches.


danking (approved):
awesome, thanks!

PR #10299, opened 2021-04-09
[query] Fix memory management in new stream emitter

tpoterba:
- add 1st round of tests for stream memory usage
- clarify streamfilter implementation (no bug here though)

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
@@ -1,6 +1,6 @@
 package is.hail.expr.ir.streams
 
-import is.hail.annotations.Region
+import is.hail.annotations.{Region, RegionPool, RegionUtils}
 import is.hail.asm4s._
 import is.hail.expr.ir._
 import is.hail.expr.ir.orderings.StructOrdering
@@ -154,7 +154,7 @@ object EmitStream {
 
     def emit(ir: IR, cb: EmitCodeBuilder, region: Value[Region] = outerRegion, env: Emit.E = env, container: Option[AggContainer] = container): IEmitCode = {
       ir.typ match {
-        case _: TStream => produce(ir, cb, outerRegion, env, container)
+        case _: TStream => produce(ir, cb, region, env, container)
         case _ => emitter.emitI(ir, cb, region, env, container, None)
       }
     }
@@ -782,6 +782,7 @@ object EmitStream {
             }
           }
 
+          val resultElementRegion = mb.genFieldThisRef[Region]("flatmap_result_region")
           // grabbing emitcode.pv weird pattern but should be safe
           val SStreamCode(_, innerProducer) = innerStreamEmitCode.pv
 
@@ -790,6 +791,7 @@ object EmitStream {
 
             override def initialize(cb: EmitCodeBuilder): Unit = {
               cb.assign(first, true)
+              cb.assign(innerUnclosed, false)
 
               if (outerProducer.requiresMemoryManagementPerElement)
                 cb.assign(outerProducer.elementRegion, Region.stagedCreate(Region.REGULAR, cb.emb.ecb.pool()))
@@ -799,7 +801,7 @@ object EmitStream {
               outerProducer.initialize(cb)
             }
 
-            override val elementRegion: Settable[Region] = innerProducer.elementRegion
+            override val elementRegion: Settable[Region] = resultElementRegion
 
             override val requiresMemoryManagementPerElement: Boolean = innerProducer.requiresMemoryManagementPerElement || outerProducer.requiresMemoryManagementPerElement
 
@@ -819,6 +821,9 @@ object EmitStream {
                 cb.ifx(innerUnclosed, {
                   cb.assign(innerUnclosed, false)
                   innerProducer.close(cb)
+                  if (innerProducer.requiresMemoryManagementPerElement) {
+                    cb += innerProducer.elementRegion.invalidate()
+                  }
                 })
 
                 cb.goto(outerProducer.LproduceElement)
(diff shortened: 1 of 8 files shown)


hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -1269,15 +1267,6 @@ object EmitStream {
             override def close(cb: EmitCodeBuilder): Unit = {}
           }
 
-          val innerStreamCode = EmitCode.fromI(mb) { cb =>
-            if (childProducer.requiresMemoryManagementPerElement)
```
tpoterba, line 1273: this code shouldn't have been here. It's on the stream consumer to set these regions -- these were being leaked!

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -468,9 +468,12 @@ object EmitStream {
 
               override val LproduceElement: CodeLabel = mb.defineAndImplementLabel { cb =>
                 val Lfiltered = CodeLabel()
+                cb.define(Lfiltered)
```
tpoterba, line 471: no bug here, just a clearer implementation

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -154,7 +153,7 @@ object EmitStream {
 
     def emit(ir: IR, cb: EmitCodeBuilder, region: Value[Region] = outerRegion, env: Emit.E = env, container: Option[AggContainer] = container): IEmitCode = {
       ir.typ match {
-        case _: TStream => produce(ir, cb, outerRegion, env, container)
```
tpoterba, line 157: this was a big one - it masked most of the rest of the bugs because nested stream elements were allocated in the partition region.

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -915,13 +939,17 @@ object EmitStream {
...
+                    else
+                      cb.assign(leftProducer.elementRegion, outerRegion)
 
                     leftProducer.initialize(cb)
                     rightProducer.initialize(cb)
                   }
 
-                  override val elementRegion: Settable[Region] = leftProducer.elementRegion
-                  override val requiresMemoryManagementPerElement: Boolean = leftProducer.requiresMemoryManagementPerElement
+                  override val elementRegion: Settable[Region] = _elementRegion
```
tpoterba, line 951: the old implementation could generate region reference cycles if the left stream didn't require memory management, and the right did.

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -481,14 +484,8 @@ object EmitStream {
                     })
 
                 if (requiresMemoryManagementPerElement)
-                  cb += childProducer.elementRegion.addReferenceTo(filterEltRegion)
-
+                  cb += filterEltRegion.trackAndIncrementReferenceCountOf(childProducer.elementRegion)
```
tpoterba, line 487: all of these were flipped.

patrick-schultz (changes requested):
Oof, good catches. The new tests are very nice.

I have a few comments, but none are critical. If you want to push this through quickly, that's okay with me.

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
tpoterba: it gets cleared every time we jump to LProduceElement. But after some time away, I think you're right - this is worse.

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -833,6 +836,11 @@ object EmitStream {
                     _ =>
                       // the inner stream/producer is bound to a variable above
                       cb.assign(innerUnclosed, true)
+                      if (innerProducer.requiresMemoryManagementPerElement)
```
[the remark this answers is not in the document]
tpoterba: oh, I think you're right! I'll make that as a follow-up PR if that's OK.

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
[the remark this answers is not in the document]
tpoterba: I added `takeOwnershipOfAndClear` on RichCodeRegion. I'll push that into the rest of the places that call `trackAndIncrement / clear` as a followup.

hail/src/main/scala/is/hail/utils/richUtils/RichCodeRegion.scala
```diff
@@ -14,6 +14,8 @@ class RichCodeRegion(val region: Code[Region]) extends AnyVal {
   def trackAndIncrementReferenceCountOf(other: Code[Region]): Code[Unit] =
     region.invoke[Region, Unit]("addReferenceTo", other)
 
+  def takeOwnershipOfAndClear(other: Code[Region]): Code[Unit] = region.invoke[Region, Unit]("move", other)
```
patrick-schultz, line 17: I think this is backwards

patrick-schultz (changes requested):

hail/src/main/scala/is/hail/utils/richUtils/RichCodeRegion.scala
patrick-schultz: `Region.move` is "move to"

tpoterba: ugh, thanks. I said it correctly in my head when I copy/pasted the line above, and then forgot to change it 🤦

patrick-schultz (approved):

PR #10301, opened 2021-04-12
[query] Simplify HailTaskContext logic

tpoterba:
Part 1 of exposing FS correctly in service backend.


danking (approved):
Seems right to me.

tpoterba:
back to you -- revised version where HailTaskContext.get is now SparkTaskContext.get -- nobody else uses the threadlocal value.

danking (approved):
Ah this is nicer

PR #10304, opened 2021-04-13
Java NDarray Safety

johnc1231:


hail/src/main/scala/is/hail/annotations/UnsafeRow.scala
```diff
@@ -244,13 +244,16 @@ object SafeRow {
...
+      case _: UnsafeNDArray => false
 
       case r: Row =>
         r.toSeq.forall(isSafe)
       case a: IndexedSeq[_] =>
         a.forall(isSafe)
       case i: Interval =>
         isSafe(i.start) && isSafe(i.end)
+      case nd: NDArray =>
+        isSafe(nd.getRowMajorElements().forall(isSafe))
```
tpoterba, line 256: seems wrong -- should check type of top level and then recur to elements?

tpoterba (dismissed):

hail/src/main/scala/is/hail/annotations/UnsafeRow.scala
johnc1231: Top level is checked above right?

tpoterba: oh! I see.

Should be         

`nd.getRowMajorElements().forall(isSafe))`

though, right? (still probably works as written, but does an extra isSafe on the Boolean returned by forall)

johnc1231:
Oh yes duh, thanks. Will fix

tpoterba (approved):

PR #10308, opened 2021-04-13
[memory] use memory as logging name, not batch

danking:


daniel-goldstein:
I included this in my memory PR

PR #10309, opened 2021-04-13
[query] synchronize the code cache

danking:
Avoid query-service races when looking at the cache.


tpoterba:
With the FS changes, we don't serialize any secrets into generated code, right?

tpoterba (approved):

danking:
I think that's right, though we serialize other potentially private information. I think we ought to have a per-organization (Hail billing project?) cache, but also not very high priority. I'd be pretty chuffed to learn we're running important enough stuff that people are attempting timing attacks on our cache to learn what queries other people are executing.

PR #10312, opened 2021-04-13
[hailtop] use the exact same error message for sync and async

danking:
Importantly, I want to see the stack trace in either case.

### DIFF
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -710,6 +710,12 @@ async def collect_agen(agen):
     return [x async for x in agen]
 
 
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/utils/utils.py
```diff
@@ -709,12 +710,6 @@ async def collect_agen(agen):
     return [x async for x in agen]
 
 
-def dump_all_stacktraces():
```
jigold, line 712: Isn't this still used somewhere? Or did Daniel move it?

jigold (dismissed):

jigold (approved):

PR #10314, opened 2021-04-13

danking:
The high-level problem: 

The high-level solution:

Hail Batch Workers can load the JAR for a given Hail version on-demand. Although not a long-term
solution, we currently start a fresh JVM for each job. As a result, we can simply start the JVM with
the correct JAR on its classpath. We cache jars on the local filesystem.

I had to abandon the old approach for two reasons:

1. Multiple JVMs race to download the JAR. In the new approach, the python worker process uses a
   lock to ensure at most one coroutine is downloading a given version of a JAR at the same time.

2. The JVM includes assumes that a child ClassLoader does not redefine a class from the parent
   ClassLoader. That's why ClassLoaders always prefer to load a class from the parent ClassLoader's
   classes.

When we decide to re-use JVMs or use a single multi-threaded JVM, we'll need to ensure the top-level
ClassLoader *does not have Hail on its classpath*. I looked briefly at this approach and found it
more work than the current approach.

---

My apologies for eliminating JVMProcess in this PR. It's an unrelated change which facilitated my
understanding of worker.py. I essentially inlined JVMProcess into JVMJob and eliminated any duplicative
code.

---

After making this change I restored the tests. Some tests had bitrotted. In the process of fixing
those tests, I found a few other bugs. Fixing these lower-level bugs unlocked a number of new
tests.

One test (which was added since the service tests were removed) had to be marked as failing. Some
Hail operations rely on writing to the local file system. Implementing that properly in the Query
Worker will take some thought.

Here are the bugs I fixed:

1. Correct the error message raised when tests are run in a non-main thread (we look for this
   message and start an event loop for Hail's async code because asyncio refuses to start an event
   loop in a non-main thread).

2. Use a `SafeRow` to copy the globals data out of a Region and into durable, GC'ed objects.

3. Re-enable serialization of GoogleStorageFS (including its private key, which we really shouldn't
   do; Tim is working on it), which was broken (presumably) when we changed Scala versions. The
   `var` modifier ensures the name is compiled as a JVM field.

4. Correctly convert from a `Byte` to an `Int`. By default `Byte` to `Int` conversion (which is done
   automatically when you return a `Byte` from a function whose return type is `Int`) is
   sign-preserving. That means that the byte `0000 1111` is converted to the `Int` 15 and the byte
   `1000 1111` is converted to the `Int` -113. The contract of
   [`InputStream.read`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read--)
   is to return the unsigned integeral value of the next `Byte` or `-1` if we've reached the end of
   the stream. `DataInputStream` treats any negative value as EOS which lead to perplexing EOSes
   when reading data from GCS.

5. Retain the `gs://` protocol when reading MTs and Ts. `uriPath` strips *all* protocols. Before the
   Query Service, these code paths were only used by the LocalBackend. In the LocalBackend, the only
   URIs generated are `file://`. However, UNIX/JVM file system operations do not support URIs, they
   want bare paths.

6. Implement missing cases for EShuffle and PShuffle.

---

---

Small changes and fixes:

  key.

  gs://hail-query`.

- `hail-ubuntu-stmp` was missing from `docker/Makefile`'s `clean` rule

- Only apply nest_asyncio when an event loop is already started.

### DIFF
--- a/batch/Dockerfile.worker
+++ b/batch/Dockerfile.worker
@@ -46,5 +46,4 @@ COPY batch/setup.py batch/MANIFEST.in /batch/
 COPY batch/batch /batch/batch/
 RUN hail-pip-install --no-deps /batch && rm -rf /batch
 
-COPY batch/hail.jar /
 COPY query/log4j.properties /
--- a/batch/Makefile
+++ b/batch/Makefile
@@ -28,9 +28,6 @@ build-batch: build-prereqs
 
 .PHONY: build-worker
 build-worker: build-prereqs
-	$(MAKE) -C ../hail shadowJar
-# janky
-	cp ../hail/build/libs/hail-all-spark.jar ./hail.jar
 	-docker pull $(BATCH_WORKER_LATEST)
 	python3 ../ci/jinja2_render.py '{"global":{"project":"$(PROJECT)"}, "service_base_image":{"image":"service-base"}}' Dockerfile.worker Dockerfile.worker.out
 	docker build -t batch-worker -f Dockerfile.worker.out --cache-from batch-worker,$(BATCH_WORKER_LATEST),service-base ..
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, Dict
 import os
 import json
 import sys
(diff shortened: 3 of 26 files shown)


danking:
I want to add some tests first.

PR #10315, opened 2021-04-13
[memory,query] Cache job function/context/outputs in memory service

daniel-goldstein:
Reading function/contexts from GCS on query workers can contribute a significant portion of the runtime for small jobs. For a simple query like `hl.utils.range_table(10).collect()`, the jobs in the batch can range in time from 5-9 seconds depending on GCS latency.

This builds on #9484 to add write-through capability to `memory` and a `ServiceCacheableFS` in Scala. The cacheable FS reads/writes through `memory` and falls back to GCS, so in the good path the ServiceBackend writes the compiled function and contexts to `memory`, workers read inputs and write outputs exclusively from/to memory, from which the ServiceBackend reads the results. From small benchmarks in dev, this cuts down read times on the workers by ~30-40% compared to the worst case GCS latencies and roughly matches the current implementation in the best case. Writing the outputs is comparable to writing through an already warmed up GCS connection.

### DIFF
--- a/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
@@ -66,10 +66,11 @@ class ServiceBackend() extends Backend {
     assert(previous == null)
   }
 
+  // TODO Do we need the sessionID to be serialized at all?
   def userContext[T](username: String, sessionID: String, timer: ExecutionTimer)(f: (ExecuteContext) => T): T = {
     val user = users.get(username)
     assert(user != null, username)
-    ExecuteContext.scoped(user.tmpdir, "file:///tmp", this, user.fs.asCacheable(sessionID), timer, null)(f)
+    ExecuteContext.scoped(user.tmpdir, "file:///tmp", this, user.fs, timer, null)(f)
   }
 
   def defaultParallelism: Int = 10
--- a/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
@@ -66,8 +66,7 @@ class ServiceBackend() extends Backend {
     assert(previous == null)
   }
 
-  // TODO Do we need the sessionID to be serialized at all?
-  def userContext[T](username: String, sessionID: String, timer: ExecutionTimer)(f: (ExecuteContext) => T): T = {
+  def userContext[T](username: String, timer: ExecutionTimer)(f: (ExecuteContext) => T): T = {
     val user = users.get(username)
     assert(user != null, username)
     ExecuteContext.scoped(user.tmpdir, "file:///tmp", this, user.fs, timer, null)(f)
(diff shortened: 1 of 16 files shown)


hail/src/main/scala/is/hail/io/fs/GoogleStorageFS.scala
```diff
@@ -5,15 +5,19 @@ import java.net.URI
...
 import com.google.cloud.{ReadChannel, WriteChannel}
 import com.google.cloud.storage.Storage.BlobListOption
 import com.google.cloud.storage.{Blob, BlobId, BlobInfo, Storage, StorageOptions}
+import is.hail.HailContext
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable
 
 object GoogleStorageFS {
+  val log = Logger.getLogger(getClass.getName())
```
danking, line 19: if you make this `private` its still accessible in GoogleStorageFS's instances but seems tidier.

hail/src/main/scala/is/hail/services/memory_client/MemoryClient.scala
```diff
@@ -0,0 +1,104 @@
...
+  private[this] val req = new HttpGet(uri.build())
+
+  private[this] val _bytes = requester.requestAsByteStream(req)
+  private[this] var _pos: Int = 0
+  private[this] val _len: Int = _bytes.length
+
+  override def read(): Int = {
+    if (_pos >= _len) -1 else {
+      _pos += 1
+      _bytes(_pos.toInt - 1)
```
danking, line 62: _pos is already an Int

danking: explicitly convert to an int and mask to the last 8 bits: `_bytes(..).toInt & 0xff`

danking (changes requested):
small things

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -68,10 +67,10 @@ class ServiceBackend() extends Backend {
     assert(previous == null)
   }
 
-  def userContext[T](username: String, timer: ExecutionTimer)(f: (ExecuteContext) => T): T = {
+  def userContext[T](username: String, sessionID: String, timer: ExecutionTimer)(f: (ExecuteContext) => T): T = {
     val user = users.get(username)
     assert(user != null, username)
-    ExecuteContext.scoped(user.tmpdir, "file:///tmp", this, user.fs, timer, null)(f)
+    ExecuteContext.scoped(user.tmpdir, "file:///tmp", this, user.fs.asCacheable(sessionID), timer, null)(f)
```
danking, line 73: Hmm. Tim has a PR that's going to make the user's code use the fs in the execute context. I think this has to be a normal FS until we can properly stream into Memory.

danking (changes requested):

danking (approved):
LGTM, let's keep in mind that we gotta do something about streaming large files soon.

PR #10316, opened 2021-04-13
[hailgenetics] update the hailgenetics images, fix Makefile

danking:
Now these images live at gcr.io/hail-vdc/hailgenetics/hail and gcr.io/hail-vdc/hailgenetics/genetics.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -260,9 +260,8 @@ class Container:
             image_ref.tag = 'latest'
 
         if image_ref.name() in HAIL_GENETICS_IMAGES:
-            repository_name_without_prefix = image_ref.name()[len(HAIL_GENETICS):]
             image_ref.domain = 'gcr.io'
-            image_ref.path = f'{PROJECT}/{repository_name_without_prefix}'
+            image_ref.path = f'{PROJECT}/{image_ref.name()}'
 
         self.image_ref = image_ref
         self.image_ref_str = str(image_ref)
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -25,7 +25,7 @@ from hailtop.utils import (time_msecs, request_retry_transient_errors, sleep_and
                            parse_docker_image_reference)
 from hailtop.httpx import client_session
 from hailtop.batch_client.parse import (parse_cpu_in_mcpu, parse_memory_in_bytes, parse_storage_in_bytes)
-from hailtop.batch.hail_genetics_images import HAIL_GENETICS, HAIL_GENETICS_IMAGES
+from hailtop.batch.hail_genetics_images import HAIL_GENETICS_IMAGES
 from hailtop import aiotools
 # import uvloop
 
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -86,28 +86,5 @@ clean:
 
 ## Public Images ###############################################################
 
-GCR_PREFIX = gcr.io/$(PROJECT)
-HAIL_IMAGE = hailgenetics/hail
-GENETICS_IMAGE = hailgenetics/genetics
-
-.PHONY: publish-hail-public-image
-publish-hail-public-image: hail-ubuntu-stmp
-	make -C ../hail python/hail/hail_pip_version
-	cp ../hail/python/hail/hail_pip_version hail/hail_pip_version
-	-docker pull $(GCR_PREFIX)/$(HAIL_IMAGE)
-	python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"hail-ubuntu"}}' Dockerfile.base Dockerfile.base.out
-	docker build hail/ -f hail/Dockerfile.out -t $(HAIL_IMAGE):$$(cat hail/hail_pip_version) \
-     --cache-from $(GCR_PREFIX)/$(HAIL_IMAGE),$(HAIL_IMAGE)
-	docker push $(HAIL_IMAGE):$$(cat hail/hail_pip_version)
-	docker tag $(HAIL_IMAGE):$$(cat hail/hail_pip_version) $(GCR_PREFIX)/$(HAIL_IMAGE):$$(cat hail/hail_pip_version)
-	docker push $(GCR_PREFIX)/$(HAIL_IMAGE):$$(cat hail/hail_pip_version)
-
-.PHONY: publish-genetics-public-image
-publish-genetics-public-image: publish-hail-public-image
-	python3 ../ci/jinja2_render.py '{"hail_public_image":{"image":"'$(HAIL_IMAGE):$$(cat hail/hail_pip_version)'"}}' genetics/Dockerfile genetics/Dockerfile.out
-	-docker pull $(GCR_PREFIX)/$(GENETICS_IMAGE)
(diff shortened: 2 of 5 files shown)


docker/Makefile
```diff
@@ -108,3 +83,31 @@ deploy: push
...
+HAIL_IMAGE = hailgenetics/hail
+GENETICS_IMAGE = hailgenetics/genetics
+
+.PHONY: publish-hail-public-image
+publish-hail-public-image: hail-ubuntu-stmp
+	make -C ../hail python/hail/hail_pip_version
+	cp ../hail/python/hail/hail_pip_version hail/hail_pip_version
+	-docker pull $(GCR_PREFIX)/$(HAIL_IMAGE)
+	python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"hail-ubuntu"}}' Dockerfile.base Dockerfile.base.out
+	docker build hail/ -f hail/Dockerfile.out -t $(HAIL_IMAGE):$$(cat hail/hail_pip_version) \
```
jigold, line 99: Can we use more variables here? You repeat `$$(cat hail/hail_pip_version)` a lot and it makes it harder to read. Maybe also a variable for `$(HAIL_IMAGE):$$(cat hail/hail_pip_version)`

docker/Makefile
```diff
@@ -108,3 +83,31 @@ deploy: push
...
+     --cache-from $(GCR_PREFIX)/$(HAIL_IMAGE),$(HAIL_IMAGE)
+	docker push $(HAIL_IMAGE):$$(cat hail/hail_pip_version)
+	docker tag $(HAIL_IMAGE):$$(cat hail/hail_pip_version) $(GCR_PREFIX)/$(HAIL_IMAGE):$$(cat hail/hail_pip_version)
+	docker push $(GCR_PREFIX)/$(HAIL_IMAGE):$$(cat hail/hail_pip_version)
+
+.PHONY: publish-genetics-public-image
+publish-genetics-public-image: publish-hail-public-image
+	python3 ../ci/jinja2_render.py '{"hail_public_image":{"image":"'$(HAIL_IMAGE):$$(cat hail/hail_pip_version)'"}}' genetics/Dockerfile genetics/Dockerfile.out
+	-docker pull $(GCR_PREFIX)/$(GENETICS_IMAGE)
+	docker build genetics/ -f genetics/Dockerfile.out -t $(GENETICS_IMAGE):$$(cat hail/hail_pip_version) \
```
jigold, line 109: Same comment as above.

batch/batch/worker/worker.py
```diff
@@ -261,8 +261,7 @@ def __init__(self, job, name, spec):
             tag = 'latest'
 
         if repository in HAIL_GENETICS_IMAGES:
-            repository_name_without_prefix = repository[len(HAIL_GENETICS):]
-            repository = f'gcr.io/{PROJECT}/{repository_name_without_prefix}'
+            repository = f'gcr.io/{PROJECT}/{repository}'
```
jigold, line 264: The batch pool executor tests will catch if this is wrong -- correct?

docker/genetics/Dockerfile
```diff
@@ -5,6 +5,9 @@ RUN hail-apt-get-install \
     libbz2-dev \
     libcurl4-openssl-dev \
     liblzma-dev \
+    libvcflib-tools \
```
jigold, line 8: Why did you add these?

docker/genetics/Dockerfile
```diff
@@ -16,8 +19,6 @@ RUN mkdir samtools && \
      ./configure --without-curses && \
      make && \
      make install)
-RUN git clone --recursive https://github.com/vcflib/vcflib.git && \
```
jigold, line 19: Why did you delete this line?

jigold (changes requested):

docker/genetics/Dockerfile
danking: Compiling VCFlib was hard and they now have an apt-get package

docker/Makefile
danking: It's difficult to do what you're looking for in Make, so I rewrote this as a shell script. Let me know what you think.

batch/batch/worker/worker.py
danking: Yeah, the images would be absent and that would cause an error. I've already uploaded the images to GCR using the first version of this code. I just ran it again now to ensure the new thing is right too.

docker/genetics/Dockerfile
danking: compiling VCFlib was hard

docker/publish-public-images.sh
```diff
@@ -0,0 +1,37 @@
...
+           -t $docker \
+           -t $google \
+           -t $latest \
+           --cache-from $latest,$base
+    docker push $docker
+    docker push $google
+    docker push $latest
+}
+
+python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"hail-ubuntu"}}' Dockerfile.base Dockerfile.base.out
```
jigold, line 32: Does this actually work? It looks like it's expecting the Dockerfile to be located at `hail/Dockerfile.out`

docker/publish-public-images.sh
```diff
@@ -0,0 +1,37 @@
...
+
+make -C ../hail python/hail/hail_pip_version
+cp ../hail/python/hail/hail_pip_version hail/hail_pip_version
+hail_pip_version=$(cat hail/hail_pip_version)
+
+build_and_push() {
+    name=$1
+    base=$2
+
+    docker=hailgenetics/$name:$hail_pip_version
```
jigold, line 15: Did you decide not to tag these with the SHA anymore?

jigold (dismissed):

docker/publish-public-images.sh
danking: ah nope, I should've cleaned before I tested.

docker/publish-public-images.sh
danking: These images were never tagged with a SHA, they're tagged with the pip version because their version of hail is from PyPI e.g. https://hub.docker.com/r/hailgenetics/genetics/tags?page=1&ordering=last_updated

jigold: Sorry. I should have been more clear. We used to have this as well: `GENETICS_PUBLIC_IMAGE = gcr.io/$(PROJECT)/genetics-public:$(shell docker images -q --no-trunc genetics-public:latest | sed -e 's,[^:]*:,,')`

docker/publish-public-images.sh
```diff
@@ -0,0 +1,37 @@
...
+           -t $docker \
+           -t $google \
+           -t $latest \
+           --cache-from $latest,$base
+    docker push $docker
+    docker push $google
+    docker push $latest
+}
+
+python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"hail-ubuntu"}}' hail/Dockerfile.base hail/Dockerfile.base.out
```
jigold, line 32: Maybe this was a different PR, but I think the Dockerfile name is not correct here.

jigold (changes requested):

jigold (approved):

PR #10317, opened 2021-04-13
[hailtop.batch] avoid serializing whole module in tests

danking:
It seems that referencing a global function serializes a module named `test` which
does not exist on the worker.


jigold (approved):

danking:
I think this is now breaking due to issues fixed in #10322

PR #10318, opened 2021-04-14
[infra] add hail_version in TLS cookbook

lgruen:
#assign services (@daniel-goldstein Thanks for implementing this, it's awesome!)

### DIFF
--- a/dev-docs/tls-cookbook.md
+++ b/dev-docs/tls-cookbook.md
@@ -1,4 +1,5 @@
 # TLS Cookbook
+
 ## Create a Self-Signed x509 Certificate in PEM Format
 
 Produce an x509 certificate. The key is a 4096-bit RSA key. The cert is valid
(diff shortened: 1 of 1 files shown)


dev-docs/tls-cookbook.md
```diff
@@ -121,6 +122,10 @@ kubectl create secret generic \
 3. Update all the service certificates:
 
 ```
+export HAIL=$HOME/hail
```
danking, line 125: Let's remove this, in dev-docs/hail-overview.md we state that $HAIL is the root of the hail repo. I think we can assume they've seen that before they come here. Some folks (myself) keep the hail repository elsewhere.

danking (approved):

dev-docs/tls-cookbook.md
lgruen: Sure, done!

PR #10320, opened 2021-04-14
[build] Fix dependencies to strip comments

tpoterba:


hail/python/setup.py
```diff
@@ -12,12 +13,19 @@
 dependencies = []
 with open('requirements.txt', 'r') as f:
     for line in f:
-        if line.startswith('pyspark') and os.path.exists('../env/SPARK_VERSION'):
+        stripped = line.strip()
+        if stripped.startswith('#') or len(stripped) == 0:
+            continue
+
+        # strip trailing comments
+        pkg = re.split('\\s+#', stripped)[0]
```
danking, line 21: Trailing comments aren't actually kosher, pip errors if they exist.

danking (changes requested):
Sorry, my bad!

danking (approved):

PR #10323, opened 2021-04-14
[batch] more info after docker close

daniel-goldstein:


batch/batch/worker/worker.py
```diff
@@ -1503,6 +1503,17 @@ async def async_main():
         finally:
             await docker.close()
             log.info('docker closed')
+            asyncio.get_event_loop().set_debug(True)
+            log.debug('Tasks immediately after docker close')
+            dump_all_stacktraces()
+            await asyncio.sleep(10 * 60)
```
danking, line 1509: I think this keeps the worker up 10 minutes even if there are no background tasks or if the background tasks terminate early.

I think I prefer something like:
```python3
other_tasks = [t for t in asyncio.tasks() if t != asyncio.current_task()]
try:
    done, pending = await asyncio.wait(other_tasks, timeout=10 * 60, return_when=asyncio.ALL_COMPLETED)
except asyncio.TimeoutError:
    for t in pending:
        t.cancel()
```

danking (changes requested):

batch/batch/worker/worker.py
```diff
@@ -1503,6 +1503,15 @@ async def async_main():
         finally:
             await docker.close()
             log.info('docker closed')
+            asyncio.get_event_loop().set_debug(True)
+            log.debug('Tasks immediately after docker close')
+            dump_all_stacktraces()
+            other_tasks = [t for t in asyncio.tasks() if t != asyncio.current_task()]
+            _, pending = await asyncio.wait(other_tasks, timeout=10 * 60, return_when=asyncio.ALL_COMPLETED)
```
danking, line 1510: You'll get an error if `other_tasks` is empty. Maybe add:
```
if not other_tasks:
    return
```

danking (dismissed):

danking (approved):

PR #10325, opened 2021-04-14
[batch] cleanup error message for unknown instance

jigold:
If instance name isn't active, look to see whether it existed in the database before printing an error message about an unknown instance.

### DIFF
--- a/batch/batch/driver/gce.py
+++ b/batch/batch/driver/gce.py
@@ -95,9 +95,7 @@ class GCEEventMonitor:
         else:
             instance = self.inst_coll_manager.get_instance(name)
             if not instance:
-                record = await self.db.select_and_fetchone(f'''
-SELECT name FROM instances WHERE name = %s;
-''',
+                record = await self.db.select_and_fetchone('SELECT name FROM instances WHERE name = %s;',
                                                            (name,))
                 if not record:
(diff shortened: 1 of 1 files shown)


daniel-goldstein (dismissed):
In what scenario would the instance name not be in the database? Is there a way to prevent receiving these events in the first place? I think I'm confused about why we're getting these events from GCE.

jigold:
It should never not be in the database. That's why it should be an error if we get an event for an instance we've never created before. The reason we keep getting these events is because when we run this query:

```
        filter = f'''
logName="projects/{PROJECT}/logs/cloudaudit.googleapis.com%2Factivity" AND
resource.type=gce_instance AND
protoPayload.resourceName:"{self.machine_name_prefix}" AND
timestamp >= "{mark}"
'''
```

We have `timestamp >= {mark}`. This means if `mark` doesn't change each time the event polling loop reruns, then we'll always keep getting the same event as the last one processed. We need that `>=` though because different events can have the same timestamp. So the driver could have been shut down in the middle of processing multiple events with the same timestamp. So we do the conservative thing and try to reprocess the event again until the mark changes to a different timestamp.

Does that make sense?

daniel-goldstein (approved):
Thanks for the clarification. Makes sense

PR #10326, opened 2021-04-14
[query] Push FS through EmitFunctionBuilder

tpoterba:


hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
```diff
@@ -27,42 +27,15 @@ class EmitModuleBuilder(val ctx: ExecuteContext, val modb: ModuleBuilder) {
...
-        val ba = b64.invoke[String, Array[Byte]]("decode", stringAssembler)
-        val bais = Code.newInstance[ByteArrayInputStream, Array[Byte]](ba)
-        val ois = cb.newLocal("ois", Code.newInstance[ObjectInputStream, InputStream](bais))
-        Code.checkcast(ois.invoke[Any]("readObject"))(fsinfo)
-      }
-      val fs = cls.newStaticField[FS]("filesystem", mb.invokeCode())
-
-      _staticFS = new StaticFieldRef(fs)
-    }
-    _staticFS
```
danking, line 63: love it

danking (commented):
I'm good with the changes to service code. Should someone from compilers team do a once over of the rest?

danking (changes requested):
x

tpoterba:
Good call. Assigning to Chris for another look, will move assignee.

tpoterba:
@chrisvittal - the motivation for this change is that we don't want to serialize the FS in the compiled function because it contains secrets. Instead, clients who call compiled functions should pass in the FS.

chrisvittal (approved):
lgtm

PR #10329, opened 2021-04-15
[batch] Fix UI to make user errors clear

jigold:


danking (approved):
Ah, this will be a really nice improvement! Thanks Jackie!

PR #10331, opened 2021-04-15
[batch] replace GCS with GoogleStorageAsyncFS in LogStore

jigold:
Also removed GCS from `batch_pool_executor` and `Batch,PythonJob`


jigold:
I left in google_storage.py in for now because Benchmark uses it and I didn't want to rewrite that right now.

batch/batch/log_store.py
```diff
@@ -11,51 +11,69 @@
...
         if not format_version.has_attempt_in_log_path():
             return f'{self.batch_log_dir(batch_id)}/{job_id}/{task}/log'
-
         return f'{self.batch_log_dir(batch_id)}/{job_id}/{attempt_id}/{task}/log'
 
     async def read_log_file(self, format_version, batch_id, job_id, attempt_id, task):
-        path = self.log_path(format_version, batch_id, job_id, attempt_id, task)
-        return await self.gcs.read_gs_file(path)
+        url = self.log_path(format_version, batch_id, job_id, attempt_id, task)
+        return await self.gcs_fs.open(url)
```
danking, line 53: open returns a file doesn't it? I think we should `.read()` and convert from bytes to a string

batch/batch/log_store.py
```diff
@@ -11,51 +11,69 @@
...
+        await self._rmtree(url)
 
     def status_path(self, batch_id, job_id, attempt_id):
         return f'{self.batch_log_dir(batch_id)}/{job_id}/{attempt_id}/status.json'
 
     async def read_status_file(self, batch_id, job_id, attempt_id):
-        path = self.status_path(batch_id, job_id, attempt_id)
-        return await self.gcs.read_gs_file(path)
+        url = self.status_path(batch_id, job_id, attempt_id)
+        return await self.gcs_fs.open(url)
```
danking, line 68: same here

memory/memory/client.py
```diff
@@ -1,17 +1,14 @@
-
 import aiohttp
-import concurrent
-import google.api_core.exceptions
```
danking, line 4: fyi @daniel-goldstein there's a merge conflict lurking here

danking (changes requested):
This is awesome! just a few small comments

memory/memory/client.py
daniel-goldstein: 🏇

jigold:
@daniel-goldstein Did your PR that would have a merge conflict go in yet? If not, do you have an ETA on that?

danking:
No, it's not approved yet

daniel-goldstein:
It's #10315, but don't let it hold you up. Won't be hard to fix the conflicts.

batch/batch/driver/main.py
```diff
@@ -903,8 +901,6 @@ async def scheduling_cancelling_bump(app):
 
 async def on_startup(app):
     app['task_manager'] = aiotools.BackgroundTaskManager()
-    pool = concurrent.futures.ThreadPoolExecutor()
-    app['blocking_pool'] = pool
```
danking, line 907: I'm so happy about this!

batch/batch/log_store.py
```diff
@@ -11,50 +11,76 @@
...
+    async def _read(self, url, binary=True):
+        async with await self.gcs_fs.open(url) as f:
+            data = await f.read()
+            if not binary:
+                data = data.decode('utf-8')
+            return data
+
+    async def _rmtree(self, url):
+        async with self.sem:
+            await self.gcs_fs.rmtree(self.sem, url)
```
danking, line 41: Why is there only a semaphore on rmtree?

batch/batch/log_store.py
```diff
@@ -11,50 +11,76 @@
 
 
 class LogStore:
```
danking, line 13: This isn't really a LogStore, its something like "GCS API" because it deals with specs and logs. When I first saw this file, I went hunting for the `SpecStore` because I assumed such a class must exist since are specs are also stored in GCS.

batch/batch/log_store.py
```diff
@@ -11,50 +11,76 @@
...
-        self.gcs = GCS(blocking_pool, project=project, credentials=credentials)
+
+        self.gcs_fs = GoogleStorageAsyncFS(credentials=credentials)
+        self.sem = asyncio.Semaphore(50)
 
         log.info(f'BATCH_LOGS_ROOT {self.batch_logs_root}')
         format_version = BatchFormatVersion(BATCH_FORMAT_VERSION)
         log.info(f'EXAMPLE BATCH_JOB_LOGS_PATH {self.log_path(format_version, 1, 1, "abc123", "main")}')
 
+    async def _write(self, url, data, binary=True):
```
danking, line 26: Hmm.

OK, so this makes me think two things:
1. The AsyncFS API is too narrow.
2. This function will reappear many time all across our code base when someone needs to read data from a filesystem.

I think we can improve the code base as a whole by pushing your new functions *into* AsyncFS. `pathlib` is the only example I can think of as inspiration for names. It has [`read_text`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text) and `write_text`. IMHO, I prefer explicit use of `encode` and `decode` in user code (it forces me to think about the encoding).

Concretely, my proposal is to add two operations to `AsyncFS`:
```python3
    @abc.abstractmethod
    async def open(self, url: str) -> ReadableStream:
        pass

    async def read(self, url: str) -> bytes:
        async with self.open(url) as reader:
            reader.read(data)
    # ...
    @abc.abstractmethod
    async def create(self, url: str, *, retry_writes: bool = True) -> WritableStream:
        pass

    async def write(self, url: str, data: bytes):
        async with self.create(url) as writer:
            writer.write(data)
```
And then to change `LogStore` to have, for example:
```python3
async def read_log_file(...):
    url = # ...
    return await self.fs.read(url).decode('utf-8')

async def write_log_file(..., data: bytes):
    url = # ...
    return await self.fs.write(url, data.encode('utf-8'))

# ...
async def read_spec_file(...):
    # ...
    await self.fs.read(url)
```

batch/batch/log_store.py
```diff
@@ -11,50 +11,76 @@
 
 
 class LogStore:
-    def __init__(self, batch_logs_bucket_name, instance_id, blocking_pool, *, project=None, credentials=None):
+    def __init__(self, batch_logs_bucket_name, instance_id, *, credentials=None):
```
danking, line 14: I prefer that we don't deeply integrate a particular choice of FS with the LogStore. Let's have the LogStore take an `fs: FS` argument instead.

batch/batch/log_store.py
danking: We could also have a LogStore and a SpecStore, 🤷 . I don't see two classes or one as obviously better or worse, as long as we use a descriptive name.

memory/memory/client.py
```diff
@@ -18,9 +18,11 @@ def __init__(self, gcs_project=None, fs=None, deploy_config=None, session=None,
         self.url = self._deploy_config.base_url('memory')
         self.objects_url = f'{self.url}/api/v1alpha/objects'
         self._session = session
+
         if fs is None:
-            fs = GCS(blocking_pool=concurrent.futures.ThreadPoolExecutor(), project=gcs_project)
+            fs = GoogleStorageAsyncFS(project=gcs_project)
         self._fs = fs
```
danking, line 24: this needs to be closed in the close method of MemoryClient.

memory/memory/memory.py
```diff
@@ -54,7 +54,7 @@ async def write_object(request, userdata):
     filepath = request.query.get('q')
     userinfo = await get_or_add_user(request.app, userdata)
     username = userdata['username']
-    data = await request.read()
+    data = bytes(await request.read())
```
danking, line 57: Can you help me understand this? `BaseRequest.read` claims it returns bytes.

memory/memory/memory.py
```diff
@@ -115,17 +117,18 @@ async def load_file(redis, files, file_key, fs, filepath):
     await cache_file(redis, files, file_key, filepath, data)
 
 
-async def persist_in_gcs(fs: GCS, files: Set[str], file_key: str, filepath: str, data: str):
+async def persist_in_gcs(fs: GoogleStorageAsyncFS, files: Set[str], file_key: str, filepath: str, data: bytes):
     try:
         log.info(f"memory: {file_key}: persisting.")
-        await fs.write_gs_file_from_string(filepath, data)
+        async with await fs.create(filepath) as f:
+            await f.write(data)
```
danking, line 124: another use case for `await fs.write(filepath, data)`.

danking (changes requested):
So excited for this!

danking:
Is there a reason to not also fix `hailtop.batch.batch` and `hailtop.batch.batch_pool_executor`? Once those are gone, we have no references left to GCS!

jigold:
We'd need a synchronous version of the file system. I thought this was something Cotton wanted to do himself. With regards to the interface changes, I'm not sure if Cotton had something else in mind.

danking:
BPE already uses async, so we can go full AsyncFS there. It looks like, for Batch, we just need a little `async_to_blocking` inside job.py. Does that not feel like an OK solution until we have a full fledged sync solution?

batch/batch/log_store.py
jigold: Because Cotton removes files in parallel. But you have to have a maximum parallelism for that. The other file system operations aren't in parallel.

jigold:
This is finally passing!!!!

memory/memory/client.py
danking: this comment has not been addressed yet

memory/memory/memory.py
danking: Bump on this comment, I thought `read` returns bytes.

memory/test/test_memory.py
```diff
@@ -33,28 +33,40 @@ def setUp(self):
...
+        async with await self.fs.create(handle) as f:
+            await f.write(str_value)
+
         return handle
 
     def test_non_existent(self):
         for _ in range(3):
             self.assertIsNone(self.client._get_file_if_exists(f'{self.test_path}/nonexistent'))
 
     def test_small_write_around(self):
```
danking, line 61: Heh. That's super frustrating. As far as I can tell, unittest doesn't provide great support for async tests. Pytest works fine:
```
pytestmark = pytest.mark.asyncio  # once per file

async def test():
    assert 3 == 4
```

In the future, when we need to add new async test suites we should use pytest, replacing class fields with pytest fixtures. That'll ensure we can easily test async code without needing to make these blocking shim wrappers.

memory/test/test_memory.py
```diff
@@ -33,28 +33,40 @@ def setUp(self):
...
         self.client = BlockingMemoryClient(fs=self.fs)
         self.temp_files = set()
 
     def tearDown(self):
-        async_to_blocking(self.fs.delete_gs_files(self.test_path))
+        async def delete_files(url):
+            async with self.sem:
+                await self.fs.rmtree(self.sem, url)
+
+        async_to_blocking(delete_files(self.test_path))
```
danking, line 46: It seems simpler to use:

```
async_to_blocking(self.fs.rmtree(None, url)
```
It also saves unnecessarily adding `self.sem` as a class field, where it seems to be unused except for in the tearDown.

danking (changes requested):
Looking good! I'm so excited for google_storage.py to go away.

hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -344,6 +345,7 @@ async def async_submit(self,
         dill.dump(functools.partial(unapplied, *args, **kwargs), pipe, recurse=True)
         pipe.seek(0)
         pickledfun_gcs = self.inputs + f'{name}/pickledfun'
+        await self.gcs_fs.makedirs(os.path.dirname(pickledfun_gcs), exist_ok=True)
```
danking, line 348: I think I'm missing something about this code. It shouldn't be necessary for an object store and certainly isn't necessary for GCS. We know `gcs_fs` is a GCS client and any future evolution of this code would at most make it a generic object store client.

I think what I'm generally reacting to is: I don't expect to have to create directories in an object store; therefore, when I read this line, I wonder what hidden subtle issues are addressed by this line. I can't seem to think of any?

danking: I suppose the alternative is that we, as Hail team, make it a rule that we will *always* use `makedirs` on any file stored in GCS. I don't quite see the benefit of that, but consistently doing it everywhere at least avoids my confusion (because it will be common and therefore not surprising).

danking (dismissed):

hail/python/hailtop/batch/batch_pool_executor.py
jigold: The issue is the file system is actually a RouterFileSystem. I'll clean up the naming of that variable. This is needed for the python job to work on the local backend. The local backend needs makedirs to be called to write the file to the correct place. We can check explicitly whether `pickledfun_gcs` is local or remote, but that seems to defeat the purpose of the router file system at least to me.

jigold: bump.

danking: This variable always a GoogleStorageAsyncFS (line 137). The BatchPoolExecutor doesn't really make sense locally. Locally, it's just a slower ProcesPoolExecutor. If we wanted to do this, I'd create the directories once in the constructor.

The BatchPoolExecutor already asserts that its using a ServiceBackend, so there's no concern that someone might try to uses the BatchPoolExecutor on the local backend.

danking (dismissed):

hail/python/hailtop/batch/batch.py
```diff
@@ -138,16 +138,15 @@ def __init__(self,
...
     @property
-    def _gcs(self):
-        if self.__gcs is None:
-            self.__gcs = GCS(blocking_pool=concurrent.futures.ThreadPoolExecutor(),
-                             project=self._project)
-        return self.__gcs
+    def _fs(self):
+        if self.__fs is None:
+            self.__fs = RouterAsyncFS('file', [LocalAsyncFS(ThreadPoolExecutor()), GoogleStorageAsyncFS(project=self._project)])
+        return self.__fs
```
danking, line 149: This needs to be closed. I think it needs to live on the backend, not the batch. The backend has a close method.

danking (dismissed):

hail/python/hailtop/batch/batch.py
jigold: I think this was addressed in #10562.

danking (approved):

PR #10332, opened 2021-04-15
[datasets] add GTEx sQTL tissue specific tables

pwc2:
Adds the 49 [GTEx](https://gtexportal.org/home/datasets) sQTL tissue-specific all SNP gene association datasets as Hail Tables.

Sourced from `gs://gtex-resources/GTEx_Analysis_v8_QTLs/GTEx_Analysis_v8_sQTL_all_associations`. 

Files were .gz so had to be rewritten as .bgz before creating the Hail Tables.

### DIFF
--- a/datasets/notebooks/gtex_datasets.ipynb
+++ b/datasets/notebooks/gtex_datasets.ipynb
@@ -3,9 +3,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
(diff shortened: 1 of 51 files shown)


danking (approved):
Awesome. Thanks! I really like the explicit, executable description of how the datasets, json, and reference pages are created.

PR #10342, opened 2021-04-16
Clarify the behavior of OnlineBoundedGather2.

cseed:
I think this should resolve the issues we were seeing with OnlineBoundedGather2.

Changes:
 - cancelled tasks (those that raise CancelledError) are ignored (we don't propagate cancelled out of background tasks)
 - Make sure all exceptions are either reraised or logged
 - The first exception is raised out of exit, not call
 - call raises PoolShutdownError if the pool is shutdown
 - _shutdown doesn't signal _done_event until all cancelled tasks are complete
 - call clears _done_event (not strictly necessary because exit checks pending, but seems safer)
 - added copious docstrings

### DIFF
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -392,7 +392,7 @@ class OnlineBoundedGather2:
         '''
 
         if self._pending is None:
-            raise PoolShutdownException
+            raise PoolShutdownError
 
         id = self._counter
         self._counter += 1
@@ -419,6 +419,7 @@ class OnlineBoundedGather2:
 
         t = asyncio.create_task(run_and_cleanup())
         self._pending[id] = t
+        self._done_event.clear()
         return t
 
     async def wait(self, tasks: List[asyncio.Task]) -> None:
@@ -454,6 +455,7 @@ class OnlineBoundedGather2:
             else:
                 log.info('discarding exception', exc_info=exc_val)
 
+        await self._done_event.wait()
         while self._pending:
             self._done_event.clear()
             await self._done_event.wait()
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -355,7 +355,12 @@ class OnlineBoundedGather2:
         self._counter = 0
         self._subsema = Subsemaphore(sema)
         self._pending: Optional[Dict[int, asyncio.Task]] = {}
+        # done if there are no pending tasks (the tasks are all
+        # complete), or if we've shutdown and the cancelled tasks are
+        # complete
         self._done_event = asyncio.Event()
+        # not pending tasks, so done
+        self._done_event.set()
         self._exception: Optional[BaseException] = None
 
     async def _shutdown(self) -> None:
@@ -379,7 +384,6 @@ class OnlineBoundedGather2:
         if tasks:
             await asyncio.wait(tasks)
 
-        # wake up if waiting
         self._done_event.set()
 
     async def call(self, f, *args, **kwargs) -> asyncio.Task:
@@ -455,9 +459,12 @@ class OnlineBoundedGather2:
             else:
                 log.info('discarding exception', exc_info=exc_val)
 
+        # wait for done and not pending _done_event.wait can return
+        # when when there are pending jobs if the last job completed
+        # (setting _done_event) and then more tasks were submitted
         await self._done_event.wait()
         while self._pending:
-            self._done_event.clear()
+            assert not self._done_event.is_set()
             await self._done_event.wait()
 
         if self._exception:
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/utils/utils.py
```diff
@@ -325,7 +325,32 @@ async def __aexit__(self,
...
+    pass
+
+
 class OnlineBoundedGather2:
+    '''`OnlineBoundedGather2` provides the capability to run background
+    tasks with bounded parallelism.  It is a context manager, and
+    waits for all background tasks to complete on exit.
+
+    `OnlineBoundedGather2` supports cancellation of background tasks.
+    When a background tasks raises `asyncio.CancelledError`, the task
```
jigold, line 338: should be task instead of tasks.

hail/python/hailtop/utils/utils.py
```diff
@@ -407,11 +448,20 @@ async def __aexit__(self,
...
-        else:
-            await self.wait_done()
+        if exc_val:
+            if self._exception is None:
+                self._exception = exc_val
+                await self._shutdown()
+            else:
+                log.info('discarding exception', exc_info=exc_val)
+
+        await self._done_event.wait()
```
jigold, line 458: Is this first wait supposed to be here? Otherwise, all the code is the same except for the refactoring.

hail/python/hailtop/utils/utils.py
```diff
@@ -345,15 +376,23 @@ async def _shutdown(self) -> None:
...
-        self._done_event.set()
-
         if tasks:
             await asyncio.wait(tasks)
 
+        # wake up if waiting
+        self._done_event.set()
+
     async def call(self, f, *args, **kwargs) -> asyncio.Task:
-        if self._exception:
```
jigold, line 355: Where does the exception get raised in `_shutdown`? I see it in `__aexit__`. I tried to see how this class is used and I couldn't find any usages of it in hailtop.

jigold (dismissed):

hail/python/hailtop/utils/utils.py
cseed: Good catch!

hail/python/hailtop/utils/utils.py
cseed: OK, I changed this a bit to be clearer and I'm happier with it.  The invariant is:

_done_event is set if (1) we are not shut down and _pending is empty, or (2) we did shutdown and all the cancelled tasks are complete.  That means we can be shutdown, _pending can be empty, but _done_event is not (yet) set.

It also means _done_event can be set and then cleared again, if the last task completes but another one is then submitted.

Then I changed the code to maintain this invariant.  If we add a task to _pending, I clear _done_event.  If the last pending task finishes, set it.  In shutdown, if all the cancelled tasks are complete, set it, etc.

Finally, __aexit__ needs to wait for _event_done to be set _and_ pending to be empty (or None).

Make sense?

hail/python/hailtop/utils/utils.py
```diff
@@ -362,41 +405,43 @@ async def run_and_cleanup():
...
+            except asyncio.CancelledError:
+                pass
             except:
-                if not self._exception:
+                if self._exception is None:
                     _, exc, _ = sys.exc_info()
                     self._exception = exc
                     await self._shutdown()
+                else:
+                    log.info('discarding exception', exc_info=True)
```
jigold, line 416: Do we want this message in user logs? Or will that be too confusing?

jigold (changes requested):

hail/python/hailtop/utils/utils.py
cseed: I'm not sure what to do here.  Once we introduce concurrency, multiple exceptions can happen simultaneously, and the question is how to resolve that.  I see a few options:

1. (This is the first path I took.)  Raise the first exception and log (using `info`) any secondary exceptions.  I felt like this was OK because the user always got the first/primary exception.  (You might argue about which exception I'm discarding.  For example, in aexit, I discard the the task manager exception in favor of an earlier task exception, and maybe we should switch that?)

2. Log the exceptions more forcefully (with `error`, say).  Problem is, I don't really know if it is an error.  This can also potentially generate a large number of exception messages if there are a large number of tasks running concurrently, and I wanted to avoid that (although maybe we shouldn't).

3. Gather all the exceptions together and raise them with a new exception type, like `MultipleExceptions(exceptions)`.  This has the benefit that it doesn't loose information, except you loose all knowledge about the original exception types so callers can't reasonably use try/except, etc.

Thoughts @danking?  It seems like it would be good to have a standard convention for this case.

jigold: I'm going to go ahead and approve this because Lindo is waiting on this fix. We should revisit this discussion later though.

jigold (approved):

PR #10344, opened 2021-04-17
[libhail] restyle includes

chrisvittal:
Includes are grouped by source (system, LLVM, internal), the groups
themselves are then ordered stdlib, dependencies, internal. Ordering
within groups is alphabetical.

Also switch internal include style from #include <> to #include ""


chrisvittal:
Feel free to reject this, it's basically one big style nit. I recognize that it's largely my preference, but if we're going to go with it (I hope) it's better to do it asap.

cseed (approved):

PR #10345, opened 2021-04-19
[batch] worker needs to await JVMJob cleanup as well

danking:


daniel-goldstein:
`await` strikes again

daniel-goldstein (approved):

PR #10348, opened 2021-04-20
[build.yaml] use consistent formatting

danking:
I ran our build.yaml through yq and am committing the results.


danking:
yq broke the script tags, so I manually fixed all of those. If you turn on ignore whitespace, it's a small diff.

The important change is standardizing how many spaces a list item has before it (two, always).

daniel-goldstein (approved):

PR #10349, opened 2021-04-20
[docker] add Boto3 dependency

cseed:
boto3 is the AWS Python client library.  This is for the upcoming S3AsyncFS.  FYI @tpoterba I also added it to the hail package requirements file because it will be exposed via `hailctl cp ...`, etc.

### DIFF
--- a/docker/requirements.txt
+++ b/docker/requirements.txt
@@ -57,7 +57,7 @@ sortedcontainers==2.1.0
 tabulate==0.8.3
 tqdm==4.42.1
 twine>=1.11.0
-urllib3==1.24.3
+urllib3==1.26.4
 uvloop==0.14.0
 Werkzeug==0.15.4
 wheel>=0.31.0
--- a/docker/requirements.txt
+++ b/docker/requirements.txt
@@ -51,7 +51,7 @@ pytest-instafail==0.4.2
 pytest-xdist==2.2.1
 python-dateutil==2.8.1
 python-json-logger==0.1.11
-requests==2.22.0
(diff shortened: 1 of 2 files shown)


cseed:
Looks like there is a version conflict on setuptools.  I'm investigating.  However, it looks like we're not pinning in our requirements file:

 > setuptools>=38.6.0

Do you know if there is a reason for that?

danking:
For `hail` we’ve tried to be flexible, but We can move the lower bound or add an upper bound if necessary. I’d this is services stuff that’s just a mistake, definitely pin if you’d like.

cseed:
The failure doesn't appear to be related to my changes.  Installing the docker requirements, which has `setuptools>=38.6.0`, is trying to upgrade to the latest setuptools (56.0.0).  Another dependency might be forcing the upgrade.  However, setuptools was installed via apt, not pip, and that is causing this:

```
  Attempting uninstall: setuptools
    Found existing installation: setuptools 45.2.0
    Not uninstalling setuptools at /usr/lib/python3/dist-packages, outside environment /usr
    Can't uninstall 'setuptools'. No files were found to uninstall.
```

So there's two things I don't understand.  I'll keep investigating.  I glad I PRed this separately!

danking (approved):
No objections, I'm good with however you solve the setup tools thing

cseed:
The setuptools thing was actually a red herring, that error isn't fatal, and there was a version conflict elsewhere.  I upgraded urllib3 and requests to resolve.

PR #10354, opened 2021-04-21
[ci] allow Dockerfile to come from inputs

danking:


daniel-goldstein (approved):
How did this work before…

danking:
The docker file always came directly from the repository rather than from the inputs.

ci/test/resources/build.yaml
```diff
@@ -55,6 +55,8 @@ steps:
     inputs:
       - from: /repo/docker/hail-ubuntu
         to: /inputs/docker/hail-ubuntu
+    dependsOn:
+      - code
```
daniel-goldstein, line 59: Isn't this from the other PR?

daniel-goldstein (changes requested):
Some dockerfiles (e.g. Grafana) depend on a templated input file to render the Dockerfile. How could we have been rendering the Dockerfile before bringing in those inputs?

ci/test/resources/build.yaml
danking: oh yep, that should be copy_files

danking:
Grafana's Dockerfile is templated, but the template, `grafana/Dockerfile.nginx` came from the repository.

daniel-goldstein:
Ah right, the dockerfile isn't actually used until after the inputs are brought in.

danking (commented):
FYI Daniel, I got super frustrated trying to make this backwards compatible, so I just added a buildImage2 step, copied buildImage and removed all the unnecessary input manipulation and repository cloning.

After this PR and the next one go in, I can remove buildImage and rename buildImage2 to buildImage again.

danking:
@daniel-goldstein ok fixed!

daniel-goldstein (approved):
LGTM

PR #10356, opened 2021-04-21
[hail-ubuntu] avoid software-properties-common, a huge package

danking:
We can directly download the key and specify the repository. I used these instructions:
- https://wiki.debian.org/DebianRepository/UseThirdParty
- https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa

This builds in like 2 minutes rather than 4.

The frontend part avoids issues in downstream docker files where the installer might try to interact with the user. I noticed that some package was updated and now pulls in `tzdata` which asks you to select a timezone at install-time (if Debian frontend is interactive).

### DIFF
--- a/docker/hail-ubuntu/Dockerfile
+++ b/docker/hail-ubuntu/Dockerfile
@@ -6,11 +6,7 @@ COPY hail-apt-get-install /bin/hail-apt-get-install
 COPY pip.conf /root/.config/pip/pip.conf
 COPY hail-pip-install /bin/hail-pip-install
 COPY controller.sh /
-RUN chmod 755 /bin/retry && \
-    chmod 755 /bin/hail-apt-get-install && \
-    chmod 755 /bin/hail-pip-install && \
-    chmod 755 /controller.sh && \
-    echo "APT::Acquire::Retries \"5\";" > /etc/apt/apt.conf.d/80-retries && \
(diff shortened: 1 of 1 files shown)


docker/hail-ubuntu/Dockerfile
```diff
@@ -1,18 +1,25 @@
...
-    update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.7 1 && \
-    python3 -m pip check && \
-    python3 -m pip --version
 COPY pip.conf /root/.config/pip/pip.conf
 COPY hail-pip-install /bin/hail-pip-install
 COPY controller.sh /
+RUN chmod 755 /bin/retry && \
+    chmod 755 /bin/hail-apt-get-install && \
+    chmod 755 /bin/hail-pip-install && \
+    chmod 755 /controller.sh && \
```
daniel-goldstein, line 12: I only see `hail-apt-get-install` used (and transitively `retry`). Do we have to chmod `hail-pip-install` and `controller`?

daniel-goldstein (changes requested):

docker/hail-ubuntu/Dockerfile
danking: Descendant Dockerfiles (like service-base) use hail-pip-install. I can remove the chmod's. I'm parceling out a bunch of changes from my full set of changes. Once we move to GCS we lose permissions. I'll remove the chmod's for now.

daniel-goldstein (approved):

PR #10357, opened 2021-04-21
[query] fix annotation db test

danking:
The test was lacking asserts, so it didn't fail even though it wasn't right. The tables were missing keys.


pwc2 (approved):
looks good, nice catch

PR #10358, opened 2021-04-21
[datasets] [annotationdb] add column for cloud/regions to tables in docs

pwc2:
Existing tables and matrix tables that we don't own/didn't create but are accessible via the datasets API are not always available on both GCS and S3 (e.g. pan UKB datasets are only on AWS S3), or are not available in an `eu` bucket on GCS (e.g. gnomAD datasets). 

This just adds a column `cloud: [regions]` to the tables on the datasets API and annotation DB docs pages to be more explicit about what datasets/versions are available on which cloud platform. 

So, for example, if a version of a dataset is in `gs://hail-datasets-us`, `gs://hail-datasets-eu`, and `s3://hail-datasets-us-east-1`, then under `cloud: [regions]` we would see `gcp: [eu,us], aws: [us]`.

### DIFF
--- a/hail/python/hail/docs/_static/annotationdb/annotationdb.html
+++ b/hail/python/hail/docs/_static/annotationdb/annotationdb.html
@@ -45,6 +45,7 @@ mt = db.annotate_rows_db(mt)
       <th>description</th>
       <th>version</th>
       <th>reference genome</th>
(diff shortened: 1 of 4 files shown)


danking (approved):
Excellent addition, thanks!

PR #10366, opened 2021-04-23
[hailgenetics/hail] add dill to the image

danking:
This allows this to be used with PythonJob.


jigold (dismissed):
Why are we removing pandas? That seems useful.

danking:
It's already a hail dependency.

jigold (approved):

PR #10367, opened 2021-04-23
[batch] avoid unnecessary locks

danking:
This PR is important for making the query service tests run in a reasonable amount of time.

1. msec_mcpu is not used any more, so don't bother. This causes a conflict with *every other job in
   the batch*, effectively serializing changes to every attempt in the same batch.

2. There is no need to eagerly take update lock on the instances table in add_attempt. A share lock
   prevents other rows from mutating that value while we're looking at it. We do not care if they
   read an old value for free_cores_mcpu between us reading `state` and us updating
   `free_cores_mcpu`.

3. In `schedule_job` there is no need to take an update lock on the `batches` table (which
   effectively serializes all scheduling on the same batch).

4. In `unschedule_job`, I made the formatting match other selects.

5. In `unschedule_job`, same thing about update locks on the `instances` table.

6. In `mark_job_creating`, same thing about update locks on the `batches` table.

7. In `mark_job_started`, same thing about update locks on the `batches` table.

8. In `mark_job_complete`, same thing about update locks on the `instances` table.


batch/sql/estimated-current.sql
```diff
@@ -360,10 +360,6 @@ BEGIN
 
   SET msec_mcpu_diff = msec_diff * job_cores_mcpu;
 
-  UPDATE batches
-  SET msec_mcpu = batches.msec_mcpu + msec_mcpu_diff
-  WHERE id = NEW.batch_id;
-
   UPDATE jobs
```
jigold, line 363: Should we just delete the update from jobs as well?

batch/sql/do-not-lock-entire-batch.sql
```diff
@@ -0,0 +1,446 @@
...
+  SELECT state INTO cur_instance_state FROM instances WHERE name = in_instance_name LOCK IN SHARE MODE;
+
+  IF cur_job_state = 'Ready' AND NOT cur_job_cancel AND cur_instance_state = 'active' THEN
+    UPDATE jobs SET state = 'Running', attempt_id = in_attempt_id WHERE batch_id = in_batch_id AND job_id = in_job_id;
+  END IF;
+
+  COMMIT;
+  SELECT 0 as rc, delta_cores_mcpu;
+END $$
+DROP PROCEDURE IF EXISTS mark_job_complete $$
```
jigold, line 291: Can you add a new line here?

batch/sql/do-not-lock-entire-batch.sql
```diff
@@ -0,0 +1,446 @@
+DELIMITER $$
+
+DROP TRIGGER IF EXISTS attempts_after_update $$
```
jigold, line 3: You have `attempts_after_update` twice in this file

jigold (changes requested):

jigold:
Good catches!

batch/sql/estimated-current.sql
danking: done

batch/test/test_scale.py
```diff
@@ -16,24 +14,15 @@ def client():
...
 
 def test_scale(client):
-    now = time.time()
-    n_batches = 100
-    batches = []
-    for _ in range(n_batches):
-        n_jobs = 100
-        batch = client.create_batch()
-        for idx in range(n_jobs):
-            sleep_time = random.uniform(0, 30)
```
jigold: Let's just delete this file. It's not used and I can write my own script if needed.

jigold (dismissed):

batch/test/test_scale.py
jigold: If you do that, then you can delete the exclusions from all of `test_batch_*` in build.yaml

danking: I reverted this, let's leave this change for a different PR.

jigold (approved):

jigold:
Approved. Merge at your convenience. There's a WIP tag.

PR #10368, opened 2021-04-23
[batch] Make initializing GCS lazy when using PythonJob

jigold:
See https://hail.zulipchat.com/#narrow/stream/223457-Batch-support/topic/GOOGLE_APPLICATION_CREDENTIALS.20issue/near/235884329

This way we maintain some backwards compatibility where project doesn't have to be defined when using the Batch constructor unless you use the new functionality.

### DIFF
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -1,5 +1,6 @@
 import re
 import dill
+import concurrent
 import os
 import functools
 from io import BytesIO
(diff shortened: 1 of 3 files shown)


hail/python/hailtop/batch/job.py
```diff
@@ -909,6 +912,10 @@ def wrapped(*args, **kwargs):
             job_path = os.path.dirname(result._get_path(remote_tmpdir))
             code_path = f'{job_path}/code{i}.p'
 
+            if self._batch._gcs is None:
+                self._batch._gcs = GCS(blocking_pool=concurrent.futures.ThreadPoolExecutor(),
+                                       project=self._batch._project)
```
danking, line 917: This feels like unnecessary deep entanglement between these classes.

Let's make `_gcs` a lazy field:

```python3
    self.__gcs: Optional[GCS] = None

    @property
    def _gcs(self):
        if self.__gcs is None:
            self.__gcs = GCS(...)
        return self.__gcs
```

danking (changes requested):

danking (approved):

PR #10373, opened 2021-04-26
[batch] Fix Python job with local backend

jigold:


daniel-goldstein (changes requested):
I feel like this could be cleanest as a `Backend.write_file` method, would that make sense?

jigold:
I tried your suggestion and I don't think this is the right solution for now and made things more complicated. When we add Cotton's file system library to hailtop.batch, then we'll get a file system that abstracts this away.

daniel-goldstein (approved):
Sounds good, thanks for the explanation

PR #10376, opened 2021-04-26
[batch] Run job containers with crun

daniel-goldstein:
This is the first step to removing batch workers' reliance on the docker daemon and docker in general, in favor of a lower level of abstraction that gives us finer control over resources on the worker like overlays and network namespaces, allowing us to shortcut and pre-configure some of the overhead that goes into running a job.

## What this does differently
Currently, the high-level process for running a job involves communicating with the docker daemon to:
1. Pull an image for a job
2. Start a container from that image
3. Run the container
4. Delete the container and its associated resources

We offload some of these responsibilities into the worker code and onto [crun](https://github.com/containers/crun), a lightweight low-level runtime with the same API as `runc`, what docker uses to run containers. Once docker has retrieved an image, if we see that the pulled image has a new digest from one we currently have cached on the worker, we extract the image's filesystem into a directory on the worker's disk. We then:

- use `mount` to create an overlay on top of the image that the container will use as its rootfs
- use `xfs_quota` to limit the container's storage in the overlay
- invoke `crun` to run a container with the overlay as its root filesystem and an appropriate network namespace that we set up at worker-start time.

Since we control the overlay, we can set the XFS quota before creating the container. So what was separate calls to docker create/start/run/delete is just a single `crun run`. Fewer steps, less back and forth with a single daemon, and pre-configuring the networks gives some sizable performance gains reliable, as well as reliable and consistent performance.

## What this doesn't solve
- Docker is still running the worker container. I don't see any real challenge to this it's just a matter of translating the docker parameters
- Still using docker to pull images and extract filesystems / environment variables from them. I don't have a substitute for this at the moment.

### DIFF
--- a/batch/Dockerfile.worker
+++ b/batch/Dockerfile.worker
@@ -1,4 +1,4 @@
-FROM {{ global.docker_prefix }}/python:3.7-slim-stretch
+FROM {{ hail_ubuntu_image.image }} AS base
 
 COPY docker/hail-ubuntu/retry /bin/retry
 COPY docker/hail-ubuntu/hail-apt-get-install /bin/hail-apt-get-install
@@ -6,6 +6,8 @@ RUN chmod 755 /bin/retry && \
     chmod 755 /bin/hail-apt-get-install && \
     mkdir -p /usr/share/man/man1 /usr/share/man/man2
 RUN hail-apt-get-install \
+    iproute2 \
+    iptables \
     openjdk-8-jre-headless \
     liblapack3
 
@@ -15,7 +17,7 @@ COPY docker/requirements.txt .
 RUN chmod 755 /bin/hail-pip-install && \
     hail-pip-install -r requirements.txt pyspark==3.1.1
 
-ENV SPARK_HOME /usr/local/lib/python3.7/site-packages/pyspark
+ENV SPARK_HOME /usr/local/lib/python3.7/dist-packages/pyspark
 ENV PATH "$PATH:$SPARK_HOME/sbin:$SPARK_HOME/bin"
 ENV PYSPARK_PYTHON python3
 
@@ -30,6 +32,23 @@ RUN echo "APT::Acquire::Retries \"5\";" > /etc/apt/apt.conf.d/80-retries && \
     curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add - && \
     hail-apt-get-install fuse gcsfuse
 
+# Install crun runtime dependencies
+RUN hail-apt-get-install libyajl-dev
+
+# Build crun in separate build step
+FROM base AS crun_builder
+RUN hail-apt-get-install make git gcc build-essential pkgconf libtool \
+   libsystemd-dev libcap-dev libseccomp-dev \
+   go-md2man libtool autoconf automake
+RUN git clone --depth 1 --branch 0.19.1 https://github.com/containers/crun.git && \
+   cd crun && \
+   ./autogen.sh && \
+   ./configure && \
+   make && \
+   make install
+
+FROM base
+COPY --from=crun_builder /usr/local/bin/crun /usr/local/bin/crun
 RUN python3 -m pip install --upgrade --no-cache-dir pip
 
 COPY hail/python/setup-hailtop.py /hailtop/setup.py
--- a/batch/batch/driver/create_instance.py
+++ b/batch/batch/driver/create_instance.py
@@ -145,7 +145,7 @@ sudo mkdir -p /etc/netns
 iptables --table nat --append POSTROUTING --source 10.0.0.0/15 --jump MASQUERADE
 
 # [public] Block public traffic to the metadata server
-iptables --table net --append PREROUTING --source 10.1.0.0/16 --destination 169.254.169.254 --jump DROP
+iptables --table net --append OUTPUT --source 10.1.0.0/16 --destination 169.254.169.254 --jump DROP
 
 CORES=$(nproc)
 NAMESPACE=$(curl -s -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/attributes/namespace")
--- a/batch/batch/driver/create_instance.py
+++ b/batch/batch/driver/create_instance.py
@@ -145,7 +145,7 @@ sudo mkdir -p /etc/netns
 iptables --table nat --append POSTROUTING --source 10.0.0.0/15 --jump MASQUERADE
 
 # [public] Block public traffic to the metadata server
-iptables --table net --append OUTPUT --source 10.1.0.0/16 --destination 169.254.169.254 --jump DROP
+iptables --insert FORWARD --source 10.1.0.0/16 --destination 169.254.169.254 --jump DROP
 
 CORES=$(nproc)
 NAMESPACE=$(curl -s -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/attributes/namespace")
--- a/batch/batch/driver/create_instance.py
+++ b/batch/batch/driver/create_instance.py
@@ -111,30 +111,6 @@ nohup /bin/bash run.sh >run.log 2>&1 &
 #!/bin/bash
 set -x
 
-# set up docker networks
-docker network create public --opt com.docker.network.bridge.name=public
-docker network create private --opt com.docker.network.bridge.name=private
-docker network create batch-worker --opt com.docker.network.bridge.name=batch-worker
-
-# wait for docker to create the DOCKER-USER chain
-while ! iptables -L DOCKER-USER ; do sleep 1; done
-
-# [all docker] ban metadata server
-iptables -I DOCKER-USER           -d 169.254.169.254 -j DROP
-
-# [all docker] override: allow udp/53 (dns) to metadata server
-iptables -I DOCKER-USER           -d 169.254.169.254 -p udp -m udp --destination-port 53 -j ACCEPT
-
-# [public docker] ban inter-container communication
-iptables -I DOCKER-USER -i public -d 172.16.0.0/12   -j DROP
-
-# [public docker] ban unused ip address range
-iptables -I DOCKER-USER -i public -d 192.168.0.0/16  -j DROP
-
-# [batch worker] override: allow metadata server for batch worker
-iptables -I DOCKER-USER -i batch-worker -d 169.254.169.254 -j ACCEPT
-
-
 WORKER_DATA_DISK_NAME="{worker_data_disk_name}"
 UNRESERVED_WORKER_DATA_DISK_SIZE_GB="{unreserved_disk_storage_gb}"
 
@@ -161,7 +137,15 @@ sudo ln -s /mnt/disks/$WORKER_DATA_DISK_NAME/logs /logs
 sudo mkdir -p /mnt/disks/$WORKER_DATA_DISK_NAME/gcsfuse/
 sudo ln -s /mnt/disks/$WORKER_DATA_DISK_NAME/gcsfuse /gcsfuse
 
-export HOME=/root
+sudo mkdir -p /etc/netns
+
+# private job network = 10.0.0.0/16
+# public job network = 10.1.0.0/16
+# [all networks] Rewrite traffic coming from containers to masquerade as the host
+iptables --table nat --append POSTROUTING --source 10.0.0.0/15 --jump MASQUERADE
+
+# [public] Block public traffic to the metadata server
+iptables --table net --append PREROUTING --source 10.1.0.0/16 --destination 169.254.169.254 --jump DROP
 
 CORES=$(nproc)
 NAMESPACE=$(curl -s -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/attributes/namespace")
@@ -264,11 +248,14 @@ docker run \
 -e BATCH_WORKER_IMAGE=$BATCH_WORKER_IMAGE \
 -e UNRESERVED_WORKER_DATA_DISK_SIZE_GB=$UNRESERVED_WORKER_DATA_DISK_SIZE_GB \
 -v /var/run/docker.sock:/var/run/docker.sock \
+-v /var/run/netns:/var/run/netns:shared \
 -v /usr/bin/docker:/usr/bin/docker \
 -v /usr/sbin/xfs_quota:/usr/sbin/xfs_quota \
 -v /batch:/batch:shared \
 -v /logs:/logs \
 -v /gcsfuse:/gcsfuse:shared \
+-v /etc/netns:/etc/netns \
+-v /sys/fs/cgroup:/sys/fs/cgroup \
 --mount type=bind,source=/mnt/disks/$WORKER_DATA_DISK_NAME,target=/host \
 --mount type=bind,source=/dev,target=/dev,bind-propagation=rshared \
 -p 5000:5000 \
@@ -278,7 +265,7 @@ docker run \
 --privileged \
 --cap-add SYS_ADMIN \
 --security-opt apparmor:unconfined \
---network batch-worker \
+--network host \
 $BATCH_WORKER_IMAGE \
 python3 -u -m batch.worker.worker >worker.log 2>&1
 
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -858,7 +858,7 @@ class Container:
 
     async def pipe_to_log(self, strm: Optional[asyncio.StreamReader]):
         if strm is not None:
-            while not strm.at_eof():
+            while not strm.at_eof() and not strm.exception():
                 self.logbuffer.extend(await strm.readline())
 
     def __str__(self):
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -148,10 +148,11 @@ class PortAllocator:
 
 
 class NetworkNamespace:
-    def __init__(self, subnet_index: int, private: bool):
+    def __init__(self, subnet_index: int, private: bool, internet_interface: str):
         assert subnet_index <= 255
         self.subnet_index = subnet_index
         self.private = private
+        self.internet_interface = internet_interface
         self.network_ns_name = uuid.uuid4().hex[:5]
         self.hostname = uuid.uuid4().hex[:10]
         self.veth_host = self.network_ns_name + '-host'
@@ -203,8 +204,8 @@ ip -n {self.network_ns_name} route add default via {self.host_ip}'''
     async def enable_iptables_forwarding(self):
         await check_shell(
             f'''
-iptables -w 10 --append FORWARD --in-interface {self.veth_host} --out-interface ens5 --jump ACCEPT && \
-iptables -w 10 --append FORWARD --out-interface {self.veth_host} --in-interface ens5 --jump ACCEPT'''
+iptables -w 10 --append FORWARD --in-interface {self.veth_host} --out-interface {self.internet_interface} --jump ACCEPT && \
+iptables -w 10 --append FORWARD --out-interface {self.veth_host} --in-interface {self.internet_interface} --jump ACCEPT'''
         )
 
     async def expose_port(self, port, host_port):
@@ -242,16 +243,21 @@ class NetworkAllocator:
         self.private_networks = asyncio.Queue()
         self.public_networks = asyncio.Queue()
 
-        if 'ens5' not in psutil.net_if_addrs().keys():
-            raise Exception('No ens5 interface detected')
+        for nic in psutil.net_if_addrs().keys():
+            if nic.startswith('ens'):
+                self.internet_interface = nic
+                break
+        else:
+            raise Exception('No ens interface detected')
 
     async def reserve(self, netns_pool_min_size: int = 64):
         for subnet_index in range(netns_pool_min_size):
-            public = NetworkNamespace(subnet_index, private=False)
+            public = NetworkNamespace(subnet_index, private=False, internet_interface=self.internet_interface)
             await public.init()
             self.public_networks.put_nowait(public)
 
-            private = NetworkNamespace(subnet_index, private=True)
+            private = NetworkNamespace(subnet_index, private=True, internet_interface=self.internet_interface)
+
             await private.init()
             self.private_networks.put_nowait(private)
 
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -178,12 +178,12 @@ class NetworkNamespace:
         # Jobs on the private network should have access to the metadata server
         # and our vdc. The public network should not so we use google's public
         # resolver.
-        with open(f'/etc/netns/{self.network_ns_name}/resolv.conf', 'w') as hosts:
+        with open(f'/etc/netns/{self.network_ns_name}/resolv.conf', 'w') as resolv:
             if self.private:
-                hosts.write('nameserver 169.254.169.254\n')
-                hosts.write('search c.hail-vdc.internal google.internal\n')
+                resolv.write('nameserver 169.254.169.254\n')
+                resolv.write('search c.hail-vdc.internal google.internal\n')
             else:
-                hosts.write('nameserver 8.8.8.8\n')
+                resolv.write('nameserver 8.8.8.8\n')
 
     async def create_netns(self):
         await check_shell(
@@ -502,7 +502,7 @@ class Container:
                 self.short_error = 'image not found'
             raise
 
-        image_config, _ = await check_exec_output('docker', 'inspect', shq(self.image_ref_str))
+        image_config, _ = await check_exec_output('docker', 'inspect', self.image_ref_str)
         image_configs[self.image_ref_str] = json.loads(image_config)[0]
 
     async def ensure_image_is_pulled(self, auth=None):
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -230,8 +230,8 @@ iptables -w 10 --append FORWARD --out-interface {self.veth_host} --jump ACCEPT''
         self.port = None
         await check_shell(
             f'''
-ip netns delete {self.network_ns_name} && \
-ip link delete {self.veth_host}'''
+ip link delete {self.veth_host} && \
+ip netns delete {self.network_ns_name}'''
         )
         await self.create_netns()
 
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1,4 +1,4 @@
-from typing import Optional, Dict, Callable, Tuple
+from typing import Optional, Dict, Callable, Tuple, Awaitable, Any
 import os
 import json
 import sys
@@ -356,6 +356,7 @@ class Container:
         self.job = job
         self.name = name
         self.spec = spec
+        self.deleted_event = asyncio.Event()
 
         image_ref = parse_docker_image_reference(self.spec['image'])
         if image_ref.tag is None and image_ref.digest is None:
@@ -402,7 +403,8 @@ class Container:
 
     async def run(self, worker):
         try:
-            with self.step('pulling'):
+
+            async def localize_rootfs():
                 async with worker.rootfs_locks[self.image_ref_str]:
                     last_fetched = worker.rootfs_last_fetched.get(self.image_ref_str)
                     five_minutes_ago = time_msecs() - 5 * 60 * 1000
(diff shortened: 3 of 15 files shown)


batch/batch/worker/worker.py
```diff
@@ -138,6 +141,78 @@ def free(self, port):
...
+    ip -n {self.network_name} link set dev lo up && \
+    ip -n {self.network_name} address add {self.job_ip}/24 dev {veth_job} && \
+    ip -n {self.network_name} route add default via {self.host_ip}'''
+        )
+
+        await check_shell(
+            f'''
+iptables -w 10 --append FORWARD --in-interface {veth_host} --jump ACCEPT && \
+    iptables -w 10 --append FORWARD --out-interface {veth_host} --jump ACCEPT && \
+    iptables -w 10 --table nat --append POSTROUTING --source {self.job_ip}/24 --jump MASQUERADE'''
```
daniel-goldstein, line 172: we could probably only make this last iptables call once with `10.0.0.0/16`

batch/batch/worker/worker.py
```diff
@@ -138,6 +141,78 @@ def free(self, port):
...
+        )
+
+        # The default resolver in `/etc/resolv.conf` is a local address (127.0.0.53) which is
+        # inaccessible in other net namespaces since they have their own loop device. So we use
+        # google's public resolver.
+        os.makedirs(f'/etc/netns/{self.network_name}')
+        await check_shell(f'echo nameserver 8.8.8.8 | tee /etc/netns/{self.network_name}/resolv.conf')
```
daniel-goldstein, line 179: This can probably just be done once since we bind mount it to /etc/resolv.conf in the container. This is the "correct" way to set up a resolv.conf for a network namespace but the container might not care about it

batch/batch/worker/worker.py
```diff
@@ -138,6 +141,78 @@ def free(self, port):
...
+        self.port = None
+        self.host_port = None
+
+    async def init(self):
+        veth_host = self.network_name + '-host'
+        veth_job = self.network_name + '-job'
+        await check_shell(
+            f'''
+ip netns add {self.network_name} && \
+    ip link add name {veth_host} type veth peer name {veth_job} && \
```
danking, line 158: I'm inclined to have this all left-aligned.

batch/batch/worker/worker.py
```diff
@@ -138,6 +141,78 @@ def free(self, port):
...
+        pass
+
+
+class NetworkAllocator:
+    def __init__(self):
+        self.netns = asyncio.Queue()
+
+    async def reserve(self, netns_pool_min_size: int = 64):
+        for subnet_index in range(10, netns_pool_min_size):
+            netns = NetworkNamespace(subnet_index)
```
danking, line 204: I'm curious to know what worker start-up time is in this PR.

danking (changes requested):
A few thoughts already, this is looking great.

As a separate PR, we can probably implement a OCI Repository REST client in aiohttp. It seems like all the standard tools are written in Go, unfortunately. I'm inclined to avoid shelling out to a go program if it's not hard to just hit the REST API.

batch/batch/worker/worker.py
daniel-goldstein: The reason I write to `/etc/netns/.../resolv.conf` is because of the way network namespaces work on their own. Inside a network namespace, processes will look to `/etc/netns/<namespace_name>/resolv.conf` for a resolver IP, and fall back to `/etc/resolv.conf` if one doesn't exist.

Say we had a namespace `mynet`, we can run ping within that network namespace by `ip netns exec mynet ping google.com`. ping looks for `/etc/netns/mynet/resolv.conf`, can't find it, and falls back to `/etc/resolv.conf` which, since we're only in a separate network namespace, is the same as that on the host. But! Each network namespace has its own loopback interface, and since our host's `/etc/resolv.conf` contains the entry `nameserver 127.0.0.53`, processes in the `mynet` namespace won't see it. The "solution" to this, per the internet, is to use a public resolver and put it in `/etc/netns/mynet/resolv.conf`. I suppose we could do some `iptables` trickery to redirect some made up IP to the resolver on the host, but I'm unsure if that's necessary or wanted.

To actually answer your question, this is great for network namespaces but our containers have their own mount namespace, so they can't see `/etc/netns/mynet/resolv.conf` on the host. But on the flip side they have their own `/etc/resolv.conf` that we can mess with without worry, so the better approach here is just to write our a single `/container_resolv.conf` on the host with the public resolver, and then mount that to `/etc/resolv.conf` in all our containers.

batch/batch/worker/worker.py
daniel-goldstein: Using a bridge might also be the way to go here. The `iptables` rules apply to just the bridge, we give it an IP address range, etc. New namespaces would just add one end of their veth-pair to the bridge. But we would have to lock it down so they can't talk to each other..

batch/batch/worker/worker.py
daniel-goldstein: Ultimately, I ended up writing separate files to `/etc/netns/.../` anyway because it makes debugging network problems easier. This way you can just ssh into a worker and run `sudo ip netns exec <ns_name> …`

daniel-goldstein:
@danking bump

batch/batch/worker/worker.py
```diff
@@ -419,135 +489,232 @@ async def batch_worker_access_token(self):
...
+                ],
+                'uidMappings': [],  # Can control host -> container uid/gid mappings here
+                'gidMappings': [],
+                'resources': {
+                    'cpu': {'shares': weight},
+                    'memory': {
+                        'limit': self.spec['memory'],
+                        'reservation': self.spec['memory'],
+                    },
+                    # 'blockIO': {'weight': min(weight, 1000)}, TODO Got permission denied
```
daniel-goldstein, line 607: Interestingly, linux 5.0 removed `blkio.weight`/`blkio.weight_device` and in our production logs right now you see a lot of
`Your kernel does not support cgroup blkio weight` and `Your kernel does not support cgroup blkio weight_device` warnings from `dockerd`, so I'm inclined to leave this as a `FIXME` and come back to it in a separate PR.

danking:
Sorry, I will take a look tomorrow.

danking:
actually looking now

batch/batch/worker/worker.py
```diff
@@ -425,138 +490,271 @@ async def batch_worker_access_token(self):
...
+                ],
+                'uidMappings': [],  # Can control host -> container uid/gid mappings here
+                'gidMappings': [],
+                'resources': {
+                    'cpu': {'shares': weight},
+                    'memory': {
+                        'limit': self.spec['memory'],
+                        'reservation': self.spec['memory'],
+                    },
+                    # 'blockIO': {'weight': min(weight, 1000)}, FIXME blkio.weight not supported
```
danking, line 622: OK, so, let's pin down exactly the issues here.

`blkio.weight` was deprecated and then removed in Liinux 5.0. The cgroups man page (what a beauty of
a man page) notes this:

> Although cgroups v2 is intended as a replacement for cgroups v1, the older system continues to
> exist (and for compatibility reasons is unlikely to be removed).  Currently, cgroups v2 implements
> only a subset of the controllers available in cgroups v1.  The two systems are implemented so that
> both v1 controllers and v2 controllers can be mounted on the same system.  Thus, for example, it
> is possible to use those controllers that are supported under version 2, while also using version
> 1 controllers where version 2 does not yet support those controllers.  The only restriction here
> is that a controller can't be simultaneously employed in both a cgroups v1 hierarchy and in the
> cgroups v2 hierarchy.

[cgroups(7)](https://man7.org/linux/man-pages/man7/cgroups.7.html)

Supposedly, the new `io` cgroup still supports `weight`:

> blkio.weight was removed in kernel 5.0: torvalds/linux@f382fb0
>
> cgroup v2 still has io.weight which can be converted from blkio.weight

[AkihiroSuda](https://github.com/docker/cli/pull/2908#issuecomment-755867349)

It appears that batch workers might support cgroup2:

```
dking@batch-worker-default-standard-n2-d7sea:~$ mount | grep cgroup
...
cgroup2 on /sys/fs/cgroup/unified type cgroup2 (rw,nosuid,nodev,noexec,relatime,nsdelegate)
...
```

Clearly some more work is needed to understand this, but I'll leave this comment here as
documentation about what we need to do next.

batch/batch/worker/worker.py
```diff
@@ -425,138 +490,271 @@ async def batch_worker_access_token(self):
...
-        )
-        self.log = "".join(logs)
-        return self.log
+        with open(f'{self.rootfs_path}/etc/subgid', 'r') as subgid:
+            for record in subgid:
+                if record.startswith(user):
+                    gid = record.split(":")[1]
+                    break
+            else:
+                raise ValueError("Container user not found in image's /etc/subgid")
```
danking, line 681: container group not found, right?

batch/batch/worker/worker.py
```diff
@@ -425,138 +490,271 @@ async def batch_worker_access_token(self):
...
+    async def _get_in_container_user(self):
+        user = self.image_config['Config']['User']
+        if not user:
+            uid, gid = 0, 0
+        elif ":" in user:
+            uid, gid = user.split(":")
+        else:
+            uid, gid = self._read_user_from_rootfs(user)
+        if uid != 0 or gid != 0:
+            await check_shell(f'chown -R {uid}:{gid} {self.rootfs_path}/home/{user}')
```
danking, line 663: Why is this necessary? Does the image transfer process not preserve permissions?

batch/batch/worker/worker.py
```diff
@@ -425,138 +490,271 @@ async def batch_worker_access_token(self):
...
+            'root': {
+                'path': '.',
+                'readonly': False,
+            },
+            'hostname': self.netns.hostname,
+            'mounts': self._mounts(),
+            'process': {
+                'user': {  # uid/gid *inside the container*
+                    'uid': int(uid),
+                    'gid': int(gid),
```
danking: Note to ourselves: user namespaces needs to happen before we get too popular.

batch/batch/worker/worker.py
```diff
@@ -425,138 +490,271 @@ async def batch_worker_access_token(self):
...
 
         if self.host_port is not None:
             port_allocator.free(self.host_port)
             self.host_port = None
 
+        if self.netns:
+            network_allocator.free(self.netns)
+            self.netns = None
+
+        # FIXME I am not sure why this does not always exist
```
danking, line 764: I'm disinclined to leave FIXMEs in the code. Can we get to the bottom of this?

batch/batch/worker/worker.py
```diff
@@ -567,30 +765,68 @@ async def delete(self):
     #   timing: dict(str, float),
     #   error: str, (optional)
     #   short_error: str, (optional)
-    #   container_status: { (from docker container state)
+    #   container_status: {
     #     state: str,
     #     started_at: str, (date)
     #     finished_at: str, (date)
-    #     out_of_memory: bool
-    #     error: str, (one of error, exit_code will be present)
```
danking, line 575: also verify that error is not used anywhere

batch/batch/worker/worker.py
```diff
@@ -567,30 +765,68 @@ async def delete(self):
...
+        status['container_status'] = await self.get_container_status()
         if self.error:
             status['error'] = self.error
         if self.short_error:
             status['short_error'] = self.short_error
         if self.container_status:
             status['container_status'] = self.container_status
-        elif self.container:
+        elif self.container_is_running():
             status['container_status'] = await self.get_container_status()
```
danking, line 800: something seems wrong there, you call get_container_status twice

batch/batch/worker/worker.py
```diff
@@ -761,7 +997,12 @@ def __init__(
         self.main_volume_mounts = []
         self.output_volume_mounts = []
 
-        io_volume_mount = f'{self.io_host_path()}:/io'
+        io_volume_mount = {
+            'source': self.io_host_path(),
+            'destination': '/io',
+            'type': 'none',
+            'options': ['rbind', 'rw'],
+        }
```
danking, line 1018: Thought: now that we fully control the filesystem, can we rid ourselves of this duplicated /io and /? Can we place the writable layers wherever the io_host_path is?

batch/batch/worker/worker.py
```diff
@@ -859,12 +1107,27 @@ def __init__(
 
         requester_pays_project = job_spec.get('requester_pays_project')
 
+        self.timings = Timings(lambda: False)
+
+        # FIXME: Remove this, it probably does not work
         if job_spec['process'].get('mount_docker_socket'):
```
danking, line 1126: not necessary anymore, afaik

batch/batch/worker/worker.py
```diff
@@ -1316,6 +1588,8 @@ def __init__(self):
         self.task_manager = aiotools.BackgroundTaskManager()
         self.jar_download_locks = defaultdict(asyncio.Lock)
 
+        self.bundle_locks = defaultdict(asyncio.Lock)
```
danking, line 1604: bundle is a bit of a surprising name. Is that the OCI terminology for an image?

batch/test/test_batch.py
```diff
@@ -594,7 +592,7 @@ def test_verify_no_access_to_metadata_server(client):
     builder.submit()
     status = j.wait()
     assert status['state'] == 'Failed', str(status)
-    assert "Connection timed out" in j.log()['main'], str(j.log()['main'], status)
+    assert "Could not resolve host" in j.log()['main'], (str(j.log()['main']), status)
```
danking, line 595: big improvement.

danking (approved):

danking (changes requested):
xxx

batch/Dockerfile.worker
```diff
@@ -1,11 +1,13 @@
-FROM {{ global.docker_prefix }}/python:3.7-slim-stretch
```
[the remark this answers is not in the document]
daniel-goldstein: I had issues building on debian but I can try this again because this was at the very beginning

batch/batch/driver/create_instance.py
```diff
@@ -161,7 +137,18 @@ async def create_instance(
...
-export HOME=/root
+# FIXME Explicitly add the buildkit user as a workaround
+sudo useradd -u 1000 user
+
+sudo mkdir -p /etc/netns
+
+# private job network = 10.0.0.0/16
+# public job network = 10.1.0.0/16
+# [all networks] Rewrite traffic coming from containers to masquerade as the host
+iptables --table nat --append POSTROUTING --source 10.0.0.0/15 --jump MASQUERADE
```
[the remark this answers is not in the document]
daniel-goldstein: IIUC 15 allows the rule to cover both the 10.0.x.x and 10.1.x.x subnets (which is what we want)

batch/batch/worker/worker.py
```diff
@@ -137,6 +138,116 @@ def free(self, port):
...
+ip -n {self.network_name} link set dev {veth_job} up && \
+ip -n {self.network_name} link set dev lo up && \
+ip -n {self.network_name} address add {self.job_ip}/24 dev {veth_job} && \
+ip -n {self.network_name} route add default via {self.host_ip}'''
+        )
+
+        await check_shell(
+            f'''
+iptables -w 10 --append FORWARD --in-interface {veth_host} --jump ACCEPT && \
+iptables -w 10 --append FORWARD --out-interface {veth_host} --jump ACCEPT'''
```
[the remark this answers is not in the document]
daniel-goldstein: This is for communicating with the internet. We create `veth_host`, which is the end of the job's veth pair in the *host* network namespace, but there is no connection yet between the internet-enabled "physical" device on the host and this device, so this forwards traffic between the two. It would be better if we specified the physical device in these forwarding rules, and I think that would do a better job of ensuring job devices can't speak to each other, but I haven't looked into programmatically getting the right device name.

batch/batch/worker/worker.py
```diff
@@ -137,6 +138,116 @@ def free(self, port):
...
+
+
+class NetworkAllocator:
+    def __init__(self):
+        self.private_networks = asyncio.Queue()
+        self.public_networks = asyncio.Queue()
+
+    async def reserve(self, netns_pool_min_size: int = 64):
+        for subnet_index in range(netns_pool_min_size):
+            public = NetworkNamespace(subnet_index + 10, private=False)
```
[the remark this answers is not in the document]
daniel-goldstein: This is probably legacy from when I originally put the subnet index at the end and got yelled at for not leaving 10.0.0.0 open. I will see if I can remove it.

batch/batch/worker/worker.py
daniel-goldstein: yes, that's the last todo once buildkit is passing

batch/batch/worker/worker.py
daniel-goldstein: Oops I think the top one should be removed

batch/batch/worker/worker.py
daniel-goldstein: Yes we can and should do that. I'm not sure what can of worms that might open up, but we can put the writeable layer anywhere. Ideally a user's xfs project should just have one root directory.

batch/batch/worker/worker.py
daniel-goldstein: Missed this when renaming bundles to `rootfs`. In OCI jargon, they consider a bundle to be a directory of the rootfs and the container's config.json (i.e. everything you need to run a container). Thankfully, crun allows us to store the config.json separately from the rootfs, so there's not much need for the term "bundle".

batch/batch/worker/worker.py
daniel-goldstein: I got mixed up with which user buildkit was trying to use this directory for. You're right and this isn't actually necessary

batch/batch/worker/worker.py
```diff
@@ -567,30 +765,68 @@ async def delete(self):
     #   timing: dict(str, float),
     #   error: str, (optional)
     #   short_error: str, (optional)
-    #   container_status: { (from docker container state)
+    #   container_status: {
     #     state: str,
     #     started_at: str, (date)
     #     finished_at: str, (date)
-    #     out_of_memory: bool
```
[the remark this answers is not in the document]
daniel-goldstein: I'll just put it back in.

batch/batch/worker/worker.py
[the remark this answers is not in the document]
daniel-goldstein: yup removed

batch/batch/worker/worker.py
```diff
@@ -315,91 +358,117 @@ def __init__(self, job, name, spec):
...
+        self.netns: Optional[NetworkNamespace] = None
+        self.proc = None
 
-        env = self.spec.get('env', [])
+    async def run(self, worker):
+        try:
+            with self.step('pulling'):
+                async with worker.bundle_locks[self.image_ref_str]:
+                    self.image_config = await self.pull_image()
+                    image_digest = self.image_config['RepoDigests'][0].split('sha256:')[1]
```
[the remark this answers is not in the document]
daniel-goldstein: I thought we couldn't put `:`s in directory names. `split('@')[1].split(':')[1]`?

batch/batch/worker/worker.py
daniel-goldstein: oops I was double deleting the container. Fixed

batch/batch/worker/worker.py
daniel-goldstein: This whole `container_status['container_status']` and repeated fields are quite confusing imo. I think it was only used in one place in the batch client so I removed it (but left in the out_of_memory), and will try to reconcile the tests. Since the inner `container_status` was essentially passing through the docker status I would like to follow up by flattening the container status structure. I don't see a scenario where the duplicate fields have/should have different values.

batch/batch/worker/worker.py
daniel-goldstein: I moved the container's overlay and crun config into the job's scratch directory, which is great because I could then remove the extra call to `xfs_quota` to add the overlay to the xfs project. Here is basically the directory tree for a job:

```
scratch (actually /batch/token)/
├─ io/
├─ input/
│  ├─ rootfs_overlay/
│  │  ├─ upperdir/
│  │  │  ├─ io/ (bind mount)
│  │  ├─ merged/
│  │  ├─ workdir/
│  ├─ volumes/
│  ├─ config/
│  │  ├─ config.json
├─ main/
│  ├─ rootfs_overlay/
│  │  ├─ upperdir/
│  │  │  ├─ io/ (bind mount)
│  │  ├─ merged/
│  │  ├─ workdir/
│  ├─ volumes/
│  ├─ config/
│  │  ├─ config.json
├─ output/
…
```

where `scratch` (like before) is marked with the user's storage quota for that job. Note that this does not contain the `lowerdir` of the overlays, because those are shared, readonly extracted rootfs's that we share across containers/overlays.

daniel-goldstein: Perhaps it's not fair to make the user pay for the size of the config, and we could store that elsewhere, but it's a tiny json file…

daniel-goldstein:
I think that should cover everything, however I am still failing one test. I think the current docker approach has a convenient quirk of retrying the docker daemon, which allows container deletion to stop a container while it's running, which in turn allows the loop that checks the running container to `raise` and exit.

In the subprocess structure that I lifted from `JVMJob`, I'm unconvinced that this is happening correctly. What I thought would happen is that the `delete_container` coroutine would `kill` the container process while the coroutine running the container is still on `await self.process.wait()`. I would hope that means `self.process.wait` would soon raise but I'm now consistently seeing `test_cancel_left_after_tail` time out, indicating that it wasn't able to successfully cancel the sleeping job.

batch/Dockerfile.worker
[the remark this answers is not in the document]
danking: Is this still an issue?

batch/batch/worker/worker.py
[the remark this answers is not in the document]
danking: I just checked all the currently live batch instances and everyone has an ens5 and an lo. ens5 == PCI-Express slot 5. I'd bet Google's hypervisor always connects the ethernet device on slot 5. If this changes, we lose network connectivity, but its not a security risk.

batch/batch/worker/worker.py
```diff
@@ -137,6 +146,128 @@ def free(self, port):
...
+
+        os.makedirs(f'/etc/netns/{self.network_ns_name}')
+        with open(f'/etc/netns/{self.network_ns_name}/hosts', 'w') as hosts:
+            hosts.write('127.0.0.1 localhost\n')
+            hosts.write(f'{self.job_ip} {self.hostname}\n')
+
+        # Jobs on the private network should have access to the metadata server
+        # and our vdc. The public network should not so we use google's public
+        # resolver.
+        with open(f'/etc/netns/{self.network_ns_name}/resolv.conf', 'w') as hosts:
```
danking, line 181: hosts -> resolv

batch/batch/worker/worker.py
```diff
@@ -119,9 +121,16 @@
...
 port_allocator: Optional['PortAllocator'] = None
+network_allocator: Optional['NetworkAllocator'] = None
 
 worker: Optional['Worker'] = None
 
+image_configs: Dict[str, Dict[str, Any]] = dict()
+
+
+def get_image_digest(image_config) -> str:
+    return image_config['RepoDigests'][0].split('@')[1].split(':')[1]
```
danking, line 132: Colons are kosher: 

```
(base) # touch :
(base) # mkdir ::
(base) # ls -al
total 0
drwxr-xr-x   4 dking  wheel  128 Jun 25 18:11 .
drwxrwxrwt  25 root   wheel  800 Jun 25 18:11 ..
-rw-r--r--   1 dking  wheel    0 Jun 25 18:11 :
drwxr-xr-x   2 dking  wheel   64 Jun 25 18:11 ::
```

We use ext4 and XFS, but basically every file system we'd ever consider (including networked ones like Lustre) support everything but `/`s. https://en.wikipedia.org/wiki/Comparison_of_file_systems

That said, I'm fine with splitting on the `:`, it gets the "encoded" part of the digest [according to the OCI format](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests)

batch/batch/worker/worker.py
```diff
@@ -315,91 +379,131 @@ def __init__(self, job, name, spec):
...
-            'out_of_memory': cstate['OOMKilled'],
-        }
-        cerror = cstate['Error']
-        if cerror:
-            status['error'] = cerror
-        else:
-            status['exit_code'] = cstate['ExitCode']
-
-        return status
+        image_config, _ = await check_exec_output('docker', 'inspect', shq(self.image_ref_str))
```
danking, line 505: With exec, we actually *can't* `shq`. This converts `foo;bar` into `foo\;bar` or `'foo;bar'`
```
In [2]: shq('foo;bar')                                                                                                         
Out[2]: "'foo;bar'"
```
`exec` completely avoids the shell, it passes those literal, unadulterated bytes to the program. So, we just want:
```
await check_exec_output('docker', 'inspect', self.image_ref_str)
```

batch/batch/worker/worker.py
```diff
@@ -425,140 +525,292 @@ async def batch_worker_access_token(self):
...
-                    if e.status == 404:
-                        if 'pull access denied' in e.message:
-                            self.short_error = 'image cannot be pulled'
-                        elif 'not found: manifest unknown' in e.message:
-                            self.short_error = 'image not found'
-                    raise
+    async def extract_rootfs(self):
+        assert self.rootfs_path
+        os.makedirs(self.rootfs_path)
+        await check_shell(f'docker export $(docker create {shq(self.image_ref_str)}) | tar -C {self.rootfs_path} -xf -')
```
danking, line 535: Good, this is quoted properly.

batch/batch/worker/worker.py
danking: Can we add that directory tree to a dev-doc? This is a great visual.

danking: And just to be clear: with this change the entire disk quota should be available at any path, right? Can we add a couple tests to that effect? Try starting a 1 cpu job with 10GB and writing nearly 15GB to `/foo` and `/io/foo`.

danking: cc: @jigold

danking (changes requested):
This is a transformational PR. Great work Daniel, I'm excited to see this merge.

re:
> This whole container_status['container_status'] and repeated fields are quite confusing imo. I think it was only used in one place in the batch client so I removed it (but left in the out_of_memory), and will try to reconcile the tests. Since the inner container_status was essentially passing through the docker status I would like to follow up by flattening the container status structure. I don't see a scenario where the duplicate fields have/should have different values.

I'm totally down for simplifying the format in a future PR.

batch/batch/worker/worker.py
daniel-goldstein: Yes. Added a test for that and added a brief dev doc for a job's layout on the filesystem.

batch/Dockerfile.worker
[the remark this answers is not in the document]
daniel-goldstein: Ya I haven't been able to successfully build on Debian. Currently the worker image is 873MB and this branch increases it to 1.24GB.

batch/batch/worker/worker.py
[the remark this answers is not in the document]
daniel-goldstein: Should we just hard code `ens5` then? We can do a sanity check at worker startup time that `ens5` exists with `psutil`.

dev-docs/batch-design.md
```diff
@@ -0,0 +1,60 @@
...
+directory that is mounted into each container. Each container directory contains
+- The upper, merged, and work directories for the overlay filesystem used in the container. For
+    a great explanation of how overlayfs works, see
+    [here](https://jvns.ca/blog/2019/11/18/how-containers-work--overlayfs/).
+- Any volumes specified in the user's image that are mounted into the container
+- The container's `config.json` that the worker creates and passes to `crun`.
+
+Batch uses [xfs_quota](https://man7.org/linux/man-pages/man8/xfs_quota.8.html) to enforce storage
+limits for jobs. Each job receives its own XFS project rooted at the scratch directory, so all storage used by
+the main container, input/output containers and anything written to `/io` count toward the overall job quota.
```
jigold, line 24: How does this work for jobs with attached disks? Can we test this still works in the case where we need more storage?

jigold (changes requested):

dev-docs/batch-design.md
daniel-goldstein: This existing test should make sure that attaching a disk still works:
```python
```

danking (approved):
I'm good whenever Jackie is good.

daniel-goldstein:
Put WIP on so it doesn't merge before Jackie takes a look

jigold (approved):

daniel-goldstein:
Looks like this is destined for next week

PR #10383, opened 2021-04-27
[query] Enable NDArray based PCA to run on local and service backend

johnc1231:
Had to fix a bug where we were not binding global when lowering scans.


johnc1231:
Also deleted `test_blanczos_against_hail`, no reason to compare them, that test can never work on the service, and we want to delete the Spark one soon.

patrick-schultz (approved):

PR #10385, opened 2021-04-27
[query] more tcode removal

chrisvittal:


hail/src/main/scala/is/hail/expr/ir/functions/ArrayFunctions.scala
```diff
@@ -328,8 +329,8 @@ object ArrayFunctions extends RegistryFunctions {
             setup1,
             setup2),
           m1 || m2 || Code(
-            a1 := v1.tcode[Long],
-            a2 := v2.tcode[Long],
+            a1 := v1.asInstanceOf[SIndexablePointerCode].a,
```
tpoterba, line 332: This isn't correct -- this won't always be an SIndexablePointerCode. The right thing is to use the indexable SType interface here

hail/src/main/scala/is/hail/expr/ir/functions/LocusFunctions.scala
```diff
@@ -422,7 +422,7 @@ object LocusFunctions extends RegistryFunctions {
             val srcRG = iT.pointType.asInstanceOf[PLocus].rg
             val destRG = rt.types(0).asInstanceOf[PInterval].pointType.asInstanceOf[PLocus].rg
             val er = EmitRegion(cb.emb, r)
-            val intervalObj = Code.checkcast[Interval](asm4s.coerce[AnyRef](wrapArg(er, iT)(interval.tcode[Long])))
+            val intervalObj = Code.checkcast[Interval](asm4s.coerce[AnyRef](wrapArg(er, iT)(interval.asInstanceOf[SIntervalPointerCode].a)))
```
tpoterba, line 425: There's a `scodeToJavaValue` method in Functions.scala -- use that instead. This is wrong for the same reason as above.

tpoterba (changes requested):

hail/src/main/scala/is/hail/expr/ir/functions/ArrayFunctions.scala
chrisvittal: I'll fix this later then. It's a larger change involving more Code[_] -> CodeBuilder.

tpoterba (approved):

PR #10386, opened 2021-04-27
[query] Speed up table join tests by not doing full cartesian product…

tpoterba:
… of params

### DIFF
--- a/hail/src/test/scala/is/hail/expr/ir/TableIRSuite.scala
+++ b/hail/src/test/scala/is/hail/expr/ir/TableIRSuite.scala
@@ -430,15 +430,34 @@ class TableIRSuite extends HailSuite {
   )
 
   @DataProvider(name = "join")
-  def joinData(): Array[Array[Any]] =
+  def joinData(): Array[Array[Any]] = {
+    val defaultLParts = 2
+    val defaultRParts = 2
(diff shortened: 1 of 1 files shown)


hail/src/test/scala/is/hail/expr/ir/TableIRSuite.scala
```diff
@@ -430,15 +430,34 @@ class TableIRSuite extends HailSuite {
   )
 
   @DataProvider(name = "join")
-  def joinData(): Array[Array[Any]] =
+  def joinData(): Array[Array[Any]] = {
+    val defaultLParts = 2
+    val defaultRParts = 2
+    val (defaultJ, defaultP) = joinTypes(0)
```
johnc1231, line 436: Is this saying to only ever test outer joins?

johnc1231 (dismissed):

hail/src/test/scala/is/hail/expr/ir/TableIRSuite.scala
tpoterba: This change tests the full span of values within each parameter class (join types, for instance, but doesn't test against the full cartesian product.

You're right that it's probably worth doing the product over all join types, though. That should still be ~10x faster than the test was before.

johnc1231 (approved):

PR #10355, opened 2021-04-21
[ci] checkout repository once

danking:
stacked on https://github.com/hail-is/hail/pull/10354, will fail until that PR merges

PR #10389, opened 2021-04-27
[ci] use kaniko

danking:


danking:
Stacked on #10355

daniel-goldstein (approved):
I see what you mean about it being awkward putting python + kaniko together, but overall not too bad.

PR #10390, opened 2021-04-28

danking:
The high-level problem:

The high-level solution:

Hail Batch Workers can load the JAR for a given Hail version on-demand. Although not a long-term
solution, we currently start a fresh JVM for each job. As a result, we can simply start the JVM with
the correct JAR on its classpath. We cache jars on the local filesystem.

I had to abandon the old approach for two reasons:

1. Multiple JVMs race to download the JAR. In the new approach, the python worker process uses a
   lock to ensure at most one coroutine is downloading a given version of a JAR at the same time.

2. The JVM assumes that a child ClassLoader does not redefine a class from the parent
   ClassLoader. That's why ClassLoaders always prefer to load a class from the parent ClassLoader's
   classes.

When we decide to re-use JVMs or use a single multi-threaded JVM, we'll need to ensure the top-level
ClassLoader *does not have Hail on its classpath*. I looked briefly at this approach and found it
more work than the current approach.

---

My apologies for eliminating JVMProcess in this PR. It's an unrelated change which facilitated my
understanding worker.py. I essentially inlined JVMProcess into JVMJob and eliminated any duplicative
code.

---

After making this change I restored the tests. Some tests had bitrotted. In the process of fixing
those tests, I found a few other bugs. Fixing these lower-level bugs unlocked a number of new
tests.

A couple tests (which were added since the service tests were removed) had to be marked as
failing.

Here are the bugs I fixed:

1. Correct the error message raised when tests are run in a non-main thread (we look for this
   message and start an event loop for Hail's async code because asyncio refuses to start an event
   loop in a non-main thread).

2. Use a `SafeRow` to copy the globals data out of a Region and into durable, GC'ed objects.

3. Re-enable serialization of GoogleStorageFS (including its private key, which we really shouldn't
   do; Tim is working on it), which was broken (presumably) when we changed Scala versions. The
   `var` modifier ensures the name is compiled as a JVM field.

4. Correctly convert from a `Byte` to an `Int`. By default `Byte` to `Int` conversion (which is done
   automatically when you return a `Byte` from a function whose return type is `Int`) is
   sign-preserving. That means that the byte `0000 1111` is converted to the `Int` 15 and the byte
   `1000 1111` is converted to the `Int` -113. The contract of
   [`InputStream.read`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read--)
   is to return the unsigned integeral value of the next `Byte` or `-1` if we've reached the end of
   the stream. `DataInputStream` treats any negative value as EOS which lead to perplexing EOSes
   when reading data from GCS.

5. Retain the `gs://` protocol when reading MTs and Ts. `uriPath` strips *all* protocols. Before the
   Query Service, these code paths were only used by the LocalBackend. In the LocalBackend, the only
   URIs generated are `file://`. However, UNIX/JVM file system operations do not support URIs, they
   want bare paths.

6. Implement missing cases for EShuffle and PShuffle.

7. BLAS and LAPACK need to be thread-local.

8. The LSM-tree used by the Shuffler needs to permit multiple values for the same key. There was
   some subtlety here around fine-grained locking. Unfortunately, the LSM doesn't expose the right
   operations (putIfAbsent) to make this easy, so I had to use a concurrent hash map of locks.

---

---

Small changes and fixes:

  key.

  gs://hail-query`.

- Remove the auto-scaling and remove the k8s grace period. Neither of these really works right. We
  will live without the auto-scaling for now. I have to fix the grace period thing before we let
  users run real pipelines.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, Dict
 import os
 import json
 import sys
@@ -17,6 +17,7 @@ from aiohttp import web
 import async_timeout
 import concurrent
 import aiodocker
+from collections import defaultdict
 from aiodocker.exceptions import DockerError
 import google.oauth2.service_account
 from hailtop.utils import (time_msecs, request_retry_transient_errors, sleep_and_backoff,
@@ -27,6 +28,7 @@ from hailtop.httpx import client_session
 from hailtop.batch_client.parse import (parse_cpu_in_mcpu, parse_memory_in_bytes, parse_storage_in_bytes)
 from hailtop.batch.hail_genetics_images import HAIL_GENETICS_IMAGES
 from hailtop import aiotools
+from hailtop.aiotools.fs import RouterAsyncFS, LocalAsyncFS
 import hailtop.aiogoogle as aiogoogle
 
 # import uvloop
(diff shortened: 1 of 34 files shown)


danking:
@daniel-goldstein, if you could find some time to take a fresh look at this on Friday, I'd really appreciate that. I think this is finally ready.

hail/python/test/hail/methods/test_impex.py
```diff
@@ -58,7 +58,7 @@ def test_info_float64(self):
         for f in _FLOAT_ARRAY_INFO_FIELDS:
             self.assertEqual(mt['info'][f].dtype, hl.tarray(hl.tfloat64))
 
-    @fails_service_backend()
+    @skip_when_service_backend('Shuffler is encoding/decoding is broken.')
```
daniel-goldstein, line 61: extra `is`

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -638,8 +643,16 @@ object ServiceBackendMain {
   def main(argv: Array[String]): Unit = {
     assert(argv.length == 1, argv.toFastIndexedSeq)
     val udsAddress = argv(0)
+    val queryGCSPathEnvVar = System.getenv("HAIL_QUERY_GCS_PATH")
+    val queryGCSBucketEnvVar = System.getenv("HAIL_QUERY_GCS_BUCKET")
```
daniel-goldstein, line 647: Why not just have one env var?

hail/src/main/scala/is/hail/backend/service/Worker.scala
```diff
@@ -100,7 +99,7 @@ object Worker {
     timer.start("executeFunction")
 
     val hailContext = HailContext(
-      ServiceBackend(), skipLoggingConfiguration = true, quiet = true)
+      new ServiceBackend(null), skipLoggingConfiguration = true, quiet = true)
```
daniel-goldstein, line 102: Could you add a comment or fixme as to why there's this dummy backend on the worker? I realize it might not be easily fixable (I think you tried on an earlier revision) but it's odd-looking

build.yaml
```diff
@@ -2424,6 +2422,40 @@ steps:
       - deploy_address_sa
       - address_image
       - create_certs
+  - kind: runImage
+    name: upload_query_jar
```
daniel-goldstein, line 2426: Is there not a canonical place where this jar already sits? Not entirely clear to me yet why we need an extra bucket.

daniel-goldstein (dismissed):
Haven't looked yet at LSM and beyond, but I will look again later today. So far only a few nits, this looks much more straightforward!

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: fixed

hail/python/test/hail/methods/test_impex.py
danking: fixed

batch/batch/worker/worker.py
```diff
@@ -1193,10 +1141,58 @@ async def run(self, worker):
...
+                                while True:
+                                    b = await jar_data.read(256 * 1024)
+                                    if not b:
+                                        break
+                                    written = await local_file.write(b)
+                                    assert written == len(b)
+
+                self.timing['downloading_jar']['finish_time'] = time_msecs()
+                self.timing['downloading_jar']['duration'] = (
+                    self.timing['downloading_jar']['finish_time'] - self.timing['downloading_jar']['start_time'])
```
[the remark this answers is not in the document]
danking: done, I made some wider cleanup of that class, it was unnecessarily complicated and unnecessarily async

hail/src/main/scala/is/hail/backend/service/Worker.scala
danking: fixme added. I forget exactly why workers need a HailContext, but to have a HailContext one must have a backend.

hail/src/main/scala/is/hail/io/fs/GoogleStorageFS.scala
```diff
@@ -98,7 +98,7 @@ class GoogleStorageFileStatus(path: String, modificationTime: java.lang.Long, si
   def getOwner: String = null
 }
 
-class GoogleStorageFS(serviceAccountKey: String) extends FS {
+class GoogleStorageFS(var serviceAccountKey: String) extends FS {
```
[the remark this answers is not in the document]
danking: I think no, because of how serialization works? I guess I can try and we'll see if it passes the tests.

build.yaml
danking: `deploy_hail` uploads the JAR to `hail-common`, but that's only done on default deploy builds, not on dev or PR builds.

I'm generally inclined to not entangle the services team's infrastructure with compiler team's infrastructure.

The bucket is probably not necessary, but there wasn't another obvious place to put Hail Query Service stuff. Putting it in batch's bucket seems wrong.

daniel-goldstein: Sounds good, I'm fine with a separate bucket if it helps prevents awkward dependencies.

daniel-goldstein (approved):

danking:
rebased, and yet another transient error added (EAI_EAGAIN, temporary failure in name resolution).

PR #10391, opened 2021-04-28
[ci] replace BuildImageStep with BuildImage2Step

danking:
Once the merges, I can rename buildImage2 to buildImage everywhere and remove the buildImage2 case.


daniel-goldstein (approved):

danking:
This PR's CI can't seem to build base_image. It crashes with 137, which I suspect means it had insufficient storage or memory. I increased the cpu, memory, and storage requirements of base_image.

danking:
(for reference the base image itself is >2 GiB, so it seems not unreasonable that Kaniko needs 2x that in storage or memory to process everything).

PR #10393, opened 2021-04-28
[ci] building the kaniko image requires a lot of memory, apparently

danking:


danking:
override merging because CI is never gonna get past the kaniko image.

PR #10398, opened 2021-04-28
[query] Add method to go from BlockMatrix to NDArray

johnc1231:
I mostly want this for debugging lowering on `LocalBackend`, but I added a SparkBackend implementation as well (goes through breeze and java literals, not going to be super performant).

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Interpret.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Interpret.scala
@@ -787,6 +787,12 @@ object Interpret {
         function.execute(ctx, child.execute(ctx))
       case BlockMatrixToValueApply(child, function) =>
         function.execute(ctx, child.execute(ctx))
+      case BlockMatrixCollect(child) =>
+        val bm = child.execute(ctx)
+        val breezeMat = bm.toBreezeMatrix()
+        val shape = IndexedSeq(bm.nRows, bm.nCols)
+        // transpose because breeze toArray is column major
+        SafeNDArray(shape, breezeMat.t.toArray)
       case x@TableAggregate(child, query) =>
         val value = child.execute(ctx)
         val fsBc = ctx.fsBc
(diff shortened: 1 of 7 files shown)


johnc1231:
cc @pwc2 , might be of interest to you, though I think you're more interested in going from NDArrays to BlockMatrices.

johnc1231:
The tests are going to fail the local backend for now because they use methods like `from_numpy` which aren't lowered yet.

hail/src/main/scala/is/hail/expr/ir/Interpret.scala
```diff
@@ -787,6 +787,12 @@ object Interpret {
         function.execute(ctx, child.execute(ctx))
       case BlockMatrixToValueApply(child, function) =>
         function.execute(ctx, child.execute(ctx))
+      case BlockMatrixCollect(child) =>
+        val bm = child.execute(ctx)
+        val breezeMat = bm.toBreezeMatrix()
+        val shape = IndexedSeq(bm.nRows, bm.nCols)
+        // transpose because breeze toArray is column major
+        SafeNDArray(shape, breezeMat.t.toArray)
```
patrick-schultz, line 795: If there's no way to avoid an actual transpose, it would be better to do it before collecting to breeze, so you're only transposing the blocks.

patrick-schultz (dismissed):

hail/src/main/scala/is/hail/expr/ir/Interpret.scala
johnc1231: In breeze transpose is just a `isTransposed` flag, so this should just be telling it which way to collect.

patrick-schultz: Right, but then the collect is the transpose. You're collecting a column major matrix in row major order, which for larger matrices will have terrible cache behavior.

johnc1231: Made the change anyway.

patrick-schultz (approved):

PR #10399, opened 2021-04-28
[ci] allow kaniko to build the kaniko image

danking:
I do not know why but /etc triggers errors about:
```
archive/tar: write too long
```
Even though /etc is not very large (1.4MB). I suspect there is some symlink
or other nonsense which is breaking Kaniko.

The solution, after much trial and error, was simple: copy over directories that do not
cause issues and copy only the necessary files out of etc. A mix of speculation and
binary search lead me to the conclusion that /etc/ld.so.* are the only files necessary
from /etc for python to run correctly. These files tell the kernel how to link python3.7
to the various libraries on which it depends (which live in lib and lib64).

Anyway, I've tested that this image can build itself, so it should be good enough for
our purposes.

### DIFF
--- a/ci/kaniko/Dockerfile
+++ b/ci/kaniko/Dockerfile
@@ -1,7 +1,29 @@
 FROM python:3.7-slim-stretch
 RUN pip3 install jinja2
-COPY jinja2_render.py /jinja2_render.py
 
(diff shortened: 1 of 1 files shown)


danking:
CI is borked currently, so I'm force merging.

PR #10401, opened 2021-04-28
[query] remove tcode from Emit.scala

chrisvittal:


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -1148,7 +1149,8 @@ class Emit[C](
           val childShape = pndVal.shapes(cb)
           val childStrides = pndVal.strides(cb)
 
-          val dataArray = childPType.dataType.loadCheapPCode(cb, childPType.dataPArrayPointer(pndVal.tcode[Long]))
+          val pndAddr = SingleCodePCode.fromPCode(cb, pndVal, region)
+          val dataArray = childPType.dataType.loadCheapPCode(cb, childPType.dataPArrayPointer(pndAddr.code.asInstanceOf[Code[Long]]))
```
johnc1231, line 1153: Isn't doing `.code.asInstanceOf` effectively the same as doing `tcode`? What's the win here?

johnc1231 (changes requested):

hail/src/main/scala/is/hail/expr/ir/Emit.scala
chrisvittal: The win is in `SingleCodePCode` which ensures the PCode has only one code. `tcode` is going away from `PCode` in order to remove the assumption that one PCode is one Code[_].

chrisvittal: Basically `SingleCodePCode` is the One True Way to get a `Code[_]` out of a `PCode`

johnc1231 (approved):

PR #10411, opened 2021-04-29
[batch] update GPG key before apt-get update on build-worker instance

daniel-goldstein:
Tested this manually to make sure that the build-worker instance ran to completion (it fails on `apt-get update` currently)


batch/build-batch-worker-image-startup.sh
```diff
@@ -5,6 +5,10 @@ set -ex
 curl --silent --show-error --remote-name --fail https://dl.google.com/cloudagents/add-logging-agent-repo.sh
 bash add-logging-agent-repo.sh
 
+
+# Get the lates GPG key as it might not always be up to date
```
jigold, line 9: There's a typo in latest

jigold (changes requested):

jigold (approved):

PR #10416, opened 2021-04-30
[query] Fix some memory leaks in lowered execution

tpoterba:
* compiled iterators use the wrong region
* TableRange generates IR that does not flag itself as memory managed
* TableParallelize generates IR that does not flag itself as memory managed
* RepartitionNoShuffle generates a mixed flatMap; make it all memory managed for simplicity.


hail/src/main/scala/is/hail/expr/ir/GenericTableValue.scala
```diff
@@ -71,6 +71,7 @@ class PartitionIteratorLongReader(
 
   def toJValue: JValue = {
     JObject(
+      "category" -> JString("PartitionIteratorLongReader"),
```
tpoterba, line 74: not an important change, just makes it a little easier to diagnose what reader is being generated from IR printouts.

patrick-schultz (approved):

PR #10417, opened 2021-04-30
[batch] enable command-click on job id cell

danking:
An onclick handler does not correctly handle command+click. Instead, we
convert the anchor to a block element (allowing it to have width and
height independent of its contents) and set the width and height to
100% of its container.


danking:
@CDiaz96 bump

CDiaz96 (approved):
It would also be good to cross check that the class `fill-td` does not have overlapping styling that is being implemented elsewhere.

PR #10419, opened 2021-04-30
Release 0.2.66

jigold:
Stacked on #10397 and #10418


jigold:
cc: @johnc1231 @tpoterba

danking (changes requested):
You gotta update the Hail change log too, this is the only one I see:

#10398 CHANGELOG: Add BlockMatrix.to_ndarray

You can find this with `git log 0.2.65..hi/main`  and search for CHANGELOG. I also usually check for other commits that maybe *should* have a changelog entry.

jigold:
I was going to let John and Tim fill those in.

johnc1231:
I'll fill in no worries.

johnc1231:
Added the two I know about, signing off for the day now though.

tpoterba:
seems fine to me!

danking (approved):

PR #10422, opened 2021-04-30
[query] Remove PCode.tcode

chrisvittal:


tpoterba (changes requested):
nice! doesn't compile though, looks like a few more overridden methods

chrisvittal:
should be fixed now

tpoterba (approved):

PR #10426, opened 2021-05-03
[query] Lower BlockMatrix.fill

johnc1231:
This was an easy one, since  we already  had `BlockMatrixBroadcast` lowered. All  `ValueToBlockMatrix` does is make a tiny 1x1 BlockMatrix that can later be broadcast. I kind  of think this is a weird  IR design and that we should actually have a  `BlockMatrixBroadcast`  that takes a value `IR` child instead of a `BlockMatrixIR`, but I'll experiment with that refactoring later. Trying to just get the current thing lowered.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/lowering/LowerBlockMatrixIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/lowering/LowerBlockMatrixIR.scala
@@ -223,7 +223,8 @@ object LowerBlockMatrixIR {
         new BlockMatrixStage(Array(elt.name -> eltValue), TTuple(TInt64, TInt64)) {
           def blockContext(idx: (Int, Int)): IR = {
             val (i, j) = x.typ.blockShape(idx._1, idx._2)
-            MakeTuple.ordered(FastSeq(I64(i), I64(j)))
+            // Block size in block matrix type is an int, so it's safe to cast.
+            MakeTuple.ordered(FastSeq(I32(i.toInt), I32(j.toInt)))
           }
           def blockBody(ctxRef: Ref): IR =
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/expr/ir/lowering/LowerBlockMatrixIR.scala
```diff
@@ -223,10 +223,10 @@ object LowerBlockMatrixIR {
         new BlockMatrixStage(Array(elt.name -> eltValue), TTuple(TInt64, TInt64)) {
           def blockContext(idx: (Int, Int)): IR = {
             val (i, j) = x.typ.blockShape(idx._1, idx._2)
-            MakeTuple.ordered(FastSeq(I64(i), I64(j)))
+            MakeTuple.ordered(FastSeq(I64(i.toInt), I64(j.toInt)))
           }
           def blockBody(ctxRef: Ref): IR =
-            MakeNDArray(ToArray(mapIR(rangeIR(GetTupleElement(ctxRef, 0) * GetTupleElement(ctxRef, 1)))(_ => elt)),
+            MakeNDArray(ToArray(mapIR(rangeIR( invoke("toInt32", TInt32, GetTupleElement(ctxRef, 0) * GetTupleElement(ctxRef, 1))) )(_ => elt)),
```
tpoterba, line 229: is this doing a int64=>int32 cast? this should just be a `Cast` node then. You can also use the alias `(ir).toI`

tpoterba (changes requested):

hail/src/main/scala/is/hail/expr/ir/lowering/LowerBlockMatrixIR.scala
johnc1231: Fixed, didn't know about `Cast` node, much easier.

tpoterba (approved):

PR #10428, opened 2021-05-03
[batch] Make unschedule job not blocking when worker is unresponsive

jigold:


batch/batch/driver/job.py
```diff
@@ -244,7 +244,7 @@ async def make_request():
             await instance.incr_failed_request_count()
             raise
 
-    await retry_transient_errors(make_request)
+    await retry_transient_errors_n_times(5, make_request)
```
danking, line 247: If we reach the maximum number of failures, what happens? Will the worker eventually learn that the job is deleted?

danking (dismissed):

batch/batch/driver/job.py
jigold: At least this way, we'll get an interrupt. But this job will tried to be unscheduled over and over again from the SQL query until the worker is gone.

batch/batch/front_end/front_end.py
```diff
@@ -156,12 +156,15 @@ async def _handle_ui_error(session, f, *args, **kwargs):
...
         set_message(session, str(e), 'error')
+        log.info(f'ui error: KeyError {e}')
         return True
     except BatchOperationAlreadyCompletedError as e:
         set_message(session, e.message, e.ui_error_type)
+        log.info(f'ui error: BatchOperationAlreadyCompletedError {e.message}')
         return True
     except BatchUserError as e:
         set_message(session, e.message, e.ui_error_type)
+        log.info(f'ui error: BatchUserError {e.message}')
```
danking, line 167: We already have a *ton* of log messages (in Konrad's job its 6.5% of all CPU use, only 1.7 is due to non-info logs). What's the developer use-case for these logs?

batch/batch/driver/job.py
```diff
@@ -258,7 +258,10 @@ async def make_request():
             await instance.incr_failed_request_count()
             raise
 
-    await retry_transient_errors(make_request)
+    def excl(e: Exception):
+        return isinstance(e, asyncio.TimeoutError)
+
+    await retry_transient_errors_with_exclusions(excl, make_request)
```
danking, line 264: I could be convinced that this is useful in general, but in this case it seems more straightforward to just do the following. What do you think?
```python3
    async def make_request():
         if instance.state in ('inactive', 'deleted'):
             return
         try:
             async with aiohttp.ClientSession(raise_for_status=True, timeout=aiohttp.ClientTimeout(total=5)) as session:
                 await session.delete(url)
                 await instance.mark_healthy()
         except asyncio.TimeoutError as err:
             return
         except aiohttp.ClientResponseError as err:
            if err.status == 404:
                await instance.mark_healthy()
                return
             await instance.incr_failed_request_count()
             raise

     await retry_transient_errors(make_request)
```

danking: It keeps the timeout error syntactically closer to the code that times out which seems good.

danking (changes requested):

batch/batch/front_end/front_end.py
jigold: I wanted an easier way to tell what error was seen (or should have been seen) by a user in the logs. This is on the front end for UI calls. I don't think it's too many to add this for the rare cases where it would be helpful. For example, when Ben's cancel operation wasn't working. I wanted to know if he should have been getting a session error message (or an error message at all).

danking (approved):

PR #10429, opened 2021-05-03
[batch] Add logs to disk operations

jigold:
Also, make sure we always delete the disk despite any failures

### DIFF
--- a/batch/batch/worker/disk.py
+++ b/batch/batch/worker/disk.py
@@ -42,14 +42,10 @@ class Disk:
             await self._delete()
 
     async def _format(self):
-        try:
-            await check_shell_output(f'mkfs.ext4 -m 0 -E lazy_itable_init=0,lazy_journal_init=0,discard {self.disk_path}')
-            await check_shell_output(f'mkdir -p {self.mount_path}')
-            await check_shell_output(f'mount -o discard,defaults {self.disk_path} {self.mount_path}')
-            await check_shell_output(f'chmod a+w {self.mount_path}')
-        except Exception as e:
-            log.error(f'error while formatting disk {self}: {e}')
-            raise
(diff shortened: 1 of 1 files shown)


batch/batch/worker/disk.py
```diff
@@ -36,14 +36,20 @@ async def create(self, labels=None):
...
-        await check_shell_output(f'mount -o discard,defaults {self.disk_path} {self.mount_path}')
-        await check_shell_output(f'chmod a+w {self.mount_path}')
+        try:
+            await check_shell_output(f'mkfs.ext4 -m 0 -E lazy_itable_init=0,lazy_journal_init=0,discard {self.disk_path}')
+            await check_shell_output(f'mkdir -p {self.mount_path}')
+            await check_shell_output(f'mount -o discard,defaults {self.disk_path} {self.mount_path}')
+            await check_shell_output(f'chmod a+w {self.mount_path}')
+        except Exception as e:
+            log.error(f'error while formatting disk {self}: {e}')
+            raise
```
danking, line 52: Why is the catch and log in DockerJob.run not sufficient?

batch/batch/worker/disk.py
```diff
@@ -57,28 +63,42 @@ async def _create(self, labels=None):
...
             }
 
-            await self.compute_client.create_disk(f'/zones/{self.zone}/disks',
-                                                  json=config)
+            try:
+                await self.compute_client.create_disk(f'/zones/{self.zone}/disks',
+                                                      json=config)
+            except Exception as e:
+                log.error(f'error while creating disk {self}: {e}')
+                raise e
```
danking, line 71: same question as above.

danking (changes requested):
I'll approve the try-finally in delete, but I don't understand why we need to log these exceptions if they all propagate up to DockerJob's run method.

danking (approved):

PR #10431, opened 2021-05-03
[query] Renovate StagedArrayBuilder and sorting IR code generators

tpoterba:
* push SingleCodeType through StagedArrayBuilder
* what was makeDependentSortingFunction, let's not do that anymore
* generate a staged quicksort instead of pass a virtual function
  as a param and copying everything many times
* Break apart emitters for different IRs. They don't share that much.

Non-randomly assigning Chris because we talked about this design in the past.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/ArraySorter.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/ArraySorter.scala
@@ -6,6 +6,8 @@ import is.hail.types.physical.{PCanonicalArray, PCanonicalDict, PCanonicalSet, P
 import is.hail.types.virtual.{TArray, TDict, TSet, Type}
 import is.hail.utils.FastIndexedSeq
 
+import scala.language.existentials
+
 class ArraySorter(r: EmitRegion, array: StagedArrayBuilder) {
   val ti: TypeInfo[_] = array.elt.ti
   val mb: EmitMethodBuilder[_] = r.mb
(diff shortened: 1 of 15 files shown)


chrisvittal (changes requested):
looks good. make the tests pass

chrisvittal (approved):

hail/src/main/scala/is/hail/expr/ir/ArraySorter.scala
```diff
@@ -2,75 +2,207 @@ package is.hail.expr.ir
...
+
+  private[this] val workingArrayInfo = arrayInfo(array.ti)
+  private[this] val workingArray1 = mb.genFieldThisRef("sorter_working_array")(workingArrayInfo)
+  private[this] val workingArray2 = mb.genFieldThisRef("sorter_working_array")(workingArrayInfo)
+
+  private[this] def arrayRef(workingArray: Code[Array[_]]): UntypedCodeArray = new UntypedCodeArray(workingArray, array.ti)
+
+  def sort(cb: EmitCodeBuilder, region: Value[Region], comparesLessThan: (EmitCodeBuilder, Value[Region], Code[_], Code[_]) => Code[Boolean]): Unit = {
+
+    val sortMB = cb.emb.ecb.newEmitMethod("arraySorter_outer", FastIndexedSeq[ParamType](classInfo[Region]), UnitInfo)
```
chrisvittal, line 25: apparently this is leading to duplicate method errors.

chrisvittal (changes requested):
one issue with the tests that I saw

hail/src/main/scala/is/hail/expr/ir/ArraySorter.scala
tpoterba: yeah this is supposed to be `genEmitMethod` which has the same signature but generates a name instead

chrisvittal (approved):

PR #10432, opened 2021-05-03
[batch] Retry disk operations

jigold:


batch/batch/worker/disk.py
```diff
@@ -1,10 +1,23 @@
 import logging
 
-from hailtop.utils import check_shell_output, LoggingTimer
+from hailtop.utils import check_shell_output, LoggingTimer, sleep_and_backoff
 
 log = logging.getLogger('disk')
 
 
+def retry_n_times(n: int, f, *args, **kwargs):
```
danking, line 8: This is too broad. Let's specifically look for the rate limit message in the body of the response ("Rate Limit Exceeded"). Google describes the rate limiting like this:

> Currently, projects are limited to the following API rate limit categories. Each limit category is counted separately, so you can achieve the maximum limit in each category simultaneously. Rate limits are enforced and automatically refilled at intervals of every 100 seconds. If you exceed the rate limit for an interval, your request should succeed once you have passed that interval. For example, 20 requests/second would translate to 2000 requests within 100 seconds. That means that if you reach a specific limit anytime within 100 seconds, you need to wait for your quota bucket to refresh to make more requests.

If we get a 403 Rate Limit Exceeded, then we might need to wait as long as 100 seconds before we get a chance to perform our operation.

If a cluster of 100 workers are each trying to mount 32 disks, that's 6400 create and attach operations. We'll fill up the first 2000 operations nearly immediately. Then we wait 100s and another 2000 operations will succeed. Then we wait 100s and another 2000 operations will succeed. Another 100s and the remaining 1400 operations will succeed.

Hmm, what is the ideal outcome? We want to mount disks as fast as possible. Batch Driver also needs to be able to manage instances. In the absence of coordination, exponential backoff hopefully allows everyone to get access. The workers should probably start at a higher delay and back off further than the driver so that the driver is more likely to succeed.

We should integrate this directly into the ComputeClient class. The ComputeClient should probably have a configurable initial delay and max delay. I think the driver is fine with the default max delay of 30s. The workers, on the other hand, should probably be allowed to back off past 30s so that batch driver is more likely to succeed.

danking (changes requested):

batch/batch/worker/disk.py
danking: I asked cotton to give this a think and get back to us.

jigold:
I wasn't sure how to get RateLimitExceeded from the exception.

jigold:
Also, the quota documentation says the error should be this:

If you exceeded a quota with an HTTP/REST request, Google Cloud returns an HTTP 429 TOO MANY REQUESTS status code.

hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -17,9 +21,15 @@ async def request_with_wait_for_done(request_f, path, params: MutableMapping[str
...
     while True:
-        resp = await request_f(path, params=params, **kwargs)
-        if resp['status'] == 'DONE':
-            return resp
+        try:
+            resp = await request_f(path, params=params, **kwargs)
+            if resp['status'] == 'DONE':
+                return resp
+        except aiohttp.ClientResponseError as e:
+            if e.status == 403:
```
danking, line 29: I think we have to also check the message for the string "Rate Limit Exceeded". 403 could also be a real permissions issue.

I think we should just add this to is_transient_errors, it could show up in other APIs and the response is always the same: exponentially back off.

danking (dismissed):

danking:
I think the docs you're referencing aren't the Compute Engine API docs: https://cloud.google.com/compute/docs/api-rate-limits. The audit logs also show 403s: https://console.cloud.google.com/logs/query;query=%22403%22%0A%22Rate%20Limit%20Exceeded%22;timeRange=2021-05-04T00:05:00.000Z%2F2021-05-04T01:06:00.000Z;pinnedLogId=2021-05-04T00:40:25.985091Z%2F-1xc71ve6z9k6;cursorTimestamp=2021-05-04T00:40:31.308803Z?project=hail-vdc&query=%0A

jigold:
The problem is the exception we actually get tells us no information about the problem other than it is Forbidden.

aiohttp.client_exceptions.ClientResponseError: 403, message='Forbidden', url=URL('https://compute.googleapis.com/compute/v1/projects/hail-vdc/zones/us-central1-a/instances/batch-worker-default-highmem-40e6w/detachDisk?deviceName=batch-disk-be5253d5361d45118350&requestId=130727e3-f96c-4f3a-ac86-1df585bc4e1c'

jigold:
The logs you're seeing are the cloud audit logs.

jigold:
bump

danking:
Sorry for the delay! Every HTTP Status Code has a "message" associated with it. That message doesn't vary by request, for 403 its "Forbidden". There's a [full list at the MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status). You have to read the body of the request to get rate limit exceeded. I'm sorry, I was playing fast and loose with words, when I said "message," I was referring to the body of the HTTP request, not the `message` field of the aiohttp ClientResponseError.

The message, the code and all that come from the HTTP Header, before the body is parsed:
```
(base) # curl -v https://google.com/123 >/dev/null 
...
> GET /123 HTTP/1.1
> Host: google.com
> User-Agent: curl/7.69.1
> Accept: */*
> 
...
< HTTP/1.1 404 Not Found
< Content-Type: text/html; charset=UTF-8
< Referrer-Policy: no-referrer
< Content-Length: 1564
< Date: Thu, 13 May 2021 15:31:43 GMT
< Alt-Svc: h3-29=":443"; ma=2592000,h3-T051=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43"
```

danking:
Basically, we need to `await resp.body()` and see if we the string `"Rate Limit Exceeded"` is present in there.

jigold:
Is this request_info here: https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponseError.request_info?

Otherwise, I don't see how we can add this to `is_transient_error` where we only have the `ClientResponseError`

danking:
AFAIK, the ClientResponseError only contains information from the original request or the headers. The RequestInfo contains information from our request to Google, like the URL.

You're right, incorporating this into is_transient_error is a bit complex. We need to somehow communicate the body, if any, to is_transient_error. I started to go down this route with httpx.py. I wanted to wrap ClientSession with a new HailClientSession whose `get`, `post` etc. methods would check for non-successful statuses themselves, read the body, and raise HailHTTPException which included the status code *and the body*. Then in is_transient_error we can look for HailHTTPException and use the body to determine if we should retry.

For now, we can just fix the compute client.

jigold:
Maybe I'm missing something, but I don't see how you can get a response object when there's been an exception.

```python3
        try:
            resp = await request_f(path, params=params, **kwargs)
            if e.status == 403:
                await resp.body()  # resp doesn't exist here!
```

jigold:
So this is the same problem you discovered last year. It's supposed to get fixed in 4.0.0, but that hasn't been released yet. https://github.com/hail-is/hail/commit/337383674697f51dd6e04a3be3acedf4ed7a59a9

jigold:
bump

jigold:
bump

danking:
Sorry! I didn't see this because of the review.

The root problem is that `raise_for_status` ignores the response body. This is a [known issue in aiohttp](https://github.com/aio-libs/aiohttp/issues/4600). It will be fixed in 4.0.0, but development on that seems slow.

There's a variety of solutions to this problem. Every solution avoids aiohttp's raise_for_status and replaces it with  something that includes the response body in the error message.

A thorough fix to this is to finish the work I started in `httpx.py`. Instead of returning an `aiohttp.ClientSession` we could return a shim class that wraps `aiohttp.ClientSession` and checks the status code itself and raises an error *with the response body*.

A smaller fix that only addresses aiogoogle would be to modify `aiogoogle.auth.Session` to:
1. Not pass `raise_for_status` on to `aiohttp.ClientSession`.
2. Store raise_for_status as a field on `aiogoogle.auth.Session`.
2. In `aiogoogle.auth.Session.request`, if `self.raise_for_status` is true and the response status is greater than or equal to 400, retrieve the response body and raise an exception (maybe `HailHTTPError`) that includes the body.
3. Ensure `is_transient_error` properly handles whatever exception we raise.

jigold:
Stacked on #10541 now.

hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -17,7 +20,7 @@ async def request_with_wait_for_done(request_f, path, params: MutableMapping[str
 
     delay = 0.2
     while True:
-        resp = await request_f(path, params=params, **kwargs)
+        resp = await retry_transient_errors(request_f, path, params=params, **kwargs)
```
danking, line 23: I don't think this is necessary, Cotton has retry by default in `aiogoogle.auth.session.Session.request`. BaseClient uses a Session by default. BaseClient's post, etc. all use Session's post, etc. Session's post uses request.

danking (changes requested):

danking (approved):

PR #10434, opened 2021-05-04
[batch] rate-limit use of the Google Compute Engine API

danking:


batch/batch/driver/main.py
```diff
@@ -891,7 +891,15 @@ async def on_startup(app):
...
-        PROJECT, credentials=aiogoogle_credentials)
+        PROJECT,
+        credentials=aiogoogle_credentials,
+        # https://cloud.google.com/compute/docs/api-rate-limits
+        # The operations of most concern are: createDisk, attachDisk, detachDisk,
+        # and deleteDisk which I believe these all qualify as simple "Queries" which
+        # have a project-wide limit of 20 requests per second. We keep 18 requests per
+        # second uses 90% of our quota. The remaining two operations per second enable
+        # developers fleixiblity to create disks and instances for debugging purposes.
+        rate_limit=RateLimit(18, 100))
```
jigold, line 902: I'm not sure if this actually solves the problem. This is one instance of the compute client running on the driver. But each worker has its own client. I think we need retry loops on the workers as well.

jigold (changes requested):

PR #10437, opened 2021-05-04
[query] remove EmitCode.value calls in CountAggregator

chrisvittal:


chrisvittal:
nevermind, folding this into another PR

PR #10438, opened 2021-05-04
[black] Add pyproject.toml and run black on services code

daniel-goldstein:
Black only supports the `pyproject.toml` configuration file. Between all our tools (mypy, flake8, black, pylint…), there's no single config file they all seem to support. I moved the black config from `.pre-commit-config.yaml` to `pyproject.toml` so black can pick up the configuration whenever it's run, so editor plugins to activate autoformatting can work by default now.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,8 +10,6 @@ repos:
     hooks:
       - id: black
         language_version: python3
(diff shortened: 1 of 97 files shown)


danking (approved):
Awesome. And you'll mark this commit as an ignored one for blame once it merges?

PR #10444, opened 2021-05-05
Parameterise docker registry in the kaniko config

vladsaveliev:


danking (approved):
oops, sorry about that!

PR #10447, opened 2021-05-05
[query] Create BlockMatrix from NDArray + Actually Running Lowering Tests

johnc1231:
In this PR:

- I add `BlockMatrix.from_ndarray`. The implementation isn't great, it just basically just evals the ndarray and adds NDArray support to `ValueToBlockMatrix`. A better version wouldn't cross the Python / Java boundary at all, but I want something that works on all backends, so for now this will have to suffice. Any solution will at least need to communicate the shape of the ndarray back to python, since it's tracked in the block matrix type. 
- With this new method, I can now get many tests in `test_linalg.py` to run on the local / service backends. Most BM lowering was apparently untested before, so some bug fixes were necessary, including:
    - Support requiredness analysis on BlockMatrix, even though the answer is always required. 
    - Use `CompileAndEvaluate`  rather than `Interpret` to evaluate the child node in `ValueToBlockMatrix`. 
    - Casting between Int32 and Int64 in various places in lowering. Almost always the culprit was a bad interaction between ndarray shapes (which are Int64) and `StreamRange` argument (which is an Int32). This is sort of a pervasive ndarray problem  that will need to be systematically addressed at some point. I don't anticipate anyone making a BlockMatrix with blocks big enough to blow 32 bits though. 
    - Lots of fixes to the `BlockMatrixBroadcast` rule for getting diagonal of a BlockMatrix, as it was clearly never run. It had `MakeStream(StreamIR)` which was not allowed, it didn't update the context appropriately,  and it used the wrong axis to determine if something was a row vector.

### DIFF
--- a/hail/python/hail/linalg/blockmatrix.py
+++ b/hail/python/hail/linalg/blockmatrix.py
@@ -508,6 +508,12 @@ class BlockMatrix(object):
 
         return BlockMatrix(ValueToBlockMatrix(hl.literal(data)._ir, [n_rows, n_cols], block_size))
 
+    @classmethod
+    def from_ndarray(cls, ndarray_expression, block_size=4096):
+        """Create a BlockMatrix from an ndarray"""
+        shape = hl.eval(ndarray_expression.shape)
+        return BlockMatrix(ValueToBlockMatrix(ndarray_expression._ir, shape, block_size))
+
     @staticmethod
     def default_block_size():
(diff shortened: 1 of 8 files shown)


johnc1231:
I feel good about most of this, but the `Requiredness` stuff I'm not sure I did right.

tpoterba (approved):

PR #10449, opened 2021-05-05
[query] Remove EmitCode.get

chrisvittal:
replace usages with toI.get instead


chrisvittal:
I know this fails, I'm not sure why yet.

tpoterba:
I bet it's usages of EmitCode.get doing the wrong thing. Just delete EmitCode.get entirely, I think.

chrisvittal:
Thanks!

tpoterba (approved):

PR #10460, opened 2021-05-06
[utils] retry transient OSError 104 connection reset by peer

jigold:


hail/python/hailtop/utils/utils.py
```diff
@@ -565,7 +565,8 @@ def is_transient_error(e):
                             errno.EHOSTUNREACH,
                             errno.ECONNRESET,
                             errno.ENETUNREACH,
-                            errno.EPIPE
+                            errno.EPIPE,
+                            errno.ENOTRECOVERABLE
```
danking, line 569: Connection reset by peer is ECONNRESET (errno 104 on Linux machines), which looks like we already retry. The issue is that ClientOSError is caught on line 553 and we check if its cause is transient. Unfortunately, aiohttp creates ClientOSError's [without a cause, sometimes](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client_proto.py#L97).

Let's move the ClientOSError check below the OSError check so that we first check if the errno is known bad before we look at the cause.

danking (changes requested):

PR #10466, opened 2021-05-07
[query] Remove PValue.value

chrisvittal:


chrisvittal:
Non-randomly assigned to @johnc1231 as all usages were in NDArray handling code.

johnc1231 (approved):

PR #10467, opened 2021-05-07
[batch] Configure XFS quotas without configuration files

daniel-goldstein:
Not sure if this passes tests yet. We'll see

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -969,7 +969,7 @@ class DockerJob(Job):
 
                 await check_shell_output(f'xfs_quota -x -c "project -s -p {self.scratch} {self.project_id}" /host/')
                 await check_shell_output(
-                    f'xfs_quota -x-c "limit -p bsoft={data_disk_storage_in_bytes} bhard={data_disk_storage_in_bytes} {self.project_id}" /host/'
+                    f'xfs_quota -x -c "limit -p bsoft={data_disk_storage_in_bytes} bhard={data_disk_storage_in_bytes} {self.project_id}" /host/'
                 )
 
                 if self.secrets:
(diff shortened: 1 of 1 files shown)


jigold (changes requested):
This looks fine to me, but can you double check the logs for either this running in your dev namespace or the test PR namespace to make sure there's no errors. I had a bunch of error messages when I was trying to get this to work that may or may not have impacted the functionality. That's why I put in all the deletes of the project ids from the files. Can you also send a link for the docs for the other interface?

daniel-goldstein:
Good idea, I'll spam it a bit and check for errors. I feel like I initially found this in deep in a redhat tutorial, but ultimately found it again at the bottom of the [man page](https://man7.org/linux/man-pages/man8/xfs_quota.8.html). I was following this example:
```
       Enabling project quota on an XFS filesystem (restrict files in
       log file directories to only using 1 gigabyte of space).

            # mount -o prjquota /dev/xvm/var /var
            # echo 42:/var/log >> /etc/projects
            # echo logfiles:42 >> /etc/projid
            # xfs_quota -x -c 'project -s logfiles' /var
            # xfs_quota -x -c 'limit -p bhard=1g logfiles' /var

       Same as above without a need for configuration files.

            # rm -f /etc/projects /etc/projid
            # mount -o prjquota /dev/xvm/var /var
            # xfs_quota -x -c 'project -s -p /var/log 42' /var
            # xfs_quota -x -c 'limit -p bhard=1g 42' /var
```

daniel-goldstein:
I checked the [logs for this PR](https://console.cloud.google.com/logs/query;query=logName:%22worker%22%0Alabels.namespace:%22pr-10467%22;timeRange=PT6H;cursorTimestamp=2021-05-07T19:35:43.101282634Z?project=hail-vdc&folder=true&organizationId=548622027621&query=%0A), both normal logs and just ERRORs and didn't see anything abnormal.

jigold (approved):

PR #10476, opened 2021-05-11
[query] Lower MatrixBlockMatrixWriter

johnc1231:
This PR lowers `MatrixBlockMatrixWriter`, which writes an entry field of a `MatrixTable` out into a `BlockMatrix`. Its main use is to support `BlockMatrix.from_entry_expr`. Accordingly,  the `from_entry_expr` test on the `LocalBackend` is now passing. 

Alright, removed all shuffles from this, should be good for review.


johnc1231:
@tpoterba One thing I haven't dealt with before that could use a double check: Do I need `requiresMemoryManagementPerElement` anywhere? I think mostly my streams get `ToArray`ed, and so can't be cleaned up between elements?

tpoterba (approved):
looks good 👍

PR #10478, opened 2021-05-12
[query] Lower BlockMatrix.entries

johnc1231:
This PR lowers `BlockMatrixToTable`, which is the implementation of Hail's `BlockMatrix.entries`. It turns a `BlockMatrix` to a table of `(i, j, entry)` triplets.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/lowering/LowerBlockMatrixIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/lowering/LowerBlockMatrixIR.scala
@@ -165,27 +165,28 @@ abstract class BlockMatrixStage(val globalVals: Array[(String, IR)], val ctxType
 }
 
 object LowerBlockMatrixIR {
-  def apply(node: IR, typesToLower: DArrayLowering.Type, ctx: ExecuteContext, r: RequirednessAnalysis, relationalLetsAbove: Map[String, IR]): IR = {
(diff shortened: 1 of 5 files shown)


hail/src/main/scala/is/hail/expr/ir/lowering/LowerBlockMatrixIR.scala
```diff
@@ -167,25 +168,89 @@ abstract class BlockMatrixStage(val globalVals: Array[(String, IR)], val ctxType
...
+
+    def tsPartitionFunction(ctxRef: Ref): IR = {
+      val s = MakeStruct(Seq("blockRow" -> GetTupleElement(GetField(ctxRef, "new"), 0), "blockCol" -> GetTupleElement(GetField(ctxRef, "new"), 1), "block" -> bmsWithCtx.blockBody(ctxRef)))
+      MakeStream(Seq(
+        s
+      ), TStream(s.typ))
     }
+    val ts = TableStage(letBindings, bcFields, Ref(globalsId, emptyGlobals.typ), RVDPartitioner.unkeyed(blocksRowMajor.size), TableStageDependency.none, contextsIR, tsPartitionFunction)
+    ts
+  }
```
johnc1231, line 230: This is the bulk of the PR. Everything else in this file is rearranging methods so they're not inner methods anymore, allowing me to call from TableStage.

chrisvittal (approved):

PR #10480, opened 2021-05-13
[batch] kill workers that are not responding

jigold:
after 5+ failed requests and 5+ minutes since last successful request

### DIFF
--- a/batch/batch/driver/instance_collection.py
+++ b/batch/batch/driver/instance_collection.py
@@ -123,9 +123,11 @@ class InstanceCollection:
                 return
             raise
 
-        if instance.failed_request_count > 10 and time_msecs() - instance.last_updated > 5 * 60 * 1000:
-            log.exception(f'removing {instance} with {instance.failed_request_count} failed request counts after more than 5 minutes')
-            await self.remove_instance(instance, 'not_responding')
+        if (instance.state == 'active'
+                and instance.failed_request_count > 5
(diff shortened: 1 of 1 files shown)


batch/batch/driver/instance_collection.py
```diff
@@ -123,6 +123,11 @@ async def check_on_instance(self, instance):
                 return
             raise
 
+        if instance.failed_request_count > 10 and time_msecs() - instance.last_updated > 5 * 60 * 1000:
+            log.exception(f'removing {instance} with {instance.failed_request_count} failed request counts after more than 5 minutes')
+            await self.remove_instance(instance, 'not_responding')
```
danking, line 128: shouldn't we delete it too?

danking: I feel like what we need to do depends on the instance state no? If the instance is stopping or terminated, we just need to deactivate.

Provisioning instances will get killed anyway.

This really seems only relevant to STAGING or RUNNING instances, right? It sounds like the if on 153 should be if pending _or_ many failed requests.

danking (changes requested):

batch/batch/driver/instance_collection.py
jigold: I made it so the check only happens for active instances. Good catch on the delete.

danking (approved):

PR #10482, opened 2021-05-14
[query] Turn on eager whole-stage codegen for SparkBackend (mk 2)

tpoterba:
Most pipelines will take the old path.


tpoterba:
I believe this is ready for review after #10746 goes in 😲

johnc1231:
#10746 is in, but this has conflicts to address

hail/src/main/scala/is/hail/rvd/AbstractRVDSpec.scala
```diff
@@ -155,7 +155,7 @@ object AbstractRVDSpec {
         val extendedNewPartitioner = np.extendKey(partitioner.kType)
         val tmpPartitioner = extendedNewPartitioner.intersect(partitioner)
 
-        val reader = PartitionZippedIndexedNativeReader(specLeft.typedCodecSpec, specRight.typedCodecSpec, indexSpecLeft, indexSpecRight, requestedKey)
+        val reader = PartitionZippedIndexedNativeReader(specLeft.typedCodecSpec, specRight.typedCodecSpec, indexSpecLeft, indexSpecRight, specLeft.key)
```
johnc1231, line 158: Why this change? Was it wrong before?

johnc1231 (changes requested):

hail/src/main/scala/is/hail/rvd/AbstractRVDSpec.scala
tpoterba: yep.

hail/src/main/scala/is/hail/expr/ir/lowering/CanLowerEfficiently.scala
```diff
@@ -0,0 +1,111 @@
...
+        case t: TableTail => fail("TableTail has no short-circuit using known partition counts")
+        case t: TableJoin =>
+        case t: TableIntervalJoin => fail(s"TableIntervalJoin has no lowered implementation")
+        case t: TableMultiWayZipJoin =>
+        case t: TableLeftJoinRightDistinct =>
+        case t: TableMapPartitions =>
+        case t: TableMapRows => if (ContainsScan(t.newRow)) fail("TableMapRows does not have a scalable implementation of scans")
+        case t: TableMapGlobals =>
+        case t: TableExplode =>
+        case t: TableUnion if t.children.length > 16 => fail(s"TableUnion lowering generates deeply nested IR if it has many children")
```
johnc1231, line 60: Could you explain where the "deeply nested IR" comes into play? Lookinag the at the lowering rule for `TableUnion` is isn't apparent to me.

tpoterba: by "nesting depth" I'm referring to the longest path between the root of the IR tree and a leaf. Since most of our compiler passes are written as recursive, it's possible/easy for us to blow stack size limits when generating deep IRs.

TableUnion generates IR with depth proportional to the number of `t.children`, due to the way the Let bindings get placed. This isn't an inherent problem, but interacts badly with the limitations of our compiler. We should continue rewriting passes to work iteratively using a stack on the JVM heap rather than using the JVM function stack directly.

tpoterba:
Benchmarks:
https://gist.github.com/tpoterba/f74d9428a58b3ab705ded53be35e207d

tpoterba:
Ready for review 🎉

hail/src/main/scala/is/hail/expr/ir/lowering/CanLowerEfficiently.scala
johnc1231: So the lowering of `TableUnion` doesn't need to change, just our individual lowering passes?

We could also introduce a `Let*` node for nodes that introduce many bindings at once to prevent this. Probably trickier to optimize around though.

tpoterba: That's right, the lowering is correct, but we need to support deeply nested IRs in other compiler infrastructure.

Yeah, tougher to optimize with that node. At some point soon, we're going to discuss the Future of IR™ with Cotton's proposal for a graph-based IR which we will probably introduce to Scala from the bottom up.

johnc1231 (approved):
Looks good!

PR #10484, opened 2021-05-14
[batch] Add infrastructure for open batches

jigold:
Batch Client:
- Added new parameter max_idle_time (seconds)
- Removed BatchBuilder and fused it with Batch
- Added 2 new operations:
  - Commit 
  - Close
- Kept submit which is the same as close for backwards compatibility reasons
- commit, close, and submit now return the batch so these methods can be chained together
- create_batch stayed the same

Tests:
- Added 3 new tests for new functionality
- Renamed a bunch of variables in the tests and cleaned up the variables

Database:
- Added time_last_updated for determining how long a batch has been idle
- Added max_idle_time
- Added a closed field and changed the possible states for a batch to created, running, and complete (removed open)
- Changed close_batch to be commit_staged_jobs. Changes were made to make sure this worked even if there were 0 jobs to actually commit. 
- Changed cancel_batch to always set cancelled = 1 and closed = 1 regardless of whether the batch is actually running. The time_completed is only set if no jobs are currently running. Otherwise, the time_completed will be set in MJC. **It also commits any jobs that are pending before cancelling the batch.** I'm not sure if we want this behavior or not.

Driver:
- Runs a loop every 60 seconds to close batches with max_idle_time greater than that specified.

UI:
- The UI changed the batches table to be time_created instead of time_closed as the Submitted/Created column. The duration is the time from time_complete - time_created for newer batches instead of time_complete - time_closed.
- Added a close button
- An open batch (even one just in the created state) can be cancelled or closed. **This might be confusing**

Other:
- Cancel is idempotent
- Getting the batch state and time_completed correct was tricky.


jigold:
I overlooked something pretty critical. I'm going to close this while I figure out what to do.

jigold:
Possible optimization is to only run the long query that figures out the ready jobs if the n_commits > 0.

jigold:
I added a new field to the jobs table and a new index. This will cause this migration to take at least 30 minutes.

danking:
Sorry, I will takea  look tomorrow.

batch/batch/front_end/front_end.py
```diff
@@ -1125,10 +1138,29 @@ async def cancel_batch(request, userdata, batch_id):  # pylint: disable=unused-a
...
+    async with client_session() as session:
+        await request_retry_transient_errors(
+            session,
+            'PATCH',
+            deploy_config.url('batch-driver', f'/api/v1alpha/batches/{user}/{batch_id}/commit'),
+            headers=app['batch_headers'],
+        )
+
+
+@routes.patch('/api/v1alpha/batches/{batch_id}/commit')
```
danking, line 1160: I don't understand the motivation for this. In my head, the conversation between client and server would go like this:
```
> create batch
< OK, batch 10 created
> create jobs ...
< OK
```
Any job with no parents is eligible for scheduling now.
```
> create jobs ...
< OK
```
Any job whose parents are completed is eligible for scheduling.
```
> batch 10 status
< NOT DONE
... exponential back off of status requests
> batch 10 status
< DONE
```

In this conversation there's only three endpoints: 
1. create batch`POST /api/v1alpha/batches/create`
2. create jobs `POST /api/v1alpha/batches/{batch_id}/jobs/create`
3. get batch `GET /api/v1alpha/batches/{batch_id}`

I think we still need batch cancel and batch delete.

I think what I'm looking for is:
1. What use-cases does commit support?
2. Why does the user need to call close? Put another way: what goes wrong if a batch is "closed" 24 hours later?

I think there are four batch states:
1. accepting-new-jobs
2. not-accepting-new-jobs
3. cancelled 
4. deleted

I think the notion of "complete" for a batch doesn't quite make sense any more. In `hailtop.batch` we can still define "complete" as: "BatchBuilder.run was called and n_jobs == n_completed_jobs"

danking (changes requested):

jigold:
Closing as this is quite stale. Will revisit in 2022.

PR #10486, opened 2021-05-14
[compiler] Some groundwork for SStackStruct

tpoterba:
Needed to add a bit more tooling to emitcode/emitsettable/emittype for stack struct purposes.


tpoterba:
Also, PresentEmitSettable wasn't doing anything other than adding another name to keep track of.

PR #10487, opened 2021-05-14
[compiler] Orderings now use STypes/SCodes only

chrisvittal:


tpoterba (changes requested):
to close the loop, I think this is the wrong approach, and the change can be done by a simple rename of PCode => SCode given that the two are now aliases.

chrisvittal:
closed in favor of #10500

PR #10495, opened 2021-05-18
[batch] Worker compaction loop

jigold:
I tested this in my namespace. We can customize the parameters -- this implementation is aggressive at making sure we're at maximum efficiency.


jigold:
I'm going to close this because it isn't the right approach.

PR #10497, opened 2021-05-18
[fs] Add higher level fs operations

jigold:
CC: @danking

### DIFF
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -540,7 +540,7 @@ class GoogleStorageMultiPartCreate(MultiPartCreate):
                     [self._part_name(i) for i in range(self._num_parts)],
                     self._dest_name)
             finally:
-                await self._fs.rmtree(self._sema, f'gs://{self._bucket}/{self._dest_dirname}_/{self._token}')
+                await self._fs.rmtree(f'gs://{self._bucket}/{self._dest_dirname}_/{self._token}', self._sema)
 
 
 class GoogleStorageAsyncFS(AsyncFS):
--- a/hail/python/hailtop/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiogoogle/client/storage_client.py
@@ -735,7 +735,7 @@ class GoogleStorageAsyncFS(AsyncFS):
             if e.status != 404:
                 raise
 
-    async def rmtree(self, sema: asyncio.Semaphore, url: str) -> None:
+    async def _rmtree(self, url: str, sema: asyncio.Semaphore) -> None:
         async with OnlineBoundedGather2(sema) as pool:
             try:
                 it = await self.listfiles(url, recursive=True)
@@ -743,6 +743,14 @@ class GoogleStorageAsyncFS(AsyncFS):
                 return
             async for entry in it:
                 await pool.call(self._remove_doesnt_exist_ok, await entry.url())
+
+    async def rmtree(self, url: str, sema: Optional[asyncio.Semaphore] = None) -> None:
+        if sema is None:
+            sema = asyncio.Semaphore(50)
+            async with sema:
+                return await self._rmtree(url, sema)
+
+        return await self._rmtree(url, sema)
 
     async def close(self) -> None:
         await self._storage_client.close()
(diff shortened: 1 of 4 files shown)


hail/python/hailtop/aiotools/fs.py
```diff
@@ -822,6 +839,18 @@ def _get_fs(self, url):
...
+        return await fs.read(url)
+
+    async def read_from(self, url: str, start: int) -> bytes:
+        fs = self._get_fs(url)
+        return await fs.read_from(url, start)
+
+    async def read_range(self, url: str, start: int, end: int) -> bytes:
+        fs = self._get_fs(url)
+        return await fs.read_range(url, start, end)
+
```
danking, line 853: I don't think these need to be here. The implementations inherited from AsyncFS should work just fine, right?

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -744,6 +744,14 @@ async def rmtree(self, sema: asyncio.Semaphore, url: str) -> None:
             async for entry in it:
                 await pool.call(self._remove_doesnt_exist_ok, await entry.url())
 
+    async def rmtree(self, url: str, sema: Optional[asyncio.Semaphore] = None) -> None:
```
cseed, line 747: I don't like moving sema because it makes the interface inconsistent.  With this change, some functions take it as the first argument, but rmtree teh last.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -130,13 +130,30 @@ async def remove(self, url: str) -> None:
...
+            return await f.read()
+
+    async def read_range(self, url: str, start: int, end: int) -> bytes:
+        n = (end - start) + 1
+        async with await self.open_from(url, start) as f:
+            return await f.read(n)
+
+    async def write(self, url: str, data: bytes, *, retry_writes: bool = True) -> None:
+        async with await self.create(url, retry_writes=retry_writes) as f:
+            await f.write(data)
```
cseed, line 155: So if you have all the data present, you shouldn't take a retry_writes argument.  You should do the rewrite with `retry_writes=False` and catch and retry yourself like it is done in `_copy_file_multi_part_main`:

```
    async def _copy_file(self, srcfile: str, destfile: str) -> None:
        ...
        dest_cm = await self.router_fs.create(destfile, retry_writes=False)
        async with dest_cm as destf:
            ...

    async def _copy_file_multi_part_main(...):
        ...
        await retry_transient_errors(self._copy_file, srcfile, destfile)
```

cseed (changes requested):

hail/python/hailtop/aiotools/fs.py
jigold: I think you need to grab the appropriate file system first. This is the RouterFS. That's what all the other methods do.

jigold:
The tests are passing. Failures are from building images.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -130,13 +130,30 @@ async def remove(self, url: str) -> None:
...
-    async def rmtree(self, sema: asyncio.Semaphore, url: str) -> None:
+    async def rmtree(self, sema: Optional[asyncio.Semaphore], url: str) -> None:
         pass
 
     async def touch(self, url: str) -> None:
         async with await self.create(url):
             pass
 
+    async def read(self, url: str) -> bytes:
+        async with await self.open(url) as f:
```
cseed, line 141: I think I agree with @danking.  This implementation should work in the RouterFS because the call to `self.open` will dispatch off the URL scheme.

hail/python/hailtop/aiotools/fs.py
cseed: So if create isn't retrying writes, you need to retry writes.  I was thinking of something like this:

```
    async def write(self, url: str, data: bytes) -> None:
        async def _write() -> None:
            async with await self.create(url, retry_writes=False) as f:
                await f.write(data)

        await retry_transient_errors(_write)
```

Let me clarify the meaning of retry writes.  By default, AsyncFS.create is retry safe.  To be retry safe with writes, the AsyncFS code has to buffer any data sent to GCS (S3, etc.) until it is acknowledged and safely stored.  That's not free.  If you pass `rewrite_retries=False`, AsyncFS.create doesn't retry writes, in which case it can fail.  `retry_writes=False` assumes that you're retrying failures at a higher level.  That's possible when you can restart the write from scratch.  That's the case here, because `data` is bytes, so if the write failed, we already have the data buffered and just write again.

cseed (changes requested):

hail/python/hailtop/aiotools/fs.py
jigold: That makes more sense. Thanks!

cseed (approved):

PR #10501, opened 2021-05-19
Fixed arbitrary maps on BlockMatrix, added test

johnc1231:
We had an old   assert  that only  applies  to Spark, moved it  into the execute method.

### DIFF
--- a/hail/python/test/hail/helpers.py
+++ b/hail/python/test/hail/helpers.py
@@ -154,8 +154,8 @@ fails_service_backend = pytest.mark.xfail(
     strict=True)
 
 def check_spark():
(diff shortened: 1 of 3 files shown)


hail/python/test/hail/helpers.py
```diff
@@ -130,6 +130,17 @@ def wrapper(func, *args, **kwargs):
 
     return wrapper
 
+def skip_when_spark_backend():
```
tpoterba, line 133: should this be a `fails_spark_backend` with xfail? because I expect this to pass once we turn on eager lowering.

tpoterba (changes requested):

tpoterba (approved):

PR #10502, opened 2021-05-19
[ci] buildkit

danking:
1. Replace kaniko with buildkit.
2. For any repository, the `cache` tag is a dump of layers that have been built before.
3. Allow ci to start docker jobs that are "unconfined" by app armor and seccomp.
4. Add `docker-build.sh` which enforces use of buildkit, knows about the `cache` tag, and ensures the cache is updated on each build.
5. Remove the `gcloud` binary from Dockerfile.base. Replace uses with either the gcloud-sdk image or with ci-utils-image, which now contains the gcloud install.
6. Move pyspark (which is huge, 100s of MB) before everything because its version rarely changes.
7. Move requirements.txt to the end of base, since it changes more often than the rest.
8. Move hailtop last in service-base because hailtop has a git SHA in it.
9. Simplify make files: always use docker-build.sh, no explicit pushes (we almost always want to push), no explicit pulls (buildkit cache doesn't need it), none of this digest nonsense (it was never accurate anyway)

When my namespace CI builds ci/test/resources/build.yaml, it finishes in 4 minutes. Still dominated by image building. Layer extraction (required when things change, e.g. hail top's SHA change or hello's python files) dominates our time. We might try collapsing the largely unchanging lower layers of service-base (pyspark, apt-get, gcs-connector, and catch2). That will hurt us when we *do* change one of those layers. Alternatively, we might make service-base based on hail-ubuntu instead of base. We could eliminate a bunch of build software like cmake, gcc, and the jdk.

I based the create-certs image on hail-ubuntu to ensure its built early and doesn't hold up service deployment.

The following is an as-cached-as-possible build. The service and hello images have to extract layers and build themselves because the SHA changed.

Screen Shot 2021-05-19 at 2 34 18 PM

I'm not even sure how much mileage we can get out of layer squashing. You can take a look at a service-base build [here](https://gist.github.com/danking/830af0688970c176ff25dbfeb4b222e7). Note the stdout comes first and then I `cat` the "trace" file which is a very weirdly formatted series of JSON objects that give more detailed information. You can clean it up a bit with `jq -c '(.Logs + .Statuses + .Vertexes) | add | {Timestamp, ID, Name}'`.


docker-build.sh
```diff
@@ -0,0 +1,31 @@
...
+       --file ${DOCKERFILE} \
+       --cache-from ${CACHE_NAME} \
+       ${EXTRA_CACHE:+--cache-from ${EXTRA_CACHE}} \
+       --build-arg BUILDKIT_INLINE_CACHE=1 \
+       --tag ${REMOTE_IMAGE_NAME} \
+       --tag ${CACHE_NAME} \
+       ${CONTEXT}
+
+time DOCKER_BUILDKIT=1 docker push ${REMOTE_IMAGE_NAME}
+time DOCKER_BUILDKIT=1 docker push ${CACHE_NAME}
```
daniel-goldstein, line 20: extra whitespace

docker/Makefile
```diff
@@ -1,96 +1,53 @@
...
-	[ "$(shell bash stat-permissions.sh core-site.xml)" = "644" ]
-	[ "$(shell bash stat-permissions.sh requirements.txt)" = "644" ]
-	[ "$(shell bash stat-permissions.sh ../pylintrc)" = "644" ]
-	[ "$(shell bash stat-permissions.sh ../setup.cfg)" = "644" ]
-	docker build -t base -f Dockerfile.base.out --cache-from base,$(BASE_LATEST),hail-ubuntu ..
-	touch base-stmp
+base: hail-ubuntu-stmp Dockerfile.base core-site.xml requirements.txt ../pylintrc ../setup.cfg
+	python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"'$$(cat hail-ubuntu-stmp)'"}}' Dockerfile.base Dockerfile.base.out
+	../docker-build.sh .. docker/Dockerfile.base.out $(BASE_IMAGE)
+	echo $(BASE_IMAGE) > base
```
daniel-goldstein, line 24: The use of the `-stmp` convention seems inconsistent to me. Looks like it's only `hail-ubuntu-stmp` now? I don't care much whether to use it or not but would prefer whichever way we can be consistent.

docker/publish-public-images.sh
```diff
@@ -10,27 +10,27 @@ hail_pip_version=$(cat hail/hail_pip_version)
...
+        ${name}
+
+    time DOCKER_BUILDKIT=1 docker push ${versioned_short}
+    time DOCKER_BUILDKIT=1 docker push ${versioned_full}
+    time DOCKER_BUILDKIT=1 docker push ${cache}
 }
 
-python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"hail-ubuntu"}}' hail/Dockerfile hail/Dockerfile.out
-build_and_push hail hail-ubuntu
+python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"gcr.io/hail-vdc/hail-ubuntu"}}' hail/Dockerfile hail/Dockerfile.out
```
daniel-goldstein, line 32: Should use `$docker_prefix`

ci/ci/build.py
```diff
@@ -315,16 +318,26 @@ def build(self, batch, code, scope):
...
-set +e
-/busybox/sh /convert-google-application-credentials-to-kaniko-auth-config
-set -e
+set +x
+/bin/sh /home/user/convert-google-application-credentials-to-docker-auth-config
+set -x
 
-exec /kaniko/executor --dockerfile={shq(dockerfile_in_context)} --context=dir://{shq(context)} --destination={shq(self.image)} --cache=true --cache-repo={shq(cache_repo)} --snapshotMode=redo --use-new-run'''
+export BUILDKITD_FLAGS=--oci-worker-no-process-sandbox
+buildctl-daemonless.sh \
```
daniel-goldstein, line 326: it's unclear to me where this script comes from.

daniel-goldstein (changes requested):
Unfortunate about Kaniko but I like how much smaller and more uniform the local and CI build systems are now. Just a few comments

danking:
I couldn't get rid of gcloud, sadly. `kubectl` is still to pervasively integrated into what build.yaml does. Although removing it would save us 100s of MB of unnecessary transfer and extraction, using the ci_utils_image for this purpose delays deployment of most services by an unacceptable amount of time.

docker/Makefile
danking: I switched everything to -image-ref. We can't use `hail-ubuntu` because that's the name of a directory containing the hail-ubuntu files.

docker/publish-public-images.sh
danking: fixed to use hail-ubuntu-image-ref

ci/ci/build.py
danking: Ah, my bad. I missed the very first commit in this series when I ripped it out of my wider batch changes.

I replaced `KANIKO_IMAGE` and `ci/kaniko/` with `BUILDKIT_IMAGE` and `ci/buildkit`. `buildctl-daemonless.sh` is part of the buildkit image.

daniel-goldstein (approved):

PR #10503, opened 2021-05-19
[batch] Catch 500 error messages in dev ui pages

jigold:
My first approach was to populate an error.html template and return that. I could get the title to be "Error", but none of the content was showing up. I couldn't figure out why, so I switched to raising HTTPErrorFound with the traceback message. This works fine, but it won't work if we need the decorator above the authentication decorators. This decorator has to be the furthest down the call stack.


danking (approved):
Super elegant solution, I love it.

PR #10504, opened 2021-05-20
[js] add scaffolding for batch front ends with react and svelte

daniel-goldstein:
Copied over from the Zulip thread:

Dan and I still have work to figure out the authentication strategy for browser-based REST requests, but as a workaround I've added a tiny aiohttp proxy that uses the python client library to fulfill the requests, which could enable local frontend work while we figure out the right way to do authentication and streaming data through websockets. Implementing the polling and separating it logically from the view components was actually a nice little case study in how to do this in React/Svelte, but is far from an honest or thorough comparison.

If you want to run it for yourself, you can pull down the branch in that PR and then do the following (which I'll write dev docs for if this is something that we actually want to check in):

Install node if you do not have it
Run npm install in the $HAIL, $HAIL/js_common, and $HAIL/batch2/react-batch (or svelte-batch) directories
In one terminal in $HAIL/batch2, run python proxy.py
In another terminal in one of the react-batch or svelte-batch directories, run npm run dev
Go to localhost:3000 in your browser if it didn't pop up automatically


batch2/proxy.py
```diff
@@ -0,0 +1,33 @@
...
+#     route = request.match_info['route']
+#     ws = web.WebSocketResponse(heartbeat=30)
+#     await ws.prepare(request)
+#     try:
+#         while True:
+#             data = await client._get(f'/api/{route}')
+#             await ws.send_json(await data.json())
+#             await asyncio.sleep(1)
+#     finally:
+#         await ws.close()
```
danking, line 28: let's axe the commented code

js_common/react/hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return value
+}
+
+// TODO This closes the websocket when the component using this
+// hook is unmounted. It is not clear to me what the best overall
+// strategy is around websockets, and whether we want to prioritize
+// simple 1:1 socket to REST url or minimize number of opening and
+// closing connections with a more complicated method of passing
+// messages.
```
danking, line 26: Let's stick with 1:1 socket:endpoint for now. Lots of unnecessary concurrency seems bad for batch, but I'm not sure how to think about it yet. Let's remove the TODO, I don't like to have those hanging around in our code.

danking (dismissed):
Man something feels wrong about package-lock files but, 🤷‍♀️ Two small comments, but otherwise I'm OK merging this. It will enable collaborative exploration of this approach between all of services team.

CDiaz96 (approved):
I in the other hand am very accustomed to package-lock files in js.

Looks good to me, apart from what Dan mentioned.

PR #10508, opened 2021-05-20
[batch] Fix local backend too many args error

jigold:
CHANGELOG: Fixed the error 'Argument list too long: '/bin/sh'' when using the LocalBackend


danking (approved):
LGTM

PR #10509, opened 2021-05-20
[compiler] Don't use bad boxed array builders

tpoterba:
BoxedArrayBuilder's type parameter needs to extend AnyRef to avoid
runtime matches on the type for all operations on the array.


hail/src/main/scala/is/hail/io/plink/LoadPlink.scala
```diff
@@ -247,7 +247,7 @@ object MatrixPLINKReader {
     }
     assert(prevEnd == nVariants)
 
-    val contexts = cb.result()
+    val contexts = cb.result().map(r => r: Any)
```
johnc1231, line 250: Can't you just do a `asInstanceOf` here?

johnc1231 (approved):

PR #10511, opened 2021-05-21
[batch] Avoid CI deadlocks based on zone

jigold:
This PR fixes the problem where there were 3 free cores on an instance not in us-central. ci can not schedule jobs in other zones so ci could not progress.

### DIFF
--- a/batch/batch/driver/pool.py
+++ b/batch/batch/driver/pool.py
@@ -165,21 +165,6 @@ WHERE name = %s;
         if instance.state == 'active' and instance.failed_request_count <= 1:
             self.healthy_instances_by_free_cores.add(instance)
 
-    async def ready_cores_mcpu_per_user(self):
-        records = self.db.execute_and_fetchall(
-            '''
-SELECT user,
-  CAST(COALESCE(SUM(ready_cores_mcpu), 0) AS SIGNED) AS ready_cores_mcpu
-FROM user_inst_coll_resources
-WHERE inst_coll = %s
-GROUP BY user;
-''',
-            (self.name,),
-            timer_description=f'in compute_ready_cores for {self.name}: ready_cores_mcpu_per_user',
-        )
-        records = {r['user']: r['ready_cores_mcpu'] async for r in records}
-        return records
-
     async def create_instance(self, cores=None, max_idle_time_msecs=None, zone=None):
         if cores is None:
             cores = self.worker_cores
@@ -225,7 +210,7 @@ GROUP BY user;
             job_private=False,
         )
 
-    async def create_instances_from_ready_cores(self, ready_cores_mcpu):
+    async def create_instances_from_ready_cores(self, ready_cores_mcpu, zone=None):
         n_live_instances = self.n_instances_by_state['pending'] + self.n_instances_by_state['active']
 
         instances_needed = (ready_cores_mcpu - self.live_free_cores_mcpu + (self.worker_cores * 1000) - 1) // (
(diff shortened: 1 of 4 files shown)


batch/batch/utils.py
```diff
@@ -248,7 +248,7 @@ def __init__(self):
         self._global_counter = WindowFractionCounter(10)
 
     def push(self, success: bool):
-        self._global_counter.push('exceeded_shares', success)
+        self._global_counter.push(secrets.token_urlsafe(6), success)
```
danking, line 251: what's this about?

batch/batch/driver/pool.py
```diff
@@ -272,6 +280,7 @@ def __init__(self, app, pool):
         self.pool = pool
         self.async_worker_pool: AsyncWorkerPool = self.app['async_worker_pool']
         self.exceeded_shares_counter = ExceededSharesCounter()
+        self.ci_no_viable_instances = Box(0)
```
danking, line 283: Why do we need a box?

batch/batch/driver/pool.py
```diff
@@ -428,16 +437,25 @@ def get_instance(user, cores_mcpu):
...
+            if user == 'ci':
+                self.ci_no_viable_instances.value += 1
+
             return None
 
         should_wait = True
-        for user, resources in user_resources.items():
+        users = sorted(user_resources.keys(), key=lambda user: user != 'ci')
+        for user in users:
+            resources = user_resources[user]
```
danking, line 458: this all seems like a mixing of concerns, the scheduler shouldn't have to inform the scaler that CI needs more instances.

batch/batch/driver/pool.py
```diff
@@ -251,6 +252,13 @@ async def create_instances(self):
                 # parallelism will be bounded by thread pool
                 await asyncio.gather(*[self.create_instance() for _ in range(instances_needed)])
 
+        if self.scheduler.ci_no_viable_instances.value > 5 and self.live_free_cores_mcpu_by_zone[GCP_ZONE] == 0:
```
danking, line 255: I feel like the right fix is to retrieve ready cores by user, and then check here:
```
if ready_cores_mcpu_by_user['ci'] > 0 and self.live_free_cores_mcpu_by_zone[GCP_ZONE] == 0:
   ...
```

batch/batch/driver/pool.py
```diff
@@ -428,16 +437,25 @@ def get_instance(user, cores_mcpu):
                 instance = self.pool.healthy_instances_by_free_cores[i]
                 assert cores_mcpu <= instance.free_cores_mcpu
                 if user != 'ci' or (user == 'ci' and instance.zone.startswith('us-central1')):
```
danking, line 439: Just seeing this now: should this be GCP_ZONE?

danking (changes requested):

batch/batch/driver/pool.py
```diff
@@ -209,6 +225,27 @@ async def create_instance(self, cores=None, max_idle_time_msecs=None):
...
+            10,
+        )
+
+        if instances_needed > 0:
+            log.info(f'creating {instances_needed} new instances')
+            # parallelism will be bounded by thread pool
+            await asyncio.gather(*[self.create_instance() for _ in range(instances_needed)])
+
     async def create_instances(self):
         ready_cores = await self.db.select_and_fetchone(
```
danking, line 250: I'm inclined to hit the database once rather than twice. The per_user query is sufficient, we can compute the sum over all users in Python. Also, this SQL expression uses a LOCK IN SHARE MODE that has no effect (no transaction is present).

danking (dismissed):

batch/batch/utils.py
jigold: The interface for the `WindowFractionCounter` takes a key for idempotency. It was originally written for the case where we didn't want to count duplicate failures for the same operation ID. I'm reusing the code, but there isn't a "key" here. So I made up a random token as the key.

danking (approved):

PR #10512, opened 2021-05-22
[compiler] Code generation infrastructure for on-demand method splitting

tpoterba:


johnc1231:
Can we fix this up and try to get it in? I'm excited to do some ndarray experiments

tpoterba:
Yes, definitely! I'll assign you as a reviewer, though I'd like to also do a quick design review with Cotton to make sure I'm not doing anything crazy.

I don't think there are any major redesigns I have in mind, barring feedback from review. I think I'll make it possible to control the default split size using a flag / env var, but other than that it works and seems to improve performance.

hail/src/main/scala/is/hail/expr/ir/analyses/CompilationUnitSize.scala
```diff
@@ -0,0 +1,21 @@
+package is.hail.expr.ir.analyses
+
+import is.hail.expr.ir._
+
+object CompilationUnitSize {
```
johnc1231, line 5: This seems like it's not used anywhere?

johnc1231 (changes requested):
I'm not sure what CompilationUnitSize is for, but I otherwise think this looks good.

hail/src/main/scala/is/hail/expr/ir/analyses/CompilationUnitSize.scala
tpoterba: oh, good catch! This was used in an earlier version.

johnc1231: Ah ok. Feel free to dismiss my review after deleting it.

johnc1231 (approved):

PR #10515, opened 2021-05-24
[annotationdb] more informative error when can't find compatible annotation dataset

pwc2:
If a compatible annotation dataset can't be found in `index_compatible_version`, show the user the available versions and reference genome builds of the requested annotation dataset in the raised `ValueError`.


danking (changes requested):
beautiful

danking (approved):
whoops

PR #10520, opened 2021-05-25
[query] Add liftover test to catch RG initialization bug

tpoterba:


tpoterba:
This catches the problem -- you can try the diff you sent me Friday. Once this is merged, new version deploys will be blocked until a fix is in.

PR #10527, opened 2021-05-26
[datasets] modify `load_dataset` to handle s3/s3a

pwc2:
To allow users on GCP to access datasets that are only available on AWS (e.g. pan-ukb LD block matrices and tables). 

If trying to read dataset from `s3` path throws `FatalError: UnsupportedFileSystemException: No FileSystem for scheme "s3"`, then will read with the `s3a` prefix instead.


danking (approved):
huh, I thought I approve this already.

PR #10529, opened 2021-05-26
[hailtop.batch] allow either remote_tmpdir or bucket to be specified in ServiceBackend

danking:
This is a bit complicated now, but seems valuable for the users.

### DIFF
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -295,7 +295,7 @@ class ServiceBackend(Backend):
     Examples
     --------
 
-    >>> service_backend = ServiceBackend('my-billing-account', 'my-bucket') # doctest: +SKIP
+    >>> service_backend = ServiceBackend('my-billing-account', bucket='my-bucket') # doctest: +SKIP
     >>> b = Batch(backend=service_backend) # doctest: +SKIP
     >>> b.run() # doctest: +SKIP
     >>> service_backend.close() # doctest: +SKIP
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/batch/backend.py
```diff
@@ -309,18 +309,51 @@ class ServiceBackend(Backend):
...
     billing_project:
         Name of billing project to use.
     bucket:
-        Name of bucket to use.  Should not include the ``gs://``
-        prefix.
+        Name of bucket to use. Should not include the ``gs://`` prefix. Cannot be used with
+        remote_tmpdir. Temporary data will be stored in the "/batch" folder of this
+        bucket. Using this parameter as a positional argument is deprecated.
+    remote_tmpdir:
+        Temporary data will be stored in this google cloud storage folder. Cannot be used with bucket. Example: "gs://
```
jigold, line 329: I don't think this example is valid.

hail/python/hailtop/batch/backend.py
```diff
@@ -330,14 +363,19 @@ def __init__(self,
...
+            if bucket is None:
+                raise ValueError(
+                    'either the bucket or remote_tmpdir parameter of ServiceBackend '
+                    'must be set or run `hailctl config set batch/bucket MY_BUCKET`')
+            if 'gs://' in bucket:
+                raise ValueError(
+                    'The bucket parameter to ServiceBackend() should be a bucket name, not a path. '
+                    'Use the remote_tmpdir parameter to specify a path.')
+            remote_tmpdir = f'gs://{bucket}/batch'
+        self.remote_tmpdir = remote_tmpdir
```
jigold, line 378: I think we should assert `remote_tmpdir` starts with `gs://`

hail/python/hailtop/batch/backend.py
```diff
@@ -309,18 +309,51 @@ class ServiceBackend(Backend):
...
 
     def __init__(self,
-                 billing_project: str = None,
-                 bucket: str = None):
+                 *args,
+                 bucket: Optional[str] = None,
+                 remote_tmpdir: Optional[str] = None
+                 ):
+        billing_project: Optional[str]
+        if len(args) > 2:
```
jigold, line 339: I'm worried about the backwards compatibility of this change. This seems complicated. Why not just keep the interface the way it was before? Otherwise, the right interface seems to be `def __init__(self, *, billing_project=None, bucket = None, remote_tmpdir = None)`

hail/python/hailtop/batch/backend.py
```diff
@@ -309,18 +309,51 @@ class ServiceBackend(Backend):
...
+            bucket = args[1]
+            warnings.warn('Use of deprecated positional argument \'bucket\' in ServiceBackend(). Specify \'bucket\' as a keyword argument instead.')
+        elif len(args) == 1:
+            billing_project = args[0]
+        elif len(args) == 0:
+            billing_project = None
+
+        if remote_tmpdir is not None and bucket is not None:
+            raise ValueError('Cannot specify both remote_tmpdir and bucket in ServiceBackend()')
+        if remote_tmpdir is None and bucket is None:
```
jigold, line 354: What happened to trying to get the bucket from the user config here? I think this logic is further down below

jigold (changes requested):

hail/python/hailtop/batch/backend.py
danking: fixed

hail/python/hailtop/batch/backend.py
danking: removed

jigold (dismissed):
This looks good now! Do we need to add a test with the remote_tmpdir argument?

danking:
Neither bucket nor remote_tmpdir had tests so I added tests for both.

jigold (approved):

PR #10542, opened 2021-06-01
[compiler] Nested Array STypes and ETypes

chrisvittal:
This implements [Apache Arrow](https://arrow.apache.org/docs/format/Columnar.html#variable-size-list-layout) style nested arrays.

A more detailed write up will appear here as this gets closer to being ready.


chrisvittal:
Just putting this up for now.

danking:

PR #10543, opened 2021-06-01
[batch] validate that gcsfuse bucket and local paths are not empty

daniel-goldstein:


jigold (dismissed):

daniel-goldstein:
Sorry got the lint rule wrong, passes pylint locally now

jigold (approved):

PR #10545, opened 2021-06-02
[batch] Make error messages clearer in the UI and formatted correctly

jigold:


jigold:
I added the error for each container to the logs output. I think that's fine, but maybe we need separate section for errors?

daniel-goldstein:
I think the layout you have makes sense (grouping logs and errors under steps vs separate logs and errors sections), though I think it would be helpful to mark where the log ends and the error begins with subheaders, e.g.
# Main
## Log
## Errors

daniel-goldstein (changes requested):
I think the layout you have makes sense (grouping logs and errors under steps vs separate logs and errors sections), though I think it would be helpful to mark where the log ends and the error begins with subheaders, e.g.
# Main
## Log
## Errors

where there's only an error header if there was an error and only a log header if there's a log to view

daniel-goldstein (approved):

PR #10553, opened 2021-06-04
CHANGELOG: Dividing two ints should return a float64, not a float32

johnc1231:


johnc1231:
That `ret_type_f` function was defined in like 3 places, deleted the extras.

patrick-schultz (approved):

PR #10554, opened 2021-06-04
[query] Make ndarray from stream

johnc1231:
Previously, we always made the stream into an array, then copied the array into the ndarray. Now we just insert the stream directly into the ndarray. This saves memory (~7% on the linear regression rows benchmark), and had no noticeable effect on speed in my experiments.


tpoterba (changes requested):
Looks good. Should add simplify rules that strip out ToArray() and ToStream() nodes around the child, though.

tpoterba (approved):

PR #10555, opened 2021-06-04
Add hl.hadoop_scheme_supported function

nawatts:
Currently, the only way to check if Hail can read URLs with a given scheme (`gs`, `s3`, etc) is to attempt to read a URL with that scheme. However, the same exception type is thrown whether the scheme is not supported or the file doesn't exist or something else went wrong and the error message is the only way to determine what went wrong.

This adds a `hl.hadoop_scheme_supported` function, which returns a boolean indicating whether or not a URL scheme is supported.

Discussed on Zulip: https://hail.zulipchat.com/#narrow/stream/123010-Hail-0.2E2.20support/topic/Get.20supported.20URL.20schemes

@johnc1231

### DIFF
--- a/hail/python/hail/fs/google_fs.py
+++ b/hail/python/hail/fs/google_fs.py
@@ -139,3 +139,6 @@ class GoogleCloudStorageFS(FS):
         if self._is_local(path):
             rmtree(path)
         self.client.rm(path, recursive=True)
+
+    def supports_scheme(self, scheme: str) -> bool:
+        return scheme in ("gs", "")
--- a/hail/python/hail/fs/local_fs.py
+++ b/hail/python/hail/fs/local_fs.py
@@ -68,3 +68,6 @@ class LocalFS(FS):
 
     def rmtree(self, path: str):
         rmtree(path)
+
+    def supports_scheme(self, scheme: str) -> bool:
+        return scheme == ""
--- a/hail/python/hail/__init__.py
+++ b/hail/python/hail/__init__.py
@@ -59,7 +59,7 @@ from . import nd  # noqa: E402
 from hail.expr import aggregators as agg  # noqa: E402
 from hail.utils import (Struct, Interval, hadoop_copy, hadoop_open, hadoop_ls,  # noqa: E402
                         hadoop_stat, hadoop_exists, hadoop_is_file,
-                        hadoop_is_dir, copy_log)
+                        hadoop_is_dir, hadoop_scheme_supported, copy_log)
 
 from .context import (init, init_local, stop, spark_context, tmp_dir, default_reference,  # noqa: E402
                       get_reference, set_global_seed, _set_flags, _get_flags, current_backend,
@@ -94,6 +94,7 @@ __all__ = [
     'hadoop_stat',
     'hadoop_exists',
     'hadoop_ls',
+    'hadoop_scheme_supported',
     'copy_log',
     'Struct',
     'Interval',
--- a/hail/python/hail/docs/utils/index.rst
+++ b/hail/python/hail/docs/utils/index.rst
@@ -15,6 +15,7 @@ utils
     hadoop_is_dir
     hadoop_stat
     hadoop_ls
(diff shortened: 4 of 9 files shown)


hail/python/hail/fs/google_fs.py
```diff
@@ -139,3 +139,6 @@ def rmtree(self, path: str):
         if self._is_local(path):
             rmtree(path)
         self.client.rm(path, recursive=True)
+
+    def supports_scheme(self, scheme: str) -> bool:
+        return scheme in ("gs", "")
```
johnc1231, line 144: Should also support `file` as a scheme.

hail/python/hail/fs/local_fs.py
```diff
@@ -68,3 +68,6 @@ def remove(self, path: str):
 
     def rmtree(self, path: str):
         rmtree(path)
+
+    def supports_scheme(self, scheme: str) -> bool:
+        return scheme == ""
```
johnc1231, line 73: Should support `file`

johnc1231 (changes requested):
All file systems should support `file://` scheme. I'm not sure if it makes sense for this function to return true for empty string (which I read as meaning no specified scheme, and therefore read file from local disk). I think those should maybe just be `'file'`? I could be convinced it should support empty string and `'file'` as inputs if you want it to work on empty string.

hail/python/hail/fs/local_fs.py
nawatts: LocalFS uses `os` methods and `open`, which don't support `file://` URLs.

hail/python/hail/fs/google_fs.py
nawatts: For URLs that don't start with `gs://`, GoogleCloudStorageFS uses `os` methods and `open`, which don't support `file://` URLs.

nawatts:
Sorry, I should have explained the rationale for returning true for empty strings.

I was thinking that you should be able to do something like this:
```python
from urllib.parse import urlparse

url = "whatever:///path/to/file"
if hl.hadoop_supports_scheme(urlparse(url).scheme):
    with hl.hadoop_open(url) as f:
        ...
```

For paths without a scheme, the ParseResult's scheme is "".
```python
from urllib.parse import urlparse

urlparse("/path/to/file.txt").scheme
''
```

> All file systems should support file:// scheme.

HadoopFS is the only one that will actually work correctly if passed a `file://` URL. For example, both LocalFS and GoogleCloudStorageFS use `os.path.exists` to implement `exists`. On macOS at least, `os.path.exists("file:///test.txt")` looks for `test.txt` in a directory named `file:` in the current working directory, not the root directory.

> I'm not sure if it makes sense for this function to return true for empty string (which I read as meaning no specified scheme, and therefore read file from local disk). I think those should maybe just be 'file'? I could be convinced it should support empty string and 'file' as inputs if you want it to work on empty string.

For LocalFS and GoogleCloudStorageFS, an empty scheme will correspond to a path to a local file. For HadoopFS, an empty scheme causes Hadoop to use its configured default file system. When running Hail locally, this would be the local file system. On a Dataproc cluster, this would be HDFS. Returning true for the empty string in `HadoopFS.supports_scheme` is based on an assumption that there would always be a working default file system.

hail/python/hail/fs/local_fs.py
johnc1231: LocalFS supports it, tested:

```
import hail as hl
hl.init_local()
hl.import_vcf("file:///Users/johnc/Code/hail/hail/1kg/1kg.vcf.bgz")
```

johnc1231 (approved):

johnc1231:
Thanks for explaining and for adding this functionality, this all makes sense.

nawatts:
Linking to the Zulip discussion for possible future reference.

https://hail.zulipchat.com/#narrow/stream/123011-Hail-Dev/topic/file.20schemes

PR #10563, opened 2021-06-07
[hailtop] bounded_gather2 takes partial functions

cseed:
This supresses "coroutines not awaited" messages in the case that a
bounded gather is cancelled before all subtasks have had a chance to
start.


cseed:
I rebased.  This should be ready to go now.

jigold (approved):

PR #10566, opened 2021-06-07
[compiler] Generate method splits at stream consumers

tpoterba:
Helps a little with benchmarks in whole stage codegen (which I'm working on in a different branch)


johnc1231:
Did you do a benchmark on this one vs main?

tpoterba:
Yeah, basically no change.

johnc1231 (approved):

PR #10569, opened 2021-06-08
[compiler] New Array PType

chrisvittal:
The purpose of this series is to create a new PArray type that will
eventually replace PCanonicalArray. It is represented as a C struct with
the following layout (T is the element type):

```
struct {
    int length;
    T *elements;
    char *missing; // present only if T is optional
}
```

This puts the PType used for arrays more in line with
the values in SIndexablePointerValue.


chrisvittal:
There are enough changes here that I'm going to make a new PType to start, set everything to always use it, then rename PCanonicalArray.

chrisvittal:
reworking approach to this. now unnecessary.

PR #10571, opened 2021-06-09
[query] Fix lowered read with intervals

tpoterba:


hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
```diff
@@ -182,8 +186,9 @@ case class SplitPartitionNativeWriter(
     context: EmitCode,
     region: Value[Region]): IEmitCode = {
     val keyType = ifIndexed { index.get._2 }
-    val iAnnotationType = PCanonicalStruct(required = true, "entries_offset" -> PInt64Required)
+    val iAnnotationType = PCanonicalStruct(required = true, "entries_offset" -> PInt64())
```
chrisvittal, line 189: Why this change? What was broken about the types of the annotations here?

tpoterba: This doesn't match the IndexSpec.defaultAnnotation, which is used to generate the types written into the metadata. Everywhere else (interpreted paths + staged metadata), the entries_offset field is optional. This change makes it consistent.

I suspect we could change them all to required as well, or thread the types through, but this is a simple fix for right now.

chrisvittal (approved):

PR #10575, opened 2021-06-09
[compiler] Relegate PType.deepInnerRequired to tests

chrisvittal:
It's only used in tests, and is not something we wish to maintain as part of the ptype interface.

### DIFF
--- a/hail/src/test/scala/is/hail/expr/ir/RequirednessSuite.scala
+++ b/hail/src/test/scala/is/hail/expr/ir/RequirednessSuite.scala
@@ -105,7 +105,7 @@ class RequirednessSuite extends HailSuite {
       MakeTuple.ordered(FastIndexedSeq(I32(5), Str("bar"))))
 
     allRequired.foreach { n =>
-      nodes += Array(n, PType.canonical(n.typ, required).deepInnerRequired(required))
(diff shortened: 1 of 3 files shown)


hail/src/test/scala/is/hail/expr/ir/RequirednessSuite.scala
```diff
@@ -519,3 +519,22 @@ class RequirednessSuite extends HailSuite {
     assert(actual == expected)
   }
 }
+
+object RequirednessSuite {
+  def deepInnerRequired(t: PType, required: Boolean): PType =
+    t match {
+      case t: PArray => PCanonicalArray(deepInnerRequired(t.elementType, true), required)
```
tpoterba, line 526: can we change this while we're here? It's wrong -- should match on the same types (PCArray, PCSet, PCDict, PCStruct.

tpoterba (changes requested):

hail/src/test/scala/is/hail/expr/ir/RequirednessSuite.scala
chrisvittal: of course

tpoterba (approved):

PR #10579, opened 2021-06-10
[compiler] Rework SIndexablePointer to be a multi-code SCode

chrisvittal:
The new implementation is comprised of three (two if the element type is required) codes:
```
length: Code[Int]
elements: Code[Long]
missing: Option[Code[Long]] // if None if element is required
```

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -1231,7 +1231,7 @@ class Emit[C](
             cb.assign(grpIdx, grpIdx + 1)
           })
 
-          dictType.construct(finishOuter(cb))
+          dictType.sType.fromCodes(finishOuter(cb).makeCodeTuple(cb))
         }
 
       case x@StreamLen(a) =>
--- a/hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
+++ b/hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
@@ -30,7 +30,7 @@ final case class PCanonicalArray(elementType: PType, required: Boolean = false) 
 
   override val byteSize: Long = 8
 
-  def setRequired(required: Boolean) = if (required == this.required) this else PCanonicalArray(elementType, required)
+  override def setRequired(required: Boolean): PCanonicalArray = if (required == this.required) this else PCanonicalArray(elementType, required)
 
   def loadLength(aoff: Long): Int =
     Region.loadInt(aoff)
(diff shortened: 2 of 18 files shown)


chrisvittal:
Once again totally broken until we can return code tuples with more than one code from `invokePCode`

tpoterba (changes requested):
This change is hard to read with both rename and semantic changes. Can you separate those and do the semantic changes first, then the rename after?

chrisvittal:
Un-renamed

tpoterba:
thanks, much easier to read.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -1231,7 +1231,7 @@ class Emit[C](
             cb.assign(grpIdx, grpIdx + 1)
           })
 
-          dictType.construct(finishOuter(cb))
```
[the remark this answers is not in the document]
tpoterba: actually, I think it's `coerceArrayCode`.

hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
```diff
@@ -377,25 +377,31 @@ final case class PCanonicalArray(elementType: PType, required: Boolean = false)
...
+  def loadCheapPCode(cb: EmitCodeBuilder, addr: Code[Long]): SIndexablePointerCode = {
+    val ptr = cb.newLocal[Long]("load_array_addr", addr)
+    val len = cb.newLocal[Int]("load_array_len", loadLength(ptr))
+    val missing = if (elementRequired) None else Some(ptr + lengthHeaderBytes)
+    new SIndexablePointerCode(sType, len, firstElementOffset(ptr, len), missing)
+  }
 
   def storeContentsAtAddress(cb: EmitCodeBuilder, addr: Value[Long], region: Value[Region], indexable: SIndexableValue, deepCopy: Boolean): Unit = {
     val length = indexable.loadLength()
+    val array = indexable.asInstanceOf[SIndexablePointerSettable]
```
tpoterba, line 391: This is invalid. Can't cast before the match.

tpoterba: (just move it inside I think)

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SIndexablePointer.scala
```diff
@@ -22,39 +21,59 @@ case class SIndexablePointer(pType: PContainer) extends SContainer {
   def elementEmitType: EmitType = EmitType(elementType, pType.elementType.required)
 
   def coerceOrCopy(cb: EmitCodeBuilder, region: Value[Region], value: SCode, deepCopy: Boolean): SCode = {
-    new SIndexablePointerCode(this, pType.store(cb, region, value, deepCopy))
+    value match {
+      case value: SIndexablePointerCode if !deepCopy => new SIndexablePointerCode(this, value.length, value.elements, value.missing)
```
tpoterba, line 25: This is invalid. Types might not match.

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SIndexablePointer.scala
```diff
@@ -22,39 +21,59 @@ case class SIndexablePointer(pType: PContainer) extends SContainer {
...
   }
 
   def codeTupleTypes(): IndexedSeq[TypeInfo[_]] = FastIndexedSeq(LongInfo)
 
   override def settableTupleTypes(): IndexedSeq[TypeInfo[_]] = FastIndexedSeq(LongInfo, IntInfo, LongInfo)
 
   def fromSettables(settables: IndexedSeq[Settable[_]]): SIndexablePointerSettable = {
-    val IndexedSeq(a: Settable[Long@unchecked], length: Settable[Int@unchecked], elementsAddress: Settable[Long@unchecked]) = settables
-    assert(a.ti == LongInfo)
+    val (length, elements, missing) = settables match {
```
tpoterba, line 37: Should assert that the missingness option-status is what we expect.

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SIndexablePointer.scala
```diff
@@ -22,39 +21,59 @@ case class SIndexablePointer(pType: PContainer) extends SContainer {
...
-    new SIndexablePointerCode(this, pType.store(cb, region, value, deepCopy))
+    value match {
+      case value: SIndexablePointerCode if !deepCopy => new SIndexablePointerCode(this, value.length, value.elements, value.missing)
+      case _ =>
+        val addr = pType.store(cb, region, value, deepCopy)
+        pType.loadCheapPCode(cb, addr)
+    }
   }
 
   def codeTupleTypes(): IndexedSeq[TypeInfo[_]] = FastIndexedSeq(LongInfo)
```
tpoterba, line 32: needs update

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SIndexablePointer.scala
```diff
@@ -22,39 +21,59 @@ case class SIndexablePointer(pType: PContainer) extends SContainer {
...
 
   override def settableTupleTypes(): IndexedSeq[TypeInfo[_]] = FastIndexedSeq(LongInfo, IntInfo, LongInfo)
```
tpoterba, line 34: needs update

tpoterba (changes requested):
Some comments, still thinking through the baseAddress stuff.

chrisvittal (commented):
baseAddress is sound if the components came from a PType with the layout of a PCanonicalArray

chrisvittal:
Compared to the baseAddress stuff, a more exotic change would be to change the layout of a PCanonicalArray, as well as update the notion of a 'pointer' to a PCanonicalArray.

The result would be something like [sds](https://github.com/antirez/sds) which is a dynamic string library, where the public api works on char pointers, and all the metadata is kept in a prefix to that pointer.

chrisvittal:
I plan on reworking this entirely.

PR #10581, opened 2021-06-10
[compiler] Use AsmTuple for multi-code SType return values

chrisvittal:
This removes the single code requirement (st.nCodes == 1) for methods
returning SCodes.

### DIFF
--- a/hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
+++ b/hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
@@ -76,6 +76,8 @@ class ClassesBytes(classesBytes: Array[(String, Array[Byte])]) extends Serializa
 }
 
 class AsmTuple[C](val cb: ClassBuilder[C], val fields: IndexedSeq[Field[_]], val ctor: MethodBuilder[C]) {
+  val ti: TypeInfo[_] = cb.ti
+
   def newTuple(elems: IndexedSeq[Code[_]]): Code[C] = Code.newInstance(cb, ctor, elems)
 
   def loadElementsAny(t: Value[_]): IndexedSeq[Code[_]] = fields.map(_.get(coerce[C](t) ))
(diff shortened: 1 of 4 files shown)


chrisvittal:
This the branch that involves the AsmTuple is not taken (obviously), but will be with the changes from #10579

tpoterba (changes requested):
I think before we merge this, we need to have a sense of the performance impact. Just running the benchmarks on the SIndexablePointer changes will include performance changes from the new representation and the asmtuple change, so isolating it here sounds like a good idea. Let's change this PR so all pcode methods go through an asm tuple (even with 1 code) and see how affects benchmark timings.

chrisvittal:
Okay, we can benchmark now.

tpoterba:
submitting benchmarks now.

chrisvittal:
I should probably fix things first.

chrisvittal:
Getting IncompatibleClassChangeError and I don't know why.

chrisvittal:
Pushed up latest version.

hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
```diff
@@ -108,18 +110,29 @@ class ModuleBuilder() {
...
+        val L = new lir.Block()
+        L.append(
+          lir.methodStmt(INVOKESPECIAL,
+            "java/lang/Object",
+            "<init>",
+            "()V",
+            false,
+            UnitInfo,
+            FastIndexedSeq(lir.load(ctor._this.asInstanceOf[LocalRef[_]].l))))
+        cb += new VCode(L, L, null)
```
chrisvittal, line 129: This is messy, but I'm not what sure the best way to improve it is.

chrisvittal (commented):
One thing I'd like to maybe change after benchmarking and restoring the `nCodes == 1` optimizations.

chrisvittal:
Pretty sure this is ready for review

tpoterba:
oh, sorry, didn't pop up on my to-do since my old review hadn't been dismissed. looks good!

tpoterba (approved):

PR #10582, opened 2021-06-10
[compiler] Rename almost every reference from PCode to SCode

chrisvittal:
There may be a few varables with the name `pc` or `pv` around, but with
this, PCode should be history (for now).


tpoterba:
rebase and remove stacked label, this should be good to go

chrisvittal:
In progress.

chrisvittal:
done

tpoterba (approved):

PR #10583, opened 2021-06-11
[query] tensor index expression code generator

patrick-schultz:
Add a codegen method `SNDArray.coiterate`, with signature
```
def coiterate(cb: EmitCodeBuilder, region: Value[Region], indexVars: IndexedSeq[String], arrays: IndexedSeq[(SNDArrayCode, IndexedSeq[Int], String)], body: IndexedSeq[SSettable] => Unit, deepCopy: Boolean): Unit
```
For example, the index expression `A[i, j] += B[j]` would be written
```
coiterate(cb, region, Seq("i", "j"), Seq((A, Seq(0, 1), "A"), (B, Seq(1), "B")), {
  case Seq(a, b) => cb.assign(a, SCode.add(cb, a, b))
})
```
This generates a loop nest, with one loop per variable in `indexVars`. The inner loop is `indexVars(0)`, so that column major traversal is when index variables are increasing, as in `(A, Seq(0, 1), "A")`. Each index variable iterates over a dimension, with the size of the dimension inferred from its use. In the inner loop, each index variable `i0, i1, ...` has a value; `body` is run, with each of the `SSettable`s bound to an element of the corresponding input in `arrays`. For example, if the first element of `arrays` is `(A, IndexedSeq(1, 3), "A")`, then the `SSettable` will be the element of `A` at index `(i1, i3)`. However, it avoids computing the address of each element from the indices from scratch in the inner loop.

This was motivated by the need to generate operations on ndarrays in the local whitening aggregator. I replaced a few uses of `forEachIndex` with `coiterate`, which may give a performance boost since it avoids index math in the inner loop.

### DIFF
--- a/hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
+++ b/hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
@@ -92,7 +92,7 @@ object SNDArray {
         }
       }
       val strides = array.strides(cb)
-      val pos = Array.tabulate(array.st.nDims) { i => cb.newLocal[Long](s"$name$i") }
+      val pos = Array.tabulate(array.st.nDims + 1) { i => cb.newLocal[Long](s"$name$i") }
       val elt = new SSettable {
         def st: SType = array.st.elementType
         val pt: PType = array.st.pType.elementType
@@ -117,7 +117,7 @@ object SNDArray {
             if (info(n).indexToDim.contains(idx)) {
               val i = info(n).indexToDim(idx)
               // FIXME: assumes array's indices in ascending order
-              cb.assign(info(n).pos(i), if (i == info(n).array.st.nDims - 1) info(n).array.firstDataAddress(cb) else info(n).pos(i+1))
+              cb.assign(info(n).pos(i), info(n).pos(i+1))
             }
           }
         }
@@ -135,6 +135,9 @@ object SNDArray {
       }
     }
 
+    for (n <- arrays.indices) {
+      cb.assign(info(n).pos(info(n).array.st.nDims), info(n).array.firstDataAddress(cb))
+    }
     recurLoopBuilder(indexVars.length - 1)
   }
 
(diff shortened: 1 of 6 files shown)


johnc1231:
Very cool! You're failing memory tests though since the tests run with the debug allocator

patrick-schultz:
Do java tests run with the debug allocator? I'm just getting a segfault.

johnc1231:
I think they're supposed to? The python tests definitely do.

patrick-schultz:
Tests are passing now

johnc1231:
You have a few `FIXME`s in here, are those for the future?

patrick-schultz:
Yeah, they're the places that bake in the assumption that each array is indexed in increasing (i.e. column-major) order. That shouldn't be hard to generalize, which would let this generate code for things like transposing, which have to iterate at least one array in non-column-major order. But I thought I'd leave that till we need it.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
```diff
@@ -41,6 +44,104 @@ object SNDArray {
...
+      val elt = new SSettable {
+        def st: SType = array.st.elementType
+        val pt: PType = array.st.pType.elementType
+
+        // FIXME: need to use `pos` of smallest index var
+        def get: SCode = pt.loadCheapPCode(cb, pt.loadFromNested(pos(0)))
+        def store(cb: EmitCodeBuilder, v: SCode): Unit = pt.storeAtAddress(cb, pos(0), region, v, deepCopy)
+        def settableTuple(): IndexedSeq[Settable[_]] = FastIndexedSeq(pos.last)
+      }
+      val indexToDim = Map(indices.indices.map(i => indices(i) -> i): _*)
```
johnc1231, line 106: This is `indices.zipWithIndex`, isn't it?

johnc1231 (changes requested):

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
johnc1231: Like, `indexToDim = Map(indices.zipWithIndex(): _*)`

patrick-schultz: Ah, yep. I'll change it.

johnc1231 (approved):

PR #10588, opened 2021-06-11
[batch] use correct quota for preemptible local ssd

danking:
The LOCAL_SSD_TOTAL_GB has nothing to do with *preemptible* local ssds.

### DIFF
--- a/batch/batch/driver/zone_monitor.py
+++ b/batch/batch/driver/zone_monitor.py
@@ -96,7 +96,7 @@ class ZoneMonitor:
 
             remaining = quota_remaining['PREEMPTIBLE_CPUS'] / worker_cores
             if worker_local_ssd_data_disk:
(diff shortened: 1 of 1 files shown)


jigold:
For reference: https://cloud.google.com/compute/quotas#quotas_for_preemptible_resources

jigold:
I think we need to address the quotas first for hail-vdc before we can merge this.

danking:
OK, update from Google: they suggest we check if the preemptible quota is non-zero and assume that if it is non-zero preemptible is in use and if it is zero normal quota is in use. In very rare cases, people can increase and then reduce their preemptible quota and Hail will not work properly. Google wasn't interested in providing an API to detect this case. I'll change this PR accordingly sometime next week.

danking:
closing until I can get back to this.

PR #10591, opened 2021-06-14
[aiotools] Fix out of memory bug when using writable streams

jigold:
The original problem we were seeing was this error message:

```
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7fe17e8c4bd0>
  File "/usr/local/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.7/site-packages/batch/copy/__main__.py", line 34, in <module>
    asyncio.run(main())
  File "/usr/local/lib/python3.7/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/usr/local/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
concurrent.futures._base.CancelledError
sys:1: RuntimeWarning: coroutine 'retry_transient_errors' was never awaited
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

I'm not sure why we didn't get a better error message from the ThreadPoolExecutor, but this fix definitely solves the OOM issue. We were assuming `close` in Python means the entire write has completed, but that's not true. `close` just means the file handle in Python has been closed, but the data is still stored in the kernel's write buffer until it gets a chance to have written the data. When the pd-ssds were slow (either network bandwidth or ext3/4 is slow) and we're trying to write a lot of data, this meant we were building up lots of data in the write buffer for previously "completed" tasks and the semaphore was not doing its job to limit the number of "active" tasks to 50.

There may still be another error here that we're not retrying, but I wasn't able to replicate the bug after making this fix with 5 replicas of downloading 80 Gi x 8 jobs per node.

### DIFF
--- a/hail/python/hailtop/aiotools/stream.py
+++ b/hail/python/hailtop/aiotools/stream.py
@@ -2,6 +2,7 @@ from typing import BinaryIO, Optional, Tuple, Type
 from types import TracebackType
 import abc
 import io
+import os
 from concurrent.futures import ThreadPoolExecutor
 import janus
 from hailtop.utils import blocking_to_async
@@ -31,7 +32,7 @@ class ReadableStream(abc.ABC):
                 self._waited_closed = True
 
     @property
-    def closed(self) -> None:
+    def closed(self) -> bool:
         return self._closed
 
     async def __aenter__(self) -> 'ReadableStream':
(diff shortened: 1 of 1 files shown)


cseed (approved):
Great work.  Very subtle, but the analysis totally makes sense!

PR #10597, opened 2021-06-16
[batch] Include currently running batches by default in billing UI

jigold:


danking (approved):
very useful

batch/sql/estimated-current.sql
```diff
@@ -152,6 +152,7 @@ CREATE INDEX `batches_user_state_cancelled` ON `batches` (`user`, `state`, `canc
 CREATE INDEX `batches_deleted` ON `batches` (`deleted`);
 CREATE INDEX `batches_token` ON `batches` (`token`);
 CREATE INDEX `batches_time_completed` ON `batches` (`time_completed`);
+CREATE INDEX `batches_time_closed` ON `batches` (`time_closed`);
```
danking, line 155: I don't think I buy that this index will improve performance. I assume you're hoping for an index merge where we fetch rows matching the left-hand-side of the OR from one index and rows matching the RHS with another index?

Seems reasonable, but it's not obvious to me that `time_closed IS NOT NULL` is a better choice than `time_completed IS NULL`. Almost all batches satisfy `time_closed IS NOT NULL`. Relatively few batches should satisfy `time_completed IS NULL` (basically: running batches and batches that are left open).

danking (changes requested):

danking (approved):

PR #10600, opened 2021-06-17
[prometheus] make middleware function instead of per-route decorator

daniel-goldstein:
We currently have a `@monitor_endpoint` decorator that we use to wrap aiohttp endpoints and produce prometheus metrics, but the same result can be achieved with fewer lines of code by using essentially the same wrapper as an aiohttp middleware. 

Separately, I fixed the wrapper to correctly catch and report on endpoints that raise exceptions, as well as keeping track of the number of concurrent connections.

### DIFF
--- a/gear/gear/metrics.py
+++ b/gear/gear/metrics.py
@@ -1,21 +1,24 @@
-from functools import wraps
+from aiohttp import web
 import prometheus_client as pc  # type: ignore
 from prometheus_async.aio import time as prom_async_time  # type: ignore
 
 REQUEST_TIME = pc.Summary('http_request_latency_seconds', 'Endpoint latency in seconds', ['endpoint', 'verb'])
 REQUEST_COUNT = pc.Counter('http_request_count', 'Number of HTTP requests', ['endpoint', 'verb', 'status'])
+CONCURRENT_REQUESTS = pc.Gauge('http_concurrent_requests', 'Number of in progress HTTP requests', ['endpoint', 'verb'])
 
 
-def monitor_endpoint(handler):
-    @wraps(handler)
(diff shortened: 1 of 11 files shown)


daniel-goldstein:
bump

gear/gear/metrics.py
```diff
@@ -1,21 +1,24 @@
...
-            REQUEST_TIME.labels(endpoint=endpoint, verb=verb), handler(request, *args, **kwargs)
-        )
+@web.middleware
+async def monitor_endpoints_middleware(request, handler):
+    # Use the path template given to @route.<METHOD>, not the fully resolved one
+    endpoint = request.match_info.route.resource.canonical
+    verb = request.method
+    CONCURRENT_REQUESTS.labels(endpoint=endpoint, verb=verb).inc()
+    try:
+        response = await prom_async_time(REQUEST_TIME.labels(endpoint=endpoint, verb=verb), handler(request))
```
jigold, line 17: What happened to *args and **kwargs?

jigold (changes requested):

gear/gear/metrics.py
daniel-goldstein: `*args` and `**kwargs` were for when this was decorating arbitrary functions of ours. Functions decorated with `web.middleware` aren't given those bare functions but strictly a `Request` and a `Handler`, an `aiohttp` type that encapsulates the call to the endpoint. I haven't looked through the code but I think it's doing something like:

```python
handler = lambda request: foo(request, *args, **kwargs)
```

before handing `handler` off to middleware

jigold (approved):

PR #10603, opened 2021-06-21
[hailtop.batch] Add ability to unset default resource values

jigold:
CHANGELOG: Added the ability to unset batch-level default values and use the service defaults instead.


daniel-goldstein (dismissed):
Could you add to the documentation what it means to set the resources to `None`? Wouldn't want someone to misinterpret `storage=None` as no storage or unlimited

daniel-goldstein (approved):

PR #10606, opened 2021-06-21
[query] Fix the branches in logreg sigmoid

johnc1231:
A user reviewing the code pointed out that the branches of this function were backwards. Is there any easy way to construct a test that would catch this? Is `sigmoid` worth wrapping up into like `hl.math.sigmoid` or something? It seems like the name `sigmoid` refers to many things, maybe it should be `expit` like it apparently is in R and scipy?

### DIFF
--- a/hail/python/hail/docs/functions/index.rst
+++ b/hail/python/hail/docs/functions/index.rst
@@ -99,11 +99,13 @@ These functions are exposed at the top level of the module, e.g. ``hl.case``.
     bit_rshift
     bit_not
     exp
+    expit
     is_nan
     is_finite
     is_infinite
     log
     log10
+    logit
     sign
     sqrt
     int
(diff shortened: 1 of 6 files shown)


patrick-schultz:
Oof, good catch! The thing we're trying to avoid is `e^x` overflowing for large positive `x`. In double precision, the smallest `x` that overflows is 710. So to test that we handle overflow correctly, you can check `sigmoid(710) == 1.0` and `sigmoid(-710) == 0.0` (using approximate equality).

Actually, after playing with this, if you just use the simple definition `sigmoid(x) = 1 / (1 + np.exp(-x))`, then `sigmoid(-710)` does overflow, but it returns the right answer since `np.exp(710)` returns `inf`, and `1 / inf == 0.0`. But `math.exp(710)` throws an exception. `hl.exp` seems to have the numpy behavior, so I think the simple version actually works. But we should add the above test.

I think wrapping this in an exposed function is a good idea. I agree it should be called `expit`, both for consistency with scipy, and because as you say, `sigmoid` really just means an S shaped function. And if we do expose `expit`, we should probably expose its inverse `logit` too.

patrick-schultz (dismissed):
See comment

patrick-schultz (approved):
Thanks John!

PR #10610, opened 2021-06-22
[query] Impute Hail type for frozensets

nawatts:
Currently, Hail cannot impute a Hail type for frozenset objects.

```python
hl.literal(frozenset(["foo"]))
# ExpressionException: Hail cannot automatically impute type of <class 'frozenset'>: frozenset({'foo'})
```

This is an issue because frozenset is Hail's Python counterpart to SetExpression. Since Hail outputs frozensets from functions like `eval` and `collect`, users are likely to expect that Hail can also accept them as input.

```python
hl.eval(hl.set([1,2,3]))
# frozenset({1, 2, 3})

hl.set([1,2,3]).collect()
# [frozenset({1, 2, 3})]
```

One example use case where this issue pops up is annotating a table with all possible values of some field. Currently, the output of aggregating using `hl.agg.collect_as_set` cannot be directly used to annotate a table, it must be explicitly cast to a set.

```python
values = ds.aggregate(hl.agg.collect_as_set(ds.field))
ds.annotate_globals(values=values)
# TypeError: annotate_globals: keyword argument 'values': expected expression of type any, found frozenset: frozenset({0, 1, 2})

ds.annotate_globals(values=set(values))
```

### DIFF
--- a/hail/python/hail/expr/expressions/base_expression.py
+++ b/hail/python/hail/expr/expressions/base_expression.py
@@ -4,7 +4,7 @@ import numpy as np
 import hail
 import hail as hl
 from hail.expr import expressions
-from hail.expr.types import HailType, is_numeric, is_compound, tint32, \
+from hail.expr.types import HailType, is_numeric, is_compound, is_setlike, tint32, \
     tint64, tfloat32, tfloat64, tstr, tbool, tarray, \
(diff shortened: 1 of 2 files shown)


johnc1231 (approved):
This is a great change, thank you.

PR #10613, opened 2021-06-24
[utils] Fix transient error for rateLimitExceeded

jigold:
```
body='{\\n \"error\": {\\n \"code\": 403,\\n \"message\": \"Quota exceeded for quota group \\'default\\' and limit \\'Queries per user per 100 seconds\\' of service \\'compute.googleapis.com\\' for consumer \\'project_number:859893752941\\'.\",\\n \"errors\": [\\n {\\n \"message\": \"Quota exceeded for quota group \\'default\\' and limit \\'Queries per user per 100 seconds\\' of service \\'compute.googleapis.com\\' for consumer \\'project_number:859893752941\\'.\",\\n \"domain\": \"usageLimits\",\\n \"reason\": \"rateLimitExceeded\"\\n }\\n ],\\n \"status\": \"PERMISSION_DENIED\"\\n }\\n}
```

### DIFF
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -559,7 +559,7 @@ def is_transient_error(e):
         # 503 service unavailable, 504 gateway timeout
         return True
     if isinstance(e, hailtop.httpx.ClientResponseError) and (
-            e.status == 403 and 'Rate Limit Exceeded' in e.body):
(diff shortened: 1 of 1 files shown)


jigold:
I'm not sure this is correct; the body might be JSON rather than a string.

danking (approved):

PR #10614, opened 2021-06-24
[query] add SNDArraySlice

patrick-schultz:
I pulled this out of my local whitening branch. It's being tested there, but I'm not sure the best way to write unit tests. Do we have any similar tests for STypes?

### DIFF
--- a/hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
+++ b/hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
@@ -3,13 +3,9 @@ package is.hail.types.physical.stypes.concrete
 import is.hail.annotations.Region
 import is.hail.asm4s._
 import is.hail.expr.ir.EmitCodeBuilder
-import is.hail.types.physical.stypes.interfaces.{SBaseStructCode, SNDArray, SNDArrayCode, SNDArrayValue}
-import is.hail.asm4s.{Code, IntInfo, LongInfo, Settable, SettableBuilder, TypeInfo, Value, const}
-import is.hail.expr.ir.orderings.CodeOrdering
-import is.hail.expr.ir.{EmitCodeBuilder, EmitMethodBuilder, SortOrder}
 import is.hail.types.physical.stypes.interfaces.{SBaseStructCode, SNDArray, SNDArrayCode, SNDArraySettable, SNDArrayValue}
-import is.hail.types.physical.stypes.{SCode, SSettable, SType, SValue}
-import is.hail.types.physical.{PCanonicalNDArray, PType}
+import is.hail.types.physical.stypes.{SCode, SType, SValue}
+import is.hail.types.physical.{PCanonicalNDArray, PPrimitive, PType}
 import is.hail.types.virtual.Type
 import is.hail.utils.{FastIndexedSeq, toRichIterable}
 
@@ -77,9 +73,14 @@ class SNDArrayPointerSettable(
  ) extends SNDArraySettable {
   val pt: PCanonicalNDArray = st.pType
 
+  def loadElementAddress(indices: IndexedSeq[Value[Long]], cb: EmitCodeBuilder): Code[Long] = {
+    assert(indices.size == pt.nDims)
+    pt.loadElementFromDataAndStrides(cb, indices, dataFirstElement, strides)
+  }
+
   def loadElement(indices: IndexedSeq[Value[Long]], cb: EmitCodeBuilder): SCode = {
     assert(indices.size == pt.nDims)
-    pt.elementType.loadCheapSCode(cb, pt.loadElementFromDataAndStrides(cb, indices, dataFirstElement, strides))
+    pt.elementType.loadCheapSCode(cb, loadElementAddress(indices, cb))
   }
 
   def settableTuple(): IndexedSeq[Settable[_]] = FastIndexedSeq(a) ++ shape ++ strides ++ FastIndexedSeq(dataFirstElement)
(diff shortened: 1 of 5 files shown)


johnc1231:
I'm a little worried that the LAPACK calls we do currently for QR, SVD, etc. might be incompatible with slices. Not a hard fix, but all of the stride arguments to those mostly just inspect the shape, which is no longer a safe thing to do.

johnc1231:
I'm thinking of things like: https://github.com/hail-is/hail/blob/78e0caa8850cd79733d8b0e84e037cf2c1ebaf5e/hail/src/main/scala/is/hail/expr/ir/Emit.scala#L1564-L1565

patrick-schultz:
Good point. In my branch I've written wrappers for the LAPACK methods that take SNDArrays and determine the leading dimension arguments. I can pull those into this PR and replace all the LAPACK calls with the wrappers.

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
```diff
@@ -66,9 +65,19 @@ class SNDArrayPointerSettable(
...
+    pt.loadElementFromDataAndStrides(cb, indices, dataFirstElement, strides)
+  }
+
   def loadElement(indices: IndexedSeq[Value[Long]], cb: EmitCodeBuilder): SCode = {
     assert(indices.size == pt.nDims)
-    pt.elementType.loadCheapSCode(cb, pt.loadElementFromDataAndStrides(cb, indices, dataFirstElement, strides))
+    pt.elementType.loadCheapSCode(cb, loadElementAddress(indices, cb))
+  }
+
+  def setElement(indices: IndexedSeq[Value[Long]], value: SCode, cb: EmitCodeBuilder): Unit = {
```
tpoterba, line 78: do we still need setElement after coiterateMutate?

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
```diff
@@ -180,24 +185,134 @@ trait SNDArray extends SType {
...
+    }
+
+    val newFirstDataAddress = loadElementAddress(firstElementIndices, cb)
+
+    val newSType = SNDArraySlice(PCanonicalNDArray(st.pType.elementType, newShape.size, st.pType.required))
+
+    new SNDArraySliceCode(newSType, newShape, newStrides, newFirstDataAddress)
+  }
+
+  def slice(cb: EmitCodeBuilder, indices: Any*): SNDArraySliceCode = {
```
tpoterba, line 287: I'm not a huge fan of this helper function -- we don't need the syntactic sugar in the compiler. I weight clarity highly above terseness in the backend, and I think it'll be easier to maintain the single slice function
```
  def slice(cb: EmitCodeBuilder, indices: IndexedSeq[NDArrayIndex]): SNDArraySliceCode = {
```

tpoterba (changes requested):
Sorry for leaving this hanging! A couple individual comments, but I want to also have a discussion about how a slice is actually different from a "canonical" ndarray, especially after #10619.

I think of `SNDArrayPointer` as a pointer to data, and variables for shape/strides. Isn't that exactly what a slice is, too? Where are the important differences? I'd think copy semantics are one, but are there others?

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
patrick-schultz: I definitely understand the concern. But a method like local whitening that's implemented using our codegen abstractions blurs the line between "in the compiler" and out. It's a complicated method that's already difficult to read and reason about in Julia, and will much harder to follow in Hail even with sugar like this.

I can drop it from this PR, and we can weigh the benefits later with some before/after code samples from the local whitening implementation.

patrick-schultz:
> I want to also have a discussion about how a slice is actually different from a "canonical" ndarray, especially after #10619.
> 
> I think of `SNDArrayPointer` as a pointer to data, and variables for shape/strides. Isn't that exactly what a slice is, too? Where are the important differences? I'd think copy semantics are one, but are there others?

I think the main difference is that `SNDArrayPointer` points to the location in memory that stores shape/strides as well as the data, while `SNDArraySlice` points directly to the first element, with shape/strides living only on the stack. #10619 doesn't look like it changes that, though perhaps after that change there's no need for `SNDArrayPointer` any more, and we can only use `SNDArraySlice`. IIRC, MLIR uses basically that design.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
tpoterba: OK, sounds good re: looking at the code samples with local whitening. And I definitely do see the need for ergonomics in the compiler!

patrick-schultz:
@johnc1231 After talking to Tim, I decided to put off ensuring the rest of the compiler is compatible with slices, and just make sure local whitening is the only user for now.

@tpoterba I think I've addressed all your comments.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
```diff
@@ -240,24 +243,46 @@ trait SNDArray extends SType {
...
-  def sameShape(other: SNDArrayValue, cb: EmitCodeBuilder): Code[Boolean]
+  def assertInBounds(indices: IndexedSeq[Value[Long]], cb: EmitCodeBuilder, errorId: Int): Code[Unit] = {
+    val shape = this.shapes(cb)
+    Code.foreach(0 until st.nDims) { dimIndex =>
+      val eMsg = const("Index ").concat(indices(dimIndex).toS)
+        .concat(s" is out of bounds for axis $dimIndex with size ")
+        .concat(shape(dimIndex).toS)
+      (indices(dimIndex) >= shape(dimIndex)).orEmpty(Code._fatalWithID[Unit](eMsg, errorId))
+    }
+  }
```
johnc1231, line 284: This one isn't your fault since you just moved this code block, but since we are touching it can we just make this a `Unit` method` and append the `Code.foreach` to the code builder?

johnc1231 (dismissed):

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
patrick-schultz: Oof, good call

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArraySlice.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  override def fromCodes(codes: IndexedSeq[Code[_]]): SNDArraySliceCode = {
+    val codesT = codes.asInstanceOf[IndexedSeq[Code[Long@unchecked]]]
+    val shape = codesT.slice(0, nDims)
+    val strides = codesT.slice(nDims, 2*nDims)
+    val dataFirstElement = codesT.last
+    new SNDArraySliceCode(this, shape, strides, dataFirstElement)
+  }
+
+  def canonicalPType(): PType = pType
```
tpoterba, line 51: no longer necessary -- subsumed by storageType/copiedType

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArraySlice.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  override def fromSettables(settables: IndexedSeq[Settable[_]]): SNDArraySliceSettable = {
+    val shape = settables.slice(1, 1 + pType.nDims).asInstanceOf[IndexedSeq[Settable[Long@unchecked]]]
+    val strides = settables.slice(1 + pType.nDims, 1 + 2 * pType.nDims).asInstanceOf[IndexedSeq[Settable[Long@unchecked]]]
+    val dataFirstElementPointer = settables.last.asInstanceOf[Settable[Long]]
+    new SNDArraySliceSettable(this, shape, strides, dataFirstElementPointer)
+  }
+
+  override def fromCodes(codes: IndexedSeq[Code[_]]): SNDArraySliceCode = {
+    val codesT = codes.asInstanceOf[IndexedSeq[Code[Long@unchecked]]]
```
tpoterba, line 44: can you assert that the number of codes is 2 * nDims + 1?

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArraySlice.scala
```diff
@@ -0,0 +1,137 @@
...
+  override def castRename(t: Type): SType = SNDArrayPointer(pType.deepRename(t))
+
+  override def _coerceOrCopy(cb: EmitCodeBuilder, region: Value[Region], value: SCode, deepCopy: Boolean): SCode = ???
+
+  override def codeTupleTypes(): IndexedSeq[TypeInfo[_]] = FastIndexedSeq(LongInfo)
+
+  def loadFrom(cb: EmitCodeBuilder, region: Value[Region], pt: PType, addr: Code[Long]): SCode = ???
+
+  override def fromSettables(settables: IndexedSeq[Settable[_]]): SNDArraySliceSettable = {
+    val shape = settables.slice(1, 1 + pType.nDims).asInstanceOf[IndexedSeq[Settable[Long@unchecked]]]
```
tpoterba, line 37: assert here too

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArraySlice.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  override lazy val virtualType: TNDArray = pType.virtualType
+
+  override def copiedType: SType = SNDArrayPointer(pType)
+
+  override def storageType(): PType = pType
+
+  override def castRename(t: Type): SType = SNDArrayPointer(pType.deepRename(t))
+
+  override def _coerceOrCopy(cb: EmitCodeBuilder, region: Value[Region], value: SCode, deepCopy: Boolean): SCode = ???
```
tpoterba, line 30: implement this by matching on stype and returning self if the stype of `value` is equal to `this` and deepCopy is false, and throwing a match error otherwise.

tpoterba (dismissed):
couple minor things. looks good!

patrick-schultz:
Thanks Tim. The from/toCodes were completely wrong, should be good now. Do you think I should leave `loadFrom` unimplemented until there's a corresponding pType?

tpoterba (approved):

PR #10618, opened 2021-06-25
[compiler] Chunk cache

ammekk:
Added Chunk Cache to facilitate faster chunk interactions through less use of malloc and free.

### DIFF
--- a/hail/src/main/scala/is/hail/annotations/ChunkCache.scala
+++ b/hail/src/main/scala/is/hail/annotations/ChunkCache.scala
@@ -104,7 +104,7 @@ class ChunkCache1 private(allocator: Long => Long, freer: Long => Unit){
     else {
       val closestSize = bigChunkCache.ceilingEntry(size)
       if (closestSize != null && (closestSize.getKey == size
-        || ((closestSize.getKey * minSpaceRequiremen) <= size))) {
+        || ((closestSize.getKey * minSpaceRequirements) <= size))) {
         cacheHits += 1
         val chunkPointer = closestSize.getValue.pop()
         if (closestSize.getValue.size == 0) bigChunkCache.remove(closestSize.getKey)
--- a/hail/src/main/scala/is/hail/annotations/ChunkCache.scala
+++ b/hail/src/main/scala/is/hail/annotations/ChunkCache.scala
@@ -31,7 +31,7 @@ private class ChunkCache (allocator: Long => Long, freer: Long => Unit){
   private[this] var chunksRequested = 0
   private[this] var cacheHits = 0
   private[this] var smallChunkCacheSize = 0
-  private[this] val smallChunkCache = new Array[LongArrayBuilder](highestSmallChunkPowerOf2 + 1)
+  private[this] val smallChunkCache = new Array[LongArrayBuilder](highestSmallChunkPowerOf2 + 2)
   (0 until highestSmallChunkPowerOf2 + 1).foreach(index => {
     smallChunkCache(index) = new LongArrayBuilder()
   })
@@ -106,6 +106,7 @@ private class ChunkCache (allocator: Long => Long, freer: Long => Unit){
 
   def getChunk(pool: RegionPool, size: Long): (Long, Long) = {
     chunksRequested += 1
+    assert(size > 0L)
     if (size <= biggestSmallChunk) {
       val closestPower = indexInSmallChunkCache(size)
       if(smallChunkCache(closestPower).size == 0 ) {
--- a/hail/src/main/scala/is/hail/annotations/ChunkCache.scala
+++ b/hail/src/main/scala/is/hail/annotations/ChunkCache.scala
@@ -6,7 +6,7 @@ import java.util.TreeMap
 import java.util.function.BiConsumer
 import scala.collection.mutable
 
-class ChunkCache1 private(allocator: Long => Long, freer: Long => Unit){
+private class ChunkCache (allocator: Long => Long, freer: Long => Unit){
   private[this] val highestSmallChunkPowerOf2 = 24
   private[this] val biggestSmallChunk = Math.pow(2,highestSmallChunkPowerOf2)
   private[this] val bigChunkCache = new TreeMap[Long, LongArrayBuilder]()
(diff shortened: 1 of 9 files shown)


hail/src/main/scala/is/hail/annotations/RegionMemory.scala
```diff
@@ -58,9 +58,9 @@ final class RegionMemory(pool: RegionPool) extends AutoCloseable {
 
   private def allocateBigChunk(size: Long): Long = {
     val o = pool.getChunk(size)
```
johnc1231, line 60: To make this clearer, do something like `val (addr, bytesAllocated) = pool.getChunk(size)`

hail/src/main/scala/is/hail/annotations/RegionMemory.scala
```diff
@@ -293,13 +295,14 @@ final class RegionMemory(pool: RegionPool) extends AutoCloseable {
     val extra = PNDArray.headerBytes
 
     // This adjusted address is where the ndarray content starts
-    val allocatedAddr = pool.getChunk(size + extra) + extra
-
+    val allocatedChunk = pool.getChunk(size + extra)
```
johnc1231, line 298: Same as above, structure like `val (a, b) = pool.getChunk(...)`

hail/src/main/scala/is/hail/annotations/RegionMemory.scala
```diff
@@ -132,20 +132,22 @@ final class RegionMemory(pool: RegionPool) extends AutoCloseable {
 
   private def releaseNDArrays(): Unit = {
     var i = 0
+    val ndArraysToFree = new LongArrayBuilder()
```
johnc1231, line 135: Like Tim said, we should try not allocating an object here every time. Let's just make a `freeChunk` method that takes one long, and change `freeChunks` to just call `freeChunk` a bunch of times or something

johnc1231 (changes requested):

hail/src/main/scala/is/hail/annotations/RegionMemory.scala
johnc1231, line 60: Here too.

hail/src/main/scala/is/hail/annotations/RegionPool.scala
```diff
@@ -13,6 +17,151 @@ object RegionPool {
...
+  def getUsage(): (Int, Int, Int) = {
+    (chunksRequested, cacheHits, deallocationsToFit)
+  }
+  def indexInSmallChunkCache(size: Long): Int = {
+    var closestPower = highestSmallChunkPowerOf2
+    while((size >> closestPower) != 1) closestPower = closestPower - 1
+    if (size % (1 << closestPower) != 0) closestPower +=1
+    closestPower
+  }
+  override def getChunk(pool: RegionPool, size: Long): (Long, Long) = {
```
johnc1231, line 118: Leave new lines between method names. Necessary for the whole file really, I just highlight some examples.

hail/src/main/scala/is/hail/annotations/RegionPool.scala
```diff
@@ -13,6 +17,151 @@ object RegionPool {
...
+  var cacheHits = 0
+  var deallocationsToFit = 0
+  var smallChunkCacheSize = 0
+  val smallChunkCache = new Array[LongArrayBuilder](highestSmallChunkPowerOf2 + 1)
+   (0 until highestSmallChunkPowerOf2 + 1).foreach(index => {
+     smallChunkCache(index) = new LongArrayBuilder()
+   })
+
+  def getChunkSize(chunkPointer: Long): Long =  chunksEncountered(chunkPointer)
+   def freeInLongArrayBuilder(pool: RegionPool, ab: LongArrayBuilder):Unit = {
```
johnc1231, line 56: These defs aren't indented the same way

tpoterba:
design looks great! Awesome work!

hail/src/test/scala/is/hail/annotations/RegionSuite.scala
```diff
@@ -232,4 +235,33 @@ class RegionSuite extends TestNGSuite {
...
+      var i = 0
+      while (i <= 9) {
+        ab += chunkCache.getChunk(pool, sizeArr(i))._1
+        i += 1
+      }
+      chunkCache.freeChunksToCache(ab)
+      chunkCache.freeAll(pool)
+
+    }
+  }
```
johnc1231, line 266: We talked in person about what this test should be. We want to look at the operations arrays to check that we use the cache when appropriate, and free when appropriate.

johnc1231 (changes requested):

hail/src/main/scala/is/hail/annotations/ChunkCache.scala
```diff
@@ -0,0 +1,155 @@
...
+  private[this] var cacheHits = 0
+  private[this] var smallChunkCacheSize = 0
+  private[this] val smallChunkCache = new Array[LongArrayBuilder](highestSmallChunkPowerOf2 + 1)
+  (0 until highestSmallChunkPowerOf2 + 1).foreach(index => {
+    smallChunkCache(index) = new LongArrayBuilder()
+  })
+
+  def getChunkSize(chunkPointer: Long): Long =  chunksEncountered(chunkPointer)
+
+  def freeChunkFromMemory(pool: RegionPool, chunkPointer: Long):Unit = {
```
tpoterba, line 41: might be easier for a cache to take a pool in the constructor, rather than on every method

hail/src/main/scala/is/hail/annotations/ChunkCache.scala
```diff
@@ -0,0 +1,155 @@
...
+  def getUsage(): (Int, Int) = {
+    (chunksRequested, cacheHits)
+  }
+
+  def indexInSmallChunkCache(size: Long): Int = {
+    var closestPower = highestSmallChunkPowerOf2
+    while((size >> closestPower) != 1) closestPower = closestPower - 1
+    if (size % (1 << closestPower) != 0) closestPower +=1
+    closestPower
+  }
```
johnc1231, line 105: This is maybe overkill, but I think I would assert size > 0 here. If we somehow get a size of 0 here, you'll have an infinite loop. It doesn't really make sense to free something that's 0 bytes, but it's a bug that's happened before and I'd rather we get an error than an infinite loop.

johnc1231 (dismissed):

johnc1231 (approved):

PR #10619, opened 2021-06-25
[query] NDArrays only reference count their underlying data, are no longer based on arrays

johnc1231:
Previously, NDArrays were represented by a pointer to reference counted memory which contained a tuple of (Tuple, Tuple, Array), which were respectively (shape, strides, data). Now, the representation is an inline (Tuple, Tuple, Long), where the Long is a pointer to a reference counted region of memory that stores the ndarray data.

### DIFF
--- a/hail/src/main/scala/is/hail/annotations/Region.scala
+++ b/hail/src/main/scala/is/hail/annotations/Region.scala
@@ -218,6 +218,13 @@ object Region {
     false
   }
 
+  val sharedChunkHeaderBytes = 16L
+  def getSharedChunkRefCount(ndAddr: Long): Long = Region.loadLong(ndAddr - sharedChunkHeaderBytes)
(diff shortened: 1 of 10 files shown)


patrick-schultz (dismissed):
This is a great change. I do have one high level design concern, that isn't really introduced by this PR, but this PR makes it easier to fix.

I think it's wrong for `RegionMemory` to have a dependency on `PNDArray`. I'd propose that instead of `allocateNDArrayData` et al, we have `allocateSharedChunk`, which returns the pointer to the beginning of the usable data (after the header), with methods like `getSharedChunkRefCount(ptr: Long)` which know to subtract an offset from the pointer.

patrick-schultz (changes requested):
Comments above

johnc1231:
Great comments, addressed both.

patrick-schultz (approved):

PR #10624, opened 2021-06-29
[query] fix _quantile_from_cdf bug

patrick-schultz:
A user was getting an index out-of-bounds error on `cdf.values[idx]`. I can't reproduce it, but this should guarantee the index is in bounds, and is a simplification besides.


tpoterba (approved):
looks great, thanks.

PR #10626, opened 2021-06-29
[query] Remove support for legacy unordered keyed tables.

tpoterba:
CHANGELOG: Removed support for reading legacy for keyed tables written by Hail versions before 0.2.0.

I'm making this change because I don't want to add infrastructure to read these tables with whole stage code generation.


tpoterba:
I removed four tables from the `backward_compatibility` directory corresponding to version 1.0.0 (written by a `devel` Hail version).

johnc1231 (approved):

PR #10627, opened 2021-06-29
[grafana] Configure datasources in ConfigMap

daniel-goldstein:
This adds the Google Cloud Monitoring and Prometheus datasources to the grafana configuration. I had done this initially by hand in the UI but this is the first step toward reproducible monitoring, and I'll eventually follow up with dashboards as code.

The one "change" I made is I exposed the prometheus port sitting behind nginx so that grafana can talk directly to prometheus. Currently, there's an nginx sitting in front of prometheus so that the prometheus UI can be exposed at prometheus.hail.is with https and dev authentication. This hasn't changed. Currently though, grafana is piggybacking on this flow by forwarding the user's session (which I set up in the UI), but I couldn't figure out an easy way to set that in the config and it seemed unnecessarily complicated. I ended up going the simpler route of just letting grafana talk to prometheus directly and not go through nginx. The 9090 endpoint is not reachable outside of the cluster.

I considered namespacing the prometheus domain (`{{ default_ns.name }}` instead of `default`), but I pretty much never find it useful to spin up my own prometheus. In the rare case I run my own grafana I just point it to the data from default.

### DIFF
--- a/grafana/Makefile
+++ b/grafana/Makefile
@@ -14,5 +14,5 @@ build:
 
 deploy: build
 	! [ -z $(NAMESPACE) ]  # call this like: make deploy NAMESPACE=default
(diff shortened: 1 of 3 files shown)


jigold (approved):
Take the WIP tag off at your leisure (holiday weekend)

PR #10630, opened 2021-06-30
[aiogoogle] Fix disk API and add catch for quota exceeded

jigold:


jigold:
I tested this by submitting 64 true, 10 Gi 0.25 core jobs on one node with a local SSD. It took 7 minutes for the whole batch to finish. This is a bit disappointing, but at least there were no errors. We can potentially revisit using get instead of wait for the polling loop, but the default sleep backoff method we have quickly blew our quota.

hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -1,29 +1,33 @@
...
 
 log = logging.getLogger('compute_client')
 
 
-async def request_with_wait_for_done(request_f, path, params: MutableMapping[str, Any] = None, **kwargs):
-    assert 'params' not in kwargs
+class GCPError:
+    def __init__(self, code, message):
+        self.code = code
+        self.message = message
```
danking, line 16: Can we maybe call this GCPErrorInfo? I'm a bit worried the name looks like it should be raise-able.

danking: Let's put types on these fields. Without the types, `code` sounds like it should be an int.

hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -77,13 +82,50 @@ async def list(self, path: str, *, params: MutableMapping[str, Any] = None, **kw
...
     async def detach_disk(self, path: str, *, params: MutableMapping[str, Any] = None, **kwargs):
-        return await request_with_wait_for_done(self.post, path, params, **kwargs)
+        return await self._request_with_zonal_operations_response(self.post, path, params, **kwargs)
 
     async def delete_disk(self, path: str, *, params: MutableMapping[str, Any] = None, **kwargs):
-        return await request_with_wait_for_done(self.delete, path, params, **kwargs)
+        return await self.delete(path, params=params, **kwargs)
+
+    async def _request_with_zonal_operations_response(self, request_f, path, params: MutableMapping[str, Any] = None, **kwargs):
+        assert 'params' not in kwargs
```
danking, line 97: Python guarantees this is true. Let's remove the assertion.

```
In [1]: def f(foo=3, **kwargs): 
   ...:     return foo 
   ...:                                                                                                                                                                                        

In [5]: f(foo=2, **{'foo':3})                                                                                                                                                                  
  File "<ipython-input-5-48a15be69d20>", line 1, in <module>
    f(foo=2, **{'foo':3})
TypeError: f() got multiple values for keyword argument 'foo'


In [6]: f(foo=2, foo=3)                                                                                                                                                                        
  File "<ipython-input-6-b0f647cac392>", line 1
    f(foo=2, foo=3)
            ^
SyntaxError: keyword argument repeated
```

hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -77,13 +82,50 @@ async def list(self, path: str, *, params: MutableMapping[str, Any] = None, **kw
...
+                    error = result.get('error')
+                    if error:
+                        errors = error['errors']
+                        errors = [GCPError(e['code'], e['message']) for e in errors]
+
+                    if http_error_status_code is not None:
+                        raise GCPOperationError(http_error_status_code, http_error_message, errors, result)
+                    return result
+
+                await asyncio.sleep(1)
```
danking, line 129: Hmm. I don't like a hardcoded 1 here. I feel like we should use exponential back off like we do elsewhere. `wait` returns if:
1. The operation finished.
2. The server doesn't want to hold open the operation (probably because its overloaded).
3. We're approaching the 2 minute deadline.

In the first case, we won't reach the sleep. In the second case, we should definitely exponentially back off. In the third case, we don't need to sleep at all. I think we can satisfy all cases by exponentially backing off with a maximum wait much smaller than 2 minutes. Maybe 15 seconds? What do you think?

hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -77,13 +82,50 @@ async def list(self, path: str, *, params: MutableMapping[str, Any] = None, **kw
...
+                    http_error_message = result.get('httpErrorMessage')
+                    errors = None
+
+                    error = result.get('error')
+                    if error:
+                        errors = error['errors']
+                        errors = [GCPError(e['code'], e['message']) for e in errors]
+
+                    if http_error_status_code is not None:
+                        raise GCPOperationError(http_error_status_code, http_error_message, errors, result)
```
danking, line 126: Hmm. I don't like the correlation between `errors is not None` and `http_error_status_code is not None`. I don't think we actually care about the HTTP status code and error message. If there are errors, then something went wrong, right?

What do you think of this?
```python3
error = result.get('error')
if error:
    raise GCPOperationError([GCPError(e['code'], e['message']) for e in error['errors']], result)
```

hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -77,13 +82,50 @@ async def list(self, path: str, *, params: MutableMapping[str, Any] = None, **kw
...
+            local_params = copy.deepcopy(params)
+
+            if local_params is None:
+                local_params = {}
+
+            request_uuid = str(uuid.uuid4())
+            if 'requestId' not in local_params:
+                local_params['requestId'] = request_uuid
+
+            resp = await request_f(path, params=local_params, **kwargs)
```
danking, line 109: Hmm. I'm wary of deep copying params.

Do you agree that it's always a bug for a user of this API to set `requestId`? If they set a requestId, there's no way for us to correctly retry transient errors.

In light of that, I think we should do this:

```python3
async def _request_with_zonal_operations_response(...):
    params = params or dict()
    assert 'requestId' not in params
    async def request_and_wait():
        params['requestId'] = str(uuid.uuid4())
        resp = ...
```

hail/python/hailtop/utils/utils.py
```diff
@@ -552,12 +550,18 @@ def is_transient_error(e):
...
+    import hailtop.httpx  # pylint: disable=import-outside-toplevel
+
     if isinstance(e, aiohttp.ClientResponseError) and (
             e.status in RETRYABLE_HTTP_STATUS_CODES):
         # nginx returns 502 if it cannot connect to the upstream server
         # 408 request timeout, 500 internal server error, 502 bad gateway
         # 503 service unavailable, 504 gateway timeout
         return True
+    if (isinstance(e, hailtop.aiogoogle.client.compute_client.GCPOperationError)
+            and (e.errors and 'QUOTA_EXCEEDED' in e.error_codes)):
```
danking, line 563: checking that e.errors is non-empty is redundant, let's just use the right-hand side of the `and`.

hail/python/hailtop/aiogoogle/client/compute_client.py
```diff
@@ -1,29 +1,33 @@
...
 
-    request_uuid = str(uuid.uuid4())
-    if 'requestId' not in params:
-        params['requestId'] = request_uuid
+class GCPOperationError(Exception):
+    def __init__(self, status: int, message: str, errors: Optional[List[GCPError]], response: Dict[str, Any]):
+        super().__init__(message)
+        self.status = status
+        self.message = message
+        self.errors = errors or []
```
danking, line 24: I don't think we need to keep the errors around, we have the codes and messages, right? In that case we probably don't need to have GCPError at all, since we're storing the information split.

hail/python/hailtop/aiogoogle/client/compute_client.py
danking: Or maybe:
```python3
error = result.get('error')
if error:
    assert result.get('httpErrorStatusCode') is not None
    assert result.get('httpErrorMessage') is not None
    raise GCPOperationError(result['httpErrorStatusCode'],
                            result['httpErrorMessage'],
                            [GCPError(e['code'], e['message']) for e in error['errors']],
                            result)
```

danking (changes requested):
I think wait is actually the right choice! `get` seems strictly worse because it's not a long-polling API.

batch/batch/driver/gce.py
```diff
@@ -184,4 +184,4 @@ async def delete_orphaned_disks(self):
                     log.exception(f'error while deleting orphaned disk {disk_name}')
 
     async def delete_orphaned_disks_loop(self):
-        await periodically_call(15, self.delete_orphaned_disks)
+        await periodically_call(60, self.delete_orphaned_disks)
```
jigold, line 187: I made this 60. We're making too many API calls listing disks for each zone. I'm open to making it 15, but I don't think it matters whether it's 15 or 60 and is better to conserve api calls.

jigold:
I'm putting a wip tag on this so we make sure we dev deploy and test the behavior one last time before merging.

jigold:
I dev deployed this and it still is working fine. It's still pretty slow and I was getting rate limit exceeded errors still trying to attch/detach 64 disks. Average operation time was still 15 seconds. I think part of the problem might be the delay starts at 0.1 for retry_transient_errors. We can consider making this a parameter and setting it to a higher number for this use case.

danking (approved):

danking:
I'm good with the 60s.

danking:
I like the idea that, in a future PR, we investigate a slightly higher initial back off for disk creation failures.

PR #10632, opened 2021-07-01
[query] Weighted linear regression rows

johnc1231:


tpoterba:
sweet!

hail/src/main/scala/is/hail/expr/ir/ndarrays/EmitNDArray.scala
```diff
@@ -128,7 +128,14 @@ object EmitNDArray {
...
                 val rightShapeValues = rightProducer.shape
 
-                val shapeArray = NDArrayEmitter.unifyShapes2(cb, leftShapeValues, rightShapeValues)
+//                cb.println(rightShapeValues.foldLeft[Code[String]](
+//                  leftShapeValues.foldLeft[Code[String]](
+//                    const("Map2 with NDArray shapes: [ ")
+//                  )((accum, v) => accum.concat(v.toS).concat(" "))
+//                    .concat("] vs [ ")
+//                )((accum, v) => accum.concat(v.toS).concat(" "))
+//                  .concat(s"] and operation ${Pretty(body)}"))
```
patrick-schultz, line 137: debug

patrick-schultz (changes requested):
This looks good to me, though I didn't think very hard about the quality of the IR being generated. I do wonder if it might be worth trying to avoid the multiplications by 1 in the unweighted case, but that could be a future optimization.

johnc1231:
I can remove multiplications by 1 in a subsequent PR. I think not hard and probably not much of a difference. At the moment, the only person using this method will be using it to enable weighted linreg anyway, so not super worried.

patrick-schultz (dismissed):

johnc1231:
Something is wrong with the `annotate_globals`. It's taking way too long when used on something modestly large. I think it's redoing work somewhere.

tpoterba:
Reassigning to myself since Patrick is OOO this week

johnc1231:
Great, thanks. Should be ready to be reviewed now, the performance is now where I want it to be.

hail/python/hail/methods/statgen.py
```diff
@@ -489,11 +530,11 @@ def process_block(block):
 
         # Processes one block group based on given idx. Returns a single struct.
         def process_y_group(idx):
-            X = hl.nd.array(block[entries_field_name].map(lambda row: mean_impute(select_array_indices(row, ht.kept_samples[idx])))).T
+            X = (hl.nd.array(block[entries_field_name].map(lambda row: mean_impute(select_array_indices(row, ht.kept_samples[idx])))) * ht.__sqrt_weight_nds[idx]).T
```
tpoterba, line 533: do we think that this line will have any performance impact on the non-weighted case?

tpoterba (changes requested):
Looks good! One question.

hail/python/hail/methods/statgen.py
johnc1231: I would think it would be minimal since it'll just broadcast out some multiply by 1s, but it'd be pretty easy to just add an if statement here and only do it if the user initially specified weights. I'll just do that.

johnc1231: Addressed

johnc1231:
Also made it so that `y` and `cov` don't get scaled when weights is None, and then `sqrt_weights` can just be missing.

tpoterba (approved):

PR #10639, opened 2021-07-06
[batch] cancel step or delete tasks when waiting for job deletion

daniel-goldstein:


jigold (approved):

batch/batch/worker/worker.py
```diff
@@ -474,11 +474,18 @@ async def localize_rootfs():
...
+        try:
+            await asyncio.wait([deleted, step], return_when=asyncio.FIRST_COMPLETED)
+            if deleted.done():
+                raise JobDeletedError()
+            assert step.done()
+            return step.result()
+        finally:
+            for t in (step, deleted):
+                if t.done() and t.exception():
+                    log.exception(t.exception())
```
jigold, line 486: This will log an exception for user tasks failing. Do we want that? Or do we catch whether it's a user error in an upstream function call?

jigold (changes requested):

batch/batch/worker/worker.py
daniel-goldstein: Good point, I can add a check here for `user_error`.

daniel-goldstein: I don't think we have a robust way of guaranteeing that upstream calls check or not, or propagate or not. Long term I think there's a better way of splitting up the error cases further but that's for another time I suppose.

batch/batch/worker/worker.py
```diff
@@ -474,11 +474,19 @@ async def localize_rootfs():
...
+        try:
+            await asyncio.wait([deleted, step], return_when=asyncio.FIRST_COMPLETED)
+            if deleted.done():
+                raise JobDeletedError()
+            assert step.done()
+            return step.result()
+        finally:
+            for t in (step, deleted):
+                if t.done():
+                    if t.exception() and not user_error(t.exception()):
```
jigold, line 486: sorry nit -- can we assign `t.exception()` to a variable and then use that?

jigold (changes requested):

jigold (approved):

daniel-goldstein:
Closing since I will work this into the second return of crun.

PR #10640, opened 2021-07-06
[batch] prevent running startup script more than once

jigold:


daniel-goldstein (dismissed):
How come we need to explicitly delete the instance and can't let the startup script run to completion? Having these potentially infinite loops in here makes me uneasy.

Regardless, the loop here is repeated at the end of the script, so I think we should at least move that into a `cleanup` function, add a `trap cleanup EXIT`, and then just `exit` in this conditional.

jigold:
The reason for the while loop is the instance keeps running even if the startup script fails.

jigold:
Also my bash is not real great, so definitely double check I got the function right.

daniel-goldstein:
Ah I didn't notice that the two loops were in different scripts (startup vs. run). If the run-script exits cleanly will the worker shut down?

batch/batch/driver/create_instance.py
```diff
@@ -109,7 +125,7 @@ async def create_instance(
                     'key': 'run_script',
                     'value': rf'''
 #!/bin/bash
-set -x
+set -ex
```
jigold, line 128: I'm not sure if we need this change.

daniel-goldstein: I think this is good we should fail early.

daniel-goldstein (approved):

PR #10641, opened 2021-07-06
[batch] Use a timeout for iptables instead of failing on contention

daniel-goldstein:


jigold (approved):

daniel-goldstein:
Closing since I'll work this into the second return of crun.

PR #10645, opened 2021-07-07
[batch] Add user error for bad gcsfuse creds

jigold:
Pending confirmation from Ben W. that this is indeed a user error...


jigold:
Ben confirmed this was a user error.

daniel-goldstein (approved):

PR #10646, opened 2021-07-07
[batch] Modify disk gc loop to be more lenient

jigold:


daniel-goldstein (approved):

jigold:
This shouldn't be necessary because of #10652. Will reopen if there's still a problem.

jigold:
Ben's disks still took over 10 minutes to be attached when the system was under load.

daniel-goldstein (approved):

PR #10648, opened 2021-07-08
[batch] run user jobs with crun: Take 2

daniel-goldstein:
Another attempt at #10376 with some fixes of bugs seen during the first deploy.


batch/batch/worker/worker.py
```diff
@@ -806,12 +837,13 @@ async def delete_container(self):
...
-                self.process.kill()
+                self.process.terminate()
+                self.process = None
+                await check_exec_output('crun', 'kill', '--all', self.container_name)
             except asyncio.CancelledError:
                 raise
             except Exception:
-                log.warning('while deleting container, ignoring', exc_info=True)
-        self.process = None
+                log.exception('while deleting container', exc_info=True)
```
jigold, line 846: Should the `self.process = None` line be in a finally clause?

batch/batch/worker/worker.py
```diff
@@ -408,33 +415,39 @@ def __init__(self, job, name, spec):
...
         self.netns: Optional[NetworkNamespace] = None
         self.process = None
 
-    async def run(self, worker):
+    async def run(self, worker: 'Worker'):
         try:
 
             async def localize_rootfs():
-                async with worker.rootfs_locks[self.image_ref_str]:
+                cond = worker.rootfs_locks[self.image_ref_str]
```
jigold, line 434: Should `worker.rootfs_locks` be `worker.rootfs_conds`? It's a bit confusing since a lock is also an asyncio object. I understand though you're using this as a lock.

batch/batch/worker/worker.py
```diff
@@ -408,33 +415,39 @@ def __init__(self, job, name, spec):
...
+                async with cond:
+                    if len(worker.rootfs_digest_refs) == 10:
+                        unused_image = await cond.wait_for(lambda: get_unused_image(worker))
+                        shutil.rmtree(f'/host/rootfs/{unused_image}')
+                        del worker.rootfs_digest_refs[unused_image]
```
jigold, line 439: Do you need to also delete the lock/condition for the unused image?

batch/batch/worker/worker.py
```diff
@@ -408,33 +415,39 @@ def __init__(self, job, name, spec):
...
 
-    async def run(self, worker):
+    async def run(self, worker: 'Worker'):
         try:
 
             async def localize_rootfs():
-                async with worker.rootfs_locks[self.image_ref_str]:
+                cond = worker.rootfs_locks[self.image_ref_str]
+                async with cond:
+                    if len(worker.rootfs_digest_refs) == 10:
```
jigold, line 436: I'm wary of us capping the number of in-use images to 10. I think we should ultimately treat this as a resource like the disks and monitor how many jobs are waiting for a root filesystem and for how long. The scheduler should probably take this number into account. Or we have a unified storage system and this won't matter anymore.

I'm still not sure why we don't use xfs copy on write here so that the root filesystem is attributed to the users quota. Then you can get rid of the underlying copy and the inodes will still exist for the in-use copy, but then you can replace the cache with another image. I think I thought about this when I was trying to design an input file cache shared between users. Take a look at `--reflink`.

batch/batch/worker/worker.py
jigold: So if two jobs with the same image are waiting for the 10th image to delete, what happens? Do they each have to wait for an image to be cleared or does one break when the other job has cleared the last image and only one image is deleted?

jigold (changes requested):

batch/batch/worker/worker.py
```diff
@@ -133,6 +135,11 @@ def get_image_digest(image_config) -> str:
     return image_config['RepoDigests'][0].split('@')[1].split(':')[1]
 
 
+def get_unused_image(worker):
+    unused = [image_digest for image_digest, count in worker.rootfs_digest_refs.items() if count == 0]
+    return unused[0] if len(unused) > 0 else False
```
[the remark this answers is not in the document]
daniel-goldstein: Your way of writing the method is definitely cleaner. As for None/False, both would probably work but I was following this description of truthiness from the Condition documentation:
```
Wait until a predicate becomes true.

The predicate must be a callable which result will be interpreted as a boolean value. The final value is the return value.
```

batch/batch/worker/worker.py
```diff
@@ -806,12 +837,13 @@ async def delete_container(self):
         if self.container_is_running():
             try:
                 log.info(f'{self} container is still running, killing crun process')
-                self.process.kill()
+                self.process.terminate()
```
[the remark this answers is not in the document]
daniel-goldstein: huh, I missed the signal. I don't quite know the best way about this, I was finding today that `terminate` wasn't very effective at actually shutting the `crun` process down, and `kill` could stop the `crun` process but not the child processes in the container. `crun kill` was the only option I tried that actually cleared all processes out of the container. Looks like `crun` [defaults to SIGTERM](https://github.com/containers/crun/blob/0b0b2887403de620c501a70db853ba7ebbdede3a/src/kill.c#L103), but I can do so explicitly since it is not documented.

batch/batch/worker/worker.py
```diff
@@ -474,17 +487,27 @@ async def localize_rootfs():
             self.state = 'error'
             self.error = traceback.format_exc()
         finally:
-            with self.step('deleting'):
-                await self.delete_container()
+            await self.delete_container()
```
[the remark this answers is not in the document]
daniel-goldstein: This is partially a separate issue that we need to address in a followup. The step manager will raise on `__enter__` if the job has been deleted, which means it will double raise and if we time anything in the `finally` block. This is part of the reason why job cancellation wasn't working properly. In the future I think the step manager should not care about job deletion since we now have an `asyncio.Event` to signal deletion. I'm unsure whether it's better to leave it in this in between and clean up deletion handling later or lump more things that don't seem entirely related into this PR, since many things use the timings blocks.

batch/batch/worker/worker.py
daniel-goldstein: Ya that's not ideal. If two jobs are waiting on the same image, the first will delete an unused image and populate the new one, and the second will delete a second unused image and not repopulate, so at the end we'll end up with 9 images in the cache. I suppose I can change this condition to be `if cache-is-full and need-to-pull-image`

daniel-goldstein: also yes I will rename to `rootfs_conds`.

batch/batch/worker/worker.py
daniel-goldstein: This should delete the entry as well

batch/batch/worker/worker.py
daniel-goldstein: Aside from providing a more accurate account of the space left on the worker disk, I'm confused about what effect this would have. The root filesystems are readonly since they are the lowerdir of the overlay so there would be no copy on write. This feels like it would primarily introduce a large copy overhead when a worker gets fully saturated with jobs from the same batch that all want to use the same image.

daniel-goldstein: I also really don't like this arbitrary 10 business, but I would have to think for longer about a reasonable solution without a unified storage system (which feels like the actual solution).

batch/batch/worker/worker.py
daniel-goldstein: Ended up reverting this to just normal locks again and have a separate GC loop for removing unused images.

batch/batch/worker/worker.py
daniel-goldstein: Now have a loop that will delete 1 unused image each minute.

batch/batch/worker/worker.py
daniel-goldstein: I'm not sure, because if we've failed before the point of marking `self.process = None` then it really isn't `None`, and it would indicate that we've ended up with a zombie process. I would rather treat the job as still running than let go of the reference to the asyncio `Process` entirely.

daniel-goldstein: In general, I'm not sure how to handle zombie processes. This is currently doing the most aggressive thing I can think of to kill off the processes.

batch/batch/worker/worker.py
```diff
@@ -408,36 +410,38 @@ def __init__(self, job, name, spec):
...
                     five_minutes_ago = time_msecs() - 5 * 60 * 1000
-                    if last_fetched and last_fetched < five_minutes_ago:
-                        assert self.image_ref_str in image_configs
-                        image_digest = get_image_digest(image_configs[self.image_ref_str])
-                        shutil.rmtree(f'/host/rootfs/{image_digest}')
                     if not last_fetched or last_fetched < five_minutes_ago:
-                        worker.rootfs_last_fetched[self.image_ref_str] = time_msecs()
                         await self.pull_image()
+                        worker.rootfs_last_fetched[self.image_ref_str] = time_msecs()
+                        log.info(f'Added image to cache: {self.image_ref_str}')
```
jigold, line 435: I'm not seeing where the XFS Quota gets set for the container's root file system. What I think I'm envisioning is we pull or use the cached image and get it's total size. We then add XFS quotas for the merged / upper directories (user writable ones) to the remainder of the quota (asserting not less than 0). For example, we have a 1 core job. We've promised the root file system (not including /io) accounts for 5Gi. The image is 2.5Gi. We then change the size of the XFS Quota for the merged/upper directories for the container to 2.5Gi. `/io` is whatever the user requested and how we calculate the quotas and mount `/io` shouldn't change with this PR.

batch/batch/worker/worker.py
```diff
@@ -474,17 +478,28 @@ async def localize_rootfs():
             self.state = 'error'
             self.error = traceback.format_exc()
         finally:
-            with self.step('deleting'):
-                await self.delete_container()
+            await self.delete_container()
+            if self.image_digest:
+                worker.image_digest_ref_count[self.image_digest] -= 1
+                assert worker.image_digest_ref_count[self.image_digest] >= 0
```
jigold, line 484: You might want to add what to print if the assert fails. Not sure if it's helpful or not.

```
assert worker.image_digest_ref_count[self.image_digest] >= 0, f'{worker.image_digest_ref_count}'
```

batch/batch/worker/worker.py
```diff
@@ -408,36 +410,38 @@ def __init__(self, job, name, spec):
...
-                        worker.rootfs_last_fetched[self.image_ref_str] = time_msecs()
                         await self.pull_image()
+                        worker.rootfs_last_fetched[self.image_ref_str] = time_msecs()
+                        log.info(f'Added image to cache: {self.image_ref_str}')
 
                     self.image_config = image_configs[self.image_ref_str]
-                    image_digest = get_image_digest(self.image_config)
-                    self.rootfs_path = f'/host/rootfs/{image_digest}'
+                    self.image_digest = get_image_digest(self.image_config)
+                    self.rootfs_path = f'/host/rootfs/{self.image_digest}'
```
jigold, line 439: Is `self.image_digest` always a 1:1 mapping with `self.image_ref_str`? It's a bit confusing to me that you have two data structures. One is indexed by image_digest and the other is by image_ref_str.

batch/batch/worker/worker.py
```diff
@@ -806,12 +829,20 @@ async def delete_container(self):
...
+                self.process = None
+                await check_exec_output('crun', 'kill', '--all', self.container_name, 'SIGTERM')
             except asyncio.CancelledError:
                 raise
             except Exception:
-                log.warning('while deleting container, ignoring', exc_info=True)
-        self.process = None
+                log.exception('while deleting container', exc_info=True)
+
+            try:
```
jigold, line 840: Did you change the structure of this from checking whether the path exists to a try/except because the path should always be there?

jigold (changes requested):

jigold:
Don't let me forget to announce this on Zulip! We need to both announce we're doing an upgrade and that we are enforcing quotas for the root file system.

batch/batch/worker/worker.py
daniel-goldstein: It's not. `image_digest` is the unique SHA for a docker image, while `image_ref_str` is the colloquial name for the image the job will use, i.e. `gcr.io/hail-vdc/batch-worker`. I agree it's confusing to use both, but they serve two different purposes: we need to lock on `image_ref_str` because that's what the jobs are specifying, and we count how many containers are running on a particular image digest because we could have multiple digests downloaded for a particular `image_ref_str`.

batch/batch/worker/worker.py
daniel-goldstein: The try catch here actually might not be necessary, I'm doing it a bit defensively. I previously had the check that the mount path exists because the way we currently use `delete_container` requires it be idempotent (which I'm not crazy about but that's for another time). The point of this change was moving it into the `if`,

daniel-goldstein:
Action items for getting this in:
- Delete Docker's copy of an image when deleting an expanded root filesystem
- Try docker save instead of docker export so docker doesn't create a container. If this doesn't work, delete the container before deleting the image
- Pull every time a job is run for authentication purposes
- Up the reserved image cache size to 30

batch/batch/worker/worker.py
```diff
@@ -2007,22 +2002,25 @@ async def activate(self):
...
+                images = list(self.image_ref_count.keys())
+                for image_ref_str, image_id in images:
+                    async with self.rootfs_locks[image_ref_str]:
+                        if self.image_ref_count[(image_ref_str, image_id)] == 0:
+                            log.info(f'Found an unused image: {image_ref_str} with ID {image_id}')
+                            shutil.rmtree(f'/host/rootfs/{image_id}')
+                            del self.image_ref_count[(image_ref_str, image_id)]
+                            await check_shell(f'docker rmi {image_id}')
+                            log.info(f'Deleted image from cache: {image_ref_str} with ID {image_id}')
```
jigold, line 2021: Do we need to catch CancelledError here and raise?

batch/batch/worker/worker.py
```diff
@@ -1642,13 +2001,38 @@ async def activate(self):
             self.headers = {'X-Hail-Instance-Name': NAME, 'Authorization': f'Bearer {resp_json["token"]}'}
             self.active = True
 
+    def get_unused_image(self) -> Optional[str]:
```
jigold, line 2004: I don't think this function is used anymore.

batch/batch/worker/worker.py
```diff
@@ -1976,6 +2006,25 @@ async def activate(self):
...
+        for image_digest, count in self.image_digest_ref_count.items():
+            if count == 0:
+                return image_digest
+        return None
+
+    async def cleanup_old_images(self):
```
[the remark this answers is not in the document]
jigold: Bump on this.

jigold (changes requested):

jigold:
I'm approving, but please fix the changes requested and announce on Zulip in advance that you're putting in a change.

jigold (approved):

PR #10650, opened 2021-07-08
[monitoring] Expose disk and instance counts to prometheus

jigold:


monitoring/monitoring/monitoring.py
```diff
@@ -29,6 +34,16 @@
...
 
+GCP_REGION = os.environ['HAIL_GCP_REGION']
+BATCH_GCP_REGIONS = set(json.loads(os.environ['HAIL_BATCH_GCP_REGIONS']))
+BATCH_GCP_REGIONS.add(GCP_REGION)
+
+PROJECT = os.environ['PROJECT']
+
+DISK_STATES = pc.Gauge('batch_disk_states', 'Batch disk states', ['state', 'namespace', 'zone'])
+DISK_SIZE_GB_SUMMARY = pc.Summary('batch_disk_size_gb', 'Summary of user batch disk sizes (GB)', ['namespace', 'zone'])
+INSTANCE_STATUSES = pc.Gauge('batch_instance_statuses', 'Batch instance statuses', ['status', 'namespace', 'zone', 'machine_type', 'preemptible'])
```
daniel-goldstein, line 45: can you remove `states` and `statuses` from the metric names? I like that you've made them tags instead of separate metrics for each, but now the name is oddly biased toward a particular tag. Just `batch_disks` and `batch_worker_instances` feels cleaner, and suggests that you can aggregate by tags other than state and status.

monitoring/monitoring/monitoring.py
```diff
@@ -227,18 +242,76 @@ async def polling_loop(app):
...
+            DISK_SIZE_GB_SUMMARY.labels(**labels).observe(size_gb)
+
+            if creation_timestamp_msecs is None:
+                DISK_STATES.labels(state='creating', **labels).inc()
+            elif last_attach_timestamp_msecs is None:
+                DISK_STATES.labels(state='created', **labels).inc()
+            elif last_attach_timestamp_msecs is not None and last_detach_timestamp_msecs is None:
+                DISK_STATES.labels(state='attached', **labels).inc()
+            elif last_attach_timestamp_msecs is not None and last_detach_timestamp_msecs is not None:
+                DISK_STATES.labels(state='detached', **labels).inc()
```
daniel-goldstein, line 274: I'm wary to clear and then increment in an async loop. Maybe it's not a realistic issue but an unfortunately timed scrape could give a very incorrect picture. Could you instead count the occurences per unique set of labels and then `observe` outside of the async loop? Then we don't have to `clear`. `DISK_SIZE_GB_SUMMARY` is fine as it is.

monitoring/monitoring/monitoring.py
```diff
@@ -227,18 +242,76 @@ async def polling_loop(app):
...
+    INSTANCE_STATUSES.clear()
+
+    for zone in app['zones']:
+        async for instance in await compute_client.list(f'/zones/{zone}/instances', params={'filter': '(labels.role = batch2-agent)'}):
+            labels = {'status': instance['status'],
+                      'zone': zone,
+                      'namespace': instance['labels']['namespace'],
+                      'machine_type': instance['machineType'].rsplit('/', 1)[1],
+                      'preemptible': instance['scheduling']['preemptible']}
+            INSTANCE_STATUSES.labels(**labels).inc()
```
daniel-goldstein, line 292: Same concern about clear/inc.

daniel-goldstein (dismissed):
Couple thoughts about the metric collection.

jigold:
This is definitely much better, but there's still some jumpiness in the plots when a pod gets restarted. This is because it can get double the metrics in that interval.

daniel-goldstein (approved):

daniel-goldstein (approved):

PR #10654, opened 2021-07-12
[ci] Dont rebuild a failed source SHA

daniel-goldstein:
If you sort the GitHub PRs by `is:pr is:open sort:updated-desc`, you'll see that some PRs which have merge conflicts are continuously tested by CI. Really old PRs that are stuck in this state reach a GitHub quota for status updates to a PR for a given SHA. It seems to me that we shouldn't retest any PR that is `source_sha_failed`, and I think this should resolve that particular issue, but am not sure if this is necessarily the best place to insert this logic.

### DIFF
--- a/ci/ci/github.py
+++ b/ci/ci/github.py
@@ -189,6 +189,7 @@ class PR(Code):
         self.sha = None
         self.batch = None
         self.source_sha_failed = None
+        self.merge_conflict = False
 
         # 'error', 'success', 'failure', None
         self.build_state = None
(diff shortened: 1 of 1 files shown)


danking:
Hmm. I think I'll let you make the determination here since I think of CI as your domain. However, let me provide context on the previous behavior. We wanted to retry PRs that failed because some functionality in main was broken and subsequently fixed by another PR. This is probably quite rare in practice. If a bad commit is merged to main, that's an example of this.

A slightly more conservative change is to not build PRs that had a merge conflict. I can't think of any normal circumstance where a new commit to the *target* resolves the merge conflict.

danking (changes requested):
see my comment, dismiss if you disagree and I'll approve

danking:

PR #10656, opened 2021-07-12
[batch] Expose billing page to non-dev users

jigold:
I tested this with dev deploy using both my broad developer account and my personal account to make sure regular users only saw their own billing information.


danking (approved):
I like this, but I think the page should change to indicate that you're only seeing *your* charges in these billing projects. Currently, I don't think that's clear. It's also a bit weird to see a table of Cost by User that only shows me.

danking (changes requested):
xxx

jigold:
Closing while I test out changes.

PR #10664, opened 2021-07-14
[hailtop.batch] Write out large commands to GCS for BashJobs

jigold:
I made any command that is larger than 10KB is written to a file instead and is downloaded as an input file to be run (same as what we do for Python jobs). I also added a new field `user_code` that represents the user's code that they want to run versus the command we have Docker run. I formatted the `user_code` for Python jobs to show the actual function definition and the arguments used to call the function. It's probably not perfect (i.e. JobResourceFile), but better than nothing.

### DIFF
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -2,11 +2,13 @@ import re
 import dill
 import os
 import functools
-from io import BytesIO
+import inspect
+from io import StringIO, BytesIO
 from typing import Union, Optional, Dict, List, Set, Tuple, Callable, Any, cast
 
 from . import backend, resource as _resource, batch  # pylint: disable=cyclic-import
 from .exceptions import BatchException
+from .globals import DEFAULT_SHELL
 
 
 def _add_resource_to_set(resource_set, resource, include_rg=True):
@@ -71,7 +73,8 @@ class Job:
         self._timeout: Optional[Union[int, float]] = None
         self._gcsfuse: List[Tuple[str, str, bool]] = []
         self._env: Dict[str, str] = dict()
-        self._command: List[str] = []
+        self._wrapper_code: List[str] = []
+        self._user_code: List[str] = []
 
         self._resources: Dict[str, _resource.Resource] = {}
         self._resources_inverse: Dict[_resource.Resource, str] = {}
@@ -398,6 +401,9 @@ class Job:
         self._gcsfuse.append((bucket, mount_point, read_only))
         return self
 
+    async def _compile(self, local_tmpdir, remote_tmpdir):
+        raise NotImplementedError
+
     def _pretty(self):
         s = f"Job '{self._uid}'" \
             f"\tName:\t'{self.name}'" \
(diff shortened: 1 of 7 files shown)


danking:
I'm am so stoked to review this. This is a really excellent step forward for Batch!

hail/python/hailtop/batch/job.py
```diff
@@ -600,48 +657,50 @@ def command(self, command: str) -> 'BashJob':
...
-                   Job._regex_pattern,
-                   batch.Batch._regex_pattern]
-        subst_command = re.sub('(' + ')|('.join(regexes) + ')',
-                               handler,
-                               command)
-        self._command.append(subst_command)
-        return self
+        pipe = StringIO()
+        pipe.write(f'#! {job_shell}\n')
+        pipe.writelines(job_command)
```
danking, line 677: job_command isn't a list so `write` suffices.

hail/python/hailtop/batch/job.py
```diff
@@ -600,48 +657,50 @@ def command(self, command: str) -> 'BashJob':
...
+        pipe.seek(0)
+
+        code_bytes = pipe.getvalue().encode()
+
+        assert self._job_id is not None
+        job_path = f'{remote_tmpdir}/{self._job_id}'
+        code_path = f'{job_path}/code.sh'
+
+        await self._batch._fs.makedirs(os.path.dirname(code_path), exist_ok=True)
+        await self._batch._fs.write(code_path, code_bytes)
```
danking, line 688: I have mild performance concerns about copying the whole job command into a StringIO for every job in a large batch. Can we instead do:
```python3
fs = self._batch._fs
async with await retry_transient_errors(fs.create, url) as out:
    await out.write(f'#! {job_shell}\n'.encode())
    await out.write(job_command.encode())
    await out.write(b'\n')
```

danking: Oh, I see we need a copy of it for the _user_code anyway. I suppose if we're copying anyway then this is what we must do.

hail/python/hailtop/batch/job.py
```diff
@@ -600,48 +657,50 @@ def command(self, command: str) -> 'BashJob':
...
-        subst_command = re.sub('(' + ')|('.join(regexes) + ')',
-                               handler,
-                               command)
-        self._command.append(subst_command)
-        return self
+        pipe = StringIO()
+        pipe.write(f'#! {job_shell}\n')
+        pipe.writelines(job_command)
+        pipe.write('\n')
+        pipe.seek(0)
```
danking, line 679: unnecessary

hail/python/hailtop/batch/job.py
```diff
@@ -600,48 +657,50 @@ def command(self, command: str) -> 'BashJob':
...
+        assert self._job_id is not None
+        job_path = f'{remote_tmpdir}/{self._job_id}'
+        code_path = f'{job_path}/code.sh'
+
+        await self._batch._fs.makedirs(os.path.dirname(code_path), exist_ok=True)
+        await self._batch._fs.write(code_path, code_bytes)
+
+        code = self._batch.read_input(code_path)
+        self._add_inputs(code)
+        self._mentioned.add(code)
```
danking, line 692: Why isn't this handled by _interpolate_command?

hail/python/hailtop/batch/job.py
```diff
@@ -600,48 +657,50 @@ def command(self, command: str) -> 'BashJob':
...
-                               command)
-        self._command.append(subst_command)
-        return self
+        pipe = StringIO()
+        pipe.write(f'#! {job_shell}\n')
+        pipe.writelines(job_command)
+        pipe.write('\n')
+        pipe.seek(0)
+
+        code_bytes = pipe.getvalue().encode()
```
danking, line 681: I think this whole bit with the pipe is just:
```
code_bytes = ('#! '  + job_shell + '\n' + job_command + '\n').encode()
```

danking (dismissed):

danking (approved):

PR #10665, opened 2021-07-14
[query] Solve triangular matrices

ammekk:
Added nd array function to facilitate solving triangular matrices using LAPACK dtrtrs

### DIFF
--- a/hail/python/hail/nd/nd.py
+++ b/hail/python/hail/nd/nd.py
@@ -7,9 +7,10 @@ from hail.expr.types import HailType, tfloat64, tfloat32, ttuple, tndarray
 from hail.typecheck import typecheck, nullable, oneof, tupleof, sequenceof
 from hail.expr.expressions import (
     expr_int32, expr_int64, expr_tuple, expr_any, expr_array, expr_ndarray,
-    expr_numeric, Int64Expression, cast_expr, construct_expr)
+    expr_numeric, Int64Expression, cast_expr, construct_expr, expr_bool)
 from hail.expr.expressions.typed_expressions import NDArrayNumericExpression
 from hail.ir import NDArrayQR, NDArrayInv, NDArrayConcat, NDArraySVD, Apply
+
 
 tsequenceof_nd = oneof(sequenceof(expr_ndarray()), expr_array(expr_ndarray()))
 shape_type = oneof(expr_int64, tupleof(expr_int64), expr_tuple())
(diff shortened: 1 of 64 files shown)


hail/python/hail/nd/nd.py
```diff
@@ -269,6 +259,52 @@ def solve(a, b, no_crash=False):
     return result
 
 
+@typecheck(nd_coef=expr_ndarray(), nd_dep=expr_ndarray(), lower=expr_bool)
+def solve_tri(nd_coef, nd_dep, lower=False):
```
johnc1231, line 263: Let's give it the same name as scipy: https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve_triangular.html

Two reasons: 
1. It's good to be consistent with other tools
2. `tri` could mean `tridiagonal`, which is a different thing.

hail/src/main/scala/is/hail/expr/ir/functions/NDArrayFunctions.scala
```diff
@@ -35,6 +36,53 @@ object NDArrayFunctions extends RegistryFunctions {
...
+      val uplo = cb.newLocal[String]("dtrtrs_uplo")
+      cb.ifx(lower.boolCode(cb), cb.assign(uplo, const("L")), cb.assign(uplo, const("U")))
+
+      val infoDTRTRSResult = cb.newLocal[Int]("dtrtrs_result")
+
+      val trans = cb.newLocal[String]("dtrtrs_trans")
+      cb.assign(trans, const("N"))
+
+      val diag = cb.newLocal[String]("dtrtrs_diag")
+      cb.assign(diag, const("N"))
```
johnc1231, line 61: No need to make values for these. You are only going to use `trans` and `diag` once.

hail/src/main/scala/is/hail/expr/ir/functions/NDArrayFunctions.scala
```diff
@@ -35,6 +36,53 @@ object NDArrayFunctions extends RegistryFunctions {
...
+
+      val n = cb.newLocal[Long]("dtrtrs_n")
+      cb.assign(n, ndDepRow)
+
+      val nrhs = cb.newLocal[Long]("dtrtrs_nrhs")
+      cb.assign(nrhs, ndDepCol)
```
johnc1231, line 67: You can just directly pass `ndDepCol` and `ndDepRow` to the `dtrtrs`, no need to create `n` and `nrhs`.

hail/src/main/scala/is/hail/linalg/LAPACK.scala
```diff
@@ -108,6 +108,16 @@ object LAPACK {
 
     INFOref.getValue()
   }
+  def dtrtrs(UPLO: String, TRANS: String, DIAG: String, N: Int, NRHS: Int,
```
johnc1231, line 111: Newline between methods

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
```diff
@@ -107,6 +107,8 @@ object SNDArray {
               val i = info(n).indexToDim(idx)
               cb.assign(info(n).pos(i), info(n).pos(i) + info(n).strides(i))
             }
+
+
```
johnc1231, line 111: Delete

johnc1231 (changes requested):

johnc1231 (changes requested):
You're going to have to add `solve_triangular` to `/python/hail/docs/nd/index.rst` for it to show up in documentation.

johnc1231 (approved):

PR #10666, opened 2021-07-14
[query] Generate better code in lowering rule for MatrixAggregate

tpoterba:
With StackStruct, now we don't allocate in the `MakeTuple` here.
Should deforest better, and be faster in general.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/LowerMatrixIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/LowerMatrixIR.scala
@@ -814,13 +814,18 @@ object LowerMatrixIR {
         val lc = lower(child, ab)
         val idx = Symbol(genUID())
         TableAggregate(lc,
-          aggExplodeIR(filterIR(zip2(GetField(Ref("row", lc.typ.rowType), entriesFieldName),
-            GetField(Ref("global", lc.typ.globalType), colsFieldName),
-            ArrayZipBehavior.AssertSameLength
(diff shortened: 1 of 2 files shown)


hail/src/main/scala/is/hail/expr/ir/LowerMatrixIR.scala
```diff
@@ -812,17 +813,20 @@ object LowerMatrixIR {
...
-              __entries_field = 'row (entriesField),
-              __cols_field = 'global (colsField)) {
-              irRange(0, '__entries_field.len)
-                .filter(idx ~> !'__entries_field (idx).isNA)
-                .aggExplode(idx ~> aggLet(sa = '__cols_field (idx), g = '__entries_field (idx)) {
-                  subst(query, matrixSubstEnv(child))
-                })
-            })
+        TableAggregate(lc,
+          aggExplodeIR(filterIR(
```
patrick-schultz, line 817: The indentation is making this hard to parse. Could you maybe start the `filterIR` on the next line and indent its arguments, and put the filter body on its own line?

patrick-schultz (changes requested):

hail/src/main/scala/is/hail/expr/ir/LowerMatrixIR.scala
tpoterba: oh, sure.

chrisvittal (approved):

PR #10671, opened 2021-07-15
Fix formatting for metadata parameter type in export_vcf docs

nawatts:
Before

Screen Shot 2021-07-15 at 4 40 48 PM

https://hail.is/docs/0.2/methods/impex.html#hail.methods.export_vcf

After

Screen Shot 2021-07-15 at 4 45 01 PM

#assign compiler


tpoterba (approved):

johnc1231:
The test failure here is spurious, happening because batch is going through a transition / upgrade and things are a little broken right now. Will make sure this merges when that's resolved.

nawatts:
Thanks @johnc1231!

PR #10673, opened 2021-07-16
[batch] accept packets from jobs addressing themselves with an external IP

daniel-goldstein:


danking (approved):

daniel-goldstein:
Force merging because this will never pass with the current workers.

daniel-goldstein:
Reverted repo settings.

PR #10679, opened 2021-07-19
[batch] Prefix job hostnames with hostname_

daniel-goldstein:
Random hex strings, while valid, are a little dangerous to use as hostnames because some tools (Spark) could attempt to interpret an entirely numeric name as an IP address. This just avoids that danger entirely.


daniel-goldstein:
Succeeded on a 1200 job batch of starting spark.

danking (approved):

PR #10693, opened 2021-07-22
[batch] Replace docker on the worker with podman

daniel-goldstein:
This introduces a new version of the batch worker instance: one without `docker`. Instead we bring in `podman` to cover the functions of pulling images, extracting expanding filesystems from those images, and running the worker container. `podman` by default uses `crun` as its low-level runtime so we can get rid of the independent `crun` installation in the worker image. `podman` is daemonless and can be run rootless. For the most part you can't tell the difference, except this makes `podman` easy to run under multiple users with different caches per user. So if you ssh into a worker, be sure to `sudo` before any `podman` (or `crun`) command or else you might think nothing is running when in fact the worker is running under root's podman configuration. The podman/crun state directories are now shared between the host and the worker so `sudo crun list` on the worker should reveal the running job containers without having to exec into the worker first.

For the most part, `podman` is a drop-in replacement for `docker`, but there are a handful of inconsistencies that comprise most of this PR. One notable change is that we no longer persist any GCR credentials in the worker VM image so we authenticate again on start up.

cc: @jigold

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1,4 +1,5 @@
 from typing import Optional, Dict, Callable, Tuple, Awaitable, Any
+from shlex import quote as shq
 import os
 import json
 import sys
@@ -377,6 +378,7 @@ class Container:
         self.rootfs_path = None
         scratch = self.spec['scratch']
         self.container_scratch = f'{scratch}/{self.name}'
+        os.makedirs(self.container_scratch)
         self.container_overlay_path = f'{self.container_scratch}/rootfs_overlay'
         self.config_path = f'{self.container_scratch}/config'
         self.log_path = f'{self.container_scratch}/container.log'
@@ -494,9 +496,8 @@ class Container:
                     # FIXME improve the performance of this with a
                     # per-user image cache.
                     auth = self.current_user_access_token()
-                os.makedirs(self.container_scratch, exist_ok=True)
                 authfile_name = f'{self.container_scratch}/podman_auth.json'
-                with open(authfile_name, 'w+') as authfile:
+                with open(authfile_name, 'w') as authfile:
                     creds = base64.b64encode(f'{auth["username"]}:{auth["password"]}'.encode())
                     auth_json = {'auths': {'gcr.io': {'auth': creds.decode()}}}
                     authfile.write(json.dumps(auth_json))
@@ -529,9 +530,10 @@ class Container:
 
     async def extract_rootfs(self):
         assert self.rootfs_path
+        assert self.image_id
         os.makedirs(self.rootfs_path)
         await check_shell(
-            f'id=$(podman create {self.image_id}) && podman export $id | tar -C {self.rootfs_path} -xf - && podman rm $id'
+            f'id=$(podman create {shq(self.image_id)}) && podman export $id | tar -C {self.rootfs_path} -xf - && podman rm $id'
         )
         log.info(f'Extracted rootfs for image {self.image_ref_str}')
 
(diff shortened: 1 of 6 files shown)


batch/batch/worker/worker.py
```diff
@@ -281,34 +272,6 @@ async def _free(self, netns: NetworkNamespace):
             self.public_networks.put_nowait(netns)
 
 
-def docker_call_retry(timeout, name):
```
jigold, line 284: This makes me so happy!!!!

danking: likewise!

batch/batch/worker/worker.py
```diff
@@ -514,38 +473,38 @@ async def pull_image(self):
...
+                    auth = await self.batch_worker_access_token()
+                else:
+                    # Pull to verify this user has access to this
+                    # image.
+                    # FIXME improve the performance of this with a
+                    # per-user image cache.
+                    auth = self.current_user_access_token()
+                os.makedirs(self.container_scratch, exist_ok=True)
+                authfile_name = f'{self.container_scratch}/podman_auth.json'
+                with open(authfile_name, 'w+') as authfile:
```
danking, line 492: why open with truncation? Shouldn't this be a brand new file?

batch/batch/worker/worker.py
```diff
@@ -514,38 +473,38 @@ async def pull_image(self):
...
-            if e.status == 404 and 'pull access denied' in e.message:
+                if is_public_image:
+                    auth = await self.batch_worker_access_token()
+                else:
+                    # Pull to verify this user has access to this
+                    # image.
+                    # FIXME improve the performance of this with a
+                    # per-user image cache.
+                    auth = self.current_user_access_token()
+                os.makedirs(self.container_scratch, exist_ok=True)
```
danking, line 490: I feel like we should do this when the container class is created, not inside the image pull function.

batch/batch/worker/worker.py
```diff
@@ -514,38 +473,38 @@ async def pull_image(self):
...
+                authfile_name = f'{self.container_scratch}/podman_auth.json'
+                with open(authfile_name, 'w+') as authfile:
+                    creds = base64.b64encode(f'{auth["username"]}:{auth["password"]}'.encode())
+                    auth_json = {'auths': {'gcr.io': {'auth': creds.decode()}}}
+                    authfile.write(json.dumps(auth_json))
+                await check_exec_output('podman', 'pull', f'--authfile={authfile_name}', self.image_ref_str)
```
danking, line 496: What does podman pull do if the image already exists? Does it still do an auth check?

batch/batch/worker/worker.py
```diff
@@ -565,7 +524,7 @@ async def extract_rootfs(self):
         assert self.rootfs_path
         os.makedirs(self.rootfs_path)
         await check_shell(
-            f'id=$(docker create {self.image_id}) && docker export $id | tar -C {self.rootfs_path} -xf - && docker rm $id'
+            f'id=$(podman create {self.image_id}) && podman export $id | tar -C {self.rootfs_path} -xf - && podman rm $id'
```
danking, line 527: We should have always been shell quoting `self.image_id`. Can we add that now?

danking (dismissed):
I'm so stoked for this.

batch/batch/worker/worker.py
daniel-goldstein: I think this was a side-effect of doing the makedirs a little janky and not realizing what wasn't created. Can take out the `+`.

batch/batch/worker/worker.py
daniel-goldstein: I tested this manually and it looks like on every `podman pull`, Podman checks the remote manifest to know if it needs to pull anything new, and does the authentication then. I pulled the batch worker image, ran `podman logout gcr.io/hail-vdc`, checked to ensure that the image was listed in `podman images`, and then tried to pull it again but received an unauthorized error.

danking (approved):
n.b. you have a WIP tag on.

PR #10695, opened 2021-07-23
[query] Fix loop memory management

johnc1231:
Recur needs to emit its args into region 1, then copy them into region 2. Also the `emitI` result needs to emit into `loopRef.r1` before being copied into the main `region`.


johnc1231:
Trying to cook up a python test that wouldn't pass without this, surprisingly annoying.

tpoterba:
I think:

```
return hl.experimental.loop(
    lambda f, array, idx: hl.if_else(idx > 10,  array, f(array.extend(hl.str(idx)), idx + 1),
    hl.tarray(hl.tstr), ['foo'], 1)

```

tpoterba:
Could also do a scala test that asserts about memory usage.

johnc1231:
Added the python test, it returns bogus results in `main` so it's a good one.

tpoterba (approved):

PR #10697, opened 2021-07-23
[query] fix read(..., _n_partitions) with missing keys

danking:
I verified the test failed with the same [error as KC's](https://discuss.hail.is/t/repartition-on-read/2148/2) before my change.


tpoterba (approved):

tpoterba:
I think this might fail tests with an xpass on local backend.

danking:
`test_read_partitions` fails on the local (I copied the annotations from that test)

danking:
Hmm. So. _n_partitions is more like "desired _n_partitions"? Because a key can't be split across multiple partitions.

PR #10699, opened 2021-07-23
[compiler] Add Columnar STypes/ETypes

chrisvittal:


hail/src/main/scala/is/hail/types/physical/stypes/concrete/SStructOfArrays.scala
```diff
@@ -0,0 +1,144 @@
...
+    )
+  }
+}
+
+class SStructOfArraysSettable(
+  val st: SStructOfArrays,
+  val length: Settable[Int],
+  val missing: Option[Settable[Long]],
+  val fields: IndexedSeq[SIndexableValue with SSettable]
+) extends SIndexableValue with SSettable {
```
chrisvittal, line 75: We're going to want to add specialized methods for accessing only one field in a tight loop.

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SStructOfArrays.scala
```diff
@@ -0,0 +1,144 @@
...
+      elementsRequired,
+      structType.types.zip(fields).map { case (v, f) => f.castRename(TArray(v)).asInstanceOf }
+    )
+  }
+}
+
+class SStructOfArraysSettable(
+  val st: SStructOfArrays,
+  val length: Settable[Int],
+  val missing: Option[Settable[Long]],
```
chrisvittal, line 73: If the elements are optional and the field is optional, I intend to have the field missingness be an `OR` of these missing bits and the array's missing bits. That way accessing a single field can be entirely local to that field.

chrisvittal (commented):
Some initial comments.

danking:

PR #10702, opened 2021-07-26
[batch] Reduce non-essential warning / error messages

jigold:
I'm trying to address three separate error messages:

```
/usr/local/lib/python3.7/dist-packages/aiomysql/cursors.py:458: Warning: This version of MySQL doesn't yet support 'sorting of non-scalar JSON values'
```

Add more debug info to warning message with the query executed

```
  File "/usr/local/lib/python3.7/dist-packages/hailtop/utils/utils.py", line 775, in retry_long_running
    return await f(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/hailtop/utils/utils.py", line 812, in loop
    await f(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/batch/driver/instance_collection.py", line 181, in monitor_instances
    await asyncio.gather(*[check(instance) for instance in instances])
  File "/usr/local/lib/python3.7/dist-packages/batch/driver/instance_collection.py", line 179, in check
    await self.check_on_instance(instance)
  File "/usr/local/lib/python3.7/dist-packages/batch/driver/instance_collection.py", line 157, in check_on_instance
    assert last_start_timestamp is not None, f'lastStartTimestamp does not exist {spec}'
```

Handle case where last_start_timestamp is None

```
Failed to collect and upload profile: [Errno 32] Broken pipe
```

This is from the google cloud profiler. I reduced the logging level from error to warning for messages from this module.

### DIFF
--- a/batch/batch/driver/instance_collection.py
+++ b/batch/batch/driver/instance_collection.py
@@ -6,7 +6,7 @@ import dateutil.parser
 import collections
 from typing import Dict
 
-from hailtop.utils import time_msecs, secret_alnum_string, periodically_call
+from hailtop.utils import time_msecs, secret_alnum_string, periodically_call, time_msecs_str
 from hailtop import aiotools, aiogoogle
 from gear import Database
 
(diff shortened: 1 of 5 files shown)


gear/gear/database.py
```diff
@@ -21,6 +22,22 @@
...
+    if ws is None:
+        return
+    for w in ws:
+        msg = w[-1]
+        last_executed = getattr(self, '_last_executed', None)
+        executed = getattr(self, '_executed', None)
+        warnings.warn(f'{msg} {last_executed} {executed}', Warning, 4)
+
+
+aiomysql.Cursor._show_warnings = _show_warnings
```
daniel-goldstein, line 38: Will this print out the query with all its parameters? That makes me uneasy from both a security standpoint and generally overriding library methods. Is this very hard to replicate in dev? If it has to be in default then we should remove it quickly after diagnosing the bug.

batch/batch/driver/instance_collection.py
```diff
@@ -154,12 +154,16 @@ async def check_on_instance(self, instance):
...
-                await self.call_delete_instance(instance, 'activation_timeout')
+            if last_start_timestamp is not None:
+                last_start_time_msecs = dateutil.parser.isoparse(last_start_timestamp).timestamp() * 1000
+
+                if instance.state == 'pending' and time_msecs() - last_start_time_msecs > 5 * 60 * 1000:
+                    log.exception(f'{instance} did not activate within 5m after starting, deleting')
+                    await self.call_delete_instance(instance, 'activation_timeout')
+            else:
+                time_elapsed = time_msecs() - instance.time_created
+                if instance.state == 'pending' and time_elapsed > 5 * 60 * 1000:
```
daniel-goldstein, line 165: nit: the use of a `time_elapsed` variable is inconsistent between the two `if`s that are otherwise the same. I like having the variable

batch/batch/driver/main.py
```diff
@@ -64,6 +64,10 @@
 
 uvloop.install()
 
+google_cloud_profiler_logger = logging.getLogger('googlecloudprofiler')
+google_cloud_profiler_logger.addFilter(adjust_logging_level_error_to_warning)
+import googlecloudprofiler
+
```
daniel-goldstein, line 70: I'm inclined to be target noisy errors very specifically. The only noise I see from the profiler is failing to collect and upload the profile. That's annoying, but I would otherwise like to know when something else goes wrong. What do you think instead about demoting that particular error to an `info` instead of dropping everything from this dependency?

daniel-goldstein (changes requested):

gear/gear/database.py
jigold: Good call! I didn't even think of that. Hopefully attempt number 2 is cleaner.

daniel-goldstein (approved):

PR #10704, opened 2021-07-26
[asyncfs] fix deadlock in S3 backend

cseed:
boto is synchronous.  The AsyncFS create interface returns a writable file-like object, which means the S3 request has to be run in a thread in the background.  Originally, I was using the thread pool for that. However, if we fill the thread pool with create requests, then there are no threads for other operations, so client code can deadlock.  Instead, run the create requests in a new thread.

### DIFF
--- a/hail/python/hailtop/aiotools/s3asyncfs.py
+++ b/hail/python/hailtop/aiotools/s3asyncfs.py
@@ -3,6 +3,7 @@ from types import TracebackType
 from concurrent.futures import ThreadPoolExecutor
 import os.path
 import urllib
+import threading
 import asyncio
 import botocore.exceptions
 import boto3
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/aiotools/s3asyncfs.py
```diff
@@ -147,31 +149,33 @@ def __init__(self, mpc, number: int):
...
             exc_value: Optional[BaseException] = None,
             exc_traceback: Optional[TracebackType] = None) -> None:
         assert self._async_writable is not None
-        assert self._put_task is not None
-        try:
-            await self._async_writable.wait_closed()
-        finally:
-            self._mpc._etags[self._number] = await self._put_task
+        assert self._put_thread is not None
+        await self._async_writable.wait_closed()
```
daniel-goldstein, line 177: How come this is no longer `try`'d?

daniel-goldstein (changes requested):
Makes sense, just one question.

hail/python/hailtop/aiotools/s3asyncfs.py
cseed: So I didn't think it was necessary since the assignment to `self._value` happened on the thread, but, separately, I realized exceptions are not automatically propagated out of the thread at the `.join()`, so I added that.

daniel-goldstein (approved):

PR #10707, opened 2021-07-26
[query]Lowering import_gen

ammekk:
Created StringTableReader to replace MatrixGenReader. Forming resulting MatrixTable takes place mostly in python plus fully lowered.


johnc1231:
Some tests failing. But there are also two good tests that are passing now:

```
test_aggregator_info_score
test_logistic_regression_wald_dosage
```

You have to remove the `fails_local_backend` and `fails_service_backend` annotations from those tests since they pass now.

hail/python/hail/methods/impex.py
```diff
@@ -1293,12 +1293,68 @@ def import_gen(path,
...
-    return MatrixTable(ir.MatrixRead(ir.MatrixGENReader(
-        path, sample_file, chromosome, min_partitions, tolerance, rg, contig_recoding, skip_invalid_loci)))
+        contig_recoding = hl.empty_dict(hl.tstr, hl.tstr)
+    else:
+        contig_recoding = hl.dict(contig_recoding)
+
+    gen_table = gen_table.transmute(data=gen_table.text.split(' '))
+
+    if chromosome is None:
+        index = 5
```
johnc1231, line 1309: index on its own is a slightly confusing name for a constant value. What is it the index of?

hail/python/hail/methods/impex.py
```diff
@@ -1293,12 +1293,68 @@ def import_gen(path,
...
+        contig_holder = gen_table.data[0]
+    else:
+        index = 4
+        contig_holder = chromosome
+
+    contig_holder = contig_recoding.get(contig_holder, contig_holder)
+
```
johnc1231, line 1316: Maybe just write here: `position = gen_table.data[index - 2]` and then use that on the next several lines where you need it.

hail/python/hail/methods/impex.py
```diff
@@ -1293,12 +1293,68 @@ def import_gen(path,
...
+    alleles = hl.array([hl.str(gen_table.data[index - 1]), hl.str(gen_table.data[index])])
+    rsid = gen_table.data[index - 3]
+    varid = gen_table.data[index - 4]
+    gen_table = gen_table.annotate(locus=locus, alleles=alleles, rsid=rsid, varid=varid)
+    gen_table = gen_table.annotate(entries=gen_table.data[index + 1:].map(lambda x: hl.float64(x))
+                                   .grouped(3).map(lambda x: hl.struct(GP=x)))
+
+    sample_table_count = sample_table.count() - 2
+    gen_table = gen_table.annotate_globals(cols=hl.range(sample_table_count).map(lambda x: hl.struct(col_idx=x)))
+    gen_table = gen_table.filter(hl.is_defined(gen_table.locus))
```
johnc1231, line 1336: Only do this if `skip_invalid_loci` I think.

hail/python/hail/methods/impex.py
```diff
@@ -1293,12 +1293,68 @@ def import_gen(path,
...
+            locus = hl.locus(contig_holder, hl.int(gen_table.data[index - 2]), rg)
+
+    alleles = hl.array([hl.str(gen_table.data[index - 1]), hl.str(gen_table.data[index])])
+    rsid = gen_table.data[index - 3]
+    varid = gen_table.data[index - 4]
+    gen_table = gen_table.annotate(locus=locus, alleles=alleles, rsid=rsid, varid=varid)
+    gen_table = gen_table.annotate(entries=gen_table.data[index + 1:].map(lambda x: hl.float64(x))
+                                   .grouped(3).map(lambda x: hl.struct(GP=x)))
+
+    sample_table_count = sample_table.count() - 2
```
johnc1231, line 1334: Maybe just add a comment here explaining this 2. (I know why it's there, but for the next person.

hail/python/test/hail/methods/test_impex.py
```diff
@@ -1093,7 +1093,7 @@ def test_import_bgen_gavin_example(self):
         bgen_file = resource('example.8bits.bgen')
         bgenmt = hl.import_bgen(bgen_file, ['GT', 'GP'], sample_file)
         self.assertTrue(
-            bgenmt._same(genmt, tolerance=1.0 / 255, absolute=True))
+            bgenmt.head(2)._same(genmt.head(2), tolerance=1.0 / 255, absolute=True))
```
johnc1231, line 1096: Take out these `head` calls, that was just for us when debugging.

hail/src/main/scala/is/hail/io/vcf/LoadVCF.scala
```diff
@@ -1727,7 +1727,8 @@ class MatrixVCFReader(
 
     val fullRowPType: PType = fullRVDType.rowType
 
-    val bodyPType = (requestedRowType: TStruct) => fullRowPType.subsetTo(requestedRowType).asInstanceOf[PStruct]
+    val bodyPType = (requestedRowType: TStruct) =>
+      fullRowPType.subsetTo(requestedRowType).asInstanceOf[PStruct]
```
johnc1231, line 1731: Don't change this, put it back how it was.

hail/python/test/hail/methods/test_impex.py
```diff
@@ -1529,19 +1534,18 @@ def test_import_gen_no_reference_specified(self):
...
                            resource('skip_invalid_loci.sample'),
                            reference_genome='GRCh37',
                            skip_invalid_loci=True)
-        self.assertEqual(mt._force_count_rows(),
-                         3)
+        self.assertEqual(mt._force_count_rows(), 3)
 
         with self.assertRaisesRegex(FatalError, 'Invalid locus'):
-            hl.import_gen(resource('skip_invalid_loci.gen'),
+             mt = hl.import_gen(resource('skip_invalid_loci.gen'),
```
johnc1231, line 1545: Indented one extra space more than it should be.

johnc1231 (changes requested):

hail/python/hail/methods/impex.py
```diff
@@ -1293,12 +1293,70 @@ def import_gen(path,
...
+    if skip_invalid_loci:
+        gen_table = gen_table.filter(hl.is_defined(gen_table.locus))
+    mt = gen_table._unlocalize_entries('entries', 'cols', ['col_idx'])
+
+    sample_table = sample_table.tail(sample_table_count).add_index()
+    sample_table = sample_table.annotate(s=sample_table.text.split(' ')[0])
+    sample_table = sample_table.key_by(sample_table.idx)
+    mt = mt.annotate_cols(s=sample_table[hl.int64(mt.col_idx)].s)
+
+    mt = mt.annotate_entries(GP=hl.bind(lambda x: hl.if_else(hl.abs(1.0 - x) > tolerance,
```
johnc1231, line 1345: Not a complaint, just an FYI: We also have `rbind`, which lets you put the thing you're binding before the lambda. I find that easier to read.

johnc1231, line 1345: Separate comment: Don't use `x` in the lambda if it could have a better name. Let's say `gp_sum` or something instead.

hail/python/hail/methods/impex.py
```diff
@@ -1293,12 +1293,70 @@ def import_gen(path,
...
+                                                             hl.missing(hl.tarray(hl.tfloat64)),
+                                                             hl.abs((1 / x) * mt.GP)),
+                                        hl.sum(mt.GP)))
+    mt = mt.annotate_entries(GT=hl.bind(lambda x: hl.if_else(hl.len(mt.GP.filter(lambda y: y == mt.GP[x])) == 1,
```
johnc1231, line 1349: Use `rbind` here, it's hard to see that `hl.argmax(mt.GP)` is the `x` in the `lambda`. In fact, let's also name `x` `max_idx` or something instead.

johnc1231 (changes requested):

johnc1231 (changes requested):

PR #10712, opened 2021-07-27
[query] Lower BlockMatrixAgg

johnc1231:
`BlockMatrixAgg` is the IR node that sums along an axis. This PR lowers `BlockMatrixAgg`. It confusingly took different arguments from `NDArrayAgg`, in that it requested a list of axes you wanted to keep around, instead of a list of axes you want to sum out. Since the `sum` function in numpy, on hail ndarrays, and on hail `BlockMatrix` takes in a list of axes you want to sum out, I decided it would be better to make things consistent and changed the arguments to be consistent.


johnc1231:
bump @patrick-schultz

patrick-schultz (approved):

patrick-schultz:
Forgot I made an earlier comment when I hit approve, but I'm fine with merging as is.

PR #10713, opened 2021-07-27
[batch] Don't use group by with a JSON array value

jigold:
This PR eliminates this warning message:

```
This version of MySQL doesn't yet support 'sorting of non-scalar JSON values'
```

I tested this change in the SQL Fiddle and still got the correct answer. We could remove all non-varying fields from the group by, but decided to leave the other sortable fields as that is the SQL standard.


danking (approved):

danking:
Good work hunting this down; these messages are always a pain to find.

PR #10716, opened 2021-07-27
[hailtop.batch] Fix compile code and directly use resource paths in code

jigold:
CHANGELOG: Fixes writing large commands to GCS

I made a couple of changes here:
1. Try and format a readable version of the code (including the user code) for the dry run mode
2. Fix `interpolate_command` was not idempotent. Thus large commands with resource files would fail.
3. Use paths directly in paths by doing a double interpolation. I'm worried about the speed of having to doubling parse the command with the regex, but I think the improvement in readability is a big win.

### DIFF
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -98,7 +98,7 @@ class Job:
         def safe_str(s):
             new_s = ''
             for c in s:
-                if c.isalnum():
+                if c.isalnum() or c == '-':
                     new_s += c
                 else:
                     new_s += '_'
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -95,7 +95,15 @@ class Job:
         self._valid: Set[_resource.Resource] = set()  # resources declared in the appropriate place
         self._dependencies: Set[Job] = set()
 
-        self._dirname = name.replace(' ', '_') + f'-{self._token}' if name else self._token
+        def safe_str(s):
+            new_s = ''
+            for c in s:
+                if c.isalnum():
+                    new_s += c
+                else:
+                    new_s += '_'
+
+        self._dirname = f'{safe_str(name)}-{self._token}' if name else self._token
 
     def _get_resource(self, item: str) -> '_resource.Resource':
         raise NotImplementedError
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -96,12 +96,13 @@ class Job:
         self._dependencies: Set[Job] = set()
 
         def safe_str(s):
-            new_s = ''
+            new_s = []
             for c in s:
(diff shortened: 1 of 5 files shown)


jigold:
So there's a double regex substitution now in this version. I couldn't figure out how to avoid this without having nice error checking at the exact line there's a problem. For example, `j.command(f'{b}')` right now immediately errors with a nice error message. But if the error checking doesn't come until the massive parallel `_compile` in `Backend.run`, then it will be harder to tell where the error is. I thought about having a `debug_mode` which is on by default that does the double check while the `debug_mode` being off is more efficient.

jigold:
I need to double check this still works with the batch pool executor code.

jigold:
ah. this should be okay with BPE. It calls `j.command()`.

danking (dismissed):

jigold:
@danking Did we decide we're okay with double interpolation?

danking:
I don't have an objection.

jigold:
I'm going to think about this some more. I'm worried we're going to make submission for the million plus batches be twice as long which is not great.

jigold:
I fixed the code so that there's one interpolation step. What I did was use the job name plus a random token as the job directory name and the batch tmp directory is an environment variable. I'm happy with this setup. Then it will be easier for users to tell where files came from if they give jobs unique names.

jigold:
This should be good to go now.

hail/python/hailtop/batch/job.py
```diff
@@ -90,6 +95,16 @@ def __init__(self,
         self._valid: Set[_resource.Resource] = set()  # resources declared in the appropriate place
         self._dependencies: Set[Job] = set()
 
+        def safe_str(s):
+            new_s = ''
+            for c in s:
+                if c.isalnum() or c == '-':
+                    new_s += c
+                else:
+                    new_s += '_'
```
danking, line 104: Since strings in Python are immutable, this function isn't O(N). You can either use a [StringIO](https://docs.python.org/3/library/io.html#io.StringIO) to accumulate the characters and then call `getvalue` or you can create a list of characters and then call `''.join(ls)`

hail/python/hailtop/batch/job.py
```diff
@@ -452,25 +467,19 @@ def handler(match_obj):
...
-            return shq(r._get_path(local_tmpdir))
-        return handler
+        subst_command = re.sub('(' + ')|('.join(regexes) + ')',
+                               handler,
+                               command)
 
-    def _subst_resource_path(self, command, local_tmpdir):
-        handler = self._subst_resource_path_handler(local_tmpdir)
-        return self._process_command(command, handler)
+        return subst_command
```
danking, line 482: The block seems to be the same as `self._process_command(handler, command)`. Do we still need `_process_command`?

danking (changes requested):

danking (approved):

PR #10721, opened 2021-07-28
[batch] Don't cleanup images too quickly

jigold:


batch/batch/worker/worker.py
```diff
@@ -436,10 +437,11 @@ async def localize_rootfs():
                     await self.pull_image()
                     self.image_config = image_configs[self.image_ref_str]
                     self.image_id = self.image_config['Id'].split(":")[1]
-                    worker.image_ref_count[self.image_id] += 1
+                    worker.image_data[self.image_id] += 1
```
daniel-goldstein, line 440: I think you'll want to stick with the `defaultdict` or you'll get a `KeyError` here. Removing `id` from `ImageData` (it's already the corresponding key in the dictionary) should allow you to make `ImageData` have no-argument constructor and then I think it should work.

daniel-goldstein (approved):
I like this middle ground. Just one comment.

daniel-goldstein (dismissed):
See prev. message.

daniel-goldstein (approved):

PR #10723, opened 2021-07-29
[ci] Add more visibility into the PR update cycle

daniel-goldstein:
CI is updating a lot of PRs in its `update` loop that don't need to be updated. I've replicated and examined one such scenario in my own dev namespace but I feel like this visibility would be nice to have by default.

### DIFF
--- a/ci/ci/github.py
+++ b/ci/ci/github.py
@@ -1,3 +1,4 @@
+from typing import Dict, Optional
 import secrets
 from shlex import quote as shq
 import json
@@ -8,6 +9,7 @@ import aiohttp
 import gidgethub
 import zulip
 import random
+import prometheus_client as pc  # type: ignore
 
 from hailtop.config import get_deploy_config
 from hailtop.batch_client.aioclient import Batch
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
Mm I need to decrement when a PR is destroyed

ci/ci/github.py
```diff
@@ -277,6 +295,8 @@ def from_gh_json(gh_json, target_branch):
             {user['login'] for user in gh_json['requested_reviewers']},
             {label['name'] for label in gh_json['labels']},
         )
+        pr.increment_pr_metric()
```
danking, line 298: What about when a PR is merged and forgotten about?

danking: Shouldn't there be a matching decrement?

danking (changes requested):

ci/ci/github.py
daniel-goldstein: yup, there should be. Updated to decrement any tracked pr that is not seen in a new github update and subsequently dropped.

tpoterba:
oh, nice! Is this the issue we talked about that one time a few months ago?

I would love to be able to sort by "recently updated" accurately again!

daniel-goldstein:
Yup! I got a better understanding why PRs stuck on merge conflicts keep getting updated all of the time in my own tinkering, but I want to be sure of the general case before following up and this will be a useful metric.

danking (approved):

PR #10727, opened 2021-07-30
[svcr] improve performance of both sample_qc methods

tpoterba:
Also add test for vds.sampleqc.


pwc2 (approved):

hail/python/test/hail/vds/test_vds.py
```diff
@@ -38,3 +38,34 @@ def test_conversion_equivalence():
     svcr_readback = hl.vds.to_merged_sparse_mt(vds)
 
     assert svcr._same(svcr_readback)
+
+
+def test_sampleqc_old_new_equivalence():
```
pwc2, line 43: I realized after approving that we probably need the failure decorators here too

tpoterba: I think this should work -- the feature that breaks local/service backend above is the VCF combiner. This test just uses plan relational query infrastructure.

pwc2: Gotcha

tpoterba: lint failures though...

PR #10731, opened 2021-07-30
[batch] Report job resource utilization to influxdb

daniel-goldstein:
Stacked on #10725. More work needs to go in to make sure these metrics are correct (especially unsure about cpu utilization), and it might be better to get at the cgroups directly rather than using `psutil`.


daniel-goldstein:
Closing since it will be a while until we can refocus on this.

PR #10732, opened 2021-07-30
[batch,monitoring] Reduce cardinality of prometheus labels

jigold:


batch/batch/driver/main.py
```diff
@@ -1140,7 +1141,8 @@ async def on_startup(app):
 
     app['task_manager'].ensure_future(periodically_call(60, scheduling_cancelling_bump, app))
 
-    app['task_manager'].ensure_future(periodically_call(15, monitor_system, app))
+    if HAIL_SHOULD_MONITOR_SYSTEM:
+        app['task_manager'].ensure_future(periodically_call(15, monitor_system, app))
```
daniel-goldstein, line 1145: I think there's another way to handle this more broadly without adding complexity to the driver code. In the prometheus config, you can filter and alter labels on metrics that the kubernetes collector adds, like we do [here](https://github.com/hail-is/hail/blob/4834afaf434dfbadc731899b8374937e1dd8ca36/prometheus/prometheus.yaml#L112-L123). Unless there are google quota issues to worry about, I think it would be simpler to always have our services aggregate and display metrics, but just tell prometheus to ignore those whose `__meta_kubernetes_namespace` matches `pr-.*`.

daniel-goldstein (changes requested):

jigold:
I'm not sure how to actually test these changes.

daniel-goldstein:
I think it could be enough to `make deploy` prometheus in your own namespace, kick off a couple PR builds if there aren't any, and check that your prometheus doesn't contain any of those entries. You can run queries through the prometheus UIO.

jigold:
I tested this in my namespace and it should be good to go.

daniel-goldstein (approved):

PR #10733, opened 2021-07-30
[compiler] read sequence list fully in tabix parser

chrisvittal:
CHANGELOG: Fix a bug in tabix parsing when the size of the list of all sequences is large

BGZipInput stream will (validly) only copy to the end of its internal
buffer on a call to `int read(byte[] b, int off, int len)`. The internal
buffer is at most 64k, so the single `read` call when reading the
sequences could often not read the entire list. Replacing it with
readFully solves the issue.


tpoterba (approved):
nice! That's dastardly.

PR #10734, opened 2021-07-30
[compiler] Remove Type.fundamentalType

tpoterba:
This was unused, but was a performance nightmare -- the reconstruction of a single type triggered a full reallocation of the nested structure.

image


chrisvittal (approved):
wat

PR #10736, opened 2021-07-30
[infra] Refresh infra setup instructions

daniel-goldstein:
A little pieced together but these are my notes from getting batch set up in a new GCP project. Most is copied from @cseed's recent branch with a few additional fixes.

One not-great thing is that gateway fails to start if it can't resolve its upstream hosts, so I had to effectively remove the server blocks for anything not batch/batch-driver/auth by editing `letsencrypt/subdomains.txt` and then redeploying the gateway. Weird thing is that I deleted the batch deployment and new gateway pods started up fine, so perhaps it's something to do with kube-dns? Planning to look into this as a follow-up.

### DIFF
--- a/ci/bootstrap.py
+++ b/ci/bootstrap.py
@@ -69,7 +69,6 @@ class LocalJob:
 async def docker_run(*args: str):
     script = ' '.join([shq(a) for a in args])
     outerr = await check_shell_output(script)
-    out, err = outerr
     print(f'Container output: {outerr[0]}\n' f'Container error: {outerr[1]}')
 
     cid = outerr[0].decode('ascii').strip()
--- a/ci/bootstrap.py
+++ b/ci/bootstrap.py
@@ -69,7 +69,8 @@ class LocalJob:
 async def docker_run(*args: str):
     script = ' '.join([shq(a) for a in args])
     outerr = await check_shell_output(script)
-    print(outerr)
+    out, err = outerr
+    print(f'Container output: {outerr[0]}\n' f'Container error: {outerr[1]}')
 
     cid = outerr[0].decode('ascii').strip()
 
@@ -330,9 +331,6 @@ class Branch(Code):
 
 git checkout {shq(self._sha)}
 '''
-
-    def repo_dir(self):
-        return '.'
 
 
 async def main():
--- a/ci/bootstrap.py
+++ b/ci/bootstrap.py
@@ -18,10 +18,11 @@ from batch.driver.k8s_cache import K8sCache
 
 KUBERNETES_SERVER_URL = os.environ['KUBERNETES_SERVER_URL']
 DOCKER_PREFIX = os.environ['HAIL_DOCKER_PREFIX']
-DOCKER_ROOT_IMAGE = f'{DOCKER_PREFIX}/ubuntu:18.04'
+DOCKER_ROOT_IMAGE = os.environ['HAIL_DOCKER_ROOT_IMAGE']
+BATCH_WORKER_IMAGE = os.environ['BATCH_WORKER_IMAGE']
 
 
-def populate_secret_host_path(host_path: str, secret_data: Union[str, bytes]):
+def populate_secret_host_path(host_path: str, secret_data: Dict[str, bytes]):
     os.makedirs(host_path)
     if secret_data is not None:
         for filename, data in secret_data.items():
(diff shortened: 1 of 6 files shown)


danking:
nginx should need the services (i.e. domain names) to exist, not the deployments.

danking:
Since Jackie needs to use this stuff when we go to Azure, I think she should review it.

daniel-goldstein:
> nginx should need the services (i.e. domain names) to exist, not the deployments.

Ah, that makes sense. I suppose that's why the service definition of `router` used to exist along with gateway's k8s config.

ci/bootstrap.yaml
```diff
@@ -69,3 +69,22 @@ roleRef:
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: default-admin-cluster-admin-binding
+subjects:
+- kind: ServiceAccount
+  name: admin
+  namespace: default
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
```
jigold, line 89: Where is this cluster role defined?

ci/create_initial_account.py
```diff
@@ -21,7 +21,7 @@ async def insert(tx):
             ('creating', username, email, 1, 0),
         )
 
-    return await insert(db)
+    return await insert()
```
jigold, line 24: You might need to add this pylint message (used in batch/front_end.py) `# pylint: disable=no-value-for-parameter`

infra/README.md
```diff
@@ -144,16 +150,20 @@ You can now install Hail:
...
 
-- Deploy the bootstrap gateway.  Run `make deploy` in
-  $HAIL/bootstrap-gateway.
+- Deploy the bootstrap gateway.  Run `make -C $HAIL/bootstrap-gateway deploy`.
 
-- Create Let's Encrypt certs. Run `make run` in $HAIL/letsencrypt.
+- Create Let's Encrypt certs. Run `make -C $HAIL/letsencrypt run`.
 
-- Deploy the gateway.  Run `make deploy` in $HAIL/gateway.
+- Deploy the internal-gateway.  Run `make -C $HAIL/internal-gateway deploy`.
```
jigold, line 159: What happened to deploying the gateway?

infra/README.md
```diff
@@ -202,11 +221,14 @@ You can now install Hail:
   python3 ci/bootstrap.py hail-is/hail:main $(git rev-parse HEAD) test_batch_0
   ```
 
+- Deploy the gateway. First, edit `$HAIL/letsencrypt/subdomains.txt` to include
```
jigold, line 224: Ah, I see. This step moved here.

ci/bootstrap.py
```diff
@@ -232,8 +239,22 @@ async def run(self):
...
+                    security_options = []
+
                 main_cid, main_ok = await docker_run(
-                    'docker', 'run', '-d', *env_options, *mount_options, j._image, *j._command
+                    'docker',
+                    'run',
+                    '-d',
+                    *env_options,
+                    *mount_options,
+                    *security_options,
```
jigold, line 253: This change makes an assumption that you can always do `--entrypoint j._command[0] IMAGE j._command[1:]`. Does this work in all cases?

jigold (changes requested):

ci/bootstrap.py
```diff
@@ -66,17 +69,17 @@ def __init__(
 async def docker_run(*args: str):
     script = ' '.join([shq(a) for a in args])
     outerr = await check_shell_output(script)
+    print(outerr)
```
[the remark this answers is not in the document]
daniel-goldstein: I can clean it up, or change it to only print errors, but I found the output to be very useful. Otherwise when anything failed it would just say `FAILURE` with no other information.

ci/bootstrap.yaml
daniel-goldstein: It's a default role in k8s https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles

ci/create_initial_account.py
daniel-goldstein: Looks like files not in `ci/ci` don't get checked by pylint. I can do a separate pr to address all the linting issues?

ci/bootstrap.py
daniel-goldstein: This should work with single element command lists so I can't think of a case where it wouldn't work in all cases. That being said this is grabbed from Cotton's deadbeef changes so I can try without it.

ci/create_initial_account.py
jigold: Yes, let's get this in. That's definitely a bigger project.

jigold (approved):

infra/README.md
```diff
@@ -66,8 +67,10 @@ Instructions:
 
    gcp_zone = "<gcp-zone>"
 
+   gcp_location = "<gcp-region>"
```
danking, line 70: @daniel-goldstein, do you recall why this was added? It seems to me that there is now to distinct variables that refer to the same thing: the region in which to deploy. Is that true? Any reason for me to not change all occurrences of gcp_location to gcp_region?

daniel-goldstein: Hm I don't recall, I believe you should be ok making that change, but I would do a grep through the code first.

PR #10738, opened 2021-08-02
[hailtop.batch] More batch pool executor fixes

jigold:
A couple of fixes to batch pool executor to both get rid of orphaned running forever jobs and exception not retrieved errors:

- `asyncio.wait` does not retrieve results. I had to change waits to gathers with return_exceptions=True to get the behavior we want.
- A timeout error with `asyncio.wait_for` cancels the task automatically. Therefore, the previous code would never cancel the batch because the task was already "cancelled".
- I made `asyncio_cancel` idempotent and made sure we cancel the batch if the task has been cancelled to address the issue above. I added a check to see if the batch is running before cancelling. I'm ambivalent on whether this change is necessary.
- I added an explicit test now to make sure all batches are terminated. I think this is a good change, but the downstream consequences could be if this runs forever on a deploy (relies on an explicit timeout). Although, `test_hailtop_batch_*` has explicit timeouts. So I think we're good.

### DIFF
--- a/hail/python/hailtop/batch/batch_pool_executor.py
+++ b/hail/python/hailtop/batch/batch_pool_executor.py
@@ -463,7 +463,6 @@ class BatchPoolFuture:
         self.output_gcs = output_gcs
         self.fetch_coro = asyncio.ensure_future(self._async_fetch_result())
         executor._add_future(self)
-        self._cancelled_batch = False
 
     def cancel(self):
         """Cancel this job if it has not yet been cancelled.
@@ -479,15 +478,8 @@ class BatchPoolFuture:
         ``True`` is returned if the job is cancelled. ``False`` is returned if
         the job has already completed.
         """
-        async def cancel_batch():
-            await self.batch.cancel()
-            self._cancelled_batch = True
 
         if self.fetch_coro.cancelled():
-            status = await self.batch.status()
-            if status['state'] == 'running':
-                await cancel_batch()
-                return True
             return False
 
         if self.fetch_coro.done():
@@ -495,7 +487,7 @@ class BatchPoolFuture:
             self.fetch_coro.result()
             return False
 
-        await cancel_batch()
+        await self.batch.cancel()
         self.fetch_coro.cancel()
         await asyncio.wait([self.fetch_coro])
         return True
@@ -556,6 +548,8 @@ class BatchPoolFuture:
             return await asyncio.wait_for(self.fetch_coro, timeout=timeout)
         except asyncio.TimeoutError as e:
             raise concurrent.futures.TimeoutError() from e
+        finally:
+            await self.batch.cancel()
 
     async def _async_fetch_result(self):
         try:
--- a/hail/python/hailtop/batch/batch_pool_executor.py
+++ b/hail/python/hailtop/batch/batch_pool_executor.py
@@ -264,14 +264,14 @@ class BatchPoolExecutor:
                         bp_futures.append(bp_future)
 
                 if bp_futures:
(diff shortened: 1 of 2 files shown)


hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -478,21 +479,31 @@ async def async_cancel(self):
...
         the job has already completed.
         """
+        async def cancel_batch():
+            await self.batch.cancel()
+            self._cancelled_batch = True
+
         if self.fetch_coro.cancelled():
+            status = await self.batch.status()
+            if status['state'] == 'running':
+                await cancel_batch()
```
danking, line 489: Isn't cancel idempotent and doesn't care if the batch is finished or not? We shouldn't add a round trip to check the status if unnecessary.

hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -478,21 +479,31 @@ async def async_cancel(self):
         ``True`` is returned if the job is cancelled. ``False`` is returned if
         the job has already completed.
         """
+        async def cancel_batch():
```
danking, line 482: I think we should just add `self.batch.cancel` to the finally block of the fetch_coro. Cancelling a completed batch is fine. That keeps everything simple: if the fetch_coro is cancelled, the BatchPoolFuture has been cancelled.

danking (dismissed):

jigold:
Not 100% sure I understood the last comment.

hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -537,9 +546,10 @@ async def async_result(self, timeout: Optional[Union[float, int]] = None):
             raise concurrent.futures.CancelledError()
         try:
             return await asyncio.wait_for(self.fetch_coro, timeout=timeout)
-        except asyncio.TimeoutError:
-            await self.async_cancel()
-            raise
+        except asyncio.TimeoutError as e:
+            raise concurrent.futures.TimeoutError() from e
+        finally:
+            await self.batch.cancel()
```
danking, line 552: This only fires if someone calls async_result, right? I'm sorry I was a bit vague. When I said `fetch_coro`, I was thinking of `self._async_fetch_result`, which is the function called to start the fetch_coro. I think you add self.batch.cancel() to its finally block, then if fetch_coro is cancelled, the batch will be cancelled. The batch will also be cancelled if any error happens during result fetching or job waiting.

danking (dismissed):

hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -478,12 +478,15 @@ async def async_cancel(self):
...
+
         if self.fetch_coro.cancelled():
             return False
+
         if self.fetch_coro.done():
             # retrieve any exceptions raised
             self.fetch_coro.result()
             return False
+
         await self.batch.cancel()
```
danking, line 490: We should remove this as well because cancelling the fetch_coro (on the next line) will cancel the batch.

hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -537,9 +546,8 @@ async def async_result(self, timeout: Optional[Union[float, int]] = None):
             raise concurrent.futures.CancelledError()
         try:
             return await asyncio.wait_for(self.fetch_coro, timeout=timeout)
```
danking, line 548: We should shield the fetch_coro here. It's perfectly valid to call `.result` again after a timeout.

danking (changes requested):

danking (approved):

PR #10739, opened 2021-08-02
[query] Partition reader no ptype

johnc1231:
We want to make a `PartitionReader` that reads in lines of text, no notion of `PType` makes sense for this.

@tpoterba, does this make sense? I find the `Requiredness` stuff confusing, but I think this should do the same thing the old code did


tpoterba:
🚀

tpoterba (approved):

PR #10744, opened 2021-08-02
[query] Perform a shufflerectomy (remove the shuffle service and code)

tpoterba:


dev-docs/hail-query-lifecycle.md
```diff
@@ -73,15 +73,8 @@ When invoked from Python, the JVM backends perform the following:
      CollectDistributedArray IR which represents the execution of a
      stage of the pipeline.  See `LowerTableIR`.
    - Shuffles (distributed sorts) are implemented on a per-backend
```
lgruen, line 75: Very interesting! Is there already a design doc?

tpoterba: Notes from our first (only) meeting here:
https://gist.github.com/tpoterba/cd590b70ec303f23353b4dc4c7fdfa7d

daniel-goldstein:
As far as I'm aware, resolving the shuffler is the only thing `address` is doing at the moment. Is there a reason to keep that service around once this is removed?

danking:
No, let's get rid of it too after this merges.

danking (approved):

PR #10748, opened 2021-08-04
[query] SeqSample

johnc1231:
This PR adds `SeqSample(totalRange, numToSample)`, a stream IR that generates a stream of `numToSample` unique random integers in the range `totalRange`.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
@@ -454,7 +454,7 @@ object EmitStream {
 
 
       case SeqSample(totalSize, numToSample, _requiresMemoryManagementPerElement) =>
-        // Implemented based on http://proceedings.mlr.press/v130/shekelyan21a/shekelyan21a.pdf
+        // Implemented based on http://www.ittc.ku.edu/~jsv/Papers/Vit84.sampling.pdf Algorithm A
         emit(totalSize, cb).flatMap(cb) { totalSizeCode =>
           emit(numToSample, cb).map(cb) { numToSampleCode =>
             val totalSizeVal = totalSizeCode.asInt.memoize(cb, "seq_sample_total_size")
@@ -472,6 +472,9 @@ object EmitStream {
 
               override def initialize(cb: EmitCodeBuilder): Unit = {
                 cb.assign(len, numToSampleVal.asInt.intCode(cb))
+                cb.assign(nRemaining, numToSampleVal.intCode(cb))
+                cb.assign(candidate, 0)
+                cb.assign(elementToReturn, -1)
               }
 
               override val elementRegion: Settable[Region] = regionVar
--- a/hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
@@ -470,14 +470,20 @@ object EmitStream {
           end
       end
          */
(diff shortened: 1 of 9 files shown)


hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -452,6 +452,55 @@ object EmitStream {
...
+
+            val nRemaining = cb.newLocal[Int]("seq_sample_num_remaining", numToSampleVal.intCode(cb))
+            val candidate = cb.newLocal[Int]("seq_sample_candidate", 0)
+            val elementToReturn = cb.newLocal[Int]("seq_sample_element_to_return", -1) // -1 should never be returned.
+
+            val producer = new StreamProducer {
+              override val length: Option[EmitCodeBuilder => Code[Int]] = Some(_ => len)
+
+              override def initialize(cb: EmitCodeBuilder): Unit = {
+                cb.assign(len, numToSampleVal.asInt.intCode(cb))
```
patrick-schultz, line 474: I think you need to initialize all your state variables here (nRemaining, candidate, elementToReturn)

hail/src/test/scala/is/hail/expr/ir/EmitStreamSuite.scala
```diff
@@ -193,6 +193,19 @@ class EmitStreamSuite extends HailSuite {
...
+    val seq = evalStream(
+      SeqSample(
+        I32(20),
+        I32(5),
+        false
+      )
+    ).map(_.asInstanceOf[Int])
+    assert(seq.size == 5)
+    assert(seq.toSet.size == 5)
+    assert(seq.forall(e => e >= 0 && e < 20))
```
patrick-schultz, line 206: It's always hard to test random methods, but we should probably have something to check that this has the right statistical properties. I would suggest repeating `SeqSample(N, n)` a large number of times (maybe compile once and run repeatedly), where n is 2 or 3, and N is smallish, then count how many times each possible two element subset gets sampled in an NxN matrix. If the sampler is uniform, they should all be nearly equal.

patrick-schultz (changes requested):
Thanks John! Just a couple things.

johnc1231:
Double check that the test is what you want. I made the bounds pretty wide to make sure we never had an unlucky sporadic test failure.

hail/src/test/scala/is/hail/expr/ir/EmitStreamSuite.scala
```diff
@@ -193,6 +193,38 @@ class EmitStreamSuite extends HailSuite {
...
+      assert(seq.toSet.size == n)
+      assert(seq.forall(e => e >= 0 && e < N))
+    }
+
+
+    (0 until N).foreach { i =>
+      (i + 1 until N).foreach { j =>
+        val entry = results(i)(j)
+        // Expected value of entry is 5000
+        assert(entry > 4000 && entry < 6000)
```
patrick-schultz, line 223: It's safe to tighten this up. I'd suggest 4880 and 5650. The probability that not all entries are within that window is < 1e-5. (Expected value is 5263.)

patrick-schultz (dismissed):

hail/src/test/scala/is/hail/expr/ir/EmitStreamSuite.scala
johnc1231: Done, thanks. I mistakenly thought it was 5000.

patrick-schultz (approved):

PR #10749, opened 2021-08-04
[query] Add properly lowered matrix reader implementations

tpoterba:


tpoterba:
Adds code for now, but after some of the infrastructure that is used by whole-stage codegen goes in, we'll remove the TableValue paths (including MaybeIndexedReadZippedIterator) entirely.

chrisvittal (approved):

PR #10753, opened 2021-08-05
Lines iterator

ammekk:
Improve performance usage of StringTableReader by implementing StringTablePartitionReader instead of using rvbs.

### DIFF
--- a/hail/python/hail/ir/ir.py
+++ b/hail/python/hail/ir/ir.py
@@ -1142,8 +1142,7 @@ class ToDict(IR):
         self._type = tdict(self.a.typ['key'], self.a.typ['value'])
 
 
-class \
-        ToArray(IR):
+class ToArray(IR):
     @typecheck_method(a=IR)
     def __init__(self, a):
         super().__init__(a)
--- a/hail/python/hail/ir/ir.py
+++ b/hail/python/hail/ir/ir.py
@@ -1142,7 +1142,8 @@ class ToDict(IR):
         self._type = tdict(self.a.typ['key'], self.a.typ['value'])
 
 
-class ToArray(IR):
+class \
+        ToArray(IR):
     @typecheck_method(a=IR)
     def __init__(self, a):
         super().__init__(a)
--- a/hail/src/main/scala/is/hail/expr/ir/StringTableReader.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/StringTableReader.scala
@@ -59,7 +59,7 @@ case class StringTablePartitionReader(lines: GenericLines) extends PartitionRead
      requestedType: Type): IEmitCode = {
 
      context.toI(cb).map(cb) { partitionContext =>
-       val iter = cb.emb.genFieldThisRef[CloseableIterator[GenericLine]]("iter")
+       val iter = cb.emb.genFieldThisRef[CloseableIterator[GenericLine]]("string_table_reader_iter")
 
        val ctxMemo = partitionContext.asBaseStruct.memoize(cb, "string_table_reader_ctx")
        val fileName = cb.emb.genFieldThisRef[String]("fileName")
@@ -125,13 +125,11 @@ class StringTableReader(
 
   override def lower(ctx: ExecuteContext, requestedType: TableType): TableStage = {
     val fs = ctx.fs
-    println(requestedType.rowType)
     val lines = GenericLines.read(fs, fileStatuses, None, None, params.minPartitions, false, true)
     TableStage(globals = MakeStruct(FastSeq()),
       partitioner = RVDPartitioner.unkeyed(lines.nPartitions),
       dependency = TableStageDependency.none,
-      //TODO figure out if stream requires memory management per element
-      contexts = ToStream(Literal.coerce(TArray(lines.contextType), lines.contexts), true),
+      contexts = ToStream(Literal.coerce(TArray(lines.contextType), lines.contexts)),
       body = { partitionContext: Ref => ReadPartition(partitionContext, requestedType.rowType, StringTablePartitionReader(lines))
       }
     )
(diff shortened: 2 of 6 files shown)


hail/src/main/scala/is/hail/expr/ir/StringTableReader.scala
```diff
@@ -34,6 +39,78 @@ object StringTableReader {
...
+             cb.goto(LendOfStream)
+           })
+         }
+         override val element: EmitCode = EmitCode.fromI(cb.emb) { cb =>
+           val reqType: TStruct = requestedType.asInstanceOf[TStruct]
+
+           IEmitCode.present(cb, SStackStruct.constructFromArgs(cb, elementRegion, reqType,
+             EmitCode.fromI(cb.emb)(cb => IEmitCode.present(cb, SJavaString.construct(fileName))),
+             EmitCode.fromI(cb.emb)(cb => IEmitCode.present(cb, SJavaString.construct(line)))))
+         }
```
johnc1231, line 103: You are getting errors, I think it is because you need to use the `requestedType` when you do this and only pass in arguments that are requested. You are currently always passing in `fileName` and `line`. Think back to the old method where we only did `RegionValueBuilder.addAnnotation` if the field was in the requested type. Same idea here

hail/src/main/scala/is/hail/expr/ir/StringTableReader.scala
```diff
@@ -44,11 +121,24 @@ class StringTableReader(
 
   override def pathsUsed: Seq[String] = params.files
 
-  override def lower(ctx: ExecuteContext, requestedType: TableType): TableStage =
-    executeGeneric(ctx).toTableStage(ctx, requestedType)
```
johnc1231, line 48: You should also delete `executeGeneric`, now that you've deleted the place that calls it.

johnc1231 (changes requested):

hail/src/main/scala/is/hail/expr/ir/StringTableReader.scala
tpoterba: 🎉 🎉 🎉 🎉 🎉 🎉 🎉

hail/src/main/scala/is/hail/expr/ir/StringTableReader.scala
```diff
@@ -44,52 +123,32 @@ class StringTableReader(
...
+  override def lower(ctx: ExecuteContext, requestedType: TableType): TableStage = {
+    val fs = ctx.fs
+    println(requestedType.rowType)
+    val lines = GenericLines.read(fs, fileStatuses, None, None, params.minPartitions, false, true)
+    TableStage(globals = MakeStruct(FastSeq()),
+      partitioner = RVDPartitioner.unkeyed(lines.nPartitions),
+      dependency = TableStageDependency.none,
+      //TODO figure out if stream requires memory management per element
+      contexts = ToStream(Literal.coerce(TArray(lines.contextType), lines.contexts), true),
```
johnc1231, line 134: I thought about this, and it's unnecessary. So you can take away the `true` at the end of line 134 and delete the TODO. We need that to be true when the streams represent rows of a table within a partition. This stream is just some helper information to make the partitions.

hail/src/main/scala/is/hail/expr/ir/StringTableReader.scala
```diff
@@ -44,52 +123,32 @@ class StringTableReader(
...
+      body = { partitionContext: Ref => ReadPartition(partitionContext, requestedType.rowType, StringTablePartitionReader(lines))
+      }
+    )
+  }
 
-  override def apply(tr: TableRead, ctx: ExecuteContext): TableValue =
-    executeGeneric(ctx).toTableValue(ctx, tr.typ)
+  override def apply(tr: TableRead, ctx: ExecuteContext): TableValue = {
+    val ts = lower(ctx, tr.typ)
+    println(tr.typ)
```
johnc1231, line 142: No prints

hail/src/main/scala/is/hail/expr/ir/StringTableReader.scala
```diff
@@ -44,52 +123,32 @@ class StringTableReader(
 
   override def pathsUsed: Seq[String] = params.files
 
-  override def lower(ctx: ExecuteContext, requestedType: TableType): TableStage =
-    executeGeneric(ctx).toTableStage(ctx, requestedType)
+  override def lower(ctx: ExecuteContext, requestedType: TableType): TableStage = {
+    val fs = ctx.fs
+    println(requestedType.rowType)
```
johnc1231, line 128: No prints.

hail/src/main/scala/is/hail/expr/ir/StringTableReader.scala
```diff
@@ -34,6 +39,80 @@ object StringTableReader {
...
+     context: EmitCode,
+     partitionRegion: Value[Region],
+     requestedType: Type): IEmitCode = {
+
+     context.toI(cb).map(cb) { partitionContext =>
+       val iter = cb.emb.genFieldThisRef[CloseableIterator[GenericLine]]("iter")
+
+       val ctxMemo = partitionContext.asBaseStruct.memoize(cb, "string_table_reader_ctx")
+       val fileName = cb.emb.genFieldThisRef[String]("fileName")
+       val line = cb.emb.genFieldThisRef[String]("line")
```
johnc1231, line 66: Let's give these generated fields more unique names. Like `stpr_iter` or `string_table_reader_iter` instead of just `iter` (you only have to change the string name, not the scala variable name).

johnc1231 (changes requested):

hail/python/hail/ir/ir.py
```diff
@@ -1142,7 +1142,8 @@ def _compute_type(self, env, agg_env):
         self._type = tdict(self.a.typ['key'], self.a.typ['value'])
 
 
-class ToArray(IR):
+class \
+        ToArray(IR):
```
johnc1231, line 1146: Typo, fix.

johnc1231 (changes requested):

johnc1231:
Looks good except for one accidentally inserted backslash.

johnc1231 (approved):

PR #10754, opened 2021-08-06
[compiler] Fix bug in streamJoinLeftDistinct with multiple key repeats.

tpoterba:
The new tests in scala and python both catch this.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
@@ -1039,6 +1039,7 @@ object EmitStream {
                 val rxOut = mb.newEmitField(rEltType.copy(required = false)) // right value in joinF (may be missing while rx is not)
 
                 val pulledRight = mb.genFieldThisRef[Boolean]("join_right_distinct_pulledRight")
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -1091,6 +1092,7 @@ object EmitStream {
                     cb.define(LpullRight)
                     if (rightProducer.requiresMemoryManagementPerElement)
                       cb += rightProducer.elementRegion.clearRegion()
+                    cb.assign(pulledRight, true)
```
patrick-schultz, line 1095: If we initialize `c` to 1, then the first thing we do is pull from the right. That should remove the need for `pulledRight` and simplify the logic.

patrick-schultz (changes requested):

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
tpoterba: I don't think that's true -- we still need a bit that indicates we have to pull from both streams.

patrick-schultz: Ah, I see it now. Thanks!

patrick-schultz (approved):

PR #10756, opened 2021-08-06
More complete instructions for building hail on spark clusters

### LINKED ISSUE #10747: Cannot build from source - hail 0.2.74
Building from source - hail 0.2.74

`$ git clone https://github.com/hail-is/hail.git`
`$ cd hail`
`$ git checkout tags/0.2.74`
`$ cd hail`
`$ make install-on-cluster HAIL_COMPILE_NATIVES=1 SCALA_VERSION=2.12 SPARK_VERSION=3.1.2`
`...`
`curl -sSL https://storage.googleapis.com/hail-common/libsimdpp-2.1.tar.gz > libsimdpp-2.1.tar.gz
tar -xzf libsimdpp-2.1.tar.gz`

`c++ -o build/ibs.o -march=sandybridge -O3 -std=c++14 -Ilibsimdpp-2.1 -Wall -Wextra -fPIC -ggdb -fno-strict-aliasing -I../resources/include -I/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home/include -I/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home/include/darwin -MD -MF build/ibs.d -MT build/ibs.o -c ibs.cpp`

`make[1]: *** No rule to make target `lz4.h', needed by `build/Decoder.o'.  Stop.`
`make: *** [native-lib-prebuilt] Error 2`

I've installed lz4 and still get this error. 

https://hail.is/docs/0.2/getting_started.html?highlight=lz4#requirements no longer documents lz4 as required, but 0.2.74 is breaking on it.

johnc1231:
Resolves #10747

I sshed into a VM and tried to do a clean install based on the issue raised above. As noted there, lz4 was missing from our cluster install docs. I also noticed `pip` returns a nonzero exit code if it tries to install something but doesn't find it, so I added some `|| true` to prevent a confusing error message. I also updated our examples to use Spark 3 by default.

### DIFF
--- a/hail/Makefile
+++ b/hail/Makefile
@@ -252,18 +252,18 @@ upload-artifacts: $(WHEEL)
 .PHONY: install-editable
 install-editable: $(PYTHON_VERSION_INFO) $(INIT_SCRIPTS)
 install-editable: python/README.md $(PYTHON_JAR)
-	-$(PIP) uninstall -y hail
+	-$(PIP) uninstall -y hail || true
 	cd python && $(PIP) install -e .
 
 .PHONY: install
 install: $(WHEEL)
-	-$(PIP) uninstall -y hail
+	-$(PIP) uninstall -y hail || true
 	$(PIP) install $(WHEEL)
 
 .PHONY: install-on-cluster
 install-on-cluster: $(WHEEL)
 	sed '/^pyspark/d' python/requirements.txt | grep -v '^#' | xargs $(PIP) install -U
-	-$(PIP) uninstall -y hail
+	-$(PIP) uninstall -y hail || true
 	$(PIP) install $(WHEEL) --no-deps
 
 .PHONY: install-hailctl
--- a/hail/Makefile
+++ b/hail/Makefile
@@ -252,18 +252,18 @@ upload-artifacts: $(WHEEL)
 .PHONY: install-editable
 install-editable: $(PYTHON_VERSION_INFO) $(INIT_SCRIPTS)
 install-editable: python/README.md $(PYTHON_JAR)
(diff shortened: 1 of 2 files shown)


hail/Makefile
```diff
@@ -252,18 +252,18 @@ upload-artifacts: $(WHEEL)
 .PHONY: install-editable
 install-editable: $(PYTHON_VERSION_INFO) $(INIT_SCRIPTS)
 install-editable: python/README.md $(PYTHON_JAR)
-	-$(PIP) uninstall -y hail
+	-$(PIP) uninstall -y hail || true
```
chrisvittal, line 255: this `-` should suppress failure due to nonzero exit status, is it not?

johnc1231: Oh I see. I didn't know about that. I saw:

```
Cannot uninstall requirement hail, not installed
Makefile:265: recipe for target 'install-on-cluster' failed
make: [install-on-cluster] Error 1 (ignored)
```

but I guess the `ignored` part is coming from that `-` and so it's all fine?

chrisvittal: It should be. This https://www.gnu.org/software/make/manual/make.html#Errors-in-Recipes describes the syntax. We shouldn't need `|| true`.

chrisvittal (dismissed):
`|| true` is unnecessary for recipe lines prefixed with `-`

johnc1231:
Fixed. Thanks for teaching me something new about Makefiles.

chrisvittal (approved):
no problem, always happy to share what I know about arcane (though standard) tooling

PR #10761, opened 2021-08-06
[batch] ensure we cancel batches after BPE tests

danking:
BPE tests with timeouts need to explicitly cancel their jobs.

### DIFF
--- a/hail/python/test/hailtop/batch/test_batch_pool_executor.py
+++ b/hail/python/test/hailtop/batch/test_batch_pool_executor.py
@@ -103,11 +103,16 @@ def test_map_timeout():
             pass
         else:
             assert False
+        finally:
+            for f in bpe.futures:
+                f.cancel()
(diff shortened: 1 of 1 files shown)


jigold:
Why would we want the behavior to be that a user has to explicitly cancel batches on a failure? This code already does that:

```python3
        async def async_result_or_cancel_all(future):
            try:
                return await future.async_result(timeout=timeout)
            except Exception as err:
                for fut in futures:
                    fut.cancel()
                raise err
        if chunksize > 1:
            return (val
                    for future in futures
                    for val in await async_result_or_cancel_all(future))
        return (await async_result_or_cancel_all(future)
                for future in futures)
```

jigold (changes requested):
see comment

danking:
Yeah, you're right, `map` works differently from `submit`.

PR #10762, opened 2021-08-06
[batch] ensure batches are cancelled on error

danking:


jigold:
This needs a rebase now.

hail/python/hailtop/batch/backend.py
```diff
@@ -681,23 +681,25 @@ async def compile_job(job):
...
+
+            if open:
+                webbrowser.open(url)
+            if wait:
+                print(f'Waiting for batch {bc_batch.id}...')
+                status = bc_batch.wait()
+                print(f'batch {bc_batch.id} complete: {status["state"]}')
+            return bc_batch
+        except:
+            bc_batch.cancel()
```
jigold, line 705: I think we should raise here.

jigold (changes requested):

hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -241,38 +244,24 @@ async def async_map(self,
...
-                bp_futures = []
-                for task in tasks:
-                    assert task.done()
-                    if not task.cancelled() and not task.exception():
-                        bp_future = task.result()
-                        bp_futures.append(bp_future)
-
-                if bp_futures:
-                    await asyncio.gather(*[bp_fut.async_cancel() for bp_fut in bp_futures], return_exceptions=True)
+            bp_futures = [await t for t in submit_tasks]
```
jigold, line 250: This is not in parallel anymore. Was it intentional to do this?

jigold (dismissed):

hail/python/hailtop/batch/batch_pool_executor.py
danking: The `ensure_future` on line 247 places all the tasks into the background. Each await here suspends execution and lets one of those background tasks make progress.

jigold (approved):

PR #10769, opened 2021-08-10
[query] refactor PCNDArray constructors

patrick-schultz:
This adds a `constructUninitialized` constructor to `PCanonicalNDArray`. Now that we have interface methods that mutate an ndarray, there's no reason to force initialization of the data to happen inside the constructor. And there are plenty of uses, like out parameters for lapack methods, that just want an uninitialized ndarray.

Also, it seemed silly to be forced to memoize the new uninitialized ndarray before filling its data, when the constructor already had everything in locals, so I changed the ndarray constructors to return `SNDArrayValue`s.


johnc1231 (approved):
This is a great change.

PR #10771, opened 2021-08-11
[aiogoogle] Fix GetObjectStream.read() to return all data

jigold:
I'm hoping this fixes the JSONDecodeError bug we saw when merging #10331. We assume that AsyncFS.read(n) always returns exactly n bytes unless the EOF is reached.


hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -310,7 +310,15 @@ def __init__(self, resp):
...
-        return await self._content.read(n)
+        data = bytearray()
+        while n == -1 or n > 0:
+            b = await self._content.read(n)
+            if not b:
+                break
+            data.extend(b)
+            if n != -1:
+                n -= len(b)
+        return bytes(data)
```
danking, line 321: Hmm. If `not b`, and `n > 0` then we must raise an UnexpectedEOF, right? 

I find it easier to convince myself I'm right if I write it in two pieces like this:

```python3
data = bytearray()

if n == -1:
    b = await self._content.read(n)
    data.extend(b)
    while b:
        b = await self._content.read(n)
        data.extend(b)
    return bytes(data)

while n > 0:
    b = await self._content.read(n)
    if not b:
        raise UnexpectedEOF()
    data.extend(b)
    n -= len(b)
assert n == 0
return bytes(data)

```

danking (changes requested):

hail/python/hailtop/aiogoogle/client/storage_client.py
jigold: Actually, the python interface just gives you at most n bytes requested. If you request 5 bytes and there's only 3, then you just get 3 bytes with no error.

danking: Something seems off to me. The contract of Python's `read(size)` is:
> Read up to size bytes from the object and return them.

The contract of the `read(n)` in this PR is:
> Read exactly size bytes from the object unless we encounter EOF in which case read fewer than size bytes.


Who uses `GetObjectStream`? Why do they want the aforementioned contract? I'd think think they either want the contract of Python's read or the "read exactly n bytes or raise" contract.

danking (changes requested):

hail/python/hailtop/aiogoogle/client/storage_client.py
jigold: `GetObjectStream` is returned by `StorageClient.get_object` which is returned by `GoogleStorageAsyncFS.read` and `GoogleStorageAsyncFS.open_from`. So when we read a file, we call `read` on the `GetObjectStream`. The LocalAsyncFS reads from a file handle. This returns all the bytes. For example, we write code like this when using the Python `open` function.

```python3
with open('file', 'r') as f:
     data = f.read()
```

So `AsyncFS.read()` can not be used like this if we don't return all bytes when reading from the stream.

```python3
    async def read(self, url: str) -> bytes:
        async with await self.open(url) as f:
            return await f.read()
```

It would need to be like this:

```python3
    async def read(self, url: str) -> bytes:
        async with await self.open(url) as f:
            data = bytearray()
            b = f.read()
            data.extend(b)
            while b:
                 b = await f.read()
                 data.extend(b)
```

jigold:
Would be great if we could hash this out. This PR / fix is blocking getting the PR in that uses the AsyncFS everywhere.

hail/python/hailtop/aiogoogle/client/storage_client.py
```diff
@@ -310,7 +310,23 @@ def __init__(self, resp):
 
     async def read(self, n: int = -1) -> bytes:
```
danking, line 311: OK, sorry about all the back and forth.

I think based on our Zulip thread the right answer is: we need a new method called `readexactly`. It should take one argument, `n` and it should read exactly that many bytes. I think we should also `assert n > 0` because it doesn't make any sense to `readexactly(-1)`.

danking: We'll then need to make corresponding changes to the other PR so that where it needs exactly n bytes, it asks for exactly n bytes.

danking (changes requested):

hail/python/hailtop/aiotools/stream.py
```diff
@@ -99,6 +106,20 @@ async def read(self, n: int = -1) -> bytes:
             return await blocking_to_async(self._thread_pool, self._f.read)
         return await blocking_to_async(self._thread_pool, self._f.read, n)
 
+    def _readexactly(self, n: int) -> bytes:
+        assert n >= 0
+        data: List[bytes] = []
+        while n > 0:
+            block = self._f.read(n)
+            if len(block) != n:
+                raise IncompleteRead()
```
danking, line 115: I think this should be if the `len(block) == 0`. [`RawIOBase.read`](https://docs.python.org/3/library/io.html#io.RawIOBase.read) is allowed to return fewer than `n` bytes. The docs indicate that a zero length return value when n is non-zero indicates EOF:
> If 0 bytes are returned, and size was not 0, this indicates end of file.

danking (changes requested):

hail/python/hailtop/aiotools/stream.py
danking: Also, I think we should use UnexpectedEOFError from fs.py. It's already used for exactly this kind of thing in the copy tool.

hail/python/hailtop/aiotools/stream.py
```diff
@@ -99,6 +106,20 @@ async def read(self, n: int = -1) -> bytes:
             return await blocking_to_async(self._thread_pool, self._f.read)
         return await blocking_to_async(self._thread_pool, self._f.read, n)
 
+    def _readexactly(self, n: int) -> bytes:
+        assert n >= 0
+        data: List[bytes] = []
+        while n > 0:
+            block = self._f.read(n)
+            if len(block) != n:
```
danking, line 114:
```suggestion
            if len(block) == 0:
```

hail/python/hailtop/aiotools/stream.py
jigold: It's not used in the copy tool. I moved it to stream.py anyways because of import issues.

jigold:
I'm still worried about this implementation of `read`:

```
class AsyncFS:
    async def read(self, url: str) -> bytes:
        async with await self.open(url) as f:
            return await f.read()

    async def read_from(self, url: str, start: int) -> bytes:
        async with await self.open_from(url, start) as f:
            return await f.read()
```

jigold:
I think this needs a while loop here then. The point of these functions was to read all of the data needed.

danking:
`read()` with no argument is `read(-1)`. Python's [`read`](https://docs.python.org/3/library/io.html#io.RawIOBase.read) and aiohttp's [`read`](https://docs.aiohttp.org/en/stable/streams.html#aiohttp.StreamReader.read) will return all bytes until EOF if their argument is `-1`.

Python:
> Read up to size bytes from the object and return them. As a convenience, if size is unspecified or -1, all bytes until EOF are returned. Otherwise, only one system call is ever made. Fewer than size bytes may be returned if the operating system call returns fewer than size bytes.
> 
> If 0 bytes are returned, and size was not 0, this indicates end of file. If the object is in non-blocking mode and no bytes are available, None is returned.
>
> The default implementation defers to readall() and readinto().

Aiohttp:
> Read up to n bytes. If n is not provided, or set to -1, read until EOF and return all read bytes.
>
> If the EOF was received and the internal buffer is empty, return an empty bytes object.

We should document that our `read` follows these same rules in `hailtop.aiotools.stream.ReadableStream.read`.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -522,10 +522,6 @@ def add_source_reports(transfer_report):
             print(f'  {sr._source}: {sr._files} files, {humanize.naturalsize(sr._bytes)}')
 
 
-class UnexpectedEOFError(Exception):
```
danking, line 525: This is going to cause a conflict when you rebase on main because of commit 89abe08339 which changes `_copy_part` to correctly throw UnexpectedEOFError when it does not read the entire part.

Let's put the exceptions into `aiotools/exceptions.py`.

danking (changes requested):

danking (approved):

PR #10772, opened 2021-08-11
[grafana] Add grafana authorization token to prometheus requests

jigold:
I finally figured out how to get the authorization bearer token for the Grafana robot into Grafana automatically. The problem I'm running into right now is when we load a datasource from a configuration file, we can not edit any of the settings in the UI. We'd want to make sure all the prometheus settings we want are inside the new config file. I also don't want to accidentally overwrite any of the existing configuration.

### DIFF
--- a/grafana/deployment.yaml
+++ b/grafana/deployment.yaml
@@ -23,10 +23,28 @@ spec:
         - name: grafana-configmap-volume
           configMap:
             name: grafana-config
+        - name: grafana-datasources-configmap-volume
+          configMap:
+            name: grafana-datasources-config
         - name: ssl-config-grafana
           secret:
             optional: false
             secretName: ssl-config-grafana
+        - name: grafana-tokens
+          secret:
+            secretName: grafana-tokens
+        - name: grafana-shared
+          emptyDir: {}
+      initContainers:
+       - name: setup-tokens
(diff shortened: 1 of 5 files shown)


grafana/deployment.yaml
```diff
@@ -47,6 +65,12 @@ spec:
             name: grafana-storage
           - mountPath: /etc/grafana
             name: grafana-configmap-volume
+          - mountPath: /etc/grafana/provisioning/datasources
+            name: grafana-datasources-configmap-volume
+          - mountPath: /grafana-tokens
```
daniel-goldstein, line 70: This shouldn't have to be mounted in the grafana container if you're moving the relevant information to `grafana-shared`, right?

daniel-goldstein (changes requested):
This looks awesome, thanks! Just one small comment. Regarding the prometheus config, I think it might be best to just pull in the additional changes from my other PR #10627.

daniel-goldstein:
It also looks like we might be able to set datasource as [editable](https://grafana.com/docs/grafana/latest/administration/provisioning/#example-data-source-config-file) in case we need to make any immediate fixes after this merges.

jigold:
I hopefully fixed this branch. @daniel-goldstein Do you want to reconcile it with #10627?

PR #10774, opened 2021-08-11
[batch] Only mount the gce deploy config for private jobs

daniel-goldstein:


batch/batch/front_end/front_end.py
```diff
@@ -832,6 +824,15 @@ async def create_jobs(request, userdata):
                             'mount_in_copy': False,
                         }
                     )
+                    if spec.get('network') == 'private':
```
lgruen, line 827: I believe the problem with not mounting _any_ deployment config for public jobs is that we can't override the domain. E.g. for the CPG deployment, we set `"domain": "hail.populationgenomics.org.au"` in the deployment config, so we don't send requests to `batch.hail.is`.

daniel-goldstein: Yes you're right, I instead need to mount the relevant deploy config for the environment that the batch is running in. I was hoping delegating to the client could be sufficient but I'm now realizing that you must not patch the client and that this doesn't even work for dev/pr namespaces. I am beginning to not like the option of the default config at all… it is one of the remaining places where `hail.is` is hard coded in the codebase.

daniel-goldstein:
Closing in favor of #10796.

PR #10777, opened 2021-08-12
[combiner] Add a combiner submodule that works with VariantDataset

chrisvittal:


hail/python/hail/experimental/vcf_combiner/vds_combiner.py
```diff
@@ -0,0 +1,190 @@
...
+        _transform_variant_function_map[mt.row.dtype] = f
+    transform_row = _transform_variant_function_map[mt.row.dtype]
+    return Table(TableMapRows(mt._tir, Apply(transform_row._name, transform_row._ret_type, TopLevelReference('row'))))
+
+
+def make_reference_table(mt) -> Table:
+    mt = mt.filter(hl.is_defined(mt.info.END))
+
+    def make_entry_struct(e, row):
+        reference_fields = {k: v for k, v in e.items() if k in ('DP', 'GQ', 'MIN_DP')}
```
tpoterba, line 92: We really can't hard-code this schema. I think the two options are to force the user to pass the reference block fields in explicitly, or compute them automatically from the data. I think I prefer the second, because it's the easiest user experience.

hail/python/hail/experimental/vcf_combiner/vds_combiner.py
```diff
@@ -0,0 +1,190 @@
...
+
+_transform_variant_function_map = {}
+_transform_reference_fuction_map = {}
+
+
+def make_variants_table(mt, info_to_keep=[]) -> Table:
+    mt = mt.filter(hl.is_missing(mt.info.END))
+
+    if mt.row.dtype not in _transform_variant_function_map:
+        def get_lgt(e, n_alleles, has_non_ref, row):
```
tpoterba, line 18: General thought, but we should make a scode function for this. It'll be way better.

hail/python/hail/experimental/vcf_combiner/vds_combiner.py
```diff
@@ -0,0 +1,190 @@
...
+                    __entries=row.__entries.map(
+                        lambda e: make_entry_struct(e, alleles_len, has_non_ref, row)))),
+            mt.row.dtype)
+        _transform_variant_function_map[mt.row.dtype] = f
+    transform_row = _transform_variant_function_map[mt.row.dtype]
+    return Table(TableMapRows(mt._tir, Apply(transform_row._name, transform_row._ret_type, TopLevelReference('row'))))
+
+
+def make_reference_table(mt) -> Table:
+    mt = mt.filter(hl.is_defined(mt.info.END))
```
tpoterba, line 89: Add type hints to these? Either this is a table and `mt` is misleading, or it's a matrix table and will fail with no such method 'filter'

hail/python/hail/experimental/vcf_combiner/vds_combiner.py
```diff
@@ -0,0 +1,190 @@
+from typing import List
```
tpoterba, line 1: I think this should be in the `hl.vds` module! We should phase out the experimental module when the new stuff is documented,tested, and performant.

tpoterba (changes requested):
Thanks for getting this up so fast!

hail/python/hail/experimental/vcf_combiner/vds_combiner.py
chrisvittal: Ok. In the update, I pushed a method to get the used fields for reference block entries. For combining we can populate it from an already extant VariantDataset, or from the first gvcf in order to not scan every gvcf.

hail/python/hail/vds/combiner.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def make_reference_matrix_table(mt: MatrixTable, entry_to_keep) -> MatrixTable:
+    mt = mt.filter_rows(hl.is_defined(mt.info.END))
+
+    if not entry_to_keep:
+        # set difference here, since all GTs are 0/0 and for a homref call,
+        # LAD and LPL are uninteresting and trivially recreatable assuming
+        # DP is defined (which it should be!).
+        entry_to_keep = defined_entry_fields(mt, sample=10_000) - {'AD', 'GT', 'PL'}
```
tpoterba, line 109: I think we shouldn't make this transformation by default, especially unchecked -- we should strive to leave the data model basically untouched with respect to the input gvcfs. This might drop data (it definitely would for the 1kg gvcfs we use in tests).

Happy to drop GT (we should probably insert a runtime check that it's 0/0 if the END field is defined), but I think we should let the others go through the default handling (leave them in ref data if they're defined at ref blocks)

chrisvittal: I did an experiment, for every 1kg gvcf, after the initial transformation, every reference LAD or LPL was either missing or `LAD == [DP]` and `LPL == [0]`

tpoterba: I realize that *our* data generally has this property, but I still object to building in this assumption. It means we're tailoring the combiner specifically to GATK output (in a certain version range likely), and leads to lossy transformations of certain inputs.

hail/python/test/hail/vds/test_vds.py
```diff
@@ -73,3 +73,32 @@ def test_sampleqc_old_new_equivalence():
     assert res.aggregate_cols(hl.all(
         *(hl.agg.all(res.sample_qc[field] == res.sample_qc_new[field]) for field in fields_to_test)
     ))
+
+
+@fails_local_backend
```
tpoterba, line 78: these need parens

hail/python/hail/vds/combiner.py
tpoterba: I do still feel strongly about this point

tpoterba: I think ready to go once we fix this

chrisvittal: We drop PL for reference calls in the current VCF combiner. I've removed the removal of AD in favor of proper handling with translation to LAD (if it's there). For PL, I could have an LPL array, which would change the behavior from the current combiner.

chrisvittal:
Depending on desired PL behavior, there's one last mismatch between the new combiner and the old one.

The new combiner doesn't compute `RGQ` at all for reference alleles, I think this is fine.

tpoterba:
> Depending on desired PL behavior, there's one last mismatch between the new combiner and the old one.

> The new combiner doesn't compute RGQ at all for reference alleles, I think this is fine.

I think this is OK. Ship it!

tpoterba (approved):

PR #10778, opened 2021-08-12
[hailtop] add hailtop.aiotools.copy

danking:
God it feels good to make a pull request again.

The following operation copies every file in gs://danking/test/ into a folder named s3://hail-test-bucket/test/.

    python3 -m hailtop.aiotools.copy null '[{"from": "gs://danking/test", "to": "s3://hail-test-bucket/test"}]'

The `null` can be replaced with a quoted GCP project id if one of the sources is a Requester Pays bucket. For example:

    python3 -m hailtop.aiotools.copy '"broad-ctsa"' '[{"from": "gs://hail-datasets-us/", "to": "s3://hail-datasets-us-east-1"}]'

As you can see, the syntax is rough, but expressive. Explicitly listing all the files to be transferred should not impair the throughput.

    python3 -m hailtop.aiotools.copy null '[
        {"from": "gs://danking/test/data1", "to": "s3://hail-test-bucket/test/data1"},
        {"from": "gs://danking/test/data2", "to": "s3://hail-test-bucket/test/data2"},
        {"from": "gs://danking/test/sub-directory1", "to": "s3://hail-test-bucket/test/sub-directory1"},
        ...
    ]'

Better syntax is forthcoming: https://github.com/hail-is/hail/pull/9913


hail/python/hailtop/aiotools/fs.py
```diff
@@ -218,6 +218,16 @@ def _copy_part_size():
         the destination filesystem is used.'''
         return 128 * 1024 * 1024
 
+    async def copy(self,
```
danking, line 221: This, oddly, was only defined on a RouterAsyncFS. It's totally FS agnostic though.

hail/python/hailtop/aiotools/fs.py
```diff
@@ -897,7 +907,7 @@ def __init__(self, default_scheme: Optional[str], filesystems: List[AsyncFS]):
                     scheme_fs[scheme] = fs
                     schemes.add(scheme)
 
-        if default_scheme not in schemes:
+        if default_scheme is not None and default_scheme not in schemes:
```
danking, line 910: this was just a bug I found; default_scheme was assumed to be non-None.

danking:
I mishandled the empty scheme case.

jigold (approved):

PR #10770, opened 2021-08-11
[query] more lapack bindings, support using mkl

patrick-schultz:
Adds support for using mkl, which is needed for some of the added lapack methods.

To use, install mkl, then set the environment variable `HAIL_MKL_PATH` to the directory containing the mkl dylib files. In particular, it should contain `libmkl_rt.dylib`.

On my laptop, I installed mkl through conda with `conda install -c intel mkl`, then set `HAIL_MKL_PATH=~/opt/miniconda3/envs/hail/lib/`.

PR #10783, opened 2021-08-16
[query] Preserve static NDArray shape information

patrick-schultz:
~Stacked on #10770~
~Stacked on #10791~

This PR attempts to allow linear algebra codegen methods, like the LAPACK wrappers and the local whitening methods I'm working on, to defensively assert shape compatibility preconditions, without generating redundant runtime checks. (I always hate when we're pushed to avoid using code generation abstractions (in this case, just factoring code into smaller functions), because they generate worse code.)

The method is pretty simple. SNDArray shapes are now arrays of `SizeValue`, which is a sum type with cases `SizeValueDyn(Value[Long])` and `SizeValueStatic(Long)`. I don't think static sizes occur very often, but it was a simple addition.

`SizeValue`s can be compared statically with `==`, or at runtime with `ceq`: the former is true only if we can prove statically that the two sizes must be equal, while the latter emits code to check equality at runtime, using static knowledge to elide dynamic checks where possible.

The way we encode static knowledge that two sizes are equal is by using the same local variable to store both. The primary interface to introduce that static knowledge (other than using the same set of sizes to construct multiple SNDArrays), is the method `coerceToShape(cb: CodeBuilder, newShape: Seq[SizeValue]): SNDArrayValue`, which emits code to dynamically assert that `this.shape` agrees with `newShape`, then returns `this` with shape replaced by `newShape`. Thus, `a.coerceToShape(cb, newShape).shape == newShape` will always be true, preserving the static knowledge about the shape of `a`.

As a simple example, `gemm` verifies its inputs with (simplifying to the case with no transposes)
```
    val Seq(m, n) = C.shapes
    val k = A.shapes(1)
    A.assertHasShape(cb, FastIndexedSeq(m, k), errMsg)
    B.assertHasShape(cb, FastIndexedSeq(k, n), errMsg)
```
If we call this with
```
val m, n, k = \\ compute expected dim sizes

\\ emit dynamic size checks once
val A_ = A.coerceToShape(cb, IndexedSeq(m, k))
val B_ = B.coerceToShape(cb, IndexedSeq(k, n))
val C_ = C.coerceToShape(cb, IndexedSeq(m, n))

\\ this generates no new dynamic shape checks
SNDArray.gemm(cb, A, B, C)
```
then we can emit one set of runtime shape assertions, and further calls to methods with shape preconditions generate no new runtime assertions.

### DIFF
--- a/hail/src/main/scala/is/hail/types/physical/stypes/concrete/SUnreachable.scala
+++ b/hail/src/main/scala/is/hail/types/physical/stypes/concrete/SUnreachable.scala
@@ -255,17 +255,19 @@ class SUnreachableNDArrayValue(val st: SUnreachableNDArray) extends SUnreachable
 
   def loadElementAddress(indices: IndexedSeq[is.hail.asm4s.Value[Long]],cb: is.hail.expr.ir.EmitCodeBuilder): is.hail.asm4s.Code[Long] = const(0L)
 
-  def shapes(cb: EmitCodeBuilder): IndexedSeq[Value[Long]] = (0 until st.nDims).map(_ => const(0L))
+  def shapes: IndexedSeq[SizeValue] = (0 until st.nDims).map(_ => SizeValueStatic(0L))
 
-  def strides(cb: EmitCodeBuilder): IndexedSeq[Value[Long]] = (0 until st.nDims).map(_ => const(0L))
+  def strides: IndexedSeq[Value[Long]] = (0 until st.nDims).map(_ => const(0L))
 
   override def outOfBounds(indices: IndexedSeq[Value[Long]], cb: EmitCodeBuilder): Code[Boolean] = const(false)
 
   override def assertInBounds(indices: IndexedSeq[Value[Long]], cb: EmitCodeBuilder, errorId: Int = -1): Unit = {}
 
-  override def sameShape(other: SNDArrayValue, cb: EmitCodeBuilder): Code[Boolean] = const(false)
+  override def sameShape(cb: EmitCodeBuilder, other: SNDArrayValue): Code[Boolean] = const(false)
 
-  def firstDataAddress(cb: EmitCodeBuilder): Value[Long] = const(0L)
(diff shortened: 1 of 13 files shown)


hail/src/main/scala/is/hail/types/physical/stypes/concrete/SUnreachable.scala
```diff
@@ -251,17 +251,19 @@ class SUnreachableNDArrayValue(val st: SUnreachableNDArray) extends SUnreachable
 
   def loadElementAddress(indices: IndexedSeq[is.hail.asm4s.Value[Long]],cb: is.hail.expr.ir.EmitCodeBuilder): is.hail.asm4s.Code[Long] = const(0L)
 
-  def shapes(cb: EmitCodeBuilder): IndexedSeq[Value[Long]] = (0 until st.nDims).map(_ => const(0L))
+  def shapes: IndexedSeq[SizeValue] = (0 until st.nDims).map(_ => SizeValueStatic(0L))
 
-  def strides(cb: EmitCodeBuilder): IndexedSeq[Value[Long]] = (0 until st.nDims).map(_ => const(0L))
+  def strides: IndexedSeq[Value[Long]] = (0 until st.nDims).map(_ => const(0L))
```
patrick-schultz, line 256: I'm reconsidering whether removing the CodeBuilders is a good thing to do. I think it made something easier, but I can't remember anymore. I'm going to try to revert that change.

patrick-schultz: Actually, revisiting this I think I prefer leaving the CodeBuilders off. The only benefit to putting them back is to make it easier to change the representation of SNDArrays to store strides in #elements, but I don't think that would be a difficult refactoring regardless. And leaving them off makes it easier to use strides without worrying about storing in locals.

Happy to put them back if others prefer.

patrick-schultz:
Rebased on #10791, because I'd like that to go in first. Otherwise should be ready for review.

PR #10785, opened 2021-08-17
[terraform] Use a module for gsa and secret creation

daniel-goldstein:
This is an attempt to modularize / refactor our terraform code regarding google service accounts and kubernetes secrets. This doesn't add any new functionality.

Currently, our use of terraform is one flat file `main.tf` where we declare every `resource` in GCP that should exist. Examples of such resources are `google_service_account`, `google_service_account_key` and `kubernetes_secret`. For each of the accounts we create for various services, we end up creating these three resources (and sometimes IAM roles) in the same way. To abstract this, we can create a custom `module`, which is just a collection of resources, a set of inputs called `variables`, and a set of outputs. A module can then be "instantiated" using a `module` block in `main.tf`, providing it the source path of the module and values for its declared variables.

Tested by hand in my own GCP project.

### DIFF
--- a/infra/main.tf
+++ b/infra/main.tf
@@ -446,7 +446,7 @@ module "auth_gsa_secret" {
   iam_roles = [
     "iam.serviceAccountAdmin",
     "iam.serviceAccountKeyAdmin",
-    "storage.admin",
+    # "storage.admin",
   ]
 }
 
--- a/infra/main.tf
+++ b/infra/main.tf
@@ -352,7 +352,7 @@ resource "google_artifact_registry_repository_iam_member" "artifact_registry_ci_
   repository = google_artifact_registry_repository.repository.name
   location = var.gcp_location
   role = "roles/artifactregistry.reader"
-  member = "serviceAccount:${google_service_account.ci.email}"
+  member = "serviceAccount:${module.ci_gsa_secret.email}"
 }
 
 resource "google_storage_bucket_iam_member" "gcr_push_admin" {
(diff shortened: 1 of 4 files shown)


infra/main.tf
```diff
@@ -435,293 +435,102 @@ resource "kubernetes_service" "ukbb_rb_static" {
...
+    "compute.instanceAdmin.v1",
+    "iam.serviceAccountUser",
+    "logging.logWriter",
+    "storage.objectCreator",
+    "storage.objectViewer",
+  ])
 
-resource "google_project_iam_member" "batch_agent_object_viewer" {
-  role = "roles/storage.objectViewer"
+  role = "roles/${each.key}"
```
jigold, line 533: Just double checking this shouldn't be `for_each` here.

infra/main.tf
```diff
@@ -435,293 +435,102 @@ resource "kubernetes_service" "ukbb_rb_static" {
...
-
-  data = {
-    "key.json" = base64decode(google_service_account_key.ci_key.private_key)
-  }
+  member = "serviceAccount:${module.ci_gsa_secret.email}"
 }
 
-resource "google_service_account" "monitoring" {
-  account_id = "monitoring"
+module "monitoring_gsa_secret" {
```
jigold, line 499: FYI, this secret in hail-vdc has access to BigQuery in broad-ctsa.

jigold (changes requested):
Couple of questions:
- How does Terraform know to import the module `gsa_k8s_secret`? I assume it just scans the directory.
- What does `${module.ci_gsa_secret....}` do? Does module refer to that terraform file?

infra/main.tf
daniel-goldstein: Ya, this is a weird quirk of the terraform language (HCL). The grammar is very strict, basically just blocks with `key = val` pairs, which doesn't exactly lend itself to writing for loops. The way they get around that is by saying that if you assign the reserved keyword `for_each` to a set or map in a resource block, it will then create a resource for each element in the assigned set, and assign that value to the reserved binding `each`.

infra/main.tf
daniel-goldstein: Ok, I'll add this to the list of things to address in terraform. I'm not sure how best to codify that right now.

daniel-goldstein:
Good questions:

> How does Terraform know to import the module gsa_k8s_secret? I assume it just scans the directory.

In short, yes. Every module is defined by a `main.tf`, `variables.tf`, and `outputs.tf`, and it finds it through the `source` path in the module blocks. When you run `terraform init` in the `infra` directory, it scans `main.tf`, sees that there are module references to the local filesystem, and sets up a watch on that directory (you can still make changes to `gsa_k8s_secret` without re-running `init`).

> What does ${module.ci_gsa_secret....} do? Does module refer to that terraform file?

The way references work in terraform is a little bizarre. There are basically three classes of value that we use right now in terraform: resources, modules and variables. You declare instances of them like so (note that I'm using class and instance colloquially):

```
resource <resource_class_name> <resource_instance_name> {
  …
}

# I like to think of `module`s as being essentially unnamed collections of resources
module <module_instance_name> {
source = '/path/to/module'
…
}

variable <variable_name> {
…
}
```

and then reference them like so:

```
<resource_class_name>.<resource_instance_name>
module.<module_instance_name>
var.<variable_name>
```

So module.ci_gsa_secret.... is how you would access `output`s of the module instance called `ci_gsa_secret`.

infra/main.tf
```diff
@@ -435,293 +435,102 @@ resource "kubernetes_service" "ukbb_rb_static" {
...
 }
 
-resource "google_service_account_key" "monitoring_key" {
-  service_account_id = google_service_account.monitoring.name
+module "test_gsa_secret" {
+  source = "./gsa_k8s_secret"
+  name = "test"
+  iam_roles = [
+    "compute.instanceAdmin.v1",
+    "iam.serviceAccountUser",
```
jigold, line 509: Can you remind me why the test user needs access to all of these roles? Is this the service account that's being used for the non-default batch driver instead of batch2-agent?

infra/main.tf
```diff
@@ -435,293 +435,102 @@ resource "kubernetes_service" "ukbb_rb_static" {
...
-resource "google_project_iam_member" "batch_logging_viewer" {
-  role = "roles/logging.viewer"
-  member = "serviceAccount:${google_service_account.batch.email}"
+module "auth_gsa_secret" {
+  source = "./gsa_k8s_secret"
+  name = "auth"
+  iam_roles = [
+    "iam.serviceAccountAdmin",
+    "iam.serviceAccountKeyAdmin",
+    "storage.admin",
```
jigold, line 449: We might be able to get rid of the `storage.admin` role. We shouldn't be creating buckets for users anymore.

infra/main.tf
```diff
@@ -352,7 +352,7 @@ resource "google_artifact_registry_repository_iam_member" "artifact_registry_ci_
   repository = google_artifact_registry_repository.repository.name
   location = var.gcp_location
   role = "roles/artifactregistry.reader"
-  member = "serviceAccount:${google_service_account.ci.email}"
+  member = "serviceAccount:${module.ci_gsa_secret.email}"
```
jigold, line 355: So terraform is able to use a resource that is defined later on in the file...

jigold (dismissed):

infra/main.tf
daniel-goldstein: ah yes.

infra/main.tf
daniel-goldstein: I actually don't know the answer to this, I was just translating from Cotton's terraform. It could be that we can remove some of these.

infra/main.tf
daniel-goldstein: I'm not to the point where I can run the full test suite in another project and validate this, but I can leave it commented out and remove it in a follow-up if there's no problem? That way if we do use it somewhere I have a clue as to what might be going wrong.

jigold (approved):
This all looks good to me, but it's hard to follow the diff to make sure all the roles are copied correctly. I think when you get this all working, we should do an audit of the roles.

PR #10786, opened 2021-08-17
[address] Remove address service

daniel-goldstein:
This is unused after the shuffler was deleted.


danking (approved):
Not sure why it failed that test.

PR #10788, opened 2021-08-17
[compiler] Rewrite lowered partition alignment with takeWhile/dropWhile 

tpoterba:
Stacked on #10773

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/functions/IntervalFunctions.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/functions/IntervalFunctions.scala
@@ -137,41 +137,114 @@ object IntervalFunctions extends RegistryFunctions {
 
 
     val endpointT = TTuple(tv("T"), TInt32)
-    registerIR2("partitionIntervalContains",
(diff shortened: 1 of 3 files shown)


tpoterba:
Reassigning to John because I've assigned a lot to Patrick lately

hail/src/main/scala/is/hail/expr/ir/functions/IntervalFunctions.scala
```diff
@@ -137,41 +137,114 @@ object IntervalFunctions extends RegistryFunctions {
...
+        val isContained = cb.newLocal[Boolean]("partitionInterval_b")
+        cb.ifx(c.ceq(0),
+          cb.assign(isContained, _containsEnd.asBoolean.boolCode(cb)),
+          cb.assign(isContained, c > 0))
+
+        primitive(isContained)
+    }
+
+
+    registerSCode2("partitionIntervalContains",
```
tpoterba, line 195: This has the same semantics as the previous IR function, but I rewrote it as an SCode function for clarity and method size considerations.

johnc1231 (approved):

PR #10790, opened 2021-08-18
[query] Fix int overflow in generic lines and add error for size limit

tpoterba:


hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
```diff
@@ -160,7 +160,15 @@ object GenericLines {
             // move scanned input from buf to lineData
             val n = bufPos - begin
             if (linePos + n > lineData.length) {
-              val newLineData = new Array[Byte]((linePos + n) * 2)
+              val copySize = linePos.toLong + n
+              if (copySize > Int.MaxValue)
```
arunbhat, line 164: Should this be copySize >= Int.MaxValue. 

scala> val newLineData = new Array[Byte](Int.MaxValue)
java.lang.OutOfMemoryError: Requested array size exceeds VM limit

hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
```diff
@@ -160,7 +160,15 @@ object GenericLines {
...
+                fatal(s"GenericLines: line size reached: cannot read a line with more than 2^31-1 bytes")
+              val newSize = Math.min(copySize * 2, Int.MaxValue).toInt
```
arunbhat, line 166: Similarly should this be (Int.MaxValue-1)?

hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
tpoterba: Hmm. This is an implementation limit, not a language limit:
https://www.baeldung.com/java-arrays-max-size

The link above suggests Int.MaxValue - 8

tpoterba: Pushed up a fix.

tpoterba: thanks for the comment!

chrisvittal (approved):

PR #10791, opened 2021-08-18
[query] defaultValue is a value

patrick-schultz:
Makes `SCode.defaultValue` return an `SValue`. To support that, I added `SCode.fromValues`, which required adding concrete `SValue` subclasses for every SType. I also did some tidying up, adding `final` and `override` keywords more consistently.

### DIFF
--- a/hail/src/main/scala/is/hail/asm4s/Code.scala
+++ b/hail/src/main/scala/is/hail/asm4s/Code.scala
@@ -1240,7 +1240,7 @@ object FieldRef {
 }
 
 object Value {
-  def fromLIR[T](v: lir.ValueX): Value[T] = new Value[T] {
+  def fromLIR[T](v: => lir.ValueX): Value[T] = new Value[T] {
     def get: Code[T] = Code(v)
   }
 }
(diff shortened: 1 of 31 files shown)


hail/src/main/scala/is/hail/asm4s/Code.scala
```diff
@@ -1239,6 +1239,12 @@ object FieldRef {
   }
 }
 
+object Value {
+  def fromLIR[T](v: lir.ValueX): Value[T] = new Value[T] {
```
tpoterba, line 1243: I think this isn't valid -- the node `v` will appear in the LIR tree by object identity each time you `get` it, which is a violation of our data model.

tpoterba: (am guessing this is what's causing some of your test failures)

tpoterba:
Looks great, aside from Value.fromLIR stuff which I think is causing test failures.

hail/src/main/scala/is/hail/asm4s/Code.scala
patrick-schultz: Ah, yup. Thanks!

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
```diff
@@ -41,27 +41,26 @@ case class SNDArrayPointer(pType: PCanonicalNDArray) extends SNDArray {
...
-  def fromCodes(codes: IndexedSeq[Code[_]]): SNDArrayPointerCode = {
+  override def fromCodes(codes: IndexedSeq[Code[_]]): SNDArrayPointerCode = {
     val IndexedSeq(a: Code[Long@unchecked]) = codes
     assert(a.ti == LongInfo)
     new SNDArrayPointerCode(this, a)
   }
 
-  def storageType(): PType = pType
+  override def fromValues(values: IndexedSeq[Value[_]]): SNDArrayPointerValue = {
+    val a = values(0).asInstanceOf[Settable[Long@unchecked]]
```
tpoterba, line 51: should be Value, not Settable

tpoterba (approved):

PR #10792, opened 2021-08-18
Import table

ammekk:
Import_table now done in python and lowered by using import_lines.


hail/src/main/scala/is/hail/expr/ir/functions/StringFunctions.scala
```diff
@@ -82,10 +85,83 @@ object StringFunctions extends RegistryFunctions {
...
+          sb.clear()
+        }
+      } else {
+        sb += c
+        i += 1
+      }
+    }
+    ab += testIfMissing(sb.result())
+    val a = ab.result()
+    println("reached end")
```
tpoterba, line 158: stray println

hail/src/main/scala/is/hail/expr/ir/functions/StringFunctions.scala
```diff
@@ -82,10 +85,83 @@ object StringFunctions extends RegistryFunctions {
...
+    val quoteC = if (quote == null) '\u0000' else quote.charAt(0)
+
+    val ab = new BoxedArrayBuilder[String]()
+    val sb = new StringBuilder()
+
+    val testIfMissing = (field_entry: String) => {
+      if (!missing.contains(field_entry)) field_entry
+      else null
+    }
+    val matchSep: Int => Int = separator.length match {
```
tpoterba, line 101: I think this is adding a lot of overhead. For each line you'll need to allocate this anonymous class, and the matchSep becomes a virtual function call on that class. 

However, I see this doesn't differ between implementations. we should fix this in the future.

danking:

danking:
I merged and fixed some compilation issues. Tim, feel free to re-assign.

danking:
I think the main issue you're facing is the lack of a path canonicalization function. I wrote a quick and dirty one up and pushed to a separate PR here: https://github.com/hail-is/hail/pull/11647. Feel free to steal those changes. It felt like a substantial enough modification that I didn't want to push directly to this branch.

danking:
Also those service backend failures are :100: because we need #11624  to merge. Feel free to @fails them and I'll deal with the conflict.

danking:
@johnc1231 passing

johnc1231 (approved):

PR #10793, opened 2021-08-19
[compiler] Add Apache avro file reading

chrisvittal:
The support is very limited at this time, only basic data types,
["null", type] unions, and record (struct) types.

The intent is that for every field in the record, generate the
appropriate code for a stack struct.


hail/src/main/scala/is/hail/io/avro/AvroPartitionReader.scala
```diff
@@ -0,0 +1,102 @@
...
+          cb.assign(record, Code.newInstance[GenericData.Record, Schema](codeSchema))
+          val is = mb.open(path.asString.loadString(), false)
+          val datumReader = Code.newInstance[GenericDatumReader[GenericRecord], Schema](codeSchema)
+          val dataFileStream = Code.newInstance[DataFileStream[GenericRecord], InputStream, GenericDatumReader[GenericRecord]](is, datumReader)
+
+          cb.assign(it, dataFileStream)
+        }
+
+        val elementRegion: Settable[Region] = region
+        val requiresMemoryManagementPerElement: Boolean = false
```
chrisvittal, line 49: Is this correct? We're not putting elements in regions here, and I don't know what that means for this parameter.

tpoterba: should be `true` probably -- "requiresMemoryManagementPerElement" really means "this stream produces a lot of elements and consumers should clear between each element". 

Another way of saying this is that streams that represent *partitions* should make sure this flag is true, and streams that represent work within a single record don't need to.

hail/src/main/scala/is/hail/io/avro/AvroPartitionReader.scala
```diff
@@ -0,0 +1,102 @@
...
+        val LproduceElement: CodeLabel = mb.defineAndImplementLabel { cb =>
+          cb.ifx(!it.invoke[Boolean]("hasNext"), cb.goto(LendOfStream))
+          cb.assign(record, it.invoke[GenericRecord, GenericRecord]("next", record))
+          cb.goto(LproduceElementDone)
+        }
+
+        val element: EmitCode = ??? // need to read all the record fields (recursively) and return a stack struct
```
tpoterba, line 56: I think you can declare a stack struct settable field above, and assign to that in LproduceElement. Then `element` is just `EmitCode.present(mb, field.get)`

chrisvittal:
I'm not sure what the class loading issues are, there's probably some dependency conflict that's been introduced by pulling in avro.

Need to investigate, and possibly relocate avro.

chrisvittal:
okay, this passes locally, is there anything I need to do with the build images if I change the dependencies?

chrisvittal:
This is complete, but I need to know how to test it, I can't find explicit partition reader unit tests anywhere

tpoterba:
You can test partition readers using the ReadPartition node, that's probably the easiest way. Make an IR that's ToArray(ReadPartition) and assert it matches.

tpoterba (approved):

PR #10794, opened 2021-08-19
[compiler] Fix requiredness bug in toInt32OrMissing

tpoterba:


hail/src/main/scala/is/hail/expr/ir/functions/UtilFunctions.scala
```diff
@@ -198,7 +198,7 @@ object UtilFunctions extends RegistryFunctions {
           val s = x.loadString()
           primitive(rt.virtualType, Code.invokeScalaObject1(thisClass, s"parse$name", s)(ctString, ct))
       }
-      registerIEmitCode1(s"to${name}OrMissing", TString, t, (_: Type, xPT: EmitType) => EmitType(rpt, xPT.required)) {
+      registerIEmitCode1(s"to${name}OrMissing", TString, t, (_: Type, _: EmitType) => EmitType(rpt, false)) {
```
tpoterba, line 201: The previous implementation incorrectly assigned the requiredness of the returned value to be the same as the requiredness of a value coming in. This is invalid because it's possible to generate a missing value from a present input (if the input is not a valid string representation of a number).

ammekk (approved):

PR #10795, opened 2021-08-19
[aiotools] Azure AsyncFS

jigold:
This PR adds the Azure AsyncFS. I added a new secret with the Azure credentials for a new service account, adopted the container `hail-test-****` as the test container, and made the container have a 1 day retention policy (we'll want to check and make sure this actually works).

The performance seemed decent. We'll probably want to benchmark this formally at some point, but I think it's good enough for now to have something that works.


daniel-goldstein (dismissed):
I might give this a second full pass but on the first read it's quite good! Surprisingly straight forward, just a couple nits and questions.

daniel-goldstein (approved):
Sorry for the super late review on this. This is great, and has me thinking about how to deal with storage and many clouds.

PR #10796, opened 2021-08-19
 [batch] Teach jobs to talk to the batch front end internally

daniel-goldstein:
Since  #10648, public jobs use external DNS and are prohibited from accessing the metadata server to resolve names such as `batch.hail` or `internal.hail`. This broke the ability to submit batches from within a job, since the batch client, recognizing it is in GCE, would attempt to use the fore-mentioned domains. This opens that communication channel by adding an `/etc/hosts` entry for the appropriate batch domain that points to the internal gateway.

Relatedly, we currently have this iptables rule https://github.com/hail-is/hail/blob/81f4b1fbedfac288c717ae65664c3cd82b25ac2f/batch/batch/worker/worker.py#L210

for every job's network namespace that `ACCEPT`s packets leaving the worker. This includes packets destined for the internet but can also include those going to internal gateway, or other GCE nodes in our internal network (so maybe `internet_interface` is a slight misnomer). This means that public jobs can currently send requests to any ip address on our internal network (they cannot use our internal DNS, however). I removed this line and instead added more explicit rules (note the order is important):
- Allow packets to the internal gateway
- Allow packets looping back to the same worker the job is on
- Allow packets from public jobs leaving the worker and NOT destined for our internal network
- Allow packets NOT from public jobs leaving the worker destined anywhere

Since the default policy is to drop packets, this should forbid public jobs from talking to **other** addresses on the network with the exception of internal gateway.


batch/test/test_batch.py
```diff
@@ -674,6 +674,27 @@ def test_verify_no_access_to_metadata_server(client):
     assert "Connection timed out" in j.log()['main'], (str(j.log()['main']), status)
 
 
+def test_submit_batch_in_job(client):
```
jigold, line 677: Is it possible to try and write a test that tries to submit a batch to the wrong batch instance and ensure it fails? Maybe build a separate client with a DeployConfig as an argument with the wrong setting.

batch/batch/driver/create_instance.py
```diff
@@ -178,6 +171,27 @@ def create_instance_config(
...
+iptables --append FORWARD --source 10.1.0.0/16 --destination 169.254.169.254 --jump DROP
+# But allow the internal gateway
+iptables --append FORWARD --destination $INTERNAL_GATEWAY_IP --jump ACCEPT
+# And this worker
+iptables --append FORWARD --destination $IP_ADDRESS --jump ACCEPT
+# For public jobs, forbid outgoing requests to cluster-internal IP addresses. For not-public jobs, allow
+# all outgoing packets
+ENS_DEVICE=$(ip link list | grep ens | awk -F": " '{{ print $2 }}')
+iptables --append FORWARD --source 10.1.0.0/16 --out-interface $ENS_DEVICE ! --destination 10.128.0.0/16 --jump ACCEPT
+iptables --append FORWARD ! --source 10.1.0.0/16 --out-interface $ENS_DEVICE --jump ACCEPT
```
jigold, line 192: Let me make sure I understand. This says forward any request not from the public network to any device in our network. I'm not sure what ENS_DEVICE means, but I assume it translate to our cluster.

Likewise, `! --denstination 10.128.0.0/16` is preventing any packet not destined to our internal IP. 

Why are both $ENS_DEVICE and `10.128.0.0/16` needed?

batch/test/test_batch.py
```diff
@@ -674,6 +674,27 @@ def test_verify_no_access_to_metadata_server(client):
...
+    builder = client.create_batch()
+    bucket_name = get_user_config().get('batch', 'bucket')
+    script = f'''import hailtop.batch as hb
```
jigold, line 680: Also, can we bump the instance version? This test is going to fail on old versions of the workers. Not sure it's worth trying to prevent the deploy failure.

jigold (changes requested):
@danking It would be great if you could double check this. The IP rules seem fine to me, but I don't have much experience with them. 

Also, did we make sure the deploy config that's mounted goes to the correct place (ie. batch.hail or internal.hail)?

batch/test/test_batch.py
daniel-goldstein: Do you mean to try to submit to a different namespace or submit through the internet instead of internal gateway? If the former, what namespace should we use when the tests run in default?

batch/batch/driver/create_instance.py
daniel-goldstein: `ENS_DEVICE` is the logical equivalent of the ethernet port that connects the worker machine to the network. The reason I'm specifying this device is to distinguish between traffic that is leaving the worker and traffic inside the worker, e.g. if a job was trying to contact another colocated job.

The first iptables rule  *permits* public traffic leaving the worker (going through `ENS_DEVICE`) so long as it is *not* destined for another node on our network (`10.128.0.0./16`). Without the `destination`, we would be permitting any public outgoing traffic, including to other internal IPs. Without the `out-interface`, we would be permitting any traffic not going to other internal IPs, which means that a job could talk to another job on the same worker since jobs are given `10.0.x.x` or `10.1.x.x` IPs.

daniel-goldstein: Correct about the second iptables rule. *Not* from the public network essentially just means private network + host.

batch/test/test_batch.py
jigold: I meant let's figure out what points of attack we are most worried about (eg. the ip tables rules get refactored and don't do what we want anymore).

We could try submitting to a different namespace. For example, if the job is running in a PR or dev namespace, we try submitting a batch to the default batch instance. This should fail. If the job is running in the default namespace, then we expect it should succeed or we just skip the test.

I don't like having namespaces be interdependent though. Another idea is to not expect any response at all for the packets you've blocked. Maybe add a timeout to the request which uses a simple ClientSession to make the request? We expect a timeout error. If we get a different kind of error 502, 504, or success then we should fail the test. I'm not sure how to test contacting an external VM. Maybe the behavior is the same?

I can try and think about this more later if these proposals aren't good enough.

daniel-goldstein: Ah I see. In this case to test the iptables rules we would have to try to contact a different internal ip that is not internal gateway, but I don't know of one that is guaranteed to exist. We could reserve an IP address in that subnet purely for testing purposes..

For the submitting to a different namespace, I suppose I can add the test that expects a different result whether it is running in a PR namespace vs default.

jigold (dismissed):
@danking 

Can you also take a look before this goes in?

batch/test/test_batch.py
```diff
@@ -674,6 +677,103 @@ def test_verify_no_access_to_metadata_server(client):
...
+
+
+def test_cannot_contact_other_internal_ips(client):
+    internal_gateway_last_byte = int(INTERNAL_GATEWAY_IP.rsplit('.', maxsplit=1)[1])
+    assert internal_gateway_last_byte <= 253
+    other_ip = f'10.128.0.{internal_gateway_last_byte + 1}'
+    other_ip_2 = f'10.128.0.{internal_gateway_last_byte + 2}'
+    builder = client.create_batch()
+    script = f'''
+if [ -z ${{HAIL_BATCH_WORKER_IP+x}} ]; then
```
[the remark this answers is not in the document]
daniel-goldstein: It could be that the current worker is neither of the given IPs, which is also fine, I just don't want to accidentally collide. I My checking to make sure it is set is a little defensive, since using `port=5000` should guarantee that it is. If someone were to remove `port=5000`, then that environment variable wouldn't be set and we wouldn't know the batch worker's IP, at which point we can't guarantee that the job won't accidentally try to ping its own worker.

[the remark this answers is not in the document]
daniel-goldstein: I thought about this a bit and I think that check just made the test harder to understand at a glance because of the guarantees we have around that environment variable being set. I've removed that check.

batch/test/test_batch.py
```diff
@@ -20,8 +20,8 @@
 
 DOCKER_PREFIX = os.environ.get('DOCKER_PREFIX')
 DOCKER_ROOT_IMAGE = os.environ.get('DOCKER_ROOT_IMAGE', 'gcr.io/hail-vdc/ubuntu:18.04')
-DOMAIN = os.environ['DOMAIN']
-NAMESPACE = os.environ['HAIL_DEFAULT_NAMESPACE']
+DOMAIN = os.environ.get('DOMAIN')
```
jigold, line 23: Can you explain this change to me?

jigold (changes requested):

batch/test/test_batch.py
daniel-goldstein: The `test_batch_invariants` test was failing because of a missing environment variable, which seemed odd to me until I realized that `test_batch_invariants` doesn't define *any* environment variables. Making these optional allowed for a test that didn't need them to omit them in `build.yaml`.

jigold: Is this a pytest issue where it runs all the code in that folder and then runs the specified test? If so, do you think this would work better?

```
python3 -m pytest --log-date-format="%Y-%m-%dT%H:%M:%S" --log-format="%(asctime)s %(levelname)s %(name)s %(filename)s:%(lineno)d:%(funcName)s %(message)s" --log-cli-level=INFO -s -vv --instafail /io/test/test_invariants.py
```

daniel-goldstein: Ya I think that could work. Could I do a separate PR where we put that test in its own file and then require that the environment variables for a test file be provided?

jigold (approved):

PR #10797, opened 2021-08-19
[query] more refactoring towards removing SCode

patrick-schultz:
~Stacked on #10791~

A few high level changes got mixed up in this PR, since I couldn't predict where a thread would lead once I started pulling. If you would like, I can try to disentangle them.

Here are the conceptual changes:
* add a CodeBuilder argument to `getEmitParam`, so that parameters which are pointers to region values can be loaded into `SValue`s with multiple locals
* make `EmitValue` a concrete class, consisting of an optional boolean value and an `SValue`
* add `valueTuple` to both `SValue` and `EmitValue`
* copy the `SCode` interface onto `SValue`, to make it easier to replace `SCode`s with `SValues` 
* add `loadToSValue` to `SingleCodeType`
* add `SStreamValue`. This loses the single use assertion, but that doesn't seem like it asserts much, since the stream producer can be freely accessed without memoizing the `SStreamCode`


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -158,56 +158,40 @@ case class EmitRegion(mb: EmitMethodBuilder[_], region: Value[Region]) {
...
+  lazy val emitType: EmitType = EmitType(v.st, required)
 
-    override def get(cb: EmitCodeBuilder): SCode = {
-      if (required) {
-        v
-      } else {
-        cb.ifx(missing.get, cb._fatal(s"Can't convert missing ${ st } to PValue"))
-        v
-      }
+  def valueTuple(): IndexedSeq[Value[_]] = {
```
tpoterba, line 172: It looks like if the `missing` is `Some(const(false))` we include the bool in the value tuple. I think that's wrong -- should match on required instead of `missing` here?

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -158,56 +158,40 @@ case class EmitRegion(mb: EmitMethodBuilder[_], region: Value[Region]) {
...
-    used = true
+    val ec = EmitCode(Code._empty,
+      if (required) const(false) else missing.get,
+      v)
+    assert(ec.required == required)
     ec
   }
 
-  override def get(cb: EmitCodeBuilder): SCode = throw new UnsupportedOperationException(s"Can't make PValue for unrealizable type ${ ec.st }")
+  def get(cb: EmitCodeBuilder): SCode = {
```
tpoterba, line 187: get could return a value instead of a code, I think? I know that could happen in a future PR.

tpoterba (changes requested):
great work. One comment about emitvalue requiredness

hail/src/main/scala/is/hail/expr/ir/Emit.scala
patrick-schultz: Thanks, good catch. I restructured things to make that kind of mistake harder to make.

hail/src/main/scala/is/hail/expr/ir/Emit.scala
patrick-schultz: Turns out `EmitValue.get` only has one use!

tpoterba (dismissed):

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
```diff
@@ -953,44 +940,16 @@ class EmitMethodBuilder[C](
...
-              override def get: SCode = st.fromCodes(ts.zipWithIndex.map { case (t, i) =>
-                mb.getArg(codeIndex + i)(t).get
-              })
-
-              override def st: SType = evSelf.st
-            }
-          }
-        }
+        val m = if (et.required) None else Some(mb.getArg[Boolean](codeIndex + ts.length))
+        val v = et.st.fromValues(ts.zipWithIndex.map { case (t, i) =>
```
patrick-schultz, line 949: I'm getting an error here with a `SIndexablePointer`. Should that have been a `SingleCodeEmitParamType`?

patrick-schultz:
@chrisvittal Would you mind reviewing at least the changes since @tpoterba's approval, since he's out this week? Summary of the original changes is at the top. The main changes since that, which I made while debugging failing tests, are:
* Add `PType.loadCheapSCodeField`. Both this and `loadCheapSCode` now return an `SValue`, with the former using fields instead of locals.
* `SingleCodeSCode.loadToSValue` uses fields

chrisvittal (approved):

tpoterba:
It's probably worth running benchmarks to make sure this didn't cause anything important to get deoptimized

patrick-schultz:
I'll run benchmarks this morning. Do we have benchmarks from an earlier commit saved somewhere, or do I need to run two?

patrick-schultz:
Never mind, found them.

PR #10798, opened 2021-08-19
[query]Table tail lower

ammekk:
Adds optimization available in lowering process if number of rows per partition from child TableIR is known


hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -748,21 +748,35 @@ object LowerTableIR {
...
+              case Some(partCounts) =>
+                println(s"partCounts: ${partCounts}")
+                println(s"target number of rows ${targetNumRows}")
+                var idx = partCounts.length - 1
+                var sumSoFar = 0L
+                while(idx >= 0 && sumSoFar < targetNumRows) {
+                  sumSoFar += partCounts(idx)
+                  idx -= 1
+                }
+                MakeArray.apply((partCounts.length - 1 until idx by -1).map{partIdx => I32(partCounts(partIdx).toInt)}, TArray(TInt32))
```
johnc1231, line 761: I was wrong to do `MakeArray` here I think. We should just use `Literal`. If you can change this use and the use in `TableHead` to use `Literal` that would be great.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -748,21 +748,35 @@ object LowerTableIR {
...
-                  .mapCollect(relationalLetsAbove)(StreamLen)
-              ) { counts =>
-                If((Cast(streamSumIR(ToStream(counts)), TInt64) >= targetNumRows) || (totalNumPartitions <= ArrayLen(counts)),
-                  counts,
-                  Recur(partitionSizeArrayFunc, FastIndexedSeq(howManyPartsToTry * 4), TArray(TInt32)))
-              })
+            child.partitionCounts match {
+              case Some(partCounts) =>
+                println(s"partCounts: ${partCounts}")
+                println(s"target number of rows ${targetNumRows}")
```
johnc1231, line 754: Stray prints

johnc1231 (changes requested):

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
ammekk: I know you said to do ` Literal(ttype, FastIndexedSeq(..))` but map creates a IndexedSeq and did not like trying to be made into a FastIndexedSeq. Obviously I could try harder and do it if its a big deal.

johnc1231: Just call `toFastIndexedSeq` on the resulting `IndexedSeq`.

johnc1231:
One last thing, we now have a file called `CanLowerEfficiently.scala` (rebase if you don't have it). In that file, you have to update the `TableTail` case to match `TableHead`.

hail/src/main/scala/is/hail/expr/ir/lowering/LoweringPass.scala
```diff
@@ -46,8 +47,14 @@ case object LiftRelationalValuesToRelationalLets extends LoweringPass {
...
-  def transform(ctx: ExecuteContext, ir: BaseIR): BaseIR = ir match {
-    case x: IR => LiftRelationalValues(x)
+  def transform(ctx: ExecuteContext, ir: BaseIR): BaseIR = {
+    val ans = ir match {
+      case x: IR => LiftRelationalValues(x)
+    }
+
+    log.info(s"After lift relational values \n ${Pretty(ans)}")
+
+    ans
```
johnc1231, line 57: Just undo this stuff, it was just for debugging.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -818,11 +839,17 @@ object LowerTableIR {
             }
           }
 
+          val letBindNewCtx = TableStage.wrapInBindings(newCtxs, loweredChild.letBindings)
+          val newCtxSeq = CompileAndEvaluate(ctx, ToArray(letBindNewCtx)).asInstanceOf[IndexedSeq[Any]]
+          val numNewParts = newCtxSeq.length
+          val oldParts = loweredChild.partitioner.rangeBounds
+          val newIntervals = (oldParts.size - numNewParts until oldParts.size).map(newPartIdx => oldParts(newPartIdx))
```
johnc1231, line 846: I think you could have just used `slice` for this?

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -735,11 +735,19 @@ object LowerTableIR {
             }
           }
 
+          val letBindNewCtx = TableStage.wrapInBindings(newCtxs, loweredChild.letBindings)
+          val bindRelationLetsNewCtx = LowerToCDA.substLets(letBindNewCtx, relationalLetsAbove)
+          val newCtxSeq = CompileAndEvaluate(ctx, ToArray(bindRelationLetsNewCtx)).asInstanceOf[IndexedSeq[Any]]
+          val numNewParts = newCtxSeq.length
+          val oldParts = loweredChild.partitioner.rangeBounds
+          val newIntervals = (0 until numNewParts).map(newPartIdx => oldParts(newPartIdx))
```
johnc1231, line 743: I think you can just slice `oldParts`

johnc1231 (changes requested):

johnc1231 (approved):

PR #10800, opened 2021-08-19
[batch] Progress bar for jobs

CDiaz96:


jigold:
Can you fix the following pylint formatting errors?

```
batch/front_end/front_end.py:1455:1: W293 blank line contains whitespace
batch/front_end/front_end.py:1462:1: W293 blank line contains whitespace
batch/front_end/front_end.py:1463:13: W291 trailing whitespace
```

CDiaz96:
@jigold I went ahead and fixed the formmating issues. Let me know if anything else pops up but I think it should be good now.

web_common/web_common/styles/main.scss
```diff
@@ -452,3 +452,39 @@ td a.fill-td {
...
+div .bar {
+  display: inline-block;
+  margin: 0;
+}
+
+.textbox {
+  display: none;
+}
+
+.progress_bar {
```
jigold, line 466: Do we need to make these changes specific to the progress bar? Something like `.progress_bar div .bar`?

jigold (changes requested):

web_common/web_common/styles/main.scss
```diff
@@ -452,3 +452,9 @@ td a.fill-td {
   width: 100%;
   height: 100%;
 }
+//Progress bar css
+
+div .bar {
```
jigold, line 457: I'm sorry. I should have been more clear. Are you sure we need this change? If so, can we add something that corresponds to the id of the div the plot is in (`plotly-timeseries`)? I think it's this:

```css
#plotly-timeseries .bar {
    display: inline-block;
    margin: 0;
}
```

jigold (changes requested):

web_common/web_common/styles/main.scss
CDiaz96: Oh Okay. I get what you mean. I am going to try it but I think that specific id is directly correlated with Plotly and the css it has attached to it, so we might not need any of the CSS specifically. But let me run it and make sure.

jigold: That's what I thought. I don't think you need this at all.

jigold (approved):

PR #10803, opened 2021-08-21
[compiler] Thread requiredness through TableIR.execute

tpoterba:
stacked on #10802

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/TableIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/TableIR.scala
@@ -50,7 +50,12 @@ abstract sealed class TableIR extends BaseIR {
 
   val rowCountUpperBound: Option[Long]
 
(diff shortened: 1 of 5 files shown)


tpoterba:
unstacked and ready for review.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -65,6 +70,8 @@ abstract sealed class TableIR extends BaseIR {
   def pyUnpersist(): TableIR = unpersist()
 }
 
+class TableRunContext(val req: RequirednessAnalysis)
```
tpoterba, line 73: Making a wrapper class because I plan on adding more analyses to this later.

tpoterba: (and that means we don't need to change method signatures when we do)

johnc1231 (approved):

PR #10804, opened 2021-08-21
[compiler] Use lowering rule to implement TableJoin.execute

tpoterba:


johnc1231:
@tpoterba Needs to be assigned?

tpoterba:
I'm waiting until the full stack is done so I can benchmark and make sure there aren't regressions. I have one more commit to go, I think.

danking:

PR #10805, opened 2021-08-23
[batch] Add container str to deletion error messages

daniel-goldstein:

PR #10806, opened 2021-08-23
[batch] Only umount a container's overlay if it was successfully created

daniel-goldstein:
Scenario: A job is cancelled while its image is being pulled. The container overlay will never be made, so we should not try to unmount it in `delete_container`. This adds a flag to prevent that scenario.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -420,6 +420,8 @@ class Container:
         self.config_path = f'{self.container_scratch}/config'
         self.log_path = f'{self.container_scratch}/container.log'
 
+        self.overlay_mounted = False
+
         self.fs = LocalAsyncFS(worker.pool)
 
         self.container_name = f'batch-{self.job.batch_id}-job-{self.job.job_id}-{self.name}'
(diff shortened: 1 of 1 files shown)


batch/batch/worker/worker.py
```diff
@@ -844,12 +847,14 @@ async def delete_container(self):
...
-        except Exception:
-            log.exception('while unmounting overlay', exc_info=True)
+        if self.overlay_mounted:
+            try:
+                await check_shell(f'umount -l {self.container_overlay_path}/merged')
+                self.overlay_mounted = False
+            except asyncio.CancelledError:
+                raise
+            except Exception:
+                log.exception('while unmounting overlay', exc_info=True)
```
jigold, line 857: Add the container name here.

jigold (changes requested):

batch/batch/worker/worker.py
daniel-goldstein: Ah I have that in #10805. I can combine them though.

jigold (approved):

PR #10807, opened 2021-08-23
[terraform] Add test bucket to the terraform global config

daniel-goldstein:
I manually added a `hail_test_gcs_bucket` field to the k8s global config and use that value wherever we have our current test bucket hard coded. I also added the necessary terraform to make that in a new cluster, though I have not done a new terraform run in my project. Once this and a couple more refactoring PRs go in I'll be able to run ci tests in a separate cluster and validate that the terraform is working correctly.


jigold (dismissed):

jigold:
There's something wrong with the environment of one of the tests.

daniel-goldstein:
I forgot the Java tests were split in the same way as the batch tests and had forgot to add the new env variable in each split.

PR #10811, opened 2021-08-24
[batch] Use separate ComputeClient per disk

daniel-goldstein:
Sharing the worker's compute client can lead to prematurely closing the client session shared by the worker and disks. Just use compute clients instead.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1310,7 +1310,9 @@ class DockerJob(Job):
                     project=PROJECT,
                     instance_name=NAME,
                     name=f'batch-disk-{uid}',
-                    compute_client=worker.compute_client,
+                    compute_client=aiogoogle.ComputeClient(
+                        PROJECT, credentials=aiogoogle.Credentials.from_file('/worker-key.json')
+                    ),
                     size_in_gb=self.external_storage_in_gib,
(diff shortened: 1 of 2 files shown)


batch/batch/worker/worker.py
```diff
@@ -1310,7 +1310,9 @@ async def setup_io(self):
                     project=PROJECT,
                     instance_name=NAME,
                     name=f'batch-disk-{uid}',
-                    compute_client=worker.compute_client,
+                    compute_client=aiogoogle.ComputeClient(
+                        PROJECT, credentials=aiogoogle.Credentials.from_file('/worker-key.json')
+                    ),
```
danking, line 1315: The Disk assumes it doesn't own the ComputeClient. I think the most future-proof move here is to have the Disk create a ComputeClient itself inside the constructor. It knows the project and the Credentials come from an expected place. Then the Disk needs a close method. That method needs to be called when the Disk is no longer needed. I think that's `DockerJob.cleanup`

danking (dismissed):

batch/batch/worker/worker.py
daniel-goldstein: ah yes you're right. Haven't started on my coffee yet.

danking (approved):

PR #10814, opened 2021-08-25
[batch] Replace GCS with GoogleStorageAsyncFS (attempt #2)

jigold:
This is the same PR as #10331 that got reverted. #10771 should fix the issue so this PR doesn't introduce `JSONDecodeError` when reading job specs.


jigold:
@danking If you don't have time to review this (should be exact same code that was already approved), then I can assign it to @daniel-goldstein

danking:
I think it's very worthwhile for Daniel to get a chance to see this stuff. It's the future of Hail's filesystem API!

benchmark-service/benchmark/benchmark.py
```diff
@@ -53,11 +53,11 @@
     oauth_token = f.read().strip()
 
 
-def get_benchmarks(app, file_path):
+async def get_benchmarks(app, file_path):
     log.info(f'get_benchmarks file_path={file_path}')
-    gs_reader = app['gs_reader']
+    gs_reader: aiogoogle.GoogleStorageAsyncFS = app['gs_reader']
```
daniel-goldstein, line 58: Why does this need to be a Google FS? If it's a temporary requirement it might be better to treat it as non-google-specific wherever possible and assert that it is a `GoogleStorageAsyncFS` in the few places where it needs to be until those are abstracted.

hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -131,8 +132,7 @@ def __init__(self, *,
         self.directory = self.backend.remote_tmpdir + f'batch-pool-executor/{self.name}/'
         self.inputs = self.directory + 'inputs/'
         self.outputs = self.directory + 'outputs/'
-        self.gcs = GCS(blocking_pool=concurrent.futures.ThreadPoolExecutor(),
-                       project=project)
+        self.gcs_fs = aiogoogle.GoogleStorageAsyncFS(project=project)
```
daniel-goldstein, line 135: Similarly would be nice to move away from `gcs_fs` for naming.

memory/memory/client.py
```diff
@@ -37,7 +39,7 @@ async def _get_file_if_exists(self, filename):
         params = {'q': filename}
         try:
             async with await request_retry_transient_errors(
-                self._session, 'get', self.objects_url, params=params, headers=self._headers
+                    self._session, 'get', self.objects_url, params=params, headers=self._headers
```
daniel-goldstein, line 42: I think this is an unnecessary indent. Below as well.

memory/memory/memory.py
```diff
@@ -101,10 +102,10 @@ async def get_file_or_none(app, username, fs, filepath):
     return None
 
 
-async def load_file(redis, files, file_key, fs, filepath):
+async def load_file(redis, files, file_key, fs: GoogleStorageAsyncFS, filepath):
```
daniel-goldstein, line 105: Same as before, can this be an `FS`?

daniel-goldstein (changes requested):
Looks great! I just had a couple questions mostly around typing and naming.

daniel-goldstein (approved):

PR #10816, opened 2021-08-25
[query] StreamDistribute

johnc1231:
Adds the `StreamDistribute` IR node, which is necessary for writing distributed sort.


patrick-schultz (dismissed):
Thanks for doing this, John. This was a tricky one.

As a high level comment: this is a sufficiently complicated method, it might be worth moving it out into its own file, and breaking the code up into smaller functions. E.g.
```scala
def constructSplitterTree(
  treeHeight: Value[Int],
  splitters: SIndexableValue
): SIndexableValue
```
But I leave that up to you.

johnc1231:
Ok, I've addressed all your comments now I believe. Thanks for being thorough.

hail/src/main/scala/is/hail/expr/ir/EmitStreamDistribute.scala
```diff
@@ -0,0 +1,303 @@
...
+import is.hail.types.physical.stypes.concrete.{SIndexablePointer, SIndexablePointerCode, SJavaString, SStackStruct, SStackStructCode}
+import is.hail.types.physical.stypes.interfaces.{SBaseStruct, SIndexableCode, SIndexableValue, SStreamCode, SStringCode, primitive}
+import is.hail.types.physical.stypes.primitives.{SBooleanCode, SInt32, SInt32Code}
+import is.hail.types.virtual.{TArray, TStruct}
+import is.hail.utils._
+import is.hail.asm4s._
+
+object EmitStreamDistribute {
+
+  def emit(emitter: Emit[_], ir: StreamDistribute, cb: EmitCodeBuilder, region: Value[Region], env: EmitEnv, typeWithReqx: IR => VirtualTypeWithReq, container: Option[AggContainer], loopEnv: Option[Env[LoopRef]]): IEmitCode = {
```
patrick-schultz, line 21: Splitting the emitter into a bunch of mutually recursive functions distributed across a bunch of files adds a lot of unnecessary complexity. I actually want to stop doing that in `EmitStream` and `EmitNDArray`. But here it's easy to avoid: just make the function in a separate file take inputs at the level of `Code/SCode`, not `IR`. E.g. in a separate file:
```scala
  def streamDistribute(
    cb: EmitCodeBuilder,
    requestedSplittersAndEndsVal: SIndexableValue,
    childStream: SStreamCode,
    pathVal: SValue,
    spec: AbstractTypedCodecSpec,
    resultType: TStruct,
    region: Value[Region]
  ): SIndexablePointerCode = {
    val mb = cb.emb
    val pivotsPType = requestedSplittersAndEndsVal.st.storageType().asInstanceOf[PCanonicalArray]
    ...
  }
```
and in Emit:
```scala
      case ir@StreamDistribute(child, pivots, path, spec) =>
        val resultType = ir.typ.asInstanceOf[TArray].elementType.asInstanceOf[TStruct]
        emitI(path).flatMap(cb) { case pathCode: SStringCode =>
          val pathVal = pathCode.memoize(cb, "stream_dist_output_path")

          emitI(pivots).flatMap(cb) { case pivotsCode: SIndexableCode =>
            emitStream(child, cb, region).map(cb) { case childStream: SStreamCode =>
              val requestedSplittersAndEndsVal = pivotsCode.memoize(cb, "stream_dist_pivots")
              EmitStreamDistribute.streamDistribute(cb, requestedSplittersAndEndsVal, childStream, pathVal, spec, resultType, region)
            }
          }
        }
```

patrick-schultz (dismissed):
Thanks for putting up with all my requests John. I just have two more small ones.

patrick-schultz (approved):
Thanks John!

PR #10820, opened 2021-08-26
Import matrix table

ammekk:
Draft lowering of import matrix table through use of import_lines


danking:

danking:
Putting you down John, but feel free to reassign.

danking:
OK, I'm pretty sure https://github.com/hail-is/hail/pull/11647 will pass. We should move reviewing to that PR. Emma will still get credit for all those commits when they merge.

PR #10828, opened 2021-08-31
[ci] Alert zulip when deploy config fails to build

daniel-goldstein:


daniel-goldstein:
@danking bump

danking (approved):
Love it.

PR #10829, opened 2021-08-31
[query] Escaping field names that don't start with a letter

johnc1231:
My understanding of this now is that we have to backtick-escape any identifiers that aren't going to be valid java/python identifiers. That means we have to change the regex that identifies those to only accept things that start with a letter.

### DIFF
--- a/hail/python/hail/utils/java.py
+++ b/hail/python/hail/utils/java.py
@@ -138,7 +138,7 @@ def jiterable_to_list(it):
         return None
 
 
-_parsable_str = re.compile(r'[\w_]+')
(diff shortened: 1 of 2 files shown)


patrick-schultz:
Starting with an underscore is safe, isn't it?

johnc1231:
Ah yes, apparently Java considers an underscore to be a "letter": https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.8

johnc1231:
Addressed

patrick-schultz (approved):

PR #10834, opened 2021-09-02
[batch] Azure boot disk image and batch-worker managed identity

jigold:
- Added a shared image gallery to terraform
- Added a managed identity `batch-worker` to terraform
- Gave `batch-worker` "acrpull" privileges for the resource group
- Added new config variables in config.mk that are specific to Azure
- Added commands to batch/Makefile to create a boot disk image
- Added an Azure-specific startup script that installs Docker and the CLI and then authenticates and pulls the base image

The disk image we create is specialized. This means it has credentials in there after publishing it. I think this is okay and I specifically used the batch-worker managed identity to login for this. I can try and double check this assumption if you think I'm not correct after reading these docs: https://docs.microsoft.com/en-us/cli/azure/vm?view=azure-cli-latest#az_vm_create

> Accept system or user assigned identities separated by spaces. Use '[system]' to refer system assigned identity, or a resource id to refer user assigned identity. Check out help for more examples.

I had to give the batch-worker managed identity in the resource group we want permissions to be an identity for a VM in the build-batch-worker-image resource group.


daniel-goldstein (dismissed):
This looks great. Just one question.

batch/Makefile
```diff
@@ -40,14 +40,60 @@ deploy: build
...
-create-worker-image:
+.PHONY: gcp-create-worker-image
+gcp-create-worker-image:
 	-gcloud -q compute --project $(PROJECT) images delete batch-worker-12
 	gcloud -q compute --project $(PROJECT) images create batch-worker-12 --source-disk=build-batch-worker-image --source-disk-zone=$(ZONE)
 	gcloud -q compute --project $(PROJECT) instances delete --zone=$(ZONE) build-batch-worker-image
+
+.PHONY: az-create-build-worker-image-instance
+az-create-build-worker-image-instance: export BATCH_WORKER_PRINCIPAL_ID := $(shell az identity show --resource-group $(RESOURCE_GROUP) --name batch-worker --query principalId --output tsv)
+az-create-build-worker-image-instance: export BATCH_WORKER_IDENTITY := $(shell az identity show --resource-group $(RESOURCE_GROUP) --name batch-worker --query id --output tsv)
```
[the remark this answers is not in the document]
jigold: I spent a long time trying to figure out how to do this properly. I'm not actually sure it's the correct way. I tried a bunch of things. The issue about putting the definition of the environment variable in the actual rule is each command is run in it's own subshell and not all together in the same shell. So I'd have to define the variable at the beginning of each command which was difficult to read. We could inline the commands instead of using environment variables, but that was also difficult to read.

https://stackoverflow.com/questions/23843106/how-to-set-child-process-environment-variable-in-makefile

@danking Can you comment on whether there's a better way to define environment variables at runtime that can then be used by multiple commands in the rule?

infra/azure/main.tf
```diff
@@ -126,6 +126,6 @@ resource "azurerm_shared_image" "batch_worker" {
   identifier {
     publisher = "Canonical"
     offer     = "UbuntuServer"
-    sku       = "18.04-LTS"
```
daniel-goldstein: It would be nice to soon upgrade everything across the board to 20.04.

infra/azure/main.tf
```diff
@@ -95,3 +95,37 @@ resource "azurerm_mysql_server" "db" {
...
+  gallery_name        = azurerm_shared_image_gallery.batch.name
+  resource_group_name = data.azurerm_resource_group.rg.name
+  location            = data.azurerm_resource_group.rg.location
+  os_type             = "Linux"
+  specialized	      = true
+
+  identifier {
+    publisher = "Canonical"
+    offer     = "UbuntuServer"
+    sku       = "20.04-LTS"
```
daniel-goldstein, line 129: It would be nice to soon upgrade everything across the board to 20.04.

daniel-goldstein:
So I definitely think this should go into a shell script, since they're just commands that don't rely on any Make features (they're all phony). Could you turn each make target that you've added into a function in a script? Then the body of the script can just be to run the functions in sequence

batch/Makefile
```diff
@@ -40,14 +40,31 @@ deploy: build
...
 	gcloud -q compute --project $(PROJECT) instances delete --zone=$(ZONE) build-batch-worker-image
+
+.PHONY: az-create-build-worker-image-instance
+az-create-build-worker-image-instance:
+	REGION=$(REGION) \
+	  RESOURCE_GROUP=$(RESOURCE_GROUP) \
+	  CONTAINER_REGISTRY_NAME=$(CONTAINER_REGISTRY_NAME) \
+	  DOCKER_ROOT_IMAGE=$(DOCKER_ROOT_IMAGE) \
+	  sh az-create-build-worker-image-instance.sh
+	# Need to SSH into the VM and run "sudo waagent -deprovision+user -force"
```
daniel-goldstein, line 62: Can you explain why this is necessary? Also could it just be done in the shell script with `run-command`?

batch/Makefile
```diff
@@ -40,14 +40,31 @@ deploy: build
...
+
+.PHONY: az-create-worker-image
+az-create-worker-image:
+	REGION=$(REGION) \
+	  RESOURCE_GROUP=$(RESOURCE_GROUP) \
+          SHARED_GALLERY_NAME=$(SHARED_GALLERY_NAME) \
```
daniel-goldstein, line 68: At this point it doesn't seem very useful to thread these variables from config.mk to here to the shell script. In the interest of cloud isolation and generally defining things close to where they are used, can you define these in the script itself?

config.mk
```diff
@@ -1,6 +1,16 @@
...
+
+SUBSCRIPTION_ID := $(shell az account list | jq -rj '.[0].id')
+REGION := eastus
+RESOURCE_GROUP := hail-dev
+CONTAINER_REGISTRY_NAME := $(RESOURCE_GROUP)
+SHARED_GALLERY_NAME := batch
+DOCKER_PREFIX := $(CONTAINER_REGISTRY_NAME).azurecr.io
+
+else
+
```
daniel-goldstein, line 11: Ya I would like if possible for this PR not to touch `config.mk`. I think it's alright for now if some of these values were hard coded into the shell script, one of my following PRs will include harmonizing all of this.

daniel-goldstein (changes requested):
Just a few more thoughts

batch/Makefile
jigold: I tried to automate it, but the command just hangs and will never finish. Works in less than a second once you login to the node.

batch/Makefile
```diff
@@ -40,14 +40,31 @@ deploy: build
...
 	gcloud -q compute --project $(PROJECT) images create batch-worker-12 --source-disk=build-batch-worker-image --source-disk-zone=$(ZONE)
 	gcloud -q compute --project $(PROJECT) instances delete --zone=$(ZONE) build-batch-worker-image
+
+.PHONY: az-create-build-worker-image-instance
+az-create-build-worker-image-instance:
+	REGION=$(REGION) \
+	  RESOURCE_GROUP=$(RESOURCE_GROUP) \
+	  CONTAINER_REGISTRY_NAME=$(CONTAINER_REGISTRY_NAME) \
+	  DOCKER_ROOT_IMAGE=$(DOCKER_ROOT_IMAGE) \
+	  sh az-create-build-worker-image-instance.sh
```
danking, line 61: Make variables declare with `export` are always exposed to subprocesses, for example:

```
export DEFINED_IN_MAKEFILE=3

foo:
	env
```
```
(base) # env -i make foo FOFOFO=3
env
MAKEOVERRIDES=${-*-command-variables-*-}
DEFINED_IN_MAKEFILE=3
MAKEFLAGS=FOFOFO=3
MFLAGS=
__CF_USER_TEXT_ENCODING=0x1F7:0:0
FOFOFO=3
MAKELEVEL=1
```
NB: `env -i cmd` runs `cmd` in an empty environment, so these 7 variables are all provided by Make

daniel-goldstein (approved):

PR #10836, opened 2021-09-03
[compiler] Use a UUID prefix instead of a suffix for partfiles

chrisvittal:
Google Cloud Storage documentation and [best practices] for object names
recommends avoiding sequential names like 'part-0nnnn'. We already use
UUIDs for randomness to avoid two tasks writing to the exact same
object, but by using the UUID as a prefix rather than a suffix we
(to a degree) uniformly distribute part file names over a range,
(hopefully) improving throughput.

[best practices]: https://cloud.google.com/storage/docs/best-practices#naming

### DIFF
--- a/hail/src/main/scala/is/hail/utils/package.scala
+++ b/hail/src/main/scala/is/hail/utils/package.scala
@@ -595,13 +595,11 @@ package object utils extends Logging
     "part-" + StringUtils.leftPad(is, numDigits, "0")
   }
 
-  def partSuffix(ctx: TaskContext): String = {
-    val rng = new java.security.SecureRandom()
-    val fileUUID = new java.util.UUID(rng.nextLong(), rng.nextLong())
-    s"${ ctx.stageId() }-${ ctx.partitionId() }-${ ctx.attemptNumber() }-$fileUUID"
-  }
+  def partSuffix(ctx: TaskContext): String =
+    s"${ ctx.stageId() }-${ ctx.partitionId() }-${ ctx.attemptNumber() }"
 
-  def partFile(d: Int, i: Int, ctx: TaskContext): String = s"${ partFile(d, i) }-${ partSuffix(ctx) }"
+  def partFile(d: Int, i: Int, ctx: TaskContext): String =
+    s"${ java.util.UUID.randomUUID() }-${ partFile(d, i) }-${ partSuffix(ctx) }"
 
   def mangle(strs: Array[String], formatter: Int => String = "_%d".format(_)): (Array[String], Array[(String, String)]) = {
     val b = new BoxedArrayBuilder[String]


hail/src/main/scala/is/hail/utils/package.scala
```diff
@@ -595,13 +595,11 @@ package object utils extends Logging
...
-    val rng = new java.security.SecureRandom()
-    val fileUUID = new java.util.UUID(rng.nextLong(), rng.nextLong())
-    s"${ ctx.stageId() }-${ ctx.partitionId() }-${ ctx.attemptNumber() }-$fileUUID"
-  }
+  def partSuffix(ctx: TaskContext): String =
+    s"${ ctx.stageId() }-${ ctx.partitionId() }-${ ctx.attemptNumber() }"
 
-  def partFile(d: Int, i: Int, ctx: TaskContext): String = s"${ partFile(d, i) }-${ partSuffix(ctx) }"
+  def partFile(d: Int, i: Int, ctx: TaskContext): String =
+    s"${ java.util.UUID.randomUUID() }-${ partFile(d, i) }-${ partSuffix(ctx) }"
```
danking, line 602: I believe this complies with Google's suggestions. I assume we keep a list of the files to read somewhere else so this is forwards and backwards compatible?

chrisvittal: We do, our metadata files keep a list of partitions. The first partition is `partFile[0]`, second is `partFile[1]`, etc.. In cases where parallel we expect the output to be consumed by tools other than hail (parallel vcf or tsv/csv export) we use spark's facilities which produce `part-0nnnn` names.

danking: Hmm. Well, that's just another reason to stop using Spark!

tpoterba:
The Google docs aren't clear about whether the hash needs to be a suffix or prefix: 

> add the hash of the sequence number as part of the object name to make it non-sequential

I'm somewhat hesitant to make this change since it means our part outputs are no longer sorted lexicographically, and this property has been very useful in the past.

tpoterba:
I see, here's the bit that does suggest the prefix:

> Use a naming convention that distributes load evenly across key ranges
> Auto-scaling of an index range can be slowed when using sequential names, such as object keys based on a sequence of numbers or timestamp. This occurs because requests are constantly shifting to a new index range, making redistributing the load harder and less effective.

> In order to maintain a high request rate, avoid using sequential names. Using completely random object names gives you the best load distribution. If you want to use sequential numbers or timestamps as part of your object names, introduce randomness to the object names by adding a hash value before the sequence number or timestamp.

danking:
I endorse this change, but I don't think I'm the right reviewer. Should we roll a compiler's team dice? Or do you want to discuss this at compilers team meeting, Tim?

tpoterba:
I'll take this change, yeah. I also think it's worth running some experiments to see if this actually makes a performance difference...

danking:
Agreed, we should understand the practical effects. I suspect this isn't relevant to Hail Query read or write at almost all scales. You'd need to issue more than 1000 write requests in a second or 5000 read requests in a second. How often are folks running clusters with 1000 cores?

I bet the limit is more apparent when doing highly concurrent operations like copying (there's essentially no compute, so I'm issuing requests as fast as I possibly can).

tpoterba (approved):
after (months, oops) of thought, this is a good change and the fact that the files are no longer lexicographically sorted by index doesn't pose any problems that don't have easy workarounds (e.g. collecting the list of files, extracting the part index, sort on that).

tpoterba (changes requested):
er, however, we need to have the same semantics in the lowered writers. Add this semantics to TableNativeWriter.lower and MatirxNativeWriter.lower

danking:

PR #10837, opened 2021-09-03
[batch] Catch bad storage requests on job submission

jigold:
This fixes the bug where someone submitted a bad storage request and we did not catch it on job submission.

### DIFF
--- a/batch/batch/inst_coll_config.py
+++ b/batch/batch/inst_coll_config.py
@@ -86,6 +86,8 @@ class PoolConfig(InstanceCollectionConfig):
 
     def convert_requests_to_resources(self, cores_mcpu, memory_bytes, storage_bytes):
         storage_gib = requested_storage_bytes_to_actual_storage_gib(storage_bytes)
+        if storage_gib is None:
+            return None
 
         cores_mcpu = adjust_cores_for_memory_request(cores_mcpu, memory_bytes, self.worker_type)
         cores_mcpu = adjust_cores_for_packability(cores_mcpu)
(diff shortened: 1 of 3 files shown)


batch/batch/inst_coll_config.py
```diff
@@ -118,6 +120,8 @@ def __init__(self, name, boot_disk_size_gb, max_instances, max_live_instances):
     def convert_requests_to_resources(self, machine_type, storage_bytes):
         # minimum storage for a GCE instance is 10Gi
         storage_gib = max(10, requested_storage_bytes_to_actual_storage_gib(storage_bytes))
```
daniel-goldstein, line 122: I think you need to catch the `None` after the call to `requested_storage_bytes_to_actual_storage_gib` before passing it to max. When I run `python3 -c 'max(10, None)'`

```python3
  File "<string>", line 1, in <module>
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

I also personally get a linter error on this line on main as it is now using pyright, saying

```
■ No overloads for "max" match the provided arguments    Argument types: (Literal[10], Unknown | int | None)
```

Maybe this would be a good thing for me to add to our lint checks?

daniel-goldstein (changes requested):

batch/batch/inst_coll_config.py
jigold: good catch.

daniel-goldstein (approved):

jigold:
I had to make a fix and would like a second review to make sure this makes sense.

daniel-goldstein (approved):
This looks correct to me.

PR #10838, opened 2021-09-03
[infra] Add tf resources for gateways and container registry

daniel-goldstein:
This adds the following terraform capabilities:
- A reserved public IP for the k8s gateway
- A container registry and pull access for the k8s cluster
- A [private link](https://azure.microsoft.com/en-us/services/private-link/) for the mysql database that makes it accessible on the k8s subnet. I haven't set up the config to use the database yet but ensured that the hostname for the database was resolvable from a pod on the cluster

I think this covers most of the azure specific resources that we need. Most of the rest of our terraform for gcp creating k8s secrets for the database config and various service accounts. I'd like to approach that in a single chunk to find out how best to abstract those into modules.

### DIFF
--- a/infra/azure/README.md
+++ b/infra/azure/README.md
@@ -35,7 +35,7 @@ and client secret credentials from another developer who has them, or reset
 them. Either to create the service principal or reset it, run the following
 
 ```
-az ad sp create-for-rbac --role="Contributor" --scopes="/subscriptions/${ARM_SUBSCRIPTION_ID}" --name="terraform-principal" > terraform_principal.json
+az ad sp create-for-rbac --role="Owner" --scopes="/subscriptions/${ARM_SUBSCRIPTION_ID}" --name="terraform-principal" > terraform_principal.json
 ```
 
 Note: Your developer Azure account must be an owner of the terraform service principal.
@@ -109,12 +109,14 @@ Next, create a `global.tfvars` file from the following template and replace in t
 
 ```
 az_resource_group_name = "<resource_group_name>"
+acr_name               = "<azure_container_registry_name>"
 ```
 
-Run `terraform apply -var-file=global.tfvars`. To sync the kubernetes config, run
+Run `terraform apply -var-file=global.tfvars`. To sync the kubernetes config and login with docker, run
 
 ```
 az aks get-credentials --name vdc --resource-group $AZ_RESOURCE_GROUP_NAME
+az acr login --name <aczure_container_registry_name>
 ```
 
 you can now use `kubectl` to communicate with AKS cluster.
--- a/infra/azure/README.md
+++ b/infra/azure/README.md
@@ -109,14 +109,16 @@ Next, create a `global.tfvars` file from the following template and replace in t
 
 ```
 az_resource_group_name = "<resource_group_name>"
+# Omit this field to have the container registry name default to the resource group name
 acr_name               = "<azure_container_registry_name>"
 ```
 
-Run `terraform apply -var-file=global.tfvars`. To sync the kubernetes config and login with docker, run
+Run `terraform apply -var-file=global.tfvars`. To sync the kubernetes config and authenticate
+docker with the container registry, run
 
 ```
 az aks get-credentials --name vdc --resource-group $AZ_RESOURCE_GROUP_NAME
-az acr login --name <aczure_container_registry_name>
+az acr login --name <azure_container_registry_name>
 ```
 
 you can now use `kubectl` to communicate with AKS cluster.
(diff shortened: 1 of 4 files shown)


infra/azure/main.tf
```diff
@@ -60,18 +62,50 @@ resource "azurerm_kubernetes_cluster" "vdc" {
...
+  name                  = "pool"
   kubernetes_cluster_id = azurerm_kubernetes_cluster.vdc.id
-  vm_size = "Standard_D2_v2"
-  vnet_subnet_id = azurerm_subnet.k8s_subnet.id
+  vm_size               = "Standard_D2_v2"
+  vnet_subnet_id        = azurerm_subnet.k8s_subnet.id
 
   enable_auto_scaling = true
 
-  node_count = 1
```
daniel-goldstein, line 70: This just shouldn't have been here in the first place, since there's already a min and max count set. I think this is what you faced when it said there were changes to be made because the node pool had naturally scaled down to 0.

infra/azure/README.md
```diff
@@ -109,12 +109,14 @@ Next, create a `global.tfvars` file from the following template and replace in t
 
 ```
 az_resource_group_name = "<resource_group_name>"
+acr_name               = "<azure_container_registry_name>"
 ```
 
-Run `terraform apply -var-file=global.tfvars`. To sync the kubernetes config, run
+Run `terraform apply -var-file=global.tfvars`. To sync the kubernetes config and login with docker, run
```
jigold, line 115: Do you mean login to the Azure Container Registry?

infra/azure/README.md
```diff
@@ -109,12 +109,14 @@ Next, create a `global.tfvars` file from the following template and replace in t
...
 
 ```
 az aks get-credentials --name vdc --resource-group $AZ_RESOURCE_GROUP_NAME
+az acr login --name <aczure_container_registry_name>
```
jigold, line 119: I think there's a typo in `<aczure_...`

infra/azure/main.tf
```diff
@@ -95,3 +129,17 @@ resource "azurerm_mysql_server" "db" {
   ssl_enforcement_enabled       = true
   public_network_access_enabled = false
 }
+
+resource "azurerm_private_endpoint" "db_endpoint" {
```
jigold, line 133: Is this what allows the database to be on the same network as the k8s cluster?

infra/azure/main.tf
```diff
@@ -60,18 +62,50 @@ resource "azurerm_kubernetes_cluster" "vdc" {
...
+}
+
+resource "azurerm_role_assignment" "vdc_to_acr" {
+  scope                = azurerm_container_registry.acr.id
+  role_definition_name = "AcrPull"
+  principal_id         = azurerm_kubernetes_cluster.vdc.kubelet_identity[0].object_id
+}
+
+resource "azurerm_role_assignment" "vdc_batch_worker_subnet" {
+  scope                = azurerm_subnet.batch_worker_subnet.id
```
jigold, line 98: Does this allow k8s to talk to the batch_worker VMs?

infra/azure/main.tf
```diff
@@ -60,18 +62,50 @@ resource "azurerm_kubernetes_cluster" "vdc" {
...
+  role_definition_name = "Network Contributor"
+  principal_id         = azurerm_kubernetes_cluster.vdc.identity[0].principal_id
+}
+
+resource "azurerm_role_assignment" "vdc_k8s_subnet" {
+  scope                = azurerm_subnet.k8s_subnet.id
+  role_definition_name = "Network Contributor"
```
jigold, line 105: I'm not sure what this block is for.

jigold (changes requested):

jigold:
Sorry I should read your description first...

infra/azure/README.md
daniel-goldstein: yes, I can clarify this.

infra/azure/main.tf
daniel-goldstein: Both this and the k8s subnet roles were necessary for me to successfully launch a k8s internal load balancer that sat between these two subnets. The k8s service would fail otherwise listing that it lacked read/write permissions on the subnet.

infra/azure/main.tf
daniel-goldstein: See above comment. It kind of surprised me that I had to do this explicitly, but I see no harm in it because it is literally a subnet just for k8s.

infra/azure/main.tf
```diff
@@ -16,6 +16,10 @@ provider "azurerm" {
   features {}
 }
 
+locals {
+  acr_name = var.acr_name == "" ? var.az_resource_group_name : var.acr_name
```
jigold, line 20: This is cool!

jigold (approved):

PR #10845, opened 2021-09-07
[gateway] Include server block for ukbb if it is listed as subdomain

daniel-goldstein:
This is a pretty simple change, though the diff is a little bizarre. I don't want running a ukbb server to be a requirement of running a hail instance. Since the ukbb app is a special case in the way we deploy apps in k8s, we would render the gateway configuration server blocks with the logic:

- For each service, if it's not ukbb, render it the usual way
- Render the ukbb block

This is a very simple change to instead make the logic

- For each service, if it's ukbb render it the ukbb way, else render it the usual way

Together with the separation of the ukbb terraform into its own module in #10842, it should be easier to deploy hail without the ukbb app and simplify the bootstrap process.

This is currently running in hail-vdc.

### DIFF
--- a/gateway/gateway.nginx.conf
+++ b/gateway/gateway.nginx.conf
@@ -73,38 +73,7 @@ server {
 }
 
 {% for service in subdomains %}
-{% if service != 'ukbb-rg' %}
-server {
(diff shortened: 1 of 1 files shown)


danking (approved):
I love that we're making this all more modular. Great work!

PR #10847, opened 2021-09-07
[query] Add a check that file exists before checking if it's a directory

### LINKED ISSUE #10843: Bad error message: MatrixTable and Table files are directories, path 'blah' is not a directory
to replicate:

```
hl.read_table('gs://hail-datasets-us/annotations/THIS_PATH_DOES_NOT_EXIST')
```

johnc1231:
Resolves #10843 

Now you'll see:

```
Hail version: 0.2.74-467a12fcbef9
Error summary: HailException: No file or directory found at gs://hail-datasets-us/annotations/THIS_PATH_DOES_NOT_EXIST
```

instead of the error about it not being a directory.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/AbstractMatrixTableSpec.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/AbstractMatrixTableSpec.scala
@@ -27,11 +27,13 @@ object RelationalSpec {
     new MatrixTypeSerializer
 
   def readMetadata(fs: FS, path: String): JValue = {
-    if (!fs.exists(path)) {
-      fatal(s"No file or directory found at ${path}")
+    if (!fs.isDir(path)) {
+      if (!fs.exists(path)) {
+        fatal(s"No file or directory found at ${path}")
+      } else {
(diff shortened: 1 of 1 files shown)


hail/src/main/scala/is/hail/expr/ir/AbstractMatrixTableSpec.scala
```diff
@@ -27,6 +27,9 @@ object RelationalSpec {
     new MatrixTypeSerializer
 
   def readMetadata(fs: FS, path: String): JValue = {
+    if (!fs.exists(path)) {
```
tpoterba, line 30: can we move this into the `!isDir` path so we don't end up with two file system calls on the fast path?

tpoterba (dismissed):

hail/src/main/scala/is/hail/expr/ir/AbstractMatrixTableSpec.scala
johnc1231: Good call.

tpoterba (approved):

PR #10849, opened 2021-09-07
Release 0.2.75

johnc1231:
@jigold @daniel-goldstein Can one of you update batch changelog? It's been a while since we released.


jigold:
I will do so.

PR #10850, opened 2021-09-07
[batch] Fix jq wiping deploy config in test

daniel-goldstein:
This test is failing because `jq` reading back to its own stream is resulting in wiping the file. This seemed the easiest workaround. I also fixed the semantics of the test. I'm not sure why I previously prepended `batch` to the domain, what I'm really testing here is that something meant to be `internal.hail.is` can't pretend to be pointing to `hail.is`.

### DIFF
--- a/batch/test/test_batch.py
+++ b/batch/test/test_batch.py
@@ -727,6 +727,10 @@ python3 -c \'{script}\'''',
     else:
         assert status['state'] == 'Failed', str(status)
 
+    if DOMAIN.startswith('internal'):
+        without_internal = DOMAIN.split('.', maxsplit=1)[1]
+    else:
+        without_internal = DOMAIN
     builder = client.create_batch()
     j = builder.create_job(
         os.environ['HAIL_HAIL_BASE_IMAGE'],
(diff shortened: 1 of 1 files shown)


batch/test/test_batch.py
```diff
@@ -727,14 +727,19 @@ def test_cant_submit_to_default_with_other_ns_creds(client):
...
+        without_internal = DOMAIN
     builder = client.create_batch()
     j = builder.create_job(
         os.environ['HAIL_HAIL_BASE_IMAGE'],
         [
             '/bin/bash',
             '-c',
             f'''
-jq '.domain = "batch.{DOMAIN}"' /deploy-config/deploy-config.json > /deploy-config/deploy-config.json
+jq '.domain = "{without_internal}"' /deploy-config/deploy-config.json > tmp.json
```
danking, line 741: Hmm. Yeah, I'm not sure I understand why we had `batch....`, this should be the domain suffix for the current Hail instance. That should always be `hail.is` except in the Aussie's setup.

batch/test/test_batch.py
```diff
@@ -727,14 +727,19 @@ def test_cant_submit_to_default_with_other_ns_creds(client):
...
+mv tmp.json /deploy-config/deploy-config.json
 python3 -c \'{script}\'''',
```
danking, line 743: I don't think this test is doing what it ought to. A deploy config in a namespace looks like this:
```
(base) # k get secrets  -n dking gce-deploy-config -o json | jq -r '.data["deploy-config.json"]' | base64 -D | jq '.'
{
  "location": "gce",
  "default_namespace": "dking",
  "domain": "hail.is"
}
```

The domain is a per-deployment-of-hail thing. The Aussies have a domain and we have a domain. I suppose Hail on Azure will soon have a domain.

If you want to create a deploy config that points at the default namespace, you can change the `default_namespace` (this name is a confusing legacy, it would be good to change this to `namespace`) to point at `default`

danking (changes requested):

daniel-goldstein:
Closing in favor of #10854

PR #10855, opened 2021-09-08
[batch] Do not try to kill a non-existent container

jigold:


batch/batch/worker/worker.py
```diff
@@ -841,7 +841,12 @@ async def delete_container(self):
                 log.info(f'{self} container is still running, killing crun process')
                 self.process.terminate()
```
daniel-goldstein, line 842: These lines still make me uneasy, because `self.process.terminate()` should be sending a SIGTERM. I am almost inclined to remove this line or put it as a later resort. So
1. Try to ask crun to stop the container with `crun kill`
2. If the above fails, try to `terminate` the crun process itself (it's unclear to me if we should do some sort of waiting to check whether the process responded to the terminate. I'm not sure what there is left to do at this point if it didn't other than log that we tried really hard and didn't succeed)
3. Finally set `process = None`

What do you think?

daniel-goldstein (changes requested):

batch/batch/worker/worker.py
```diff
@@ -841,7 +841,12 @@ async def delete_container(self):
...
                 self.process = None
-                await check_exec_output('crun', 'kill', '--all', self.container_name, 'SIGTERM')
+                try:
+                    await check_exec_output('crun', 'state', self.container_name)
+                except CalledProcessError:
+                    log.info(f'container {self} cannot be killed because it does not exist', exc_info=True)
+                else:
+                    await check_exec_output('crun', 'kill', '--all', self.container_name, 'SIGTERM')
```
daniel-goldstein, line 849: We are *still* not guaranteed that the container will exist at this point, but it is probably worth checking to see how often that arises in practice.

batch/batch/worker/worker.py
jigold: I think this seems reasonable, but I changed the signal for `crun kill` to SIGKILL as I think responding to a SIGTERM is optional. Maybe we call `crun kill` twice with two different signals.

batch/batch/worker/worker.py
```diff
@@ -839,9 +839,19 @@ async def delete_container(self):
...
             try:
                 log.info(f'{self} container is still running, killing crun process')
-                self.process.terminate()
-                self.process = None
-                await check_exec_output('crun', 'kill', '--all', self.container_name, 'SIGTERM')
+
+                try:
+                    await check_exec_output('crun', 'state', self.container_name)
+                except CalledProcessError:
+                    log.info(f'container {self} cannot be killed because it does not exist', exc_info=True)
```
daniel-goldstein, line 846: Wouldn't we want this branch to get `terminate`d too? Is there a possibility that the crun process is still setting up the container, so it doesn't exist yet, then we set `self.process = None` which leaves a zombie process?

With that in mind, it pretty much makes the except blocks here identical, and we `log.info` regardless in both cases. `crun state` erroring out is essentially telling us the same information as `crun kill` erroring out, it just perhaps feels kinder. I also think these nested try blocks are starting to get confusing.

I see this code as essentially the same, albeit with slightly less clean error messages:

```python
try:
    await check_exec_output('crun', 'state', self.container_name)
    await check_exec_output('crun', 'kill', '--all', self.container_name, 'SIGKILL')
except CalledProcessError:
    log.info(f'Could not delete container {self} because it does not exist', exc_info=True)
    self.process.terminate()
finally:
    self.process = None
```

The exc_info here could still tell us which one failed, but calling state starts to look not quite as useful as before. For the most part what we've done is change the order of `self.process.terminate`, which I think is much better now, and downgraded the error from before to info.

daniel-goldstein: I'm still unsure of what I think the best approach is here, but I do think that as long as we're positive there are no leaks, we'll hit delete-nonexistent-container errors that we will have to ignore. I'm alright with trying out something like the above, though I'm unsure how to verify that the state of the system has improved and we're not just silencing errors.

We could have a monitoring loop that ensures the output of `crun list` approximately matches in length the number of containers we expect are running.

daniel-goldstein (changes requested):
I'm not sure if my comment was a step backward or forward. I offered a counterpoint that questions whether calling state really tells us anything. If you think it would give us useful information, I'm fine with it as long as we `self.process.terminate()` in all failure modes.

batch/batch/worker/worker.py
daniel-goldstein: We might also use something like an `asyncio.wait_for` with a reasonable timeout and wait for `process` to exit. If it doesn't do so within a few seconds then we log an `exception`. This might be the easiest approach to make sure that we see something if it's really borked.

PR #10863, opened 2021-09-12
Dataproc: allow upload of python files in a zip archive

vladsaveliev:
Sometimes it's useful to send the dataproc a whole python package with python modules to import. However, listing files comma-separated in `--pyfiles` ends up dropping the folder name, which prevents importing them as modules within a package. To demonstrate:

```
hailctl dataproc submit --pyfiles mypackage/mymodule.py,mypackage/__init__.py ...
```

Translates to:

```
gcloud dataproc jobs submit pyspark myscript.py \
    --py-files=/tmp/pyscripts_4b2_1oej.zip
   ...
```

Which is extracted on the cluster into

```
> ls pyscripts_pqrklaet
__init__.py  mymodule.py
```

And `myscript.py` ends up not being able to import `mymodule.py`:

```
  File "/tmp/d6b95a71fea84fd2a32e6b523dedea2d/myscript.py", line 20, in <module>
    from mypackage import mymodule
ModuleNotFoundError: No module named 'mypackage'
```

This PR is a quick fix to that: it allows uploading `--py-files` in a zip archive, so the folder structure is preserved after extracting.


johnc1231 (approved):

vladsaveliev:
@johnc1231, fixed this lint issue!

PR #10866, opened 2021-09-13
[infra] Add necessary infra for running CI tests in a new project

daniel-goldstein:
This is a bit of a mess so feel free to ask that I break it down or explain more in depth. In short, this:

- adds terraform for forgotten bits and pieces necessary for running PR tests like test buckets and the necessary permissions on those resources
- Adds a couple of flags that allow a CI `WatchedBranch` to be considered mergeable or not. This shouldn't change anything in default CI, but it allows you to specify that a secondary CI should run PRs, post statuses, and deploy new commits, but never actually commit anything to GitHub. Similarly there's a flag for turning off zulip notifications, but annoyingly the zulip config is still a required secret. I plan to make that nicer in the future.
- Fixes a lot of previously-unreached syntax errors in the batch tests.

Following PRs will have relatively less functionality but probably a fair bit of cleanup and reorganization, e.g. getting rid of config.mk and generating it from terraform output, making scripts of the bootstrapping process etc.


daniel-goldstein:
I'm going to need to do a bit of manual work with our global config and submit a prequel PR so will mark as WIP until that goes through, but most of the context is here.

daniel-goldstein:
Stacked on #10870

infra/main.tf
```diff
@@ -449,7 +462,29 @@ module "ci_gsa_secret" {
...
+}
+
+resource "google_storage_bucket_iam_member" "gcr_ci_viewer" {
+  bucket = google_container_registry.registry.id
+  role = "roles/storage.objectViewer"
+  member = "serviceAccount:${module.ci_gsa_secret.email}"
+}
+
+resource "kubernetes_secret" "hail_ci_0_1_service_account_key" {
+  metadata {
```
jigold, line 478: I've always found the name of this confusing.

infra/main.tf
```diff
@@ -582,3 +691,31 @@ resource "google_dns_record_set" "internal_gateway" {
 
   rrdatas = [google_compute_address.internal_gateway.address]
 }
+
+resource "kubernetes_cluster_role" "batch" {
```
jigold, line 695: Can we delete batch/service-account.yaml?

infra/main.tf
```diff
@@ -473,11 +513,60 @@ module "test_gsa_secret" {
...
+  metadata {
+    name = "zulip-config"
+  }
+
+  data = {
+    ".zuliprc" = file("~/.hail/.zuliprc")
+  }
+}
+
+resource "kubernetes_secret" "auth_oauth2_client_secret" {
```
jigold, line 559: Can we rename this as `google_oauth2_client_secret`?

infra/README.md
```diff
@@ -214,14 +233,15 @@ You can now install Hail:
   export HAIL_BUILDKIT_IMAGE=$HAIL_DOCKER_PREFIX/hail-buildkit:cache
   export BATCH_WORKER_IMAGE=$HAIL_DOCKER_PREFIX/batch-worker:cache
   export HAIL_CI_BUCKET_NAME=dummy
+  export HAIL_TEST_GCS_BUCKET=<hail-test-gcs-bucket>
```
jigold, line 236: Do you need the new `HAIL_TEST_REQUESTER_PAYS_GCS_BUCKET`?

infra/README.md
```diff
@@ -77,17 +86,22 @@ Instructions:
    # If set to true, pull the base ubuntu image from Artifact Registry.
    # Otherwise, assumes GCR.
    use_artifact_registry = false
+
```
jigold, line 89: I can't comment above, but do we need the requester pays bucket up here?

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -621,8 +625,8 @@ def test_gcsfuse_empty_string_bucket_fails(self):
             j.gcsfuse(self.bucket_name, '')
 
     def test_requester_pays(self):
-        b = self.batch(requester_pays_project='hail-vdc')
-        input = b.read_input('gs://hail-services-requester-pays/hello')
+        b = self.batch(requester_pays_project=PROJECT)
+        input = b.read_input(f'gs://{HAIL_REQUESTER_PAYS_GCS_BUCKET}/hello')
```
jigold, line 629: **Don't do this now**, but the tests will need to have `HAIL_REQUESTER_PAYS_GCS_BUCKET` include the `gs://` scheme. If there's no equivalent of the requester pays in Azure, we'll need to add some sort of decorator to not run this test in Azure.

daniel-goldstein:
Going to stack this on #10869 as well. That will allow me to clean up this PR substantially.

ci/ci/github.py
```diff
@@ -596,10 +596,12 @@ def checkout_script(self):
 
 
 class WatchedBranch(Code):
-    def __init__(self, index, branch, deployable):
+    def __init__(self, index, branch, deployable, mergeable, notify_zulip):
         self.index = index
         self.branch = branch
         self.deployable = deployable
+        self.mergeable = mergeable
```
jigold, line 603: I double checked the global config has these two new parameters.

jigold (changes requested):
This looks great!

infra/README.md
```diff
@@ -38,8 +38,17 @@ Instructions:
...
+- Go to the Google Cloud console, API & Services, Credentials.
+  Configure the consent screen.  Add the scope:
+  https://www.googleapis.com/auth/userinfo.email.  Back in Credentials, create an OAuth
+  client ID.  Authorize the redirect URIs:
+
+   - https://auth.<domain>/oauth2callback
+   - http://127.0.0.1/oauth2callback
+
+  Download the client secret to $HOME/.hail/client_secret.json.
```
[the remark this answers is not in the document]
daniel-goldstein: This is intended to be downloading the secret from the console UI (it is fairly obvious when you go through the process). This secret is now managed by terraform so creating it explicitly through kubectl is no longer required.

infra/main.tf
daniel-goldstein: Me too

infra/main.tf
daniel-goldstein: This ClusterRole and the `batch/service-account.yaml` serve different purposes. The cluster role is necessary for *default* batch because it is essentially responsible for creating and managing other namespaces. The roles in `batch/service-account.yaml` are not cluster roles, but roles within a namespace, and are applied with `kubectl -f apply -n PR_OR_DEV_NAMESPACE` and that allows the batch in a particular namespace access to the secrets and service accounts within the same namespace. This ClusterRole was added by hand years ago.

infra/main.tf
daniel-goldstein: I suppose this might depend on what we hear back about azure oauth, but I imagined that oauth2 being a standard would mean we would have the same secret just filled by an azure oauth2 secret, whatever form that may take. In which case, why make the name of this secret specifically google?

infra/README.md
daniel-goldstein: I reuse the `hail_test_gcs_bucket_*` variables as idk why we would want a different configuration for the test bucket vs the test requester pays bucket.

infra/README.md
```diff
@@ -38,8 +38,17 @@ Instructions:
        artifactregistry.googleapis.com
    ```
 
-- Delete the default network if it exists. Enabling the networking
-  API creates it.
+- Check to make sure that there does not exist a default network.
```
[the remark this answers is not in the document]
daniel-goldstein: Added a comment on finding it with the cli.

infra/README.md
daniel-goldstein: I actually think some of these environment variables are unnecessary. I'll make a todo in asana to remove the ones that are no longer needed.

daniel-goldstein:
Broke this up into many PRs, the last of which is #11053

PR #10869, opened 2021-09-13
[k8s] Use global config values from k8s secret when possible

daniel-goldstein:
There are many of occurences of k8s templating of variables like GCP project or domain that never change and exist in the global config. The process of adding a field to the global config sometimes then requires adding it to `config.mk`, then the jinja of a deployment template that needs it, and then templating that in the deployment.yaml. These are nearly always environment variables (but not always), which can and sometimes are read from kubernetes secrets. This is a sweep of every such occurence I could find so that these variables are just read directly from the k8s secret.

Though it adds lines to the deployments, it reduces the complexity of our Makefile process and makes adding variables to the global config much easier. This also *dramatically* reduces the dependencies on `config.mk` and most of its variables. I think  I'll address config.mk specifically in another PR, but I believe keeping it from ballooning with multi-cloud configuration will be valuable in keeping the complexity of our build/deployment system in check.


daniel-goldstein:
Stacked on #10870

daniel-goldstein:
@jigold bump

jigold (approved):

memory/Makefile
```diff
@@ -23,5 +23,5 @@ deploy: build
 	! [ -z $(NAMESPACE) ]  # call this like: make deploy NAMESPACE=default
 	kubectl -n $(NAMESPACE) apply -f service-account.yaml
 	python3 ../ci/jinja2_render.py '{"default_ns":{"name":"$(NAMESPACE)"}}' service-account-batch-pods.yaml service-account-batch-pods.yaml.out
```
danking, line 25: Hmm. This file doesn't exist. It seems it was mistakenly deleted by https://github.com/hail-is/hail/pull/9718/files

danking: Huh. I have no idea how the memory service works in PRs since that PR also removed the service account for memory.

danking: Ahah, the file is called service-account.yaml

danking: We shouldn't do this line at all, we never use the output!

memory/Makefile
```diff
@@ -23,5 +23,5 @@ deploy: build
...
-	python3 ../ci/jinja2_render.py '{"code":{"sha":"$(shell git rev-parse --short=12 HEAD)"},"deploy":$(DEPLOY),"memory_image":{"image":"$(MEMORY_IMAGE)"},"global":{"docker_prefix":"$(DOCKER_PREFIX)","domain":"$(DOMAIN)"}}' deployment.yaml deployment.yaml.out
```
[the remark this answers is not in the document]
daniel-goldstein: Addressed these two comments in #10891

PR #10870, opened 2021-09-13
[ci] Add requester pays bucket for tests

daniel-goldstein:
I manually added a field to the global-config for the requester pays bucket used in batch tests. Adding it here to build.py's view of global fields so that CI can template #10866 in the future and actually test it.


jigold (dismissed):
Can we rename this to `HAIL_TEST_REQUESTER_PAYS_GCS_BUCKET`?

jigold (approved):

PR #10873, opened 2021-09-13
[query] add `pc_relate_bm` method (BlockMatrix implementation)

pwc2:
Adds the `BlockMatrix` implementation of `pc_relate` and associated tests. Currently, this will only run on the local backend.


danking:
@johnc1231, @patrick-schultz  I think this slipped through the reviewing cracks, would one or both of you mind adopting this PR?

danking:
bump

hail/python/hail/methods/relatedness/pc_relate_bm.py
```diff
@@ -0,0 +1,536 @@
...
+    beta = (BlockMatrix.from_ndarray(((1 / S) * V).T, block_size=block_size) @ g.T) \
+        .checkpoint(new_temp_file("pc_relate_bm/beta", "bm"))
+    mu = (0.5 * (BlockMatrix.from_ndarray(V * S, block_size=block_size) @ beta).T) \
+        .checkpoint(new_temp_file("pc_relate_bm/pre-mu", "bm"))
+
+    # Define NaN to use instead of missing, otherwise cannot go back to block matrix
+    nan = hl.float64(float("NaN"))
+
+    # Replace bad entries in g and pre_mu with NaNs
+    g = g._map_dense(lambda x: hl.if_else(_bad_gt(x), nan, x)).checkpoint(new_temp_file("pc_relate_bm/g", "bm"))
```
johnc1231, line 458: Trying to follow this. 

`g` comes from `mean_imputed_gt`, which has fields that are floats derived from mean imputing data. Do we intend to throw those mean imputed values out if they don't land exactly on 0.0, 1.0, or 2.0?

johnc1231 (changes requested):

hail/python/hail/methods/relatedness/pc_relate_bm.py
pwc2: This is a good point, I had to take a look at the paper again to refresh my memory. In the GRM definition in the beginning (equation 1) it says

> where S_ij is the subset of SNPs for which individuals i and j both have non-missing genotype data

so maybe it's more appropriate to skip the mean imputation all together. Otherwise we could potentially end up keeping mean imputed values that happen to be exactly 0, 1, or 2, as you point out.

johnc1231: Based on your quote from the paper, I think it would be correct to skip the mean imputing.

johnc1231: Probably should just introduce nans at the time where you were doing mean imputing, and then you don't need this `_bad_gt` method at all

pwc2: Cool, I'll make that change

danking:
@patrick-schultz can you adopt reviewing of this PR in light of John C's departure? Thanks!

danking:
Bump

patrick-schultz:
Sorry for dragging my feet on this. Looking now.

jigold:
@patrick-schultz bump

patrick-schultz:
Sorry for taking so long with this. There are some changes I think will be needed for this to be production ready, but at this point it's probably best to get it merged and iterate from there.

@pwc2 It looks like the tolerance might need to be relaxed in test_pc_relate. And there are some python linter errors. Once those are fixed, I'm ready to approve. Thanks again for doing the work on this, and thanks for your patience!

PR #10878, opened 2021-09-14
[query] Fix lowered coercers with empty partitions

tpoterba:


hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -243,20 +239,24 @@ object LoweredTableReader {
...
+              ArrayRef(scanResult, i),
+              FastIndexedSeq("i" -> i))
+          }) { row => ArrayLen(GetField(row, "minkey")) > 0 }
+      ) { row =>
+        InsertFields(row, FastSeq(
+          ("minkey", ArrayRef(GetField(row, "minkey"), I32(0))),
+          ("maxkey", ArrayRef(GetField(row, "maxkey"), I32(0)))))
+      }
+    }) { (l, r) =>
+      ApplyComparisonOp(LT(keyType), GetField(l, "minkey"), GetField(r, "minkey"))
```
patrick-schultz, line 257: You didn't introduce this, but shouldn't this sort on `(key, i)` lexicographically? Say the partitions are `[0, 2], [2, 2], [2, 5]`, and it sorted by min key to `[0, 2], [2, 5], [2, 2]`. Then we would decide the data was unsorted.

patrick-schultz (changes requested):
Looks good, other than one preexisting issue. Since it doesn't seem to be causing problems, I'm okay if you want to get this in and fix the other bug later.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
tpoterba: I can fix this. I think it should sort on `(minkey, maxkey)` though -- the index shouldn't matter.

I believe the adversarial case is:
```
[0, 2], [2, 5], [2, 2]
```
This will be sorted to itself, and will fail the sortedness check in the next phase.

patrick-schultz: Yup, you're right.

tpoterba: done

patrick-schultz (approved):

PR #10885, opened 2021-09-20
[query] Fold Aggregator

johnc1231:
Introduces `hl.agg.fold`, which allows writing of generic aggregations in Python

### DIFF
--- a/hail/python/hail/ir/ir.py
+++ b/hail/python/hail/ir/ir.py
@@ -2823,6 +2823,9 @@ def subst(ir, env, agg_env):
         return ApplyAggOp(ir.agg_op,
                           subst_init_op_args,
                           subst_seq_op_args)
+    elif isinstance(ir, AggFold):
+        subst_seq_op = subst(ir.seq_op, agg_env, {})
+        return AggFold(ir.zero, subst_seq_op, ir.comb_op, ir.accum_name, ir.other_accum_name, ir.is_scan)
     elif isinstance(ir, AggArrayPerElement):
         return AggArrayPerElement(_subst(ir.array, agg_env),
                                   ir.element_name,
--- a/hail/python/hail/ir/ir.py
+++ b/hail/python/hail/ir/ir.py
@@ -1887,6 +1887,13 @@ class ApplyScanOp(BaseApplyAggOp):
         return i == 1
 
 
+class AggFold(IR):
+    @typecheck_method(zero=IR, seq_op=IR, comb_op=IR, element_name=str, accum_name=str, other_state_name=str)
+    def __init__(self, zero, seq_op, comb_ob, element_name, accum_name):
+        super().__init__(zero, seq_op, comb_ob)
+        # TODO: Fill in
+
+
 class Begin(IR):
     @typecheck_method(xs=sequenceof(IR))
     def __init__(self, xs):
--- a/hail/python/hail/ir/ir.py
+++ b/hail/python/hail/ir/ir.py
@@ -1891,7 +1891,6 @@ class AggFold(IR):
     @typecheck_method(zero=IR, seq_op=IR, comb_op=IR, accum_name=str, other_accum_name=str, is_scan=bool)
     def __init__(self, zero, seq_op, comb_op, accum_name, other_accum_name, is_scan):
         super().__init__(zero, seq_op, comb_op)
-        # TODO: Fill in
         self.zero = zero
         self.seq_op = seq_op
         self.comb_op = comb_op
--- a/hail/python/hail/ir/ir.py
+++ b/hail/python/hail/ir/ir.py
@@ -1888,10 +1888,44 @@ class ApplyScanOp(BaseApplyAggOp):
 
 
 class AggFold(IR):
-    @typecheck_method(zero=IR, seq_op=IR, comb_op=IR, element_name=str, accum_name=str, other_state_name=str)
-    def __init__(self, zero, seq_op, comb_ob, element_name, accum_name):
-        super().__init__(zero, seq_op, comb_ob)
+    @typecheck_method(zero=IR, seq_op=IR, comb_op=IR, accum_name=str, other_accum_name=str)
+    def __init__(self, zero, seq_op, comb_op, accum_name, other_accum_name):
+        super().__init__(zero, seq_op, comb_op)
         # TODO: Fill in
+        self.zero = zero
+        self.seq_op = seq_op
+        self.comb_op = comb_op
+        self.accum_name = accum_name
+        self.other_accum_name = other_accum_name
+
+    def copy(self, zero, seq_op, comb_op):
+        return AggFold(zero, seq_op, comb_op, self.accum_name, self.other_accum_name)
+
+    def _compute_type(self, env, agg_env):
+        self.zero._compute_type(env, agg_env)
+        self.seq_op._compute_type(_env_bind(env, self.bindings(1)), agg_env)
+        self.comb_op._compute_type(_env_bind(env, self.bindings(2)), agg_env)
+
+        assert self.zero._type == self.seq_op._type
+        assert self.zero._type == self.comb_op._type
+
+        self._type = self.zero._type
+
+    def renderable_bindings(self, i: int, default_value=None):
+        dict_so_far = {}
+        if i == 1 or i == 2:
+            if default_value is None:
+                dict_so_far[self.accum_name] = self.zero.typ
+            else:
+                raise ValueError("TODO: Not sure")
+
+        if i == 2:
+            if default_value is None:
+                dict_so_far[self.other_accum_name] = self.zero.typ
+            else:
+                raise ValueError("TODO: Still note sure.")
+
+        return dict_so_far
 
 
 class Begin(IR):
--- a/hail/python/hail/ir/ir.py
+++ b/hail/python/hail/ir/ir.py
@@ -1888,8 +1888,8 @@ class ApplyScanOp(BaseApplyAggOp):
 
 
 class AggFold(IR):
-    @typecheck_method(zero=IR, seq_op=IR, comb_op=IR, accum_name=str, other_accum_name=str)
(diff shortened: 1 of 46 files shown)


hail/python/hail/ir/ir.py
```diff
@@ -1887,6 +1887,56 @@ def renderable_uses_scan_context(self, i: int):
...
+    def __init__(self, zero, seq_op, comb_op, accum_name, other_accum_name, is_scan):
+        super().__init__(zero, seq_op, comb_op)
+        self.zero = zero
+        self.seq_op = seq_op
+        self.comb_op = comb_op
+        self.accum_name = accum_name
+        self.other_accum_name = other_accum_name
+        self.is_scan = is_scan
+
+        if self.comb_op.free_vars != set([accum_name, other_accum_name]):
```
tpoterba, line 1901: I think this is unnecessarily strict. It should be possible to *not* use one of the vars. I think you want set subtraction == empty set.

hail/python/test/hail/expr/test_expr.py
```diff
@@ -482,6 +482,49 @@ def test_counter(self):
...
+        mt = hl.utils.range_matrix_table(15, 10, 5)
+        mt = mt.annotate_rows(s=hl.scan.fold(0, lambda a: a + mt.row_idx, lambda a, b: a + b))
+        mt = mt.annotate_rows(x=hl.scan.fold(0, lambda s: s + 1, lambda a, b: a + b))
+        self.assertEqual(mt.s.collect(), [0, 0, 1, 3, 6, 10, 15, 21, 28, 36, 45, 55, 66, 78, 91])
+        self.assertEqual(mt.rows().collect(),
+                         [hl.Struct(row_idx=0, s=0, x=0), hl.Struct(row_idx=1, s=0, x=1), hl.Struct(row_idx=2, s=1, x=2),
+                          hl.Struct(row_idx=3, s=3, x=3), hl.Struct(row_idx=4, s=6, x=4), hl.Struct(row_idx=5, s=10, x=5),
+                          hl.Struct(row_idx=6, s=15, x=6), hl.Struct(row_idx=7, s=21, x=7), hl.Struct(row_idx=8, s=28, x=8),
+                          hl.Struct(row_idx=9, s=36, x=9), hl.Struct(row_idx=10, s=45, x=10), hl.Struct(row_idx=11, s=55, x=11),
+                          hl.Struct(row_idx=12, s=66, x=12), hl.Struct(row_idx=13, s=78, x=13), hl.Struct(row_idx=14, s=91, x=14)])
```
tpoterba, line 527: Could you add a test for the following:

```
hl.agg.fold(hl.int32(0), lambda acc: acc + hl.float32(ht.idx), lambda acc1, acc2: hl.float64(acc1) + hl.float64(acc2))
```

This should work and result in a state type of float64, right?

hail/src/main/scala/is/hail/expr/ir/agg/FoldAggregator.scala
```diff
@@ -0,0 +1,40 @@
...
+    val pEnv = Env.apply((accumName, stateEV.st.storageType()), (otherAccumName, otherEV.st.storageType()))
+
+    val emitCtx = EmitContext.analyze(ctx, combOpIR, pEnv)
+    val emit = new Emit[Any](emitCtx, cb.emb.ecb.asInstanceOf[EmitClassBuilder[Any]])
+    val ec = emit.emit(combOpIR, cb.emb.asInstanceOf[EmitMethodBuilder[Any]], env, None)
+    ec.toI(cb).consume(cb, state.storeMissing(cb), sv => state.storeNonmissing(cb, sv))
+  }
+
+  override protected def _result(cb: EmitCodeBuilder, state: State, region: Value[Region]): IEmitCode = {
+    state.get(cb).map(cb){ sv => sv.copyToRegion(cb, region, sv.st)}  // TODO: Confirm this is necessary
```
tpoterba, line 38: This is necessary, yes.

tpoterba (changes requested):
Such a cool change! I have some comments, let me know if you want to discuss any of them.

hail/python/hail/ir/ir.py
johnc1231: Great point, thanks

hail/python/test/hail/expr/test_expr.py
johnc1231: Good test suggestion, caught a but where I was not casting the `init_op` value. Now it works correctly.

hail/src/main/scala/is/hail/expr/ir/agg/Extract.scala
```diff
@@ -97,6 +101,9 @@ case class NDArrayMultiplyAddStateSig(nda: VirtualTypeWithReq) extends AggStateS
   require(!nda.r.required)
 }
 
+// TODO: What is t? I am giving an empty array, is that right? wrong?
+case class FoldStateSig(initOpTypes: IndexedSeq[Type], seqOpTypes: IndexedSeq[Type], resultEmitType: EmitType, accumName: String, otherAccumName: String, combOpIR: IR, r: RequirednessAnalysis) extends AggStateSig(Array[VirtualTypeWithReq](), None)
```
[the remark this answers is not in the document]
johnc1231: Fixing `agg_state_signature` is straightforward. For the state types, is `t` supposed to be the type of the result of the aggregator then? Not sure why some aggregators get to leave it as empty list.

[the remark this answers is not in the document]
tpoterba: which are leaving it as an empty list?

The type here is basically a parameterization of the agg state using requiredness.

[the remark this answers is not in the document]
johnc1231:
```
case class CallStatsStateSig() extends AggStateSig(Array[VirtualTypeWithReq](), None)
case class ImputeTypeStateSig() extends AggStateSig(Array[VirtualTypeWithReq](), None)
```

hail/src/main/scala/is/hail/expr/ir/agg/Extract.scala
```diff
@@ -391,6 +400,22 @@ object Extract {
           i
         })
         GetTupleElement(result, idx)
+      case x@AggFold(zero, seqOp, combOp, accumName, otherAccumName, isScan) =>
+        val idx = memo.getOrElseUpdate(x, {
+          val i = ab.length
+          val initOpArgs = IndexedSeq(zero)
+          val seqOpArgs = IndexedSeq(seqOp)
+          val op = Fold()
+          val resultEmitType = EmitType(SType.canonical(zero.typ), false) // TODO: Can we ever decide this is required?
```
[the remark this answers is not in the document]
johnc1231: I tried this, ran into issues with the fact that the `resultEmitType` gets used for the requiredness of the Refs to the bindings as well.

johnc1231:
Fixed the Pretty / serialization issues.

hail/src/main/scala/is/hail/expr/ir/agg/Extract.scala
```diff
@@ -391,6 +399,22 @@ object Extract {
           i
         })
         GetTupleElement(result, idx)
+      case x@AggFold(zero, seqOp, combOp, accumName, otherAccumName, isScan) =>
+        val idx = memo.getOrElseUpdate(x, {
+          val i = ab.length
+          val initOpArgs = IndexedSeq(zero)
+          val seqOpArgs = IndexedSeq(seqOp)
+          val op = Fold()
+          val resultEmitType = EmitType(SType.canonical(zero.typ), false) // FIXME: Should be able to be true sometimes.
```
tpoterba, line 408: This produces a deeply-optional state type. that's going to hurt us for sure.

Can we talk through the issue with the requiredness and refs?

johnc1231: I haven't dug into it yet, I can try this afternoon. What I know so far is that sometimes the analysis is deciding that `AggFold` is required, but that at least one of the `Ref`s inside the `seqOp` or `combOp` aren't. The `Requiredness` rule for `AggFold` probably just isn't right.

johnc1231: What do you mean by "deeply optional", I thought it was just optional at top level?

tpoterba: This piece:
```
SType.canonical(zero.typ)
```

That's currently implemented as 
```
  def canonical(virt: Type): SType = {
    PType.canonical(virt).sType
  }
```

which aliases
```
PType.canonical(t, required=false, innerRequired=false)
```

johnc1231: Great, I see. Changed it to:

```
val resultSType = r(x).canonicalPType(x.typ).sType
val resultEmitType = EmitType(resultSType, false)
```

Which should work since it's based on the requiredness analysis.

tpoterba: Can we do `r(x).canonicalEmitType`?

tpoterba: or does that hit the requiredness mismatch issue above?

johnc1231: Currently we can't, working on understanding why.

johnc1231: Fixed, see: https://github.com/hail-is/hail/pull/10885/commits/9c940c7afd9c6a2e33c51436eda27bc88b41cda6

johnc1231:
Should be all set now, the problem was I was using `Stype.storageType` when compiling the `combOp` in the `FoldAggregator` which caused everything to be optional.

tpoterba:
Very nice!

tpoterba (approved):

PR #10892, opened 2021-09-22
[vds] Implement resumable VariantDataset combiner state machine

chrisvittal:
Create a new class `VariantDatasetCombiner` that implements a simple state
machine for running a VariantDataset combiner pipeline from start to
finish.

The `VariantDatasetCombiner` structure contains pretty much everything the
old `experimental.vcf_combiner.run_combiner` method took as arguments:
`temp_path`, `output_path`, list of gvcf paths, etc.. What is new is the
addition of a `save_path` and `vds_paths`. The `save_path` is used as
a filesystem or object storage location to save the state of the
combiner so that it may be resumed. The vds paths implement the long
awaited mixed hail and gvcf inputs, that was never completed for the VCF
combiner. End users should not call `VariantDatasetCombiner.__init__`
instead preferring the free function new_combiner. The
`VariantDatasetCombiner` API is fairly simple:

    * @property finished
    * save: serialize the current state to save_path
    * step: does the next unit of work for the combiner (described in more detail below)

The public api is present in 3 free functions:

    * run_combiner: just `new_combiner` followed by `step` in a loop until done
    * new_combiner: public VariantDatasetCombiner constructor
    * load_combiner

if `new_combiner` is not given a `save_path`, it will compute one based on
the SHA-256 of it's inputs. If the file at `save_path` is present, and
a valid `VariantDatasetCombiner`, it will load it and use it, unless the
`force` argument to `new_combiner` is True. This way, every combiner has
a `save_path`.

## The `step` algorithm

If there are any gvcf paths still uncombined, they are combined
`branch_factor` at a time, using `batch_size` parallel matrix table
writes to achive good parallelism. If there is only one merged vds, and
no vds arguments remaining to be combined it is written to
`output_path`. Otherwise the new vds paths are appended to the vds list
and the vds list is sorted by number of samples.

The vds paths are always kept sorted by samples and new items are only
ever added to the end of the list and then sorted. Since python's sort
function is stable, this will produce a stable output for a given set of
inputs and batch size. In order to preserve the gvcf sample ordering
from the paths, we work from the start of the gvcf list.

Once all the gvcfs are combined, we combine the variant datasets in
reverse order (smallest number of samples to largest number of samples)
inserting the new vds into the same sorted order until there are no
remaining input vdses, writing it to `output_path`

## Not Implemented Yet

Other stepping and planning strategies may be preferable in the future,
for example we could work the vds arguments into a list of lists based
on the log base `branch_factor` of the number of samples to create
a tiered list that is processed in order rather than reinserting new
datasets into a sorted list. This may create opportunites for large
scale parallelism on a future service backend.

### DIFF
--- a/hail/python/hail/vds/combiner/variant_dataset_combiner.py
+++ b/hail/python/hail/vds/combiner/variant_dataset_combiner.py
@@ -2,6 +2,7 @@ import collections
 import hashlib
 import itertools
 import json
+import math
 import os
 import sys
 import uuid
@@ -61,7 +62,7 @@ class VariantDatasetCombiner:  # pylint: disable=too-many-instance-attributes
         'gvcf_import_intervals',
     ]
 
-    __slots__ = tuple(__serialized_slots__ + ['_uuid', '_job_id'])
+    __slots__ = tuple(__serialized_slots__ + ['_uuid', '_job_id', '__intervals_cache'])
 
     def __init__(self,
                  *,
@@ -114,6 +115,7 @@ class VariantDatasetCombiner:  # pylint: disable=too-many-instance-attributes
         self.gvcf_import_intervals = gvcf_import_intervals
         self._uuid = uuid.uuid4()
         self._job_id = 0
+        self.__intervals_cache = {}
 
     def __eq__(self, other):
         if other.__class__ != VariantDatasetCombiner:
@@ -203,9 +205,16 @@ class VariantDatasetCombiner:  # pylint: disable=too-many-instance-attributes
         largest_vds = files_to_merge[0]
         vds = hl.vds.read_vds(largest_vds.path)
         # we use the reference data since it generally has more rows than the variant data
-        intervals, intervals_dtype = calculate_new_intervals(vds.reference_data,
-                                                             self.target_records,
-                                                             os.path.join(temp_path, 'interval_checkpoint.ht'))
+
+        new_n_samples = sum(f.n_samples for f in files_to_merge)
+        interval_bin = math.ceil(math.log(new_n_samples, self.branch_factor))
+        intervals, intervals_dtype = self.__intervals_cache.get(interval_bin, (None, None))
+
+        if intervals is None:
+            intervals, intervals_dtype = calculate_new_intervals(vds.reference_data,
+                                                                 self.target_records,
+                                                                 os.path.join(temp_path, 'interval_checkpoint.ht'))
+            self.__intervals_cache[interval_bin] = (intervals, intervals_dtype)
 
         paths = [f.path for f in files_to_merge]
         vdss = read_variant_datasets(paths, intervals, intervals_dtype)
@@ -217,7 +226,6 @@ class VariantDatasetCombiner:  # pylint: disable=too-many-instance-attributes
 
         new_path = os.path.join(temp_path, 'dataset.vds')
         combined.write(new_path, overwrite=True)
-        new_n_samples = sum(f.n_samples for f in files_to_merge)
         self.vdses.append(VDSMetadata(path=new_path, n_samples=new_n_samples))
         self.vdses.sort(key=lambda x: x.n_samples, reverse=True)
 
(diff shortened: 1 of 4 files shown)


tpoterba (changes requested):
some initial thoughts. Will probably have more as I digest

hail/python/hail/vds/combiner/variant_dataset_combiner.py
```diff
@@ -0,0 +1,297 @@
...
+            combiner = json.load(stream, cls=Decoder)
+            if combiner.save_path != path:
+                warning('path/save_path mismatch in loaded VariantDatasetCombiner, using '
+                        f'{path} as the new save_path for this combiner')
+                combiner.save_path = path
+            return combiner
+
+    @property
+    def finished(self) -> bool:
+        return not self.gvcfs and not self.vdses
```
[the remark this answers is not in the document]
chrisvittal: I wasn't planning on having the final path in the vds array. I can though.

hail/python/hail/vds/combiner/variant_dataset_combiner.py
```diff
@@ -0,0 +1,297 @@
...
+        if other.__class__ != VariantDatasetCombiner:
+            return False
+        for slot in self.__serialized_slots__:
+            if getattr(self, slot) != getattr(other, slot):
+                return False
+        return True
+
+    def save(self):
+        fs = hl.current_backend().fs
+        with fs.open(self.save_path, 'w') as out:
```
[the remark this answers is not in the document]
chrisvittal: There's no atomic operations on the cloud, there will always be some failure condition that we can't account for, but I'll make this more robust.

hail/python/hail/vds/combiner/variant_dataset_combiner.py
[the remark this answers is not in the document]
tpoterba: no, that's OK

hail/python/hail/vds/combiner/variant_dataset_combiner.py
```diff
@@ -0,0 +1,445 @@
...
+        paths = [md.path for md in merge_metadata]
+        hl.vds.write_variant_datasets(merge_vds, paths, overwrite=True)
+        self.vdses.extend(merge_metadata)
+        self.vdses.sort(key=lambda x: x.n_samples, reverse=True)
+
+    def _temp_out_path(self, extra):
+        return os.path.join(self.temp_path, 'combiner-intermidiates', f'{self._uuid}_{extra}')
+
+
+def run_combiner(**kwargs):
```
tpoterba, line 260: I think we should remove this -- it's hard to document and confusing to have two entry points to creating a new combiner. Can we just add a `run()` method on the combiner it self that does the loop?

hail/python/hail/vds/combiner/variant_dataset_combiner.py
```diff
@@ -0,0 +1,445 @@
...
+
+    def _step_vdses(self):
+        files_to_merge = self.vdses[-self.branch_factor:]
+        self.vdses = self.vdses[:-self.branch_factor]
+
+        temp_path = self._temp_out_path(f'vds-combine_job{self._job_id}')
+        largest_vds = files_to_merge[0]
+        vds = hl.vds.read_vds(largest_vds.path)
+        # we use the reference data since it generally has more rows than the variant data
+        intervals, intervals_dtype = calculate_new_intervals(vds.reference_data,
```
tpoterba, line 198: This is going to get executed for each VDS combine 'job', rather than once per 'phase'. How can we avoid this? Maybe we can cache the intervals in buckets that refer to ranges of sample sizes?

tpoterba (changes requested):
This looks great. A couple of comments, one of which might be tricky.

hail/python/hail/vds/combiner/variant_dataset_combiner.py
chrisvittal: see 89f66e2

tpoterba: beautiful.

tpoterba (approved):

PR #10894, opened 2021-09-24
Refactor ResultOp to return a single element instead of a tuple.

johnc1231:
This PR:

1. Refactors `ResultOp` to just take an index and return whatever agg result is at that index. No more returning a suffix of the total aggregator tuple based on a starting index. This is necessary for me to effectively implement my Fold aggregator, which will be included in a subsequent PR. 
2. Pushes `EmitType` through aggregators, uses them as the basis for analyzing the requiredness of aggregator results.
3. Changes `_storeResult` on aggregators to instead just be `_result`, which directly returns an `IEmitCode`. No reason that `ResultOp` had to be so wound up in `PType`s, and for the most part this made the code simpler.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/agg/CallStatsAggregator.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/agg/CallStatsAggregator.scala
@@ -5,6 +5,8 @@ import is.hail.asm4s._
 import is.hail.expr.ir.{EmitClassBuilder, EmitCode, EmitCodeBuilder, IEmitCode}
 import is.hail.io.{BufferSpec, InputBuffer, OutputBuffer, TypedCodecSpec}
 import is.hail.types.physical._
+import is.hail.types.physical.stypes.EmitType
+import is.hail.types.physical.stypes.concrete.SBaseStructPointer
 import is.hail.types.virtual.{TCall, TInt32, Type}
 import is.hail.types.physical.stypes.interfaces._
 import is.hail.utils._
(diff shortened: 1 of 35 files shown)


hail/src/main/scala/is/hail/expr/ir/agg/CallStatsAggregator.scala
```diff
@@ -16,7 +18,7 @@ object CallStatsState {
   val callStatsInternalArrayType = PCanonicalArray(PInt32Required, required = true)
   val stateType: PCanonicalTuple = PCanonicalTuple(true, callStatsInternalArrayType, callStatsInternalArrayType)
 
-  val resultType = PCanonicalStruct(required = true,
+  val resultPType = PCanonicalStruct(required = false,
```
johnc1231, line 21: Is this the right thing to do? I'm doing it because I can't construct a `SBaseStructPointer` of a `PCanonicalStruct` that is `required` (we have an assertion forbidding it). This is causing a small test failure in a requiredness test of nested aggregators where we have hardcoded that we expect it to be required. I still don't understand why we have these assertions.

johnc1231: @tpoterba , Patrick said he is good with this PR but doesn't know the answer to this.

tpoterba: this is the right thing to do, yes, if you're still using a PCanonicalStruct. I think we should follow up this change (which is fine as-is0 with an audit to ensure the results of the aggregators aren't doing undue allocation -- here we can certainly be using a stack struct!

patrick-schultz (approved):

PR #10896, opened 2021-09-27
[vds] Add validate() method to verify assumptions about the VDS representation

tpoterba:


hail/python/hail/vds/variant_dataset.py
```diff
@@ -115,3 +115,96 @@ def write(self, path, **kwargs):
...
+            error(f"expect reference data to have a single row key 'locus' of type locus, found {rd_row_key}")
+
+        vd_row_key = vd.row_key.dtype
+        if (not isinstance(vd_row_key, hl.tstruct)
+                or len(vd_row_key) != 2
+                or not vd_row_key.fields == ('locus', 'alleles')
+                or not isinstance(vd_row_key.types[0], hl.tlocus)
+                or vd_row_key.types[1] != hl.tarray(hl.tstr)):
+            error(
+                f"expect variant data to have a row key {{'locus': locus<rg>, alleles: array<str>}}, found {vd_row_key}")
```
chrisvittal, line 142: In many cases, a VDS will have row keys of just `locus`, it may be better to warn the user and rekey rather than failing here.

chrisvittal (changes requested):

hail/python/hail/vds/variant_dataset.py
tpoterba: I think including this as a requirement of having a VDS will make our lives easier. What are the cases where it won't have the alleles key? The combiner? It should be easy for us to return a keyed result from each step.

tpoterba: actually, the combiner methods are broken if the alleles are part of the key. This is another reason to have all VDSes be keyed by both locus and alleles.

chrisvittal (approved):

PR #10898, opened 2021-09-27
[aiocloud] Add hailtop.aiocloud and Azure Client Infrastructure

jigold:
- Reorganized hailtop to no longer have aiogoogle. Instead there is aiocloud.{aws,azure,gcp}.
- Moved AzureAsyncFS and S3AsyncFS to aiocloud modules from aiotools
- Added a new Azure implementation of the Compute Client (ActivityLogsClient, NetworkClient will follow)


jigold:
FYI @danking. If you don't like how this is structured, then please give me high level feedback on what you would do differently.

daniel-goldstein (changes requested):
I think there are a couple related points to address here. I had a hard time gathering my thoughts succinctly so I hope I'm not talking in circles. Please let me know if I should better articulate any of my thoughts.

1. I think there's a danger of over-abstracting to the point where interfaces and reuse start getting in the way of understanding. Particularly, I point out that I prefer the azure `AccessToken` class be called `AzureAccessToken`, because that `AccessToken` can't and shouldn't be used for anything other than Azure. The name supposes a wider utility than it can deliver. Despite being made unambiguous by the directory structure (which I think is good), I don't like having to rely on the filesystem to understand the contents of a file. It makes such things like searching the codebase for `AccessToken(` more difficult to interpret and could introduce collisions if we were to say export multiple `AccessToken` classes from the same module.

2. This might be slightly out of scope of this PR and admittedly comes from a distaste for heavily OO designs so let me know what you think is a reasonable line to draw. I think it is important to identify the visibility boundaries of a class and stop using indirection when it is no longer necessary to hide a particular implementation detail.

The boundary between the `aiocloud` module and the rest of our code seems to me to be the clients. Once I get down to the level of the specific client, say a `GCPClient`, we've torn away the cloud abstraction. But this then *reintroduces* that abstraction by upcasting the `GCPSession` and subsequently the `AccessToken`. Once I'm in a `GCP*`, I think any `GCPAccessToken` I see should be treated unambiguously as a `GCPAccessToken`.

I think something at the root of this is that this PR takes what was previously a concrete class with no indirection `Session` and inserted an interface at `AccessToken`. It's unclear to me now whether `Session` should ever be instantiated. It (and `AccessToken`) appear in some places to be used as interfaces, because they are used as the types of fields in a class. But in practice they look more like abstract classes meant to consolidate functionality. I generally don't find the `AccessToken` interface very compelling, because it is quite small and swapping implementations would be disastrous! Here's one idea that I think would streamline this a bit more and at least make it easier for me to follow.

By virtue of having GCPSession and AzureSession, `Session` is demoted to an abstract class (it shouldn't be instantiated because there are more specific extensions, and it's not the root interface, so it should only exist to consolidate functionality). So I say to remove access_token as a field of `Session`, rename `Session` to something like `AbstractAuthenticatedSession`, and to the same effect we can write `GCPSession` as:

```python
class GCPSession(AbstractAuthenticatedSession):
    def __init__(self, *, credentials: GCPCredentials = None, params: Optional[Mapping[str, str]] = None, **kwargs):
        self.access_token = GCPAccessToken(credentials or GCPCredentials.default_credentials())
        super().__init__(params=params, **kwargs)
        
    async def request(self, method: str, url: str, **kwargs):
         headers = await self.access_token.auth_headers(self._session)
        return await super().request(method, url, headers=headers, **kwargs)
```

This doesn't add much code, but it removes an indirection, and we no longer need the inheritance tree for `AccessToken`. I think one less hierarchy is worth the couple more lines of code. It also allows us to more easily consolidate GCP-specific code because the `AccessToken` is an implementation detail of the `GCPSession`, not the more broad `Session`. I think this could be taken farther, and that the authentication work that `Session` is doing could probably just be moved into a function that adds headers when given the result of `auth_headers`, called in the above `request` implementation as one extra line, and then the `Session` class could be removed entirely and further simplify the hierarchy. But I think this illustrates my point all the same.

atgu/atgu/atgu.py
```diff
@@ -12,7 +12,7 @@
 from hailtop.hail_logging import AccessLogger
 from hailtop.tls import internal_server_ssl_context
 from hailtop.config import get_deploy_config
-from hailtop import aiogoogle
+from hailtop.aiocloud import gcp as aiogoogle
```
danking, line 15: minor point: I think import with rename is more confusing in the long term than picking a consistent, simple naming system.

I think I'm roughly fine with `hailtop.aiocloud import gcp`. Using `gcp` as a variable name seems rare. Naming the module `aiogoogle` (and naming the other ones `aioazure`, `aioaws`) also seems fine.

hail/python/hailtop/aiocloud/aioazure/client/compute_client.py
```diff
@@ -1,9 +1,11 @@
+from typing import Optional
+
 from ..session import AzureSession
 from .base_client import AzureBaseClient
 
 
 class ComputeClient(AzureBaseClient):
-    def __init__(self, subscription_id, resource_group_name, session=None):
+    def __init__(self, subscription_id, resource_group_name, session: Optional[AzureSession] = None):
```
daniel-goldstein, line 8: For consistency I think this ought to be called `AzureComputeClient`.

daniel-goldstein (changes requested):
I find this much easier to understand now, thanks! Only one small naming comment.

daniel-goldstein (approved):
Looks good! Just a lint check and import issue in memory

PR #10900, opened 2021-09-28
There was no --region flag in submit preventing submission to clusters

Fedja:
created in different regions


Fedja:
Completely missed this 

> This is failing style checks. There need to be two blank lines before the main function and only one blank line after the added if statement.

sorry completely missed this rev. comment. Just noticed when colleague tried to launch a pipeline with this feature assumed in the submitter. Fixed style issues now

tpoterba:
running tests now!

PR #10907, opened 2021-09-30
[compiler] (S)Code->(S)Value in invoke

patrick-schultz:
~Stacked on #10906~

This PR refactors `MethodBuilder.invoke` and `EmitCodeBuilder.invoke(S)Code` to take/return values.

* `invoke` now takes a `CodeBuilderLike`. It is used in places where there is only access to a `CodeBuilder` (not an `EmitCodeBuilder`), so I had to use the generic interface, and had to move a couple methods on `EmitCodeBuilder` to `CodeBuilderLike`. I have never understood this Emit/non-Emit split; would be a great simplification if we could collapse it.
* This change pushed some (S)Code->(S)Value refactorings inside some aggregator implementations, which generate and invoke internal methods.
* I had to keep a version of `MethodBuilder.invoke` that doesn't take a CodeBuilder, for use in `ThisLazyFieldRef.get`. Will have to think more about how this should work when Code is (mostly) gone. Maybe lazy fields should not subclass Value, and to access a lazy field requires an explicit `load(cb: CodeBuilder): Value[T]`.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/agg/TakeByAggregator.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/agg/TakeByAggregator.scala
@@ -195,7 +195,8 @@ class TakeByRVAS(val valueVType: VirtualTypeWithReq, val keyVType: VirtualTypeWi
     }
   }
 
-  private def elementOffset(i: Value[Int]): Code[Long] = ab.elementOffset(i)
+  private def elementOffset(cb: EmitCodeBuilder, i: Value[Int]): Value[Long] =
+    ab.elementOffset(cb, i)
 
   private def keyIsMissing(cb: EmitCodeBuilder, offset: Code[Long]): Value[Boolean] =
     indexedKeyType.isFieldMissing(cb, offset, 0)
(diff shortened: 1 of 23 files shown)


tpoterba:
> Maybe lazy fields should not subclass Value, and to access a lazy field requires an explicit load(cb: CodeBuilder): Value[T].

I think this is probably the right model.

hail/src/main/scala/is/hail/expr/ir/agg/TakeByAggregator.scala
```diff
@@ -469,47 +466,51 @@ class TakeByRVAS(val valueVType: VirtualTypeWithReq, val keyVType: VirtualTypeWi
...
-                swap(indexOffset(low), indexOffset(high)),
-                low := low + 1,
-                high := high - 1))),
+        def indexOffset(cb: EmitCodeBuilder, idx: Value[Int]): Value[Long] =
+          cb.memoize(indices + idx.toL * 4L)
+
+        def indexAt(cb: EmitCodeBuilder, idx: Value[Int]): Value[Int] =
+          cb.memoize(Region.loadInt(indexOffset(cb, idx)))
+
+        mb.emitWithBuilder { cb =>
```
tpoterba, line 475: yay! The fact this was all still Code[T] instead of CodeBuilder bothered me :D

tpoterba (approved):

PR #10911, opened 2021-09-30
[ci] Mostly pass through global-config in CI instead of constructing from env variables

daniel-goldstein:
There are many global config fields that CI needs in order to template build.yaml jobs that are threaded through to CI with environment variables. However, these variables are never actually used by CI and they introduce some needless dependencies to run CI (you need a GCP_PROJECT, for example, even though CI doesn't care at all). Instead of setting specific environment variables for each field that build.yaml steps need, I instead mount the global-config (read-only) to the CI container and read in the whole thing. This does potentially expose more variables to the build.yaml environment than there were previously, but I argue that none of those should be sensitive anyway or maybe don't belong in the global-config (which shouldn't be sensitive). This in part makes the process of adding global config fields easier, since right now you need separate PRs to 1) introduce the field to CI and then 2) use it in a new build.yaml step. It also makes the CI deployment.yaml cloud-agnostic. 

cc @jigold

### DIFF
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -23,8 +23,10 @@ log = logging.getLogger('ci')
 
 global_config = {}
 for field in os.listdir('/global-config/'):
-    with open(f'/global-config/{field}') as value:
-        global_config[field] = value.read()
+    # Kubernetes inserts some hidden files that we don't care about
+    if not field.startswith('.'):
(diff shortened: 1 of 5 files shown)


daniel-goldstein:
@danking bump

danking (approved):

PR #10915, opened 2021-09-30
[query] remove sl4fj dependency

danking:
This might be the cause of https://discuss.hail.is/t/hail-py4jerror-while-calling-z-is-hail-backend-spark-sparkbackend-apply/2272/14.

### DIFF
--- a/hail/build.gradle
+++ b/hail/build.gradle
@@ -174,8 +174,6 @@ dependencies {
         transitive = false
     }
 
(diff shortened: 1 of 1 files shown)


johnc1231 (approved):

danking:
This had no effect.

johnc1231:
How do you know?

danking:
The user on the forum tried a wheel I made for them and they encountered the exact same error.

PR #10916, opened 2021-09-30
[gear/auth] do not log on cancelled error

danking:


jigold (approved):

jigold (changes requested):
x

gear/gear/auth.py
```diff
@@ -24,6 +24,8 @@ def maybe_parse_bearer_header(value: str) -> Optional[str]:
 async def _userdata_from_session_id(session_id):
     try:
         return await async_get_userinfo(deploy_config=deploy_config, session_id=session_id)
+    except CancelledError:
```
jigold, line 27: Shouldn't this be `asyncio.CancelledError`?

jigold (changes requested):

PR #10918, opened 2021-09-30
[batch] use correct not found error

danking:
The `GoogleStorageAsyncFS` raises `FileNotFoundError` when a file is not found.


jigold (approved):
good catch

johnc1231 (approved):

PR #10919, opened 2021-10-04
[infra] Add scripts for azure setup and sql/global configs

daniel-goldstein:
This adds the following pieces of infrastructure:

- Fully scripted bootstrapping process, from creating a managed identity to run terraform through (instead of the current service principal), to creating a VM to run the bootstrap process off of, through all following steps until running bootstrap.py
- Adds the root CA certificate that azure uses to sign the MySQL server certificates so that we can connect to the database with `VERIFY_CA`. Unlike gcp, however, this still doesn't allow us to use mTLS since it doesn't look like we can request a client cert/key for our database. Still this is not so bad for now.
- Creates a separate k8s module for terraform. This currently just holds the global-config and sql-config resources, but establishes a boundary between the cloud-specific terraform and purely k8s terraform. Later on I'll refactor the GCP terraform to use the k8s module so that different clouds can use the same k8s configuration.
- Adds a pool of spot instances to the AKS cluster and adds the required toleration to all of our preemptible deployments. Part of the node selection process for a pod requires that exist a toleration on the pod for every taint on the node. In other words, it is ok for a pod to have redundant tolerations, so it's fine to have azure-specific tolerations even if we're running in gcp.
- Refactor the az-create-worker-image.sh script to complete the entire batch worker image creation process from start to finish. This involved sending a command over ssh that previously had to be executed by hand. This meant we could combine the two-script process into one shell script.

Stacked on #10911


infra/azure/main.tf
```diff
@@ -18,6 +18,7 @@ provider "azurerm" {
 
 locals {
   acr_name = var.acr_name == "" ? var.az_resource_group_name : var.acr_name
+  internal_ip = "10.128.255.254"
```
jigold, line 21: Where did this come from?

infra/azure/outputs.tf
```diff
@@ -0,0 +1,27 @@
...
+  value     = azurerm_kubernetes_cluster.vdc.kube_config_raw
+  sensitive = true
+}
+
+output "global_config" {
+  value = {
+    cloud = "azure",
+    domain = var.domain,
+    docker_prefix = azurerm_container_registry.acr.login_server,
+    internal_ip = local.internal_ip,
```
jigold, line 11: Do we also need resource group and subscription id here? I need that for the AzureComputeClient etc.

jigold (changes requested):
Looks great!

PR #10920, opened 2021-10-04
[batch] Make batch cloud-agnostic (Attempt 2)

jigold:
This PR tries to address the comments in #10860. However, there are a bunch of additional changes as well. I did everything necessary to make the entire batch code base cloud-agnostic. Before I fully test the entire system and someone proofreads this PR carefully, I would like your input on the design and get high level comments. It's a lot of work to test this branch!!!! Once everyone is happy with this structure, then I can start writing the Azure components.

List of changes:
- Created a new CloudResourceManager that has the interface for interacting with VMs and managing cloud resources
- Renamed WorkerConfig to InstanceConfig
- Created a new Disk interface in batch/worker/disk.py
- Renamed the instance's zone field to location as Azure's zones mean something different.
- Moved all resource utility calculations to its own module
- resource_utils.py contains a set of functions that calls the appropriate implementation based on the cloud parameter

I moved all GCP specific implementations into the batch/gcp module.

There is one cyclic dependency I couldn't break right now which is the InstanceCollectionManager -> InstanceCollection -> Pool/JPIM all need a CloudResourceManager while the CloudResourceManager needs to know the list of active instances and the global number of cores from InstanceCollectionManager to appropriately clean up resources and select a zone. I think of the InstanceCollectionManager hierarchy as being the manager of virtual instances and the CloudResourceManager manages the physical VM instances.

I had to put the constructor functions for the InstanceConfig in utils.py because of cyclic imports even though the import was inside the static method constructor function in my original attempt.


jigold:
I didn't quite everything done in worker.py or managing the different secrets. However, I'd still appreciate feedback on the structure before I work on this more.

danking:
OK, I have a version of this branch that's nearly ready and addresses the add_instance issue as well as some other issues I hit along the way to fixing that. Unfortunately, it's not compatible with hail-is/main? What happened to batch_configuration.py? It seems to be using the global config even though the main branch batch_configuration does not.

jigold:
This is what I was trying to tell you last week. Daniel had a PR that changed the storage URI that didn't add them from the global config yet and there were going to be merge conflicts based on who merged first. Sorry :(

#11014

danking:
OK, I resolve those issues.

danking:
OK, so the big insight is that "InstanceConfig" is really just "ResourcesForAParticularInstance" (well, and, sometimes, "ResourcesOfARepresentativeInstance"). I trimmed the InstanceConfig down *significantly* removing the "vm_config". Now the InstanceConfig is cheap and easy to create and there's no circularity between vm_config and instance config.

I pushed that through everywhere and then abstracted the common create_instance logic for pool and job-private into InstanceCollection.

With both of those changes, I was able to modify the ResourceManager's API to expose methods for constructing instance configs. However, the instance config isn't critical to the operation of the ResourceManager. It's just an interface for communicating an instance's resources to the rest of the code base.

danking:
I also converted stress into a proper test which we can start running on every PR. I'll PR that separately. I'm currently running that test on my changes. You can see all the changes here: https://github.com/hail-is/hail/pull/11029/commits/8d4c7a22a19ae0a79527eae790d537cf020c1cca. The diff is bad because I merged in main. I removed all the global_config things and tried to keep the minimal diff from main. We can make those changes, if desired, at a later date.

danking:
Erm, sorry, the right place to look for a diff is: https://github.com/hail-is/hail/compare/main...danking:batch-cloud-agnostic-rewrite-all

jigold:
The reason for the circularity is because I added this code to get the monitoring in Grafana/Prometheus of instance costs.

https://github.com/hail-is/hail/commit/8d4c7a22a19ae0a79527eae790d537cf020c1cca#diff-0eb2a9c198baaf288c71080bf314f821303f734f13b9b15bb9767cad8ba9bbe7R1012-R1022

jigold:
If you look at the driver logs, there should be a bunch of errors.

danking:
I'll keep an eye out for that, but I nix'ed the circularity by taking the vm_config off of the instance & instance config. Every instance still has an instance_config which I made non-optional. It basically has exactly one function now: tell me how much this instance costs per hour.

jigold:
Ok. This exact scenario is what I was worried about when we merge PRs without checking the logs by hand in a full testing scenario. I want a way to check the PR driver, front-end, and worker logs automatically that they don't have ERROR messages. Like test_invariants. For example, I'm still looking at your change for time_since_last_state_change. When I had the code you wanted, there were errors because time_since_last_state_change was None. The current tests would not have caught that. I think we need either a white list of acceptable front-end/driver errors or some kind of threshold for error types. I'll think about it some more once the batch porting is done.

danking:
Yes! I definitely want to assert no errors in the logs. I think we're very close to being able to assert that.

danking:
OK, I tested my branch twice with stress. No error logs. Many warnings due to known deadlock errors. Otherwise it looks clean. I've pushed those changes onto this branch. Let's merge tomorrow first thing.

danking (dismissed):

danking:
Reminder to future self: this isn't merging because it is WIP.

batch/batch/cloud/gcp/instance_config.py
```diff
@@ -0,0 +1,114 @@
...
+            disks = data['disks']
+            assert len(disks) == 2
+            assert disks[0]['boot']
+            boot_disk_size_gb = disks[0]['size']
+            assert not disks[1]['boot']
+            local_ssd_data_disk = disks[1]['type'] == 'local-ssd'
+            data_disk_size_gb = disks[1]['size']
+            job_private = data['job-private']
+            return GCPSlimInstanceConfig(
+                data['vm_config']['machineType'],
```
jigold, line 48: I'm not sure this is correct. I think `vm_config` is something I added in this PR and didn't exist before. I'm not 100% sure though.

batch/batch/cloud/gcp/driver/resource_manager.py
```diff
@@ -0,0 +1,133 @@
...
+                        data_disk_size_gb: int,
+                        boot_disk_size_gb: int,
+                        preemptible: bool,
+                        job_private: bool,
+                        location: str,
+                        machine_type: str,
+                        instance_config: InstanceConfig,
+                        ) -> List[Dict[str, Any]]:
+        if local_ssd_data_disk:
+            assert data_disk_size_gb == 375
```
jigold, line 112: Did you change the semantics here? It should be 0.

danking: The term "data_disk_size_gb" now means the size of the data disk, regardless of its type. I did this so I could centralize the definition of how big a local ssd is [in one place](https://github.com/hail-is/hail/pull/10920/files#diff-b3aa93ab91e349f8d38b2c10efeddd451eef1e1f884bb8d5527ed96cb19bf083R89). I think I still need to push that through everywhere. I saw a lot of code that was like this:
```
if is_local_ssd:
    return XYZ 375 ABC
else:
    return XYZ pd_ssd_size ABC
```
which said to me that there was a concept, data disk size, which wasn't represented directly in the code.

danking: I'll push it through fully in a separate PR. There's enough changes in here already.

batch/batch/cloud/gcp/instance_config.py
danking: Yeah,  you're right this isn't present.

batch/batch/cloud/gcp/instance_config.py
```diff
@@ -0,0 +1,105 @@
...
+        else:
+            data_disk_product = 'disk/pd-ssd/1'
+        return [
+            {'name': f'compute/{self._instance_family}-{preemptible}/1', 'quantity': cpu_in_mcpu},
+            {'name': f'memory/{self._instance_family}-{preemptible}/1', 'quantity': memory_in_bytes // 1024 // 1024},
+            {'name': 'disk/pd-ssd/1', 'quantity': self.boot_disk_size_gb * worker_fraction_in_1024ths},  # the factors of 1024 cancel between GiB -> MiB and fraction_1024 -> fraction
+            {'name': data_disk_product, 'quantity': self.data_disk_size_gb * worker_fraction_in_1024ths},  # the factors of 1024 cancel between GiB -> MiB and fraction_1024 -> fraction
+            {'name': 'disk/pd-ssd/1', 'quantity': extra_storage_in_gib * 1024},  # storage is in units of MiB
+            {'name': 'service-fee/1', 'quantity': cpu_in_mcpu},
+            {'name': 'ip-fee/1024/1', 'quantity': worker_fraction_in_1024ths},
```
danking, line 103: I really like how this turned out.

batch/batch/driver/instance_collection/job_private.py
```diff
@@ -235,58 +263,25 @@ def allocate_jobs(user, mark):
...
-            activation_token=activation_token,
-            max_idle_time_msecs=WORKER_MAX_IDLE_TIME_MSECS,
-            worker_local_ssd_data_disk=False,
-            worker_pd_ssd_data_disk_size_gb=storage_gb,
-            boot_disk_size_gb=self.boot_disk_size_gb,
-            preemptible=preemptible,
             job_private=True,
+            location=None,
+            preemptible=preemptible,
+            max_idle_time_msecs=None,
```
jigold, line 278: Why did you make the max idle time here None?

batch/batch/driver/main.py
```diff
@@ -455,7 +472,7 @@ async def pool_config_update(request, userdata):  # pylint: disable=unused-argum
 
     if not worker_local_ssd_data_disk:
         unreserved_disk_storage_gb = unreserved_worker_data_disk_size_gib(
-            worker_local_ssd_data_disk, worker_pd_ssd_data_disk_size_gb, worker_cores
+            pool.cloud, worker_local_ssd_data_disk, worker_pd_ssd_data_disk_size_gb, worker_cores
```
jigold, line 475: See line 465 above. I'm worried the input for `worker_pd_ssd_data_disk_size_gb` here is 0 when there's a local ssd. This is also the value in the database. I'm not sure how you reconciled that it should be 375 later on in the code where is here it is 0.

batch/batch/inst_coll_config.py
```diff
@@ -85,91 +83,56 @@ def __init__(
...
-                'type': self.worker_type,
-                'cores': self.worker_cores,
-                'preemptible': True,
-            },
-            'disks': disks,
-            'job-private': False,
-        }
-
-        return WorkerConfig(config)
+            return 375
```
jigold, line 91: I see. You're doing the conversion here.

batch/batch/worker/worker.py
```diff
@@ -107,35 +101,38 @@ def deeper_stack_level_warn(*args, **kwargs):
...
-ZONE = os.environ['ZONE'].rsplit('/', 1)[1]
 DOCKER_PREFIX = os.environ['DOCKER_PREFIX']
 PUBLIC_IMAGES = publicly_available_images(DOCKER_PREFIX)
-WORKER_CONFIG = json.loads(base64.b64decode(os.environ['WORKER_CONFIG']).decode())
+INSTANCE_CONFIG = json.loads(base64.b64decode(os.environ['INSTANCE_CONFIG']).decode())
+if CLOUD == 'gcp':
+    LOCATION = os.environ['LOCATION'].rsplit('/', 1)[1]
+else:
+    assert False, (CLOUD, os.environ.get('LOCATION'))
+PROJECT = os.environ['PROJECT']
```
jigold, line 111: Grabbing the project here will have to change in Azure but fine for now.

batch/batch/driver/instance.py
```diff
@@ -68,11 +75,11 @@ async def create(app, inst_coll, name, activation_token, instance_config: Instan
...
                 INSTANCE_VERSION,
-                instance_config.location,
+                location,
                 inst_coll.name,
-                instance_config.machine_type,
-                instance_config.preemptible,
-                base64.b64encode(json.dumps(instance_config.config).encode()).decode()
+                machine_type,
+                preemptible,
+                json.dumps(instance_config.to_dict()),
```
jigold, line 82: I'm worried about backwards compatibility here.

jigold:
It's fine except for the `max_idle_time_msecs` in JPIM and the backwards compatibility with the instance config in the database.

batch/batch/driver/instance.py
danking: ah the base64 bit? I should restore that, though I'm not sure why we did that.

jigold: I thought it was more compact to store.

danking: Regarding the null issue, no active instances will have null, so I don't think we need to handle that any more.

batch/batch/driver/instance_collection/job_private.py
danking: I suppose that's a bit confusing. None is the default value. I centralized that in InstanceCollection._create_instance. I'll inline the constant.

danking: Erm. I won't. I did it this way because in pool.py None is used to indicate the default max idea time. I pushed that convention everywhere for consistency

batch/batch/driver/main.py
danking: Ah, I see, sometimes we use PoolConfig to configure a pool and sometimes we use configure to configure a pool. I'll fix Pool.configure.

Your'e right that this function, `unreserved_worker_data_disk_size_gib` hasn't been updated to the new style yet. I'll do that in a subsequent PR.

danking: Ah, grr, this is annoying. The whole point was to centralize where we define ssd == 375, but since we use PoolConfig sometimes and configure other times, there are inherently two locations.

danking: OK, I consolidated all DB manipulation into PoolConfig. PoolConfig is basically our API to the database tables inst_colls and pools. Once the information comes out of the database, we create a Pool at which point we're just working with data_disk_size_gb.

danking: Good catch, this would've been a subtle bug to fix.

danking (approved):

danking:
I'm good. WIP is still on.

batch/batch/driver/instance.py
danking: Yeah, I suppose it should be more compact. Now that we're just storing a few fields I don't think it matters much.

PR #10923, opened 2021-10-05
[batch] aioclient.BatchClient must be created from async

danking:
I ran into issues in the development of the query service wherein, if the
aiohttp `ClientSession` is not created in the same event pool as the one
from which the request is made, then unusual errors occur.

This change makes it harder to accidentally create a BatchClient in the
wrong event loop because the factory method, `BatchClient.create` is
itself an async function. It is still possible to create the BatchClient
in one event loop and use it in another (do not do this!!), but that seems
to be an unlikely mistake.

cc: @daniel-goldstein


danking:
bump

PR #10932, opened 2021-10-05
[httpx] re-use client sessions, particularly for auth

danking:
While developing Hail Query on Hail Batch, I discovered that, for very fast requests,
I spent significant percentages of the time constructing aiohttp client sessions. This
change makes all the `XXX_authenticated_YYY_only` methods use a single, shared client
session from the `app`.

This change requires every server to have a `client_session` key in the `app` which
points at a `httpx.ClientSession`. While making this change, I also updated every
instance I saw of adhoc client session creation.


hail/python/hailtop/auth/auth.py
```diff
@@ -1,25 +1,26 @@
...
-    if client_session is None:
-        client_session = http_client_session()
-
     if session_id is None:
         headers = service_auth_headers(deploy_config, 'auth')
     else:
         headers = {'Authorization': f'Bearer {session_id}'}
 
     userinfo_url = deploy_config.url('auth', '/api/v1alpha/userinfo')
-    async with client_session as session:
```
daniel-goldstein, line 22: huh, how did this not prematurely close client sessions before?

daniel-goldstein (approved):
sweet

PR #10940, opened 2021-10-05
[hailtop] avoid code duplication in process.py

danking:


hail/python/hailtop/utils/process.py
```diff
@@ -42,6 +29,12 @@ async def check_exec_output(command, *args, echo=False):
     return outerr
 
 
+async def check_shell_output(script, echo=False):
+    if echo:
+        print(script)
```
daniel-goldstein, line 34: This'll double print here the script and then /bin/bash -c script. Seems like it'd be fine to just delete this echo?

daniel-goldstein (changes requested):

daniel-goldstein (approved):

PR #10947, opened 2021-10-05
[aiogoogle] make aiogoogle tests resilient

danking:
I received a 503 in one of my other PRs from one of these tests.

### DIFF
--- a/hail/python/test/hailtop/test_aiogoogle.py
+++ b/hail/python/test/hailtop/test_aiogoogle.py
@@ -76,10 +76,10 @@ async def test_compose():
     part_data = [b'a', b'bb', b'ccc']
 
     async with GoogleStorageClient() as client:
+        async def upload(i, b):
+            async with await client.insert_object(bucket, f'{token}/{i}') as f:
+                await f.write(b)
         for i, b in enumerate(part_data):
-            async def upload(i, b):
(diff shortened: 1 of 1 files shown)


hail/python/test/hailtop/test_aiogoogle.py
```diff
@@ -73,8 +77,10 @@ async def test_compose():
 
     async with GoogleStorageClient() as client:
         for i, b in enumerate(part_data):
-            async with await client.insert_object(bucket, f'{token}/{i}') as f:
-                await f.write(b)
+            async def upload(i, b):
+                async with await client.insert_object(bucket, f'{token}/{i}') as f:
+                    await f.write(b)
+            await retry_transient_errors(upload, i, b)
```
daniel-goldstein, line 83: Bit of a nit but this function either doesn't need inputs or should be defined outside the for loop. My guess is you intended to define this outside the loop.

daniel-goldstein (changes requested):

hail/python/test/hailtop/test_aiogoogle.py
danking: I can't elide the inputs because Python closures don't capture values, they capture references to the variables. I moved outside the for loop.

daniel-goldstein (approved):

PR #10951, opened 2021-10-06
[devbin] use sh instead of bash for kssh

daniel-goldstein:
our prometheus image is based on busybox


danking:
eck, the busybox shell is annoying under featured. Can you add an environment variable instead, maybe KSSH_SHELL?

```
kubectl -n ${2:-default} exec -it "$(kfind1 "$1" "$2")" ${3:+--container="$3"} -- ${KSSH_SHELL:-/bin/sh}
```
That way I can override it by default to give me a good shell.

PR #10960, opened 2021-10-08
[build.yaml] Add test_blob_storage_uri for cloud-agnostic test storage

daniel-goldstein:
I added a new field to the global config that is gs:// + hail_test_gcs_bucket named test_blob_storage_uri and use that wherever it doesn't matter that the backend be google storage, which is essentially everywhere except for the FS/copy tests, where we specifically want a test gcs bucket.


daniel-goldstein:
This is currently blocked by `hailctl config set batch/bucket` not accepting a full GCS uri and only accepts the bucket

daniel-goldstein:
Replaced by another PR.

Issue #10963, opened 2021-10-08
Cache `reference_entry_fields_to_keep` in the VDS combiner


chrisvittal:
Running TableHead on thousands of gvcfs is not good. Don't do it.

PR #10964, opened 2021-10-09
[vds/combiner] Various Improvements

chrisvittal:
Notably:

* Add options for `gvcf_info_to_keep` and `gvcf_reference_entry_fields_to_keep` to control those parameters to `transform_gvcf`.
* Limit gvcf merge tasks to 150,000 partitions.
* Auto-detect `gvcf_reference_entry_fields_to_keep` if it is not provided. If any VDS argument is present, use the reference_data entry type, otherwise, compute it from the first gvcf.

### DIFF
--- a/hail/python/test/hail/vds/test_vds.py
+++ b/hail/python/test/hail/vds/test_vds.py
@@ -3,6 +3,7 @@ import pytest
 
 import hail as hl
 from hail.utils import new_temp_file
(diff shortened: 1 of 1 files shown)


chrisvittal:
It has to do with #10963 . Without something to keep track of entry fields that are present in ref blocks, we'll run a 10_000 row table head on every gvcf.

PR #10969, opened 2021-10-12
[query] ExecuteContext is responsible for broadcasts and cleans them up

tpoterba:


tpoterba:
Almost all of the lines in this change are threading through ExecuteContexts through BlockMatrix stuff so that we can call `ctx.broadcast` instead of `HailContext.get.backend.broadcast`.

The important stuff is in `ExecuteContext.scala` and `SparkBackend.scala` (other `BroadcastValue[T]` implementations on Local/Service backends have no-op cleanups).

tpoterba:
closed until I can take another look at how broadcasts are used in the code cache.

PR #10974, opened 2021-10-13
[hailtop] restore aiogoogle

danking:
Restore backwards compatibility of aiogoogle.

With depreciation warning:
```
(base) # python3 -c 'from hailtop.aiogoogle import GoogleStorageAsyncFS'                        
-c:1: DeprecationWarning: importing hailtop.aiogoogle is deprecated, please use hailtop.aiocloud.aiogoogle
```


danking:
dammit pylint

PR #10985, opened 2021-10-18
[batch] eliminate schedule_job and mark_job_started deadlocks

danking:
Under ideal conditions, I observed this code complete 12 jobs per second in a 1000 job
batch. However, it's usually closer to 6 jobs per second. There's still a deadlock in
`mark_job_complete` which hamstrings performance. I haven't thought carefully about why my metric
(jobs per second) is much lower than our usual metric (MJC per second). I will eventually squash
the `mark_job_complete` deadlock, but I haven't the time currently.

The two main issues were the `batches` field `cancelled` and the `instances` field
`free_cores_mcpu`. We frequently locked the entire batch row just to prevent a batch from being
cancelled while other work was done. This effectively serialized all operation on jobs in the same
batch. We had the dual issue with free_cores_mcpu: we lock the entire instance record everywhere to
prevent the instance changing state while we update job information. This serialized updates to
free_cores_mcpu which serialized job operation for any job on the same instance.

We also had some unnecessarily lock-heavy code inside `add_attempt`. This block of code is trying to
insert a new attempt record and update the associated instance's free cores.

I also fixed some out of date values and bad syntax in `estimated-current.sql`.


batch/batch/front_end/front_end.py
```diff
@@ -1038,21 +1040,31 @@ async def create_batch(request, userdata):
...
+        if bp is None:
             raise web.HTTPForbidden(reason=f'Unknown Hail Batch billing project {billing_project}.')
-        assert billing_projects[0]['status'] is not None
-        if billing_projects[0]['status'] in {'closed', 'deleted'}:
+        if bp['status'] in {'closed', 'deleted'}:
             raise web.HTTPForbidden(reason=f'Billing project {billing_project} is closed or deleted.')
 
-        bp = billing_projects[0]
+        cost = await tx.execute_and_fetchone('''
+SELECT COALESCE(SUM(`usage` * rate), 0) AS cost
```
jigold, line 1057: This logic is not quite the same as what I have in `query_billing_projects`. This code ignores that the billing for really old batches is in the `msec_mcpu` field.

jigold (changes requested):
One last thing.

batch/batch/front_end/front_end.py
danking: Does this mean _get_batch currently has a bug?

jigold: No. Because `batches.msec_mcpu` is passed through to `batch_record_to_dict` which does the right thing:

```
    cost = format_version.cost(record['msec_mcpu'], record['cost'])
    d['cost'] = cost
```

danking: ah, never mind I see how that avoids it.

danking: Ugh. OK, I need to eliminate msec_mcpu. I fear reading all those others tables will just restore a bunch of deadlocks.

danking: Yeah, if we touch every other batch in the billing project, we effectively serialize batch creation and stall all jobs trying to complete.

danking:
closing while I address the msec_mcpu issue

danking:
I was wrong about needing to eliminate msec_mcpu entirely. I think I address your concern directly.

build.yaml
```diff
@@ -475,7 +475,7 @@ steps:
       set -ex
       export NAMESPACE={{ default_ns.name }}
       export CLOUD={{ global.cloud }}
-      python3 /io/create_initial_account.py {{ code.username }} {{ code.login_id }}
```
jigold, line 478: Can you clean up this cruft?

jigold (changes requested):
Is there anything other than the accrued cost that I need to look at carefully? I feel like we did a bunch of passes over this PR already and I can't see the diff from what you had in 2021.

build.yaml
danking: done

danking:
I don't think so. I rebased and then fixed the msec_mcpu issue where your comment was.

jigold (approved):
I put the WIP tag on. Please wait to merge this until a point where you have time to do the migration since we're at home now and have time to handle any bugs that may occur. Also, the migration might take a bit since it's O(n_batches) and O(n_instances).

daniel-goldstein:
Shockingly no merge conflicts but looks like formatting check fails now. @danking

danking:
all builds are failing because the repository the hosts the shadow jar plugin is down

PR #10986, opened 2021-10-18
[auth] Support Azure in Auth

jigold:
- Added the AzureGraphClient which creates applications and service principals
- Modified Auth to create service principals and delete them
- Added two fields to the auth database that optionally store the application ID and the credentials secret name
- Had to modify AzureCredentials a bit to account for a different scope (one of the Azure Credentials types cannot take multiple scopes for some reason)
- There's an auth database migration here!
- I tried to figure out what API calls result in the same result in the portal. It's possible the exact calls are not quite right (ex: addPassword on the application versus the service principal)

TODO:
- Figure out how to use the global config in auth/Makefile to template global.cloud
- Double check the service principal creation is correct (I know the appID and password end up working to create resources at least)

cc: @danking

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -796,21 +796,11 @@ WHERE user = %s AND id = %s AND NOT deleted;
 
                 spec['secrets'] = secrets
 
-                # backwards compatibility
                 secrets.append(
                     {
                         'namespace': DEFAULT_NAMESPACE,
                         'name': userdata['hail_credentials_secret_name'],
                         'mount_path': '/gsa-key',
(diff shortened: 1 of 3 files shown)


jigold:
This is ready for a second look. I basically tried to do the minimum to change nomenclature for services that aren't auth. I had to put in some backwards compatibility in `batch/job.py` because there could be batches with the old format for userdata in the database with `gsa_key_secret_name` instead of `hail_credentials_secret_name`. I also kept the old field in `hail/python/hailtop/hailctl/auth/user.py`. I'm not sure exactly what depends on that.

jigold:
@danking back to you.

PR #10989, opened 2021-10-19
[query] add functionality to `hardy_weinberg_test` to perform one-sided test of excess heterozygosity

pwc2:
Adds functionality to allow `hardy_weinberg_test` to return p-value from one-sided exact test of excess heterozygosity by passing `one_sided=True`. By default, `one_sided=False` and the p-value returned is from the two-sided exact test of HWE.


tpoterba:
nice!

PR #10994, opened 2021-10-19
[query] Speed up struct decoding

johnc1231:
I'm trying to stop having us call `Region.loadBit` everywhere in `EBaseStruct.decode`. First step of that is not calling `setFieldMissing` and `setFieldPresent` everywhere. PRing for tests right now on first round of doing this, there are still some calls that need to be removed.


johnc1231:
Closing for now

PR #11003, opened 2021-10-21
[query] add local whitening method

patrick-schultz:
~Stacked on #11045~

Implements a method to orthogonalize each row of (an entry field of) a matrixtable against all previous rows within a specified window size. The method has time complexity linear in the window size, and is blocked to arrange for most work to occur inside BLAS level 3 subroutines.


danking:
@patrick-schultz Are there any blockers to getting this PR in? It looks like it just needs a review and a merge.

I'll close in the meantime to alleviate some CI issues.

johnc1231:
I'll be reviewing this shortly.

johnc1231 (changes requested):
Haven't dug into the math yet, but looked at most of the scaffolding. Here's some initial comments.

danking:
@tpoterba, I assigned to you in light of John C's departure.

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArraySlice.scala
```diff
@@ -82,6 +82,30 @@ class SNDArraySliceValue(
...
 
+  // FIXME: only optimized for column major
+  override def setToZero(cb: EmitCodeBuilder): Unit = {
+    val eltType = pt.elementType.asInstanceOf[PNumeric with PPrimitive]
+
+    val contiguousDims = contiguousDimensions(cb)
+
+    def recur(startPtr: Value[Long], dim: Int): Unit =
+      if (dim > 0) {
+        cb.ifx(contiguousDims.ceq(dim), {
```
tpoterba, line 93: It seems like making this a check inside each loop is wrong -- we'll evaluate this check many times. Can we evaluate continuousDims per dim from 0 to nDims until we get a False, and put those in an array or something, then look that up?

I'm fine with this being optimized for column-major, but it's *really* bad for not-column-major.

tpoterba (changes requested):

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
```diff
@@ -782,3 +926,305 @@ trait SNDArrayCode extends SCode {
 
   def memoize(cb: EmitCodeBuilder, name: String): SNDArrayValue
 }
+
+class LocalWhitening(cb: EmitCodeBuilder, vecSize: SizeValue, _w: Value[Long], chunksize: Value[Long], _blocksize: Value[Long], region: Value[Region]) {
```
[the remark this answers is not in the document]
patrick-schultz: I think you were looking at an old version somehow.

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArraySlice.scala
patrick-schultz: Good catch, though I don't see how building an array of bools helps, you still have the branches in the inner loops.

Only option I see is to lift the branches to the top level at the cost of a bit of code growth. See what you think.

danking:
It seems like this is realistically delayed until after the native compiler sprint. Is that a fair assessment, @patrick-schultz ?

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -1166,6 +1167,62 @@ object EmitStream {
...
+                state.initializeWindow(cb, prevWindow.asNDArray)
+              })
+              val block = row.loadField(cb, newChunkName).get(cb, "StreamWhiten: missing chunk").asNDArray
+              val whitenedBlock = LinalgCodeUtils.checkColMajorAndCopyIfNeeded(block, cb, elementRegion)
+              state.whitenBlock(cb, whitenedBlock)
+              // the 'newChunkName' field of 'row' is mutated in place and given
+              // to the consumer
+              val result = row.insert(cb, elementRegion, eltType.virtualType.asInstanceOf[TStruct], newChunkName -> EmitValue.present(whitenedBlock))
+              resultField = mb.newPField("StreamWhiten_result", result.st)
+              cb.assign(resultField, result)
```
patrick-schultz, line 1208: This feels pretty hacky, would welcome ideas on better ways to do this.

In the old version, this just mutated the incoming block in place. But for some reason, after merging with current main, the incoming block is row major (at least in my tests), forcing us to copy it. I was then having trouble getting the types right for the result after inserting the modified field, which is why I wait to define the result field until after I have the svalue to be stored.

tpoterba: that seems fine. We do that elsewhere.

tpoterba (approved):

danking:
@patrick-schultz I resolve the conflict by keeping main which seems to have encompassed your change but with a more descriptive error message.

danking:
@patrick-schultz bump

danking:
HUZZAH

PR #11004, opened 2021-10-22
[compiler] add simple copy propagation

patrick-schultz:
I added this when I was hitting class too large exceptions in a previous PR, to see if it would help. I didn't end up using it there, but it still seems like a good optimization. @tpoterba I haven't thought about this very hard, would be interested in your thoughts.


danking:
@patrick-schultz , I'm gonna close since this is from October 2021, but if you address the merge conflict and open a new PR that's great!

PR #11016, opened 2021-10-26
[hailctl] Configure Hail/Spark to use proper memory allocations

tpoterba:


johnc1231 (dismissed):

tpoterba:
OK, this is working and ready for review. I tested manually that on a variety of node types, we both (1) get the expected number of containers (all the cores in the cluster are used) and (2) we get the right OOM error instead of container crashing.

Your comment is addressed -- we always call increment before actually allocating, so we won't exceed the threshold.

hail/python/hailtop/hailctl/dataproc/start.py
```diff
@@ -326,6 +335,45 @@ def disk_size(size):
...
+        worker_memory = MACHINE_MEM[args.worker_machine_type]
+
+        # A Google support engineer recommended the strategy of passing the YARN
+        # config params, and the default value of 95% of machine memory to give to YARN.
+        # yarn.nodemanager.resource.memory-mb - total memory per machine
+        # yarn.scheduler.maximum-allocation-mb - max memory to allocate to each container
+        available_memory_fraction = args.yarn_memory_fraction
+        available_memory_mb = int(worker_memory * available_memory_fraction * 1024)
+        cores_per_machine = int(args.worker_machine_type.split('-')[-1])
+        executor_cores = min(cores_per_machine, 4)
```
johnc1231, line 349: Confused about the min here. Is it the case that when using larger worker machines, we assign more executors per machine and always end up with 4 cores per executor?

tpoterba: that's correct, yes. Larger executors than 4 cores tend to slow things down because there's contention for shared resources (broadcasts, static object stuff, etc).

This keeps the same semantics that we've had, but makes it explicit.

johnc1231 (approved):
Great change, should make our lives much easier.

PR #11020, opened 2021-10-29
[iam] Add rotation status and colors to key rotation info

daniel-goldstein:
Some additional asserts and quality of life messages that I found necessary for confidently rotating keys.

### DIFF
--- a/devbin/rotate_keys.py
+++ b/devbin/rotate_keys.py
@@ -1,4 +1,3 @@
-import aiohttp
 import asyncio
 import base64
 import itertools
@@ -6,7 +5,8 @@ from dateutil.parser import isoparse
 from datetime import datetime, timedelta
 import pytz
 import json
-from typing import List, Optional, Tuple, Generator
+from typing import List, Tuple, Generator, Optional
+from typing_extensions import Literal
 import warnings
 import sys
 import kubernetes_asyncio as kube
@@ -14,6 +14,15 @@ from hailtop.aiocloud.aiogoogle import GoogleIAmClient
 from hailtop.utils import retry_transient_errors
 
 warnings.simplefilter('always', UserWarning)
+
+RotationState = Literal['up-to-date', 'in-progress', 'ready-for-delete', 'expired']
+
+OK_BLUE = '\033[94m'
+OK_GREEN = '\033[92m'
+WARNING_YELLOW = '\033[93m'
+ERROR_RED = '\033[91m'
+GREY = '\033[37m'
+ENDC = '\033[0m'
 
 
 class GSAKeySecret:
@@ -27,6 +36,9 @@ class GSAKeySecret:
 
     def private_key_id(self):
         return self.key_data['private_key_id']
+
+    def matches_iam_key(self, k: 'IAMKey'):
+        return self.private_key_id() == k.id
 
     def __str__(self):
         return f'{self.name} ({self.namespace})'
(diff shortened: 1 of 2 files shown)


devbin/rotate_keys.py
```diff
@@ -87,15 +109,73 @@ def add_new_key(self, k: IAMKey):
...
+            assert kube_key is not None
+            assert kube_key.user_managed
+            return kube_key
+        # We assume that if the key is not used in kubernetes, the most recently
+        # created user-managed key should be considered "active".
+        return next((k for k in self.keys if k.user_managed), None)
+
+    def redundant_user_keys(self) -> List[IAMKey]:
+        active_key = self.active_user_key()
+        return list(k for k in self.keys if k is not active_key and k.user_managed)
```
danking, line 174: just use `[k for k ...]` no list.

devbin/rotate_keys.py
```diff
@@ -87,15 +109,73 @@ def add_new_key(self, k: IAMKey):
         self.keys.insert(0, k)
 
     def list_keys(self) -> str:
```
danking, line 111: another not particularly important performance note: I'd probably write a function like this to take a `io.TextIOBase` and return None. You can pass StringIO if you want a string back, but you could also just pass the stream of interest.

devbin/rotate_keys.py
```diff
@@ -87,15 +109,73 @@ def add_new_key(self, k: IAMKey):
...
+            return 'expired'
+        elif active_key.recently_created() and len(old_user_keys) > 0:
+            return 'in-progress'
+        elif not active_key.recently_created() and len(old_user_keys) > 0:
+            return 'ready-for-delete'
+        else:
+            assert len(old_user_keys) == 0
+            return 'up-to-date'
+
+    def rotation_status(self) -> str:
```
danking, line 150: I'm not a fan of the words status and state meaning almost the same thing. I'd probably have called this `rotation_state_str`

danking: Alternatively, RotationState could be an enum with a custom str.

danking (changes requested):
A few comments. The only one that absolutely needs addressing is the bit about user_managed and expired.

devbin/rotate_keys.py
```diff
@@ -70,7 +82,17 @@ def __init__(self, key_json):
         self.created_readable = key_json['validAfterTime']
         self.expiration_readable = key_json['validBeforeTime']
         self.created = isoparse(self.created_readable)
-        self.expiration = isoparse(self.expiration_readable)
+        assert key_json['keyType'] in ('USER_MANAGED', 'SYSTEM_MANAGED')
+        self.user_managed = key_json['keyType'] == 'USER_MANAGED'
+
+    def expired(self) -> bool:
+        return self.user_managed and self.older_than(90)
```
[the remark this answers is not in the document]
daniel-goldstein: That's fine. I originally had `self.user_managed` because I didn't want a system-managed key to show up as red, but looks like that was before I made system-managed keys greyed-out. Now looking through I don't call `expired` on any system-managed keys so shouldn't change anything.

devbin/rotate_keys.py
daniel-goldstein: I'll just make this change it feels cleaner.

danking (approved):

PR #11023, opened 2021-10-29
[query] Fix bug in call decoding, add error ID to lgt_to_gt

tpoterba:


hail/src/main/scala/is/hail/variant/Call.scala
```diff
@@ -43,6 +43,19 @@ object Call2 {
 
     Call(alleleRepr, phased, ploidy = 2)
   }
+
+  def withErrorID(aj: Int, ak: Int, phased: Boolean, errorID: Int): Call = {
+    if (aj < 0 || ak < 0)
+      fatal(s"allele indices must be >= 0. Found j=$aj and k=$ak.")
```
patrick-schultz, line 49: Should this be tagged with the errorID too?

tpoterba: yep, good catch

johnc1231:
Thanks for catching this, sorry about that.

johnc1231:
@tpoterba failing the test you wrote

tpoterba:
yep, will fix when done with docs

tpoterba:
Commented out the phased test -- this was meant to fail, and I didn't intend to fix that in this PR (it's hard, we need to change the Call representation...)

tpoterba:
Addressed Patrick's comment.

patrick-schultz (approved):

PR #11025, opened 2021-10-29
[ci] Use storage uri for ci module

daniel-goldstein:
Another small effort toward getting `gs://` out of our code. I'll follow up with a PR that puts the prod CI bucket into terraform. The `test_storage_uri` field of the global config was introduced in #11014

### DIFF
--- a/ci/ci/ci.py
+++ b/ci/ci/ci.py
@@ -144,8 +144,9 @@ async def get_pr(request, userdata):  # pylint: disable=unused-argument
 
 
 def storage_uri_to_url(uri: str) -> str:
-    assert uri.startswith('gs://')
-    path = uri.split('gs://')[1]
+    protocol = 'gs://'
+    assert uri.startswith(protocol)
+    path = uri[len(protocol) :]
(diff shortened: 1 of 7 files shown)


ci/ci/ci.py
```diff
@@ -141,6 +143,12 @@ async def get_pr(request, userdata):  # pylint: disable=unused-argument
     return await render_template('ci', request, userdata, 'pr.html', page_context)
 
 
+def storage_uri_to_url(uri: str) -> str:
+    assert uri.startswith('gs://')
+    path = uri.split('gs://')[1]
```
danking, line 148: This is a bit of a funky way to do it. What you actually want is [`removeprefix`](https://docs.python.org/3/library/stdtypes.html#str.removeprefix), but that's in 3.9. I prefer something a bit more explicit:
```python3
protocol = 'gs://'
assert uri.startswith(protocol)
path = uri[len(protocol):]
```

danking (dismissed):

danking (approved):

PR #11027, opened 2021-11-01
[infra] Add batch worker network security group

jigold:
I didn't test this yet -- do you want me to try the updated docs or should we wait until we redeploy the infrastructure next? Documentation is [here](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/network_security_group).

### DIFF
--- a/infra/azure/main.tf
+++ b/infra/azure/main.tf
@@ -211,6 +211,24 @@ resource "tls_self_signed_cert" "db_client_cert" {
   ]
 }
 
+resource "azurerm_network_security_group" "batch_worker" {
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
You should be able to apply this yourself by going into `infra/azure` and running `./bootstrap.sh run_azure_terraform dgoldste`. The rest of the infrastructure is running.

daniel-goldstein (changes requested):
Can you move this to sit next to the rest of the batch worker resources?

daniel-goldstein (approved):

PR #11029, opened 2021-11-01
test only

danking:


danking:
https://github.com/jigold/hail/compare/batch-cloud-agnostic-rewrite-all...danking:batch-cloud-agnostic-rewrite-all

PR #11030, opened 2021-11-01
[infra] Put azure database in its own subnet

daniel-goldstein:
Since the batch workers will ultimately need to communicate with the database for CI jobs, it seems asymmetrical to put the private database connection in the k8s subnet. I've created a third small subnet just for the private database connection. Traffic across subnets is allowed by default so VMs in k8s and the batch workers should be able to access the database IP. This way we can put more restrictive rules in later (like batch workers should not be able to address VMs in the k8s subnet) without interfering with the database connection.

I've applied this change to the azure instance and verified that auth and the admin pod work with the new IP.

### DIFF
--- a/infra/azure/bootstrap.sh
+++ b/infra/azure/bootstrap.sh
@@ -49,7 +49,6 @@ run_terraform() {
         return
     fi
 
-    # export_terraform_vars
     setup_terraform_backend "$@"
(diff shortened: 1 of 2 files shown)


jigold (approved):
When you're done with everything (i.e. things are stable), can you write a dev doc that explains how you architected the networking and why you made the choices you did?

PR #11033, opened 2021-11-02
[infra] Use the same name for azure and gcp registry push credentials

daniel-goldstein:
#11005 Added support for the BuildImage job in build.py to work with `GOOGLE_APPLICATION_CREDENTIALS` or `AZURE_APPLICATION_CREDENTIALS`, but that doesn't mean that the backing secret behind those environment variables has to be different in each case. Ultimately, we would like the same k8s terraform to generate secrets for both clouds, so this changes both `gcr-push-service-account-key` and `acr-push-credentials` to just be `registry-push-credentials`. I have copied the `gcr-push-service-account-key` into a new secret with the appropriate name in `hail-vdc`.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -44,12 +44,7 @@ steps:
     public: true
     secrets:
       - name: auth-oauth2-client-secret
-      - name: gcr-push-service-account-key
-        clouds:
-          - gcp
-      - name: acr-push-credentials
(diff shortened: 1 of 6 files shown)


jigold:
Put the WIP tag on.

jigold (approved):

PR #11038, opened 2021-11-03
[query] expose `lower_tail` and `log_p` parameters in normal and chi-squared statistical functions

pwc2:
This PR exposes the `lower_tail` and `log_p` parameters in `hl.pnorm`, `hl.qnorm`, `hl.pchisqtail`, and `hl.qchisqtail`, per this [feature request](https://hail.zulipchat.com/#narrow/stream/127634-Feature-Requests/topic/log_p.20option.20for.20pnorm.2Fpchisqtail). 

While I was at it I made a few other changes, including:
- Added `hl.dnorm` and `hl.dchisq` to allow for computation of normal and chi-squared probability densities.
- Exposed `mu` and `sigma` in `hl.dnorm`, `hl.pnorm`, and `hl.qnorm` so users can specify different mean/sd values if they wish. By default these functions will still use `mu=0` and `sigma=1` for standard normal densities/cumulative probabilities/quantiles if not otherwise specified by the user.
- Added `ncp` parameter to `hl.qchisqtail` to allow users to specify a non-centrality parameter when computing quantiles.
- Added tests for all of these functions to `test_expr.py`.

### DIFF
--- a/hail/python/hail/docs/functions/stats.rst
+++ b/hail/python/hail/docs/functions/stats.rst
@@ -8,6 +8,7 @@ Statistical functions
     fisher_exact_test
     contingency_table_test
     dbeta
+    dnorm
     dpois
     hardy_weinberg_test
     binom_test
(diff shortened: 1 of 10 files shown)


tpoterba (approved):
great work Patrick, thank you!

PR #11040, opened 2021-11-04
[aiotools][batch][memory] Fix the secret cache

danking:
Our secret cache fails on ~1 in 10000 jobs. I observed this while running some
large scale tests which will soon become standard PR tests. In anticipation of this,
I fixed the k8s_cache. In particular, note how *everyone* who wins the lock tries to
remove it from the dictionary; however, only *one* task can do that successfully.

The new code avoids locks entirely. It is a bit longer because I eagerly remove
out of date keys when I see them and use a future to notify all waiter simultaneouly.

I also updated memory to use this cache for user credentials.

### DIFF
--- a/batch/batch/driver/k8s_cache.py
+++ b/batch/batch/driver/k8s_cache.py
@@ -32,4 +32,4 @@ class K8sCache:
         return await self.secret_cache.lookup((name, namespace))
 
     async def read_service_account(self, name, namespace):
-        return await self.secret_cache.lookup((name, namespace))
+        return await self.service_account_cache.lookup((name, namespace))
--- a/batch/batch/driver/k8s_cache.py
+++ b/batch/batch/driver/k8s_cache.py
@@ -5,10 +5,13 @@ from hailtop.aiotools.time_limited_max_size_cache import TimeLimitedMaxSizeCache
 from ..batch_configuration import KUBERNETES_TIMEOUT_IN_SECONDS
 
 
+FIVE_SECONDS_NS = 5 * 1000 * 1000 * 1000
+
+
 class K8sCache:
     def __init__(self,
                  client,
-                 refresh_time_ns: int = 5 * 1000 * 1000 * 1000,
+                 refresh_time_ns: int = FIVE_SECONDS_NS,
                  max_size: int = 100):
         self.client = client
         self.secret_cache = TimeLimitedMaxSizeCache(
--- a/batch/batch/driver/k8s_cache.py
+++ b/batch/batch/driver/k8s_cache.py
@@ -6,7 +6,10 @@ from ..batch_configuration import KUBERNETES_TIMEOUT_IN_SECONDS
 
 
 class K8sCache:
-    def __init__(self, client, refresh_time_ns: int, max_size: int = 100):
+    def __init__(self,
+                 client,
+                 refresh_time_ns: int = 5 * 1000 * 1000 * 1000,
+                 max_size: int = 100):
         self.client = client
         self.secret_cache = TimeLimitedMaxSizeCache(
             self._get_secret_from_k8s, refresh_time_ns, max_size)
@@ -25,8 +28,8 @@ class K8sCache:
             *name_and_namespace,
             _request_timeout=KUBERNETES_TIMEOUT_IN_SECONDS)
 
-    async def read_secret(self, name, namespace, timeout):
+    async def read_secret(self, name, namespace):
         return await self.secret_cache.lookup((name, namespace))
 
-    async def read_service_account(self, name, namespace, timeout):
+    async def read_service_account(self, name, namespace):
         return await self.secret_cache.lookup((name, namespace))
--- a/batch/batch/driver/k8s_cache.py
+++ b/batch/batch/driver/k8s_cache.py
@@ -1,74 +1,32 @@
-import time
-import asyncio
-import sortedcontainers
-
+from typing import Tuple
 from hailtop.utils import retry_transient_errors
+
+from hailtop.aiotools.time_limited_max_size_cache import TimeLimitedMaxSizeCache
+from ..batch_configuration import KUBERNETES_TIMEOUT_IN_SECONDS
 
 
 class K8sCache:
-    def __init__(self, client, refresh_time, max_size=100):
+    def __init__(self, client, refresh_time_ns: int, max_size: int = 100):
         self.client = client
-        self.refresh_time = refresh_time
-        self.max_size = max_size
+        self.secret_cache = TimeLimitedMaxSizeCache(
+            self._get_secret_from_k8s, refresh_time_ns, max_size)
+        self.service_account_cache = TimeLimitedMaxSizeCache(
+            self._get_service_account_from_k8s, refresh_time_ns, max_size)
 
-        self.secrets = {}
-        self.secret_ids = sortedcontainers.SortedSet(key=lambda id: self.secrets[id][1])
-        self.secret_locks = {}
+    async def _get_secret_from_k8s(self, name_and_namespace: Tuple[str, str]):
+        return await retry_transient_errors(
+            self.client.read_namespaced_secret,
+            *name_and_namespace,
+            _request_timeout=KUBERNETES_TIMEOUT_IN_SECONDS)
 
-        self.service_accounts = {}
-        self.service_account_ids = sortedcontainers.SortedSet(key=lambda id: self.service_accounts[id][1])
-        self.service_account_locks = {}
+    async def _get_service_account_from_k8s(self, name_and_namespace: Tuple[str, str]):
+        return await retry_transient_errors(
+            self.client.read_namespaced_service_account,
+            *name_and_namespace,
+            _request_timeout=KUBERNETES_TIMEOUT_IN_SECONDS)
 
     async def read_secret(self, name, namespace, timeout):
-        id = (name, namespace)
-
-        lock = self.secret_locks.get(id)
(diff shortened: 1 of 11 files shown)


danking:
I suppose another difference is that, when the cache is too large, I remove the oldest key rather than the lexicographically earliest key. Not an issue at our level of usage, but would certainly be unfair if we were actively evicting secrets.

hail/python/hailtop/aiotools/time_limited_max_size_cache.py
```diff
@@ -0,0 +1,77 @@
...
+                self._futures[k].set_exception(exc)
+                self._futures[k].cancel()  # prevent "unretrieved exception" logs
+                del self._futures[k]
+                raise exc
+
+            expiry_time = time.monotonic_ns() + self.lifetime_ns
+            self._futures[k].set_result(v)
+            self._cache[k] = (v, expiry_time)
+            self._keys_with_expiry.add((k, expiry_time))
+            del self._futures[k]
```
daniel-goldstein, line 72: Could this ever delete a future that another coroutine is `await`ing?

hail/python/test/hailtop/aiotools/test_time_limited_max_size_cache.py
```diff
@@ -0,0 +1,146 @@
...
+            raise ValueError("already loaded secret twice")
+        load_counts += 1
+        return k ** 2
+
+    c = TimeLimitedMaxSizeCache(load_secret_only_twice, one_day_ns, 2)
+    assert await c.lookup(3) == 9
+    assert await c.lookup(10) == 100
+    assert await c.lookup(3) == 9
+    assert await c.lookup(3) == 9
+    assert await c.lookup(10) == 100
```
daniel-goldstein, line 41: can use `with pytest.raises(ValueError)...`

hail/python/test/hailtop/aiotools/test_time_limited_max_size_cache.py
```diff
@@ -0,0 +1,146 @@
...
+async def test_exception_propagates_everywhere():
+    async def boom(_: int):
+        await asyncio.sleep(2)  # seconds
+        raise ValueError('boom')
+
+    c = TimeLimitedMaxSizeCache(boom, one_day_ns, 3)
+    x = asyncio.create_task(c.lookup(0))  # this task will sleep then boom
+    y = asyncio.create_task(c.lookup(0))
+    z = asyncio.create_task(c.lookup(0))
+
```
daniel-goldstein, line 126: can also use the `pytest.raises` context manager

hail/python/test/hailtop/aiotools/test_time_limited_max_size_cache.py
```diff
@@ -0,0 +1,146 @@
...
+    assert await c.lookup(3) == 9
+    assert load_counts == 3
+
+    assert await c.lookup(5) == 25
+    assert load_counts == 4
+
+    assert await c.lookup(10) == 100
+    assert load_counts == 5
+
+    # these two are a bit racy, but should be fine
```
daniel-goldstein, line 76: Perhaps just make the lifetime 4 seconds and sleep for 5 ealier? It shouldn't take _that_ long

memory/memory/memory.py
```diff
@@ -140,10 +129,23 @@ async def on_startup(app):
...
     k8s_client = kube.client.CoreV1Api()
     app['k8s_client'] = k8s_client
+
+    async def get_secret_from_k8s(secret_name: str) -> GoogleStorageAsyncFS:
+        secret = await retry_transient_errors(
+            k8s_client.read_namespaced_secret, secret_name, DEFAULT_NAMESPACE, _request_timeout=5.0
+        )
+        gsa_key = json.loads(base64.b64decode(secret.data['key.json']).decode())
+        credentials = GoogleCredentials.from_credentials_data(gsa_key)
+        return GoogleStorageAsyncFS(credentials=credentials)
```
daniel-goldstein, line 143: Is this constructor expensive? If not, it seems cleaner to just use a `K8sCache` here for the gsa key and construct a `GoogleStorageAsyncFS` using the cached secret in `get_or_add_user`.

memory/memory/memory.py
```diff
@@ -161,11 +163,11 @@ async def on_cleanup(app):
...
-                        for items in app['users'].values():
-                            try:
-                                await items['fs'].close()
-                            except:
-                                pass
+
+                        async def close_fs(fs: GoogleStorageAsyncFS):
+                            await fs.close()
+
+                        await app['user_credentials_cache'].shutdown(close_fs)
```
daniel-goldstein, line 170: Ah, I see we need to close these… in that case maybe my above comment on what to store in the cache is moot. Regardless, there's a problem here where we don't properly clean up evicted cache items. The cache itself would need an optional cleanup function, not just one passed into the `shutdown` function.

memory/memory/memory.py
```diff
@@ -60,33 +61,21 @@ async def write_object(request, userdata):
...
+    return await user_credentials_cache.lookup(userdata['hail_credentials_secret_name'])
 
 
-def make_redis_key(username, filepath):
+def make_redis_key(username: str, filepath: str):
     return f'{ username }_{ filepath }'
 
 
-async def get_file_or_none(app, username, fs: AsyncFS, filepath):
+async def get_file_or_none(app, username: str, fs: AsyncFS, filepath: str):
```
daniel-goldstein, line 78: Can this get a return type?

memory/memory/memory.py
```diff
@@ -60,33 +61,21 @@ async def write_object(request, userdata):
...
-            k8s_client.read_namespaced_secret,
-            userdata['hail_credentials_secret_name'],
-            DEFAULT_NAMESPACE,
-            _request_timeout=5.0,
-        )
-        gsa_key = json.loads(base64.b64decode(hail_identity_secret.data['key.json']).decode())
-        credentials = GoogleCredentials.from_credentials_data(gsa_key)
-        users[username] = {'fs': GoogleStorageAsyncFS(credentials=credentials)}
-    return users[username]
+async def get_or_add_user(app, userdata: dict) -> GoogleStorageAsyncFS:
```
daniel-goldstein, line 69: I think `get_user_fs` is a better name here.

hail/python/test/hailtop/aiotools/test_time_limited_max_size_cache.py
daniel-goldstein: It is a good comment to leave but can you add a bit more noting the lifetime being so short? I missed that and it took me a minute to understand why it's racy

batch/batch/driver/k8s_cache.py
```diff
@@ -1,74 +1,35 @@
...
 
+from hailtop.aiotools.time_limited_max_size_cache import TimeLimitedMaxSizeCache
+from ..batch_configuration import KUBERNETES_TIMEOUT_IN_SECONDS
+
 
 class K8sCache:
-    def __init__(self, client, refresh_time, max_size=100):
+    def __init__(self,
+                 client,
+                 refresh_time_ns: int = 5 * 1000 * 1000 * 1000,
```
daniel-goldstein, line 11: nit: can you put this in a constant called `FIVE_SECONDS_NS`? The math doesn't immediately happen in my head

daniel-goldstein (dismissed):
Love the generic lockless data structure.

hail/python/hailtop/aiotools/time_limited_max_size_cache.py
```diff
@@ -0,0 +1,77 @@
...
+import asyncio
+import sortedcontainers
+
+
+T = TypeVar('T')
+U = TypeVar('U')
+
+
+class TimeLimitedMaxSizeCache(Generic[T, U]):
+    def __init__(self, load: Callable[[T], Awaitable[U]], lifetime_ns: int, max_size: int):
```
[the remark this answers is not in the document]
danking: I like it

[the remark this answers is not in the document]
daniel-goldstein: Ya this is really an `LeastRecentlyAddedAssociativeCache`

memory/memory/memory.py
danking: OK, I threaded memory's client session to all the GoogleStorageAsyncFSes. That avoids the need to close them.

danking (commented):
Great review. I addressed all your comments. A couple things came up:

1. I needed to thread the httpx.ClientSession through the google client/session/etc. so that those classes didn't need to be closed (their only resource is a client session). I centralized the http_session into GoogleCredentials rather than its subclasses as well.
2. I added types to aiogoogle.credentials which encouraged me to introduce a helper class that I think you'll like.
3. I've eliminated the exception based control flow in time_limited_max_size_cache.py. I also eliminated the explicit asyncio.Future in favor of a task. This simplifies the code a fair bit because exception propagation is handled by asyncio. A task/future can be `await` ed many times by many actors. Once the task/future is complete, the `await` will just return immediately with the value. I considered eliminated _cache entirely and using a `Dict[str, asyncio.Future]`, but I suspect that will be slower and the current code is already written.

danking:
I'm going to abandon the changes to memory. The current design is super tangled up and the design of aiogoogle makes it difficult to share one httpx.ClientSession across many instances of GoogleStorageAysncFS, unfortunately.

danking:
@daniel-goldstein I anticipate it will pass the tests now.

daniel-goldstein (approved):

PR #11041, opened 2021-11-04
[services] eliminate unclosed client sesssions from k8s

danking:


daniel-goldstein:
I shelved this while the big batch PR was still up but otherwise looks good. Luckily only one merge conflict!

danking:
OK, rebased. Minor procedural point for next time, if you request changes, I'll be notified that I need to take action.

daniel-goldstein (approved):

PR #11044, opened 2021-11-05
[batch] introduce create-fast

danking:
This change allows batchces which fit in one bunch to be created in one
request instead of three. I found this saved a few hundred milliseconds
for batches with 1 job. This path is already well tested because most
of our test batches fit in one job.

To be clear: all the savings here is from avoiding two round-trips to front-end.

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -726,6 +726,7 @@ WHERE user = %s AND id = %s AND NOT deleted;
                 if machine_type and ('cpu' in resources or 'memory' in resources):
                     raise web.HTTPBadRequest(reason='cannot specify cpu and memory with machine_type')
 
+                req_memory_bytes: Optional[int]
                 if machine_type is None:
                     if 'cpu' not in resources:
                         resources['cpu'] = BATCH_JOB_DEFAULT_CPU
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1007,7 +1007,7 @@ VALUES (%s, %s, %s);
                         f'jobs_args={json.dumps(jobs_args)}'
                         f'job_parents_args={json.dumps(job_parents_args)}'
                     ) from err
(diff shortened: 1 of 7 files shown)


danking:
Ah, rebase fail, should be fixed now.

danking:
bump

batch/batch/front_end/front_end.py
```diff
@@ -888,122 +889,139 @@ async def create_jobs(request, userdata):
...
-                        ),
-                    )
-                    await tx.execute_update(
+                        batches_inst_coll_staging_args)
+
+                    batch_inst_coll_cancellable_resources_args = [
+                        (batch_id,
+                         inst_coll,
+                         rand_token,
+                         resources['n_ready_cancellable_jobs'],
```
jigold, line 961: I think you're missing a value here. I count 5 and there should be 6.

jigold (dismissed):

batch/batch/front_end/front_end.py
danking: The diff is a bit confusing. Each list is associated with the subsequent SQL, so this is associated with the `%s`'s on line 967.

batch/test/test_accounts.py
```diff
@@ -209,7 +209,7 @@ async def test_close_billing_project_with_open_batch_errors(
     project = new_billing_project
     await dev_client.add_user("test", project)
     client = await make_client(project)
-    b = await client.create_batch()._create()
+    b = await client.create_batch()._open_batch()
```
jigold, line 212: I'm concerned that we've taken away the tests for the code where we use the slower code path. Can we add a test that makes the max_bunch_size 1 and submit a batch with 2 or 3 jobs?

jigold (dismissed):

danking:
Rebased.

jigold:
Putting the WIP tag on this so the Azure PRs can get priority merging.

jigold (approved):

PR #11045, opened 2021-11-05
[query] refactor pca, add spectral moments estimators

patrick-schultz:
Refactors `_blanczos_pca` into reusable and composable pieces, and uses those pieces to implement spectral moments estimators.

# Krylov factorization
The core iteration of `_blanczos_pca` is factored out into `_krylov_factorization`. `_krylov_factorization(A, V0, p)` takes a matrix `A` (represented as a table of ndarrays) and a starting block `V0` (a local ndarray), and computes matrices (for now local) `U`, `R`, and `V`, such that:
* `U` and `V` are orthonormal matrices (i.e. `U'U =  V'V = I`)
* the columns of `V` are a basis for the block Krylov subspace `K_p(A'A, V_0)`, where `K_p(X, Y) = span(Y, XY, ... X^pY)`
* `UR = AV`, and hence `U` is a basis for the block Krylov subspace `A K_p(A'A, V_0) = K_p(AA', AV_0)`
* `V` is an extension of `V_0`, i.e. `V = hcat(V_0, ...)`
* `R` is upper triangular

# Reduced SVD
From a Krylov factorization, a reduced SVD can be easily computed: If `R = U_1 S V'_1` is a full SVD of `R` (which is small and easily computable), then `(U U_1[:, :k]) S[:k, :k] (V V_1[:, :k])'` is a reduced SVD of `A`. This is implemented in `KrylovFactorization.reduced_svd`.

# Spectral moments
We can also easily compute estimates of spectral moments, i.e. moments of the set of all eigenvalues of `A'A`. The estimator exploits the following key facts:
* If `v` is a random vector of independent entries with mean 0, std. dev. 1 (equivalently `E(v) = 0`, `E(vv') = I`), then `E(v'Xv) = tr(X)`
* `tr(X)` equals the sum of the eigenvalues of `X`, `∑_i 𝜆_i`. More generally, if `f` is any matrix function, `tr(f(X)) = ∑_i f(𝜆_i)`.
* If `w` is a unit-norm vector, and `UR = AV` is the factorization `_krylov_factorization(A, w, p)`, then `w' f(A'A) w` is well-approximated by `w' f(VV'A'AVV') w = w'V f(R'U'UR) V'w = w'V f(R'R) V'w`, and is exact if `f` is a degree `2p+1` polynomial. Moreover, since `w` is the first column of `V`, i.e. `w = Ve_1`, the above further simplifies `w'V f(R'R) V'w = e'_1 f(R'R) e_1`. Finally, if `R = U_1 S V'_1` is an SVD, this reduces to `e'_1 f(R'R) e_1 = e'_1 f(V_1 S^2 V'_1) e_1 = e'_1 V_1 f(S^2) V'_1 e_1 = V_1[0, :] f(S^2) V_1[0, :]'`.
* The previous bullet has a block generalization. If `W` is an orthonormal matrix with `k` columns, and `UR = AV` is the factorization `_krylov_factorization(A, W, p)`, and `R = U_1 S V'_1` is an SVD, then `W' f(A'A) W` is well-approximated by `V_1[:k, :] f(S^2) V_1[:k, :]'`, and this is exact if `f` is a degree-p polynomial.

We can combine the above into an estimator for the p-th spectral moment, `𝜇_p = ∑_i 𝜆_i^p`. We get an unbiased estimator by generating a random vector `v`, and using `E(v' (A'A)^p v) = tr((A'A)^p) = 𝜇_p`. The estimator can be computed exactly using the Krylov factorization as above, i.e. `v' (A'A)^p v = V_1[0, :] S^{2p} V_1[0, :]'`. But this estimator has large variance, so we can just average over many independent estimators. We combine `k` random vectors into a random matrix `V_0`, compute `_krylov_factorization(A, V_0, p)`, and then `V'_0 (A'A)^p V_0 = V_1[:k, :] f(S^2) V_1[:k, :]'`, from which we can average the individual estimates.

This spectral moments estimator is implemented in `KrylovFactorization.spectral_moments`.

# SVD and moments
Finally, in practice we may need to average over too many estimators to get the desired accuracy, so we use one more trick to bring down the variance. This depends on the following property:
* If `Q` is an orthonormal matrix, then `tr(X) = tr(QQ'XQQ') + tr((I-QQ')X(I-QQ'))`, i.e. the trace decomposes into a sum of the traces on the projection onto the subspace spanned by `Q`, and the projection onto its complement. To see this, use the additivity and cyclicity of the trace (`tr(X+Y) = tr(X) + tr(Y)`, `tr(XY) = tr(YX)`). In particular, cyclicity implies `tr(QQ'XQQ') = tr(QQ'X) = tr(XQQ')`.

Now we do two passes. In the first pass, we use some method (such as a Krylov factorization) to compute an orthonormal basis `V` for a subspace which approximates the dominant subspace. Even if this is not a very accurate approximation, the result is that `tr(QQ' X^p QQ')` is much larger than the complement `tr((I-QQ')X^p(I-QQ'))`. But the former we can compute exactly, and all the error comes from our estimate of the much smaller piece.

This variance reduction trick is implemented in `_pca_and_moments`, which computes two Krylov factorizations. The first is used to compute the principal components, and the `V` factor is also used to decompose the moments estimator as above. The second factorization is used to estimate the remaining component of the moments estimator.


johnc1231 (changes requested):

hail/python/hail/methods/pca.py
```diff
@@ -207,13 +207,251 @@ def pca(entry_expr, k=10, compute_loadings=False) -> Tuple[List[float], Table, T
     return hl.eval(g.eigenvalues), scores, None if t is None else t.drop('eigenvalues', 'scores')
 
 
-@typecheck(entry_expr=expr_float64,
+class TallSkinnyMatrix:
+    def __init__(self, block_table, block_expr, source_table, col_key):
+        self.col_key = col_key
+        first_block = block_expr.take(1)[0]
+        self.nrows = first_block.shape[0]
```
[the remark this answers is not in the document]
patrick-schultz: You're right, not sure what my thinking was there. It isn't used, I'll just delete.

hail/python/hail/methods/pca.py
```diff
@@ -207,13 +207,251 @@ def pca(entry_expr, k=10, compute_loadings=False) -> Tuple[List[float], Table, T
...
+    compute_U
+
+    Returns
+    -------
+
+    """
+    t = A.block_table
+    A_expr = A.block_expr
+
+    # FIXME: compute ncols of V0, use to compute final ncols
```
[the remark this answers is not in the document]
patrick-schultz: Oops, already did it and forgot to delete.

hail/python/test/hail/methods/test_pca.py
```diff
@@ -166,48 +162,156 @@ def spectra_helper(spec_func):
...
+        min_dim = min(m, n)
+        sigma = np.diag([spec_func(i+1, k) for i in range(min_dim)])
+        seed = 1025
+        np.random.seed(seed)
+        U = np.linalg.qr(np.random.normal(0, 1, (m, min_dim)))[0]
+        V = np.linalg.qr(np.random.normal(0, 1, (n, min_dim)))[0]
+        A = U @ sigma @ V.T
+        mt_A = matrix_table_from_numpy(A)
+
+        moments, stdevs = hl._spectral_moments(_make_tsm(mt_A.ent, 128), 7)
```
johnc1231, line 240: Should we be checking the stddevs too?

hail/python/hail/methods/pca.py
```diff
@@ -207,13 +207,260 @@ def pca(entry_expr, k=10, compute_loadings=False) -> Tuple[List[float], Table, T
...
+    if not isinstance(A, TallSkinnyMatrix):
+        check_entry_indexed('_spectral_moments/entry_expr', A)
+        A = _make_tsm_from_call(A, block_size)
+
+    n = A.ncols
+
+    if p is None:
+        p = min(num_moments // 2, 10)
+
+    # TODO: Breaks when moment_samples > n. But in that case, we should just
```
johnc1231, line 383: Maybe let's add an assert or something for this until it's resolved, or one telling people they ought to just compute eigenvalues directly?

johnc1231 (changes requested):

hail/python/test/hail/methods/test_pca.py
patrick-schultz: I'm not sure how to test them (they're just an estimate for how accurate the moment estimates are), and from hand testing it looks like they might be off anyways. I alerted Alex and Patrick C., so I'll wait for feedback from them on how the error estimates are working.

hail/python/hail/methods/pca.py
```diff
@@ -207,13 +207,249 @@ def pca(entry_expr, k=10, compute_loadings=False) -> Tuple[List[float], Table, T
...
+        mt = mt.unfilter_entries()
+
+        mt = mt.select_entries(__x=hl.or_else(mt.__gt - mt.__mean_gt, 0.0))
+
+        if hwe_normalize:
+            mt = mt.annotate_rows(
+                __hwe_scaled_std_dev=hl.sqrt(mt.__mean_gt * (2 - mt.__mean_gt) * n_variants / 2))
+            mt = mt.select_entries(__x=mt.__x / mt.__hwe_scaled_std_dev)
+    else:
+        mt = mt.select_entries(__x=mt.__gt)
```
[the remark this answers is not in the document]
patrick-schultz: One of my goals here was to refactor the `hwe_normalize` method into a more usable form. Right now, it's not much good outside of `hwe_normalized_pca`, because `ht.annotate(x=hwe_normalize(ht.GT))` doesn't work (the expression returned by `hwe_normalize` has a different base table). After this PR, `hwe_normalize` is only used in the Spark pca, so it will go away when that does.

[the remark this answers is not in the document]
johnc1231: That makes sense, disregard my earlier comment then.

hail/python/hail/methods/pca.py
patrick-schultz: Ah, thanks, that's why I left the todo.

hail/python/test/hail/methods/test_pca.py
johnc1231: Are we still waiting for feedback from them, or should I go ahead and do next round of review?

patrick-schultz: Sorry, I meant based on feedback I might need to adjust how stdevs are computed, but I don't think there's any need to hold up this PR for that. This is all experimental, and will only have one user for a while.

johnc1231 (approved):

hail/python/hail/methods/pca.py
```diff
@@ -207,13 +207,261 @@ def pca(entry_expr, k=10, compute_loadings=False) -> Tuple[List[float], Table, T
...
+            U = (self.U @ self.U1[:, :k])._persist()
+
+        if self.V is None:
+            V = None
+        else:
+            V = (self.V @ self.V1t.T[:, :k])._persist()
+
+        return U, S, V
+
+    def spectral_moments(self, num_moments, R):
```
johnc1231, line 281: Actually, one note. I think maybe you ought to return a single struct from this method instead of 2 separate expressions. You're making it very easy to accidentally do double work by forking the IR.

johnc1231 (changes requested):

hail/python/hail/methods/pca.py
johnc1231: In the place you use the method, you actually recombine the results into a struct, presumably for this reason, so why not just return a struct here.

johnc1231 (approved):

PR #11046, opened 2021-11-05
[ubuntu] Update default image from 18.04 to 20.04

daniel-goldstein:
Python 3.6 is end-of-life in just over a month, and since it's the default version on an `ubuntu:18.04` image, it seemed a good time to upgrade across the board to 20.04. I'm not sure if there's a better time to do this or if we should make an announcement to users that the default image is changing. Feel free to let it sit here until we want to make the switch.

After this, the only occurrence of `ubuntu:18.04` in the codebase is in the third-party images that we mirror in our registry (so it will still be there, just not used by default).

cc: @jigold


daniel-goldstein:
Stacked on #11047.

danking:
Looks like a few build steps do not have DOCKER_ROOT_IMAGE set.

danking:
needs rebase when the underlying PR merges.

johnc1231:
Worth noting that this is also going to update JDK from 8 -> 11. That should be fine, as far as I can tell both work now, but we don't run hail tests with JDK 11, we do so with JDK 8. 

Probably we should take some time to drop support for Spark 2, Python 3.6, and Java 8.

johnc1231:
Ah, but Google Dataproc still uses Java 8, and that's the main platform we run our software on. Annoying.

daniel-goldstein:
Could we add a specific test for compatibilty on JDK 8 and move the default forward? We're on a crash course of incompatibility with the current python3.6 / ubuntu:18.04 / Java 8 restrictions. From Oracle's pages it feels like Java 8 will simply never die...

danking:
Upgrading is a good instinct to have. I appreciate this PR.

The story here is long, complex, and mundane. I've tried to include everything below. Python 3.6, (should) only appear in the linting image for python 3.6. The `hail-ubuntu` image explicitly installs Python 3.7. I'm happy to drop linting for Python 3.6 from build.yaml if compilers team is OK with that (ask Tim).

We are already using Ubuntu 20.04 for our services and tests. See below for details, but `hail-ubuntu` is based on 20.04. We explicitly install JDK 8 in the `base` image.

---

This PR doesn't change the hail-ubuntu image which is the basis for nearly all our images. `DOCKER_ROOT_IMAGE` seems to have been introduced [here](https://github.com/hail-is/hail/pull/9660). That's my bad for not flagging in the review that this isn't actually a "root image". `DOCKER_ROOT_IMAGE` is just some Linux image with the standard utils. It was introduced as a distinct global concept when Docker Hub began enforcing rate limits (so we needed all our test images to live in GCR).

If you look at the occurrences of `DOCKER_ROOT_IMAGE` or `docker_root_image` you'll find it's almost exclusively used in the tests except for one occurrence in `build-batch-worker-image-startup-gcp.sh`. We could just remove that line. That line is an attempt to keep a relatively recent version of the ubuntu image cached on the worker VM so that we can save some time when pulling the worker Docker image. In practice, the ubuntu image is extraordinarily tiny and quickly becomes out of date (because we rarely rebuild the VM).

hail-ubuntu uses a timestamped ubuntu 20.04 tag: `ubuntu:focal-20201106`. I did this because we kept getting screwed by new ubuntu images getting released which were incompatible with us. We would only find out later when we changed the hail-ubuntu Dockerfile and triggered a refetch of the latest image at the `ubuntu:18.04` tag which included the breaking changes.

You're correct that this is technical debt of ours. DOCKER_ROOT_IMAGE should really be SOME_LINUX_IMAGE and we shouldn't bother trying to fetch it in the worker image startup script. You're fighting the good fight, but, man, keeping a growing software project clean is hard.

danking (approved):

PR #11049, opened 2021-11-08
[hailtop.batch] Deprecate bucket in favor of remote_tmpdir

jigold:
I did not fix the Hail version of the Service backend. I didn't quite understand where bucket was being used and it's not on the critical path for Azure.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2622,7 +2622,7 @@ steps:
       cpu: '1'
     script: |
       export GOOGLE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
-      hailctl config set batch/bucket {{ global.hail_test_gcs_bucket }}
+      hailctl config set batch/remote_tmpdir gs://{{ global.hail_test_gcs_bucket }}/batch/
       python3 -m pytest --log-cli-level=INFO -s -vv --instafail --durations=50 /io/test/
     timeout: 600
     secrets:
@@ -2679,7 +2679,7 @@ steps:
       export HAIL_TEST_DEV_TOKEN_FILE=/dev-tokens/tokens.json
       export HAIL_TOKEN="{{ token }}"
       export HAIL_SCOPE="{{ scope }}"
-      hailctl config set batch/bucket {{ global.hail_test_gcs_bucket }}
+      hailctl config set batch/remote_tmpdir gs://{{ global.hail_test_gcs_bucket }}/batch/
       python3 -m pytest \
               --log-date-format="%Y-%m-%dT%H:%M:%S" \
               --log-format="%(asctime)s %(levelname)s %(name)s %(filename)s:%(lineno)d:%(funcName)s %(message)s" \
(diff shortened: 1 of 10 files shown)


build.yaml
```diff
@@ -2625,7 +2625,7 @@ steps:
       cpu: '1'
     script: |
       export GOOGLE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
-      hailctl config set batch/bucket {{ global.hail_test_gcs_bucket }}
+      hailctl config set batch/remote_tmpdir gs://{{ global.hail_test_gcs_bucket }}/batch/
```
daniel-goldstein, line 2628: Now that this uses `remote_tmpdir` this can instead use the recently-added `test_storage_uri`:

```
hailctl config set batch/remote_tmpdir {{ global.test_storage_uri }}/batch/
```

hail/python/hailtop/aiocloud/aiogoogle/session.py
```diff
@@ -6,7 +6,11 @@
...
 class GoogleSession(BaseSession):
-    def __init__(self, *, credentials: Optional[GoogleCredentials] = None, params: Optional[Mapping[str, str]] = None, **kwargs):
+    def __init__(self, *, credentials: Optional[GoogleCredentials] = None, credentials_file: Optional[str] = None,
+                 params: Optional[Mapping[str, str]] = None, **kwargs):
         if credentials is None:
-            credentials = GoogleCredentials.default_credentials()
+            if credentials_file:
+                credentials = GoogleCredentials.from_file(credentials_file)
+            else:
+                credentials = GoogleCredentials.default_credentials()
```
danking, line 15: if credentials is not None, assert credentials_file is None.

hail/python/hailtop/batch/backend.py
```diff
@@ -409,29 +399,45 @@ def __init__(self,
...
+                    'The \'remote_tmpdir\' parameter of ServiceBackend must be set. '
+                    'Run `hailctl config set batch/remote_tmpdir REMOTE_TMPDIR`')
             if 'gs://' in bucket:
                 raise ValueError(
                     'The bucket parameter to ServiceBackend() should be a bucket name, not a path. '
                     'Use the remote_tmpdir parameter to specify a path.')
             remote_tmpdir = f'gs://{bucket}/batch'
         else:
-            if not remote_tmpdir.startswith('gs://'):
+            schemes = {'gs', 'hail-az', 's3'}
```
danking, line 429: we shouldn't include hail-az if the router does not include an Azure FS.

hail/python/hailtop/hailctl/config/cli.py
```diff
@@ -1,15 +1,22 @@
...
                           'should be valid Google Bucket identifier, with no gs:// prefix'),
+    ('batch', 'remote_tmpdir'): (lambda x: any([re.fullmatch(fr'^{scheme}://.*', x) is not None for scheme in {'gs', 's3', 'hail-az'}]),
+                                 'should be valid cloud storage URI such as gs://my-bucket/batch-tmp/'),
     ('email',): (lambda x: re.fullmatch(r'.+@.+', x) is not None, 'should be valid email address')
 }
 
+deprecated_paths = {
+    ('batch', 'bucket'): '\'batch/bucket\' has been deprecated. Use \'batch/remote_tmpdir\' instead.'
+}
+
```
danking, line 19: good idea

danking (changes requested):

hail/python/hailtop/aiocloud/aiogoogle/session.py
jigold: I want it to fallback on the default credentials instead.

danking: What I meant is that it's a bug to pass both credentials and credentials_file to GoogleSession (or at least, it should be a bug).

e.g.
```python3
assert credentials is None or credentials_file is None, f'specify only one of credentials or credentials_file: {(credentials, credentials_file)}'
```

danking (changes requested):
x

danking (approved):

PR #11051, opened 2021-11-08
[batch] Use higher subnet range for job subnets

daniel-goldstein:
Currently we use the `10.0.0.0/15` range to assign local IPs to jobs on a worker. This is problematic if other resources in our infrastructure collide with this range. Since these are low address ranges it is quite likely that this could happen, and indeed has happened to me in a dev project where the database was assigned a `10.1.x.x` ip and CI jobs that require database access could not run. I don't have a particularly good reason to pick `10.150.0.0/15` as the new range other than I have not seen anything else use it.

### DIFF
--- a/batch/batch/cloud/gcp/driver/create_instance.py
+++ b/batch/batch/cloud/gcp/driver/create_instance.py
@@ -181,14 +181,14 @@ DOCKER_PREFIX=$(curl -s -H "Metadata-Flavor: Google" "http://metadata.google.int
 
 INTERNAL_GATEWAY_IP=$(curl -s -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/attributes/internal_ip")
 
-# private job network = 172.34.0.0/16
-# public job network = 172.35.0.0/16
+# private job network = 172.20.0.0/16
+# public job network = 172.21.0.0/16
 # [all networks] Rewrite traffic coming from containers to masquerade as the host
-iptables --table nat --append POSTROUTING --source 172.34.0.0/15 --jump MASQUERADE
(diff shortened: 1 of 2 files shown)


danking (approved):

danking:
@daniel-goldstein is it possible to prevent google from allocating IPs in a given range?

daniel-goldstein:
We can [reserve specific IP addresses](https://cloud.google.com/compute/docs/ip-addresses/reserve-static-internal-ip-address) for use later (like we do with internal gateway), but as far as I can tell the model is you assign a network or subnet an IP address range and that should be available to any resources created in it. I guess the proper way to ensure this would be to then use a range for job IPs that is outside the internal IP address range given to our default network.

danking:
It looks to me like none of our subnets use, say, 10.255.0.0/24. Any reason we can't use that subnet for jobs?

daniel-goldstein:
Now that I'm looking at our [subnets](https://console.cloud.google.com/networking/networks/details/default?project=hail-vdc&pageTab=SUBNETS), I'm seeing that this is actually a bad choice. We also are only properly blocking VM communication within the us-central1-a region..

Perhaps we should choose an entirely different private IP range than the 10.0.0.0/8, such as 172.16.0.0

danking:
Or that

daniel-goldstein:
beat me to it. Ya I think I would prefer the latter as it makes it easier to distinguish between job and worker IPs

danking:
Yeah, it seems to me that we should be dropping traffic from the public jobs to anything in 10.0.0.0/8, 172.16.0.0/12, and 192.168.0.0/16.

daniel-goldstein:
Would you like me to add that here or in a followup?

danking:
Follow up

daniel-goldstein:
Oops wait I chose poorly.

PR #11053, opened 2021-11-09
[ci] Add CI infrastructure for testing outside hail-vdc

daniel-goldstein:
This introduces the necessary pieces of infrastructure to run CI in GCP and a couple of small changes such that it can run as a secondary CI. This is currently failing because one of the secrets I introduce here in the terraform does not exist in hail-vdc. If you approve of the approach I can add it in manually. 

##  Terraform changes
This adds a new CI terraform module that adds a CI bucket, sets some permissions for the CI service account and adds some K8s secrets like github tokens and the zulip config. This allows the terraform deployment to optionally include resources needed for CI. This was the best way I could think to introduce this infra with the least changes, but it's not what I want in the long term. Right now we have one monolithic root module that includes all the resources necessary to run batch, with the option for tacking on CI. I would rather extract most of our root module into a `batch` module (and while we're break down the innards into modules like vdc, db, etc.) and have the root module be something that can be easily pieced together from the library of modules. This would be a decently big refactor and more importantly would require existing deployments to manually overhaul their terraform state, so it's something I want to do carefully but also sooner is better than later. Given how terraform state is indexed, I believe more modularity will be easier to manage in the long term.

## CI changes
This adds the following features to CI
- Watched branches can be marked as `mergeable`. `mergeable=true` should be the default behavior and `false` prevents CI from merging a PR on GitHub. This allows multiple CI's to run tests in different environments without stepping on each others' toes. This *does not*, however, consider statuses from multiple CIs when making the decision to merge a PR. That is currently based on the build status, and later should be changed to consider the collection of statuses on GitHub.
- Custom Deploy Steps: This is a collection of build.yaml steps that CI should run on `deploy`. This allows a deployment to only deploy a subset of the infrastructure to default, though it will run the entire test suite. Based on the semantics, `requested_step_names`, the empty list will deploy everything.
- Custom github context: This is just so CIs running in different clouds/environments don't step on each other in the GitHub statuses.

### DIFF
--- a/ci/deployment.yaml
+++ b/ci/deployment.yaml
@@ -45,14 +45,10 @@ spec:
            - name: HAIL_CI_OAUTH_TOKEN
              value: /secrets/oauth-token/oauth-token
            - name: HAIL_CI_GITHUB_CONTEXT
-{% if deploy %}
              valueFrom:
                secretKeyRef:
                  name: ci-config
                  key: github_context
-{% else %}
-             value: "ci-test"
-{% endif %}
            - name: HAIL_WATCHED_BRANCHES
              valueFrom:
                secretKeyRef:
--- a/ci/deployment.yaml
+++ b/ci/deployment.yaml
@@ -45,10 +45,14 @@ spec:
            - name: HAIL_CI_OAUTH_TOKEN
              value: /secrets/oauth-token/oauth-token
            - name: HAIL_CI_GITHUB_CONTEXT
+{% if deploy %}
              valueFrom:
                secretKeyRef:
                  name: ci-config
                  key: github_context
+{% else %}
+             value: "ci-test"
+{% endif %}
            - name: HAIL_WATCHED_BRANCHES
 {% if deploy %}
              valueFrom:
--- a/ci/deployment.yaml
+++ b/ci/deployment.yaml
@@ -54,23 +54,20 @@ spec:
              value: "ci-test"
 {% endif %}
            - name: HAIL_WATCHED_BRANCHES
-{% if deploy %}
              valueFrom:
                secretKeyRef:
                  name: ci-config
                  key: watched_branches
-{% else %}
-             value: '[["hail-ci-test/ci-test-{{create_ci_test_repo.token}}:master", true, true]]'
-{% endif %}
            - name: HAIL_CI_DEPLOY_STEPS
-{% if deploy %}
              valueFrom:
                secretKeyRef:
                  name: ci-config
                  key: deploy_steps
-{% else %}
-             value: '[]'
-{% endif %}
+           - name: HAIL_CI_STORAGE_URI
+             valueFrom:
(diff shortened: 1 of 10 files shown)


daniel-goldstein:
Stacked on #11051

daniel-goldstein:
I'm currently running this branch of CI on a pull request of itself on my own fork of hail, and it nearly passes all tests except for hailtop_batch_* because of requester pays permissions issues and monitoring, because I don't have a service account in my project with all the permissions for broad-ctsa. So unfortunately haven't fully validated that it will _not_ merge a passing PR, but this seemed good enough that we can push it through for azure (since both of these errors are gcp-dependent). If this goes through I can put in a follow-up PR that mirrors the infra resources that CI needs in azure (bucket, permissions, etc.)

danking (changes requested):
I'm roughly OK with this, but can you elaborate on what resources in broad-ctsa are required for the tests to pass? My goal is for there to be ~one command that a user can run which terraforms their GCP project, deploys Batch, and runs the tests, which pass.

daniel-goldstein:
It would need BigQuery access to broad-ctsa, since that's where the data for the monitoring service lives. You can also look at the billing-monitor service account. I'm not sure much beyond that because I don't seem to be able to see broad-ctsa anymore in the console? And I only have IAM permissions in hail-vdc.

danking:
Huh. I added you back to broad-ctsa.

I'm curious to nail down the requester pays issues.

Mmm. The billing monitoring is a bit of a mess. The BQ table should really be in hail-vdc and terraform should be able to create the necessary billing sinks to dump billing data into BQ. I'm not exactly sure what the equivalent tools are in Azure to get billing information.

ci/deployment.yaml
```diff
@@ -44,12 +44,32 @@ spec:
...
+             value: '[["hail-ci-test/ci-test-{{create_ci_test_repo.token}}:master", true, true]]'
+{% endif %}
+           - name: HAIL_CI_DEPLOY_STEPS
+{% if deploy %}
+             valueFrom:
+               secretKeyRef:
+                 name: ci-config
+                 key: deploy_steps
+{% else %}
+             value: '[]'
```
daniel-goldstein, line 72: It would probably be better actually to just have a build.yaml step that creates a `ci-config` in the non-default namespace that looks like:

```
github_context="ci-test"
watched_branches="[["hail-ci-test/ci-test-{{create_ci_test_repo.token}}:master", true, true]]"
deploy_steps="[]"
storage_uri="{{ global.test_storage_uri }}"
```

And then remove all of this logic in the deployment.yaml. Then the CI Makefile wouldn't need all of this repo token business.

jigold:
I'm going to add a clouds scope for RunImage steps today. Maybe that will help for the monitoring situation for now?

danking (approved):

PR #11055, opened 2021-11-09
[batch] Add mvp azure implementation

daniel-goldstein:
Bringing over the azure-specific changes from #10970, this reflects the test batch instance that is currently running in azure.


batch/batch/cloud/azure/resource_utils.py
```diff
@@ -0,0 +1,135 @@
...
+    if match is None:
+        return match
+    return MachineTypeParts.from_dict(match.groupdict())
+
+
+def azure_machine_type_to_worker_type_and_cores(machine_type: str) -> Tuple[str, int]:
+    maybe_machine_type_parts = azure_machine_type_to_parts(machine_type)
+    if maybe_machine_type_parts is None:
+        raise ValueError(f'bad machine_type: {machine_type}')
+    # local_ssd = maybe_machine_type_parts.additive_features == 'd' or maybe_machine_type_parts.family == 'F'
```
jigold, line 84: I ended up not needing to return the local SSD, but it might be good to add it as a return type in order to assert that the machine type has the right option for the local ssd when creating instances, etc. in Azure.

batch/batch/cloud/azure/worker/utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+async def azure_worker_access_token(session: aiohttp.ClientSession) -> Dict[str, str]:
+    global acr_refresh_token, expiration_time
+    if acr_refresh_token is None or time_msecs() >= expiration_time:
+        acr_url = os.environ['DOCKER_PREFIX']
+        assert acr_url.endswith('azurecr.io'), acr_url
+        aad_access_token = await get_aad_access_token(session)
+        acr_refresh_token = await get_acr_refresh_token(session, acr_url, aad_access_token)
+        expiration_time = time_msecs() + 60 * 60 * 1000  # token expires in 3 hours so we refresh after 1 hour
+    return {'username': '00000000-0000-0000-0000-000000000000', 'password': acr_refresh_token}
```
jigold, line 57: I forgot to add back the documentation reference for the username and password: https://docs.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli#az-acr-login-with---expose-token

batch/sql/add_azure_tables.py
```diff
@@ -0,0 +1,66 @@
...
+        scope = os.environ['HAIL_SCOPE']
+
+        await db.just_execute('''
+DELETE FROM `pools`;
+''')
+        await db.just_execute('''
+DELETE FROM `inst_colls`;
+''')
+
+        enable_standing_worker = scope != 'dev'
```
jigold, line 25: This needs to be removed here and used down below where the pools are defined. The `enable_standing_worker` flag should only be used for the 'standard' pool.

batch/sql/add_azure_tables.py
```diff
@@ -0,0 +1,66 @@
...
+        ]
+
+        await db.execute_many('''
+INSERT INTO `inst_colls` (name, is_pool, boot_disk_size_gb, max_instances, max_live_instances, cloud)
+VALUES (%s, %s, %s, %s, %s, %s)
+''',
+                              inst_colls)
+
+        pools = [
+            ('highcpu', 'F', worker_cores, True, 0, enable_standing_worker, 8),
```
jigold, line 51: For the `F` pool, `worker_local_ssd_data_disk` needs to be False and the `worker_external_ssd_data_disk_size_gb` needs to be at least `n_cores * 5 + 30` for the largest value of n_cores between the standing worker and the worker_cores.

hail/python/hailtop/batch/backend.py
```diff
@@ -425,9 +425,11 @@ def __init__(self,
                     'Use the remote_tmpdir parameter to specify a path.')
             remote_tmpdir = f'gs://{bucket}/batch'
         else:
-            if not remote_tmpdir.startswith('gs://'):
+            schemes = {'file', 'gs', 'hail-az', 's3'}
```
jigold, line 428: I think Dan wants us to remove `s3` here for now.

jigold (changes requested):

jigold:
It looks like the tests are close to passing!

jigold:
Also, do we want to add the delete VM privilege here? We can always comment out the `az vm delete` command in create_instance when debugging.

hail/python/hailtop/batch/backend.py
jigold: The RouterAsyncFS definition above needs to either have Azure added or I think it's an item on the to-do list already.

jigold (dismissed):

batch/batch/cloud/azure/driver/create_instance.py
```diff
@@ -0,0 +1,471 @@
...
+
+    make_global_config = ['mkdir /global-config']
+    global_config = get_global_config()
+    for name, value in global_config.items():
+        make_global_config.append(f'echo -n {shq(value)} > /global-config/{name}')
+    make_global_config_str = '\n'.join(make_global_config)
+
+    assert instance_config.is_valid_configuration(resource_rates.keys())
+
+    startup_script = r'''#cloud-config
```
danking, line 79: admittedly somewhat confusing that Azure calls this a startup *script* when its a configuration file. 🤷

batch/batch/worker/disk.py
```diff
@@ -13,6 +13,7 @@ async def __aenter__(self, labels=None):
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         await self.delete()
+        await self.close()
```
danking, line 16: good fix

batch/batch/worker/worker.py
```diff
@@ -124,15 +125,23 @@ def deeper_stack_level_warn(*args, **kwargs):
...
+log.info(f'INTERNET_INTERFACE {INTERNET_INTERFACE}')
 log.info(f'UNRESERVED_WORKER_DATA_DISK_SIZE_GB {UNRESERVED_WORKER_DATA_DISK_SIZE_GB}')
 
 instance_config = instance_config_from_config_dict(INSTANCE_CONFIG)
 assert instance_config.cores == CORES
 assert instance_config.cloud == CLOUD
 
-deploy_config = DeployConfig('gce', NAMESPACE, {})
+if CLOUD == 'gcp':
+    deploy_location = 'gce'
```
danking, line 136: This feels like a rake waiting to slam us in the face.

danking: Yeah, sorry, I missed this earlier. There are only three "deploy locations":

1. Inside a pod inside a Kubernetes cluster
2. Inside a Batch job
3. Everywhere else

The right fix here is to use better names, such as: `k8s`, `batch-job`, and `external`. The short term fix is to just pervasively use the word `gce` and eliminate `azure` as an option for DeployConfig.

danking: The reasoning for this is because there are three (classes of) names for any service inside k8s:

1. service.namespace: this is only accessible inside k8s
2. service.hail: this is a special DNS name used when you're in the cloud but outside k8s (the internal gateway is cheaper & does not expose traffic to the internet, so it's important to distinguish (2) from (3))
3. service.hail.is: this is accessible from the Internet

batch/batch/worker/worker.py
```diff
@@ -195,7 +204,8 @@ async def init(self):
         with open(f'/etc/netns/{self.network_ns_name}/resolv.conf', 'w') as resolv:
             if self.private:
                 resolv.write('nameserver 169.254.169.254\n')
-                resolv.write('search c.hail-vdc.internal google.internal\n')
+                if CLOUD == 'gcp':
+                    resolv.write('search c.hail-vdc.internal google.internal\n')
```
danking, line 208: Are we absolutely certain there is no parallel version of this in Azure?

batch/batch/worker/worker.py
```diff
@@ -318,6 +322,11 @@ async def wrapper(f, *args, **kwargs):
     return wrapper
 
 
+async def pull_docker_image(pool: concurrent.futures.ThreadPoolExecutor, image_ref_str: str, auth: dict):
+    assert docker_sync
+    return await blocking_to_async(pool, docker_sync.images.pull, image_ref_str, auth_config=auth)
```
danking, line 327: I'm confused by this. Why do we wrap a synchronous docker client with blocking_to_async instead of using aiodocker?

danking: If this is unavoidable, I think a cleaner fix is to wrap the aiodocker instance and override image pull to do this.

batch/batch/worker/worker.py
```diff
@@ -553,14 +563,22 @@ async def pull_image(self):
...
-                    docker.images.pull, self.image_ref_str, auth=auth
+                    pull_docker_image, worker.pool, self.image_ref_str, auth=auth
                 )
         except DockerError as e:
             if e.status == 404 and 'pull access denied' in e.message:
                 self.short_error = 'image cannot be pulled'
             elif 'not found: manifest unknown' in e.message:
                 self.short_error = 'image not found'
             raise
+        # errors is just a top level python file and not actually exposed in docker's __init__.py
```
danking, line 574: I don't think we need this comment.

batch/deployment.yaml
```diff
@@ -185,6 +170,10 @@ spec:
          secret:
            optional: false
            secretName: ssl-config-batch-driver
+       - name: ssh-public-key
+         secret:
+           secretName: batch-worker-ssh-public-key
```
danking, line 175: AFAICT, this does not exist in gcp k8s.

gear/gear/clients.py
```diff
@@ -1,6 +1,6 @@
 from typing import Optional
```
danking, line 1: This file also feels like the sort of thing that should be an interface with an instance per cloud and you allocate it once at start up time.

gear/gear/cloud_config.py
```diff
@@ -3,6 +3,20 @@
...
+    resource_group: str
+    region: str
+
+    @staticmethod
+    def from_global_config(global_config):
+        conf = AzureConfig()
+        conf.subscription_id = global_config['azure_subscription_id']
+        conf.resource_group = global_config['azure_resource_group']
+        conf.region = global_config['azure_location']
+        return conf
```
danking, line 17:
```suggestion
class AzureConfig:
    @staticmethod
    def from_global_config(global_config):
        return AzureConfig(global_config['azure_subscription_id'],
                           global_config['azure_resource_group'],
                           global_config['azure_location'])
    
    def __init__(self, subscription_id: str, resource_group: str, region: str):
        self.subscription_id = subscription_id
        self.resource_group = resource_group
        self.region = region
```

gear/gear/cloud_config.py
```diff
@@ -3,6 +3,20 @@
 import json
 
 
+class AzureConfig:
```
danking, line 6: Is it possible to unify this with AzureInstanceEnvironment? I'd prefer not to have lots of subtly different classes if we can help it.

infra/azure/outputs.tf
```diff
@@ -16,6 +16,8 @@ output "global_config" {
       test_storage_uri = "hail-az://${azurerm_storage_account.test.name}/${azurerm_storage_container.test.name}"
     }
     azure = {
+      # FIXME Just for testing
+      gsuite_organization = "broadinstitute.org"
```
danking, line 20: should we eliminate this?

danking (changes requested):
This looks pretty good on the whole. I think we can get it in quickly.

batch/sql/add_azure_tables.py
daniel-goldstein: > largest value of n_cores between the standing worker and the worker_cores.

Not sure I understand this sentence. There's no standing worker for highcpu, should `n_cores` just be `worker_cores`?

batch/batch/worker/worker.py
daniel-goldstein: Looking at `resolv.conf`, I see what looks like
`<random_id>.bx.internal.cloudapp.net`. We can look into figuring out how that's generated.

batch/batch/worker/worker.py
daniel-goldstein: Jackie would know the exact details, but I believe it had to do with aiodocker not always properly handling credentials in a way that messed up the azure credentials, and this was chosen in favor of monkey-patching aiodocker in the interim before we move over to podman.

batch/deployment.yaml
daniel-goldstein: That's why it's marked as optional. It does rely on the good faith that gcp won't try to use it. In this case it would fail by reading an empty mount (though the mount should still exist).

gear/gear/cloud_config.py
daniel-goldstein: I intended these config objects mainly as a way of getting python types on the k8s secret for each cloud. I agree that it is better to have a robust abstraction that can provide functionality, like AzureInstanceEnvironment was turned into, rather than passing a dict around everywhere informing different components how to behave. But after your earlier comments, we've transformed AzureInstanceEnvironment to the AzureWorkerAPI, which seems like it shouldn't be used more widely than the worker.

If we had something else like a AzureDriverAPI, and both API abstractions used the AzureConfig just to initialize themselves, would that be reasonable?

infra/azure/outputs.tf
daniel-goldstein: yes I can remove it. I think you'll have to address this in the AAD oauth work.

gear/gear/cloud_config.py
daniel-goldstein: Overall I think the cloud configs had a bit of a tangled history and I have yet to be convinced of their ratio of utility : danger. I think as long as the pieces of configuration (resource_group, etc.) are hidden behind powerful abstractions like the AzureWorkerAPI or AzureDriver, whether they get them this way or just through env variables is not as important.

daniel-goldstein: Unless you see a clearly better solution, I'm happy to do a separate pass once this is all in to see what's actually useful here.

danking (approved):

PR #11058, opened 2021-11-10
[build.yaml] Add delete azure instances build step

jigold:
Stacked on #11057. I tested the query string and made sure the list commands worked. However, I did not test the actual delete with xargs. This was copied almost verbatim from the gcp delete instances step. I'm pretty sure by not adding the `-x` flag at the top of the script, the password won't be printed to the command line output in the logs, but I'm not 100% sure.

Also, for this to work, the test-gsa-key needs to be able to delete VM and network resources. Do we have this in GCP as well? I guess we must in order for Batch to be able to remove its instances.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -3879,7 +3879,7 @@ steps:
       - test_hailtop_batch_3
       - test_hailtop_batch_4
   - kind: runImage
-    name: delete_batch_instances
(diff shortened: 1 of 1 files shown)


daniel-goldstein (approved):
LGTM. Should be easy enough to verify once we have CI up and running in Azure.

PR #11062, opened 2021-11-12
[compiler] Refactor EncodedLiteral / TableLiteral to use multiple byte arrays

tpoterba:
Support encoded values greater than 2GB.


patrick-schultz (approved):
Thanks!

PR #11059, opened 2021-11-10
[fs] Reorganize the hailtop.aiotools.fs module

jigold:
I was trying to figure out how to implement the lazy router and just couldn't stand the huge hailtop/aiotools/fs.py file anymore. This is just a reorganization. Should be no actual code changes.

Feel free to can this if you think it's unnecessary.

FYI @danking

PR #11063, opened 2021-11-15
[router_fs] Make RouterAsyncFS lazy

jigold:
I made some changes in the second commit that were necessary to avoid the circular imports. Closing #11059 in favor of this PR.

FYI: @danking


hail/python/hailtop/aiotools/fs/copier.py
```diff
@@ -1,446 +1,20 @@
...
-
-        def f():
-            try:
-                shutil.rmtree(path)
-            except FileNotFoundError:
-                # no error if path does not exist
-                pass
-
-        await blocking_to_async(self._thread_pool, f)
+if TYPE_CHECKING:
```
danking, line 15: We should treat `if TYPE_CHECKING` as a code smell. This code can be restructured to avoid the circularity: https://github.com/jigold/hail/compare/lazy-router-pt2...danking:dk-lazy-router-pt2 (I also included the copy part fix)

danking (changes requested):

hail/python/hailtop/aiotools/fs/copier.py
danking: Erm, I didn't mention the copy part fix before. The copier shouldn't directly depend on RouterAsyncFS. We might want to use it to copy between the same file system. We can avoid this by making `copy_part_size` a function that takes the URL. From the URL it determines the scheme and returns an appropriate copy part size.

jigold:
Back to you.

danking (approved):

PR #11065, opened 2021-11-15
[batch] Add Azure GC loops for nics and public ip addresses

jigold:


hail/python/hailtop/aiocloud/aioazure/client/network_client.py
```diff
@@ -5,7 +5,12 @@
 
 
 class AzureNetworkClient(AzureBaseClient):
-    def __init__(self, subscription_id, resource_group_name, session: Optional[AzureSession] = None):
-        session = session or AzureSession()
+    def __init__(self, subscription_id, resource_group_name, session: Optional[AzureSession] = None, **kwargs):
+        if 'params' not in kwargs:
+            kwargs['params'] = {}
+        params = kwargs['params']
```
daniel-goldstein, line 11: I'm a little confused here. Shouldn't params be per request?

hail/python/hailtop/aiocloud/aioazure/client/base_client.py
```diff
@@ -16,15 +16,17 @@ def __init__(self, base_url: str, *, session: Optional[AzureSession] = None,
             session = AzureSession(**kwargs)
         super().__init__(base_url, session, rate_limit=rate_limit)
 
-    async def delete(self, path: str, **kwargs) -> aiohttp.ClientResponse:
-        async with await self._session.delete(
-                f'{self._base_url}{path}', **kwargs) as resp:
+    async def delete(self, path: Optional[str] = None, *, url: Optional[str] = None, **kwargs) -> aiohttp.ClientResponse:
```
daniel-goldstein, line 19: See comment about keeping the url just on `get`.

daniel-goldstein (changes requested):
I think it'll be good to keep an eye on whether all of these requests eat up resources on the driver, but I'm totally fine doing the simpler list and delete unless we know it's a problem. Most of my comments are just about how to consolidate some of the boilerplate.

hail/python/hailtop/aiocloud/common/base_client.py
```diff
@@ -16,24 +16,32 @@ def __init__(self, base_url: str, session: BaseSession, *, rate_limit: RateLimit
...
 
-    async def get(self, path: str, **kwargs) -> Any:
-        async with await self._session.get(
-                f'{self._base_url}{path}', **kwargs) as resp:
+    async def get(self, path: Optional[str] = None, *, url: Optional[str] = None, **kwargs) -> Any:
+        if url is None:
+            assert path
+            url = f'{self._base_url}{path}'
+        async with await self._session.get(url, **kwargs) as resp:
             return await resp.json()
```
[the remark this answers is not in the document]
jigold: I had `get_nextlink` which was essentially the same thing and you and Dan both shot that down. I can move the url to only `get`, but I did like the symmetry across all API calls.

hail/python/hailtop/aiocloud/aioazure/client/network_client.py
jigold: This sets the default params in the Session. You can override them with a request.

hail/python/hailtop/aiocloud/common/base_client.py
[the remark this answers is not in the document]
daniel-goldstein: ah, I don't think I had enough context, I never knew that this was for paging. Sorry I should have asked for more details.

daniel-goldstein (approved):

PR #11071, opened 2021-11-17
[batch] Fixes to get tests to start running on Azure

daniel-goldstein:
Mostly random bugs that didn't get flexed until trying to run CI jobs and batch tests within jobs.
- The IP addresses used for jobs immediately got out of sync with GCP and I needed to add an `internal.hail` entry to the worker and job's `/etc/hosts` so that default batch could submit to dev batch.
- GCP's metadata server and DNS nameserver are both 169.254.169.254. Azure has a separate IP address for the latter, so I added this configuration to the CloudWorkerAPI.

Something that's not addressed here is that I needed to comment out the resource requirements for build image jobs to make them run on standards. The common 2 vCPU / 10 Gi storage / 7.5 Gi Mem lands on standards in GCP but highcpu on azure, which doesn't have disks implemented yet. I'm not sure what the correct step forward on that front is. Otherwise, dev deploying batch should be possible! I ran into multiple issues where my user's sql config was messed up because it was created from a buggy branch. I tried to fix these for the other dev namespaces (dan's which was made later was fine) but there may be some bits I missed.

I got as far as running `test_batch_0` and the tests start (!) but fail quickly because of a blob permission issue on the dev driver.

### DIFF
--- a/batch/batch/cloud/azure/driver/create_instance.py
+++ b/batch/batch/cloud/azure/driver/create_instance.py
@@ -163,25 +163,26 @@ DOCKER_PREFIX=$(jq -r '.docker_prefix' userdata)
 
 INTERNAL_GATEWAY_IP=$(jq -r '.internal_ip' userdata)
 
-# private job network = 10.0.0.0/16
-# public job network = 10.1.0.0/16
+# private job network = 172.20.0.0/16
+# public job network = 172.21.0.0/16
 # [all networks] Rewrite traffic coming from containers to masquerade as the host
-sudo iptables --table nat --append POSTROUTING --source 10.0.0.0/15 --jump MASQUERADE
+iptables --table nat --append POSTROUTING --source 172.20.0.0/15 --jump MASQUERADE
 
 # [public]
 # Block public traffic to the metadata server
-sudo iptables --append FORWARD --source 10.1.0.0/16 --destination 169.254.169.254 --jump DROP
+iptables --append FORWARD --source 172.21.0.0/16 --destination 169.254.169.254 --jump DROP
 # But allow the internal gateway
-sudo iptables --append FORWARD --destination $INTERNAL_GATEWAY_IP --jump ACCEPT
+iptables --append FORWARD --destination $INTERNAL_GATEWAY_IP --jump ACCEPT
 # And this worker
-sudo iptables --append FORWARD --destination $IP_ADDRESS --jump ACCEPT
+iptables --append FORWARD --destination $IP_ADDRESS --jump ACCEPT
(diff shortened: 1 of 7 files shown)


batch/batch/cloud/azure/driver/create_instance.py
```diff
@@ -163,25 +163,26 @@ def create_vm_config(
 
 INTERNAL_GATEWAY_IP=$(jq -r '.internal_ip' userdata)
 
-# private job network = 10.0.0.0/16
-# public job network = 10.1.0.0/16
+# private job network = 172.20.0.0/16
+# public job network = 172.21.0.0/16
```
jigold, line 167: What's up with the 172 change? Was this because we were conflicting with something? Didn't we also use these IP address ranges in some tests in test_batch.py?

jigold:
> The common 2 vCPU / 10 Gi storage / 7.5 Gi Mem lands on standards in GCP but highcpu on azure

I'm a bit worried about this. Shouldn't it still be standard?

jigold (changes requested):

batch/batch/cloud/azure/driver/create_instance.py
daniel-goldstein: We realized that GCP uses the whole [Class A](https://www.ibm.com/docs/en/networkmanager/4.2.0?topic=translation-private-address-ranges) private IP range for nodes on the network, not just 10.128.x.x which turns out to be just our main zone. So we instead chose to switch to Class B to ensure job IPs could not conflict with node IPs. When I spun up my own GCP project for the n'th time, the database got assigned an IP in the old range and jobs couldn't talk to it.

There is a test that uses a 10.128.x.x to ensure that a job cannot communicate with a different worker on the network, but since that's a worker IP not a job IP that shouldn't need to change.

daniel-goldstein:
> > The common 2 vCPU / 10 Gi storage / 7.5 Gi Mem lands on standards in GCP but highcpu on azure
> 
> I'm a bit worried about this. Shouldn't it still be standard?

I thought so too, maybe it's a bug?

daniel-goldstein:
Prior to this we add an entry to /etc/hosts so that jobs can contact the batch front end to submit batches. This entry in default is `batch.hail`, and in a dev/pr namespace is `internal.hail`. The batch tests are jobs that run in default but submit test batches to the dev deployment and therefore need an entry for `internal.hail`. This essentially changes it to "everything knows about internal but only default knows about default.


We haven't seen this before because cloud dns in google provides these entries as a fallback, but I never set up in azure because it didn't seem necessary. If this succeeds, we can delete cloud dns in google

jigold (approved):

PR #11072, opened 2021-11-17
[hailtop] add hailtop.aiotools.delete

danking:
1. Implement hailtop.aiotools.delete.
2. Unify most definitions of rmtree and actually use parallelism.
3. Remove unnecesary and confusing async annotation on OnlineBoundedGather2.call.
4. Substantially increase the complexity of the rmtree test.

I had to fix a circularity caused by RouterAsyncFS referencing the other clouds. The fix was easy, I don't re-expose it in hailtop.aiotools.


danking:
I have a subsequent PR that uses it in build.yaml. Using it in batch is going to take a bit more thinking because we need an image with Hail available.

danking:
ready

hail/python/hailtop/aiotools/local_fs.py
```diff
@@ -205,14 +208,49 @@ async def remove(self, url: str) -> None:
...
+            listener(-1)
+
+        async with OnlineBoundedGather2(sema) as pool:
+            tasks = []
+            rm_directory_task_by_url: Dict[str, asyncio.Task] = {}
+            for dirpath, dirnames, filenames in os.walk(url, topdown=False):
+                contents_tasks = [
+                    pool.call(rm_file, os.path.join(dirpath, filename))
+                    for filename in filenames
+                ] + [
```
jigold, line 247: How does this work if you can't guarantee tasks will complete in a given order. i.e. I'm worried about the case where `rmdir(/tmp/foo)` finishes before `remove('/tmp/foo/bar')`?

Ok. I think I see what you are doing. You're waiting for one level at a time.

jigold (changes requested):

hail/python/hailtop/aiocloud/aioaws/fs.py
```diff
@@ -383,7 +384,11 @@ async def _listfiles_flat(self, bucket: str, name: str) -> AsyncIterator[FileLis
                 for item in contents:
                     yield S3FileListEntry(bucket, item['Key'], item)
 
-    async def listfiles(self, url: str, recursive: bool = False) -> AsyncIterator[FileListEntry]:
+    async def listfiles(self,
+                        url: str,
+                        recursive: bool = False,
+                        exclude_trailing_slash_files: bool = True  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
danking: The aioaws implementation always included trailing slash files. We didn't have tests for that. We also didn't have tests for Azure, but Azure excluded them, though somewhat differently from GCP (but in the same way as S3).

I enabled the tests for Azure and S3 and fixed S3.

hail/python/hailtop/aiocloud/aioazure/fs.py
```diff
@@ -374,7 +374,10 @@ async def _listfiles_flat(self, client: ContainerClient, name: str) -> AsyncIter
                 url = f'hail-az://{client.account_name}/{client.container_name}/{item.name}'
                 yield AzureFileListEntry(url, item)
 
-    async def listfiles(self, url: str, recursive: bool = False) -> AsyncIterator[FileListEntry]:
+    async def listfiles(self,
+                        url: str, recursive: bool = False,
+                        exclude_trailing_slash_files: bool = True
```
jigold, line 379: Can you move the `recursive` argument to a new line?

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -423,10 +429,10 @@ async def test_file_overwrite_dir(copy_test_context):
 
 
 @pytest.mark.asyncio
-async def test_file_and_directory_error(router_filesystem):
+async def test_file_and_directory_error(router_filesystem: Tuple[asyncio.Semaphore, AsyncFS, Dict[str, str]], cloud_scheme):
```
jigold, line 432: If you're going to annotate these, might as well do `cloud_scheme` as well.

hail/python/test/hailtop/aiotools/test_copy.py
```diff
@@ -494,9 +500,8 @@ async def test_file_and_directory_error_with_slash_empty_file(router_filesystem)
             exp_dest = f'{dest_base}foo'
             await expect_file(fs, exp_dest, 'foo')
 
-
 @pytest.mark.asyncio
-async def test_file_and_directory_error_with_slash_non_empty_file(router_filesystem):
+async def test_file_and_directory_error_with_slash_non_empty_file_for_google_non_recursive(router_filesystem: Tuple[asyncio.Semaphore, AsyncFS, Dict[str, str]]):
```
jigold, line 504: Can you explain this test to me and why you needed to split it into two?

hail/python/test/hailtop/test_fs.py
```diff
@@ -267,18 +269,60 @@ async def test_rmtree(filesystem):
...
+    assert await fs.isdir(subdir3)
+    # subdir4_empty: in cloud fses, empty dirs do not exist and thus are not dirs
 
     await fs.rmtree(sema, dir)
 
     assert not await fs.isdir(dir)
 
 
+@pytest.mark.asyncio
+async def test_rmtree_empty_dir(filesystem):
```
jigold, line 316: Do we need a test for a non-empty file that ends in a slash?

jigold (dismissed):
This kind of PR makes me nervous because of the user implications. I think the tests are thorough, but I was trying to think where this could have unintended consequences. What do you think of a test that creates a tree of directories but only deletes a subtree. Then assert the top level directories and the other subtrees exist. This will convince me that we're not just blindly deleting everything even unintended files which will pass the tests..

hail/python/hailtop/aiotools/copy.py
```diff
@@ -58,7 +29,7 @@ async def copy(*,
         if 'thread_pool' not in s3_kwargs:
             s3_kwargs['thread_pool'] = thread_pool
 
-        async with RouterAsyncFS(default_scheme,
+        async with RouterAsyncFS(default_scheme='file',
```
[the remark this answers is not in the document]
danking: You mean if we delete the LocalAsyncFS class? I think we can safely assume that anywhere our code is running has some local file system.

hail/python/hailtop/aiotools/local_fs.py
```diff
@@ -205,14 +208,46 @@ async def remove(self, url: str) -> None:
...
+            listener(-1)
+
+        async with OnlineBoundedGather2(sema) as pool:
+            contents_tasks_by_dir: Dict[str, List[asyncio.Task]] = {}
+            for dirpath, dirnames, filenames in os.walk(path, topdown=False):
+                contents_tasks = [
+                    pool.call(rm_file, os.path.join(dirpath, filename))
+                    for filename in filenames
+                ] + [
+                    pool.call(rm_dir, pool, contents_tasks_by_dir[fulldirname], fulldirname)
```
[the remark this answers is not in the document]
danking: Yeah, topdown=False ensures that directories come after their contents.

hail/python/test/hailtop/aiotools/test_copy.py
danking: These tests check that we get errors when there is a file named `not-empty/` and a file named `not-empty/bar`. In the cloud, the existence of `not-empty/bar` implies the existence of a directory named `not-empty`. This should trigger a `FileAndDirectoryError`.

Google differs from Azure and S3 in how it lists files matching a prefix. Google considers the file named `gs://bucket/not-empty/` to be inside `gs://bucket` but not inside the "folder" named `gs://bucket/not-empty/`. Azure and AWS act differently. In Azure and AWS, the file named `/bucket/not-empty/` is considered to be inside the folder named `/bucket/not-empty/` (if such a folder exists).

As a result, on any cloud, you *should* get an error when recursively listing `gs://bucket` if it contains a file and directory with the same name. On AWS and Azure, you will not get an error when non-recursively listing `gs://bucket` because you never see both the file and the directory in the same list. If you list `/bucket` you only see the folder. If you list `/bucket/not-empty/` you only see the file. On GCP, if you list `/bucket` you see both `/bucket/not-empty/` the file and `/bucket/not-empty/` the folder.

In any cloud, if there exists a file-and-directory situation, an error will always occur when copying from the cloud to the local file system.

To accommodate the differing behavior of GCP, I moved the assertions that are only true for GCP into a pair of `...google_non_recursive` tests.

hail/python/test/hailtop/test_fs.py
```diff
@@ -267,18 +269,60 @@ async def test_rmtree(filesystem):
...
+        functools.partial(fs.touch, f'{subdir}a{i:02}')
+        for subdir in [dir, subdir1, subdir2, subdir3, subdir1subdir1, subdir1subdir2, subdir1subdir3]
+        for i in range(30)])
+
     assert await fs.isdir(dir)
+    assert await fs.isdir(subdir1)
+    assert await fs.isdir(subdir1subdir1)
+    assert await fs.isdir(subdir1subdir2)
+    assert await fs.isdir(subdir1subdir3)
+    # subdir1subdir4_empty: in cloud fses, empty dirs do not exist and thus are not dirs
```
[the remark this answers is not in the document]
danking: I will add that, but I think a comment is warranted because its confusing that something named subdir1subdir3 is not a directory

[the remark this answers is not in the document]
danking: Erm, actually I can't add that assertion because that's *not* true for the local filesystem.

hail/python/hailtop/aiotools/copy.py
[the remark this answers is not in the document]
jigold: This will break this init method if there's no explicit local filesystems given.

```
class RouterAsyncFS(AsyncFS):
    def __init__(self, default_scheme: Optional[str], filesystems: List[AsyncFS]):
        scheme_fs = {}
        schemes = set()
        for fs in filesystems:
            for scheme in fs.schemes():
                if scheme not in schemes:
                    scheme_fs[scheme] = fs
                    schemes.add(scheme)

        if default_scheme not in schemes:
            raise ValueError(f'default scheme {default_scheme} not in set of schemes: {", ".join(schemes)}')

        self._default_scheme = default_scheme
        self._filesystems = filesystems
        self._schemes = schemes
        self._scheme_fs = scheme_fs
```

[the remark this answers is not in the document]
jigold: I was looking at old code. Nevermind

hail/python/test/hailtop/test_fs.py
danking: Done.

danking:
Great suggestion, I added an assertion and some code to the rmtree test to verify that deleting a subdirectory does not affect the sibling or avuncular directories.

hail/python/test/hailtop/test_fs.py
```diff
@@ -307,13 +309,35 @@ async def test_rmtree(filesystem):
...
+    assert await fs.isfile(f'{dir}a')
+    assert await fs.isfile(f'{dir}b')
+
+    assert await fs.isdir(subdir1)
+    assert await fs.isfile(f'{subdir1}a00')
+
+    assert await fs.isdir(subdir1subdir1)
+    assert await fs.isfile(f'{subdir1subdir1}a00')
+
+    assert not await fs.isdir(subdir1subdir2)
```
jigold, line 324: One more:

```python3
assert not await fs.isfile(f'{subdir1subdir2}a00')
```

jigold (dismissed):

jigold (approved):

PR #11077, opened 2021-11-20
Return paths from write_matrix_tables

jmartasek:
I think this little change adds a lot of usability to the method.

You do not need to know how the names of the tables are constructed in the function.


tpoterba:
good change, thanks!

tpoterba (approved):

PR #11078, opened 2021-11-22
[query] import_table better error empty missing

johnc1231:
Previously, accidentally passing an empty string to the list of missing values throws an inscrutable assertion error. This checks in python.

### DIFF
--- a/hail/python/hail/methods/impex.py
+++ b/hail/python/hail/methods/impex.py
@@ -1602,6 +1602,9 @@ def import_table(paths,
     comment = wrap_to_list(comment)
     missing = wrap_to_list(missing)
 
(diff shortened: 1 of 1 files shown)


tpoterba:
Chris is OOO, I'll take this.

johnc1231:
Maybe? We have an assertion in `TextMatrixReader.parseOptionalValue` that the missing values are not empty strings. I didn't trace through why, I just moved the error up.

johnc1231:
Looks like we have a test where we pass `missing=""` that used to work, so I'm doing something wrong here or misunderstanding something. Will investigate and get back to you.

johnc1231:
Closing for now

PR #11082, opened 2021-11-22
[batch] Ignore job deletion when uploading the logs

jigold:
I'm not 100% sure I know the implications of this change with regards to race conditions. I think it's okay? Also, why do we not include the timings if the job is deleted in `__exit__`? This is the current code:

```python3
class ContainerStepManager:
    def __init__(self, timing: Dict[str, float], is_deleted: Callable[[], bool], ignore_job_deletion: bool = False):
        self.timing: Dict[str, float] = timing
        self.is_deleted = is_deleted
        self.ignore_job_deletion = ignore_job_deletion

    def __enter__(self):
        if self.is_deleted() and not self.ignore_job_deletion:
            raise JobDeletedError()
        self.timing['start_time'] = time_msecs()

    def __exit__(self, exc_type, exc, tb):
        if self.is_deleted() and not self.ignore_job_deletion:
            return
        finish_time = time_msecs()
        self.timing['finish_time'] = finish_time
        self.timing['duration'] = finish_time - self.timing['start_time']
```

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -350,27 +350,28 @@ class Timings:
         self.timings: Dict[str, Dict[str, float]] = dict()
         self.is_deleted = is_deleted
 
-    def step(self, name: str):
+    def step(self, name: str, ignore_job_deletion: bool = False):
         assert name not in self.timings
         self.timings[name] = dict()
-        return ContainerStepManager(self.timings[name], self.is_deleted)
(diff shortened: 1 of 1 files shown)


daniel-goldstein (approved):
This seems fine. Ultimately I want to change this such that `ContainerStepManager` never raise and only times things, and job deletions are explicitly accounted for by wrapping cancellable steps of the job-running process with `run_until_done_or_deleted`.

PR #11083, opened 2021-11-22
[batch] Monkey patch aiodocker to be urlsafe

jigold:


batch/batch/worker/worker.py
```diff
@@ -84,6 +86,19 @@ def deeper_stack_level_warn(*args, **kwargs):
 
 warnings.warn = deeper_stack_level_warn
 
+
```
daniel-goldstein, line 89: Can you put a comment here explaining this patch, with a link to the broken line of code in aiodocker?

daniel-goldstein (dismissed):

daniel-goldstein (approved):

PR #11084, opened 2021-11-23
[infra] Add terraform for CI infrastructure in Azure

daniel-goldstein:
This adds terraform modules for CI azure resources and CI k8s resources. It also many roles to the test account that the batch account has so that we can run PR and test namespaces (internal batch instances use the test account for all the services so the test account needs tons of privilege. It looks scary, but this is the model we currently have). Thanks to #11053, which is the current version of the CI deployment in Azure, this required no change to the CI application code.

### Sidebar
It might look weird that I've added a block here for the kubernetes provider. That is because up until now I've kept all the k8s and azure terraform in separate root modules, so that they need to be `terraform apply`'d separately and therefore their provider blocks were separated as well. While keeping the code isolated is good (the k8s modules can be reused for GCP), putting them in separate apply's was purely because of [this bug](https://github.com/hashicorp/terraform-provider-kubernetes/issues/1028) in the kubernetes provider. I ran into it when experimenting tearing clusters down and putting them back up again. However, it has since proven very cumbersome to manage two different terraform states where one relies heavily on the other and I've changed my mind. The bug in question has a PR forthcoming and is really only a problem when tearing a K8s cluster down and rebuilding it while preserving other terraform state, which isn't something I see us dealing with often past initial development. Thus, I've added the k8s provider block so that I can directly invoke the CI k8s module. I'll follow up with a PR that moves the other k8s module invocations in here as well. If it would help, I can first start with a dev doc detailing our terraform structure (or where I want it to be).

### DIFF
--- a/infra/azure/main.tf
+++ b/infra/azure/main.tf
@@ -23,6 +23,20 @@ provider "azurerm" {
   features {}
 }
 
+provider "kubernetes" {
+  host = "https://${azurerm_kubernetes_cluster.vdc.fqdn}"
+
+  cluster_ca_certificate = base64decode(
+    azurerm_kubernetes_cluster.vdc.kube_config[0].cluster_ca_certificate
+  )
+  client_certificate = base64decode(
+    azurerm_kubernetes_cluster.vdc.kube_config[0].client_certificate
+  )
+  client_key = base64decode(
+    azurerm_kubernetes_cluster.vdc.kube_config[0].client_key
+  )
+}
+
 locals {
   acr_name = var.acr_name == "" ? var.az_resource_group_name : var.acr_name
   # An IP toward the top of the batch-worker-subnet IP address range
@@ -96,6 +110,11 @@ resource "azurerm_kubernetes_cluster" "vdc" {
       enabled = true
       log_analytics_workspace_id = azurerm_log_analytics_workspace.logs.id
     }
+  }
+
+  # https://github.com/hashicorp/terraform-provider-azurerm/issues/7396
+  lifecycle {
+    ignore_changes = [addon_profile.0]
   }
 }
 
@@ -356,6 +375,12 @@ resource "azurerm_storage_management_policy" "test" {
   }
 }
 
+resource "azurerm_role_assignment" "test_test_container_contributor" {
+  scope                = azurerm_storage_container.test.resource_manager_id
+  role_definition_name = "Storage Blob Data Contributor"
+  principal_id         = module.test_sp.principal_id
+}
+
 resource "azurerm_role_assignment" "batch_test_container_contributor" {
   scope                = azurerm_storage_container.test.resource_manager_id
   role_definition_name = "Storage Blob Data Contributor"
@@ -366,6 +391,12 @@ resource "azurerm_role_assignment" "batch_worker_test_container_contributor" {
   scope                = azurerm_storage_container.test.resource_manager_id
   role_definition_name = "Storage Blob Data Contributor"
   principal_id         = azurerm_user_assigned_identity.batch_worker.principal_id
+}
+
+resource "azurerm_role_assignment" "ci_test_container_contributor" {
+  scope                = azurerm_storage_container.test.resource_manager_id
+  role_definition_name = "Storage Blob Data Contributor"
+  principal_id         = module.ci_sp.principal_id
 }
 
 provider "azuread" {}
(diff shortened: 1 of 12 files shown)


infra/k8s/ci/main.tf
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+
+resource "kubernetes_secret" "hail_ci_0_1_service_account_key" {
+  metadata {
+    name = "hail-ci-0-1-service-account-key"
+  }
+
+  data = {
+    "user1" = var.github_user1_oauth_token
```
danking, line 40: we have not much control over these poorly named secrets, but we can pick good names for the variables. This one should be `ci_test_repo_creator_github_ouath_token`. I'm pretty sure the `test_ci` build step shouldn't mount the `hail-ci-0-1-service-account-key`. The other two mounts are because we create/delete a GitHub test repo.

infra/k8s/ci/main.tf
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+
+resource "kubernetes_secret" "hail_ci_0_1_github_oauth_token" {
+  metadata {
+    name = "hail-ci-0-1-github-oauth-token"
+  }
+
+  data = {
+    "oauth-token" = var.github_oauth_token
```
danking, line 30: See below message for context, but let's name this `ci_and_deploy_github_oauth_token`.

infra/azure/variables.tf
```diff
@@ -24,3 +24,14 @@ variable k8s_machine_type {
...
+variable "ci_config" {
+  type = object({
+    github_oauth_token = string
+    github_user1_oauth_token = string
+    watched_branches = list(tuple([string, bool, bool]))
+    deploy_steps = list(string)
+    github_context = string
+  })
+  default = null
+}
```
danking, line 37: It feels odd to me that these duplicate the variables in ci. What's your plan for these two things?

infra/azure/main.tf
```diff
@@ -475,3 +552,24 @@ module "grafana_sp" {
...
+  source = "./ci"
+  count = var.ci_config != null ? 1 : 0
+
+  resource_group        = data.azurerm_resource_group.rg
+  ci_principal_id       = module.ci_sp.principal_id
+  container_registry_id = azurerm_container_registry.acr.id
+}
+
+module "ci_k8s_resources" {
+  source = "../k8s/ci"
```
danking, line 566: Hmm. So, in terraform, declaring something else as a "module" is similar to invoking/calling that "code"? I guess I'd have expected to declare that k8s "depends on" a certain set of cloud resources. Then, I guess, you'd have to specify which cloud in a variable? Can you comment a bit on why my intuitions are wrong?

infra/azure/main.tf
```diff
@@ -442,13 +479,53 @@ module "ci_sp" {
...
+}
+
+resource "azurerm_role_assignment" "test_subscription_reader" {
+  scope                = data.azurerm_subscription.primary.id
+  role_definition_name = "Reader"
+  principal_id         = module.test_sp.principal_id
+}
+
+resource "azurerm_role_assignment" "test_shared_gallery_reader" {
+  scope                = azurerm_shared_image_gallery.batch.id
```
danking, line 518: why no data. here?

infra/azure/main.tf
```diff
@@ -442,13 +479,53 @@ module "ci_sp" {
...
   display_name = "${data.azurerm_resource_group.rg.name}-test"
+
+  required_resource_access {
+    resource_app_id = "00000003-0000-0000-c000-000000000000"
+
+    resource_access {
+      # Application.ReadWrite.All
+      id   = "1bfefb4e-e0b5-418b-a88f-73c46d2cc8e9"
+      type = "Role"
+    }
```
danking, line 490: Huh. Can you give some context on why we write a resource_access here? I would have expected an azurerm_role_assignment. Also, what does "Application.ReadWrite.All" mean? Is that full access to Azure AD Applications?

infra/azure/main.tf
```diff
@@ -356,6 +375,12 @@ resource "azurerm_storage_management_policy" "test" {
   }
 }
 
+resource "azurerm_role_assignment" "test_test_container_contributor" {
```
danking, line 378: hmm. I think I find these easier to read when grouped by principal rather than by role. What do you think?

infra/azure/main.tf
```diff
@@ -97,6 +111,11 @@ resource "azurerm_kubernetes_cluster" "vdc" {
       log_analytics_workspace_id = azurerm_log_analytics_workspace.logs.id
     }
   }
+
+  # https://github.com/hashicorp/terraform-provider-azurerm/issues/7396
+  lifecycle {
+    ignore_changes = [addon_profile.0]
```
danking, line 117: the issue suggests:
```
    ignore_changes = [addon_profile.0.kube_dashboard]
```
Should we use that?

danking (changes requested):
Overall I'm good with this, but I need a bit of education on terraform.

danking (approved):

PR #11087, opened 2021-11-29
[batch] Make test_batch_* pass in Azure

jigold:
The coalesce is necessary right now because Azure has no resources so the result is None rather than 0. Since the cost is always 0, then the tests that test billing limits fail.

In addition, the tests that choose whether we select the cheapest machine won't work because all machines cost $0/hr right now due to no billing setup. Instead, we get the first pool alphabetically (highcpu). 

The reason for the database upgrade is because we're selecting the first pool alphabetically which is highcpu. The default settings don't leave enough disk space and thus require an external disk. To circumvent this, I just made the worker data disk size the same as in GCP -- 375Gi -- for the highcpu pool.

### DIFF
--- a/batch/batch/batch_format_version.py
+++ b/batch/batch/batch_format_version.py
@@ -7,7 +7,7 @@ class BatchFormatVersion:
     def __init__(self, format_version):
         self.format_version = format_version
 
-    def has_full_spec_in_gcs(self):
+    def has_full_spec_in_cloud(self):
         return self.format_version > 1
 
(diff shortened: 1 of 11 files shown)


daniel-goldstein (commented):
LGTM but I'll let dan chime in re the database-related changes

danking (approved):
LGTM

danking:
Just move delete_all_test_billing_projects into some file that isn't also full of tests and import it into test_accounts.py

PR #11088, opened 2021-11-29
[build] Mark steps that cannot currently run on Azure as GCP only

daniel-goldstein:
There might be more that we won't be able to run but saving for the gcs_bucket/gsutil steps, these are the steps that I found after a couple scans through that we can't expect to run currently on azure.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2587,6 +2587,8 @@ steps:
       - memory_image
       - deploy_memory_sa
       - create_certs
+    clouds:
+      - gcp
   - kind: runImage
     name: test_lsm
     image:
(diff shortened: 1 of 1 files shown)


build.yaml
```diff
@@ -2648,6 +2676,8 @@ steps:
       - merge_code
       - memory_image
       - deploy_memory
+    clouds:
```
jigold, line 2679: Does `deploy_memory` work in Azure? Or is it just the test is too specific?

jigold (dismissed):

build.yaml
daniel-goldstein: memory might deploy but won't work because it specifically uses the GoogleStorage FS.

jigold: Let's set gcp-only for that step as well then.

jigold (approved):

PR #11089, opened 2021-11-29
[build.yaml] use hailtop.aiotools.copy (and re-enable FS tests)

danking:
Bigger than I expected, but:
1. Re-enable the FS tests and create a Gradle target for them so they can be run locally.
2. Allow the FS tests to be easily used locally by not hardcoding a particular key file path.
3. Skip GoogleStorageFSSuite when `CLOUD` is not `gcp`
4. Remove irrelevant env vars from non-FS Scala tests.
5. Eliminate the "hail_repl" image and deployment which was scoped dev anyway and never used.
6. Add hail_pip_installed_image which can be used to execute `hailtop.aiotools.copy`.
7. Use copy in two places in build.yaml.
8. Add a command line argument for configuring the number of concurrent transfers which sets an upper bound on the number of open source files (and, additionally, open destination files). On my MacBook, I can't seem to open 100 local files simultaneously. I set the default low enough that local use should work by default.


danking:
erm, this is missing some changes, hold on

build.yaml
```diff
@@ -882,30 +882,23 @@ steps:
...
     image:
-      valueFrom: base_image.image
+      valueFrom: hail_pip_installed_image.image
     script: |
       set -ex
       cd /io/repo/hail/
-      gcloud -q auth activate-service-account --key-file=/test-gsa-key/key.json
-      gsutil -m cp -r src/test/resources/* gs://{{ global.hail_test_gcs_bucket }}/{{ token }}/test/resources/
-      gsutil -m cp -r python/hail/docs/data/* gs://{{ global.hail_test_gcs_bucket }}/{{ token }}/doctest/data/
+      export GOOGLE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
```
daniel-goldstein, line 895: Will this always use `GOOGLE_APPLICATION_CREDENTIALS` or will it `AZURE_APPLICATION_CREDENTIALS`?

danking:
OK, companion changes in https://github.com/hail-is/hail/pull/11091.

build.yaml
```diff
@@ -3807,6 +3798,55 @@ steps:
...
+      cpu: '2'
+    script: |
+      set -ex
+      cd /io
+      tar xzf resources.tar.gz -C src/test
+
+      export CLOUD={{ global.cloud }}
+      export GOOGLE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
+      export HAIL_FS_TEST_CLOUD_RESOURCES_URI={{ global.test_storage_uri }}/{{ upload_test_resources_to_blob_storage.token }}/test/resources/fs
+      export HAIL_TEST_GCS_BUCKET={{ global.hail_test_gcs_bucket }}
```
daniel-goldstein, line 3816: This currently doesn't exist in azure, but I can see that it's necessary for the FS tests. I'll lump it in with the test keys to get this into the infrastructure, but maybe we should mark this as gcp-only for now? Jinja will barf about this being undefined.

daniel-goldstein (changes requested):

build.yaml
danking: Ah, OK, from Zulip I see we need to set both:
```
export AZURE_APPLICATION_CREDENTIALS =/test-gsa-key/key.json
```
even though that looks weird. We'll never try to use the google ones.

build.yaml
danking: Ah, hmm. I think Jinja will not complain if I hide it behind an if.

daniel-goldstein (approved):

build.yaml
```diff
@@ -3843,6 +3825,58 @@ steps:
...
+    name: test_hail_fs
+    image:
+      valueFrom: hail_run_tests_image.image
+    resources:
+      memory: standard
+      cpu: '2'
+    script: |
+      set -ex
+      cd /io
+      tar xzf resources.tar.gz -C src/test
```
daniel-goldstein, line 3838: Might need to `mkdir` here?

PR #11090, opened 2021-11-29
[hailctl] allow describe on s3, hail-az, gs, and local

danking:
Tested this locally after uploading a tiny dataset to all three clouds.


tpoterba:
ooh, nice!

tpoterba (approved):

PR #11091, opened 2021-11-29
[hailtop.batch] cloud agnostic copy and delete

danking:
Instead of using `gsutil` we use hailtop.aiotools.copy from the new `HAILGENETICS_HAIL_IMAGE`.  Previously, deploying the pip-versioned image was a manual asynchronous step that mostly happened in response to user requests.

1. Actually build and test hailgenetics/hail and hailgenetics/genetics on every build.
2. On deploy, push the newly built hailgenetics/hail and hailgenetics/genetics images to both docker hub and gcr.io/hail-vdc/
3. Provide the built-for-this-PR hailgenetics/hail as an env var to the tests.
4. By default use the hailgenetics/hail image for the currently published pip version for FS operations. Allow overriding by environment variable.
5. Remove now unnecessary publish-public-images.sh.


danking:
Ok, there were some issues with the local backend but I think that's resolved now.

hail/python/hailtop/batch/backend.py
```diff
@@ -184,18 +190,17 @@ def run_code(code):
...
+        requester_pays_project_json = orjson.dumps(batch.requester_pays_project).decode('utf-8')
 
         def copy_input(job, r):
             if isinstance(r, resource.InputResourceFile):
                 if r not in copied_input_resource_files:
                     copied_input_resource_files.add(r)
 
                     if r._input_path.startswith('gs://'):
-                        return [f'gsutil {requester_pays_project} cp -r {shq(r._input_path)} {shq(r._get_path(tmpdir))}']
+                        transfers_bytes = orjson.dumps([{"from": r._input_path, "to": r._get_path(tmpdir)}])
```
jigold, line 201: Do we still need the `if r._input_path.startswith('gs://'):` here or should this work everywhere now?

jigold (changes requested):
Overall this looks good. I need some food before I can take a closer look at the transfers for the hailtop.batch code.

hail/python/hailtop/batch/backend.py
jigold: bump

danking: It looks like, in the local backend, we sometimes use a symlink instead of a copy, so I tried to preserve that behavior. Admittedly, I didn't change the local-to-local-outside-docker to use copy tool, it just uses cp still.

jigold: ok. I think we'll want some global variable for the available schemes and check if the input path starts with one of the valid schemes.

hail/python/hailtop/batch/job.py
```diff
@@ -89,8 +89,8 @@ def __init__(self,
         self._job_id: Optional[int] = None
 
         self._inputs: Set[_resource.Resource] = set()
-        self._internal_outputs: Set[_resource.Resource] = set()
-        self._external_outputs: Set[_resource.Resource] = set()
+        self._internal_outputs: Set[_resource.ResourceFile] = set()
```
jigold, line 92: Are you sure this can't be a ResourceGroup or a PythonResult?

jigold (dismissed):

hail/python/hailtop/batch/backend.py
danking: Ah, sorry, I misunderstood your comment. Hmm. Yeah, you're right this should support hail-az, s3, and gs.

hail/python/hailtop/batch/job.py
danking: Yeah, you're right, this can also be a PythonResult, but AFAICT it is never a group. Groups add their constituent files to this variable.

hail/python/hailtop/batch/backend.py
danking: OK, here's a fix: https://github.com/hail-is/hail/pull/11091/commits/878177ab4363ddd5466790dce9d56d91d8af798b. If it has no scheme e.g. `/foo/bar`, then we presume its local. Otherwise, even `file:/`, we use the copy tool. That makes sense because `cp` doesn't understand `file:/` anyway.

hail/python/hailtop/batch/job.py
danking: Fixed.

danking:
@jigold, bump, it is passing

jigold (approved):

PR #11092, opened 2021-11-29
[batch] Infrastructure for billing with time and location varying resources

jigold:
I'd like your initial feedback before I start testing this on Azure. A substantially earlier version seemed to work fine on GCP with dev deploy.

The major conceptual change I made is a `resource` now contains a `prefix` and a `version`. The `resource_name` is just `{prefix}/{version}`. The prefixes for GCP are the same as they were before and don't vary by region. However, the new prefixes for Azure are region specific. The version is `1` for all current resources. 

I added a `latest_resource_versions` table that has the prefix mapped to the latest version. This is used to generate the current resource names.

There is a new CloudResourceManager that is in charge of managing the spot billing pricing cache and updating the prices in the cache and the database from the cloud provider's API. Since I couldn't easily rename resources to products everywhere in the database due to anonymous foreign key constraints, I had to rename the existing `CloudResourceManager` to `CloudDriverAPI`. Feel free to suggest a better name. The GCPResourceManager is a skeleton right now, but we'll have to flesh it out in the new year when GCP moves to spot billing with varying prices.

For the `AzureResourceManager`, I use a new pricing client to grab the latest vm and disk prices. I support all possible disk prices, but for now, I limited the VM query to just get the machine types we support right now. In the future, we could get all VM prices, but the query is around 40 seconds for that compared to 2 seconds now. I was worried if we had such a slow query that blocked driver startup, that would be bad and this is fine for now.

There are two classes I added: a `Resource` and a `Price`. The Price is only implemented for Azure and is used to store cost results from the pricing API. The resource has a couple of different mixin classes with an abstract method to generate the quantified resource depending on the type (ex: ComputeResourceMixin). Then there's `AzureDiskResource`, `AzureVMResource`, etc. I had to add a Resource specifically for external disks which made things a little more complicated because you don't know the size of the disk ahead of time. The reason for storing the resources in the InstanceConfig now is because we can change the implementation or add new products without breaking the billing for existing instances. I think this is a positive change and will give us more flexibility. I also worked hard to make sure the definition of a resource prefix was only defined in a consistent place for each resource type.

The `AzureExternalDiskResource` is complicated because we have to store all possible disk resource names that are current at the time of Instance creation for that disk family. This is because Azure charges different rates per disk type (ex: P20) and the cost per GiB is not consistent across products. Therefore, one job could use a P4 while another job uses a P50, so we have to store all disk types with their latest resource names.


danking (changes requested):

jigold:
Everything seems to be working. Ready for another look.

batch/batch/cloud/azure/resources.py
```diff
@@ -0,0 +1,228 @@
...
+                     location: str,
+                     redundancy_type: str = 'LRS',
+                     ) -> 'AzureDiskResource':
+        assert redundancy_type in ('LRS', 'ZRS'), redundancy_type
+
+        # Azure bills for specific disk sizes so we must round the storage_in_gib to the nearest power of two
+        disk = azure_disk_from_storage_in_gib(disk_type, storage_in_gib)
+        assert disk, f'disk_type={disk_type} storage_in_gib={storage_in_gib}'
+
+        prefix = AzureDiskResource.generate_prefix(disk.name, redundancy_type, location)
```
danking, line 44: I say this below, but I again find the term "prefix" obfuscating here. This is really the name of the kind/type of resource with the given disk type, redundancy type, and location. Then, on the next line, we say "get me the latest *fully versioned* resource of this name".

"generate_prefix" gives me the "kind" or "type" of resource which fits my needs. "latest_resource_name" finds the latest "version" of that "kind/type" of resource.

---

Hmm.

OK, here's a question: why are we concatenating the version into the resource's name in the database? That feels like the opposite of how databases are supposed to work. Each distinct piece of information should be in a column.

This feels more in the spirit of SQL and also less obfuscating of the difference between resource types and resource versions:

```sql
CREATE TABLE IF NOT EXISTS `resources` (
  `name` VARCHAR(100) NOT NULL,
  `version` VARCHAR(100) NOT NULL,
  `rate` DOUBLE NOT NULL,
   PRIMARY KEY (`name`, `version`)
 ) ENGINE = InnoDB;

 CREATE TABLE IF NOT EXISTS `latest_resource_versions` (
   `name` VARCHAR(100) NOT NULL,
   `version` VARCHAR(100) NOT NULL,
   PRIMARY KEY (`name`)
 ) ENGINE = InnoDB;
```

We should probably make name a `UNIQUE KEY`. That will ensure there's always exactly one latest version. We couldn't achieve that if we used a boolean flag on the resources table; therefore, I do like having two tables.

Now we can pervasively use the term "name" to refer to the kind/type of resource and "version" to refer to the version of it (which in the Azure case is always a date stamp).

batch/batch/cloud/azure/driver/billing_manager.py
```diff
@@ -0,0 +1,103 @@
...
+        resource_version_updates = []
+
+        for price in await fetch_prices(self.pricing_client, self.regions):
+            if isinstance(price, AzureVMPrice):
+                resource_prefix = AzureVMResource.generate_prefix(price.machine_type, price.preemptible, price.region)
+                latest_resource_rate = rate_instance_hour_to_fraction_msec(price.cost_per_hour, 1024)
+            else:
+                assert isinstance(price, AzureDiskPrice)
+                resource_prefix = AzureDiskResource.generate_prefix(price.disk_name, price.redundancy_type, price.region)
+                latest_resource_rate = rate_gib_month_to_mib_msec(price.cost_per_gib_month)
```
danking, line 59: When we use isinstance to switch between two classes, that's usually a sign we should use an abstract method instead. How does the code look if we have a `resource_prefix` method and a `resource_rate` method?

batch/batch/cloud/azure/resources.py
danking: We might even want a `VersionedResource` which is a pair of a version and a name. We could have a `VersionedRatedResource` which has a version, rate, and a name.

batch/batch/cloud/azure/driver/billing_manager.py
```diff
@@ -0,0 +1,103 @@
...
+
+            latest_resource_version = price.version
+            resource_name = resource_prefix_version_to_name(resource_prefix, latest_resource_version)
+
+            current_resource_rate = self.resource_rates.get(resource_name)
+            current_resource_version = self.resource_versions.latest_version(resource_prefix)
+
+            if current_resource_rate is None:
+                resource_updates.append((resource_name, latest_resource_rate))
+            elif abs(current_resource_rate - latest_resource_rate) >= 1E-20:
```
danking, line 69: In 2022, we need to sit down and think about why our rates are floating point numbers. This makes me deeply uncomfortable.

danking: Why can't this be equality? The latest resource rate and the current resource rate should absolutely be the exact same floating point number. How would they not be the same? Don't the latest and current both ultimately come from the database?

batch/batch/cloud/azure/driver/create_instance.py
```diff
@@ -38,10 +38,14 @@ def create_vm_config(
         subscription_id: str,
         resource_group: str,
         ssh_public_key: str,
+        max_price: Optional[float],
         instance_config: InstanceConfig,
 ) -> dict:
     _, cores = azure_machine_type_to_worker_type_and_cores(machine_type)
 
+    if max_price is not None and not preemptible:
+        log.exception(f'max price given for a nonpreemptible machine {max_price}')
```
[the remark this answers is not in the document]
danking: If we raise an exception, do we still need a log statement? The exception will propagate somewhere and get logged there? (I'm trying to fight the flood of exceptional logs we have).

batch/batch/cloud/azure/driver/pricing.py
```diff
@@ -0,0 +1,165 @@
...
+    time_updated: int
+
+    def is_current_price(self):
+        now = time_msecs()
+        return (now >= self.effective_start_date
+                and (self.effective_end_date is None or now <= self.effective_end_date))
+
+    @property
+    def version(self) -> str:
+        return datetime.datetime.utcfromtimestamp(self.effective_start_date / 1000).strftime('%Y-%m-%d')
```
danking, line 31: It makes me a bit uncomfortable that we throw away information here. Why don't we just use the actual effective_start_date in (it seems) milliseconds since the epoch as the version? I feel like there's all sorts of opportunities for confusion if Azure ever sends us a start date that's, say, at noon. Is there a guarantee in their documentation that the start date is always midnight or something?

batch/batch/cloud/azure/driver/pricing.py
```diff
@@ -0,0 +1,165 @@
...
+        preemptible = ('Spot' in sku_name)
+        vm_cost_per_hour = float(data['retailPrice'])
+
+        start_date = parse_timestamp_msecs(data['effectiveStartDate'])
+        end_date = parse_timestamp_msecs(data.get('effectiveEndDate'))
+
+        if sku_name in seen_vm_names:
+            seen_data = seen_vm_names[sku_name]
+            log.exception(f'already seen pricing for vm {sku_name}; {seen_data} vs {data}; aborting')
+            raise Exception()
```
danking, line 116: If we raise, why do we log? The caller is obligated to either handle the exception or log the issue. I want to reduce the amount of repeated exception logs we get. Also, shouldn't the exception include the error information?

batch/batch/cloud/azure/driver/pricing.py
```diff
@@ -0,0 +1,165 @@
...
+        seen_vm_names[sku_name] = data
+
+        vm_price = AzureVMPrice(machine_type, preemptible, region, vm_cost_per_hour, start_date, end_date)
+        prices.append(vm_price)
+
+    return prices
+
+
+async def get_managed_disk_prices_by_region(pricing_client: aioazure.AzurePricingClient, region: str) -> List[AzureDiskPrice]:
```
danking, line 125: prefer no get_ since it doesn't convey any extra information.

danking (changes requested):

batch/batch/cloud/azure/resources.py
jigold: I agree with you that when we created the database resources table, we should have added a version field. If we change the table structure, there's a couple of things that I worry about:

1. Backwards compatibility with old workers. This should be easy to fix
2. We'll have to figure out how to split the string in the resources table to get rid of the version (which is always 1 right now) which is not straightforward.
3. We'll also need to account for the aggregated_*_resources tables.

In all, I thought it was easier to do what I did and have the full name still in the database.

batch/batch/cloud/azure/driver/pricing.py
jigold: We can do the full timestamp. We might want a resource_versions table instead that contains when the version was added. That would fulfill the purpose of easily being able to tell what's going on.

However, you said something about not having a boolean field to say which is the latest version.

batch/batch/cloud/azure/resources.py
jigold: There actually is a function that we can use: `SUBSTRING_INDEX`

jigold: If you are okay with a huge migration (we should backup the database ahead of time!!!!) and we'll need to take Batch offline for a long time, then we can add the proper tables.

batch/sql/populate_latest_resource_versions.py
```diff
@@ -0,0 +1,32 @@
...
+    latest_versions = []
+    for resource in resources:
+        prefix, version = resource.rsplit('/', maxsplit=1)
+        latest_versions.append((prefix, version))
+
+    await db.execute_many('''
+INSERT INTO `latest_resource_versions` (prefix, version)
+VALUES (%s, %s)
+''',
+                          latest_versions)
```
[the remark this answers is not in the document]
jigold: I don't quite understand the `INSERT INTO ... FROM` part of your comment.

jigold:
So the nomenclature I settled on is we have a (product, version) = resource

Eventually, we can do the right thing with the tables.

danking (dismissed):
just a few more small things

batch/batch/cloud/azure/resources.py
```diff
@@ -0,0 +1,233 @@
...
+    def product_name(disk_product: str, redundancy_type: str, location: str) -> str:
+        return azure_disk_product(disk_product, redundancy_type, location)
+
+    @staticmethod
+    def from_dict(data: Dict[str, Any]) -> 'AzureExternalDiskResource':
+        assert data['type'] == AzureExternalDiskResource.TYPE
+        return AzureExternalDiskResource(data['disk_type'], data['location'], data['latest_disk_versions'])
+
+    @staticmethod
+    def new_resource(product_versions: ProductVersions,
```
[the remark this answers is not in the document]
jigold: Weird. I thought I did change this, but clearly I didn't.

batch/batch/cloud/azure/driver/billing_manager.py
```diff
@@ -0,0 +1,96 @@
...
+        @transaction(self.db)
+        async def insert_or_update(tx):
+            if resource_updates:
+                await tx.execute_many('''
+INSERT INTO `resources` (resource, rate)
+VALUES (%s, %s)
+''',
+                                      resource_updates)
+
+            if resource_version_updates:
```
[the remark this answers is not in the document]
jigold: Good catch!

batch/batch/driver/billing_manager.py
```diff
@@ -0,0 +1,73 @@
...
+        for product, old_version in self._product_versions.items():
+            new_version = data.get(product)
+            if new_version is None:
+                log.exception(f'product {product} does not appear in new data; keeping in product versions')
+                continue
+
+            if new_version != old_version:
+                log.info(f'updated product version for {product} from {old_version} to {new_version}')
+
+        if self._product_versions:
```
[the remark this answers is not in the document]
jigold: I was trying to distinguish between adding a new product to the database versus an empty data structure in Python on init that hasn't been populated from the database yet. Let me think if there's a better way to do this that's clearer.

batch/batch/resources.py
```diff
@@ -0,0 +1,109 @@
...
+                               worker_fraction_in_1024ths: int,
+                               external_storage_in_gib: int) -> Optional[QuantifiedResource]:
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def to_dict(self) -> dict:
+        raise NotImplementedError
+
+
+class DiskResourceMixin(Resource, abc.ABC):
```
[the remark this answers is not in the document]
jigold: It's meant to capture a disk that has a known size at creation time versus a disk where the size is unknown at creation time. For example, we know the boot disk image is 32 Gi in Azure, but we don't know what each user will mount until the caller wants to quantify the resource with the external storage amount.

I'll pick a better name.

jigold:
This is finally passing in both clouds.

danking (approved):

PR #11094, opened 2021-11-30
[batch] Fix validate_int function to not reraise

jigold:
I'm pretty sure this is why the session messages were gone when trying to configure the pools pages.


daniel-goldstein (changes requested):
linters seem to want you to use `from`. I don't actually know what the semantic difference is here. How come this will fix the error?

jigold:
I'm not sure, but it's the only difference I could find between session messages that worked on the front end and the messages not showing up from the driver. I want to test this explicitly, but my namespace is currently messed up in Azure with previous database changes.

daniel-goldstein (approved):

PR #11095, opened 2021-11-30
[hailgenetics] genetics

danking:
Restore the genetics image, which I removed from #11091 for expediency.


danking:
closing while I figure out the build time issue.

PR #11098, opened 2021-12-01
[batch] Add auth.hail to job /etc/hosts

daniel-goldstein:
Discovered in azure when it couldn't run the auth copy-paste token tests. We don't see this in google because we use Cloud DNS for resolving *.hail to internal gateway for private jobs (we use /etc/hosts and a public resolver for public jobs). In Azure I decided to forego azure dns and just use the /etc/hosts route for all of our jobs.

### DIFF
--- a/batch/Dockerfile.worker
+++ b/batch/Dockerfile.worker
@@ -66,3 +66,4 @@ COPY batch/batch /batch/batch/
 RUN hail-pip-install --no-deps /batch && rm -rf /batch
 
 COPY query/log4j.properties /
+COPY letsencrypt/subdomains.txt /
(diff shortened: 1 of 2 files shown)


daniel-goldstein:
If this works completely with azure we could drop GCP's Cloud DNS.

danking (dismissed):
This is great, but let's copy subdomains.txt in at build time so that we both use the complete list and don't duplicate the list. I think it will be a pretty easy change to this PR.

danking:
Man, you got some *weird* errors in this PR. Why is asyncio giving us a `None` for a loop?

danking (approved):

PR #11099, opened 2021-12-01
[atgu] Delete atgu site

daniel-goldstein:


danking (approved):
I'm unreasonably jealous of this +3 -1131 commit.

Carla confirmed we can delete it. @daniel-goldstein can you do that once this deploys?

PR #11100, opened 2021-12-01
[ci] Add image cleanup for PR runs in Azure

daniel-goldstein:
Implemented image untagging for image cleanup steps (like is done in GCR) for Azure. Since old layers still should be used for caching, this just removes the tag used for an image in a test build. We can then do something like [here](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-auto-purge#run-in-an-on-demand-task) where you can purge untagged layers that are older than some number of weeks where we believe they're no longer relevant to the layer cache.

I also switched out the `registry-push-credentials` that CI uses to build images from the ACR admin login to CI's service principal and eliminated the admin login from the ACR terraform resource.

I dev deployed CI and manually verified after a deploy that a tag that was cleaned up no longer showed up in acr

### DIFF
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -352,15 +352,37 @@ cat /home/user/trace
         )
 
     def cleanup(self, batch, scope, parents):
-        if CLOUD == 'azure':
-            log.warning('Image cleanup in ACR is not yet supported')
-            return
-        assert CLOUD == 'gcp'
-
         if scope == 'deploy' and self.publish_as and not is_test_deployment:
             return
 
-        script = f'''
+        if CLOUD == 'azure':
+            image = 'mcr.microsoft.com/azure-cli'
+            assert self.image.startswith(DOCKER_PREFIX)
+            image_name = self.image[len(f'{DOCKER_PREFIX}/') :]
+            script = f'''
+set -x
+date
+
+set +x
+USERNAME=$(cat /secrets/registry-push-credentials/credentials.json | jq -j '.appId')
+PASSWORD=$(cat /secrets/registry-push-credentials/credentials.json | jq -j '.password')
+TENANT=$(cat /secrets/registry-push-credentials/credentials.json | jq -j '.tenant')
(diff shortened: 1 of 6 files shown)


ci/ci/build.py
```diff
@@ -352,15 +352,37 @@ def build(self, batch, code, scope):
...
+date
+
+set +x
+USERNAME=$(cat /secrets/registry-push-credentials/credentials.json | jq -j '.appId')
+PASSWORD=$(cat /secrets/registry-push-credentials/credentials.json | jq -j '.password')
+TENANT=$(cat /secrets/registry-push-credentials/credentials.json | jq -j '.tenant')
+az login --service-principal -u $USERNAME -p $PASSWORD --tenant $TENANT
+set -x
+
+until az acr repository untag -n {shq(DOCKER_PREFIX)} --image {shq(image_name)} || ! az acr repository show -n {shq(DOCKER_PREFIX)} --image {shq(image_name)}
```
jigold, line 373: Let's make sure there's a timeout on this build step of say 5 or 10 minutes.

ci/ci/build.py
```diff
@@ -352,15 +352,37 @@ def build(self, batch, code, scope):
...
-        assert CLOUD == 'gcp'
-
         if scope == 'deploy' and self.publish_as and not is_test_deployment:
             return
 
-        script = f'''
+        if CLOUD == 'azure':
+            image = 'mcr.microsoft.com/azure-cli'
+            assert self.image.startswith(DOCKER_PREFIX)
+            image_name = self.image[len(f'{DOCKER_PREFIX}/') :]
```
jigold, line 361: formatting here? Should there be the extra space before the colon?

jigold (dismissed):

ci/ci/build.py
daniel-goldstein: yes, `black` put it there.

jigold (approved):

PR #11102, opened 2021-12-01
[build.yaml] eliminate hail_query_gcs_path

danking:
After this deploys and we re-run terraform, I can remove the deprecated entry.

### DIFF
--- a/infra/azure/main.tf
+++ b/infra/azure/main.tf
@@ -306,6 +306,12 @@ resource "azurerm_storage_container" "batch_logs" {
   container_access_type = "private"
 }
 
+resource "azurerm_storage_container" "query" {
+  name                  = "query"
+  storage_account_name  = azurerm_storage_account.batch.name
+  container_access_type = "private"
+}
+
 resource "azurerm_role_assignment" "batch_batch_account_contributor" {
   scope                = azurerm_storage_account.batch.id
   role_definition_name = "Storage Blob Data Contributor"
--- a/query/deployment.yaml
+++ b/query/deployment.yaml
@@ -96,13 +96,13 @@ spec:
            - name: HAIL_QUERY_WORKER_IMAGE
              value: {{ query_image.image }}
 {% if deploy %}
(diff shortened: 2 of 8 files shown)


infra/azure/main.tf
```diff
@@ -306,6 +306,12 @@ resource "azurerm_storage_container" "batch_logs" {
   container_access_type = "private"
 }
 
+resource "azurerm_storage_container" "query" {
+  name                  = "query"
+  storage_account_name  = azurerm_storage_account.batch.name
+  container_access_type = "private"
+}
```
daniel-goldstein, line 313: As is, only the batch service principal will have access to this container. Do any other service accounts need access?

query/deployment.yaml
```diff
@@ -96,13 +96,13 @@ spec:
            - name: HAIL_QUERY_WORKER_IMAGE
              value: {{ query_image.image }}
 {% if deploy %}
-           - name: HAIL_QUERY_GCS_PATH
+           - name: HAIL_QUERY_STORAGE_UI
```
daniel-goldstein, line 99: nit: can move the name out of the `if` to avoid having to write the same name twice.

daniel-goldstein (changes requested):
Looks fine just a question about access control

infra/azure/main.tf
danking: No, in the long run Batch workers will fetch the JAR.

daniel-goldstein: In that case you'll need to give the batch worker managed identity access to the container

daniel-goldstein (dismissed):
See comment about batch worker permissions

infra/azure/main.tf
danking: This is a new container, not a new account. AFAIK, the batch worker already has access to the batch storage account.

daniel-goldstein: Ah, you're right! Never mind.

daniel-goldstein (approved):

PR #11103, opened 2021-12-01
[utils] Retry when rate limited

daniel-goldstein:
A little surprised there's no prior mention of 429's in here which made me think I might have put this in the wrong place but afaict this is the way we would retry when being rate limited.


danking (approved):
I think google might just go 503 service unavailable.

PR #11106, opened 2021-12-01
[query] Make flattened version of to_pandas work without Spark

johnc1231:
I might want collecting to pandas for plotting, so let's figure out how to completely avoid spark when going to Pandas.


johnc1231:
Temporarily closing

PR #11113, opened 2021-12-02
[deploy_config] Add domain support to hailctl config

jigold:
I added the capability for the deploy config to find the domain from setting it in the config.ini file. This way users only use `hailctl config set domain` rather than `hailctl dev config set domain`. In addition, we use this new capability to make a test in Batch work on Azure.

CC: @danking

### DIFF
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -1,4 +1,4 @@
-from typing import Callable, TypeVar, Awaitable, Optional, Type, List, Dict, Any
+from typing import Callable, TypeVar, Awaitable, Optional, Type, List, Dict
 from types import TracebackType
 import concurrent
 import subprocess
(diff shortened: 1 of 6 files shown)


hail/python/hailtop/utils/utils.py
```diff
@@ -59,6 +59,13 @@ def first_extant_file(*files: Optional[str]) -> Optional[str]:
     return None
 
 
+def first_defined_value(*values: Optional[Any]) -> Optional[Any]:
```
danking, line 62: If you use the type var `T`, mypy will have an easier time guessing types.

hail/python/hailtop/utils/utils.py
```diff
@@ -1,4 +1,4 @@
-from typing import Callable, TypeVar, Awaitable, Optional, Type, List, Dict
+from typing import Callable, TypeVar, Awaitable, Optional, Type, List, Dict, Any
```
danking: remove any

danking (changes requested):

danking (approved):
once you fix the import, it will pass and merge

PR #11115, opened 2021-12-02
[auth] Support create_initial_user in dev deploy for dev namespaces

daniel-goldstein:
These are the alterations that Jackie and I made to be able to add a developer account to a dev namespace. If the namespace isn't default then it grabs the credentials for the user from default instead of creating a new credential. I don't like that the extra config lives on the branch but this is how it is otherwise supported in bootstrap.py, and ultimately we should just use the auth account to hit the auth endpoint and delete the `create_initial_account` script entirely.

### DIFF
--- a/auth/auth/driver/driver.py
+++ b/auth/auth/driver/driver.py
@@ -307,6 +307,7 @@ class K8sNamespaceResource:
         self.name = name
 
     async def create(self, name):
+        assert name != 'default'
         assert self.name is None
 
         await self._delete(name)
@@ -315,7 +316,7 @@ class K8sNamespaceResource:
         self.name = name
 
     async def _delete(self, name):
-        assert namespace != 'default'
+        assert name != 'default'
         try:
             await self.k8s_client.delete_namespace(name)
         except kube.client.rest.ApiException as e:
--- a/auth/auth/driver/driver.py
+++ b/auth/auth/driver/driver.py
@@ -315,6 +315,7 @@ class K8sNamespaceResource:
         self.name = name
 
     async def _delete(self, name):
+        assert namespace != 'default'
         try:
             await self.k8s_client.delete_namespace(name)
         except kube.client.rest.ApiException as e:
--- a/auth/auth/driver/driver.py
+++ b/auth/auth/driver/driver.py
@@ -547,7 +547,7 @@ async def delete_user(app, user):
         await hail_identity_secret.delete()
 
     namespace_name = user['namespace_name']
-    if namespace_name is not None:
+    if namespace_name is not None and namespace_name != DEFAULT_NAMESPACE:
         assert user['is_developer'] == 1
 
         # don't bother deleting database-server-config since we're
--- a/auth/auth/driver/driver.py
+++ b/auth/auth/driver/driver.py
@@ -307,7 +307,7 @@ class K8sNamespaceResource:
         self.name = name
 
     async def create(self, name):
-        assert name != 'default'
+        assert name not in ('default', DEFAULT_NAMESPACE)
         assert self.name is None
 
         await self._delete(name)
@@ -316,7 +316,7 @@ class K8sNamespaceResource:
         self.name = name
 
     async def _delete(self, name):
-        assert name != 'default'
+        assert name not in ('default', DEFAULT_NAMESPACE)
         try:
             await self.k8s_client.delete_namespace(name)
         except kube.client.rest.ApiException as e:
--- a/ci/create_initial_account.py
+++ b/ci/create_initial_account.py
@@ -1,9 +1,41 @@
 import argparse
+import base64
+import json
+import kubernetes_asyncio as kube
+import os
+
 from hailtop.utils import async_to_blocking
 from gear import Database, transaction
 
 
-async def insert_user_if_not_exists(db, username, email):
+NAMESPACE = os.environ['NAMESPACE']
+
+
+async def copy_identity_from_default(hail_credentials_secret_name: str) -> str:
+    cloud = os.environ['CLOUD']
+    await kube.config.load_kube_config()
+    k8s_client = kube.client.CoreV1Api()
+
+    secret = await k8s_client.read_namespaced_secret(hail_credentials_secret_name, 'default')
+
+    await k8s_client.create_namespaced_secret(
+        NAMESPACE,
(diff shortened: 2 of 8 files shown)


auth/auth/driver/driver.py
```diff
@@ -315,6 +315,7 @@ async def create(self, name):
         self.name = name
 
     async def _delete(self, name):
+        assert namespace != 'default'
```
jigold, line 318: I don't see where namespace is defined. I think you meant `HAIL_DEFAULT_NAMESPACE`. Also, should we assert the same thing for `create`?

jigold (changes requested):

auth/auth/driver/driver.py
daniel-goldstein: Oops, I think that got lost in the rebase. I'll add that though.

auth/auth/driver/driver.py
```diff
@@ -307,6 +307,7 @@ def __init__(self, k8s_client, name=None):
         self.name = name
 
     async def create(self, name):
+        assert name != 'default'
```
jigold, line 310: What is your reason for not making this stronger? I don't think you want to try and create and delete the namespace you are currently running in. I was thinking `assert name not in ('default', HAIL_DEFAULT_NAMESPACE)`

jigold (changes requested):

ci/create_initial_account.py
```diff
@@ -12,19 +44,37 @@ async def insert(tx):
                 return None
             return row['id']
 
+        hail_credentials_secret_name = f'{username}-gsa-key'
```
jigold, line 47: Should this be None if we're in default? This is a change in behavior from the current code.

jigold (changes requested):

ci/create_initial_account.py
daniel-goldstein: You're right, this is different. I'll keep the same behavior as default.

hail/python/hailtop/hailctl/dev/deploy/cli.py
```diff
@@ -16,6 +16,8 @@ def init_parser(parser):
                         help="Comma or space-separated list of steps to run.", required=True)
     parser.add_argument("--excluded_steps", "-e", nargs='+', action='append', default=[],
                         help="Comma or space-separated list of steps to forcibly exclude. Use with caution!")
+    parser.add_argument("--extra-config", "-c", nargs="+", action='append', default=[],
+                        help="Comma or space-separate list of key=value pairs to add as extra config parameters.")
```
danking, line 20: this can't be space separated. `unpack_comma_delimited_inputs` splits on `,`. I suspect you're trying to indicate that you can do `-c x=y -c z=a`. Maybe:

> Extra configuration key-value pairs. Format as: key=value. May specify multiple pairs separated by commas.

danking (dismissed):

hail/python/hailtop/hailctl/dev/deploy/cli.py
daniel-goldstein: I don't think this is true. Here's `unpack_comma_delimited_inputs`:

```python
    return [s.strip()
            for steps in inputs
            for step in steps
            for s in step.split(',') if s.strip()]
```

If you have `-s deploy_batch,deploy_ci , test_ci, test_batch_0 -s test_batch_1`, you'll get an `inputs` list of `[['deploy_batch,deploy_ci', ',', 'test_ci,', 'test_batch_0'], 'test_batch_1']`, which will split accordingly:

```python
In [4]: def unpack_comma_delimited_inputs(inputs):
   ...:     return [s.strip()
   ...:             for steps in inputs
   ...:             for step in steps
   ...:             for s in step.split(',') if s.strip()]
   ...:

In [5]: unpack_comma_delimited_inputs([['deploy_batch,deploy_ci', ',', 'test_ci,', 'test_batch_0'], ['test_batch_1']])
   ...:
Out[5]: ['deploy_batch', 'deploy_ci', 'test_ci', 'test_batch_0', 'test_batch_1']

In [6]:
```

daniel-goldstein: Certainly non-obvious and I had to run it in ipython to be very sure, but I always space-delimit by dev deploys, not comma

daniel-goldstein: I would honestly rather just space delimit everything, the code would be far easier to reason about and the example I showed is very ugly, but that would be breaking

danking:
I have no objections but I'll let @jigold  approve.

hail/python/hailtop/hailctl/dev/deploy/cli.py
jigold: I didn't know you could use the same flag multiple times, so I coded this for this case which is probably bad. So I'm okay if you change my code:

`-c x=y,a=b,c=d`

jigold: I've been using `-e deploy_auth,create_accounts` with dev deploy

daniel-goldstein: What do you want me to change it to? What you wrote is consistent with the other dev deploy options, and I would above all prefer that they be consistent with each other.

jigold: What makes sense to you and is consistent.

daniel-goldstein: > What makes sense to you and is the right way to do it.

I'm not sure how to interpret this sentence. I said I'd prefer to stay consistent with the other CLI options, which this PR currently is.

auth/auth/driver/driver.py
```diff
@@ -315,6 +316,7 @@ async def create(self, name):
         self.name = name
 
     async def _delete(self, name):
+        assert name not in ('default', DEFAULT_NAMESPACE)
```
jigold, line 319: On line 550 in this file, we need to change this line to this:

```
if namespace_name is not None and namespace_name != DEFAULT_NAMESPACE
```

This way we can delete a developer from a namespace.

jigold (changes requested):

jigold (approved):

PR #11118, opened 2021-12-02
[auth] Add current Azure display name to CLI and UI

jigold:
I still need to test this works with dev deploy.


danking (dismissed):
LGTM. We could just merge and check in default. No one is bothered if it breaks.

jigold:
This is not actually working.

jigold:
I'll revisit this once #11115 is merged.

danking (changes requested):
dismiss when you're ready for a new look

jigold:
Closing in favor of #11147

PR #11122, opened 2021-12-02
[hailgenetics] automate building and publishing of hailgenetics/python-dill

danking:
Previously python-dill lived both at gcr.io/hail-vdc/python-dill and gcr.io/hail-vdc/hailgenetics/python-dill
and hailgenetics/python-dill. This PR eliminates the first home because it is incongruous with
the other hailgenetics images. I also duplicated the Dockerfiles because that is the easiest
way to create multiple images in build.yaml.


docker/python-dill/build-and-push.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -ex
+
+for version in 3.6 3.6-slim 3.7 3.7-slim 3.8 3.8-slim
+do
+    python3 ../../../ci/jinja2_render.py '{"global":{"docker_prefix":'${DOCKER_PREFIX}'}}' python-dill/Dockerfile.$version python-dill/Dockerfile.$version.rendered
+    ../../docker-build.sh . Dockerfile.$version.rendered $(DOCKER_PREFIX)/hailgenetics/python-dill:$version
+    DOCKER_BUILDKIT=1 docker tag $(DOCKER_PREFIX)/hailgenetics/python-dill:$version docker.io/hailgenetics/python-dill:$version
```
jigold, line 9: Doesn't this assume whomever uses this script has access to our dockerhub hailgenetics? Couldn't someone want to run this script in their own project and not push the image to docker hub? Or is this the purpose of putting it in build.yaml?

jigold (changes requested):

build.yaml
```diff
@@ -3732,13 +3794,39 @@ steps:
...
+                             /io/github-oauth
+
+      retry() {
+          "$@" ||
+              (sleep 2 && "$@") ||
+              (sleep 5 && "$@");
+      }
+
+      retry skopeo copy {{ hailgenetics_hail_image.image }} docker.io/hailgenetics/hail:$HAIL_PIP_VERSION
+      retry skopeo copy {{ hailgenetics_hail_image.image }} gcr.io/hail-vdc/hailgenetics/hail:$HAIL_PIP_VERSION
```
daniel-goldstein, line 3811: The `gcr.io/hail-vdc` should be `{{ global.docker_prefix }}`

danking: ah yes.

docker/python-dill/build-and-push.sh
danking: Yeah, I think this should be hail-team only. Third parties should use build.yaml.

danking:
OK, yeah, I need a coherent strategy for someone deploying a fresh Hail install.

If folks are not continuously deploying every commit, they might miss a tagged commit. @daniel-goldstein , thoughts on how we might support things that should be run on a tagged commit for users who are not deploying every commit?

daniel-goldstein:
> If folks are not continuously deploying every commit, they might miss a tagged commit. @daniel-goldstein , thoughts on how we might support things that should be run on a tagged commit for users who are not deploying every commit?

I don't have a good plan for this at the moment and have not spent much time thinking about it. I will say though that I expect not deploying on every commit while running CI to be painful given our current build process. Sometimes we make "backwards compatible" changes (to CI or build.yaml) through multiple PRs that would not be backward compatible if those commits were squashed. We should definitely have a discussion about it, probably with AUS folks since they might be able to offer insight as to what they're planning with deploys.

daniel-goldstein:
> If folks are not continuously deploying every commit, they might miss a tagged commit.

Do you mean someone watching `hail-is/hail:main` but only deploying occasionally? I feel like we need to separate out publish steps entirely from our CD steps, track our most recently published tag/sha, and spin off publish batches when the tag is no longer the most recent on GitHub.

daniel-goldstein:
This all would be a lot easier to reason about if at the top of CI it showed a DEPLOY batch and a PUBLISH batch.

danking:
> 

Yeah this is what I had in mind. Agreed, a separate publish vs deploy flow could help. Though, it doesn't address the backwards compatibility issue you mentioned. Ideally we could get all of this right rather than just fixing part of the problem.

daniel-goldstein:
Not sure I see a fix for the backward compatibility problem other than releasing/versioning our build setup. I know #10911 eliminated a large class of backwards incompatible changes to CI, and I have a hunch that there are few scenarios that are inherently backwards incompatible, we just need a better approach. I think a good first step would be to start tracking any PRs that need to be split up into 2 for compatibility reasons.

danking:
I'll think about this more, but making the CI version explicit rather than implicit would at least provide a clear progression of necessary deploys. We'd want to tag the CI versions in git so that folks know which commits are necessary to achieve the step-wise transition.

Let's find some time to chat next next week.

danking:
Or if the transitions were explicit the way our database versions are explicit, that would help.

build.yaml
danking: OK, so here's my plan:

I'll make a separate step which copies into `{{ global.docker_prefix }}` (my ultimate goal is for python-dill to be tagged with hail version and include hail). I'll keep the docker.io/hailgenetics/hail stuff in deploy.

danking: Hmm. this ain't right.

danking:
need more thought

PR #11123, opened 2021-12-02
[batch] Use premium SSD data disks in Azure

jigold:


jigold:
Bump. This needs to go into production before we do massive scale tests since Standard_SSD charges per IO operation.

jigold:
Actually, this seems to cause instances to hang when using an actual data disk. I'm going to put the WIP tag on until I get a chance to investigate

danking (changes requested):
dismiss when you're ready for a new look

jigold:
I'm not sure if we need a Premium SSD on the os disk or not.

jigold:
See question, but otherwise tests are passing.

batch/batch/cloud/azure/resource_utils.py
```diff
@@ -90,7 +89,7 @@ def azure_worker_properties_to_machine_type(worker_type: str, cores: int, local_
         return f'Standard_F{cores}s_v2'
 
     if local_ssd_data_disk:
-        additive_features = 'd'
+        additive_features = 'ds'
```
jigold, line 92: d means use a local SSD data disk. s means support premium disks.

danking (dismissed):
x

danking (approved):

PR #11127, opened 2021-12-03
[batch] Delete dead token code

daniel-goldstein:
I moved these functions into methods on the worker API in `instance_env.py` and forgot to delete the old file.


danking:
I am increasingly jealous of your lines deleted count.

PR #11128, opened 2021-12-03
[bootstrap] Publish python-dill images to container registry

daniel-goldstein:
Currently `docker/python-dill` images are untested and deployed manually, but used in tests and by our users. #11122 aims to both test and automatically publish those images on every release, but to move forward testing in Azure for now this pushes those images as part of the bootstrapping process.

Also removed the `export HAIL` line because operators should be able to define where their hail home is.

### DIFF
--- a/infra/bootstrap_utils.sh
+++ b/infra/bootstrap_utils.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-export HAIL="$HOME/hail"
-
 get_global_config_field() {
     kubectl get secret global-config --template={{.data.$1}} | base64 --decode
 }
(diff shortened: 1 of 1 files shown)


infra/bootstrap_utils.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-export HAIL="$HOME/hail"
```
jigold, line 3: Can we have some sort of assert that $HAIL is defined?

jigold (changes requested):

jigold (approved):

PR #11136, opened 2021-12-05
[batch] skip disk mount test in azure because it hangs

danking:


danking:
example, which I cancelled: https://batch.daniel-azure.hail.is/batches/3251/jobs/1

daniel-goldstein (approved):

jigold:
This is fine for now to get things moving, but the underlying problem needs to be fixed this week. It's not acceptable to have our system hang because a user requests too much storage.

PR #11142, opened 2021-12-09
[batch] fix job private jobs

danking:
AFAIK, we do not need a distinct name anyway


danking:
This PR causes all the batch tests to pass!

jigold (approved):

PR #11143, opened 2021-12-09
[azure] clean up deployments

danking:
Sure it is not this easy?


danking:
we have a better solution forthcoming

PR #11147, opened 2021-12-09
[auth] Add support for Azure Active Directory oauth flows

jigold:
I think it's going to be too hard to test this PR in dgoldste. I think we should combine getting this PR to work with standing up the Azure production cluster. Until then, I need your help to finish this draft:

@danking -- Can you look over the oauth flow?
@daniel-goldstein -- Next week, can you help out with the necessary Terraform changes?

### DIFF
--- a/auth/auth/auth.py
+++ b/auth/auth/auth.py
@@ -200,6 +200,7 @@ async def callback(request):
     caller = session['caller']
     next_page = session.pop('next', nb_url)
     flow_dict = session['flow']
+    flow_dict['callback_uri'] = deploy_config.external_url('auth', '/oauth2callback')
     cleanup_session(session)
 
     try:
@@ -323,7 +324,11 @@ async def rest_login(request):
     else:
         assert CLOUD == 'gcp'
         host = '127.0.0.1'
-    flow_data = request.app['flow_client'].initiate_flow(f'http://{host}:{callback_port}/oauth2callback')
+
+    callback_uri = f'http://{host}:{callback_port}/oauth2callback'
+    flow_data = request.app['flow_client'].initiate_flow(callback_uri)
+    flow_data['callback_uri'] = callback_uri
+
     return web.json_response({'flow': flow_data})
 
 
--- a/auth/auth/auth.py
+++ b/auth/auth/auth.py
@@ -573,7 +573,7 @@ async def on_startup(app):
     await db.async_init(maxsize=50)
     app['db'] = db
     app['client_session'] = httpx.client_session()
-    app['flow_client'] = get_flow_client('/auth-oauth2-client-secret/client_secret.json')
+    app['flow_client'] = get_flow_client('/auth-oauth2-client-secret/client_secret_v2.json')
 
 
 async def on_cleanup(app):
--- a/auth/auth/auth.py
+++ b/auth/auth/auth.py
@@ -327,7 +327,9 @@ async def rest_login(request):
     callback_port = request.query['callback_port']
     flow = get_flow(f'http://127.0.0.1:{callback_port}/oauth2callback')
     flow_data = flow.initialize_flow()
-    return web.json_response({'authorization_url': flow_data.authorization_url, 'state': flow_data.state, 'flow': flow.as_dict()})
+    return web.json_response(
+        {'authorization_url': flow_data.authorization_url, 'state': flow_data.state, 'flow': flow.as_dict()}
+    )
 
 
 @routes.get('/roles')
@@ -449,7 +451,10 @@ async def rest_callback(request):
 
     db = request.app['db']
     users = [
-        x async for x in db.select_and_fetchall("SELECT * FROM users WHERE login_id = %s AND state = 'active';", flow_result.login_id)
+        x
+        async for x in db.select_and_fetchall(
+            "SELECT * FROM users WHERE login_id = %s AND state = 'active';", flow_result.login_id
+        )
     ]
 
     if len(users) != 1:
--- a/auth/auth/auth.py
+++ b/auth/auth/auth.py
@@ -10,6 +10,7 @@ from hailtop.config import get_deploy_config
 from hailtop.tls import internal_server_ssl_context
 from hailtop.hail_logging import AccessLogger
 from hailtop.utils import secret_alnum_string
+from hailtop import httpx
 from gear import (
     setup_aiohttp_session,
     rest_authenticated_users_only,
@@ -162,17 +163,16 @@ async def signup(request):
     next_page = request.query.get('next', deploy_config.external_url('notebook', ''))
 
     flow = get_flow(deploy_config.external_url('auth', '/oauth2callback'))
-
-    authorization_url, state = flow.authorization_url_and_state()
+    flow_data = flow.initialize_flow()
 
     session = await aiohttp_session.new_session(request)
     cleanup_session(session)
-    session['state'] = state
+    session['state'] = flow_data.state
     session['next'] = next_page
     session['caller'] = 'signup'
     session['flow'] = flow.as_dict()
 
-    return aiohttp.web.HTTPFound(authorization_url)
+    return aiohttp.web.HTTPFound(flow_data.authorization_url)
 
 
 @routes.get('/login')
@@ -180,18 +180,17 @@ async def login(request):
     next_page = request.query.get('next', deploy_config.external_url('notebook', ''))
 
     flow = get_flow(deploy_config.external_url('auth', '/oauth2callback'))
-
-    authorization_url, state = flow.authorization_url_and_state()
+    flow_data = flow.initialize_flow()
 
     session = await aiohttp_session.new_session(request)
 
     cleanup_session(session)
-    session['state'] = state
+    session['state'] = flow_data.state
     session['next'] = next_page
     session['caller'] = 'login'
     session['flow'] = flow.as_dict()
 
-    return aiohttp.web.HTTPFound(authorization_url)
+    return aiohttp.web.HTTPFound(flow_data.authorization_url)
 
 
 @routes.get('/oauth2callback')
@@ -211,8 +210,8 @@ async def callback(request):
     flow = get_flow(deploy_config.external_url('auth', '/oauth2callback'), state=state)
 
     try:
-        token = flow.fetch_and_verify_token(request, session.get('flow'))
-        login_id, email = flow.login_id_email_from_token(token)
+        flow_result = flow.finish_flow(request, session.get('flow'))
+        login_id = flow_result.login_id
     except asyncio.CancelledError:
         raise
     except Exception as e:
@@ -230,7 +229,7 @@ async def callback(request):
 
         assert caller == 'signup'
 
-        username, domain = email.split('@')
+        username, domain = flow_result.email.split('@')
         username = ''.join(c for c in username if c.isalnum())
 
         if domain != ORGANIZATION_DOMAIN:
@@ -327,8 +326,8 @@ async def logout(request, userdata):
 async def rest_login(request):
     callback_port = request.query['callback_port']
     flow = get_flow(f'http://127.0.0.1:{callback_port}/oauth2callback')
-    authorization_url, state = flow.authorization_url_and_state()
-    return web.json_response({'authorization_url': authorization_url, 'state': state, 'flow': flow.as_dict()})
+    flow_data = flow.initialize_flow()
+    return web.json_response({'authorization_url': flow_data.authorization_url, 'state': flow_data.state, 'flow': flow.as_dict()})
 
 
 @routes.get('/roles')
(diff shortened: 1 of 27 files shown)


infra/gcp/main.tf
```diff
@@ -269,6 +270,7 @@ resource "kubernetes_secret" "global_config" {
     internal_ip = google_compute_address.internal_gateway.address
```
danking, line 270: lets remove gsuite organization as completely as possible.

infra/azure/main.tf
```diff
@@ -418,13 +418,24 @@ resource "azuread_application" "auth" {
...
+    }
+  }
 }
 module "auth_sp" {
   source = "./service_principal"
   application_id = azuread_application.auth.application_id
   object_id      = azuread_application.auth.object_id
 }
 
+# FIXME: Create secret /auth-oauth2-client-secret/client_secret.json which is just a copy of the auth service principal credentials
```
danking, line 437: So, k8s stuff goes in ../k8s/main.tf. There's several examples of `resource kuberenets_secret` there. Pursuant to my other comment, this should be a fresh app registration (not auth's) and not a user's service principal. As such, it should not use the `for_each` thing that `gsa_key` does.

danking (changes requested):
Looks great! Just a few things.

auth/auth/flow.py
```diff
@@ -0,0 +1,125 @@
...
+
+class AzureFlow(Flow):
+    @staticmethod
+    def get_flow(credentials_file: str, redirect_uri: str, state: Optional[str] = None) -> 'AzureFlow':
+        with open(credentials_file) as f:
+            data = json.loads(f.read())
+
+        tenant_id = data['tenant']
+        authority = f'https://login.microsoftonline.com/{tenant_id}'
+        client = msal.ConfidentialClientApplication(data['appId'], data['password'], authority)
```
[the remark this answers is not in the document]
jigold: I used the IDE to help me. I knew the `msal` library was what we needed, but as you say the docs/examples are not very helpful. Then I saw which Client class was the correct one for our use case (there were only two to choose from). Then I saw what methods were available and looked at the docstrings for each one. I saw one method `initiate_auth_code_flow` which was similar to what we were using for Google's oauth flow and had the same outputs (state, authorization_url). The documentation is this:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/8bdb1ef8a3ee2ba3da0c50090683c555d7020198/msal/application.py#L625

and this for the finalizer which also had what we needed (email):

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/8bdb1ef8a3ee2ba3da0c50090683c555d7020198/msal/application.py#L791

The last part was just figuring out how to glue the pieces together and I think the right way is putting the "flow" dict into the session based on this example:

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/8bdb1ef8a3ee2ba3da0c50090683c555d7020198/msal/application.py#L830-L839

infra/azure/main.tf
```diff
@@ -406,6 +406,22 @@ resource "azurerm_role_assignment" "ci_test_container_contributor" {
 
 provider "azuread" {}
 
+resource "azuread_application" "oauth2" {
```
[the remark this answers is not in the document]
daniel-goldstein: I suppose there's no harm in creating a new application and that app will have fewer privileges. This is fine.

daniel-goldstein (changes requested):
Terraform looks fine except a couple comments about the GCP module. The terraform is in a difficult to understand transitory state right now, which is why refactoring it is a high priority for me. I have all that ready to go and can stack it on top of this change, so I think that if you've validated the oauth in azure manually we can merge this and I'll start spinning up prod instead of running this against the current cluster.

infra/azure/main.tf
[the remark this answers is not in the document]
jigold: This is what Dan requested.

jigold:
It's failing in Azure in `create_accounts` because the credentials secrets are all missing the new field `appObjectId`

daniel-goldstein:
Terraform looks good to me

auth/auth/flow.py
```diff
@@ -0,0 +1,123 @@
...
+
+        flow = client.initiate_auth_code_flow(scopes=[], redirect_uri=redirect_uri, state=state)
```
danking, line 86: This should happen in `initialize_flow`, not in this constructor. We don't want to initiate an auth flow the second time we call `get_flow` in the oauth2callback.

auth/auth/flow.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class FlowResult:
+    def __init__(self, login_id: str, email: str, token: dict):
+        self.login_id = login_id
+        self.email = email
+        self.token = token
+
+
+class Flow(abc.ABC):
```
danking, line 27: OK, I've thought about this a bit more and I think we can streamline these classes a bit more.

Before this PR a login session had a `state` parameter. After this PR a login session will have a `state` parameter and a `flow` parameter. The `flow` parameter is only used by Azure. The value of the `state` parameter in Azure is duplicated in the `flow` parameter. My conclusion: we should just have one `flow` session parameter which is a dictionary. In Google, it will have one field: `state`. In Azure, it will have whatever is returned by `initiate_auth_code_flow`. This removes a few lines across the code base (woo!).

Below, I note that in Azure we unnecessarily initiate the auth flow the second time we call `get_flow` (the call to `get_flow` in oauth2callback). I think this mistake was made easier by the structure of this class. Let's instead make the structure of the class mirror the steps of the oauth2 flow:
```python3
class Flow(abc.ABC):
    @abc.abstractmethod
    def initiate_flow(redirect_uri: str) -> dict:
        """Initiates the OAuth2 flow. Usually run in response to a user clicking a login button.

           The returned dict should be stored in a secure session so that the server can
           identify to which OAuth2 flow a client is responding. In particular, the server must
           pass this dict to :meth:`.receive_callback` in the OAuth2 callback."""
        raise NotImplementedError

    @abc.abstractmethod
    def receive_callback(self, flow_dict: dict) -> FlowResult:
        """Concludes the OAuth2 flow by returning the user's identity and credentials."""
        raise NotImplementedError
```

Note that there should be exactly one `Flow` class for the entire lifetime of the server. It will need, I think, just one field: `credentials_file`.

In auth.py, I think this produces substantial simplifications, for example login and signup both only need one line for OAuth2 related activity:
```python3
sessoin['flow'] = app['the_oauth2_flow'].initiate_flow(deploy_config.external_url('auth', '/oauth2callback'))
```
And oauth2callback is likewise simplified to two lines:
```python3
flow_result = app['the_oauth2_flow'].receive_callback(session['flow'])
login_id = flow_result.login_id
```

danking (changes requested):

infra/azure/main.tf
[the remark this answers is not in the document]
danking: Conceptually, the auth app registration is the auth robot user's identity within Batch. The fact that an app registration can both be an identity and an OAuth2 participant is, IMHO, confusing. I think it's simpler to treat them as distinct: the auth app registration is *just* its service principal. The oauth2 app registration is *just* an OAuth2 participant. The privileges granted to the auth *service principal* have nothing to do with the ability to verify someone's identity.

auth/auth/flow.py
```diff
@@ -0,0 +1,114 @@
...
+            'authorization_url': authorization_url,
+            'redirect_uri': redirect_uri,
+            'state': state,
+        }
+
+    def receive_callback(self, request: aiohttp.web.Request, flow_dict: dict) -> FlowResult:
+        flow = google_auth_oauthlib.flow.Flow.from_client_secrets_file(
+            self._credentials_file, scopes=self.scopes, state=flow_dict['state']
+        )
+        flow.redirect_uri = flow_dict['redirect_uri']
```
jigold, line 64: This is a change from the previous behavior. This reuses the redirect uri from the initial initialize flow call. In the past, this would have been hard coded to `deploy_config.external_url('auth', '/oauth2callback')` in `/oauth2callback`

auth/auth/auth.py
```diff
@@ -341,11 +318,13 @@ async def logout(request, userdata):
...
-
-    flow = get_flow(f'http://127.0.0.1:{callback_port}/oauth2callback')
-    authorization_url, state = flow.authorization_url(access_type='offline', include_granted_scopes='true')
-
-    return web.json_response({'authorization_url': authorization_url, 'state': state})
+    if CLOUD == 'azure':
+        host = 'localhost'
+    else:
+        assert CLOUD == 'gcp'
+        host = '127.0.0.1'
```
danking, line 325: This shouldn't differ between clouds? `127.0.0.1` is defined as the loopback address ([RFC1122 3.2.1.3 (g)](https://datatracker.ietf.org/doc/html/rfc1122#page-31) and [RFC6890 2.2.2](https://datatracker.ietf.org/doc/html/rfc6890#page-7)). Did 127.0.0.1 not work when using the Hail CLI to log into Azure?

auth/auth/flow.py
danking: Huh. Certainly this seems better than hardcoding it. It does seem like it needs to be set because its ultimately used in [`fetch_token`](https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth2_session.py#L321). 🤷‍♀️

danking: I bet it just has to be one of the valid redirect URIs for your application, but it doesn't matter which one. Or maybe requests-oauthlib is just unnecessarily sending extra data?

danking (changes requested):
just nits now

auth/auth/auth.py
jigold: Azure won't let you use `127.0.0.1`. It breaks their formatting requirements for a URL. I can try again.

hail/python/hailtop/hailctl/auth/login.py
```diff
@@ -69,7 +72,8 @@ async def auth_flow(deploy_config, default_ns, session):
             params={
                 'callback_port': port,
                 'code': code,
-                'state': state
+                'state': state,
```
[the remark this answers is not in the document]
jigold: This is required. I get this error if I don't include it in the request back to the oauth2callback.

```
  File "/usr/loca
l/lib/python3.7/dist-packages/auth/auth.py", line 439, in rest_callback
    flow_result = request.app['flow_client'].receive_callback(request, flow_dict)
  File "/usr/local/lib/python3.7/dist-packages/auth/flow.py", line 103, in receive_callback

  token = self._client.acquire_token_by_auth_code_flow(flow_dict['flow'], query_dict)
  File "/usr/local/lib/python3.7/dist-packages/msal/application.py", line 841, in acquire_token_by_auth_code_flow
    **kwargs))
  File "/usr/local/lib/python3.7/
dist-packages/msal/application.py", line 136, in obtain_token_by_auth_code_flow
    auth_code_flow, auth_response, headers=headers, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/msal/oauth2cli/oidc.py", line 205, in obtain_token_by_auth_c
ode_flow
    auth_code_flow, auth_response, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/msal/oauth2cli/oauth2.py", line 543, in obtain_token_by_auth_code_flow
    auth_code_flow.get("state"), auth_response.get("state")))
ValueError
: state mismatch: XXXXXXXX vs None
```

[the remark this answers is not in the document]
jigold: This is in Azure.

auth/auth/auth.py
jigold: This is the error you get. The URL has to match exactly

Screen Shot 2021-12-15 at 12 12 12 PM

jigold: > AADSTS50011: The reply URL specified in the request does not match the reply URLs configured for the application:

jigold:
https://developers.google.com/identity/protocols/oauth2/web-server#exchange-authorization-code

So the google oauth flow example shows this when getting the token:

```python3
flow.redirect_uri = flask.url_for('oauth2callback', _external=True)
```

Let me see if I can incorporate this behavior instead of what I did.

auth/auth/auth.py
daniel-goldstein: Huh interesting, the portal requires that the reply URL use https unless the domain is specfically `localhost`. I tried to add 127.0.0.1 under https, does that work? Seems we should be able to use either the IP or domain for both clouds.

daniel-goldstein: ah right we run a local server for hailctl auth login

jigold: If I change the Google one to 'localhost', I get a similar error message. The option is to keep this code or I go and duplicate all the callbacks in GCP to use localhost instead.

jigold:
Addressed comments.

auth/auth/auth.py
danking: Man, this chaps my ass. Let's add a comment noting that this is specifically because azure does not support `http://127.0.0.1`. Otherwise, I'm definitely going to forget and try to change this in a year.

hail/python/hailtop/hailctl/auth/login.py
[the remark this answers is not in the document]
danking: Huh... well ok.

danking (approved):
OK, I have no remaining issues with this PR. Thanks for teaching me about Azure's auth libraries! I put the WIP tag on because I think we have to use localhost for everyone, not 127.0.0.1 for everyone. If we change to localhost for google, we have to update the valid Redirect URIs in our google oauth2 client registration.

PR #11148, opened 2021-12-09
[batch] Clarify DockerHub error message

daniel-goldstein:
and only use user credentials to pull from same cloud


danking (approved):
love `hosted_in`

PR #11151, opened 2021-12-13
[query] Check for distinct keys when writing files to disk

johnc1231:
This is part one of an effort to do joins more efficiently in the common case where keys are distinct. This PR:

1. Adds a flags component to allow us somewhere to write flags like "isDistinctlyKeyed"
2. Does an analysis when writing out Tables to determine if the keys are in fact distinct.


tpoterba (changes requested):
said this in 1:1 but just so we remember after the holidays :)

johnc1231:
Ok, I've addressed your comment, corrected requiredness inference for `PartitionNativeWriter` and `SplitPartitionWriter` and revamped things so that the key is not determined based on whether or not there is an index. I've also bumped file version and regenerated files. 

I'd be interested to see what you think about testing / whether there's more testing you would do. This is harmless change right now if all the normal tests pass since it's just adding a new metadata field that's unused save for a few tests, but we don't want to start marking files "distinctlyKeyed" that aren't and then run into problems down the road when we implement the new join behavior.

tpoterba:
great, will look!

hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
```diff
@@ -158,17 +164,23 @@ case class TableNativeWriter(
   }
 }
 
-case class PartitionNativeWriter(spec: AbstractTypedCodecSpec, partPrefix: String, index: Option[(String, PStruct)], localDir: Option[String]) extends PartitionWriter {
+case class PartitionNativeWriter(spec: AbstractTypedCodecSpec, optKeyType: Option[TStruct], partPrefix: String, index: Option[(String, PStruct)], localDir: Option[String]) extends PartitionWriter {
```
tpoterba, line 167: can the key just be an `IndexedSeq[String]`?

tpoterba (changes requested):
also need to change the n_files assertion in test_backward_compatibility (this exists so that failures to find the files don't result in silent successes).

That test took ~3m to run -- we should break it up, but that can be done in a separate PR.

hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
johnc1231: So that would then be indistinguishable from keying a table by no fields. If I have a table with a single row that has an empty key, should we say it's distinctly keyed? I guess it doesn't really matter since you can't join on such a thing.

Yeah, I'll change it to IndexedSeq[String]

tpoterba: I think an unkeyed table is a table keyed by no fields, yes. Can't join on that so distinctness doesn't matter.

johnc1231:
Addressed comments. My question about it being sufficiently tested still stands.

tpoterba (changes requested):
Yeah, you're right about testing. I think we can add a test in Scala that writes a distinctly keyed and non-distinctly-keyed table (plus one with edge case where there's a duplicate key spread across partitions) and asserts that the property is what we expect.

johnc1231:
I think the current Scala tests I've written roughly do that

tpoterba (approved):
oh, missed that among the new compatibility files! looks good.

PR #11153, opened 2021-12-14
[dataproc] fix notebooks

danking:
mistune 2.0.0 was released which breaks nbconvert. nbconvert 5.6.0 introduced
appropriate pinning. https://github.com/jupyter/nbconvert/commit/365683f40f85dc9304b6f1c23ca35eeb4d63b597.


danking:
We need to release after this. I suspect release will be rocky due to some deploy changes I made recently, so make sure I'm online when we do it.

tpoterba (approved):

PR #11154, opened 2021-12-14
[vds] Add `filter_chromosomes`

tpoterba:


hail/python/hail/vds/methods.py
```diff
@@ -443,6 +443,101 @@ def filter_variants(vds: 'VariantDataset', variants_table: 'Table', *, keep: boo
...
+    if mode == 'variants_only':
+        variant_data = hl.filter_intervals(vds.variant_data, intervals, keep)
+        return VariantDataset(vds.reference_data, variant_data)
+    if mode == 'split_at_boundaries':
+        if not keep:
+            raise ValueError("filter_intervals mode 'split_at_boundaries' not implemented for keep=False")
+        par_intervals = hl.Table.parallelize([hl.Struct(interval=x) for x in intervals],
+                                             schema=hl.tstruct(interval=intervals.dtype.element_type), key='interval')
+        ref = segment_reference_blocks(vds.reference_data, par_intervals)
+        return VariantDataset(hl.filter_intervals(ref, intervals, keep),
```
patrick-schultz, line 464: Is the filter on the segmented ref blocks needed?

patrick-schultz (changes requested):
Just one question

hail/python/hail/vds/methods.py
```diff
@@ -443,6 +443,101 @@ def filter_variants(vds: 'VariantDataset', variants_table: 'Table', *, keep: boo
     return VariantDataset(vds.reference_data, variant_data)
 
 
+@typecheck(vds=VariantDataset,
+           intervals=expr_array(expr_interval(expr_any)),
+           keep=bool,
+           mode=enumeration('variants_only', 'split_at_boundaries', 'unchecked_filter_both'))
+def _parameterized_filter_intervals(vds: 'VariantDataset',
```
chrisvittal, line 450: What is the point of having this be a function at all? It's hidden from the public API and used in one place. Furthermore, the modes are short and seem like they can be inlined where necessary.

chrisvittal (changes requested):

hail/python/hail/vds/methods.py
tpoterba: ah, you've pointed out that I failed to call this from `hl.vds.filter_intervals`!

hail/python/hail/vds/methods.py
tpoterba: you're right, it's not.

patrick-schultz (approved):

PR #11156, opened 2021-12-15
[ci] Track all hail CI statuses and merge only when they all pass

daniel-goldstein:
What level of confidence do we need in this? Definitely not something our CI test suite supports.

### DIFF
--- a/ci/ci/github.py
+++ b/ci/ci/github.py
@@ -1,5 +1,5 @@
+from enum import Enum
 from typing import Dict, Optional
-from typing_extensions import Literal, get_args
 import secrets
 from shlex import quote as shq
 import json
@@ -34,7 +34,10 @@ zulip_client = zulip.Client(config_file="/zulip-config/.zuliprc")
 TRACKED_PRS = pc.Gauge('ci_tracked_prs', 'PRs currently being monitored by CI', ['build_state', 'review_state'])
 
 
-GithubStatus = Literal['success', 'pending', 'failure']
+class GithubStatus(Enum):
+    SUCCESS = 'success'
+    PENDING = 'pending'
+    FAILURE = 'failure'
 
 
 def select_random_teammate(team):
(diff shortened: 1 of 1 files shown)


ci/ci/github.py
```diff
@@ -385,18 +387,19 @@ async def _update_github(self, gh):
...
-            f'github sent multiple status summaries for our one '
-            f'context {GITHUB_STATUS_CONTEXT}: {hail_status}\n\n{statuses_json}'
-        )
+        hail_statuses = {}
+        for s in statuses:
+            context = s['context']
+            if context == GITHUB_STATUS_CONTEXT or context.startswith('hail-ci'):
+                if context in hail_statuses:
+                    raise ValueError(
+                        f'github sent multiple status summaries for our one '
```
danking, line 398: this message is mislead now. maybe just remove "our one"

ci/ci/github.py
```diff
@@ -385,18 +387,19 @@ async def _update_github(self, gh):
...
+                        f'context {context}: {s}\n\n{statuses_json}'
+                    )
+                hail_statuses[context] = s['state']
+        return hail_statuses
```
danking, line 402: this can't possibly type check. Are we not running mypy on this file?

danking (changes requested):
Seems right to me. We lack a testing strategy for this, so I usually think hard before merging changes and monitor it a bit afterwards.

ci/ci/github.py
daniel-goldstein: I don't fully understand this either, but I ran mypy by hand and it does seem to be ok with this. I can add an assert to make this more clear?

danking: It seems the issue is that statuses_json is `Any` so all its children are `Any` and can be used for anything you like.

If you add a `:str` type annotation to context and status, mypy appropriately complains.

daniel-goldstein: Ah, I would be surprised that it's so permissive of Any, I would have assumed that would behave the same as str in this context.

danking (changes requested):
let's use enum

ci/ci/github.py
```diff
@@ -566,9 +572,14 @@ def is_up_to_date(self):
         return self.batch is not None and self.target_branch.sha == self.batch.attributes['target_sha']
 
     def is_mergeable(self):
+        if self.last_known_github_status[GITHUB_STATUS_CONTEXT] == 'success':
```
danking, line 575: heh, mypy isn't checking this method because of the lack of any type signature (try adding `-> bool`). This needs to be `== GithubStatus.SUCCESS`. Same below in the `all(..)` expression.

danking (dismissed):

PR #11159, opened 2021-12-15
[ci] Catch cloud-incompatible steps at build config time

daniel-goldstein:


ci/ci/build.py
```diff
@@ -92,40 +92,39 @@ def __init__(self, code, config_str, scope, *, requested_step_names=(), excluded
...
+            self.steps = [step for step in all_steps if step in visited]
+        else:
+            self.steps = [step for step in all_steps if not step.run_if_requested and step.can_run_in_current_cloud()]
 
     def build(self, batch, code, scope):
         assert scope in ('deploy', 'test', 'dev')
 
         for step in self.steps:
-            if (step.scopes is None or scope in step.scopes) and (step.clouds is None or CLOUD in step.clouds):
+            if step.can_run_in_scope(scope):
```
danking, line 127: let's add an assert that it can run in current cloud, the logic is currently nonlocal and thus a bit hard to follow.

danking (changes requested):

danking (approved):

PR #11160, opened 2021-12-15
[batch] Add updating provisioning state in Azure

jigold:
Looking at the logs, I think these two new states are because we added the log analytics agent based on when the PR merged and the absence of these errors before December 10th.

```
Unknown azure statuses [{'code': 'ProvisioningState/updating', 'level': 'Info', 'displayStatus': 'Updating'}, {'code': 'PowerState/running', 'level': 'Info', 'displayStatus': 'VM running'}] for instance batch-worker-default-standard-166xu
```


daniel-goldstein (dismissed):

danking (approved):
I'm OK with this. I'm not sure what Daniel's message means. I've added WIP. We can discuss in January.

jigold:
bump @daniel-goldstein

jigold:
Ah, I see you're good with this change.

PR #11168, opened 2021-12-16
[compiler] Fix bad export_vcf filters error message

tpoterba:


johnc1231:
This doesn't compile

johnc1231:
Should we not just crash when someone does this? It feels like this ought to be a python enforced type error?

johnc1231 (approved):

PR #11169, opened 2021-12-16
[hailtop] add new transient error

danking:


jigold:
The docker image doesn't have aiodocker in it.

jigold (approved):

PR #11174, opened 2021-12-16
[batch] Add GC loop for Azure deployments

jigold:


hail/python/hailtop/aiocloud/aioazure/client/arm_client.py
```diff
@@ -10,3 +12,21 @@ def __init__(self, subscription_id: str, resource_group_name: str, **kwargs):
             params['api-version'] = '2021-04-01'
         super().__init__(f'https://management.azure.com/subscriptions/{subscription_id}/resourceGroups/{resource_group_name}/providers/Microsoft.Resources',
                          **kwargs)
+
+    async def _paged_get(self, path, **kwargs) -> AsyncGenerator[Any, None]:
```
danking, line 16: This already exists on the AzureResourcesClient. Move it to AzureBaseClient

danking (changes requested):

hail/python/hailtop/aiocloud/aioazure/client/arm_client.py
jigold: I moved it since all the clients we have now have the same paging format. The activity logs had different names for the JSON schema, but we can worry about that later.

danking (approved):

PR #11175, opened 2021-12-16
[infra] Break up Azure terraform into modules

daniel-goldstein:
This diff is pretty inscrutable but this entire PR is just a reorganization to better make use of terraform modules. The only things that are not shifting code into different modules and invoking those modules are:

- Made the default AKS node pool the nonpreemptible pool instead of having two custom pools in addition to default
- Added subscription/resource group level roles to be added to the service_principal module which removed a fair amount of repetitive role blocks.
- The azure module directly invokes Kubernetes resources instead of managing two separate states (one for azure and one for k8s). This simplifies the process to one `terraform apply` instead of two and removes the need for the janky copying of outputs to inputs that happened in bootstrap.sh

Sorry for the nearly 1k line PR but I think maintaining this terraform is going to be easier going forward and better to be done atomically.


danking (approved):
As discussed, I'm OK with this change. We can review in detail in January.

PR #11176, opened 2021-12-16
[grafana] Make deployable in Azure with Makefile

daniel-goldstein:
The Makefile was hardcoded for GCP. Grafana is still currently able to be deployed through CI, this just changes the Makefile so it's not hardcoded to `gcp` and moves the gcp project from a Makefile-set value to one set through kubernetes so the Makefile is more cloud agnostic (doesn't reference a gcp project).

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1923,6 +1923,7 @@ steps:
     config: grafana/deployment.yaml
     scopes:
       - deploy
+      - test
       - dev
     dependsOn:
       - default_ns
--- a/grafana/deployment.yaml
+++ b/grafana/deployment.yaml
@@ -60,6 +60,13 @@ spec:
           - name: GF_SERVER_ROOT_URL
             value: "%(protocol)s://%(domain)s/{{ default_ns.name }}/grafana/"
 {% endif %}
+{% if global.cloud == 'gcp' %}
+          - name: GCP_PROJECT
+            valueFrom:
+              secretKeyRef:
+                name: global-config
+                key: gcp_project
+{% endif %}
          volumeMounts:
           - mountPath: /var/lib/grafana
             name: grafana-storage
@@ -138,7 +145,7 @@ data:
         isDefault: true
         jsonData:
           authenticationType: gce
-          defaultProject: {{ global.gcp_project }}
+          defaultProject: $GCP_PROJECT
         editable: true
 {% endif %}
       - name: Prometheus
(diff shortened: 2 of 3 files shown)


build.yaml
```diff
@@ -1923,6 +1923,7 @@ steps:
     config: grafana/deployment.yaml
     scopes:
       - deploy
+      - test
```
jigold, line 1926: Was there a reason for not including this scope initially? Maybe something about needing disks?

grafana/deployment.yaml
```diff
@@ -138,7 +145,7 @@ data:
         isDefault: true
         jsonData:
           authenticationType: gce
-          defaultProject: {{ global.gcp_project }}
+          defaultProject: $GCP_PROJECT
```
jigold, line 148: I'm trying to figure out why this change would work. Can you explain it to me?

jigold (changes requested):

build.yaml
daniel-goldstein: My guess is that this was removed along with the `test` scope for prometheus, which was removed iirc because of it needs an entire k8s node for itself. But based off this PR run it looks like grafana doesn't need prometheus to exist for it to successfully start. We might even be able to deploy prometheus in tests too and just change the resource requirements for test deploys, since it'll be starting from scratch and not have to load a ton of data in PR runs. I can try that out too if you want.

grafana/deployment.yaml
daniel-goldstein: I just learned that grafana allows [environment variables](https://grafana.com/docs/grafana/latest/administration/provisioning/#using-environment-variables) in their datasource templates and it will fill them in when grafana loads the yaml. So as long as `GCP_PROJECT` is set as an environment variable in the grafana container it should just work.

build.yaml
daniel-goldstein: hm going back in the git history it looks like this just never had the `test` scope

jigold: I'm all for testing grafana in CI, but I do want to be mindful of resources. If you can give it sensible parameters in testing, then this seems fine to me. If we don't like the behavior, then we can revert.

daniel-goldstein: It's requesting a modest amount of resources, and its disk is only 10Gi. I think it should have a negligible impact on our testing resources. Prometheus however gives me more pause, since it accumulates a lot of data quickly and I have yet to test how modest we can be within the timeline that the PR deployment would live. Maybe I'll save testing prometheus for the new year and just stick to grafana for now?

jigold: that seems fine.

jigold (approved):

PR #11180, opened 2021-12-17
[hailctl hdinsight] start a Spark cluster with SparkMonitor and Hail

danking:
Azure already has a Jupyter system in place, so I worked within that. As a result, I took a very different approach from `hailctl dataproc`. I'm not sure how many of the configuration settings done in `hailctl dataproc` are necessary in Azure. I also do not plan to add special support for any special parameters from Azure. If a user wants to, for example, configure auto-scaling, they can use pass through arguments.

There are three files that need to be hosted somewhere: two startup scripts and an Azure-specific wheel file. For the startup scripts, I just rely on GitHub tagged raw files. For the wheel file, I placed it in hail common and use the public HTTP endpoint. For development, you have to manually upload the files you want to override and invoke `hailctl hdinsight` like this:
```
hailctl hdinsight
        start \
        clustername \
        password \
        password \
        storageaccount \
        --install-hail-uri https://raw.githubusercontent.com/danking/hail/dk-hdinsight-test/hail/python/hailtop/hailctl/hdinsight/resources/install-hail.sh \
        --install-native-deps-uri https://raw.githubusercontent.com/danking/hail/dk-hdinsight-test/hail/python/hailtop/hailctl/hdinsight/resources/install-native-deps.sh \
        --wheel-uri https://storage.googleapis.com/hail-common/dking/hail-0.2.79-py3-none-any.whl
```
We could make this easier, but I'd rather spend that time on the query service.

cc: @jigold


hail/python/hailtop/hailctl/__main__.py
```diff
@@ -98,6 +101,9 @@ def main():
         elif module == 'dataproc':
             from hailtop.hailctl.dataproc import cli  # pylint: disable=import-outside-toplevel
             cli.main(args)
+        elif module == 'hdinsight':
```
tpoterba, line 104: 🔥

hail/python/hail/backend/spark_backend.py
```diff
@@ -135,14 +135,20 @@ def __init__(self, idempotent, sc, spark_conf, app_name, master,
...
                 jars.append(os.path.join(os.path.dirname(sparkmonitor.__file__), 'listener.jar'))
                 conf.set("spark.extraListeners", "sparkmonitor.listener.JupyterSparkMonitorListener")
 
             conf.set('spark.jars', ','.join(jars))
-            conf.set('spark.driver.extraClassPath', ','.join(jars))
-            conf.set('spark.executor.extraClassPath', './hail-all-spark.jar')
+            if os.environ.get('AZURE_SPARK') == '1':
+                print('AZURE_SPARK environment variable is set to "1", assuming you are in HDInsight.')
+                # Setting extraClassPath in HDInsight overrides the classpath entirely so you can't
+                # load the Scala standard library. Interestingly, setting extraClassPath is not
```
tpoterba, line 147: huh, weird. How do we pick up the sparkmonitor jar then?

danking: So, this reminds me of a frustration I had years ago. The Spark docs stated that adding a JAR to the `spark.jars` configuration parameter *should* be sufficient to get the JAR onto the class path. In practice, this was not true. In particular, we added this bizarre `spark.executor.extraClassPath=./hail-all-spark.jar` setting because we knew the JAR file would be copied into the worker directory of the executor.

HDInsight seems to be broken in a slightly different way: the "extra" class path is actually the "whole" class path. Luckily the `spark.jars` parameter works as advertised in HDInsight (it both copies the JAR and modifies the class path).

danking: Erm, huh, I realize now I answered a different question. I have no idea how we find the spark monitor JAR.

danking:
closing in favor of the whole shebang: https://github.com/hail-is/hail/pull/11187

PR #11183, opened 2021-12-20
[curl] pervasively use retries, report errors, follow redirects

danking:
I set a `.curlrc` file to ensure we use the right options by default. In a few places that are
intended to be run on a developer's laptop, I had to explicitly specify the necessary
options.

I replaced all the short flags with long flags, the important conversions to know:
- `-L` `--location`, follow redirects
- `-O` `--remote-name`, use the basename of the URL as the name of a local file in which to save the data
- `-s` `--silent`, do not print anything to stdout/stderr
- `-S` `--show-error`, when used with `-s`, shows errors on stderr
- `-f` `--fail`, return non-zero exit code if the response is 4xx or 5xx (normally the response would be printed / saved and curl will exit successfully)


jigold:
Can you rebase this before I look at it?

danking:
done

docker/genetics/Dockerfile
```diff
@@ -11,28 +11,28 @@ RUN hail-apt-get-install \
...
-     curl -LO https://github.com/samtools/samtools/releases/download/1.9/samtools-1.9.tar.bz2 && \
+     curl --remote-name https://github.com/samtools/samtools/releases/download/1.9/samtools-1.9.tar.bz2 && \
      tar -xf samtools-1.9.tar.bz2 && \
      rm -rf samtools-1.9.tar.bz2 && \
      cd samtools-1.9 && \
      ./configure --without-curses && \
      make && \
      make install)
-RUN curl -LO http://s3.amazonaws.com/plink2-assets/plink2_linux_avx2_latest.zip && \
+RUN curl --remote-name http://s3.amazonaws.com/plink2-assets/plink2_linux_avx2_latest.zip&& \
```
jigold, line 21: Missing a space after `.zip`

jigold (approved):

jigold (dismissed):
See comment.

jigold (approved):

PR #11184, opened 2021-12-20
[hail] loosen requirements

danking:
A user requested we loosen requirements on fsspec and tqdm. TQDM appears to
follow semver (https://github.com/tqdm/tqdm/issues/33). GCSFS is one of the
packages that specifically caused us versioning problems before. It does not
claim a versioning policy anywhere, but they do now pin fsspec versions, so
I think we can trust them to not break in that way, at least.


tpoterba (changes requested):
this doesn't match your comment, which suggests that we would remove the fsspec explicit dependency, and loosen the gcsfs one. Am I reading that wrong?

tpoterba (approved):

PR #11186, opened 2021-12-22
[vds] Fix lgt_to_gt/splitmulti to ignore LA if LGT is reference

tpoterba:


tpoterba:
After making a tiny change to LGT, I chased a thread of deviations between `n_singleton` in `hl.vds.sample_qc` and `hl.sample_qc`. The difference had to do with counting singleton ref alleles -- hl.sample_qc counted singleton ref alleles if the ref was the only allele in the dense matrix, hl.vds.sample_qc was counting singleton ref alleles if the ref allele was the only allele in the sparse variant data. 

I changed both sample_qc algorithms to define n_singleton as never counting reference alleles.

chrisvittal (approved):

PR #11192, opened 2022-01-04
[auth] Use client_secret instead of client_secret_v2

daniel-goldstein:
I can apply the terraform to azure after this merges and new auth pods are running.


jigold (approved):

jigold:
You can also get rid of the extra secret in GCP hail-vdc as well.

PR #11193, opened 2022-01-04
[ci] Dont add requested step dependencies that are run_if_requested

daniel-goldstein:
I mistakenly assumed in #11159 that _all_ dependencies of a requested step should run, but in fact `run_if_requested` steps can be dependents of normal steps.

### DIFF
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -95,12 +95,13 @@ class BuildConfiguration:
         if requested_step_names:
             log.info(f"Constructing build configuration with steps: {requested_step_names}")
 
-        all_steps: List[Step] = []
+        runnable_steps: List[Step] = []
         name_step: Dict[str, Step] = {}
         for step_config in config['steps']:
             step = Step.from_json(StepParameters(code, scope, step_config, name_step))
-            name_step[step.name] = step
-            all_steps.append(step)
+            if step.name not in excluded_step_names and step.can_run_in_current_cloud():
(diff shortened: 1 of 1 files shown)


ci/ci/build.py
```diff
@@ -112,7 +112,8 @@ def visit_dependent(step: Step):
                         raise BuildConfigurationError(f'Step {step.name} cannot be run in cloud {CLOUD}')
                     visited.add(step)
                     for s2 in step.deps:
-                        visit_dependent(s2)
+                        if not s2.run_if_requested:
+                            visit_dependent(s2)
```
danking, line 116: Hmm, I think this is right because if it is requested, we'll see it in the requeted_step_names loop.

I think we now have duplicate logic though. Check out build.py:159. `name_step[k]` previously had three possible states:
1. key not present, which means invalid step name
2. key mapped to None, which means step is skipped for some reason
3. key mapping to Step, which means step is active

Let's get rid of the if statement from 159, if it is no longer necessary.

danking (changes requested):

ci/ci/build.py
daniel-goldstein: hmph, I'm starting to see now why `name_step` used to be this value-optional dictionary. I still don't love that the logic is non-local, I'll try to rework it.

danking (approved):
This seems right to me. Would be nice to have more testing around this.

PR #11194, opened 2022-01-04
Query on Batch

danking:
# Overview

Query on Batch fundamentally eliminates Spark in favor of Hail Batch. Each Hail pipeline starts a
single job batch which acts as the "driver" or "compiler". The driver executes new batches whenever
it encounters a parallelizeAndComputeWithIndex. You can try this branch out in your developer
namespace by executing this:

```
hailctl dev config set default_namespace default
hailctl dev deploy --branch danking/hail:nqs-copy --steps delete_batch_tables,deploy_batch
make -C query ipython
```

The last command compiles and uploads a JAR to GCS. These JARs are deleted in 24 hours. If you have
already uploaded a JAR for theee current SHA in the last 24 hours, you can use `make -C query
ipython-no-deps` to start a new session without uploading a new JAR.

You can interact at ipython like this:

```
In [1]: import hail as hl
   ...: hl.utils.range_table(10).collect()
Initializing Hail with default parameters...
/Users/dking/projects/hail/hail/python/hail/utils/java.py:54: UserWarning: When using the query service backend, use `await Env._async_hc()'
  warnings.warn('When using the query service backend, use `await Env._async_hc()\'')
Welcome to
     __  __     <>__
    / /_/ /__  __/ /
   / __  / _ `/ / /
  /_/ /_/\_,_/_/_/   version 0.2.81-edcde5c1b324
LOGGING: writing to /Users/dking/projects/hail/query/hail-20220113-1844-0.2.81-edcde5c1b324.log
Out[1]: 
[Struct(idx=0),
 Struct(idx=1),
 Struct(idx=2),
 Struct(idx=3),
 Struct(idx=4),
 Struct(idx=5),
 Struct(idx=6),
 Struct(idx=7),
 Struct(idx=8),
 Struct(idx=9)]
```

The very first time you execute this comand, it will run four batches to generate the four default
reference genomes and download those to your local machine. Those four reference genomes are cached
per-SHA on your local machine, so future Hail pipelines will have lower latency.

If you have the "standing working" enabled, you should expect a latency of ~8s for the above job. I
think we can approximately halve this by re-using classloaders. Ask me more about this.

After running this job a few times, the Batch UI looks like this:

Screen Shot 2022-01-14 at 11 59 41 AM

When you call `collect`, the Python-side `ServiceBackend` writes a file to cloud storage containing
the temporary directory, billing project, and the IR. In addition to the `execute` command used for
`collect`, there are command for getting the table type and references
genomes. `ServiceBackend._rpc` defines this API on the Python-side and
`is.hail.backend.service.ServiceBackendSocketAPI2` defines this API on the Scala-side.

After writing to cloud storage, the ServiceBackend submits the one-job driver batch to Hail
Batch. It then waits for the driver job to complete. When the driver job is finished, it reads the
outputs from google cloud storage and returns the results to the user.

All the meaningful changes to batch are in the worker. The worker starts one jvm per core on
startup. The mainclass is a new class called `JVMEntryway`. This entryway starts a UNIX socket and
speaks a very simple binary protocol. It accepts only one type of message:

```
int32 the number of strings to expect
(
int32 the number of bytes in the next string
byte* UTF-8 string
)*
```

The array of strings is interpreted as:

```
comma-spearated-classpath
main-class-name
arg0
arg1
...
```

The entryway constructs a URLClassLoader with the given classpath, reflectively allocates an
instance of the mainclass and invokes the `main` method with the remaining arguments. This is
obviously a security risk. The system bans JARs from locations not controlled (and locked down) by
Hail Team. You should require me to hardcode the mainclass as
`is.hail.backend.service.ServiceBackendSocketAPI2` before we merge; however, this flexibility was
useful during development.

The JVMEntryway will eventually be useful because we will keep a ClassLoader full of a bunch of
JIT-optimized Hail classes. I did not include that in this PR because we need to finish eliminating
global state used by Hail. Currently, two executions would try to re-use compiled class names for
different code, leading to very weird errors.

# Changes to File Systems

Hail has three four file system interfaces:

| File System Interface   | Public | Language | Async |
| ----------------------- | ------ | -------- | ----- |
| hail.utils.hadoop_utils | Yes    | Python   | no    |
| hail.fs                 | Yes    | Python   | no    |
| hailtop.aiotools.fs     | No     | Python   | yes   |
| is.hail.io.fs           | No     | Scala    | no    |

`hail.fs` is technically in the public API (via `hl.current_backend().fs`), but I doubt anyone uses
it. `hail.utils.hadoop_utils` is a shim over `hail.fs`, there are no direct concrete implementations of
it.

This PR adds `hail.fs.RouterFS` to `hail.fs`, a synchronous wrapper around
`hailtop.aiotools.fs.AsyncRouterFS`. A "router" file system is one which operates on URLs instead of
paths. It uses the URL's protocol to determine which concrete file system to use. For example, a
router fs can `open` both `gs://danking/abc` and `s3a://danking/abc`.

Each Hail Query Python Backend is associated with one file system class. This PR associates the
ServiceBackend with `RouterFS`, enabling `hl.current_backend().fs.open`, `hl.hadoop_open`, etc. to
read from S3, GCS, ABS, and the local file system. We should deprecate `hail.utils.hadoop_utils`
because it is not Hadoop-specific. We should instead advertise the class-based `hail.fs` or create a
new function-based interface (e.g. `hl.fs.open(...)`.

# Test Clean-up

The Hail Query local and spark tests should now work in Azure. I moved all the `hail.fs` and
`hailtop.aiotools.fs` tests into two build.yaml steps: `test_hail_python_fs` and
`test_hail_scala_fs`. These tests are exhaustive: they test every file system: S3, ABS, and GCS.

The only file system tests that remain in the Hail Query tests are the tests of
`hail.utils.hadoop_utils`. The hadoop tests are not exhaustive: they only test the *current* file
system. In Azure, they test ABS. In Google, they test GCS.

I have not decided yet if we should enable the hail python tests in Azure. It seems mostly wasteful.

# Local Cache

I added a local cache directory. It defaults to `$XDG_CONFIG_HOME/hail/cache` or
`~/.config/hail/cache` if `XDG_CONFIG_HOME` is not set. I store Python reference genome metadata
here.

# Batch Attributes

The ServiceBackend `batch_attributes` attribute specifies the attributes for any batch created by
the ServiceBackend. I modified the tests so that the test function name is use the "name" of the
batch. When a Hail Query driver job executes `parallelizeAndComputeWithIndex`, it uses its name with
a unique suffix as the name of the batch of worker jobs.

# Changes to Hail Scala

I think the only changes outside of `is.hail.backend.service` and `is.hail.services` are:

1. More retries in GoogleStorageFS.

2. Make MatrixSpecHelper and TableSpecHelper serializable.


hail/python/hail/context.py
```diff
@@ -98,9 +125,20 @@ def __init__(self, log, quiet, append, tmpdir, local_tmpdir,
         if global_seed is None:
             global_seed = 6348563392232659379
         Env.set_seed(global_seed)
+        Env._hc = self
+
+    def initailize_references(self, references, default_reference):
```
tpoterba, line 130: typo: initialize

danking:
I have a follow-up PR that ~halves the run-time of the service backend tests:
https://github.com/hail-is/hail/pull/11212

batch/batch/worker/worker.py
```diff
@@ -251,6 +251,7 @@ async def create_netns(self):
 
     async def enable_iptables_forwarding(self):
         await check_shell(
+            # FIXME: this just accumulates on the host, right?
```
daniel-goldstein, line 254: Not really. When we recycle network namespaces, we delete and recreate the namespace but reuse the same range of IPs and therefore the iptables rules. We reserve 128 network namespaces (64 each for public/private) at worker startup and don't `init` any more after that, so this is run a fixed number of times. We don't clean them up, but the rules do not grow unbounded. The below method should probably not be called `cleanup`, as it also recreates the necessary components.

daniel-goldstein: The reason why we recycle the namespace is in case a bad actor set some nefarious iptables rules *inside* the namespace, but they should not be able to affect iptables on the host.

batch/batch/worker/worker.py
```diff
@@ -132,6 +135,10 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
 INTERNET_INTERFACE = os.environ['INTERNET_INTERFACE']
 UNRESERVED_WORKER_DATA_DISK_SIZE_GB = int(os.environ['UNRESERVED_WORKER_DATA_DISK_SIZE_GB'])
 assert UNRESERVED_WORKER_DATA_DISK_SIZE_GB >= 0
+ACCEPTABLE_QUERY_JAR_URL_PREFIX = os.environ['ACCEPTABLE_QUERY_JAR_URL_PREFIX']
+assert (
```
jigold, line 139: This doesn't seem like a very strong assert to me. The environment variable should always be a string.

batch/batch/worker/worker.py
jigold: How hard is it to up this number to 64*4 or 96*4. We might have larger workers in the future. Maybe it should just be N_CORES * 4 and not be hard coded.

batch/batch/worker/worker.py
```diff
@@ -1605,56 +1610,52 @@ def __init__(
...
-        user_command_string = job_spec['process']['command']
-        assert len(user_command_string) >= 3, user_command_string
-        self.revision = user_command_string[1]
-        self.jar_url = user_command_string[2]
-        classpath = f'{find_spark_home()}/jars/*:/hail-jars/{self.revision}.jar:/log4j.properties'
+        self.user_command_string = job_spec['process']['command']
+        assert len(self.user_command_string) >= 3, self.user_command_string
+        self.revision = self.user_command_string[1]
+        self.jar_url = self.user_command_string[2]
+        # self.classpath = f'{find_spark_home()}/jars/*:/hail-jars/{self.revision}.jar:/log4j.properties'
```
jigold, line 1617: delete.

batch/batch/worker/worker.py
```diff
@@ -1605,56 +1610,52 @@ def __init__(
...
                 log.info(f'{self}: initializing')
                 self.state = 'initializing'
 
-                os.makedirs(f'{self.scratch}/')
-
-                await check_shell_output(f'xfs_quota -x -c "project -s -p {self.scratch} {self.project_id}" /host/')
-                await check_shell_output(
-                    f'xfs_quota -x -c "limit -p bsoft={self.data_disk_storage_in_gib} bhard={self.data_disk_storage_in_gib} {self.project_id}" /host/'
-                )
+                # await check_shell_output(f'xfs_quota -x -c "project -s -p {self.scratch} {self.project_id}" /host/')
```
jigold, line 1651: delete. Also should we make storage requests illegal for JVMJobs?

batch/batch/worker/worker.py
```diff
@@ -1664,6 +1665,7 @@ async def run(self):
                     async with self.worker.jar_download_locks[self.revision]:
                         local_jar_location = f'/hail-jars/{self.revision}.jar'
                         if not os.path.isfile(local_jar_location):
+                            self.verify_is_acceptable_query_jar_url(self.jar_url)
                             async with await self.worker.fs.open(self.jar_url) as jar_data:
                                 await self.worker.fs.makedirs('/hail-jars/', exist_ok=True)
                                 async with await self.worker.fs.create(local_jar_location) as local_file:
```
jigold, line 1671: We need to shield the **entire** download operation here. See the change we had to make when downloading root fs.

batch/batch/worker/worker.py
```diff
@@ -1676,38 +1678,42 @@ async def run(self):
...
                 raise
             except Exception:
+                # FIXME: this can also be a Hail Query driver error, not a Hail Batch error
                 log.exception(f'while running {self}')
 
                 self.state = 'error'
                 self.error = traceback.format_exc()
                 await self.cleanup()
-            await self.cleanup()
+            else:
```
jigold, line 1695: Again, let's see what the logic is for making sure logs get uploaded when the job can be cancelled.

batch/batch/worker/worker.py
```diff
@@ -1717,21 +1723,22 @@ async def cleanup(self):
         log.info(f'{self}: cleaning up')
         try:
             await check_shell(f'xfs_quota -x -c "limit -p bsoft=0 bhard=0 {self.project_id}" /host')
-
```
jigold, line 1720: So the reason we can delete this is because there's no scratch directory for JVMJobs? I'm really confused about the storage story here. There's an xfsquota command existing above, but deleted in another place.

batch/batch/worker/worker.py
```diff
@@ -1789,6 +1796,323 @@ def __str__(self):
         )
 
 
+def write_int(writer: asyncio.StreamWriter, v: int):
```
jigold, line 1799: Is there anyway to move all of this new code into another file in the module? I already didn't like how long this file was getting.

batch/batch/worker/worker.py
```diff
@@ -1664,6 +1665,7 @@ async def run(self):
                     async with self.worker.jar_download_locks[self.revision]:
                         local_jar_location = f'/hail-jars/{self.revision}.jar'
                         if not os.path.isfile(local_jar_location):
+                            self.verify_is_acceptable_query_jar_url(self.jar_url)
```
jigold, line 1668: We had a bug in DockerJob here by checking only whether the file exists and not explicitly storing whether the file finished downloading successfully.

batch/batch/worker/worker.py
```diff
@@ -1789,6 +1796,323 @@ def __str__(self):
...
+class BufferedOutputProcess:
+    @classmethod
+    async def create(cls, *args, **kwargs):
+        assert 'stdout' not in kwargs
+        assert 'stderr' not in kwargs
+
+        process = await asyncio.create_subprocess_exec(
+            *args, **kwargs, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
+        )
+        stop_event = asyncio.Event()  # FIXME: this seems unused
```
jigold, line 1881: ???

batch/batch/worker/worker.py
```diff
@@ -1789,6 +1796,323 @@ def __str__(self):
...
+
+    def kill(self):
+        return self.process.kill()
+
+    @property
+    def returncode(self) -> Optional[int]:
+        return self.process.returncode
+
+    def close(self):
+        self.kill()
```
jigold, line 1918: Try - Finallys???

batch/batch/worker/worker.py
```diff
@@ -1789,6 +1796,323 @@ def __str__(self):
...
+    FINISH_ENTRYWAY_EXCEPTION = 1
+    FINISH_NORMAL = 2
+    FINISH_CANCELLED = 3
+    FINISH_JVM_EOS = 4
+
+    @classmethod
+    async def create_process(cls, socket_file: str) -> BufferedOutputProcess:
+        return await BufferedOutputProcess.create(
+            'java',
+            '-Xmx3500M',
```
jigold, line 1936: I'm a bit worried about the memory situation. How is memory shared amongst users? Should we require JVM jobs are "standard" with 2 cores or "highmem" (whatever the exact numbers work out to)?

batch/batch/worker/worker.py
```diff
@@ -1812,6 +2137,22 @@ def __init__(self, client_session: httpx.ClientSession):
         self.headers = None
         self.compute_client = None
 
+        self.jvm_initializer_task = asyncio.ensure_future(self._initialize_jvms())
+        self.jvms: List[JVM] = []
+
+    async def _initialize_jvms(self):
+        self.jvms = await asyncio.gather(*[JVM.create(i) for i in range(CORES)])
```
jigold, line 2144: Should this be `4 * CORES`? If not, should JVM Jobs have a minimum CPU of 1?

jigold (changes requested):
I looked at Python file system changes, Batch code, and build.yaml. I need to spend more time on the JVM class in the worker and reread your PR summary. You also pushed new changes since I started looking at the code, so ignore comments that are no longer relevant.

tpoterba (commented):
No issues with the stuff I understand. I do want to talk through the system design so I can review how the new scala code fits into that.

batch/batch/worker/worker.py
danking: added

batch/batch/worker/worker.py
danking: Ah, we should just make this the correct memory. Where can I get that from the worker?

batch/batch/worker/worker.py
danking: Ah, JVM jobs are always 1 core no matter what. I'll add to todo list to verify that in front end.

batch/deployment.yaml
```diff
@@ -111,15 +111,40 @@ spec:
          - name: STANDING_WORKER_MAX_IDLE_TIME_SECS
            value: "7200" # 2 hours
 {% else %}
-         - name: HAIL_SHOULD_CHECK_INVARIANTS
```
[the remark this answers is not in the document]
danking: This is *very very very very* slow when you have more than a handful of batches and it was constantly becoming a problem for me when developing. I guess we can re-enable for tests only (not dev) if it doesn't increase test time dramatically.

batch/deployment.yaml
```diff
@@ -111,15 +111,40 @@ spec:
...
+               name: global-config
+               key: test_storage_uri
+         - name: HAIL_QUERY_ACCEPTABLE_JAR_SUBFOLDER
+           value: "/{{ token }}/jars"
+{% elif scope == "dev" %}
+         - name: HAIL_QUERY_STORAGE_URI
+           valueFrom:
+             secretKeyRef:
+               name: global-config
+               key: test_storage_uri
```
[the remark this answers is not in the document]
danking: How do you figure? default should always use query_storage_uri.

danking (commented):
Responses so far, will keep going through. thanks!

batch/batch/worker/worker.py
jigold: xfs_quota is fine to use outside of a container. We set the quota on the host directories and then mount them as /io. There's nothing special. I think the code should just work.

batch/batch/worker/worker.py
jigold: I still think you need to shield it if multiple jobs might try and download the same jar version.

jigold: This is where the /exec not found bug was.

batch/batch/worker/worker.py
jigold: `job_spec['resources']['memory_bytes']`

daniel-goldstein (changes requested):
Mostly a worker.py pass, I'll also take a look at FS but want to get some comments in

batch/batch/worker/worker.py
```diff
@@ -1789,6 +1793,327 @@ def __str__(self):
...
+    async def new_connection(self):
+        while True:
+            try:
+                interim_output = self.process.retrieve_and_clear_output()
+                if len(interim_output) > 0:
+                    log.warning(f'{self}: unexpected output between jobs: {interim_output}')
+
+                return await asyncio.open_unix_connection(self.socket_file)
+            except ConnectionRefusedError:
+                log.warning(f'{self}: unexpected exit between jobs', extra=dict(output=self.process.output()))
```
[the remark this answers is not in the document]
daniel-goldstein: To clarify, it's not clear from a glance who you're talking to on the other end (trusted hail batch code, trusted hail query code, use hail query code)

batch/batch/worker/worker.py
danking: I'm on a random worker and I see this:
```
# iptables -L
...
Chain FORWARD (policy DROP)
target     prot opt source               destination         
DOCKER-USER  all  --  anywhere             anywhere            
DOCKER-ISOLATION-STAGE-1  all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere             ctstate RELATED,ESTABLISHED
DOCKER     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
DROP       all  --  172.21.0.0/16        metadata.google.internal 
ACCEPT     all  --  anywhere             10.128.0.57         
ACCEPT     all  --  anywhere             batch-worker-default-standard-2rw6o.c.hail-vdc.internal 
ACCEPT     all  --  anywhere            !10.128.0.0/16       
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
```

Doesn't that seem wrong? Do we know why we have 259 accepts from anywhere to anywhere on all protocols?

daniel-goldstein: That dump is a bit misleading because it omits network interfaces, and the default for IPs is anywhere. Try `iptables -L -v`. The rules I set work at the interface level to allow:
- The ethernet interface to forward packets to the job's network interface
- The job's network interface to forward packets to the ethernet interface
- The job's network interface to forward packets back to itself (a very special situation where we curl the worker's 10.128.x.x IP from inside the job container)

daniel-goldstein: I think this can be substantially cleaned up if we just allow eth -> job IP address subnet and job IP address subnet -> eth.

We'd still be left with job -> itself and I'm unsure how to specify "itself" in iptables rules. The main point of this is not allowing inter-job communication.

daniel-goldstein: Looks like rule number 2 there is already implemented as the general rule, so we only write rules 1 & 3 per network namespace. Since we create 128 network namespaces on each worker (64 private and 64 public), that makes 256 rules in addition to the few that we put in in create_instance

daniel-goldstein: I'll PR at least the # 1 fix

daniel-goldstein: See #11250

danking (commented):
Thanks Tim, Daniel, & Jackie for great comments so far! I addressed a bunch of comments. There's a couple outstanding conversations. I've run out of time today, but Monday I'll plan to get a inter-team master plan for QoB which captures any known issues with this PR that we plan to address later.

batch/batch/worker/worker.py
```diff
@@ -1717,21 +1720,22 @@ async def cleanup(self):
...
         except asyncio.CancelledError:
             raise
         except Exception:
             log.exception('while deleting volumes')
 
     async def get_log(self):
-        return {'main': self.logbuffer.decode()}
+        if self.log is not None:
+            return {'main': self.log}
+        return {'main': self.jvm.output()}
```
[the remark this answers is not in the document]
danking: In practice it loads pretty quickly and I'm hesitant to duplicate an entire Hail log.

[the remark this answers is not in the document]
danking: (b/c its currently held in memory)

batch/batch/worker/worker.py
```diff
@@ -1812,6 +2138,22 @@ def __init__(self, client_session: httpx.ClientSession):
...
+        self.jvms: List[JVM] = []
+
+    async def _initialize_jvms(self):
+        self.jvms = await asyncio.gather(*[JVM.create(i) for i in range(CORES)])
+        log.info(f'JVMs initialized {self.jvms}')
+
+    async def borrow_jvm(self) -> JVM:
+        await self.jvm_initializer_task
+        assert self.jvms
+        return self.jvms.pop()
```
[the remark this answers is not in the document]
danking: Hmm. I don't think a Queue makes sense because it's an error if a Job wants to borrow a JVM and the JVM does not exist. The design of this system is intended to ensure there are never more JVMJobs than JVMs on a machine.

Fair point about the global synchronization point. I guess I could make self.jvms a list of `Future[JVM]` and do a `gather(..., FIRST_COMPLETED)`. I'm open to this change if JVM startup time is variable and a bottleneck for Query on Batch.

hail/python/hail/context.py
danking: done

batch/batch/worker/worker.py
danking: Ah, great, the -v makes a world of difference.

batch/batch/worker/worker.py
danking: Actually, you'll need Java 11, the JVMEntryway.class file, and the junixsocket JAR. I think that plus the aforementioned mounts is sufficient.

batch/batch/worker/worker.py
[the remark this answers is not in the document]
daniel-goldstein: I mostly want to make sure that we're not accidentally logging anything sensitive from the user, so assuring that it could not be the case here.

danking:
QoB master plan did not fit into today. Might happen Wednesday.

danking:
OK, after a rather rocky period caused by my changes in response to comments, I anticipate quiescence tonight.

danking:
I would super appreciate another round of comments so we can get this PR merged around the end of the month.

hail/python/hail/backend/service_backend.py
```diff
@@ -1,157 +1,370 @@
...
         if typ == tvoid:
-            value = None
-        else:
-            value = typ._convert_from_json_na(resp['value'])
-        # FIXME put back timings
+            return None
+        return typ._convert_from_json_na(resp['value'])
 
-        return (value, None) if timed else value
+    def execute_many(self, *irs, timed=False):
```
tpoterba, line 312: Could wrap in a tuple and submit a single query, but I think this is really just used for the tests, right? If so then we should leave as-is to give each query the authentic isolation experience.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -70,8 +83,18 @@ class ServiceBackend(
...
+  def broadcast[T: ClassTag](_value: T): BroadcastValue[T] = {
+    using(new ObjectOutputStream(new ByteArrayOutputStream())) { os =>
+      try {
+        os.writeObject(_value)
+      } catch {
+        case e: Exception =>
+          fatal(_value.toString, e)
+      }
+    }
+    new BroadcastValue[T] with Serializable {
```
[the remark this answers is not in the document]
tpoterba: I agree with your understanding of the current code. However, I think this probably won't work for real-scale user pipelines and we should have a non-serializing approach in place for QoB release. 

Broadcasts are not used only for fs/RG, they'er also used for literals: https://github.com/hail-is/hail/blob/2ab242915c2c7d7b640074658fea6df06da9097b/hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala#L654

I think we'll also need to revisit how globals (broadcast values) are handled in the CDA emitter.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -82,146 +105,206 @@ class ServiceBackend(
...
   def stop(): Unit = ()
 
-  def valueType(username: String, s: String): String = {
-    ExecutionTimer.logTime("ServiceBackend.valueType") { timer =>
-      userContext(username, timer) { ctx =>
-        val x = IRParser.parse_value_ir(ctx, s)
-        x.typ.toString
-      }
-    }
+  def valueType(
```
tpoterba, line 211: I wonder if should support only a single execute endpoint by adding functions / ir nodes to look up the types and return them as strings. Might be a bit simpler. Maybe not.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -250,49 +333,33 @@ class ServiceBackend(
...
+    sessionId,
+    billingProject,
+    remoteTmpDir
+  ) { ctx =>
+    log.info(s"executing: ${token}")
+
+    execute(ctx, IRParser.parse_value_ir(ctx, code)) match {
+      case Some((v, t)) =>
+        JsonMethods.compact(
+          JObject(List("value" -> JSONAnnotationImpex.exportAnnotation(v, t.virtualType),
```
tpoterba, line 355: Fine for now, but we should switch to the encoded protocol John added pretty much immediately after this goes in.

hail/src/main/scala/is/hail/backend/service/Worker.scala
```diff
@@ -89,18 +101,21 @@ object Worker {
...
     timer.start("executeFunction")
 
-    val hailContext = HailContext(
-      // FIXME: workers should not have backends, but some things do need hail contexts
-      new ServiceBackend(null), skipLoggingConfiguration = true, quiet = true)
+    if (HailContext.isInitialized) {
+      HailContext.get.backend = new ServiceBackend(null, null, null)
+    } else {
+      HailContext(
+        // FIXME: workers should not have backends, but some things do need hail contexts
```
tpoterba, line 116: worker code shouldn't need this, I think. Let's follow up when the PR goes in to remove this.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
johnc1231: Agreed, and it shouldn't take long to switch I think

johnc1231:
Read through what I thought was relevant to me, and didn't have any concerns that Tim didn't point out. Exciting that this is working

batch/batch/worker/worker.py
```diff
@@ -1789,6 +1778,259 @@ def __str__(self):
...
+
+            if message == JVM.FINISH_NORMAL:
+                log.info(f'{self}: finished normally (interrupted: {wait_for_interrupt.done()})')
+            elif message == JVM.FINISH_CANCELLED:
+                assert wait_for_interrupt.done()
+                log.info(f'{self}: was cancelled')
+            elif message == JVM.FINISH_USER_EXCEPTION:
+                log.info(f'{self}: user exception encountered (interrupted: {wait_for_interrupt.done()})')
+                exception = await read_str(reader)
+                raise ValueError(exception)
```
daniel-goldstein, line 2023: I think this could still log higher up in the callstack, and I don't think we should log user exceptions, both from a noise and security POV. Maybe we should have a `UserError` exception and ensure that we never log exceptions of those types.

daniel-goldstein (changes requested):
Looks good to me, just a little concerned about what we log when we get exceptions from the JVM.

batch/src/main/java/is/hail/JVMEntryway.java
```diff
@@ -0,0 +1,222 @@
...
+                } catch (IllegalAccessException | InvocationTargetException e) {
+                  throw new RuntimeException(e);
+                } finally {
+                  Thread.currentThread().setContextClassLoader(oldClassLoader);
+                }
+              }
+            }, null);
+          shouldCancelThread = gather.submit(new Runnable() {
+              public void run() {
+                ClassLoader oldClassLoader = Thread.currentThread().getContextClassLoader();
```
jigold, line 105: Can you explain the resetting of class loaders and why that's important? I assume it's a security reason. But you're making a new thread each time so why is this important?

batch/src/main/java/is/hail/JVMEntryway.java
```diff
@@ -0,0 +1,222 @@
...
+      t.printStackTrace(pw);
+      return sw.toString();
+    }
+  }
+
+  private static int FINISH_USER_EXCEPTION = 0;
+  private static int FINISH_ENTRYWAY_EXCEPTION = 1;
+  private static int FINISH_NORMAL = 2;
+  private static int FINISH_CANCELLED = 3;
+
```
jigold, line 25: There's no JVM_EOS (4) here, but it's in the worker.py code.

hail/python/hail/backend/service_backend.py
```diff
@@ -1,157 +1,370 @@
...
+                await inputs(infile, token)
+
+            batch_attributes = self.batch_attributes
+            if 'name' not in batch_attributes:
+                batch_attributes = {**batch_attributes, 'name': name}
+            bb = self.async_bc.create_batch(token=token, attributes=batch_attributes)
+
+            j = bb.create_jvm_job([
+                'is.hail.backend.service.ServiceBackendSocketAPI2',
+                os.environ['HAIL_SHA'],
```
jigold, line 229: Where do these environment variables come from when a regular user is using the ServiceBackend?

hail/python/hail/backend/service_backend.py
```diff
@@ -1,157 +1,370 @@
...
+                os.environ['HAIL_JAR_URL'],
+                batch_attributes['name'],
+                iodir + '/in',
+                iodir + '/out',
+            ], mount_tokens=True)
+            b = await bb.submit(disable_progress_bar=self.disable_progress_bar)
+            try:
+                status = await b.wait(disable_progress_bar=self.disable_progress_bar)
+                (j, b, status) = (j, b, status)
```
jigold, line 238: What does this line do?

jigold (changes requested):
Mostly questions for my understanding.

batch/batch/worker/worker.py
```diff
@@ -1656,58 +1630,71 @@ async def run(self):
...
 
                 log.info(f'{self}: running jvm process')
                 with self.step('running'):
-                    self.process = await asyncio.create_subprocess_exec(
-                        *self.command_string,
-                        stdout=asyncio.subprocess.PIPE,
-                        stderr=asyncio.subprocess.PIPE,
-                        env={envvar['name']: envvar['value'] for envvar in self.env},
+                    await self.jvm.execute(
+                        f'{local_jar_location}', self.user_command_string[0], self.scratch, self.user_command_string[1:]
```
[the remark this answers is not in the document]
danking: Yeah, it is confusing and insecure. The command string is structured like this:

```
[mainclass, revision, jar_url, *more_args]
```

I should remove the main class and force everyone to go through one WorkerAPI. I'll do that. Then this line will change to something like:

```
await self.jvm.execute(local_jar_location, 'HailWorkerAPI', self.scratch, self.user_command_string)
```

batch/batch/worker/worker.py
```diff
@@ -1746,14 +1735,14 @@ async def delete(self):
...
+    #   resources: list of dict, {name: str, quantity: int},
+    #   jvm: str
     # }
     async def status(self):
         status = await super().status()
         status['container_statuses'] = dict()
         status['container_statuses']['main'] = {'name': 'main', 'state': self.state, 'timing': self.timings.to_dict()}
-        if self.process is not None and self.process.returncode is not None:
-            status['container_statuses']['main']['exit_code'] = self.process.returncode
+        status['jvm'] = self.jvm_name
```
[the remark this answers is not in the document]
danking: No, dictfix will not modify fields you don't tell it about. The UI doesn't display this field directly. This field is not/will-never-be sometimes optional so we shouldn't ever need to fix it.

batch/batch/worker/worker.py
danking: OK, I do not start JVMs except on standard or D

batch/batch/worker/worker.py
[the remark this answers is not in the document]
danking: Until we segregate logs per user, that will always be a risk. We should do that soon but I don't think it should gate this PR.

batch/batch/worker/worker.py
```diff
@@ -1789,6 +1778,259 @@ def __str__(self):
...
+            self.kill()
+        finally:
+            try:
+                self.stdout_pump.cancel()
+            finally:
+                self.stderr_pump.cancel()
+
+
+class JVM:
+    SPARK_HOME = find_spark_home()
```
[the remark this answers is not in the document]
danking: We set SPARK_HOME in the base image. `find_spark_home` either uses that environment variable or calls the python program `find_spark_home.py` which is a python program installed by `pyspark`. It returns the path to the spark directory. The spark directory contains a jars directory which contains the JARs which define the Spark classes. As long as Hail is entangled with Spark, we need to at least define those classes, even if we don't use them.

batch/batch/worker/worker.py
```diff
@@ -1789,6 +1778,259 @@ def __str__(self):
...
+            try:
+                interim_output = self.process.retrieve_and_clear_output()
+                if len(interim_output) > 0:
+                    log.warning(f'{self}: unexpected output between jobs')
+
+                return await asyncio.open_unix_connection(self.socket_file)
+            except ConnectionRefusedError:
+                log.warning(f'{self}: unexpected exit between jobs', extra=dict(output=self.process.output()))
+                os.remove(self.socket_file)
+                process, startup_output = await self.create_process_and_connect(self.index, self.socket_file)
```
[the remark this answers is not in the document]
danking: Yeah, if the JVM dies during a job or in between jobs, the subsequent job will restart the JVM.

batch/batch/worker/worker.py
danking: Agreed, the exception handling is a bit of a mess. It also doesn't get delivered correctly to the user. I've marked a to-do to follow up on this.

batch/batch/worker/worker.py
```diff
@@ -2186,7 +2447,7 @@ async def async_main():
     finally:
         try:
             await worker.shutdown()
-            log.info('worker shutdown')
+            log.info('worker shutdown', exc_info=True)
```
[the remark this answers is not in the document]
danking: If exceptions contain user data, then yes it can. We should ensure exceptions do not contain user data, not avoid logging exceptions on shutdown. If a worker unexpectedly shuts down, we need to know why. The lack of this information was an issue for me when debugging this PR.

batch/src/main/java/is/hail/JVMEntryway.java
danking: I added it here with a note that it should never be used here because the JVM cannot know that is unexpectedly sending EOS.

batch/src/main/java/is/hail/JVMEntryway.java
danking: It is not a security reason. It's just that some old/poorly-written Java/Scala code will look at `Thread.currentThread().getContextClassLoader()` when it needs to load a class. I want that code to be able to see the Hail classes that we loaded.

It is not a new thread each time. The `gather` is an `ExecutorCompletionService` which is just a way to wait on multiple Runnables at once. That `ExecutorCompletionService` sends Runnables to the `executor`. The `executor` is a fixed size thread pool with exactly two threads. Since I always set the current thread's context class loader on startup, I don't really need to revert it (the next Runnable will always set it properly), but it seems prudent.

hail/python/hail/backend/service_backend.py
danking: A regular user needs to set these environment variables for the service backend to work properly. When we're ready to actually release this to users, we'll need to be able to fall back on a sha and JAR url which are read from a file generated when we generate the wheel. I haven't decided exactly how to do that yet, but it will be something similar to how hailctl currently works.

hail/python/hail/backend/service_backend.py
danking: Nothing. Good point. I have removed it.

hail/python/hail/backend/service_backend.py
danking: Yes and agreed. This is really only useful for backends that can run queries in parallel.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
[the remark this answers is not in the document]
danking: OK, I have marked this in my todo list.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: I am game for whatever API you all like. The current one is not particularly onerous for me to support.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: I have marked this in my todo.

hail/src/main/scala/is/hail/backend/service/Worker.scala
danking: OK. I've added to my todo.

danking:
Thank you all for another round of detailed critique!

OK, I think the only remaining critical fix is to hard-code a mainclass. This is a wee bit complicated because I need to multiplex the ServiceBackendSocketAPI2 and the Worker. I hope to do this tomorrow AM. I'll then dismiss reviews.

I also have a list of todos generated by this process which will feedback into some master QoB doc that integrates the two teams necessary todos.

config.mk
```diff
@@ -1,13 +1,8 @@
...
-HAIL_TEST_GCS_BUCKET := hail-test-dmk9z
-DOMAIN := hail.is
+DOCKER_PREFIX := gcr.io/hail-vdc
 INTERNAL_IP := 10.128.0.57
 IP := 35.188.91.25
-KUBERNETES_SERVER_URL := https://104.198.230.143
-REGION := us-central1
-ZONE := us-central1-a
+DOMAIN := hail.is
+CLOUD := gcp
```
daniel-goldstein, line 5: revert this

danking:
OK, this PR basically works except we're encountering OOMs somewhat often. I'm trying to track down which tests are triggering the OOMs, but its a bit tricky because the OOMKiller doesn't necessarily kill the test which is using a ton of memory.

danking:
Huh! I seem to have reliably broken unicode ordering and/or en/de-coding. I've marked as fails for now. Hopefully all the remaining tests will now pass.
```
_________________________ Tests.test_unicode_ordering __________________________
[gw1] linux -- Python 3.7.12 /usr/bin/python3

self = <test.hail.table.test_table.Tests testMethod=test_unicode_ordering>

    def test_unicode_ordering(self):
        a = hl.literal(["é", "e"])
        ht = hl.utils.range_table(1, 1)
        ht = ht.annotate(fd=hl.sorted(a))
>       assert ht.fd.collect()[0] == ["e", "é"]
E       AssertionError: assert ['?', 'e'] == ['e', 'é']
E         At index 0 diff: '?' != 'e'
E         Full diff:
E         - ['e', 'é']
E         + ['?', 'e']

```

danking:
(FWIW: I can't replicate in my namespace when issuing commands from my laptop.)

PR #11197, opened 2022-01-05
[batch] Fix exec not found bug on the worker

jigold:
Not 100% sure this will solve the issue if it's actually more nuanced, but this code should at least be an improvement. For reference, original discussion is here: https://hail.zulipchat.com/#narrow/stream/300487-Hail-Batch.20Dev/topic/Bugs/near/258563670

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -500,7 +500,7 @@ class Container:
                                     raise
                                 except Exception:
                                     log.exception(f'while extracting image {self.image_ref_str}, ID: {self.image_id}')
-                                    await check_shell(f'rm -Rf {self.rootfs_path}')
+                                    await blocking_to_async(worker.pool, shutil.rmtree, self.rootfs_path)
                 await asyncio.shield(_localize_rootfs())
 
             with self.step('pulling'):
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -478,17 +478,30 @@ class Container:
     async def run(self):
         try:
             async def localize_rootfs():
-                async with image_lock.reader_lock:
-                    # FIXME Authentication is entangled with pulling images. We need a way to test
-                    # that a user has access to a cached image without pulling.
-                    await self.pull_image()
-                    self.image_config = image_configs[self.image_ref_str]
-                    self.image_id = self.image_config['Id'].split(":")[1]
-                    self.worker.image_data[self.image_id] += 1
+                async def _localize_rootfs():
+                    async with image_lock.reader_lock:
+                        # FIXME Authentication is entangled with pulling images. We need a way to test
+                        # that a user has access to a cached image without pulling.
+                        await self.pull_image()
+                        self.image_config = image_configs[self.image_ref_str]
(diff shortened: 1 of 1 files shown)


batch/batch/worker/worker.py
```diff
@@ -2135,14 +2145,15 @@ async def cleanup_old_images(self):
...
                     image_data = self.image_data[image_id]
-                    if image_data.ref_count == 0 and (now - image_data.last_accessed) > 10 * 60 * 1000:
-                        assert image_id != BATCH_WORKER_IMAGE_ID
-                        log.info(f'Found an unused image with ID {image_id}')
-                        await check_shell(f'docker rmi -f {image_id}')
-                        image_path = f'/host/rootfs/{image_id}'
-                        await blocking_to_async(self.pool, shutil.rmtree, image_path)
-                        del self.image_data[image_id]
-                        log.info(f'Deleted image from cache with ID {image_id}')
+                    async with image_data.lock:
```
daniel-goldstein, line 2148: This makes me think that we're not shielding enough. In the good case, this should be redundant because the `writer_lock` means that no coroutines are currently pulling or extracting any filesystems (since `localize_fs` grabs `image_lock.reader_lock`). I feel like we should instead fully shield `localize_fs` and be certain that locks pertaining to pulling/extracting filesystems are properly accounted for during cancellation. What do you think?

daniel-goldstein (changes requested):

batch/batch/worker/worker.py
```diff
@@ -478,19 +478,30 @@ def __init__(self, job, name, spec, client_session: httpx.ClientSession, worker:
...
+                            if not image_data.extracted:
+                                try:
+                                    await self.extract_rootfs()
+                                    image_data.extracted = True
+                                    log.info(f'Added expanded image to cache: {self.image_ref_str}, ID: {self.image_id}')
+                                except asyncio.CancelledError:
+                                    raise
+                                except Exception:
+                                    log.exception(f'while extracting image {self.image_ref_str}, ID: {self.image_id}')
+                                    await check_shell(f'rm -Rf {self.rootfs_path}')
```
daniel-goldstein, line 503: I appreciate this making the extraction idempotent. Why not do this in the case of the cancelled error as well?

daniel-goldstein (dismissed):
I think this looks good but I'm still having a hard time understanding under which circumstances we would hit line 499 vs line 501. Could you explain / potentially leave a comment describing the scenario in which we would reach each except?

batch/batch/worker/worker.py
jigold: My understanding is that the CancelledError was for when the application is closing and not us cancelling the task since we shield it. If we catch all exceptions including that one, then we'll try and delete files upon exiting. I could be wrong though!

daniel-goldstein: Ah I see, I was forgetting that the CancelledError could come from application shutdown. Thanks!

daniel-goldstein: One last thing: can we use `await blocking_to_async(self.pool, shutil.rmtree, image_path)` instead of shelling out to `rm`?

PR #11200, opened 2022-01-05
[grafana] Use templated domain instead of hail.is

daniel-goldstein:
This issue didn't manifest until I needed to re-authenticate in azure, at which point grafana tried to bring me to gcp auth and got hella confused.


danking (approved):
indeed very confusing

PR #11201, opened 2022-01-05
[memory] implement azure support

danking:
I also added factories for Google and Azure so that we only
check the global-config cloud setting once, not repeatedly
every time we create an AsyncFS.

### DIFF
--- a/memory/memory/memory.py
+++ b/memory/memory/memory.py
@@ -72,13 +72,13 @@ async def get_or_add_user(app, userdata):
     username = userdata['username']
     if username not in users:
         k8s_client = app['k8s_client']
-        hail_credentials_secret_name = await retry_transient_errors(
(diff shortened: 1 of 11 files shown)


memory/memory/memory.py
```diff
@@ -70,15 +72,14 @@ async def get_or_add_user(app, userdata):
     username = userdata['username']
     if username not in users:
         k8s_client = app['k8s_client']
-        hail_identity_secret = await retry_transient_errors(
+        hail_credentials_secret_name = await retry_transient_errors(
```
jigold, line 75: Can you remove the `_name` from the suffix here? You're returning a secret and not a name.

jigold (changes requested):

jigold (approved):

PR #11202, opened 2022-01-05
[memory] several fixes

danking:
1. I taught `get_or_add_user` to serialize loading of a single user's secrets
   from kubernetes. The old code would issue multiple requests in parallel for
   a user not yet in `users`.

2. The whole situation with the worker pool and the files in progress set did
   not seem right to me. If a write comes in, we create a future corresponding
   to the write and store the future in a dictionary. If a read comes in, we
   first check redis, then we check if there is already a write or read of cloud
   storage in progress. If there is no write nor read in progress, we initiate
   a read. Before a read or write completes, it removes itself from the in-progress
   files (future reads will now see the file in redis). Writes are permitted to
   overwrite other writes that are already in progress.

### DIFF
--- a/gear/gear/clients.py
+++ b/gear/gear/clients.py
@@ -1,4 +1,3 @@
-import abc
 from typing import Optional
 
 from gear.cloud_config import get_azure_config, get_gcp_config, get_global_config
--- a/gear/gear/clients.py
+++ b/gear/gear/clients.py
@@ -1,3 +1,4 @@
+import abc
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


danking:
stacked on https://github.com/hail-is/hail/pull/11201

danking:
cc: @jigold some changes to memory to simplify it and address some weird behavior. sent over to Daniel since I feel like I've been hitting you with a bunch of PRs lately 😉 !

daniel-goldstein (approved):
Big fan of just waiting on futures instead here.

PR #11203, opened 2022-01-06
[batch] address some pylint issues

danking:
Lines 381-390 had a real bug: we re-used the "b1" and "b2" variables for the rest of wait which is a dictionary, not a true handle to the batch. I went ahead and fixed all the mypy / pylons issues I found in this file.


jigold:
I see. It's a bug in that the debug information if the test had failed would have been wrong / thrown an error. But the actual test right now was testing the right thing. Is this correct?

danking:
That's correct. I saw an intermittent failure in Azure on one of my PRs. That's how I found out the debug info was wrong.

jigold (approved):

PR #11205, opened 2022-01-07
[batch] Enforce black formatting

daniel-goldstein:
The result of running `black $HAIL` and added `black --check` to the `check` targets for the affected services: batch, gear, website, benchmark-service

cc: @danking


danking:
love it; let's make sure we follow up with the git blame ignore for this commit SHA whatever it may be

daniel-goldstein:
@jigold bump

jigold (dismissed):
Do we need Black on for query?

jigold (approved):

PR #11207, opened 2022-01-07
[aiogoogle] Use anonymous credentials when not in GCE

daniel-goldstein:
and give the user a warning that they might want to log in. Anonymous credentials will allow us to access public buckets without authenticating, which was not possible until now. I contemplated whether we should only suggest the login if they've received a 401 but I assume most everyone using this will want to be authenticated and we already log a warning as it is.

This is reasonably testable, but I couldn't find where it should go. Does `test_fs` test credential-loading at all?

### DIFF
--- a/hail/python/hailtop/aiocloud/aiogoogle/credentials.py
+++ b/hail/python/hailtop/aiocloud/aiogoogle/credentials.py
@@ -3,6 +3,7 @@ import os
 import json
 import time
 import logging
+import socket
 from urllib.parse import urlencode
 import jwt
 from hailtop.utils import request_retry_transient_errors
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/aiocloud/aiogoogle/credentials.py
```diff
@@ -160,3 +165,25 @@ async def _get_access_token(self) -> GoogleExpiringAccessToken:
...
+        pass
+
+    async def auth_headers(self) -> Dict[str, str]:
+        return {}
+
+    async def _get_access_token(self) -> GoogleExpiringAccessToken:
+        raise NotImplementedError
+
+    async def close(self):
+        pass
```
danking, line 189: You can just use `AnonymousCloudCredentials` from aiocloud.common.

danking (changes requested):

danking:
AFAIK, we don't test credential discovery. A little tricky because you want to modify the file system, which isn't kosher on a developer's laptop. Also not clear what to do about, e.g., the metadata server. You could mock it, but I've always felt pretty negative about the value of mocking. You might try adding some build.yaml steps that use docker to replicate some common environments?

daniel-goldstein:
Ah, I forgot this is potentially a laptop-running test and not in a batch job! Definitely would require some more thought.

danking (approved):

PR #11209, opened 2022-01-07
to_pandas

ammekk:
Changed function implementation to not rely on spark

### DIFF
--- a/hail/python/hail/table.py
+++ b/hail/python/hail/table.py
@@ -3239,6 +3239,18 @@ class Table(ExprContainer):
 
     @typecheck_method(flatten=bool)
     def to_pandas(self, flatten=True):
+        """Converts this table to a Pandas DataFrame.
+
+        Parameters
+        ----------
+        flatten : :obj:`bool`
+            If ``True``, :meth:`flatten` before converting to Pandas DataFrame.
+
+        Returns
+        -------
+        :class:`.pandas.DataFrame`
+
+        """
         table = self.flatten() if flatten == True else self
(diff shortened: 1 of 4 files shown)


hail/python/hail/table.py
```diff
@@ -3255,7 +3251,28 @@ def to_pandas(self, flatten=True):
         :class:`.pandas.DataFrame`
 
         """
-        return Env.spark_backend('to_pandas').to_pandas(self, flatten)
+        table = self.flatten() if flatten == True else self
+        dtypes_struct = table.row.dtype
+        collect_dict = {key: hl.agg.collect(value) for key, value in table.row.items()}
+        table.aggregate(hl.struct(**collect_dict))
```
johnc1231, line 3257: Delete this line, it's not doing anything (you use same expression on next line)

hail/python/hail/table.py
```diff
@@ -3255,7 +3251,28 @@ def to_pandas(self, flatten=True):
...
+        column_struct_array = table.aggregate(hl.struct(**collect_dict))
+        columns = list(column_struct_array.keys())
+        data_dict = {}
+
+        for data_idx, column in enumerate(columns):
```
johnc1231, line 3262: I don't think you use `data_idx` for anything.

johnc1231 (changes requested):
Looks great! Two small changes, plus you should fix the style checks, and you're good to go.

johnc1231 (dismissed):

johnc1231 (approved):

PR #11210, opened 2022-01-07
[hailctl] Move slow imports out of the top level

daniel-goldstein:
tqdm.notebook and pkg_resources take .55s and .37s to import, respectively. humanize imports pkg_resources which is why that is moved to. Moving these out of the top level improve cold-start time from 1.5s to .5s, and never even need to be imported for invocations such as `hailctl config list`, whose runtime is entirely import time.

### DIFF
--- a/hail/python/hailtop/__init__.py
+++ b/hail/python/hailtop/__init__.py
@@ -1,6 +1,12 @@
+_VERSION = None
+
+
 def version() -> str:
-    import pkg_resources  # pylint: disable=import-outside-toplevel
-    return pkg_resources.resource_string(__name__, 'hail_version').decode().strip()
+    global _VERSION
+    if _VERSION is None:
+        import pkg_resources  # pylint: disable=import-outside-toplevel
+        _VERSION = pkg_resources.resource_string(__name__, 'hail_version').decode().strip()
+    return _VERSION
 
 
 def pip_version() -> str:
--- a/hail/python/hailtop/__init__.py
+++ b/hail/python/hailtop/__init__.py
@@ -1,12 +1,7 @@
-import pkg_resources
-
-
-_VERSION = pkg_resources.resource_string(__name__, 'hail_version').decode().strip()
+def version() -> str:
+    import pkg_resources  # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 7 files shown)


hail/python/hailtop/utils/tqdm.py
```diff
@@ -1,10 +1,8 @@
...
-TQDM_DEFAULT_DISABLE = False if tqdm_auto == tqdm_notebook else None
-
-
-def tqdm(*args, disable=TQDM_DEFAULT_DISABLE, **kwargs):
+def tqdm(*args, disable=False, **kwargs):
+    from tqdm.notebook import tqdm as tqdm_notebook  # pylint: disable=import-outside-toplevel
+    from tqdm.auto import tqdm as tqdm_auto  # pylint: disable=import-outside-toplevel
+    # To tqdm_notebook, None means do not display. To standard tqdm, None means
+    # display only when connected to a TTY.
+    if disable is False and tqdm_auto != tqdm_notebook:
```
daniel-goldstein, line 6: I think this is technically semantically different as it means you cannot pass `disable=False` to standard tqdm…

daniel-goldstein: Maybe this is fine, or else I introduce some other `disable` value that means "whatever the default disable is"

hail/python/hailtop/__init__.py
```diff
@@ -1,12 +1,7 @@
-import pkg_resources
-
-
-_VERSION = pkg_resources.resource_string(__name__, 'hail_version').decode().strip()
+def version() -> str:
+    import pkg_resources  # pylint: disable=import-outside-toplevel
+    return pkg_resources.resource_string(__name__, 'hail_version').decode().strip()
```
danking, line 3: Let’s still cache this value, but do it lazily.

hail/python/hailtop/utils/tqdm.py
danking: I prefer that we do not change the semantics of disable. Use a unique distinct value for the default value of the disable parameter. Like:
```
DEFAULT_DISABLE = object()
```

danking (dismissed):

danking (dismissed):

danking (approved):

PR #11211, opened 2022-01-10
[batch] Add blobfuse support in Azure

jigold:
- Added support for blobfuse, which is Azure's equivalent of gcsfuse.
- Renamed operations/variables to cloudfuse to be generic
- The validator is slightly more complicated because there is a double renaming of gcsfuse to cloud fuse to gcsfuse for old clients. This is to maintain backwards compatibility with old instances. Once v21 instances die, then we can remove the extra backwards compatibility step.
- renamed instance_env.py files to worker_api.py files

@daniel-goldstein @danking Can you decide amongst yourselves who should review this? Maybe Dan takes a first glance and then Daniel reads it in more detail?


jigold:
I forgot about the formatting change. Let's wait for that to go in before reviewing this.

jigold:
This should be ready for review. Only error was black formatting which hopefully I fixed now.

daniel-goldstein (commented):
This is great, glad to see that it was mostly naming changes and slotting in `blobfuse` instead of `gcsfuse`. I'll let Dan do a review too but my main concern is that I keep forgetting what `location` means in this context. Curious what your thoughts are.

batch/batch/worker/worker.py
```diff
@@ -1187,16 +1178,17 @@ def __init__(
...
+        if cloudfuse:
+            for config in cloudfuse:
+                config['mounted'] = False
+                location = config.get('location') or config['bucket']
+                assert location
                 self.main_volume_mounts.append(
                     {
-                        'source': self.gcsfuse_path(b["bucket"]),
-                        'destination': b['mount_path'],
+                        'source': f'{self.cloudfuse_data_path(location)}',
```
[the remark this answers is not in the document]
jigold: It's supposed to be `bucket` or `storage_account/container_name` where both are a string. I definitely agree it's confusing. I had `cloud_location` for awhile, but felt the `cloud_` didn't add anything. I thought about the type being Union[str,Tuple[str,str]]. Maybe `storage_source` or `blob_location`? Not sure.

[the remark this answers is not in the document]
daniel-goldstein: Hmph, I feel like this is sacrificing a lot of clarity in the name of generality. Mostly I blame Azure for breaking from the unique and evocative name "bucket" and using "storage account" and "container" both of which are much more confusing terms. I feel like the string `storage_account/container_name` is semantically what we think of as a bucket, maybe we could just call it a bucket and document that?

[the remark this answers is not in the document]
daniel-goldstein: Alternatively, `blob_location` isn't bad, though I might prefer something like `blob_root_path` (still not great). I just have a hard time severing "location" from geography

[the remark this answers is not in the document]
jigold: @danking This is where the naming conversation is.

batch/Dockerfile.worker
```diff
@@ -25,12 +25,26 @@ COPY batch/download-gcs-connector.py /
...
+{% endif %}
+
+{% if global.cloud == "azure" %}
+# https://github.com/Azure/azure-storage-fuse/issues/603
+RUN hail-apt-get-install ca-certificates pkg-config libfuse-dev cmake libcurl4-gnutls-dev libgnutls28-dev uuid-dev libgcrypt20-dev wget && \
+    wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb && \
+    dpkg -i packages-microsoft-prod.deb && \
+    apt-get update && \
+    hail-apt-get-install blobfuse
+{% endif %}
```
danking, line 44: I have a mild preference for:
```
{% if global.cloud == "gcp" %}
...
{% elif global.cloud == "azure" %}

{% else %}
RUN echo "!!! UNEXPECTED CLOUD {{global.cloud}} !!!" && exit 1
{% endif %}
```

Just to be absolutely certain that we do not accidentally create a bad image.

batch/batch/cloud/azure/driver/create_instance.py
```diff
@@ -250,12 +253,13 @@ def create_vm_config(
 -v /batch:/batch:shared \
 -v /logs:/logs \
 -v /global-config:/global-config \
--v /gcsfuse:/gcsfuse:shared \
+-v /cloudfuse:/cloudfuse:shared \
```
danking, line 256: I don't understand this bit. Why do we mount `/cloudfuse` from the VM into the *worker* container at `/cloudfuse`? Neither the VM nor the worker container use cloud fuse, right?

batch/batch/front_end/validate.py
```diff
@@ -39,17 +39,18 @@
...
     {
         'always_run': bool_type,
         'attributes': dictof(str_type),
         'env': listof(keyed({'name': str_type, 'value': str_type})),
-        'gcsfuse': listof(
+        'cloudfuse': listof(
             keyed(
                 {
-                    required('bucket'): non_empty_str_type,
+                    required('location'): non_empty_str_type,
```
danking, line 53: let's just keep this as bucket. Location is pretty overloaded in our code-base already. I know "bucket" isn't really an Azure term, but I think it's clear that it means a fully qualified container (i.e. "$account/$container").

batch/batch/worker/worker.py
```diff
@@ -1088,36 +1063,50 @@ def get_next_xfsquota_project_id():
...
+    def io_host_path(self) -> str:
         return f'{self.scratch}/io'
 
-    def gcsfuse_path(self, bucket):
+    def cloudfuse_base_path(self):
+        # Make sure this path isn't in self.scratch to avoid accidental bucket deletions!
+        return f'/cloudfuse/{self.token}'
+
+    def cloudfuse_data_path(self, location: str) -> str:
+        # Make sure this path isn't in self.scratch to avoid accidental bucket deletions!
```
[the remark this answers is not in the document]
danking: The real point here is that we can't `rm -rf` this path. It's not trivial to assert that one path does not resolve to a folder that is "inside" another because of hard links and symlinks.

While I sympathize with the need for a good assertion, I'm not sure how to do that here.

batch/batch/worker/worker.py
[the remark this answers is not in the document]
danking: Oh, oops, I commented above. Let's call it bucket, that's what it is, regardless of the added complications of Azure's naming system.

batch/batch/worker/worker.py
```diff
@@ -1430,18 +1422,35 @@ async def run(self):
...
-                                bucket=bucket,
-                                key_file=self.credentials_host_file_path(),
-                                read_only=b['read_only'],
+                with self.step('adding cloudfuse support'):
+                    if self.cloudfuse:
+                        os.makedirs(self.cloudfuse_base_path())
+
+                        await check_shell_output(
+                            f'xfs_quota -x -c "project -s -p {self.cloudfuse_base_path()} {self.project_id}" /host/'
+                        )
```
danking, line 1431: This seems not right to me. Wouldn't this effectively include the entire bucket's contents in the user's xfs quota? I'm not sure I even understand how that would work. It does seem important to put the temporary directory under a quota restriction.

batch/batch/worker/worker_api.py
```diff
@@ -0,0 +1,61 @@
...
+
+from hailtop import httpx
+from hailtop.utils import check_shell, CalledProcessError, sleep_and_backoff
+
+from .disk import CloudDisk
+from .credentials import CloudUserCredentials
+from ..instance_config import InstanceConfig
+
+
+class CloudWorkerAPI(abc.ABC):
```
danking, line 12: This is really cool and I like it a lot.

hail/python/hailtop/batch/job.py
```diff
@@ -408,15 +411,57 @@ def gcsfuse(self, bucket, mount_point, read_only=True):
         Same job object set with a bucket to mount with gcsfuse.
         """
 
+        warnings.warn("The 'gcsfuse' method has been deprecated. Use the 'cloudfuse' method instead.")
+        return self.cloudfuse(bucket, mount_point, read_only=read_only)
+
+    def cloudfuse(self, location: str, mount_point: str, *, read_only: bool = True):
```
danking, line 417: same point about location here.

hail/python/hailtop/batch/job.py
```diff
@@ -408,15 +411,57 @@ def gcsfuse(self, bucket, mount_point, read_only=True):
...
+        There are performance and cost implications of using `gcsfuse <https://cloud.google.com/storage/docs/gcs-fuse>`__
+        or `blobfuse <https://github.com/Azure/azure-storage-fuse#considerations>`__.
+
+        Examples
+        --------
+
+        >>> b = Batch(backend=backend.ServiceBackend('test'))
+        >>> j = b.new_job()
+        >>> (j.cloudfuse('my-bucket', '/my-bucket')
+        ...   .command(f'cat /my-bucket/my-file'))
```
danking, line 437: give an example in Azure as well with a storage account / container pair.

danking (changes requested):
Good stuff! some thoughts

batch/batch/cloud/azure/driver/create_instance.py
jigold: The batch worker creates directories for a specific job in /cloudfuse. This needs to be on the host's file system.

batch/batch/worker/worker.py
jigold: I'm not 100% sure, but my understanding was that the contents on disk were just cached files. I don't think it's the entire bucket. Nonetheless, maybe the fuse file system returns the full size of a directory which would be a problem if that's what xfs uses to track storage space. My guess is they are completely independent.

batch/batch/worker/worker_api.py
jigold: This was Daniel's addition :)

batch/batch/worker/worker.py
jigold: Just to be clear. The temp directory is for Azure only. It's an unused directory for GCP

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -406,9 +406,11 @@ def setUp(self):
         self.remote_tmpdir = remote_tmpdir
 
         if remote_tmpdir.startswith('gs://'):
-            self.bucket_name = re.fullmatch('gs://(?P<bucket_name>[^/]+).*', remote_tmpdir).groupdict()['bucket_name']
+            self.cloud_location = re.fullmatch('gs://(?P<bucket_name>[^/]+).*', remote_tmpdir).groupdict()['bucket_name']
         else:
-            self.bucket_name = None
+            assert remote_tmpdir.startswith('hail-az://')
+            storage_account, container_name = re.fullmatch('hail-az://(?P<storage_account>[^/]+)/(?P<container_name>[^/]+).*', remote_tmpdir).groups()
+            self.cloud_location = f'{storage_account}/{container_name}'
```
danking, line 413: sorry this is a nit, but can we we switch the words back to bucket for the tests too? Otherwise I'm good with this PR.

danking (changes requested):

danking (approved):

PR #11212, opened 2022-01-10
[query] Parameterize Hail Query Class Loading by ClassLoader

danking:
The current main version of the Query Service uses a fresh class loader for every query. This means each driver job and worker job starts with uncompiled classes for any class in Hail.

This change uses a shared class loader for all jobs with the same SHA. This enables use of previously JIT'ed Hail classes. This noticeably improves no-op performance from ~8 seconds to ~3 seconds. Most of that remaining 3 seconds is due to Query-on-Batch and Batch, not Query.

Currently, Hail generates classes using a counter. When a driver or worker re-uses an old class loader, it would mistakenly re-use classes generated by a previous Hail Query-on-Batch job because they share the same name. This PR avoids that entirely by using a fresh class loader per job for *generated* classes.

This PR parameterizes the entire Hail Query system by a class loader. This class loader is passed in from the initiator of the driver or worker job. We could, eventually, re-use class loaders:
- across jobs for a single batch
- across jobs for a single user
- across jobs for a single billing project
- across all jobs

I think the first three are somewhat uncontroversial but we need to fix the class naming problem. The fourth introduces a new security risk. I think we have a lot of performance to squeeze out of QoB before we need to take that step.


danking:
@daniel-goldstein , I tagged you because the QoB java code (everything `is.hail.backend.service`) should really be owned by services team. @johnc1231 , I tagged you because this is a pervasive change to Query. You may want to confer with Chris and Patrick S before approving in light of Tim's absence.

I'll apply WIP. Whoever approved second, please remove the WIP tag.

I expect the test time to roughly halve with this change.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -581,21 +584,6 @@ class ServiceBackendSocketAPI2(
             writeString(formatException(t))
         }
 
-      case MATRIX_TABLE_TYPE =>
```
daniel-goldstein, line 584: Was this an accidental delte?

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -428,7 +429,9 @@ object ServiceBackendSocketAPI2 {
     val input = argv(5)
     val output = argv(6)
 
-    val backend = new ServiceBackend(revision, jarLocation, name, scratchDir)
+    // FIXME: when can the classloader be shared? (optimizer benefits!)
```
daniel-goldstein, line 432: Is this because of class-naming issues? I assume anything "client"-side (i.e. anything that's not batch code) should be able to share as much as it wants from a security perspective.

daniel-goldstein (changes requested):
My parts look mostly fine but were really simple

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: Yep great find. Merge fail.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: I removed the FIXME. Explaining this above has made pretty clear to me when it can be shared. There is a hierarchy of security perimeters:
1. Global
2. Billing Project
3. User
4. Batch
5. Job

And a tree of sets of code. Each code set relies on code from one level higher. The levels of the tree are:
1. Java stdlibs
2. Spark JARs
2. Hail JAR for a given SHA
3. Hail generated code for a given SHA and batch and job

The question is: in which perimeter does each code-set live? In main the answer is:
1. Java stdlibs: Global
2. Spark JARs: Global
3. Hail JAR: Job
4. Hail generated code: Job

In this PR the answer is:
1. Java stdlibs: Global
2. Spark JARs: Global
3. Hail JAR: Global
4. Hail generated code: Job

The higher up the security hierarchy a code-set lives, the higher the chance that someone else has already paid the cost of JIT'ing a class you need. Descending the security hierarchy obviously gives fewer chances for bad code to screw with you.

Honestly, I think everything should be in the global hierarchy, because all this code is either written by Hail team or generated by code written by Hail team. However, we need to fix the naming issue with generated classes before code-set 4 can be shared across different batches.

danking: @johnc1231 I think this framing was really instructive for me, you might read over it.

johnc1231: When you say everything should be in the global hierarchy, I think that's not quite true, since there's still the issue of different hail shas. The Java stdlib stuff should be sharable globally, but that's not true of two different hail SHAs with different versions of some random SType class.

danking:
This only yielded a ~20% reduction. Test times are around 16 minutes now down from around 20 minutes.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: You're definitely right John. I think my picture is not quite right.

PR #11213, opened 2022-01-10
[query] from_pandas

ammekk:
Updated from_pandas to not rely on spark.


johnc1231 (changes requested):
We are going to grow the scope of this a bit.

Right now, you'd run into a problem if the first row of your dataframe contained an empty python list. The problem is, hail won't know what the element type is. Same for empty dict, empty set. Ndarrays are covered since numpy always stores an element type.

Probably the right thing to do is to scan over all the data in a type inference pass where we look at the elements and decide the type. If one column has an array of ints and an array of floats, we should realize that the more general type is floats and accept that. 

It might be best to extend `Table.parallelize` to do this. Imagine if `parallelize` took a `type_hints` argument, which was a dictionary of field names to hail types. For any types specified in `type_hints`, we'd know the type already. Then we'd infer the rest. Then `from_pandas` could just generate a set of type hints based on the non `object` columns, and let `parallelize` infer the type of the remaining columns.

hail/python/hail/table.py
```diff
@@ -450,8 +450,9 @@ def _select_globals(self, caller, s) -> 'Table':
     @typecheck_method(rows=anytype,
                       schema=nullable(hail_type),
                       key=table_key_type,
-                      n_partitions=nullable(int))
-    def parallelize(cls, rows, schema=None, key=None, n_partitions=None) -> 'Table':
+                      n_partitions=nullable(int),
+                      partial_type=nullable(dict))
+    def parallelize(cls, rows, schema=None, key=None, n_partitions=None, partial_type=None) -> 'Table':
```
danking, line 455: Let's make partial_type keyword only: `..., n_partitions=None, *, partial_type=None`

hail/python/test/hail/table/test_table.py
```diff
@@ -1771,4 +1842,6 @@ def test_to_pandas_nd_array():
     }
 
     df_from_python = pd.DataFrame(python_data)
-    pd.testing.assert_frame_equal(df_from_hail, df_from_python)
\ No newline at end of file
+    pd.testing.assert_frame_equal(df_from_hail, df_from_python)
+
+
```
danking, line 1847: nix empty lines at end of file

danking (changes requested):
love it!

johnc1231:
@ammekk You have to fix documentation for this method. The documentation contains an example which is failing tests

hail/python/hail/table.py
```diff
@@ -3290,7 +3336,25 @@ def from_pandas(df, key=[]) -> 'Table':
...
+        columns = {fields[col_idx]: df[col].tolist() for col_idx, col in enumerate(df.columns)}
+        data = [{} for _ in range(len(columns[fields[0]]))]
+
+        for data_idx in range(len(columns[fields[0]])):
+            for field in fields:
+                data[data_idx][field] = columns[field][data_idx]
+
+        for data_idx, field in enumerate(fields):
+            type_hint = dtypes_from_pandas(pd_dtypes[field])
+            if type_hint:
```
johnc1231, line 3353: Prefer `if type_hint is not None` if that's the intended check being done here.

johnc1231 (changes requested):

johnc1231 (approved):

PR #11214, opened 2022-01-10
[fs] Add azure blob path to FileNotFound error in read

daniel-goldstein:
The Azure blob storage client does not include the blob name in file not found errors. This adds that information to the `FileNotFoundError` that we raise on top of the azure error in the azure fs's `read`.


daniel-goldstein:
Dan ended up taking this over so I'm closing

PR #11217, opened 2022-01-12
[vds] Fix inverted reference block inversion in `hl.vds.to_dense_mt`

tpoterba:


hail/python/hail/vds/methods.py
```diff
@@ -58,21 +58,23 @@ def coalesce_join(ref, var):
...
-                merged_fields[field] = hl.coalesce(var[field], ref[field])
+        shared_fields = [call_field] + list(f for f in ref.dtype if f in var.dtype)
+        shared_field_set = set(shared_fields)
+        var_fields = [f for f in var.dtype if f not in shared_field_set]
 
-        return hl.struct(**merged_fields).annotate(**{f: var[f] for f in var if f not in merged_fields})
+        return hl.if_else(hl.is_defined(var),
+                          var.select(*shared_fields, *var_fields),
+                          ref.annotate(**{call_field: hl.call(0, 0)})
+                          .select(*shared_fields, **{f: hl.null(var[f].dtype) for f in var_fields}))
```
patrick-schultz, line 68: Would it be better to leave the variant fields in their original order, or do you think it doesn't matter?

patrick-schultz (changes requested):

hail/python/hail/vds/methods.py
tpoterba: I think this doesn't matter. the shared fields are usually in front by convention anyway.

patrick-schultz (approved):

PR #11219, opened 2022-01-12
Update docs to remove 3.6 from supported versions

johnc1231:
Follow up to #11206. Python 3.6 is EOL, so let's not claim to support it or encourage its use anymore.

### DIFF
--- a/hail/python/hail/docs/install/linux.rst
+++ b/hail/python/hail/docs/install/linux.rst
@@ -3,7 +3,7 @@ Install Hail on GNU/Linux
 =========================
 
 - Install Java 8.
-- Install Python 3.6+.
+- Install Python 3.7 or later.
 - Install a recent version of the C and C++ standard libraries. GCC 5.0, LLVM
   version 3.4, or any later versions suffice.
 - Install BLAS and LAPACK.
@@ -16,8 +16,8 @@ On a recent Debian-like system, the following should suffice:
    apt-get install -y \
        openjdk-8-jre-headless \
        g++ \
-       python3.6 python3-pip \
+       python3.7 python3-pip \
        libopenblas-base liblapack3
-   python3.6 -m pip install hail
+   python3.7 -m pip install hail
 
 `Now let's take Hail for a spin! <try.rst>`__
--- a/hail/python/hail/docs/install/macosx.rst
+++ b/hail/python/hail/docs/install/macosx.rst
@@ -11,6 +11,6 @@ Install Hail on Mac OS X
     brew cask install adoptopenjdk8
     brew install --cask adoptopenjdk8
 
-- Install Python 3.6+. We recommend `Miniconda <https://docs.conda.io/en/latest/miniconda.html#macosx-installers>`__.
+- Install Python 3.7 or later. We recommend `Miniconda <https://docs.conda.io/en/latest/miniconda.html#macosx-installers>`__.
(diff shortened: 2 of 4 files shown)


daniel-goldstein:
hail/python/setup.py:    python_requires=">=3.6",
hail/python/hail/backend/spark_backend.py:    raise EnvironmentError('Hail with spark {} requires Python 3.6 or 3.7, found {}.{}'.format(
hail/python/requirements.txt:orjson==3.6.4
hail/python/hailtop/batch/batch.py:        3.6, 3.7, or 3.8.
hail/python/hailtop/batch/docker.py:            f'Python versions other than 3.6, 3.7, or 3.8 (you are using {major_version}.{minor_version}) are not supported')
hail/python/hailtop/batch/batch_pool_executor.py:        include Python 3.6 or later and must have the ``dill`` Python package
hail/python/hailtop/batch/batch_pool_executor.py:                    f'You must specify an image if you are using a Python version other than 3.6, 3.7, or 3.8 (you are using {version})')
hail/python/hailtop/batch/docs/cookbook/random_forest.rst:choose a suitable image for you if your Python version is 3.6, 3.7, or 3.8.
hail/python/hailtop/batch/docs/docker_resources.rst:exists to base your image on. An image that has Python preinstalled is `python:3.6-slim-stretch` and
hail/python/hailtop/batch/docs/tutorial.rst:f-strings were added to Python in version 3.6 and are denoted by the 'f' character
hail/python/hailtop/batch/backend.py:                        f"You must specify 'image' for Python jobs if you are using a Python version other than 3.6, 3.7, or 3.8 (you are using {version})")
docker/python-dill/push.sh:for version in 3.6 3.6-slim 3.7 3.7-slim 3.8 3.8-slim
docker/third-party/images.txt:python:3.6
docker/third-party/images.txt:python:3.6-slim
benchmark/python/setup.py:    python_requires=">=3.6",


Found a few more invocations of 3.6. I'm not sure when you want to update the `python_requires` and makes sense to leave the 3.6 docker images around for a little while, but the other places in the batch documentation and warnings seem a good thing to update now too.

johnc1231:
I think this covers most spots. I updated the python_requires, left the docker images, tried to update batch docs.

hail/python/hail/backend/spark_backend.py
```diff
@@ -25,7 +25,7 @@
 from ..hail_logging import Logger
 
 if pyspark.__version__ < '3' and sys.version_info > (3, 8):
```
daniel-goldstein, line 27: If the `>` symbol there is to be believed should we require 3.7 or 3.8? Or should that be `>=`?

daniel-goldstein (changes requested):

hail/python/hail/backend/spark_backend.py
daniel-goldstein: If it's `>=` then maybe that fix is out of scope here since that would be breaking?

johnc1231: Huh, this must not actually run, since we definitely allow users to use 3.8. I'll investigate, good catch.

johnc1231: Oh wait, I remember now. This is because Spark 2 can't use > 3.8. So yeah, 3.7 is the only ok version, that's fine. I plan on dropping Spark 2 support shortly anyway.

daniel-goldstein (approved):

PR #11220, opened 2022-01-13
[query] include JRE 11 as an acceptable Java version in the docs

danking:
I also fixed a reference to the JDK. Users do not need `javac` and friends, just a JRE (i.e. `java`).


hail/python/hail/docs/install/other-cluster.rst
```diff
@@ -10,7 +10,7 @@ Hail should work with any Spark 3.1.1 cluster built with Scala 2.12.
 Hail needs to be built from source on the leader node. Building Hail from source
 requires:
 
-- Java 8 JDK.
+- Java 8 or 11 JRE.
```
johnc1231, line 13: Building Hail from source requires a JDK.

johnc1231 (dismissed):

johnc1231 (approved):

PR #11223, opened 2022-01-14
[ci] Show github statuses in addition to build status

daniel-goldstein:
Currently if I visit `ci.hail.is` it shows the status of the build that CI ran in GCP. Now that we enforce all CI's pass, we can end up with situations on `ci.hail.is` where the PR is approved and the build was successful but it's not merging because the tests failed over on azure. This shows all the github statuses in addition to the cloud-local build information.

### DIFF
--- a/ci/ci/templates/pr-table.html
+++ b/ci/ci/templates/pr-table.html
@@ -6,6 +6,7 @@
     <tr>
       <th>PR</th>
       <th>Build State</th>
+      <th>Github Statuses</th>
       <th>Labels</th>
       <th>Branch</th>
       <th>Review State</th>
@@ -29,6 +30,11 @@
         {% if pr.build_state is not none and pr.out_of_date %}
           *
         {% endif %}
+      </td>
+      <td>
+        {% for context, status in pr['gh_statuses'].items() %}
+          {{ context }}: {{ status }}
+        {% endfor %}
       </td>
       <td>
         {{ pr.labels|join(", ") }}
--- a/ci/ci/templates/pr-table.html
+++ b/ci/ci/templates/pr-table.html
@@ -32,9 +32,11 @@
         {% endif %}
       </td>
       <td>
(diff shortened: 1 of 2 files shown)


ci/ci/templates/pr-table.html
```diff
@@ -30,6 +31,11 @@
           *
         {% endif %}
       </td>
+      <td>
+        {% for context, status in pr['gh_statuses'].items() %}
+          {{ context }}: {{ status }}
```
daniel-goldstein, line 36: Should I be making rows instead here to stack these on top of each other? Not too familiar with jinja.

danking: The question is less  to do with ninja and more to do with HTML.

WRT stacking rows, do you want to achieve this?
```
+----------------------+
|         | azure: yes |
|  #12345 +------------+ 
|         | gcp: no    |
+----------------------+
```

danking: I don't think this is possible unless we fix the number of other CIs. If we fix the number of other CIs, we should just create two columns, one for azure and one for GCP.

I think that's probably the right thing to do here. Something like this
```python3
def get_gh_status_str(statuses, status_key):
    x = statuses(status_key)
    if x is None:
        return x
    return x.value
# ...
'gh_statuses': {get_gh_status_str(pr.last_known_github_status, k) for k in ('hail-ci', 'ci-test-azure')},
```

You could also dynamically figure out the number of distinct status across all PRs and create that many columns, etc. Either approach seems fine to me.

danking (dismissed):
forgot to request changes

danking:
@daniel-goldstein bump, If you want to display AZ and GCP, then do two columns. If you want to show an arbitrary number of statuses, then change this to an `<ol>` so that it's a bit more readable.

danking (approved):

PR #11226, opened 2022-01-14
[batch] create load test to stress the batch-driver at a maximum sche…

vrautela:
…duling rate

### DIFF
--- a/batch/load-test/test-ten-fast-jobs.py
+++ b/batch/load-test/test-ten-fast-jobs.py
@@ -1,0 +1,30 @@
+import hailtop.batch as hb
+import math
+import argparse
+
+
+parser = argparse.ArgumentParser(description='Load test the Batch service using for a given max scheduling rate')
+parser.add_argument('msr', type=int, nargs=1, help='the maximum scheduling rate for the Batch Driver')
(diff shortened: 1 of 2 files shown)


batch/load-test/test-ten-fast-jobs.py
```diff
@@ -0,0 +1,30 @@
...
+
+# Need to set hailctl remote_tmpdir
+backend = hb.ServiceBackend('test')
+b = hb.Batch(backend=backend, name='load-test')
+
+# the number of quarter-core jobs that can be running on n machines (assuming each machines has 16 cores)
+max_concurrent_quarter_core_jobs = 64 * n
+
+# TODO: should the factor of 10 be user-specificable?
+# this load test will require the driver to schedule 10 times as many jobs as can possibly be running at once
```
daniel-goldstein, line 22: I think we should base the calculation of `n_jobs` off of the desired duration of the load test. Let's add a third parameter being the number of minutes the test should last, and use our scheduling rate to figure how large our batch should be.

daniel-goldstein (changes requested):
Good first pass, see my comment about specifying the duration.

daniel-goldstein (approved):

PR #11227, opened 2022-01-14
[aiotools] copy into directory

CDiaz96:
Added `into` functionality for the copy tool. Created test in a new file to show this functionality works.


jigold:
Don't forget to remind me that we need to look at the CI logs and make sure the test suite actually got ran.

hail/python/test/hailtop/aiotools/test_into_copy.py
```diff
@@ -0,0 +1,41 @@
...
+@pytest.mark.asyncio
+async def test_copy_dir():
+    with tempfile.TemporaryDirectory() as test_dir:
+        os.makedirs(f"{test_dir}/subdir1")
+        write_file(f"{test_dir}/subdir1/file1", "hello world\n")
+
+        inputs = [{"from": f"{test_dir}/subdir1", "into": f"{test_dir}/subdir2"}]
+
+        await copy_from_dict(files=inputs)
+        assert read_file(f"{test_dir}/subdir2/subdir1/file1") == "hello world\n"
```
danking, line 41: I have another test I would like to add. If there are two files:

```
(base) # ls -l 
total 0
-rw-r--r--  1 dking  wheel  0 Jan 14 16:04 bar
-rw-r--r--  1 dking  wheel  0 Jan 14 16:04 foo
```

And we try to copy bar *into* foo, then it must fail because foo is already a file. It cannot also be a directory. Let's add a this test which:
1. creates two files 
2. tries to copy one file *into* the other file
3. and expects a `NotADirectoryError`

The pattern for "expect an error" tests is this:
```python3
try:
    # ... try to do the thing that causes an error here ...
    dangerous_function()
except TheErrorYouExpect:
    pass
else:
    assert False  # we come to "else:" if there was no error
```

danking (changes requested):
Looks good, just one more test to add!

danking:
I checked CI, the test is run in test_hail_python.

jigold:
@danking I'm not sure I like this new test. We're only testing it works locally. I'm not sure this `NotADirectoryError` is actually raised in the cloud. This is a bit worrisome to me and I think we should move the tests to the test_copy suite where we can test against all clouds once Carolin has finished what you asked for.

danking:
Fair point. I think asserting this behavior works on a local FS is good. I think a follow up to this PR that parameterized the other tests by filesystem is also great. That follow up PR can just leave the local-only test unchanged.

hail/python/test/hailtop/aiotools/test_into_copy.py
CDiaz96: Added additional test

danking (approved):
excellent!

PR #11231, opened 2022-01-19
[style] Add isort for import sorting

daniel-goldstein:
See `pyproject.toml` for the isort configuration. This adds isort as a part of the `check` target for the services. To manually run isort, just run `isort .` anywhere in the hail repo. It will properly find the config file and ignore non-services code. isort has a `black` setting so it should be compatible with subsequent format checks. It doesn't immediately play well with pre-commit (ignores file excludes and runs where it shouldn't e.g. migrations) so I will separately have to look into running it there. When run, it does the following:

- Inserts line breaks to keep long imports in the desired line length
- Sorts items imported from a given module, it appears first by data type (classes before functions) and then alphabetically
- Groups and orders imports by: stdlib, third party, first party (set in pyproject.toml), and local (inferred from imports that start with `.`). I like this a lot
- Sorts imports within each group by `import`vs`from … import`, then alphabetically by module

### DIFF
--- a/auth/Makefile
+++ b/auth/Makefile
@@ -12,6 +12,7 @@ BLACK := $(PYTHON) -m black . --line-length=120 --skip-string-normalization
 check:
 	$(PYTHON) -m flake8 auth
 	$(PYTHON) -m pylint --rcfile ../pylintrc auth --score=n
+	$(PYTHON) -m isort . --check-only --diff
 	$(BLACK) --check --diff
 	curlylint .
 	bash ../check-sql.sh
--- a/auth/auth/auth.py
+++ b/auth/auth/auth.py
@@ -1,32 +1,34 @@
-import os
-import logging
 import asyncio
+import json
+import logging
+import os
+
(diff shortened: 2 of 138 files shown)


daniel-goldstein:
I need to make this more robustly ignore migration files

danking:
@daniel-goldstein , I love this change, can we just make isort run on `batch/batch`, `ci/ci`, etc. that avoids all the sql migrations.

daniel-goldstein:
So the issue was less ignoring the SQL files and more pre-commit overriding preferences. I just removed the pre-commit hook and can come back to it later. Running `isort .` in $HAIL should fix any issues, but I also hope that just having everything in order will make it easier to identify where new imports should go.

Happy to follow-up with a git blame ignore PR but idk how much anyone cares about import history.

danking:
Let's do the git blame ignore PR, it seems like good housekeeping practice..

danking (approved):

PR #11233, opened 2022-01-19
[query] fix storeEmitParam

patrick-schultz:
Make sure result of `storeEmitParam` doesn't contain any references to the parameter, since it may be used in a different function due to method splitting.


tpoterba (approved):
very nice!

PR #11240, opened 2022-01-20
[query] Remove SCode.memoize

patrick-schultz:
Replace uses of `new SFooCode(...).memoize(cb)` with one of
* `new SFooValue(...)`
* for pointer types, `pType.loadCheapSCode(cb, addr)`

With `memoize` no longer used, all `SCode` methods are unused. I delete all the class bodies, but leave the concrete `SCode` classes, because `SSettable.store(cb, SValue)` is still implemented as `SSettable.store(cb, sv.get)`. Fixing that will be the next PR.

Delete `loadCheapSCodeField`, replace uses with `cb.memoizeField(loadCheapSCode(...))`.


tpoterba (approved):
nice!

PR #11241, opened 2022-01-20
[query] Finish removing SCode

patrick-schultz:
Stacked on #11240

Rewrite all `SSettable.store` implementations to take an `SValue`. With that, `SCode` and subclasses are safe to delete.


johnc1231 (approved):
Nice!

PR #11244, opened 2022-01-20
[query] Make TableKeyByAndAggregate bufferSize parameter required

chrisvittal:
Allow the default value to be controlled by a context flag.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/TableIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/TableIR.scala
@@ -2498,7 +2498,7 @@ case class TableKeyByAndAggregate(
   expr: IR,
   newKey: IR,
   nPartitions: Option[Int] = None,
-  bufferSize: Int = 50) extends TableIR {
+  bufferSize: Int) extends TableIR {
   require(expr.typ.isInstanceOf[TStruct])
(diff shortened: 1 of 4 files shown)


chrisvittal:
This is barely for power users. I hate it. And it should die as soon as it can.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -2498,7 +2498,7 @@ case class TableKeyByAndAggregate(
   expr: IR,
   newKey: IR,
   nPartitions: Option[Int] = None,
-  bufferSize: Int = 50) extends TableIR {
+  bufferSize: Int = -1) extends TableIR {
```
tpoterba, line 2501: just take out the default value?

tpoterba (changes requested):
thanks!

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
chrisvittal: didn't know you could _do_ that in scala. Have default parameters and non-default parameters after them in a parameter list.

tpoterba (approved):

PR #11243, opened 2022-01-20
[vds] Change definitions of coverage in interval_coverage and sample_qc

tpoterba:
Bins now use `>=` comparisons instead of `>`

PR #11245, opened 2022-01-21
[vds] Introduce DP calculations to vds.sample_qc, change field names

tpoterba:
Match interval_coverage, and annotate a global field with the bins used.

Stacked on #11243


chrisvittal (approved):
LGTM, fix style and rebase and we're good to go.

PR #11246, opened 2022-01-21
General properies json object instead of string -> string map

johnc1231:
This makes it so `properties` can be an arbitrary json dict.


johnc1231:
I rewrote all of the 1.6 backwards compatibility files to reflect the change to use boolean values to store whether a table is `distinctlyKeyed`

johnc1231:
This is an ok thing to do since we haven't released since those 1.6 files were created in #11151

tpoterba (approved):

PR #11247, opened 2022-01-21
Introduce Hail version of ggplot

johnc1231:


hail/python/hail/gg/geoms.py
```diff
@@ -0,0 +1,501 @@
...
+class FigureAttribute:
+    pass
+
+
+class Geom(FigureAttribute):
+
+    def __init__(self, aes):
+        self.aes = aes
+
+    @abc.abstractmethod
```
danking, line 20: If you subclass abc.ABC you get some sanity checks:
```
In [1]: import abc                                                                                                                                                                                                                             

In [2]: class A: 
   ...:     @abc.abstractmethod 
   ...:     def foo(): 
   ...:         pass 
   ...:                                                                                                                                                                                                                                        

In [3]: A()                                                                                                                                                                                                                                    
Out[3]: <__main__.A at 0x7fe48736ab90>

In [4]: class B(abc.ABC): 
   ...:     @abc.abstractmethod 
   ...:     def foo(): 
   ...:         pass 
   ...:                                                                                                                                                                                                                                        

In [5]: B()                                                                                                                                                                                                                                    
  File "<ipython-input-5-a9cb5187b7ad>", line 1, in <module>
    B()
TypeError: Can't instantiate abstract class B with abstract methods foo

```

hail/python/hail/gg/scale.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class PositionScale(Scale):
+    def __init__(self, aesthetic_name, name, breaks, labels):
+        super().__init__(aesthetic_name)
+        self.name = name
+        self.breaks = breaks
+        self.labels = labels
+
+    def update_axis(self, fig):
```
danking, line 32: this seems like its not meant to be used only by subclasses, right? Let's prefix it with an underscore to indicate that.

hail/python/hail/genetics/reference_genome.py
```diff
@@ -226,8 +226,8 @@ def contig_length(self, contig):
         else:
             raise KeyError("Contig `{}' is not in reference genome.".format(contig))
 
-    @typecheck_method(contig=str)
-    def _contig_global_position(self, contig):
+    @property
+    def _global_positions_dict(self):
```
danking, line 230: Hmm. Any reason to not make this part of the public API (make it not underscore)? If we're referencing it in the plotting functionality, then it must be generally useful, yeah?

danking (commented):
a few drive-by thoughts. It'd be great to add type annotations as well! I find they're useful when exploring a new library, but I realize they're pretty rare in Hail Query code.

hail/python/hail/genetics/reference_genome.py
johnc1231: Forgot this was in there, modified quickly as part of a "genomic" scale experiment. Will clean up and make it not `_`.

hail/python/hail/gg/geoms.py
johnc1231: Good call, will do.

hail/python/hail/gg/scale.py
johnc1231: It's meant to be used by `GGPlot` in order to update the axes of a plot with things like ticks, labels, etc.

johnc1231:
Addressed @danking's comments, though I'm not going to attempt to add type information in this PR, given intended timeline for release and other commitments.

hail/python/hail/ggplot/utils.py
```diff
@@ -0,0 +1,46 @@
+import plotly
+import hail as hl
+
+
+def check_scale_continuity(scale, dtype, aes_key):
+
+    if scale.is_discrete() and not is_discrete_type(dtype):
+        raise ValueError(f"Aesthetic {aes_key} has continuous dtype but non continuous scale")
+    if not scale.is_discrete() and is_discrete_type(dtype):
```
patrick-schultz, line 9: Should this check `not is_continuous_type(dtype)` instead? More generally, `is_discrete_type` and `is_continuous_type` are not complements, and it's not clear to me if that is intentional.

patrick-schultz (changes requested):
This is really great! I know we want to release this today, so just a few small things.

hail/python/hail/ggplot/utils.py
johnc1231: Yeah, this is true. Just went and added a `is_continuous` method on scale to address this. I think they were briefly complementary though that was an oversimplification

johnc1231:
@patrick-schultz Addressed your comments, and made the docs a little better.

patrick-schultz (approved):
🎉

PR #11248, opened 2022-01-21
[compiler] IndexWriter: don't flush after every record

tpoterba:


danking (changes requested):
We appear to have two load-bearing flushes.

Hmm. Do we ever close the OutputBuffer? That seems like a bug.

hail/src/main/scala/is/hail/io/index/IndexWriter.scala
```diff
@@ -370,6 +370,8 @@ class StagedIndexWriter(branchingFactor: Int, keyType: PType, annotationType: PT
   }
   def close(cb: EmitCodeBuilder): Unit = {
     val off = flush.invokeCode[Long](cb)
+    cb += ob.flush()
+    cb += ob.close()
```
danking, line 374: I think we lost the removal of the other flushes in writeLeafNode and writeInternalNode.

danking (changes requested):

hail/src/main/scala/is/hail/io/index/IndexWriter.scala
tpoterba: 🤦 

no wonder it worked...

tpoterba:
understand the issue -- need to fix the buffer spec used by the index writer.

PR #11249, opened 2022-01-21
[auth] Add rest endpoint and cli for create / delete users

jigold:
I wanted to get this feature done quickly for Michael. But I do think we should think about having consistency in route names across services. Like when do we have `/delete` in the route name versus a 'DELETE' request or both?

Another question is whether to add the polling / waiting for the user to be created and if so, where should the polling go? I put it in the auth service for now rather than the CLI because I don't think we expose the state of the user in our API.

I tested this as much as I could in my namespace.

The most important thing is to make sure the decorators are correct and only developers can create / delete users. I want to add auth tests that make sure the routes are protected, but that's more work than I wanted to do in this PR.

### DIFF
--- a/auth/auth/auth.py
+++ b/auth/auth/auth.py
@@ -2,6 +2,7 @@ import asyncio
 import json
 import logging
 import os
+from typing import List, Optional
 
 import aiohttp
 import aiohttp_session
@@ -11,10 +12,12 @@ from prometheus_async.aio.web import server_stats  # type: ignore
 
 from gear import (
     Database,
+    Transaction,
     check_csrf_token,
     create_session,
     maybe_parse_bearer_header,
     monitor_endpoints_middleware,
+    rest_authenticated_developers_only,
     rest_authenticated_users_only,
     setup_aiohttp_session,
     transaction,
@@ -30,6 +33,18 @@ from hailtop.tls import internal_server_ssl_context
 from hailtop.utils import secret_alnum_string
 from web_common import render_template, set_message, setup_aiohttp_jinja2, setup_common_static_routes
 
+from .exceptions import (
+    AuthUserError,
+    DuplicateLoginID,
+    DuplicateUsername,
+    EmptyLoginID,
+    InvalidType,
+    InvalidUsername,
+    MultipleExistingUsers,
+    MultipleUserTypes,
+    PreviouslyDeletedUser,
+    UnknownUser,
+)
 from .flow import get_flow_client
 
 log = logging.getLogger('auth')
@@ -50,6 +65,81 @@ async def user_from_login_id(db, login_id):
         return users[0]
     assert len(users) == 0, users
     return None
+
+
+async def users_with_username_or_login_id(tx: Transaction, username: str, login_id: Optional[str]) -> List[dict]:
+    where_conditions = ['username = %s']
+    where_args = [username]
+
+    if login_id is not None:
+        where_conditions.append('login_id = %s')
+        where_args.append(login_id)
+
+    existing_users = [
+        x
+        async for x in tx.execute_and_fetchall(
+            f"SELECT * FROM users WHERE {' OR '.join(where_conditions)} LOCK IN SHARE MODE;", where_args
+        )
+    ]
+
+    return existing_users
+
+
+async def check_valid_new_user(tx: Transaction, username, login_id, is_developer, is_service_account) -> Optional[dict]:
+    if not isinstance(username, str):
+        raise InvalidType('username', username, 'str')
+    if login_id is not None and not isinstance(login_id, str):
+        raise InvalidType('login_id', login_id, 'str')
+    if not isinstance(is_developer, bool):
+        raise InvalidType('is_developer', is_developer, 'bool')
+    if not isinstance(is_service_account, bool):
+        raise InvalidType('is_service_account', is_service_account, 'bool')
+    if is_developer and is_service_account:
+        raise MultipleUserTypes(username)
+    if not is_service_account and not login_id:
+        raise EmptyLoginID(username)
+    if not username or not all(c for c in username if c.isalnum()):
+        raise InvalidUsername(username)
+
+    existing_users = await users_with_username_or_login_id(tx, username, login_id)
+
+    if len(existing_users) > 1:
+        raise MultipleExistingUsers(username, login_id)
+
+    if len(existing_users) == 1:
+        existing_user = existing_users[0]
+        expected_username = existing_user['username']
+        expected_login_id = existing_user['login_id']
+        if username != expected_username:
+            raise DuplicateLoginID(expected_username, login_id)
+        if login_id != expected_login_id:
+            raise DuplicateUsername(username, expected_login_id)
+        if existing_user['state'] in ('deleting', 'deleted'):
+            raise PreviouslyDeletedUser(username)
+        return existing_user
+
+    return None
+
+
+async def insert_new_user(
+    db: Database, username: str, login_id: Optional[str], is_developer: bool, is_service_account: bool
+) -> bool:
+    @transaction(db)
+    async def _insert(tx):
+        existing_user = await check_valid_new_user(tx, username, login_id, is_developer, is_service_account)
+        if existing_user is not None:
+            return False
+
+        await tx.execute_insertone(
+            '''
+INSERT INTO users (state, username, login_id, is_developer, is_service_account)
+VALUES (%s, %s, %s, %s, %s);
+''',
+            ('creating', username, login_id, is_developer, is_service_account),
+        )
+
+    await _insert()  # pylint: disable=no-value-for-parameter
+    return True
 
 
 def cleanup_session(session):
@@ -231,13 +321,11 @@ async def callback(request):
         if domain != ORGANIZATION_DOMAIN:
             raise web.HTTPUnauthorized()
 
-        await db.execute_insertone(
-            '''
-        INSERT INTO users (state, username, login_id, is_developer)
-        VALUES (%s, %s, %s, %s);
-        ''',
-            ('creating', username, login_id, False),
-        )
+        try:
+            await insert_new_user(db, username, login_id, is_developer=False, is_service_account=False)
+        except AuthUserError as e:
+            set_message(session, e.message, 'error')
+            return web.HTTPFound(deploy_config.external_url('notebook', ''))
 
         session['pending'] = True
         session['login_id'] = login_id
@@ -263,6 +351,25 @@ async def callback(request):
     session_id = await create_session(db, user['id'])
     session['session_id'] = session_id
     return aiohttp.web.HTTPFound(next_page)
+
+
+@routes.post('/api/v1alpha/users/{user}/create')
+@rest_authenticated_developers_only
+async def create_user(request: web.Request, userdata):  # pylint: disable=unused-argument
+    db: Database = request.app['db']
+    username = request.match_info['user']
+
+    body = await request.json()
+    login_id = body['login_id']
+    is_developer = body['is_developer']
+    is_service_account = body['is_service_account']
+
+    try:
+        await insert_new_user(db, username, login_id, is_developer, is_service_account)
+    except AuthUserError as e:
+        raise e.http_response()
+
+    return web.json_response()
 
 
 @routes.get('/user')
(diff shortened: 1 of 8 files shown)


auth/auth/auth.py
```diff
@@ -263,6 +272,66 @@ async def callback(request):
...
+        if username != expected_username:
+            raise web.HTTPBadRequest(body=f"user '{expected_username}' already exists with login id '{login_id}'")
+        if login_id != expected_login_id:
+            raise web.HTTPBadRequest(
+                body=f"user '{username}' already exists with a different login id '{expected_login_id}'"
+            )
+    elif len(existing_users) == 0:
+        await db.execute_insertone(
+            '''
+        INSERT INTO users (state, username, login_id, is_developer)
```
illusional, line 316: Thanks for putting this together @jigold! It doesn't look like this handles the `is_service_account` case: https://github.com/hail-is/hail/blob/9aa1e6acd7e2e6c443f0c501ae6c97359641f50c/auth/auth/auth.py#L465-L466

Is it worth combining the logic within these two functions? It doesn't look like `post_create_user` explicitly checks for existing users.

auth/auth/auth.py
```diff
@@ -263,6 +272,66 @@ async def callback(request):
...
+        )
+    else:
+        raise web.HTTPInternalServerError()
+
+    delay = 5
+    while True:
+        user = await user_from_login_id(db, login_id)
+        if user['state'] != 'creating':
+            break
+        delay = await sleep_and_backoff(delay, max_delay=30)
```
illusional, line 329: We've seen cases where the user doesn't move out of a 'creating' state (and we have to check the Hail batch logs directly) eg, this might hang until a timeout is reached:

- A quota has been reached for project number <project-id>: Service accounts per project
- The service account name was too long.

lgruen: (I think another case was about invalid characters in the service account name (e.g. `michael.franklin` having a dot in the user name, which then got translated to an invalid service account email address), but maybe that has already been fixed.)

auth/auth/auth.py
```diff
@@ -263,6 +272,66 @@ async def callback(request):
...
+    username = request.match_info['user']
+
+    body = await request.json()
+    login_id = body['login_id']
+    is_developer = body['is_developer']
+    is_service_account = body['is_service_account']
+
+    if not username:
+        raise web.HTTPBadRequest(body='username must be a non-zero length string')
+    if not isinstance(login_id, str) or not login_id:
```
illusional, line 288: I think login_id can be blank for for service-accounts here.

auth/auth/auth.py
```diff
@@ -229,6 +296,12 @@ async def callback(request):
         if domain != ORGANIZATION_DOMAIN:
             raise web.HTTPUnauthorized()
 
+        try:
+            await check_valid_new_user(db, username, login_id, is_developer=False, is_service_account=False)
+        except AuthUserError as e:
+            set_message(session, e.message, 'error')
+            return web.HTTPFound(deploy_config.external_url('notebook', ''))
```
jigold, line 303: I will fix the notebook redirect in a future PR. I just wanted it to be consistent for now.

jigold:
@illusional @lgruen Thanks for the feedback. I think the new changes address your comments, but let me know if you see something else.

auth/auth/auth.py
```diff
@@ -263,6 +336,48 @@ async def callback(request):
...
+
+    try:
+        await check_valid_new_user(db, username, login_id, is_developer, is_service_account)
+    except AuthUserError as e:
+        raise e.http_response()
+
+    await db.execute_insertone(
+        '''
+INSERT INTO users (state, username, login_id, is_developer, is_service_account)
+VALUES (%s, %s, %s, %s);
```
danking, line 358: need five %s's.

auth/auth/auth.py
```diff
@@ -263,6 +336,48 @@ async def callback(request):
...
+            user = await user_from_login_id(db, login_id)
+            if user['state'] != 'creating':
+                break
+            delay = await sleep_and_backoff(delay, max_delay=30)
+        return user
+
+    try:
+        user = await asyncio.wait_for(_create(), 300)
+    except asyncio.TimeoutError:
+        raise web.HTTPInternalServerError(body=f"timeout creating user '{username}' with login_id '{login_id}'")
```
danking, line 375: As a general design principle, I think HTTP endpoints should return immediately. We should add a separate endpoint `GET /api/v1alpha/users/{user}` which returns the user dict to a developer. The developer can poll that if waiting for creation is necessary.

auth/auth/auth.py
```diff
@@ -263,6 +336,48 @@ async def callback(request):
     return aiohttp.web.HTTPFound(next_page)
 
 
+@routes.post('/api/v1alpha/user/{user}/create')
```
danking, line 339: We use the plural form of the resource: `/api/v1alpha/users/{user}/create`

auth/auth/auth.py
```diff
@@ -431,6 +545,35 @@ async def delete_user(request, userdata):  # pylint: disable=unused-argument
     return web.HTTPFound(deploy_config.external_url('auth', '/users'))
 
 
+@routes.delete('/api/v1alpha/user/{user}')
```
danking, line 548: This is the correct way to write an API endpoint. The billing projects API endpoint should never have been a POST. POSTs should only be used for the web endpoints so that a web form can be used (which only supports GET and POST for silly historical reasons).

auth/auth/auth.py
```diff
@@ -431,6 +545,35 @@ async def delete_user(request, userdata):  # pylint: disable=unused-argument
...
+    if n_rows != 1:
+        assert n_rows == 0
+        raise web.HTTPBadRequest(body=f"Delete failed, no such user '{username}'")
+
+    delay = 5
+    while True:
+        user = await user_from_username(db, username)
+        if user['state'] != 'deleting':
+            break
+        delay = await sleep_and_backoff(delay, max_delay=30)
```
danking, line 571: No polling, the request should return immediately. If the user wants to poll for deletion, they can do that separately.

auth/auth/exceptions.py
```diff
@@ -0,0 +1,52 @@
...
+
+class MultipleUserTypes(AuthUserError):
+    def __init__(self, username):
+        super().__init__(f"User '{username}' cannot be both a developer and a service account.", 'error')
+
+
+class MultipleExistingUsers(AuthUserError):
+    def __init__(self, username, login_id):
+        super().__init__(
+            f"User with user name '{username}' and login id '{login_id}' appear multiple times in the database.",
```
danking, line 47: text seems a bit odd. maybe "Multiple users with user name ... appear in the database.

auth/auth/exceptions.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class InvalidUsername(AuthUserError):
+    def __init__(self, username):
+        super().__init__(f"Invalid username '{username}'. Must be a non-empty alphanumeric string.", 'error')
+
+
+class InvalidType(AuthUserError):
+    def __init__(self, field_name, input, expected_type):
+        super().__init__(f"Expected '{field_name}' is type {expected_type}. Found type {type(input)}", 'error')
```
danking, line 36: is of type

auth/auth/auth.py
```diff
@@ -50,6 +63,60 @@ async def user_from_login_id(db, login_id):
...
+    existing_users = await users_with_username_or_login_id(db, username, login_id)
+
+    if len(existing_users) == 1:
+        existing_user = existing_users[0]
+        expected_username = existing_user['username']
+        expected_login_id = existing_user['login_id']
+        if username != expected_username:
+            raise DuplicateLoginID(username, login_id)
+        if login_id != expected_login_id:
+            raise DuplicateUsername(username, login_id)
```
danking, line 115: In the message you claim that the login_id is the one that already exists, so I think this should be

```
raise DuplicateUsername(username, existing_user['login_id'])
```

auth/auth/auth.py
```diff
@@ -50,6 +63,60 @@ async def user_from_login_id(db, login_id):
...
+        raise InvalidUsername(username)
+
+    existing_users = await users_with_username_or_login_id(db, username, login_id)
+
+    if len(existing_users) == 1:
+        existing_user = existing_users[0]
+        expected_username = existing_user['username']
+        expected_login_id = existing_user['login_id']
+        if username != expected_username:
+            raise DuplicateLoginID(username, login_id)
```
danking, line 113: In the message you claim that the username is the username of the already extant user, so I think this should be:

```
raise DuplicateLoginID(existing_user['username'], login_id)
```

danking (changes requested):
Looks good, but I strongly prefer non-blocking endpoints.

auth/auth/auth.py
```diff
@@ -381,27 +487,86 @@ async def post_create_user(request, userdata):  # pylint: disable=unused-argumen
...
+
+    user = await db.select_and_fetchone(
+        '''
+SELECT id, username, login_id, state, is_developer, is_service_account FROM users
+WHERE username = %s;
 ''',
-        ('creating', username, login_id, is_developer, is_service_account),
+        (username,),
     )
+    return web.json_response(user)
```
danking, line 540: prefer `if user is None: raise aiohttp.HTTPNotFound` or w/e that exception is called

auth/auth/auth.py
```diff
@@ -381,27 +487,86 @@ async def post_create_user(request, userdata):  # pylint: disable=unused-argumen
...
+
+                if id is not None:
+                    where_conditions.append('id = %s')
+                    where_args.append(id)
+
+                n_rows = await tx.execute_update(
+                    f'''
+UPDATE users
+SET state = 'deleting'
+WHERE {' AND '.join(where_conditions)};
```
danking, line 560: you don't lock the users table, so the user could change from deleting to deleted before you run this query and then you'd change it back to deleting, which seems wrong to me. username is a unique key on the users table, so it seems to me that we should just do:
```
UPDATE users
SET state = 'deleting'
WHERE state != 'deleted' AND username = %s
```
that's guaranteed to be atomic, so no other operation can race in and change deleting to deleted and then you change deleted back to deleting.

We can use the updated row count to throw an unknown user error.

danking (dismissed):

jigold:
bump

auth/auth/auth.py
```diff
@@ -263,6 +341,36 @@ async def callback(request):
...
+    if existing_user is not None:
+        return web.json_response()
+
+    await db.execute_insertone(
+        '''
+INSERT INTO users (state, username, login_id, is_developer, is_service_account)
+VALUES (%s, %s, %s, %s, %s);
+''',
+        ('creating', username, login_id, is_developer, is_service_account),
+    )
```
danking, line 369: This races with the select in `check_valid_new_user`. MySQL's [repeatable read](https://dev.mysql.com/doc/refman/8.0/en/innodb-consistent-read.html) isolation level does not guarantee that, if we a `SELECT` returns no rows that a subsequent `INSERT` will not cause a duplicate key error. See their example about `SELECT` and `UPDATE` where the update affects rows not seen by the select.

The `SELECT` in `users_with_username_or_login_id` needs a `LOCK IN SHARE MODE` to prevent other transactions from modifying/inserting the relevant rows.

I also think we should rework this so that both user creation functions use a common insertion function:

```python3
async def insert_new_user(tx, username, login_id, is_developer, is_service_account) -> Optional[dict]:
    existing_user = check_valid_new_user(tx, ...)
    if existing_user is not None:
        return existing_user
    db.execute_insertone(...)
    return None
```

The UI call should convert login_id to None before the call to `check_valid_new_user` anyway. I I think both the UI call and the API call can be implemented using `insert_new_user`.

auth/auth/auth.py
```diff
@@ -381,27 +489,84 @@ async def post_create_user(request, userdata):  # pylint: disable=unused-argumen
...
+    if user is None:
+        raise web.HTTPNotFound()
+    return web.json_response(user)
 
-    set_message(session, f'Created user {user_id} {username} {login_id}.', 'info')
 
-    return web.HTTPFound(deploy_config.external_url('auth', '/users'))
+async def _delete_user(db: Database, username: str, id: Optional[str]):
+    @transaction(db)
+    async def _delete(tx):
```
danking, line 549: there's only one operation in this block so there's no need for a transaction.

auth/auth/auth.py
```diff
@@ -414,23 +579,29 @@ async def delete_user(request, userdata):  # pylint: disable=unused-argument
...
+@routes.delete('/api/v1alpha/users/{user}')
+@rest_authenticated_developers_only
+async def rest_delete_user(request: web.Request, userdata):  # pylint: disable=unused-argument
+    db = request.app['db']
+    username = request.match_info['user']
+
+    try:
+        await _delete_user(db, username, None)
+    except UnknownUser as e:
+        return e.http_response()
```
danking, line 600: I think aiohttp warns if you return an error instead of `raise`-ing it.

hail/python/hailtop/hailctl/auth/create_user.py
```diff
@@ -0,0 +1,46 @@
...
+            delay = 5
+            while True:
+                user = await async_get_user(args.username, args.namespace)
+                if user['state'] == 'active':
+                    print(f"Created user '{args.username}'")
+                    return
+                assert user['state'] == 'creating'
+                delay = await sleep_and_backoff(delay)
+
+        await asyncio.wait_for(_poll(), 300)
```
danking, line 39: I'm open to arguments in favor, but I don't think we should have this timeout. The HTTP requests themselves already have time outs. This only triggers if it takes longer than 5 minutes to create a user account. I'm not sure there's a compelling reason for that. I would expect a CLI tool to just wait until it completes. I can always ctrl-c if I want to abort. If this is used inside a test or w/e, that test will have a timeout.

danking (dismissed):

auth/auth/auth.py
jigold: From the docs:
https://docs.aiohttp.org/en/v0.14.3/web.html#exceptions

You can do either. I'll fix these to raise, but we ran into problems with the session message showing up if we raised errors.

danking:
@jigold needs rebase

PR #11250, opened 2022-01-21
[batch] Simplify iptables rules for worker network namespaces

daniel-goldstein:
# Commit 1

This rule that we add per job network namespace

> iptables -w {IPTABLES_WAIT_TIMEOUT_SECS} --append FORWARD --out-interface {self.veth_host} --in-interface {self.internet_interface} --jump ACCEPT

is functionally equivalent to this single line 

> iptables --append FORWARD --in-interface $INTERNET_INTERFACE --destination 172.20.0.0/15 --jump ACCEPT

so better have 1 rule than O(slots) rules.

It would be nice to eliminate the one remaining rule that we append per-network namespace. Without that rule [this test](https://github.com/hail-is/hail/blob/9da1c759efe9fcbbd9ad02ee7c8bf725974270e9/batch/test/test_dag.py#L117) will fail because the iptables rules won't allow `curl $HAIL_BATCH_WORKER_IP:$HAIL_BATCH_WORKER_PORT` from inside the job container. The request leaves the job's network namespace to the host (whose IP it is), and then returns right back to the network namespace. I have this strict rule such that jobs cannot communicate with each other, but so far don't know how in iptables to specify that the sending/receiving interfaces or IPs must match without listing them explicitly.

That being said, this is also a side-effect of `HAIL_BATCH_WORKER_IP` being an in-cluster ip address `10.128.x.x` and thus known to the host. I believe if we made this the public IP (forcing the request to leave through the ethernet device I think) the test could work without this rule. Jobs can still communicate with themselves within their own network namespace (localhost still works fine), and we strictly prohibit workers talking to each other, so I don't see any reason why we shouldn't go the public ip approach.

# Commit 2
This gives the external ip address to a job container that opens a port instead of using the private one. This just seems to make more sense to me and also allows us to delete that iptables rule. This ensures that iptables rules don't grow linearly with number of jobs (they'll be 1:1 with jobs that open ports but that makes more sense).

### DIFF
--- a/batch/batch/cloud/azure/driver/create_instance.py
+++ b/batch/batch/cloud/azure/driver/create_instance.py
@@ -202,6 +202,8 @@ iptables --append FORWARD --destination $IP_ADDRESS --jump ACCEPT
 # Forbid outgoing requests to cluster-internal IP addresses
 INTERNET_INTERFACE=eth0
 iptables --append FORWARD --out-interface $INTERNET_INTERFACE ! --destination 10.128.0.0/16 --jump ACCEPT
+# Allow incoming packets to job network namespaces only from the internet
+iptables --append FORWARD --in-interface $INTERNET_INTERFACE --destination 172.20.0.0/15 --jump ACCEPT
 
 cat >> /etc/hosts <<EOF
 $INTERNAL_GATEWAY_IP batch-driver.hail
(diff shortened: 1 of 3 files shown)


batch/batch/cloud/azure/driver/create_instance.py
```diff
@@ -202,6 +203,8 @@ def create_vm_config(
 # Forbid outgoing requests to cluster-internal IP addresses
 INTERNET_INTERFACE=eth0
 iptables --append FORWARD --out-interface $INTERNET_INTERFACE ! --destination 10.128.0.0/16 --jump ACCEPT
+# Allow incoming packets to job network namespaces only from the internet
+iptables --append FORWARD --in-interface $INTERNET_INTERFACE --destination 172.20.0.0/15 --jump ACCEPT
```
danking, line 207: Mmmm. I'm a bit confused. How does the internet interface receive a packet destined for 172.20.0.1? The internet does not know this machine can serve that IP address.

danking: OK, I think I get it. This page helped: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/4/html/security_guide/s1-firewall-ipt-fwd

The PREROUTING chain effectively rewrites the IP address *and then* puts the packet through the FORWARD chain. That's how we the public internet interface generates an internal IP address. In fact, since Google's routing system would never deliver us packets for 172.20.0.1, we *know* that this packet came from our own PREROUTING rules. Does that all sound right to you?

danking (approved):

danking:
Can we follow up with a test that creates one job which echos its IP and port and then listens, then the test will try to curl the public IP and port?

PR #11251, opened 2022-01-21
[batch] Dont block on all netns creation on startup

daniel-goldstein:
`NetworkAllocator.reserve` creates a bunch of network namespaces on startup and adds them to an `asyncio.Queue`. There's no reason to wait for all of them before accepting jobs.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -2256,7 +2256,8 @@ class Worker:
         await site.start()
 
         try:
-            await asyncio.wait_for(self.activate(), MAX_IDLE_TIME_MSECS / 1000)
+            startup_tasks = asyncio.gather(self.activate(), network_allocator.reserve())
+            await asyncio.wait_for(startup_tasks, MAX_IDLE_TIME_MSECS / 1000)
         except asyncio.TimeoutError:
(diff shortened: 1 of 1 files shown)


batch/batch/worker/worker.py
```diff
@@ -2178,9 +2178,9 @@ async def async_main():
 
     port_allocator = PortAllocator()
     network_allocator = NetworkAllocator()
-    await network_allocator.reserve()
 
     worker = Worker(httpx.client_session())
+    worker.task_manager.ensure_future(network_allocator.reserve())
```
danking, line 2183: If there's an error in reserve, we'll never see it (we throw away the future). How much time do we spend waiting on this? Maybe do it in parallel with activating (i.e. `gather` the two tasks)?

danking (dismissed):

danking:
@daniel-goldstein Bump, let's resolve some old PRs. I think the gather change is relatively small and will yield an improvement.

danking (dismissed):
yep, that's exactly what I had in mind

danking:
This PR is failing because we activate (telling driver we can accept jobs) but don't immediately start the web server (causing requests to fail). Activation implies we're ready for jobs. Are we actually ready for jobs before the netns stuff is created?

danking (dismissed):
see comment

danking (approved):

PR #11253, opened 2022-01-24
[batch] Support Azure disks mounted per job

jigold:
@danking Can you do an initial pass and make sure there aren't any major issues with the Python code? I am most concerned with how I setup the disk manager and the disk creation / deletion code to avoid costly mistakes. I am aware of the FIXMEs. I haven't tested this at all yet. It would be great if I could have feedback on Wednesday so I can have this PR completely done on Friday.

### DIFF
--- a/batch/batch/cloud/azure/resource_utils.py
+++ b/batch/batch/cloud/azure/resource_utils.py
@@ -2,7 +2,7 @@ import re
 import logging
 from typing import Optional, Tuple, Dict
 import sortedcontainers
-
+import math
 from ...globals import RESERVED_STORAGE_GB_PER_CORE
 
 log = logging.getLogger('resource_utils')
@@ -102,6 +102,12 @@ def azure_disk_from_storage_in_gib(disk_type: str, storage_in_gib: int) -> Optio
     return disks[index]
 
 
+def azure_disk_from_storage_in_bytes(disk_type: str, storage_in_bytes: int) -> Optional[AzureDisk]:
+    storage_in_gib = storage_in_bytes / 1024 / 1024 / 1024
+    storage_in_gib = math.ceil(storage_in_gib)
+    return azure_disk_from_storage_in_gib(disk_type, storage_in_gib)
+
+
 class MachineTypeParts:
     @staticmethod
     def from_dict(data: dict) -> 'MachineTypeParts':
@@ -197,8 +203,13 @@ def azure_requested_to_actual_storage_bytes(storage_bytes, allow_zero_storage):
         return None
     if allow_zero_storage and storage_bytes == 0:
         return storage_bytes
+
     # actual minimum storage size is 4 Gi on Azure, but keeping 10 to be consistent with gcp
-    return max(10 * 1024 ** 3, storage_bytes)
+    storage_bytes = max(10 * 1024 ** 3, storage_bytes)
+
+    # assuming Premium disks because this is used for computing user's external storage
+    disk = azure_disk_from_storage_in_gib('P', storage_bytes)
+    return disk.size_in_gib * 1024 ** 3
 
 
 def azure_is_valid_storage_request(storage_in_gib: int) -> bool:
--- a/batch/batch/cloud/azure/resource_utils.py
+++ b/batch/batch/cloud/azure/resource_utils.py
@@ -103,8 +103,8 @@ def azure_disk_from_storage_in_gib(disk_type: str, storage_in_gib: int) -> Optio
 
 
 def azure_disk_from_storage_in_bytes(disk_type: str, storage_in_bytes: int) -> Optional[AzureDisk]:
-    storage_in_gib = storage_in_bytes / 1024 / 1024 / 1024
-    storage_in_gib = math.ceil(storage_in_gib)
+    one_gib = 1024 * 1024 * 1024
+    storage_in_gib = (storage_in_bytes - 1 + one_gib) // one_gib
     return azure_disk_from_storage_in_gib(disk_type, storage_in_gib)
 
 
@@ -203,9 +203,6 @@ def azure_requested_to_actual_storage_bytes(storage_bytes, allow_zero_storage):
         return None
     if allow_zero_storage and storage_bytes == 0:
         return storage_bytes
-
-    # actual minimum storage size is 4 Gi on Azure, but keeping 10 to be consistent with gcp
-    storage_bytes = max(10 * 1024 ** 3, storage_bytes)
 
     # assuming Premium disks because this is used for computing user's external storage
     disk = azure_disk_from_storage_in_gib('P', storage_bytes)
--- a/batch/batch/cloud/azure/worker/disk.py
+++ b/batch/batch/cloud/azure/worker/disk.py
@@ -1,27 +1,185 @@
-import logging
+from typing import Dict, List, Optional
 
-from ....worker.disk import CloudDisk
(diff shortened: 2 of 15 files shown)


danking:
looking now

batch/batch/worker/worker.py
```diff
@@ -1496,14 +1501,12 @@ async def run(self):
                 with self.step('post-job finally block'):
                     if self.disk:
                         try:
-                            await self.disk.delete()
+                            await asyncio.wait_for(await CLOUD_WORKER_API.delete_disk(self.disk), DISK_OPERATION_TIMEOUT_SECONDS)
```
danking, line 1504: wait_for should take an awaitable. As written, this code would wait for delete disk and then, I think, error because `wait_for` expects an awaitable.

batch/batch/worker/worker.py
```diff
@@ -167,6 +167,8 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
 
 image_lock = aiorwlock.RWLock()
 
+DISK_OPERATION_TIMEOUT_SECONDS = 10 * 60
```
danking, line 170: do these operations actually take ten minutes? That seems... wrong? Do these endpoints not return immediately? Do they wait for the disk to actually be created or deleted?

batch/batch/worker/instance_env.py
```diff
@@ -9,23 +9,25 @@
...
         raise NotImplementedError
 
     @abc.abstractmethod
-    def create_disk(self, instance_name: str, disk_name: str, size_in_gb: int, mount_path: str) -> CloudDisk:
+    async def new_disk(self, instance_name: str, disk_name: str, size_in_gb: int, mount_path: str) -> CloudDisk:
         raise NotImplementedError
 
     @abc.abstractmethod
-    def user_credentials(self, credentials: Dict[str, bytes]) -> CloudUserCredentials:
+    async def delete_disk(self, disk: CloudDisk):
```
danking, line 24: The way to do this is a type parameter:
```python3
TDisk = TypeVar('TDisk', bound=CloudDisk)
class CloudWorkerAPI(abc.ABC, Generic[TDisk]):
    # ...
    def new_disk(...) -> TDisk
        # ...
    def delete_disk(self, disk: TDisk):
        # ...
```

batch/batch/cloud/gcp/worker/disk.py
```diff
@@ -94,19 +76,41 @@ async def _attach(self):
...
+        return GCPDisk(disk_name, self.zone, self.project, instance_name, size_in_gb, mount_path)
+
+    async def delete_disk(self, disk: GCPDisk):  # type: ignore[override]
+        try:
+            await disk.delete()
+        finally:
+            try:
+                self.disk_slots.release()
+            finally:
+                await disk.close()
```
danking, line 116: I don't see any reason to delete the disk before release but not close it. I think we should do both before releasing the semaphore.

batch/batch/cloud/azure/resource_utils.py
```diff
@@ -197,8 +203,13 @@ def azure_requested_to_actual_storage_bytes(storage_bytes, allow_zero_storage):
         return None
     if allow_zero_storage and storage_bytes == 0:
         return storage_bytes
+
     # actual minimum storage size is 4 Gi on Azure, but keeping 10 to be consistent with gcp
-    return max(10 * 1024 ** 3, storage_bytes)
+    storage_bytes = max(10 * 1024 ** 3, storage_bytes)
+
+    # assuming Premium disks because this is used for computing user's external storage
+    disk = azure_disk_from_storage_in_gib('P', storage_bytes)
```
danking, line 211: We need to make sure the docs are crystal clear about what `.storage` means, because this suggests to me that it means "you will receive *at least* this much disk space, but you may pay for twice as much"

To be clear, I don't think that's a problem, in fact, I think powers of two of disk is a good idea. We might be able to do some kind of packing thing with powers of two disks.

batch/batch/cloud/azure/resource_utils.py
```diff
@@ -102,6 +102,12 @@ def azure_disk_from_storage_in_gib(disk_type: str, storage_in_gib: int) -> Optio
     return disks[index]
 
 
+def azure_disk_from_storage_in_bytes(disk_type: str, storage_in_bytes: int) -> Optional[AzureDisk]:
+    storage_in_gib = storage_in_bytes / 1024 / 1024 / 1024
+    storage_in_gib = math.ceil(storage_in_gib)
```
danking, line 107: ehhhh I don't like going through floating point for this kind of calculation. seems wrong when we can do this with integral arithmetic.

```
one_gib = 1024 * 1024 * 1024
storage_in_gib = (storage_in_bytes - 1 + one_gib) // one_gib
```

batch/batch/cloud/azure/worker/disk.py
```diff
@@ -1,27 +1,185 @@
...
+        matches = [disk['name'] == self.name for disk in disks]
+        if len(matches) == 0:
+            return None
+        return matches[0]
+
+    async def _create_and_attach(self):
+        async with LoggingTimer(f'creating disk {self.name}'):
+            async with self.disk_manager.update_lock:
+                new_vm_disk_config = self.disk_manager._add_disk(self)
+                await self.compute_client.patch(f'/virtualMachines/{self.instance_name}', json=new_vm_disk_config)
```
danking, line 66: Oof, it's really annoying that we have to track all the disks.

batch/batch/cloud/azure/worker/disk.py
```diff
@@ -1,27 +1,185 @@
...
+                }
+            }
+        }
+
+    def _add_disk(self, disk: AzureDisk) -> dict:
+        self.disk_mapping[disk.name] = disk.spec()
+        return self.vm_disk_config()
+
+    def _remove_disk(self, disk: AzureDisk) -> dict:
+        self.disk_mapping.pop(disk.name)
```
danking, line 168: `del self.disk_mapping[disk.name]` should have the same semantics and has pleasing symmetry with line 164.

batch/batch/cloud/azure/worker/disk.py
```diff
@@ -1,27 +1,185 @@
...
+        try:
+            await disk.delete()
+        finally:
+            try:
+                if not disk.attached:
+                    self.lun_queue.put_nowait(disk.lun)
+                else:
+                    log.exception(f'disk {disk.name} did not detach cleanly. not releasing lun {disk.lun}')
+            finally:
+                await disk.close()
```
danking, line 185: same bit here about close coming before the semaphore.

danking (changes requested):
A few thoughts, overall looks good!

batch/batch/worker/worker.py
jigold: In GCP, it can take a really long time. I'm not sure why. We could start with 5 minutes?

batch/batch/cloud/azure/resource_utils.py
jigold: I agree that we need to make this clear when we have Azure specific docs. It's the same behavior in GCP right now where you get exactly what you requested.

jigold: I guess it's not hard to add this to the docs right now. Will do so

danking:
fml azure

PR #11259, opened 2022-01-24
[hailctl] permit multiple distinct clusters

danking:
`gettempdir` seems to return the same temporary directory, even for new
Python processes. If the user data dir is re-used, chrome will ignore the
proxy setting passed on the command line.


chrisvittal:
Do you mind if we go with #11264? There's a few improvements there, but I think this is the correct solution rather than using `mkdtemp`.

PR #11260, opened 2022-01-24
[query] Begin work to lower import_gvcfs

chrisvittal:
Minor cleanups, remove unused/unnecessary variables, force/forceBgz have
no use, all files must be bgzip compressed and tabix indexed for
import_gvcfs.


hail/python/hail/methods/impex.py
```diff
@@ -2399,8 +2397,6 @@ def import_vcf(path,
            _external_header=nullable(str))
 def import_gvcfs(path,
                  partitions,
-                 force=False,
-                 force_bgz=False,
```
patrick-schultz, line 2403: Should update the docstring about "the only difference" with the args of `import_vcf` to mention the lack of `force, force_bgz`.

patrick-schultz (changes requested):

patrick-schultz (approved):

PR #11262, opened 2022-01-25
[batch-driver] Add UI Button to Deactivate workers

vrautela:


batch/batch/driver/main.py
```diff
@@ -273,13 +273,37 @@ async def deactivate_instance_1(instance):
...
+
+
+@routes.post('/instances/{instance_name}/deactivate')
+@check_csrf_token
+@web_authenticated_developers_only()
+async def kill_instance(request, userdata):
+    instance_name = request.match_info['instance_name']
+
+    inst_coll_manager: InstanceCollectionManager = request.app['driver'].inst_coll_manager
+    instance = inst_coll_manager.get_instance(instance_name)
```
daniel-goldstein, line 292: I'd like to strive for IDE-navigable types in new code that we introduce. If `get_instance` had types in its signature, then the language server will be able to infer the type of `instance` and we could then navigate to the definition of methods like `instance.kill` below. Could you add a type signature to `get_instance`? It should return an optional instance.

batch/batch/driver/main.py
```diff
@@ -273,13 +273,37 @@ async def deactivate_instance_1(instance):
...
 
 @routes.post('/api/v1alpha/instances/deactivate')
 @active_instances_only
 async def deactivate_instance(request, instance):  # pylint: disable=unused-argument
-    return await asyncio.shield(deactivate_instance_1(instance))
+    await asyncio.shield(deactivate_instance_1(instance))
+    return web.Response()
+
+
+@routes.post('/instances/{instance_name}/deactivate')
```
daniel-goldstein, line 285: Let's change this to kill instead of deactivate to disambiguate what we're actually doing here.

batch/batch/driver/main.py
```diff
@@ -273,13 +273,37 @@ async def deactivate_instance_1(instance):
...
+@check_csrf_token
+@web_authenticated_developers_only()
+async def kill_instance(request, userdata):
```
daniel-goldstein, line 288: You'll need a comment here like on line 280 to tell `pylint` that it's ok we're not using the userdata parameter.

batch/batch/driver/main.py
```diff
@@ -273,13 +273,37 @@ async def deactivate_instance_1(instance):
...
+    instance_name = request.match_info['instance_name']
+
+    inst_coll_manager: InstanceCollectionManager = request.app['driver'].inst_coll_manager
+    instance = inst_coll_manager.get_instance(instance_name)
+
+    if instance:
+        pool_name = instance.inst_coll.name
+        pool_url_path = f'/inst_coll/pool/{pool_name}'
```
daniel-goldstein, line 296: let's define these pool variables directly before they are used.

batch/batch/driver/main.py
```diff
@@ -273,13 +273,37 @@ async def deactivate_instance_1(instance):
...
+        session = await aiohttp_session.get_session(request)
+
+        if instance.state == 'active':
+            await asyncio.shield(instance.kill())
+            set_message(session, f'Deactivated instance {instance_name}', 'info')
+        else:
+            set_message(session, 'Cannot deactivate a non-active instance', 'error')
+    else:
+        return web.HTTPNotFound()
+    return web.HTTPFound(deploy_config.external_url('batch-driver', pool_url_path))
```
daniel-goldstein, line 306: The logic here for the two code paths is unnecessarily interleaved, i.e. I have to skip over lines 304-305 in the case where the instance is not None. I'd prefer we make this as continuous as possible, so move this return into the if block, and then we don't need an `else`, our default return will just be `HTTPNotFound`. Alternatively, we could avoid a large block of indentation by flipping this if around and then having the bulk of the code indented at the normal level:

```python
if instance is None:
    return web.HTTPNotFound()
    
… rest of the function …
```

What do you think?

batch/batch/driver/templates/pool.html
```diff
@@ -115,6 +115,15 @@ <h2>Instances</h2>
     <td class="numeric-cell">{{ instance.failed_request_count }}</td>
     <td>{{ instance.time_created_str() }}</td>
     <td>{{ instance.last_updated_str() }} ago</td>
+    <td>
+      <form action="{{ base_path }}/instances/{{ instance.name }}/deactivate" method="post">
+        <input type="hidden" name="_csrf" value="{{ csrf_token }}" />
+        {% if q is not none %}
+        <input type="hidden" name="q" value="{{ q }}" />
+        {% endif %}
```
daniel-goldstein, line 123: Can delete these three lines involving `q`.

batch/batch/driver/templates/pool.html
```diff
@@ -115,6 +115,15 @@ <h2>Instances</h2>
...
+        <button>Deactivate</button>
```
daniel-goldstein, line 124: This is a bit misleading now since deactivate means something very specific in our application and this is calling `instance.kill`. Delete is a less ambiguous synonym for `kill` that is also a natural button name.

daniel-goldstein (changes requested):
Great job! Mostly aesthetic nits.

batch/batch/driver/templates/pool.html
```diff
@@ -118,10 +118,7 @@ <h2>Instances</h2>
     <td>
       <form action="{{ base_path }}/instances/{{ instance.name }}/deactivate" method="post">
```
daniel-goldstein, line 119: I think this needs to change to `kill` as well.

daniel-goldstein (changes requested):
One last thing

daniel-goldstein (approved):

PR #11265, opened 2022-01-25
[ci] Turn off Azure required status

jigold:


danking (approved):

jigold:
Did we still want this change?
@danking @daniel-goldstein

danking:
I think no. The frequency of failures is a bit lower than it was.

danking:
I restored hail-ci-azure as a GitHub required status.

johnc1231:
I'm fine with this, though I think frequency of failures is still relatively high. I have been hitting retry on PRs when i notice.

PR #11268, opened 2022-01-26
[query] Fix old hl.plot.histogram log argument

johnc1231:
Struct fields are not mutable, so the log argument of hl.plot.histogram would never work. Also have to handle the case where there are 0s

### DIFF
--- a/hail/python/hail/plot/plots.py
+++ b/hail/python/hail/plot/plots.py
@@ -419,8 +419,20 @@ def histogram(data, range=None, bins=50, legend=None, title=None, log=False, int
         legend = ""
 
     if log:
+        bin_freq = []
+        count_problems = 0
+        for x in data.bin_freq:
+            if x == 0.0:
+                count_problems += 1
+                bin_freq.append(x)
+            else:
+                bin_freq.append(math.log10(x))
+
+        if count_problems > 0:
+            warning(f"There were {count_problems} bins with height 0, those cannot be log transformed and were left as 0s.")
+
         changes = {
-            "bin_freq": [math.log10(x) if x > 0.0 else x for x in data.bin_freq],
+            "bin_freq": bin_freq,
             "n_larger": math.log10(data.n_larger) if data.n_larger > 0.0 else data.n_larger,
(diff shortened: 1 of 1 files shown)


hail/python/hail/plot/plots.py
```diff
@@ -419,9 +419,12 @@ def histogram(data, range=None, bins=50, legend=None, title=None, log=False, int
...
     if log:
-        data.bin_freq = [math.log10(x) for x in data.bin_freq]
-        data.n_larger = math.log10(data.n_larger)
-        data.n_smaller = math.log10(data.n_smaller)
+        changes = {
+            "bin_freq": [math.log10(x) if x > 0.0 else x for x in data.bin_freq],
+            "n_larger": math.log10(data.n_larger) if data.n_larger > 0.0 else data.n_larger,
+            "n_smaller": math.log10(data.n_smaller) if data.n_smaller > 0.0 else data.n_smaller
+        }
+        data = data.annotate(**changes)
```
patrick-schultz, line 427: I think this is wrong, because it plots count-0 and count-1 bins the same. I think the right thing to do is `log(x) if x > 0.0 else shift`, where `shift` is arbitrary, but `shift=-1` is as good as anything.

patrick-schultz (dismissed):

johnc1231:
@patrick-schultz , shift = -1 can lead to some pretty ugly graphs when the range of log transformed y values is small (i.e. between 0 and 1)
Screen Shot 2022-01-26 at 2 12 03 PM

johnc1231:
The negative 1 doesn't seem right, but we should probably give a warning or something when a user does this. I could log a warning telling user how many points were 0 and so weren't plotted?

patrick-schultz:
That seems good. Perhaps we could also suggest using `hl.plot.pdf` instead. Just about to test that out.

patrick-schultz:
`hl.plot.pdf` still plots histogram style, not line-plot style, so it still has the issue of where the bottom of the bars go. It would be easy to add a line-plot option, but lets just add the warning for now.

patrick-schultz (approved):

PR #11270, opened 2022-01-26
[query] New Shuffler with Arbitrary Sort Order

johnc1231:
This PR:

- Introduces `StructComparisonOp`, a subclass of `ComparisonOp` that is just for comparing structs. They have an array of sort fields for this purpose.
- Extends `StreamDistribute` to take in a `ComparisonOp`, so that it doesn't just assume data sorted in ascending order.
- Updates `LowerDistributedSort.scala` to actually use the `SortFields` that get passed into it, allowing for arbitrary sort orders.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/ComparisonOp.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/ComparisonOp.scala
@@ -94,59 +94,34 @@ case class Compare(t1: Type, t2: Type) extends ComparisonOp[Int] {
 }
 object Compare { def apply(typ: Type): Compare = Compare(typ, typ) }
 
-case class StructLT(t1: Type, t2: Type, sortFields: Array[SortField]) extends ComparisonOp[Boolean] {
-  val op: CodeOrdering.Op = CodeOrdering.StructLt()
+trait StructComparisonOp[T] extends ComparisonOp[T] {
+  val sortFields: Array[SortField]
 
-  override def codeOrdering(ecb: EmitClassBuilder[_], t1p: SType, t2p: SType): F[op.ReturnType] = {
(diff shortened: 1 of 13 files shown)


hail/src/main/scala/is/hail/expr/ir/ComparisonOp.scala
```diff
@@ -89,4 +92,36 @@ case class Compare(t1: Type, t2: Type) extends ComparisonOp[Int] {
...
+
+  override def codeOrdering(ecb: EmitClassBuilder[_], t1: SType, t2: SType): F[op.ReturnType] = {
+    ComparisonOp.checkCompatible(t1.virtualType, t2.virtualType)
+    ecb.getStructOrderingFunction(t1.asInstanceOf[SBaseStruct], t2.asInstanceOf[SBaseStruct], sortFields, op).asInstanceOf[CodeOrdering.F[op.ReturnType]]
+  }
+}
+
+case class StructCompare(t1: Type, t2: Type, sortFields: Array[SortField]) extends StructComparisonOp[Int] {
+  val op: CodeOrdering.Op = CodeOrdering.StructCompare()
+  override val strict: Boolean = false
```
johnc1231, line 108: Admittedly am not 100% sure what `strict` means in this context.

tpoterba: strict refers to missing semantics -- strict=true means the result is missing if any argument is missing.

tpoterba: I think this all looks correct.

tpoterba (approved):
🚀

PR #11271, opened 2022-01-26
Allocate more workers to testing dataproc

johnc1231:


johnc1231:
Probably we should run the Grch37 and Grch38 parts of this in parallel, but reducing wall clock time by giving some more resources is also a good idea, as some of these tests have 10k partitions.

PR #11273, opened 2022-01-26
[build] Parallelize test-dataproc

johnc1231:
test-dataproc was taking over an hour, for two reasons:

1. Starting two vep clusters sequentially is slow.
2. A very small cluster was working on submits that sometimes had 10k partitions.

To solve this, I've:

1. Split test-dataproc into test-dataproc-37 and test-dataproc-38
2. Added a few preemptibles to cluster create script.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -3905,7 +3905,7 @@ steps:
       - blog_nginx_image
       - create_certs
   - kind: runImage
-    name: test_dataproc
+    name: test_dataproc-37
     image:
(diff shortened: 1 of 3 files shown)


johnc1231:
Since this doesn't get tested by CI normally, running a dev deploy here: https://ci.hail.is/batches/403111

If it succeeds, we should feel good I think.

danking (approved):
LGTM.

PR #11275, opened 2022-01-26
[query] Update deploy to depend on new test dataprocs

johnc1231:
Correcting an oversight from #11273 

Probably we should test that all build.yaml dependencies exist in the future (Services team seems to be replacing build.yaml with something new?)


daniel-goldstein (approved):
This might have cropped up when making certain steps cloud-optional. I'll make sure to look at it.

PR #11278, opened 2022-01-27
[batch] fix n cancelled creating jobs

danking:
I ran the select ... from (select ...) part of the UPDATE which fixes
this issue. It completed in ~72 seconds, so I think this change is actually
plenty fast to run.

### DIFF
--- a/batch/sql/fix-n-cancelled-creating-jobs.sql
+++ b/batch/sql/fix-n-cancelled-creating-jobs.sql
@@ -201,15 +201,15 @@ select t.user,
        0,
        t.n
 FROM (
-  select batches.user, jobs.inst_coll, count(*) as n
-  from jobs
-  left join batches_cancelled
-  on jobs.batch_id = batches_cancelled.id
-  left join batches
-  on jobs.batch_id = batches.id
-  where jobs.state = 'Creating'
-    and (NOT jobs.always_run AND (jobs.cancelled OR (batches_cancelled.id is not null)))
-  group by batches.user, jobs.inst_coll
+  SELECT batches.user, jobs.inst_coll, count(*) AS n
+  FROM jobs
+  LEFT JOIN batches_cancelled
+  ON jobs.batch_id = batches_cancelled.id
+  LEFT JOIN batches
+  ON jobs.batch_id = batches.id
+  WHERE jobs.state = 'Creating'
+    AND (NOT jobs.always_run AND (jobs.cancelled OR (batches_cancelled.id IS NOT NULL)))
+  GROUP BY batches.user, jobs.inst_coll
   LOCK IN SHARE MODE
 ) as t
 ON DUPLICATE KEY UPDATE n_cancelled_creating_jobs = t.n;
--- a/batch/sql/fix-n-cancelled-creating-jobs.sql
+++ b/batch/sql/fix-n-cancelled-creating-jobs.sql
@@ -189,7 +189,7 @@ INSERT INTO user_inst_coll_resources(
   `n_cancelled_running_jobs`,
   `n_cancelled_creating_jobs`
 )
-select t.user,
+SELECT t.user,
        t.inst_coll,
        0,
        0,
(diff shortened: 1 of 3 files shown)


jigold:
Can you also fix `check_incremental` in batch/batch/driver/main.py to actually throw an exception if there's a problem?

batch/sql/fix-n-cancelled-creating-jobs.sql
```diff
@@ -0,0 +1,217 @@
...
+       t.n
+FROM (
+  select batches.user, jobs.inst_coll, count(*) as n
+  from jobs
+  left join batches_cancelled
+  on jobs.batch_id = batches_cancelled.id
+  left join batches
+  on jobs.batch_id = batches.id
+  where jobs.state = 'Creating'
+    and (NOT jobs.always_run AND (jobs.cancelled OR (batches_cancelled.id is not null)))
```
jigold, line 211: caps lock please

danking:
Yep! https://github.com/hail-is/hail/pull/11279

batch/sql/fix-n-cancelled-creating-jobs.sql
```diff
@@ -0,0 +1,217 @@
...
+        n_creating_cancellable_jobs = n_creating_cancellable_jobs - 1;
+    END IF;
+
+    # state = 'Creating' jobs cannot be cancelled at the job level
+    IF NOT OLD.always_run AND cur_batch_cancelled THEN
+      # cancelled
+      INSERT INTO user_inst_coll_resources (user, inst_coll, token, n_cancelled_creating_jobs)
+      VALUES (cur_user, OLD.inst_coll, rand_token, -1)
+      ON DUPLICATE KEY UPDATE
+        n_cancelled_creating_jobs = n_cancelled_creating_jobs - 1;
```
danking, line 86: this is the important line.

batch/sql/fix-n-cancelled-creating-jobs.sql
jigold, line 211: I think this is right.

jigold (changes requested):

batch/sql/fix-n-cancelled-creating-jobs.sql
```diff
@@ -201,15 +201,15 @@ select t.user,
...
-  select batches.user, jobs.inst_coll, count(*) as n
-  from jobs
-  left join batches_cancelled
-  on jobs.batch_id = batches_cancelled.id
-  left join batches
-  on jobs.batch_id = batches.id
-  where jobs.state = 'Creating'
-    and (NOT jobs.always_run AND (jobs.cancelled OR (batches_cancelled.id is not null)))
-  group by batches.user, jobs.inst_coll
+  SELECT batches.user, jobs.inst_coll, count(*) AS n
```
jigold, line 204: So this only updates the row with token = 0. What happens to the other rows? I think we need to set all other rows to 0.

jigold (changes requested):

batch/sql/fix-n-cancelled-creating-jobs.sql
danking: I do, the pervious statement in this transaction is an UPDATE that sets them all to 0.

batch/sql/fix-n-cancelled-creating-jobs.sql
```diff
@@ -0,0 +1,217 @@
...
+  `n_ready_jobs`,
+  `n_running_jobs`,
+  `n_creating_jobs`,
+  `ready_cores_mcpu`,
+  `running_cores_mcpu`,
+  `n_cancelled_ready_jobs`,
+  `n_cancelled_running_jobs`,
+  `n_cancelled_creating_jobs`
+)
+select t.user,
```
jigold, line 192: One more

jigold (changes requested):

jigold:
I have to go, but what happens when you update the rows to 0. Does the trigger run? I presume the n_cancelled_creating_jobs stay at 0 because now the additions / subtractions are in sync?

danking:
@jigold there are no user_inst_coll_resources table triggers in the database. Are you referring to something other than a db trigger?

jigold (approved):

jigold:
I was getting confused where the trigger was located.

jigold:
I put the WIP back on -- it's too late in the day to react if there's an issue and the database is in a bad state.

PR #11279, opened 2022-01-27
[batch] raise error in check incremental

danking:
Without this change, no error is raised and we never record
that there was a check incremental failure.


jigold (approved):

jigold:
I put the WIP tag on

jigold (approved):

PR #11281, opened 2022-01-27
[batch] add sql_query_count and sql_query_latency metrics

vrautela:
- Edit LoggingTimer class to use prometheus client to time execution of SQL queries


batch/batch/driver/instance.py
```diff
@@ -159,7 +159,7 @@ async def activate(self, ip_address, timestamp):
         assert self._state == 'pending'
 
         rv = await check_call_procedure(
-            self.db, 'CALL activate_instance(%s, %s, %s);', (self.name, ip_address, timestamp)
+            self.db, 'CALL activate_instance(%s, %s, %s);', (self.name, ip_address, timestamp), 'activate'
```
daniel-goldstein, line 162: Let's make these names match the SQL procedure that we're CALLing here (`activate_instance`) and below with deactivate as well.

batch/batch/front_end/front_end.py
```diff
@@ -1361,22 +1364,22 @@ async def _get_job(app, batch_id, job_id):
...
+    LEFT JOIN instances
+      ON attempts.instance_name = instances.name
+    LEFT JOIN aggregated_job_resources
+      ON jobs.batch_id = aggregated_job_resources.batch_id AND
+         jobs.job_id = aggregated_job_resources.job_id
+    LEFT JOIN resources
+      ON aggregated_job_resources.resource = resources.resource
+    WHERE jobs.batch_id = %s AND NOT deleted AND jobs.job_id = %s
+    GROUP BY jobs.batch_id, jobs.job_id;
+    ''',
```
daniel-goldstein, line 1382: Looks like this got indented for some reason. Can you undo this?

gear/gear/database.py
```diff
@@ -199,26 +200,26 @@ async def just_execute(self, sql, args=None):
...
-    async def execute_and_fetchall(self, sql, args=None, timer_description=None):
+    async def execute_and_fetchall(self, sql, args=None, query_name=None):
         assert self.conn
         async with self.conn.cursor() as cursor:
-            if timer_description is None:
+            if query_name is None:
                 await cursor.execute(sql, args)
             else:
-                async with LoggingTimer(f'{timer_description}: execute_and_fetchall: execute', threshold_ms=20):
+                async with LoggingTimer(query_name, threshold_ms=20):
```
daniel-goldstein, line 219: We can remove `threshold_ms` since we will always time.

gear/gear/metrics.py
```diff
@@ -1,10 +1,13 @@
 from aiohttp import web
 import prometheus_client as pc  # type: ignore
 from prometheus_async.aio import time as prom_async_time  # type: ignore
+from hailtop.utils import time_msecs
```
daniel-goldstein, line 4: this is unused

gear/gear/metrics.py
```diff
@@ -1,10 +1,13 @@
...
 
 REQUEST_TIME = pc.Summary('http_request_latency_seconds', 'Endpoint latency in seconds', ['endpoint', 'verb'])
 REQUEST_COUNT = pc.Counter('http_request_count', 'Number of HTTP requests', ['endpoint', 'verb', 'status'])
 CONCURRENT_REQUESTS = pc.Gauge('http_concurrent_requests', 'Number of in progress HTTP requests', ['endpoint', 'verb'])
+SQL_QUERY_COUNT = pc.Counter('sql_query_count', 'Number of SQL Queries', ['query_name'])
+SQL_QUERY_LATENCY = pc.Summary('sql_query_latency', 'SQL Query latency in seconds', ['query_name'])
```
daniel-goldstein, line 10: Let's make the naming consistent with the above metrics, so change `sql_query_latency` to `sql_query_latency_seconds`.

gear/gear/metrics.py
```diff
@@ -25,3 +28,38 @@ async def monitor_endpoints_middleware(request, handler):
         raise e
     finally:
         CONCURRENT_REQUESTS.labels(endpoint=endpoint, verb=verb).dec()
+
+
+class LoggingTimer:
```
daniel-goldstein, line 33: `LoggingTimer` is a bit of a misnomer now, perhaps `PrometheusSQLTimer`?

gear/gear/metrics.py
```diff
@@ -25,3 +28,38 @@ async def monitor_endpoints_middleware(request, handler):
...
+    def __init__(self, query_name, threshold_ms=None):
+        self.query_name = query_name
+        self.threshold_ms = threshold_ms
+        self.timing = {}
+        self.start_time = None
```
daniel-goldstein, line 38: All of this time/threshold stuff is dead code now.

gear/gear/metrics.py
```diff
@@ -25,3 +28,38 @@ async def monitor_endpoints_middleware(request, handler):
...
+        self.sql_query_latency_manager = None
+
+    def step(self, name):
+        return LoggingTimerStep(self, name)
```
daniel-goldstein, line 42: Given our refactor, this method is probably unused. Can you make sure that it is and then delete it? Along with `LoggingTimerStep`

gear/gear/metrics.py
```diff
@@ -25,3 +28,38 @@ async def monitor_endpoints_middleware(request, handler):
...
+
+    async def __aenter__(self):
+        SQL_QUERY_COUNT.labels(query_name=self.query_name).inc()
```
daniel-goldstein, line 45: This is fine for now, but let's follow up later with another PR that gets us closer to matching the information that we have for HTTP endpoints. Specifically, I would want to examine the `exc` in `__aexit__` and add a sort of `status` label to these metrics. It would be nice to know, for example, if a query is slow only when it is failing or encountering deadlocks, but is otherwise fast.

daniel-goldstein (changes requested):
I think it would be good to follow up a little later and get finer-grain information, but this is a great step forward.

gear/gear/metrics.py
vrautela: I'm realizing now that the LoggingTimer is used in other places that don't have anything to do with monitoring SQL queries. I think we might have to design this a little differently to take care of the different use cases

batch/batch/cloud/gcp/worker/disk.py
```diff
@@ -1,8 +1,9 @@
 import logging
 from typing import Dict, Optional
 
-from hailtop.utils import check_shell_output, LoggingTimer, retry_all_errors_n_times
+from hailtop.utils import check_shell_output, retry_all_errors_n_times
 from hailtop.aiocloud import aiogoogle
+from hailtop.utils.utils import LoggingTimer
```
daniel-goldstein, line 6: This import can be undone. To change this file back to what it currently is in `main`, do `git checkout main batch/**/disk.py`

batch/batch/driver/instance.py
```diff
@@ -159,7 +159,7 @@ async def activate(self, ip_address, timestamp):
         assert self._state == 'pending'
 
         rv = await check_call_procedure(
-            self.db, 'CALL activate_instance(%s, %s, %s);', (self.name, ip_address, timestamp)
+            self.db, 'CALL activate_instance(%s, %s, %s);', (self.name, ip_address, timestamp), 'activate_instance'
```
daniel-goldstein, line 162: I don't believe this works with this function signature. The `check_call_procedure` also seems a bit random where it is currently defined. Let's turn that function into a method on the `Database` class, make that method signature work with the query timing, and delete the file it's currently defined in.

hail/python/hailtop/utils/__init__.py
```diff
@@ -4,7 +4,7 @@
     bounded_gather, grouped, sync_sleep_and_backoff, sleep_and_backoff, is_transient_error,
     request_retry_transient_errors, request_raise_transient_errors,
     collect_agen, retry_all_errors, retry_transient_errors,
-    retry_long_running, run_if_changed, run_if_changed_idempotent, LoggingTimer,
```
daniel-goldstein, line 7: This file should also be reverted to its state in `main`.

batch/batch/front_end/front_end.py
```diff
@@ -1364,16 +1367,16 @@ async def _get_job(app, batch_id, job_id):
 SELECT jobs.*, user, billing_project, ip_address, format_version, COALESCE(SUM(`usage` * rate), 0) AS cost
 FROM jobs
 INNER JOIN batches
-  ON jobs.batch_id = batches.id
+    ON jobs.batch_id = batches.id
```
daniel-goldstein, line 1370: still some formatting left to fix.

batch/batch/front_end/front_end.py
```diff
@@ -12,6 +12,7 @@
 import signal
 from aiohttp import web
 import aiohttp_session
+from hailtop.utils.utils import LoggingTimer
```
daniel-goldstein, line 15: This import should be reverted to what it was previously.

daniel-goldstein (dismissed):

daniel-goldstein (approved):

PR #11282, opened 2022-01-27
Update location and schema for gnomAD mitochondria datasets

nawatts:
I noticed that the gnomAD mitochondria datasets were pointing at the `gs://gnomad-public-requester-pays` bucket. The Google Cloud Public Datasets version should be up to date now.

Also updated the documented schema for chrM sites. Some information about those changes is available in the gnomAD change log: https://gnomad.broadinstitute.org/news/2021-08-rename-filter-in-mitochondria-dataset-and-minor-format-changes/.

And finally, since these were the only two datasets that reference `gnomad-public-requester-pays`, removed `gnomad-public-requester-pays` from the list of annotation DB buckets used by `hailctl dataproc`.


nawatts:
@gtiao @klaricch fyi

danking (approved):
Awesome! Thank you.

I verified the `describe` outputs are the same as listed in this PR. Everything looks good to me

PR #11290, opened 2022-01-28
[ci] Make WIP a PR that requires action

daniel-goldstein:
This adds all WIP-labelled PRs to the Awaiting Action table on the CI's /me page. CI has automated testing in-so-far as we create a test repository on GitHub with some toy code in it and see that CI can deploy that code into Kubernetes. This stuff not so much, but I can for example put a CI in my namespace and make a PR to its test repository, add a wip tag, and see the table change at `internal.hail.is/dgoldste/ci/me`

### DIFF
--- a/ci/ci/ci.py
+++ b/ci/ci/ci.py
@@ -27,7 +27,7 @@ from typing import Dict, Any, Optional, List
 from web_common import setup_aiohttp_jinja2, setup_common_static_routes, render_template, set_message
 
 from .environment import STORAGE_URI
-from .github import Repo, FQBranch, WatchedBranch, UnwatchedBranch, MergeFailureBatch, PR, select_random_teammate
+from .github import Repo, FQBranch, WatchedBranch, UnwatchedBranch, MergeFailureBatch, PR, select_random_teammate, WIP
 from .constants import AUTHORIZED_USERS, TEAMS
 
 with open(os.environ.get('HAIL_CI_OAUTH_TOKEN', 'oauth-token/oauth-token'), 'r') as f:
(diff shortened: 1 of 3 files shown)


ci/ci/ci.py
```diff
@@ -253,7 +253,8 @@ def pr_requires_action(gh_username, pr_config):
     build_state = pr_config['build_state']
     review_state = pr_config['review_state']
     return (
-        is_pr_author(gh_username, pr_config) and (build_state == 'failure' or review_state == 'changes_requested')
+        is_pr_author(gh_username, pr_config)
+        and (build_state == 'failure' or review_state == 'changes_requested' or WIP in pr_config['labels'])
```
vrautela, line 257: What is the type of pr_config? I think it might be good to have types in the function signature

daniel-goldstein: I added a good deal of types, with a small bit of refactoring to help with the typing. The major addition here is the use of `TypedDict`, which allows us to enforce a fixed schema on a python dictionary. The conveniences and serialization of a dictionary with the type information we want from a class.

ci/ci/ci.py
```diff
@@ -237,52 +262,51 @@ async def get_job(request, userdata):
     return await render_template('ci', request, userdata, 'job.html', page_context)
 
 
-def filter_wbs(wbs, pred):
+def filter_wbs(wbs, pred: Callable[[PRConfig], bool]):
```
vrautela, line 265: Should `wbs` have the type `List[WatchedBranchConfig]`?

vrautela (dismissed):
The changes make sense to me. I noticed a couple things that I think should be fixed, but it looks to good besides that

ci/ci/ci.py
```diff
@@ -68,7 +83,17 @@ async def pr_config(app, pr: PR) -> Dict[str, Any]:
...
+    sha: Optional[str]
+    deploy_batch_id: Optional[int]
+    deploy_state: Optional[str]
+    repo: str
+    prs: List[PRConfig]
+
+
+async def watched_branch_config(app, wb: WatchedBranch, index: int) -> WatchedBranchConfig:
     if wb.prs:
         pr_configs = [await pr_config(app, pr) for pr in wb.prs.values()]
```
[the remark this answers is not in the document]
daniel-goldstein: So notice that this element types in this list comprehension are `PRConfig`s. We're iterating over the values of `WatchedBranch.prs` which in the definition of `WatchedBranch` is an `Optional[Dict[str, PR]]`. The names aren't great here, but there's a difference between `WatchedBranch` and WatchedBranchConfig`, where the former is more the internal model, and the latter is the public data  that we're sending to the front-end (in this case the html template).

PR #11293, opened 2022-01-28
[hailctl] Move aiohttp imports in deploy_config out of top level

daniel-goldstein:
Python CLI tools like `hailctl` suffer from slow startup times, which infuriate me. This is in part because the first thing that happens is python has to recursively load all imported packages, since imports are traditionally done at the top-level. Very conveniently, setting the `PYTHONPROFILEIMPORTTIME` environment variable will cause python to emit a profile to stderr, which you can visualize with tools like [tuna](https://github.com/nschloe/tuna). So running

```
PYTHONPROFILEIMPORTTIME=1 hailctl dev config show 2> profile.log
tuna profile.log
```

gave me this

Screen Shot 2022-01-28 at 2 58 28 PM

showing that importing `aiohttp` is responsible for half the time it takes me to run `hailctl dev config show`, which is literally just printing a local file!! There's no reason this shouldn't be instantaneous, but reducing it to ~300ms, which this change did, is fine enough for me for now. Generally people don't care about import time because most applications are long-lived and what does a few seconds matter, so `pylint` by default wants us to put imports at the top level. I would say this is a valid exception.

### DIFF
--- a/hail/python/hailtop/config/deploy_config.py
+++ b/hail/python/hailtop/config/deploy_config.py
@@ -1,10 +1,8 @@
 from typing import List, Tuple, Dict
-import aiohttp
 import random
 import os
 import json
 import logging
-from aiohttp import web
 from ..utils import retry_transient_errors, first_extant_file
 from ..tls import internal_client_ssl_context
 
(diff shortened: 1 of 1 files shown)


vrautela (approved):
Why were these imports moved into the functions?

daniel-goldstein:
When python loads a file it "runs" everything at the top level. That means running statements like `import aiohttp`, `MY_CONSTANT = 5`, and `def foo: …`. So importing files can define functions, so they can be used later, but those functions are not run (how could they be?). By putting the imports inside the functions themselves, we defer the import from when the deploy_config is imported to when the functions that depend on `aiohttp` are actually used. If they are never used, `aiohttp` will never be imported! This is why it made `hailctl dev config list` much faster, since it doesn't need to use those functions.

PR #11294, opened 2022-01-28
[deploy] Publish the website we are building to hail_common on deploys. 

johnc1231:
I changed the `deploy` step to now take the built docs and publish them to a folder in hail-common. In the future, `make_pip_versioned_docs` will be deleted, so that we stop rebuilding an old version of the docs on every commit. We will simply serve the most recent version of the docs from here.


johnc1231:
I decided to try to do this in two passes since making changes to deploy logic always finnicky on its own. I think this does the right thing though. 

Does build.yaml support a way to say "depend on this step x if we are doing x at all"? Redeploying the website will have to happen after the `deploy` step runs in the future and publishes the latest version of the docs to hail-common.

hail/scripts/deploy.sh
```diff
@@ -8,15 +8,16 @@ retry() {
         (sleep 5 && "$@");
```
danking, line 8: update hail/Makefile as well with the extra argument. That's used in an emergency situation when we need to hand deploy.

danking (changes requested):
I think the only way to switch on deploy vs not deploy is to have a step which checks if this is a deploy or not (you can use `{% if global.scope == "deploy" %}...{% endif %}`). That step should depend on make_docs. If it *is* a deploy, make the `www.tar.gz`  from make_docs your output. If it is not a deploy, then `gsutil cp ...` the docs from the most recent tagged version and make *those* your output instead.

hail/scripts/deploy.sh
johnc1231: Done!

hail/Makefile
```diff
@@ -292,7 +292,7 @@ test-dataproc-38: install-hailctl
 # set WHEEL_FOR_AZURE to the path to that wheel file
 .PHONY: deploy
 deploy: test-dataproc $(WHEEL)
-	bash scripts/deploy.sh $(HAIL_PIP_VERSION) $(HAIL_VERSION) $(REVISION) $(DEPLOY_REMOTE) $(WHEEL) $(GITHUB_OAUTH_HEADER_FILE) $(HAIL_GENETICS_HAIL_IMAGE) $(WHEEL_FOR_AZURE)
+	bash scripts/deploy.sh $(HAIL_PIP_VERSION) $(HAIL_VERSION) $(REVISION) $(DEPLOY_REMOTE) $(WHEEL) $(GITHUB_OAUTH_HEADER_FILE) $(HAIL_GENETICS_HAIL_IMAGE) $(WHEEL_FOR_AZURE) $(WEBSITE_TAR)
```
danking, line 295: add a comment about how to generate WEBSITE_TAR

danking (changes requested):

danking (approved):

PR #11296, opened 2022-01-31
[query] Shuffler Outputs Approximately Evenly Sized Partitions

johnc1231:

1. For some reason, I forgot `CompileAndEvaluate` exists, and was separately calling `Compile`, applying it, then doing `SafeRow.read` in two places. Ripped those out in favor of `CompileAndEvaluate`.
2. When all segments are either already sorted or sufficiently small, I now sort and write out all the sufficiently small ones. Now I have a set of files that taken together fully represent the table, and I create a new `TableStage` by assembling those chunks into approximately evenly sized partitions.


patrick-schultz (approved):
looks good!

PR #11300, opened 2022-01-31
[query][ggplot] geom_histogram infers min and max values automatically

johnc1231:
This PR:

1. Adds a notion of `precomputes`, things that need to be computed in advance before we can do the main aggregation to render the plot. Right now, only `geom_histogram` has these, and they're only present if the user does not specify a minimum or maximum value. 

2. Adjusts the tooltip for histograms to show the range of values that a bar represents, instead of just the x coordinate of the center of the bar. See image for example:
Screen Shot 2022-01-31 at 4 01 54 PM


hail/python/test/hail/ggplot/test_ggplot.py
```diff
@@ -24,3 +24,11 @@ def test_manhattan_plot():
...
     fig = ggplot(ht, aes(x=ht.locus, y=-hl.log10(ht.pval))) + geom_point() + geom_hline(yintercept=-math.log10(5e-8))
     fig.to_plotly()
+
+
+def test_histogram():
+    ht = hl.utils.range_table(10)
+    fig = (ggplot(ht, aes(x=ht.idx)) +
+           geom_histogram()
+           )
+    fig.to_plotly()
```
ammekk, line 34: needs newline character at end of file

ammekk (approved):

PR #11302, opened 2022-01-31
[query] ssa style IR in logs

patrick-schultz:
This adds a new IR pretty-printer that uses an mlir inspired printed form. It is currently not rigorously tested, so using the new format in logs is gated behind the flag `HAIL_USE_SSA_LOGS`.

### DIFF
--- a/hail/src/main/scala/is/hail/HailContext.scala
+++ b/hail/src/main/scala/is/hail/HailContext.scala
@@ -483,7 +483,8 @@ object HailFeatureFlags {
     ("method_split_ir_limit", ("HAIL_DEV_METHOD_SPLIT_LIMIT" -> "16")),
     ("use_new_shuffle", ("HAIL_USE_NEW_SHUFFLE" -> null)),
     ("shuffle_cutoff_to_local_sort", ("HAIL_SHUFFLE_CUTOFF" -> null)),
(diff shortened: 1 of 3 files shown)


chrisvittal:
Do our tests pass with the flag on?

hail/src/main/scala/is/hail/HailContext.scala
```diff
@@ -483,7 +483,8 @@ object HailFeatureFlags {
     ("method_split_ir_limit", ("HAIL_DEV_METHOD_SPLIT_LIMIT" -> "16")),
     ("use_new_shuffle", ("HAIL_USE_NEW_SHUFFLE" -> null)),
     ("shuffle_cutoff_to_local_sort", ("HAIL_SHUFFLE_CUTOFF" -> null)),
-    ("grouped_aggregate_buffer_size", ("HAIL_GROUPED_AGGREGATE_BUFFER_SIZE" -> "50"))
+    ("grouped_aggregate_buffer_size", ("HAIL_GROUPED_AGGREGATE_BUFFER_SIZE" -> "50")),
+    ("use_ssa_logs", "HAIL_USE_SSA_LOGS" -> null),
```
chrisvittal, line 487: trailing comma, won't build under previous Scala version.

chrisvittal (changes requested):
Looks good to me. Just need to make sure that it works across the whole test suite when turned on. Manual is fine here.

chrisvittal (approved):

PR #11307, opened 2022-02-01
Add JSONEncoder that supports Hail types

nawatts:
Motivated by https://hail.zulipchat.com/#narrow/stream/123000-general/topic/exporting.20histogram.20aggregation.

This adds a [JSONEncoder](https://docs.python.org/3/library/json.html#json.JSONEncoder) class that supports Hail types such as Struct and frozendict.

```python
json.dumps(hl.Struct(foo="bar'))
# TypeError: Object of type Struct is not JSON serializable

json.dumps(hl.Struct(foo="bar"), cls=hl.utils.JSONEncoder)
# '{"foo": "bar"}'
```

This doesn't fully solve the issue in the Zulip thread since the encoding is not reversible. But it can still be useful for serializing Hail objects.


danking:
Whoops, sorry this got dropped! I'll review now.

hail/python/hail/utils/json.py
```diff
@@ -0,0 +1,31 @@
+import json
```
danking, line 1: Hrm. This might be superstition, but I feel like naming files in our project with the same name as standard Python packages has caused issues in the past for us. Can we name this `jsonx.py`? We do a similar thing with `httpx.py`.

danking (changes requested):

danking:
@johnc1231 , I'm roughly OK with this proposal, but I feel like you always have good thoughts about how changes like this could unintentionally affect things. Any apprehensions?

hail/python/hail/utils/json.py
nawatts: Renamed in [e76ca57](https://github.com/hail-is/hail/pull/11307/commits/e76ca57a3bdfa444019168d79be2cbc2323c0590).

nawatts:
> Whoops, sorry this got dropped!

No problem. This wasn't important, just a potential "nice to have".

> how changes like this could unintentionally affect things

This shouldn't change anything unless a user intentionally opts into using it to encode JSON. It was intended as a convenience: it's easy enough for a user to implement this themselves, but it might take a few rounds of trial and error to catch all the classes that need to be handled.

johnc1231 (approved):

PR #11313, opened 2022-02-02
[infra] always use valid storage account names

danking:
Resource groups are permitted to use dashes, underscores, uppercase letters,
and probably other characters not permitted in storage account names. This
PR cahanges `bootstrap.sh` to:

1. Ignore invalid characters in the resouce group.

2. Ensure (via randomness) that the generated name is unique.

3. Do not try to create a new storage account if `backend-config.tfvars` exists.

I lightly tested this. Here is an example of how it sanitizes a resource group name:

```
RESOURCE_GROUP=bu__ild-batch-worker-i32mage
possibly_invalid_storage_account_name="$(cat /dev/urandom | LC_ALL=C tr -dc 0-9 | head -c 4)${RESOURCE_GROUP}"
STORAGE_ACCOUNT_NAME=$(LC_ALL=C tr -dc a-z0-9 <<< "${possibly_invalid_storage_account_name}" | head -c 24)
echo $STORAGE_ACCOUNT_NAME
7241buildbatchworkeri32m
```

### DIFF
--- a/infra/azure/bootstrap.sh
+++ b/infra/azure/bootstrap.sh
@@ -16,50 +16,53 @@ setup_az() {
     az login --identity
 }
 
-setup_terraform_backend() {
-    RESOURCE_GROUP=$1
-    STORAGE_CONTAINER_NAME="tfstate"
+create_terraform_remote_storage() {
+    RESOURCE_GROUP=${RESOURCE_GROUP:-$1}
+    STORAGE_CONTAINER_NAME=${STORAGE_CONTAINER_NAME:-"tfstate"}
 
-    if [ -e backend-config.tfvars ]
-    then
-        echo "Found backend-config.tfvars. I assume storage account and container for terraform are already created."
-    else
-        echo "No backend-config.tfvars found. I will create storage account and container for terraform."
-        # NOTE: This name is assumed to be globally unique
-        # NOTE: must be 3-24 lowercase letters and numbers only
-        possibly_invalid_storage_account_name="$(cat /dev/urandom | LC_ALL=C tr -dc '0-9' | head -c 4)${RESOURCE_GROUP}"
-        STORAGE_ACCOUNT_NAME=$(LC_ALL=C tr -dc 'a-z0-9' <<< "${possibly_invalid_storage_account_name}" | head -c 24)
+    # NOTE: This name is assumed to be globally unique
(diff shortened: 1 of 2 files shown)


infra/azure/bootstrap.sh
```diff
@@ -20,28 +20,35 @@ setup_terraform_backend() {
...
+        possibly_invalid_storage_account_name="$(cat /dev/urandom | LC_ALL=C tr -dc '0-9' | head -c 4)${RESOURCE_GROUP}"
+        STORAGE_ACCOUNT_NAME=$(LC_ALL=C tr -dc 'a-z0-9' <<< "${possibly_invalid_storage_account_name}" | head -c 24)
+
+        # This storage account/container will be created if it does not already exist
+        az storage account create -n $STORAGE_ACCOUNT_NAME -g $RESOURCE_GROUP
+        STORAGE_ACCOUNT_KEY=$(az storage account keys list \
+            -g $RESOURCE_GROUP \
+            --account-name $STORAGE_ACCOUNT_NAME \
+            | jq -rj '.[0].value'
+        )
```
daniel-goldstein, line 39: I agree the old naming scheme was brittle, but one of the things I liked about it was that we didn't have to pass a `backend-config.tfvars` file around. I guess since we must do that for `global.tfvars`, this isn't much different, but I feel like it would be nice to keep secret keys out of these files. For example, for `global.tfvars` we could move any secrets (like the github oauth token) into vault and the input variable would just reference it.

I mostly agree with this approach, but perhaps we could have an env variable for the storage account / container name and trust `az` to grab the key? One less secret that we're passing around?

daniel-goldstein (dismissed):
I think my comment is one step further than this PR, but relevant nonetheless.

infra/azure/bootstrap.sh
daniel-goldstein: What concerns me most about this PR is someone accidentally running terraform without having the `backend-config.tfvars` in place and generating a new state bucket when they really should be using an existing one. It would probably break part of the way through a terraform apply because it tries to create an AKS cluster called "vdc" when one already exists.

daniel-goldstein: It might be nice to instead have a separate step to generate the bucket and make it obvious that it should only be run the very first time.

daniel-goldstein (approved):

PR #11314, opened 2022-02-02
[website] Restore the broken tutorials so they actually run

johnc1231:
This PR:

1. Fixes a mistake in deploy.sh so that files get written to predictable paths. I manually put a file in the hail-common bucket for the purposes of testing this PR. 
2. Replaces `make_pip_versioned_docs` with `get_pip_versioned_docs`, a step that will usually just pull prerendered docs from `hail-common`, but if it's a deploy, it will pass along the docs that it gets from its input, `make-docs`. 

The effect should be that as soon as this merges, tutorials will correctly render again.


build.yaml
```diff
@@ -2082,37 +2082,40 @@ steps:
...
-      # 0.2.63 was the last old-style docs
-      if [[ "$(cat hail/python/hail/hail_pip_version)" != "0.2.63" ]]
+      if git ls-remote --exit-code --tags origin $(cat /io/hail_pip_version)
       then
-          # downloading files from GCS changes permissions, so we reset all of that first
-          git reset --hard HEAD
-          # if the tag doesn't exist, this commit is the release
-          git checkout $(cat hail/python/hail/hail_pip_version) || true
+        # In this case, we want to get the docs from Google Storage.
+        export PIP_VERSION=$(cat hail/python/hail/hail_pip_version)
```
danking, line 2101: call this HAIL_PIP_VERSION as we do elsewhere

build.yaml
```diff
@@ -2082,37 +2082,40 @@ steps:
       - hail_base_image
       - build_hail
   - kind: runImage
-    name: make_pip_versioned_docs
+    name: get_pip_versioned_docs
     image:
       valueFrom: hail_pip_installed_image.image
     script: |
-      set -ex
+      set -x
```
danking, line 2089: We can restore the `-e`, right? Failing on an error doesn't apply to the expression inside the `if` command.

danking (dismissed):

danking (approved):

PR #11316, opened 2022-02-04
[compiler] StreamJoinRightDistinct handles intervals on right

patrick-schultz:
This will be used to lower TableIntervalJoin.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/TypeCheck.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/TypeCheck.scala
@@ -341,9 +341,16 @@ object TypeCheck {
         assert(coerce[TStream](x.typ).elementType == join.typ)
         assert(lKey.forall(lEltTyp.hasField))
         assert(rKey.forall(rEltTyp.hasField))
-        assert((lKey, rKey).zipped.forall { case (lk, rk) =>
-          lEltTyp.fieldType(lk) == rEltTyp.fieldType(rk)
-        })
+        if (x.isIntervalJoin) {
+          val lKeyTyp = lEltTyp.fieldType(lKey(0))
+          val rKeyTyp = rEltTyp.fieldType(rKey(0)).asInstanceOf[TInterval]
+          assert(lKeyTyp == rKeyTyp.pointType)
+          assert((joinType == "left") || (joinType == "inner"))
+        } else {
+          assert((lKey, rKey).zipped.forall { case (lk, rk) =>
+            lEltTyp.fieldType(lk) == rEltTyp.fieldType(rk)
+          })
+        }
       case x@StreamFor(a, valueName, body) =>
         assert(a.typ.isInstanceOf[TStream])
         assert(body.typ == TVoid)
--- a/hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
@@ -1298,12 +1298,11 @@ object EmitStream {
               if (x.isIntervalJoin) {
                 val lhs = cb.memoize(lelt.toI(cb).flatMap(cb)(_.asBaseStruct.loadField(cb, lKey(0))))
                 val rhs = cb.memoize(relt.toI(cb).flatMap(cb)(_.asBaseStruct.loadField(cb, rKey(0))))
-                val pointOrdering = cb.emb.ecb.getOrderingFunction(lhs.st, rhs.st.asInstanceOf[SInterval].pointType, CodeOrdering.Compare(missingEqual=false))
                 val result = cb.newLocal[Int]("SJRD-interval-compare-result")
                 rhs.toI(cb).consume(cb, {
                   cb.assign(result, -1)
                 }, { case interval: SIntervalValue =>
-                  cb.assign(result, IntervalFunctions.pointIntervalCompare(cb, lelt, interval))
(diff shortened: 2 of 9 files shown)


johnc1231:
Looks like some kind of issue with PTypes on `CodeOrdering`

patrick-schultz:
Ah, just a typo

hail/src/main/scala/is/hail/expr/ir/TypeCheck.scala
```diff
@@ -341,9 +341,16 @@ object TypeCheck {
...
         assert(lKey.forall(lEltTyp.hasField))
         assert(rKey.forall(rEltTyp.hasField))
-        assert((lKey, rKey).zipped.forall { case (lk, rk) =>
-          lEltTyp.fieldType(lk) == rEltTyp.fieldType(rk)
-        })
+        if (x.isIntervalJoin) {
+          val lKeyTyp = lEltTyp.fieldType(lKey(0))
+          val rKeyTyp = rEltTyp.fieldType(rKey(0)).asInstanceOf[TInterval]
+          assert(lKeyTyp == rKeyTyp.pointType)
+          assert((joinType == "left") || (joinType == "inner"))
```
johnc1231, line 348: Since these asserts are here, shouldn't the `isIntervalJoin` function on `StreamJoinRightDistinct` have being `"left" or "inner" as conditions of it being true?

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -1547,7 +1555,7 @@ object EmitStream {
                 val pulledRight = mb.genFieldThisRef[Boolean]("left_join_right_distinct_pulledRight]")
 
                 val producer = new StreamProducer {
-                  override val length: Option[EmitCodeBuilder => Code[Int]] = leftProducer.length
+                  override val length: Option[EmitCodeBuilder => Code[Int]] = None
```
johnc1231, line 1558: Was this just a bug before?

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -1294,19 +1292,29 @@ object EmitStream {
...
-              val rhs = relt.map(cb)(_.asBaseStruct.subset(rKey: _*))
-              StructOrdering.make(lhs.st.asInstanceOf[SBaseStruct], rhs.st.asInstanceOf[SBaseStruct],
-                cb.emb.ecb, missingFieldsEqual = false)
-                .compare(cb, lhs, rhs, missingEqual = false)
+              if (x.isIntervalJoin) {
+                val lhs = cb.memoize(lelt.toI(cb).flatMap(cb)(_.asBaseStruct.loadField(cb, lKey(0))))
+                val rhs = cb.memoize(relt.toI(cb).flatMap(cb)(_.asBaseStruct.loadField(cb, rKey(0))))
+                val result = cb.newLocal[Int]("SJRD-interval-compare-result")
+                rhs.toI(cb).consume(cb, {
+                  cb.assign(result, -1)
```
johnc1231, line 1303: Should the interval comparison functions just support taking in missing values? That way the result of comparison against missing is consistent if used in other places in code? Not sure if it should always be handled the same way, this is just me asking.

johnc1231 (dismissed):
Looks great! My three comments are more like questions, so it may be that there's nothing to actually change depending on your answers.

hail/src/main/scala/is/hail/expr/ir/TypeCheck.scala
patrick-schultz: I agree it's a little weird. My thought process was that `isIntervalJoin` assumes a well-typed node. If the join type isn't "left" or "inner", but the right side is interval keyed while the left isn't, the right answer isn't "isIntervalJoin = false", it's just a malformed IR. I think that's the right general philosophy: outside of TypeCheck, assume things are well-typed, so we don't have to litter the same typechecking assertions everywhere.

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
patrick-schultz: Yeah, an inner join certainly doesn't need to have the same length as either input.

johnc1231: Agreed

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
patrick-schultz: I also find this unsatisfying. The problem is that the interval itself could be missing, and the endpoints could be missing, and it's not obvious to me that those layers of missingness will always be handled the same way. Right now, the interval comparison functions handle the inner missingness.

In fact, I'm now not confident that `missingEqual = false` is right for the inner comparisons. We use that in joins so that we don't join missing keys on the left with missing keys on the right. But should we consider a key `NA(Int32)` on the left as being in the interval `[5, NA]` on the right? I'll go look for what we do currently.

patrick-schultz:
Okay, I've rewritten all the interval functions so that a missing endpoint leads to a missing result, which I think is the only reasonable user facing semantics. Inside an interval join, if an interval has a missing endpoint, it is considered less than any point, so that the join just skips past it, and the point has a chance to match with later intervals.

johnc1231 (approved):

PR #11317, opened 2022-02-04
[query] Refactor ggplot to use pandas, support alpha on histograms and bar charts

johnc1231:
This PR:

1. Refactors `listify` to return pandas dataframes, and every geom to take in pandas dataframes. This significantly simplifies the code and should also speed things up.
2. Refactors the `apply_to_fig` method of most of the geoms to rely on a specification dict and then just loop over that. This simplifies adding a new argument / aesthethic. In the future, it may be best to just make all of the geoms take `**kwargs` so that arguments added to that dict will immediately be used for plotting, as right now I have to add something to `geom_bar`, `GeomBar.__init__`, and that dictionary for it to start showing up in plots. 
3. Adds `identity` as a bar position, which means to plot bars on top of each other. This is useful along with the `alpha` argument added to bars and histograms, which sets transparency of points.

### DIFF
--- a/hail/python/hail/ggplot/ggplot.py
+++ b/hail/python/hail/ggplot/ggplot.py
@@ -13,7 +13,7 @@ from .scale import Scale, ScaleContinuous, ScaleDiscrete, scale_x_continuous, sc
     scale_x_discrete, scale_y_discrete, scale_color_discrete, scale_color_continuous, scale_fill_discrete, \
     scale_fill_continuous
 from .aes import Aesthetic, aes
-from .utils import is_continuous_type, is_genomic_type, check_scale_continuity
+from .utils import is_continuous_type, is_genomic_type, check_scale_continuity, should_use_scale_for_grouping
 
 
 class GGPlot:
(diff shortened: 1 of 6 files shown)


johnc1231:
I'm still figuring out what the abstractions should be here. This refactoring has resulted in the `apply_to_figure` method of most geoms looking very similar, but still with little differences. Reluctant to commit to a very stringent class hierarchy yet and didn't want to further complicate this PR, but other than `hline` and `vline`, feels close to a world where each geom just overrides `plot_one_group` and the general looping over groups is handled by the parent class.

hail/python/hail/ggplot/stats.py
```diff
@@ -25,10 +27,20 @@ def make_agg(self, mapping, precomputed):
 
     def listify(self, agg_result):
         # Collect aggregator returns a list, nothing to do.
```
patrick-schultz, line 29: delete comment

hail/python/hail/ggplot/ggplot.py
```diff
@@ -172,30 +172,27 @@ def get_aggregation_result(selected, mapping_per_geom, precomputed):
...
                     group_counter += 1
                     return group_counter - 1
                 numberer = defaultdict(increment_and_return_old)
-                numbering = [numberer[data_tuple] for data_tuple in subsetted_to_discrete]
-
-                numbered_result = []
-                for s, i in zip(listified_agg_result, numbering):
-                    numbered_result.append(s.annotate(group=i))
-                listified_agg_result = numbered_result
+                df_agg_result["group"] = [numberer[tuple(x)] for _, x in subsetted_to_discrete.iterrows()]
```
patrick-schultz, line 193: You can leave this for future improvement, but I suspect you could do this more easily and faster with pandas `groupby`.

patrick-schultz (dismissed):
Looks good! Just an obsoleted comment.

hail/python/hail/ggplot/ggplot.py
johnc1231: I am going to do this in future improvement, since I'm currently in the midst of refactoring this a bit due to a bug report about running out of colors to plot with.

patrick-schultz (approved):

PR #11321, opened 2022-02-07
[query] Better error message on locus_windows

johnc1231:
Trying to get to the bottom of https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query.200.2E2.20support/topic/locus_windows.20Error/near/270984913


johnc1231:
I haven't tested this error message, as I'm not sure how to replicate the bug scenario, but I think it should work.

tpoterba (approved):

PR #11323, opened 2022-02-07
[query] Lower TextTableWriter

chrisvittal:
To do this add an intermidiate abstract PartitionWriter,
SimplePartitionWriter, and extend it with TextTablePartitionWriter.

SimplePartitionWriter manages the output stream for its subclasses. The
subclasses need to implement consumeElement, and optionally preConsume
and postConsume.

TextTablePartitionWriter handles the delimiter delimited writing per
element item and the line delimited writing per element of the stream.
It will also optionally write the header if ExportType.PARALLEL_HEADER_IN_SHARD
has been requested.

TextTableFinalizer is the 'MetadataWriter' (name change pending), for
TextTableWriter. That's where the header is written for
PARALLEL_SEPARATE_HEADER and CONCATENATED, and files are merged for
CONCATENATED.

There are currently a few idosyncracies that don't replicate bug-for-bug
compatibility with the non-lowered version.

    1. We avoid the use of fs.copyMerge, as such, we don't check for the
       existence of the _SUCCESS file (even though we create it).
    2. ExportType.PARALLEL_COMPOSABLE is unsupported.
    3. We rely on the temporary file cleaner to clean up the files
       created for CONCATENATED export, rather than deleting them
       explicitly.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/IR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/IR.scala
@@ -15,11 +15,15 @@ import is.hail.types.{RIterable, TypeWithRequiredness}
 import is.hail.types.encoded._
 import is.hail.types.physical._
 import is.hail.types.physical.stypes.{BooleanSingleCodeType, Float32SingleCodeType, Float64SingleCodeType, Int32SingleCodeType, Int64SingleCodeType, PTypeReferenceSingleCodeType, SType}
+import is.hail.types.physical.stypes.concrete.SJavaString
+import is.hail.types.physical.stypes.interfaces._
 import is.hail.types.virtual._
 import is.hail.utils.{FastIndexedSeq, _}
 import org.json4s.{DefaultFormats, Extraction, Formats, JValue, ShortTypeHints}
 
 import scala.language.existentials
+
+import java.io.OutputStream
 
 sealed trait IR extends BaseIR {
   private var _typ: Type = null
@@ -733,6 +737,7 @@ object PartitionWriter {
   implicit val formats: Formats = new DefaultFormats() {
     override val typeHints = ShortTypeHints(List(
       classOf[PartitionNativeWriter],
+      classOf[TableTextPartitionWriter],
       classOf[AbstractTypedCodecSpec],
       classOf[TypedCodecSpec]), typeHintFieldName = "name"
     ) + BufferSpec.shortTypeHints
@@ -750,6 +755,7 @@ object MetadataWriter {
       classOf[RVDSpecWriter],
       classOf[TableSpecWriter],
       classOf[RelationalWriter],
+      classOf[TableTextFinalizer],
       classOf[RVDSpecMaker],
       classOf[AbstractTypedCodecSpec],
       classOf[TypedCodecSpec]),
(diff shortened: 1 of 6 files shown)


tpoterba (dismissed):
To summarize our design discussion:

- we think that in the future, it might be cleaner for WriteMetadata to be functional (return the path of the file committed) and to rewrite the MatrixWriter lowering in terms of a single commit with side effects so it is clear that the various side-effecting operations happen atomically. 
- here, instead of relying on a heavily-IR lowering, we will dispatch more into codegen. All of the driver operations will happen in the codegen of a single parameterized MetadataWriter
- For the stream consumers, we will create a new abstract PartitionWriter that handles an output stream, and an abstract implementation that handles text output specifically which will be used by TableExport.

tpoterba:
The structure looks great, thanks for doing the redesign with me! I'm happy and will pass off to John for a full review.

hail/src/main/scala/is/hail/expr/ir/IR.scala
```diff
@@ -786,6 +792,37 @@ abstract class PartitionWriter {
   def toJValue: JValue = Extraction.decompose(this)(PartitionWriter.formats)
 }
 
+abstract class SimplePartitionWriter extends PartitionWriter {
+  def ctxType: Type = TString
+  def returnType: Type = TString
+  def unionTypeRequiredness(r: TypeWithRequiredness, ctxType: TypeWithRequiredness, streamType: RIterable): Unit = {
+    r.union(ctxType.required)
+    r.union(streamType.required)
```
johnc1231, line 800: I think you're supposed to `unionFrom` here? The `ctxType` itself could be a struct with fields that are potentially missing.

hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
```diff
@@ -433,7 +436,115 @@ case class TableTextWriter(
   exportType: String = ExportType.CONCATENATED,
   delimiter: String
 ) extends TableWriter {
+
   def apply(ctx: ExecuteContext, tv: TableValue): Unit = tv.export(ctx, path, typesFile, header, exportType, delimiter)
+
+  override def canLowerEfficiently: Boolean = true
+  override def lower(ctx: ExecuteContext, ts: TableStage, t: TableIR, r: RTable, relationalLetsAbove: Map[String, IR]): IR = {
+    if (exportType == ExportType.PARALLEL_COMPOSABLE)
+      throw new LowererUnsupportedOperation(s"cannot lower parallel_composable export")
```
johnc1231, line 445: Probably you want `canLowerEfficiently` to only be true if `exportType != ExportType.PARALLEL_COMPOSABLE`. Otherwise the thing that looks at pipelines and decides if it can lower them instead of doing old Spark execution model is going to decide that it can, only to error when it tries.

johnc1231 (changes requested):

hail/src/main/scala/is/hail/expr/ir/IR.scala
chrisvittal: I don't really understand this part of the code. I basically just copied it from the PartitionNativeWriter.

hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
chrisvittal: done

hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
```diff
@@ -433,7 +436,114 @@ case class TableTextWriter(
...
+          cb += (allFiles(0) = const(headerFilePath))
+          cb += Code.invokeStatic5[System, Any, Int, Any, Int, Int, Unit](
+            "arraycopy", jFiles /*src*/, 0 /*srcPos*/, allFiles /*dest*/, 1 /*destPos*/, jFiles.length /*len*/)
+          allFiles
+        } else {
+          files.asInstanceOf[SJavaArrayStringValue].array
+        }
+
+        cb += cb.emb.getFS.invoke[Array[String], String, Unit]("concatenateFiles", jFiles, const(outputPath))
+        // can I rely on context managers to clean up?
```
johnc1231, line 529: The only thing I'd say now is that we need to answer this question and delete this comment. I'm not sure what the answer is. What are you worried about cleaning up specifically?

johnc1231 (changes requested):

chrisvittal:
tests are green now

johnc1231 (approved):

PR #11325, opened 2022-02-07
SQL Query monitoring active_instances_only decorator

vrautela:
- Add PrometheusSQLTimer logging capabilities to select_and_fetchone and execute_and_fetchone methods in database.py
- Add parameter to use monitoring in active_instances_only decorator


daniel-goldstein:
The previous PR has merged. Could you remove the Stacked PR label and rebase / drop any redundant commits? I thought GitHub wouldn't show the extra diff at this point but *shrug*

PR #11333, opened 2022-02-08
[hailctl] support missing domain in deploy config

danking:


jigold:
@danking Which do you prefer? I created the same PR just now as well.

jigold (approved):

PR #11345, opened 2022-02-10
[vds] Fix and test `hl.vds.filter_intervals`

tpoterba:
Accept a table or a list.

### DIFF
--- a/hail/python/hail/vds/methods.py
+++ b/hail/python/hail/vds/methods.py
@@ -482,13 +482,22 @@ def filter_variants(vds: 'VariantDataset', variants_table: 'Table', *, keep: boo
 
 
 @typecheck(vds=VariantDataset,
-           intervals=expr_array(expr_interval(expr_any)),
+           intervals=oneof(Table, expr_array(expr_interval(expr_any))),
            keep=bool,
            mode=enumeration('variants_only', 'split_at_boundaries', 'unchecked_filter_both'))
 def _parameterized_filter_intervals(vds: 'VariantDataset',
-                                    intervals: 'ArrayExpression',
+                                    intervals,
                                     keep: bool,
                                     mode: str) -> 'VariantDataset':
+    intervals_table = None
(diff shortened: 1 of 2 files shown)


danking:
Old PR, closing. @chrisvittal can you take a look at the additional changes Tim made here? It looks like you only made one of them in https://github.com/hail-is/hail/pull/11400 Do we need the rest of this?

chrisvittal:
nah, this is a better PR. I missed it when looking for fixes from zulip.

chrisvittal:
I'll fix.

danking:
Think you @chrisvittal !

chrisvittal (dismissed):
I only fixed merge conflicts.

danking:
There's something fishy happening in the service backend. It keeps timing out the same three test jobs. We'll need to do a little grep+diff to figure which tests are getting started but not terminating. I suspect we'll need to either skip something or ensure that it errors early.

danking:
I'm also seeing this in my PR that removes most spark-only annotations (in favor of fails_service). I think we must have some logic that spins forever.

danking:
Dismissed the review so that CI would focus on other PRs, but feel free to re-enable. We can chat tomorrow about how to test the query service with a custom JAR

danking:
I'm on the train until about noon, but I can peek at this in the afternoon and let you know what I find.

danking:
Sorry, had some unexpected meetings yesterday. Digging in now.

danking (approved):

danking (changes requested):
(ci is broke, need to do this to unstick)

danking (approved):

danking:
@chrisvittal, these tests actually pass the local backend. I changed the tests to reflect this, but wanted to ping you to make sure that seems reasonable.

danking:
Actually they both passed on the service backend too.

danking:
Chris, could you do a release next week to get this fix out? Thank you much!

PR #11346, opened 2022-02-10
[batch] active_instances_only name-token cache

vrautela:
- Create a cache that stores an instance's token which can be looked up by the instance's name
- Use this cache in the active_instances_only decorator to avoid making DB request on every invocation
- Add monitoring of caches' hits, misses, evictions, and load latencies


batch/batch/driver/instance_collection/base.py
```diff
@@ -87,6 +96,15 @@ def get_instance(self, inst_name) -> Optional[Instance]:
             return inst_coll.name_instance.get(inst_name)
         return None
 
+    async def get_token_from_instance_name(self, name):
+        record: Dict[str, Any] = await self.db.select_and_fetchone(
+            'SELECT token FROM instances WHERE name = %s',
+            (name),
+        )
+        if not record or 'token' not in record:
```
daniel-goldstein, line 104: If the record exists is not None, then you should be guaranteed to have the token field in it.

batch/batch/driver/instance_collection/base.py
```diff
@@ -87,6 +96,15 @@ def get_instance(self, inst_name) -> Optional[Instance]:
...
+            return None
```
daniel-goldstein, line 105: Hm, there is a type error here that reveals a sticky question. The cache has a value type of `str`, which `None` is not. Should it remain `str` or be `Optional[str]`? Is `None` a value that we *want* in the cache? Would I be able to submit a bunch of garbage instance names and essentially purge the cache of all useful values?

I think we want to guarantee here that the `load` function we give to our cache will produce a valid result. i.e. we should be able to assert that the record we retrieve from the database is not None. Can we do that?

hail/python/hailtop/aiotools/time_limited_max_size_cache.py
```diff
@@ -2,20 +2,29 @@
 
 import time
 import asyncio
+import prometheus_client as pc  # type: ignore
+from prometheus_async.aio import time as prom_async_time  # type: ignore
```
daniel-goldstein, line 6: On further inspection, hailtop doesn't depend on prometheus (it's in `docker/requirements.txt` which is for services code). We might need to move this into gear, I've started a thread in zulip.

daniel-goldstein (changes requested):
Nice work! I've made a couple small comments and have one further request. Could you run a load test with and without this change and post a screenshot of how the number of invocations for the active instances query goes down? I would like to have proof for the records that this is reducing the number of queries we're performing. It doesn't need to be more than a handful of workers, so long as they're submitting many requests to the driver.

daniel-goldstein:
Now that we have the SQL query monitoring, I would love to also see just the simple comparison of the total number of queries we perform over a 1-minute period under high load. We have the tools now to see just what chunk of overall database communication we are cutting down on, which is an achievement in itself. Just need to run the test!

batch/batch/driver/instance_collection/base.py
vrautela: I think we would have to change the logic in the TimeLimitedMaxSizeCache.lookup function. Currently, it awaits the result of calling load, puts the result in the cache, and then returns the result. It looks like it doesn't handle the case where a user looks up an item that isn't present in the data source

daniel-goldstein: You're right about the behavior of `lookup`, but our *need* for this behavior isn't apparent to me, at least not yet. Based on our usage of this cache, it feels over-protective to think that `record` could be None. Since we keep old workers in the database, `record` should be not None so long as `name` is a real name of an instance that existed. If you look at `active_instances_only`, we throw unauthorized if a request submits an instance name that does not exist. By the time we get to using the cache, we've created an invariant that the input we've received is a good one, and we should take advantage of it! I think we can simply assert here that the record isn't None.

daniel-goldstein (changes requested):

daniel-goldstein (approved):
Nice work!

PR #11347, opened 2022-02-10
[hailtop.utils]new test added for grouped

CDiaz96:


hail/python/hailtop/utils/utils.py
```diff
@@ -112,6 +112,8 @@ def grouped(n, ls):
         group = ls[:n]
         ls = ls[n:]
         yield group
+        if n < 1:
+           raise ValueError ('invalid value for n')
```
danking, line 116: I like the spirit of this, but it won't work as written. Try copying this definition into `ipython` and trying it out. You'll find it doesn't error when you pass the argument `0` for the parameter `n`.

There's a really powerful technique for reasoning about programs that I call "stepping through the program". To "step" a program, you pretend to be the computer. Below, I give an example of stepping through the Python expression `grouped(0, [1,2,3,4,5,6])`.

Note that instead of raising an error when I call `grouped(0, [1,2,3,4,5,6])`, I just return from the function without any error!

---

### Step 1
#### Variables
```
```
(none)
#### current line of code
```
grouped(0, [1,2,3,4,5,6])
```

---

### Step 2
#### Variables
```
n = 0
ls = [1,2,3,4,5,6]
```
#### current line of code
```
while len(ls) != 0
```

---

### Step 3
#### Variables
```
n = 0
ls = [1,2,3,4,5,6]
```
#### current line of code
```
while 6 != 0
```

---

### Step 4
#### Variables
```
n = 0
ls = [1,2,3,4,5,6]
```
#### current line of code
We reached the end of the function, so there's not really a particular "line". We'll return from grouped without erroring!

hail/python/test/hailtop/utils/test_utils.py
```diff
@@ -126,3 +127,61 @@ def test_parse_docker_image_reference():
...
+
+
+def test_grouped_1():
+    try:
+        actual = list(grouped(0, [1,2,3,4,5,6,7,8,9]))
+        expected = [ 1,2,3,4,5,6,7,8,9]
+    except ValueError:
+        pass
+    else:
+        assert actual == expected
```
danking, line 141: In plain-English this test says:

- run `grouped(0, ...)`
- if there's a `ValueError` do nothing. (note!!! `pass` has *absolutely nothing to do with tests* it is just the Python expression that means "do nothing")
- if there's no error, check that actual equals expected.

But that's not what we want! *if there is no error*, then the test should fail! We *expect* `grouped(0, [1,2,3,4,5,6,7,8,9])` to fail. It doesn't make sense to make groups of size zero.

Here's an example of a test which verifies that dividing by zero is an error:
```
def test_dividing_by_zero_should_raise_an_error():
    try:
        123 / 0
    except ZeroDivisionError:
        pass  # recall: this just means "do nothing"
    else:
        assert False  # if we get here, then there was *not* an error, so we "assert False" which fails the test
```

I think an important thing to keep in mind is:
- a test fails if the function contains an `assert XXX` where `XXX` evaluates to false.
- otherwise, a test passes.

---

Can you change this test to fail when there is no error?

hail/python/test/hailtop/utils/test_utils.py
```diff
@@ -126,3 +127,61 @@ def test_parse_docker_image_reference():
...
+
+def test_grouped_2():
```
danking, line 143: Let's make the test function names more descriptive. For example, this test tests that grouping into groups of 1 works, so we might call it: `test_grouped_with_groups_of_one`

danking (changes requested):

hail/python/test/hailtop/utils/test_utils.py
CDiaz96: Done

hail/python/test/hailtop/utils/test_utils.py
CDiaz96: Changed the names

hail/python/hailtop/utils/utils.py
CDiaz96: Fixed the function

CDiaz96 (commented):
Addressed comments and made changes.

hail/python/test/hailtop/utils/test_utils.py
```diff
@@ -126,3 +127,59 @@ def test_parse_docker_image_reference():
     assert x.digest is None
     assert x.name() == 'us-docker.pkg.dev/my-project/my-repo/test-image'
     assert str(x) == 'us-docker.pkg.dev/my-project/my-repo/test-image'
+
+
+def test_grouped_size_0_groups_9_elements():
+    try:
+        actual = list(grouped(0, [1,2,3,4,5,6,7,8,9]))
```
jigold, line 134: This should throw an error. So there is no `actual` or `expected` values here. It should just be this:

```python3
try:
    list(grouped(0, [1,2,3,4,5,6,7,8,9]))
except ValueError:
    pass
else:
    assert False
```

hail/python/hailtop/utils/utils.py
```diff
@@ -108,11 +108,13 @@ def digits_needed(i: int):
...
 
 def grouped(n, ls):
+    if n < 1:
+        raise ValueError ('invalid value for n')
     while len(ls) != 0:
         group = ls[:n]
         ls = ls[n:]
         yield group
-
+
```
jigold, line 117: Can you get rid of this extra spacing here? This change should not show up in the diff. Try putting your cursor to the end of the line and then hitting backspace to get rid of the extra spaces.

hail/python/hailtop/utils/utils.py
```diff
@@ -108,11 +108,13 @@ def digits_needed(i: int):
 
 
 def grouped(n, ls):
+    if n < 1:
+        raise ValueError ('invalid value for n')
```
jigold, line 112: There shouldn't be a space between `ValueError` and `('invalid...`

jigold (changes requested):

hail/python/hailtop/utils/utils.py
CDiaz96: Fixed

hail/python/hailtop/utils/utils.py
CDiaz96: Fixed

hail/python/test/hailtop/utils/test_utils.py
```diff
@@ -126,3 +127,57 @@ def test_parse_docker_image_reference():
     assert x.digest is None
     assert x.name() == 'us-docker.pkg.dev/my-project/my-repo/test-image'
     assert str(x) == 'us-docker.pkg.dev/my-project/my-repo/test-image'
+
+
+
+
+def test_grouped_1():
+    actual = list(grouped(0, [1,2,3,4,5,6,7,8,9]))
+    expected = [[1], [2], [3], ...]
```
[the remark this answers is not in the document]
CDiaz96: I had previously removed this

hail/python/test/hailtop/utils/test_utils.py
```diff
@@ -126,3 +127,57 @@ def test_parse_docker_image_reference():
     assert x.digest is None
     assert x.name() == 'us-docker.pkg.dev/my-project/my-repo/test-image'
     assert str(x) == 'us-docker.pkg.dev/my-project/my-repo/test-image'
+
+
+
+
+def test_grouped_1():
+    actual = list(grouped(0, [1,2,3,4,5,6,7,8,9]))
```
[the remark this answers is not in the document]
CDiaz96: @jigold Should I change the current ValueError we have for this Exception? Or should I add the Exception in addition to the ValueError?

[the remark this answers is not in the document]
jigold: My reviews might have gotten messed up. Ignore this comment.

CDiaz96:
Created a new PR in favor of #11372

PR #11349, opened 2022-02-10
add additional steps to dev-docs based on Vedant's onboarding

vrautela:


dev-docs/development_process.md
```diff
@@ -35,13 +35,19 @@ make install-dev-deps
 
 to install the python dependencies.
 
+To install the python dependencies for Docker run
```
daniel-goldstein, line 38: I think this deserves a bit of an explanation, which I might not have given properly at the time. `make install-dev-deps` installs what you need to do truly "local" development, which is basically limited to Hail Query. These additional requirements are not so much for Docker as they are shared dependencies for our dockerized applications, i.e. our cloud services code. It would be useful to elaborate here that if you are doing batch or other services development you should install the following dependencies

daniel-goldstein (dismissed):
Just one nit

daniel-goldstein (approved):

PR #11352, opened 2022-02-10
[batch] Move volative n_completed and friends state out of batch table

daniel-goldstein:
Currently, the `mark_job_complete` SQL procedure deadlocks with itself, because the transaction takes a shared lock on the relevant row of the `batches` table, which it proceeds to try to upgrade to an exclusive lock later in the transaction when it calls `UPDATE batches …`. The lock upgrade is not an issue in itself, but it will cause a deadlock if we run multiple `mark_job_complete` transactions for the same batch at the same time, which we clearly would like to do.

One way to avoid this deadlock is to never issue an UPDATE to the batches table during the `mark_job_complete` procedure. Currently, we conduct the following updates to the batch row:
- increment number of completed jobs
- mark the batch as complete if the number of completed jobs is equal to the total number of jobs
- increment the number of successful/failed/cancelled jobs depending on the completion type of the job

This deadlock is a symptom of the problem that the `batches` table holds both very static information (like the billing project) and very volatile information like the number of completed jobs. It's not an issue to have many transactions holding read-only locks on rows of the batches table so long as it contains mostly static data. Therefore, it seems appropriate to move the job counters out of the batch table and into a new small table that just contains these counters. This eliminates all but one of the UPDATEs to the batch table in the procedure. The one sticky issue is marking the batch as complete. This is a much more popular column and I was hesitant to move it around just yet. I ended up moving the completeness update into a second transaction, which I think is not too bad for now. I also think that it could merge nicely with the following transaction that does the callback.

Resolving this deadlock uncovered yet another deadlock underneath, which I want to tackle next. This PR does NOT resolve the issue that all mark job complete transactions are serialized due to the way we increment these counters. This could be solved possibly through using tokens to use more rows of the new table per batch. I could try to tackle that in this PR if you think that is appropriate, but it shouldn't be related to deadlock issues.

@jigold


daniel-goldstein:
This is also my first migration PR so please take a very critical eye! I ran a load test on my namespace and noticed that while previously it looked like we only get deadlocks in MJC, the deadlock that this uncovered might be between MJC and SJ. Happy to provide details about the following deadlock to decide whether it's something we'd want to lump in here or tackle all at once.

daniel-goldstein:
Screen Shot 2022-02-10 at 4 59 10 PM
The SQL query monitoring that @vrautela added makes this really stark: the MJC and SJ SQL procedures are having to retry multiple times due to deadlocks, while the deactivate instance procedure is just slow for some reason.

jigold:
I still haven't looked at the code carefully, but one thing I'd say with database migrations is to make sure that your incremental change won't make the next change you want to make in the future more difficult or impossible. I don't think that's the case here, but keep in mind that this migration is O(n_batches). Anything that is O(n_jobs) right now is not great on hail-vdc as it could take hours to complete.

How big of a lift is it to the next change you wanted to try out? Where is the new deadlock error coming from?

daniel-goldstein:
Here's the deadlock that I'm observing now on this branch. I don't have a  great understanding of what's happening. I believe Transaction 1 to be [here](https://github.com/hail-is/hail/blob/40d8882470af71f2d08dd1aa6b723357ca8a1245/batch/sql/estimated-current.sql#L1186-L1188) in MJC, and Transaction 2 to be [here](https://github.com/hail-is/hail/blob/40d8882470af71f2d08dd1aa6b723357ca8a1245/batch/sql/estimated-current.sql#L449-L453) in the jobs_after_update trigger. Looking at the second transaction in context now, it looks like that is probably another MJC transaction toward the end of its run after it updated the jobs table. I think it would make sense then that T2 would still hold the lock for `instances_free_cores_mcpu` but I'm not sure where the contention for `batch_inst_coll_cancellable_res
ources` is coming from, as I don't see how T1 could be holding any form of lock on it. Either way it seems like how we use these tables is similarly a mess.

```
*** (1) TRANSACTION:
TRANSACTION 644409381, ACTIVE 0 sec starting index read
mysql tables in use 1, locked 1
LOCK WAIT 39 lock struct(s), heap size 3520, 50 row lock(s), undo log entries 28
MySQL thread id 1941960, OS thread handle 140297909716736, query id 1869168359 10.32.3.8 dgoldste updating
UPDATE instances_free_cores_mcpu
 SET free_cores_mcpu = free_cores_mcpu + cur_cores_mcpu
 WHERE instances_free_cores_mcpu.name = in_instance_name
*** (1) WAITING FOR THIS LOCK TO BE GRANTED:
RECORD LOCKS space id 1263041 page no 3 n bits 264 index PRIMARY of table `dgoldste`.`instances_free_cores_mcpu` trx i
d 644409381 lock_mode X locks rec but not gap waiting
Record lock, heap no 192 PHYSICAL RECORD: n_fields 4; compact format; info bits 0
 0: len 30; hex 62617463682d776f726b65722d64676f6c647374652d7374616e64617264; asc batch-worker-dgoldste-standard; (tot
al 36 bytes);
 1: len 6; hex 00002668e81a; asc &h ;;
 2: len 7; hex 710000071136b3; asc q 6 ;;
 3: len 4; hex 800029fe; asc ) ;;

*** (2) TRANSACTION:
TRANSACTION 644409370, ACTIVE 0 sec inserting
mysql tables in use 6, locked 6
39 lock struct(s), heap size 3520, 51 row lock(s), undo log entries 30
MySQL thread id 1941930, OS thread handle 140298159240960, query id 1869168731 10.32.3.8 dgoldste update
INSERT INTO batch_inst_coll_cancellable_resources (batch_id, inst_coll, token, n_running_cancellable_jobs, running_can
cellable_cores_mcpu)
 VALUES (OLD.batch_id, OLD.inst_coll, rand_token, -1, -OLD.cores_mcpu)
 ON DUPLICATE KEY UPDATE
 n_running_cancellable_jobs = n_running_cancellable_jobs - 1,
 running_cancellable_cores_mcpu = running_cancellable_cores_mcpu - OLD.cores_mcpu
*** (2) HOLDS THE LOCK(S):
RECORD LOCKS space id 1263041 page no 3 n bits 264 index PRIMARY of table `dgoldste`.`instances_free_cores_mcpu` trx i
d 644409370 lock_mode X locks rec but not gap
Record lock, heap no 192 PHYSICAL RECORD: n_fields 4; compact format; info bits 0
 0: len 30; hex 62617463682d776f726b65722d64676f6c647374652d7374616e64617264; asc batch-worker-dgoldste-standard; (tot
al 36 bytes);
 1: len 6; hex 00002668e81a; asc &h ;;
 2: len 7; hex 710000071136b3; asc q 6 ;;
 3: len 4; hex 800029fe; asc ) ;;

*** (2) WAITING FOR THIS LOCK TO BE GRANTED:
RECORD LOCKS space id 1263027 page no 15 n bits 232 index PRIMARY of table `dgoldste`.`batch_inst_coll_cancellable_res
ources` trx id 644409370 lock_mode X locks rec but not gap waiting
Record lock, heap no 23 PHYSICAL RECORD: n_fields 10; compact format; info bits 0
 0: len 8; hex 8000000000000007; asc ;;
 1: len 8; hex 7374616e64617264; asc standard;;
 2: len 4; hex 80000041; asc A;;
 3: len 6; hex 00002668e825; asc &h %;;
 4: len 7; hex 78000007ac2bcc; asc x + ;;
 5: len 4; hex 7ffffff9; asc ;;
 6: len 8; hex 7ffffffffffff92a; asc *;;
 7: len 4; hex 7ffffffc; asc ;;
 8: len 8; hex 7ffffffffffffc18; asc ;;
 9: len 4; hex 80000000; asc ;;
 ```

**Comment by danking** [comment]
OK, this is now higher priority for me. The Query-on-Batch tests are absolutely hammering the database with huge spikes in deadlock errors during working hours (when deploys trigger tests).

**Comment by jigold** [comment]
I don't know how to fix the deadlock errors. But for context, the next change Daniel had in mind is to have a table with the n_jobs counts (n_pending, n_failed, etc.) as well as time_completed and then use aggregation queries when the batch state is needed in the Python services. The time the batch is completed is just taken as the latest timestamp from the tokenized table.

**Comment by danking** [comment]
I went down that route once before and the main issue is how to trigger the change in batch state to completed. I couldn't figure out how to do that correctly and efficiently.

**Comment by jigold** [comment]
I thought it's just an aggregation when you query the batches table where the state is implicitly n_jobs == n_completed. When we create the batch, we can just populate the tokenized table with token 0 for the total n_jobs.

**Comment by danking** [comment]
Yeah, I think that should work. I remember corrupting the database and not being sure where I was missing a lock.

**Comment by danking** [comment]
In particular, isn't it possible that you have the n-1th and nth job racing to complete. Everyone else is already done. Call the n-1th job's transaction T1 and the nth job's transaction T2.

Both race down to this statement in MJC:
```
    UPDATE batches
      SET time_completed = new_timestamp,
          `state` = 'complete'
      WHERE id = in_batch_id AND n_completed = batches.n_jobs;
```

That will now need to have a sum(n_completed) over all tokens. The isolation level is repeatable read. Assume T1 and T2 generate non equal tokens.

T1 and T2 may both snapshot the state of the database before either T1 or T2 executes. T1 and T2 will necessarily see the changes they've made (which affect distinct rows because they have distinct tokens), but neither is required to see the changes the other has made.

I think the only way to guarantee that at least one of T1 or T2 sees the database with sum(n_completed) == n_jobs is for both of them to LOCK IN SHARE MODE when doing the sum(n_completed). That will cause lock contention. Maybe that's OK? In the worst case you could have this happen:

1. Job 1 executes all the way to just before the sum(n_completed).
2. Job 2 executes all the way to modifying the volatile state.
3. Job 1 blocks waiting for Job 2 to modify the volatile state.
4. Job 3 executes all the way to modifying the volatile state.
5. Job 1 and 2 now wait for Job 3 to modifying the volatile state.
6. ...
7. Job 1, 2, 3, n-1 now all wait for Job n to modify the volatile state.
8. Job 1...n finally execute the sum, all in parallel.

I guess that's not terrible, it just means that the latency of Job 1 is extended as long as other jobs can race in before it grabs a shared lock on all the rows.

File: batch/sql/estimated-current.sql [hunk 1]
```diff
@@ -1205,18 +1215,19 @@ BEGIN
...
 
     IF new_state = 'Cancelled' THEN
-      UPDATE batches SET n_cancelled = n_cancelled + 1 WHERE id = in_batch_id;
+      UPDATE batches_n_jobs_in_complete_states SET n_cancelled = n_cancelled + 1 WHERE id = in_batch_id;
     ELSEIF new_state = 'Error' OR new_state = 'Failed' THEN
-      UPDATE batches SET n_failed = n_failed + 1 WHERE id = in_batch_id;
+      UPDATE batches_n_jobs_in_complete_states SET n_failed = n_failed + 1 WHERE id = in_batch_id;
     ELSE
-      UPDATE batches SET n_succeeded = n_succeeded + 1 WHERE id = in_batch_id;
+      UPDATE batches_n_jobs_in_complete_states SET n_succeeded = n_succeeded + 1 WHERE id = in_batch_id;
```
danking, line 1230: Let's coalesce all this into the n_completed + 1 update:

```
UPDATE ...
SET n_completed = n_completed + 1
    n_cancelled = n_cancelled + (new_state = 'Cancelled')
    ...
```

jigold:
I was thinking we would get rid of that update completely in MJC. If you want to know the state or the time completed, then you query the tokenized table. We won't store these values explicitly.

batch/sql/estimated-current.sql
```diff
@@ -1205,18 +1215,19 @@ BEGIN
...
-    UPDATE batches
-      SET time_completed = new_timestamp,
-          `state` = 'complete'
-      WHERE id = in_batch_id AND n_completed = batches.n_jobs;
+    UPDATE batches_n_jobs_in_complete_states
+      SET n_completed = n_completed + 1
+      WHERE id = in_batch_id;
+    SELECT n_completed INTO new_n_completed
+    FROM batches_n_jobs_in_complete_states
+    WHERE id = in_batch_id;
```
danking, line 1223: You've dropped the modification of the state of a batch, which I think is a bug.

danking: In particular, for this to work I think you *must* move `state` into the volatile table.

I think the only things that a job should be permitted to place shared read locks on are:
- the user which owns the batch
- the billing project in which a batch resides
- whether a batch is cancelled
The first two are immutable properties of a batch.

danking: Ah, I see you moved the update to state into a separate transaction. I don't think that will help much.

I'm not sure we can do this, but if we can move the update of the parent jobs to before the update to the volatile state, then we'll only lock the batch row for a very short period of time at the very end of MJC. That shouldn't cause deadlocks and should have a small effect on latency.

In particular I imagine this:

```sql
UPDATE batches_n_jobs_in_complete_states, batches
SET n_completed = n_completed + 1,
     n_cancelled = n_cancelled + (new_state = 'Cancelled'),
     n_failed = n_failed + (new_state = 'Error' OR new_state = 'Failed'),
     n_succeeded = n_succeeded + (new_state != 'Cancelled' AND new_state != 'Error' AND new_state != 'Failed'),
    state = IF(n_completed + 1 = batches.n_jobs, 'complete', state)
WHERE id = in_batch_id
```

danking:
Oh, interesting idea Jackie. It has always bothered me that the batches state is duplicated as both the state column and the four job state count columns.

@daniel-goldstein , responding specifically to your point about the new deadlock. I agree, that seems strange. I'd bump up the number of tokens and hope that contention for batch_inst_coll_cancellable_resources goes down.

danking (approved):
I placed WIP.

After noodling further, I think your changes are an improvement. Please read over my comments and consider if we can further improve this in subsequent PRs. In the meantime, let's get this PR merged and see if it alleviates some of our database load. Please remove WIP at your earliest convenience.

danking (dismissed):
Actually, Jackie is right. If the batch-driver dies in between MJC and your update batches command, then that batch will never be marked complete.

batch/sql/estimated-current.sql
```diff
@@ -1205,18 +1215,24 @@ BEGIN
...
           `state` = 'complete'
-      WHERE id = in_batch_id AND n_completed = batches.n_jobs;
-
-    IF new_state = 'Cancelled' THEN
-      UPDATE batches SET n_cancelled = n_cancelled + 1 WHERE id = in_batch_id;
-    ELSEIF new_state = 'Error' OR new_state = 'Failed' THEN
-      UPDATE batches SET n_failed = n_failed + 1 WHERE id = in_batch_id;
-    ELSE
-      UPDATE batches SET n_succeeded = n_succeeded + 1 WHERE id = in_batch_id;
+      WHERE id = in_batch_id;
```
danking, line 1235: You don't need to fix here b/c it's already a present bug, but something I just noticed that you should keep in mind going forward: I think time_completed can jitter. It really should only change from NULL to a timestamp *once*.

1. T1 does n++
2. T2 does n++
3. T1 reads n_jobs == n_completed
4. T2 reads n_jobs == n_completed
5. T1 updates with its new_timestamp
6. T2 updates with its new_timestamp

I'm not sure how to solve it the "right" way without a `max` over all jobs attempts.

daniel-goldstein: Would 3 be true? I thought the point of repeatable read would be that T1 should always be able to select after an update and see its own write. It should see some stale view of the table and only the last job to grab the X lock on n_completed will actually see n_jobs == n_completed

daniel-goldstein: Or actually even stronger, the procedure gets serialized starting at the first `UPDATE`, since it must hold an exclusive lock through the end of the transaction (in case it needs to be rolled back) and then your order of 1-6 wouldn't actually be interleaved in that way

batch/sql/estimated-current.sql
```diff
@@ -1205,18 +1215,24 @@ BEGIN
...
+          n_succeeded = n_succeeded + (new_state != 'Cancelled' AND new_state != 'Error' AND new_state != 'Failed')
+      WHERE id = in_batch_id;
+
+    SELECT n_completed INTO new_n_completed
+    FROM batches_n_jobs_in_complete_states
+    WHERE id = in_batch_id;
+
+    # Grabbing an exclusive lock on batches here could deadlock,
+    # but this IF should only execute for the last job
+    IF new_n_completed = total_jobs_in_batch THEN
```
danking, line 1231: shouldn't this be two more spaces to the right? It's inside the 
```
ELSEIF cur_job_state = 'Ready' OR cur_job_state = 'Creating' OR cur_job_state = 'Running' THEN
```

danking: nope disregard.

danking:
It is mighty fishy that both azure and google failed the callback test. What are we missing? If MJC returns, then the database was clearly updated. Subsequent DB queries should see those changes.

total_jobs_in_batch won't change during the lifetime of the batch, so that should be correct (though we should probably LOCK IN SHARE MODE anyway).

Assuming I'm reading the [reference manual](https://dev.mysql.com/doc/refman/5.7/en/innodb-consistent-read.html) correctly, that select should see the result of the UPDATE *or a later state*. The updates to a single row are serial. So there must exist a transaction that takes it from n_jobs-1 to n_jobs. That transaction thus must see n_jobs for new_n_completed. That transaction thus ought to update batches. Once that transaction is committed the subsequent query for notification should see the changes...

batch/sql/estimated-current.sql
danking: T1 is able to see its own write, but might also see subsequent writes, which it does in my example. There's a really clear example of this here: https://dev.mysql.com/doc/refman/5.7/en/innodb-consistent-read.html.

danking:
@daniel-goldstein The logs show this:

```
insertId: "88db8ey9nom69366"
jsonPayload: {
asctime: "2022-02-16 23:07:40,794"
exc_info: "Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/batch/driver/job.py", line 65, in notify_batch_job_complete
    await request(session)
  File "/usr/local/lib/python3.7/dist-packages/batch/driver/job.py", line 56, in request
    await session.post(callback, json=batch_record_to_dict(record))
  File "/usr/local/lib/python3.7/dist-packages/batch/batch.py", line 18, in batch_record_to_dict
    elif record['n_failed'] > 0:
KeyError: 'n_failed'"
filename: "job.py"
funcNameAndLine: "notify_batch_job_complete:69"
hail_log: 1
levelname: "ERROR"
message: "callback for batch 1731 failed, will not retry."
```

`notify_batch_job_complete` just needs to join against the volatile table.

danking:
FWIW, I often grab the PR name from the `default_ns` job and search like this in the logs viewer:
```
resource.labels.namespace_name="PR_NAME"
labels."k8s-pod/app":"batch"
```

daniel-goldstein:
Ah, duh, thanks! And ya I do want to make it easier to get to error logs from the PR namespace.

danking (approved):

danking:
I'm leaving so slapping a WIP tag on for now.

Issue #11335, opened 2022-02-08
lowered table coercers construct invalid partitioners with prefix-sorted mode


tpoterba:
Example stack trace:
```
2022-02-08 18:09:30 root: ERROR: IllegalArgumentException: requirement failed
From java.lang.IllegalArgumentException: requirement failed
	at scala.Predef$.require(Predef.scala:268)
	at is.hail.rvd.RVDPartitioner.<init>(RVDPartitioner.scala:52)
	at is.hail.rvd.RVDPartitioner.extendKeySamePartitions(RVDPartitioner.scala:141)
	at is.hail.expr.ir.LoweredTableReader$$anon$2.coerce(TableIR.scala:382)
	at is.hail.expr.ir.GenericTableValue.toTableStage(GenericTableValue.scala:162)
	at is.hail.io.vcf.MatrixVCFReader.lower(LoadVCF.scala:1790)
	at is.hail.expr.ir.lowering.LowerTableIR$.applyTable(LowerTableIR.scala:581)
	at is.hail.expr.ir.lowering.LowerTableIR$.lower$2(LowerTableIR.scala:561)
	at is.hail.expr.ir.lowering.LowerTableIR$.applyTable(LowerTableIR.scala:1304)
	at is.hail.expr.ir.lowering.LowerTableIR$.lower$2(LowerTableIR.scala:561)
	at is.hail.expr.ir.lowering.LowerTableIR$.applyTable(LowerTableIR.scala:1035)
	at is.hail.expr.ir.lowering.LowerTableIR$.lower$1(LowerTableIR.scala:394)
	at is.hail.expr.ir.lowering.LowerTableIR$.apply(LowerTableIR.scala:547)
	at is.hail.expr.ir.lowering.LowerToCDA$.lower(LowerToCDA.scala:69)
	at is.hail.expr.ir.lowering.LowerToCDA$.apply(LowerToCDA.scala:18)
	at is.hail.expr.ir.lowering.LowerToDistributedArrayPass.transform(LoweringPass.scala:77)

```

called from here:
https://github.com/hail-is/hail/blob/d2f87d81dd1af43617740309e354d4bac8c672e0/hail/src/main/scala/is/hail/expr/ir/TableIR.scala#L382

PR #11354, opened 2022-02-11
[wip] replicate rvd partitioner failure

tpoterba:
@patrick-schultz 

This is a test that catches #11335. Can you take this and patch the problem? I think something about the allowed overlap.


patrick-schultz:
I don't seem to have permission to push directly to @tpoterba's branch, so I opened a new PR #11355.

PR #11357, opened 2022-02-14
[batch] Don't do O(jobs) logging in scheduling loop when no machines found

daniel-goldstein:
I'm happy to see if I can replace this with a more efficient observability solution, but according to the profiler these lines combined can take up to 8% of the driver's overall CPU time, which just seems like something we shouldn't do. The `get_instance` logging will be especially bad with big clusters because it builds up a whole histogram which then needs to get formatted and printed.

I'm not sure how useful these log statements currently are. I never look at them but maybe others do.

cc: @danking

### DIFF
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -159,11 +159,6 @@ WHERE removed = 0 AND inst_coll = %s;
             if user != 'ci' or (user == 'ci' and instance.location == self._default_location()):
                 return instance
             i += 1
-        histogram = collections.defaultdict(int)
-        for instance in self.healthy_instances_by_free_cores:
-            histogram[instance.free_cores_mcpu] += 1
-        log.info(f'schedule {self}: no viable instances for {cores_mcpu}: {histogram}')
-
         return None
 
     async def create_instance(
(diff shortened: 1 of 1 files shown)


batch/batch/driver/instance_collection/pool.py
```diff
@@ -159,11 +159,6 @@ def get_instance(self, user, cores_mcpu):
             if user != 'ci' or (user == 'ci' and instance.location == self._default_location()):
                 return instance
             i += 1
-        histogram = collections.defaultdict(int)
-        for instance in self.healthy_instances_by_free_cores:
-            histogram[instance.free_cores_mcpu] += 1
-        log.info(f'schedule {self}: no viable instances for {cores_mcpu}: {histogram}')
```
jigold, line 165: The original purpose of this logging statement was to make sure the fair share algorithm was working properly and understand why we weren't scheduling for some reason. Two users could deadlock if the share they were each entitled to was greater than the amount of free cores even if theoretically both jobs could fit in what free cores was available. We fixed the deadlock by putting a random bump into the code if the deadlock was too frequent. So I think we can do without this logging statement as it's not helpful.

```python3
                if scheduled_cores_mcpu + record['cores_mcpu'] > allocated_cores_mcpu:
                    if random.random() > self.exceeded_shares_counter.rate():
                        self.exceeded_shares_counter.push(True)
                        self.scheduler_state_changed.set()
                        break
                    self.exceeded_shares_counter.push(False)
```

jigold (approved):
I'm okay with this change, but I am open to other cheaper ways to detect the same deadlock issue and logging exceptions in that case.

PR #11359, opened 2022-02-14
[hailtop.utils]Test for digits needed

CDiaz96:


danking (changes requested):
Each PR should be for just one change. Can you modify this branch so that it only contains the digits needed test and none of the `grouped` tests?

CDiaz96:
@danking  Yes, I did not mean to do this. I am currently working on fixing it.

johnc1231:
Easiest way is probably to use git interactive rebase: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

And you can just delete the commits you don't want.

CDiaz96:
Thanks @johnc1231, I will read through this right now.

PR #11360, opened 2022-02-14
[query/service] use orjson and track timings

danking:
orjson is substantially faster.

I promise it looks nice with whitespace changes hidden.

### DIFF
--- a/hail/python/hail/backend/service_backend.py
+++ b/hail/python/hail/backend/service_backend.py
@@ -11,7 +11,7 @@ from pathlib import Path
 
 from hail.context import TemporaryDirectory, tmp_dir
 from hail.utils import FatalError
-from hail.expr.types import dtype, tvoid
+from hail.expr.types import dtype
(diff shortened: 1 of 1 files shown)


hail/python/hail/backend/service_backend.py
```diff
@@ -342,7 +344,7 @@ async def inputs(infile, _):
             await write_str(infile, self.billing_project)
             await write_str(infile, self.remote_tmpdir)
             await write_str(infile, self.render(tir))
-        _, resp = await self._rpc('value_type(...)', inputs)
+        _, resp, _ = await self._rpc('value_type(...)', inputs)
```
daniel-goldstein, line 347: Looks like an old bug, but should be `table_type`

daniel-goldstein (changes requested):
Unrelated to this change but one fix I noticed

danking:
I also fixed load_references_from_dataset.

daniel-goldstein (approved):

PR #11368, opened 2022-02-15
[batch] Save job memory and CPU statistics

jigold:
Closes https://github.com/hail-is/hail-tasks/issues/6

Stacked on #11353. This PR does everything necessary to give the user the CPU and memory statistics except expose them on the jobs UI page.

### DIFF
--- a/batch/batch/file_store.py
+++ b/batch/batch/file_store.py
@@ -48,7 +48,9 @@ class FileStore:
         url = self.log_path(format_version, batch_id, job_id, attempt_id, task)
         await self.fs.write(url, data.encode('utf-8'))
 
-    async def read_resource_usage_file(self, format_version, batch_id, job_id, attempt_id, task) -> Optional[pd.DataFrame]:
+    async def read_resource_usage_file(
+        self, format_version, batch_id, job_id, attempt_id, task
+    ) -> Optional[pd.DataFrame]:
         url = self.resource_usage_path(format_version, batch_id, job_id, attempt_id, task)
         data = await self.fs.read(url)
         return ResourceUsageMonitor.decode_to_df(data)
--- a/batch/batch/file_store.py
+++ b/batch/batch/file_store.py
@@ -1,10 +1,14 @@
 import asyncio
 import logging
+from typing import Optional
+
+import pandas as pd
 
 from hailtop.aiotools.fs import AsyncFS
 
 from .batch_format_version import BatchFormatVersion
 from .globals import BATCH_FORMAT_VERSION
+from .resource_usage import ResourceUsageMonitor
 from .spec_writer import SpecWriter
 
 log = logging.getLogger('logstore')
@@ -30,6 +34,11 @@ class FileStore:
             return f'{self.batch_log_dir(batch_id)}/{job_id}/{task}/log'
         return f'{self.batch_log_dir(batch_id)}/{job_id}/{attempt_id}/{task}/log'
 
+    def resource_usage_path(self, format_version, batch_id, job_id, attempt_id, task):
+        if not format_version.has_attempt_in_log_path():
+            return f'{self.batch_log_dir(batch_id)}/{job_id}/{task}/resource_usage'
+        return f'{self.batch_log_dir(batch_id)}/{job_id}/{attempt_id}/{task}/resource_usage'
+
     async def read_log_file(self, format_version, batch_id, job_id, attempt_id, task):
         url = self.log_path(format_version, batch_id, job_id, attempt_id, task)
         data = await self.fs.read(url)
@@ -38,6 +47,15 @@ class FileStore:
     async def write_log_file(self, format_version, batch_id, job_id, attempt_id, task, data):
         url = self.log_path(format_version, batch_id, job_id, attempt_id, task)
         await self.fs.write(url, data.encode('utf-8'))
+
+    async def read_resource_usage_file(self, format_version, batch_id, job_id, attempt_id, task) -> Optional[pd.DataFrame]:
+        url = self.resource_usage_path(format_version, batch_id, job_id, attempt_id, task)
+        data = await self.fs.read(url)
+        return ResourceUsageMonitor.decode_to_df(data)
+
+    async def write_resource_usage_file(self, format_version, batch_id, job_id, attempt_id, task, data):
+        url = self.resource_usage_path(format_version, batch_id, job_id, attempt_id, task)
+        await self.fs.write(url, data)
 
     async def delete_batch_logs(self, batch_id):
         url = self.batch_log_dir(batch_id)
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -10,7 +10,7 @@ import signal
 import traceback
 from functools import wraps
 from numbers import Number
-from typing import Dict, Optional, Union
+from typing import Any, Awaitable, Callable, Dict, Optional, Union
 
 import aiohttp
 import aiohttp_session
@@ -74,6 +74,7 @@ from ..exceptions import (
 from ..file_store import FileStore
 from ..globals import BATCH_FORMAT_VERSION, HTTP_CLIENT_MAX_SIZE
 from ..inst_coll_config import InstanceCollectionConfigs
+from ..resource_usage import ResourceUsageMonitor
 from ..spec_writer import SpecWriter
 from ..utils import accrued_cost_from_cost_and_msec_mcpu, coalesce, query_billing_projects
 from .validate import ValidationError, validate_and_clean_jobs, validate_batch
@@ -330,63 +331,7 @@ WHERE id = %s AND NOT deleted;
     return web.json_response(resp)
 
 
-async def _get_job_log_from_record(app, batch_id, job_id, record):
-    client_session: httpx.ClientSession = app['client_session']
-    batch_format_version = BatchFormatVersion(record['format_version'])
-
-    state = record['state']
-    ip_address = record['ip_address']
-
-    spec = json.loads(record['spec'])
-    tasks = []
-
-    has_input_files = batch_format_version.get_spec_has_input_files(spec)
-    if has_input_files:
-        tasks.append('input')
-
-    tasks.append('main')
-
-    has_output_files = batch_format_version.get_spec_has_output_files(spec)
-    if has_output_files:
-        tasks.append('output')
-
-    if state == 'Running':
-        try:
-            resp = await request_retry_transient_errors(
-                client_session, 'GET', f'http://{ip_address}:5000/api/v1alpha/batches/{batch_id}/jobs/{job_id}/log'
-            )
-            return await resp.json()
-        except aiohttp.ClientResponseError:
-            log.exception(f'while getting log for {(batch_id, job_id)}')
-            return {task: 'ERROR: encountered a problem while fetching the log' for task in tasks}
-
-    if state in ('Pending', 'Ready', 'Creating'):
-        return None
-
-    if state == 'Cancelled' and record['last_cancelled_attempt_id'] is None:
-        return None
-
-    assert state in ('Error', 'Failed', 'Success', 'Cancelled')
-
-    attempt_id = record['attempt_id'] or record['last_cancelled_attempt_id']
-    assert attempt_id is not None
-
-    file_store: FileStore = app['file_store']
-    batch_format_version = BatchFormatVersion(record['format_version'])
-
-    async def _read_log_from_cloud_storage(task):
-        try:
-            data = await file_store.read_log_file(batch_format_version, batch_id, job_id, attempt_id, task)
-        except FileNotFoundError:
-            id = (batch_id, job_id)
-            log.exception(f'missing log file for {id} and task {task}')
-            data = 'ERROR: could not find log file'
-        return task, data
-
-    return dict(await asyncio.gather(*[_read_log_from_cloud_storage(task) for task in tasks]))
-
-
-async def _get_job_log(app, batch_id, job_id):
+async def _get_job_record(app, batch_id, job_id):
     db: Database = app['db']
 
     record = await db.select_and_fetchone(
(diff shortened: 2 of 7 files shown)


danking (changes requested):
I haven't looked at front_end.py yet

batch/batch/worker/worker.py
```diff
@@ -548,18 +624,25 @@ async def _localize_rootfs():
             if not isinstance(e, (JobDeletedError, JobTimeoutError)) and not user_error(e):
                 log.exception(f'while running {self}')
 
-            self.state = 'error'
-            self.error = traceback.format_exc()
+            if isinstance(e, JobDeletedError):
+                self.state = 'cancelled'
+            else:
```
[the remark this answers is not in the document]
jigold: It's in the PR this is stacked on.

batch/batch/worker/worker.py
```diff
@@ -1701,9 +1820,20 @@ async def cleanup(self):
...
+        full_status = await retry_all_errors(f'error while getting status for {self}')(self.status)
+
+        if self.format_version.has_full_status_in_gcs():
+            await retry_all_errors(f'error while writing status file to cloud storage for {self}')(
+                self.worker.file_store.write_status_file,
+                self.batch_id,
+                self.job_id,
+                self.attempt_id,
+                json.dumps(full_status),
+            )
```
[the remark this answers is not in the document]
jigold: See stacked PR

batch/batch/worker/worker.py
```diff
@@ -178,6 +180,77 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
...
+        self.last_cpu_nanosecs = now_cpu_nanosecs
+
+        return cpu_usage
+
+    def memory_usage_bytes(self) -> Optional[int]:
+        try:
+            with open(f'/sys/fs/cgroup/memory/{self.container_name}/memory.usage_in_bytes', 'r') as f:
+                return int(f.read().rstrip())
+        except FileNotFoundError:
+            return None
```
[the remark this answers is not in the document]
jigold: There's probably a race condition where there's a delay in the container being created and when we start measuring because we don't know when the container has actually started. All we know is we have started the process that calls `crun run`. Would you rather make another `crun` call to get whether the container exists yet before starting to measure?

FWIW, the FileNotFoundError happened every time.

batch/batch/worker/worker.py
```diff
@@ -178,6 +180,77 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
...
+            with open(f'/sys/fs/cgroup/cpu/{self.container_name}/cpuacct.usage', 'r') as f:
+                now_cpu_nanosecs = int(f.read().rstrip())
+        except FileNotFoundError:
+            return None
+
+        if self.last_time_nanosecs is not None:
+            assert self.last_cpu_nanosecs is not None
+            cpu_usage = (now_cpu_nanosecs - self.last_cpu_nanosecs) / (now_time_nanosecs - self.last_time_nanosecs)
+        else:
+            cpu_usage = None
```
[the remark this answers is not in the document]
jigold: We don't actually know this time. See comment below.

batch/batch/worker/worker.py
```diff
@@ -178,6 +180,77 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
 image_lock = aiorwlock.RWLock()
 
 
+async def read_file(fs: LocalAsyncFS, path: str) -> str:
+    if os.path.exists(path):
+        stream = await fs.open(path)
+        async with stream:
+            return (await stream.read()).decode()
```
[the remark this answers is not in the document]
jigold: This doesn't seem right. I didn't change the existing code at all.

batch/batch/front_end/front_end.py
```diff
@@ -1626,7 +1668,8 @@ async def ui_get_job(request, userdata, batch_id):
         'job_status_str': json.dumps(job, indent=2),
         'step_errors': step_errors,
         'error': job_status.get('error'),
-        'plot_json': plot_json,
+        'plot_json': plot_job_durations(container_statuses, batch_id, job_id),
```
danking, line 1671: Let's change this Jinja2 name if we now have two plots.

danking: it's used in job.html

batch/batch/worker/worker.py
```diff
@@ -171,6 +174,68 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
...
+        return cpu_usage
+
+    def memory_usage_bytes(self) -> Optional[int]:
+        with open(f'/sys/fs/cgroup/memory/{self.container_name}/memory.usage_in_bytes', 'r') as f:
+            return int(f.read().rstrip())
+
+    async def measure(self):
+        data = [time_msecs(), self.percent_cpu_usage(), self.memory_usage_bytes()]
+        data = [str(d) for d in data]
+        with open(self.output_file_path, 'a') as f:
```
danking, line 224: After this merges, we should peek at the workers and make sure their CPU use under heavy load is not too bad. Opening and closing files takes some time. If we have 64 jobs in parallel that's ~12 files per second. I feel like that should be OK, but it would be good to check that a fully loaded worker still monitors at a reasonable pace.

batch/batch/worker/worker.py
```diff
@@ -664,18 +734,39 @@ async def run_container(self) -> bool:
...
                         self.container_name,
                         stdout=container_log,
                         stderr=container_log,
                     )
-                    await self.process.wait()
+
+                    async def wait_for_pid_file():
+                        delay = 0.01
+                        while not os.path.exists(pid_file):
+                            delay = await sleep_and_backoff(delay)
```
danking, line 756: Does the pid file not immediately exist after `crun run` returns? I believe `crun state` will also print the pid.

danking: crun's tests appear to immediately read the pid file after `crun run` returns.

https://github.com/containers/crun/blob/2d58f66d931cca0454fbd10529665922112bb98f/tests/test_pid_file.py#L27-L31

hail/python/hailtop/utils/time.py
```diff
@@ -7,6 +7,10 @@ def time_msecs() -> int:
     return int(time.time() * 1000 + 0.5)
 
 
+def time_ns() -> int:
+    return int(time.monotonic_ns() * 1000**3 + 0.5)
```
danking, line 11: I'm confused by the use of 1000**3, isn't it already in nanoseconds?

batch/batch/worker/worker.py
```diff
@@ -171,6 +174,68 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
...
+
+    def memory_usage_bytes(self) -> Optional[int]:
+        with open(f'/sys/fs/cgroup/memory/{self.container_name}/memory.usage_in_bytes', 'r') as f:
+            return int(f.read().rstrip())
+
+    async def measure(self):
+        data = [time_msecs(), self.percent_cpu_usage(), self.memory_usage_bytes()]
+        data = [str(d) for d in data]
+        with open(self.output_file_path, 'a') as f:
+            f.write('\t'.join(data) + '\n')
```
danking, line 225: Hmm. I'm uneasy about generating text. It just seems an unnecessary ~2x inflation in size. We could write a simple binary format that included a version number in the eight bytes. The remaining bytes could just be 64-bit long time, 64-bit float cpu use, 64-bit long memory usage.

Let's talk about this tomorrow maybe? If we don't do this right now I want to make sure we have a versioning plan for the resource usage files so that we could switch to a compact binary format in the future.

batch/batch/worker/worker.py
```diff
@@ -171,6 +174,68 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
...
+
+
+class ResourceUsageMonitor:
+    VERSION = 1
+
+    def __init__(self, container_name: str, output_file_path: str, create_time_secs: int):
+        self.container_name = container_name
+        self.output_file_path = output_file_path
+
+        self.last_time_ns: Optional[int] = int(create_time_secs * 10**3)
```
danking, line 190: this is never set to None

batch/batch/worker/worker.py
```diff
@@ -171,6 +174,68 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
...
+        self.last_cpu_ns: Optional[int] = 0  # assume zero CPU usage at container start time
```
danking, line 191: this is never set to None (so the type is overly broad)

batch/batch/worker/worker.py
```diff
@@ -171,6 +174,68 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
...
+
+    def cpu_ns(self) -> int:
+        with open(f'/sys/fs/cgroup/cpu/{self.container_name}/cpuacct.usage', 'r') as f:
+            return int(f.read().rstrip())
+
+    def percent_cpu_usage(self) -> float:
+        now_time_ns = time_ns()
+        now_cpu_ns = self.cpu_ns()
+
+        assert self.last_time_ns is not None and self.last_cpu_ns is not None
```
danking, line 208: this assert is never false

batch/batch/worker/worker.py
```diff
@@ -171,6 +174,68 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
...
+
+    def __enter__(self):
+        self.write_header()
+        self.task = asyncio.ensure_future(
+            retry_long_running(f'monitor {self.container_name} resource usage', periodically_call, 5, self.measure)
+        )
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self.task:
```
danking, line 235: assert self.task is not none. We should never get here unless we already entered.

danking (changes requested):

danking:
Hmm, somehow I made to separate reviews. Sorry about that, should just be one set of comments

batch/batch/worker/worker.py
jigold: I was trying to make this easy to get in for Carolin to be able to do what I want her to do. Would gzip solve the problem?

batch/batch/worker/worker.py
jigold: Would it be better to have an open file handle?

jigold: Actually, the reason why I didn't have an open file handle is we need it to flush to disk after every line is written. I can directly add a flush.

batch/batch/worker/worker.py
jigold: I thought about this some more and I think you are right and we should have our own format now with proper versioning. I can write a function that converts the data to a Pandas data frame.

batch/batch/worker/worker.py
jigold: I just tried this and am getting file not found error. Will revert back to my original code.

jigold:
I'm having a pretty hard time with the binary format. It's also incredibly slow to load the data in the UI when the data is being recorded properly. My approach must not be right. Can you double check this is what you had in mind?

batch/batch/worker/worker.py
danking: Ah, sure, we don't wait for the crun run process to exit so of course there's a race. OK.

danking:
Did the resource_usage.py file not get pushed?

jigold:
yep whoops

batch/batch/resource_usage.py
```diff
@@ -0,0 +1,108 @@
...
+        while offset < len(data):
+            timestamp_msecs, cpu_usage, memory_in_bytes = struct.unpack_from('>qdq', data, offset)
+            decoded_data.append(
+                {'time_msecs': timestamp_msecs, 'cpu_usage': cpu_usage, 'memory_in_bytes': memory_in_bytes}
+            )
+            offset += 24
+
+        if len(decoded_data) == 0:
+            return ResourceUsageMonitor.empty_df()
+        return pd.DataFrame(data=decoded_data)
```
danking, line 46: do you need to specify the column names here?

batch/batch/resource_usage.py
```diff
@@ -0,0 +1,108 @@
...
+        (version,) = struct.unpack_from('>i', data, offset)
+        offset += 4
+        assert version == ResourceUsageMonitor.VERSION, version
+
+        while offset < len(data):
+            timestamp_msecs, cpu_usage, memory_in_bytes = struct.unpack_from('>qdq', data, offset)
+            decoded_data.append(
+                {'time_msecs': timestamp_msecs, 'cpu_usage': cpu_usage, 'memory_in_bytes': memory_in_bytes}
+            )
+            offset += 24
```
danking, line 42: I think the main issue is that we're spending too much time in this Python while loop and not enough time in optimized C code.

Use the built-in repetition system (described under [Format Characters](https://docs.python.org/3/library/struct.html#format-characters)):
> A format character may be preceded by an integral repeat count. For example, the format string '4h' means exactly the same as 'hhhh'.

Super tiny minor thing that probably doesn't matter, but let's use a long for the version. That ensures everything is aligned on 8-byte boundaries.

danking: You should be able to reshape efficiently with numpy:
```python3
n_records = (len(data) - HEADER_LENGTH) // 3
np.array(data).reshape((n_records, 3))
```

danking (changes requested):

danking:
If with my changes it is still slower, then let's abandon. My bad for heading us down this route. Seems natural to use binary, but given that pandas probably has native code for parsing TSVs, maybe we're fighting a losing battle.

batch/batch/resource_usage.py
jigold: Using numpy helped a lot!

jigold:
The other issue is I'm getting a small negative value for the first time point on CPU. Not sure how to fix that. Maybe we just ignore it for now? Or revert back to the simpler code where we omit the first timepoint? Although I don't know how to represent NaN in binary data.

batch/batch/file_store.py
```diff
@@ -39,6 +45,15 @@ async def write_log_file(self, format_version, batch_id, job_id, attempt_id, tas
         url = self.log_path(format_version, batch_id, job_id, attempt_id, task)
         await self.fs.write(url, data.encode('utf-8'))
 
+    async def read_resource_usage_file(self, format_version, batch_id, job_id, attempt_id, task):
+        url = self.resource_usage_path(format_version, batch_id, job_id, attempt_id, task)
+        data = await self.fs.read(url)
+        return ResourceUsageMonitor.decode_to_df(data).to_json()
```
danking, line 51: I think we should just keep the data in a DataFrame. The output of this function goes straight into `plot_resource_usage`, right? `plot_resource_usage` needs to convert it back to a data frame anyway so that it can send the data into plotly.

danking (changes requested):
Can you elaborate on the multipart thing? I don't follow where that's involved

batch/batch/file_store.py
jigold: The multipart is to essentially send 3 binary files together. Same as if you were downloading multiple images at the same time. This way we work with the raw binary and don't use the `to_json()` anywhere.

jigold: If I can't get the multipart to work easily, then I think I should just make 3 separate requests to get the data.

danking: Hmm. I still don't follow. Who is making a multipart request to whom?

Regarding, `to_json`, I don't think we need to convert the data frame to json. We use the data frame to construct a plotly plot, right? And then the plotly plot will be converted to json, but not the data frame itself.

jigold:
Stacked on #11397

danking (approved):

PR #11370, opened 2022-02-15
properly import submodules from kubernetes_asyncio

daniel-goldstein:
I use `pyright` as my LSP server and it gets mad at me with our old form of importing saying that `client is not a known member of kubernetes_asyncio`,  which as far as I can tell is because they don't actually use `__all__` and just import their submodules in their init? Anyway this seemed better to me and allows me to actually get the LSP hints for this library now, but let me know if this is just a Me thing or if I don't understand python modules

### DIFF
--- a/auth/auth/driver/driver.py
+++ b/auth/auth/driver/driver.py
@@ -6,7 +6,9 @@ import logging
 import secrets
 import asyncio
 import aiohttp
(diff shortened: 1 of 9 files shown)


danking (approved):
I have no qualms either way

PR #11371, opened 2022-02-16
Bootstrap render_config_mk: add missing variables

vladsaveliev:
Few more variables are expected to be in `config.mk` for manual bootstrap:

* `DOCKER_ROOT_IMAGE` used to build batch workers and benchmark
* `HAIL_TEST_GCS_BUCKET` used to build query
* `KUBERNETES_SERVER_URL` used to build amundsen
* `PROJECT`, `ZONE`, `REGION` are probably not need, but might make sense to add for consistency.

Also match the order from `global-config`.

### DIFF
--- a/infra/bootstrap_utils.sh
+++ b/infra/bootstrap_utils.sh
@@ -8,18 +8,29 @@ fi
 source $HAIL/devbin/functions.sh
 
 render_config_mk() {
+    CLOUD=$(get_global_config_field cloud)
     DOCKER_PREFIX=$(get_global_config_field docker_prefix)
+    DOCKER_ROOT_IMAGE=$(get_global_config_field docker_root_image)
+    DOMAIN=$(get_global_config_field domain)
+    PROJECT=$(get_global_config_field gcp_project)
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
Hi Vlad, thanks for the PR! I'm afraid there are some internal migrations we're making that are probably not clear from just looking at the codebase. Are you up to date on our `main`? We've found working with `config.mk` cumbersome because it can be stale if you switch between different instances of Batch (e.g. one deployed in azure and the other in GCP).

> DOCKER_ROOT_IMAGE used to build batch workers and benchmark

I've recently updated the scripts for building the batch worker VM image to query kubernetes directly and we should probably do the same for benchmark.

> HAIL_TEST_GCS_BUCKET used to build query
KUBERNETES_SERVER_URL used to build amundsen

These are both currently deleted in our `main`.

> PROJECT, ZONE, REGION are probably not need, but might make sense to add for consistency.

These will fail in an Azure deployment, and while we want to move away from `config.mk` entirely, we would at least want it to contain configurations that are valid across clouds.

PR #11372, opened 2022-02-16
[hailtop.utils]test for grouped

CDiaz96:


hail/python/test/hailtop/utils/test_utils.py
```diff
@@ -126,3 +127,58 @@ def test_parse_docker_image_reference():
     assert x.digest is None
     assert x.name() == 'us-docker.pkg.dev/my-project/my-repo/test-image'
     assert str(x) == 'us-docker.pkg.dev/my-project/my-repo/test-image'
+
+
+
```
danking, line 132: only two spaces between function definitions please!

hail/python/hailtop/utils/utils.py
```diff
@@ -108,6 +108,8 @@ def digits_needed(i: int):
 
 
 def grouped(n, ls):
+    if n < 1:
+        raise ValueError('invalid value for n')
```
danking, line 112: Let's also include the bad `n` value in the error message. You can use a [Python format string](https://docs.python.org/3/tutorial/inputoutput.html#fancier-output-formatting) to include the number `n` in the string itself.

danking (changes requested):

hail/python/test/hailtop/utils/test_utils.py
CDiaz96: Fixed

danking (approved):
Looks great!

PR #11376, opened 2022-02-16
[batch] Dont log user errors from JVM jobs

daniel-goldstein:
I'll admit I didn't test this since I'm currently debugging the deadlock PR, but this should quiet down the error logs that are being emitted whenever a test runs that intentionally throws an exception in Query.


jigold (approved):

daniel-goldstein:
Closing in favor of Dan's

PR #11377, opened 2022-02-17
[batch] do not log user errors

danking:


batch/batch/worker/worker.py
```diff
@@ -1670,8 +1670,11 @@ async def run(self):
                 log.info(f'{self} main: {self.state}')
             except asyncio.CancelledError:
                 raise
+            except JVMUserError:
+                self.state = 'failed'
+                self.error = traceback.fromat_exc()
```
jigold, line 1675: format is still misspelled

jigold (changes requested):

jigold (approved):

PR #11378, opened 2022-02-17
[hailtop.batch] allow for remote-only dependencies, allow nested reso…

danking:
…urces

TJ has raised a bunch of very legitimate concerns about the expressivity of
hailtop.batch. In this PR, I basically rebuilt hailtop.batch from scratch as
a learning exercise. Once I understood how it worked, I added two valuable
features:

1. `hb.remote` creates a "remote" resource which is never localized but still
   creates dependency relationships.

2. `ResourceToStringPickler` which pickles resources appropriately even if
   they are nested within other structures.


danking:
Closing, I will come back to this

PR #11381, opened 2022-02-17
[query] LowerDistributedSort cutoff based on bytes not rows

johnc1231:
With this, we should theoretically be able to arbitrarily sized rows.

### DIFF
--- a/hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
+++ b/hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
@@ -4,8 +4,9 @@ import is.hail.annotations.Region
 import is.hail.asm4s._
 import is.hail.expr.ir.EmitCodeBuilder
 import is.hail.types.physical.stypes.interfaces._
+import is.hail.types.physical.stypes.primitives.SInt64Value
 import is.hail.types.physical.stypes.{SCode, SType, SValue}
-import is.hail.types.physical.{PCanonicalNDArray, PType}
+import is.hail.types.physical.{PCanonicalNDArray, PNDArray, PType}
 import is.hail.types.virtual.Type
 import is.hail.utils.{FastIndexedSeq, toRichIterable}
 
@@ -103,6 +104,12 @@ class SNDArrayPointerValue(
       }
       pt.elementType.storeAtAddress(cb, ptrs.head, region, body(codes), deepCopy)
     }
+  }
+
+  override def sizeInBytes(cb: EmitCodeBuilder): SInt64Value = {
+    val storageType = st.storageType().asInstanceOf[PNDArray]
+    storageType.byteSize
+    new SInt64Value(???)
   }
 }
 
--- a/hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
+++ b/hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
@@ -105,12 +105,6 @@ class SNDArrayPointerValue(
       pt.elementType.storeAtAddress(cb, ptrs.head, region, body(codes), deepCopy)
     }
   }
-
-  override def sizeInBytes(cb: EmitCodeBuilder): SInt64Value = {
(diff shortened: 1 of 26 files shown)


johnc1231:
Chris, the main thing that needs reviewing is the implementation of `sizeToStoreInBytes`. The `EmitStreamDistribute` stuff is just adding `sizeInBytes` to every write out, and the `LowerDistributedSort` stuff is mostly logic to track sizes and make sure that if a single row is bigger than our size cutoff, we don't get stuck in an infinite loop

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SStackStruct.scala
```diff
@@ -1,11 +1,14 @@
 package is.hail.types.physical.stypes.concrete
 
 import is.hail.annotations.Region
-import is.hail.asm4s.{Settable, TypeInfo, Value}
+import is.hail.asm4s.{Settable, TypeInfo, Value, const}
```
chrisvittal, line 4: unused/unnecessary with the glob import later

hail/src/main/scala/is/hail/types/physical/stypes/concrete/SNDArrayPointer.scala
```diff
@@ -4,8 +4,9 @@ import is.hail.annotations.Region
 import is.hail.asm4s._
 import is.hail.expr.ir.EmitCodeBuilder
 import is.hail.types.physical.stypes.interfaces._
+import is.hail.types.physical.stypes.primitives.SInt64Value
 import is.hail.types.physical.stypes.{SCode, SType, SValue}
-import is.hail.types.physical.{PCanonicalNDArray, PType}
+import is.hail.types.physical.{PCanonicalNDArray, PNDArray, PType}
```
chrisvittal, line 9: unused?

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SInterval.scala
```diff
@@ -31,4 +32,21 @@ trait SIntervalValue extends SValue {
...
+
+    loadStart(cb).consume(cb, {}, {sv =>
+      sv.sizeToStoreInBytes(cb)
+    })
+
+    loadEnd(cb).consume(cb, {}, {sv =>
+      sv.sizeToStoreInBytes(cb)
+    })
+
+    new SInt64Value(???)
```
chrisvittal, line 50: unimplemented value?

chrisvittal (changes requested):
Looks good to me. Just a few nits.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SBaseStruct.scala
```diff
@@ -57,6 +57,23 @@ trait SBaseStructValue extends SValue {
     new SInt32Value(hash_result)
   }
 
+  override def sizeToStoreInBytes(cb: EmitCodeBuilder): SInt64Value = {
+    // Size in bytes of the struct that must represent this thing, plus recursive call on any non-missing children.
+    val pStructSize = this.st.storageType().byteSize
+    val sizeSoFar = cb.newLocal[Long]("sstackstruct_size_in_bytes", pStructSize)
+    (0 until st.size).foreach { idx =>
+      if (!this.st.fieldTypes(idx).isPrimitive) {
```
chrisvittal, line 65: I think there are non-primitive types that are stored inline, so I think that we should switch on if the value points outside the struct, rather than if it's primitive.

johnc1231: Hmm, do we have a way to check if the value points outside the struct?

johnc1231: `containsPointers` check if something has pointers inside it, but it doesn't tell me if a particular element of this struct is inline or not.

johnc1231:
Failing tests are resolved, should be good to review again.

chrisvittal (approved):

PR #11383, opened 2022-02-17
[batch] demote callback failure to info

danking:
This is not a real error. It is not the fault of batch that the other
end of the connection did not respond. If batch is having general
network connectivity issues, I would expect to see many other
error log statements.


danking:
@daniel-goldstein this is a one-line change that will alleviate some of our log alerts.

daniel-goldstein (approved):

PR #11384, opened 2022-02-17
[batch] wait unless we have more work

danking:
The return value of these functions indicates if the containing loop
should wait or if we should immediately re-call the function. This
is intended to be used to allow functions which *know* they have more
work to eagerly invoke themselves again.

The use of this variable seems to have been changed to basically always
eagerly re-run during the Azure work.

This change restores the original behavior:
1. Do not wait in job private if we saw 300 records (seems likely there were
   301 or more records in the db).
2. Do not wait in pool scheduler if we exhaust a user's share. I do not
   fully follow the pool scheduler's logic. There might be something
   smarter we can do. I think we should really only re-call if we believe
   the db contains more ready jobs and we have available cores.

### DIFF
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -459,7 +459,6 @@ LIMIT %s;
                     instance.adjust_free_cores_in_memory(-record['cores_mcpu'])
                     scheduled_cores_mcpu += record['cores_mcpu']
                     n_scheduled += 1
-                    should_wait = False
 
                     async def schedule_with_error_handling(app, record, id, instance):
                         try:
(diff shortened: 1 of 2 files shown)


danking:
Hey @daniel-goldstein , Jackie is off this afternoon, could you take a quick look here? This has cropped up a couple times now.

batch/batch/driver/instance_collection/pool.py
```diff
@@ -471,11 +470,12 @@ async def schedule_with_error_handling(app, record, id, instance):
...
                 remaining.value -= 1
                 if remaining.value <= 0:
+                    should_wait = False
                     break
 
         await waitable_pool.wait()
 
         end = time_msecs()
-        log.info(f'schedule: scheduled {n_scheduled} jobs in {end - start}ms for {self.pool}')
+        log.info(f'schedule: attemptd to schedule {n_scheduled} jobs in {end - start}ms for {self.pool}')
```
daniel-goldstein, line 479: attempted spelled wrong

daniel-goldstein (dismissed):
Good find, just a typo

daniel-goldstein (approved):

danking:
Reducing pro because other PRs will avoid triggering this and are more important.

PR #11389, opened 2022-02-18
[batch] Incorporate WebSockets into driver-worker communication

vrautela:


batch/batch/worker/worker.py
```diff
@@ -2271,12 +2276,38 @@ async def run(self):
...
+        )
+        aenter = type(self.ws_connection_manager).__aenter__
+        self.ws_connection = await aenter(self.ws_connection_manager)
+
+    async def close_websocket_connection(self):
+        if self.ws_connection_manager is not None and self.ws_connection is not None:
+            aexit = type(self.ws_connection_manager).__aexit__
+            await aexit(self.ws_connection_manager, None, None, None)
+
+    def send_request_via_ws_connection(self, url, json, headers):
```
jigold, line 2294: Is this used?

batch/batch/worker/worker.py
```diff
@@ -2271,12 +2276,38 @@ async def run(self):
...
+    async def open_websocket_connection(self):
+        self.ws_connection_manager = self.client_session.ws_connect(
+            deploy_config.url('batch-driver', '/api/v1alpha/worker_wss')
+        )
+        aenter = type(self.ws_connection_manager).__aenter__
+        self.ws_connection = await aenter(self.ws_connection_manager)
+
+    async def close_websocket_connection(self):
+        if self.ws_connection_manager is not None and self.ws_connection is not None:
+            aexit = type(self.ws_connection_manager).__aexit__
```
jigold, line 2291: Why do you need to get the `type` here?

batch/batch/worker/worker.py
```diff
@@ -2271,12 +2276,38 @@ async def run(self):
...
+            await aexit(self.ws_connection_manager, None, None, None)
```
jigold, line 2292: Set `self.ws_connection_manager` to None as well as `self.ws_connection`.

batch/batch/worker/worker.py
```diff
@@ -2064,6 +2064,10 @@ def __init__(self, client_session: httpx.ClientSession):
         self.headers = None
         self.compute_client = None
 
+        # initialized after activation
+        self.ws_connection_manager = None
```
jigold, line 2068: Can you add types to the attributes here? It would look something like:

```
self.foo: Optional[int] = None
```

batch/batch/driver/main.py
```diff
@@ -375,6 +376,30 @@ async def job_started(request, instance):
...
+    async for msg in ws:
+        if msg.type == aiohttp.WSMsgType.TEXT:
+            if msg.data == 'close':
+                await ws.close()
+            else:
+                print(f"Received message: {msg.data}")
+                await ws.send_str(msg.data + '/answer')
+        elif msg.type == aiohttp.WSMsgType.ERROR:
+            print('ws connection closed with exception %s' % ws.exception())
+
```
jigold, line 396: I think you want a `finally` to close the `ws` and decrement the `HTTP_ACTIVE_WEBSOCKETS`. I'm going off of the code in `auth/auth.py` for `_wait_websocket`.

jigold (changes requested):
How big of an improvement does this feature provide compared to the overhead of having an additional code path such that the old code path is never tested by CI and we have added complexity to maintain?

When Batch is under significant load, at what point do we make the behavior of communication between the driver and the worker even worse? For example, every request from the worker to the driver will require 2 calls since the first to initiate the web socket will fail every time?

What do we expect to happen with web sockets when there are too many active connections? Will there be deadlocks like we experienced with the MySQL database connection pool? Will connections be so short-lived that they provide no extra benefit?

This will make things more complicated, but we should probably track on the worker the number of times we fail to establish a connection and then fallback to the original code path. I'm thinking this is similar to how we use a `WindowFractionCounter` and the probability of choosing a particular connection type is based on the success rate of previous calls with some kind of decay over time so we don't assume one code path will always fail.

PR #11390, opened 2022-02-18
Change default styles for hail ggplot

johnc1231:
- Better defaults for professional looking plots
- Adjusted title size defaults

The old defaults were somewhat arbitrary, the new defaults are requested by Alicia's group. It's likely defaults will continue to move around a bit until we arrive at something stable and have good options in place to tweak plot aesthetics.

### DIFF
--- a/hail/python/hail/ggplot/ggplot.py
+++ b/hail/python/hail/ggplot/ggplot.py
@@ -171,11 +171,6 @@ class GGPlot:
 
         fig = go.Figure()
 
-        fig.update_layout(
-            template="simple_white",
-            font_family="Ariel"
-        )
-
         for geom, (geom_label, agg_result) in zip(self.geoms, aggregated.items()):
             df_agg_result = labels_to_stats[geom_label].listify(agg_result)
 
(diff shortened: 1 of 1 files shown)


hail/python/hail/ggplot/ggplot.py
```diff
@@ -206,6 +206,14 @@ def increment_and_return_old():
         if self.coord_cartesian is not None:
             self.coord_cartesian.apply_to_fig(fig)
 
+        fig = fig.update_xaxes(title_font_size=18)
+        fig = fig.update_yaxes(title_font_size=18)
+        fig = fig.update_layout(
+            template="simple_white",
+            font_family="Ariel",
```
chrisvittal, line 213: what's the exact syntax here? If a system doesn't have Ariel, what will happen?

chrisvittal (changes requested):
1 Question. Other than that, ready to go.

hail/python/hail/ggplot/ggplot.py
chrisvittal: Okay, just looked it up, the font stack used by default is: `'"Open Sans", verdana, arial, sans-serif'`, this should be good enough for all cases, if you'd like, youcan reorder it to prefer Arial, but we definitely should not be deleting sans-serif as a fallback to the system sans-serif in case only basic system fonts are installed.

johnc1231: Really good catch, will fix.

johnc1231:
Thank you for catching that Chris.

chrisvittal (approved):
You're welcome! Font stacks are difficult.

PR #11391, opened 2022-02-18
[batch] Don't raise 500s for jobs with missing job specs

jigold:
I can try and be clever with how to test this by hand by not writing the spec to cloud storage, but before I do that, I'd like feedback first.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1,3 +1,4 @@
+import abc
 from typing import Optional, Dict, Callable, Tuple, Awaitable, Any, Union, MutableMapping, List, Iterator
 import errno
 import os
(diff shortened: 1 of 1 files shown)


daniel-goldstein (changes requested):
I'm having a hard time understanding the approach here. It feels like a workaround to need a new class instead of being able to just report the job as error'd and drop it. Can you explain why this approach?

jigold:
Working backwards, we need to not return a 500 on error. We could return a BadRequest error code with the message 'invalid spec' and then handle the MJC database call on the driver. I chose instead to have the worker to post job complete so we get the error message with the stack trace showing up in the UI as having the normal job flow seemed cleaner to me last week then special casing `schedule_job` on the driver. `post job complete` needs a job object to get the status to send back to the driver. However, a `Job` has two concrete implementations and we don't know which the bad job is because we can't get the spec. Furthermore, the `Job` class does a lot of work based on the spec right now. So I thought it was clearer to just create a new class that had the status, but nothing else.

After writing this out, it's probably better to have the driver MJC upon error rather than from the worker. The code below would be more complicated. We'd have to get the traceback / error message from the response from the worker.

```python3
        try:
            await client_session.post(
                f'http://{instance.ip_address}:5000/api/v1alpha/batches/jobs/create',
                json=body,
                timeout=aiohttp.ClientTimeout(total=2),
            )
            await instance.mark_healthy()
            await instance.mark_healthy()
            if e.status == 403:
                log.info(f'attempt already exists for job {id} on {instance}, aborting')
            if e.status == 503:
                log.info(f'job {id} cannot be scheduled because {instance} is shutting down, aborting')
            raise e
        except Exception:
            await instance.incr_failed_request_count()
            raise
```

And the error handling would look something like this:

```python3
        try:
            body = await job_config(app, record, attempt_id)
        except Exception:
            log.exception('while making job config')
            status = {
                'version': STATUS_FORMAT_VERSION,
                'worker': None,
                'batch_id': batch_id,
                'job_id': job_id,
                'attempt_id': attempt_id,
                'user': record['user'],
                'state': 'error',
                'error': traceback.format_exc(),
                'container_statuses': {k: None for k in tasks},
            }

```

danking:
I think we shouldn't do this now that we've fixed the system to properly upload specs.

PR #11392, opened 2022-02-18
[batch] Fix monitor instances loop to delete inactive instances

daniel-goldstein:
Currently, an instance that is `inactive` in batch and `Terminated` per the cloud will enter the second branch and we will call `deactivate`, which since the instance is already inactive will be a no-op. We really want the third branch to be executed in which we call delete on the instance, so that the inactive instances don't hang around forever.

Vedant and I paired on this and did some case analysis to restructure the conditions here a little bit. The order of the conditions now doesn't matter and is hopefully more explicit. We also decided that an unspoken case (deleted but not terminated) should be an error scenario, let me know if you think that is an appropriate exception to log here.

### DIFF
--- a/batch/batch/driver/instance_collection/base.py
+++ b/batch/batch/driver/instance_collection/base.py
@@ -15,7 +15,14 @@ from ...instance_config import QuantifiedResource
 from ...batch_configuration import WORKER_MAX_IDLE_TIME_MSECS
 from ..instance import Instance
 from ..location import CloudLocationMonitor
-from ..resource_manager import CloudResourceManager, VMStateCreating, VMStateRunning, VMStateTerminated, VMDoesNotExist
+from ..resource_manager import (
+    CloudResourceManager,
+    VMStateCreating,
+    VMStateRunning,
+    VMStateTerminated,
+    VMDoesNotExist,
+    UnknownVMState,
+)
 
 SIXTY_SECONDS_NS = 60 * 1000 * 1000 * 1000
 CACHE_CAPACITY = 1000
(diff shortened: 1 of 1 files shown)


batch/batch/driver/instance_collection/base.py
```diff
@@ -294,19 +294,19 @@ async def check_on_instance(self, instance: Instance):
...
+            if vm_state.time_since_last_state_change() > 5 * 60 * 1000:
+                log.exception(f'{instance} (state: {str(vm_state)}) has made no progress in last 5m, deleting')
+                await self.call_delete_instance(instance, 'activation_timeout')
+        elif instance.state in ('pending', 'active') and isinstance(vm_state, VMStateTerminated):
             log.info(f'{instance} live but stopping or terminated, deactivating')
             await instance.deactivate('terminated')
         elif instance.state == 'inactive':
             log.info(f'{instance} (vm_state: {vm_state}) is inactive, deleting')
             await self.call_delete_instance(instance, 'inactive')
+        elif instance.state == 'deleted' and not isinstance(vm_state, VMStateTerminated):
```
jigold, line 308: This is great! I'm glad you formalized the cases. Is it worth adding asserts at the end or no? Something like

```python3
if instance.state == 'active':
     assert is instance(vm_state, VMStateRunning)
```

Actually, I'm a little worried about doing that because I think the OMS update state is treated as VMStateCreating right now.

jigold (changes requested):

batch/batch/driver/instance_collection/base.py
daniel-goldstein: I believe this permits active with Creating or Running, but ya idk how the OMS updates play into this situation.

daniel-goldstein: Happy to add asserts for the other cases, there's also the `UnknownVMState` which just gets silently propagated, didn't quite know how to handle that or how much attention to give it. It seems kind of odd to me for `get_vm_state` to return a value in that situation but I'm just not very clued in to this part of the codebase.

jigold: Unknown VM state is literally that we don't know how to interpret the instance state from the cloud provider and it doesn't match one of the known states we know how to interpret

jigold: I think asserts would be good if it's not too onerous. I feel like we've messed up this part of the code base multiple times now and it's always been hard to easily verify it has the correct behavior.

daniel-goldstein: Will add asserts then

jigold (approved):

PR #11395, opened 2022-02-22
Stream buffered Aggregator used in lowering of TableKeyByAndAggregate

ammekk:
Used streamed buffered aggregate to optimize in lowering ofTableKeyByAndAggregate


tpoterba:
went in as a separate PR. Emma should still have GH credit as the author of the base commit!

PR #11396, opened 2022-02-22
[batch] Refactor worker Container class into Image and Container

jigold:
This PR splits the old Container class into an Image class, a Task class, and a Container class. The Image class has all the code necessary for pulling, localizing, and extracting images. There are two public methods `pull` and `prune`. The Container class has all the code for setting up networks, ports, overlays, and running the container. The two important public methods are `run` and `get_logs` which subclasses can overwrite. The Task class is a subclass of the Container class. In the future I will add `JVMContainer` which is also a subclass of a Container.

There is an open question about what the code is doing with `spec['network']` right now. The validator says it's an optional string, but our code allows booleans and doesn't handle the 'public' case.

The last thing to make sure of is the `Timings` class is still correct. I had trouble with getting that right in earlier attempts.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1,7 +1,6 @@
 import asyncio
 import base64
 import concurrent
-import contextlib
 import errno
 import json
 import logging
@@ -557,10 +556,10 @@ def user_error(e):
 class Container:
     def __init__(
         self,
+        worker: 'Worker',
         name: str,
         image: Image,
         scratch_dir: str,
-        pool: concurrent.futures.ThreadPoolExecutor,
         command: List[str],
         cpu_in_mcpu: int,
         memory_in_bytes: int,
@@ -571,6 +570,9 @@ class Container:
         volume_mounts: Optional[List[dict]] = None,
         env: Optional[List[str]] = None,
     ):
+        self.worker = worker
+        assert self.worker
+
         self.name = name
         self.image = image
         self.command = command
@@ -609,8 +611,7 @@ class Container:
         # regarding no-member: https://github.com/PyCQA/pylint/issues/4223
         self.process: Optional[asyncio.subprocess.Process] = None  # pylint: disable=no-member
 
-    @contextlib.asynccontextmanager
-    async def run(self):
+    async def run(self, on_completion: Callable[..., Awaitable[Any]], *args, **kwargs):
         try:
             with self._step('pulling'):
                 await self._run_until_done_or_deleted(self.image.pull)
@@ -653,11 +654,13 @@ class Container:
             self.state = 'error'
             self.error = traceback.format_exc()
         finally:
-            yield
             try:
-                await self.delete_container()
+                await on_completion(*args, **kwargs)
             finally:
-                self.image.release()
+                try:
+                    await self.delete_container()
+                finally:
+                    self.image.release()
 
     async def _run_until_done_or_deleted(self, f: Callable[..., Awaitable[Any]]):
         try:
@@ -972,9 +975,6 @@ class Container:
             network_allocator.free(self.netns)
             self.netns = None
 
-        if self.fs:
-            await self.fs.close()
-
     async def delete(self):
         log.info(f'deleting {self}')
         self.deleted_event.set()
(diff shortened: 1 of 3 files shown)


batch/batch/worker/worker.py
```diff
@@ -1002,8 +1052,56 @@ def container_is_running(self):
...
+                 client_session: httpx.ClientSession,
+                 pool: concurrent.futures.ThreadPoolExecutor):
+        self.job = job
+        self.name = name
+        self.spec = spec
+
+        container_name = f'batch-{self.job.batch_id}-job-{self.job.job_id}-{self.name}'
+        image = Image(spec['image'], job.credentials, client_session, pool)
+
+        super().__init__(
```
danking, line 1080: `super().__init__(` should be the first line of the subclass's `__init__`.

batch/batch/worker/worker.py
```diff
@@ -1002,8 +1052,56 @@ def container_is_running(self):
...
+class Task(Container):
+    def __init__(self,
+                 job: 'DockerJob',
+                 name: str,
+                 spec: dict,
+                 client_session: httpx.ClientSession,
+                 pool: concurrent.futures.ThreadPoolExecutor):
+        self.job = job
+        self.name = name
+        self.spec = spec
```
danking, line 1075: This appears to be unused.

batch/batch/worker/worker.py
```diff
@@ -1012,15 +1110,8 @@ async def upload_log(self):
             await self.get_log(),
```
danking, line 1110: Ah, hmm, uploading the log requires loading the entire log into RAM. As a follow up PR, let's fix this so that we either use the aiotools Copier or we just manually read and write a fixed size block at a time.

batch/batch/worker/worker.py
```diff
@@ -1002,8 +1052,56 @@ def container_is_running(self):
...
+            stream = await self.fs.open(self.log_path)
+            async with stream:
+                return (await stream.read()).decode()
+        return ''
+
+    def __str__(self):
+        return f'container {self.name}'
+
+
+class Task(Container):
```
danking, line 1066: I'm not a fan of this name. We already think of these as containers. Can we maybe just call this `LogUploadingContainer`? And then we don't need to change the name of `copy_container` etc.

danking: Alternatively, this entire class could just be a `Job` method:
```
def run_then_upload_log(self, c: Container):
    try:
        await c._run()
    finally:
        self._step(...):
            # ... copy the log ...
```

danking: Actually, I substantially prefer the method approach. Let's avoid subclasses whenever possible.

danking (changes requested):
Moving the image into a self-contained class is a good innovation. I also noticed a few issues we should fix in follow up PRs. I think we can avoid introducing a Task subclass.

jigold:
Stacked on #11429

danking:
now stacked on https://github.com/hail-is/hail/pull/11482

batch/batch/worker/worker.py
```diff
@@ -1387,6 +1441,19 @@ async def setup_io(self):
...
+            with container._step('uploading_log'):
+                assert self.worker.file_store
+                await self.worker.file_store.write_log_file(
+                    self.format_version,
+                    self.batch_id,
+                    self.job_id,
+                    self.attempt_id,
+                    task_name,
+                    await container.get_log(),
+                )
```
danking, line 1455: I think this might be a bit *too* clever a use of a context manager. What do you think of just adding a parameter to the Container constructor which is an async function `upload_log`?

danking (changes requested):

jigold:
I made the changes, but it might take a bit to get through CI before I know it's all good to go.

jigold:
Ok. Everything's passing. I think this PR is ready for another look. Would be great to get it in the next day or two because I have two PRs that need to be stacked on this.

danking (approved):

jigold:
Putting the WIP tag until I understand why test_ci keeps failing for the JVM containerization branch.

danking:
OK, the changes here seem correct. I'm gonna focus just on #11397

jigold:
Should we close this in favor of #11397?

danking:
lifted WIP so we can get this and the JVM change in.

PR #11397, opened 2022-02-22
[batch] Containerize the JVM

jigold:
Closes https://github.com/hail-is/hail-tasks/issues/3

Stacked on #11396. I haven't tested this code at all.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -2084,7 +2084,7 @@ class JVM:
         root_dir: str,
         client_session: httpx.ClientSession,
         pool: concurrent.futures.ThreadPoolExecutor,
-    ) -> Tuple[JVMContainer, str]:
+    ) -> JVMContainer:
         try:
             container = await JVMContainer.create_and_start(index, socket_file, root_dir, client_session, pool)
 
@@ -2125,10 +2125,9 @@ class JVM:
                 output_file = root_dir + '/output'
                 should_interrupt = asyncio.Event()
                 await blocking_to_async(worker.pool, os.makedirs, root_dir)
-                container, startup_output = await cls.create_container_and_connect(
+                container = await cls.create_container_and_connect(
                     index, socket_file, root_dir, worker.client_session, worker.pool
                 )
-                log.info(f'JVM-{index}: startup output: {startup_output}')
                 return cls(
                     index,
                     socket_file,
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -661,9 +661,6 @@ class Container:
         except asyncio.CancelledError:
             raise
         except Exception as e:
-            if not isinstance(e, ContainerDeletedError) and not user_error(e):
-                log.exception(f'while creating {self}')
-
             if isinstance(e, ImageNotFound):
                 self.short_error = 'image not found'
             elif isinstance(e, ImageCannotBePulled):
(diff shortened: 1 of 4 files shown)


jigold:
My plan is to add the JVM logs to disk for each job feature and debug the entire stack at once. If you see problems with the underlying stack, then it would be great to know that sooner rather than later.

jigold:
I think I need to add the scratch directory for the user where all their secrets are located to the volume mounts.

danking (dismissed):

jigold:
I tested everything by hand in my namespace including the UI and it all looked good.

jigold:
I might still have a race condition when cleaning up the container despite using a lock. I'll see if I can reason about it tomorrow.

danking:
Do I need to take action on this PR? I'm OK with this change if it passes the test.

jigold:
I can't figure out why the tests aren't passing yet. It's failing on test_ci because the code to remove files is failing (KeyError???). Otherwise, I'm good with it, except I might take out the run_if_done_or_deleted for now so we don't get infinite loops while this code is still in flux.

danking:
I tossed this up in my namespace and this is what seems to be the issue:
```

Job Step	Image Pulling Time (s)	Running Time (s)	Error Type	State
main	0.135	30.011	timed out	error
Logs
Main
Log
executable file `sleep 5` not found in $PATH: No such file or directory
Error
  File "/usr/local/lib/python3.7/dist-packages/batch/worker/worker.py", line 680, in _run
    raise ContainerTimeoutError(f'timed out after {self.timeout}s')
ContainerTimeoutError: timed out after Nones

**Comment by danking** [comment]
An error is being treated as a timeout but then the job is still considered running?

**Comment by jigold** [comment]
Do you need any action from me on this? I bet I messed up where I define the container state.

**Comment by danking** [comment]
Still debugging, no action necessary

**Comment by danking** [comment]
@jigold Huzzah it passes!

File: batch/batch/worker/worker.py [hunk 1]
```diff
@@ -1787,59 +1945,113 @@ def scoped_ensure_future(coro_or_future, *, loop=None) -> Iterator[asyncio.Futur
...
+        self.fs = fs
+        self.log_offset = 0
+
+    async def retrieve_and_clear_output(self) -> str:
+        logs = await self.get_log()
+        self.log_offset += len(logs)
+        return logs
+
+    async def get_log(self):
+        return await self.container.get_log(offset=self.log_offset)
```
danking, line 2042: Because you successfully fixed the Hail Query logging problem already, I don't think we need offset at all. The JVMEntryway stdout & stderr are never displayed in the UI. Longer term, we should add a worker status page that shows the JVM logs, but that can show the *entire* log. I think we can completely rip out the offset stuff now.

danking: In fact we can rip out all of the JVM log stuff because there's no way to access it right now.

danking (changes requested):
OK, just a few small changes necessary!

So excited for this to land!

batch/batch/worker/worker.py
```diff
@@ -610,8 +632,16 @@ def __init__(
         self.netns: Optional[NetworkNamespace] = None
         # regarding no-member: https://github.com/PyCQA/pylint/issues/4223
         self.process: Optional[asyncio.subprocess.Process] = None  # pylint: disable=no-member
+        self.pid: Optional[int] = None
```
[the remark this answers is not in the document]
danking: I removed this in my commit, so ignore this comment

batch/batch/worker/worker.py
jigold: It's used when logging error messages for the JVM. Can you double check you don't want this functionality?

jigold: If you don't like the last commit, then I can revert it.

danking (approved):
🐎 💨

PR #11399, opened 2022-02-23
[hail.utils]adding new example to select

CDiaz96:


hail/python/hail/utils/struct.py
```diff
@@ -143,6 +143,14 @@ def select(self, *fields, **kwargs):
         -------
         :class:`.Struct`
             Struct containing specified existing fields and computed fields.
+
+        Examples
+        --------
+
+        Define s
+        >>> s = hl.Struct(**{'foo': 5, '1kg': 10})
```
danking, line 151: On your previous PR I said we should avoid `**` syntax. I still think we should avoid it but, you're using an example of the kind of Struct field name that *must* use the `**` syntax:
```
hl.Struct(**{'foo': 5, '1kg': 10})
```
Python rejects this expression:
```
hl.Struct(foo=5, 1kg=10)
```
```
In [1]: hl.Struct(foo=5, 1kg=10) 
   ...:                                                                                                                                                       
  File "<ipython-input-1-611e92d5a31f>", line 1
    hl.Struct(foo=5, 1kg=10)
                       ^
SyntaxError: invalid syntax
```

This is because python does not allow us to user variable names which start with a digit. `1kg` starts with a `1` so its invalid.

---

Let's change the field name from `1kg` to `bar` or `apple` or w/e. Anything without a leading digit. If you want to have two examples one without `**` and one *with* `**`, then I'm OK with that! I just want the *first* example to be an easy one.

danking (dismissed):

hail/python/hail/utils/struct.py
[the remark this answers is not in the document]
CDiaz96: Added Line

hail/python/hail/utils/struct.py
```diff
@@ -143,6 +143,14 @@ def select(self, *fields, **kwargs):
...
+
+
```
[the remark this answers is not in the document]
CDiaz96: Added the examples

hail/python/hail/utils/struct.py
CDiaz96: Addressed changes

hail/python/hail/utils/struct.py
```diff
@@ -143,6 +143,29 @@ def select(self, *fields, **kwargs):
...
+        Define s
+
+        >>> s = hl.Struct(foo=5, apple=10)
+
+        Prints the original field
+
+        >>> s.select('foo')
+        hl.Struct(foo=5)
+
+        Add one new fields and prints old field
```
danking, line 159: Instead of "prints", let's say "keeps"

hail/python/hail/utils/struct.py
```diff
@@ -143,6 +143,29 @@ def select(self, *fields, **kwargs):
...
             Struct containing specified existing fields and computed fields.
+
+        Examples
+        --------
+
+        Define s
+
+        >>> s = hl.Struct(foo=5, apple=10)
+
+        Prints the original field
```
danking, line 154: Also here, let's say "keep just one original field"

danking (dismissed):
Awesome, I think we just need these two narrative changes and then we're good to merge!

hail/python/hail/utils/struct.py
CDiaz96: Fixed

hail/python/hail/utils/struct.py
CDiaz96: Fixed

hail/python/hail/utils/struct.py
```diff
@@ -143,6 +143,29 @@ def select(self, *fields, **kwargs):
...
+        Define a Struct 's'
+
+        >>> s = hl.Struct(foo=5, apple=10)
+
+        Keep just one original field
+
+        >>> s.select('foo')
+        hl.Struct(foo=5)
+
+        Add one new fields and keeps old field
```
danking, line 159: change "add one new fields" to "Add one new field"

change "keeps old field" to "keeps one old field"

danking (changes requested):

hail/python/hail/utils/struct.py
CDiaz96: Fixed

danking (approved):
Looking great!

danking:
CI is in a bad place right now. Ignore that until tomorrow!

johnc1231:
@CDiaz96 Failing style checks.

PR #7949, opened 2020-01-22
[batch] move spec, status to gcs

jigold:

PR #11404, opened 2022-02-24
[batch] db must rollback if spec write fails or is cancelled

danking:
Closes hail-is/hail-production-issues#9

Details [here](https://github.com/hail-is/hail-production-issues/issues/9#issuecomment-1049356188).

This is my bad. One of the query service PRs allowed spec writing to happen in
parallel with DB insertion (which reduces latency a bit), but if the spec fails
to write or is cancelled, then the DB has a spec token that points at a
cloud object which does not necessarily exist.

I think we do not need the spec token, but removing it does not seem likely
to improve performance much. We still need to hit the DB to get the start_job_id.
There was some discussion about the necessity of the token [here](https://github.com/hail-is/hail/pull/7949#discussion_r370406517).
I think that discussion came to the wrong conclusion. GCS is atomic and globally
consistent. Writing to an already present spec object is atomic. The only issue
I forsee is the possibility that the spec is different the second time. The spec
should have the same semantic content, but if the characters are different the
spec index could be very briefly wrong. We could fix this by storing both the
spec and the index in one GCS file.

cc: @jigold

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1033,7 +1033,7 @@ VALUES (%s, %s, %s);
             # must rollback. See https://github.com/hail-is/hail-production-issues/issues/9
             await asyncio.gather(write_spec_to_cloud(), insert_jobs_into_db(tx))
 
-        write_and_insert()  # pylint: disable=no-value-for-parameter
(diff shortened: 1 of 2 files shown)


daniel-goldstein (approved):

danking:
I added the Jinja2 fix here.

PR #11406, opened 2022-02-24
[batch] ensure cost is up-to-date when a batch is marked complete

danking:
Closes hail-is/hail-production-issues#3

Details [here](https://github.com/hail-is/hail-production-issues/issues/3).

The high-level takeaway is that if the system is moving quite fast MJS
can happen *after* MJC. When MJC happens first, there are not yet any
resources. As a result, the state of a batch is observable as
`complete` *before* the resources are present. A lack of
resources implies an incorrect total cost for a batch. Linking
resource insertion with MJC ensures a batch may be complete if-and-only-if
the cost of the batch is accurate.

### DIFF
--- a/batch/batch/driver/job.py
+++ b/batch/batch/driver/job.py
@@ -130,6 +130,7 @@ async def mark_job_complete(
             raise
         await add_attempt_resources(tx, batch_id, job_id, attempt_id, resources)
         return rv
+
     rv = mjc_and_add_attempt_resources()  # pylint: disable=no-value-for-parameter
 
     scheduler_state_changed.notify()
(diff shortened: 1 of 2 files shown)


batch/batch/driver/job.py
```diff
@@ -201,23 +207,25 @@ async def mark_job_creating(
...
+                (batch_id, job_id, attempt_id, instance.name, start_time),
+            )
+        except Exception:
+            log.info(f'error while marking job {id} creating on {instance}')
+            raise
+        await add_attempt_resources(tx, batch_id, job_id, attempt_id, resources)
+        return rv
+    rv = mjc_and_add_attempt_resources()  # pylint: disable=no-value-for-parameter
 
-    log.info(rv)
```
danking, line 215: I'm not sure why this was here. Neither MJC nor MJS log their return value without any context.

danking:
The diff is more readable if you ignore whitespace changes.

jigold (approved):

PR #11410, opened 2022-02-24
[query] Simplify Elasticsearch Dependencies

johnc1231:
This PR drastically simplifies the hail/elasticsearch story. We no longer require any special custom built JAR from google cloud (was necessary because there was a time when Spark 3 compatible ES wasn't on Maven). Also allow ES version 7 or 8 when on Spark 3, and don't require that it's Spark 3.1

### DIFF
--- a/hail/Makefile
+++ b/hail/Makefile
@@ -66,9 +66,6 @@ GRADLE_ARGS += -Dscala.version=$(SCALA_VERSION) -Dspark.version=$(SPARK_VERSION)
 .PHONY: shadowJar
 shadowJar: $(SHADOW_JAR)
 
-$(JAR_SOURCES): $(ELASTICSEARCH_311_JAR)
-$(JAR_TEST_SOURCES): $(ELASTICSEARCH_311_JAR)
-
 ifdef HAIL_COMPILE_NATIVES
 $(SHADOW_JAR): native-lib-prebuilt
 endif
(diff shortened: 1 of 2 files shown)


hail/Makefile
```diff
@@ -60,17 +60,9 @@ SHADOW_TEST_JAR := build/libs/hail-all-spark-test.jar
...
 
-.PHONY: elasticsearchJar
-elasticsearchJar: $(ELASTICSEARCH_311_JAR)
-
-$(ELASTICSEARCH_311_JAR):
-	@mkdir -p libs
-	curl https://storage.googleapis.com/hail-common/elasticsearch-libs/elasticsearch-spark-30_2.12-8.0.0-SNAPSHOT-custom-hail-spark311.jar >  libs/elasticsearch-spark-30_2.12-8.0.0-SNAPSHOT-custom-hail-spark311.jar
-
 .PHONY: shadowJar
 shadowJar: $(SHADOW_JAR)
```
danking, line 68: The JAR_SOURCES and JAR_TEST_SOURCES reference ELASTICSEARCH_311_JAR

danking (changes requested):

johnc1231:
Building locally confirmed it ran for Spark 3.2 with commons lang fix. Will follow up later with some kind of compilation test against Spark 3.2, but for now let's just do this.

danking (approved):
Nice. I searched for and found no other errant references to the old SNAPSHOT jar.

PR #11411, opened 2022-02-24
[hail.utils]Added examples for Struct.drop

CDiaz96:


hail/python/hail/utils/struct.py
```diff
@@ -166,6 +166,23 @@ def drop(self, *args):
         -------
         :class:`.Struct`
             Struct without certain fields.
+
+        Examples
+        --------
+
+        Define a Struct `s`
+
+         >>> s = hl.Struct(food=8, fruit=5, bar=2, apple=10)
```
jigold, line 175: I think the alignment of the code examples with lines starting with `>>>` and the result shown below need to be aligned with the text above.

danking, line 175: I think you have one extra space before all your code blocks.

hail/python/hail/utils/struct.py
```diff
@@ -166,6 +166,23 @@ def drop(self, *args):
...
+
+        Drop one field from `s`
+
+         >>> s.drop(bar=2)
```
danking, line 179: When we drop, we just drop the name of the field: `s.drop('bar')`.

hail/python/hail/utils/struct.py
```diff
@@ -166,6 +166,23 @@ def drop(self, *args):
...
+         Struct(food=8, fruit=5, apple=10)
+
+        Drop two fields from `s`
+
+         >>> s.annotate(food=8, fruit=5)
```
danking, line 184: looks like this is an annotation instead of a drop!

danking (dismissed):

hail/python/hail/utils/struct.py
jigold, line 179: I think this is not correct. What is the argument type `drop` requires? Remember drop is deleting a field. We don't need to assign a value at the same time.

hail/python/hail/utils/struct.py
jigold, line 184: See comment above and this should be `delete` and not `annotate`.

jigold (dismissed):

hail/python/hail/utils/struct.py
CDiaz96: Fixed

CDiaz96: Switched to `drop`

danking (approved):

PR #11414, opened 2022-02-24
[infra] Remove hard-coded values from config.mk

daniel-goldstein:
I was hoping that a greater restructure would lead to a more elegant solution, but after so many footguns it just felt worth making this change. The main point is shelling out in the Makefiles to ask Kubernetes directly for what the global config values are instead of hard-coding them. Specifically, the changes are:

- `KUBERNETES_SERVER_URL` was simply not used anymore in the Makefiles so I deleted it.
- `DOCKER_ROOT_IMAGE`, `INTERNAL_IP`, `IP` and `CLOUD` were only used in a couple Makefiles so I moved the `kubectl` invocation into where they're used so that Makefiles that don't depend on those variables won't incur the cost of querying them
- `DOCKER_PREFIX` and `DOMAIN` were used pretty widely, so I kept them in config.mk because most Makefiles will need to query those values anyway. The added startup time for `make deploy` feels pretty insignificant.
- With this change in place, there's no need to render config.mk anymore so I deleted the shell function for doing so

### DIFF
--- a/infra/bootstrap_utils.sh
+++ b/infra/bootstrap_utils.sh
@@ -7,29 +7,6 @@ fi
 
 source $HAIL/devbin/functions.sh
 
-render_config_mk() {
-    DOCKER_PREFIX=$(get_global_config_field docker_prefix)
-    INTERNAL_IP=$(get_global_config_field internal_ip)
-    IP=$(get_global_config_field ip)
-    DOMAIN=$(get_global_config_field domain)
-    CLOUD=$(get_global_config_field cloud)
-    cat >$HAIL/config.mk <<EOF
-DOCKER_PREFIX := $DOCKER_PREFIX
-INTERNAL_IP := $INTERNAL_IP
-IP := $IP
-DOMAIN := $DOMAIN
-CLOUD := $CLOUD
-
-ifeq (\$(NAMESPACE),default)
-SCOPE = deploy
-DEPLOY = true
-else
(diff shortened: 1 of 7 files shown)


infra/bootstrap_utils.sh
```diff
@@ -7,29 +7,6 @@ fi
 
 source $HAIL/devbin/functions.sh
 
-render_config_mk() {
-    DOCKER_PREFIX=$(get_global_config_field docker_prefix)
-    INTERNAL_IP=$(get_global_config_field internal_ip)
-    IP=$(get_global_config_field ip)
-    DOMAIN=$(get_global_config_field domain)
-    CLOUD=$(get_global_config_field cloud)
```
jigold, line 15: Are SCOPE and DEPLOY not needed anymore?

jigold (changes requested):

infra/bootstrap_utils.sh
daniel-goldstein: They're still in config.mk, but they don't vary by cloud so there's no reason to re-render the file.

daniel-goldstein: I forgot to include those in the PR message, they haven't changed.

jigold: My bad. I see now.

jigold (approved):

PR #11415, opened 2022-02-24
[auth] Use drop user if exists instead of select and drop

daniel-goldstein:
We use `DROP USER IF EXISTS` in other places so I assume there's no longer a reason to keep this compatibility workaround.


danking:
I love these little tidy-up PRs.

danking:

PR #11423, opened 2022-02-25
[infra] Move azure from mysql single server to mysql flexible server

daniel-goldstein:
This changes the Azure database from Azure database for MySQL Single Server to Azure database for MySQL Flexible Server. The major changes are:

- Fixed several small rots across the bootstrap code
- Altered the database module in terraform to use flexible server. This configuration mostly matches what we had with single server, importantly that it is only accessible on our vnet.
- Makes the client key/certificate in the SQLConfig optional (the current use of SQLConfig is a little repetitious and could probably use a refactor).


auth/auth/driver/driver.py
```diff
@@ -275,14 +279,14 @@ def secret_data(self):
...
-            user=self.name if CLOUD != 'azure' else f'{self.name}@{server_config.instance}',
+            user=self.name,
             password=self.password,
             instance=server_config.instance,
             connection_name=server_config.connection_name,
             db=self.name,
-            ssl_ca='/sql-config/server-ca.pem',
-            ssl_cert='/sql-config/client-cert.pem',
-            ssl_key='/sql-config/client-key.pem',
+            ssl_ca=server_config.ssl_ca,
```
daniel-goldstein, line 287: This might actually be very incorrect if the current config is mounted as database-server-config which it looks like it is.

danking: Can you describe why this change is necessary?

These paths are hardcoded into the MYSQL config file, meaning we can't arbitrary relocate the secret. That was frustrating when I first built this stuff. In the end, I embraced the fixed location of the secret and that seemed to mostly avoid issues.

batch/az-create-worker-image.sh
```diff
@@ -17,7 +17,7 @@ BUILD_IMAGE_RESOURCE_GROUP="${RESOURCE_GROUP}-build-batch-worker-image"
 VM_NAME=build-batch-worker-image
 WORKER_VERSION=0.0.12
 
-USERNAME=$(az ad signed-in-user show --output tsv --query mailNickname)
+USERNAME=$(whoami)
```
danking, line 20: This seems dangerous, it might not be the case that, e.g. the Australians have the same usernames locally as in AZ.

hail/python/hailtop/auth/sql_config.py
```diff
@@ -84,20 +88,21 @@ def from_dict(d: Dict[str, Any]) -> 'SQLConfig':
...
+                                   client_cert: Optional[str],
+                                   client_key: Optional[str]
                                    ) -> Dict[str, str]:
     secret_data = dict()
     secret_data['sql-config.json'] = config.to_json()
     secret_data['sql-config.cnf'] = config.to_cnf()
     secret_data['server-ca.pem'] = server_ca
-    secret_data['client-cert.pem'] = client_cert
-    secret_data['client-key.pem'] = client_key
+    if client_cert is not None and client_key is not None:
```
danking, line 105: maybe an assert here as well

danking (changes requested):

batch/az-create-worker-image.sh
daniel-goldstein: I think I might be misunderstanding exactly how we authenticate. AFAIK now, azure credentials aren't connected to SSH'ing into azure VMs the way that it is in GCP. When creating a VM with the `--generate-ssh-keys` flag (which honestly I'm thinking we maybe shouldn't use), Azure uses your `~/.ssh/id_rsa.pub` public key as a trusted key for the vm, and creates one if it doesn't exist. So the following ssh command should use `~/.ssh/id_rsa` and should work for the current user.

auth/auth/driver/driver.py
daniel-goldstein: I think I was trying to lump in the optionality of the `ssl_cert`/`ssl_key` fields with what I thought was some needless repetition, but I see now that what I replaced these literals with are not necessarily the same. I reverted this but think I do need to have the conditional on whether to make `ssl_cert` and `ssl_key` None.

batch/az-create-worker-image.sh
danking: oh, dang, OK. In that case `whoami` seems exactly right.

danking (approved):
I'm good with this but you're failing a lint.

danking:

PR #11427, opened 2022-02-28
[git] Add isort pr to blame ignore

daniel-goldstein:


danking:

PR #11428, opened 2022-02-28
[batch] Add option to show docker output when building images

daniel-goldstein:
This adds support to optionally view the docker output in `build_python_image`. I'm not sure what our general strategy is for adding inputs to user-facing functions. I tried to maintain backwards compatibility, but it feels very weird and similarly wrong to put a normal parameter after an underscore parameter. I feel like it would be best to have `_tmp_dir` be keyword-only. That would also technically be breaking, but does the underscore prefix relinquish us from that responsibility?

### DIFF
--- a/hail/python/hailtop/batch/docker.py
+++ b/hail/python/hailtop/batch/docker.py
@@ -3,7 +3,7 @@ import sys
 import os
 from typing import Optional, List
 
-from ..utils import secret_alnum_string, sync_check_shell
+from ..utils import secret_alnum_string, sync_check_exec
 
 
 def build_python_image(fullname: str,
@@ -82,15 +82,15 @@ RUN pip install --upgrade --no-cache-dir -r requirements.txt && \
     python3 -m pip check
 ''')
 
-            sync_check_exec('docker', 'build', '-t', fullname, docker_path, capture_output=False)
+            sync_check_exec('docker', 'build', '-t', fullname, docker_path, capture_output=(not show_docker_output))
             print(f'finished building image {fullname}')
         else:
-            sync_check_exec('docker', 'pull', base_image, capture_output=False)
-            sync_check_exec('docker', 'tag', base_image, fullname, capture_output=False)
+            sync_check_exec('docker', 'pull', base_image, capture_output=(not show_docker_output))
+            sync_check_exec('docker', 'tag', base_image, fullname, capture_output=(not show_docker_output))
             print(f'finished pulling image {fullname}')
 
         if '/' in fullname:
-            sync_check_shell('docker', 'push', fullname, capture_output=False)
+            sync_check_shell('docker', 'push', fullname, capture_output=(not show_docker_output))
             print(f'finished pushing image {fullname}')
     finally:
         shutil.rmtree(docker_path, ignore_errors=True)
--- a/hail/python/hailtop/batch/docker.py
+++ b/hail/python/hailtop/batch/docker.py
@@ -10,6 +10,7 @@ def build_python_image(fullname: str,
                        requirements: Optional[List[str]] = None,
                        python_version: Optional[str] = None,
                        _tmp_dir: str = '/tmp',
+                       *,
                        show_docker_output: bool = False) -> str:
     """
     Build a new Python image with dill and the specified pip packages installed.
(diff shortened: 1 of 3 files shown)


hail/python/hailtop/batch/docker.py
```diff
@@ -78,15 +81,15 @@ def build_python_image(fullname: str,
...
         else:
-            sync_check_shell_output(f'docker pull {base_image}')
-            sync_check_shell_output(f'docker tag {base_image} {fullname}')
+            sync_check_shell(f'docker pull {base_image}', inherit_std_out_err=show_docker_output)
+            sync_check_shell(f'docker tag {base_image} {fullname}', inherit_std_out_err=show_docker_output)
             print(f'finished pulling image {fullname}')
 
         if '/' in fullname:
-            sync_check_shell_output(f'docker push {fullname}')
+            sync_check_shell(f'docker push {fullname}', inherit_std_out_err=show_docker_output)
```
danking, line 92: These really should not be check shell, or at least we need to shell quote the paths and full name.

---

That suggests these should really be `sync_check_exec`, which does not yet exist.

I feel like we're building up a parallel process API that diverges from the [`subprocess.run` API](https://docs.python.org/3/library/subprocess.html?highlight=subprocess).

If we need to change our asyncio process API, I think we should prefer starting an asyncio copy of `subprocess.run`. In particular, `inherit_std_out_err` is known in `subprocess.run` as `capture_output=False`.

---

That said, this is synchronous code, so maybe we should just use `subprocess.run`? Unfortunately subprocess.run does not include the stdout and stderr in the CalledProcessError:
```
In [33]: subprocess.run(['/bin/bash', '-c', 'echo hello ; echo >&2 \'ERROR!\' ; false'], capture_output=True, check=True) 
    ...:                                                                                                                                          
  File "<ipython-input-33-3dac52dfa019>", line 1, in <module>
    subprocess.run(['/bin/bash', '-c', 'echo hello ; echo >&2 \'ERROR!\' ; false'], capture_output=True, check=True)
  File "/Users/dking/miniconda3/lib/python3.7/subprocess.py", line 512, in run
    output=stdout, stderr=stderr)
CalledProcessError: Command '['/bin/bash', '-c', "echo hello ; echo >&2 'ERROR!' ; false"]' returned non-zero exit status 1.
```

Unfortunately the printed form of CalledProcessError does not include the stdout and stderr. We might want to write our own shim around `subprocess.run` which just raises a Hail CalledProcessError with includes the stdout and stderr (which are available on subprocess.CalledProcessError as `stdout` and `stderr`).

hail/python/hailtop/batch/docker.py
```diff
@@ -3,13 +3,14 @@
...
 
-from ..utils import secret_alnum_string, sync_check_shell_output
+from ..utils import secret_alnum_string, sync_check_shell
 
 
 def build_python_image(fullname: str,
                        requirements: Optional[List[str]] = None,
                        python_version: Optional[str] = None,
-                       _tmp_dir: str = '/tmp') -> str:
+                       _tmp_dir: str = '/tmp',
```
jigold, line 12: I have no idea why we named the argument to start with an underscore. That seems like it was a mistake on my part when I first put this functionality in. It could have been a bad renaming operation that did more than I intended. 

Not sure what @danking thinks, but I think I'd be okay with renaming the argument to `tmp_dir` or `tmpdir` and placing a `*` after the renamed `tmp_dir`so that all other arguments after this are keyword argument only. I'd also change the type to Optional[str] and set the default to `/tmp` inside the function.

Sorry about that! Also, why not make the show docker output always True and punt on the tmp_dir problem?

jigold (changes requested):

danking (changes requested):

hail/python/hailtop/batch/docker.py
danking: You have a better sense of how folks are using this API than I, if you're comfortable with changing tmp_dir, then I am also comfortable changing it. What are you thoughts on making tmp_dir keyword only? Does that feel a step too far?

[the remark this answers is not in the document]
daniel-goldstein: Is there a list anywhere of "Breaking changes we want to make in 0.3"?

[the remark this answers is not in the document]
johnc1231: Query has one that exists in Asana, I don't know if there's services stuff in there too.

danking:
@daniel-goldstein bump, let's get this resolved and merged.

daniel-goldstein:
@jigold this could use a look

hail/python/hailtop/batch/docker.py
```diff
@@ -78,15 +82,15 @@ def build_python_image(fullname: str,
     python3 -m pip check
 ''')
 
-            sync_check_shell_output(f'docker build -t {fullname} {docker_path}')
+            sync_check_exec('docker', 'build', '-t', fullname, docker_path, capture_output=(not show_docker_output))
```
jigold, line 85: Why is it `(not show_docker_output)`? I would have thought it should be the opposite.

hail/python/hailtop/utils/process.py
```diff
@@ -49,3 +50,13 @@ def sync_check_shell_output(script: str, echo=False) -> Tuple[bytes, bytes]:
 
 def sync_check_shell(script: str, echo=False) -> None:
     sync_check_shell_output(script, echo)
+
+
+def sync_check_exec(command: str, *args: str, echo: bool = False, capture_output: bool = False) -> None:
```
jigold, line 55: What's the reason for passing `command` and `args` separately? It doesn't look like command is used by itself anywhere.

jigold (changes requested):

hail/python/hailtop/batch/docker.py
daniel-goldstein: `capture_output` means to collect the stdout/stderr with a pipe instead of inheriting the standard streams of the calling process. If we want to show the docker output on the user's stdout/err, we want not to capture the output.

daniel-goldstein:
Tested with

```python
hb.build_python_image('my_image', requirements=['hail'], show_docker_output=True)

jigold (approved):

PR #11429, opened 2022-02-28
[batch] Use Timings with no consideration of job deletion

jigold:
I'm not quite sure why there was the deletion logic intertwined with timings. Can you make sure this change is okay or is there something I am missing here?

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -36,6 +36,7 @@ from hailtop.config import DeployConfig
 from hailtop.hail_logging import configure_logging
 from hailtop.utils import (
     CalledProcessError,
+    Timings,
     blocking_to_async,
     check_exec_output,
     check_shell,
@@ -364,39 +365,6 @@ class JobTimeoutError(Exception):
     pass
 
 
-class Timings:
-    def __init__(self, is_deleted: Callable[[], bool]):
-        self.timings: Dict[str, Dict[str, float]] = dict()
-        self.is_deleted = is_deleted
-
-    def step(self, name: str, ignore_job_deletion: bool = False):
-        assert name not in self.timings
-        self.timings[name] = dict()
-        return ContainerStepManager(self.timings[name], self.is_deleted, ignore_job_deletion=ignore_job_deletion)
-
-    def to_dict(self):
-        return self.timings
-
-
-class ContainerStepManager:
-    def __init__(self, timing: Dict[str, float], is_deleted: Callable[[], bool], ignore_job_deletion: bool = False):
-        self.timing: Dict[str, float] = timing
-        self.is_deleted = is_deleted
-        self.ignore_job_deletion = ignore_job_deletion
-
-    def __enter__(self):
-        if self.is_deleted() and not self.ignore_job_deletion:
-            raise JobDeletedError
-        self.timing['start_time'] = time_msecs()
-
-    def __exit__(self, exc_type, exc, tb):
-        if self.is_deleted() and not self.ignore_job_deletion:
-            return
-        finish_time = time_msecs()
-        self.timing['finish_time'] = finish_time
-        self.timing['duration'] = finish_time - self.timing['start_time']
-
-
 def worker_fraction_in_1024ths(cpu_in_mcpu):
     return 1024 * cpu_in_mcpu // (CORES * 1000)
 
@@ -456,7 +424,7 @@ class Container:
         self.started_at: Optional[int] = None
         self.finished_at: Optional[int] = None
 
-        self.timings = Timings(self.is_job_deleted)
+        self.timings = Timings()
 
         self.logbuffer = bytearray()
         self.overlay_path = None
@@ -546,7 +514,7 @@ class Container:
             self.error = traceback.format_exc()
         finally:
             try:
-                with self.step('uploading_log', ignore_job_deletion=True):
+                with self.step('uploading_log'):
                     await self.upload_log()
             finally:
                 try:
@@ -576,8 +544,8 @@ class Container:
     def is_job_deleted(self) -> bool:
         return self.job.deleted
 
-    def step(self, name: str, ignore_job_deletion: bool = False):
-        return self.timings.step(name, ignore_job_deletion=ignore_job_deletion)
+    def step(self, name: str):
+        return self.timings.step(name)
 
     async def pull_image(self):
         is_cloud_image = (CLOUD == 'gcp' and self.image_ref.hosted_in('google')) or (
@@ -1306,7 +1274,7 @@ class DockerJob(Job):
 
         requester_pays_project = job_spec.get('requester_pays_project')
 
-        self.timings = Timings(lambda: False)
+        self.timings = Timings()
 
         if self.secrets:
             for secret in self.secrets:
@@ -1610,7 +1578,7 @@ class JVMJob(Job):
         self.jar_url = self.user_command_string[2]
 
         self.deleted = False
-        self.timings = Timings(lambda: self.deleted)
+        self.timings = Timings()
         self.state = 'pending'
         self.log: Optional[str] = None
 
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1675,22 +1675,20 @@ class JVMJob(Job):
 
     async def cleanup(self):
         if self.jvm is not None:
-            # I really want this to be a timed step but I can't skip this ITS CLEAN UP
-            # with self.step('retrieve_output'):
-            log.info(f'{self}: retrieving log')
-            self.log = self.jvm.retrieve_and_clear_output()
+            with self.step('retrieve_output'):
+                log.info(f'{self}: retrieving output')
+                self.log = self.jvm.retrieve_and_clear_output()
             worker.return_jvm(self.jvm)
             self.jvm = None
 
         job_log = self.log
         if job_log is None:
             job_log = ''
-        # I really want this to be a timed step but I CANT RAISE EXCEPTIONS IN CLEANUP!!
-        # with self.step('uploading_log'):
-        log.info(f'{self}: uploading log')
-        await worker.file_store.write_log_file(
-            self.format_version, self.batch_id, self.job_id, self.attempt_id, 'main', job_log
-        )
+        with self.step('uploading_log'):
+            log.info(f'{self}: uploading log')
+            await worker.file_store.write_log_file(
+                self.format_version, self.batch_id, self.job_id, self.attempt_id, 'main', job_log
+            )
 
         log.info(f'{self}: cleaning up')
         try:
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -344,6 +344,29 @@ def docker_call_retry(timeout, name):
     return wrapper
 
 
+class TaskCancelledError(Exception):
+    pass
+
+
+async def run_until_done_or_cancelled(event: asyncio.Event, f: Callable[[...], Awaitable[Any]], *args, **kwargs):
+    task = asyncio.ensure_future(f(*args, **kwargs))
+    cancelled = asyncio.ensure_future(event.wait())
+    try:
+        await asyncio.wait([cancelled, task], return_when=asyncio.FIRST_COMPLETED)
+        if cancelled.done():
+            raise TaskCancelledError
+        assert task.done()
+        return task.result()
+    finally:
+        for t in (task, cancelled):
+            if t.done():
+                e = t.exception()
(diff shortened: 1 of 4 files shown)


daniel-goldstein:
There's not a good reason, this is just how it was originally designed. Whenever a job was cancelled, it would take until the start of the next step for a container's execution to be stopped. I replaced the dependency on this in `Container` with `run_until_done_or_deleted`, but stopped short of deleting the functionality entirely because there were other parts of the worker, specifically `JVMJob` that still relied on it. Hopefully that is no longer the case after the QoB changes, but @danking  would know better. We've also both lamented about how it's impossible to use timings currently inside cleanup blocks because it could accidentally re-raise a deleted error. This is a great change, I would just take extra care to test job cancellation to make sure there isn't anywhere that's still relying on this functionality.

jigold:
> how it's impossible to use timings currently inside cleanup blocks because it could accidentally re-raise a deleted error

Can you explain this to me? I saw Dan had a comment as well in the JVMJob that stated this. Our code seems to be always this:

```
try:
    with self.step('running'):
        self.run_until_deleted or completed()
finally:
    with self.step('uploading log'):
         self.upload_log()
```

The upload log step should propagate the deletion error from above.

jigold:
Also, I double checked and the only place the deletion logic was used was `upload_logs`.

daniel-goldstein:
I'm looking at `worker.py` now and it looks like you worked around this with the addition to `ignore_job_deletion`, which maybe Dan wasn't aware of but is still a workaround since Timings just shouldn't care about deletion in the first place. Without that flag you'd get this:

1. Running step would start
2. Job is cancelled, so `Job.deleted` would be set to `True`.
3. Job would set the Container's `deleted_event`, which would abort the run function inside `run_until_done_or_deleted`
4. Container would jump to the uploading logs step, which would raise a job deleted error before running `upload_log`.

jigold:
oh I see. Thanks for the explanation! So my current code is not quite right. Let me figure out a fix.

jigold:
Ok. I think this code has the correct behavior now. (4) from above should never depend on the deleted event because we don't handle the deletion inside the regular timings now.

batch/batch/worker/worker.py
```diff
@@ -576,8 +544,8 @@ async def run_until_done_or_deleted(self, f: Callable[[], Awaitable[Any]]):
     def is_job_deleted(self) -> bool:
         return self.job.deleted
```
daniel-goldstein, line 545: This should hopefully be unused now. Can you delete?

batch/batch/worker/worker.py
```diff
@@ -1306,7 +1274,7 @@ def __init__(
 
         requester_pays_project = job_spec.get('requester_pays_project')
 
-        self.timings = Timings(lambda: False)
```
daniel-goldstein, line 1309: Ah great. I was worried that `DockerJob` would depend on its `deleted` field here. The fact that it doesn't makes me more at ease.

batch/batch/worker/worker.py
```diff
@@ -1610,7 +1578,7 @@ def __init__(
         self.jar_url = self.user_command_string[2]
 
         self.deleted = False
-        self.timings = Timings(lambda: self.deleted)
+        self.timings = Timings()
```
daniel-goldstein, line 1581: I think this effectively removes the ability to stop `JVMJob`s mid-execution. With the existing functionality, every `with self.step` in the `run` method is an opportunity to raise a `JobDeletedException` and skip to the cleanup step. As it is now, I don't see a way in which `run` can be interrupted when the job is cancelled. I think that in order to eliminate this, we need to make the `run` method look more like the `Container.run` method, which uses `run_until_done_or_deleted` to allow for deletion mid-run.

daniel-goldstein (changes requested):

batch/batch/worker/worker.py
jigold: I think my new changes actually enable this behavior. I don't think it was there before.

jigold:
The method to my madness for having a `TaskCancelledError` is because in future PRs, it's going to be a ContainerDeletedError. I think we could even move the `run_until_...` function to hailtop.utils.

@danking Can you also take a look? I added interruptability to the JVMJob run tasks.

batch/batch/worker/worker.py
daniel-goldstein: I'm confused, maybe we're not talking about the same thing? I'm saying that stopping `run` between steps worked here when `Timings` still threw a `JobDeletedError`. The more fine-grained approach that you're implementing here is definitely better though, but that didn't exist when I wrote the top comment.

batch/batch/worker/worker.py
```diff
@@ -1741,6 +1726,7 @@ async def get_log(self):
     async def delete(self):
         log.info(f'deleting {self} {self.jvm}')
         self.deleted = True
```
daniel-goldstein, line 1728: `self.deleted` should be removed now that we have the event.

hail/python/hailtop/utils/utils.py
```diff
@@ -982,3 +983,21 @@ def find_spark_home():
...
+
+    @contextlib.contextmanager
+    def step(self, name: str):
+        assert name not in self.timings
+        d: Dict[str, int] = dict()
+        self.timings[name] = d
+        d['start_time'] = time_msecs()
+        yield
+        d['finish_time'] = time_msecs()
+        d['duration'] = d['finish_time'] - d['start_time']
```
daniel-goldstein, line 1000: I like this contextlib implementation, very clean

batch/batch/worker/worker.py
```diff
@@ -343,6 +344,29 @@ async def wrapper(f, *args, **kwargs):
...
+    pass
+
+
+async def run_until_done_or_cancelled(event: asyncio.Event, f: Callable[[...], Awaitable[Any]], *args, **kwargs):
+    task = asyncio.ensure_future(f(*args, **kwargs))
+    cancelled = asyncio.ensure_future(event.wait())
+    try:
+        await asyncio.wait([cancelled, task], return_when=asyncio.FIRST_COMPLETED)
+        if cancelled.done():
+            raise TaskCancelledError
```
daniel-goldstein, line 357: I fear this naming convention is going to make communication more confusing. "Task" and "cancel" are both asyncio concepts, and while this is similar in intention to an `asyncio.CancelledError`, it is not, and it operates on a different level of abstraction (our application code vs library code). I also really hesitate to call anything in our system a Task, to prevent confusion with asyncio Tasks and the discrepancy with the usual use of Task in the literature (from the papers that we've read so far in systems cs I believe Task is generally equivalent to our notion of a Job).

daniel-goldstein: If you want to generalize from deleted, I think stopped is better, but it might be best not to do the abstraction until it is necessary.

daniel-goldstein (changes requested):
I think semantically this is good, but I want to pin down terminology.

batch/batch/worker/worker.py
jigold: Oh I see. I was looking at `__aexit__` and not `__aenter__` where the exception is raised in both.

daniel-goldstein (dismissed):
Awesome

batch/batch/worker/worker.py
```diff
@@ -1220,6 +1189,10 @@ def attempt_id(self):
     def id(self):
         return (self.batch_id, self.job_id)
 
+    @property
+    def deleted(self):
+        return self.deleted_event.is_set()
```
danking, line 1194: very nice

danking (changes requested):
hold, I need to verify the JVM is still correct with these changes

batch/batch/worker/worker.py
```diff
@@ -1656,32 +1656,14 @@ async def run(self):
...
-                                    if err.errno != errno.ENOENT:
-                                        raise
+                    local_jar_location = await self.run_until_done_or_deleted(self.download_jar)
 
                 log.info(f'{self}: running jvm process')
                 with self.step('running'):
-                    await self.jvm.execute(local_jar_location, self.scratch, self.user_command_string)
+                    await self.run_until_done_or_deleted(
+                        self.jvm.execute, local_jar_location, self.scratch, self.user_command_string
+                    )
```
danking, line 1665: cc @daniel-goldstein 

---

Just to recap for clarity how the JVMEntryway works: there's a JVM process. That process shares a socket with the worker. The worker-JVMEnryway protocol follows:
1. Worker sends the number of "arguments". [1]
2. Worker sends that number of "arguments" as strings (strings are encoded as length in bytes then data in utf-8)
3. The worker may now optionally send the number 0 to stop execution immediately.
4. The JVMEntryway sends a message (an integer) indicating why it has stopped executing this request. If the worker sent a 0 (requesting immediate stop to execution) we expect to receive the JVM.FINISH_CANCELLED message. The JVM can also stop executing due to a JVMEntryway exception, a user exception, or normal successful exit. We might also receive End of Stream (EOS) which indicates the JVM has died.

---

I think this change can leave the worker in a bad state. For example, if we cancel while `JVM.execute` is half-way through the protocol with the JVM. Unlike, say, an HTTP request, the JVM won't somehow timeout and let the worker start a new session. It's important that the three required steps above are followed. Moreover, we need to tell the JVM to stop processing a user's job early if we want to re-use that JVM.

---

I think we shouldn't change `JVM.execute` to rely on task cancellation. Task cancellation is a blunt instrument and I think it's really easy for us to not handle it correctly. `JVM.execute` instead is explicit about when it can exit early: while it is waiting for the JVM. It's also easy to see that the protocol is followed correctly. We don't have to reason about control jumping from somewhere in the middle of the method into an `except` or `finally` block.

We should carefully inspect any Task with which we plan to use `run_until_done_or_deleted`. That task needs to be written such that cancellation does not leave the worker in a bad state.

`run_until_done_or_deleted` on `borrow_jvm` seems OK. That will let us cancel a job if JVM initialization is hung.

`run_until_done_or_deleted` on `download_jar` is also, I think, OK. The rename is atomic. The finally will cleanup any temporary resources.

---

[1] The "arguments" are: colon-separated classpath, main class name, scratch directory, and then an arbitrary number of arguments which are passed along to the main class's main method.

danking: Other than this, I think this PR is excellent and I'm really glad that we can actually measure the timing of writing logs and retrieving output.

I had not intended for those comments to make it into the final PR. I had screwed myself up several times where an exception occurred in `retrieve_and_clear_output` and the logs were lost. Thanks for cleaning up my mess 😅

danking (changes requested):

batch/batch/worker/worker.py
danking: Ah, you know what, this must explain why we had issues with bad images. That `localize_rootfs` method really shouldn't be cancelled once it starts extracting the filesystem.

I think `run_container` is fine. I am less certain about aborting a `mount` or modification of the network namespaces. What's the harm in letting those operations run to completion? Thinking back to explicit opportunities to stop early rather than bluntly cancelling tasks, it seems like the slow operations that we might want to interrupt early are:
1. image pulling
2. rootfs extraction (is this actually slow?)
3. waiting on the container to complete

Are there any other long running steps that we should be able to abort?

jigold: > run_until_done_or_deleted on borrow_jvm seems OK. That will let us cancel a job if JVM initialization is hung.

What is the implication of cancelling a coroutine that is awaiting the future `self._jvm_initializer_task`? This means the jvm_initializer_task will be cancelled even if it's part way through. I think we should still have `run_until_done_or_deleted`, but we should do this instead shielding the finish of the _jvm_intializer_task:

```python3
    async def borrow_jvm(self) -> JVM:
        if instance_config.worker_type() not in ('standard', 'D'):
            raise ValueError(f'JVM jobs not allowed on {instance_config.worker_type()}')
        await asyncio.shield(self._jvm_initializer_task)
        assert self._jvms
        return self._jvms.pop()
```

jigold: We have a shield on `localize_rootfs` now so we don't have the problem with bad images anymore.

danking: Ah, yes, good eye. That would've cancelled the JVM initializer task. This is exactly what gives me the heebie-jeebies about `run_until_done_or_deleted`, it's so easy to miss these things. The only benefit it gives for borrow_jvm is that if the JVMs are stuck coming up we can cancel JVM jobs. But if the JVMs are stuck coming up, there's probably something very wrong with that worker.

danking (approved):

danking:

PR #11433, opened 2022-02-28
[batch] Close the db pool after the task manager and worker pool

daniel-goldstein:
If we close the db pool before shutting down the task manager and worker pool, then we get coroutines trying to `acquire` a connection after the connection pool is already closed.


jigold (approved):

danking:

PR #11434, opened 2022-03-01
Always Exclude Spark from Build

johnc1231:
- Keep ES transitive = false
- Explicitly exclude spark from import


danking (approved):

danking:
This change can't possibly be causing the website_image issues. I'm going to force merge this and force merge a release to address the pressing hadoop connector issues. I will address the website_image issue after that is done.

PR #11435, opened 2022-03-01
Support passing an authorization token to the ServiceBackend.

vladsaveliev:
This is useful for launching batches on behalf of other users.

### DIFF
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -375,7 +375,9 @@ class ServiceBackend(Backend[bc.Batch]):
         If specified, the project to use when authenticating with Google
         Storage. Google Storage is used to transfer serialized values between
         this computer and the cloud machines that execute Python jobs.
-
+    token:
+        The authorization token to pass to the batch client.
+        Should only be set for user delegation purposes.
     """
 
     def __init__(self,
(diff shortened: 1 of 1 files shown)


vladsaveliev:
#assign services

danking:
@jigold this seems roughly fine to me, though we should make the token argument a non-underscore one on BatchClient. They should probably all be kwargs too. I'm not sure how public we consider the aioclient.

jigold (approved):

jigold (changes requested):
See comment.

vladsaveliev:
Thanks a lot for looking into the PR!

Made `Optional[str]`, and leaving the parameter with the underscore for this PR.

danking:
@vladsaveliev I dismissed Jackie's review. That will make it show up in her reviewing queue in our CI system. In the future feel free to dismiss reviews if you've addressed the concerns!

jigold (approved):

vladsaveliev:
Thanks you :)

Could you also make a release soon after this PR is merged, if reasonable? Would be handy to be able to try the package in our setup!

danking:
Sure

vladsaveliev:
@jigold and @danking, I can't see why `ci-test` is failing, unfortunately. Can you take a look?

(and I should look into setting up Hail CI in our infra as well :-))

danking:
@vladsaveliev , we've had a bit of CI instability here which is causing tests to fail. Looks like it's merged now, I'll do a release.

danking:
https://github.com/hail-is/hail/pull/11487

vladsaveliev:
Thank you!!

PR #11437, opened 2022-03-01
[infra] Install skopeo on bootstrap vm

daniel-goldstein:
Allows `skopeo` to be used instead of docker when copying third-party images, which should substantially speed up the bootstrap process.


danking (changes requested):
sorry to hold this up, but I want to verify: this is strictly a performance benefit for copy_images.sh, right?

danking (approved):

PR #11438, opened 2022-03-01
[hailtop]grouped type annotation

CDiaz96:


hail/python/hailtop/utils/utils.py
```diff
@@ -107,7 +107,7 @@ def digits_needed(i: int) -> int:
     return 1 + digits_needed(i // 10)
 
 
-def grouped(n, ls):
+def grouped(n:int, ls:list[Any]) -> list:
```
jigold, line 110: A couple of things

1. Type annotations have an int after the colon
2. Types are all lowercase for things like int, str, float, bytes, but are in upper case (imported from Typing above) for things like Any, List, Dict, Tuple, Iterable, etc.
3. Think about what `grouped` is returning. Take a look at your test cases you wrote a week or two ago.

jigold (changes requested):

danking:
Also for names fo PR's we like to use `[component] blah blah`. So for this it would be `[hailtop] added type annotation`

hail/python/hailtop/utils/utils.py
```diff
@@ -107,7 +108,7 @@ def digits_needed(i: int) -> int:
     return 1 + digits_needed(i // 10)
 
 
-def grouped(n: int, ls: List[Any]) -> List:
+def grouped(n: int, ls: List[Any]) -> Generator[Any]:
```
jigold, line 111: We want the type to be as specific as we know. What are we generating? i.e. what is the type of `group` in the code below?

jigold (changes requested):

hail/python/hailtop/utils/utils.py
CDiaz96: We are generating a list that could have either integers or strings.

hail/python/hailtop/utils/utils.py
```diff
@@ -1,4 +1,5 @@
-from typing import Callable, TypeVar, Awaitable, Optional, Type, List, Dict, Iterable, Tuple
+from email.generator import Generator
+from typing import Any, Callable, TypeVar, Awaitable, Optional, Type, List, Dict, Iterable, Tuple
```
jigold, line 2: The first import is not needed. We want the Generator type from `typing`. You can just add it in Alphabetical order to the line below.

hail/python/hailtop/utils/utils.py
jigold: Correct, so what should `Any` be in this case then?

jigold (changes requested):

CDiaz96:
Looking through the google doc I read Dan mentioning Iterators as returns types and since this is a list we are iterating through, I thought I should test it to see if the return type for Iterable would work and it did pass.

hail/python/hailtop/utils/utils.py
```diff
@@ -107,7 +107,7 @@ def digits_needed(i: int) -> int:
     return 1 + digits_needed(i // 10)
 
 
-def grouped(n, ls):
+def grouped(n: int, ls: List[Any]) -> Iterable:
```
danking, line 110: Let's go with `Iterable[Any]`. `Iterable` is just like `List`: it's an iterable *of something*.

danking (changes requested):

hail/python/hailtop/utils/utils.py
CDiaz96: Okay. Yes, I went ahead and added it.

danking (approved):
nice

PR #11441, opened 2022-03-01
[build.yaml] steps using database-server-config must depend on its creation

danking:
I'm honestly not sure how we managed to go so long without this fix.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -466,6 +466,7 @@ steps:
       - create_deploy_config
       - deploy_auth_driver_service_account
       - create_test_gsa_keys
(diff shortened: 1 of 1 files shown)


jigold:
What is `create_database_server_config` supposed to be doing? It looks like it's just a secret from a `test_database_instance`, but I don't know what `test_database_instance` is supposed to be used for.

danking:
In PR namespaces, it created the database-server-config which is the root user which we use to set up everyone else. test_database_instance is cooked up by the createDatabase for the test database.

jigold (approved):

PR #11443, opened 2022-03-01
[dev-docs] kubernetes operations: altering a node pool

danking:
cc: @jigold

I am honestly not sure how this should interact with terraform. We do specify the
node pools there. I worry that using terraform to change the node pools would create
downtime.

### DIFF
--- a/dev-docs/kubernetes-operations.md
+++ b/dev-docs/kubernetes-operations.md
@@ -1,0 +1,49 @@
+# Kubernetes Operations
+
+## Altering a Node Pool
+
+We will have the old node pool and the new node pool active simultaneously. We will use `cordon` and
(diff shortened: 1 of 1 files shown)


daniel-goldstein (approved):
This is a good thing to consider when we circle back to running hail-vdc under terraform. I haven't looked into it much but I think we would add and remove the nodepools with terraform but then we could always do the rest of the steps as we describe here. Terraform also gives you the option to ignore if certain features of the resource change outside of terraform (e.g. if a pool no longer has the number of nodes specified in terraform due to autoscaling, don't suddenly force the number of nodes to change).

lgruen:
At least on GCP, the GKE Autopilot offering might also be worth a look? One advantage would be that you're only charged for the pods you use, not the nodes (https://cloud.google.com/kubernetes-engine/pricing).

PR #11445, opened 2022-03-02
[query/service] convert most spark-only tests to fails_service & fails_local tests

danking:


danking:
I think I might have flipped all the remaining spark-only ones to fails and removed the ones that are passing. CI is backed up a bit so it hasn't been run in a while.

johnc1231:
Looks like: 

test_combiner_manual_filtration is failing in an obvious way, not ready for ServiceBackend

test_plot_roc_curve is failing in a non-obvious way, not sure what's going on.

danking:
It's a vaguely confusing message, but the part in the brackets indicates that we expected this to fail but instead it passed. The part after the brackets is the default message associated with the `fails_service_backend` annotation
```
[XPASS(strict)] doesn't yet work on service backend
```

danking:
bump a bump bump

danking:
This does not substantially change test time, so let's approve and merge this.

johnc1231 (approved):

PR #11420, opened 2022-02-25
[batch] Mitigate accrued_costs test failure

jigold:
Mitigation for hail-is/hail-production-issues#3

PR #11448, opened 2022-03-02
[batch] mitigate some of the cost test failures

danking:
These are failing a bunch of PRs which is impairing our ability
to get features merged. Hopefully these delays give the DB enough
time to process the add_attempt_resources.

We can properly fix this in Q2.

### DIFF
--- a/batch/test/test_accounts.py
+++ b/batch/test/test_accounts.py
@@ -1,5 +1,4 @@
 import os
-import time
 import secrets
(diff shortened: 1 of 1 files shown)


jigold:
I had a mitigation already. Which do you prefer? https://github.com/hail-is/hail/pull/11420

danking:
@jigold I removed the duplicated mitigation, the second part is also necessary, AFAIK.

jigold (approved):

PR #11449, opened 2022-03-02
[query] reduce mean test time

danking:
More tests => lower mean => more even distribution of time across jobs => lower mean test job time

### DIFF
--- a/hail/python/test/hail/methods/test_pca.py
+++ b/hail/python/test/hail/methods/test_pca.py
@@ -1,4 +1,5 @@
 import math
+import pytest
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


hail/python/test/hail/methods/test_pca.py
```diff
@@ -200,28 +198,33 @@ def spec5(j, k):
 
 
 @fails_service_backend(reason='persist_ir')
-def test_spectra_1():
-    spectra_helper(spec1)
+@pytest.mark.parameterize("k,m,n", dim_triplets)
+def test_spectra_1(k, m, n):
+    spectra_helper(spec1, k, m, n)
```
patrick-schultz, line 203: Can't we combine these all into one test, with the spectrum another parameter?

patrick-schultz, line 203: And the spectral_moments tests below could get the same treatment.

patrick-schultz (changes requested):

patrick-schultz (approved):

danking:
I will revisit this next week

PR #11453, opened 2022-03-02
[batch] load credentials from file once instead of thrice

danking:
I'm also not sure how the Azure compute client ever worked, it never received the credentials_file argument.


jigold (approved):

danking:
I'll revisit when I return

PR #11472, opened 2022-03-03
[gear] Add metric for number of queries waiting on a connection

daniel-goldstein:
Added a panel for it [here](https://grafana.hail.is/d/TVytxvb7z/batch-driver-performance?orgId=1&var-namespace=dgoldste&from=1646327460490&to=1646328244432). I was kind of surprised to see that we don't have a queue building up, but the back-pressure could be somewhere else. I can also bring up the rate limit now that the DB can handle it

### DIFF
--- a/gear/gear/database.py
+++ b/gear/gear/database.py
@@ -9,7 +9,7 @@ from typing import Optional
 import aiomysql
 import pymysql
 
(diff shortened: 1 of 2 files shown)


danking (approved):
Nice.

PR #11473, opened 2022-03-03
added partition type annotation

CDiaz96:


hail/python/hailtop/utils/utils.py
```diff
@@ -116,7 +116,7 @@ def grouped(n: int, ls: List[Any]) -> Iterable[Any]:
         yield group
 
 
-def partition(k, ls):
+def partition(k: int, ls: List[int]) -> Iterable[Any]:
```
jigold, line 119: What is the type of the object we are yielding on line 134? Replace `Iterable[Any]` with `Iterable[.......]` where `......` is the type of the object on line 134 being yielded.

hail/python/hailtop/utils/utils.py
```diff
@@ -116,7 +116,7 @@ def grouped(n: int, ls: List[Any]) -> Iterable[Any]:
         yield group
 
 
-def partition(k: int, ls: List[int]) -> Iterable[Any]:
+def partition(k: int, ls: List[int]) -> Iterable[List]:
```
jigold, line 119: What is the type of the elements of the List? Take a look at the type of `ls` for a hint.

jigold (dismissed):

hail/python/hailtop/utils/utils.py
```diff
@@ -116,7 +116,7 @@ def grouped(n: int, ls: List[Any]) -> Iterable[Any]:
         yield group
 
 
-def partition(k, ls):
+def partition(k: int, ls: List[int]) -> Iterable[int]:
```
jigold, line 119: `ls` doesn't have to contain integers only. It can be any kind of value. So for `ls` and the return type (but not `k`) replace `int` with `Any`.

jigold (changes requested):

hail/python/hailtop/utils/utils.py
```diff
@@ -116,7 +116,7 @@ def grouped(n: int, ls: List[Any]) -> Iterable[Any]:
         yield group
 
 
-def partition(k, ls):
+def partition(k: int, ls: List[Any]) -> Iterable[Any]:
```
jigold, line 119: The return type should be `Iterable[List[Any]]`. Can you tell me why?

jigold (changes requested):

hail/python/hailtop/utils/utils.py
CDiaz96: It's because `partition` returns a list with either `int` or `str`.

jigold (approved):

CDiaz96:
@danking

CDiaz96:
Changed function in another branch with more current updates, so I closed this one.

PR #11479, opened 2022-03-03
[web_common] search bar should ignore all header rows

danking:


danking:
I went a bit overboard and switched things over to for/of instead of explicit iteration.

daniel-goldstein (approved):

PR #11509, opened 2022-03-05
[batch] use a filesystem that still exists

danking:
A front_end may request a log as long as the job state is "Running". Ergo, we
must have a valid FS until after the job complete HTTP request to
the driver completes successfully. Seems easiest to just use the
worker's file system which can read local as well as remote files.


jigold:
Closing in favor of #11512

PR #11529, opened 2022-03-08
[hailtop.utils]unzip annotation type

CDiaz96:


hail/python/hailtop/utils/utils.py
```diff
@@ -136,7 +136,7 @@ def generator():
     return generator()
 
 
-def unzip(lst):
+def unzip(lst: List[Any]) -> Tuple[Any]:
```
jigold, line 139: So the way Tuple types work is if they return a tuple of two integers, then the type is `Tuple[int, int]`. If it returns a tuple of an integer and a string, then it would be `Tuple[int, str]`. So what should `Tuple[Any]` be? Take a look at line 145. What is the type of `a`? What is the type of `b`?

Hint: `lst` should have the type `Iterable[Tuple[Any, Any]]`. You can see this is true because of line 142. `for x, y in lst:`

CDiaz96: Should the return type then be `Tuple[any, Any] since it returns a tuple that can be either integers or string.

jigold: What is the type of `a`? What is the type of `b`? You're returning a tuple that consists of the values `a` and `b`. Once you know the types of `a` and `b` then you can fill in what the return type should be.

CDiaz96: `a` and `b` are both list that contain either Strings or Integers

jigold: Great. So `Any` is shorthand for any possible type. Since the lists `a` and `b` can contain any element type (strings, integers, floats, booleans, and even more complex types like lists, tuples, dictionaries), then we use the type `Any` to allow any possible type.

This is why the type of `lst` is `Iterable[Any]`.

hail/python/hailtop/utils/utils.py
```diff
@@ -136,7 +136,7 @@ def generator():
     return generator()
 
 
-def unzip(lst: List[Any]) -> Tuple[Any]:
+def unzip(lst: Iterable[Tuple[Any, Any]]) -> Tuple[List(Any),List(Any)]:
```
jigold, line 139: Almost perfect except this `Tuple[List(Any),List(Any)]` should be this `Tuple[List[Any], List[Any]]`

jigold (changes requested):

hail/python/hailtop/utils/utils.py
```diff
@@ -136,7 +136,7 @@ def generator():
     return generator()
 
 
-def unzip(lst):
+def unzip(lst: Iterable[Tuple[Any, Any]]) -> Tuple[List[Any], List[Any]]:
```
jigold, line 139: This is great except you have extra changes with `partition`

jigold (changes requested):

jigold (approved):

jigold (changes requested):
This looks like it was reverted.

CDiaz96:
There was merging issues and that is why I rebased it, but the last commit is the most recent one.

jigold:
Right, but you got rid of the changes that we wanted. You've replaced T, U with Any, Any.

CDiaz96:
Oh you're confusing this pull request with the one I created today that already merged.

jigold:
Then is this PR still needed or can you close it?

CDiaz96:
I mean I can close it since we re-updated the same function. It had just been a previous task of the day.

CDiaz96:
Should I delete the branch as well?

jigold:
Up to you.

PR #11530, opened 2022-03-08
[ci] Use Database interface instead of connection pool directly

daniel-goldstein:


daniel-goldstein:
After the first commit, lots of unrelated type errors popped up from mypy. I think this is because I have the new mypy installed and it's actually catching more errors that were there all along. It might help to see the errors that it gave me (also visible in old CI builds of the PR:

```
ci/github.py:508: error: Incompatible types in assignment (expression has type "bool", variable has type "Optional[str]")
ci/github.py:551: error: Incompatible types in assignment (expression has type "bool", variable has type "Optional[str]")
ci/github.py:554: error: Incompatible types in assignment (expression has type "bool", variable has type "Optional[str]")
ci/github.py:574: error: Unsupported operand types for > ("int" and "None")
ci/github.py:574: note: Left operand is of type "Optional[int]"
ci/github.py:575: error: Unsupported operand types for + ("None" and "int")
ci/github.py:575: note: Left operand is of type "Optional[int]"
ci/github.py:817: error: Item "None" of "Optional[Dict[str, PR]]" has no attribute "values"
ci/github.py:828: error: Item "None" of "Optional[Dict[str, PR]]" has no attribute "values"
ci/github.py:840: error: Item "None" of "Optional[Dict[str, PR]]" has no attribute "values"
ci/github.py:842: error: Item "None" of "Optional[Dict[str, PR]]" has no attribute "values"
ci/github.py:849: error: Item "MergeFailureBatch" of "Union[Batch, Any, MergeFailureBatch]" has no attribute "id"
ci/github.py:849: error: Item "None" of "Optional[Dict[str, PR]]" has no attribute "values"
Found 11 errors in 1 file (checked 19 source files)
```

It might be helpful to look at the first commit and last commit in isolation. Or if you'd like I can make a separate PR.

ci/ci/ci.py
```diff
@@ -368,7 +364,7 @@ async def push_callback(event):
         branch_name = ref[len('refs/heads/') :]
         branch = FQBranch(Repo.from_gh_json(data['repository']), branch_name)
         for wb in watched_branches:
-            if wb.branch == branch or any(pr.branch == branch for pr in wb.prs.values()):
```
danking, line 371: huh, I don't understand how this didn't cause attribute errors.

ci/ci/github.py
```diff
@@ -628,7 +623,7 @@ def __init__(self, index, branch, deployable, mergeable):
         self.deployable: bool = deployable
         self.mergeable: bool = mergeable
 
-        self.prs: Optional[Dict[str, PR]] = None
+        self.prs: Dict[str, PR] = {}
```
danking, line 626: huh this was silly.

ci/ci/github.py
```diff
@@ -640,7 +635,7 @@ def __init__(self, index, branch, deployable, mergeable):
         self.batch_changed: bool = True
         self.state_changed: bool = True
 
-        self.n_running_batches: Optional[int] = None
+        self.n_running_batches: int = 0
```
danking, line 638: I think the idea was that None meant we had no information yet. For example, we maybe shouldn't run any PRs until we know how many running batches we have? I'm, OK with this change too if its simpler.

danking (approved):

PR #11552, opened 2022-03-09
[batch] fix flaky test

danking:
Saw this in https://ci.azure.hail.is/batches/38760/jobs/99.

I think the log of a container is actually never None any more. It can
be empty if bash has started but the echo command has not yet run.

### DIFF
--- a/batch/test/test_batch.py
+++ b/batch/test/test_batch.py
@@ -57,7 +57,7 @@ def test_job_running_logs(client: BatchClient):
         status = j.status()
         if status['state'] == 'Running':
             log = j.log()
-            if log is not None:
+            if log is not None and log != '':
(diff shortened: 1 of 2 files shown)


batch/test/test_batch.py
```diff
@@ -57,7 +57,7 @@ def test_job_running_logs(client: BatchClient):
         status = j.status()
         if status['state'] == 'Running':
             log = j.log()
-            if log is not None:
+            if log is not None and log != '':
```
daniel-goldstein, line 60: Shouldn't this be `log['main'] != ''`?

daniel-goldstein: I can push a fix after lunch

PR #11553, opened 2022-03-09
[hailtop.utils]cost_str type annotation

CDiaz96:


hail/python/hailtop/utils/utils.py
```diff
@@ -75,7 +75,7 @@ def first_extant_file(*files: Optional[str]) -> Optional[str]:
     return None
 
 
-def cost_str(cost):
+def cost_str(cost: Optional[int]) -> Optional[int]:
```
jigold, line 78: What should the return type be? Hint: look at the name of the function and what is being returned on line 81.

jigold (changes requested):

jigold (approved):

PR #11557, opened 2022-03-10
[batch] Time mark job creating and close batch

daniel-goldstein:
We time most of our stored SQL procedures (anything with a `CALL`). I'd like to time these two as well, the `mark_job_creating` should hopefully inform the deadlock situation.


daniel-goldstein:
@vrautela bump

vrautela (approved):
🚢

PR #11558, opened 2022-03-10
add grouped type variable

CDiaz96:


hail/python/hailtop/utils/utils.py
```diff
@@ -107,8 +107,8 @@ def digits_needed(i: int) -> int:
         return 1
     return 1 + digits_needed(i // 10)
 
-
-def grouped(n: int, ls: List[Any]) -> Iterable[Any]:
+T =TypeVar('T')
+def grouped(n: int, ls: List[T]) -> Iterable[T]:
```
jigold, line 111: The type variable definition already exists on line 48. You don't need to put it here.

What is the return type of grouped? Such as on line 115?
Going back to your examples:

`list(grouped(3, [1, 2, 3, 4, 5, 6, 7, 8, 9]))` will return `[[1, 2, 3], [4, 5, 6], [7, 8, 9]]`.

jigold (dismissed):

hail/python/hailtop/utils/utils.py
jigold: And just to be clear, this is a generator function with `yield`. So the outer part of the return type should still be `Iterable`

CDiaz96: It returns a list of list that contain either strings or numbers, that you can iterate through. Oh so should the return Type be `Iterable[List[T]]`

jigold: Yes.

jigold (approved):

PR #11559, opened 2022-03-11
[batch] Add `/version` endpoint (originally in [query])

vladsaveliev:
The endpoint was originally added into the Query service in https://github.com/hail-is/hail/pull/10085, but it disappeared after the Query service was removed. Re-adding it into Batch.

Let me know if it better belongs somewhere else. Our use case was to request the version of currently running Batch service, so adding it into Batch made sense.)


jigold:
@danking @daniel-goldstein Do we want this endpoint to be public? I think we should add the authorized users only decorator.

PR #11561, opened 2022-03-11
[hailtop.utils]added precise annotation type

CDiaz96:


hail/python/hailtop/utils/utils.py
```diff
@@ -137,7 +137,7 @@ def generator():
     return generator()
 
 
-def unzip(lst):
+def unzip(lst: Iterable[Tuple[T]]) -> Tuple[List[T], List[T]]:
```
jigold, line 140: This is slightly tricky. The Tuple type takes `n` types where `n` is the number of elements in the tuple. This function expects lists with two elements, but they don't necessarily have to be the same.

On line 48, can you add this `U = TypeVar('U') # pylint: disable=invalid-name` so we have a second generic type.

Then you can have this be the input type for `lst`: `Iterable[Tuple[T, U]]`. How does this change the return type?

jigold (dismissed):

hail/python/hailtop/utils/utils.py
CDiaz96: So for the return type were going to have two generic types. So it'll look more like this `Tuple[List[T], List[U]]`?

jigold: yes

PR #11564, opened 2022-03-13
[query] explicitly thread HailFeatureFlags through ExecuteContext

danking:
NB: I don't verify that the flags are valid flags. I created [a ticket for that](https://github.com/hail-is/hail-tasks/issues/16) and will return to it later. I don't think we need it for march 28 because only developers should use flags.


hail/python/hail/backend/py4j_backend.py
```diff
@@ -38,6 +39,7 @@ def deco(*args, **kwargs):
 
 
 class Py4JBackend(Backend):
+    _jbackend: py4j.java_gateway.JavaObject
```
danking, line 42: This satisfies mypy.

hail/src/main/scala/is/hail/expr/ir/Pretty.scala
```diff
@@ -14,8 +15,8 @@ import org.json4s.jackson.{JsonMethods, Serialization}
 import scala.collection.mutable
 
 object Pretty {
-  def apply(ir: BaseIR, width: Int = 100, ribbonWidth: Int = 50, elideLiterals: Boolean = true, maxLen: Int = -1, allowUnboundRefs: Boolean = false): String = {
-    val useSSA = HailContext.getFlag("use_ssa_logs") != null
+  def apply(ctx: ExecuteContext, ir: BaseIR, width: Int = 100, ribbonWidth: Int = 50, elideLiterals: Boolean = true, maxLen: Int = -1, allowUnboundRefs: Boolean = false): String = {
+    val useSSA = ctx != null && ctx.getFlag("use_ssa_logs") != null
```
danking, line 19: There's one place where we use `null`: `MatrixBGENReaderParameters.toJValue`. You should interrogate that use. Maybe toJValue should be forced to take an ExecuteContext?

hail/src/main/scala/is/hail/lir/X.scala
```diff
@@ -65,8 +65,6 @@ class Classx[C](val name: String, val superName: String, var sourceFile: Option[
       SimplifyControl(m)
     }
 
-    val writeIRs = HailContext.isInitialized && HailContext.getFlag("write_ir_files") != null
```
danking, line 68: This is  where the HailContext leaked in asm4s.

hail/src/main/scala/is/hail/io/bgen/LoadBgen.scala
```diff
@@ -421,7 +421,11 @@ case class MatrixBGENReaderParameters(
       }.toList),
       "nPartitions" -> nPartitions.map(JInt(_)).getOrElse(JNull),
       "blockSizeInMB" -> blockSizeInMB.map(JInt(_)).getOrElse(JNull),
-      "includedVariants" -> includedVariants.map(t => JString(Pretty(t))).getOrElse(JNull)))
+      // FIXME: feels like a hack that Pretty needs execute context
+      // FIXME: feels like a hack that I use null here
+      // FIXME: feels like a hack that toJValue uses Pretty?
+      // Q: can we parse SSA'ed pretty table IR?
+      "includedVariants" -> includedVariants.map(t => JString(Pretty(null, t))).getOrElse(JNull)))
```
danking, line 428: This is my biggest concern with this PR. I'm not sure if there should be an execute context available. It also seems not great that serialization of a MatrixBGENReaderParameters depends on hail feature flags?

hail/src/test/scala/is/hail/HailSuite.scala
```diff
@@ -83,4 +92,205 @@ class HailSuite extends TestNGSuite {
       hc.sparkBackend("HailSuite.withExecuteContext").withExecuteContext(timer)(f)
     }
   }
+
+  def assertEvalsTo(
```
danking, line 96: I moved several methods from TestUtils into HailSuite so that the execute context would be available.

hail/src/test/scala/is/hail/asm4s/ASM4SSuite.scala
```diff
@@ -13,37 +14,37 @@ import scala.language.postfixOps
 
 trait Z2Z { def apply(z:Boolean): Boolean }
 
-class ASM4SSuite extends TestNGSuite {
+class ASM4SSuite extends HailSuite {
```
danking, line 17: This feels not great, but I wanted a single source of truth for `shouldWriteIRFiles`

hail/src/main/scala/is/hail/backend/ExecuteContext.scala
```diff
@@ -113,6 +133,14 @@ class ExecuteContext(
...
   }
 
+  def getFlag(name: String): String = flags.get(name)
+
+  def shouldWriteIRFiles(): Boolean = getFlag("write_ir_files") != null
+
+  def shouldNotLogIR(): Boolean = flags.get("no_ir_logging") != null
+
+  def shouldLogIR(): Boolean = !shouldNotLogIR()
+
```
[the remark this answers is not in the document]
tpoterba: this is great.

hail/src/main/scala/is/hail/backend/local/LocalBackend.scala
```diff
@@ -62,12 +62,19 @@ class LocalBackend(
       + "is.hail.io.compress.BGzipCodecTbi,"
       + "org.apache.hadoop.io.compress.GzipCodec")
 
+  private[this] val flags = HailFeatureFlags.fromEnv()
   private[this] val theHailClassLoader = new HailClassLoader(getClass().getClassLoader())
 
+  def getFlag(name: String): String = flags.get(name)
```
tpoterba, line 68: might discover this as I keep reading, but why do we need this if ExecuteContext has it?

hail/src/main/scala/is/hail/expr/ir/Pretty.scala
tpoterba: I've always thought that usage was super weird -- a TableIR inside of a table reader parameters (and not treated as a child or anything). Passing null there is fine; we're going to fix it up at some point anyway. I think fixing this as part of the new functionality (TableReadWithFilter) intended to be added for fast web portal backends is the right strategy (this is there to support reading a BGEN with filter, the same functionality!)

hail/src/main/scala/is/hail/lir/X.scala
tpoterba: fix looks good.

tpoterba (dismissed):
point of discussion around backends exposing flags. Otherwise great.

hail/src/main/scala/is/hail/backend/local/LocalBackend.scala
danking: When in use, the LocalBackend and SparkBackend are the source of truth for flags. I use getFlag/setFlag on the backend to implement `hl._set_flag` and `hl._get_flag`. Put another way: the Python client might want to know a flag value when there is no execution occurring.

tpoterba (approved):

danking:
Rebased and fixed.

PR #11566, opened 2022-03-14
added precise type annotation to partition

CDiaz96:


hail/python/hailtop/utils/utils.py
```diff
@@ -1,4 +1,4 @@
-from typing import Callable, TypeVar, Awaitable, Optional, Type, List, Dict, Iterable, Tuple
+from typing import Any, Callable, TypeVar, Awaitable, Optional, Type, List, Dict, Iterable, Tuple
```
danking, line 1: Now that we're using the type variable `T`, we don't need to import `Any`. Let's remove it here.

CDiaz96: Okay perfect, yes now we don't use it.

danking (approved):

PR #11567, opened 2022-03-14
[auth] Remove redirects to notebook

jigold:
Closes https://github.com/hail-is/hail-tasks/issues/7

I tested this in my namespace.


danking (approved):
💯

PR #11568, opened 2022-03-14
[deploy] fix sed for updating HAIL_VERSION

danking:
I could not get the `\d` to work for me on an Ubuntu machine. The `[0-9]`
range seems to work with extended regexps.


chrisvittal (approved):
should have caught this, oops.  thanks!

PR #11581, opened 2022-03-14
[query] Use RFC3339/ISO8601 for Py4jUtils.stat data

chrisvittal:
This is a sane choice, and while I do not expect java to ever change the
toString of Date, it's better to have a consistent format across all
file systems and interfaces.


tpoterba (approved):
good change!

PR #11589, opened 2022-03-14
[query/service] fix fs-related issues

danking:
Do not approve this if test_backwards_compatibility dramatically inflates test time

### DIFF
--- a/hail/python/hail/fs/router_fs.py
+++ b/hail/python/hail/fs/router_fs.py
@@ -240,7 +240,8 @@ class RouterFS(FS):
         async def _ls():
             return await bounded_gather(
                 *[functools.partial(self._fle_to_dict, fle)
-                  async for fle in await self.afs.listfiles(path)])
+                  async for fle in await self.afs.listfiles(path)],
+                parallelism=_max_simultaneous_files)
         return async_to_blocking(_ls())
 
     def mkdir(self, path: str):
--- a/hail/python/hail/fs/router_fs.py
+++ b/hail/python/hail/fs/router_fs.py
@@ -1,13 +1,14 @@
 from typing import Dict, List, Any, AsyncContextManager, BinaryIO
 import asyncio
 import gzip
-import hurry
+from hurry.filesize import size as hurry_filesize_size
 import io
 import nest_asyncio
+import functools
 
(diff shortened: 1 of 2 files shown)


hail/python/hail/fs/router_fs.py
```diff
@@ -230,10 +238,9 @@ async def size():
 
     def ls(self, path: str, _max_simultaneous_files: int = 50) -> List[Dict]:
         async def _ls():
-            async with OnlineBoundedGather2(asyncio.Semaphore(_max_simultaneous_files)) as pool:
-                tasks = [pool.call(self._fle_to_dict, fle)
-                         async for fle in await self.afs.listfiles(path)]
-                return [await t for t in tasks]
+            return await bounded_gather(
+                *[functools.partial(self._fle_to_dict, fle)
```
jigold, line 242: Do we not care how many simultaneous queries we do?

danking: Ah, yes, I dropped the parallelism argument. I restored it

johnc1231:
Has conflict

danking:
should work now

danking:
@jigold since you already took a look I sent back to you. It passed tests on last run other than the MovieLens issue.

jigold:
Can you give me some context for this change? It seems fine to me, but I don't see how it was different than what was there before.

jigold:
Can you give me some context for this change? It seems fine to me, but I don't see how it was different than what was there before.

danking:
`AsyncFS.exist(url)` is true if and only if `url` is a blob (aka a file). Hail Query's `FS` file system API mimics Hadoop's file system API which mimics a UNIX-like file system api, so we gotta check for both isdir and isfile.

jigold:
Great! Can you check whether the issue with backwards compatibility you were worried about is indeed a problem?

danking:
There's another bug which means that test still fails and fails in ~0 seconds. I'm gonna put off fixing that bug for now because I'm not sure what to do about that test. It's currently marked as fails_service_backend.

danking:
merged into https://github.com/hail-is/hail/pull/11599 because I need both fixes together to pass all the tests.

PR #11591, opened 2022-03-15
[query] Rework python FS.stat to return an object

chrisvittal:
The new object, a StatResult (analogous to os.stat_result) is minimal
for now. Notably, it uses a fractional UNIX timestamp for modification
time. StatResult should be moderately extensible, but since it
subclasses NamedTuple, care may need to be taken to preserve field
order.


chrisvittal:
stacked on #11581

tpoterba (approved):
LGTM!

PR #11592, opened 2022-03-15
Check that the pip wheel is less than 100MB, the pypi upload size limit. 

johnc1231:


danking (changes requested):

johnc1231:
Bump, let's get this in before we break it again.

PR #11594, opened 2022-03-15
[batch] Add nonpreemptible pools

jigold:
I don't think I need extra tests for the additional nonpreemptible pools because the service tests will do so. However, they won't use the highmem or highcpu pools. We can add tests for these, but it will add additional time to the batch tests. We might want to split the batch tests up more in that case. On second thought, maybe it would be good to specifically have tests for these inside the batch module.


daniel-goldstein:
Just driving by to say very satisfying that this is under 100 LoC

danking (approved):
Agreed, this is a beautifully short PR.

PR #11597, opened 2022-03-16
[query/service] implement index_bgen

danking:


hail/python/hail/backend/service_backend.py
```diff
@@ -444,8 +451,52 @@ async def inputs(infile, _):
...
+        async def inputs(infile, _):
+            await write_int(infile, ServiceBackend.INDEX_BGEN)
+            await write_str(infile, tmp_dir())
+            await write_str(infile, self.billing_project)
+            await write_str(infile, self.remote_tmpdir)
+            await write_int(infile, len(files))
+            for fname in files:
+                await write_str(infile, fname)
+            await write_int(infile, len(index_file_map))
+            for k, v in index_file_map.values():
```
daniel-goldstein, line 483:
```suggestion
            for k, v in index_file_map.items():
```

hail/python/hail/backend/service_backend.py
```diff
@@ -444,8 +451,52 @@ async def inputs(infile, _):
...
+                await write_str(infile, k)
+                await write_str(infile, v)
+            if referenceGenomeName is None:
+                await write_bool(infile, False)
+            else:
+                await write_bool(infile, True)
+                await write_str(infile, referenceGenomeName)
+            await write_int(infile, len(contig_recoding))
+            for k, v in contig_recoding.values():
```
daniel-goldstein, line 492:
```suggestion
            for k, v in contig_recoding.items():
```

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -715,6 +744,63 @@ class ServiceBackendSocketAPI2(
...
+        while (i < nContigRecoding) {
+          val k = readString()
+          val v = readString()
+          contigRecoding(k) = v
+          i += 1
+        }
+        val skipInvalidLoci = readBool()
+
+        try {
+          val result = backend.indexBgen(
```
daniel-goldstein, line 784: `result` in unused so might as well not assign

daniel-goldstein (changes requested):
Small things

tpoterba:
I think that for a lot of this kind of functionality -- index_bgen, vcf_metadata, etc -- it's going to be easier to just convert them to IR functions and then they hit the same endpoint.

tpoterba:
for now, though, full steam with whatever's easiest.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: I made indexBgen return the string instead for symmetry with the other methods.

hail/python/hail/backend/service_backend.py
daniel-goldstein: I don't think this changed, did a commit go on the wrong branch?

daniel-goldstein: Unclear to me how this is passing tests, or else the types are wrong.

daniel-goldstein (changes requested):

daniel-goldstein (approved):

PR #11598, opened 2022-03-16
[query/service] implement parse_vcf_metadata

danking:


danking:
No new tests are passing because we need to implement VCF export first.

tpoterba:
I have another proposal here to delete this endpoint entirely and replace it with an IR function

chrisvittal:
I'm personally in favor of this change adding a switch to the protocol. It avoids running the compiler.

tpoterba (approved):
I can't really reject a working fix. I think in the future I might want to implement this as an IR function because we could delete all of this code, but 🤷

PR #11599, opened 2022-03-16
[hail.fs] fix RouterFS.remove

danking:
I also added some tests that would have caught this infinite recursion.


danking:
Forgot about the difference between rmtree and remove. I fixed and improved the tests to distinguish between these two.

chrisvittal (approved):

danking:
This will fail until #11589 merges.

danking:
I had to merge https://github.com/hail-is/hail/pull/11589 into this one to pass all the tests

PR #11605, opened 2022-03-16
[compiler] First pieces of infrastructure for reproducible randomness

patrick-schultz:
This PR begins to implement the infrastructure needed for reproducible randomness.

The main components are:
* An implementation of the Threefish block cipher, reduced to 20 rounds as in Threefry [1], but keeping the tweak from Threefish (really just the first 64 bits, the second 64 bits are always 0). The specification for Threefish can be found in [2].
* An implementation of the `jdistlib.RandomEngine` interface using Threefish, so that we can continue using the `jdistlib` implementations of sampling from various distributions.
  * This has some improvements over the standard Java RNG implementations of random floating point numbers, and of random integers from a specified interval. See comments in the code for details.
* The beginnings of a new type `(S/T)RNGState`. This implements a splittable RNG interface, similarly to [3], but instead of the cascade construction, we use a modification of PMAC [4] to build a psuedo-random function from a blockcipher. This allows us to reorder the processing of blocks of the input, in particular moving computation to compile time as much as possible.
* A simple test suite for the new RNG using a chi-square test.

[1] "Parallel random numbers: as easy as 1, 2, 3", http://www.thesalmons.org/john/random123/papers/random123sc11.pdf
[2] "The Skein Hash Function Family", https://www.schneier.com/wp-content/uploads/2015/01/skein.pdf
[3] "Splittable pseudorandom number generators using cryptographic hashing", https://publications.lib.chalmers.se/records/fulltext/183348/local_183348.pdf
[4] "Efficient Instantiations of Tweakable Blockciphers and Refinements to Modes OCB and PMAC", https://www.cs.ucdavis.edu/~rogaway/papers/offsets.pdf

### DIFF
--- a/hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
+++ b/hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
@@ -241,7 +241,8 @@ class ClassBuilder[C](
 
   val lazyFieldMemo: mutable.Map[Any, Value[_]] = mutable.Map.empty
 
-  val lInit = lclass.newMethod("<init>", FastIndexedSeq(), UnitInfo)
+  val lInitBuilder = new MethodBuilder[C](this, "<init>", FastIndexedSeq(), UnitInfo)
+  val lInit = lInitBuilder.lmethod
 
   var initBody: Code[Unit] = {
     val L = new lir.Block()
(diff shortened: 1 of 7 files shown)


tpoterba (approved):
This all looks good in the context of what we've talked about. I think it's hard to do a full critique of some of the totally new stuff; I'd love to go through this together when we have fewer balls in the air.

PR #11606, opened 2022-03-16
[query] Ensure we can import custom VCF with custom RG successfully.

tpoterba:


danking:
fails in the service due to lack of ref genome.

danking (approved):

PR #11617, opened 2022-03-18
Implement azure storage filesystem in scala

vrautela:
Created AzureStorageFS which implements the FS trait's methods so users can interface with Azure Blob Storage as though it were a filesystem.


hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+          closed = true
+        }
+      }
+    }
+
+    new WrappedPositionedDataOutputStream(os)
+  }
+
+  def delete(filename: String, recursive: Boolean): Unit = {
+    if (this.exists(filename)) {
```
daniel-goldstein, line 220: I think this is being a little overprotective here. Errors can sometimes be a good thing! If I try to delete something that doesn't exist, I expect to get an error saying "Hey, this doesn't actually exist". Silently doing nothing is surprising behavior and could prevent me from realizing my mistake earlier.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+          }
+        }
+        deleteRecursively()
+
+        if (fileStatus.isFile && blobClient.exists()) blobClient.delete()
+      }
+      else {
+          if (fileStatus.isDirectory) {
+            throw new IllegalArgumentException("Cannot non-recursively delete a directory")
+          }
```
daniel-goldstein, line 250: I think we should delegate to the underlying library here. If it takes issue with this, it will probably provide a richer error than one that we write on their behalf. That is -- let the user do what they said they want.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+            if (seenBlobItems.contains(blobItem.getName)) return
+            seenBlobItems += blobItem.getName
+            val blobFileName = s"hail-az://$account/$container/${blobItem.getName}"
+            // recursively delete all files/subfolders
+            this.delete(blobFileName, recursive)
+          }
+        }
+        deleteRecursively()
+
+        if (fileStatus.isFile && blobClient.exists()) blobClient.delete()
```
daniel-goldstein, line 245: Similarly here we should not check for existence before issuing the delete. Remember that every method that you call on the client is potentially another request going over the wire.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+          while (iter.hasNext) {
+            val blobItem: BlobItem = iter.next
+            if (seenBlobItems.contains(blobItem.getName)) return
+            seenBlobItems += blobItem.getName
+            val blobFileName = s"hail-az://$account/$container/${blobItem.getName}"
+            // recursively delete all files/subfolders
+            this.delete(blobFileName, recursive)
+          }
+        }
+        deleteRecursively()
```
daniel-goldstein, line 243: Hm, I find this a bit hard to reason about. Your `seenBlobItems` is defined as a local variable in `delete`, but `delete` is then called recursively without passing on this set of seen items. I think this does n^2 work, where each item in the tree tries to delete the entire subtree. If `listBlobsByHierarchy` does a DFS/BFS for you, is there a need to recur?

daniel-goldstein: We also Must make a big comment about the bug in `listBlobsByHierarchy` here, and diagnose exactly what the failure conditions are. I suspect that it's looping from the beginning and we don't need to keep a whole set, but even tracking just one item is a workaround and should not be confused for ideal behavior. Please put a comment above the Set definition explaining why it is this way.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+    val (account, container, path) = getAccountContainerPath(filename)
+    val blobContainerClient: BlobContainerClient = getContainerClient(filename)
+
+    val statusList: mutable.MutableList[FileStatus] = mutable.MutableList()
+    var blobNameSet: Set[String] = Set()
+
+    var seenBlobItems: Set[String] = Set()
+
+    val prefix = dropTrailingSlash(path) + "/"
+    val directoryContents = blobContainerClient.listBlobsByHierarchy(prefix)
```
daniel-goldstein, line 266: Let's cordon this buggy behavior and not repeat our workaround by creating a function that given a `blobContainerClient` and a prefix, produces an iterator of the results of `listBlobsByHierarchy` and encapsulates this short-circuiting workaround.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
daniel-goldstein: I think I might only be comfortable letting in this bug if we knew its limitations.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+        var blobName = blobItem.getName
+        // if there is a slash after the prefix, then this blob is under a subdirectory
+        if (indexOfFirstSlashAfterPrefix != -1) {
+          blobName = blobItem.getName.substring(0, indexOfFirstSlashAfterPrefix)
+        }
+
+        if (!blobNameSet.contains(blobName)) {
+          blobNameSet = blobNameSet + blobName
+          val blobFileName = s"hail-az://$account/$container/$blobName"
+          statusList += fileStatus(blobFileName)
```
daniel-goldstein, line 287: We already have the `BlobItem` at this point, so we should take advantage of the information it holds and assemble a `BlobStorageFileStatus` directly from the `BlobProperties`.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+        }
+      }
+    }
+    populateStatList()
```
daniel-goldstein, line 291: I don't think we need a local function here, can just run the body directly.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+          if (blobClient.exists()) blobClient.delete()
+      }
+    }
+  }
+
+  def listStatus(filename: String): Array[FileStatus] = {
+    val (account, container, path) = getAccountContainerPath(filename)
+    val blobContainerClient: BlobContainerClient = getContainerClient(filename)
+
+    val statusList: mutable.MutableList[FileStatus] = mutable.MutableList()
```
daniel-goldstein, line 260: Let's use an `ArrayBuffer` here instead of a `MutableList`, as the former is more like a java `ArrayList` and `MutableList` uses a linked list underneath.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+          statusList += fileStatus(blobFileName)
+        }
+      }
+    }
+    populateStatList()
+
+    statusList.toArray
+  }
+
+  def glob(filename: String): Array[FileStatus] = {
```
daniel-goldstein, line 296: This is pretty much the same as the Google implementation, they just use different implementations of `listStatus`. Let's abstract these. Maybe just put them on the trait and have hadoop override?

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+        if (seenBlobItems.contains(name)) return
+        seenBlobItems += name
+        numSubBlobs += 1
+      }
+    }
+    countSubBlobs()
+
+    if (numSubBlobs == 0 && !blobClient.exists()) throw new FileNotFoundException("file isn't found")
+
+    val isDir = numSubBlobs > 0
```
daniel-goldstein, line 371: If we're only comparing 0 or more than 0, we can just ask if the iterator `hasNext` instead of going through it all.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+      val blobProperties: BlobProperties = blobClient.getProperties
+      val modificationTime = blobProperties.getLastModified.toEpochSecond
+      val size = blobProperties.getBlobSize
+
+      new BlobStorageFileStatus(dropTrailingSlash(filename), modificationTime, size, isDir)
+    }
+  }
+
+  def makeQualified(filename: String): String = {
+    assert(filename.startsWith("hail-az://"))
```
daniel-goldstein, line 385: Same here with raising a `ValueError`?

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
daniel-goldstein: See my below comment about making an `apply` method for `AzureStorageFileStatus`

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
daniel-goldstein: Also, if you look at the python implementation, `is_file` and `is_dir` is determined based on the existence of bob properties. Let's do that instead.

hail/src/main/scala/is/hail/io/fs/FS.scala
```diff
@@ -41,6 +41,16 @@ trait FileStatus {
   def getOwner: String
 }
 
+class BlobStorageFileStatus(path: String, modificationTime: java.lang.Long, size: Long, isDir: Boolean) extends FileStatus {
```
daniel-goldstein, line 44: If you look at the `GoogleStorageFileStatus` *object*, it uses an apply method to extract these values from the google object. We can do that too with `BlobProperties` if we create a similar `AzureStorageFileStatus`.

hail/src/main/scala/is/hail/io/fs/FS.scala
```diff
@@ -118,23 +208,30 @@ trait FS extends Serializable {
...
   def listStatus(filename: String): Array[FileStatus]
 
   def glob(filename: String): Array[FileStatus]
 
-  def globAll(filenames: Iterable[String]): Array[String]
+  def globAll(filenames: Iterable[String]): Array[String] =
+    globAllStatuses(filenames).map(_.getPath)
 
-  def globAllStatuses(filenames: Iterable[String]): Array[FileStatus]
+  def globAllStatuses(filenames: Iterable[String]): Array[FileStatus] = filenames.flatMap(glob).toArray
```
daniel-goldstein, line 222: Nice

hail/src/main/scala/is/hail/io/fs/FS.scala
```diff
@@ -60,6 +70,86 @@ object BGZipCompressionCodec extends CompressionCodec {
...
+    }
+
+    val toTransfer = math.min(len, bb.remaining())
+    bb.get(bytes, off, toTransfer)
+    pos += toTransfer
+    toTransfer
+  }
+
+  def getPosition: Long = pos
+}
```
daniel-goldstein, line 120: Nice abstracting these

hail/src/main/scala/is/hail/io/fs/FS.scala
```diff
@@ -60,6 +70,86 @@ object BGZipCompressionCodec extends CompressionCodec {
   def makeOutputStream(os: OutputStream): OutputStream = new BGzipOutputStream(os)
 }
 
+object FSUtil {
+  def runOnExit(f: Runnable): Unit = {
+    Runtime.getRuntime.addShutdownHook(new Thread(f))
+  }
```
daniel-goldstein, line 76: I think this is unused as you've defined this default on the `FS` trait. I'm fine putting it on the trait so this can be deleted.

daniel-goldstein (changes requested):
Good work putting this together. I think there is more work we can do to tighten up the implementation from primarily from an organizational and performance standpoint. Way to hit the ground running in a new language!

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+
+      def fill(): Unit = {
+        bb.clear()
+
+        // read some bytes
+        val outputStream: ByteArrayOutputStream = new ByteArrayOutputStream()
+        val blobProperties = client.getProperties
+        val blobSize = blobProperties.getBlobSize
+
+        val count = Math.min(blobSize - pos, bb.remaining() - 1)
```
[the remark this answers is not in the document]
vrautela: I think I must have confused myself at some point because I thought there would be an off by one error which would cause the buffer to overflow. However I reread the documentation for the [BlobRange class](https://javadoc.io/static/com.azure/azure-storage-blob/12.8.0/com/azure/storage/blob/models/BlobRange.html) and it seems to be what you'd expect where count is the number of bytes being downloaded. I changed it to `bb.remaining()`and all tests are passing so I think that's fine

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+        if (count <= 0) {
+          eof = true
+          return
+        }
+
+        // calculate the minimum of the bytes remaining in the buffer and the bytes remaining
+        // to be read in the blob
+        client.downloadWithResponse(
+          outputStream, new BlobRange(pos, count),
+          null, null, false, Duration.ofMinutes(1), null)
```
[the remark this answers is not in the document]
vrautela: I don't think I can. I tried adding the parameter names and IntelliJ didn't like that. I think since the actual library is written in Java which doesn't have named arguments, it might not be possible to add them in

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,388 @@
...
+        pos += count
+        bb.put(outputStream.toByteArray)
```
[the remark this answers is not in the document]
vrautela: I think I have implemented an OutputStream which does this and passes existing tests

hail/src/main/scala/is/hail/io/fs/FS.scala
vrautela: I don't think we can exactly mimic that behavior since I don't believe that a directory has a BlobProperties object in the Azure API, but I've implemented something similar

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
vrautela: I'm not sure if I can implement fileStatus in that way. If you call getProperties on a BlobClient and you get a 404 response code you can deduce that the path you called it on is the path to a directory or that the path does not correspond to a blob in Azure Storage, but you can't distinguish those two outcomes

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
vrautela: This is not as easy as I would like it to be since AzureStorage will return a 404 error if you call getProperties on a path that corresponds to a directory (and not directly to a blob)

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
vrautela: Actually the way the filenames are deconstructed and the prefix is constructed are also different, so I don't know if this abstraction makes sense (it also looks like only 2 methods are scheme-specific)

daniel-goldstein: I think we can reuse this if we change the signature slightly. What does glob actually need of the filename here? It looks like it needs a prefix, so the bucket or account/container pair, and it needs the path because it slits it into `components`. So what if we make a helper with the following signature `glob(prefix: String, path: String)` and the Google/Azure implementations call that after splitting it appropriately. It would also avoid this weird re-assembling of the string on line 286 happening inside the glob implementation.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,320 @@
...
+
+    val blobClient: BlobClient = getBlobClient(filename)
+    val blobContainerClient: BlobContainerClient = getContainerClient(filename)
+
+    val prefix = dropTrailingSlash(path) + "/"
+    val options: ListBlobsOptions = new ListBlobsOptions().setPrefix(prefix)
+    val prefixMatches = blobContainerClient.listBlobs(options, null)
+    val isDir = prefixMatches.iterator().hasNext
+
+    if (!isDir && !blobClient.exists()) throw new FileNotFoundException("file isn't found")
```
daniel-goldstein, line 305: nit: can you add `filename` to the error message here and capitalize the first word?

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
daniel-goldstein: We should raise `ValueError` here instead so that clients have an opportunity to `catch` it and recover instead of the program crashing.

daniel-goldstein: That would be a good test to add actually

hail/src/main/scala/is/hail/io/fs/FS.scala
```diff
@@ -60,7 +70,115 @@ object BGZipCompressionCodec extends CompressionCodec {
...
+      remaining -= toTransfer
+      pos += toTransfer
+    }
+  }
+
+  def getPosition: Long = pos
+}
+
 trait FS extends Serializable {
+  def containsWildcard(path: String): Boolean = {
```
daniel-goldstein, line 163: Can you put this on the `FSUtil` object? It's more a helper than part of the API

hail/src/test/scala/is/hail/fs/azure/AzureStorageFSSuite.scala
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }
+
+  @Test def testDropTrailingSlash(): Unit = {
+    assert(dropTrailingSlash("") == "")
+    assert(dropTrailingSlash("/foo/bar") == "/foo/bar")
+    assert(dropTrailingSlash("foo/bar/") == "foo/bar")
+    assert(dropTrailingSlash("/foo///") == "/foo")
+    assert(dropTrailingSlash("///") == "")
+  }
```
daniel-goldstein, line 50: Now that these are the same method, we don't need the test in two different places.

daniel-goldstein (dismissed):
This is looking a lot better, mainly have a few clarity suggestions. After this we can kick this to another Services person.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
vrautela: Gotcha, I can add that test. I'm also going to change this behavior in the GoogleStorageFS

danking:
Which PR is this stacked on?

daniel-goldstein:
#11699, still trying to sort out the shaded azure dependency, this is operating under the assumption that we can use the azure dependency specified in the build.gradle

danking:
What's the status of this PR?

daniel-goldstein:
I'm still working out the best workaround for the library version. I tried a mix of gradle sub-project-ing and shading with John and they all didn't work out for various different reasons. Azure library was doing some weird classLoader stuff that didn't like being bulk-relocated, and force-upgrading jackson didn't work because pyspark bundles in the buggy version of jackson without shading it. We can try to work out the former but I decided to pause on that front and try instead to just not use the buggy Azure method from the compatible library version and hit the REST endpoint directly, which involves some XML parsing. Happy to discuss the various approaches at our 1:1 tomorrow or earlier.

daniel-goldstein:
@danking This is passing tests now

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,280 @@
...
+        pos = newPos
+      }
+    }
+
+    new WrappedSeekableDataInputStream(is)
+  }
+
+  def createNoCompression(filename: String): PositionedDataOutputStream = {
+    val (account, container, path) = getAccountContainerPath(filename)
+    val appendClient: AppendBlobClient = getBlobClient(account, container, path).getAppendBlobClient
```
danking, line 170: This type isn't required, right? AFAIK, we generally avoid type annotations in Scala if they're unnecessary.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,280 @@
...
+    if (!appendClient.exists()) appendClient.create()
```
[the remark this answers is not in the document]
danking: Yeah this is wrong it needs to be `appendClient.create(true)` with no `if`.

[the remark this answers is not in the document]
danking: I'm honestly surprised Azure lets you append to a blob, I don't think Google permits this.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,280 @@
...
+    val blobClient: BlobClient = getBlobClient(account, container, path)
+
+    val fileStatus = this.fileStatus(filename)
+
+    if (recursive) {
+      val blobContainerClient = getContainerClient(account, container)
+
+      val options = new ListBlobsOptions()
+      options.setPrefix(dropTrailingSlash(path) + "/")
+      val prefixMatches = blobContainerClient.listBlobs(options, Duration.ofMinutes(1))
```
danking, line 208: 5 second timeout. Make this a static field

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,280 @@
...
+
+      prefixMatches.forEach(blobItem => {
+        getBlobClient(account, container, blobItem.getName).delete()
```
danking, line 211: assert that `!blobItem.isPrefix`

danking, line 211: assert that `!blobItem.isPrefix`

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    new WrappedPositionedDataOutputStream(os)
+  }
+
+  def delete(filename: String, recursive: Boolean): Unit = {
+    val (account, container, path) = getAccountContainerPath(filename)
+    val blobClient: BlobClient = getBlobClient(account, container, path)
+
+    val fileStatus = this.fileStatus(filename)
```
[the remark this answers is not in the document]
danking: call the variable `stat` or something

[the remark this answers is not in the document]
danking: call the variable `stat` or something

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,280 @@
...
+    var (account, container, path) = getAccountContainerPath(filename)
+    path = dropTrailingSlash(path)
+
+    globWithPrefix(prefix = s"hail-az://$account/$container", path = path)
+  }
+
+  def fileStatus(account: String, container: String, path: String): FileStatus = {
+    if (path == "") {
+      return new BlobStorageFileStatus(s"hail-az://$account/$container", null, 0, true)
+    }
```
danking, line 248: This seems bizarre. Why not error if you ask for the blob with no name?

danking, line 248: This seems bizarre. Why not error if you ask for the blob with no name?

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,280 @@
...
+    val blobClient: BlobClient = getBlobClient(account, container, path)
+    val blobContainerClient: BlobContainerClient = getContainerClient(account, container)
+
+    val prefix = dropTrailingSlash(path) + "/"
+    val options: ListBlobsOptions = new ListBlobsOptions().setPrefix(prefix)
+    val prefixMatches = blobContainerClient.listBlobs(options, null)
+    val isDir = prefixMatches.iterator().hasNext
+
+    val filename = dropTrailingSlash(s"hail-az://$account/$container/$path")
+    if (!isDir && !blobClient.exists()) throw new FileNotFoundException(s"File not found: $filename")
```
danking, line 259: prefer three lines with brackets for consistency across the code base

danking, line 259: prefer three lines with brackets for consistency across the code base

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,280 @@
...
+
+    if (isDir) {
+      new BlobStorageFileStatus(path = filename, -1, -1, isDir = true)
+    }
+    else {
```
danking, line 264: prefer `} else {` for consistency across the code base

danking, line 264: prefer `} else {` for consistency across the code base

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,280 @@
...
+    val prefix = dropTrailingSlash(path) + "/"
+    val options: ListBlobsOptions = new ListBlobsOptions().setPrefix(prefix)
+    val prefixMatches = blobContainerClient.listBlobs(options, null)
+    val isDir = prefixMatches.iterator().hasNext
+
+    val filename = dropTrailingSlash(s"hail-az://$account/$container/$path")
+    if (!isDir && !blobClient.exists()) throw new FileNotFoundException(s"File not found: $filename")
+
+    if (isDir) {
+      new BlobStorageFileStatus(path = filename, -1, -1, isDir = true)
```
danking, line 262: consistency above all else: we use size zero for directories elsewhere.

danking, line 262: consistency above all else: we use size zero for directories elsewhere.

danking: and null for modification time

danking (changes requested):
Overall this looks great but some important changes are needed.

build.yaml
```diff
@@ -793,7 +793,8 @@ steps:
       HAIL_PIP_VERSION=$(cat python/hail/hail_pip_version)
       WHEEL_PATH="build/deploy/dist/hail-$HAIL_PIP_VERSION-py3-none-any.whl"
       du -h $WHEEL_PATH
-      $(python3 -c "import os; exit(1) if (os.path.getsize('$WHEEL_PATH')) > 100_000_000 else exit(0)")
+      python3 -c "import os; print(os.path.getsize('$WHEEL_PATH'))"
+      $(python3 -c "import os; exit(1) if (os.path.getsize('$WHEEL_PATH')) > (100 * 1024 * 1024) else exit(0)")
```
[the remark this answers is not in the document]
daniel-goldstein: ya we should stack this on #11814

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,280 @@
...
+      def fill(): Unit = {
+        bb.clear()
+
+        val outputStreamToBuffer: OutputStream = (i: Int) => {
+          bb.put(i.toByte)
+        }
+
+        // calculate the minimum of the bytes remaining in the buffer and the bytes remaining
+        // to be read in the blob
+        val count = Math.min(blobSize - pos, bb.remaining())
```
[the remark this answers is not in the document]
vrautela: I like the let-binding option

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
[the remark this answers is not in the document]
vrautela: Why don't we like the "this" keyword?

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
vrautela: This is done because if a user asks for the status of the root blob (which is identified by hail-az://accountName/containerName/) then I don't think you should error

daniel-goldstein: I'm with Vedant here but I wouldn't call it the root blob, since it is not a blob. This would in my mind be akin to `ls /`

hail/src/main/scala/is/hail/io/fs/FS.scala
```diff
@@ -60,6 +73,114 @@ object BGZipCompressionCodec extends CompressionCodec {
...
+    if (bb.remaining() == 0) {
+      fill()
+      if (eof)
+        return -1
+    }
+
+    val toTransfer = math.min(len, bb.remaining())
+    bb.get(bytes, off, toTransfer)
+    pos += toTransfer
+    toTransfer
```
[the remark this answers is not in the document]
daniel-goldstein: I believe this will result in a short read, and it's up to the callee to invoke `read` until `EOF`. Unclear to me which semantics we want but this is lifted from the current `GoogleStorageFS`.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -0,0 +1,280 @@
...
+        if (count <= 0) {
```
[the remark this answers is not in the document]
daniel-goldstein: It appears so. Should we disallow seeking past the end?

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
daniel-goldstein: Ah wait no, that would be `path = "/"`. Ya Dan I think this should error. However, this is the current behavior.

build.yaml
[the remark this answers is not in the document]
danking: ping me when https://github.com/hail-is/hail/pull/11814 merges and you've rebased.

danking (dismissed):

danking (approved):

PR #11620, opened 2022-03-19
[query/service] implement the function registry in the ServiceBackend

danking:
This PR keeps a list of registered function client-side. When we send
an execute operation, we also send all the functions with the IR-to-execute.

The bulk of the deletions are removing fails_service_backend markers.

### DIFF
--- a/hail/python/hail/backend/py4j_backend.py
+++ b/hail/python/hail/backend/py4j_backend.py
@@ -84,7 +84,7 @@ class Py4JBackend(Backend):
 
         Env.hail().expr.ir.functions.IRFunctionRegistry.pyRegisterIR(
             name,
-            [ta._parsable_string() for ta in type_parameter],
+            [ta._parsable_string() for ta in type_parameters],
             value_parameter_names,
             [pt._parsable_string() for pt in value_parameter_types],
             return_type._parsable_string(),
--- a/hail/python/hail/backend/py4j_backend.py
+++ b/hail/python/hail/backend/py4j_backend.py
@@ -1,4 +1,4 @@
-from typing import Mapping
+from typing import Mapping, Union, Tuple, List
 import abc
 
 import py4j
@@ -10,6 +10,8 @@ from hail.ir import JavaIR
 from hail.ir.renderer import CSERenderer
 from hail.utils.java import FatalError, Env
 from .backend import Backend, fatal_error_from_java_error_triplet
+from ..expr import Expression
+from ..expr.types import HailType
 
 
 def handle_java_exception(f):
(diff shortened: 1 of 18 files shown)


danking:
Stacked on https://github.com/hail-is/hail/pull/11646

danking:
rebased

hail/python/hail/backend/py4j_backend.py
```diff
@@ -75,15 +75,22 @@ def _parse_value_ir(self, code, ref_map={}, ir_map={}):
     def _to_java_value_ir(self, ir):
         pass
 
-    def register_ir_function(self, name, type_parameters, argument_names, argument_types, return_type, body):
+    def register_ir_function(self,
+                             name,
```
jigold, line 79: Is it worth adding types here as well?

hail/src/main/scala/is/hail/expr/ir/functions/Functions.scala
```diff
@@ -82,6 +83,39 @@ object IRFunctionRegistry {
...
+    argTypeStrs: Array[String],
+    returnType: String,
+    bodyStr: String
+  ): Unit = {
+    requireJavaIdentifier(name)
+
+    val typeParameters = typeParamStrs.map(IRParser.parseType).toFastIndexedSeq
+    val valueParameterTypes = argTypeStrs.map(IRParser.parseType).toFastIndexedSeq
+    val refMap = argNames.zip(valueParameterTypes).toMap
+    val body = IRParser.parse_value_ir(
```
jigold, line 100: Why is this function `parse_value_ir` not camel case?

hail/src/main/scala/is/hail/expr/ir/functions/Functions.scala
```diff
@@ -82,6 +83,39 @@ object IRFunctionRegistry {
...
+      bodyStr,
+      IRParserEnvironment(ctx, refMap, Map())
+    )
+
+    userAddedFunctions += ((name, (body.typ, typeParameters, valueParameterTypes)))
+    addIR(
+      name,
+      typeParameters,
```
jigold, line 108: Can someone on compilers team check this block?

jigold (changes requested):

hail/python/hail/backend/py4j_backend.py
danking: I'm pretty sure mypy finds and uses the types from the super method, so it's not necessary for type checking, but I will add for documentary purposes.

hail/src/main/scala/is/hail/expr/ir/functions/Functions.scala
danking: Because they call it from Python, I presume.

hail/src/main/scala/is/hail/expr/ir/functions/Functions.scala
danking: @johnc1231  can you take a look? My intent is for this to be exactly the same as above except `argNames` is now an `Array` instead of a `java.utils.ArrayList`.

johnc1231:
@danking Needs a rebase

hail/src/main/scala/is/hail/expr/ir/functions/Functions.scala
danking: @johnc1231 bump

johnc1231: I'm fine with this :+1:

jigold (approved):

PR #11621, opened 2022-03-19
[query/service] proposal to ease testing in default

danking:
Now you should be able to do:
```
make -C query test NAMESPACE=default PYTEST_ARGS='-k test_foo'
```


danking:
Likewise:

```
make -C query ipython NAMESPACE=default
```

If you know the JAR & resources are uploadedc (or you don't need the resources), you can run:
```
make -C query ipython-no-deps NAMESPACE=default
```

chrisvittal (approved):

PR #11622, opened 2022-03-20
[hailtop/utils] Allow OnlineBoundedGather2.call to return values

chrisvittal:
This was breaking RouterFS.ls

Tested in the export_vcf lowering branch


chrisvittal:
I have no idea why `test_range_matrix_table_n_lt_partitions` is failing on this branch. Running it locally hangs.

danking (approved):

chrisvittal:
Why won't this merge?

danking:
I retried

danking:
My index bgen PR won the race, but this should be next.

danking:
a test job got preempted so that added ~20 minutes to total test time

danking:
Ugh, and in Azure, the batch-driver was too slow coming alive and it failed the build. Retried.

PR #11623, opened 2022-03-20
[query service] ServiceBackend rpc calls should raise FatalError

chrisvittal:


chrisvittal:
This probably needs more work to return a 'nice' error for user errors and internal errors (for bug reporting), though with the service, we should have visibility into compiler errors.

danking (approved):
I have a PR stacked on a couple others which properly handles error_id, so we'll have that support soon!

PR #11624, opened 2022-03-20
[query/service] use error id to raise user-friendly errors

danking:
The main change is to the communication protocol between the client and
the driver and between the driver and the worker.

In main, both the driver and the client send messages back to the client
and driver (respectively) by writing to a file in cloud storage. In both
cases, the file (in main) has one of these two structures:

    0x00                  # is_success (False)
    UTF-8 encoded string  # the stack trace

    0x01                  # is_success (True)
    UTF-8 encoded string  # JSON message to send back to the client or driver

In this PR, the success case does not change. The failure case becomes:

    0x00                  # is_sucess (False)
    UTF-8 encoded string  # short message
    UTF-8 encoded string  # expanded message
    4-byte signed integer # error id

The Python client (in `service_backend.py`) and the Driver (in
`ServiceBackendSocketAPI2`) changes to read this and raise the right error if
an error id is present.

I also uncovered three unrelated problems that are fixed in this PR:
1. PlinkVariant needs to be serializable because it is broadcasted.
2. We open an input stream in LoadPlink which ought to be closed, but there is no mechanism to do so in the ServiceBackend. I just ignore it for now. cc: @tpoterba, I'm not sure what the right answer is here.
3. Two uses of the broadcasted file system that should use the ExecuteContext's file system.

### DIFF
--- a/hail/python/hail/backend/backend.py
+++ b/hail/python/hail/backend/backend.py
@@ -2,6 +2,22 @@ from typing import Mapping, List, Dict, Optional
 import abc
 from ..fs.fs import FS
 from ..expr import Expression
+from ..expr.types import HailType
+from ..ir import BaseIR
+from ..utils.java import FatalError, HailUserError
+
+
+def fatal_error_from_java_error_triplet(short_message, expanded_message, error_id):
+    from .. import __version__
+    if error_id != -1:
+        return FatalError(f'Error summary: {short_message}', error_id)
+    return FatalError(f'''{short_message}
+
+Java stack trace:
+{expanded_message}
+Hail version: {__version__}
+Error summary: {short_message}''',
+                      error_id)
 
 
 class Backend(abc.ABC):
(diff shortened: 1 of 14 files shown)


hail/src/main/scala/is/hail/expr/ir/functions/Functions.scala
```diff
@@ -82,6 +83,39 @@ object IRFunctionRegistry {
...
 
+  def pyRegisterIRForServiceBackend(
+    ctx: ExecuteContext,
+    name: String,
+    typeParamStrs: Array[String],
+    argNames: Array[String],
+    argTypeStrs: Array[String],
+    returnType: String,
+    bodyStr: String
+  ): Unit = {
```
danking, line 94: This is just a minor modification to the above function:
1. Directly consume Scala Arrays instead of ArrayList. 
2. Construct a refMap from the arguments and parameters and use that to parse the body.

Unlike the Py4J backends, in the service backend, the IR will never be already constructed in Java, so it makes sense to do all the type, argument, and body parsing in one method all together.

danking:
Stacked on https://github.com/hail-is/hail/pull/11646

johnc1231:
@danking Needs to be unstacked and conflicts addressed

PR #11626, opened 2022-03-21
[query/service] implement globbing for RouterFS

danking:
The `hadoop_ls` interface has traditionally supported globbing. This improves the service backend to also support Python-style globbing. The strategy is straightforward:
1. Break the object name into components (the bits between `/`s)
2. Find the components with glob patterns. Call these pattern-components.
3. For each pattern-component, list the files in its `dirname`. Find entries which match the pattern-component.

cc: @tpoterba @chrisvittal , y'all have a vested interest in the Query Python FS API.

### DIFF
--- a/hail/python/hail/fs/router_fs.py
+++ b/hail/python/hail/fs/router_fs.py
@@ -1,6 +1,7 @@
 from typing import List, AsyncContextManager, BinaryIO, Optional
 import asyncio
 import io
+from hailtop.aiotools.local_fs import LocalAsyncFSURL
 import nest_asyncio
 import os
 import functools
@@ -261,6 +262,8 @@ class RouterFS(FS):
             raise ValueError(f'glob pattern only allowed in path (e.g. not in bucket): {path}')
 
         blobpath = url.path
+        if isinstance(url, LocalAsyncFSURL) and blobpath[0] != '/':
+            blobpath = './' + blobpath
 
         components = blobpath.split('/')
         assert len(components) > 0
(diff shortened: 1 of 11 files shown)


hail/python/hailtop/aiocloud/aiogoogle/client/storage_client.py
```diff
@@ -288,7 +288,7 @@ async def insert_object(self, bucket: str, name: str, **kwargs) -> WritableStrea
             params = {}
             kwargs['params'] = params
         assert 'name' not in params
-        params['name'] = name
+        params['name'] = urllib.parse.quote(name, safe="")
```
danking, line 291: this was a latent bug

hail/python/hail/fs/router_fs.py
```diff
@@ -232,13 +237,92 @@ async def size():
...
+                running_prefix.append(component)
+            else:
+                glob_components.append((running_prefix, component))
+                running_prefix = []
+
+        suffix_components: List[str] = running_prefix
+        first_prefix = '/'.join([url.scheme +':', '', *url.bucket_parts])
+        cumulative_prefixes = [first_prefix]
+
+        for intervening_components, single_component_glob_pattern in glob_components:
```
daniel-goldstein, line 286: Hard for me to tell what intervening means. I feel like `no_pattern_components` is more straightforward.

daniel-goldstein (changes requested):
Nice. It took me half-implementing this another way to understand why you went to the trouble of grouping non-pattern components instead of just one loop over the components listing as you go. Perhaps a performance comment to that effect would help, because my first impression was that it was a bit more complicated than it needed to be.

hail/python/hailtop/aiocloud/aiogoogle/client/storage_client.py
```diff
@@ -513,13 +537,19 @@ def __init__(self, *,
...
     @staticmethod
     def get_bucket_name(url: str) -> Tuple[str, str]:
         parsed = urllib.parse.urlparse(url)
         if parsed.scheme != 'gs':
             raise ValueError(f"invalid scheme, expected gs: {parsed.scheme}")
 
         name = parsed.path
+        if parsed.query:
+            # gs:// URLs don't have query parameters, so just include it as part of the object name
+            name += '?' + parsed.query
```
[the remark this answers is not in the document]
danking: We should talk about this.

hail/python/hail/fs/router_fs.py
daniel-goldstein: preceeding is fine

daniel-goldstein (approved):

PR #11627, opened 2022-03-21
[hailtop] add types to AsyncThrottledGather and bounded_gather

danking:


hail/python/hailtop/utils/utils.py
```diff
@@ -159,24 +160,30 @@ async def blocking_to_async(thread_pool: concurrent.futures.Executor,
         thread_pool, lambda: fun(*args, **kwargs))
 
 
-async def bounded_gather(*pfs, parallelism=10, return_exceptions=False):
-    gatherer = AsyncThrottledGather(*pfs,
-                                    parallelism=parallelism,
-                                    return_exceptions=return_exceptions)
+async def bounded_gather(*pfs: Callable[[], Awaitable[T]],
+                         parallelism=10,
+                         return_exceptions=False
```
jigold, line 165: Can you add the types here as well for parallelism and return_exceptions?

jigold (changes requested):

hail/python/hailtop/utils/utils.py
danking: good point done.

danking:
bump

jigold (approved):

PR #11630, opened 2022-03-21
[hail] fix some mypy types

danking:
Dict is generally supposed to have type arguments, we can use `dict`, lowercase, if
we do not want to specify type arguments.


johnc1231 (approved):

danking:
Thank you John!

PR #11631, opened 2022-03-21
[query/service] fix idempotent test

danking:
The service backend has no SparkContext, but we can still check for idempotence.


danking:
This has to wait for the register_ir_function to land. More on that soon.

PR #11634, opened 2022-03-21
[query] Improve Makefile Compatibility

chrisvittal:
When using GNU Make 4.3, the slashes in the json string in
query/Makefile were being passed unmodified as arguments to the python
copy process. In order to fix this, I had to put the json input on one
line. Furthermore, I converted all indents to tabs for consistency.

I also modified hailtop.aiotools.copy to read from standard input if '-'
or no 'files' argument is provided. I belive this will allow easier use
of the copy tool by end users, for example, using [jo], the copy
invocation looks like:

```make
	jo -a \
		$$(jo from=../hail/src/test/resources to=$(HAIL_TEST_RESOURCES_DIR)) \
		$$(jo from=../hail/python/hail/docs/data to=$(HAIL_DOCTEST_DATA_DIR)) | \
		python3 -m hailtop.aiotools.copy 'null'
```

[jo]: https://jpmens.net/2016/03/05/a-shell-command-to-create-json-jo/

### DIFF
--- a/hail/python/hailtop/aiotools/copy.py
+++ b/hail/python/hailtop/aiotools/copy.py
@@ -1,8 +1,9 @@
 from typing import List, Optional, Dict
-import json
-import asyncio
-import logging
 import argparse
+import asyncio
+import json
+import logging
+import sys
 import uvloop
 from concurrent.futures import ThreadPoolExecutor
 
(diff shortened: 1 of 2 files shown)


danking (approved):
This is really nice and `jo` is neat. Thank you much!

PR #11636, opened 2022-03-21
[infra] Update k8s to 8-core nodes

daniel-goldstein:
There's a few things happening here:

### Node pool updates through terraform
I extended the node pool update documentation with how to deal with terraform-managed node pools. This is what I did on Azure and worked fine. The only real change in terraform other than changing the machine type is making the node pool name configurable to adhere to the naming guidelines and allow us to do a rolling migration.

### Updated the kubernetes and azurerm providers
I updated the azurerm provider without thinking much about it and even though it's a minor version had some breaking changes that after a half-successful `apply` made it hard to downgrade. So I decided just to appease the breaking change and leave us at the new version, which is what all the `blob_properties` changes are for. They are in no way related to the node pools.

### Troubleshooting
I added a section for a bug that I've seen a couple of times (and encountered again today) but never documented how I got around it.

### DIFF
--- a/dev-docs/kubernetes-operations.md
+++ b/dev-docs/kubernetes-operations.md
@@ -2,6 +2,8 @@
 
 ## Altering a Node Pool
 
+### When managing node pools manually
+
 We will have the old node pool and the new node pool active simultaneously. We will use `cordon` and
 `drain` to move all load from the old node pool to the new node pool. Then we will delete the old
 node pool.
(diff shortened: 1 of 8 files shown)


dev-docs/kubernetes-operations.md
```diff
@@ -47,3 +47,48 @@ kubectl drain --delete-emptydir-data --ignore-daemonsets --selector="cloud.googl
 ```
 gcloud container node-pools delete $OLD_POOL_NAME --cluster $CLUSTER_NAME
 ```
+
+### When using terraform
```
danking, line 51: Can you add a H3 above as well to distinguish these two different approaches?

danking (changes requested):
Just one change

danking (approved):

PR #11637, opened 2022-03-21
[query] Lowered BlockMatrixBinaryReader and BlockMatrixBinaryWriter

johnc1231:
This allows the use of `from_numpy` and `to_numpy` on the local / service backends.


johnc1231:
Problem with this one is that I have to use the right file api for service backend

danking:
passes on local now: test_rectangles_to_numpy

I think the easiest thing to do is to just switch on spark vs non-spark. For spark its important we're not using HDFS for tofile, but for Local and Service, ideally, we just use a normal `hl.TemporaryFilename` and load it via `hl.current_backend().fs.open`. I'm not sure how finicky numpy.fromfile / tofile are about their file objects though

johnc1231:
So yeah, there are two problems with this one. The second is that mysterious `test_rectangles_to_numpy` thing which shouldn't work and fails locally.

danking:
I reproduced this by running all the tests together. There's some inter-test leakage.

danking:
It does fail if I just run test_linalg.py, unfortunately.

danking:
repro
```
make -C hail pytest PYTEST_ARGS="-k 'test_init_hail_context_twice or test_rectangles_to_numpy'" HAIL_QUERY_BACKEND=local
```

danking:
Got it.

johnc1231:
You can review Tim, mostly straightforward. The only thing I'm going to change is special casing on the backend to figure out where to write this numpy file

tpoterba:
Looking now.

hail/src/main/scala/is/hail/expr/ir/functions/NDArrayFunctions.scala
```diff
@@ -145,5 +145,11 @@ object  NDArrayFunctions extends RegistryFunctions {
         cb.ifx(info cne 0, cb._fatalWithError(errorID,s"hl.nd.solve: Could not solve, matrix was singular. dtrtrs error code ", info.toS))
         resPCode
     }
+
+    registerSCode2("export_doubles", TNDArray(TFloat64, Nat(2)), TString, TVoid, (t, ndSt, pathSt) => SVoid) { case (er, cb, rt, snd: SNDArrayValue, sstring: SStringValue, errorID) =>
```
tpoterba, line 149: delete?

hail/src/main/scala/is/hail/types/encoded/ENumpyBinaryNDArray.scala
```diff
@@ -0,0 +1,77 @@
...
+    cb.forLoop(cb.assign(i, 0L), i < n, cb.assign(i, i + 1L), {
+      readElemF(cb, region, currElementAddress, in)
+      cb.assign(currElementAddress, currElementAddress + pt.elementType.byteSize)
+    })
+
+    tFinisher(cb)
+  }
+
+  def _buildSkip(cb: EmitCodeBuilder, r: Value[Region], in: Value[InputBuffer]): Unit = {
+    ???
```
tpoterba, line 64: fine to go in like this.

tpoterba (changes requested):
one piece of dead code to GC, otherwise looks great and elegant.

hail/src/main/scala/is/hail/types/encoded/ENumpyBinaryNDArray.scala
johnc1231: It's never going to have a skipper. I should throw an error instead.

hail/src/main/scala/is/hail/expr/ir/functions/NDArrayFunctions.scala
johnc1231: Yes will delete

danking:
I pushed fixes for the filesystem issues.

danking (approved):
I addressed Tim's comment about deleting a function.

PR #11638, opened 2022-03-21
[batch] Use nginx reverse proxy to terminate TLS for batch-driver

daniel-goldstein:
This change adds an nginx sidecar to the batch-driver pod for terminating TLS. TLS negotiation has proven a major bottleneck to scheduling performance, as the batch-driver currently spends up to 60% of its CPU time in handshakes with workers. Moving TLS termination into a sidecar that can leverage additional cores both reduced CPU pressure on the driver and allowed for a 3-4x increase in job-scheduling throughput.

##  Benchmarking
Below are before-and-after profiles of the same benchmark (30,000 1s jobs) under the proposed higher rate limit, showing CPU time:
Screen Shot 2022-03-21 at 5 10 13 PM
Screen Shot 2022-03-21 at 5 12 33 PM

Looking at Wall time, the before profile is nearly identical because at the current rate limit the driver uses 100% of its CPU shares under this benchmark. On this branch, CPU utilization drops to 40-60%, giving the following wall time profile:
Screen Shot 2022-03-21 at 5 30 39 PM

So we can be confident that driver CPU is no longer a bottleneck even in the increased rate limit.

## So what's the bottleneck now?
Since the higher rate limit still leaves the driver plenty of CPU room (I've seen it peak at 60% of a vCPU), why not crank it higher? Well, we're increasing concurrency so latent deadlocks start to be a bigger issue again. We start to see tens of deadlocks per second in the proposed rate limit and hundreds at higher rate limits. As a result, we're spending more cycles repeating queries instead of actually scheduling faster. Next steps should focus on eliminating deadlocks before we can continue to max out CPU use.

## Miscellaneous
We've lumped in a few other monitoring changes that were helpful in this process and tried to leave the commits tidy. The one potentially rude change is enforcing a minimum wait time of half a second for `run_if_changed` loops. This dramatically reduced the number of scheduling loop invocations we were executing, greatly reducing the number of `compute_fair_share` queries, while maintaining the same scheduling ability. This feels like a fine requirement, but I'm unfamiliar with other use cases and can introduce a less invasive change if desired.

cc @vrautela

### DIFF
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -5,6 +5,8 @@ from typing import Optional
 
 import sortedcontainers
 
+import prometheus_client as pc
+
 from gear import Database
 from hailtop import aiotools
 from hailtop.utils import (
@@ -27,6 +29,8 @@ from ..resource_manager import CloudResourceManager
 from .base import InstanceCollection, InstanceCollectionManager
 
 log = logging.getLogger('pool')
+
+SCHEDULING_LOOP_RUNS = pc.Counter('scheduling_loop_runs', '')
 
 
 class Pool(InstanceCollection):
@@ -373,6 +377,7 @@ HAVING n_ready_jobs + n_running_jobs > 0;
 
         log.info(f'schedule {self.pool}: starting')
         start = time_msecs()
+        SCHEDULING_LOOP_RUNS.inc()
         n_scheduled = 0
 
         user_resources = await self.compute_fair_share()
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -29,7 +29,11 @@ from .base import InstanceCollection, InstanceCollectionManager
 
 log = logging.getLogger('pool')
 
-SCHEDULING_LOOP_RUNS = pc.Counter('scheduling_loop_runs', '')
+SCHEDULING_LOOP_RUNS = pc.Counter(
+    'scheduling_loop_runs',
+    'Number of scheduling loop executions per pool',
+    ['pool_name'],
+)
 
 
 class Pool(InstanceCollection):
@@ -376,7 +380,7 @@ HAVING n_ready_jobs + n_running_jobs > 0;
 
         log.info(f'schedule {self.pool}: starting')
         start = time_msecs()
-        SCHEDULING_LOOP_RUNS.inc()
+        SCHEDULING_LOOP_RUNS.labels(pool_name=self.pool.name).inc()
         n_scheduled = 0
 
         user_resources = await self.compute_fair_share()
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -3,9 +3,8 @@ import logging
 import random
 from typing import Optional
 
-import sortedcontainers
-
 import prometheus_client as pc
+import sortedcontainers
 
 from gear import Database
 from hailtop import aiotools
--- a/batch/deployment.yaml
+++ b/batch/deployment.yaml
@@ -34,6 +34,26 @@ spec:
          value: "spot"
 {% endif %}
       containers:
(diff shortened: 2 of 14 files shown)


daniel-goldstein:
Interesting, looks like I broke the same two tests on GCP and Azure. I'll dig into that tomorrow.

danking:
> Next steps should focus on eliminating deadlocks before we can continue to max out CPU use.

👍

batch/batch/driver/instance_collection/pool.py
```diff
@@ -28,6 +29,8 @@
 
 log = logging.getLogger('pool')
 
+SCHEDULING_LOOP_RUNS = pc.Counter('scheduling_loop_runs', '')
```
danking, line 32: Hmm. Should this be per-pool or something?

batch/deployment.yaml
```diff
@@ -34,6 +34,26 @@ spec:
...
       containers:
+      - name: nginx
+        image: {{ batch_driver_nginx_image.image }}
+        resources:
+          requests:
+            cpu: "4"
+            memory: "2G"
+          limits:
+            cpu: "4.5"
+            memory: "4G"
```
danking, line 45: This is gonna increase our testing burden a fair bit. I guess that's what needs to be done to get performance, but we should go into this eyes open.

danking (changes requested):

batch/batch/driver/instance_collection/pool.py
daniel-goldstein: Good point, I'll add a label with the pool name

batch/deployment.yaml
daniel-goldstein: Ya I was thinking about this, and contemplated if we should run smaller PR deployments, but I'm not sure I see a way to drive test times down if we don't give the resources we can to PR drivers.

daniel-goldstein:
Not exactly mandatory, but should probably be stacked on #11636

gear/gear/database.py
```diff
@@ -235,10 +236,15 @@ async def execute_update(self, sql, args=None):
         async with self.conn.cursor() as cursor:
             return await cursor.execute(sql, args)
 
-    async def execute_many(self, sql, args_array):
+    async def execute_many(self, sql, args_array, query_name=None):
```
jigold, line 239: Just to be clear, we're only monitoring `execute_many`. Is there a plan to add more monitoring of queries in the future?

jigold (changes requested):
This is awesome!

gear/gear/database.py
daniel-goldstein: Right now we do the following:
1. Count the total number of sql transactions performed per second
2. Time explicitly monitored queries

In terms of timing more queries, it's currently on an as-needed basis. It felt potentially overbearing to require every query have a unique name and add timing, rather we've been targeting hot-path queries as needed to address big performance issues.

It could be nice to have a more complete sql monitoring solution in place, but I think would require some more thought to make it easy to analyze from an operational perspective.

hail/python/hailtop/utils/utils.py
```diff
@@ -837,6 +837,7 @@ async def run_if_changed(changed, f, *args, **kwargs):
     while True:
         changed.clear()
         should_wait = await f(*args, **kwargs)
+        await asyncio.sleep(0.5)
```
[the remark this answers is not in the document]
daniel-goldstein: Ya there's no rigorous reason here, just that it seemed frequent enough that it shouldn't bother anyone but infrequent enough that the looping alone isn't a bottleneck. I will make a note.

Also note that it is after the invocation, not after waiting on the event. So "cold start" time after an event is set is unaffected.

danking (approved):
I'm good with this. I slapped WIP on so that we can do the following before merging:
1. ensure the node pools are ready in both GCP and Azure?
2. ensure we've communicated to AUS et al. and received confirmation of awareness


Do you mind taking lead on the above two activities?

PR #11641, opened 2022-03-22
[query] Add nicer errors for unsupported functionality in the service

tpoterba:


danking (approved):
❤️

tpoterba:
note this doesn't add the errors everywhere, just the infrastructure.

PR #11643, opened 2022-03-22
[query/service] automatically determine JAR url based on SHA

danking:
Alright, so the goal of this PR is to make this work:

```
HAIL_QUERY_BACKEND=service \
  python3 -c 'import hail as hl; hl.utils.range_table(10).write("gs://foo/bar.t")`
```

In particular, the Hail Query JAR is stored in a well known location. If we know the desired git revision, (we do, it should be the same git revision as our wheel), then we can deduce the JAR URL for the user. Moreover, if we're pointed at a namespace, we can still determine the correct location [1].

This PR provides three escape hatches to this behavior:
1. Specify the `jar_url` parameter to `ServiceBackend`.
2. Specify the `HAIL_JAR_URL` environment variable.
3. Specify a JAR url in the user config: `hailctl config set query/jar_url gs://...`.

This PR is unfortunately entangled with one other minor change. In `main`, we send the git revision *and* the JAR URL to the driver and the worker as a part of the "command string" (the JVMEntryway passes this array of strings to the `main` method of `ServiceBackendSocketAPI2` or `Worker`. After this change, the backend does not necessarily know the git revision. That's fine. The git revision was only ever used as:
1. a cache key for the JAR cache, and
2. a unique name for the JAR
Both of these uses are buggy anyway! If you re-use a HAIL_SHA with a different HAIL_JAR_URL and you land on a worker that previously pulled that HAIL_SHA, you'll get the previously pulled JAR, not the newly specified one. Instead I use the JAR URL directly as a cache key and unique name.

---

[1] Odds are good that the developer has not uploaded a JAR to this location, but they can do so by dev deploying `upload_query_jar` or by running `make -C query push-jar`.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1578,9 +1578,8 @@ class JVMJob(Job):
             raise Exception("i/o not supported")
 
         self.user_command_string = job_spec['process']['command']
-        assert len(self.user_command_string) >= 3, self.user_command_string
-        self.revision = self.user_command_string[1]
-        self.jar_url = self.user_command_string[2]
+        assert len(self.user_command_string) >= 2, self.user_command_string
+        self.jar_url = self.user_command_string[1]
 
         self.timings = Timings()
         self.state = 'pending'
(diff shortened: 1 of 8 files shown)


danking:
Need to figure some stuff out first.

PR #11646, opened 2022-03-23
[query/service] simplify ServiceBackend by abstracting over ExecuteContext

danking:
Previously, I read the tmpdir, billingProject, and remoteTmpdir arguments
separately for each comand in its branch of the switch. Then, I called
a method for that command, passing the three aformentioned arguments. The
method then constructs an execute context. Every method constructed the
context in the exact same way. This was just naughty copy paste job on my
part.

I rectify that here by abstracting over the creation of the execute
context and moving the reading of the three execute context relevant
parameters to outside of the switch.

I did not change any functionality. This strictly a refactor.


danking:
Stacked on https://github.com/hail-is/hail/pull/11597

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -618,150 +500,143 @@ class ServiceBackendSocketAPI2(
...
+          withExecuteContext(
+            "ServiceBackend.referenceGenome",
+            backend.referenceGenome(_, name)
+          )
+        case EXECUTE =>
+          val code = readString()
+          val token = readString()
+          withExecuteContext("ServiceBackend.execute", { ctx =>
+            backend.execute(ctx, code, token)
+          })
```
daniel-goldstein, line 574: nit: inconsistent with the way all the other lambdas are written

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -547,7 +430,6 @@ class ServiceBackendSocketAPI2(
   private[this] val PARSE_VCF_METADATA = 8
   private[this] val INDEX_BGEN = 9
   private[this] val IMPORT_FAM = 10
-  private[this] val GOODBYE = 254
```
daniel-goldstein, line 550: Why is this removed?

daniel-goldstein (changes requested):
looks great, one q

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: It's not used. It was used ages ago when the query-service was an actual service that opened a TCP connection, issued several commands and then closed the connection cleanly by saying GOODBYE.

daniel-goldstein (approved):

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: missed this message but its now fixed too

PR #11648, opened 2022-03-23
[query] Use lowered logistic regression on non Spark backends as needed

johnc1231:


johnc1231:
I already had cases in tests that ran the underscore version tests if it's on the service or local backends, so tests didn't need changing.

patrick-schultz (approved):

danking (approved):

PR #11656, opened 2022-03-23
 [query] Replace ExportVCF.apply with lowered version

chrisvittal:
This removes support for parallel composable vcf export. I believe this
to be okay however as parallel composable export is undocumented and was
added for one specific internal project and may no longer be needed.


tpoterba (approved):

tpoterba:
:shipit:

PR #11662, opened 2022-03-24
Geom bar weight

ammekk:
Added support for a weight aesthetic in geom_bar. Also made small fix to geom_tile's color.

### DIFF
--- a/hail/python/test/hail/ggplot/test_ggplot.py
+++ b/hail/python/test/hail/ggplot/test_ggplot.py
@@ -88,3 +88,16 @@ def test_scale_color_manual():
     pfig = fig.to_plotly()
 
     assert set([scatter.marker.color for scatter in pfig.data]) == colors
+
+
+def test_weighted_bar():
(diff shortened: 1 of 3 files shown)


hail/python/test/hail/ggplot/test_ggplot.py
```diff
@@ -88,3 +88,16 @@ def test_scale_color_manual():
     pfig = fig.to_plotly()
 
     assert set([scatter.marker.color for scatter in pfig.data]) == colors
+
+
```
johnc1231, line 92: You need to change `geom_bar` to add the `weight` argument right? And then add documentation of what it does?

johnc1231 (changes requested):

hail/python/test/hail/ggplot/test_ggplot.py
johnc1231: I'm wrong, handled in stat

johnc1231 (approved):

PR #11665, opened 2022-03-24
[query] Support repartition on lowered backends

tpoterba:


danking (approved):

danking:
😢

tpoterba:
Should work for empty keys now.

danking (approved):
moral of the story: always have a key

danking:
We gotta implement `TableCalculateNewPartitions`

tpoterba:
this was blocked on no lowering for TableCalculateNewPartitions. Added that; testing now.

tpoterba:
This proved to be a bit of a rabbit hole with how we do sampling. Should be ready now.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -466,6 +467,65 @@ object LowerTableIR {
...
           stage.mapCollect(relationalLetsAbove)(rows => foldIR(mapIR(rows)(row => Consume(row)), 0L)(_ + _)))
 
+      case TableToValueApply(child, TableCalculateNewPartitions(nPartitions)) =>
+        val stage = lower(child)
+        val sampleSize = math.min(nPartitions * 20, 1000000)
+        val samplesPerPartition = sampleSize / stage.numPartitions
+        val keyType = child.typ.keyType
+        val samplekey = AggSignature(TakeBy(),
+          FastIndexedSeq(TInt32),
+          FastIndexedSeq(keyType, TFloat64))
```
patrick-schultz, line 477: We should be able to collect the sample faster using SeqSample instead of TakeBy, but that's not urgent. Also not sure the purpose of inserting the partition mins/maxs into the sample.

patrick-schultz (approved):

hail/python/hail/matrixtable.py
```diff
@@ -3326,6 +3326,20 @@ def repartition(self, n_partitions: int, shuffle: bool = True) -> 'MatrixTable':
...
+        if hl.current_backend().requires_lowering:
+            tmp = hl.utils.new_temp_file()
+
+            if len(self.row_key) == 0:
+                uid = Env.get_uid()
+                tmp2 = hl.utils.new_temp_file()
+                self.checkpoint(tmp2)
+                ht = hl.read_matrix_table(tmp2).add_row_index(uid).key_rows_by(uid)
+                ht.checkpoint(tmp)
+                return hl.read_matrix_table(tmp, _n_partitions=n_partitions).drop(uid)
```
patrick-schultz, line 3338: Tests failing because you can't drop a key field.

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -466,6 +467,65 @@ object LowerTableIR {
         invoke("sum", TInt64,
           stage.mapCollect(relationalLetsAbove)(rows => foldIR(mapIR(rows)(row => Consume(row)), 0L)(_ + _)))
 
+      case TableToValueApply(child, TableCalculateNewPartitions(nPartitions)) =>
+        val stage = lower(child)
+        val sampleSize = math.min(nPartitions * 20, 1000000)
+        val samplesPerPartition = sampleSize / stage.numPartitions
```
patrick-schultz: Have to guard against division by zero

danking: I stared at this for a bit and I'm not sure what we should do if there are no rows. You'll get failures when you try to pass round an empty array of intervals. For example, ir.TableNativeReader will complain.

danking:
I pushed a couple fixes for the errant tests and the drop key issue

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
danking: I feel like we should figure out how to pervasively support empty intervals, but for now maybe we just check and return an empty table in repartition?

patrick-schultz: It looks like we're getting another divide-by-zero, which I think comes from the StreamRange with stepSize=0, which happens when there are no rows. It should be easy to fix to create an empty partitioner in that case. Is there any reason TableNativeReader can't support reading with zero partitions?

danking: I'm unaware of a true reason. I might have fixed it in my [most recent commit](https://github.com/hail-is/hail/pull/11665/commits/d2eb207acd4b5053dae62ee8235bfa68ea6aaab4). We'll see what CI says.

PR #11666, opened 2022-03-24
[ci] Retry image builds

daniel-goldstein:
@johnc1231 Had this idea for dealing with registry flakiness. I was hesitant at first because really I want buildkit to retry more transient errors, but maybe with its local cache it would be quick?

### DIFF
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -9,7 +9,7 @@ import jinja2
 import yaml
 
 from gear.cloud_config import get_global_config
-from hailtop.utils import flatten
+from hailtop.utils import flatten, RETRY_FUNCTION_SCRIPT
 
 from .environment import BUILDKIT_IMAGE, CI_UTILS_IMAGE, CLOUD, DEFAULT_NAMESPACE, DOCKER_PREFIX, DOMAIN, STORAGE_URI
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
This will also retry things that are our fault, but that might just be worth it if it avoids most transient errors

ci/ci/build.py
```diff
@@ -311,9 +311,13 @@ def build(self, batch, code, scope):
 /bin/sh /home/user/convert-cloud-credentials-to-docker-auth-config
 set -x
 
+retry() {{
+    "$@" || (sleep 2 && "$@") || (sleep 4 && "$@")
+}}
+
```
danking, line 317: Totally down for this, but let's use our `RETRY_FUNCTION_SCRIPT` it is in `hailtop.utils.utils`.

danking (dismissed):

ci/ci/build.py
daniel-goldstein: I'm all for reusing that, but the retry script requires bash, which the buildkit image doesn't have. I don't think `/bin/sh` is what we want in the retry script because our images have that as `dash`. What do you think is the best way to deal with that?

danking (approved):

PR #11670, opened 2022-03-25
[query/service] enable collecting 10 million rows

danking:
1. Use encoded bytes to transfer result from Scala to Python
2. Use encoded bytes for RelationalLet literals
3. Optimize after lifting relational operations to eliminate trivial lets
4. Avoid the memory service for large jobs (eventually we need to scale memory)

### DIFF
--- a/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
@@ -303,7 +303,7 @@ class ServiceBackend(
       val retPType = pt.asInstanceOf[PBaseStruct]
       val off = f(ctx.theHailClassLoader, ctx.fs, 0, ctx.r)(ctx.r)
       val codec = TypedCodecSpec(
-        EType.fromTypeAllOptional(retPType),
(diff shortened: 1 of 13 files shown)


hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -254,7 +263,9 @@ class ServiceBackend(
     ReferenceGenome.getReference(name).toJSONString
   }
 
-  private[this] def execute(ctx: ExecuteContext, _x: IR): Option[(Annotation, PType)] = {
+  private[this] def execute(ctx: ExecuteContext, _x: IR, bufferSpecString: String): Array[Byte] = {
+    // FIXME: do we need TypeCheck(ctx, _x)?
```
johnc1231, line 267: When you `Compile` on line 272, that calls `TypeCheck`, so I think you're ok without `TypeCheck`

danking: And Validate?

tpoterba (approved):
looks good. will leave the fixme for now.

PR #11672, opened 2022-03-25
[query] warn on using randomness on lowered backends

patrick-schultz:
stacked on #11665

The new deterministic randomness design will not quite be ready by the 28th. This PR adds a warning when using seeded randomness on lowered backends. The plan is to add the new randomness semantics incrementally, and to show the warning only when an expression is constructed which still depends on the old randomness semantics, which should quickly become uncommon.


danking:
I pushed a merge with the latest version of the underlying PR.

danking (approved):

PR #11677, opened 2022-03-26
[query/service] fast path for checkpoint and persist

danking:
As much as possible, avoid network requests. In particular, we know the type of tables that are read in checkpoint and Expression.persist.


johnc1231 (approved):
Good idea

PR #11678, opened 2022-03-26
[query] Fix memory not being resolved in default

jigold:
I'm not sure if we need to redeploy anything by hand as well.

cc: @daniel-goldstein


daniel-goldstein (approved):
Shouldn't need to redeploy

danking:
Thank you both for handling this!

PR #11679, opened 2022-03-26
[build.yaml] cancel all running test batches after tests timeout/complete

danking:


danking:
I tested this by creating 100 batches with my username and then running exactly the code here but with `dking` in place of `test`. It cancelled all 100 batches quickly (a couple seconds).

chrisvittal (approved):

PR #11681, opened 2022-03-26
[query] use encoded literals instead of annotations in LowerToCDA

danking:
I separated this change from #11670 because I believe it somehow triggers encoder
issues. I am not sure what is wrong.

PR #11683, opened 2022-03-27
[query] Lowering uses encoded literals instead of java lits

tpoterba:


danking:
In my PR (#11681), I used `GetTupleElement` so that missingness propagated correctly. Should we do that here as well?

I also pasted the encoder/decoder issue uncovered by this change in #11681.

danking (changes requested):
taking off my queue, dismiss when we've resolved the encoding issue

tpoterba:
@danking this works now!

tpoterba:
(was failing due to the same encoder bug)

PR #11684, opened 2022-03-27
[query] propagate type information when we have it in BlockMatrix

danking:
Same principle as #11677.


hail/python/hail/linalg/blockmatrix.py
```diff
@@ -1,4 +1,4 @@
-import os
+nimport os
```
johnc1231, line 1: No n

johnc1231 (approved):
Approving so I don't have to look again, but there's an n to remove.

PR #11685, opened 2022-03-27
[query/service] add docs, use consistent branding

danking:
Writing the ATGU presentation made clear that we need consistent branding. I am
going with "Hail Query-on-Batch" and the "Batch backend".

### DIFF
--- a/hail/python/hail/context.py
+++ b/hail/python/hail/context.py
@@ -283,7 +283,7 @@ def init(sc=None, app_name='Hail', master=None, local='local[*]',
     backend = (
         backend
         or os.environ.get('HAIL_QUERY_BACKEND', None)
-        or get_user_config().get('query', 'backend')
+        or get_user_config().get('query', 'backend', fallback=None)
         or 'spark'
     )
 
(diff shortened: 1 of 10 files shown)


hail/python/hail/context.py
```diff
@@ -219,20 +224,22 @@ def init(sc=None, app_name='Hail', master=None, local='local[*]',
     Parameters
     ----------
     sc : pyspark.SparkContext, optional
-        Spark context. By default, a Spark context will be created.
+        Spark Backend only. Spark context. If none is specified, the Spark backend
+        will create a Spark context will be created.
```
jigold, line 228: wording here.

hail/python/hail/docs/cloud/query_on_batch.rst
```diff
@@ -0,0 +1,65 @@
...
+===================
+
+.. warning::
+
+    Hail Query-on-Batch (the Batch backend) is currently in beta. This means some functionality is
+    not yet working. Please `contact us <https://discuss.hail.is>`__ if you would like to use missing
+    functionality on Query-on-Batch!
+
+
+Hail Query-on-Batch uses Hail Batch instead of Apache Spark. Instead of a Dataproc cluster, you will
```
jigold, line 12: `Hail Query-on-Batch uses Hail Batch instead of Apache Spark to execute jobs.`

jigold (dismissed):

jigold:
Does the query Makefile need to change at all?

danking:
good point yes

jigold (approved):

PR #11652, opened 2022-03-23
[batch]memory cpu plot

CDiaz96:

PR #11690, opened 2022-03-28
Revert "[batch]memory cpu plot"

jigold:
There's some issues that need to be fixed when @CDiaz96 comes back from vacation... The units on the plot are wrong and we're not handling missing data well.

Reverts hail-is/hail#11652

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -18,10 +18,8 @@ import humanize
 import pandas as pd
 import plotly
 import plotly.express as px
-import plotly.graph_objects as go
 import pymysql
 from aiohttp import web
-from plotly.subplots import make_subplots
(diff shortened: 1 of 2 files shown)


jigold:
Screen Shot 2022-03-28 at 12 02 10 AM
Screen Shot 2022-03-28 at 12 07 14 AM

jigold:
I'm also skeptical that we're getting any actual data so we'd need to check the resource usage code as well.

daniel-goldstein (approved):

danking:
We’re definitely getting data though the displayed units are swapped between cpu and men: https://batch.hail.is/batches/1382976/jobs/65

danking:
I think you can get weird plots when there’s only one data point. We should probably hide the plots if there’s fewer than 12 points (I think that’s one minute at our sampling rate)

PR #11691, opened 2022-03-28
[query] fix init

danking:
I dropped a bunch of args to spark and accidentally prepended every Hail QoB pipeline name with "Hail"


johnc1231:
Can approve this now if you want, but really we should have a test that we respect Spark confs being passed in at least.

johnc1231 (approved):

johnc1231:
@danking Failing check_services, let's fix and release

PR #11692, opened 2022-03-28
[batch] Fix logs for running jobs

jigold:
Fixes the test where 'ERROR: could not find file' is in the message.

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -427,7 +427,7 @@ async def _get_job_log(app, batch_id, job_id) -> Optional[Dict[str, str]]:
     async def handle_running_response(resp: aiohttp.ClientResponse) -> Dict[str, str]:
         return await resp.json()
 
-    maybe_data = await _get_resource_from_record(
+    return await _get_resource_from_record(
         app,
         batch_id,
         job_id,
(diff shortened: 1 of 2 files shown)


batch/batch/front_end/front_end.py
```diff
@@ -1660,13 +1650,19 @@ async def ui_get_job(request, userdata, batch_id):
...
-    job, attempts, job_log, resource_usage = await asyncio.gather(
+    job, attempts, maybe_job_log, resource_usage = await asyncio.gather(
         _get_job(app, batch_id, job_id),
         _get_attempts(app, batch_id, job_id),
         _get_job_log(app, batch_id, job_id),
         _get_job_resource_usage(app, batch_id, job_id),
     )
 
+    job_log = {}
+    for task, log in maybe_job_log.items():
```
danking, line 1661: won't this raise an exception when maybe_job_log is None?

batch/test/test_batch.py
```diff
@@ -364,20 +364,6 @@ def test_unknown_image(client: BatchClient):
...
-
-    while True:
-        if j.status()['state'] == 'Running' or j.is_complete():
-            break
-
-    j.log()
-    # FIXME after batch1 goes away, check running status
-    b.cancel()
-
-
```
danking, line 380: Why are we removing this test?

danking (changes requested):

PR #11695, opened 2022-03-28
[batch] Upgrade crun to 1.4.4

daniel-goldstein:
Mainly bug fixes and criu support. I don't like that we're still building from source and hope to replace that soon, but Vedant will need this to start playing with checkpoint/restore.


danking (approved):
LGTM

danking:
ugh more 503s from ACR

PR #11701, opened 2022-03-29
[vds] Add option to VDS impute sex to use only the variant data

jkgoodrich:
During testing of our gnomAD v4 exome dataset we determined that there was better separation of x-ploidy values by using only the variant dataset to determine the mean DP of chromosome X rather than the reference blocks.

This PR is my attempt to add an option to the the VDS `impute_sex_chromosome_ploidy` method that will only compute the ploidy estimates using the variant dataset instead of the default option of using the reference blocks.

The code should still filter to variants that fall within the `calling_intervals`.

### DIFF
--- a/hail/python/hail/vds/methods.py
+++ b/hail/python/hail/vds/methods.py
@@ -358,14 +358,16 @@ def filter_samples(vds: 'VariantDataset', samples_table: 'Table', *,
 
 @typecheck(vds=VariantDataset,
            calling_intervals=oneof(Table, expr_array(expr_interval(expr_locus()))),
-           normalization_contig=str
+           normalization_contig=str,
+           use_variant_dataset=bool
            )
 def impute_sex_chromosome_ploidy(
         vds: VariantDataset,
         calling_intervals,
-        normalization_contig: str
+        normalization_contig: str,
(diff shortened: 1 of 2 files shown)


danking:
Assigning myself until we identify an appropriate reviewer.

tpoterba:
Julia, sorry I've been letting this sit. I'm totally fine with these changes, but what I want to sort out is the following issue -- 

Most of the VDS functions are really built to directly support gnomAD / CCDG activities, and the default function usages should reflect the standard or best practice applications that your team converges on. I think if gnomAD is going to typically use the `use_variant_dataset` mode, we should make that default. Is that the plan? Does it depend on genome/exome data?

tpoterba:
Julia, sorry I've been letting this sit. I'm totally fine with these changes, but what I want to sort out is the following issue -- 

Most of the VDS functions are really built to directly support gnomAD / CCDG activities, and the default function usages should reflect the standard or best practice applications that your team converges on. I think if gnomAD is going to typically use the `use_variant_dataset` mode, we should make that default. Is that the plan? Does it depend on genome/exome data?

tpoterba (approved):

PR #11711, opened 2022-03-30
[query] Add assertion that methods have fewer than 255 parameters

chrisvittal:
According to the JVM specification, methods may only have 255 parameters
(including this).

### DIFF
--- a/hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
+++ b/hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
@@ -488,6 +488,10 @@ class MethodBuilder[C](
   val returnTypeInfo: TypeInfo[_],
   val isStatic: Boolean = false
 ) extends WrappedClassBuilder[C] {
+  require(parameterTypeInfo.length + isStatic.toInt <= 255,
+    s"""Invalid method, methods may have at most 255 arguments, found ${parameterTypeInfo.length + isStatic.toInt}
+       |Return Type Info: $returnTypeInfo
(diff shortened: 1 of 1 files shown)


johnc1231 (approved):

tpoterba (changes requested):
can we print the type infos of the params in the message too, not just the number?

hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
```diff
@@ -488,6 +488,10 @@ class MethodBuilder[C](
   val returnTypeInfo: TypeInfo[_],
   val isStatic: Boolean = false
 ) extends WrappedClassBuilder[C] {
+  require(parameterTypeInfo.length + isStatic.toInt <= 255,
+    s"""Invalid method, methods may have at most 255 arguments, found ${parameterTypeInfo.length + isStatic.toInt}
+       |Return Type Info: $returnTypeInfo
+       |Parameter Type Info: ${parameterTypeInfo.mkString}""".stripMargin)
```
tpoterba, line 494: `mkString(", ")`? no separator is default.

tpoterba (changes requested):

hail/src/main/scala/is/hail/asm4s/ClassBuilder.scala
chrisvittal: TypeInfo's toString, uses jvm field descriptors representation, no separator makes it a JVM method descriptor parameter list.

https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3.3

PR #11715, opened 2022-03-30
[batch] Don't double log exceptions in job scheduling

daniel-goldstein:
The `log.exception` in the wrapping try/except means we log anything that raises as an error, even things like 503's and 403's from the workers which we explicitly log as info. I think we're abusing exception handling here to catch a potentially non-exceptional failure mode which is "we couldn't schedule, we need to add back those cores". Didn't rework things though, just pushed the exception logging in to the only chunk of code that wasn't already in a nested try/except.

### DIFF
--- a/batch/batch/driver/job.py
+++ b/batch/batch/driver/job.py
@@ -465,15 +465,18 @@ async def schedule_job(app, record, instance):
 
         log.info(f'schedule job {id} on {instance}: called create job')
 
-        rv = await db.execute_and_fetchone(
-            '''
-CALL schedule_job(%s, %s, %s, %s);
-''',
-            (batch_id, job_id, attempt_id, instance.name),
-            'schedule_job',
-        )
+        try:
+            rv = await db.execute_and_fetchone(
+                '''
+    CALL schedule_job(%s, %s, %s, %s);
+    ''',
+                (batch_id, job_id, attempt_id, instance.name),
+                'schedule_job',
+            )
+        except Exception:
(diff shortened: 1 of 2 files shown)


batch/batch/driver/job.py
```diff
@@ -465,15 +465,18 @@ async def schedule_job(app, record, instance):
...
+    CALL schedule_job(%s, %s, %s, %s);
+    ''',
+                (batch_id, job_id, attempt_id, instance.name),
+                'schedule_job',
+            )
+        except Exception:
+            log.exception(f'error while scheduling job {id} on {instance}')
+            raise
     except Exception:
-        log.exception(f'error while scheduling job {id} on {instance}')
```
danking, line 476: this would muffle non client response errors from the `post`.

Two thoughts:
1. the adjustment of free cores in memory should happen locally to where the free cores are incremented.
   1. The job private instances never decrement free cores anyway, so this isn't needed
   2. The pool should use a try-except to adjust the free cores (without logging) in `schedule_loop_body` right where the free cores are adjusted
2. Now this outer try-except is irrelevant. Push logging into inner try-excepts if needed (e.g. exactly as you've done with the SQL code)

danking (dismissed):

daniel-goldstein:
@danking bump

danking (approved):
OK, I think this is right, but, man, it seems odd that the job privates don't need any core manipulation...

danking:
Let's carefully watch the cluster after this lands.

PR #11719, opened 2022-03-31
[batch] Add Cancel button to Batch page

vrautela:
Add a cancel button to the /batch/batches/{batch_id} page for convenience
Manually tested  cancelling behavior


danking (approved):
love it

PR #11720, opened 2022-03-31
geom_boxplot

ammekk:
Added a geom to support boxplots.


johnc1231:
Ok, so the main problem with this PR is that it uses `precomputed`, which doesn't respect faceting right now. So it'll end up plotting the same plot on each facet since it'll just compute the overall quantiles once.

danking:
@tpoterba we'll need a champion and a reviewer for this one as well

danking:
@patrick-schultz , does this PR have a path to merging or does it require a rethink of our ggplot implementation? If the former, perhaps we could pull Iris in to finish this?

patrick-schultz:
I haven't had a chance to look into this and understand how precomputed and faceting work and interact. If Iris wants to give that some thought, I'd be happy to advise. If there is a real obstruction to doing this in the current design, it would be good for both Iris and I to understand it before starting a redesign.

ammekk:
Thats a great point
________________________________

danking:
@iris-garden reopen if this is still a relevant PR to your work

PR #11725, opened 2022-04-04
[query] Introducing Faceting with Plotting

johnc1231:
This PR introduces `facet_wrap`, which will allow creating plots based on a specified facet.

It also adds 
`     def _add_aesthetics_to_trace_args(self, trace_args, df):`
and
`     def _update_legend_trace_args(self, trace_args, legend_cache):`

two helper methods which let me clean up some of the redundant plotting work.

A `legend_cache` was introduced to make sure we put traces that should have the same legend point in a `legendgroup` together. Without it, if I draw one red line in each of 4 different subplots created by a facet, then 4 entries appear in the legend.


patrick-schultz (approved):
very nice!

PR #11732, opened 2022-04-05
Update installation commands for using homebrew

gilmorera:
`adoptopenjdk` has been deprecated.  `temurin` is now the official successor as of 2021-08-01 (Aug 01, 2021)
https://github.com/AdoptOpenJDK/homebrew-openjdk
https://formulae.brew.sh/cask/adoptopenjdk#default


tpoterba (approved):
good fix, thanks!

PR #11733, opened 2022-04-05
[query] Lower BlockMatrix to Table

johnc1231:
This isn't working yet, but I think it's pretty close. It transforms a BlockMatrix into a row major table of entries. 

Things to note:

1. I'm not respecting `partitionRanges` right now, need to do that for same behavior as unlowered. 
2. `sliceSize` 4 is arbitrary, it should be computed based on the known width of the BlockMatrix.


danking:
@patrick-schultz would you be able to adopt this PR after randomness?

danking:
@tpoterba should we keep this PR or do you have a new approach in mind for BlockMatrixToTable?

danking:
Bump. @tpoterba @patrick-schultz

PR #11734, opened 2022-04-05
[deps] update mypy to 900 series

chrisvittal:
Compilation of the typed-ast version range that mypy 780 used was broken
in python 3.10.

### DIFF
--- a/docker/requirements.txt
+++ b/docker/requirements.txt
@@ -39,7 +39,7 @@ Jinja2==3.0.3
 keyrings.alt>=3.1
 kubernetes-asyncio==19.15.1
 libsass==0.19.2
-mypy==0.780
+mypy>=0.900,<0.1000
 nest_asyncio==1.5.4
 parsimonious==0.8.1
 pandas==1.3.0
--- a/docker/requirements.txt
+++ b/docker/requirements.txt
@@ -69,3 +69,16 @@ wheel>=0.31.0
 # yarl 1.6.0 broke query string parsing in aiohttp 3.6.[012] https://github.com/aio-libs/aiohttp/issues/4972#issuecomment-700290809
 yarl<1.6.0
 zulip==0.6.3
+
+# mypy library stubs
+types-Deprecated
+types-PyMySQL
+types-PyYAML
+types-chardet
+types-decorator
+types-python-dateutil
+types-requests
+types-setuptools
+types-six
+types-tabulate
+types-urllib3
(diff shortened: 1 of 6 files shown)


daniel-goldstein (dismissed):

danking:
Thank you for making this fix Chris!

danking:
Are you running this locally?
```
(cd hail/python && python3 -m mypy --config-file ../../setup.cfg hailtop)
```

danking:
I get these errors locally with mypy 0.934

danking:
I think it's transitively checking types in the dependencies of hailtop and we probably import hail in hailtop somewhere even though we really ought not to.

danking:
Ah, the batch docs import hail

danking:
add this to the setup.cfg:
```
exclude = hailtop/batch/docs
```

daniel-goldstein:
Looks like the Azure library does something kind of lame here where `commit_block_list`'s signature accepts an argument of type `List[BlobBlock]` but they actually accept either the `BlobBlock` or the string id:

```python
    def _commit_block_list_options( # type: ignore
            self, block_list,  # type: List[BlobBlock]
            content_settings=None,  # type: Optional[ContentSettings]
            metadata=None,  # type: Optional[Dict[str, str]]
            **kwargs
        ):
        # type: (...) -> Dict[str, Any]
        if self.require_encryption or (self.key_encryption_key is not None):
            raise ValueError(_ERROR_UNSUPPORTED_METHOD_FOR_ENCRYPTION)
        block_lookup = BlockLookupList(committed=[], uncommitted=[], latest=[])
        for block in block_list:
            try:
                if block.state.value == 'committed':
                    block_lookup.committed.append(encode_base64(str(block.id)))
                elif block.state.value == 'uncommitted':
                    block_lookup.uncommitted.append(encode_base64(str(block.id)))
                else:
                    block_lookup.latest.append(encode_base64(str(block.id)))
            except AttributeError:
                block_lookup.latest.append(encode_base64(str(block)))
```

IIUC, the `str` type there is correct in how we treat the field, but I'm guessing mypy was complaining?

chrisvittal:
correct

daniel-goldstein (changes requested):
I think I might prefer then leaving the type as `str` since we control it, and add a `# type: ignore` comment on the offending line with a very brief explanation as to the Azure signature bug.

daniel-goldstein (approved):

PR #11736, opened 2022-04-05
[query/service] use authority, not host

danking:
This feels like a bug in `java.net.URI`, but who am I to upend decades of software:

```
scala> import java.net.URI
import java.net.URI

scala> new URI("gs://nd_qc/preimp_qc/GWASpy/Preimp_QC/wave1_pilot_qced.mt").getAuthority
res0: String = nd_qc

scala> new URI("gs://nd_qc/preimp_qc/GWASpy/Preimp_QC/wave1_pilot_qced.mt").getHost
res1: String = null

scala> new URI("gs://nd_qc/preimp_qc/GWASpy/Preimp_QC/wave1_pilot_qced.mt").getPort
res2: Int = -1

scala> new URI("gs://nd_qc/preimp_qc/GWASpy/Preimp_QC/wave1_pilot_qced.mt").getUserInfo()
res3: String = null

```

### DIFF
--- a/hail/src/main/scala/is/hail/io/fs/GoogleStorageFS.scala
+++ b/hail/src/main/scala/is/hail/io/fs/GoogleStorageFS.scala
@@ -44,8 +44,8 @@ object GoogleStorageFS {
     val scheme = uri.getScheme
     assert(scheme != null && scheme == "gs", (uri.getScheme, filename))
 
-    val bucket = uri.getHost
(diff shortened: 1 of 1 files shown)


danking:
@daniel-goldstein a tiny PR, I think buckets with underscores aren't parsed properly by java.net.URI. This is an ongoing issue with some buckets I've been looking at for other reasons.

PR #11740, opened 2022-04-07
Better windows support

chrisvittal:
For the `hail` python package, even if some things don't work great/at all, I think it would be nice if it at least installed on windows. `uvloop` is unsupported on windows, so I add a little logic to ensure that it's not requried on windows and the copy tool doesn't fail if it's not found.

### DIFF
--- a/hail/python/hailtop/aiotools/copy.py
+++ b/hail/python/hailtop/aiotools/copy.py
@@ -19,7 +19,7 @@ except ImportError as e:
     if not sys.platform.startswith('win32'):
         raise e
 
-    def uvloop_install(*args, **kwargs):
+    def uvloop_install():
         pass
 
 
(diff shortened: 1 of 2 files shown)


hail/python/hailtop/aiotools/copy.py
```diff
@@ -4,14 +4,24 @@
...
 from .utils import make_tqdm_listener
 
+try:
+    import uvloop
+    uvloop_install = uvloop.install
+except ImportError as e:
+    if not sys.platform.startswith('win32'):
+        raise e
+
+    def uvloop_install(*args, **kwargs):
```
chrisvittal, line 22:
```suggestion
    def uvloop_install():
```

danking (approved):

PR #11743, opened 2022-04-08
[letsencrypt] no secret logging and no blowing away secret

danking:
1. Do not leak the private key for lets encrypt into the renewal container's stdout and consequently into the logs.
2. Do not revert the secret to an empty state before renewing the certificate. Doing so causes a failed renewal (e.g. 500s from lets encrypt) to destroy the extant keys.

---

Anyone using the Hail infrastructure should both regenerate their lets encrypt certificates with the changes in that PR. To do so they can execute the following from the root of the Hail repository:

    make -C letsencrypt run

To take advantage of this vulnerability, someone would need access to the k8s container logs and the ability to redirect the relevant domain name to an IP they control. We have no evidence anyone has done this with Hail’s certs.


daniel-goldstein (approved):
Regenerated azure certs from this branch 👍

PR #11745, opened 2022-04-08
[query/service] streamline & standardize configuration for query

danking:
These changes make it straightforward to always use QoB by default.


hail/python/hail/backend/service_backend.py
```diff
@@ -135,17 +135,6 @@ def __repr__(self):
         return f'GitRevision({self.revision})'
 
 
-def _get_jar_specification(jar_url: Optional[str]) -> JarSpec:
-    user_config = get_user_config()
-
-    jar_url = jar_url or os.environ.get('HAIL_JAR_URL')
```
daniel-goldstein, line 141: I think this is the only env var which doesn't conform with the `HAIL_{SECTION}_{OPTION}` and should be `HAIL_QUERY_JAR_URL`. Seems a fine breaking change to make but I think you need to update it in Makefiles and the like.

hail/python/hail/backend/service_backend.py
```diff
@@ -230,21 +218,13 @@ async def create(*,
...
-            or os.environ.get('HAIL_QUERY_BATCH_DRIVER_MEMORY', None)
-            or get_user_config().get('query', 'batch_driver_memory', fallback=None)
-            or 'standard'
-        )
+        driver_cores = configuration_of('query', 'batch_driver_cores', driver_cores, '1')
+        driver_memory = configuration_of('query', 'batch_driver_memory', driver_memory, 'standard')
+        name_prefix = configuration_of('query', 'name_prefix', name_prefix, '')
 
         return ServiceBackend(
             billing_project=billing_project,
```
daniel-goldstein, line 230: Can remove the `name_prefix or ''` down below in the constructor.

daniel-goldstein (dismissed):
Very nice. Only one configuration that didn't seem quite right.

daniel-goldstein:
I think there's also an awkwardness here where `ServiceBackend.create` handles setting defaults for values like `driver_cores` which are ultimately optional fields on the `ServiceBackend`, and in such a scenario would get set to be the default cores for any old batch job in the batch front end. Currently these two different places that set default values are the same (1 standard core), but feels footgun-y. I feel like so long as the constructor accepts optional kwargs, resetting those kwargs should happen in the constructor and not in `ServiceBackend.create`

daniel-goldstein (approved):

PR #11749, opened 2022-04-11
[batch] Add dill support for 3.9 and 3.10

daniel-goldstein:
Python 3.6 has been EOL for over three months, and around that time we dropped support in the docs but silently kept support in the code. I'd like to drop support for 3.6 but our stance seems inconsistent on how long we support something after EOL (I would like to not ever) so I wanted to double check.

Regardless we should support new versions of python.

I pushed these into our internal registry but I don't believe I have access to dockerhub's hailgenetics, so haven't been able to push them there yet.

### DIFF
--- a/docker/python-dill/push.sh
+++ b/docker/python-dill/push.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-for version in 3.6 3.6-slim 3.7 3.7-slim 3.8 3.8-slim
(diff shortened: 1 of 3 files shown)


danking (approved):
I'm good with this change. As to EOLs, hmm, if Python is EOL, then I think we don't need to support it. Agree?

PR #11750, opened 2022-04-11
[batch] Reattempt adding resource usage plots

jigold:
Second attempt for #11368 

I'm not sure what the easiest way is to get the changes for commit 22d3644 to be attributed to Carolin instead of me.


danking:
Git has special support for coauthors in the commit message https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

hail/python/hailtop/utils/utils.py
```diff
@@ -862,8 +862,10 @@ async def periodically_call(period, f, *args, **kwargs):
     async def loop():
         log.info(f'starting loop for {f.__name__}')
         while True:
-            await f(*args, **kwargs)
-            await asyncio.sleep(period)
+            try:
+                await f(*args, **kwargs)
+            finally:
+                await asyncio.sleep(period)
```
danking, line 868: This means there's a delay before we see the error message. retry_long_running already has a delay built in. Do we need to make that more configurable?

danking (changes requested):
Let's have Daniel do a review. I'm feeling a bit overwhelmed with work right now and don't have the time to think critically about the implication of using multipart writer. Can we also recalculate the amount of disk space we anticipate using for this data and include that in the commit message? I just want to make sure we can refer back to that.

danking:
That said, the high-bit of my thinking is: this is great we should do it; just need a standard review.

hail/python/hailtop/utils/utils.py
jigold: @daniel-goldstein I thought this could be where the problem was with using a ton more disk space than expected if we reach an error and are constantly writing to the file with no sleep in between. I might be wrong that this is true and whether this is the function that needs to change.

The delay / sleep in retry_long_running is a different delay than the periodically call delay. For example, the sleep we want is 5 seconds while retry_long_running has a sleep backoff starting at 0.1 seconds up to 30 seconds.

daniel-goldstein: I think I would prefer here that we push the try/except into the `f` that we're passing to periodically_call. If we expect it to raise and don't want it to, we can silence those errors closer to the source and then not change the periodically_call semantics.

batch/batch/worker/worker.py
```diff
@@ -2423,26 +2456,46 @@ async def create_job(self, request):
...
+        resp = web.StreamResponse(
+            status=200, reason='OK', headers={'Content-Type': f'multipart/mixed;boundary={boundary}'}
+        )
+        await resp.prepare(request)
+
+        with aiohttp.MultipartWriter('mixed', boundary=boundary) as mpwriter:
+            for task, data in resource_usage.items():
+                part = mpwriter.append(data)
+                part.set_content_disposition('attachment', filename=task)
+            await mpwriter.write(resp)
```
daniel-goldstein, line 2491: I'm not really clued in on why we need multipart writer here, what's the reason for using it this way and not using a `FileResponse` or something?

daniel-goldstein (changes requested):
This looks good, I think I just need a bit more context on the file we're passing around.

batch/batch/front_end/front_end.py
```diff
@@ -352,13 +398,13 @@ async def _get_job_log_from_record(app, batch_id, job_id, record):
...
-            return await resp.json()
+            resp = await request_retry_transient_errors(client_session, 'GET', f'http://{ip_address}:5000{endpoint}')
+            return await handle_running_response(resp)
         except aiohttp.ClientResponseError:
-            log.exception(f'while getting log for {(batch_id, job_id)}')
-            return {task: 'ERROR: encountered a problem while fetching the log' for task in tasks}
+            log.exception(f'while getting resource for {(batch_id, job_id)}')
+            return {
+                task: GetResourceError('ERROR: encountered a problem while fetching the resource') for task in tasks
+            }
```
daniel-goldstein, line 407: This abstraction feels a little clunky and like it's not strictly beneficial. I think this is a situation where saving lines of code came at the sacrifice of local interpretability. As an example, let's say I see this error in the logs (which now just say "getting resource" instead of "getting log", which is less informative). If I want to understand what happened such that line number 404 triggered, I have to look at the invocation of this function to see which `endpoint` was used, then jump back here to see how it was used, then immediately jump back to the invocation to see what `handle_running_response` is. I think that this abstraction was too surgical instead of a natural way to break up responsibilities.

daniel-goldstein: I also think we should get rid of `GetResourceError`.

```
'ERROR: encountered a problem while fetching the resource'
```

This is what the user will see if we can't get the job log, which is going to be less intuitive than a message saying we couldn't get the log. I feel like this method should return `None` for any task that it can't find, and the client of this method should figure out how to interpret missing items.

hail/python/hailtop/utils/utils.py
```diff
@@ -882,12 +882,16 @@ async def run_if_changed_idempotent(changed, f, *args, **kwargs):
...
+    log.info(f'starting loop for {f.__name__}')
+    while True:
+        try:
             await f(*args, **kwargs)
+        except asyncio.CancelledError:
+            raise
+        except Exception:
+            log.exception(f'in {f.__name__}')
+        finally:
             await asyncio.sleep(period)
```
daniel-goldstein, line 894: I think this will sleep even if a `CancelledError` was raised, which I don't think we want. I also think this suffers from a similar problem of putting logic in multiple different places. I think we should push any specific error handling closer to where that error might occur, so whatever function we pass to `periodically_call`. I don't think `periodically_call` has to change here if we can properly suppress errors inside of our resource-usage collecting function.

batch/batch/worker/worker.py
daniel-goldstein: Does this not work? I feel like we should be able to just send a File as bytes and not do this ourselves.

daniel-goldstein: Also I think webkit is a safari thing, does that matter?

batch/batch/front_end/front_end.py
daniel-goldstein: I feel like this should just be broken up into separate functions: one that communicates with the worker and returns the `ClientResponse` that the callee can then process however it wants, and another which reads from cloud storage. The two callees (getting log file and getting resource file) will have to duplicate the `if state == 'running'`, but that seems a small duplication and I think will significantly help clarity as then we won't need this `handle_running_response` input.

daniel-goldstein (changes requested):

jigold:
Expected resource usage:

```
32 bytes per measurement * 12 measurements per minute * 60 minutes / 1024 = 22.5KB
```

hail/python/hailtop/utils/utils.py
jigold: I abandoned fixing periodically call at all and pushed the retry logic to the `measure` function.

jigold:
I can't leave a reply on your webkit comment. I abandoned the Multipart Reader/Writer completely so those comments are no longer relevant. Instead, I created my own response format that has the task names encoded along with the data lengths.

hail/python/hailtop/utils/utils.py
jigold: Just to be clear. The semantics I want are to never execute that function more than once every 5 seconds no matter what. This way we guarantee that there's no possible way to fill the disk with writes more than every 5 seconds.

batch/batch/front_end/front_end.py
```diff
@@ -374,46 +401,83 @@ async def _get_job_log_from_record(app, batch_id, job_id, record):
...
+
+                task_name = data[offset : offset + task_name_size].decode('utf-8')
+                assert task_name in ('input', 'main', 'output'), task_name
+                offset += task_name_size
+
+                (task_data_size,) = struct.unpack_from('>i', data, offset)
+                offset += 4
+
+                task_data = data[offset : offset + task_data_size]
+                offset += task_data_size
```
daniel-goldstein, line 458: One more idea. If it's not possible then this is fine. Can we just make the message the following JSON?

```
{
  'input': '<base64_encoded_dataframe_data>',
  'main':  '<base64_encoded_dataframe_data>',
  'output': '<base64_encoded_dataframe_data>',
}
```

It seems to me like we should be able to take whatever on-disk representation we want, base64 encode it, and stuff it in a json instead of creating our own binary format that mixes data with metadata.

batch/batch/front_end/front_end.py
```diff
@@ -351,14 +385,7 @@ async def _get_job_log_from_record(app, batch_id, job_id, record):
...
     if state == 'Running':
-        try:
-            resp = await request_retry_transient_errors(
-                client_session, 'GET', f'http://{ip_address}:5000/api/v1alpha/batches/{batch_id}/jobs/{job_id}/log'
-            )
-            return await resp.json()
-        except aiohttp.ClientResponseError:
-            log.exception(f'while getting log for {(batch_id, job_id)}')
-            return {task: 'ERROR: encountered a problem while fetching the log' for task in tasks}
+        return await handle_running_job(batch_id, job_id, tasks, client_session, ip_address)
```
daniel-goldstein, line 388: Ya I think I still find this structure quite confusing. I'd prefer not to pass in these `handle_running_job` and `handle_completed_job` pseudo-continuations and not have a function called `_get_resource_from_record`. I think we should be able to write the `get_job_log` and `get_job_resource_usage` functions without creating these handle callbacks. I'm not against breaking these pieces down into helper functions, but I don't think that these warrant passing functions in as arguments here as it requires jumping back and forth between function definitions when scanning the code linearly. Not sure if I'm making sense but happy to provide an example if I'm not.

daniel-goldstein (changes requested):

batch/batch/front_end/front_end.py
jigold: I got this approach to work. My only concern is that it could be less performant than the other method I used with a hand-rolled format. It seemed fine on my hour long test example.

@danking Do you have any intuition on whether using base64 encoded strings of the binary data frame data is a bad idea.

jigold:
Let me know if you want me to dev deploy this one last time and test the changes. I haven't done so recently.

danking:
YESSSSSSS

PR #11751, opened 2022-04-11
[ci] Add RunImage type to support cleanup only jobs

jigold:
This is intended to solve the issue with `delete_batch_instances` running before the batch deployment has been deleted. I added a "cleanup" job type to RunImage jobs. The default is the current behavior.


danking (changes requested):
Can you explain a bit more about how you'll use this to stop the batch deployment from creating new instances after the delete instances step has run?

jigold:
I don't think this design is going to work.

PR #11753, opened 2022-04-11
[batch] Use stdin for copy container specs

jigold:
https://hail.zulipchat.com/#narrow/stream/223457-Hail-Batch-support/topic/how.20to.20use.20sample.20list.20file.20as.20an.20argument.20in.20Hail.20batch/near/278603415


danking (approved):
Seems right to me.

PR #11759, opened 2022-04-12
[query] add max_iterations and tolerance parameters to all iterative regressions

danking:
CHANGELOG: `hl.logistic_regression_rows`, `hl.poisson_regression_rows`, and `hl.skat` all now support configuration of the maximum number of iterations and the tolerance.

I had to make some fixes to how we count the number of iterations. It was quite screwy. Now it should reliably report the correct number of iterations regardless of failure, non-convergence, or explosion.

This was requested [on Zulip](https://hail.zulipchat.com/#narrow/stream/127634-Feature-Requests/topic/Convergence.20issues.20with.20hl.2Elogistic_regression_rows).


tpoterba (approved):
looks good. This PR would have been so tiny if we could delete the Scala implementation today :(

PR #11760, opened 2022-04-12
[docker] Dont store the wheel tar in hailgenetics image

daniel-goldstein:
Putting the tar in its own `COPY` layer means we drag around the tar in the resultant image. This uses a read-only mount for the subsequent `RUN` command to extract the tar without adding it to the image. This should save close to 100MB on the image size.

Also realized that this happens in our pip-installed images.


daniel-goldstein:
@danking bump

danking (approved):

PR #11766, opened 2022-04-14
[batch] Create pool scheduling loops after loading existing instances

daniel-goldstein:
Currently, tasks to schedule new instances are put on the event loop inside the `Pool` and `JobPrivateInstanceManager` constructors. `Pool.create` and `JobPrivateInstanceManager.create` first instantiate an object of their respective type and then load existing instances from the database into the in-memory instance collection. This could potentially cause the create instances loop to trigger while we're drawing "existing" instances, which causes the assertion error in https://github.com/hail-is/hail-tasks/issues/24 when the create instances loop and load instances query race to add the instance to the in-memory data structure.

This change moves the task creation from the constructor to the `create` method, so we don't start creating instances until all existing instances are accounted for. I think I would have liked to simply pass the constructor a list of instances, but we can't create an `Instance` without an `InstanceCollection`.

Resolves hail-is/hail-tasks#24

### DIFF
--- a/batch/batch/cloud/azure/driver/driver.py
+++ b/batch/batch/cloud/azure/driver/driver.py
@@ -64,7 +64,7 @@ class AzureDriver(CloudDriver):
                 app['async_worker_pool'],
                 task_manager,
             )
-            for pool_name, config in inst_coll_configs.name_pool_config.items()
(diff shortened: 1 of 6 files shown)


daniel-goldstein:
@danking bump

danking (approved):
love it

PR #11770, opened 2022-04-15
[httpx] use timeouts and remove footgun

danking:
`httpx.ClientSession` did not include the logic to set up the timeout. I do
not recall why I had a factory function distinct from the class itself. That
was a mistake. I leave the factory function because it is used prevasively.
However, now, allocating the class itself is also OK.

### DIFF
--- a/hail/python/hailtop/httpx.py
+++ b/hail/python/hailtop/httpx.py
@@ -83,9 +83,34 @@ class ClientResponse:
 
 
 class ClientSession:
-    def __init__(self, *args, **kwargs):
-        self.raise_for_status = kwargs.pop('raise_for_status', False)
-        self.client_session = aiohttp.ClientSession(*args, raise_for_status=False, **kwargs)
+    def __init__(self,
+                 *args,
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/httpx.py
```diff
@@ -83,9 +83,34 @@ async def __aexit__(
...
+            tls = external_client_ssl_context()
+        elif location == 'k8s':
+            tls = internal_client_ssl_context()
+        else:
+            assert location in ('gce', 'azure')
+            # no encryption on the internal gateway
+            tls = external_client_ssl_context()
+
+        assert 'connector' not in kwargs
+        kwargs['connector'] = aiohttp.TCPConnector(ssl=tls)
```
daniel-goldstein, line 102: Nit: a little awkward here putting `connector` back in `kwargs` but not `timeout`. Can we leave `kwargs` as it is and pass `connector` explicitly to `ClientSession`?

daniel-goldstein (dismissed):

daniel-goldstein (approved):

PR #11771, opened 2022-04-15
[build.yaml] Allow numSplits in RunImageStep

daniel-goldstein:
I haven't tested this yet because I wanted to see if you liked the idea. Not as flexible as a python loop, but possibly the easiest way to get rid of *a lot* of yaml.

### DIFF
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -497,7 +497,7 @@ class RunImageStep(Step):
                 for i in range(self.num_splits)
             ]
 
-    def _build_job(self, batch, code, scope, name, env):
+    def _build_job(self, batch, code, scope, name, env, output_prefix):
         template = jinja2.Template(self.script, undefined=jinja2.StrictUndefined, trim_blocks=True, lstrip_blocks=True)
         rendered_script = template.render(**self.input_config(code, scope))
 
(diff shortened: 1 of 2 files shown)


danking:
This seems like a good interim solution on the way to a `build.py`. Maybe just allow any RunImage step to be a VectorRunImage step though? We control these env vars and can change them if necessary.

daniel-goldstein:
Ah, totally thought `PYTEST_SPLITS` was a pytest env var. In that case I can add a generic index to RunImages

danking:
It's sadly duplicated in each module's conftest, but yeah this is a classic Tim Python Hack. https://github.com/hail-is/hail/blob/main/batch/test/conftest.py

daniel-goldstein:
Ok so something like this I guess. Happy to change the env variable name to something more generic if you think it appropriate.

danking (changes requested):
hmm.... Yeah this seems about right. I think we should change the name so we can use it for test_hail_java_n as well. Those functions just read a particular "split" file whose indices run from 0 to N. I guess if we use HAIL_RUN_IMAGE_SPLITS and HAIL_RUN_IMAGE_SPLIT_INDEX, that's good. And then we can fix the three conftest's to use that. And the test_hail_java_n jobs will use it in the filename for `testng-splits-$HAIL_RUN_IMAGE_SPLIT_INDEX.xml`.

danking:
BTW, this is an extremely awesome change. I didn't think it would be so simple!

daniel-goldstein:
> BTW, this is an extremely awesome change. I didn't think it would be so simple!

I am simply here for the negative LoC points

danking (approved):

PR #11776, opened 2022-04-18
[copy] make --verbose show a progress bar

danking:
CHANGELOG: `hailtop.aiotools.copy` will always show a progress bar when `--verbose` is passed.


danking:
When I run copy inside a batch job to copy some files, it's really annoying to get no feedback at all.

daniel-goldstein (approved):

PR #11777, opened 2022-04-19
[query] Add worker_cores and worker_memory options to hl.init when using batch

daniel-goldstein:
CHANGELOG: `hl.init` now supports `worker_cores` and `worker_memory` options when using Hail Batch backend.

### DIFF
--- a/hail/python/test/hail/backend/test_service_backend.py
+++ b/hail/python/test/hail/backend/test_service_backend.py
@@ -33,7 +33,7 @@ def test_big_driver_has_big_memory():
 @skip_unless_service_backend()
 def test_tiny_worker_has_tiny_memory():
     try:
-        t = hl.utils.range_table(1).annotate(nd=hl.nd.ones((30_000, 30_000)))
+        t = hl.utils.range_table(1, n_partitions=2).annotate(nd=hl.nd.ones((30_000, 30_000)))
         t = t.annotate(nd_sum=t.nd.sum())
         t.aggregate(hl.agg.sum(t.nd_sum))
     except Exception as exc:
--- a/hail/python/test/hail/backend/test_service_backend.py
+++ b/hail/python/test/hail/backend/test_service_backend.py
@@ -50,7 +50,7 @@ def test_big_worker_has_big_memory():
     try:
         backend.worker_cores = 8
         backend.worker_memory = 'highmem'
-        t = hl.utils.range_table(1).annotate(nd=hl.nd.ones((30_000, 30_000)))
+        t = hl.utils.range_table(1, n_partitions=2).annotate(nd=hl.nd.ones((30_000, 30_000)))
         t = t.annotate(nd_sum=t.nd.sum())
         # We only eval the small thing so that we trigger an OOM on the worker
         # but not the driver or client
--- a/hail/python/test/hail/backend/test_service_backend.py
+++ b/hail/python/test/hail/backend/test_service_backend.py
@@ -1,6 +1,7 @@
 import hail as hl
 
 from ..helpers import skip_unless_service_backend
+from hail.backend.service_backend import ServiceBackend
 
 @skip_unless_service_backend()
 def test_tiny_driver_has_tiny_memory():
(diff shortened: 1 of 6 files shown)


daniel-goldstein:
Tested only by hand. IIUC the current driver options are tested by Hail Query tests that need them? Should we do the same here?

danking (dismissed):
This looks right to me, but let's indeed add a test. See `test_service_backend.py`. We'll want to do something like allocate a huge string per-row. Something big enough to blow the 3.75 GB memory limit.

danking (changes requested):
This seems to have hit a timeout. Can you try running this on default from your laptop?  Maybe something isn't actually blowing up as we expect and is instead causing JVM memory thrashing.

hail/python/test/hail/backend/test_service_backend.py
```diff
@@ -9,20 +10,51 @@ def test_tiny_driver_has_tiny_memory():
...
+    old_driver_cores = backend.driver_cores
+    old_driver_memory = backend.driver_memory
+    try:
+        backend.worker_cores = 8
+        backend.worker_memory = 'highmem'
+        t = hl.utils.range_table(1).annotate(nd=hl.nd.ones((30_000, 30_000)))
+        t = t.annotate(nd_sum=t.nd.sum())
+        # We only eval the small thing so that we trigger an OOM on the worker
+        # but not the driver or client
+        hl.eval(t.aggregate(hl.agg.sum(t.nd_sum)))
```
danking, line 57: you need not use `hl.eval`, `t.aggregate` returns a Python value.

danking (changes requested):
That sounds like it worked then, right? I'm disappointed by how slow it is, but is there a reason to not merge this?

danking (approved):

hail/python/test/hail/backend/test_service_backend.py
```diff
@@ -9,20 +10,51 @@ def test_tiny_driver_has_tiny_memory():
...
+    old_driver_cores = backend.driver_cores
+    old_driver_memory = backend.driver_memory
+    try:
+        backend.worker_cores = 8
+        backend.worker_memory = 'highmem'
+        t = hl.utils.range_table(1).annotate(nd=hl.nd.ones((30_000, 30_000)))
+        t = t.annotate(nd_sum=t.nd.sum())
+        # We only eval the small thing so that we trigger an OOM on the worker
+        # but not the driver or client
+        t.aggregate(hl.agg.sum(t.nd_sum))
```
danking, line 57: @daniel-goldstein , Tim recently added an optimization that uses the driver machine if there's exactly one partition. In this instance, that violates your expectations about what a worker is.

This all feels a bit messy, but the time savings of running a single partition job on the driver seems worth the mess. If you switch this to `hl.utils.range_table(2, n_partitions=2)` the test will pass.

daniel-goldstein: Thanks! That saved me a world of confusion

daniel-goldstein:
@jigold this should be simple. Last thing Dan wanted was that I fix the test

jigold (dismissed):

hail/python/test/hail/backend/test_service_backend.py
```diff
@@ -9,20 +10,51 @@ def test_tiny_driver_has_tiny_memory():
...
+        assert False
+
+@skip_unless_service_backend()
+def test_big_worker_has_big_memory():
+    backend = hl.current_backend()
+    assert isinstance(backend, ServiceBackend)
+    old_driver_cores = backend.driver_cores
+    old_driver_memory = backend.driver_memory
+    try:
+        backend.worker_cores = 8
```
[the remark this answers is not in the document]
daniel-goldstein: `hl.current_backend()` is shared with subsequent test methods, so I save the driver cores and memory so that we can reset those values in the `finally` block to whatever they were before we ran the test.

jigold (approved):

PR #11778, opened 2022-04-19
[copy] finally eliminate all stalls from the copy tool

danking:
No changelog because the copy tool is not properly public, though
maybe it can be public now.

The main problem was that we were saturating our network bandwidth
with data we did not need. Why? As far as I can tell, `release` on
an `aiohttp.ClientResponse` is insufficient to stop the receipt of
more bytes.

I realized one proper fix was to simply tell the cloud storage vendors
how much data we wanted to receive. That is the change to `open_from`
which I have made pervasively to all clouds.

I also changed `release` to `close` because that seems more correct.
I do not understand why we only `release`d it before.

### DIFF
--- a/hail/python/hailtop/aiocloud/aioaws/fs.py
+++ b/hail/python/hailtop/aiocloud/aioaws/fs.py
@@ -307,13 +307,16 @@ class S3AsyncFS(AsyncFS):
         except self._s3.exceptions.NoSuchKey as e:
             raise FileNotFoundError(url) from e
 
-    async def open_from(self, url: str, start: int) -> ReadableStream:
+    async def open_from(self, url: str, start: int, *, length: Optional[int] = None) -> ReadableStream:
         bucket, name = self.get_bucket_and_name(url)
+        range_str = f'bytes={start}-'
+        if length is not None:
+            range_str += str(start + length)
         try:
             resp = await blocking_to_async(self._thread_pool, self._s3.get_object,
                                            Bucket=bucket,
                                            Key=name,
-                                           Range=f'bytes={start}-')
+                                           Range=range_str)
             return blocking_readable_stream_to_async(self._thread_pool, cast(BinaryIO, resp['Body']))
         except self._s3.exceptions.NoSuchKey as e:
             raise FileNotFoundError(url) from e
--- a/hail/python/hailtop/aiocloud/aioaws/fs.py
+++ b/hail/python/hailtop/aiocloud/aioaws/fs.py
@@ -307,10 +307,11 @@ class S3AsyncFS(AsyncFS):
         except self._s3.exceptions.NoSuchKey as e:
             raise FileNotFoundError(url) from e
 
-    async def open_from(self, url: str, start: int, *, length: Optional[int] = None) -> ReadableStream:
+    async def _open_from(self, url: str, start: int, *, length: Optional[int] = None) -> ReadableStream:
         bucket, name = self.get_bucket_and_name(url)
         range_str = f'bytes={start}-'
         if length is not None:
+            assert length >= 1
             range_str += str(start + length - 1)
         try:
             resp = await blocking_to_async(self._thread_pool, self._s3.get_object,
--- a/hail/python/hailtop/aiocloud/aioaws/fs.py
+++ b/hail/python/hailtop/aiocloud/aioaws/fs.py
@@ -311,7 +311,7 @@ class S3AsyncFS(AsyncFS):
         bucket, name = self.get_bucket_and_name(url)
         range_str = f'bytes={start}-'
         if length is not None:
-            range_str += str(start + length)
+            range_str += str(start + length - 1)
         try:
             resp = await blocking_to_async(self._thread_pool, self._s3.get_object,
                                            Bucket=bucket,
--- a/hail/python/hailtop/aiocloud/aioaws/fs.py
+++ b/hail/python/hailtop/aiocloud/aioaws/fs.py
@@ -14,6 +14,7 @@ import boto3
 from hailtop.utils import blocking_to_async
 from hailtop.aiotools.fs import (FileStatus, FileListEntry, ReadableStream, WritableStream, AsyncFS,
                                  AsyncFSURL, MultiPartCreate, FileAndDirectoryError)
+from hailtop.aiotools.fs.exceptions import UnexpectedEOFError
 from hailtop.aiotools.fs.stream import (
     AsyncQueueWritableStream,
     async_writable_blocking_readable_stream_pair,
@@ -321,6 +322,10 @@ class S3AsyncFS(AsyncFS):
             return blocking_readable_stream_to_async(self._thread_pool, cast(BinaryIO, resp['Body']))
         except self._s3.exceptions.NoSuchKey as e:
             raise FileNotFoundError(url) from e
+        except botocore.exceptions.ClientError as e:
+            if e.response['Error']['Code'] == 'InvalidRange':
+                raise UnexpectedEOFError from e
+            raise
 
     async def create(self, url: str, *, retry_writes: bool = True) -> S3CreateManager:  # pylint: disable=unused-argument
         # It may be possible to write a more efficient version of this
--- a/hail/python/hailtop/aiocloud/aioazure/fs.py
+++ b/hail/python/hailtop/aiocloud/aioazure/fs.py
@@ -343,6 +343,8 @@ class AzureAsyncFS(AsyncFS):
         return stream
 
     async def open_from(self, url: str, start: int, *, length: Optional[int] = None) -> ReadableStream:
+        if length == 0:
+            raise ValueError(f'Azure Blob Storage does not suport length=0.')
         client = self.get_blob_client(url)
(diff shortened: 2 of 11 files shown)


hail/python/hailtop/aiocloud/aioaws/fs.py
```diff
@@ -281,13 +281,16 @@ async def open(self, url: str) -> ReadableStream:
         except self._s3.exceptions.NoSuchKey as e:
             raise FileNotFoundError(url) from e
 
-    async def open_from(self, url: str, start: int) -> ReadableStream:
+    async def open_from(self, url: str, start: int, *, length: Optional[int] = None) -> ReadableStream:
         bucket, name = self._get_bucket_name(url)
+        range_str = f'bytes={start}-'
+        if length is not None:
+            range_str += str(start + length)
```
jigold, line 288: Can we double check the definitions of the range are correct for all clouds? I vaguely recall Google might have different semantics and the end point is inclusive. 

https://cloud.google.com/storage/docs/samples/storage-download-byte-range#storage_download_byte_range-python

hail/python/hailtop/aiocloud/aioazure/fs.py
```diff
@@ -156,7 +157,7 @@ async def read(self, n: int = -1) -> bytes:
 
         if n == -1:
             try:
-                downloader = await self._client.download_blob(offset=self._offset)
+                downloader = await self._client.download_blob(offset=self._offset, length=self._length)
```
jigold, line 160: I can't remember exactly, but I might have run into issues with Azure when the file is empty and you ask for 0 bytes.

hail/python/hailtop/httpx.py
```diff
@@ -38,8 +38,8 @@ class ClientResponse:
     def __init__(self, client_response: aiohttp.ClientResponse):
         self.client_response = client_response
 
-    def release(self) -> None:
-        return self.client_response.release()
+    async def release(self) -> None:
```
jigold, line 41: What is the reason for this change? I looked at `aiohttp.ClientResponse` and `release` is not async.

hail/python/test/hailtop/inter_cloud/test_fs.py
```diff
@@ -115,6 +115,20 @@ async def test_open_from(filesystem: Tuple[asyncio.Semaphore, AsyncFS, str]):
...
+@pytest.mark.asyncio
+async def test_open_from_with_length(filesystem: Tuple[asyncio.Semaphore, AsyncFS, str]):
+    sema, fs, base = filesystem
+
+    file = f'{base}foo'
+
+    async with await fs.create(file) as f:
+        await f.write(b'abcde')
+
+    async with await fs.open_from(file, 2, 2) as f:
```
jigold, line 127: Can you add a test with an empty file?

jigold (changes requested):
Rebase needed as well.

hail/python/hailtop/aiocloud/aioaws/fs.py
danking: Yeah, you're totally right it's inclusive. I'll fix.
- https://cloud.google.com/storage/docs/xml-api/reference-headers#range
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax
- https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35

hail/python/hailtop/aiocloud/aioazure/fs.py
danking: Totally right. I can add a helpful error message for azure.

```
HttpResponseError: The range specified is invalid for the current size of the resource.
RequestId:314c0805-301e-001f-64f6-547b9b000000
Time:2022-04-20T20:38:34.1301964Z
ErrorCode:InvalidRange
```

hail/python/hailtop/httpx.py
danking: release is async: https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.release

I returns "noop" which is a generator which makes the function async.

hail/python/test/hailtop/inter_cloud/test_fs.py
danking: Sure.

hail/python/hailtop/aiocloud/aioazure/fs.py
danking: Hrm. that's annoying to have one give an error, so I'll just have it return an empty readable.

danking:
Actually, I decided to just properly implement truncated files in local fs too. I can't use `.truncate` because I'm fairly certain that literally truncates the file on disk.

hail/python/test/hailtop/inter_cloud/test_fs.py
```diff
@@ -175,7 +225,29 @@ async def test_read_range(filesystem: Tuple[asyncio.Semaphore, AsyncFS, str]):
...
+
+@pytest.mark.asyncio
+async def test_read_range_end_exclusive_empty_file(filesystem: Tuple[asyncio.Semaphore, AsyncFS, str]):
+    sema, fs, base = filesystem
+
+    file = f'{base}foo'
+
+    await fs.write(file, b'')
+
+    assert await fs.read_range(file, 0, 0, end_inclusive=False) == b''
```
jigold, line 239: What happens if `end_inclusive=True`? Error?

hail/python/hailtop/aiotools/local_fs.py
```diff
@@ -110,6 +110,84 @@ def __str__(self) -> str:
...
+        return self.bio.fileno()
+
+    def flush(self) -> None:
+        raise NotImplementedError
+
+    def isatty(self) -> bool:
+        return self.bio.isatty()
+
+    def read(self, n: int = -1):
+        assert self.n <= self.limit
```
jigold, line 153: I find `self.n` confusing. Can we name it `self.offset`?

hail/python/hailtop/aiotools/local_fs.py
```diff
@@ -110,6 +110,84 @@ def __str__(self) -> str:
...
+
+        if n == -1:
+            n = self.limit - self.n
+        else:
+            n = min(self.limit - self.n, n)
+        b = self.bio.read(self.limit - self.n)
```
jigold, line 159: shouldn't this be `b = self.bio.read(n)`?

jigold (changes requested):
There's also CI errors for check_services etc.

jigold:
Bump.

hail/python/test/hailtop/inter_cloud/test_fs.py
danking: That will try to read the zeroth byte from an empty file and will trigger an error. I've added a new test for that case, `test_read_range_end_inclusive_empty_file_should_error`. In Google, you get
```
hailtop.httpx.ClientResponseError: 416, message='Requested range not satisfiable'
```
In local, you get:
```
hailtop.aiotools.fs.exceptions.UnexpectedEOFError
```

danking: Hmm. I suppose we should catch and convert these all to UnexpectedEOF errors

danking: OK, all now UnexpectedEOFError and a test for that.

hail/python/hailtop/aiotools/local_fs.py
danking: yes

danking:
Let's try to get this merged today, I don't have meetings so I can respond quickly to changes

hail/python/hailtop/aiotools/fs/fs.py
```diff
@@ -108,7 +108,8 @@ async def open(self, url: str) -> ReadableStream:
 
     async def open_from(self, url: str, start: int, *, length: Optional[int] = None) -> ReadableStream:
         if length == 0:
-            isfile, isdir = await asyncio.gather(self.isfile(url), self.isdir(url))
+            assert not url.endswith('/')
```
jigold, line 111: Was this not working correctly before? I'm not sure I like this change. What happens if the file ends in a slash?

Ah I see. We can't distinguish an empty file that's a directory or a file?

jigold (changes requested):

hail/python/hailtop/aiotools/fs/fs.py
danking: This is because `isdir` fails an assert if the url doesn't end with a slash. I kind of hate that we assert that, but it seems to be a pervasive assumption. If that's the assumption, assuming we never do `open_from('gs://bucket/foo/')` also seems reasonable.

jigold (approved):

danking:
The assertion causes a bunch of tests to fail which expect to get an `IsADirectoryError`. I removed the assertion and just construct two urls: file_url and dir_url which don't (resp do) end in a slash.

PR #11782, opened 2022-04-20
[query] import_table: simplify code and improve speed

danking:
CHANGELOG: `hl.import_table` is up to twice as fast for small tables.

The big change is optimizing for the single file, no filters case in which
we need not scan for the first extant row, that row *must* be in the first
partition, if it exists at all. Unfortunately there is no zero-RPC way to
determine the number of partitions in a table, so I must catch an error
about the lack of a zeroth partition.

I also did some refactoring:

1. Move some functions to a utility file and add lots of indents and newlines to make them readable.
2. Use `hl.format` for constructing strings.
3. Make `should_filter_line` into `should_remove_line` for clarity of name.
4. Modify `should_remove_line` to use short-circuiting and/or instead of array folds.
5. Modify `should_remove_line` to indicate (via returning None) when there are no filters enabled.
6. Add types.
7. Fix a bug where we assumed that `.collect()[0]` would be `None` if there were no values in the table. (It raises an error)
8. Deduplicate `hail.utils.deduplicate` (haha: I mean, there is already code for doing field dedupe)


chrisvittal (approved):
LGTM

PR #11783, opened 2022-04-20
[query/shuffler] bind CDA to ensure it is run once

danking:
This seems to me like a footgun. Perhaps the methods in LowerTableIR should all
bind the CDA before handing it to the body? That will substantially change the
structure of some of the writers which assume that they receive an IR which
represents the execution of CDA which they place inside a
`RelationalWriter.scoped`. I took that approach at first and worried that it was
too large a change.

This change reduces one particular pipeline I was investigating from ~20 stages
to ~5 stages.


danking:
Closing in favor of #11786.

PR #11784, opened 2022-04-20
[query/shuffler] somewhat tune branching factor to dataset & log phases

danking:
For example, a tiny table with one partition really ought not to use 64
branches. In the future, we should track the byte-size of partitions and
just use a local sort for one partition tables.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/lowering/LowerDistributedSort.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/lowering/LowerDistributedSort.scala
@@ -109,7 +109,13 @@ object LowerDistributedSort {
 
     val oversamplingNum = 3
     val seed = 7L
-    val defaultBranchingFactor = ctx.getFlag("shuffle_max_branch_factor").toInt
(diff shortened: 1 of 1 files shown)


tpoterba (changes requested):
One of my biggest concerns is that we cannot judge size simply from number of partitions. Consider the case of a VCF imported with `force_gz=True` -- it could be arbitrarily large, but with a single partition. Once we write the initial input in the shuffler, we can track byte sizes and make decisions about branching. We can't make decisions before then.

For instance, this code:

```scala
    val defaultBranchingFactor = if (inputStage.numPartitions < maxBranchingFactor) {
      Math.max(2, inputStage.numPartitions)
    } else {
      maxBranchingFactor
    }
```
    
will do something pretty bad in the case above.

danking:
The bad thing being that it will have log_2 layers rather than log_64 layers, right?

The real issue here is that for tables with one extremely tiny partition (the usual in our tests), this creates 64 nearly no-op jobs, which has unavoidable cost.

Is there any number less than 64 that you would accept? Maybe 8?

PR #11787, opened 2022-04-20
[ci] Use numSplits instead of copy pasting test steps

daniel-goldstein:
The diff is horrendous but I just deleted the copy-pasted steps 1-n and changed the zero'th to be the new step.

Also fixed a bug in the brand new vectorized RunImage where the job name wasn't getting the index suffix.


build.yaml
```diff
@@ -1007,143 +1008,7 @@ steps:
...
-      - from: /resources.tar.gz
-        to: /io/resources.tar.gz
-      - from: /hail-debug-test.jar
-        to: /io/hail-test.jar
-      - from: /splits.tar.gz
-        to: /io/splits.tar.gz
-    outputs:
-      - from: /io/test-output
-        to: /test-output-4
+        to: /test-output
```
danking, line 1011: Hmm this seems incorrect. All the test outputs will race to overwrite one another. I don’t think we care about the outputs though. Just remove them? Or add a template variable to outputs?

danking (dismissed):
Minor point. Looks great!

build.yaml
daniel-goldstein: It's not exactly explicit so happy to hear alternatives

danking: Ah. Hmm. A bit confusing. I don't have a better alternative though.

danking (approved):

PR #11788, opened 2022-04-21
[hailtop.aiotools.diff] add hailtop.aiotools.diff

danking:
Recursively compare two paths searching for:
1. missing files
2. files with different sizes

The output may be sent directly to hailtop.aiotools.copy to eliminate
the differences. For example:

```
python3 -m hailtop.aiotools.diff gs://abc/ s3://def/ | tee diff.json
cat diff.json | python3 -m hailtop.aiotools.copy 'null' -
```


danking:
I also included a few minor fixes to S3AsyncFS and a typing addition to tqdm.

hail/python/hailtop/aiocloud/aioaws/fs.py
```diff
@@ -275,11 +276,14 @@ def __str__(self) -> str:
 class S3AsyncFS(AsyncFS):
     schemes: Set[str] = {'s3'}
 
-    def __init__(self, thread_pool: Optional[ThreadPoolExecutor] = None, max_workers: Optional[int] = None):
+    def __init__(self, thread_pool: Optional[ThreadPoolExecutor] = None, max_workers: Optional[int] = None, *, max_pool_connections: int = 10):
```
daniel-goldstein, line 279: Not for this PR, but it could be worth a discussion on making optional arguments keyword-only before making this API "public"

daniel-goldstein (changes requested):
This is neat, but I think I take a bit of issue with size-based `diff`. Sure the etags aren't the same? I thought they were the MD5sum of the object but they could be slightly different.

daniel-goldstein (dismissed):

hail/python/test/hailtop/inter_cloud/test_diff.py
```diff
@@ -0,0 +1,148 @@
...
+
+    expected = [
+        {'from': f'{src_base}diff', 'to': f'{dest_base}diff', 'from_size': 3, 'to_size': 1},
+        {'from': f'{src_base}src-only', 'to': f'{dest_base}src-only', 'from_size': 3, 'to_size': None},
+        {'from': f'{src_base}a/diff', 'to': f'{dest_base}a/diff', 'from_size': 3, 'to_size': 1},
+        {'from': f'{src_base}a/src-only', 'to': f'{dest_base}a/src-only', 'from_size': 3, 'to_size': None},
+        {'from': f'{src_base}b/diff', 'to': f'{dest_base}b/diff', 'from_size': 3, 'to_size': 1},
+        {'from': f'{src_base}b/src-only', 'to': f'{dest_base}b/src-only', 'from_size': 3, 'to_size': None},
+    ]
+    actual = diff(source=src_base, target=dest_base)
```
daniel-goldstein, line 124: I think you just need to await these diffs

danking (commented):
@daniel-goldstein I needed to lift frozendict into hailtop to fix the tests. Can you take a look and make sure you're good with that?

danking:
bump

daniel-goldstein:
@danking Fine by me but I think you forgot to git add hailtop/frozendict?

danking:
pushed

PR #11790, opened 2022-04-25
[docker] move pyspark installation into hail-run image

daniel-goldstein:
This PR achieves a couple of things in preparation for #11617, which requires that QoB jars be built against Spark 3.2. QoB doesn't actually *use* spark, but it will use the Azure Blob Storage client library which requires a version of jackson that conflicts with that of pyspark 3.1.1. Until such a day that we can remove pyspark from the QoB environment, we upgrade it to appease our java dependencies. Non-QoB builds of hail query remain on Spark 3.1.1/3.0.1 as it is what dataproc/hdinsight support, on which we don't need to use the Azure storage fs.

This PR moves the installation of pyspark out of the base image so we can make a spark 3.2 image on which to run the scala fs suite. We also build a spark 3.2 jar to be used in that test.

Also update a couple of dependencies to be compatible with spark 3.2 and stop pulling in jackson from the gcs dependency, which just so happens to match spark 3.1


danking (approved):
LGTM

PR #11797, opened 2022-04-26
[docker] Build hailtop-slim image with just hailtop

daniel-goldstein:
This creates another hailgenetics image with just hailtop and its python dependencies installed (no hail query, no java, no gcloud), primarily useful for using the copy tool. In contrast to hailgenetics/hail's 2.89GB on disk, hailgenetics/hailtop-slim is 289MB.


hail/python/hailtop/batch/backend.py
```diff
@@ -27,8 +27,8 @@
 from .globals import DEFAULT_SHELL
 
 
-HAIL_GENETICS_HAIL_IMAGE = os.environ.get('HAIL_GENETICS_HAIL_IMAGE',
-                                          f'hailgenetics/hail:{pip_version()}')
+HAIL_GENETICS_HAILTOP_SLIM_IMAGE = os.environ.get('HAIL_GENETICS_HAILTOP_SLIM_IMAGE',
+                                                  f'hailgenetics/hail:{pip_version()}')
```
danking, line 31: Shouldn't we use `hailgenetics/hailtop-slim` here?

danking, line 31: I remember we had an issue where, when I added `hailgenetics/hail` as the, I think, cleanup image, that broke the Australians. Can you dig that up in Zulip and make sure this change won't affect them? I think the issue might be related to how we handle the transaction from "public" images like `hailgenetics/hailtop-slim` to GCR images.

danking (changes requested):

hail/python/hailtop/requirements.txt
```diff
@@ -0,0 +1,13 @@
+aiohttp==3.8.1
```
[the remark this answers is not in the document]
daniel-goldstein: Hmph ya I think we should use these requirements for the python FS tests

hail/python/hailtop/batch/backend.py
daniel-goldstein: They would indeed need to manually deploy this image into their GCR.

jigold (dismissed):
This looks fine to me as long as we communicate clearly with the Australians.

hail/Makefile
```diff
@@ -301,8 +301,10 @@ install-dev-deps:
 install-deps: install-dev-deps
 	sed "s/^pyspark.*/pyspark==$(SPARK_VERSION)/" python/requirements.txt | xargs $(PIP) install -U
 
-.PHONY: generate-pip-lockfile
-generate-pip-lockfile:
+.PHONY: generate-pip-lockfiles
```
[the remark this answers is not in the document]
daniel-goldstein: No, and consistency is probably more important. I'll revert this edit.

jigold (dismissed):

jigold:
I think you need to rebase again.

jigold:
Does this need to be incorporated some how into the bootstrapping process?

jigold (changes requested):
See comment about bootstrapping.

daniel-goldstein:
This has diverged significantly and contains parts of some changes to handling hailgenetics images that was handled in separate PRs. Closing in favor of #12446 which is much simpler.

PR #11798, opened 2022-04-26
[batch] Compare and reject stale driver config changes

vrautela:
Currently, multiple users updating the pool config on the pool pages on the batch-driver website can have unexpected behavior. For instance, if two users simultaneously go to the batch-driver website and update the pool config one after the other, the net result is that the second config update will override the first with neither user knowing that two updates happened. This PR addresses this issue by ensuring that the current pool config that a client sees is consistent with what the pool config that the server has, before processing an update.

- Add hidden input to pool.html to store the current client config json
- Change config-update route to get current client pool config and proposed pool config from POST request body
- Compare current client config with current server config and if they're different, then reject the proposed pool config


docker/hail-ubuntu/Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN chmod 755 /bin/retry && \
          >> /etc/apt/sources.list && \
     echo 'deb-src [signed-by=/usr/share/keyrings/deadsnakes-ppa-archive-keyring.gpg] http://ppa.launchpad.net/deadsnakes/ppa/ubuntu focal main' \
          >> /etc/apt/sources.list && \
-    hail-apt-get-install python3.7-minimal python3.7-dev python3-pip && \
+    hail-apt-get-install python3.7-minimal python3.7-dev python3-pip python3.7-distutils && \
```
jigold, line 23: just curious. Why was this change necessary?

jigold (dismissed):
This looks great! Can you make a separate PR to do the same for job private instances?

jigold:
Also, just to make sure, you tested this with dev deploy?

docker/hail-ubuntu/Dockerfile
daniel-goldstein: Interesting that this is showing up in GitHub, this `distutils` fix is already in main and unrelated, I just gave Vedant the patch so that my PR wouldn't hold up development on this branch.

PR #11802, opened 2022-04-27
[batch] Only remove a job from the worker's dictionary after delete 

daniel-goldstein:
Currently, when a job-deletion event arrives on a worker, it removes the job from its running list of jobs and enqueues a future to run that job's deletion/cleanup. If a deletion removes all running jobs from a worker, and the worker does not receive new work, it could potentially idle out and shutdown before the job has finished cleanup in another task. One specific example of where this could go wrong is that worker shutdown can delete the compute client that the job's cleanup step needs to detach/delete a disk.

The worker should never shut down while operations are still in progress, so we should only ever remove a job once its deletion has completed.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -2463,15 +2463,18 @@ class Worker:
         job_id = int(request.match_info['job_id'])
         id = (batch_id, job_id)
 
+        if id not in self.jobs:
+            raise web.HTTPNotFound()
+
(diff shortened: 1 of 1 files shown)


jigold (approved):

jigold (approved):
I was worried we might limit the number of jobs and block accepting new jobs even though the deleted job has "finished", but that's not the case.

PR #11804, opened 2022-04-27
[query] Remove IntPacker

chrisvittal:
It's dead. We can re-add it if we ever revive packed integer
representations.


tpoterba (approved):

hail/src/main/java/is/hail/utils/IntPacker.java
```diff
@@ -1,108 +0,0 @@
...
-                Region.storeBytes(addr, data, di, 3);
-                Region.storeByte(addr + 3, (byte)0);
-                di += 3;
-                break;
-            case 3:
-                Region.storeBytes(addr, data, di, 4);
-                di += 4;
-                break;
-            default:
-                throw new HailException("unreachable code reached, this is a bug");
```
tpoterba, line 104: 👍

PR #11805, opened 2022-04-27
[query] Fix latent bug in ArrayElementAggregator copy

tpoterba:
I spruced up the entire method, but the bug was in the original
L153. One of the Code=>Value changes in the last couple of months
added the memoize to eltOffset. This was a ticking time bomb --
the usage of eltOffset inside `seq` must be evaluated inline, since
the value of `idx` changes between iterations of the loop where the
code argument to `seq` is located.

I suspect this has worked until now because `idx` is 0 every time
this method is called in typical runtimes. It exploded for the
buffered aggregator PR, where we are calling the copy with a stale
value of `idx` presumably equal to the length of the array, so we're
reading memory off the end of the array allocation.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/agg/ArrayElementLengthCheckAggregator.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/agg/ArrayElementLengthCheckAggregator.scala
@@ -147,18 +147,13 @@ class ArrayElementState(val kb: EmitClassBuilder[_], val nested: StateTuple) ext
   }
 
   def copyFromAddress(cb: EmitCodeBuilder, src: Value[Long]): Unit = {
-    // FIXME: does this really need to be a field?
-    val srcOff = cb.newField("aelca_copyfromaddr_srcoff", src)
-    val initOffset = cb.memoize(typ.loadField(srcOff, 0))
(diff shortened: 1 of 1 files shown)


tpoterba:
actually not the pointer type of the agg state. That correctly threaded through copy methods.

chrisvittal (approved):

patrick-schultz:
I'm confused by your description. The memoize occurs inside `copyFromAddress`, and `idx` never changes in the body. If `copyFromAddress` is called inside a loop in which `idx` is changing, the memoize will be reevaluated each iteration.

Is the issue that the memoize forces `typ.loadField(srcOff, 1)` outside of the `typ.isFieldMissing(cb, srcOff, 1)` check?

tpoterba:
The problem is that memoizing eltOffset outside of `seq` uses teh value of `idx` at the point of memoization (garbage) versus inside of `seq` (well defined):

```scala
  def seq(cb: EmitCodeBuilder, init: => Unit, initPerElt: => Unit, seqOp: => Unit): Unit = {
    init
    cb.assign(idx, 0)
    cb.whileLoop(idx < lenRef, {
      initPerElt
      seqOp
      store(cb)
      cb.assign(idx, idx + 1)
    })
  }
```

PR #11806, opened 2022-04-27
[style] Pin mypy

daniel-goldstein:
I'm not sure if this is the right fix here or if we can get rid of the deprecated FS but this is the lint error that's breaking main.

### DIFF
--- a/hail/python/hailtop/batch/batch.py
+++ b/hail/python/hailtop/batch/batch.py
@@ -1,3 +1,4 @@
+import asyncio
 import os
 import warnings
 import re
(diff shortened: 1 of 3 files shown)


hail/python/hailtop/batch/batch.py
```diff
@@ -591,7 +592,7 @@ def schedule_job(j):
         run_result = self._backend._run(self, dry_run, verbose, delete_scratch_on_exit, **backend_kwargs)  # pylint: disable=assignment-from-no-return
         if self._DEPRECATED_fs is not None:
             # best effort only because this is deprecated
-            self._DEPRECATED_fs.close()
+            asyncio.get_event_loop().run_until_complete(self._DEPRECATED_fs.close())
```
danking, line 595: We usually use `async_to_blocking` for this

danking (dismissed):

danking (approved):

PR #11811, opened 2022-05-02
[query] Add adaptive branching to LowerDistributedSort

chrisvittal:
For each large segment in each loop iteration, we pick a branch factor
based on the size of the chunks in each segment.

Future work includes choosing a branch factor even for the first loop
iteration, which we currently cannot do without knowing the byte sizes
of the incoming data (not something comupted at present)


chrisvittal:
gonna benchmark

chrisvittal:
The failure here looked intermittent, a jar failed to download for a batch. Ready for review.

chrisvittal:
bump

PR #11816, opened 2022-05-03
[query] Enable Query on Batch on Azure

daniel-goldstein:
This adds the necessary glue to use the Scala AzureStorageFS when running QoB in Azure and turns on the QoB tests in Azure CI.


hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -143,14 +142,18 @@ class AzureStorageFS(val credentialsJSON: Option[String] = None) extends FS {
...
         }
 
-        client.downloadStreamWithResponse(
+        val response = client.downloadStreamWithResponse(
           outputStreamToBuffer, new BlobRange(pos, count),
           null, null, false, timeout, null)
-        pos += count
-        bb.flip()
+        AzureStorageFS.log.info(s"Tried to retrieve ${count} bytes from position ${pos}")
+        AzureStorageFS.log.info(s"Got status code ${response.getStatusCode}, and headers ${response.getHeaders}")
```
danking, line 149: Can we nix this now?

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -143,14 +142,18 @@ class AzureStorageFS(val credentialsJSON: Option[String] = None) extends FS {
...
+        if (response.getStatusCode / 100 == 2) {
```
danking, line 150: prefer >= 200 <300

danking (changes requested):

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
daniel-goldstein: Yup, haven't cleaned up some debugging

danking (approved):

PR #11819, opened 2022-05-05
[query] Execute single-partition scatters on driver

tpoterba:


danking (approved):
wow that was way easier than I expected

tpoterba:
I could probably do this at the CDA code generator to obviate the globals/context serialization but those are pretty fast anyway.

PR #11823, opened 2022-05-10
[batch] Update GCP disk resources to account for np local-ssd

jigold:


batch/sql/update_local_ssd_pricing.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # https://cloud.google.com/compute/disks-image-pricing#localssdpricing
+    # https://cloud.google.com/compute/disks-image-pricing#persistentdisk
+    rates = [
+        ('disk/local-ssd/preemptible/1', rate_gib_month_to_mib_msec(0.048)),
+        ('disk/local-ssd/nonpreemptible/1', rate_gib_month_to_mib_msec(0.08)),
+        ('compute/n1-preemptible/2', rate_cpu_hour_to_mcpu_msec(0.00698)),
+        ('memory/n1-preemptible/2', rate_gib_hour_to_mib_msec(0.00094)),
+        ('compute/n1-nonpreemptible/2', rate_cpu_hour_to_mcpu_msec(0.033174)),
+        ('memory/n1-nonpreemptible/2', rate_gib_hour_to_mib_msec(0.004446)),
```
danking, line 20: Why aren't we updated instance_config.py too? (And if we do that we should update the docs too, right?)

danking (changes requested):

batch/sql/update_local_ssd_pricing.py
jigold: I'll update the docs. 

The instance config that has the product names hard coded is for backwards compatibility. The new code from November looks up what the latest product version is for each resource and uses that. So in this case, it should see 2. I'll double check though with dev deploy.

jigold:
There's another issue I need to address.

jigold:
Closing in favor of #11840

PR #11824, opened 2022-05-10
[batch] Make MJC wait on MJS

daniel-goldstein:
Currently, the MJS and MJC requests from the worker to the driver for a given job can race, as they are run as independent asyncio tasks. This results in unnecessary database load and deadlocks between the MJS and MJC SQL procedures. Rather than address the procedures directly, we enforce that we will never run MJS and MJC concurrently. The system is resilient to never receiving an MJS (as MJC will add any attempt data if not present), so we can make the following changes to the worker:
- Serialize the submission of MJS and MJC requests by having the MJC task wait on the MJS future
- Give up retrying MJS once the job has completed because we will instead just send an MJC. This could potentially reduce the database load for very short jobs.

I ran a load test of 10k `true` jobs and `sleep 5` jobs a few times against my namespace and saw 0 deadlocks 🎉

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1356,6 +1356,8 @@ class Job:
 
         self.project_id = Job.get_next_xfsquota_project_id()
 
+        self.mjs_fut: Optional[asyncio.Future] = None
+
     @property
     def job_id(self):
         return self.job_spec['job_id']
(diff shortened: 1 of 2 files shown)


batch/batch/worker/worker.py
```diff
@@ -2606,6 +2613,8 @@ async def post_job_complete_1(self, job, full_status):
             delay_secs = min(delay_secs * 2, 2 * 60.0)
 
     async def post_job_complete(self, job, full_status):
+        if job.mjs_fut is not None:
```
jigold, line 2616: Should we cancel the future instead? I'm worried about the case where the driver is overloaded and this is just retrying transient errors. Can you remind me why this change is correct?

jigold (approved):

PR #11829, opened 2022-05-11
[query] Allow tracking of the full deserialized byte size of a partition

chrisvittal:
Add an extra (currently unused) flag on PartitionNativeWriter and an
extra field on its return type. When the flag is true, we compute the
sum of the byte sizes of every row and return it as the extra field
`partitionByteSize`.

We will use this to compute branch factors in LowerDistributedSort and
depending on computational cost, possible turn this on by default to
allow more optimization around deserializing/processing more than one
row at a time.


tpoterba (dismissed):
I think leaving the field in but uncomputed (as 0) if the flag is false is going to confuse us down the road. Can we branch on the flag in the return type, and where we create/use the variable?

tpoterba (approved):

PR #11831, opened 2022-05-12
[spring cleaning] remove libhail

chrisvittal:
Some repository cleanup. The libhail experiment has lain dormant for
a long time now. It is probably not the direction that we will take the
hail project in, it was a useful experiment but at this point, it is
extra bulk in the repository that will remain in the project history,
but need not be present in the current worktree.


danking (approved):
We can always revert the PR if we want this back.

PR #11832, opened 2022-05-12
[query] Retry read from GCS channel

daniel-goldstein:
AFAIK, the `retryTransientErrors` in `createNoCompression` would only retry any errors in creating the `ReadChannel`, but we still want to retry transient errors when calling read later down the line.


daniel-goldstein:
Ya I think your retry is right, this feels dangerous.

PR #11833, opened 2022-05-13
[query] default rand_unif to [0, 1]

danking:
CHANGELOG: `hl.rand_unif` now has default arguments of 0.0 and 1.0

### DIFF
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -2569,7 +2569,7 @@ def rand_pois(lamb, seed=None) -> Float64Expression:
 
 
 @typecheck(lower=expr_float64, upper=expr_float64, seed=nullable(int))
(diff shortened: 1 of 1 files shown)


hail/python/hail/expr/functions.py
```diff
@@ -2569,13 +2569,16 @@ def rand_pois(lamb, seed=None) -> Float64Expression:
 
 
 @typecheck(lower=expr_float64, upper=expr_float64, seed=nullable(int))
-def rand_unif(lower, upper, seed=None) -> Float64Expression:
+def rand_unif(lower=0.0, upper=0.1, seed=None) -> Float64Expression:
```
tpoterba, line 2572: 1.0

tpoterba (changes requested):

tpoterba (approved):

PR #11842, opened 2022-05-16
[pip] Use pip-compile to create fully-pinned lock files for python dependencies

daniel-goldstein:


danking (changes requested):
Can we do docker/requirements.txt too?

daniel-goldstein:
This will enforce even stricter requirements on downstream projects, right? Do we want to do this only for our own builds and use unpinned dependencies in the pip package's setup.py?

danking (dismissed):
Hmm, yeah, so the in versus not-in seems backwards to me. The ".in" files are the relaxed pins that we want our users to experience. Can we just change the names? Our docker images all install from the requirements files directly, so the looser requirements that appear in setup.py should have no ill effect.

danking (changes requested):
Do we need a build.yaml change to continuously verify the pinned file is compatible with the public requirements?

daniel-goldstein:
I guess it depends whether you want up to date or just compatible, the maintainers seem to be of the opinion that you should either always update the lock file immediately or set upper bounds if you're not ok with a certain upgrade, seen [here](https://github.com/jazzband/pip-tools/issues/882).

Continuous work, but maybe the right way to go honestly. In that case trivially make a build.yaml step that asserts the lock file is valid and up to date.

danking:
Ah, I misunderstood. So you can't ask pip-compile to accept as input a requirements.txt and a pinned-requirements.txt and output a new pinned-requirements.txt? That's what I was imagining. With the main effect being if we add something to requirements.txt without changing pinned-requirements.txt, we'd get an error.

Having pip-compile constantly bump requirements in every PR seems not ideal. What I'm trying to achieve is that our pinned-requirements always represent *a* version-compatible, transitive closure of requirements.txt

daniel-goldstein:
I got what you originally meant, I'm just not sure how to do that, and looks like the folks writing the tool don't seem to agree. I agree it also seems like a lot to always be bumping our requirements.

daniel-goldstein:
> Ah, I misunderstood. So you can't ask pip-compile to accept as input a requirements.txt and a pinned-requirements.txt and output a new pinned-requirements.txt?

Oh wait actually we can totally do this

daniel-goldstein:
Tested this out by running it locally and changing `aiohttp-session` in `docker/requirements.txt` from `2.70` to `2.8.0` and got the following:

```
+ pip-compile --quiet docker/requirements.txt docker/pinned-requirements.txt --output-file=new-pinned.txt
Could not find a version that matches aiohttp-session==2.7.0,==2.8.0 (from -r docker/requirements.txt (line 4))
Tried: 0.0.1, 0.0.1, 0.1.0, 0.1.0, 0.1.1, 0.1.1, 0.1.2, 0.1.2, 0.2.0, 0.2.0, 0.3.0, 0.3.0, 0.4.0, 0.4.0, 0.5.0, 0.5.0, 0.7.0, 0.7.0, 0.7.1, 0.7.1, 0.8.0, 0.8.0, 1.0.0, 1.0.0, 1.0.1, 1.0.1, 1.1.0, 1.1.0, 1.2.0, 1.2.0, 1.2.1, 1.2.1, 2.0.0, 2.0.0, 2.0.1, 2.0.1, 2.1.0, 2.1.0, 2.2.0, 2.2.0, 2.3.0, 2.3.0, 2.4.0, 2.4.0, 2.5.1, 2.5.1, 2.6.0, 2.6.0, 2.7.0, 2.7.0, 2.8.0, 2.8.0, 2.9.0, 2.9.0, 2.10.0, 2.10.0, 2.11.0, 2.11.0
Skipped pre-versions: 2.10.0a0, 2.10.0a0
There are incompatible versions in the resolved dependencies:
  aiohttp-session==2.7.0 (from -r docker/pinned-requirements.txt (line 20))
  aiohttp-session==2.8.0 (from -r docker/requirements.txt (line 4))
```

and another example where I added an unrelated pip dependency in the requirements but didn't update the lock file

danking (approved):
This is really nice. Thanks!

daniel-goldstein:
@danking bump

daniel-goldstein:
@jigold This PR allows us to create fully-pinned, deterministic requirements files for our pip dependencies. This uses those fully-pinned versions in our docker images so that breaking releases of transitive dependencies don't spontaneously break our CI pipelines. If we want to add or upgrade a dependency, someone needs to:

1. Update the relevant `requirements.txt` file just as we would do currently.
2. Run `pip-compile` to generate new lock `pinned-requirements.txt` files and check in the changes to their PR.

The added check step verifies that you can't do 1 and forget to do 2.

jigold (dismissed):
> There's a slight hiccup that you should be clued in on. The previous build failed because pip-compile resolves the minimal dependency graph for the host OS it was run on, so running it on MacOS excluded a couple linux-specific packages that are transitively required in the docker requirements. I instead generated the docker requirements pinned file inside our base container and committed that, with a clarifier that these are linux-pinned-requirements.txt

Can we put a note somewhere in dev-docs on how to properly update dependencies, especially the fact that you need to run it on a Linux machine?

check_pip_requirements.sh
```diff
@@ -0,0 +1,15 @@
...
+pinned=$2
+new_pinned=$(mktemp)
+pinned_no_comments=$(mktemp)
+new_pinned_no_comments=$(mktemp)
+
+pip-compile --quiet $reqs $pinned --output-file=$new_pinned
+# Get rid of comments that might differ despite requirements being the same
+cat $pinned | sed '/#/d' > $pinned_no_comments
+cat $new_pinned | sed '/#/d' > $new_pinned_no_comments
+diff $pinned_no_comments $new_pinned_no_comments
```
jigold, line 15: Do you need to exit with the exit code from `diff`?

hail/python/dev/requirements.txt
```diff
@@ -1,3 +1,5 @@
+-c ../requirements.txt
```
jigold, line 1: Can you double check this is right?

jigold (changes requested):

hail/python/dev/requirements.txt
daniel-goldstein: Yup, this will "constrain" the requirements in this file with those in the upper requirements.txt, so we know the two requirements files won't diverge/conflict.

check_pip_requirements.sh
daniel-goldstein: The script should inherit the exit code of its last command, and because we have `set -e`, then it should fail if diff comes up different. Here's an example locally after I added an unrelated dependency, `msprime` to the hail/python/requirements.txt:

```
./check_pip_requirements.sh hail/python/requirements.txt hail/python/pinned-requirements.txt
+ reqs=hail/python/requirements.txt
+ pinned=hail/python/pinned-requirements.txt
++ mktemp
+ new_pinned=/var/folders/2c/0b40ckvn7bq_g0zz9qtkxwr40000gp/T/tmp.FxFFOFRJ
++ mktemp
+ pinned_no_comments=/var/folders/2c/0b40ckvn7bq_g0zz9qtkxwr40000gp/T/tmp.70s9RDbL
++ mktemp
+ new_pinned_no_comments=/var/folders/2c/0b40ckvn7bq_g0zz9qtkxwr40000gp/T/tmp.IjRW9sSq
+ pip-compile --quiet hail/python/requirements.txt hail/python/pinned-requirements.txt --output-file=/var/folders/2c/0b40ckvn7bq_g0zz9qtkxwr40000gp/T/tmp.FxFFOFRJ
+ cat hail/python/pinned-requirements.txt
+ sed /#/d
+ cat /var/folders/2c/0b40ckvn7bq_g0zz9qtkxwr40000gp/T/tmp.FxFFOFRJ
+ sed /#/d
+ diff /var/folders/2c/0b40ckvn7bq_g0zz9qtkxwr40000gp/T/tmp.70s9RDbL /var/folders/2c/0b40ckvn7bq_g0zz9qtkxwr40000gp/T/tmp.IjRW9sSq
21a22
> demes==0.2.2
33a35,36
> importlib-metadata==4.11.4
> importlib-resources==5.7.1
37a41
> jsonschema==4.6.0
40a45
> msprime==1.2.0
43a49
> newick==1.3.2
59a66
> pyrsistent==0.18.1
67a75,76
> ruamel-yaml==0.17.21
> ruamel-yaml-clib==0.2.6
71a81
> svgwrite==1.4.2
75a86
> tskit==0.4.1
80a92
> zipp==3.8.0
make: *** [check-pip-dependencies] Error 1
```

hail/python/dev/requirements.txt
daniel-goldstein: I double-checked by pip installing the hail dependencies and dev dependencies

jigold (approved):

PR #11847, opened 2022-05-17
[query] python side infrastructure for deterministic rngs

patrick-schultz:
High level design:
* `ApplySeeded` takes a child of type `trngstate`. It is always constructed as a free reference `__rng_state`, which must be defined by a parent, who has the responsibility to ensure the seeded function is given a distinct rng state on each invocation.
* Any node with a child which may be executed more than once (e.g. `TailLoop`, stream nodes, table nodes) must use `Let('__rng_state', RNGSplit(Ref('__rng_state'), uid), child)`, where `uid` is a long or tuple of longs which is guaranteed to be distinct on every execution of `child`.
* Uids are typically created at the leaves of pipelines (`TableRead`, `StreamRange`, etc.), and propagated upwards.

There was a phase-ordering conflict that had to be worked around:
* IRs must be given explict rng state and uid semantics as early as possible, to ensure determinism.
* The transformation to explicitly pass rng states and uids must happen during IR construction. If it happened later, it would create new IR objects, which would defeat the python CSE pass (which only recognizes equivalent subexpressions when they are represented by the same python object).
* The rng explication requires some type information.
* Types on the python IR are assigned after the IR is fully constructed.

To fix this:
* `Ref`'s must be given a type at construction
  * `TopLevelReference`s are the only case that needs to be constructed before a type is known. But they are always constructed wrapped in a `SelectFields` or `GetField`, whose type is known at construction. I added new IR classes `SelectedTopLevelReference` and `ProjectedTopLevelReference` for these two cases, which are thin wrappers which don't appear in the rendered IR.
* `construct_expr` always assigns a type to the ir. Bottom-up type construction will later assert equality with the assigned type. This caught some existing bugs, where expression type and ir type didn't agree.
* At construction of the root node of a stream/table/matrixtable pipeline (i.e. a non-stream value ir with at least one stream/table/matrixtable child), recursively rewrite the contained pipeline(s) to make rng states and uids explicit. This is safe, since stream/table/matrixtable IRs won't be CSE'd, because they may only be evaluated once. Contained value IRs are not rewritten, only wrapped with bindings which define the rng state.

These are currently non-functional changes, as `ApplySeeded` still uses the old rng, and will ignore the rng state.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -2048,18 +2048,18 @@ class Emit[C](
         val impl = x.implementation
         val unified = impl.unify(Array.empty[Type], x.argTypes, rt)
         assert(unified)
-        if (fn == "rand_unif") {
-          val pureImpl = x.pureImplementation
-          assert(pureImpl.unify(Array.empty[Type], rngState.typ +: x.argTypes, rt))
-          val codeArgsMem = codeArgs.map(_.memoize(cb, "ApplySeeded_arg"))
-          emitI(rngState).consumeI(cb, {
-            impl.applySeededI(seed, cb, region, impl.computeReturnEmitType(x.typ, codeArgs.map(_.emitType)).st, codeArgsMem.map(_.load): _*)
-          }, { state =>
-            pureImpl.applyI(EmitRegion(cb.emb, region), cb, impl.computeReturnEmitType(x.typ, codeArgs.map(_.emitType)).st, Seq[Type](), const(0), EmitCode.present(mb, state) +: codeArgsMem.map(_.load): _*)
-          })
-        } else {
+//        if (fn == "rand_unif") {
+//          val pureImpl = x.pureImplementation
+//          assert(pureImpl.unify(Array.empty[Type], rngState.typ +: x.argTypes, rt))
+//          val codeArgsMem = codeArgs.map(_.memoize(cb, "ApplySeeded_arg"))
+//          emitI(rngState).consumeI(cb, {
+//            impl.applySeededI(seed, cb, region, impl.computeReturnEmitType(x.typ, codeArgs.map(_.emitType)).st, codeArgsMem.map(_.load): _*)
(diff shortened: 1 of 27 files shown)


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -2048,18 +2048,18 @@ class Emit[C](
...
+//          val codeArgsMem = codeArgs.map(_.memoize(cb, "ApplySeeded_arg"))
+//          emitI(rngState).consumeI(cb, {
+//            impl.applySeededI(seed, cb, region, impl.computeReturnEmitType(x.typ, codeArgs.map(_.emitType)).st, codeArgsMem.map(_.load): _*)
+//          }, { state =>
+//            pureImpl.applyI(EmitRegion(cb.emb, region), cb, impl.computeReturnEmitType(x.typ, codeArgs.map(_.emitType)).st, Seq[Type](), const(0), EmitCode.present(mb, state) +: codeArgsMem.map(_.load): _*)
+//          })
+//        } else {
           impl.applySeededI(seed, cb, region, impl.computeReturnEmitType(x.typ, codeArgs.map(_.emitType)).st, codeArgs: _*)
-        }
+//        }
```
chrisvittal, line 2062: Remember to restore/remove this

patrick-schultz:
Ok @chrisvittal, tests are passing now

hail/src/main/scala/is/hail/expr/ir/Emit.scala
patrick-schultz: I uncommented this, but left it as dead code for now. Before we enable use of the new RNG, we need to at least hook up key generation. Right now the RNG key is hard-coded in a few places to a low entropy dummy key.

I would welcome input on the best way to fix a single RNG key per hail session, with the ability for users to manually set the key.

chrisvittal: How are we defining a hail session. We probably need to use something like how we pass flags to query on batch, make passing an RNG seed part of the protocol.

PR #11849, opened 2022-05-18
[query] Fix GoogleCloudStorageFS delete

chrisvittal:
When our scala implemention of  GoogleCloudStorageFS attempted to delete
a large number of files, we would only delete the first page of results.
This was because we were using getValues which only returns the blobs in
the first page.

Now, for recursive delete, we properly iterate through the pages, and
furthermore, use the Storage.delete method that takes an iterable in
order to take advantage of batch processing and reduce the number of RPC
requests.


danking (approved):

chrisvittal:
Ok. This has to retry azure.

PR #11850, opened 2022-05-18
[shuffle] Track bytes from initial write; adaptive branching all the …

tpoterba:
…way through


hail/src/main/scala/is/hail/expr/ir/lowering/LowerDistributedSort.scala
```diff
@@ -178,7 +178,7 @@ object LowerDistributedSort {
 
       val perPartStatsCDAContextData = partitionDataPerSegment.flatten.zip(numSamplesPerPartition).map { case (partData, numSamples) =>
                                                                                          /* let's just call the missing byteSize 'infinity' */
```
chrisvittal, line 180: Delete this note

chrisvittal (dismissed):

PR #11851, opened 2022-05-18
[aiotools] Make Azure asyncfs support blobs with ? in the blob name

daniel-goldstein:
urlparse isn't really what we want for breaking down a blob storage URL. There are special characters that delimit different parts of a typical URL that don't translate to a blob. Specifically, if a blob name contains a `?` (which is totally valid if albeit horrendous), `parsed.path` will terminate before the `?` and drop the rest because it views those as query params. We really just want everything after the container to be the blob name, so we do a bit more manual work in the name of simplicity. Dan already handled this in the Google async fs but this was never changed in the azure implementation. They now resemble each other more closely.


danking (approved):

danking:
@daniel-goldstein it looks like this might not work on GCS as well?

daniel-goldstein:
Closing in favor of #11816

PR #11854, opened 2022-05-19
[make] Add target for $PYTHON_JAR and version-info

tpoterba:
Sufficient to run local backend.

### DIFF
--- a/hail/Makefile
+++ b/hail/Makefile
@@ -142,6 +142,8 @@ python/README.md: ../README.md
 $(PYTHON_JAR): shadowJar
 	cp -f $(SHADOW_JAR) $@
 
+jar-and-version: $(PYTHON_JAR) python-version-info
+
 .PHONY: pytest
 pytest: PYTEST_TARGET ?= test/hail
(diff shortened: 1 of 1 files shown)


hail/Makefile
```diff
@@ -142,6 +142,8 @@ python/README.md: ../README.md
 $(PYTHON_JAR): shadowJar
```
chrisvittal, line 142: Outside of scope here, but this rule should be:
```suggestion
$(PYTHON_JAR): $(SHADOW_JAR)
```

hail/Makefile
```diff
@@ -142,6 +142,8 @@ python/README.md: ../README.md
...
 	cp -f $(SHADOW_JAR) $@
 
+jar-and-version: $(PYTHON_JAR) python-version-info
```
chrisvittal, line 145: There should be `.PHONY: jar-and-version`

chrisvittal (changes requested):
I don't like this.

I don't think that proliferation of rules like this is helpful or useful. We probably should have a .PHONY rule with `$(PYTHON_JAR)` as a prerequisite. Given that, this rule could just be `make python-jar python-version-info`

chrisvittal (approved):

PR #11855, opened 2022-05-19
[query] don't create a future for each task

tpoterba:


danking (approved):

danking:
Approving but we should remove the val r =

PR #11859, opened 2022-05-19
[ci] Build skopeo from source

daniel-goldstein:


danking (approved):

lgruen:
`skopeo` is part of the standard Ubuntu 2**2**.04 repo, but probably that's a bit more difficult to update to? Building Hail on 22.04 seems to work fine though.

daniel-goldstein:
Ya this would be more ideal, but since 22.04 is so new I anticipate there might be a couple of bumps. For example, I think GCP's logging agent we've been using for years might not be supported on the newer Ubuntu, since the agent is now considered legacy.

daniel-goldstein:
The Red Hat folks appear to have restored the package, so we could remove this workaround. However, they're not going to maintain it so we will either have to build from source or ultimately upgrade.

danking:
Do we have a logging agent installed inside the containers themselves? I'd have thought that was on the worker VM. Seems like we could go to 22.04 in all the containers. It does mean our worker VM is out of sync with the containers.

daniel-goldstein:
Ah ya that's a good point, we could indeed move forward for the containers.

PR #11861, opened 2022-05-20
[query] fix AzureStorageFS openNoCompression 

vrautela:
- fix double-counting error in openNoCompression method 
- add tests to read/write more bytes than can be held in the ByteBuffer
- abstract out seek method into FS

### DIFF
--- a/hail/src/main/scala/is/hail/io/fs/FS.scala
+++ b/hail/src/main/scala/is/hail/io/fs/FS.scala
@@ -155,10 +155,7 @@ abstract class FSSeekableInputStream extends InputStream with Seekable {
     bb.clear()
     bb.limit(0)
     pos = newPos
-    adjustSeekInternalState(newPos)
   }
-
-  def adjustSeekInternalState(newPos: Long): Unit
 
   def getPosition: Long = pos
 }
--- a/hail/src/main/scala/is/hail/io/fs/FS.scala
+++ b/hail/src/main/scala/is/hail/io/fs/FS.scala
@@ -109,7 +109,7 @@ object FSUtil {
 
 abstract class FSSeekableInputStream extends InputStream with Seekable {
   protected[this] var closed: Boolean = false
-  protected[this] var pos: Long = 0
+  private[this] var pos: Long = 0
   private[this] var eof: Boolean = false
 
   protected[this] val bb: ByteBuffer = ByteBuffer.allocate(64 * 1024)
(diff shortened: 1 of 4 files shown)


hail/src/main/scala/is/hail/io/fs/FS.scala
```diff
@@ -151,6 +151,15 @@ abstract class FSSeekableInputStream extends InputStream with Seekable {
     toTransfer
   }
 
+  def seek(newPos: Long): Unit = {
+    bb.clear()
+    bb.limit(0)
+    pos = newPos
+    adjustSeekInternalState(newPos)
+  }
```
daniel-goldstein, line 159: So I think I've slightly changed my mind here about this extra method approach. I think we should just keep seek like it is here without the call to `adjustSeekInternalState`, and then the `GoogleStorageFS` should override `seek` with a call to `super.seek` and whatever additional bookkeeping we're currently putting in `adjustSeekInternalState`.

hail/src/test/scala/is/hail/fs/FSSuite.scala
```diff
@@ -271,6 +271,60 @@ trait FSSuite {
...
+
+    fs.delete(f, false)
+
+    assert(!fs.exists(f))
+  }
+
+  @Test def testReadWriteBytesLargerThanBuffer(): Unit = {
+    val f = t()
+
+    val byteBufferSize = 64 * 1024
```
daniel-goldstein, line 295: This test is a bit brittle because it depends on an exact value of an implementation detail of the FS. If we were to change the implementation, the test could become invalid, and we might forget to update the tests. I think it is unlikely that we would ever use a buffer size greater than, say ten megabytes, so let's instead ensure that our tests cover a spread of different sizes that are likely to catch this edge case regardless of the exact buffer size.

daniel-goldstein (changes requested):

hail/src/test/scala/is/hail/fs/FSSuite.scala
```diff
@@ -271,6 +271,60 @@ trait FSSuite {
...
+    assert(fs.exists(f))
+
+    using(fs.open(f)) { is =>
+      assert(is.read() == 1)
+      assert(is.read() == 127)
+      assert(is.read() == 255)
+
+      var i = 0
+      while (i < byteBufferSize + 10) {
+        assert(is.read() == i)
```
daniel-goldstein, line 318: This value will quickly exceed the maximum value for a byte. What would you expect to read back in such a case?

daniel-goldstein:
Merging this inside #11816

PR #11867, opened 2022-05-25
[batch] Add more logging for disk formatting errors

jigold:
I'm not sure why the disk formatting failed. So I've added more logging to see if we can understand why this is happening.

### DIFF
--- a/batch/batch/cloud/gcp/worker/disk.py
+++ b/batch/batch/cloud/gcp/worker/disk.py
@@ -1,8 +1,9 @@
 import logging
+import os.path
 from typing import Dict, Optional
 
 from hailtop.aiocloud import aiogoogle
-from hailtop.utils import LoggingTimer, check_shell_output, retry_all_errors_n_times
+from hailtop.utils import CalledProcessError, LoggingTimer, check_shell_output, retry_all_errors_n_times
 
 from ....worker.disk import CloudDisk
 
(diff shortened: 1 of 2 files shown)


jigold:
Goal is to get more information for hail-is/hail-production-issues/#16

batch/batch/cloud/gcp/worker/disk.py
```diff
@@ -57,12 +60,17 @@ async def _unmount(self):
...
+            try:
+                await check_shell_output(
+                    f'mkfs.ext4 -m 0 -E lazy_itable_init=0,lazy_journal_init=0,discard {self.disk_path}'
+                )
+                await check_shell_output(f'mkdir -p {self.mount_path}')
+                await check_shell_output(f'mount -o discard,defaults {self.disk_path} {self.mount_path}')
+                await check_shell_output(f'chmod a+w {self.mount_path}')
+            except CalledProcessError:
+                outerr = await check_shell_output(f'ls {os.path.dirname(self.disk_path)}')
+                log.info(f'debugging info while formatting disk for {self.name}: {outerr}\n{self.last_response}')
```
daniel-goldstein, line 72: Can we capture the exception info that got us here? Also what if the `ls` errors?

daniel-goldstein (changes requested):
Just one change

PR #11870, opened 2022-05-26
[query] Fully rip out query-gsa-key

daniel-goldstein:
We no longer have a service with the identity of Query, so no longer have a `query-gsa-key`. This was left over from an incomplete deletion, and since Batch now owns the query jar bucket, we should use its identity to upload jars to it.


danking (dismissed):
X

daniel-goldstein:
Here's the terraform configurations in GCP and Azure:

- GCP: Batch has admin storage permissions, as granted here https://github.com/hail-is/hail/blob/1f5e1540c04abfde58ead1084841fec5aa6e0ed3/infra/gcp/main.tf#L415-L424

We also grant it a Viewer role on the query bucket after that which seems redundant. We should really not grant it global storage admin and instead give it admin for just the query bucket and other associated batch buckets.

I checked in hail-vdc and batch does not have the global storage admin role, and the Viewer role on the query bucket. I've changed that role now to admin on the query bucket.


- Azure: Story is simpler. The `query` storage container is part of the `batch` storage account. The batch SP has ownership over the `batch` storage account and by extension all of the containers inside it.

daniel-goldstein:
So I don't think that AUS will run into any issues, but we should follow up with restricting the batch GSA permissions in GCP.

danking (approved):

PR #11874, opened 2022-05-31
[aiogoogle] Increase chunk size when copying files to GCS

jigold:
The docs say we should use at a minimum 8 MB chunks. With 50 way parallelism that's 400 MB in memory which should be fine.


jigold:
The tests failed spectacularly for what I assume was because the database was overloaded. Will retry if you have no objections to the PR as is.

PR #11876, opened 2022-05-31
[internal-gateway] Higher limit for internal-gateway cpu

daniel-goldstein:
nginx can use more than one core and at max load + PRs we're maxing out internal-gateway cycles, which add latency, timeouts, retries, and generally degrade the experience in one namespace based on activity in others. Allowing nginx to use more cores (in this case this is up to half our node size) got our system back into its intended state with graceful throttling.

I'll admit, other than being half a node size, 4 is a bit arbitrary here. I think our k8s nodes are annoyingly underutilized enough that we shouldn't see issues in practice with letting internal-gateway use cores that are very likely idle.

### DIFF
--- a/internal-gateway/deployment.yaml
+++ b/internal-gateway/deployment.yaml
@@ -42,7 +42,7 @@ spec:
              cpu: "20m"
              memory: "200M"
            limits:
-             cpu: "1"
+             cpu: "4"
              memory: "1G"
          ports:
           - containerPort: 80


jigold:
Can you double check we don't need to explicitly tell nginx to use more than 1 core? I'm looking here:

https://www.nginx.com/blog/thread-pools-boost-performance-9x/#Configuring-Thread-Pools
https://www.nginx.com/blog/thread-pools-boost-performance-9x/#Benchmarking

Otherwise, I think this change is fine, although we do already have a minimum of two copies of internal-gateway at any time. Is this change better than increasing the number of copies of internal-gateway? I assume that the response time to increases in load will be faster with your proposed change.

jigold (changes requested):
See comment.

daniel-goldstein:
Closing this in light of what we've learned about keepalive

PR #11878, opened 2022-06-01
[batch] Use Spot VMs on GCP

lgruen:
This will allow preemptible VMs to run longer than 24h.

See https://cloud.google.com/compute/docs/instances/create-use-spot#create.

#assign services

### DIFF
--- a/batch/batch/cloud/gcp/driver/create_instance.py
+++ b/batch/batch/cloud/gcp/driver/create_instance.py
@@ -96,7 +96,13 @@ def create_vm_config(
                 'accessConfigs': [{'type': 'ONE_TO_ONE_NAT', 'name': 'external-nat'}],
             }
         ],
-        'scheduling': {'automaticRestart': False, 'onHostMaintenance': "TERMINATE", 'preemptible': preemptible},
+        'scheduling': {
+            'automaticRestart': False,
+            'instanceTerminationAction': 'DELETE',
+            'onHostMaintenance': 'TERMINATE',
+            'preemptible': preemptible,
+            'provisioningModel': 'SPOT' if preemptible else 'STANDARD',
+        },
         'serviceAccounts': [
             {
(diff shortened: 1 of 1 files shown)


jigold (dismissed):

jigold:
The error on Azure is unrelated and will post on Zulip.

lgruen:
> The error is here:
> `Invalid value for field \'resource.scheduling.instanceTerminationAction\': \'DELETE\'. You cannot specify a termination action for a VM instance that has the standard provisioning model (default).`

Thanks very much for finding that and sorry for not seeing this issue earlier! This is fixed now.

jigold:
Still needs a black reformat. You can run `make -C batch/ check` ahead of time to catch these errors or add pre-commit hooks.

```
PYTHONPATH=${PYTHONPATH:+${PYTHONPATH}:}../hail/python:../gear:../web_common python3 -m black . --line-length=120 --skip-string-normalization --check --diff
--- batch/cloud/gcp/driver/create_instance.py	2022-06-02 12:28:49.199357 +0000
+++ batch/cloud/gcp/driver/create_instance.py	2022-06-02 12:31:14.836500 +0000
@@ -78,15 +78,17 @@
             'automaticRestart': False,
             'onHostMaintenance': 'TERMINATE',
         }
 
         if preemptible:
-            result.update({
-                'provisioningModel': 'SPOT',
-                'instanceTerminationAction': 'DELETE',
-                'preemptible': True,
-            })
+            result.update(
+                {
+                    'provisioningModel': 'SPOT',
+                    'instanceTerminationAction': 'DELETE',
+                    'preemptible': True,
+                }
+            )
 
         return result
 
     return {
         'name': machine_name,
would reformat batch/cloud/gcp/driver/create_instance.py
```

danking:
I ran black and rebased, let's continue on my PR https://github.com/hail-is/hail/pull/11945 where we can make whatever changes are necessary to merge it. Jackie, you should have push privileges to my branch.

PR #11884, opened 2022-06-02
[batch] Add always_copy_output option

jigold:
CHANGE_LOG: 
- Added ``Job.always_copy_output`` when using the ``ServiceBackend``. The default behavior is `False` which
  is a breaking change from the previous behavior to always copy output files irregardless of the job's
  completion state.

Thoughts on introducing this "breaking" change? I think it's okay and we will announce the change on Zulip. I think the benefits of getting rid of this not so intuitive behavior where we copy files despite the main container's completion state outweighs possibly breaking someone's pipeline that relied on files being there when the copy step could have failed as well.

### DIFF
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -585,7 +585,7 @@ class Job:
 
                     if self._always_run:
                         warnings.warn('A job marked as always run has a resource file dependency on another job. If the dependent job fails, '
-                                      f'the always run job with the following command will not run:\n{command}')
+                                      f'the always run job with the following command may not succeed:\n{command}')
 
                     self._dependencies.add(r._source)
                     r._source._add_internal_outputs(r)
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -582,6 +582,11 @@ class Job:
                         raise BatchException(f"undefined resource '{name}'\n"
                                              f"Hint: resources must be defined within "
                                              f"the job methods 'command' or 'declare_resource_group'")
+
+                    if self._always_run:
+                        warnings.warn('A job marked as always run has a resource file dependency on another job. If the dependent job fails, '
+                                      f'the always run job with the following command will not run:\n{command}')
+
                     self._dependencies.add(r._source)
                     r._source._add_internal_outputs(r)
             else:
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -78,6 +78,7 @@ class Job:
         self._machine_type: Optional[str] = None
         self._timeout: Optional[Union[int, float]] = None
         self._cloudfuse: List[Tuple[str, str, bool]] = []
+        self._always_copy_output: bool = False
         self._env: Dict[str, str] = {}
         self._wrapper_code: List[str] = []
         self._user_code: List[str] = []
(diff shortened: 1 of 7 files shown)


daniel-goldstein:
I think the proposed new default and the option to change it is much more intuitive than the current behavior and worth a change. Though, I think it would be most polite to announce it on zulip/email list a week or two in advance of the release.

daniel-goldstein:
I think the proposed new default and the option to change it is much more intuitive than the current behavior and worth a change. Though, I think it would be most polite to announce it on zulip/email list a week or two in advance of the release.

daniel-goldstein:
Another thing I just thought of, I think we should add raise some form of error if a job that is `alwaysRun` has inputs from a job that is not `always_copy_output`. I can't imagine something like that being intentional, but if there is a valid use case, maybe we have a warning instead of an error.

daniel-goldstein (dismissed):
Needs rebase fix

jigold:
I'm not sure how I feel about the warning / error suggestion when always run jobs have inputs that aren't always copied out. Cleanup jobs shouldn't care whether the outputs don't get copied out on failure.

daniel-goldstein:
Do cleanup jobs take input files from the parent job? IIUC, if I have an always_run job that takes an input from a job that failed, and whose outputs were not written, the input container for the always_run job would fail.

jigold:
Hopefully the new warning gets at the case you were thinking of as well as a more general problem beyond `always_copy_output`

hail/python/hailtop/batch/job.py
```diff
@@ -539,6 +539,11 @@ def handler(match_obj):
                         raise BatchException(f"undefined resource '{name}'\n"
                                              f"Hint: resources must be defined within "
                                              f"the job methods 'command' or 'declare_resource_group'")
+
+                    if self._always_run:
+                        warnings.warn('A job marked as always run has a resource file dependency on another job. If the dependent job fails, '
+                                      f'the always run job with the following command will not run:\n{command}')
```
daniel-goldstein, line 545: I think this should be `the always run job with the following command may not succeed`. It *will* run, right?

daniel-goldstein (changes requested):

jigold:
Let's leave the WIP tag on, but this should be all set.

daniel-goldstein (dismissed):

hail/python/hailtop/batch/docs/change_log.rst
```diff
@@ -3,6 +3,12 @@
 Change Log
 ==========
 
+**Version 0.2.96**
+
+- Added ``Job.always_copy_output`` when using the ``ServiceBackend``. The default behavior is `False` which
+  is a breaking change from the previous behavior to always copy output files irregardless of the job's
```
danking, line 9: irregardless => regardless

danking (dismissed):

daniel-goldstein (dismissed):

jigold:
We cannot take off WIP and merge this until November 1st.

danking (dismissed):
Why are we making the server side change? If we only change the client *and also always send the setting* then we preserve the behavior of old versions of the hailtop.batch library while also switching to a different behavior.

daniel-goldstein:
Agreed I think this properly preserves behavior for old clients

daniel-goldstein:
@danking This is assigned to me but since you're the last person to leave `changes_requested` it seems appropriate that you re-review.

danking (approved):
LGTM

PR #11895, opened 2022-06-07
[batch] Turn off logging syslog on workers for GCP

jigold:
I don't think we need this anymore for GCP only. I'll need to figure out the equivalent in Azure with the OMS agent.

cc: @daniel-goldstein


danking:
Why do we not want the syslog? It seems useful to know what's going on in the system log.

jigold:
I'll close for now until we see what the impact of other changes are.

PR #11900, opened 2022-06-08
[k8s] Steps towards optimizing k8s

jigold:
cc: @daniel-goldstein


danking (approved):
LGTM, we should also migrate the node pools to 4 core ones to take advantage of the shrunken driver. I think before going through that pain we should ascertain the effect of this change on batch-driver maximum throughput.

PR #11902, opened 2022-06-09
[batch] aggressively reduce cpus per PR

danking:
Screen Shot 2022-06-09 at 10 32 21 AM

### DIFF
--- a/batch/sql/set-test-and-dev-pools-to-8-core-max-2.py
+++ b/batch/sql/set-test-and-dev-pools-to-8-core-max-2.py
@@ -1,0 +1,31 @@
+import os
+import asyncio
+from gear import Database
+
+
+async def main():
+    if os.environ['HAIL_SCOPE'] == 'deploy':
+        return
+
+    worker_cores = 8
+    max_instances = 2
+    max_live_instances = 2
+
+    db = Database()
+    await db.async_init()
+
+    await db.execute_update(
+        '''
+UPDATE pools
+SET worker_cores = %s
+''', (worker_cores))
+
+    await db.execute_update(
+        '''
+UPDATE inst_colls
+SET max_instances = %s, max_live_instances = %s
+''', (max_instances, max_live_instances))
+
+
+loop = asyncio.get_event_loop()
+loop.run_until_complete(main())
--- a/batch/sql/set-test-and-dev-pools-to-8-core-max-2.py
+++ b/batch/sql/set-test-and-dev-pools-to-8-core-max-2.py
@@ -4,7 +4,7 @@ from gear import Database
 
 
 async def main():
-    if os.environ['HAIL_SCOPE'] == 'deploy':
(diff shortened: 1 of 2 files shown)


danking:
ugh that tail is all service backend stuff; it has terrible utilization.

jigold:
Is part of the issue that the jobs are slow and there's a lot of them or is it the scheduler not keeping up? One hack we could consider is partial core JVM jobs. This would trivially get us 4x utilization.

danking:
I'm not sure

danking:
Screen Shot 2022-06-10 at 10 08 59 AM

Average utilization is marginally improved (maybe 2.5% -> 5%), but total number of wasted cores goes way down because the total number of cores is 16 rather than 16 * 8 = 128.

This also suggests the autoscaler/scheduler could be made substantially smarter.

danking:
While fixing a bug in globbing I introduced a bug in non-globbed code. I'll fix that and then get this PR in.

danking:
OK, I moved the file format test changes to https://github.com/hail-is/hail/pull/11906. This change can go in independently, but #11906 will even out the test job times and make developer experience better.

Service backend tests on #11904 which should be representative of a normal PR:

id | name | state | exit_code | duration
-- | -- | -- | -- | --
118 | test_hail_python_service_backend_0 | Success | Success 🎉 | 24 minutes
119 | test_hail_python_service_backend_1 | Success | Success 🎉 | 27 minutes
120 | test_hail_python_service_backend_2 | Success | Success 🎉 | 24 minutes
121 | test_hail_python_service_backend_3 | Success | Success 🎉 | 41 minutes
122 | test_hail_python_service_backend_4 | Success | Success 🎉 | 21 minutes

Service backend tests on this PR (albeit with #11906 which evens out test times):


id | name | state | exit_code | duration
-- | -- | -- | -- | --
118 | test_hail_python_service_backend_0 | Failed | Failure 🤷‍♀️ (1) | 31 minutes
119 | test_hail_python_service_backend_1 | Success | Success 🎉 | 31 minutes
120 | test_hail_python_service_backend_2 | Success | Success 🎉 | 28 minutes
121 | test_hail_python_service_backend_3 | Success | Success 🎉 | 33 minutes
122 | test_hail_python_service_backend_4 | Success | Success 🎉 | 26 minutes


I think there is almost no effect on service backend test times! We should really see if there's a way to improve the autoscaler & schedule to achieve this on its own.

batch/sql/set-test-and-dev-pools-to-8-core-max-2.py
```diff
@@ -0,0 +1,31 @@
+import os
+import asyncio
+from gear import Database
+
+
+async def main():
+    if os.environ['HAIL_SCOPE'] == 'deploy':
```
jigold, line 7: Can we add "dev" here?

jigold (dismissed):

jigold (approved):

PR #11903, opened 2022-06-09
[k8s] Reduce the number of replicas for other services

jigold:
Prometheus was using quite a lot of resources in the test namespaces so this should fix that.


danking (dismissed):
If it passes the tests I'm fine with this, but we should ask Daniel when returns if this seems reasonable

danking (approved):

PR #11904, opened 2022-06-09
[batch] Cleanup logging on the front end and driver

jigold:
The scheduling loop started log statements are pretty substantial. If you want to keep them, then we can revisit this once Daniel is back because he might be recording these metrics somewhere else.

### DIFF
--- a/batch/batch/driver/instance_collection/job_private.py
+++ b/batch/batch/driver/instance_collection/job_private.py
@@ -138,7 +138,6 @@ WHERE name = %s;
             log.info(f'not scheduling any jobs for {self}; batch is frozen')
             return True
 
-        log.info(f'starting scheduling jobs for {self}')
         waitable_pool = WaitableSharedPool(self.async_worker_pool)
 
         n_records_seen = 0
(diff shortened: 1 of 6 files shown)


batch/batch/driver/instance_collection/job_private.py
```diff
@@ -180,7 +179,8 @@ async def schedule_with_error_handling(app, record, id, instance):
 
         await waitable_pool.wait()
 
-        log.info(f'attempted to schedule {n_records_seen} jobs for {self}')
+        if n_records_seen > 0:
+            log.info(f'attempted to schedule {n_records_seen} jobs for {self}')
```
danking, line 183: As a general comment, I don't think gating log statements on these conditions will help much. I suspect most of our log volume comes when the system is under heavy load, i.e. n_records_seen >>> 0.

danking (approved):

jigold:
@daniel-goldstein Can you take a quick look at this and make sure none of the deleted log statements are used in your monitoring infrastructure? Thanks!

daniel-goldstein:
Seems fine to me. We pretty much just use MJS/MJC and rate limit logs from NGINX for monitoring

PR #11910, opened 2022-06-14
[query] Fix text export part file overwrite due to speculation

chrisvittal:
We do not make guarantees about the exact contents of a folder after
a parallel export. We do guarantee that all data is in the folder
however, so we must make it speculation safe. We do so by adding a UUID
to each part file we write for in lowered text export.


tpoterba:
looks good!

PR #11915, opened 2022-06-15
[query] Lower naive_coalesce

tpoterba:


hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -1597,6 +1597,24 @@ object LowerTableIR {
...
+          letBindings = lc.letBindings,
+          broadcastVals = lc.broadcastVals,
+          globals = lc.globals,
+          partitioner = lc.partitioner.copy(rangeBounds = lc.partitioner
+            .rangeBounds
+            .grouped(groupSize)
+            .toArray
+            .map(arr => Interval(arr.head.left, arr.last.right))),
+          dependency = lc.dependency,
+          contexts = mapIR(StreamGrouped(lc.contexts, groupSize)) { group => ToArray(group) },
```
danking, line 1614: @tpoterba fyi, I'm not sure if this is the right move, but I had to realize each group as an array in order to make the tests pass. In particular, at some point, we appear to stick a single context inside a structure and that obviously doesn't work unless its a concrete type like an Array.

danking (approved):

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
tpoterba: that's totally right. I think I made this change and then somehow deleted it when refactoring the code...

tpoterba: I'll PR a constructor requirement on TableStage that makes sure we never get a nested stream for contexts.

PR #11928, opened 2022-06-17
[batch] enable 5 JP instances in tests and alleviate failures

danking:


build.yaml
```diff
@@ -2035,6 +2035,8 @@ steps:
         script: /io/sql/add-nonpreemptible-pools.sql
       - name: set-test-and-dev-pools-to-8-core-max-2
         script: /io/sql/set-test-and-dev-pools-to-8-core-max-2.py
+      - name: set-test-and-dev-jpim-to-max-5
```
jigold, line 2038: Add `online: true`

jigold (changes requested):

jigold (approved):

PR #11931, opened 2022-06-21
[batch] Mount job secrets as read-only

daniel-goldstein:
These should never have been read-write. Caught this because a CI job I was modifying overwrote `/gsa-key/key.json` with `/test-gsa-key/key.json` which caused the Output step to use the test credentials instead of CI credentials.

I also removed an overriding definition of `secret_host_path` in `JVMJob`. I don't see why it should be different than what's defined in `Job` and using `host_path` seems quite dangerous.

Added a test that we can't `mv` a secret path and updated some existing tests that assumed we can overwrite secrets.

TODO: Update `build.yaml` to not `mv` any secrets or PRs will fail when this joins the mainline.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1364,8 +1364,7 @@ class Job:
                     }
                 )
 
-        secrets = job_spec.get('secrets')
-        self.secrets = secrets
+        self.secrets = job_spec.get('secrets')
         self.env = job_spec.get('env', [])
 
         self.project_id = Job.get_next_xfsquota_project_id()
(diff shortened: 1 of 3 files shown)


danking:
This seems good to me, any reason we shouldn't assign a reviewer and get this merged?

daniel-goldstein:
Interestingly I was have a consistent test failure here but never got the chance to diagnose it. Would like to come back to it once the glut of PRs is through. Would you rather keep it open or closed?

danking:
Let's keep open!

danking (approved):
This looks fine to me.

danking:
There's a merge conflict which I'm not confident I can resolve.

danking:
bump

danking:
Question: aren't the secrets written to a unique directory per-job? Why not let the user mutate as they see fit?

daniel-goldstein:
> Question: aren't the secrets written to a unique directory per-job? Why not let the user mutate as they see fit?

hm this is a fair point. I originally opened this because while it's per-job it is not unique *per-container* which is the original problem I found if you look at the PR body. I thought might as well make this read-only because I can't imagine wanting to mutate them but we could instead just make the credentials files per-container.

danking:
@daniel-goldstein maybe we can close since we now have an alternative path to not having keys at all?

PR #11933, opened 2022-06-21
[query] avoid rare test collection bug

danking:
Pytest sometimes uses a background thread to collect tests. That interacts badly
with asyncio. We avoid this by explicitly managing the event loop.


chrisvittal (approved):

danking:
this is actually somewhat common

PR #11936, opened 2022-06-21
[query] persist is checkpoint

danking:
I am not sure what persist should mean in the service backend. For some linear
algebra work, persist appears to be used to store the results of an expensive
query. In that setting, we should clearly checkpoint the dataset.

cc: @tpoterba

### DIFF
--- a/hail/python/hail/backend/service_backend.py
+++ b/hail/python/hail/backend/service_backend.py
@@ -10,7 +10,7 @@ import re
 import yaml
 from pathlib import Path
 
-from hail.context import TemporaryDirectory, tmp_dir, TemporaryFilename, revision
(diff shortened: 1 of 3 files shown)


hail/python/hail/backend/service_backend.py
```diff
@@ -679,10 +680,32 @@ def register_ir_function(self,
 
     def persist_expression(self, expr):
         # FIXME: should use context manager to clean up persisted resources
-        fname = TemporaryFilename().name
+        fname = TemporaryFilename(prefix='persist_expression').name
         write_expression(expr, fname)
         return read_expression(fname, _assert_type=expr.dtype)
```
danking, line 685: I'm not sure why we don't have unpersist_expression.

PR #11937, opened 2022-06-21
[query] Fix key pruning bug with respect to table/matrix row scans.

tpoterba:
Can lead to keys being pruned upstream, producing invalid scans.

CHANGELOG: Fixed correctness bug in scan order for `Table.annotate` and `MatrixTable.annotate_rows` in certain circumstances.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
@@ -1605,8 +1605,8 @@ object PruneDeadFields {
         val newRowType = newRow2.typ.asInstanceOf[TStruct]
         val child2Keyed = if (child2.typ.key.exists(k => !newRowType.hasField(k))) {
           val upcastKey = child2.typ.key.takeWhile(newRowType.hasField)
-          assert(requestedType.key.startsWith(upcastKey))
-          TableKeyBy(child2, child2.typ.key.takeWhile(newRowType.hasField))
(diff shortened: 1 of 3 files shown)


chrisvittal (approved):

chrisvittal:
This seems very broken.

hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
```diff
@@ -1593,9 +1603,11 @@ object PruneDeadFields {
         val child2 = rebuild(ctx, child, memo)
         val newRow2 = rebuildIR(ctx, newRow, BindingEnv(child2.typ.rowEnv, scan = Some(child2.typ.rowEnv)), memo)
         val newRowType = newRow2.typ.asInstanceOf[TStruct]
-        val child2Keyed = if (child2.typ.key.exists(k => !newRowType.hasField(k)))
+        val child2Keyed = if (child2.typ.key.exists(k => !newRowType.hasField(k))) {
+          val upcastKey = child2.typ.key.takeWhile(newRowType.hasField)
+          assert(requestedType.key.startsWith(upcastKey))
```
chrisvittal, line 1608: This new assertion is failing a bunch

chrisvittal (approved):
We need this fix, if my assertion is wrong, it doesn't make things more broken and it can be fixed later.

tpoterba:
yep.

tpoterba:
thanks!

PR #11938, opened 2022-06-21
[batch] Remove unneeded foreign key constraints

jigold:
Too many foreign key constraints are bad because they add compute time for every insert into a referenced table. Foreign key constraints also cause any insert on a referenced table to take out exclusive write locks on the referenced row. For example, every time we insert a new row into `attempt_resources` we're locking both `batches` and `jobs` unnecessarily which makes inserts effectively serial.

### DIFF
--- a/batch/sql/estimated-current.sql
+++ b/batch/sql/estimated-current.sql
@@ -249,7 +249,6 @@ CREATE TABLE IF NOT EXISTS `attempts` (
   `end_time` BIGINT,
   `reason` VARCHAR(40),
   PRIMARY KEY (`batch_id`, `job_id`, `attempt_id`),
-  FOREIGN KEY (`batch_id`) REFERENCES batches(id) ON DELETE CASCADE,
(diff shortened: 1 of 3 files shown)


jigold:
I verified that the right foreign key constraints were deleted in my namespace with dev deploy and I didn't touch the foreign key constraints in the other databases.

danking:
What kind of lock does a foreign key constraint take on the foreign row (or table)? I'm surprised it's an exclusive lock, I'd think you just need a shared lock on the foreign table. If it's shared, how is that serializing the inserts?

jigold:
You're right. I must have gotten confused with another db operation. These changes will still save some computation time and make it easier to reason about who is taking out locks on which tables when we are trying to understand where deadlocks are coming from.

danking:
The foreign key constraints are also a safety belt which protect us from a corrupted database. How much does this improve performance? Let's explicitly record the cost-benefit analysis in the PR comments so we can refer back to it if necessary.

daniel-goldstein (changes requested):
Could you put the performance results here when you're done with them (or link to zulip)?

jigold:
I'm closing this for now.

PR #11942, opened 2022-06-22
[batch] Update docker client timeout transient error message

daniel-goldstein:
Looks like they changed their error message. See [here](https://ci.hail.is/batches/3181369/jobs/80) for an example.


jigold (approved):
Is it worth keeping both cases for backwards compatibility reasons?

PR #11944, opened 2022-06-22
[batch] add pool labels

danking:


danking:
@jigold Do we still need this PR?

lgruen:
We still use this on our side, so it would be nice to get this merged. Unless there's a different mechanism now for similar functionality?

jigold:
I'm happy with this as long as the interface is not public in hailtop.batch. I'll look over it once more carefully to make sure nothing has changed since it was first opened and then take care of the rebase.

jigold:
@lgruen Can you remind me why you used the empty string for a pool with no label? Is it possible to use NULL or is that too much of a lift?

lgruen:
> @lgruen Can you remind me why you used the empty string for a pool with no label? Is it possible to use NULL or is that too much of a lift?

It was just easier to handle in the UI. With an empty string, everything "just worked", whereas for `NULL` you'd need to add a bit of special handling to allow specifying this as a label (and handling it separately from the empty string).

jigold:
Would you be okay if I either (1) modified your PR to use NULL and a slightly different database schema (separate table for `pool_labels`) or (2) put your PR as is and then did a second PR to do a database migration to fix it to have a separate table `pool_labels` so a pool can have multiple labels?

lgruen:
Of course, absolutely! Either way is perfectly fine, we can change this easily on our side. (2) would be slightly more convenient for us, as we'll be able to just do a standard upstream merge.

danking:
closing in favor of #12501

PR #11953, opened 2022-06-23
[build.yaml] on difference, dump up-to-date pinned requirements

danking:
Makes it easier for those of us not developing on linux.


daniel-goldstein (changes requested):
Does this lose the error code from diff if it's not 0? If so we should exit 1 or something so that the job still fails appropriately.

daniel-goldstein (approved):

PR #11954, opened 2022-06-23
[ci] allow core AUS devs to see CI pages

danking:
They all have broad emails now, which is what I used for their Batch usernames.

### DIFF
--- a/ci/ci/constants.py
+++ b/ci/ci/constants.py
@@ -32,4 +32,7 @@ AUTHORIZED_USERS = [
     User('patrick-schultz', 'pschultz', [COMPILER_TEAM]),
     User('pwc2', 'pcumming'),
     User('tpoterba', 'tpoterba', [COMPILER_TEAM]),
+    User('lgruen', 'lgruensc', []),
+    User('vladsaveliev', 'vsavelye', []),
+    User('illusional', 'mfrankli', []),
 ]


jigold (approved):

lgruen:
@danking Thanks a lot! Is it possible that our Hail Batch accounts aren't fully set up yet? @illusional and I are getting a 401 when trying to log into https://batch.hail.is, typically after a redirect loop.

danking:
Ah, hmm, I added your Broad account names, not your populationgenomics names.

If it's not too much trouble, I think the easiest thing to do is to visit http://auth.hail.is/signup and use your Broad credentials. Michael appears to have never had a Hail Batch account under his populationgenomics email anyway.

lgruen:
> I think the easiest thing to do is to visit http://auth.hail.is/signup and use your Broad credentials

That worked, thanks! I think I accidentally tried `http://batch.hail.is/signup` (or similar) before, which sent me into a redirect loop.

lgruen:
@danking I can now access https://auth.hail.is/user successfully (for the `lgruensc` user), but https://ci.hail.is/batches/3654512 or similar still results in a "redirected you too many times" followed by a 401.

daniel-goldstein:
Vlad should also contact us after signing up so we can fix his developer status.

lgruen:
@daniel-goldstein Yay, it's working now, thank you! Vlad is still on leave, but I'll let him know.

vladsavelyev:
Hey hey everyone, I'm back! I created a user (`vsavelye`, no last `v` - the Broad knows better!).
cc @danking @daniel-goldstein
Thank you! :)

daniel-goldstein:
@vladsavelyev Made you a developer on our GCP instance. Let me know if you run into any issues!

PR #11969, opened 2022-06-28
[query] Fix force_bgz partitioning for import_table/import_lines

tpoterba:


tpoterba:
we do for correctness, but not partitioning/performance it seems.

PR #11974, opened 2022-06-28
[pip] Unpin python dependencies in requirements.txt files

daniel-goldstein:


danking:
I'm good with this, shall we convert from draft?

danking (approved):
People are gonna love this, it's been a frequent request!

danking:
new lint problems

daniel-goldstein:
@danking Do you know what's up with this test? https://ci.hail.is/batches/5921822/jobs/68 Is the test wrong? I didn't change anything about the test in this PR so I feel like I must have somehow broken a dependency, but I'm unsure how because we don't use the google client library and I doubt it's a problem with aiohttp.

danking:
Oh, woah, that test does look wrong. It's concerning that its suddenly failing.

I'm not sure I care too much about tracking down exactly which dependency change caused this. We should fix the test obviously. We should add a test that verifies both `?a` and `?` have the expected data (in particular, that we didn't overwrite one with the other!).

daniel-goldstein:
Heh, so turns out that `test_weird_urls` is missing the `@pytest.mark.asyncio` decorator, and so it was getting skipped with a warning this whole time. The pytest upgrade added auto-detection of async tests and so it ran this broken test for the first time. I'm PR'ing to treat most warnings as errors in #12322 which should uncover the failed test and require a fix in that PR. Not sure why a myriad of other tests seem to failing but will tackle those next.

PR #11977, opened 2022-06-29
[auth] remove userinfo logs

danking:
Not helpful for debugging, we still get logs on errors

### DIFF
--- a/auth/auth/auth.py
+++ b/auth/auth/auth.py
@@ -759,8 +759,11 @@ async def on_cleanup(app):
 class AuthAccessLogger(AccessLogger):
     def __init__(self, logger: logging.Logger, log_format: str):
         super().__init__(logger, log_format)
-        self.exclude = [
-            ('GET', re.compile('/api/v1alpha/userinfo')),
+        self.exclude = [ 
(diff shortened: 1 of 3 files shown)


auth/auth/auth.py
```diff
@@ -755,6 +756,21 @@ async def on_cleanup(app):
...
+class AuthAccessLogger(AccessLogger):
+    def __init__(self, logger: logging.Logger, log_format: str):
+        super().__init__(logger, log_format)
+        self.exclude = [
+            ('GET', re.compile('/api/v1alpha/userinfo')),
+        ]
+
+    def log(self, request, response, time):
+        for method, path_expr in self.exclude:
+            if path_expr.fullmatch(request.path) and method == request.method:
```
jigold, line 768: I messed up this and was going to fix in another PR if you want to do the same here. `fullmatch` doesn't work for dev and pr namespaces. I was going to try with an end anchor and `match` or just set it to `match`.

jigold (changes requested):

auth/auth/auth.py
danking: I fixed all three.

jigold (approved):

PR #11983, opened 2022-06-29
[batch] Dont use worker_processes auto in nginx sidecar

daniel-goldstein:
The improvement here was really remarkable. Went from a shaky ~80 jobs per second to ~120 and NGINX max CPU use from 100% to 73%.


daniel-goldstein:
I imagine we'll want to do the same for our other uses of NGINX, do you want me to lump in `internal-gateway` to this change?

jigold:
Wow! That's great. Let's do a separate PR.

PR #11995, opened 2022-07-03
[batch] Undo trigger change to ignore updates for format_version < 3

jigold:
To make future modifications to the `attempt_resources_after_insert` trigger easier to follow, I have this PR which reverts the change to ignore any updates for batch format_version < 3.

Stacked on #11990

PR #11996, opened 2022-07-03
[batch] Add billing tables partitioned by date

jigold:
Stacked on #11995 

This PR creates all of the new `aggregated_*_resources_by_date` tables that will be used for real time billing and making our billing queries fast and also populates them! It will probably run for ~7-8 hours (online migration) and add an estimated 200 GB to the database. It will take around 5-6 hours to populate the tables and the remainder of the time is doing an audit.

I think we should whiteboard what is going on in person, but the general idea is as follows:
1. Revert any previous work and set the trigger back to the original state (idempotent)
2. Find the latest complete or open batch id. We know that a complete batch will not have updates to the attempts table. This is extremely important because the next steps can be done in parallel rather than serially.
3. Find offsets for complete batches up to the batch id from Step 1 in groups of 100 attempts
4. Randomize the offsets and have a burn in period of 5000 to avoid the birthday problem where we populate the `aggregated_*_resources_by_date` tables.
5. In 10-way parallelism (maxes out a 4 core database), randomly populate the tables for each chunk.
6. From the last offset (original first running batch id), we sequentially process attempts in groups of 100. We take note of where we are at with tracking any updates to the attempts table (`attempts_time_msecs_diff`), populate the `aggregated_*_resources_by_date` tables, and then do a final catchup step where we apply any updates from `attempts_time_msecs_diff` for any attempts that we have already processed.
7. Once we have reached the "end" of the attempts table, we lock all tables of interest especially the `attempts` table, and do one last final processing step before we add the new triggers that will auto-populate the `aggregated_*_resources_by_date` tables.
8. Then we perform an audit and make sure things look correct. (I might need to change or eliminate the billing_project audit query because there are 5 batches with ~20 jobs that aren't perfectly tracked when we did the original switch over to the new billing tables).
9. If there are any failures, we revert the triggers back to the original state.

Also to note, is the new table for billing_projects is keyed by (billing_project, user) which will make queries much faster so they don't have to scan the batches aggregated resources table.

I ran the migration successfully on a full test database and the audit was clean for jobs and batches except for the 5 batches that were running right when we started populating the original aggregated billing tables. 

I'd like to gather all feedback and then will run the migration one more time to do a final test.

Note, that most of the queries in the migration are not tested.

The key thing to double check is the triggers will continue to insert data into the old tables and we get the end points correct (start is inclusive and end is exclusive) as unlike the previous migration, this set of updates is NOT idempotent.

### DIFF
--- a/batch/sql/add-agg-billing-by-date.sql
+++ b/batch/sql/add-agg-billing-by-date.sql
@@ -1,7 +1,7 @@
-ALTER TABLE attempts ADD COLUMN added_to_per_day_rollups BOOLEAN DEFAULT FALSE, ALGORITHM=INSTANT;
+ALTER TABLE attempts ADD COLUMN migrated BOOLEAN DEFAULT FALSE, ALGORITHM=INSTANT;
 
-DROP TABLE IF EXISTS `aggregated_billing_project_user_resources`;
-CREATE TABLE IF NOT EXISTS `aggregated_billing_project_user_resources` (
+DROP TABLE IF EXISTS `aggregated_billing_project_user_resources_v2`;
+CREATE TABLE IF NOT EXISTS `aggregated_billing_project_user_resources_v2` (
   `billing_project` VARCHAR(100) NOT NULL,
   `user` VARCHAR(100) NOT NULL,
   `resource_id` INT NOT NULL,
@@ -11,10 +11,10 @@ CREATE TABLE IF NOT EXISTS `aggregated_billing_project_user_resources` (
   FOREIGN KEY (`billing_project`) REFERENCES billing_projects(name) ON DELETE CASCADE,
   FOREIGN KEY (`resource_id`) REFERENCES resources(`resource_id`) ON DELETE CASCADE
 ) ENGINE = InnoDB;
-CREATE INDEX aggregated_billing_project_user_resources ON `aggregated_billing_project_user_resources` (`user`);
+CREATE INDEX aggregated_billing_project_user_resources_v2 ON `aggregated_billing_project_user_resources_v2` (`user`);
 
-DROP TABLE IF EXISTS `aggregated_billing_project_user_resources_by_date`;
-CREATE TABLE IF NOT EXISTS `aggregated_billing_project_user_resources_by_date` (
+DROP TABLE IF EXISTS `aggregated_billing_project_user_resources_by_date_v2`;
+CREATE TABLE IF NOT EXISTS `aggregated_billing_project_user_resources_by_date_v2` (
   `billing_timestamp` DATE NOT NULL,
   `billing_project` VARCHAR(100) NOT NULL,
   `user` VARCHAR(100) NOT NULL,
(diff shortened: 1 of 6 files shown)


jigold:
Note to self to make sure there are no usages of deprecated SQL usage to make sure we are forward compatible.

batch/sql/populate_agg_billing_by_date.py
```diff
@@ -0,0 +1,797 @@
...
+  {where_cond}
+) AS t
+ON DUPLICATE KEY UPDATE aggregated_job_resources_by_date.`usage` = aggregated_job_resources_by_date.`usage` + new_usage;
+''',
+        query_args
+    )
+
+    record = await tx.execute_and_fetchone(
+        '''
+SELECT COUNT(*) AS counter FROM attempts_time_msecs_diff;
```
jigold, line 294: We should use LIMIT 1 with order by desc instead.

jigold:
Daniel -- I haven't tested this beyond dev deploy. Can you take a look first to make sure everything makes sense before I do the last test of both the attempt_resources format version < 3 PR changes and these changes to populate the by_date tables?

jigold:
Just for reference, I had to make the following changes:
1. There's a new table that tracks attempts that have been already aggregated. The reason I can't put this variable into the attempts table is due to a circular update that's not allowed.
2. I added a dummy variable to the attempts table that is just a way to get the update trigger to run.

We find all complete attempts from the batches state and update the dummy variable for the attempts table in chunks. Within the new trigger, if the attempt has not been aggregated yet, then we use `NEW.end - NEW.start` as the time to aggregate with. If the attempt has been aggregated, then we use the original way which is just to take the difference between New and Old.

At the end, we find all attempts that have not been aggregated (not in the new table that keeps track of whether an attempt has been aggregated) and update the dummy variable to run the trigger. I don't think we need to take any locks on the attempts table because any future writes that occur while we do the last processing will already be aggregated.

daniel-goldstein:
This looks great. All good to test I think

jigold:
cc: @danking

batch/sql/add-agg-billing-by-date.sql
```diff
@@ -0,0 +1,235 @@
...
+CREATE TABLE IF NOT EXISTS `aggregated_billing_project_user_resources_by_date` (
+  `billing_project` VARCHAR(100) NOT NULL,
+  `user` VARCHAR(100) NOT NULL,
+  `start_time` BIGINT NOT NULL,
+  `end_time` BIGINT NOT NULL,
+  `resource_id` INT NOT NULL,
+  `token` INT NOT NULL,
+  `usage` BIGINT NOT NULL DEFAULT 0,
+  PRIMARY KEY (`billing_project`, `user`, `start_time`, `end_time`, `resource_id`, `token`),
+  FOREIGN KEY (`billing_project`) REFERENCES billing_projects(name) ON DELETE CASCADE,
```
daniel-goldstein, line 22: Perhaps not on the critical path, but should we also make `billing_project` an integer ID here and not varchar?

batch/sql/add-agg-billing-by-date.sql
```diff
@@ -0,0 +1,235 @@
...
+  WHERE batch_id = NEW.batch_id AND job_id = NEW.job_id;
+
+  SELECT billing_project INTO cur_billing_project FROM batches WHERE id = NEW.batch_id;
+
+  SET msec_diff = (GREATEST(COALESCE(NEW.end_time - NEW.start_time, 0), 0) -
+                   GREATEST(COALESCE(OLD.end_time - OLD.start_time, 0), 0));
+
+  # do not want to add to the original billing tables if we are forcing an update
+  IF OLD.dummy_aggregated_by_date = NEW.dummy_aggregated_by_date THEN
+    INSERT INTO aggregated_billing_project_resources (billing_project, resource, token, `usage`)
```
daniel-goldstein, line 88: How come `usage` is quoted here?

batch/sql/populate_agg_billing_by_date.py
```diff
@@ -0,0 +1,308 @@
...
+        # find first batch id where the state is running
+
+        records = tx.execute_and_fetchall(
+            '''
+SELECT id FROM batches
+WHERE state = %s
+ORDER BY id
+LIMIT 1;
+''',
+            ('running',))
```
daniel-goldstein, line 171: Nit: I don't see why we need to use `%s` here, can we just inline `'running'`?

daniel-goldstein (changes requested):
This looks great just a couple clarifying questions. Also left one nit but if we wouldn't otherwise change/retest the PR it's not worth addressing the nit.

batch/sql/add-agg-billing-by-date.sql
jigold: I think it's a special word in MySQL.

batch/sql/add-agg-billing-by-date.sql
```diff
@@ -0,0 +1,235 @@
...
+
+  SELECT start_time, end_time INTO cur_start_time, cur_end_time
+  FROM attempts
+  WHERE batch_id = NEW.batch_id AND job_id = NEW.job_id AND attempt_id = NEW.attempt_id
+  LOCK IN SHARE MODE;
+
+  SET msec_diff = GREATEST(COALESCE(cur_end_time - cur_start_time, 0), 0);
+
+  SET start_time_agg_key = UNIX_TIMESTAMP(CAST(FROM_UNIXTIME(cur_end_time / 1000) AS DATE)) * 1000;
+  SET end_time_agg_key = UNIX_TIMESTAMP(ADDDATE(CAST(FROM_UNIXTIME(cur_end_time / 1000) AS DATE), INTERVAL 1 DAY)) * 1000;
```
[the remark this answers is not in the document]
jigold: The `FROM_UNIXTIME` is the equivalent of `datetime.datetime.from_timestamp()` in Python. `cur_end_time` is UTC time in msecs. We cast the DATETIME result to a DATE so the timestamp is actually midnight in UTC time on that date. `UNIX_TIMESTAMP` then converts that DATE object back to a UTC timestamp starting at midnight. The end time just adds 1 day to the time. Example output is here. Can you verify they line up with midnight to midnight in UTC time in mecs.

```
+----------+---------------+---------------+-------------+-------+-------------+
| batch_id | start_time    | end_time      | resource_id | token | usage       |
+----------+---------------+---------------+-------------+-------+-------------+
|   185348 | 1613606400000 | 1613692800000 |           2 |     0 |  4144294000 |
|   185348 | 1613606400000 | 1613692800000 |           5 |     0 | 84875141120 |
|   185348 | 1613606400000 | 1613692800000 |           6 |     0 |  4243757056 |
|   185348 | 1613606400000 | 1613692800000 |           7 |     0 | 15914088960 |
|   185348 | 1613606400000 | 1613692800000 |           9 |     0 |  4144294000 |
+----------+---------------+---------------+-------------+-------+-------------+
```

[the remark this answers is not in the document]
jigold: End point is assumed to be exclusive.

batch/sql/add-agg-billing-by-date.sql
jigold: Hmmm.... my inclination is the storage savings is not going to be huge as this table is substantially smaller than the other aggregated_*_resources_by_date tables. But it would be good to avoid future headaches. Let's add it to this PR.

jigold: Actually, we don't have an easy way to add an auto increment column for the `user` column as well since there's no users table. I think we should hold off on this change.

batch/sql/add-agg-billing-by-date.sql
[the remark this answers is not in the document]
daniel-goldstein: > We cast the DATETIME result to a DATE so the timestamp is actually midnight in UTC time on that date.

This is what I was missing. Without that I was really confused how we could get uniform buckets. Thanks!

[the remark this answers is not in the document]
daniel-goldstein: Yup, `1613606400000` is `Thursday, February 18, 2021 12:00:00 AM`

batch/sql/add-agg-billing-by-date.sql
daniel-goldstein: Fine by me, we can keep it in mind for the future

daniel-goldstein (dismissed):

jigold:
I made some changes. I had to rewrite the audits to not fill up the temp disk space and account for a bug in billing that was fixed for job private instances #10069. I'll test this afternoon after I figure out how to revert my first attempt.

jigold:
This should be good to go assuming the tests pass.

daniel-goldstein (approved):

PR #11997, opened 2022-07-03
[batch] Use the new aggregated by date billing tables

jigold:
Stacked on #12006 

This PR actually uses the new tables within the client application. The billing page should now load quickly. Note that the old aggregated billing tables are still in the database and being populated.

The key thing to note is I switched how we are computing the cost. First, I aggregate the usage by resource before multiplying by the resource rate and summing. Therefore, the old and new numbers should be close, but not identical in the UI.

The reviewer should double check that there are no references to `aggregated_*_resources$` tables (the ones that do not have the date interval) within the application code.

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1176,14 +1176,14 @@ LOCK IN SHARE MODE''',
 
         bp_cost_record = await tx.execute_and_fetchone(
             '''
-SELECT COALESCE(SUM(`usage` * rate), 0) AS cost
+SELECT COALESCE(SUM(t.`usage` * t.rate), 0) AS cost
 FROM (
   SELECT resource_id, CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
   FROM aggregated_billing_project_user_resources_v2
   LEFT JOIN resources on resources.resource_id = aggregated_billing_project_user_resources_v2.resource_id
   WHERE billing_project = %s
   GROUP BY resource_id
-)
+) AS t;
 ''',
             (billing_project,),
         )
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1176,22 +1176,16 @@ LOCK IN SHARE MODE''',
 
         bp_cost_record = await tx.execute_and_fetchone(
             '''
-SELECT billing_project, COALESCE(SUM(`usage` * rate), 0) AS cost
-FROM billing_projects
-LEFT JOIN (
-  SELECT billing_project, resource_id, CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
-  FROM billing_projects
-  LEFT JOIN aggregated_billing_project_user_resources_v2
-    ON billing_projects.name = aggregated_billing_project_user_resources_v2.billing_project
-  WHERE billing_projects.name = %s
-  GROUP BY aggregated_billing_project_user_resources_v2.billing_project, aggregated_billing_project_user_resources_v2.resource_id
-) AS usage_t ON usage_t.billing_project = billing_projects.name
-LEFT JOIN resources
-  ON resources.resource_id = usage_t.resource_id
-WHERE billing_projects.name = %s
-GROUP BY usage_t.billing_project;
+SELECT COALESCE(SUM(`usage` * rate), 0) AS cost
+FROM (
+  SELECT resource_id, CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
+  FROM aggregated_billing_project_user_resources_v2
+  LEFT JOIN resources on resources.resource_id = aggregated_billing_project_user_resources_v2.resource_id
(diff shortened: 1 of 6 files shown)


batch/batch/front_end/front_end.py
```diff
@@ -1157,18 +1156,18 @@ async def insert(tx):
 
         bp_cost_record = await tx.execute_and_fetchone(
             '''
-SELECT billing_projects.msec_mcpu, COALESCE(SUM(`usage` * rate), 0) AS cost
+SELECT COALESCE(SUM(`usage` * rate), 0) AS cost
 FROM billing_projects
-INNER JOIN aggregated_billing_project_resources
-  ON billing_projects.name = aggregated_billing_project_resources.billing_project
+INNER JOIN aggregated_billing_project_user_resources_by_date
+  ON billing_projects.name = aggregated_billing_project_user_resources_by_date.billing_project
```
jigold, line 1162: I need to fix all of these queries to aggregate by resource first before computing the cost.

jigold:
cc: @danking

jigold:
I checked all the queries are fast now even with the common table expressions.

batch/batch/front_end/front_end.py
```diff
@@ -1786,21 +1830,15 @@ async def parse_error(msg):
     if end is not None and start > end:
         return await parse_error('Invalid search; start must be earlier than end.')
 
-    where_conditions = ["billing_projects.`status` != 'deleted'"]
-    where_args = []
+    where_conditions = [
+        "billing_projects.`status` != 'deleted'",
+        "billing_timestamp >= %s",
```
daniel-goldstein, line 1835: I think this should be `billing_date`

batch/batch/front_end/front_end.py
```diff
@@ -1158,18 +1176,25 @@ async def insert(tx):
...
+  FROM billing_projects
+  LEFT JOIN aggregated_billing_project_user_resources_v2
+    ON billing_projects.name = aggregated_billing_project_user_resources_v2.billing_project
+  WHERE billing_projects.name = %s
+  GROUP BY aggregated_billing_project_user_resources_v2.billing_project, aggregated_billing_project_user_resources_v2.resource_id
+) AS usage_t ON usage_t.billing_project = billing_projects.name
+LEFT JOIN resources
+  ON resources.resource_id = usage_t.resource_id
 WHERE billing_projects.name = %s
+GROUP BY usage_t.billing_project;
```
daniel-goldstein, line 1192: Are the two `WHERE billing_projects.name = %s` here redundant? Might not matter but could we get rid of the outer WHERE statement by doing an inner join with the table produced by the subquery?

daniel-goldstein (changes requested):
So nice to get rid of those format version transformations. Just a couple points

batch/batch/front_end/front_end.py
jigold: I messed up by forgetting to switch the name of the column (got the variable name, but not the actual table schema name). We can fix this by either online renaming the column, but that means the two triggers will be out of sync and fail for potentially a second assuming no issues in the migration. Or we do a quick offline migration and guarantee no issues. Thoughts?

batch/batch/front_end/front_end.py
jigold: I'm not sure if we can guarantee order of how things are evaluated. This is a question for @danking

batch/batch/front_end/front_end.py
daniel-goldstein: Ah ok. The rename would be instantaneous? If it's basically just a blip then I think safest to do it offline and nip it in the bud, otherwise I think whatever is simplest and the least headache to keep things consistent.

jigold: My guess is it would be quick, but it will block the entire table. I think we just take it offline in case there's a problem we don't bork the system.

daniel-goldstein: Ok. I think if it would be quick better to get it out of the way sooner rather than later.

batch/batch/front_end/front_end.py
danking: I agree that an inner join avoids the need for a where, but it's not obvious to me that the optimizer is smart enough to propagate the where condition up to the outer-most inner join.

Hmm. Why are we joining and grouping by billing project if we're actually computing a point query? Seems like we should just:

```sql
select coalesce(sum(`usage` * rate)) as cost
from (
  select resource_id, cast(coalesce(sum(`usage`), 0) AS SIGNED) as `usage`
  from aggregated_billing_project_user_resources_v2
  left join resources on resources.resource_id = aggregated_billing_project_user_resources_v2.resource_id
  where billing_project = %s
  group by resource_id
)
```

jigold: You're right. I was over complicating this.

daniel-goldstein (approved):
LGTM

jigold:
I'll retry the PR Monday morning. Let's not merge on a Friday.

PR #12009, opened 2022-07-07
[dev-docs] add more context for kubernetes operations

danking:


dev-docs/kubernetes-operations.md
```diff
@@ -34,6 +34,11 @@ This will likely fail because the metrics-server uses some local disk to store m
 the *only* pod listed, then you can re-run the command with `--delete-emptydir-data`. You may lose a
 short period of k8s metrics. This will also impair the HorizontalPodAutoscaler.
 
+Other pods, such as grafana and memory may also be listed here. You can use `--delete-emptydir-data`
+to force them to be deleted as well. Deleting Grafana will cause a loss of metrics for a short
+period of time. Deleting memory will cause a loss of cache for Hail Query on Batch jobs using
```
daniel-goldstein, line 39: This wouldn't actually cause loss of *metrics*, right? AFAIK the only thing grafana is storing is its dashboard coniguration/settings, or is it also caching metrics data?

daniel-goldstein (changes requested):

dev-docs/kubernetes-operations.md
danking: Ah, you're right. It looks like the emptydir is just for a token. It's completely safe to blow away Grafana's tempdir.

PR #12010, opened 2022-07-07
[batch] Add support for open batches

jigold:
Stacked on #12120 

This PR implements open batches. Future PRs will expose the functionality to users in the Query Service and hailtop.batch. There's a [design document](https://docs.google.com/document/d/168Mq5nNATmSrwzL4h1oYGBIFmcNlgFyHr_Vwjx59Zss/edit#heading=h.ghe60pdzl3mv) that specified all of the changes.

To briefly summarize, there are now the concept of batch updates. Each job belongs inside an "update". The BatchClient has two types of builders now: UpdateBatchBuilder and CreateBatchBuilder. I play some tricks with the job ids being allowed to be negative numbers denoting relative to an offset to make things more efficient when updating a batch because you don't have to make multiple API calls to get the current job offset in the batch.

There are only two batch states in the database: `running` and `complete`. A batch starts out as `complete` until an update is committed at which point if the n_jobs > 0, it will change to `running`.

The main thing to look at implementation-wise is the new stored procedure `commit_batch_update` with a nasty update that will block progress on the batch while the update is in progress.

I added the updates to the UI. We can get rid of it if it's too confusing. There's also a `Time Updated` column now in the UI instead of `Time Closed`.
Screen Shot 2022-07-07 at 5 33 50 PM

Screen Shot 2022-07-07 at 5 34 07 PM

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -47,6 +47,7 @@ from hailtop.utils import (
     request_retry_transient_errors,
     retry_long_running,
     run_if_changed,
+    secret_alnum_string,
     time_msecs,
     time_msecs_str,
 )
@@ -708,7 +709,9 @@ async def _create_jobs(userdata: dict, job_specs: dict, batch_id: int, app: aioh
 
     record = await db.select_and_fetchone(
         '''
-SELECT `state`, format_version FROM batches
+SELECT `state`, format_version, update_id
+FROM batches
+LEFT JOIN batch_updates ON batches.id = batch_updates.batch_id
 WHERE user = %s AND id = %s AND NOT deleted;
 ''',
         (user, batch_id),
@@ -718,7 +721,9 @@ WHERE user = %s AND id = %s AND NOT deleted;
         raise web.HTTPNotFound()
     if record['state'] != 'open':
         raise web.HTTPBadRequest(reason=f'batch {batch_id} is not open')
+
     batch_format_version = BatchFormatVersion(record['format_version'])
+    update_id = record['update_id']
 
     try:
         validate_and_clean_jobs(job_specs)
@@ -1024,6 +1029,7 @@ VALUES (%s, %s, %s, %s);
             batches_inst_coll_staging_args = [
                 (
                     batch_id,
+                    update_id,
                     inst_coll,
                     rand_token,
                     resources['n_jobs'],
@@ -1034,8 +1040,8 @@ VALUES (%s, %s, %s, %s);
             ]
             await tx.execute_many(
                 '''
-INSERT INTO batches_inst_coll_staging (batch_id, inst_coll, token, n_jobs, n_ready_jobs, ready_cores_mcpu)
-VALUES (%s, %s, %s, %s, %s, %s)
+INSERT INTO batches_inst_coll_staging (batch_id, update_id, inst_coll, token, n_jobs, n_ready_jobs, ready_cores_mcpu)
+VALUES (%s, %s, %s, %s, %s, %s, %s)
 ON DUPLICATE KEY UPDATE
   n_jobs = n_jobs + VALUES(n_jobs),
   n_ready_jobs = n_ready_jobs + VALUES(n_ready_jobs),
@@ -1047,6 +1053,7 @@ ON DUPLICATE KEY UPDATE
             batch_inst_coll_cancellable_resources_args = [
                 (
                     batch_id,
+                    update_id,
                     inst_coll,
                     rand_token,
                     resources['n_ready_cancellable_jobs'],
@@ -1056,8 +1063,8 @@ ON DUPLICATE KEY UPDATE
             ]
             await tx.execute_many(
                 '''
-INSERT INTO batch_inst_coll_cancellable_resources (batch_id, inst_coll, token, n_ready_cancellable_jobs, ready_cancellable_cores_mcpu)
-VALUES (%s, %s, %s, %s, %s)
+INSERT INTO batch_inst_coll_cancellable_resources (batch_id, update_id, inst_coll, token, n_ready_cancellable_jobs, ready_cancellable_cores_mcpu)
+VALUES (%s, %s, %s, %s, %s, %s)
 ON DUPLICATE KEY UPDATE
   n_ready_cancellable_jobs = n_ready_cancellable_jobs + VALUES(n_ready_cancellable_jobs),
   ready_cancellable_cores_mcpu = ready_cancellable_cores_mcpu + VALUES(ready_cancellable_cores_mcpu);
@@ -1225,6 +1232,32 @@ VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s);
                 batch_spec.get('cancel_after_n_failures'),
             ),
         )
+
+        update_id = await tx.execute_and_fetchone(
+            '''
+SELECT update_id FROM batch_updates WHERE batch_id = %s;
+''',
+            (id,),
+        )
+
+        if update_id is None:
+            update_id = secret_alnum_string(8)
+
+        await tx.execute_insertone(
+            '''
+INSERT INTO batch_updates (batch_id, update_id, start_job_id, n_jobs, time_created)
+VALUES (%s, %s, %s, %s, %s)
+ON DUPLICATE KEY UPDATE n_jobs = n_jobs;
+''',
+            (
+                id,
+                update_id,
+                1,
+                batch_spec['n_jobs'],
+                now,
+            ),
+        )
+
         await tx.execute_insertone(
             '''
 INSERT INTO batches_n_jobs_in_complete_states (id) VALUES (%s);
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1333,7 +1333,7 @@ FROM batch_updates
 LEFT JOIN batches ON batch_updates.batch_id = batches.id
 LEFT JOIN batches_cancelled
   ON batch_updates.batch_id = batches_cancelled.id
-WHERE user = %s AND batch_updates.batch_id = %s AND batch_updates.update_id = %s AND NOT batches.deleted;            
+WHERE user = %s AND batch_updates.batch_id = %s AND batch_updates.update_id = %s AND NOT batches.deleted;
 ''',
             (batch_id, batch_id, update_id, user, batch_id, update_id),
         )
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -813,16 +813,18 @@ WHERE user = %s AND id = %s AND NOT deleted;
     bunch_start_job_id = None
 
     for spec in job_specs:
-        job_id = spec['job_id']
-        if job_id < 0:
-            job_id = update_start_job_id + abs(job_id) - 1
-            assert job_id >= update_start_job_id
-            spec['job_id'] = job_id
+        relative_job_id = spec.pop('relative_job_id')
+        job_id = update_start_job_id + relative_job_id - 1
+        spec['job_id'] = job_id
 
-        parent_ids = spec.pop('parent_ids', [])
-        parent_ids = [update_start_job_id + abs(pid) - 1 if pid < 0 else pid for pid in parent_ids]
+        absolute_parent_ids = spec.pop('absolute_parent_ids', [])
+        relative_parent_ids = spec.pop('relative_parent_ids', [])
+        relative_parent_ids = [update_start_job_id + pid - 1 for pid in relative_parent_ids]
+        parent_ids = absolute_parent_ids + relative_parent_ids
+
         for pid in parent_ids:
-            assert pid > 0
+            if not 0 < pid < job_id:
+                raise web.HTTPBadRequest(reason=f'invalid parent ids found in the spec: parent {pid}, job {job_id}')
 
         always_run = spec.pop('always_run', False)
 
@@ -1225,6 +1227,35 @@ async def create_batch(request, userdata):
     return web.json_response({'id': id, 'update_id': update_id, 'start_job_id': start_job_id})
 
 
+@routes.post('/api/v1alpha/batches/{batch_id}/update/{update_id}/update-fast')
+@rest_billing_project_users_only
+async def update_batch_fast(request, userdata, batch_id):
+    app = request.app
+    db: Database = app['db']
+    update_id = request.match_info['update_id']
+
+    if app['frozen']:
+        log.info('ignoring batch update request; batch is frozen')
+        raise web.HTTPServiceUnavailable()
+
+    user = userdata['username']
+    update_and_bunch = await request.json()
+    update_spec = update_and_bunch['update']
+    bunch = update_and_bunch['bunch']
+
+    try:
+        validate_batch_update(update_spec)
+    except ValidationError as e:
+        raise web.HTTPBadRequest(reason=e.reason)
+
+    n_update_jobs = update_spec['n_jobs']
+
+    start_job_id, n_jobs, token, attributes = await _create_batch_update(batch_id, update_id, user, n_update_jobs, db)
+    await _create_jobs(userdata, bunch, batch_id, update_id, app, update_start_job_id=start_job_id)
+    await _commit_update(app, batch_id, update_id, user, db)
+    return web.json_response({'start_job_id': start_job_id, 'n_jobs': n_jobs, 'token': token, 'attributes': attributes})
+
+
 @routes.post('/api/v1alpha/batches/{batch_id}/update/{update_id}')
 @rest_billing_project_users_only
 async def update_batch(request, userdata, batch_id):
@@ -1246,13 +1277,14 @@ async def update_batch(request, userdata, batch_id):
         raise web.HTTPBadRequest(reason=e.reason)
 
     n_update_jobs = update_spec['n_jobs']
-    start_job_id, new_n_jobs, token, attributes = await _create_batch_update(batch_id, update_id, user, n_update_jobs, db)
+    start_job_id, _, _, _ = await _create_batch_update(batch_id, update_id, user, n_update_jobs, db)
 
-    return web.json_response({'update_id': update_id,
-                              'start_job_id': start_job_id,
-                              'n_jobs': new_n_jobs,
-                              'token': token,
-                              'attributes': attributes})
+    return web.json_response(
+        {
+            'update_id': update_id,
+            'start_job_id': start_job_id,
+        }
+    )
 
 
 @routes.patch('/api/v1alpha/batches/{batch_id}/update/{update_id}/commit', name='commit-batch-update')
@@ -1278,7 +1310,7 @@ async def _commit_update(app: aiohttp.web.Application, batch_id: int, update_id:
     async def commit(tx):
         record = await tx.execute_and_fetchone(
             '''
-SELECT batches_cancelled.id IS NOT NULL AS cancelled, committed
+SELECT batches_cancelled.id IS NOT NULL AS cancelled, committed, start_job_id
 FROM batch_updates
 LEFT JOIN batches ON batches.id = batch_updates.id
 LEFT JOIN batches_cancelled
@@ -1297,7 +1329,7 @@ WHERE user = %s AND batch_updates.id = %s AND batch_updates.update_id = %s AND N
                 (batch_id, update_id, now),
             )
         except CallError as e:
-            # 2: wrong number of jobs
+            # 1: wrong number of jobs
             if e.rv['rc'] == 1:
                 expected_n_jobs = e.rv['expected_n_jobs']
                 actual_n_jobs = e.rv['actual_n_jobs']
@@ -1306,7 +1338,29 @@ WHERE user = %s AND batch_updates.id = %s AND batch_updates.update_id = %s AND N
                 )
             raise
 
-    await commit()  # pylint: disable=no-value-for-parameter
+        batch_record = await tx.select_and_fetchone(
+            '''
+SELECT `state`, format_version, batches_cancelled.id IS NOT NULL AS cancelled, token,
+  JSON_OBJECTAGG(batch_attributes.`key`, batch_attributes.`value`) as attributes, n_jobs
+FROM batches
+LEFT JOIN batches_cancelled
+  ON batches.id = batches_cancelled.id
+LEFT JOIN batch_attributes
+  ON batches.id = batch_attributes.batch_id
+WHERE user = %s AND batches.id = %s AND NOT deleted
+GROUP BY batches.id;
+''',
+            (user, batch_id),
+        )
+
+        return {
+            'start_job_id': record['start_job_id'],
+            'n_jobs': batch_record['n_jobs'],
+            'token': batch_record['token'],
+            'attributes': batch_record['attributes'],
+        }
+
+    result = await commit()  # pylint: disable=no-value-for-parameter
 
     await request_retry_transient_errors(
         client_session,
@@ -1315,7 +1369,7 @@ WHERE user = %s AND batch_updates.id = %s AND batch_updates.update_id = %s AND N
         headers=app['batch_headers'],
     )
 
-    return web.Response()
+    return web.json_response(result)
 
 
 async def _create_batch_update(
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1368,17 +1368,13 @@ async def _create_batch_update(batch_id: int, update_token: str, n_update_jobs: 
         assert n_update_jobs > 0
         record = await tx.execute_and_fetchone(
             '''
-SELECT update_id, committed, start_job_id FROM batch_updates
+SELECT update_id, start_job_id FROM batch_updates
 WHERE batch_id = %s AND token = %s;
 ''',
             (batch_id, update_token),
         )
 
         if record:
-            if record['committed']:
-                raise web.HTTPBadRequest(
-                    reason=f'update {record["update_id"]} for batch {batch_id} has already been committed'
-                )
             return record['update_id'], record['start_job_id']
 
         now = time_msecs()
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -75,7 +75,7 @@ from ..globals import BATCH_FORMAT_VERSION, HTTP_CLIENT_MAX_SIZE
 from ..inst_coll_config import InstanceCollectionConfigs
 from ..spec_writer import SpecWriter
 from ..utils import query_billing_projects
-from .validate import ValidationError, validate_and_clean_jobs, validate_batch
+from .validate import ValidationError, validate_and_clean_jobs, validate_batch, validate_batch_update
 
 # import uvloop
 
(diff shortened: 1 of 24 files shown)


jigold:
cc: @danking

jigold:
It's probably not going to pass CI on the first try after the rebase. I'm mainly looking for feedback on the API first.

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -410,21 +414,20 @@ async def delete(self):
...
-        self._jobs = []
-        self._submitted = False
-        self.attributes = attributes
-        self.callback = callback
+class BatchBuilder(abc.ABC):
+    _submitted: bool
+    _job_idx: int
+    _jobs: List[Job]
+    _job_specs: List[dict]
+    _client: 'BatchClient'
```
daniel-goldstein, line 422: I think this is better done with a constructor and the deriving class calling `super().__init__()`. I believe these fields are [shared by all instances of the class](https://stackoverflow.com/questions/1537202/difference-between-variables-inside-and-outside-of-init).

daniel-goldstein (changes requested):
I've only given a light read to the client/server interactions but I think I need some additional context. My biggest concern is definitely the negative job ids, could you explain the approach there?

Possibly relatedly, I'm suspicious of needing these separate subclasses for the `CreateBatchBuilder` and `UpdateBatchBuilder`. I feel that this bifurcation will lead to more special-casing than is necessary and we should have a slick enough abstraction to allow a BatchBuilder to easily create *and* update batches. I think I'll better be able to comment on the organization once I better understand the spec that we're submitting on updates.

jigold:
I think we should find a time to discuss this in person if the following explanation doesn't make sense. 

Right now, for small batches, we send one REST request to the server to both create the batch and create the jobs. However, if we want one REST request for an update (ideal for the query service and low latency jobs?), we have to use relative job ids because (1) we don't know the absolute start index of the jobs until we've gotten the start id of the update back from the server and (2) the job dependencies can be a mix of known job ids that have already been previously submitted in a previous creation/update. The negative job IDs are a way to deal with a mix of relative ids within an update and known, submitted job ids.

We can simplify things if we require all updates make two requests to the server to (1) get the start id and establish the update and then (2) submit new jobs with all absolute job IDs. I'd have to make sure this will actually simplify things because I also ran into a bifurcation in how the job IDs are handled in `BatchBuilder.create_job()`. We currently populate the spec with a job id before we've made any requests to the server. We need to know how many total jobs there are before we can figure out the job ids because the API for creating a new update requires reserving a block of job IDs which then returns the start id. This complexity is because we allow multiple updates to occur simultaneously to a batch.

daniel-goldstein:
This is indeed tricky, but I think there are other options here. For example, we could expand the spec so that a job specifies *either* a job_id or update_relative_job_id, and same for parents. This way the relative vs absolute job id is baked into the schema of the spec, and not inferred from the sign of the job id. Though similar in concept, I think that would be much less confusing. However,

> We can simplify things if we require all updates make two requests to the server to (1) get the start id and establish the update and then (2) submit new jobs with all absolute job IDs.

I'd like to try this first. I feel like if we get a really solid API and it has a couple of superfluous requests in some edge cases, we will be able to come up with good performance shortcuts that don't muddle the normal path. Since the Query Driver currently lives the full life of the batch and is likely to stay that way for a while, it will satisfy these conditions without making any extra requests.

daniel-goldstein:
> This complexity is because we allow multiple updates to occur simultaneously to a batch.

Right I left this out… I am unsure if we used relative IDs how we would prepare the following update without knowing the IDs of the jobs I'm submitting in this current update, so I think some supplementary request for a reservation of IDs will be necessary anyway.

daniel-goldstein:
Another option we might consider is specifying job dependencies with UUIDs instead of numerical IDs.

jigold:
Ok. I made a new draft of the client, but didn't touch `_create_jobs` on the front end at all. Hopefully this is better because job ids are no longer negative. If not, I can try again.

daniel-goldstein:
Ok, I think I'm getting a much better understanding of the situation, thank you! I think this is a lot easier for at least me to understand and if I'm not mistaken, it's mostly just moving around of the same code into `JobSpec`, right? I think `submit` is a lot easier to follow now, though I do sympathize with the pain of the three round-trips for small updates. What do you think about the following proposal?

1. Jobs are always submitted to the server with job_ids relative to the update that is being submitted (sorry for the complete reversal, this is just an idea!). This shouldn't have any affect on the current create/create-fast since only 1 update means relative and absolute job ids are the same. This also means that the client doesn't need to know what the `start_job_id` is ahead of submitting a bunch. Submitting the update could return the `start_job_id` such that the client can rectify its local `Job` objects with absolute IDs like you do in `_get_job_specs_as_json` after the fact. The server will take care of doing `absolute_job_id = update_start_job_id + in_update_job_id`.
2. parent_ids are represented as positive integers that are tagged with a type, either `in-update` or `absolute`. This is very similar to what you were previously doing with negative numbers, but I think baking it into the schema is going to be less foot-gunny than negative numbers.

Sorry if it's similar to what you were doing before but I think it has taken me a while to fully understand the limitations here. I also think it wouldn't take much changes to this current client implementation to do this.

jigold:
What do you think of rather than negative numbers or written out types we just do this? `1a` and `1r`. `a` is absolute and `r` is relative. `1` implies absolute so we don't break old clients. This way we encode the type of the job id without adding a lot more heft to the job spec which would be both more expensive to transmit and store long term.

daniel-goldstein:
In a similar vein, what if we just allow the client to either specify `parent_id` (which we keep the same and treat as absolute) or `rel_parent_id`? Even though we're using JSON and everything is a string anyway, I think I would like to keep anything that can be strictly numeric as numeric and not do additional parsing.

jigold:
ok. that's fine. I was worried you wanted something like this, which I was going to strongly oppose:

```
parent_ids = [
  {'id': 7, 'type': 'absolute'},
  {'id': 8, 'type': 'in_update'}
]
```

daniel-goldstein:
ya I think I wanted something like that in concept but agree with your sentiments about the implementation. having `parent_ids` and `rel_parent_ids` that are both just arrays of numbers, where the latter gets transformed and concatenated to the former seems like an efficient and nicely backwards-compatible implementation.

jigold:
Great! I think we are good on the client. I'll modify it later today to use the new scheme. Before we get into the nitty gritty of the actual implementation, can we move on to the UI components and the semantics of using the client in `test_batch.py`? There's also a change to how the batch fields `time_closed` and `time_created` are used. I also added `time_updated`. I think the new semantics are:

- time_created -- time the batch was created
- time_updated -- time the last update was committed
- time_completed -- time the last time n_completed == n_jobs regardless of whether there are outstanding updates that haven't been committed

For old batches:
- time_created => same
- time_closed => time_updated
- time_completed => same

I also changed what the batch state means:
> There are only two batch states in the database: running and complete. A batch starts out as complete until an update is committed at which point if the n_jobs > 0, it will change to running. There are no longer "open" batches. 

It's possible I didn't actually implement exactly what I described above as I was having a hard time figuring out whether time_updated should be equal to time_completed if the batch has no outstanding jobs to run. That's why I'd like to take a step back and make sure we agree on the interface.

daniel-goldstein:
I will take a look at `test_batch.py`, but I largely agree with the semantics you've written here. The one thing I'm unsure about is `time_completed`. Happy to have my mind changed but my initial instinct is that

- It's weird to call a batch complete if it has no jobs. It's elegant, and maybe the right thing, but it feels weird.
- A job should not have a time_completed if it is still running, i.e. there are pending/running jobs. My instinct says that `time_completed` should be optional, and should only have a value if there are >0 committed jobs in the batch and they are all complete.

daniel-goldstein:
As a caveat to my complete statement, maybe we consider batch with 0 jobs as "complete" internally, but just word it differently to the user on the front-end. I think my unease is primarily a user-facing one

daniel-goldstein:
Ok, this feels to me mostly a UI issue, and that we can make any adjustments in how these edge cases are presented to the user as a secondary concern, with the primary concern being a sensible database model which I think we agree on. I'll take a look at the code.

batch/test/test_batch.py
```diff
@@ -1056,3 +1056,99 @@ def test_job_private_instance_cancel(client: BatchClient):
...
     status = j.wait()
     assert status['state'] == 'Cancelled', str((status, b.debug_info()))
+
+
+def test_update_batch_no_deps(client: BatchClient):
+    b = client.create_batch()
+    b.create_job(DOCKER_ROOT_IMAGE, ['false'])
+    b = b.submit()
+
+    u = client.update_batch(b.id)
```
daniel-goldstein, line 1066: With the changes we've made to the client, is it possible now to elide this line? Is there any scenario where after the  first `b.submit()`, we would not be in the `update` mode? I think of the batch client now as a buffer of jobs where `submit` acts like `flush`, so we could ideally write:

```python
b = client.create_batch()
b.create_job(DOCKER_ROOT_IMAGE, ['false'])
b.submit()
j = b.create_job(DOCKER_ROOT_IMAGE, ['true'])
b.submit()
status = j.wait()
```

jigold: Sure! Good call.

jigold:
I think I addressed the changes in the client, but I'm sure it's not 100% correct yet. Will deal with any bugs later. Can we move on to the implementation? I'm thinking the best place to start is `_create_jobs` and then the SQL implementation of what it means to create and utilize the new updates infrastructure, but happy to do what's easiest for you.

daniel-goldstein (changes requested):
I like how `_create_jobs` is pretty unchanged, very clean. I like the approach of how the updates are stored and the staged resources, but I wonder if we need the additional table. What do you think about adding `update_id` and `committed` (bool) columns to the existing *_inst_coll_resources tables? Then we wouldn't copy over rows and delete them from the staging tables, and instead commit would just flip the `committed` flag on the rows that match the `update_id`. The main difference I see between these approaches is the tokenization in the staged tables and committing seems to be collapsing all of those into a single row (token = 0). How important is it to tokenize the staging tables? How come there's a token on the committed tables if we're setting it to 0?

batch/batch/front_end/front_end.py
```diff
@@ -663,8 +693,52 @@ async def create_jobs(request: aiohttp.web.Request, userdata: dict):
         raise web.HTTPServiceUnavailable()
 
     batch_id = int(request.match_info['batch_id'])
+    user = userdata['username']
+    db: Database = app['db']
+
+    record = await db.select_and_fetchone(
+        '''
+SELECT token AS update_id FROM batches
```
daniel-goldstein, line 701: Not sure I understand why the token of the batch is the update_id.

daniel-goldstein: ah, I see. This is the token that we use for idempotency of the REST API, and subsequent updates will use their own distinct token.

daniel-goldstein:
I think I'm seeing more where this approach is coming from, specifically we put batches as they exist today in a special category of having no updates and avoid the new code path in that case. An alternative which pairs with my above suggestion of not adding new staging tables is that all batches have at least 1 update. I feel like if we can force all batches down the new code path we'll be incentivized to make it really low overhead for batches that only submit jobs once, and that will benefit all batches. I may be wrong that we can do this with minimal performance tradeoff, but I'd like to try it first.

batch/sql/add-open-batches.sql
```diff
@@ -0,0 +1,167 @@
...
+CREATE TABLE IF NOT EXISTS `batch_updates` (
+  `id` BIGINT NOT NULL AUTO_INCREMENT,
+  `update_id` VARCHAR(40) NOT NULL,
+  `start_job_id` INT,
+  `n_jobs` INT NOT NULL,
+  `committed` BOOLEAN NOT NULL DEFAULT FALSE,
+  `time_created` BIGINT,
+  `time_committed` BIGINT,
+  PRIMARY KEY (`id`, `update_id`),
+  FOREIGN KEY (`id`) REFERENCES batches(id) ON DELETE CASCADE
```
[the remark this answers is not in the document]
jigold: You're right. Copy paste error from the original batches table.

batch/sql/add-open-batches.sql
```diff
@@ -0,0 +1,167 @@
...
+    INTO staging_n_jobs, staging_n_ready_jobs, staging_ready_cores_mcpu
+    FROM batch_updates_inst_coll_staging
+    WHERE batch_id = in_batch_id AND update_id = in_update_id
+    FOR UPDATE;
+
+    SELECT user INTO cur_user FROM batches WHERE id = in_batch_id;
+
+    IF staging_n_jobs = expected_n_jobs THEN
+      UPDATE batches SET `state` = IF(expected_n_jobs != 0, 'running', state),
+        time_updated = in_timestamp,
```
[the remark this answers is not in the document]
jigold: Good idea!

batch/batch/front_end/front_end.py
jigold: I think this is a backwards compatibility thing for old clients.

jigold: The code path for new clients is the function below.

jigold:
> I think I'm seeing more where this approach is coming from, specifically we put batches as they exist today in a special category of having no updates and avoid the new code path in that case. An alternative which pairs with my above suggestion of not adding new staging tables is that all batches have at least 1 update. I feel like if we can force all batches down the new code path we'll be incentivized to make it really low overhead for batches that only submit jobs once, and that will benefit all batches, as well as simplifying the mental model. I may be wrong that we can do this with minimal performance tradeoff, but I'd like to try it first.

Can you elaborate more? I'm not sure which code paths you are referring to.

jigold:
All tables that are written to in `_create_jobs` must be tokenized or the job insertion will be serial. This is the reason for the tokens in the staging tables. I had to add another layer of staging which is why the original cancellable resources table has a token, but now the data is collapsed into a single row with token 0 in the new stored procedure.

I think your idea of adding new columns to the existing tables is better and is now something we can do. When I wrote this PR, MySQL v8.0 wasn't an option.

jigold:
Also, I'm not opposed to doing a migration of the existing batches to having rows in the updates table. It will just take some work to write a chunking script because we don't want to do the migration all in one transaction. Luckily this code doesn't have to be as well tested as other migrations.

jigold:
I'm adding an additional migration PR that will put in the backwards compatibility for the updates table etc.

daniel-goldstein:
> > I think I'm seeing more where this approach is coming from, specifically we put batches as they exist today in a special category of having no updates and avoid the new code path in that case. An alternative which pairs with my above suggestion of not adding new staging tables is that all batches have at least 1 update. I feel like if we can force all batches down the new code path we'll be incentivized to make it really low overhead for batches that only submit jobs once, and that will benefit all batches, as well as simplifying the mental model. I may be wrong that we can do this with minimal performance tradeoff, but I'd like to try it first.
> 
> Can you elaborate more? I'm not sure which code paths you are referring to.

Mainly that the commit procedure branches on whether the start id is 1 and that we sometimes grab the update id from the batch token and sometimes from the update table. Not very different code paths but slightly different, which could lead to some confusion.

jigold:
Sorry for not getting this done quicker. There's two new soon to be PRs in the stack that you can see as commits here:
- [Add infrastructure for updates](https://github.com/hail-is/hail/pull/12010/commits/72ff68e628b97bae439d04d4cb45e8508941e8bb)
- [Cleanup adding update id infrastructure](https://github.com/hail-is/hail/pull/12010/commits/6364402e965a4f33248eba21639642e14a6f82be)

I'll make PRs for them on Monday once you give me the green light that no other major database changes are needed.

jigold:
Also, having the migration be online and in multiple stages was making my head hurt a lot. Maybe on Monday we should sit down and white board this and make sure there's no edge cases.

daniel-goldstein:
Felt quick to me! Ya we can talk this over on Monday but from a quick glance seems like a reasonable approach.

jigold:
> Mainly that the commit procedure branches on whether the start id is 1

I was trying to avoid an expensive query to update all of the n_pending_parents when we know it's the first update being committed. I think I'd prefer to keep this branching there.

daniel-goldstein (changes requested):
Just some small comments. Going to move over to reviewing the real time billing.

batch/batch/front_end/front_end.py
jigold: How do we deprecate it? The old clients are going to use this endpoint.

daniel-goldstein: I just meant leave a comment so engineers know that this endpoint isn't going to be used in the future.

In the future we might want to do something where we add a "Deprecated" response header for deprecated endpoints which our python client can identify and then spit out a warning that the user might want to upgrade their hail installation. That's a separate issue though.

batch/batch/front_end/front_end.py
```diff
@@ -1109,15 +1204,17 @@ async def create_batch_fast(request, userdata):
...
-    try:
-        await _create_jobs(userdata, bunch, batch_id, app)
-    except web.HTTPBadRequest as e:
-        if f'batch {batch_id} is not open' == e.reason:
-            return web.json_response({'id': batch_id})
-        raise
-    await _close_batch(app, batch_id, user, db)
-    return web.json_response({'id': batch_id})
+
+    update_id = batch_spec.get('update_id') or secret_alnum_string(8)
```
jigold, line 1210: These four operations should probably use the same transaction. This is probably the same for `update-fast` as well.

jigold:
I was having trouble figuring out how to handle the token and the attributes in hailtop.batch_client.aioclient.Batch. When we create an update from a Batch that already existed perhaps in a different process, we don't have the attributes and token. I made a contract where `commit_update` always returns the token and attributes regardless of whether the BatchBuilder already has that infromation. However, we could also get that information available lazily and cache the result. In addition, the `n_jobs` returned to the client are the number of jobs that are committed and not the same as the `n_jobs` in the batches table.

Things to do before merging:
1. Get rid of the batch updates additions to the UI2. 
2. Double check the GCP LogsExplorer to make sure there are no silent error messages especially with regards to cancellation.
3. Have @danking look over the SQL stored procedure for `commit_batch_update` to make sure that query is going to perform as good as what is possible given the complexity of the check.
4. Run a test batch with the old client (I just checked out the current version of main). You need to make sure both create and create-fast are accounted for and succeed. I've been using the following script to make sure we're using the slow path in addition to the fast path with a regular small test job:

```python3
from hailtop.batch import ServiceBackend, Batch
import secrets

backend = ServiceBackend(billing_project='hail')
b = Batch(backend=backend)
# 8 * 256 * 1024 = 2 MiB > 1 MiB max bunch size
for i in range(8):
    j1 = b.new_job()
    long_str = secrets.token_urlsafe(256 * 1024)
    j1.command(f'echo "{long_str}" > /dev/null')
batch = b.run()
```

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -383,9 +491,13 @@ async def last_known_status(self):
             return await self.status()  # updates _last_known_status
         return self._last_known_status
 
+    # TODO I think there's a decision here for what the semantics should be
+    # in a world of multiple concurrent batch updates.
+    # Should this be wait on the whole batch to reach completion or just
+    # the most recently-submitted update?
```
daniel-goldstein, line 497: Heh ya this is Danger if the job submits more jobs to its own batch. It would never complete

batch/test/test_accounts.py
```diff
@@ -75,12 +75,12 @@ async def test_bad_token():
...
-        b = bc.create_batch()
-        j = b.create_job(DOCKER_ROOT_IMAGE, ['false'])
-        await b.submit()
+        builder = bc.create_batch()
+        builder.create_job(DOCKER_ROOT_IMAGE, ['false'])
+        b = await builder.submit()
         assert False, str(await b.debug_info())
     except aiohttp.ClientResponseError as e:
-        assert e.status == 401, str((e, await b.debug_info()))
+        assert e.status == 401
```
danking, line 83: Why remove the debugging information?

batch/test/test_accounts.py
```diff
@@ -400,7 +383,7 @@ async def test_billing_limit_zero(
         bb = client.create_batch()
         batch = await bb.submit()
     except aiohttp.ClientResponseError as e:
-        assert e.status == 403 and 'has exceeded the budget' in e.message, str(await batch.debug_info())
+        assert e.status == 403 and 'has exceeded the budget' in e.message
```
danking, line 386: same question here

batch/test/test_accounts.py
```diff
@@ -172,23 +172,6 @@ async def test_close_reopen_billing_project(dev_client: BatchClient, new_billing
     ], r
 
 
-async def test_close_billing_project_with_open_batch_errors(
```
danking, line 175: This raises an interesting question. What happens when a billing project is closed but a batch still has running jobs? A closed billing project should not receive any new bills. I think closing a billing project needs to cancel any "running" batches.

How should I think about the states of a batch after this change?

batch/test/test_batch.py
```diff
@@ -258,19 +258,19 @@ def test_unsubmitted_state(client: BatchClient):
         j.wait()
 
     builder.submit()
-    with pytest.raises(ValueError):
-        builder.create_job(DOCKER_ROOT_IMAGE, ['echo', 'test'])
+    builder.create_job(DOCKER_ROOT_IMAGE, ['echo', 'test'])
+    assert j.id == 1
```
danking, line 262: This test needs a new/better name. I might also ask, what is this test really testing now? It seems a bit scattered.

batch/test/test_batch.py
```diff
@@ -258,19 +258,19 @@ def test_unsubmitted_state(client: BatchClient):
...
 
 
 def test_list_batches(client: BatchClient):
     tag = secrets.token_urlsafe(64)
-    b1 = client.create_batch(attributes={'tag': tag, 'name': 'b1'})
-    b1.create_job(DOCKER_ROOT_IMAGE, ['sleep', '3600'])
-    b1 = b1.submit()
+    builder1 = client.create_batch(attributes={'tag': tag, 'name': 'b1'})
```
danking, line 267: total aside, but I'm a fan of `bb` for batch builder and `b` for the batch.

batch/test/test_batch.py
```diff
@@ -310,13 +310,13 @@ def assert_batch_ids(expected: Set[int], q=None):
 
 
 def test_list_jobs(client: BatchClient):
-    b = client.create_batch()
-    j_success = b.create_job(DOCKER_ROOT_IMAGE, ['true'])
-    j_failure = b.create_job(DOCKER_ROOT_IMAGE, ['false'])
-    j_error = b.create_job(DOCKER_ROOT_IMAGE, ['sleep 5'], attributes={'tag': 'bar'})
-    j_running = b.create_job(DOCKER_ROOT_IMAGE, ['sleep', '1800'], attributes={'tag': 'foo'})
+    builder = client.create_batch()
```
danking, line 313: Ideally all these, which I assume are to fix type lints, would be in a separate PR not included in this PR for ease of review.

hail/python/hailtop/batch_client/aioclient.py
danking: I was just planning to type out this question. I don't think wait means anything sensible any more. Should we deprecate it and encourage waiting on a specific job?

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -453,89 +566,103 @@ def _create_job(self,
...
-            job_spec['network'] = network
-        if unconfined:
-            job_spec['unconfined'] = unconfined
-        if user_code:
-            job_spec['user_code'] = user_code
-
-        self._job_specs.append(job_spec)
-
-        j = Job.unsubmitted_job(self, self._job_idx)
+        job_spec = JobSpec(
```
danking, line 603: In the future, AFAICT, this change can happen independently of the open batches change and should happen in a separate PR for ease of review.

batch/batch/front_end/front_end.py
```diff
@@ -671,18 +695,59 @@ def check_service_account_permissions(user, sa):
...
+''',
+            (user, batch_id),
+        )
+    ]
+
+    if not records:
+        raise web.HTTPNotFound()
+
+    if len(records) > 1:
+        raise web.HTTPBadRequest(reason='Use /api/v1alpha/batches/{batch_id}/updates/{update_id}/jobs/create instead')
```
danking, line 730: I think this message should have a bit more context. Maybe: "Updating an extant batch can only be done with the updates endpoint, not this endpoint.".

batch/batch/front_end/front_end.py
```diff
@@ -671,18 +695,59 @@ def check_service_account_permissions(user, sa):
...
+
+@routes.post('/api/v1alpha/batches/{batch_id}/updates/{update_id}/jobs/create')
+@rest_billing_project_users_only
+async def update_create_jobs(request: web.Request, userdata: dict, batch_id: int):
+    app = request.app
+    db: Database = app['db']
+
+    if app['frozen']:
+        log.info('ignoring batch create request; batch is frozen')
+        raise web.HTTPServiceUnavailable()
```
danking, line 745: Aside, this feels like it should be a decorator or something.

batch/batch/front_end/front_end.py
```diff
@@ -284,9 +285,13 @@ async def _query_batch_jobs(request, batch_id):
...
-  ON jobs.batch_id = job_attributes.batch_id AND
-    jobs.job_id = job_attributes.job_id AND
-    job_attributes.`key` = 'name'
+    ON jobs.batch_id = job_attributes.batch_id AND
+      jobs.job_id = job_attributes.job_id AND
+      job_attributes.`key` = 'name'
+  LEFT JOIN batch_updates
+    ON jobs.batch_id = batch_updates.batch_id AND
+      jobs.job_id >= batch_updates.start_job_id AND
+      jobs.job_id < batch_updates.start_job_id + batch_updates.n_jobs
```
danking, line 294: This will scan every batch_update whose start id is before this job id, for the given batch. Ergo, this page is now O(n_updates), not O(1).

You haven't told the database that these are disjoint intervals so it can't do anything smarter. I think the only way to make this efficient is to store the update id in the job record. What the arguments against putting the update id in the job table?

danking (changes requested):

batch/batch/front_end/front_end.py
daniel-goldstein: I don't believe we've had this discussion, but I think that's a very reasonable approach and more intuitive than looking at the range.

batch/test/test_accounts.py
daniel-goldstein: b (as a Batch, not a BatchBuilder) won't actually exist if there's an error raised in this block, so the debug information will always be a None dereference.

daniel-goldstein: We should perhaps add debug_info to a BatchBuilder.

batch/batch/front_end/front_end.py
jigold: Just make sure if you need to add an index on the jobs table for the update id that it will be instant. I’m guessing it’s probably not and will shut batch down for at least an hour.

daniel-goldstein: aren't y'all supposed to be on vacation

jigold:
Closing in favor of #12199

PR #12015, opened 2022-07-08
[docker] Use updated cache tags in Makefiles

daniel-goldstein:
Follow up for #11999

cc @danking

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -2,6 +2,11 @@
 
 set -ex
 
+if [ -z "${NAMESPACE}" ]; then
+    echo "Must specify a NAMESPACE environment variable"
+    exit 1;
+fi
+
 CONTEXT="$(cd $1 && pwd)"
 DOCKERFILE="$CONTEXT/$2"
 REMOTE_IMAGE_NAME=$3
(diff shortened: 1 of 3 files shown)


docker-build.sh
```diff
@@ -7,16 +7,26 @@ DOCKERFILE="$CONTEXT/$2"
 REMOTE_IMAGE_NAME=$3
 EXTRA_CACHE=$4
 
-CACHE_NAME=$(echo $REMOTE_IMAGE_NAME | sed -E 's/(:[^:]+)(@[^@]+)?$//'):cache
+WITHOUT_TAG=$(echo $REMOTE_IMAGE_NAME | sed -E 's/(:[^:]+)(@[^@]+)?$//')
+MAIN_CACHE=${WITHOUT_TAG}:cache
+
+NAMESPACE=${NAMESPACE:-"default"}
```
jigold, line 13: Can we make NAMESPACE a required variable? I'm worried about pushing to default without intending to.

jigold (changes requested):

jigold (approved):

PR #12017, opened 2022-07-08
[query] import_gvs

tpoterba:


hail/python/hail/methods/impex.py
```diff
@@ -3206,8 +3218,48 @@ def add_reference_allele(mt):
...
+    )
+
+    ft_criteria = (hl.range(vd.LGT.ploidy)
+                   .map(lambda idx: vd.LA[vd.LGT[idx]])
+                   .filter(lambda x: x != 0)
+                   .fold(lambda tuple, called_idx: hl.struct(
+        any_no=tuple[0] | vd.allele_NO[called_idx-1],
+        any_yes=tuple[1] | vd.allele_YES[called_idx-1],
+        all_ok_vqslod=tuple[2] & vd.allele_OK[called_idx-1])
+    , hl.struct(any_no=False, any_yes=False, all_ok_vqslod=False)))
```
RoriCremer, line 3257: Instead of True/False/NA, would it be possible to do FAIL/PASS/. ?

tpoterba: Replacing an entry-level boolean (1 byte in Hail representations) with a string is going to be quite expensive. The gnomAD team has used filters this way (boolean fields) in the past and has had few problems.

hail/python/hail/methods/impex.py
```diff
@@ -3161,13 +3171,15 @@ def add_reference_allele(mt):
...
         var_ht = var_ht.transmute(locus=translate_locus(var_ht.location),
-                          local_alleles=var_ht.alt.split(','),
-                          LGT=hl.parse_call(var_ht.GT),
-                          GQ=hl.int32(var_ht.GQ),
-                          RGQ=hl.int32(var_ht.RGQ))
+                                  local_alleles=var_ht.alt.split(','),
+                                  LGT=hl.parse_call(var_ht.GT),
+                                  LAD=var_ht.AD.split(',').map(lambda x: hl.int32(x)),
+                                  GQ=hl.int32(var_ht.GQ),
+                                  RGQ=hl.int32(var_ht.RGQ))
```
RoriCremer, line 3178: AS_QUALapprox and QUALapprox would be welcome additions, I will add them to the spec/ query

hail/python/hail/methods/impex.py
RoriCremer: iiiinteresting. Okay, this is helpful to know! And maybe this would mean adding the "culprit" string would be very expensive as well

tpoterba: Yes, though we could use an integer to indicate state (we've been meaning to add enumeration types but haven't had a chance, that would make it easier).

hail/python/hail/methods/impex.py
```diff
@@ -2972,6 +2972,325 @@ def import_avro(paths, *, key=None, intervals=None):
...
+    vd = vd.annotate_rows(
+        allele_NO=vd.alleles[1:].map(
+            lambda allele: hl.coalesce(vd.as_vqsr.get(allele).yng_status == 'N', False)),
+        allele_YES=vd.alleles[1:].map(
+            lambda allele: hl.coalesce(vd.as_vqsr.get(allele).yng_status == 'Y', True)),
+        allele_OK=vd.alleles[1:].map(
+            lambda alt: hl.coalesce(vd.as_vqsr.get(alt).vqslod > hl.if_else(hl.is_snp(vd.alleles[0], alt),
+                                                                            truth_sensitivity_snp_threshold,
+                                                                            truth_sensitivity_indel_threshold),
+                                    True)),
```
RoriCremer, line 3277: we want to be comparing vd.as_vqsr.get(alt).vqslod and snp_vqslod_threshold.min_vqslod which we just created with the method above.

hail/python/hail/methods/impex.py
```diff
@@ -2972,6 +2972,325 @@ def import_avro(paths, *, key=None, intervals=None):
...
+                             )
+
+    vd = vd.annotate_rows(
+        allele_NO=vd.alleles[1:].map(
+            lambda allele: hl.coalesce(vd.as_vqsr.get(allele).yng_status == 'N', False)),
+        allele_YES=vd.alleles[1:].map(
+            lambda allele: hl.coalesce(vd.as_vqsr.get(allele).yng_status == 'Y', True)),
+        allele_OK=vd.alleles[1:].map(
+            lambda alt: hl.coalesce(vd.as_vqsr.get(alt).vqslod > hl.if_else(hl.is_snp(vd.alleles[0], alt),
+                                                                            truth_sensitivity_snp_threshold,
```
RoriCremer, line 3275:
```suggestion
                                                                            vd.snp_vqslod_threshold.min_vqslod,
```

hail/python/hail/methods/impex.py
```diff
@@ -2972,6 +2972,325 @@ def import_avro(paths, *, key=None, intervals=None):
...
+                                                                            truth_sensitivity_indel_threshold),
```
RoriCremer, line 3276:
```suggestion
                                                                            vd.indel_vqslod_threshold.min_vqslod),
```

hail/python/hail/methods/impex.py
RoriCremer: (and I did not test it out, but presumably the same for the indels)

hail/python/hail/methods/impex.py
```diff
@@ -2972,6 +2972,325 @@ def import_avro(paths, *, key=None, intervals=None):
...
+    )
+
+    ft_criteria = (hl.range(vd.LGT.ploidy)
+                   .map(lambda idx: vd.LA[vd.LGT[idx]])
+                   .filter(lambda x: x != 0)
+                   .fold(lambda tuple, called_idx: hl.struct(
+        any_no=tuple[0] | vd.allele_NO[called_idx-1],
+        any_yes=tuple[1] | vd.allele_YES[called_idx-1],
+        all_ok_vqslod=tuple[2] & vd.allele_OK[called_idx-1])
+    , hl.struct(any_no=False, any_yes=False, all_ok_vqslod=True)))
```
RoriCremer, line 3287: if 1 is a good allele and 2 is a bad allele, we want the following:

1/1  pass (true)
1/2 pass (true) <-- I dont think this is being handled correctly. We dont want to remove the one good allele, even if the other is bad
2/2 fail (false)
0/1 pass (true)
0/2 fail (false)

RoriCremer: I think we could take out "no" for now, since we dont have a use case for it

I think we would want this to be: all_no and any_yes and any_ok_vqslod
so the "|"s and "&"s would swap and the default values too

hail/python/hail/methods/impex.py
```diff
@@ -2972,6 +2972,337 @@ def import_avro(paths, *, key=None, intervals=None):
...
+    is_snp = vd.alleles[1:].map(lambda alt: hl.is_snp(vd.alleles[0], alt))
+    vd = vd.annotate_rows(
+        allele_NO=vd.alleles[1:].map(
+            lambda allele: hl.coalesce(vd.as_vqsr.get(allele).yng_status == 'N', False)),
+        allele_YES=vd.alleles[1:].map(
+            lambda allele: hl.coalesce(vd.as_vqsr.get(allele).yng_status == 'Y', True)),
+        allele_is_snp=is_snp,
+        allele_OK=hl._zip_func(is_snp, vd.alleles[1:],
+                               f=lambda is_snp, alt:
+                               hl.coalesce(vd.as_vqsr.get(alt).vqslod >
```
RoriCremer, line 3278:
```suggestion
                               hl.coalesce(vd.as_vqsr.get(alt).vqslod >=
```

RoriCremer: image
https://files.slack.com/files-pri/T0166C33N6A-F042M04CG74/screen_shot_2022-09-15_at_3.27.50_pm.png

Greater than or equal to should pass

danking:
@tpoterba , should we get a reviewer on this and get it merged? It seems to me that development on import_gvs might have reached a stable state now?

tpoterba:
Let's wait until the full callset goes through, should be this week.

danking:
I presume this has stabilized enough to be merged now?

tpoterba:
Need to make a smaller set of avro example files to check into the repo for automated tests. I have been writing/running tests on 10 full samples, can't automate that. I suppose maybe that shouldn't block this from being merged, though I don't want to leave that thread too long.

tpoterba:
has been split into a separate repo and necessary hail improvements 🎉

PR #12022, opened 2022-07-12
[query] Fix Let construction in MatrixFilterCols randomness handling

chrisvittal:


patrick-schultz (approved):
:facepalm: Thanks Chris!

PR #12023, opened 2022-07-12
[batch] Cache tokens for job bunches on the driver to avoid db queries

daniel-goldstein:
The capacity on the cache is pretty arbitrary, but given that bunches are going to get churned through very quickly and then never used again, it seemed nice to have the assertion that every layer of the cache is always small and shouldn't be an issue to search through in a blocking manner. I tested this with a dev-deployed load-test and observed the number of `get_token_start_id` queries drop from O(jobs) to ~4 per second at max throughput. No difference in profiling, this is just an attempt to reduce the number of queries we're hitting the database with.

### DIFF
--- a/batch/batch/spec_writer.py
+++ b/batch/batch/spec_writer.py
@@ -43,7 +43,7 @@ class SpecWriter:
         assert index < len(in_batch_cache)
         if index >= 0:
             token, start, end = in_batch_cache[index]
-            if (job_id > start and end is None) or job_id in range(start, end):
+            if (job_id >= start and end is None) or job_id in range(start, end):
                 return (token, start)
 
         token, start_job_id, next_start_job_id = await SpecWriter._get_token_start_id_and_next_start_id(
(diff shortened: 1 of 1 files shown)


batch/batch/spec_writer.py
```diff
@@ -27,20 +35,50 @@ def get_spec_file_offsets(offsets):
...
+
+    @staticmethod
+    async def _get_token_start_id_and_next_start_id(db, batch_id, job_id) -> Tuple[str, int, int]:
         bunch_record = await db.select_and_fetchone(
             '''
-SELECT start_job_id, token FROM batch_bunches
+SELECT
+start_job_id,
+token,
+(SELECT start_job_id FROM batch_bunches WHERE batch_id = %s AND start_job_id > %s LIMIT 1) AS next_start_job_id
```
jigold, line 69: You need an ORDER BY here.

```
SELECT start_job_id
FROM batch_bunches
WHERE batch_id = %s AND start_job_id > %s
ORDER BY start_job_id
LIMIT 1
```

jigold (changes requested):

PR #11985, opened 2022-06-29
[gke] optimize utilization autoscaling and shrink nodes

danking:
I need to get GKE costs down further. The driver is now 3 CPU anyway.

@daniel-goldstein, I am not sure how to modify the Azure terraform. It appears this
is controlled by a variable which is already set to a 4 core machine?

PR #12025, opened 2022-07-13
[batch] Fix autoscaling policy in GCP Terraform config

lgruen:
Related to https://github.com/hail-is/hail/pull/11985.

Also update the MySQL version while we're at it.

#assign services


daniel-goldstein (approved):
Thanks @lgruen!

PR #12026, opened 2022-07-13
[batch] Define "scope" in auth template

lgruen:
Seems to be required since https://github.com/hail-is/hail/pull/11900.

#assign services


daniel-goldstein:
@lgruen I believe #11993 fixed this by removing the need for the scope.

lgruen:
> @lgruen I believe #11993 fixed this by removing the need for the scope.

Great, thanks! I'll close this PR.

PR #12027, opened 2022-07-13
[batch] Add resource_limits to enable new autoscaling_policy

lgruen:
Follow-up to https://github.com/hail-is/hail/pull/12025. Without this setting, the Terraform cluster upgrade will fail eventually.

These are cluster-wide limits. I picked rather generous limits, but I'm not sure how many dev deployments you have or whether you generally need a lot of nodes (the current node pool max is 200 nodes).

#assign services

### DIFF
--- a/infra/gcp/main.tf
+++ b/infra/gcp/main.tf
@@ -122,6 +122,16 @@ resource "google_container_cluster" "vdc" {
   cluster_autoscaling {
     enabled = true
     autoscaling_profile = "OPTIMIZE_UTILIZATION"
+    resource_limits {
+      resource_type = "cpu"
+      minimum       = 4
+      maximum       = 100
+    }
+    resource_limits {
+      resource_type = "memory"
+      minimum       = 8
+      maximum       = 500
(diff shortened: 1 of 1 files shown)


jigold:
@danking What's the minimum number of cores we want in the cluster? Is 4 okay?

danking:
And as for lower bound, 4 seems fine, we'll always be above that with 1 non-preemptible + 3 preemptible

danking (approved):

lgruen:
https://ci.hail.is/batches/3823676/jobs/139 looks like an unrelated timeout? Can I trigger a rerun of the `ci-test` somehow?

danking:
Indeed, I'm not sure what happened there. If you navigate to the relevant PR page, https://ci.hail.is/watched_branches/0/pr/12027, you should see a "retry" button. I've clicked it for this PR.

lgruen:
Hmm, I retried this again, but still no luck...

danking:
I'll look

PR #12029, opened 2022-07-13
[batch] Modify attempt resources trigger to prepare for future migrations

jigold:
Stacked on #12028

This PR just does some reconfiguring to enable us to remove the `resource` column in the `attempt_resources` table in the next PR with an online migration.

### DIFF
--- a/batch/batch/driver/job.py
+++ b/batch/batch/driver/job.py
@@ -83,14 +83,12 @@ async def add_attempt_resources(db, batch_id, job_id, attempt_id, resources):
             # This must be sorted in order to match the order of values in the actual SQL table!
             _resources = dict(sorted(_resources.items()))
 
-            resource_args = [
-                (batch_id, job_id, attempt_id, name, quantity, name) for name, quantity in _resources.items()
-            ]
+            resource_args = [(batch_id, job_id, attempt_id, quantity, name) for name, quantity in _resources.items()]
 
             await db.execute_many(
                 '''
-INSERT INTO `attempt_resources` (batch_id, job_id, attempt_id, resource, resource_id, quantity)
-SELECT %s, %s, %s, %s, resource_id, %s
+INSERT INTO `attempt_resources` (batch_id, job_id, attempt_id, resource_id, quantity)
+SELECT %s, %s, %s, resource_id, %s
 FROM resources
 WHERE resource = %s
 ON DUPLICATE KEY UPDATE quantity = quantity;
--- a/batch/sql/estimated-current.sql
+++ b/batch/sql/estimated-current.sql
@@ -584,9 +584,9 @@ DROP TRIGGER IF EXISTS attempt_resources_before_insert $$
 CREATE TRIGGER attempt_resources_before_insert BEFORE INSERT ON attempt_resources
 FOR EACH ROW
 BEGIN
-  DECLARE cur_resource_id INT;
-  SELECT resource_id INTO cur_resource_id FROM resources WHERE resource = NEW.resource;
-  SET NEW.resource_id = cur_resource_id;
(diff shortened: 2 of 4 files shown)


batch/sql/estimated-current.sql
```diff
@@ -296,8 +296,9 @@ CREATE INDEX batch_attributes_key_value ON `batch_attributes` (`key`, `value`(25
 CREATE TABLE IF NOT EXISTS `aggregated_billing_project_resources` (
   `billing_project` VARCHAR(100) NOT NULL,
   `resource` VARCHAR(100) NOT NULL,
+  `token` INT NOT NULL,
   `usage` BIGINT NOT NULL DEFAULT 0,
-  PRIMARY KEY (`billing_project`, `resource`),
+  PRIMARY KEY (`billing_project`, `resource`, `token`),
```
daniel-goldstein, line 301: Is this supposed to be here? I don't see it used elsewhere.

daniel-goldstein (changes requested):
I think this looks good, and fairly straightforward now that I'm used to the burn-in / chunks / audit approach you're applying to these PRs. Really just one thing stuck out to me.

batch/sql/estimated-current.sql
daniel-goldstein: Ah, sweet.

batch/batch/driver/job.py
```diff
@@ -83,12 +83,14 @@ async def add_attempt_resources(db, batch_id, job_id, attempt_id, resources):
...
             _resources = dict(sorted(_resources.items()))
 
-            resource_args = [(batch_id, job_id, attempt_id, name, quantity) for name, quantity in _resources.items()]
+            resource_args = [(batch_id, job_id, attempt_id, quantity, name) for name, quantity in _resources.items()]
 
             await db.execute_many(
                 '''
-INSERT INTO `attempt_resources` (batch_id, job_id, attempt_id, resource, quantity)
-VALUES (%s, %s, %s, %s, %s)
+INSERT INTO `attempt_resources` (batch_id, job_id, attempt_id, resource_id, quantity)
```
jigold, line 90: @danking Can you double check this query for me? I looked in the database and everything seems to be inserted properly (each resource has the right quantity), but I couldn't find documentation to make sure the execute many is handled correctly. I think it must be batching 5 individual insert statements?

jigold:
All set. This should be quick.

batch/batch/driver/job.py
danking: The join-insert defeats the bulk insert optimization, so this is just a loop over `execute`. That might slow down job insertion, but I kind of doubt this is the bottleneck of insertion. It can also be parallelized if necessary. Anyway, we should measure things like job insertion time!

https://github.com/PyMySQL/PyMySQL/blob/46d17402afaa07369b954eee026f68c5b96207ba/pymysql/cursors.py#L162-L197

jigold:
Ready for another look now.

batch/batch/driver/job.py
daniel-goldstein: I do think that this non-trivially increases load on the driver as this is on the critical path for MJS/MJC. Can we simply load the resource -> resource id mapping into memory on the driver and remove the join? Since I don't believe we would ever alter a row in the resources table, that mapping shouldn't ever be stale, right?

jigold: There's infrastructure to be able to do this in the BillingManagers where they have the resources in memory and update it from the database. Let's get #11840 merged in first before we add this though unless it's on the critical path.

daniel-goldstein: It's not urgent, I can hack around it for now

PR #12030, opened 2022-07-13
[batch] Remove resource names from the attempt resources table

jigold:
Stacked on #12029 

This PR rips out the `resource` column from the `attempt_resources` table and used the `resource_id` instead. Notice that the triggers need to use the `resource` and not the `resource_id` because we didn't convert the column in the `aggregated_{batch, jobs,billing_project}_resources` tables.


jigold:
This PR is all set. It will probably take around 1.5-2 hours. I verified that the attempt resources table should not be blocked while the primary key is being changed, but I didn't test it while in the process of dropping the column. At peak, the migration added an additional 20 GB or so of storage and then ended up with a net increase of 5 GB from the very beginning of the migration series. I'm not worried about increasing our storage space temporarily because the future migrations will use that extra space.

daniel-goldstein (approved):

PR #12036, opened 2022-07-18
[ci] Add frozen mode and only merge if no active deploys

jigold:
Adding a frozen mode in case the database migrations go off the rails and also to ensure we don't merge unless the previous deploy has finished.


ci/ci/github.py
```diff
@@ -707,8 +707,12 @@ async def _update(self, app):
...
                 if self.state_changed:
                     self.state_changed = False
-                    await self._heal(batch_client, db, gh)
-                    if self.mergeable:
+                    await self._heal(app, batch_client, gh)
+                    if (
+                        (self.deploy_batch is None or self.deploy_state is not None)
+                        and not app['frozen']
+                        and self.mergeable
+                    ):
```
danking, line 715: It seems to me that the goal of this change is to force CI to *attempt* to deploy every commit of main. Is that correct? In particular, a failure or checkout_failure is sufficient to allow CI to try to merge the next commit. These failure modes can happen well before the database is deployed.

I think this change successfully makes it very rare that an instance of Batch is running that is incompatible with the database, but it doesn't eliminate that event. Really eliminating that event needs a more sophisticated deployment process and some thought.

Using a WIP tag can eliminate that event with added human involvement. I'm curious what you think about that alternative. Regardless, I think a "stop deploying mode" for CI is worthwhile.

danking (changes requested):

jigold:
bump

danking (approved):

PR #12037, opened 2022-07-18
[query] Interval shuffle memory management

tpoterba:
This seems unlikely to be causing Lindo's issue, but does seem like a good change?


patrick-schultz:
It doesn't seem like a bad change. I suspect it's rare for this range to be more than 2 long, and I'd hope that the InsertFields avoids copying the entire row, so I'd be surprised if this ever made a significant difference. But I also doubt adding memory management would slow it down much, so better to be safe I guess.

PR #12040, opened 2022-07-21
[query] Fix leaked Array[Byte] for serialized aggregator states

tpoterba:
I suspect this is why the `sum_table_of_ndarrays` OOMs. Will retest.


tpoterba:
aha, you're totally right.

tpoterba:
This is a leak of heap memory, that's right. We leave serialized agg state byte arrays around.

tpoterba:
(or at least 1, but they're big)

patrick-schultz (approved):

PR #12043, opened 2022-07-22
[ci] Use 4 cores for db migration jobs in deploys

jigold:
This is to ensure when we're getting the offsets of the chunks that we don't run out of memory on the client.

### DIFF
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -1023,6 +1023,8 @@ EOF
                 parents=self.deps_parents(),
             )
 
+        n_cores = 1 if is_test_deployment else 4
+
         self.create_database_job = batch.create_job(
             CI_UTILS_IMAGE,
             command=['bash', '-c', create_database_script],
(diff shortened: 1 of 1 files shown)


ci/ci/build.py
```diff
@@ -1023,6 +1023,8 @@ def build(self, batch, code, scope):  # pylint: disable=unused-argument
                 parents=self.deps_parents(),
             )
 
+        n_cores = 1 if is_test_deployment else 4
```
daniel-goldstein, line 1026: Do you mean to use 4 also for PRs in default CI? This will only use 1 if it is a deploy in a Dev CI

daniel-goldstein (changes requested):

ci/ci/build.py
jigold: I want 4 only for deploys in default. Is that what I am doing now?

daniel-goldstein: No, this is using 4 for deploys in default and PRs in default. I think you want

```
4 if params.scope == 'deploy' and not is_test_deployment else 1
```

daniel-goldstein (approved):

PR #12053, opened 2022-07-26
[batch] Switch attempt resources primary keys and triggers to use resource ids

jigold:
Stacked on #12029. This is basically the same as #12030, but I separated dropping the final trigger and the resource column of the table from this PR so we can double check everything looks good and do that last dangerous step separately.

I think this will take about an hour, but I haven't looked timed it recently.

@danking Can you double check none of the SQL operations are blocking. If they are, I think we just have to live with it.

### DIFF
--- a/batch/sql/rm-resource-names-agg-resources.sql
+++ b/batch/sql/rm-resource-names-agg-resources.sql
@@ -1,0 +1,92 @@
+ALTER TABLE attempt_resources DROP PRIMARY KEY, ADD PRIMARY KEY (batch_id, job_id, attempt_id, resource_id), ALGORITHM=INPLACE, LOCK=NONE;
+ALTER TABLE attempt_resources MODIFY resource VARCHAR(100), ALGORITHM=INPLACE, LOCK=NONE;
+
+DELIMITER $$
+
+DROP TRIGGER IF EXISTS attempts_after_update $$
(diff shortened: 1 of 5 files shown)


batch/sql/rm-resource-names-agg-resources.sql
```diff
@@ -0,0 +1,92 @@
+ALTER TABLE attempt_resources DROP PRIMARY KEY, ADD PRIMARY KEY (batch_id, job_id, attempt_id, resource_id), ALGORITHM=INPLACE, LOCK=NONE;
+ALTER TABLE attempt_resources MODIFY resource VARCHAR(100), ALGORITHM=INPLACE, LOCK=NONE;
```
daniel-goldstein, line 2: What does this `MODIFY` here do?

jigold: Changes the column type to be allowed to be NULL. Right now it can't be null.

jigold:
This should be ready to go. I'm estimating it will probably take around an hour or two. I'd like to get it in later tonight or tomorrow morning.

jigold:
cc: @danking Goal is to merge this one tomorrow morning.

daniel-goldstein (dismissed):
Looks like there's an issue here with the batch invariants?

jigold:
Should be fixed now.

daniel-goldstein (approved):

PR #12055, opened 2022-07-26
[batch] Fix bad /batches query again

jigold:


danking (changes requested):
I think you need a left join on billing projects users

jigold:
fixed

danking (approved):

PR #12059, opened 2022-07-27
[ci] Fix missing api groups in pr namespace admin role

jigold:
Error message was this:

```
+ kubectl -n pr-12053-default-chqnmsebv8fq scale deployment batch-driver --replicas=0
Error from server (Forbidden): deployments.apps "batch-driver" is forbidden: User "system:serviceaccount:pr-12053-default-chqnmsebv8fq:admin" cannot get resource "deployments" in API group "apps" in the namespace "pr-12053-default-chqnmsebv8fq"
```

### DIFF
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -647,6 +647,9 @@ rules:
 - apiGroups: [""]
   resources: ["*"]
   verbs: ["*"]
+- apiGroups: ["apps"]
+  resources: ["deployments"]
+  verbs: ["*"]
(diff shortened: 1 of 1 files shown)


jigold:
unfortunately, we won't know if this works until it merges because the CI adding the new permissions hasn't been deployed yet. I don't think this change will do any harm though if it's not right.

jigold:
bump

daniel-goldstein (approved):

PR #12060, opened 2022-07-28
[compiler] Fix RNGSplit tuple requiredness

patrick-schultz:
The RNGSplit emitter assumed the components of `dynBitstring` were always inferred to be required, but it seems to be impossible to construct the bitstrings in a way that the requiredness inference can always see that.

This first makes the type of `dynBitstring` more constrained to simplify the emitter. Before it was allowed to be a single long, or an arbitrarily nested tuple all of whose leaves are longs. But I believe only flat tuples are actually being generated, so this makes that the enforced type.

Then the emitter asserts all fields are present at runtime.


tpoterba (approved):
good change.

PR #12061, opened 2022-07-28
[fs] Dont clear the buffer when seeking within bounds

daniel-goldstein:


hail/src/main/scala/is/hail/io/fs/FS.scala
```diff
@@ -152,8 +152,14 @@ abstract class FSSeekableInputStream extends InputStream with Seekable {
   }
 
   def seek(newPos: Long): Unit = {
-    bb.clear()
-    bb.limit(0)
+    val distance = (newPos - pos).toInt
```
danking, line 155: assert fits in int

danking (dismissed):

danking (approved):

PR #12062, opened 2022-07-28
[query] add `phased` parameter to `hl.balding_nichols_model`

danking:
CHANGELOG: `hl.balding_nichols_model` now supports an optional boolean parameter, `phased`, to control the phasedness of the generated genotypes.

This will fail until https://github.com/hail-is/hail/pull/12060 merges.


danking:
Daniel, I figure we should start pulling you more into compilers team orbit if you might dip your toes into all that.

daniel-goldstein (approved):

PR #12063, opened 2022-07-29
Revert "[QoB] disable import_plink using tests"

daniel-goldstein:
Reverts hail-is/hail#11987

Now that the FS seek bug is fixed these shouldn’t run up huge GCS charges


danking:
Can you run these and then check the BigQuery export to verify that we don’t have an explosion of class A operations? We can jump on zoom too id you wanna go over it together

danking (approved):

PR #12065, opened 2022-08-01
Azure upstream

violetbrina:
## Bug fixes to enable Azure deployment

Most of these bugs were discovered in deploying the MySQL server from scratch, specifically deploying version 8.0. 
Some were encountered when we hit certificate issues in trying to run the `./bootstrap.sh deploy_unmanaged` step multiple times within 24hrs. Documentation was clarified in order to resolve this issue.

- build.yaml
  - Step one fails on rerun since the /repo directory exists, -p to fix
- ci/create_database.py
  - In MySQL 8 a new error was introduced [4006](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_cannot_user_referenced_as_definer)
  - This error gets triggered on the CREATE USER IF NOT EXISTS commands for both user and admin if the user was previously created and set a a definer on any events/triggers.
  - Really this statement should be a no-op given that the user exists, but for some reason the error triggers anyway.
  - To get around this I added a manual check if the user/admin exists and if they do simply skip the create user command. This fixes the bug and allows the MySQL db deploy to finish properly

- dev-docs/letsencrypt.md
  - Debugging was confusing since the revoke command addressed ids we were unable to find.
  - After extensive searching I added to the documentation how to find your existing cert IDs if you need to revoke them.

- infra/azure/README.md
  - Added clarity to the Azure deployment documentation

- infra/azure/bootstrap.sh
  - Added the passing of additional flag arguments to terraform
  - In our case the passing of the `-upgrade` flag to the terraform init step was required in order to continue

- infra/azure/main.tf, infra/azure/modules/batch/main.tf, infra/azure/modules/batch/variables.tf infra/azure/variables.tf
  - Add additional argument for the az_storage_account.
  - The name must be globally unique in Azure, so the original argument failed on our deployment since it shared the name with the Hail team's Azure deployment


infra/azure/modules/batch/main.tf
```diff
@@ -74,7 +74,7 @@ resource "azurerm_storage_container" "query" {
 }
 
 resource "azurerm_storage_account" "test" {
-  name                     = "${var.resource_group.name}test"
+  name                     = "${var.az_storage_account_name}test"
```
daniel-goldstein, line 77: Good find. We definitely shouldn't dictate the name of your resource group just because we snagged the `hailtest` storage account. Can we make this name a bit more specific, more along the lines of `batch_test_user_storage_account_name`?

daniel-goldstein: This is likely also a problem for the other storage accounts (see the `batch` storage account above, we just haven't clashed on those).

daniel-goldstein (changes requested):
Thanks for the PR! I've long been frustrated with `create_database.py` and this is a good step forward. I have a few questions regarding the changes but overall this looks great. Sorry you ran into trouble but we appreciate the fixes!

build.yaml
```diff
@@ -12,7 +12,7 @@ steps:
     script: |
       set -ex
       cd /io
-      mkdir repo
+      mkdir -p repo
```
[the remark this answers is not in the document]
violetbrina: I can confirm that it still does happen. I suspect it's because the new container is mounting to the same location on the vm where the repo folder resides. So this issue only occurs if you re-run. Which we've had to do often when it fails part way through.

ci/create_database.py
```diff
@@ -95,16 +95,42 @@ async def create_database():
...
+
+    user_exists = await db.execute_and_fetchone(
+        f"SELECT user FROM mysql.user WHERE user='{user_username}'"
+    )
+    user_exists = user_exists.get('user') == user_username
+
+    create_admin = (
+        f"CREATE USER IF NOT EXISTS '{admin_username}'@'%' IDENTIFIED BY '{admin_password}';"
+        if not admin_exists
+        else ""
```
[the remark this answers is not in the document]
violetbrina: Sure thing, I'll update the PR to reflect this suggestion

infra/azure/README.md
```diff
@@ -30,6 +37,7 @@ terraform state. The following command creates these and stores their names in
 Initialize terraform:
 
 ```
+export STORAGE_ACCOUNT_NAME="<insert from remote_storage.tfvars>"
```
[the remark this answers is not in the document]
violetbrina: Ah good point. I'll remove that comment then.

infra/azure/modules/batch/main.tf
violetbrina: Well I'll leave the other potential clashes for now, but I'll take the name change suggestion.

violetbrina:
Apologies for the delay in addressing the feedback. Had some urgent projects to attend to. Here are the changes you've suggested. Let me know if there's anything else I can do to help with this PR.

danking:
@daniel-goldstein back to you

infra/azure/README.md
```diff
@@ -106,6 +122,9 @@ Deploy unmanaged resources by running
 ./bootstrap.sh deploy_unmanaged
 ```
 
+If the final letsencrypt step in deploy_unmangaed fails, copy the kubectl secret from
+stdout and apply it manually.
```
daniel-goldstein, line 126: Could you elaborate on what error you saw here and what secret needs to be copied?

ci/create_database.py
```diff
@@ -95,16 +95,39 @@ async def create_database():
     with open(create_database_config['user_password_file']) as f:
         user_password = f.read()
 
+
+    admin_exists = await db.execute_and_fetchone(
+        f"SELECT user FROM mysql.user WHERE user='{admin_username}'"
```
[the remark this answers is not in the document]
violetbrina: Done!

infra/azure/README.md
violetbrina: The elaboration is quite verbose so if I need to simplify again let me know. Basically the letsencrypt.sh only can be run a certain number of times before it permanently fails within a certain window. Without the intervention specified above we were unable to retrieve the secret and have it applied properly to the kubernetes cluster.

So we manually retrieved it through stdout and applied it reverting the changes. If the deploy_unmanaged step runs successfully first time then this is not required. However, we had to do this multiple times while debugging the deploy. Whether you want this in the README or not I'm not sure. 

Let me know if further clarification is needed.

violetbrina: Oh and it's this secret in the letsencrypt.sh script:
```
apiVersion: v1
kind: Secret
metadata:
  name: letsencrypt-config
  namespace: default
type: Opaque
data: ...
```

daniel-goldstein:
@violetbrina there are just a couple of lint errors left on this branch. If you locally run `make check-ci` you should see the couple of suggestions to fix. I'm also happy to push the fixes to `populationgenomics:azure-upstream` if that's fine.

danking:
Closing in favor of https://github.com/hail-is/hail/pull/12521 which has the lints fixed.

PR #12067, opened 2022-08-01
[query] add rand_int32 and rand_int64

danking:
CHANGELOG: Add `rand_int32` and `rand_int64` for generating random 32-bit and 64-bit integers, respectively.


danking:
I'm pretty sure this will fail the doctests due to the issue I noted [in Zulip](https://hail.zulipchat.com/#narrow/stream/123011-Hail-Query-Dev/topic/seeded.20randomness.20in.20Spark). Once that's resolved, I think it's good for review.

patrick-schultz (approved):

PR #12073, opened 2022-08-03
[datasets] update pan-UKB datasets and make available via GCS

pwc2:
Currently, the Hail datasets API only has the pan-UKB datasets available via the Amazon S3 bucket. This PR makes the pan-UKB datasets available via GCS as well (from the `gs://ukb-diverse-pops-public` bucket), and updates the summary statistics and meta-analysis MatrixTables to reflect the most up-to-date release.


danking (approved):
Oops sorry for the delay in review!

PR #12076, opened 2022-08-04
[query] Rewrite ld prune to use java objects for filtering

chrisvittal:
This simplifies the code and means that region bookeeping for the prune
queue is unnecessary.


tpoterba (approved):
Looks good.

PR #12078, opened 2022-08-05
[batch] Use region requirements when choosing a location

jigold:
I added the concept of a "region". A "location" is a zone in GCP and the region in Azure. In Azure, a region and location are identical. For the first 1k cores, we create VMs within the default region rather than the default zone as before.

### DIFF
--- a/batch/batch/cloud/azure/driver/driver.py
+++ b/batch/batch/cloud/azure/driver/driver.py
@@ -48,7 +48,7 @@ class AzureDriver(CloudDriver):
 
         region_monitor = await RegionMonitor.create(region)
         billing_manager = await AzureBillingManager.create(db, pricing_client, regions)
-        inst_coll_manager = InstanceCollectionManager(db, machine_name_prefix, region_monitor)
+        inst_coll_manager = InstanceCollectionManager(db, machine_name_prefix, region_monitor, region)
         resource_manager = AzureResourceManager(
             subscription_id, resource_group, ssh_public_key, arm_client, compute_client, billing_manager
         )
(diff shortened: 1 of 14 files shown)


daniel-goldstein:
I think I have some quibbles on dropping the granularity of tracking cpus by zone and the terminology but I'm going to try deploying this in my project and see that we get the behavior we want.

jigold:
Feel free to modify the terminology to something that makes sense to you. I agree I had a hard time figuring out what the correct interface should be.

daniel-goldstein:
I made some edits that I think helped to reduce the number of places that are aware of the notion of a default_region. It's really now just isolated to the `InstanceCollectionManager`, since that's the piece of code that's making the decision "use the default region when the cluster is small". I didn't quite like the pattern of retrieving a default from a `LocationMonitor` just to give it right back to the location monitor in the next line. I think this way the `LocationMonitor` API is much simpler, and we can actually remove its `default_location` method entirely as I believe it is no longer used. I can do that in a follow-up PR if you like this approach.

One other thing is I wanted to articulate the distinction between the "region CI needs its jobs in" and the "default region that batch will spin up workers in for small clusters". While they are in practice the same, I found that treating them both as the "default_region" tied the logic around jobs for CI closely with internal Batch decisions. I tried to separate out these two concepts so that in the future when jobs support locations it will be easier to excise the CI-specific code from the Batch scheduler.

Another thing that I realized during this process is that Azure has regions and zones just like GCP, we just have this "Location" terminology since that's what azure Resource Groups call the region that the resource group stores its metadata in (nice job with the consistency there Azure). I am fine with using the term "location" to mean either a region or zone since we only talk about regions in Azure, but I would also like to follow up with a sweep that makes this language more precise where possible. For example, the `possible_cloud_locations` function is really just `possible_cloud_regions`, and we could even go so far as mandating a `region` field in the global config instead of having `azure_location` and `gcp_region`, which are synonymous even though named differently. It also leads me to wonder, why are we only single-region in Azure?

jigold:
Thank you for cleaning this up!

jigold:
I can't approve it, but I think it's good to go.

daniel-goldstein (approved):

PR #12086, opened 2022-08-15
[query] cache RVD specs to avoid lots of file reads

tpoterba:
cc @danking @daniel-goldstein 

I bet this is a QoB hotspot for small tables.


tpoterba:
(for context, I found my way here from `AbstractTableSpec.rowAndGlobalPTypes`, which is called once per *optimization pass*.

patrick-schultz (approved):

PR #12093, opened 2022-08-17
[ci] Track active namespaces in CI database

daniel-goldstein:
Adds functionality for CI to track which namespaces in the cluster are currently active and which services they are running. See #12095 for more context on the motivation for this change.

As an added bonus, this is one step toward more flexible internal namespace management, i.e. automatic cleanup of dev namespaces or keeping PR namespaces open for X days.

### DIFF
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -3,7 +3,7 @@ import json
 import logging
 from collections import Counter, defaultdict
 from shlex import quote as shq
-from typing import Dict, List
+from typing import Dict, List, Tuple
 
 import jinja2
 import yaml
@@ -143,12 +143,18 @@ class BuildConfiguration:
             if step.can_run_in_scope(scope):
                 step.cleanup(batch, scope, parent_jobs)
 
-    def deployed_services(self) -> Dict[str, List[str]]:
+    def deployed_services(self) -> Tuple[str, List[str]]:
         services = defaultdict(list)
         for s in self.steps:
             if isinstance(s, DeployStep):
                 services[s.namespace].extend(s.services())
-        return services
+        # build.yaml allows for multiple namespaces, but
+        # in actuality we only ever use 1 and make many assumptions
+        # around there being a 1:1 correspondence between builds and namespaces
+        namespaces = list(services.keys())
+        assert len(namespaces) == 1
+        ns = namespaces[0]
+        return ns, services[ns]
 
 
 class Step(abc.ABC):
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -143,6 +143,13 @@ class BuildConfiguration:
             if step.can_run_in_scope(scope):
                 step.cleanup(batch, scope, parent_jobs)
 
+    def deployed_services(self) -> Dict[str, List[str]]:
+        services = defaultdict(list)
+        for s in self.steps:
+            if isinstance(s, DeployStep):
+                services[s.namespace].extend(s.services())
+        return services
+
 
 class Step(abc.ABC):
     def __init__(self, params):
@@ -786,6 +793,11 @@ class DeployStep(Step):
             json.get('link'),
             json.get('wait'),
         )
+
+    def services(self):
+        if self.wait:
+            return [w['name'] for w in self.wait]
+        return []
 
     def config(self, scope):  # pylint: disable=unused-argument
         return {'token': self.token}
--- a/ci/ci/ci.py
+++ b/ci/ci/ci.py
@@ -402,10 +402,11 @@ async def batch_callback_handler(request):
                         assert 'dev' not in attrs
                         namespace = json.loads(attrs['namespace'])
                         assert namespace != 'default'
-                        await db.execute_update(
-                            'DELETE FROM active_namespaces WHERE namespace = %s',
-                            (namespace,),
-                        )
+                        if DEFAULT_NAMESPACE == 'default':
+                            await db.execute_update(
+                                'DELETE FROM active_namespaces WHERE namespace = %s',
+                                (namespace,),
+                            )
 
                     await wb.notify_batch_changed(app)
 
--- a/ci/ci/ci.py
+++ b/ci/ci/ci.py
@@ -21,11 +21,11 @@ from hailtop.batch_client.aioclient import Batch, BatchClient
 from hailtop.config import get_deploy_config
 from hailtop.hail_logging import AccessLogger
 from hailtop.tls import internal_server_ssl_context
-from hailtop.utils import collect_agen, humanize_timedelta_msecs
+from hailtop.utils import collect_agen, humanize_timedelta_msecs, periodically_call
 from web_common import render_template, set_message, setup_aiohttp_jinja2, setup_common_static_routes
 
 from .constants import AUTHORIZED_USERS, TEAMS
-from .environment import STORAGE_URI
+from .environment import DEFAULT_NAMESPACE, STORAGE_URI
 from .github import PR, WIP, FQBranch, MergeFailureBatch, Repo, UnwatchedBranch, WatchedBranch, select_random_teammate
 
 with open(os.environ.get('HAIL_CI_OAUTH_TOKEN', 'oauth-token/oauth-token'), 'r', encoding='utf-8') as f:
@@ -395,6 +395,17 @@ async def batch_callback_handler(request):
             for wb in watched_branches:
                 if wb.branch.short_str() == target_branch:
                     log.info(f'watched_branch {wb.branch.short_str()} notify batch changed')
+
+                    if 'test' in params and params['complete']:
+                        assert 'deploy' not in params
+                        db: Database = app['db']
+                        namespaces = json.loads(params['namespaces'])
+                        await db.execute_update(
+                            f'''
+DELETE FROM active_namespaces
+WHERE {" OR ".join("namespace=" + ns for ns in namespaces)}'''
+                        )
+
                     await wb.notify_batch_changed(app)
 
 
@@ -488,7 +499,7 @@ async def dev_deploy_branch(request, userdata):
     batch_client = app['batch_client']
 
     try:
-        batch_id = await unwatched_branch.deploy(batch_client, steps, excluded_steps=excluded_steps)
+        batch_id = await unwatched_branch.deploy(app['db'], batch_client, steps, excluded_steps=excluded_steps)
     except asyncio.CancelledError:
         raise
     except Exception as e:  # pylint: disable=broad-except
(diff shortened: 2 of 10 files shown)


ci/ci/ci.py
```diff
@@ -400,6 +400,17 @@ async def batch_callback_handler(request):
             for wb in watched_branches:
                 if wb.branch.short_str() == target_branch:
                     log.info(f'watched_branch {wb.branch.short_str()} notify batch changed')
+
+                    if 'test' in params and params['complete']:
+                        assert 'deploy' not in params
```
jigold, line 405: Can we assert not dev as well?

ci/ci/ci.py
```diff
@@ -558,6 +569,75 @@ async def unfreeze_deploys(request, userdata):  # pylint: disable=unused-argumen
...
+@check_csrf_token
+@web_authenticated_developers_only()
+async def add_namespaced_service(request, userdata):  # pylint: disable=unused-argument
+    db: Database = request.app['db']
+    post = await request.post()
+    service = post['service']
+    namespace = request.match_info['namespace']
+
+    await db.execute_insertone(
+        'INSERT INTO deployed_services VALUES (%s, %s)',
```
jigold, line 605: I think you need some sort of check here or nice error if the namespace/service already exists.

ci/ci/utils.py
```diff
@@ -1,9 +1,34 @@
...
+    db: Database,
+    services_per_namespace: Dict[str, List[str]],
+    expiration_time: Optional[datetime.datetime],
+):
+    expiration = expiration_time.strftime('%Y-%m-%d %H:%M:%S') if expiration_time else None
+    for namespace, services in services_per_namespace.items():
+        await db.execute_insertone(
+            '''
+INSERT INTO active_namespaces (`namespace`, `expiration_time`)
+VALUES (%s, %s)
```
jigold, line 26: Not sure, but you might get a warning with the current code you have.

This might also work if you do:

```
INSERT INTO active_namespaces (`namespace`, `expiration_time`)
VALUES (%s, %s) new
ON DUPLICATE KEY UPDATE expiration_time = new.expiration_time
```

ci/ci/utils.py
```diff
@@ -1,9 +1,34 @@
...
+ON DUPLICATE KEY UPDATE expiration_time = VALUES(expiration_time)
+            ''',
+            (namespace, expiration),
+        )
+        await db.execute_many(
```
jigold, line 31: INSERT IGNORE will add a warning message to the logs if the row exists. I think it's fine here, but we can't use it in Batch.

ci/sql/delete-ci-tables.sql
```diff
@@ -6,3 +6,4 @@ DROP TABLE IF EXISTS `ci_migrations`;
 DROP TABLE IF EXISTS `authorized_shas`;
 DROP TABLE IF EXISTS `invalidated_batches`;
 DROP TABLE IF EXISTS `globals`;
+DROP TABLE IF EXISTS `active_namespaces`;
```
jigold, line 9: You need `deployed_services` here.

jigold (dismissed):

ci/ci/build.py
```diff
@@ -143,6 +143,13 @@ def build(self, batch, code, scope):
             if step.can_run_in_scope(scope):
                 step.cleanup(batch, scope, parent_jobs)
 
+    def deployed_services(self) -> Dict[str, List[str]]:
```
jigold, line 146: When would you have multiple namespaces with deployed services in the build configuration?

ci/ci/ci.py
```diff
@@ -395,6 +396,18 @@ async def batch_callback_handler(request):
             for wb in watched_branches:
                 if wb.branch.short_str() == target_branch:
                     log.info(f'watched_branch {wb.branch.short_str()} notify batch changed')
+
+                    if 'test' in params and params['complete']:
```
jigold, line 400: I think you want to test for 'test' in `attrs`. Same for `namespaces` below.

ci/ci/ci.py
```diff
@@ -395,6 +396,18 @@ async def batch_callback_handler(request):
...
+                        assert 'deploy' not in params
+                        assert 'dev' not in params
+                        db: Database = app['db']
+                        namespaces = json.loads(params['namespaces'])
+                        await db.execute_update(
+                            f'''
+DELETE FROM active_namespaces
+WHERE {" OR ".join("namespace=" + ns for ns in namespaces)}'''
```
jigold, line 408: I'd feel more comfortable here if we used a query string with `%s` and gave the namespace names as query args.

ci/ci/ci.py
```diff
@@ -553,6 +566,85 @@ async def unfreeze_deploys(request, userdata):  # pylint: disable=unused-argumen
...
+@auth.web_authenticated_developers_only()
+async def add_namespaced_service(request, userdata):  # pylint: disable=unused-argument
+    db: Database = request.app['db']
+    post = await request.post()
+    service = post['service']
+    namespace = request.match_info['namespace']
+
+    try:
+        await db.execute_insertone(
+            'INSERT INTO deployed_services VALUES (%s, %s)',
```
jigold, line 603: Suggestion: Check for whether the record already exists and give a nice error message. I think for this page it's not essential, but might save someone some pain in the future.

ci/ci/ci.py
```diff
@@ -553,6 +566,85 @@ async def unfreeze_deploys(request, userdata):  # pylint: disable=unused-argumen
...
+
+@routes.post('/namespaces/add')
+@check_csrf_token
+@auth.web_authenticated_developers_only()
+async def add_namespace(request, userdata):  # pylint: disable=unused-argument
+    db: Database = request.app['db']
+    post = await request.post()
+    namespace = post['namespace']
+
+    try:
```
jigold, line 622: Same suggestion about nicer error message.

ci/ci/ci.py
```diff
@@ -553,6 +566,85 @@ async def unfreeze_deploys(request, userdata):  # pylint: disable=unused-argumen
...
+    except pymysql.err.IntegrityError as err:
+        # 1062 ER_DUP_ENTRY https://dev.mysql.com/doc/refman/5.7/en/server-error-reference.html#error_er_dup_entry
+        if err.args[0] != 1062:
+            raise
+
+    return web.HTTPFound(deploy_config.external_url('ci', '/namespaces'))
+
+
+async def cleanup_expired_namespaces(db: Database):
+    expired_namespaces = [
```
jigold, line 636: do we need an assert here that the DEFAULT NAMESPACE == "default"? I don't think this function is called in test deployments based on the if statement where we ensure the future for this function in the loop.

ci/sql/estimated-current.sql
```diff
@@ -15,3 +15,21 @@ CREATE TABLE IF NOT EXISTS `globals` (
...
+
+CREATE TABLE IF NOT EXISTS `deployed_services` (
+  `namespace` VARCHAR(100) NOT NULL,
+  `service` VARCHAR(100) NOT NULL,
+  PRIMARY KEY (`namespace`, `service`),
+  FOREIGN KEY (`namespace`) REFERENCES active_namespaces(namespace) ON DELETE CASCADE
+) ENGINE = InnoDB;
+
+INSERT INTO `active_namespaces` (`namespace`) VALUES (`default`);
+INSERT INTO `deployed_services` (`namespace`, `service`) VALUES
```
jigold, line 34: What about auth-driver? Should this be all possible services including benchmark?

ci/sql/active-namespaces.sql
```diff
@@ -0,0 +1,17 @@
...
+
+CREATE TABLE IF NOT EXISTS `deployed_services` (
+  `namespace` VARCHAR(100) NOT NULL,
+  `service` VARCHAR(100) NOT NULL,
+  PRIMARY KEY (`namespace`, `service`),
+  FOREIGN KEY (`namespace`) REFERENCES active_namespaces(namespace) ON DELETE CASCADE
+) ENGINE = InnoDB;
+
+INSERT INTO `active_namespaces` (`namespace`) VALUES ('default');
+INSERT INTO `deployed_services` (`namespace`, `service`) VALUES
```
jigold, line 16: See comment below about which services should be here (e.g auth-driver?)

ci/ci/utils.py
```diff
@@ -1,9 +1,38 @@
...
     alnum = string.ascii_lowercase + string.digits
     return secrets.choice(alpha) + ''.join([secrets.choice(alnum) for _ in range(size - 1)])
+
+
+async def add_deployed_services(
+    db: Database,
+    services_per_namespace: Dict[str, List[str]],
+    expiration_time: Optional[datetime.datetime],
+):
+    expiration = expiration_time.strftime('%Y-%m-%d %H:%M:%S') if expiration_time else None
```
jigold, line 21: Is there a reason why you decided to use timestamps rather than BIGINT? Can you double check that the time zones here are going to be consistent such as with daylight savings?

ci/ci/utils.py
```diff
@@ -1,9 +1,38 @@
...
+    for namespace, services in services_per_namespace.items():
+        await db.execute_insertone(
+            '''
+INSERT INTO active_namespaces (`namespace`, `expiration_time`)
+VALUES (%s, %s) new
+ON DUPLICATE KEY UPDATE expiration_time = NEW.expiration_time
```
jigold, line 27: The NEW being in caps works here I think, but it would be less confusing if it's lower case to match the declaration in the line above.

ci/ci/ci.py
```diff
@@ -395,6 +396,18 @@ async def batch_callback_handler(request):
             for wb in watched_branches:
                 if wb.branch.short_str() == target_branch:
                     log.info(f'watched_branch {wb.branch.short_str()} notify batch changed')
+
+                    if 'test' in params and params['complete']:
+                        assert 'deploy' not in params
+                        assert 'dev' not in params
+                        db: Database = app['db']
```
jigold, line 403: Is your thinking here that if the batch callback is missed, then the cleanup loop will eventually cleanup the database once the namespace expires in a day?

jigold (changes requested):

ci/ci/build.py
daniel-goldstein: build.yaml technically permits multiple `CreateNamespace` steps, but I can't imagine us using that anytime soon. I'll make this stricter and leave a comment explaining the assertion.

ci/ci/ci.py
daniel-goldstein: oops, totally

ci/ci/ci.py
daniel-goldstein: I'll add it for peace of mind

ci/sql/estimated-current.sql
daniel-goldstein: My goal here was to restrict to the set of required services in a Hail cluster (that's running CI), which seemed to me like `auth` and `batch`. I don't think the australians are running benchmark, and wouldn't want to muck up their config unnecessarily. Services like benchmark and grafana, etc. will be added in for our clusters once CI does its first deploy.

Regarding `auth-driver`, there's not actually a kubernetes `Service` for that deployment.

ci/ci/utils.py
daniel-goldstein: I think this was an oversight on my point in terms of consistency with the rest of our database infrastructure. I went with timestamp because it seemed easiest to "add a day" to. Happy to use BIGINT and `time_msecs` if that would be more consistent.

ci/ci/ci.py
daniel-goldstein: Yup!

ci/ci/ci.py
```diff
@@ -395,6 +396,17 @@ async def batch_callback_handler(request):
             for wb in watched_branches:
                 if wb.branch.short_str() == target_branch:
                     log.info(f'watched_branch {wb.branch.short_str()} notify batch changed')
+
+                    if 'test' in attrs and params['complete']:
+                        assert 'deploy' not in attrs
+                        assert 'dev' not in attrs
+                        namespace = json.loads(attrs['namespace'])
+                        assert namespace != 'default'
```
jigold, line 404: Can we assert when this callback gets executed? Mainly I want to make sure this doesn't run in the wrong scope (ie. dev or test).

ci/ci/utils.py
jigold: I think either is fine as long as we're okay if it's not exactly 24 hours when there's daylight savings time. We do use DATETIME with session ids I believe.

jigold (changes requested):

ci/ci/ci.py
daniel-goldstein: I think it's valid for this callback to run in dev (dev deploy against a dev CI), but I'll add a guard to only change the database in default

daniel-goldstein:
@jigold bump

jigold (approved):

PR #12094, opened 2022-08-17
[k8s] Move to headless services and make pods listen on 443

daniel-goldstein:
Moves multi-pod deployments over to using Headless Services, which enables client-side load-balancing to the underlying pods. See #12095 for more context.

The reason I put this in its own PR is that Kubernetes won't let me apply the `clusterIP: None` changes to existing `Services`, and I must delete the `Service` resources first. I can manually delete and apply new headless services in a way that is compatible with what is currently on main and with just a few seconds of downtime, but I should do this manually just before this PR merges.

### DIFF
--- a/auth/auth/auth.py
+++ b/auth/auth/auth.py
@@ -716,7 +716,7 @@ async def get_session_id(request):
     return session.get('session_id')
 
 
-@routes.get('/api/v1alpha/verify_dev_credentials')
+@routes.route('*', '/api/v1alpha/verify_dev_credentials')
 async def verify_dev_credentials(request):
     session_id = await get_session_id(request)
     if not session_id:
@@ -728,7 +728,7 @@ async def verify_dev_credentials(request):
     return web.Response(status=200)
 
 
-@routes.get('/api/v1alpha/verify_dev_or_sa_credentials')
+@routes.route('*', '/api/v1alpha/verify_dev_or_sa_credentials')
 async def verify_dev_or_sa_credentials(request):
     session_id = await get_session_id(request)
     if not session_id:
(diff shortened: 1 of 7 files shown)


daniel-goldstein:
Here's the link to the [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) documentation and here's an [example blog post](https://techdozo.dev/grpc-load-balancing-on-kubernetes-using-headless-service/) where someone encountered the same issues we were facing with normal services. I think the documentation is motivation enough though: Envoy's `STRICT_DNS` endpoint setting would be considered a form of service discovery done through DNS. In order for Envoy to correctly make load balancing decisions, that DNS request should return all the viable IPs for an upstream instead of a single IP that points to kube proxy. Headless services do just that.

jigold (dismissed):
This all seems fine, but just so it's here for reference, can you remind me why we need the `*` in the routes definition. I know you had this described somewhere, but I can't find it. I think port 443 is so we don't need root privileges in Envoy?

daniel-goldstein:
The `*` means that route will be triggered for any request matching the specified URL for any method, be it GET or POST, etc. The reason I needed to make that change is that when envoy makes an authentication request to that endpoint, it uses the HTTP method of the original request. E.g. If I make a POST to https://internal.hail.is/dgoldste/batch/batches/create envoy will authenticate me with a POST request to auth:443/api/v1alpha/verify_dev_credentials. So I can't set that endpoint to be any one method.

daniel-goldstein:
> I think port 443 is so we don't need root privileges in Envoy?

This is related to the way headless services expose the pod itself, but as I'm writing this I feel like I want more clarity on exactly why, so I will do a bit of digging and come back with a better response.

daniel-goldstein:
Dan mentioned it would be good to keep the pods on unprivileged ports so I'm going to think of a way to keep them there

PR #12095, opened 2022-08-17
[ci,gateway,internal-gateway] Move gateways from nginx to envoy

daniel-goldstein:
This is a multi-stage overhaul of our Kubernetes load balancers / service discovery. This involves moving off of NGINX onto Envoy, but more importantly involves better control of what namespaces and services are active in our cluster at a given point in time.

TL;DR Switching from NGINX to Envoy with CI acting as the "control plane" for our internal networking allows us to more easily dynamically configure our Kubernetes networking and achieve proper connection pooling/load-balancing over TLS, which translates to less resource consumption and lower request latencies.

## Motivation
This is primarily a performance-motivated change, and one largely based on our (ab)use of NGINX in order to work with our dynamically-generated Kubernetes test namespaces. Currently, we configure NGINX by creating server blocks that dynamically resolve and dispatch requests based on matching regular expressions on the host and path headers. This is in large part due that at gateway deploy time we do not statically know all of the namespaces and namespace-service combinations that will exist in the cluster in the future. This is true for `default`, but not test namespaces, and NGINX will refuse to start with statically-configured clusters that it cannot reach. Making the server blocks make the routing decisions dynamically circumvents this limitation.

However, this prevents usage of NGINX [upstream](http://nginx.org/en/docs/http/ngx_http_upstream_module.html) blocks that provide connection pooling, at least in the community edition, and as a result the gateways will create and terminate a TCP connection per http request. This likely causes minor delays on the front-end through gateway, but this hampers performance greatly in job scheduling. The batch driver is forced to establish a new TCP connection and do an SSL handshake with the internal-gateway multiple times per job, which is expensive and slow. We currently have to dedicate a 2-core NGINX sidecar for the batch-driver just to terminate TLS with internal-gateway and free up cycles in the batch-driver python process. By using proper persistent connections, we can reduce the TLS overhead to single-digit percents of a core.

This leads to the first goal of this transition: configure our load balancers to know the full cluster configuration at any point in time so they can properly maintain connection pools with upstream services.

However, this is not the only problem. Each "upstream" Service in Kubernetes may consist of multiple underlying pods but Kubernetes Services as we use them don't provide proper load-balancing when mixed with persistent connections. When we declare a Service for say, batch in default, Kubernetes adds a DNS record for `batch.default` that resolves to a single IP pointing at kube-proxy. When a new TCP connection is established with kube-proxy for that IP, it rolls the dice (using `iptables`) and assigns that connection to a particular pod to which it will forward all subsequent packets. From the load-balancer's perspective, there is only one IP address, and only one place to open connections. The load-balancer doesn't have the information to actually load-balance once we have a functioning connection pool. This can lead to really unbalanced scenarios when preemptible pods come and go.

This leads to our second goal: instead of routing all requests through kube-proxy, use Kubernetes Headless Services to expose all pod IPs underlying a Service so that our proxies can properly load-balance across persistent connections.

## Solution

This PR addresses the two goals outlined above and does so through using Envoy, a load-balancer/proxy that is well-suited to this sort of highly-dynamic cluster configuration. Envoy does not have the constraint that all upstream services must be available at start-time, and has a very convenient API for updating the cluster configuration without the need for restarting the process or dropping traffic. This makes regularly updating the cluster configuration whenever new test namespaces are created relatively straightforward and non-disruptive to traffic in other namespaces. The high-level approach is as follows:

1.  Envoy-based gateways and internal-gateways will load their routing configuration from a Kubernetes ConfigMap, which they watch for changes and reconcile their configuration when the ConfigMap changes. The ConfigMap can be populated with a manual deploy and is populated from the beginning with production routes (i.e. batch.hail.is gets routed to batch.default)
2. When running CI, CI will regularly update the ConfigMap with additional routes based on which internal namespaces (dev and PR) are currently active. This requires relatively small changes to CI to track active namespaces but overall is a pretty small change. Note that this does not introduce a dependency on CI to support production traffic, only development traffic.
3. Deployments that run more than 1 replica (but really can be all of them) are run behind Headless Services, which expose the underlying pod IPs so Envoy can handle load-balancing instead of kube-proxy. This allows Envoy to make smart load-balancing decisions and correctly enforce rate-limiting when using connection pools.

The namespace tracking in CI in Point 2 is possible before we make any changes to our networking, so that comes first in #12093. Point 3 is taken care of in #12094, and the rest of Point 2 and Point 1, everything to do with Envoy, is in this PR.

### Additional QoL improvements
- Envoy by default exposes Prometheus metrics that we can use to easily monitor things like rate-limiting, request failures and durations
- Since all Envoy configuration is in the configmap, we don't need to build any images. I suppose we could have done this with NGINX, so this isn't something to fault NGINX for. Just another small win buried in these changes.

cc @danking


auth/auth/auth.py
```diff
@@ -717,7 +717,7 @@ async def get_session_id(request):
     return session.get('session_id')
 
 
-@routes.get('/api/v1alpha/verify_dev_credentials')
+@routes.route('*', '/api/v1alpha/verify_dev_credentials')
```
daniel-goldstein, line 720: This change is because Envoy's built-in auth request mechanism sends the whole incoming request as is to the "auth cluster". So if I'm sending a POST to batch.hail.is, it will also send a POST to authenticate that request. The easiest way I could find to get around this was to register this authentication endpoint under all HTTP methods.

internal-gateway/deployment.yaml
```diff
@@ -36,25 +36,50 @@ spec:
...
+           - $(hostname)
          resources:
            requests:
              cpu: "20m"
              memory: "200M"
            limits:
              cpu: "1"
              memory: "1G"
+         env:
+         - name: ENVOY_UID
```
jigold, line 58: Why do we need to be the root user?

internal-gateway/deployment.yaml
```diff
@@ -36,25 +36,50 @@ spec:
...
-         image: "{{ internal_gateway_image.image }}"
+         image: "{{ global.docker_prefix }}/envoy:v1.22.3"
+         command:
+           - /usr/local/bin/envoy
+           - --config-path
+           - /etc/envoy/envoy.yaml
+           - --concurrency
+           - "2"
+           - --service-cluster
+           - gateway
```
jigold, line 47: I think this should be` internal-gateway`

internal-gateway/envoy.yaml
```diff
@@ -0,0 +1,60 @@
...
+                - name: envoy.filters.http.local_ratelimit
+                  typed_config:
+                    "@type": type.googleapis.com/envoy.extensions.filters.http.local_ratelimit.v3.LocalRateLimit
+                    stat_prefix: http_local_rate_limiter
+                - name: envoy.filters.http.router
+                  typed_config:
+                    "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
+    admin:
+      address:
+        socket_address:
```
jigold, line 42: Is there a simple example you can point me to so I can work my way up to understanding what's in the yaml?

internal-gateway/Makefile
```diff
@@ -1,21 +1,20 @@
...
 
-export NAMESPACE := default
-
-build:
-	$(MAKE) -C ../docker hail-ubuntu
-	python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"'$$(cat ../docker/hail-ubuntu-image-ref)'"}}' Dockerfile Dockerfile.out
-	../docker-build.sh . Dockerfile.out $(INTERNAL_GATEWAY_IMAGE)
+envoy-config:
+	HAIL_DOMAIN=$(DOMAIN) python3 ../ci/ci/envoy.py internal-gateway ${HAIL}/letsencrypt/subdomains.txt ${HAIL}/internal-gateway/cds.yaml ${HAIL}/internal-gateway/rds.yaml
+	kubectl -n default apply -f envoy.yaml
```
jigold, line 9: Can we get a NAMESPACE check here like we do in `batch/ deploy`? We could theoretically allow other namespaces for example testing?, but I want to make it explicit from the user and not hard code default here.

gateway/deployment.yaml
```diff
@@ -53,6 +63,15 @@ spec:
...
             mountPath: /ssl-config
             readOnly: true
+          - name: gateway-envoy-config
+            mountPath: /etc/envoy
+            readOnly: true
+          - name: gateway-xds-config
+            mountPath: /config_map/xds
+            readOnly: true
+         env:
+         - name: ENVOY_UID
```
jigold, line 73: Same question about being root as for internal-gateway.

gateway/Makefile
```diff
@@ -1,22 +1,21 @@
...
-
-build:
-	$(MAKE) -C ../docker hail-ubuntu
-	python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"'$$(cat ../docker/hail-ubuntu-image-ref)'"}}' Dockerfile Dockerfile.out
-	python3 ../ci/jinja2_render.py '{"domain": "$(DOMAIN)","subdomains":[$(shell sed -e 's/\(.*\)/"\1"/' ../letsencrypt/subdomains.txt | paste -s -d, -)]}' gateway.nginx.conf gateway.nginx.conf.out
-	../docker-build.sh . Dockerfile.out $(GATEWAY_IMAGE)
+envoy-config:
+	HAIL_DOMAIN=$(DOMAIN) python3 ../ci/ci/envoy.py gateway ${HAIL}/letsencrypt/subdomains.txt ${HAIL}/gateway/cds.yaml ${HAIL}/gateway/rds.yaml
+	python3 ../ci/jinja2_render.py '{"domain":"$(DOMAIN)"}' envoy.yaml envoy.yaml.out
+	kubectl -n default apply -f envoy.yaml.out
```
jigold, line 10: Same comment as for internal gateway about not hardcoding default here.

gateway/envoy.yaml
```diff
@@ -0,0 +1,146 @@
...
+              http_filters:
+                - name: envoy.filters.http.ext_authz
+                  typed_config:
+                    "@type": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthz
+                    http_service:
+                      server_uri:
+                        uri: https://auth:443
+                        cluster: auth
+                        timeout: 0.25s
+                      path_prefix: /api/v1alpha/verify_dev_credentials?url=
```
jigold, line 75: Same comment as internal-gateway about providing a simpler example or documentation on how this is all structured.

jigold (dismissed):
I looked at the Envoy changes. I have not looked at the ci changes yet, but figured I'd give you this feedback now.

internal-gateway/deployment.yaml
daniel-goldstein: This was to be able to bind to port 80, but now that I'm thinking about it, we might not need the envoy to be listening on port 80 on the pod. I'll try an unprivileged port and try to get the load balancer service to send port 80 to the unprivileged port on the {internal-}gateway pod.

internal-gateway/deployment.yaml
daniel-goldstein: Good catch

internal-gateway/envoy.yaml
daniel-goldstein: There are lots of examples in the envoy repo and some explanations of terminology in their [documentation](https://www.envoyproxy.io/docs/envoy/v1.23.1/intro/life_of_a_request#terminology), here is one of the smaller examples I could find that shows a [proxy that supports https and sends all traffic to a single http upstream](https://github.com/envoyproxy/envoy/blob/main/examples/tls/envoy-https-http.yaml).

There's a lot of boilerplate but hopefully looking at the `virtual_hosts` section and the `clusters` section give a sense of how things work.

gateway/envoy.yaml
daniel-goldstein: Here's an example from their repo on using the [ext_authz](https://github.com/envoyproxy/envoy/blob/main/examples/ext_authz/config/http-service.yaml) filter.

internal-gateway/Makefile
daniel-goldstein: > We could theoretically allow other namespaces for example testing?

Deploying either of these in namespaces other than default would require allocating a GCP / Azure load balancer and a reserved IP address, which we don't have anything in the Makefiles to support doing that. For this particular target, deploying this into a different namespace would do nothing which I feel is confusing behavior. And we definitely don't want to accidentally point the external load balancer to a different namespace.

If what you're worried about is deploying by accident, would an "Are You Sure?" help? I don't love specifying NAMESPACE when doing anything but default wouldn't make sense, but another option would be to fail if you provide any value other than default

jigold: Yes an "Are you Sure" would be great! I think you had this functionality somewhere else already. If it's too much work, then don't bother.

jigold:
Do you need to replace all of the port numbers in the actual services (i.e. batch/front_end.py) as well?

daniel-goldstein:
> Do you need to replace all of the port numbers in the actual services (i.e. batch/front_end.py) as well?

Do you mean this because of the `ENVOY_UID` change? Restricting Envoy to an unprivileged user is something they've done in their docker image, and it's not something that we do for our k8s services (though maybe we should?). As such, our python services can bind to 443 all they like, and I have changed them to do so already in this PR, though we can put them on any port we want as long as we're consistent.

danking:
@daniel-goldstein After this lands in main, can we do a lets encrypt rotation just to smoke test that flow while this is all front of find? Thanks!

jigold:
> > Do you need to replace all of the port numbers in the actual services (i.e. batch/front_end.py) as well?
> 
> Do you mean this because of the `ENVOY_UID` change? Restricting Envoy to an unprivileged user is something they've done in their docker image, and it's not something that we do for our k8s services (though maybe we should?). As such, our python services can bind to 443 all they like, and I have changed them to do so already in this PR, though we can put them on any port we want as long as we're consistent.

I meant you changed ports to 8443 and 8080. Do you need to change the services ports?

gateway/service.yaml
```diff
@@ -10,11 +10,11 @@ spec:
   - name: http
     port: 80
     protocol: TCP
-    targetPort: 80
+    targetPort: 8080
```
daniel-goldstein, line 13: I change them here

auth/auth/auth.py
```diff
@@ -789,7 +789,7 @@ def run():
     web.run_app(
         deploy_config.prefix_application(app, 'auth'),
         host='0.0.0.0',
-        port=5000,
+        port=443,
```
jigold, line 792: I'm not being clear enough. I meant do you need to change this to 8443?

daniel-goldstein: No, 8443 is the port that gateway is listening on. It then forwards requests to auth:443

daniel-goldstein:
@jigold I stood up batch/ci in my own project from `hail-is/hail:main` and then deployed this branch, taking notes of any changes I needed to make and all seemed to work out OK. I think that's about as much as I can properly test this without trying things out in haildev/hail-vdc. The steps were as follows:

1. Generate the configmaps used by gateway/internal-gateway. These will have the routing configuration for production services (I've edited the bootstrap instructions to match)
    `make -C gateway envoy-xds-config && make -C internal-gateway envoy-xds-config`
2. … wait a few seconds for CI to quietly update these configmaps with information about testing namespaces … (can manually verify changes with `download-configmap gateway-xds-config`)
3. Deploy the new versions of gateway/internal-gateway
    `make -C gateway deploy NAMESPACE=default && make -C internal-gateway NAMESPACE=default`

This worked for me in my project with no downtime, but either way I would probably do the same thing as with the previous PR where I test it in azure before making changes to hail-vdc.

daniel-goldstein:
Mitigations if anything goes wrong is pretty trivial: just redeploy gateway/internal-gateway from an earlier commit on `main`

jigold:
Sounds good. Will review now and hopefully have it done early this afternoon.

ci/ci/ci.py
```diff
@@ -643,18 +650,57 @@ async def add_namespace(request, userdata):  # pylint: disable=unused-argument
...
     ]
     for namespace in expired_namespaces:
         assert namespace != 'default'
         log.info(f'Cleaning up expired namespace: {namespace}')
-        await db.execute_update('DELETE FROM active_namespaces WHERE namespace = %s', (namespace,))
+        await remove_namespace_from_db(db, namespace)
+
+
+async def update_envoy_configs(db: Database, k8s_client):
+    api_response = await k8s_client.list_namespace()
```
jigold, line 668: Can we get an assert that this is executing only in default namespace?

gateway/Makefile
```diff
@@ -1,20 +1,22 @@
...
+.PHONY: envoy-xds-config envoy-config deploy-no-xds deploy
 
-GATEWAY_IMAGE := $(DOCKER_PREFIX)/gateway:$(TOKEN)
 IP := $(shell kubectl get secret global-config --template={{.data.ip}} | base64 --decode)
+SHA := $(shell git rev-parse --short=12 HEAD)
 
-export NAMESPACE := default
+envoy-xds-config:
+	HAIL_DOMAIN=$(DOMAIN) python3 ../ci/ci/envoy.py gateway ${HAIL}/letsencrypt/subdomains.txt ${HAIL}/gateway/cds.yaml ${HAIL}/gateway/rds.yaml
+	kubectl -n default create configmap gateway-xds-config \
```
jigold, line 10: Would it be too much work to add `.out` to the ends of the `rds.yaml` and `cds.yaml` files so it's clear these are temp files in the directory?

gateway/Makefile
```diff
@@ -1,20 +1,22 @@
...
 
-build:
-	$(MAKE) -C ../docker hail-ubuntu
-	python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"'$$(cat ../docker/hail-ubuntu-image-ref)'"}}' Dockerfile Dockerfile.out
-	python3 ../ci/jinja2_render.py '{"domain": "$(DOMAIN)","subdomains":[$(shell sed -e 's/\(.*\)/"\1"/' ../letsencrypt/subdomains.txt | paste -s -d, -)]}' gateway.nginx.conf gateway.nginx.conf.out
-	../docker-build.sh . Dockerfile.out $(GATEWAY_IMAGE)
-
-deploy: build
-	python3 ../ci/jinja2_render.py '{"code":{"sha":"$(shell git rev-parse --short=12 HEAD)"},"global":{"ip":"$(IP)"}}' service.yaml service.yaml.out
+deploy:
```
jigold, line 16: Why doesn't this depend on `envoy-xds-config`?

gateway/Makefile
```diff
@@ -1,20 +1,22 @@
 include ../config.mk
 
-.PHONY: build push deploy
+.PHONY: envoy-xds-config envoy-config deploy-no-xds deploy
```
jigold, line 3: Looks like there's no `envoy-config` anymore. Is it missing because it looks like it's expected in the deployment. Ah I see, `envoy.yaml` is static in the repo.

infra/azure/README.md
```diff
@@ -128,9 +131,8 @@ Directory user's object ID.
 ./bootstrap.sh bootstrap <REPO>/hail:<BRANCH> create_initial_user <USERNAME> <OBJECT_ID>
 ```
 
-Deploy the gateway service. First trim down `$HAIL/letsencrypt/subdomains.txt`
-to only the services that are deployed and then run
+Deploy the gateway:
 
 ```
-make -C $HAIL/gateway deploy
+make -C $HAIL/gateway envoy-xds-config deploy
```
jigold, line 137: Ah I see. You didn't fold `envoy-xds-config` into `deploy`. Why is that?

internal-gateway/Makefile
```diff
@@ -1,19 +1,20 @@
...
 
-.PHONY: build push deploy
+.PHONY: envoy-xds-config envoy-config deploy
 
-INTERNAL_GATEWAY_IMAGE := $(DOCKER_PREFIX)/internal-gateway:$(TOKEN)
 INTERNAL_IP := $(shell kubectl get secret global-config --template={{.data.internal_ip}} | base64 --decode)
 
-export NAMESPACE := default
+envoy-xds-config:
+	HAIL_DOMAIN=$(DOMAIN) python3 ../ci/ci/envoy.py internal-gateway ${HAIL}/letsencrypt/subdomains.txt ${HAIL}/internal-gateway/cds.yaml ${HAIL}/internal-gateway/rds.yaml
```
jigold, line 8: Same questions as for `gateway`.

jigold (changes requested):
This looks great!

gateway/Makefile
daniel-goldstein: Nope I can do that

gateway/Makefile
daniel-goldstein: Ya so this is a great question that maybe deserves better documentation (your comments below I think all get at this point). The `envoy-xds-config` target creates the yaml for defining the routes and clusters for production services. The reason I didn't make it a dependency of deploy is that if you're running CI, then running this target will temporary erase the routes for test namespaces (this target and CI both overwrite the same configmap). I don't think it's a big deal if that happens, CI rewrites the config every 10 seconds anyway so the worst thing that would happen is tests get stalled for a few seconds. But, I thought it best to be explicit about running that target anyway.

daniel-goldstein: For anyone *not* running test namespaces, running `envoy-xds-config` doesn't do any damage. The only time you'd need to re-run it is if you decided to add another service.

gateway/Makefile
daniel-goldstein: Cleaned this up.

infra/azure/README.md
daniel-goldstein: Addressed in above comment. Happy to take any suggestions. The other option here is to go back to what I had before where `deploy` means "do everything" and there's another special target that does not overwrite the config. But this felt to me like the "safer" option that is less of a footgun.

gateway/Makefile
jigold: Can we make this more idiot-proof with a warning or check of some sort? I'm just worried that all of our other services are `make deploy` and we're going to forget there's an extra target that might need to have been run. I guess if the config maps don't exist initially then the deploy will fail anyways. If that's the case, then this is fine. Otherwise, maybe we add an extra make target which combines both of those and name it with the intention for its use case. i.e. `deploy_with_config`?

jigold: In other words, redeploying gateway and internal gateway already makes me nervous. I want to make sure I don't mess something up because the only time I'd probably need to do so is when the system is not under control.

infra/azure/README.md
jigold: See comment above.

gateway/Makefile
daniel-goldstein: >  I guess if the config maps don't exist initially then the deploy will fail anyways. If that's the case, then this is fine.

Yup! This should fail because the deployment itself depends on these configmaps existing. So the way it would work in practice is:

- `make -C gateway deploy` without ever doing `make -C gateway envoy-xds-config` will fail because K8s will complain that `gateway-envoy-xds-config` doesn't exist.
- `make -C gateway deploy` won't ever *lose* any configuration.
- If you add a production service *and* you're not running CI, you need to do the extra step of running the `envoy-xds-config` target when you `make deploy`. This also has no danger of overwriting any test config because you're not running CI.


Also happy to add an additional target `deploy_with_config`, but I don't really see much of a difference between that and `make -C gateway envoy-xds-config deploy`.

jigold: ok. I leave the decision in your hands. I think this is idiot proof enough for me.

daniel-goldstein: Ok, I've tried this a couple ways and I think this is the most reasonable option I could come up with for now. If there are complaints on ergonomics can definitely put in some time to make it smoother.

jigold:
I'm good with this.

@danking Did you want to take one last look especially at the feature parity list with Nginx that Daniel put together?

jigold (approved):

PR #12096, opened 2022-08-17
[boostrap-gateway] Remove bootstrap-gateway

daniel-goldstein:


daniel-goldstein:
Closing this for now. Didn't realize the bootstrap dependence of not having a letsencrypt config. I don't think we can entirely delete this just yet (but could trivially move it to envoy)

PR #12098, opened 2022-08-18
[query] Retry GCS GetObjectStream.read automatically

lgruen:
Context: https://hail.zulipchat.com/#narrow/stream/300487-Hail-Batch-Dev/topic/retries.20for.20.22read.20output.22.20in.20service.20backend.3F/near/292352578

Adds a generic `retry_transient_errors_wrapper` decorator, which could potentially be applied to more methods in the storage client implementations. This PR is just meant to fix a particular transient issue we've already encountered.

I ran into some issues getting a local dev deploy working, so I'm mostly relying on CI for this.

#assign services


daniel-goldstein (approved):
Thanks @lgruen! I was initially a little concerned about messing up any internal storage-client state but I think this will be fine

PR #12116, opened 2022-08-22
[ci] Fix retryable cloning of ci test repo

daniel-goldstein:
This step uses `set -e` so a failed clone fails the whole step. I fixed this using `retry` but also use a fresh temporary directory for each clones so if a clone fails part of the way through it doesn't reuse the broken directory.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2544,7 +2544,7 @@ steps:
       # destination on successful clone
       set +x
       clone() {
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
Example https://ci.azure.hail.is/watched_branches/0/pr/12115

danking (approved):

PR #12122, opened 2022-08-22
[auth] Cache user sessions in memory

daniel-goldstein:
This prevents decorators like `@rest_authenticated_users_only` from making an additional request to `auth` (which includes a database query for the `userinfo` endpoint on every API request. I also realized that auth wasn't getting scraped by prometheus so I added the `grafanak8sapp` label to fix that.

### DIFF
--- a/gear/gear/auth.py
+++ b/gear/gear/auth.py
@@ -12,17 +12,32 @@ from hailtop import httpx
 from hailtop.auth import async_get_userinfo
 from hailtop.config import get_deploy_config
 
+from .time_limited_max_size_cache import TimeLimitedMaxSizeCache
+
 log = logging.getLogger('gear.auth')
 
 deploy_config = get_deploy_config()
 
 BEARER = 'Bearer '
 
+TEN_SECONDS_NANOSECONDS = int(1e10)
+
+USERDATA_CACHE = None
+
 
 def maybe_parse_bearer_header(value: str) -> Optional[str]:
     if value.startswith(BEARER):
         return value[len(BEARER) :]
     return None
+
+
+def _init_userdata_cache(client_session: httpx.ClientSession):
+    global USERDATA_CACHE
+
+    async def load(session_id):
+        return _userdata_from_session_id(session_id, client_session)
+
+    USERDATA_CACHE = TimeLimitedMaxSizeCache(load, TEN_SECONDS_NANOSECONDS, 100, 'session_userdata_cache')
 
 
 async def _userdata_from_session_id(session_id: str, client_session: httpx.ClientSession):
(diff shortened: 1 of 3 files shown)


gear/gear/auth.py
```diff
@@ -12,19 +12,34 @@
...
 
+from .time_limited_max_size_cache import TimeLimitedMaxSizeCache
+
 log = logging.getLogger('gear.auth')
 
 deploy_config = get_deploy_config()
 
 BEARER = 'Bearer '
 
+TEN_SECONDS_NANOSECONDS = int(1e10)
```
jigold, line 23: Should this just be `TEN_NANOSECONDS`?

gear/gear/auth.py
```diff
@@ -53,8 +71,11 @@ async def userdata_from_rest_request(request):
     auth_header = request.headers['Authorization']
     session_id = maybe_parse_bearer_header(auth_header)
     if not session_id:
-        return session_id
-    return await _userdata_from_session_id(auth_header[7:], request.app['client_session'])
+        return None
+
```
jigold, line 75: Where did the `auth_header[7:]` go?

jigold (changes requested):

gear/gear/auth.py
daniel-goldstein: It's 10 seconds worth of nanoseconds, not just 10 nanoseconds. Should I name it `TEN_SECONDS_IN_NANOSECONDS` or maybe `TEN_SECONDS_IN_NS`

gear/gear/auth.py
daniel-goldstein: I'm pretty sure `session_id == auth_header[7:]`, since the slice should be taking out `Bearer: `

gear/gear/auth.py
jigold: I like `TEN_SECONDS_IN_NANOSECONDS`.

jigold (changes requested):
looks good other than variable name change

gear/gear/auth.py
```diff
@@ -12,19 +12,34 @@
...
 log = logging.getLogger('gear.auth')
 
 deploy_config = get_deploy_config()
 
 BEARER = 'Bearer '
 
+TEN_SECONDS_IN_NANOSECONDS = int(1e10)
+
+USERDATA_CACHE = None
```
danking, line 25: Hmm. Ok, I understand why you lazily initialize but I’m just generally a bit hesitant with mutation-based initialization like this. How bad is it if we have to explicitly initialize a class that exposes this file’s functions as an interface? And then annotations are like @authclient.web_authenticated_user etc.

danking (changes requested):
Not a requirement but I’d like to at least consider the cost-benefits here

gear/gear/auth.py
daniel-goldstein: Ya I don't like this either, but I needed some way to get the client session into the lookup function without it being the key. I'm not sure I see how using the class would help here because we use the decorators at the top-level and we only have the client session once we've entered main?

Maybe this is orthogonal to your point of having a class, which I'm not at all opposed to, but I suppose I could make the key a tuple of session_id and session, and get it to the lookup function that way. Both that and the current approach feel a little janky.

Is there one you prefer?

PR #12127, opened 2022-08-23
[infra] Move GCP preemptible pool to spot

daniel-goldstein:
The real change here is changing the preemptible pool config from `preemptible = true` to `spot = true`, but the `spot` config was only available in the new provider which involved a major version upgrade. The only incompatibility was the addition of an explicit `project` input to `google_project_iam_member`, as opposed to picking it up from the provider configuration.

Tested just now in my own project.

If one wants to apply this change without incurring downtime for preemptible deployments, they should follow the instructions outlined in the [migrating node pools dev-docs](https://github.com/hail-is/hail/blob/main/dev-docs/kubernetes-operations.md#when-using-terraform).

### DIFF
--- a/infra/gcp/main.tf
+++ b/infra/gcp/main.tf
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     google = {
       source = "hashicorp/google"
-      version = "3.48.0"
+      version = "4.32.0"
     }
     kubernetes = {
       source = "hashicorp/kubernetes"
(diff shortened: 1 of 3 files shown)


infra/gcp/main.tf
```diff
@@ -149,7 +149,7 @@ resource "google_container_node_pool" "vdc_preemptible_pool" {
   }
 
   node_config {
-    preemptible = true
+    spot = true
     machine_type = "n1-standard-4"
 
     labels = {
```
danking, line 155: We should also change the preemptible label and taint to reduce confusion.

Probably best to do that in three steps. Add a spot label to the nodes, add spot toleration to our code. Merge  a PR that node selects spot. Remove the preemptible label and add the spot taint.

We can do that in a separate PR series though.

danking (approved):

PR #12132, opened 2022-08-24
Upstream v0.2.98

illusional:
Awaiting DB migration advice before merging: https://hail.zulipchat.com/#narrow/stream/300487-Hail-Batch-Dev/topic/Stack.20of.20Database.20Migrations/near/295134561

- Includes a hefty DB migration which requires up-resourcing the DB instance
- Enables realtime DB billing (in USD for US-costed resources).

@lgruen, maybe we should update the resource costs somewhere to get a better approximation in the Batch UI?


illusional:
Sorry, wrong fork :(

PR #12133, opened 2022-08-25
[query][qob] implement requester pays in GoogleStorageFS

danking:
This enables Query-on-Batch pipelines to read from requester pays buckets.

@tpoterba curious for your thoughts on the flag situation. I suspect this PR will induce the Australians to start including requester pays config in their pipelines. If you describe an API you like, I can implement it for this PR.

Otherwise, I think this is ready. It works, it is tested. The changes to GoogleStorageFS suck, but its due to the reality of the GCS API.


hail/src/main/scala/is/hail/io/fs/GoogleStorageFS.scala
```diff
@@ -51,9 +54,77 @@ object GoogleStorageFileStatus {
...
+
+  private[this] def handleRequesterPays[T, U](
+    makeRequest: Seq[U] => T,
+    makeUserProjectOption: String => U,
+    bucket: String
+  ): T = {
+    try {
+      makeRequest(Seq())
+    } catch {
+      case exc: StorageException =>
```
tpoterba, line 118: This is pretty gross. Why don't we always include the RP logic?

tpoterba (changes requested):
Question on RP requests

hail/src/main/scala/is/hail/io/fs/GoogleStorageFS.scala
danking: If you supply a userProject it will be billed for the charges, even if the bucket is not requester pays. Consider, for example, a non-requester-pays bucket B1 owned by project P1. If a user (with read access to B1) issues a request to get the object identified by gs://B1/O1 with no userProject, then the class A operation will be billed to P1. If, instead, the user issues a request with userProject set to P2, the class A operation will be billed to P2.

Google does not provide a method for a non-owning-user to determine the requester pays status of a bucket. Ergo, the only way to know that a bucket is requires the userProject argument is to make the request, catch the error, and check the reason or the message. Apparently the API can alternately produce a message with a "reason" set to "userProjectMissing" or produce a message with an uninformative reason but include something about requester pays in the message.

It's a terrible API.

danking: I suppose an alternative we could pursue is this: the user must specify, a priori, all requester pays buckets they intend to use and the project to which to bill that usage. e.g.:

```python3
hl.init(requester_pays_config=(['bucket1', 'bucket2', 'bucket3', ...], 'my-project'))
```

We can then provide the userProject unconditionally on any request to bucket1, bucket2, bucket3, etc.

danking:
Also, I'm gonna add parameters to `hl.init` and set the flags in there. That punts the interface decision down the road by slightly restricting users (you can't change user project mid-pipeline).

danking:
OK, I don't love specifying it in `hl.init`, but that is implemented here. Assuming the tests pass, then this should work and it also means that users don't have to restart the Dataproc clusters to turn on requester pays, they can `hl.stop` and `hl.init`.

danking:
bump

danking:
disregard this still seems broken.

danking:
The layers of wtf really seem to have no end here.

Hadoop at least *appears* to [include the configuration in the cache key](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L3833-L3841) for its FileSystem cache, but it is actually just ignored by the constructor. Ergo, even if you stop the Hail context and try to start a new hail context with a new Hadoop Configuration, you'll get a filesystem configured by the first configuration.

I'm looking for a way around this now.

danking:
OK, I added some code to empty the cache when we `hl.stop`.

danking:
ripping my hair out

danking:
Truly a cursed PR. I added a new file with 10 rows so we don't have any empty partitions.

danking:
FWIW, Hadoop documents this behavior even though it seems out of step with other FS implementations like [FileInputStream](https://docs.oracle.com/javase/7/docs/api/java/io/FileInputStream.html). I can't find documentation on what happens in, say, Linux, but the internet suggests it is fine.

tpoterba:
Seeking to end of file is valid in multiple other implementations (checked Rust and CPython)

tpoterba (approved):

PR #12138, opened 2022-08-30
[qob] Upload function and contexts in parallel

daniel-goldstein:
This might have just gotten lost in the Big QoB PR, but I'm assuming that the reason to make these futures in the first place was to await them together, instead of in sequence.

### DIFF
--- a/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
@@ -149,8 +149,8 @@ class ServiceBackend(
       }
     }
 
-    scalaConcurrent.Await.result(uploadFunction, scalaConcurrent.duration.Duration.Inf)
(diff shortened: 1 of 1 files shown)


danking (changes requested):
A Future starts running as soon as you create it, e.g.:
```scala
scala> import scala.concurrent._
import scala.concurrent._

scala> implicit val ec = scala.concurrent.ExecutionContext.global
ec: scala.concurrent.ExecutionContextExecutor = scala.concurrent.impl.ExecutionContextImpl@77d680e6

scala> val x = Future { System.out.println("Hi Daniel") }
Hi Daniel
x: scala.concurrent.Future[Unit] = Future(<not completed>)

scala> x
res0: scala.concurrent.Future[Unit] = Future(Success(()))

```

daniel-goldstein:
Ah, whoops! Good to know

PR #12139, opened 2022-08-30
[compiler] enable new rng

patrick-schultz:
~Stacked on #12331 and #12309.~

Design doc [here](https://github.com/patrick-schultz/hail/blob/prng/enable/dev-docs/hail-query/randomness.md).


hail/python/hail/docs/functions/random.rst
```diff
@@ -69,50 +70,83 @@ All random functions can take a specified seed as an argument. This guarantees
...
-    >>> hl.set_global_seed(0)
-    >>> hl.eval(hl.array([hl.rand_unif(0, 1), hl.rand_unif(0, 1)]))  # doctest: +SKIP_OUTPUT_CHECK
-    [0.6830630912401323, 0.4035978197966855]
+    >>> hl.reset_global_randomness()
+    >>> hl.eval(hl.array([hl.rand_unif(0, 1), hl.rand_unif(0, 1)]))
+    [0.9828239225846387, 0.49094525115847415]
 
-    >>> hl.set_global_seed(0)
-    >>> hl.eval(hl.array([hl.rand_unif(0, 1), hl.rand_unif(0, 1)]))  # doctest: +SKIP_OUTPUT_CHECK
-    [0.6830630912401323, 0.4035978197966855]
```
danking, line 115: The text above references `set_global_seed` which the code does not.

I think I'm a bit confused. The text above says that the global seed "is fixed for the entire session", but we appear to modify it here.

patrick-schultz: Ah thanks, meant to remove all mentions of `set_global_seed`

danking:
I resolved conflicts. There were two issues in ir.py and one in aggregators.py. The former were resolved by: replace if with assert, add a new field about needing randomness. The latter is visible in the diff and switches from frozenset to set.

danking:
Bumping this PR, I'd like it to land so I can nail down exactly why my flags PR is causing BN tests to fail.

hail/python/hail/methods/pca.py
```diff
@@ -383,7 +381,7 @@ def _spectral_moments(A, num_moments, p=None, moment_samples=500, block_size=128
     # TODO: When moment_samples > n, we should just do a TSQR on A, and compute
     # the spectrum of R.
     assert moment_samples < n, '_spectral_moments: moment_samples must be smaller than num cols of A'
-    G = hl.nd.zeros((n, moment_samples)).map(lambda n: hl.if_else(hl.rand_bool(0.5), -1, 1))
+    G = hl.rand_unif(-1, 1, size=(n, moment_samples)).map(lambda x: hl.sign(x))
```
tpoterba, line 384: good change.

hail/python/hail/utils/java.py
```diff
@@ -113,14 +112,15 @@ def dummy_table():
         return Env._dummy_table
 
     @staticmethod
-    def set_seed(seed):
-        Env._seed_generator = hail.utils.HailSeedGenerator(seed)
+    def next_static_rng_uid():
+        result = Env._static_rng_uid
+        assert(result <= 0xFFFF_FFFF_FFFF_FFFF)
```
tpoterba, line 117: This should keep Konrad contained for at least a few years

hail/src/main/scala/is/hail/types/physical/PCanonicalNDArray.scala
```diff
@@ -174,6 +174,14 @@ final case class PCanonicalNDArray(elementType: PType, nDims: Int, required: Boo
     constructByCopyingDataPointer(shape, strides, this.allocateData(shape, region), cb, region)
   }
 
+  def constructUnintialized(
```
tpoterba, line 177: typo: uninitialized

tpoterba (changes requested):
Tiny comments. Python IR and Scala RNG changes are super clear, great work.

hail/python/hail/ir/ir.py
```diff
@@ -187,12 +187,12 @@ def __init__(self, typ):
         self._typ = typ
 
     def _handle_randomness(self, create_uids):
-        if create_uids:
-            if isinstance(self.typ.element_type, tstruct):
-                new_elt_typ = self.typ.element_type._insert_field(uid_field_name, tint64)
-            else:
-                new_elt_typ = ttuple(tint64, self.typ.element_type)
-            return NA(tstream(new_elt_typ))
+        assert create_uids
```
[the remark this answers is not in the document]
patrick-schultz: Calls to `_handle_randomness` are guarded in `IR.handle_randomness`, which should be the only caller:
```
        if (create_uids == self.has_uids) and not self.needs_randomness_handling:
            return self
        new = self._handle_randomness(create_uids)
```
so the only reason this will be called is if `create_uids` is true.

hail/python/hail/expr/functions.py
```diff
@@ -43,10 +43,17 @@ def _func(name, ret_type, *args, type_args=()):
 
 
 def _seeded_func(name, ret_type, seed, *args):
-    seed = seed if seed is not None else Env.next_seed()
+    if seed is None:
+        static_rng_uid = Env.next_static_rng_uid()
+    else:
+        if Env._hc is None or not Env._hc._user_specified_rng_nonce:
+            warning('To ensure reproducible randomness across Hail sessions, '
```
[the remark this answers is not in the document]
patrick-schultz: I didn't realize, but I think that's fine. The warning is saying you should really initialize manually before doing anything else, not to worry about what functions do or don't implicitly initialize.

patrick-schultz:
This is still stacked on #12309 as well

danking:
This seems to be triggering hail initialization somewhere it not to?

patrick-schultz:
Ah, it was just a new test file that used `setUpModule = startTestHailContext`. Btw, after this merges, no need to do that anymore. Now pytest fixtures always ensure there's a hail context.

tpoterba:
Ship it!

tpoterba (approved):

danking:
Thank you both!

PR #12145, opened 2022-08-31
[batch] Make frozen 503 a middleware for POST and PATCH

daniel-goldstein:


danking (changes requested):
this also prevents cancellation; should we prevent that? I haven't thought it through but I'm curious why we didn't include it originally.

daniel-goldstein:
Good question, I was under the interpretation that frozen means batch should become read-only and clients shouldn't be able to effect any changes in the state of the system. If that's the case, I think this is the right way to make sure we don't accidentally leave out endpoints, but if it's more nuanced than that (i.e. we want to allow cancellation), then we should be explicit and use a decorator. I'll add Jackie for review and see what she thinks.

daniel-goldstein:
@jigold No rush to get here but I wasn't sure if you would see this since Dan has Changes Requested on it

jigold:
I don't think it matters too much, but we should aim for consistency. In frozen mode, cancelling a batch will flip the database state and I think the canceller loop will cancel running jobs and pending jobs despite being in frozen mode. The scheduler will not schedule Ready jobs that are always_run though. If we did not allow cancellation in frozen mode, then the existing running jobs would run until completion which might cost users money that is unnecessary.

I guess writing this out now I'm in favor of giving the users the option to cancel as it's not really impacted by the driver being in frozen mode and will save the user money.

Although, the reason we'd put Batch in frozen mode is because there's a major problem and we don't want to add stress to the system.

Thoughts now that I've written this out?

daniel-goldstein:
I feel like being able to cancel but then `always_run` jobs not running is surprising behavior, and

> Although, the reason we'd put Batch in frozen mode is because there's a major problem and we don't want to add stress to the system.

this is the biggest thing. Maybe I'm overcautious but I feel like when there is a major problem it will be nice to have the assurance that a user cannot affect the system while we are investigating.

jigold:
Ok. Let's be consistent everywhere.

danking:
I can approve if Jackie consents to this PR.

jigold (approved):

PR #12146, opened 2022-08-31
[batch] Dont include open-ended job ranges in token cache

daniel-goldstein:
We keep an in-memory cache on the driver for which job ranges correspond to which job spec token. Currently, we cache the open range [N, None) which corresponds to the last bunch in the batch. This is fine when the batch is all submitted at once, but can become stale if there are updates to the batch. Here, we just use the range [N, n_jobs), which even if the number of jobs in the batch changes should still be correct.

Ultimately, I think we should just add the end id to the batch_bunches table instead of doing this subquery and join against the batches table. This is a bit of a hack.

### DIFF
--- a/batch/batch/spec_writer.py
+++ b/batch/batch/spec_writer.py
@@ -1,6 +1,6 @@
 import collections
 import logging
-from typing import Dict, Tuple, Optional
+from typing import Dict, Tuple
 
 import sortedcontainers
 
--- a/batch/batch/spec_writer.py
+++ b/batch/batch/spec_writer.py
@@ -79,7 +79,7 @@ LIMIT 1;
         )
         token = bunch_record['token']
         start_job_id = bunch_record['start_job_id']
-        end_job_id = bunch_record['next_start_job_id'] or bunch_record['n_jobs']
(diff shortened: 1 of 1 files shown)


batch/batch/spec_writer.py
```diff
@@ -57,30 +55,32 @@ async def get_token_start_id(db, batch_id, job_id) -> Tuple[str, int]:
...
 LIMIT 1;
 ''',
             (batch_id, job_id, batch_id, job_id),
             'get_token_start_id',
         )
         token = bunch_record['token']
         start_job_id = bunch_record['start_job_id']
-        next_start_job_id = bunch_record['next_start_job_id']
-        return (token, start_job_id, next_start_job_id)
+        end_job_id = bunch_record['next_start_job_id'] or bunch_record['n_jobs']
```
danking, line 82: it's inclusive of n_jobs, isn't it?

danking (dismissed):
ugh, we should really migrate this table, this whole subquery thing feels error prone.

jigold:
Is this part of the open batches changes or a critical fix?

daniel-goldstein:
This removes an assumption that is valid now but is invalidated by open batches. So can go in now but must go in before open batches.

danking (approved):
Just a lament.

PR #12147, opened 2022-08-31
[batch] Remove lint errors from test_batch and test_accounts

daniel-goldstein:
Basically all naming, rids these test files of linting errors. We do a lot of reassigning a `BatchBuilder` variable to a `Batch` and so I consolidated around `bb` and `b`. A couple instances where I remove debug_info from an assert statement is because the associated `Batch` object would not exist, since that assert is triggered by an error that's raised before the `Batch` object is created.


danking (approved):
❤️

PR #12159, opened 2022-09-06
[query] Stage IndexReader

tpoterba:
look at that deleted code! Good riddance.


hail/src/main/scala/is/hail/io/index/StagedIndexReader.scala
```diff
@@ -95,15 +95,11 @@ class StagedIndexReader(emb: EmitMethodBuilder[_], spec: AbstractIndexSpec) {
       endQuerySettable.asBaseStruct.loadField(cb, 0).get(cb).asInt64.value -
         startQuerySettable.asBaseStruct.loadField(cb, 0).get(cb).asInt64.value
     )
+    cb.assign(startLeaf, cb.memoize(startQuerySettable.asBaseStruct.loadField(cb, 1)))
```
chrisvittal, line 98: is the memoize call needed, aren't we assigning it to a variable anyways?

chrisvittal (dismissed):
one nit, lgtm otherwise

hail/src/main/scala/is/hail/io/index/StagedIndexReader.scala
tpoterba: nope, fixed

chrisvittal (approved):
🎉

PR #12164, opened 2022-09-08
[query] Allow seeks more than MAX_INT bytes away

daniel-goldstein:
This assert is here because the buffer we use in the FS is a Java array and can't be longer than MAX_INT, but clearly the position in the blob can be more than that. This currently breaks range reads into indexes on data that's more than 2GiB.

### DIFF
--- a/hail/src/test/scala/is/hail/fs/FSSuite.scala
+++ b/hail/src/test/scala/is/hail/fs/FSSuite.scala
@@ -318,11 +318,12 @@ trait FSSuite {
   @Test def testSeekMoreThanMaxInt(): Unit = {
     val f = t()
     using (fs.create(f)) { os =>
-      val eight_mib = Array.fill(8 * 1024 * 1024){0.toByte}
+      val eight_mib = 8 * 1024 * 1024
+      val arr = Array.fill(){0.toByte}
       var i = 0
       // 256 * 8MiB = 2GiB
       while (i < 256) {
-        os.write(eight_mib, 0, Int.MaxValue)
+        os.write(arr, 0, eight_mib)
         i = i + 1
       }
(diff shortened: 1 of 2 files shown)


hail/src/test/scala/is/hail/fs/FSSuite.scala
```diff
@@ -314,6 +314,33 @@ trait FSSuite {
...
+
+  @Test def testSeekMoreThanMaxInt(): Unit = {
+    val f = t()
+    using (fs.create(f)) { os =>
+      val eight_mib = Array.fill(8 * 1024 * 1024){0.toByte}
+      val i = 0
+      // 256 * 8MiB = 2GiB
+      while (i < 256) {
+        os.write(eight_mib, 0, Int.MaxValue)
+        i += 1
```
daniel-goldstein, line 326: I clearly can't write scala

danking (approved):

hail/src/test/scala/is/hail/fs/FSSuite.scala
```diff
@@ -314,6 +314,33 @@ trait FSSuite {
     assert(dropTrailingSlash("/foo///") == "/foo")
     assert(dropTrailingSlash("///") == "")
   }
+
+  @Test def testSeekMoreThanMaxInt(): Unit = {
+    val f = t()
+    using (fs.create(f)) { os =>
+      val eight_mib = Array.fill(8 * 1024 * 1024){0.toByte}
+      val i = 0
```
danking, line 322:
```suggestion
      var i = 0
```

danking (changes requested):

danking (approved):

PR #12168, opened 2022-09-08
[query] implement bit_count operation

iris-garden:
This change adds the `bit_count` operation for 32- and 64-bit `int`s (counts the number of 1s in the binary representation of the `int`).


hail/python/hail/expr/functions.py
```diff
@@ -6193,6 +6193,28 @@ def bit_not(x):
     return construct_expr(ir.ApplyUnaryPrimOp('~', x._ir), x.dtype, x._indices, x._aggregations)
 
 
+@typecheck(x=expr_oneof(expr_int32, expr_int64))
+def bit_count(x):
+    """Count the number of 1s in the binary representation of `x`.
```
patrick-schultz, line 6198: For completeness, should probably say
```
in the `two's complement <https://en.wikipedia.org/wiki/Two%27s_complement>__` binary representation ...
```

hail/python/hail/expr/functions.py
```diff
@@ -6193,6 +6193,28 @@ def bit_not(x):
...
+
+    Parameters
+    ----------
+    x : :class:`.Int32Expression` or :class:`.Int64Expression`
+
+    Returns
+    -------
+    :class:`.Int32Expression`
+    """
+    return construct_expr(ir.ApplyUnaryPrimOp('BitCount', x._ir), x.dtype, x._indices, x._aggregations)
```
patrick-schultz, line 6215: This should always return an int32 expression, not `x.dtype`.

It looks like you'll need to fix `ApplyUnaryPrimOp._compute_type` (in python) as well. Currently it assumes all unary ops preserve the type of the argument. Now you'll need to handle `BitCount` as a special case, and always return `tint32`.

hail/python/test/hail/expr/test_expr.py
```diff
@@ -3616,6 +3616,9 @@ def test_bit_ops_types(self):
         assert hl.bit_not(1).dtype == hl.tint32
         assert hl.bit_not(hl.int64(1)).dtype == hl.tint64
 
+        assert hl.bit_count(1).dtype == hl.tint32
+        assert hl.bit_count(hl.int64(1)).dtype == hl.tint64
```
patrick-schultz, line 3620: Should be `hl.tint32` here too.

patrick-schultz (changes requested):
Looks great! Just a few small things.

hail/python/hail/expr/functions.py
iris-garden: Updated!

hail/python/hail/expr/functions.py
iris-garden: Updated!

hail/python/test/hail/expr/test_expr.py
iris-garden: Updated!

hail/python/hail/expr/functions.py
```diff
@@ -6193,6 +6193,28 @@ def bit_not(x):
...
+
+    >>> hl.eval(hl.bit_count(7))
+    3
+
+    Parameters
+    ----------
+    x : :class:`.Int32Expression` or :class:`.Int64Expression`
+
+    Returns
+    ------hl.-
```
patrick-schultz, line 6212: typo

patrick-schultz (dismissed):
Just fix the typo in the docs and this looks good to go!

hail/python/hail/expr/functions.py
iris-garden: fixed 😅

patrick-schultz (approved):

PR #12177, opened 2022-09-12
[auth] Cache user sessions in memory

daniel-goldstein:
A revival of #12122, which was reverted because the web-based code path had an error. The last commit here fixes that bug and I tested that the web path works through a dev deploy.

### DIFF
--- a/gear/gear/auth.py
+++ b/gear/gear/auth.py
@@ -20,7 +20,7 @@ deploy_config = get_deploy_config()
 
 BEARER = 'Bearer '
 
-TEN_SECONDS_NANOSECONDS = int(1e10)
+TEN_SECONDS_IN_NANOSECONDS = int(1e10)
 
 USERDATA_CACHE = None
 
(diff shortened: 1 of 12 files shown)


gear/gear/auth.py
```diff
@@ -12,110 +12,79 @@
...
     return None
 
 
-async def _userdata_from_session_id(session_id: str, client_session: httpx.ClientSession):
-    try:
-        return await async_get_userinfo(
-            deploy_config=deploy_config, session_id=session_id, client_session=client_session
+class AuthClient:
+    def __init__(self):
+        self._userdata_cache = TimeLimitedMaxSizeCache(
```
danking, line 34: I'm disappointed pylint didn't catch the type error. `TimeLimitedMaxSizeCache` has the right type parameters. Maybe we're not linting gear? Or maybe it's not smart enough to figure out the type parameters from `_load_userdata`?

danking (approved):

gear/gear/auth.py
daniel-goldstein: My assumption is that it's because it has no idea what type `session['session_id']` is, so it could have been a tuple for all it knows

PR #12185, opened 2022-09-15
[lowering] Lower BGEN export

chrisvittal:


chrisvittal:
This is green now. Testing it fully is difficult since we don't have lowered import. However, manually testing both lowered and unlowered versions has produced bit-for-bit identical output, and [test_bgen_export_from_vcf](https://github.com/chrisvittal/hail/blob/e0de6b3f5c91820be94591e95d1a49b67c83cd7f/hail/python/test/hail/methods/test_impex.py#L1521-L1536) does use the lowered execution for export and passes.

tpoterba:
awesome!

tpoterba (changes requested):
tiny comment, otherwise looks great.

tpoterba (approved):

PR #12189, opened 2022-09-15
[docker] Delete unused file and gitignore

daniel-goldstein:
Couldn't find any usages of `stat-permissions` or `stmp`. I believe the latter has been superseded by the `*-image-ref`?


danking (approved):
Yeah I think that's right.

stat-permissions.sh was an attempt by Cotton to get Docker image caching to work in the face of some weird permissions issues we saw. I think, in the end, it wasn't the root issue.

PR #12192, opened 2022-09-16
[monitoring] disable grafana-specific login screen 

iris-garden:
Currently, the Grafana service deployed with the Hail environment is behind two layers of authentication, since the Grafana NGINX configuration proxies requests to it through the `/auth` route, and the login screen built into Grafana also displays. This change removes the second login screen.

### DIFF
--- a/grafana/deployment.yaml
+++ b/grafana/deployment.yaml
@@ -131,10 +131,8 @@ data:
     root_url = %(protocol)s://%(domain)/{{ default_ns.name }}/grafana/
     serve_from_sub_path = true
 {% endif %}
-    [auth.anonymous]
-    enabled = true
-    org_name = Hail
-    org_role = Admin
+    [auth]
+    disable_login_form
 ---
 apiVersion: v1
 kind: ConfigMap
--- a/grafana/deployment.yaml
+++ b/grafana/deployment.yaml
@@ -131,8 +131,10 @@ data:
     root_url = %(protocol)s://%(domain)/{{ default_ns.name }}/grafana/
     serve_from_sub_path = true
 {% endif %}
-    [auth]
-    oauth_automatic_login = true
+    [auth.anonymous]
+    enabled = true
+    org_name = Hail
+    org_role = Admin
 ---
 apiVersion: v1
 kind: ConfigMap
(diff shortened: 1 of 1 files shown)


grafana/deployment.yaml
```diff
@@ -131,6 +131,15 @@ data:
     root_url = %(protocol)s://%(domain)/{{ default_ns.name }}/grafana/
     serve_from_sub_path = true
 {% endif %}
```
daniel-goldstein, line 133: I'm not super familiar with the config here. Could you explain why we need all of disabling basic auth, enabling anonymous auth, and hiding the forms?

daniel-goldstein (changes requested):

grafana/deployment.yaml
iris-garden: sure! per [the grafana docs](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/grafana/), basic auth (logging in directly with a username and a password) is enabled by default. i had to trial-and-error this, but it turns out that disabling all auth simply creates a grafana instance that cannot be accessed at all, so it's also necessary to enable anonymous auth, which allows access to the instance without logging in. however, just enabling anonymous auth leaves basic auth still enabled, which apparently causes grafana to choose the latter, keeping the instance behind the basic auth after all. disabling the signout menu is recommended for our use case of putting grafana behind the nginx auth, since the button wouldn't do what it claims (the user can't actually log into grafana in the first place, so logging them out wouldn't do anything). i'm not actually sure it's necessary to disable the login form, since it might actually just not show it if the basic auth and anonymous auth configs are correct? i will give it a try without that and let you know whether it needs to stay like this!

daniel-goldstein: > however, just enabling anonymous auth leaves basic auth still enabled, which apparently causes grafana to choose the latter, keeping the instance behind the basic auth after all.

Love to see it

iris-garden: removed the `disable_login_form` setting and it still works!

daniel-goldstein (approved):
Thanks for the explanation! LGTM

PR #12193, opened 2022-09-16
[batch] Cleanup adding update id infrastructure

jigold:
Replacement for #12120. Stacked on #12117. 

The `foreign_key_checks=0` is necessary in order to avoid locking the entire table while copying the whole thing.

PR #12199, opened 2022-09-19
[batch] Open Batches!

jigold:
Based off of #12154. Stacked on #12193.

### DIFF
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -415,9 +415,7 @@ WHERE user = %s AND `state` = 'running';
                     '''
 SELECT job_id, spec, cores_mcpu
 FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_inst_coll_cancelled)
-INNER JOIN batch_updates
-ON jobs.batch_id = batch_updates.batch_id AND jobs.update_id = batch_updates.update_id
-WHERE jobs.batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s AND batch_updates.committed
+WHERE jobs.batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s
 LIMIT %s;
 ''',
                     (batch['id'], self.pool.name, remaining.value),
@@ -433,9 +431,7 @@ LIMIT %s;
                         '''
 SELECT job_id, spec, cores_mcpu
 FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_cancelled)
-INNER JOIN batch_updates
-ON jobs.batch_id = batch_updates.batch_id AND jobs.update_id = batch_updates.update_id
-WHERE jobs.batch_id = %s AND state = 'Ready' AND always_run = 0 AND inst_coll = %s AND cancelled = 0 AND batch_updates.committed
+WHERE jobs.batch_id = %s AND state = 'Ready' AND always_run = 0 AND inst_coll = %s AND cancelled = 0
 LIMIT %s;
 ''',
                         (batch['id'], self.pool.name, remaining.value),
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -411,16 +411,8 @@ WHERE user = %s AND `state` = 'running';
                 (user,),
                 "user_runnable_jobs__select_running_batches",
             ):
-                # TODO Probably need to index jobs.update_id
                 async for record in self.db.select_and_fetchall(
                     '''
-SELECT job_id, spec, cores_mcpu
-FROM batch_updates
-STRAIGHT_JOIN jobs FORCE INDEX(jobs_batch_id_state_always_run_inst_coll_cancelled)
-ON batch_updates.batch_id = jobs.batch_id AND batch_updates.update_id = jobs.update_id
-WHERE batch_updates.batch_id = %s AND batch_updates.committed AND jobs.batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s
-LIMIT %s;
-
 SELECT job_id, spec, cores_mcpu
 FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_inst_coll_cancelled)
 INNER JOIN batch_updates
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -411,11 +411,21 @@ WHERE user = %s AND `state` = 'running';
                 (user,),
                 "user_runnable_jobs__select_running_batches",
             ):
+                # TODO Probably need to index jobs.update_id
                 async for record in self.db.select_and_fetchall(
                     '''
 SELECT job_id, spec, cores_mcpu
+FROM batch_updates
+STRAIGHT_JOIN jobs FORCE INDEX(jobs_batch_id_state_always_run_inst_coll_cancelled)
+ON batch_updates.batch_id = jobs.batch_id AND batch_updates.update_id = jobs.update_id
+WHERE batch_updates.batch_id = %s AND batch_updates.committed AND jobs.batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s
+LIMIT %s;
+
+SELECT job_id, spec, cores_mcpu
 FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_inst_coll_cancelled)
-WHERE batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s
+INNER JOIN batch_updates
+ON jobs.batch_id = batch_updates.batch_id AND jobs.update_id = batch_updates.update_id
+WHERE jobs.batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s AND batch_updates.committed
 LIMIT %s;
 ''',
                     (batch['id'], self.pool.name, remaining.value),
@@ -431,7 +441,9 @@ LIMIT %s;
                         '''
 SELECT job_id, spec, cores_mcpu
 FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_cancelled)
-WHERE batch_id = %s AND state = 'Ready' AND always_run = 0 AND inst_coll = %s AND cancelled = 0
+INNER JOIN batch_updates
+ON jobs.batch_id = batch_updates.batch_id AND jobs.update_id = batch_updates.update_id
+WHERE jobs.batch_id = %s AND state = 'Ready' AND always_run = 0 AND inst_coll = %s AND cancelled = 0 AND batch_updates.committed
 LIMIT %s;
 ''',
                         (batch['id'], self.pool.name, remaining.value),
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -415,7 +415,7 @@ WHERE user = %s AND `state` = 'running';
                     '''
 SELECT job_id, spec, cores_mcpu
 FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_inst_coll_cancelled)
-WHERE jobs.batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s
+WHERE batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s
 LIMIT %s;
 ''',
                     (batch['id'], self.pool.name, remaining.value),
@@ -431,7 +431,7 @@ LIMIT %s;
                         '''
 SELECT job_id, spec, cores_mcpu
 FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_cancelled)
-WHERE jobs.batch_id = %s AND state = 'Ready' AND always_run = 0 AND inst_coll = %s AND cancelled = 0
+WHERE batch_id = %s AND state = 'Ready' AND always_run = 0 AND inst_coll = %s AND cancelled = 0
 LIMIT %s;
 ''',
                         (batch['id'], self.pool.name, remaining.value),
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -611,15 +611,12 @@ WITH base_t AS (
     batches_n_jobs_in_complete_states.n_completed,
     batches_n_jobs_in_complete_states.n_succeeded,
     batches_n_jobs_in_complete_states.n_failed,
-    batches_n_jobs_in_complete_states.n_cancelled,
-    MAX(batch_updates.time_committed) AS time_updated
+    batches_n_jobs_in_complete_states.n_cancelled
   FROM batches
   LEFT JOIN batches_n_jobs_in_complete_states
     ON batches.id = batches_n_jobs_in_complete_states.id
   LEFT JOIN batches_cancelled
     ON batches.id = batches_cancelled.id
-  LEFT JOIN batch_updates
-    ON batches.id = batch_updates.batch_id
   STRAIGHT_JOIN billing_project_users ON batches.billing_project = billing_project_users.billing_project
   WHERE {' AND '.join(where_conditions)}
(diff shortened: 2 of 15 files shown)


batch/batch/driver/instance_collection/pool.py
```diff
@@ -415,7 +415,9 @@ async def user_runnable_jobs(user, remaining):
                     '''
 SELECT job_id, spec, cores_mcpu
 FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_inst_coll_cancelled)
-WHERE batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s
+INNER JOIN batch_updates
```
jigold, line 418:
```
mysql> EXPLAIN SELECT job_id, spec, cores_mcpu -> FROM batch_updates -> STRAIGHT_JOIN jobs FORCE INDEX(jobs_batch_id_state_always_run_inst_coll_cancelled) -> ON batch_updates.batch_id = jobs.batch_id AND batch_updates.update_id = jobs.update_id -> WHERE batch_updates.batch_id = 1000000 AND batch_updates.committed AND jobs.batch_id = 1000000 AND state = 'Ready' AND always_run = 1 AND inst_coll = "standard" -> LIMIT 5; +----+-------------+---------------+------------+------+---------------------------------------------------------------------+----------------------------------------------------+---------+-------------------------+------+----------+-------------+ | id | select_type | table | partitions | type | possible_keys | key | key_len | ref | rows | filtered | Extra | +----+-------------+---------------+------------+------+---------------------------------------------------------------------+----------------------------------------------------+---------+-------------------------+------+----------+-------------+ | 1 | SIMPLE | batch_updates | NULL | ref | PRIMARY,batch_id,batch_updates_committed,batch_updates_start_job_id | PRIMARY | 8 | const | 1 | 90.00 | Using where | | 1 | SIMPLE | jobs | NULL | ref | jobs_batch_id_state_always_run_inst_coll_cancelled | jobs_batch_id_state_always_run_inst_coll_cancelled | 899 | const,const,const,const | 1 | 10.00 | Using where | +----+-------------+---------------+------------+------+---------------------------------------------------------------------+----------------------------------------------------+---------+-------------------------+------+----------+-------------+ 2 rows in set, 1 warning (0.01 sec)
```

batch/batch/driver/instance_collection/pool.py
```diff
@@ -431,7 +433,9 @@ async def user_runnable_jobs(user, remaining):
                         '''
 SELECT job_id, spec, cores_mcpu
 FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_cancelled)
-WHERE batch_id = %s AND state = 'Ready' AND always_run = 0 AND inst_coll = %s AND cancelled = 0
+INNER JOIN batch_updates
```
jigold, line 436:
```
mysql> EXPLAIN SELECT job_id, spec, cores_mcpu
    ->  FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_cancelled)
    ->  INNER JOIN batch_updates
    ->  ON jobs.batch_id = batch_updates.batch_id AND jobs.update_id = batch_updates.update_id
    ->  WHERE jobs.batch_id = 1000000 AND state = 'Ready' AND always_run = 0 AND inst_coll = "standard" AND cancelled = 0 AND batch_updates.committed
    ->  LIMIT 5;
+----+-------------+---------------+------------+--------+---------------------------------------------------------------------+------------------------------------------+---------+----------------------------+------+----------+-------------+
| id | select_type | table         | partitions | type   | possible_keys                                                       | key                                      | key_len | ref                        | rows | filtered | Extra       |
+----+-------------+---------------+------------+--------+---------------------------------------------------------------------+------------------------------------------+---------+----------------------------+------+----------+-------------+
|  1 | SIMPLE      | jobs          | NULL       | ref    | jobs_batch_id_state_always_run_cancelled                            | jobs_batch_id_state_always_run_cancelled | 132     | const,const,const,const    |    1 |    10.00 | Using where |
|  1 | SIMPLE      | batch_updates | NULL       | eq_ref | PRIMARY,batch_id,batch_updates_committed,batch_updates_start_job_id | PRIMARY                                  | 12      | const,batch.jobs.update_id |    1 |    90.00 | Using where |
+----+-------------+---------------+------------+--------+---------------------------------------------------------------------+------------------------------------------+---------+----------------------------+------+----------+-------------+
2 rows in set, 1 warning (0.01 sec)
```

daniel-goldstein: Let's revert this now that the `commit_batch_update` procedure guarantees jobs in a pending update won't be `Ready`.

batch/batch/front_end/front_end.py
```diff
@@ -1228,10 +1255,10 @@ async def insert(tx):
...
                 json.dumps(attributes),
                 batch_spec.get('callback'),
-                batch_spec['n_jobs'],
+                0,
                 now,
                 token,
-                'open',
+                'complete',
                 BATCH_FORMAT_VERSION,
                 batch_spec.get('cancel_after_n_failures'),
```
daniel-goldstein, line 1263: I think this is something that I accidentally left out from your original changes: the `time_completed` should be set here to `now`

batch/test/test_batch.py
```diff
@@ -1070,3 +1070,101 @@ def test_job_private_instance_cancel(client: BatchClient):
...
+
+
+def test_empty_update(client: BatchClient):
+    bb = client.create_batch()
+    bb.create_job(DOCKER_ROOT_IMAGE, ['true'])
+    b = bb.submit()
+    b = bb.submit()
+    status = b.wait()
+
+    assert status['state'] == 'success', str((status, b.debug_info()))
```
daniel-goldstein, line 1093: Let's right the following test case:

1. Submit with an hour long sleep job, `j1`
2. Submit another update with an always-run `true` job, `j2` which is a child of `j1`
3. Verify that `j1` is `Running` and `j2` is `Pending`
4. Cancel `j1`
5. Wait on `j2` and verify that it completed successfully

daniel-goldstein: Another similar test case that just does a sanity check that jobs from different updates aren't somehow serialized

1. Submit an hour long sleep job, `j1`
2. Submit another update with a `true` job `j2`
3. Wait on `j2` and see that it completes successfully
4. Verify that `j1` is still `Running`
5. Cancel `j1` so the job ends and developers aren't sad waiting for testing to finish

batch/test/test_batch.py
```diff
@@ -1070,3 +1070,101 @@ def test_job_private_instance_cancel(client: BatchClient):
...
+    bb = client.create_batch()
+    j1 = bb.create_job(DOCKER_ROOT_IMAGE, ['true'])
+    b = bb.submit()
+    j1.wait()
+    j2 = bb.create_job(DOCKER_ROOT_IMAGE, ['true'], parents=[j1])
+    b = bb.submit()
+    status = j2.wait()
+
+    assert status['state'] == 'Success', str((status, b.debug_info()))
+
```
daniel-goldstein, line 1106: I think we also need tests here that try to submit updates to a batch that has been cancelled / deleted, in which case I suspect we get a 400 from the front-end

jigold:
@daniel-goldstein Let's see if the tests pass now, but should be really close!!!

jigold:
I forgot to add the UI time_updated field... Will do so now.

batch/batch/driver/instance_collection/pool.py
```diff
@@ -415,7 +415,7 @@ async def user_runnable_jobs(user, remaining):
                     '''
 SELECT job_id, spec, cores_mcpu
 FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_inst_coll_cancelled)
-WHERE batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s
+WHERE jobs.batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s
```
daniel-goldstein, line 418:
```suggestion
WHERE batch_id = %s AND state = 'Ready' AND always_run = 1 AND inst_coll = %s
```

I think this was most of the way reverted but might as well take it 100% so this file doesn't show up in the PR

daniel-goldstein: Same for the below query

batch/batch/front_end/front_end.py
```diff
@@ -1353,12 +1354,13 @@ async def update(tx: Transaction):
         if record:
             return record['update_id']
 
-        now = time_msecs()
         # We use FOR UPDATE so that we serialize batch update insertions
-        # This is necessary to reserve job id ranges
+        # This is necessary to reserve job id ranges.
+        # We don't allow updates to batches that have been cancelled
+        # but do allow updates to batches with jobs that have been cancelled.
```
daniel-goldstein, line 1363: 👍

batch/test/test_batch.py
```diff
@@ -1128,8 +1121,88 @@ def test_update_batch_w_deps_in_update(client: BatchClient):
...
+    bb.submit()
+
+    j.wait()
+
+    j1 = bb.create_job(DOCKER_ROOT_IMAGE, ['true'], always_run=True)
+    j2 = bb.create_job(DOCKER_ROOT_IMAGE, ['true'], parents=[j, j1])
+    b = bb.submit()
+    status = j2.wait()
+
+    assert status['state'] == 'Success', str((status, b.debug_info()))
```
daniel-goldstein, line 1140: Shouldn't `j1.wait()['state'] == 'Success'` and `j2.wait()['state'] == 'Cancelled'`? Can we group the call to wait and the assertion together so it's clear which job state we're looking at? I like how the test below this one is organized

batch/test/test_batch.py
```diff
@@ -1128,8 +1121,88 @@ def test_update_batch_w_deps_in_update(client: BatchClient):
...
+    j2 = bb.create_job(DOCKER_ROOT_IMAGE, ['true'], parents=[j, j1])
+    b = bb.submit()
+    j2_status = j2.wait()
+
+    assert j.status()['state'] == 'Success', str((j.status(), b.debug_info()))
+    assert j2_status['state'] == 'Cancelled', str((j2_status, b.debug_info()))
+    assert b.status()['state'] == 'failure', str((b.status(), b.debug_info()))
+
+
+def test_update_with_always_run(client: BatchClient):
```
daniel-goldstein, line 1159: I think the name sells the test a little short here as it's also testing that a job will wait until a parent in another update completes. Not sure if there's a succinct way to express that also in the name here.

batch/test/test_batch.py
```diff
@@ -1128,8 +1121,88 @@ def test_update_batch_w_deps_in_update(client: BatchClient):
...
+            break
+        delay = await sync_sleep_and_backoff(delay)
+
+    assert j2.is_pending(), str(j2.status(), b.debug_info())
+
+    b.cancel()
+    j2.wait()
+
+    assert j2.status()['state'] == 'Success', str((j2.status(), b.debug_info()))
+    assert b.status()['state'] == 'cancelled', str((b.status(), b.debug_info()))
```
daniel-goldstein, line 1181: This breaks the abstraction a bit here, so we would have to reach into the underlying aioclient methods, but I think it's worth it to have the following test as well:

1. Submit an hour long sleep job, `j1`
2. Create an update and submit a `true` job, `j2` that depends on `j1`, but don't commit the update yet
3. Cancel `j1`
4. Verify that `j2` is `Pending`
5. Commit the update
6. Verify that `j2` is `Cancelled`

jigold:
I put the time_updated field in. I'll test it with dev deploy once everything else is ready to go.

daniel-goldstein (changes requested):
Ya this looks super close! Just had a couple more ideas about testing

daniel-goldstein (changes requested):

batch/batch/front_end/front_end.py
```diff
@@ -611,12 +611,15 @@ async def _query_batches(request, user, q):
     batches_n_jobs_in_complete_states.n_completed,
     batches_n_jobs_in_complete_states.n_succeeded,
     batches_n_jobs_in_complete_states.n_failed,
-    batches_n_jobs_in_complete_states.n_cancelled
+    batches_n_jobs_in_complete_states.n_cancelled,
+    MAX(batch_updates.time_committed) AS time_updated
```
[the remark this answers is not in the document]
jigold: I like the first suggestion a lot.

batch/test/test_batch.py
jigold: Unfortunately, I can't write this test for two reasons: 
1. we can't cancel a specific job -- only a batch.
2. we can't query a job that hasn't been committed 

FYI - I had to modify more queries to get the desired behavior.

batch/batch/front_end/front_end.py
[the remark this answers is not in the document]
jigold: Offline we decided that adding time_updated right now is too complex and unnecessary.

jigold:
Stacked on #12208

batch/test/test_batch.py
daniel-goldstein: All good, I think the test that submits an update with a job dependent on one that has already failed is good enough to test that on commit we will cancel a job with failed parents.

batch/batch/front_end/front_end.py
```diff
@@ -1551,6 +1545,22 @@ async def commit_update(request: web.Request, userdata):
...
+FROM batches
+WHERE id = %s AND user = %s AND NOT deleted
+FOR UPDATE;
+''',
+        (batch_id, user),
+    )
+    if not record:
+        raise web.HTTPNotFound()
+    if record['cancelled']:
+        raise web.HTTPBadRequest(reason='Cannot commit and update to a cancelled batch')
```
daniel-goldstein, line 1562: Can you instead of adding this query also grab `cancelled` in the query of the `commit_update` function (the one that is actually the REST endpoint)? This function is used in multiple places including places like `create-fast` where this is redundant. I'd rather consolidate authorization checks like these as close to the endpoint as possible so these internal reused functions don't pile on redundant checks to the lifetime of a request.

daniel-goldstein, line 1562:
```suggestion
        raise web.HTTPBadRequest(reason='Cannot commit an update to a cancelled batch')
```

daniel-goldstein (changes requested):

jigold:
Note to self to check the logs for error messages before merging!

jigold:
I changed that the `_update_token` is no longer cached. I thought it was the source of a bug, but it was something else that was an issue. However, I felt it was confusing and I didn't see what value it provided as we have retries on all of our client operations that would need the token and the token is not used in future operations to submit the update.

daniel-goldstein:
duh you're totally right, I don't know why I thought retrying was happening higher up and not further down in the http client.

daniel-goldstein:
Sorry this was a lot more broken than I thought. I didn't remember everything I stripped down for the previous PRs and didn't add back in. The commit update and update-fast endpoints need to return the `start_job_id`. Hopefully that's the last of these issues

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -596,15 +654,25 @@ def _batch_spec(self):
...
+        return {'n_jobs': len(self._jobs), 'token': update_token}
+
+    async def _create_update(self, batch_id: int) -> int:
+        update_spec = self._update_spec()
+        update_json = await (await self._client._post(f'/api/v1alpha/batches/{batch_id}/updates/create', json=update_spec)).json()
+        return int(update_json['update_id'])
+
+    async def _commit_update(self, batch_id: int, update_id: int) -> int:
+        commit_json = await (await self._client._patch(f'/api/v1alpha/batches/{batch_id}/updates/{update_id}/commit')).json()
+        return int(commit_json['start_job_id'])
```
daniel-goldstein, line 675: I originally had the commit return the `start_job_id` because it's not actually needed by the client until the update is committed. It might have been premature optimization on my part hiding that information from the client until the update is committed, and it is probably clearer to provide it in the create update endpoint along with the update id, so I'm fine with your version. But this is probably the cause of some of the errors that it's currently hitting.

daniel-goldstein (approved):

PR #12202, opened 2022-09-19
[batch] Get rid of default standing workers in dev

jigold:


batch/sql/no_dev_standing_workers_by_default.py
```diff
@@ -0,0 +1,21 @@
...
+    if os.environ['HAIL_SCOPE'] != 'dev':
+        return
+
+    db = Database()
+    await db.async_init()
+
+    await db.execute_update(
+        '''
+UPDATE pools
+SET enable_standing_worker = %s
```
daniel-goldstein, line 16: Nit: I think this can just be `FALSE` instead of using query args

daniel-goldstein (dismissed):
Needs a line adding this script to the migrations in build.yaml

daniel-goldstein (approved):

PR #12204, opened 2022-09-19
[docker] Add script for mirroring dockerhub images

daniel-goldstein:
This consolidates the three types of images we publish to DockerHub: hail, genetics, and python-dill under `docker/hailgenetics` and provides a script to copy all of the python-dill images as well as the hail and genetics images for a particular pip version to a specified registry.

This also fixes a small bootstrap bug where we never mirrored any of the hailgenetics images and the python-dill images were getting pushed to our private registry under an incorrect name.


daniel-goldstein:
I originally added a `build.yaml` step to this that ran the script on every deploy, but I think there are some subtleties in there around releases that are best discussed in follow-up PR and may distract from some immediate goals around getting AR up and working.

PR #12212, opened 2022-09-21
[batch] Use a simulated job queue to estimate the ready cores in the control loop

jigold:
I added a new Grafana panel without alerts that hopefully will let us catch problems if jobs aren't getting scheduled in a timely manner. I think to have an alert, we'd want to measure what the average wait time of a job in the queue is which would require more infrastructure (keeping track of last state change). We can consider adding that now -- not sure how much work it would be.


batch/batch/driver/instance_collection/pool.py
```diff
@@ -34,6 +35,46 @@
...
+    'Number of cores attempted to schedule in one scheduling execution loop',
+    ['pool_name'],
+)
+
+NO_QUEUE_READY_CORES = pc.Gauge(
+    'control_loop_ready_cores_no_queue', 'Number of ready cores per control loop run without a job queue', ['pool_name']
+)
+JOB_QUEUE_READY_CORES = pc.Gauge(
+    'control_loop_ready_cores_job_queue', 'Number of ready cores per control loop run with a job queue', ['pool_name']
+)
```
daniel-goldstein, line 64: Can we use different terminology here? In my mind we're already using a job queue: the list of Ready jobs in the database, the difference here being that we currently look at the *entire* job queue but the new change looks at the *head* of the job queue

batch/batch/driver/instance_collection/pool.py
```diff
@@ -34,6 +35,46 @@
...
+
+TOTAL_CORES_SCHEDULED = pc.Gauge(
+    'total_cores_scheduled', 'Number of cores scheduled in one scheduling execution loop', ['pool_name']
+)
+TOTAL_CORES_ATTEMPTED_TO_SCHEDULE = pc.Gauge(
+    'total_cores_attempted_to_schedule',
+    'Number of cores attempted to schedule in one scheduling execution loop',
+    ['pool_name'],
+)
```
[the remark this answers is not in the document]
daniel-goldstein: I also think `total` here is a little redundant

batch/batch/driver/instance_collection/pool.py
```diff
@@ -34,6 +35,46 @@
...
+    'Number of scheduling loop executions per pool that did not use all of the available free cores',
+    ['pool_name'],
+)
+
+CONTROL_LOOP_RUNS = pc.Counter(
+    'control_loop_runs',
+    'Number of control loop executions per pool',
+    ['pool_name'],
+)
+
```
daniel-goldstein, line 49: There's an idiom with prometheus metrics to standardize a set of common prefixes (like I use `http` and `sql`). I might have messed this up on some other queries, but can we use `scheduler_` and `autoscaler_` prefixes for these metrics?

batch/batch/driver/instance_collection/pool.py
```diff
@@ -196,8 +237,8 @@ async def create_instance(
             boot_disk_size_gb=self.boot_disk_size_gb,
         )
 
-    async def create_instances_from_ready_cores(self, ready_cores_mcpu, region=None):
-        n_live_instances = self.n_instances_by_state['pending'] + self.n_instances_by_state['active']
+    def compute_n_instances_needed(self, ready_cores_mcpu, n_instances_by_state, region=None):
```
daniel-goldstein, line 240: Can we add types to this signature

batch/batch/driver/instance_collection/pool.py
```diff
@@ -227,15 +270,65 @@ async def create_instances_from_ready_cores(self, ready_cores_mcpu, region=None)
...
+            ready_cores_mcpu,
+            self.n_instances_by_state,
+            region,
+        )
+        await self._create_instances(instances_needed, region)
+        return instances_needed
+
+    async def ready_cores_mcpu_from_estimated_job_queue(self):
+        n_free_cores = int(
+            10 * 8 * self.worker_cores
```
daniel-goldstein, line 288: Let's make `10` (and its usage above in `compute_n_instances_needed`) a named constant

daniel-goldstein: Also, I either don't understand what we're computing here or I don't think it's correct. If we're trying to compute the upper bound on how many cores we would have available to schedule on in a few minutes, then we can write:

```python
autoscaler_runs_per_minute = 60 / AUTOSCALER_LOOP_PERIOD_SECONDS
max_new_cores_in_two_and_a_half_minutes = int(2.5 * MAX_INSTANCES_PER_AUTOSCALER_LOOP * autoscaler_runs_per_minute)

max_possible_future_available_cores = self.worker_cores + max_new_cores_in_two_and_a_half_minutes
```

I would rather use a name like `max_possible_future_available_cores` than `n_free_cores`

batch/batch/driver/instance_collection/pool.py
```diff
@@ -493,4 +623,11 @@ async def schedule_with_error_handling(app, record, instance):
         if n_scheduled > 0:
             log.info(f'schedule: attempted to schedule {n_scheduled} jobs in {end - start}ms for {self.pool}')
 
+        if n_scheduled != n_attempted_to_schedule:
+            SCHEDULING_LOOP_RUNS_UNUSED_FREE_CORES.labels(pool_name=self.pool.name).inc()
+
+        pool_labels = PoolLabels(self.pool.name)
+        set_gauge_value(TOTAL_CORES_SCHEDULED, {pool_labels: n_cores_mcpu_scheduled / 1000})
+        set_gauge_value(TOTAL_CORES_ATTEMPTED_TO_SCHEDULE, {pool_labels: n_cores_mcpu_attempted_to_schedule / 1000})
```
daniel-goldstein, line 631: I think I find this function more confusing than it is worth. Can we just do:

```python
TOTAL_CORES_SCHEDULED.labels(pool_name=self.pool.name).set(n_cores_mcpu_scheduled / 1000)
```

I find this reads much more clearly. There's also no reason to clear these gauges before setting them

batch/batch/driver/instance_collection/pool.py
```diff
@@ -252,7 +345,30 @@ async def create_instances(self):
...
+        estimated_ready_cores_mcpu_from_job_queue = await self.ready_cores_mcpu_from_estimated_job_queue()
+        n_instances_with_job_queue = self.compute_n_instances_needed(
+            estimated_ready_cores_mcpu_from_job_queue, self.n_instances_by_state
+        )
+
+        pool_label = PoolLabels(self.name)
+        set_gauge_value(NO_QUEUE_READY_CORES, {pool_label: ready_cores_mcpu / 1000})
+        set_gauge_value(JOB_QUEUE_READY_CORES, {pool_label: estimated_ready_cores_mcpu_from_job_queue / 1000})
+        set_gauge_value(N_INSTANCES_NO_QUEUE, {pool_label: n_instances_no_queue})
+        set_gauge_value(N_INSTANCES_WITH_JOB_QUEUE, {pool_label: n_instances_with_job_queue})
```
daniel-goldstein, line 371: Same comment as below about not using `set_gauge_value`

daniel-goldstein (dismissed):
This looks pretty good, it does seem like a small departure from our current implementation that gets us most of the way to basic region-based scaling. Comments I made are pretty minor, I think it all comes down to how that query performs.

I'm a little nervous about the length of the query string scaling in O(users), but there's many parts of this scheduler/scaler that aren't quite suited to many users anyway. We'll have to address those all separately down the road.

jigold:
I changed the control loop to use the old method of estimating ready cores for now. I'll change it to the new way in a second PR. This way we can make sure we're happy with the metrics in Grafana before I change the behavior.

jigold:
This is ready for a look. The tests are passing except one service backend test timed out.

batch/batch/driver/instance_collection/pool.py
```diff
@@ -227,15 +274,88 @@ async def create_instances_from_ready_cores(self, ready_cores_mcpu, region=None)
...
-            log.info(f'not creating instances for {self}; batch is frozen')
-            return
+    async def create_instances_from_ready_cores(self, ready_cores_mcpu, region=None):
+        instances_needed = self.compute_n_instances_needed(
+            ready_cores_mcpu,
+            self.n_instances_by_state,
+            region,
+        )
+        await self._create_instances(instances_needed, region)
+        return instances_needed
```
daniel-goldstein, line 288: I think this doesn't need to return anything anymore, right?

batch/batch/driver/instance_collection/pool.py
```diff
@@ -35,6 +35,50 @@
     ['pool_name'],
 )
 
+CONTROL_LOOP_RUNS = pc.Counter(
```
daniel-goldstein, line 38: Nit: match the name of the constant to the name of the query

batch/test/test_batch.py
```diff
@@ -1291,3 +1291,16 @@ def test_submit_update_to_deleted_batch(client: BatchClient):
...
+    dev_client = BatchClient('test', token_file=os.environ['HAIL_TEST_DEV_TOKEN_FILE'])
+
+    try:
+        bb = dev_client.create_batch()
+        j = bb.create_job(DOCKER_ROOT_IMAGE, ['echo', 'test'])
+        b = bb.submit()
+        status = b.wait()
+        assert status['state'] == 'success', str((status, b.debug_info()))
+    finally:
+        dev_client.close()
```
daniel-goldstein, line 1306: Should this test have another batch that is running at the same time by a different user as well?

daniel-goldstein (changes requested):

daniel-goldstein (approved):

PR #12218, opened 2022-09-23
[query] Support missing values in primitive columns in `Table.to_pandas`

tpoterba:
CHANGELOG: Support missing values in primitive columns in `Table.to_pandas`.


danking (approved):

danking:
It looks like there's a few tests whose expected values are using the old dtypes.

PR #12219, opened 2022-09-23
[gateway] use Mozilla recommended TLS intermediate settings

danking:
I think we should do this with the Envoy gateway as well. If Envoy is gonna land this week, then we can just close this and fold in over there instead.


danking:
We already do this 😆 see letsencrypt.sh

PR #12222, opened 2022-09-23
[qob] Add capability for worker jobs to go in the same batch as the driver job

daniel-goldstein:
This leverages open batches to submit all QoB stages as updates to the same batch the Query Driver is running in. Most of this implementation feels uncontroversial, but there is a backwards-incompatible change to the JVMJob that I don't love. I needed to let the Query Driver know which batch it's running in and did so by adding that as another argument to main. I initially wanted this as an environment variable, but considering that these containers are long-lived I wasn't quite sure how to do that.

I didn't stack this on open batches so this won't work until that goes in but the changes are entirely disjoint

### DIFF
--- a/hail/python/hail/backend/service_backend.py
+++ b/hail/python/hail/backend/service_backend.py
@@ -366,6 +366,7 @@ class ServiceBackend(Backend):
                     ],
                     mount_tokens=True,
                     resources=resources,
+                    attributes={'name': 'driver'},
                 )
                 b = await bb.submit(disable_progress_bar=True)
 
@@ -382,7 +383,7 @@ class ServiceBackend(Backend):
                     raise
 
             with timings.step("parse status"):
-                if status['n_succeeded'] != 1:
+                if status['n_succeeded'] != status['n_jobs']:
                     job_status = await j.status()
                     if 'status' in job_status:
                         if 'error' in job_status['status']:
--- a/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
@@ -63,11 +63,14 @@ class ServiceBackend(
   val jarLocation: String,
   var name: String,
   val theHailClassLoader: HailClassLoader,
-  val scratchDir: String = sys.env.get("HAIL_WORKER_SCRATCH_DIR").getOrElse("")
+  val batchClient: BatchClient,
+  val curBatchId: Option[Long],
+  val scratchDir: String = sys.env.get("HAIL_WORKER_SCRATCH_DIR").getOrElse(""),
 ) extends Backend {
   import ServiceBackend.log
 
-  private[this] var batchCount = 0
+  private[this] var stageCount = 0
+  private[this] var totalNumWorkerJobs: Int = 0
   private[this] implicit val ec = scalaConcurrent.ExecutionContext.fromExecutorService(
     Executors.newCachedThreadPool())
   private[this] val MAX_AVAILABLE_GCS_CONNECTIONS = 100
@@ -152,7 +155,6 @@ class ServiceBackend(
     scalaConcurrent.Await.result(uploadFunction, scalaConcurrent.duration.Duration.Inf)
     scalaConcurrent.Await.result(uploadContexts, scalaConcurrent.duration.Duration.Inf)
 
-    val batchClient = BatchClient.fromSessionID(backendContext.sessionID)
     val jobs = new Array[JObject](n)
     var i = 0
     while (i < n) {
@@ -166,7 +168,7 @@ class ServiceBackend(
       jobs(i) = JObject(
         "always_run" -> JBool(false),
         "job_id" -> JInt(i + 1),
-        "parent_ids" -> JArray(List()),
+        "in_update_parent_ids" -> JArray(List()),
         "process" -> JObject(
           "jar_spec" -> JObject(
             "type" -> JString("jar_url"),
@@ -178,6 +180,9 @@ class ServiceBackend(
             JString(s"$i"),
             JString(s"$n"))),
           "type" -> JString("jvm")),
+        "attributes" -> JObject(
+          "name" -> JString(name + "_" + stageCount + "_" + i),
+        ),
         "mount_tokens" -> JBool(true),
         "resources" -> resources,
       )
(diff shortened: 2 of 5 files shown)


jigold:
@danking What do you think about having a version ID inside the JAR file (MANIFEST???). We already download the JAR file on the worker. Not sure how much extra time it would be to look for the version inside the JAR (maybe cache this?) and then pass the write argument configuration.

danking:
Let me think about this further.

hail/python/hail/backend/service_backend.py
```diff
@@ -382,7 +383,7 @@ async def _rpc(self,
                     raise
 
             with timings.step("parse status"):
-                if status['n_succeeded'] != 1:
+                if status['n_succeeded'] == 0:
```
jigold, line 386: Can you explain this change to me? I would have thought it should have been `if status['n_failed'] > 0`

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -186,21 +191,15 @@ class ServiceBackend(
...
+    val updateId = batchClient.update(batchId, token, jobs)
+    // The driver is in this batch too
+    val batch = batchClient.waitForBatch(batchId, totalNumWorkerJobs)
+    updateCount += 1
     implicit val formats: Formats = DefaultFormats
-    val batchID = (batch \ "id").extract[Int]
     val batchState = (batch \ "state").extract[String]
-    if (batchState != "success") {
-      throw new HailBatchFailure(s"$batchID")
+    if (batchState == "failed") {
```
jigold, line 201: Shouldn't we handle `"cancelled"` as well? Let's also keep the batch id in the message.

hail/src/main/scala/is/hail/services/batch_client/BatchClient.scala
```diff
@@ -159,14 +166,21 @@ class BatchClient(
...
+  def waitForBatch(batchID: Long, nJobsToComplete: Int): JValue = {
     implicit val formats: Formats = DefaultFormats
 
     val start = System.nanoTime()
 
     while (true) {
       val batch = retryTransientErrors { get(s"/api/v1alpha/batches/$batchID") }
-      if ((batch \ "complete").extract[Boolean])
+      val n_completed = (batch \ "n_completed").extract[Int]
+      // Remember the driver is itself in the batch
```
jigold, line 182: Can we add to this message `The caller has accounted for the extra driver job.`

batch/batch/worker/worker.py
```diff
@@ -2218,7 +2220,10 @@ async def kill(self):
         if self.container is not None:
             await self.container.remove()
 
-    async def execute(self, classpath: str, scratch_dir: str, log_file: str, jar_url: str, argv: List[str]):
+    # TODO This is not backwards compatible
```
jigold, line 2223: How hard is it to make a new release with the client changes first that accept an optional batch id argument? If the batch id is not there, then revert back to the old submission method. Otherwise, use the new submission method. I just want to make sure we don't break anyone's code before a release is made. Regardless, we'll need to make a new release ASAP after this merges.

jigold (changes requested):

danking:
I'd like to avoid the backwards incompatibility here if possible, but if we're going to do it, we should make sure future changes can be done backwards compatibility.

daniel-goldstein:
I'll make this backwards compatible. Will probably need a tiny PR before this one to remove restrictions on the ServiceBackend

hail/python/hail/backend/service_backend.py
daniel-goldstein: you're right, will change this

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
daniel-goldstein: If the batch is `cancelled`, the job that's running this code would get cancelled as well, in which case it's not guaranteed that we'd even reach that point in the code. The python client controlling this QoB should be the one to pick up on the cancellation. AFAIK, the only states that the Query Driver can reliably observe here are failed and running.

batch/batch/worker/worker.py
daniel-goldstein: We can instead sidestep all of this and do this with no incompatible changes. What do you think about adding an optional parameter to the deploy config that is the batch the job is currently running in? That would work both for JVMJobs and DockerJobs, and no worries about compatibility or reverting

jigold: Can you point me to where the JVM jobs deploy config is created?

daniel-goldstein: Looks like it's created for all jobs here https://github.com/hail-is/hail/blob/7e644d28b53c59d5339073576257f0fb27bdef4f/batch/batch/front_end/front_end.py#L943. The deploy config is currently the same for every job so we just have it stored in the secret `worker-deploy-config`, but it doesn't seem unreasonable to craft a deploy-config that is job-specific.

jigold: I like this train of thought. Why don't we write a new file like `batch-config` into the job's container disk space where the log file goes already and the new code can read from this file to get the batch id?

daniel-goldstein:
Made this change backwards compatible. Note that I have not made any changes to worker.py in this PR anymore, so there's no danger of incompatibility. I tested the JAR from this PR against default and ran a simple hail query to see that it behaved as usual.

Separately, I made #12246, dev deployed it, then ran this same JAR against my dev namespace to see that it added all worker jobs to the same batch as the driver job.

hail/python/hail/backend/service_backend.py
```diff
@@ -366,6 +366,7 @@ async def _rpc(self,
                     ],
                     mount_tokens=True,
                     resources=resources,
+                    attributes={'name': 'driver'},
```
danking, line 369: Take a look at line 350, there's some name logic already included. I think you should add your changes there.

danking (changes requested):

hail/python/hail/backend/service_backend.py
daniel-goldstein: Do you want the job name to be $batch_name + 'driver'? 350 sets the batch attributes whereas here I'm just naming the driver job

jigold (approved):

PR #12226, opened 2022-09-26
[dependabot] also monitor hail/build.gradle

danking:
Curious to hear your reaction to this, Daniel, as you've effectively become the dependency czar. For this, we might need to pull in @chrisvittal or @tpoterba to resolve dependency issues.


chrisvittal:
I'm all for trying to keep things up to date.

daniel-goldstein (approved):
Happy to try this out and see how it goes. I wonder what the success rate here will be given the various relocating we've had to do for simple updates in the past.

PR #12228, opened 2022-09-26
[batch] serialize user-defined python functions once per batch

iris-garden:
Currently, for each function call that a Python job has, it serializes the entire function definition to a file, to be deserialized and called when the job is run. This change moves the serialization to the batch level, so if the same job calls the same function multiple times, or if multiple jobs within a single batch all call the same function, it will only have its definition serialized once.

### DIFF
--- a/hail/python/hailtop/batch/batch.py
+++ b/hail/python/hailtop/batch/batch.py
@@ -1,7 +1,9 @@
 import os
 import warnings
 import re
-from typing import Optional, Dict, Union, List, Any, Set
+from typing import Callable, Optional, Dict, Union, List, Any, Set
+from io import BytesIO
+import dill
 
 from hailtop.utils import secret_alnum_string, url_scheme, async_to_blocking
 from hailtop.aiotools import AsyncFS
@@ -148,6 +150,40 @@ class Batch:
         self._DEPRECATED_fs: Optional[RouterAsyncFS] = None
 
         self._cancel_after_n_failures = cancel_after_n_failures
+
+        self._python_function_defs: Dict[int, Callable] = {}
+        self._python_function_files: Dict[int, _resource.InputResourceFile] = {}
+
+    def _register_python_function(self, function: Callable) -> int:
+        function_id = id(function)
+        self._python_function_defs[function_id] = function
+        return function_id
+
+    async def _serialize_python_to_input_file(
+        self, path: str, subdir: str, file_id: int, code: Any, dry_run: bool = False
+    ) -> _resource.InputResourceFile:
+        pipe = BytesIO()
+        dill.dump(code, pipe, recurse=True)
+        pipe.seek(0)
+
+        code_path = f"{path}/{subdir}/code{file_id}.p"
+
+        if not dry_run:
+            await self._fs.makedirs(os.path.dirname(code_path), exist_ok=True)
+            await self._fs.write(code_path, pipe.getvalue())
+
+        code_input_file = self.read_input(code_path)
+
+        return code_input_file
+
+    async def _serialize_python_functions_to_input_files(
+        self, path: str, dry_run: bool = False
+    ) -> None:
+        for function_id, function in self._python_function_defs.items():
+            file = await self._serialize_python_to_input_file(
+                path, "functions", function_id, function, dry_run
+            )
+            self._python_function_files[function_id] = file
 
     def _unique_job_token(self, n=5):
         token = secret_alnum_string(n)
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -1,15 +1,11 @@
 import asyncio
-import functools
 import inspect
 import os
 import re
 import textwrap
 import warnings
-from io import BytesIO
 from shlex import quote as shq
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union, cast
-
-import dill
 
 from . import backend, batch  # pylint: disable=cyclic-import
 from . import resource as _resource  # pylint: disable=cyclic-import
@@ -828,7 +824,7 @@ class PythonJob(Job):
         super().__init__(batch, token, name=name, attributes=attributes, shell=None)
         self._resources: Dict[str, _resource.Resource] = {}
         self._resources_inverse: Dict[_resource.Resource, str] = {}
-        self._functions: List[Tuple[_resource.PythonResult, Callable, Tuple[Any, ...], Dict[str, Any]]] = []
+        self._function_calls: List[Tuple[_resource.PythonResult, int, Tuple[Any, ...], Dict[str, Any]]] = []
         self.n_results = 0
 
     def _get_resource(self, item: str) -> '_resource.PythonResult':
@@ -1030,7 +1026,9 @@ class PythonJob(Job):
         result = self._get_resource(f'result{self.n_results}')
         handle_arg(result)
 
-        self._functions.append((result, unapplied, args, kwargs))
+        unapplied_id = self._batch._register_python_function(unapplied)
+
+        self._function_calls.append((result, unapplied_id, args, kwargs))
 
         return result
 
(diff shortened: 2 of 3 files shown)


hail/python/hailtop/batch/batch.py
```diff
@@ -149,6 +151,58 @@ def __init__(self,
...
+            )
+        return function
+
+    def _get_python_function_file_name(self, function_id: int) -> str:
+        _, file_name = self._python_functions.get(function_id, (None, None))
+        if file_name is None:
+            raise ValueError(
+                f"No file found for Python function registered with id: {function_id}."
+            )
+        return file_name
```
daniel-goldstein, line 170: This single dictionary feels like it's doing a lot of different things, which leads to each of those things being a little clunky. More specifically, I don't think that we ever need both of the pieces inside of the dictionary value here at the same time, so putting them into a tuple together is a bit confusing. I think the two tasks that we're trying to accomplish here are:

- Keep track of a set of function ids that have been accounted for so far
- Once we're ready to serialize all the functions, do so and keep a map of id -> filepath.

We *could* merge these into one Map[int, Optional[String]], but I personally would prefer to have two separate fields since they don't really have much to do with one another and once we write the input files `None` really shouldn't be a valid value in the map. This should make these register and get functions much simpler.

hail/python/hailtop/batch/job.py
```diff
@@ -1045,60 +1043,25 @@ def prepare_argument_for_serialization(arg):
...
-            @functools.wraps(f)
-            def wrapped(*args, **kwargs):
-                args = [deserialize_argument(arg) for arg in args]
-                kwargs = {kw: deserialize_argument(arg) for kw, arg in kwargs.items()}
-                return f(*args, **kwargs)
-            return wrapped
-
-        for i, (result, unapplied, args, kwargs) in enumerate(self._functions):
+        for i, (result, unapplied_id, args, kwargs) in enumerate(self._function_calls):
+            unapplied = self._batch._get_python_function(unapplied_id)
```
daniel-goldstein, line 1047: I don't think this is used, can we delete this?

hail/python/hailtop/batch/job.py
```diff
@@ -1107,15 +1070,31 @@ def wrapped(*args, **kwargs):
 import traceback
 import json
 import sys
-
-with open(\\"{result}\\", \\"wb\\") as dill_out:
+from time import perf_counter
```
daniel-goldstein, line 1073: Guessing this was for testing. Can remove this right?

hail/python/hailtop/batch/job.py
```diff
@@ -1107,15 +1070,31 @@ def wrapped(*args, **kwargs):
...
+
+with open('{result}', 'wb') as dill_out:
     try:
-        with open(\\"{code}\\", \\"rb\\") as f:
-            result = dill.load(f)()
-            dill.dump(result, dill_out, recurse=True)
-            {json_write}
-            {str_write}
-            {repr_write}
+        func = args = kwargs = None
```
daniel-goldstein, line 1086: These don't need to be set outside the context-manager

hail/python/hailtop/batch/batch.py
```diff
@@ -149,6 +151,58 @@ def __init__(self,
...
+
+        code_path = f"{path}/code{file_id}.p"
+
+        if not dry_run:
+            await self._fs.makedirs(os.path.dirname(code_path), exist_ok=True)
+            await self._fs.write(code_path, pipe.getvalue())
+
+        code_input_file = self.read_input(code_path)
+
+        return str(code_input_file)
```
daniel-goldstein, line 195: How come the cast this to a `str` here? The return of `read_input` should be fine to interpolate, and it would be great to preserve more expressive types where possible.

hail/python/hailtop/batch/batch.py
```diff
@@ -149,6 +151,58 @@ def __init__(self,
...
+        return function_id
+
+    async def _serialize_python_to_input_file(
+        self, path: str, file_id: int, code: Any, dry_run: bool = False
+    ) -> str:
+        pipe = BytesIO()
+        dill.dump(code, pipe, recurse=True)
+        pipe.seek(0)
+
+        code_path = f"{path}/code{file_id}.p"
```
daniel-goldstein, line 187: It would be nice for debugging if the functions were stored more "separately" from the args. It's all blobs in the end but it would be nice to have something like `{path}/functions/` and `{path}/args/`

daniel-goldstein (dismissed):
Awesome performance improvement, thanks for the detailed example! I've made some comments but they are largely organizational / style, this looks functionally great.

Also, sorry I took so long to review this! Been one hell of a week.

hail/python/hailtop/batch/batch.py
iris-garden: Refactored into two fields on the class!

hail/python/hailtop/batch/job.py
iris-garden: It's still used to generate the `self._user_code` below, which I'm under the impression is used for debugging, but we can move the definition down to where it's actually used at least.

hail/python/hailtop/batch/job.py
iris-garden: Removed!

hail/python/hailtop/batch/job.py
iris-garden: Removed!

hail/python/hailtop/batch/batch.py
iris-garden: Fixed to preserve the `InputResourceFile` until it is interpolated into the f-string for the command!

hail/python/hailtop/batch/batch.py
```diff
@@ -149,6 +151,64 @@ def __init__(self,
...
+                f"No Python function registered with id: {function_id}."
+            ) from error
+
+    def _register_python_function(self, function: Callable) -> int:
+        function_id = id(function)
+        try:
+            self._get_python_function(function_id)
+        except ValueError:
+            self._python_function_defs[function_id] = function
+        return function_id
```
daniel-goldstein, line 184: Let's remove the try-except and just run the line that sets the value in the dictionary, shouldn't matter if it's already in there

hail/python/hailtop/batch/batch.py
```diff
@@ -149,6 +151,64 @@ def __init__(self,
 
         self._cancel_after_n_failures = cancel_after_n_failures
 
+        self._python_function_defs: Dict[int, Callable] = {}
+        self._python_function_files: Dict[int, Optional[_resource.InputResourceFile]] = {}
```
daniel-goldstein, line 155: Is `None` ever a valid value in this dictionary? From how it's used below I think if a key exists in the dictionary it'll point to a valid `InputResourceFile`

hail/python/hailtop/batch/batch.py
```diff
@@ -149,6 +151,64 @@ def __init__(self,
...
+            file = self._python_function_files[function_id]
+            if file is None:
+                raise ValueError(
+                    f"No file found for Python function registered with id: {function_id}."
+                )
+            return file
+        except KeyError as error:
+            raise ValueError(
+                f"No Python function registered with id: {function_id}."
+            ) from error
```
daniel-goldstein, line 176: Both this method and the field it protects are private and this PR creates an invariant where we'll only make valid accesses into this `_python_function_files` dictionary (and if we didn't, the underlying `KeyError` would be informative enough). So I think we don't need this function and can just access the dictionary directly where this is used.

daniel-goldstein (dismissed):
Just a few more nits.

hail/python/hailtop/batch/batch.py
iris-garden: Updated!

hail/python/hailtop/batch/batch.py
iris-garden: Removed!

daniel-goldstein (approved):
Woot, this looks awesome and ended up being a surprisingly small change!

PR #12230, opened 2022-09-27
[batch] All public images must be under DOCKER_PREFIX/hailgenetics

daniel-goldstein:
We should just treat `python-dill` like other images in the `hailgenetics` DockerHub repo and not hard-code our own registry into the docs

### DIFF
--- a/batch/batch/publicly_available_images.py
+++ b/batch/batch/publicly_available_images.py
@@ -2,9 +2,4 @@ from typing import List
 
 
 def publicly_available_images(docker_prefix: str) -> List[str]:
-    # the worker cannot import batch_configuration because it does not have all the environment
-    # variables
-    return [
-        f'{docker_prefix}/{name}'
-        for name in ('hailgenetics/hail', 'hailgenetics/genetics', 'python-dill', 'batch-worker')
-    ]
+    return [f'{docker_prefix}/hailgenetics/{name}' for name in ('hail', 'genetics', 'python-dill')]
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -795,7 +795,7 @@ class PythonJob(Job):
 
         # Create a batch object with a default Python image
 
-        b = Batch(default_python_image='gcr.io/hail-vdc/python-dill:3.7-slim')
+        b = Batch(default_python_image='hailgenetics/python-dill:3.7-slim')
 
         def multiply(x, y):
             return x * y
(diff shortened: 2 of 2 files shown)


batch/batch/publicly_available_images.py
```diff
@@ -2,9 +2,4 @@
 
 
 def publicly_available_images(docker_prefix: str) -> List[str]:
-    # the worker cannot import batch_configuration because it does not have all the environment
-    # variables
-    return [
-        f'{docker_prefix}/{name}'
-        for name in ('hailgenetics/hail', 'hailgenetics/genetics', 'python-dill', 'batch-worker')
-    ]
+    return [f'{docker_prefix}/hailgenetics/{name}' for name in ('hail', 'genetics', 'python-dill')]
```
jigold, line 5: We might need to special case the authentication / docker pull for the batch-worker image to make sure the JVM containers can use it.

jigold (changes requested):

batch/batch/publicly_available_images.py
daniel-goldstein: We already do https://github.com/hail-is/hail/blob/25bafaea4305b21da67f9997bd8b9951a08668b6/batch/batch/worker/worker.py#L432-L433

jigold: Sorry one last thing. Do we need to add backwards compatibility for `python-dill`?

jigold (changes requested):

batch/batch/publicly_available_images.py
daniel-goldstein: This depends on whether or not you think what was there is a bug. I think it's a bug that we exposed `gcr.io/hail-vdc` to users. Currently, specifying `j.image('hailgenetics/python-dill:3.7')` doesn't work, but specifying `j.image('gcr.io/hail-vdc/python-dill:3.7')` does. It is the only instance where a user can explicitly access an image in `gcr.io/hail-vdc`. It seems really hacky to add in a compatibility layer that specifically rewrites `gcr.io/hail-vdc/python-dill:3.7` to `{DOCKER_PREFIX}/hailgenetics/python-dill:3.7`, especially considering that we're moving off of `gcr.io/hail-vdc` entirely.

daniel-goldstein: I feel like an announcement of a breaking bug fix is more appropriate, but if you think this is too much breakage then the other option is to do the rewrite.

jigold: Let me double check with KC and the rest of our users to see how many people were actually using this.

daniel-goldstein: I imagine many were building their own images on top, in which case that is unaffected by this change

jigold (approved):

daniel-goldstein:
Not quite sure why this was re-marked WIP. Anyway, switching to AR meant users lost access to `gcr.io/hail-vdc/python-dill` anyway so this is PR is no longer the source of a breaking change and now a fix so that users can use `hailgenetics/python-dill`.

jigold:
I added the WIP because I thought from Zulip that it wasn't necessary to make the change for the artifact registry conversion and Dan thought it might be rude to do so, hence why I thought we might need to discuss it more. You can take the WIP tag off, but remind me to add this breaking change to the announcement to users about the changes for this week.

PR #12232, opened 2022-09-27
[ci] Add production prefix to tags deployed in production

daniel-goldstein:


jigold (approved):

jigold (changes requested):
Can we make this `prod`? There's no negative lookahead with Go Regex so I have to write something manually that is O(n_char).

jigold:
Actually. It will be easier if we do the opposite and tag the rest with "tmp" or something.

jigold:
I need a single GO regex that matches the tags that we are okay deleting

> If specified, any image where all tags match this given regular expression will be deleted. The image will not be delete if it has other tags that do not match the given regular expression. The regular expressions are parsed according to the [Go regexp package](https://golang.org/pkg/regexp/syntax/).

jigold:
I'm not sure if you can easily specify multiple groups.

jigold:
This also leaves me a bit nervous as it means we'll never delete production images.

daniel-goldstein:
> 

Does this not work? `((dev)|(pr)|(test))-.*`

daniel-goldstein:
> This also leaves me a bit nervous as it means we'll never delete production images.

We can still do this, no?

jigold:
Closing in favor of only deleting untagged images.

PR #12236, opened 2022-09-28
[query] Change indexed reads with filter_intervals=True to repartitio…

tpoterba:
…n back to the original partitioning

This means reading a 10-partition table with 100K intervals doesn't give you a 100k-partition table with tiny intervals.

Improved repartitionNoShuffle to have a fast path that drops partitions and produces no key comparisons.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
@@ -241,7 +241,7 @@ class TableStage(
       repartitionNoShuffle(partitioner.strictify)
   }
 
-  def repartitionNoShuffle(newPartitioner: RVDPartitioner, allowDuplication: Boolean = false): TableStage = {
+  def repartitionNoShuffle(newPartitioner: RVDPartitioner, allowDuplication: Boolean = false, allowPartitionFilter: Boolean = false): TableStage = {
     if (newPartitioner == this.partitioner) {
       return this
     }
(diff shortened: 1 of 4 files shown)


hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -241,7 +241,7 @@ class TableStage(
       repartitionNoShuffle(partitioner.strictify)
   }
 
-  def repartitionNoShuffle(newPartitioner: RVDPartitioner, allowDuplication: Boolean = false): TableStage = {
+  def repartitionNoShuffle(newPartitioner: RVDPartitioner, allowDuplication: Boolean = false, allowPartitionFilter: Boolean = false): TableStage = {
```
patrick-schultz, line 244: Can you explain the difference in behavior if `allowPartitionFilter` is set or not?

patrick-schultz: Is it: if true, the returned partitioner drops partitions of `newPartitioner` that would be empty, otherwise the returned partitioner is `newPartitioner`?

tpoterba: This allows the output to be partitioned to a subset of `newPartitioner`, with empty partitions dropped.

If `allowPartitionFilter` is `false`, the output will always be partitioned to `newPartitioner`, even with empty partitions.

patrick-schultz: Got it, thanks. It might be clearer to call it `dropEmptyPartitions`.

tpoterba: good change 👍

tpoterba: did the rename.

tpoterba:
@patrick-schultz made the changes, can you look again?

patrick-schultz (approved):

patrick-schultz:
Sorry! Fell off my radar.

PR #12245, opened 2022-09-29
[batch] Use copy step credentials when pulling images

jigold:


batch/sql/delete-batch-tables.sql
```diff
@@ -42,6 +42,7 @@ DROP TABLE IF EXISTS `batch_inst_coll_cancellable_resources`;
 DROP TABLE IF EXISTS `globals`;
 DROP TABLE IF EXISTS `attempts`;
 DROP TABLE IF EXISTS `batch_attributes`;
+DROP TABLE IF EXISTS `job_regions`;
```
daniel-goldstein, line 45: I think this commit got added to the wrong branch

daniel-goldstein (changes requested):
This JVMUserCredentials / CopyStepCredentials feels like a hack around a not-quite-right abstraction -- we use these exclusively for the batch worker image which we already know is installed so we shouldn't call pull in the first place. However, that refactor is beside the point of this fix. I'll put it on the list of things I wish to clean up.

Only change I see needed here is to take out the sql commit.

batch/batch/worker/worker.py
```diff
@@ -1987,6 +1987,12 @@ def __init__(self):
         self.password = None
 
 
+class CopyStepCredentials:
+    def __init__(self):
+        self.username = None
+        self.password = None
```
daniel-goldstein, line 1993: Actually, one thing to change. There's no point to these `username` and `password` being here, can we remove these?

daniel-goldstein: We should relatedly add an assert in `_current_user_access_token` that `self.credentials` is not only present, but that it is an instance of `CloudUserCredentials`

daniel-goldstein (dismissed):

daniel-goldstein (approved):

PR #12248, opened 2022-09-29
[batch] Allow finer granularity for JVM memory settings

lgruen:
This is useful for tuning resource consumption of large QoB jobs.

#assign services


jigold:
@danking Do you have an opinion on this change? I thought we didn't want to create too many JVMs on the worker.

danking:
@lgruen are y'all running with this change now? I was vaguely concerned that with ~32 JVMs alive that would negatively impact the machine. Do you find that the JVM's RAM gets swapped out and the machines are generally stable?

lgruen:
> @lgruen are y'all running with this change now? I was vaguely concerned that with ~32 JVMs alive that would negatively impact the machine. Do you find that the JVM's RAM gets swapped out and the machines are generally stable?

Yes, we haven't seen any issues with this so far, but also have only been running with this change for a few days.

danking (approved):
fyi @daniel-goldstein , @jigold

PR #12250, opened 2022-09-30
[batch] Allow python 3.9 and 3.10 in python jobs without specifying an image

daniel-goldstein:
CHANGELOG: Users running python 3.9 and 3.10 no longer have to specify a `default_python_image` for PythonJobs when using the default image.


jigold (approved):
Let's make sure this change ends up in the change log. Can you edit the commit message to add `COMMITLOG` or whatever the magic word was?

PR #12251, opened 2022-09-30
[lowering] Write local LD prune as its own stream IR.

chrisvittal:


hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
```diff
@@ -2761,6 +2765,85 @@ object EmitStream {
...
+
+                cb.goto(childProducer.LproduceElement)
+                cb.define(childProducer.LproduceElementDone)
+
+                childProducer.element.toI(cb).consume(cb,
+                  cb.goto(Lpruned),
+                  { case sc: SBaseStructValue =>
+                    val locus = sc.loadField(cb, "locus").get(cb).asLocus
+                    val locusObj = locus.getLocusObj(cb)
+                    val alleles = sc.loadField(cb, "alleles").get(cb).castTo(cb, elementRegion, SJavaArrayString(true), false).asInstanceOf[SJavaArrayStringValue]
```
tpoterba, line 2809: after tracking through the BitPackedVectorBlah classes, I think there's actually no reason that the alleles need to be passed through here. Can we strip that out?

hail/src/main/scala/is/hail/methods/LocalLDPrune.scala
```diff
@@ -22,62 +22,69 @@ object BitPackedVector {
...
+    pack = 0L
+    packOffset = BITS_PER_PACK - 2
+  }
 
-  def load(ptr: Long): Option[BitPackedVector] = {
-    require(nSamples >= 0)
-    hcView.set(ptr)
+  def addGT(call: Int): Unit = {
+    require(idx < nSamples)
+    val gt = Call.nNonRefAlleles(call)
```
tpoterba, line 49: This should probably throw an exception on non-diploid calls. Won't be semantically correct if it's not.

hail/src/main/scala/is/hail/methods/LocalLDPrune.scala
```diff
@@ -88,15 +95,52 @@ private case class BitPackedVectorLoader(fullRowPType: PStruct, callField: Strin
...
       val gtSumSqAll = gtSumSq + nMissing * gtMean * gtMean
       val gtCenteredLengthRec = 1d / math.sqrt(gtSumSqAll - (gtSumAll * gtSumAll / nSamples))
 
-      Some(BitPackedVector(locus, alleles, packs.result(), nSamples, gtMean, gtCenteredLengthRec))
+      BitPackedVector(locus, alleles, packs.result(), nSamples, gtMean, gtCenteredLengthRec)
+    }
+  }
+}
+
+private case class BitPackedVectorLoader(fullRowPType: PStruct, callField: String, nSamples: Int) {
```
tpoterba, line 110: This is only used in the unstaged implementation, right?

tpoterba (dismissed):
Two comments and one question to address. Can't wait to have LD prune available in QoB!

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
chrisvittal: You're correct. However, we do still need the parameter to handle the unstaged code (for now).

hail/src/main/scala/is/hail/methods/LocalLDPrune.scala
chrisvittal: Ok

hail/src/main/scala/is/hail/methods/LocalLDPrune.scala
chrisvittal: correct

hail/src/main/scala/is/hail/expr/ir/streams/EmitStream.scala
tpoterba: Can we just delete the unstaged one? You can define the lowering to TableToTableFunction, then implement the execute with staged boundaries.

chrisvittal: finally did this, we'll see if tests pass, if they do, this is ready to go in.

tpoterba: great!

danking:
bump for @tpoterba, I also can't wait for ld prune in QoB!

tpoterba (approved):

chrisvittal:
I don't understand a) why this isn't merging, and b) why we're seeing an intermittent StackOverflowError.

danking:
Barring bugs in CI, it probably just loses the race to other PRs. I'm not quite sure the ordering of PRs when there are multiple approved and passing ones, @daniel-goldstein can probably comment. I placed high:prio on so that it will merge next.

daniel-goldstein:
Unless its most recent build failed in one of gcp or azure, it should merge over the course of a day (as Dan said, it can get beat out by other PRs. Was there a time where it was all green and not merging?

PR #12253, opened 2022-09-30
[batch] simple & understandable metrics 

danking:
Eliminate `batch_utilization` in favor of `batch_total_cores`. Total cores can be used
in combination with free or user (aka used) cores to determine the cluster utilization.

I also simplified the code for generating the metrics to make it easier to read.


batch/batch/driver/instance.py
```diff
@@ -3,7 +3,7 @@
 import json
 import logging
 import secrets
-from typing import Optional
+from typing import Optional, Dict
```
jigold, line 6: This needs to be sorted.

batch/batch/driver/main.py
```diff
@@ -1152,7 +1152,7 @@ async def cancel_fast_failing_batches(app):
 USER_CORES = pc.Gauge('batch_user_cores', 'Batch user cores', ['state', 'user', 'inst_coll'])
 USER_JOBS = pc.Gauge('batch_user_jobs', 'Batch user jobs', ['state', 'user', 'inst_coll'])
 FREE_CORES = pc.Summary('batch_free_cores', 'Batch instance free cores', ['inst_coll'])
-UTILIZATION = pc.Summary('batch_utilization', 'Batch utilization rates', ['inst_coll'])
+TOTAL_CORES = pc.Summary('batch_total_cores', 'Batch instance free cores', ['inst_coll'])
```
jigold, line 1155: Can you change the metric description to not be the same as the one above it?

batch/batch/driver/main.py
```diff
@@ -1218,33 +1206,26 @@ def set_value(gauge, data):
...
-                billed_rate = instance.instance_config.cost_per_hour_from_cores(
-                    driver.billing_manager.resource_rates, utilized_cores_mcpu
-                )
-                cost_per_hour[billed_cost_per_hour_labels].append(billed_rate)
-
-                inst_coll_labels = InstCollLabels(inst_coll=instance.inst_coll.name)
-                free_cores[inst_coll_labels].append(instance.free_cores_mcpu / 1000)
-                utilization[inst_coll_labels].append(utilized_cores_mcpu / instance.cores_mcpu)
+                free_cores[inst_coll_labels].append(instance.free_cores_mcpu_nonnegative / 1000)
+                total_cores[inst_coll_labels].append(instance.cores_mcpu)
```
jigold, line 1224: I think you need to divide by 1000 here.

batch/batch/driver/main.py
```diff
@@ -1152,7 +1152,7 @@ async def cancel_fast_failing_batches(app):
 USER_CORES = pc.Gauge('batch_user_cores', 'Batch user cores', ['state', 'user', 'inst_coll'])
 USER_JOBS = pc.Gauge('batch_user_jobs', 'Batch user jobs', ['state', 'user', 'inst_coll'])
 FREE_CORES = pc.Summary('batch_free_cores', 'Batch instance free cores', ['inst_coll'])
```
jigold, line 1154: I'm still having trouble with why we use a  Summary here rather than a Gauge.

jigold (dismissed):

batch/batch/driver/main.py
danking: whoops, yep

batch/batch/driver/main.py
danking: Hmm. @daniel-goldstein , you're the resident prometheus expert, do you have thoughts? Upon further consideration, it seems to me that this and TOTAL_CORES are both gauges. I don't think I care about the sum over time of the TOTAL_CORES or FREE_CORES (whereas, maybe I do care about that for request latency, like total time users spent waiting over all time?)

batch/batch/driver/main.py
danking: Indeed! Good catch.

batch/batch/driver/main.py
daniel-goldstein: I'll further note that for this reason `TOTAL_CORES` feels like a not-very-useful Summary, since the further granularity it is showing you over a Gauge is the distribution of machines with different core-counts, which is most likely to always be 100% 16 with maybe a small standing worker. This information is probably best captured with a Gauge that is the number of cores in the collection.

danking: Just to be crystal clear here: how does a `Summary` with no labels differ from a `Gauge` with no labels? I think the original code here had hoped that we could use a `Summary` to aggregate over many observations within a small time window. I tried and tried but never figured out how to express that query in prometheus. Maybe I'm missing some key feature? I think our current "batch_utilization" summary *would* be useful if we could somehow say: "collapse all observations that happen at the same time" (but maybe that's note even true because we don't bulk-observe at one time, we observe repeatedly in a loop).

danking:
OK, here's my understanding of prometheus and our use of it after chatting with @daniel-goldstein :
1. The Python client treats Summary as just a pair of a Gauge and Counter. It's only useful if you need to track the number of times you set/increments/decrement a value and the value itself. For example, the total number of visits to a web page. Some clients for other languages treat Summaries as histogram-like things, but the Python client does not.
2. A Histogram, with a domain-relevant array of buckets, is the right tool for visualizing a distribution of values. In this PR, we treat percent-of-cores-in-use-on-instance as a Histogram. This should let us see the distribution of instances according to what percent of the cores are revenue-generating versus not.
3. A Gauge is the right tool for visualizing any other value. In this PR, we use a gauge to measure the total jobs, the used cores, the total free cores, the total cores, the total cost per hour (ignoring disk), and the total revenue per hour (ignoring disk). 
4. It is important to use `remove` for metrics whose label set changes over time. For example, if all the jobs owned by a particular user finish, then the metrics labelled with that user ought to become `0`. The database query will elide such records; therefore, it is important to `remove` such labels from the `USER_CORES` and `USER_JOBS`. We prefer `remote` to `clear` so as to avoid the case where prometheus collects metrics in between a call to `clear` and a call to `set` which restores the value for a still valid label.

jigold (approved):

PR #12254, opened 2022-09-30
[query/ggplot] Replaces cross-product legends with grouped ones

iris-garden:
Currently, `hail.ggplot.geom_point` generates cross-product-style legends when the `shape` and `color` aesthetics each have multiple groups. This change makes that behavior consistent with `ggplot2`'s implementation, producing legends where the groups are factored out. Because the strategy used to accomplish this does not preserve the interactive features of the legends, such as toggling groups' visibility, this change also makes all plots generated via `hail.ggplot.geom_point` [static](https://plotly.com/python/configuration-options/#making-a-static-chart).

For example, before this change, this code:

```python
import hail as hl
ht = hl.utils.range_table(10)
ht = ht.annotate(threeven=hl.if_else(ht.idx % 3 == 0, "good", "bad"))
fig = (
    + geom_point(aes(color=ht.even, shape=ht.threeven))
)
fig.show()
```

Generates this legend:

Screen Shot 2022-09-29 at 12 22 57

After the change, this legend is generated:

Screen Shot 2022-10-03 at 13 56 01

Custom labels can be used by updating the code like so:
```python
...
from hail.ggplot import ggplot, aes, geom_point, labs
...
fig = (
    + geom_point(aes(color=ht.even, shape=ht.threeven))
    + labs(color="Even", shape="Threeven")
)
...
```

Generating this legend:

Screen Shot 2022-10-03 at 13 58 34

For more information on the current behavior, see #12244 and #12207.

### DIFF
--- a/hail/python/hail/ggplot/labels.py
+++ b/hail/python/hail/ggplot/labels.py
@@ -2,10 +2,11 @@ from .geoms import FigureAttribute
 
 
 class Labels(FigureAttribute):
-    def __init__(self, title=None, xlabel=None, ylabel=None, **kwargs):
+    def __init__(self, title=None, xlabel=None, ylabel=None, group_labels={}, **kwargs):
         self.title = title
         self.xlabel = xlabel
         self.ylabel = ylabel
+        self.group_labels = group_labels
 
     def apply_to_fig(self, fig_so_far):
         layout_updates = {}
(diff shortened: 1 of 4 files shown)


danking:
Just wanted to say this is dope! Particularly considering [this issue](https://github.com/plotly/plotly.js/issues/5099) has been open since 2020.

patrick-schultz:
Really nice work! This is a great change that I'm happy to merge now, but this suggests a number of follow-up improvements:
* we should plan to abstract out the legend handling code to apply uniformly to all geoms
* in the common case where an aesthetic is mapped to a field (as opposed to a compound expression), we should use the field name as the default legend group name. E.g. your second example with custom labels should be the default (but without capitalizing). This matches ggplot behavior.
* if multiple aesthetics are mapped to the same value, they should only generate one legend group. E.g. `geom_point(aes(color=ht.even, shape=ht.even))` should generate one group with title "even", with values "no" a blue circle and "yes" a red square.

hail/python/hail/ggplot/labels.py
```diff
@@ -72,3 +77,19 @@ def ylab(label):
         Label object to change the y-axis label.
     """
     return Labels(ylabel=label)
+
+
+def group_labs(**group_labels):
```
patrick-schultz, line 82: Lets call this `labs` for consistency with ggplot.

patrick-schultz (dismissed):

hail/python/hail/ggplot/labels.py
iris-garden: Updated!

patrick-schultz:
Sorry, could you also add an example to the docstring for `labs`? It isn't obvious that the keys need to be aesthetic names.

danking:
Another follow up PR: update the ggplot tutorial to include the use of cross-product legends!

patrick-schultz (approved):

happydangoe:
Hey everyone! Thank you so much for working on this. As the requestor of many ggplot features -- including this one -- I am so thankful for all your effort in making this a great graphics language :)

PR #12256, opened 2022-10-03
[build.yaml] Make the deploy step work with AR

daniel-goldstein:


jigold:
I think this can go in now. I'm approving, but put the WIP tag on it just in case it's not ready to go and that's why no one was assigned.

jigold (approved):

daniel-goldstein:
Ah I forgot to assign, thanks

PR #12259, opened 2022-10-03
[batch] Fix idempotent update-fast

daniel-goldstein:


batch/batch/front_end/front_end.py
```diff
@@ -1301,7 +1301,7 @@ async def update_batch_fast(request, userdata):
         await _create_jobs(userdata, bunch, batch_id, update_id, app)
     except web.HTTPBadRequest as e:
         if f'update {update_id} is already committed' == e.reason:
-            return web.json_response({'id': batch_id, 'start_job_id': start_job_id})
+            return web.json_response({'update_id': batch_id, 'start_job_id': start_job_id})
```
jigold, line 1304: Should the value for update_id be `update_id` instead of `batch_id`?

jigold (dismissed):

batch/batch/front_end/front_end.py
daniel-goldstein: Yup totally

jigold (approved):

PR #12262, opened 2022-10-04
[lowering] Remove uses of RVDToTableStage in matrix writers

chrisvittal:
Based on [this comment](https://github.com/hail-is/hail/pull/12185#discussion_r985974224)


tpoterba (approved):

tpoterba:
Do you want to go delete all the unstaged implementations of things that have good staged implementations next? There's no reason we still need TableNativeWriter.apply, for instance 😲  💯

chrisvittal:
Sure, sounds great. I am once again asking if I can remove local staging functionality (stageLocally).

tpoterba:
Nope, we should add that. You're right, that's the reason we can't get rid of those unstaged ones yet.

tpoterba:
(I think it's a super quick project)

PR #12269, opened 2022-10-05
[security] enable CodeQL for main and PRs to main

danking:
I mostly want to see what kinds of things this reports. We can shut it off it is annoying.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,14 +1,3 @@
-# For most projects, this workflow file will not need changing; you simply need
-# to commit it to your repository.
-#
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
-#
-# ******** NOTE ********
-# We have attempted to detect the languages in your repository. Please check
-# the `language` matrix defined below to confirm you have the correct set of
-# supported CodeQL languages.
-#
 name: "CodeQL"
 
 on:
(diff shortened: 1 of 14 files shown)


danking:
OK, I think this is actually ready for a real review. Almost everything was spurious (I marked as such, so hopefully we won't have to do that on every PR). There were a few real things:
1. use integrity checks for CDN javascript libraries
2. don't let edits to the search textbox modify the URL arbitrarily
3. don't let the target pages of anchor tags mutate the source page's DOM (wtf, how is this the default behavior???)
4. don't send the IntegrityError from mysql back to the users. I think this is basically safe because of how restrictive we are with which error is printed, but it's also not necessary.

daniel-goldstein:
@danking Looks like there are still some failing CDN fetches and a kind of annoying false-positive complaint about URL sanitization.

danking:
I'll wait for benchmark-service to get rm -rf'ed. I addressed the other CDN issue and I dismissed the false positive. I'll check-in again once benchmark-service is removed.

danking:
Ugh GitHub actions don't re-run when main changes. The state of modern CI is shocking.

danking:
oh wow, it doesn't even merge main before running the security checks. 🤦‍♀️

daniel-goldstein:
> 

gross. Is this status something you want required or tracked by CI?

danking:
Yeah, I'd like to make it required. It's one of the things that NIST 800-53 asks for: automated vulnerability detection.

To be honest, I'm not impressed by a tool that finds variables whose names include the character string "secret" and warns if they're printed, but, 🤷.

danking:
This should now be cleared to merge. I might need to go dismiss all of these alerts on `main` once that merges the first time. Let's leave it on for a bit and assess how much of a headache it is.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,57 @@
...
+        # Prefix the list here with "+" to use these queries and those in the config file.
+
+        # Details on CodeQL's query packs refer to : https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
+        # queries: security-extended,security-and-quality
+
+    # ℹ️ Command-line programs to run using the OS shell.
+    # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
+
+    #   If the Autobuild fails above, remove it and uncomment the following three lines.
+    #   modify them (or add more) to build your code if your project, please refer to the EXAMPLE below for guidance.
```
daniel-goldstein, line 48: Could you remove these various comments?

daniel-goldstein (dismissed):

daniel-goldstein (approved):

PR #12275, opened 2022-10-05
[batch] Check for case sensitive billing project names

jigold:
This is a mitigation and does not solve the issue where we have existing duplicated billing project names. I'm not entirely sure adding BINARY in front of the search value is correct, but I was going off of this post. Apparently, if you use `BINARY name = %s` instead of `name = BINARY %s`, then you pay a performance hit because you cannot use the index any longer.

https://stackoverflow.com/questions/5629111/how-can-i-make-sql-case-sensitive-string-comparison-on-mysql

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -568,7 +568,7 @@ async def _query_batches(request, user, q):
         elif t.startswith('billing_project:'):
             k = t[16:]
             condition = '''
-(batches.`billing_project` = %s)
+(batches.`billing_project` = BINARY %s)
 '''
             args = [k]
         elif t == 'open':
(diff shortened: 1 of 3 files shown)


danking (changes requested):
I'm pretty sure this will prevent us from using the index with any of these where conditions. It seems like fixing the tables themselves is the right move, yeah?

danking:
I like the tests

jigold:
Let me do some experiments and see whether there is a performance cost. The third post down seemed to think the way I wrote it there shouldn't be a performance penalty and it should still use the index. I'm not sure whether I believe that.

jigold:
I think I should try writing the full correct solution with the correct collation type and see how painful it will be.

jigold:
I'm going to close this in favor of solving it a better way now that the job regions PR is basically done.

PR #12276, opened 2022-10-05
[auth] Case sensitive username column

jigold:
Can you point me to where you found `utf8mb4_0900_as_cs` is the correct collation? I'm pretty sure this change doesn't break any foreign key references (I don't think there are any) and just makes it safer for us that a user with a different case name can't impersonate someone else.


danking (approved):
This seems right to me

auth/sql/change-username-collation.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users MODIFY COLUMN username varchar(255) NOT NULL COLLATE utf8mb4_0900_as_cs;
+ALTER TABLE users MODIFY COLUMN login_id varchar(255) NOT NULL COLLATE utf8mb4_0900_as_cs;
```
danking, line 2: you promoted this from `null` to `not null`, I think it will pass if you switch back to `NULL`.

danking (dismissed):

auth/sql/change-username-collation.sql
danking: it needs a `DEFAULT NULL` actually

danking (approved):

danking:
Do we have a similar fix for billing project names?

jigold:
#12327

PR #12278, opened 2022-10-05
[query] teach query to JSON encode numpy types correctly

danking:
CHANGELOG: Fix bug made more likely by 0.2.101 in which Hail errors when interacting with a NumPy integer or floating point type.


danking:
Interestingly, we don't correctly roundtrip. pd.NA will become None and int64 and int32 will become int. I don't think we can round trip as long as we have two Python-side representations for NA.

PR #12285, opened 2022-10-06
[ci] Require publishAs to preserve per-image caching

daniel-goldstein:
I think this is what was wrong with the `git_make_bash_image` taking a minute each time. Since every image without a `publishAs` uses `ci-intermediate`, the `ci-intermediate:cache-PR-X` tag is left pointing to whichever anonymous image built last in the PR run. This is certainly never `git_make_bash_image`, so every time it gets rebuilt, the cache-from that it is using points to an an image whose layers do not include a layer that is `RUN apt-get update && apt-get install -y git make bash`. If this PR runs twice, hopefully we'll see the first step go super quick.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1,6 +1,7 @@
 steps:
   - kind: buildImage2
     name: git_make_bash_image
+    publishAs: git-make-bash
     dockerFile:
       inline: |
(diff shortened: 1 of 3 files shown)


danking (approved):
great find

PR #12299, opened 2022-10-11
[query] allow other kwargs in export

danking:
CHANGELOG: Teach BaseExpression.export to support `parallel` and file `types_file`.


chrisvittal (approved):

danking:
Heh. This has revealed that `hl.import_matrix_table` doesn't work with glob patterns. Fixing this seems subtle to get right.

danking:
This need substantial work. I will create a new PR which revamps import_matrix_table to be simpler and to correctly handle glob patterns.

PR #12300, opened 2022-10-11
[grafana] Update to 9.1.4

daniel-goldstein:
Also added it to third-party images so we're not pulling from DockerHub. Turns out the \ufeff bug we were seeing is hitting a lot of people and is addressed in this release. I put this up in my namespace to see that I can load it without error (though didn't try copying over dashboards and such).


danking (approved):
This seems good but I think we need a better system for these third party images. It seems like we should write the SHAs into the images.txt and have build.yaml's main deploys ensure the registry has the expected images? Or maybe some other automated system.

PR #12301, opened 2022-10-11
[docker] Reorder hailgenetics hail docker image layers

daniel-goldstein:
Installing hail from the whl is going to break this cache every time so we end up installing the gcloud sdk on every build, which takes 40 seconds. It'd be nice to also move the extra pip packages further up the image but I think they're ordered this way so that pip gives us hail-compatible versions. I think the best fix would be to track these packages in a requirements.txt that is pinned and constrained on the hail package's requirements.

### DIFF
--- a/docker/hailgenetics/hail/Dockerfile
+++ b/docker/hailgenetics/hail/Dockerfile
@@ -11,18 +11,6 @@ RUN hail-apt-get-install \
     rsync \
     unzip bzip2 zip tar tabix lz4 \
     vim pv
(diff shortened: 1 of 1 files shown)


danking:
I don't think there's any good reason why the other pip packages are in a separate RUN step. In fact, using one pip invocation should ensure we get compatible versions whereas what we have now doesn't ensure that.

I have no objections to using the hail/python/pinned-requirements.txt. If we install that and all these extra dependencies in one layer then install the hail wheel after (without -U), I *think* it should see all the dependencies are met and just install the hail package.

danking (approved):

PR #12305, opened 2022-10-12
[query] Redesign staged index reading to compute index then query by index

tpoterba:
CHANGELOG: Fixes a rare crash reading tables/matrixtables with _intervals


hail/src/main/scala/is/hail/io/index/StagedIndexReader.scala
```diff
@@ -147,31 +143,69 @@ class StagedIndexReader(emb: EmitMethodBuilder[_], spec: AbstractIndexSpec) {
   }
 
   // returns queryType
```
patrick-schultz, line 145: out of date comment

patrick-schultz (changes requested):
looks good, just one comment needs fixing

hail/src/main/scala/is/hail/io/index/StagedIndexReader.scala
tpoterba: fixed!

patrick-schultz (approved):

PR #12307, opened 2022-10-12
[qob] fix ServiceBackend.get_flags

danking:
I just plainly implemented this wrong the first time.


tpoterba (approved):

danking:
Hmm, so, I see that the py4j backend doesn't error if you `get_flags` for a flag that isn't set. I think it gives you the default value?

The service backend is a bit screwy when it comes to flags because I don't want to run a batch job just to figure out what the Scala-side flags are. I think the right fix is to move all the flag information into Python. What do you think of that @tpoterba ?

tpoterba:
fine by me. We should maybe list the names on the scala side and check we don't get anything outside that set from Python too, though?

tpoterba (changes requested):
requesting changes to get this off my queue, LMK when ready for review

danking:
Superseded by #12423

PR #12309, opened 2022-10-12
[query] persist=checkpoint in all backends

patrick-schultz:
~stacked on #12344~


tpoterba (dismissed):
can you pull the agg fix out into a separate change? If we have issues with that it will be hard/confusing to bisect that into a change with the commit message "persist == checkpoint"

danking:
This appears to be passing, shall we approve? cc: @tpoterba

danking:
bump this for @tp

tpoterba (approved):

PR #12317, opened 2022-10-13
[query] Improve split_multi warning

tpoterba:


chrisvittal (approved):

danking (approved):

hail/python/hail/methods/statgen.py
```diff
@@ -1975,7 +1977,9 @@ def split_multi_hts(ds, keep_star=False, left_aligned=False, vep_root='vep', *,
     Warning
     -------
     This method assumes `ds` contains one non-split variant per locus. This assumption permits the
```
yfarjoun, line 1979:
```suggestion
    This method assumes `ds` contains at most one non-split variant per locus. This assumption permits the
```

yfarjoun (commented):
Thanks. This is much clearer.

PR #12318, opened 2022-10-13
[QOB] Redesign exception propagation from workers

tpoterba:
1. HailException now causes job failure
2. ServiceBackend doesn't short-circuit on batch failure, but instead finds the first bad job (either success with exception or failed with no output) and throws a contextual exception from that.


danking:
I'm broadly OK with this, but I suspect we need to do something more careful to ensure that the worker exception is connected to the driver exception. It appears that some tests are failing for this reason.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -217,9 +217,6 @@ class ServiceBackend(
     stageCount += 1
     implicit val formats: Formats = DefaultFormats
     val batchState = (batch \ "state").extract[String]
-    if (batchState == "failed") {
-      throw new HailBatchFailure(s"Update $updateId for batch $batchId failed")
-    }
```
danking, line 222: @daniel-goldstein you have a bit more recent context on the batch API than I do; can you help Tim get this to the right place?

I think what we want here is to still check for batchState failure, but instead of throwing an exception unconditionally, we want to ask batch which job failed and then try to read that job's result file. If the result file doesn't exist, we probably had the JVM crash.

danking (dismissed):

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
tpoterba: My mental model of my change in this file is that instead of throwing an exception unconditionally, I throw an exception when we get to the first failing job (HailUserException or otherwise). Is that not what's going on?

How is this changing what exceptions propagate to Python? Is something else looking for failing statuses in Python and ignoring the propagated user error if the batch fails?

hail/src/main/scala/is/hail/backend/service/Worker.scala
```diff
@@ -160,6 +160,8 @@ object Worker {
         writeString(dos, shortMessage)
         writeString(dos, expandedMessage)
         dos.writeInt(errorId)
+        log.info(s"job $i/$n failed with user exception: $shortMessage (error id $errorId)\n  $expandedMessage")
+        throw userError
```
daniel-goldstein, line 164: Nice catch here. I was wondering for a bit why worker jobs weren't failing on error.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
daniel-goldstein: My guess for why the tests are failing here is because this difference in exception handling:

https://github.com/hail-is/hail/blob/35e7c4915377da3264955215da2b389a15e18fa6/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala#L691-L701

daniel-goldstein: Er, actually, not entirely sure.

danking:
@iris-garden @daniel-goldstein can you two take a peek of this in light of Iris' changes? This seems OK to merge to me, but y'all have better context to make that call than me.

daniel-goldstein (approved):
Was just reminded of this. This looks correct to me.

tpoterba:
I'll test manually when I rebase.

danking:
bump

danking:
I did a merge for you since I created the conflicts.

iris-garden:
Closing in favor of https://github.com/hail-is/hail/pull/12468.

PR #12321, opened 2022-10-13
[batch] Cleanup regions bit shift loop

jigold:
I'm pretty sure this is still correct.

### DIFF
--- a/batch/batch/utils.py
+++ b/batch/batch/utils.py
@@ -193,10 +193,9 @@ def regions_to_bits_rep(selected_regions, all_regions_mapping):
 def regions_bits_rep_to_regions(regions_bits_rep, all_regions_mapping):
     if regions_bits_rep is None:
         return None
-    reverse_all_regions_mapping = {idx: region for region, idx in all_regions_mapping.items()}
     result = []
-    for i in range(64):
-        selected_region = bool((regions_bits_rep >> i) & 1)
+    for region, idx in all_regions_mapping.items():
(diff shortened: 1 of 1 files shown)


batch/batch/utils.py
```diff
@@ -193,10 +193,9 @@ def regions_to_bits_rep(selected_regions, all_regions_mapping):
 def regions_bits_rep_to_regions(regions_bits_rep, all_regions_mapping):
     if regions_bits_rep is None:
         return None
-    reverse_all_regions_mapping = {idx: region for region, idx in all_regions_mapping.items()}
     result = []
-    for i in range(64):
-        selected_region = bool((regions_bits_rep >> i) & 1)
+    for region, idx in all_regions_mapping.items():
+        selected_region = bool((regions_bits_rep >> idx - 1) & 1)
```
daniel-goldstein, line 198: I feel like this makes more sense to me as `regions_bits_rep & (1 << (idx - 1))` which I think is equivalent?

daniel-goldstein (changes requested):

batch/batch/utils.py
daniel-goldstein: More closely matches the function that goes in the other direction.

jigold: It's not quite the same thing. You have to check whether your result is greater than 0. For example, if regions_bits_rep is 3 and idx is 2, then you get selected_region = 2. I'm not great at bit shift operations so I trust if this way makes more sense to you, then we should do that.

daniel-goldstein (approved):

PR #12322, opened 2022-10-13
[pytest] Treat most warnings as errors

daniel-goldstein:


danking:
warnings are errors is the bees knees

danking:
@daniel-goldstein that svd should have compute_uv=False since it ignores them anyway.

danking (changes requested):
x

danking (approved):

PR #12323, opened 2022-10-14
[ci] Prefix image tags with scope

daniel-goldstein:


jigold (approved):
This is great! Thanks!

PR #12324, opened 2022-10-14
[batch] do not require `mkdir` for bash jobs

danking:
This local temporary directory is only needed for Python jobs. In the bash job case, these directories are created by the input step. Using `mkdir` for all jobs means all images must have `mkdir` which, for example, `nixery.dev/shell/git/htop` does not.

### DIFF
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -622,13 +622,10 @@ class ServiceBackend(Backend[bc.Batch]):
                 if verbose:
                     print(f"Using image '{default_image}' since no image was specified.")
 
-            make_local_tmpdir = f'mkdir -p {local_tmpdir}/{job._dirname}'
-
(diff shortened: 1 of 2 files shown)


jigold:
Can we double check with @bw2 that this won't break his pipelines?

danking:
In 0.2.75 (03f387af5128a6bef10c22cc63494dec46cb01eb) we backwards incompatibly changed the directory from the job id to a 5 character random string. Before that, in 0.2.48 (399e43e7e8c) we backwards incompatibly changed the directory from something like `__JOB__0` to the job id.

Code using Hail 0.2.75 or later which assumes that `/io/batch` or `/io/batch/SIX_RANDOM_CHARACTERS` or `/io/batch/SIX_RANDOM_CHARACTERS/FIVE_RANDOM_CHARACTERS` exists, would be affected by this change.

@bw2 , do you ever assume the existence of a directory called `/io/batch`? (NB: the `/io` directory always exists because its mounted in by the worker.)

PR #12327, opened 2022-10-14
[batch] Mitigation for case insensitive billing projects and usernames

jigold:
FYI @danking

### DIFF
--- a/batch/sql/case-sensitive-billing-project.sql
+++ b/batch/sql/case-sensitive-billing-project.sql
@@ -13,5 +13,3 @@ ALTER TABLE billing_project_users MODIFY COLUMN user_cs VARCHAR(100) NOT NULL CO
 CREATE UNIQUE INDEX `billing_project_name_cs` ON `billing_projects` (`name_cs`);
 CREATE INDEX `billing_project_name_cs_status` ON `billing_projects` (`name_cs`, `status`);
 CREATE INDEX `billing_project_users_billing_project_user_cs` ON `billing_project_users` (`billing_project`, `user_cs`);
-
-
--- a/batch/sql/case-sensitive-billing-project.sql
+++ b/batch/sql/case-sensitive-billing-project.sql
@@ -9,3 +9,9 @@ SET user_cs = user;
 
 ALTER TABLE billing_projects MODIFY COLUMN name_cs VARCHAR(100) NOT NULL COLLATE utf8mb4_0900_as_cs;
 ALTER TABLE billing_project_users MODIFY COLUMN user_cs VARCHAR(100) NOT NULL COLLATE utf8mb4_0900_as_cs;
+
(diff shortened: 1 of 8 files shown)


daniel-goldstein (changes requested):
This looks good, just one thing that felt odd. Additionally, I just want to make sure that we're planning not to add any new users/billing projects in the time between migration and deploy for this PR right? I don't think a fully-safe multi-step PR is worth it, just checking to see if that was the plan

daniel-goldstein:
One more thing that just popped into mind: should there be indexes on these new columns? Probably doesn't matter much on the billing_projects table but I was just wondering

daniel-goldstein (approved):

batch/sql/case-sensitive-billing-project.sql
```diff
@@ -0,0 +1,15 @@
...
+
+UPDATE billing_project_users
+SET user_cs = user;
+
+ALTER TABLE billing_projects MODIFY COLUMN name_cs VARCHAR(100) NOT NULL COLLATE utf8mb4_0900_as_cs;
+ALTER TABLE billing_project_users MODIFY COLUMN user_cs VARCHAR(100) NOT NULL COLLATE utf8mb4_0900_as_cs;
+
+CREATE UNIQUE INDEX `billing_project_name_cs` ON `billing_projects` (`name_cs`);
+CREATE INDEX `billing_project_name_cs_status` ON `billing_projects` (`name_cs`, `status`);
+CREATE INDEX `billing_project_users_billing_project_user_cs` ON `billing_project_users` (`billing_project`, `user_cs`);
```
danking, line 15: What's the plan for dropping the old columns and fixing up the foreign key references?

danking (dismissed):

batch/sql/case-sensitive-billing-project.sql
jigold: There is no plan to do so. I wrote out what needs to happen and it's complicated, is going to take a lot of work / testing to get right, and didn't seem as important as other things we could be doing. If you feel differently, then I can go ahead and do it.

jigold: I forgot how far I got on this, but here was a pilot branch I believe: https://github.com/hail-is/hail/commit/32fe47906dc7c189703ba713cf0e3794084af475

jigold:
@danking bump

jigold:
@danking Can we come up with a plan for whether this short term fix is ok or whether you want me to do the full transformation?

danking:
Ah sorry this wasn't in my CI queue

danking:
We should put the mitigation in and also start a separate PR/project to do the proper fix. Batch is pretty complex and hard to manage already, let's make sure we don't leave cruft in for long periods. I think having this migrated properly in the next 2-3 months is reasonable.

danking (approved):

PR #12334, opened 2022-10-17
[query] Don't tree aggregate unnecessarily

tpoterba:
Making the change we talked about last week in a fresh PR so PCA is faster on the service.


patrick-schultz:
Does PCA do a scan? Either way, do you to do the same optimization for TableAggregate?

tpoterba:
d'oh. Thanks for thinking about this more than I did...

Still a good change, I believe, but won't address PCA. Also added memory management to the non-big cases, can't hurt.

patrick-schultz (approved):

PR #12342, opened 2022-10-17
[qob] Make running development jars easier for QoB

daniel-goldstein:
This will hopefully make it easier to develop QoB. Now, running `make -C hail install-for-qob NAMESPACE=default` will push a jar and configure hailctl to use that jar. So `make -C hail install-for-qob NAMESPACE=default && ipython` will drop you into an ipython session pointed at production where any hail queries reflects your local code. `make -C hail pytest-qob NAMESPACE=default PYTEST_ARGS='-k test_foo'` runs `test_foo` with any of your current changes.

Things to note:
- Variables like `QUERY_STORAGE_URI` are lazy so this should hopefully not break any current usages of the file for non-developers
- You are at no risk of overwriting a release jar unless you specify `UPLOAD_RELEASE_JAR=true` in the make invocation.


tpoterba (approved):
awesome!

PR #12346, opened 2022-10-17
[qob][batch] replace tqdm with rich

danking:
There's a kind of unrelated thing: Fix reading of configuration information to not ignore a hailctl configuration value of `''` 

The big change is to introduce 3 progress bar systems:
1. SimpleRichProgressBar. One progress bar active at a time.
2. RichProgressBar. More than one progress bar active at a time.
3. BatchProgressBar. Same as RichProgressBar but with default columns good for monitoring 1 or more Hail Batch batches.


daniel-goldstein:
@danking I think you have to add `rich` to the `hail/python` dependencies. Docker gets it by accident through `twine`.

daniel-goldstein (dismissed):
This all seems reasonable. I think if we can change the color of the completed portion of the bar from red to green that would be great (and even better if the failed jobs can show up as red but also happy to defer that change)

danking:
batch-pbar
qob-pbar

daniel-goldstein (approved):

PR #12350, opened 2022-10-18
[batch] Add indices for faster pool scheduling

jigold:
I tried benchmarking this change and didn't see much of a difference in timings in my contrived high throughput example. However, I do think this index is better because I believe MySQL does the order by first and then filters records.

@danking Can you take a look at this and make sure the index is actually an improvement.

The speed of the query is linearly related to the number of records in the limit. So I think if we need to get the query speed back to under 10ms then we revert back to pulling a smaller number of records rather than 1000. I think 300 is fine and gets us to 10ms. I just didn't want to pull 10 jobs and then none of them are schedulable but the 100th one is. We can revisit this if the scheduler becomes the bottleneck after your changes to the gateway.

### DIFF
--- a/batch/sql/estimated-current.sql
+++ b/batch/sql/estimated-current.sql
@@ -257,7 +257,7 @@ CREATE TABLE IF NOT EXISTS `jobs` (
 CREATE INDEX `jobs_batch_id_state_always_run_inst_coll_cancelled` ON `jobs` (`batch_id`, `state`, `always_run`, `inst_coll`, `cancelled`);
 CREATE INDEX `jobs_batch_id_state_always_run_cancelled` ON `jobs` (`batch_id`, `state`, `always_run`, `cancelled`);
 CREATE INDEX `jobs_batch_id_update_id` ON `jobs` (`batch_id`, `update_id`);
-CREATE INDEX `jobs_batch_id_always_run_n_regions_regions_bits_rep_job_id` ON `jobs` (`batch_id`, `always_run`, `n_regions`, `regions_bits_rep`, `job_id`);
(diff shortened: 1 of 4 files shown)


batch/sql/estimated-current.sql
```diff
@@ -257,7 +257,7 @@ CREATE TABLE IF NOT EXISTS `jobs` (
 CREATE INDEX `jobs_batch_id_state_always_run_inst_coll_cancelled` ON `jobs` (`batch_id`, `state`, `always_run`, `inst_coll`, `cancelled`);
 CREATE INDEX `jobs_batch_id_state_always_run_cancelled` ON `jobs` (`batch_id`, `state`, `always_run`, `cancelled`);
 CREATE INDEX `jobs_batch_id_update_id` ON `jobs` (`batch_id`, `update_id`);
-CREATE INDEX `jobs_batch_id_always_run_n_regions_regions_bits_rep_job_id` ON `jobs` (`batch_id`, `always_run`, `n_regions`, `regions_bits_rep`, `job_id`);
```
daniel-goldstein, line 260: Should this change drop this index?

daniel-goldstein (changes requested):

batch/sql/estimated-current.sql
jigold: In thinking it through further, I think I'd like to do that in a separate PR. It seems dangerous to delete a critical index until the service is using the new one in case there's problems. What do you think?

daniel-goldstein: Makes sense! Can do it that way

jigold:
The tests are failing for unrelated reasons. I'll retry once approved.

jigold:
Also putting the WIP tag on since this database migration is 15 minutes or so

daniel-goldstein (approved):

PR #12354, opened 2022-10-19
[ci] Permit deployment of batch and auth at the same time

daniel-goldstein:
The serial deployment of auth and batch can lengthen the critical path of CI pipelines by a few minutes if k8s needs to spin up new nodes. While auth is necessary for batch to function correctly, it's not necessary to deploy batch, so I think it's more appropriate to not have `deploy_batch` depend on `deploy_auth` but have anything that depends on `deploy_batch` also depend on `deploy_auth`. There's already a precedent for this in that the service backend tests depend on `deploy_batch` and `deploy_memory` as opposed to `deploy_batch` being dependent on `deploy_memory`.

I also removed the dependency of `upload_query_jar` on `deploy_batch`, no idea why that was there, but maybe it should be the other way around?

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2096,7 +2096,6 @@ steps:
       - batch_worker_image
       - batch_driver_nginx_image
       - batch_database
-      - deploy_auth
       - create_certs
   - kind: runImage
     name: create_billing_projects
@@ -2157,6 +2156,7 @@ steps:
       - service_base_image
       - create_deploy_config
       - create_accounts
+      - deploy_auth
       - deploy_batch
   - kind: runImage
     name: upload_query_jar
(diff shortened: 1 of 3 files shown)


jigold:
Is it worth just adding a sink no-op node for deploy_batch and deploy_auth which is like deploy_batch_system (I know terrible name)? I'm just worried about the case where we someday have a storage service or other multiple components in addition to auth.

daniel-goldstein:
That's a nice idea, it would help with e.g. forgetting to `deploy_memory` whenever I want to mess with QoB in my namespace, which would admittedly become more of a problem if I make this change. I'll try that out

jigold (changes requested):
see comment.

danking:
bump

daniel-goldstein:
I think I have a slightly different idea for this that less invasive. Will open a new PR with that

PR #12355, opened 2022-10-19
[memory,batch] Reduce logging output

jigold:
FYI @danking 

We needed to reduce logging output again. The memory service was printing logs for every request so I got rid of that. We were missing a new endpoint in the batch driver that didn't need to be logged. Finally, a lot of logs looked like syslog from the worker so I made sure to turn off the syslog Fluentd logging as the first thing we do on the worker. Not sure if it will help, but worth a try.

### DIFF
--- a/batch/batch/cloud/gcp/driver/create_instance.py
+++ b/batch/batch/cloud/gcp/driver/create_instance.py
@@ -153,6 +153,56 @@ nohup /bin/bash run.sh >run.log 2>&1 &
 #!/bin/bash
 set -x
 
+# Setup fluentd
+touch /worker.log
+touch /run.log
+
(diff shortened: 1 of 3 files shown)


danking:
Let's not turn off memory logs entirely. As we progress to NIST 800-53 compliance we need to log requests that use credentials like this. Instead, let's reduce the volume of bytes per-request.

How much money do we spend on logs? We have to view them as necessary cost of business, but one that we can reduce in terms of reducing log volume.

daniel-goldstein:
Rate limits will also ensure we don't exceed a certain log volume, so we can use that to figure out what our target bytes-per-log should be

daniel-goldstein (approved):

PR #12361, opened 2022-10-20
[query] Update JSON serialized relational type names

chrisvittal:
This makes the keys we produce in serialized representation the same as the field names in the python objects.


danking (changes requested):
Do I understand correctly that this is used for Python-Java communication but is not serialized to disk?

chrisvittal:
I looked through the code. As far as I can tell, this is transient and is never saved.

danking (approved):

PR #12364, opened 2022-10-20
[grafana,prometheus] Switch from nginx to envoy

daniel-goldstein:
Despite the pretty daunting diff I think this is a pretty "small" change and perhaps an easier one start out with in the migration from nginx -> envoy than gateway/internal-gateway (certainly less risky). I've tested these manually by make deploying and they work both in dev namespaces and default

Currently, the grafana and prometheus pods have two containers: the app itself (grafana or prometheus) and an nginx container that sits in front of it. The flow is as follows, and since this works the exact same for both prometheus and grafana I will just talk about grafana as the example and the same thing should apply to both:

1. User sends request to grafana.hail.is
2. Gateway sees an HTTP request going to a production service and forwards that request to the grafana k8s Service port 443
3. The grafana K8s Service forwards that request to the grafana pod port 443
4. Nginx is listening on port 443 in the grafana pod and receives that request. It makes an authorization check to auth to make sure that the request is coming from a developer
5. Nginx forwards that request to 127.0.0.1:3000, which is where grafana is listening

This PR does not change any behavior, just replaces Nginx with Envoy. Currently, building the nginx container involves running jinja on its config files and building a docker image. With envoy, we can just use the `envoyproxy/envoy` image from DockerHub (which I have copied into our container registries) and feed it a single configmap. The big mess of yaml that is the new configmap for envoy has a lot of boilerplate, but it comprises of the following sections which hopefully on their own are not too bad.

### Envoy config
1. The top of the `listeners` section shows that Envoy is listening on port 8443 (which is the port that the k8s `Service` will now forward traffic to)
2. The `virtual_hosts` section shows that Envoy will send all paths (prefix "/") to the cluster `grafana`
3. The `http_filters` section says that Envoy will first send an authorization request to the `auth` cluster before allowing the request to pass to grafana
4. The `clusters` section says that there are two other services Envoy knows about and can send traffic to, one named `auth` that can be found at address `auth` (same as `https://auth`), and a cluster called `grafana` which can be found at `127.0.0.1:3000` (same as `localhost:3000`)


jigold:
I think this needs a rebase.

grafana/deployment.yaml
```diff
@@ -1,3 +1,113 @@
...
+                address:
+                  socket_address:
+                    address: 127.0.0.1
+                    port_value: 3000
+    admin:
+      address:
+        socket_address:
+          address: 127.0.0.1
+          port_value: 8001
+---
```
jigold, line 110: There's an extra `---` here.

grafana/deployment.yaml
```diff
@@ -1,3 +1,113 @@
...
+                - name: envoy.filters.http.ext_authz
+                  typed_config:
+                    "@type": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthz
+                    http_service:
+                      server_uri:
+                        uri: https://auth:443
+                        cluster: auth
+                        timeout: 0.25s
+{% if deploy %}
+                      path_prefix: /api/v1alpha/verify_dev_credentials?url=
```
jigold, line 45: What's the `url=` for? Is that hardcoded somewhere?

grafana/deployment.yaml
```diff
@@ -1,3 +1,113 @@
...
+          - lb_endpoints:
+            - endpoint:
+                address:
+                  socket_address:
+                    address: 127.0.0.1
+                    port_value: 3000
+    admin:
+      address:
+        socket_address:
+          address: 127.0.0.1
```
jigold, line 108: I assume this admin block is for envoy?

jigold (changes requested):

jigold:
Putting the WIP tag so you can merge at your leisure.

jigold (approved):

danking:
🎉

danking:
I am eager to change the node pools. Are we ready to that?

PR #12369, opened 2022-10-21
[query] More utilities for relational types

chrisvittal:
Add `__slots__` to prevent assignment bugs and a `_to_json` method for `ttable`, `tmatrix`, and `tblockmatrix`.

### DIFF
--- a/hail/python/hail/expr/blockmatrix_type.py
+++ b/hail/python/hail/expr/blockmatrix_type.py
@@ -29,6 +29,12 @@ class tblockmatrix(object):
         self.is_row_vector = is_row_vector
         self.block_size = block_size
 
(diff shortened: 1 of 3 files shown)


chrisvittal:
The only use of the `_to_json` method I have is in a forthcoming combiner update. Theoretically I can use what I put in the `_to_json` method for tmatrix directly where I need it.

danking (changes requested):
Services team has preferred to call these methods `to_dict` since they don't produce JSON strings. I have a mild preference for consistency across the entire codebase.

danking:
e.g. https://github.com/hail-is/hail/blob/main/hail/python/hailtop/auth/sql_config.py#L19-L36

PR #12370, opened 2022-10-21
[query] Add milliseconds to log output

daniel-goldstein:
It was the change to `HailContext.scala` that finally affected the output for QoB jobs, but I tried to stay consistent.


danking (approved):
ugh. Yeah I think the JSON stuff is now unused because I got rid of the "query" service.

PR #12371, opened 2022-10-21
[ci] Narrow what parts of repo some steps need to localize

daniel-goldstein:


danking:
Maybe we should just cloud fuse the relevant bucket?

I am curious how much of a difference trimming a few files makes. I kinda figured this was mostly latency bound not bandwidth bound.

daniel-goldstein:
Ya I was wondering too which is why I tried this out. It was helpful in observing how much number of files is the bottleneck for the copy tool. Takes between 15-20 seconds to copy down all 5k files in the repo while jobs that copy more MB but in fewer files are [much better](https://ci.hail.is/batches/6235246/jobs/52).

I don't exactly prefer how, but it feels silly to spend 25% of the time of `auth_image` downloading the entire repo when you need just the auth directory

danking (approved):
I'm good with this. Class A operations are killing our finances!

danking:
Needs conflict resolution.

PR #12372, opened 2022-10-24
[vds combiner] Save dataset type in combiner state 

chrisvittal:
The dataset type should not change with each stage in the combiner, so
we are able to assert the type of the combiner datasets at each stage
and prevent round trips to the backend during reading. As such, we can
remove a call to read_multiple_matrix_tables which requires spark.

This sets up more changes so that we can switch on the backend and run
the combiner in query on batch.

### DIFF
--- a/hail/python/hail/vds/combiner/variant_dataset_combiner.py
+++ b/hail/python/hail/vds/combiner/variant_dataset_combiner.py
@@ -1,25 +1,24 @@
 import collections
 import hashlib
-import itertools
 import json
 import os
 import sys
 import uuid
 
 from math import floor, log
-from typing import Collection, Dict, List, Optional, Union, Tuple
+from typing import Collection, Dict, List, NamedTuple, Optional, Union
 
 import hail as hl
 
+from hail.expr import HailType, tmatrix
 from hail.utils import Interval
-from hail.utils.java import Env, info, warning
+from hail.utils.java import info, warning
 from hail.experimental.vcf_combiner.vcf_combiner import calculate_even_genome_partitioning, \
     calculate_new_intervals
-from hail.vds.variant_dataset import VariantDataset
 from .combine import combine_variant_datasets, transform_gvcf, defined_entry_fields
 
 
-class VDSMetadata:
+class VDSMetadata(NamedTuple):
     """The path to a Variant Dataset and the number of samples within.
 
     Parameters
@@ -30,14 +29,14 @@ class VDSMetadata:
         Number of samples contained within the Variant Dataset at `path`.
 
     """
+    path: str
+    n_samples: int
 
-    def __init__(self, path: str, n_samples: int):
-        self.path = path
-        self.n_samples = n_samples
 
-    def to_tuple(self) -> Tuple[str, int]:
-        """Convert to a serializable representation."""
-        return (self.path, self.n_samples)
+class CombinerOutType(NamedTuple):
+    """A container for the types of a VDS"""
+    reference_type: tmatrix
+    variant_type: tmatrix
 
 
 FAST_CODEC_SPEC = """{
@@ -54,18 +53,6 @@ FAST_CODEC_SPEC = """{
     }
   }
 }"""
-
-
-def read_variant_datasets(inputs: List[str], intervals: List[Interval], intervals_dtype):
-    n_inputs = len(inputs)
-    paths = list(itertools.chain(
-        (VariantDataset._reference_path(path) for path in inputs),
-        (VariantDataset._variants_path(path) for path in inputs)))
-    mts = Env.spark_backend("read_variant_datasets").read_multiple_matrix_tables(
-        paths, intervals, intervals_dtype)
-    vdss = [VariantDataset(reference_data=ref, variant_data=var)
-            for ref, var in zip(mts[:n_inputs], mts[n_inputs:])]
-    return vdss
 
 
 class VariantDatasetCombiner:  # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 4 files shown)


hail/python/hail/vds/combiner/variant_dataset_combiner.py
```diff
@@ -30,14 +29,17 @@ class VDSMetadata:
...
+    n_samples: int
 
-    def __init__(self, path: str, n_samples: int):
-        self.path = path
-        self.n_samples = n_samples
 
-    def to_tuple(self) -> Tuple[str, int]:
-        """Convert to a serializable representation."""
-        return (self.path, self.n_samples)
+class VDSType(NamedTuple):
```
tpoterba, line 36: I feel like VDS being a type makes it somehow official.

hail/python/hail/vds/variant_dataset.py
```diff
@@ -7,7 +7,8 @@
 from hail.genetics import ReferenceGenome
 
 
-def read_vds(path, *, intervals=None, n_partitions=None) -> 'VariantDataset':
+def read_vds(path, *, intervals=None, n_partitions=None,
+             _assert_reference_type=None, _assert_variant_type=None) -> 'VariantDataset':
```
tpoterba, line 11: I don't remember the plan for these -- were you going to implement a checked read in a stacked PR?

hail/python/hail/vds/combiner/variant_dataset_combiner.py
```diff
@@ -612,20 +618,48 @@ def maybe_load_from_saved_path(save_path: str) -> Optional[VariantDatasetCombine
...
+        mt = hl.import_vcf(gvcf_paths[0], header_file=gvcf_external_header, force_bgz=True,
+                           array_elements_required=False, reference_genome=reference_genome,
+                           contig_recoding=contig_recoding)
+        gvcf_type = mt._type
+        rmt = mt.filter_rows(hl.is_defined(mt.info.END))
+        gvcf_reference_entry_fields_to_keep = defined_entry_fields(rmt, 100_000) - {'GT', 'PGT', 'PL'}
+
+        vds = transform_gvcf(mt._key_rows_by_assert_sorted('locus'), gvcf_reference_entry_fields_to_keep, gvcf_info_to_keep)
+        dataset_type = VDSType(reference_type=vds.reference_data._type, variant_type=vds.variant_data._type)
+    elif vds_paths:
```
tpoterba, line 639: combine with the first case?

tpoterba (changes requested):
couple questions

hail/python/hail/vds/combiner/variant_dataset_combiner.py
chrisvittal: fair enough, renamed

hail/python/hail/vds/variant_dataset.py
chrisvittal: In many ways, these options are the whole reason for this PR. We want to be able to create a MatrixTable without a blocking round trip to the backend to get it's type. The async PR uses these options.

danking:
@chrisvittal if you dismiss the reviewer that'll trigger it to show up in Tim's CI page. cc: @tpoterba

hail/python/hail/vds/combiner/variant_dataset_combiner.py
```diff
@@ -612,20 +615,46 @@ def maybe_load_from_saved_path(save_path: str) -> Optional[VariantDatasetCombine
...
-        mt = mt.filter_rows(hl.is_defined(mt.info.END))
-        gvcf_reference_entry_fields_to_keep = defined_entry_fields(mt, 100_000) - {'GT', 'PGT', 'PL'}
+        gvcf_type = mt._type
+        vds = transform_gvcf(mt._key_rows_by_assert_sorted('locus'), gvcf_reference_entry_fields_to_keep, gvcf_info_to_keep)
+        dataset_type = CombinerOutType(reference_type=vds.reference_data._type, variant_type=vds.variant_data._type)
+    if gvcf_type is None and gvcf_paths:
+        mt = hl.import_vcf(gvcf_paths[0], header_file=gvcf_external_header, force_bgz=True,
+                           array_elements_required=False, reference_genome=reference_genome,
+                           contig_recoding=contig_recoding)
+        gvcf_type = mt._type
```
danking, line 648: I did some fiddling with this and I think the following is both correct and easier to reason about. What do you think of it? It still doesn't ensure that the extant VDS type matches the GVCF type. I presume that'll fail later?

```python3
    vds = None
    gvcf_paths = None
    if vds_paths:
        vds = hl.vds.read_vds(vds_paths[0])
        if gvcf_reference_entry_fields_to_keep is None:
            gvcf_reference_entry_fields_to_keep = set(vds.reference_data.entry) - {'END'}
    if gvcf_paths:
        gvcf_mt = hl.import_vcf(
            gvcf_paths[0],
            header_file=gvcf_external_header,
            force_bgz=True,
            array_elements_required=False,
            reference_genome=reference_genome,
            contig_recoding=contig_recoding
        )
        gvcf_type = gvcf_mt._type
        if gvcf_reference_entry_fields_to_keep is None:
            rmt = gvcf_mt.filter_rows(hl.is_defined(gvcf_mt.info.END))
            gvcf_reference_entry_fields_to_keep = defined_entry_fields(rmt, 100_000) - {'GT', 'PGT', 'PL'}
        if vds is None:
            vds = transform_gvcf(
                gvcf_mt._key_rows_by_assert_sorted('locus'),
                gvcf_reference_entry_fields_to_keep,
                gvcf_info_to_keep
            )

    dataset_type = CombinerOutType(
        reference_type=vds.reference_data._type,
        variant_type=vds.variant_data._type
    )
```

danking (changes requested):

danking (approved):
Tim's got a lot on his plate. I'll accept responsibility for approval. AFAICT, this looks right!

lgruen:
Does this mean the combiner now works on QoB or will that require additional changes?

danking:
The combiner uses an IR that writes many matrix tables in parallel. This is not yet lowered. As far as I know, that's the last missing piece. @chrisvittal can correct me if I'm wrong about that.

PR #12375, opened 2022-10-25
[query] Finally get tabix vcf reading working

chrisvittal:
This does, what I believe to be the minimum necessary to get Tabix reading working in a standard VCF reader.
 
* Add GenericLines readTabix
* Harmonize import_gvcfs arguments and MatrixVCFReader parameters, as such, it is now possible to directly override the sample ids.
* Remove the undocumented (and until this change, dead) _partitions argument to import_vcf

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
@@ -383,8 +383,8 @@ object GenericLines {
       "partitionIndex" -> TInt32,
       "file" -> TString,
       "chrom" -> TString,
-      "start" -> TInt64,
-      "end" -> TInt64)
+      "start" -> TInt32,
+      "end" -> TInt32)
     new GenericLines(contextType, contexts, body)
   }
 
--- a/hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
@@ -343,10 +343,10 @@ object GenericLines {
         CloseableIterator.empty
       } else {
         new CloseableIterator[GenericLine] {
-          val lines = new TabixLineIterator(fs, file, reg)
-          private var l = lines.next()
-          private var curIdx: Long = lines.getCurIdx()
-          val inner = new Iterator[GenericLine] {
+          private[this] val lines = new TabixLineIterator(fs, file, reg)
+          private[this] var l = lines.next()
+          private[this] var curIdx: Long = lines.getCurIdx()
+          private[this] val inner = new Iterator[GenericLine] {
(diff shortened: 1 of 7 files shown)


hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
```diff
@@ -298,6 +309,85 @@ object GenericLines {
...
+        val r = new TabixReader(file, fs)
+        val tid = r.chr2tid(chrom)
+        r.queryPairs(tid, start - 1, end)
+      }
+      if (reg.isEmpty) {
+        CloseableIterator.empty
+      } else {
+        new CloseableIterator[GenericLine] {
+          val lines = new TabixLineIterator(fs, file, reg)
+          private var l = lines.next()
```
danking, line 347: `private[this]` will ensure this is a JVM field and not accessor methods.

hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
```diff
@@ -15,6 +18,14 @@ import scala.annotation.meta.param
 
 trait CloseableIterator[T] extends Iterator[T] with AutoCloseable
 
+object CloseableIterator {
+  def empty[T]: CloseableIterator[T] = new CloseableIterator[T] {
```
danking, line 22: a shame the iterator interface isn't properly covariant so this could just be a `val empty: CloseableIterator[AnyRef]`.

hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
```diff
@@ -298,6 +309,85 @@ object GenericLines {
...
+      val index = contextRow.getAs[Int](0)
+      val file = contextRow.getAs[String](1)
+      val chrom = contextRow.getAs[String](2)
+      val start = contextRow.getAs[Int](3)
+      val end = contextRow.getAs[Int](4)
+
+      val reg = {
+        val r = new TabixReader(file, fs)
+        val tid = r.chr2tid(chrom)
+        r.queryPairs(tid, start - 1, end)
```
danking, line 340: Is this robust to `start` being the first position? i.e. passing `0` to queryPairs is fine?

danking (changes requested):
Is the point of this change that we can read a variant subset of a VCF in time proportional to that subset not proportional to the entire file?

Can we add a python-level test on some VCF we have in the tests?

hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
chrisvittal: Yes. We do so all the time in the combiner with default genome intervals.

danking (approved):

PR #12376, opened 2022-10-25
[query] Add `hl.query_table` function for reading tables with indices…

tpoterba:
… from Python

### DIFF
--- a/hail/python/hail/ir/ir.py
+++ b/hail/python/hail/ir/ir.py
@@ -2870,61 +2870,68 @@ class MatrixAggregate(IR):
 class PartitionReader(object):
     pass
 
-    def set_uid_field(self, uid_field):
+    def with_uid_field(self, uid_field):
         pass
 
     def render(self):
         pass
 
     def _eq(self, other):
+        pass
+
+    def row_type(self):
         pass
 
 
 class PartitionNativeIntervalReader(PartitionReader):
-    def __init__(self, path):
+    def __init__(self, path, table_row_type, uid_field=None):
         self.path = path
-        self.uid_field = '__dummy'
-
-    def set_uid_field(self, uid_field):
+        self.table_row_type = table_row_type
         self.uid_field = uid_field
+
+    def with_uid_field(self, uid_field):
+        return PartitionNativeIntervalReader(self.path, uid_field)
 
     def render(self):
         return escape_str(json.dumps({"name": "PartitionNativeIntervalReader",
                                       "path": self.path,
-                                      "uidFieldName": self.uid_field}))
+                                      "uidFieldName": self.uid_field if self.uid_field is not None else '__dummy'}))
 
     def _eq(self, other):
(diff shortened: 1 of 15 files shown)


tpoterba:
Important infrastructure for Seqr.

hail/python/hail/ir/ir.py
```diff
@@ -2865,6 +2866,65 @@ def renderable_bindings(self, i, default_value=None):
...
+        self.row_typ = row_typ
+        self.row_typ_with_uid = row_typ
+        self.drop_uid = True
+        self.reader = reader
+
+    def _handle_randomness(self, create_uids):
+        if create_uids:
+            self.row_typ_with_uid = tstruct(**self.row_typ, __uid=ttuple(tint64, tint64))
+            self.drop_uid = False
+            self.reader.set_uid_field('__uid')
```
patrick-schultz, line 2913: All the other `_handle_randomness` methods return new nodes, they don't modify the node. I think that's necessary, since an IR can be reused in different contexts.

hail/python/test/hail/table/test_table.py
```diff
@@ -2016,3 +2016,60 @@ def test_indexed_read_boundaries(branching_factor):
         ])
 
         assert t1.idx.collect() == [141, 142, 143, 144, 152]
+
+
+def test_query_table():
+    f = new_temp_file(extension='ht')
```
patrick-schultz, line 2022: Maybe include a few tests with a compound key?

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -712,6 +717,245 @@ case class PartitionNativeReader(spec: AbstractTypedCodecSpec, uidFieldName: Str
...
+          ).search(cb, partitionerRuntime, EmitCode.present(cb.emb, startAndSignTuple)))
+          cb.assign(currPartitionIdx, startPartitionIndex)
+
+          cb.assign(lastIncludedPartitionIdx, new BinarySearch(cb.emb,
+            partitionerRuntime.st,
+            EmitType(endAndSignTuple.st, true),
+            ((cb, elt) => elt),
+            bound = "lower",
+            compareEnd(_, _, _, _ < 1)
+          ).search(cb, partitionerRuntime, EmitCode.present(cb.emb, endAndSignTuple)))
```
patrick-schultz, line 834: From a quick look, I think you've reimplemented the registered function `partitionerFindIntervalRange`. If so, you should just factor out the implementation to a staged function (as most of the interval registered functions are) and call that. I think we should try to keep all the staged interval ordering stuff in one place, since it's relatively complicated.

patrick-schultz (changes requested):

tpoterba:
good comments; addressed these and did a fair bit of redesigning of the Python stuff. Refactored the scala to use that method you mentioned, also got mad that we didn't already have a SStackInterval and added that.

PR #12378, opened 2022-10-25
[hail-ubuntu] use focal-20221019

danking:


danking:
@daniel-goldstein I want to do this regularly. I also don't want to create a headache for AUS every time I do it. What if I added a build.yaml step that used skopeo to run copy_images.sh on every PR? What is docker_prefix set to in PRs and namespaces? I think this should almost never copy any information, just give new names to old images.

daniel-goldstein:
docker_prefix is the same for all namespaces: `us-docker.pkg.dev/hail-vdc/hail`. The steps sounds fine, though I'd like to see how much time it would take. Skopeo has been pretty great on no-op copies in my experience.

danking:
Hmm.

The fact that we're all commingling in the same ACR namespace means that a PR could copy a new image for a given tag. That seems to be a nonstarter.

If we namespaced our ACR (e.g. `hail-vdc/default/hail` and `hail-vdc/pr-99999-fdsaf/hail`), do copies copy data or just create new tags pointing to old data?

danking:
PRs having a fresh namespace for images (perhaps with some explicitly seeded ones) also avoids adding a new image that isn't in images.txt which seems like a feature.

daniel-goldstein:
I believe it works like a bucket where `us-docker.pkg.dev/hail-vdc/hail` is the bucket name and everything after that is the blob path, so it should be fine if you make the docker_prefix `us-docker.pkg.dev/hail-vdc/hail/<NAMESPACE>`, see [here](https://console.cloud.google.com/artifacts/docker/hail-vdc/us/hail?project=hail-vdc).

danking:
Hmm. OK, this might have legs.

daniel-goldstein:
@danking I think this should be mergeable now without AUS needing to make manual changes

danking:
@daniel-goldstein OK, I think this is ready for a final review.

daniel-goldstein (approved):

PR #12379, opened 2022-10-26
[auth] fix this cloud cookie nonsense

danking:
Heh. The fix is almost trivial, just gotta RTFM for http. Users will need to log out / clear cookies before this will take effect.


danking:
ugh this won't actually work. We want the cookie to be sent to batch.hail.is, auth.hail.is, etc.

jigold:
I was just going to say that.

PR #12380, opened 2022-10-26
[auth] fix this cloud cookie nonsense

danking:
My frustration boiled over and I went to go fix this. This change will cause everyone to become logged out because they will lack the right cookie. That seems fine.

### DIFF
--- a/ci/test/resources/deployment.yaml
+++ b/ci/test/resources/deployment.yaml
@@ -49,6 +49,9 @@ spec:
            - mountPath: /deploy-config
              name: deploy-config
              readOnly: true
+           - name: global-config
+             mountPath: /global-config
+             readOnly: true
            - name: session-secret-key
              mountPath: /session-secret-key
              readOnly: true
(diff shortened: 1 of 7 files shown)


jigold (changes requested):
Do we need to test this by hand or are you confident it will work? I'm sure this is fine, but I do want to avoid a case where we lock ourselves out of our system.

jigold:
Also, this might break grafana as it sends a cookie to prometheus.

danking:
I addressed the grafana situation.

I dev deployed into my namespace and I'm able to log in.

I cannot test that I've addressed the issue because the default namespace is still using the old version and thus I hit 401s there before even reaching the dev namespace.

jigold:
The tests failed. Can't tell if it's from this commit. Can you double check everything is good or retry?

danking:
OK, should be resolved now. I had a lint error and needed to mount the global config into hello and website.

jigold (approved):

PR #12383, opened 2022-10-27
[batch] Migrate TLS sidecar from nginx to envoy

daniel-goldstein:
In the spirit of consolidation, this is just translating the current nginx config that is handling TLS for the batch driver to envoy. Doesn't have any special dependencies or deployment complications. Note that this time running Envoy as root is required to listening on the same port (443) that nginx currently listens on.


jigold:
Can you bump me on this when you want another review?

daniel-goldstein:
@jigold passing now

jigold (approved):

PR #11800, opened 2022-04-26
[query/docs] add databricks docs with disclaimers

williambrandler:
Signed-off-by: William Brandler <william.brandler@databricks.com>

Moving the Databricks documentation for Hail to the open source Hail documentation. This project is not part of the Databricks product, and therefore instructions for using the Hail open source project on Databricks should be maintained by the open source community moving forward

Also spilt `other_cloud_providers` into three, AWS, Databricks and Others

PR #12386, opened 2022-10-27
[query/docs] add databricks docs with disclaimers

iris-garden:
See https://github.com/hail-is/hail/pull/11800.


danking (approved):

danking:
thank you!

PR #12389, opened 2022-10-28
[query] Lower and stage `import_bgen`

tpoterba:
- add infrastructure for producing streams as iterators
- move index attributes map to a runtime, rather than compile-time parameter
- redesign index_bgen to be a registered function


danking:
What is this stacked on?

patrick-schultz (approved):

PR #12392, opened 2022-10-28
Fixes for envoy

daniel-goldstein:
- Using the full domain name instead of the shorthand `<service>.<namespace>` isn't strictly necessary but just made me nervous and I opted for the full unambiguous domain for in-cluster services (what if we had a namespace named `com`?? I feel like that would break some things)
- I got the configuration wrong on how to tell envoy *not* to worry about certs of internal namespaces (I'd recommend using the `split` view for the diff because otherwise it's pretty hard to read)
- For internal namespaces, using the `prefix` parameter for matching a route allowed `/foo/batch` to match a route like `/foo/batch-driver` which is obviously not great. the `path_separated_prefix` parameter actually does what we want
- Fixed the batch tests not to look at the HTTP 1.1 reason phrase and instead look at the response body to determine the error

### DIFF
--- a/batch/test/test_batch.py
+++ b/batch/test/test_batch.py
@@ -4,9 +4,9 @@ import secrets
 import time
 from typing import Set
 
-import aiohttp
 import pytest
 
+from hailtop import httpx
 from hailtop.auth import service_auth_headers
 from hailtop.batch_client.client import BatchClient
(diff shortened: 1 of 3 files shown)


danking (approved):

danking:
🚢it

PR #12395, opened 2022-10-31
[docker] Remove unused service-java-run-base image

daniel-goldstein:
Pretty sure this is unused (rg shows no more mentions in the repo)


danking (approved):
Yeah this as for ye olde query service when it was really a service.

PR #12396, opened 2022-10-31
[test] hail-run image should depend on base not service base

daniel-goldstein:
Wonder what your thoughts are on this. I think that we have too deep an image hierarchy and that it's only stalling our time-to-test in CI. I propose here getting rid of the run-tests image and just putting plink in the hail-run image. I also remove the dependency on service-base and just install hailtop in the hail-run image -- I don't think anything hail/hail should depend on services code. This change actually revealed an accidental dependency where there was a gear (services) test in the hail/python tests. I moved that over to the batch suite as batch is the primary user of that class and we don't have a gear suite atm.

Ultimately, I think we should have a pretty flat docker image hierarchy. Each of our Dockerfiles does a decent job of following the docker mantra of dependencies first, code later. But if you consider the fact that we stack images on top of each other then you can see that in reality that doesn't translate to our images! We instead get an interleaving of hail code and dependencies such that it's very easy to break the cache for images like this one.

### DIFF
--- a/batch/test/test_time_limited_max_size_cache.py
+++ b/batch/test/test_time_limited_max_size_cache.py
@@ -1,8 +1,8 @@
-import pytest
 import asyncio
 
-from gear.time_limited_max_size_cache import TimeLimitedMaxSizeCache
+import pytest
 
+from gear.time_limited_max_size_cache import TimeLimitedMaxSizeCache
 
 pytestmark = pytest.mark.asyncio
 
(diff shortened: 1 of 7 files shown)


daniel-goldstein:
@tpoterba bump

danking (approved):
I support this and the long-term vision of a flat hierarchy. Tim's plate is quite full right now.

PR #12397, opened 2022-10-31
[batch] Fix JVM file exists error

jigold:
Docker jobs were cleaning up after themselves by including the entire directory for that job which removed the old container files. This PR makes it so `Container.remove()` removes the directory for the container.

Before we merge this, I want to check the logs for this PR and make sure there aren't error messages that don't show up in tests.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -2192,6 +2192,8 @@ class JVM:
                 if self.container:
                     await self.container.remove()
 
+                await blocking_to_async(self.pool, shutil.rmtree, f'{self.root_dir}/container', ignore_errors=True)
+
                 container = await self.create_container_and_connect(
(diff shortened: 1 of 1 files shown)


batch/batch/worker/worker.py
```diff
@@ -816,6 +819,8 @@ async def _cleanup(self):
             if self.netns:
                 network_allocator.free(self.netns)
                 self.netns = None
+
+            await blocking_to_async(self.pool, shutil.rmtree, self.container_scratch, ignore_errors=True)
```
daniel-goldstein, line 823: Isn't the Job the one that creates the scratch directory? I would think that whoever creates the directory should be the one to destroy it.

daniel-goldstein (changes requested):

daniel-goldstein (approved):

PR #12398, opened 2022-10-31
[batch] Kill any remaining jobs on worker shutdown

jigold:
This should fix the issue where we were trying to use a nonexistent GoogleAsyncFS when a worker is killed in the UI. I tested this in my namespace.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1296,6 +1296,7 @@ class Job:
         self.worker = worker
 
         self.deleted_event = asyncio.Event()
+        self.killed = False
 
         self.token = uuid.uuid4().hex
         self.scratch = f'/batch/{self.token}'
@@ -1368,6 +1369,7 @@ class Job:
         self.project_id = Job.get_next_xfsquota_project_id()
 
         self.mjs_fut: Optional[asyncio.Future] = None
+        self.run_job_fut: Optional[asyncio.Future] = None
 
     def write_batch_config(self):
         os.makedirs(f'{self.scratch}/batch-config')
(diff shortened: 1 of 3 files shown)


batch/batch/worker/worker.py
```diff
@@ -1400,10 +1402,22 @@ async def delete(self):
...
         self.deleted_event.set()
 
+    async def kill(self):
+        log.info(f'killing {self}')
+        self.killed = True
+        if self.run_job_fut is not None:
+            self.run_job_fut.cancel()
+            await asyncio.wait([self.run_job_fut])
+            assert self.run_job_fut.done()
+            self.run_job_fut = None
```
daniel-goldstein, line 1412: I like this `kill` option on the Job, but it's not clear to me why it's only concerned about the run job future. What about `mjs_fut` or the `post_job_complete` future that we don't even bind to a variable? I think the cancellation of those futures is handled by `task_manager.shutdown`, so it feels weird to me that some are handled implicitly there and this future is handled explicitly here. It also feels a little weird that `run_job_fut` gets cancelled twice, once here and once in the task_manager. I guess I would prefer to either not have a `Job.kill` and make the cancellation of the tasks in the task_manager Just Work, or have `Job.kill` manage all tasks the jobs produces.

I think the first option is the least work and not so bad, I think the main problem here is that `Job.run` runs `mark_complete` regardless of whether the task was cancelled or if there was some other exception. Could we instead update `Job.run` so that it doesn't run `mark_complete` if the task was cancelled? (Can inspect `sys.exc_info` inside the finally block).

daniel-goldstein (changes requested):

daniel-goldstein (approved):

PR #12399, opened 2022-10-31
[query] improved index search algorithm

patrick-schultz:
~Stacked on #12376~

Reworks the staged index reader query algorithm. Before, `queryInterval` did three traversals from root to leaf: one each for the start and end of the interval, and one to find the first record. Now it always does a single traversal, visiting one or two nodes at each level (two if the paths for the start and end keys diverge), and doing a single binary search at each visited node.

This is probably not a significant performance improvement when doing a single query per partition, but could be beneficial when using the new `query_table` per row.


danking:
@patrick-schultz this needs a
```
git rebase --onto hi/main  24943c85b index-reader
```

tpoterba (approved):

PR #12400, opened 2022-10-31
[httpx] Replace inspection of e.message with e.body 

daniel-goldstein:
for error responses coming from our services.

I think I got all occurrences of this by searching for `e.message` across the codebase. The only instances I see left of this are error-handling withing `front_end.py` (which isn't a problem because this is an issue of the reason phrase not making it between server -> client) and exceptions from aiodocker which is not within our control.

### DIFF
--- a/auth/auth/driver/driver.py
+++ b/auth/auth/driver/driver.py
@@ -362,20 +362,20 @@ class BillingProjectResource:
 
         try:
             await self.batch_client.create_billing_project(billing_project)
-        except aiohttp.ClientResponseError as e:
-            if e.status != 403 or 'already exists' not in e.message:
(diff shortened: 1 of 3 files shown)


jigold (changes requested):
I don't see that you change the Client anywhere. Was it always the one from `httpx` or did that happen in another PR?

daniel-goldstein:
It always was. All of the `ClientSession`s that we use are `httpx.ClientSession` which subclasses `aiohttp.ClientSession` and wraps all the method calls. When there are errors it returns a `httpx.ClientResponseError` which subclasses `aiohttp.ClientResponseError` which is why these excepts were still valid.

jigold (approved):

PR #12402, opened 2022-11-01
[envoy] Set higher rate limits for most services and add test_ci services

daniel-goldstein:
- CI needs to account not only for the services that it is deploying but the services that the test-ci might later deploy into that same namespace
- I had earlier applied batch-driver's rate limit to everything and soon found that it is a bit too restrictive w.r.t. many simultaneous client submissions to the batch front end. With nginx we just didn't have rate limits on anything else, which seems less good than setting high rate limits. I'll admit that 200 was incredibly arbitrary (let's call it optimistic) and I'm happy to take more precise suggestions.

### DIFF
--- a/ci/ci/envoy.py
+++ b/ci/ci/envoy.py
@@ -93,7 +93,7 @@ def gateway_default_host(service: str) -> dict:
                 'match': {'prefix': '/'},
                 'route': route_to_cluster(service),
                 'typed_per_filter_config': {
-                    'envoy.filters.http.local_ratelimit': rate_limit_config(),
+                    'envoy.filters.http.local_ratelimit': rate_limit_config(service, 'default'),
                     'envoy.filters.http.ext_authz': auth_check_exemption(),
                 },
             }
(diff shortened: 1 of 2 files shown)


danking (changes requested):

ci/ci/envoy.py
```diff
@@ -172,13 +172,17 @@ def auth_check_exemption() -> dict:
     }
 
 
-def rate_limit_config() -> dict:
+def rate_limit_config(service: str, namespace: str) -> dict:
+    # FIXME We didn't set a rate limit before with nginx so I'm making this
+    # sort of unreasonably high. This should be tuned per service
```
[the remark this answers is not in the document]
daniel-goldstein: Good point, I forgot about our pod autoscalers and have just been thinking 1-3 pods. The limit is per service but also per internal-gateway/gateway pod, so at least 3x the given number (and possibly more from the gateway since we allow that to autoscale).

ci/ci/envoy.py
```diff
@@ -172,13 +172,15 @@ def auth_check_exemption() -> dict:
     }
 
 
-def rate_limit_config() -> dict:
+def rate_limit_config(service: str, namespace: str) -> dict:
+    max_rps = 60 if service == 'batch-driver' and namespace == 'default' else 200
```
danking, line 176: Shouldn't we rate limit batch-driver in every namespace? Differential rate limiting by namespace seems like it will make understanding test speed and test slowness harder.

danking (changes requested):

ci/ci/envoy.py
daniel-goldstein: fair point

daniel-goldstein:
@danking Apologies for the rapid bump, but it would be great if I could test the envoy changes again tonight if you have no further requests.

danking (approved):

danking:
I'm terrible with PRs these days, bump away!

PR #12410, opened 2022-11-01
[native query] Build LLVM and native query in ci

chrisvittal:
~~Need to get sccache working for building LLVM~~

Install llvm/mlir from apt.llvm.org. Need to get llvm-lit working for the tests.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1441,6 +1441,46 @@ steps:
       - make_docs
     clouds:
       - gcp
+  - kind: buildImage2
+    name: query_build_image
+    dockerFile: /io/query/Dockerfile.query-build
+    contextPath: /io/query
+    publishAs: query-build
+    inputs:
+      - from: /repo/query
+        to: /io/query
+    dependsOn:
+      - hail_ubuntu_image
+      - merge_code
+  - kind: runImage
+    name: build_native_query
+    image:
+      valueFrom: query_build_image.image
+    script: |
+      set -ex
+      export LLVM_DIR=/usr/lib/llvm-15
+      cd /io/repo/query
+      mkdir build
+      cd build
+      cmake .. \
+        -G Ninja \
+        -DCMAKE_C_COMPILER=$LLVM_DIR/bin/clang \
+        -DCMAKE_CXX_COMPILER=$LLVM_DIR/bin/clang++ \
+        -DCMAKE_C_FLAGS=-Werror \
+        -DCMAKE_CXX_FLAGS=-Werror \
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DMLIR_DIR=$LLVM_DIR/lib/cmake/mlir
+        # -DLLVM_BUILD_BINARY_DIR=$LLVM_SRC_DIR/build/bin # FIXME need llvm-lit for tests
+      ninja
+    resources:
+      memory: standard
+      cpu: '2'
+    inputs:
+      - from: /repo
+        to: /io/repo
+    dependsOn:
+      - query_build_image
+      - merge_code
   - kind: deploy
     name: deploy_grafana
     namespace:
--- a/query/CMakeLists.txt
+++ b/query/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.18)
+cmake_minimum_required(VERSION 3.16)
 project(hail-mlir-dialect LANGUAGES CXX C)
 
 set(CMAKE_BUILD_WITH_INSTALL_NAME_DIR ON)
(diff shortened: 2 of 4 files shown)


chrisvittal:
Ok, so, llvm-dev requires python3-yaml. Attempting to pip install version 6 (our dep) onto the system fails. So, that's fun.

docker/hail-ubuntu/Dockerfile
```diff
@@ -16,10 +16,18 @@ RUN chmod 755 /bin/retry && \
...
+    curl 'https://apt.llvm.org/llvm-snapshot.gpg.key' \
+        | gpg --dearmor >> /usr/share/keyrings/llvm-snapshot-keyring.gpg && \
     echo 'deb [signed-by=/usr/share/keyrings/deadsnakes-ppa-archive-keyring.gpg] http://ppa.launchpad.net/deadsnakes/ppa/ubuntu focal main' \
          >> /etc/apt/sources.list && \
     echo 'deb-src [signed-by=/usr/share/keyrings/deadsnakes-ppa-archive-keyring.gpg] http://ppa.launchpad.net/deadsnakes/ppa/ubuntu focal main' \
          >> /etc/apt/sources.list && \
+    echo 'deb [signed-by=/usr/share/keyrings/llvm-snapshot-keyring.gpg] http://apt.llvm.org/focal/ llvm-toolchain-focal-15 main' \
+         >> /etc/apt/sources.list && \
+    echo 'deb-src [signed-by=/usr/share/keyrings/llvm-snapshot-keyring.gpg] http://apt.llvm.org/focal/ llvm-toolchain-focal-15 main' \
+         >> /etc/apt/sources.list && \
```
daniel-goldstein, line 30: Are these working? I thought the implicit behavior of ubuntu in docker nowadays is to clear the sources list after install to reduce layer size. Either way, I think it'd make more sense to keep these in the same `RUN` directive as what installs the llvm toolchain.

build.yaml
```diff
@@ -1441,6 +1441,35 @@ steps:
...
+      cd build
+      cmake .. \
+        -G Ninja \
+        -DCMAKE_C_COMPILER=$LLVM_DIR/bin/clang \
+        -DCMAKE_CXX_COMPILER=$LLVM_DIR/bin/clang++ \
+        -DCMAKE_C_FLAGS=-Werror \
+        -DCMAKE_CXX_FLAGS=-Werror \
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DMLIR_DIR=$LLVM_DIR/lib/cmake/mlir
+        # -DLLVM_BUILD_BINARY_DIR=$LLVM_SRC_DIR/build/bin # FIXME need llvm-lit for tests
```
daniel-goldstein, line 1462: See first comment. This would ultimately be necessary as a layer in a build docker image, right?

daniel-goldstein (changes requested):

docker/hail-ubuntu/Dockerfile
chrisvittal: These were working. The build breaks because we use pip to install PyYAML 6 and llvm-15-dev depends on python3-yaml.

build.yaml
chrisvittal: I don't think so. Can you clarify? This is just to build the native code in CI to make sure no one breaks the build as we develop. I expect it to be expanded upon soon as we get the test suite working and built up.

daniel-goldstein: Nvm as I'm re-reading it this makes sense.

docker/hail-ubuntu/Dockerfile
daniel-goldstein: Could we solve that by not building the `query-build` image on base_image and instead base it on hail_ubuntu? Should be lighter weight just will probably need to replicate a few apt installs

chrisvittal: good point. I'll do that.

docker/hail-ubuntu/Dockerfile
```diff
@@ -14,6 +14,8 @@ RUN chmod 755 /bin/retry && \
     echo "APT::Acquire::Retries \"5\";" > /etc/apt/apt.conf.d/80-retries && \
     mkdir -p /usr/share/keyrings/ && \
     hail-apt-get-install curl gpg && \
+    curl 'https://keyserver.ubuntu.com/pks/lookup?search=0xF23C5A6CF475977595C89F51BA6932366A755776&hash=on&exact=on&options=mr&op=get' \
+         | gpg --dearmor > /usr/share/keyrings/deadsnakes-ppa-archive-keyring.gpg && \
```
daniel-goldstein, line 18: This is the same as the line below

daniel-goldstein (changes requested):

chrisvittal:
Thanks all fixed!

daniel-goldstein (approved):

daniel-goldstein:
Screen Shot 2022-11-02 at 3 58 14 PM
Love this

PR #12411, opened 2022-11-01
[query] Retry transient errors in HadoopFS

daniel-goldstein:


hail/src/main/scala/is/hail/io/fs/HadoopFS.scala
```diff
@@ -120,19 +121,19 @@ class HadoopFS(private[this] var conf: SerializableHadoopConfiguration) extends
...
   }
 
-  override def mkDir(dirname: String): Unit = {
+  override def mkDir(dirname: String): Unit = retryTransientErrors {
     getFileSystem(dirname).mkdirs(new hadoop.fs.Path(dirname))
   }
 
-  def remove(fname: String): Unit = {
+  def remove(fname: String): Unit = retryTransientErrors {
     getFileSystem(fname).delete(new hadoop.fs.Path(fname), false)
```
danking, line 129: For retry transient to work you'll need to ensure idempotency. I think it's sufficient to try-catch and smother FileNotFoundException Same for delete. Not sure about rmtree. You'll have to try calling that on a directory that doesn't exist.

hail/src/main/scala/is/hail/io/fs/HadoopFS.scala
```diff
@@ -76,15 +77,15 @@ class HadoopFS(private[this] var conf: SerializableHadoopConfiguration) extends
...
-  def createNoCompression(filename: String): PositionedDataOutputStream = {
+  def createNoCompression(filename: String): PositionedDataOutputStream = retryTransientErrors {
     val fs = getFileSystem(filename)
     val hPath = new hadoop.fs.Path(filename)
     val os = fs.create(hPath)
     new WrappedPositionedDataOutputStream(
       HadoopFS.toPositionedOutputStream(os))
   }
 
-  def openNoCompression(filename: String): SeekableDataInputStream = {
```
danking, line 87: If the streams hiccup we'll still get errors, but this PR is a worthwhile improvement. I'm not sure how to robustly retry the input streams we get from Hadoop anyway.

danking (changes requested):

danking:
bump

daniel-goldstein:
@danking Have you actually seen anymore of these errors after your fix of the hadoop tests erasing each others' files? I considering closing this because I've not seen further evidence that hadoop isn't properly addressing transient errors.

danking:
Let's close until we have more evidence.

PR #12412, opened 2022-11-02
[release] Update changelog for 0.2.106

jigold:
This is for releasing the feature added by #11884. We need to merge this PR soon after #11884 goes in as it's a breaking change and the release will give users the option to revert back to the current behavior.


hail/python/hailtop/batch/docs/change_log.rst
```diff
@@ -3,6 +3,12 @@
 Change Log
 ==========
 
+**Version 0.2.106**
+
+- Added Job.always_copy_output when using the ServiceBackend. The default behavior is False which
+is a breaking change from the previous behavior to always copy output files irregardless of the job's
```
danking, line 9: irregardless => regardless

danking (changes requested):

hail/python/hailtop/batch/docs/change_log.rst
jigold: sorry. blindly copied from the original PR

PR #12415, opened 2022-11-02
[query] Fix text partition writers to use UUIDs and avoid speculation…

tpoterba:
… overwrites

CHANGELOG: Fixed bug sometimes leading to dropped rows from exported text files with Spark speculation.


chrisvittal (changes requested):
These should already have UUIDs in the paths in the contexts. Make sure we aren't 'doubling up' unnecessarily.

tpoterba:
Did this! #11910

PR #12417, opened 2022-11-02
[native] teach Emacs that query/ is C++ code

danking:


chrisvittal (approved):
\<holywar\>use vim instead\</holywar\>

PR #12422, opened 2022-11-03
[native] Move headers into a hail directory

chrisvittal:


chrisvittal:
This is a bikeshed, but I do think it's good organization to have all our headers with a `hail` prefix.

tpoterba:
Patrick already did a reorganization in the development branch we're using, so this is just going to introduce rebase conflicts right now. Can we postpone until next week?

chrisvittal:
Yeah! Not a problem.

PR #12425, opened 2022-11-04
[envoy] Add gzip compression for requests to upstreams

daniel-goldstein:
Also included useful changes that made it possible to diagnose and fix this problem including:

- No longer dropping metrics for test namespaces. I checked the prometheus disk and we have plenty of space to add these additional metrics. Very useful for diagnosing test time latencies.
- Add prometheus scraping for envoy pods. Gives us many great metrics like number of 2xx, 3xx, 4xx and 5xx requests per upstream, rate limit enforcement, even time until the cert expires
- Made `Connection reset` a retry-once error. A connection reset can sometimes be indistinguishable from non-transient errors when the client is not able to inspect the response code before the reset clears the TCP buffer. We take multiple consecutive resets to mean an intentional action from the server indicating that the client is doing something wrong and retrying will not help.

I have put off adding gzip compression to gateway in this PR. It was working fine with all of our services except for grafana, in which it was messing up the websocket connection for some reason. I'll dig into that separately.


daniel-goldstein:
Azure is currently running this internal-gateway, and PRs seem to be doing no worse, and sometimes better afaict, than the nginx in GCP. If we decide to merge this I would feel confident rolling out envoy to gcp as well (and can resize the k8s pool immediately after!)

hail/src/main/scala/is/hail/services/package.scala
```diff
@@ -41,6 +41,8 @@ package object services {
     // true error.
     val e = reactor.core.Exceptions.unwrap(_e)
     e match {
+      case e: SocketException =>
+        e.getMessage.contains("Connection reset")
```
danking, line 45: check that the message is not null

danking (changes requested):

danking (approved):

PR #12428, opened 2022-11-04
[qob] VEP for GRCh37 and GRCh38

jigold:
Already done:
- Made a public bucket in us-central1 with requester pays on (gs://qob-vep-grch37-us-central1)
- Uploaded the configuration file for grch37 in us-central1 to gs://hail-common/qob-vep/
- Documentation

To-Do items:
- Mirror the loftee base image in our artifact registry
- Replicate the data in Azure and create an Azure configuration file (can't make this requester pays, so not sure what to do here)
- Add instructions for setting up a config file in the respective infra READMEs
- Write tests
- Get GRCh38 working
- Modify the cloud run functions for ACR cleanup to cleanup the vep-grch37 and eventually vep-grch38 images


jigold:
Looks like there's quite a few rebase errors to touch up as well.

docker/vep/Dockerfile.GRCh37
```diff
@@ -0,0 +1,36 @@
+FROM konradjk/vep85_loftee:1.0.3
+
+ENV VEP_VERSION 85
+ENV VEP_DIR /vep_bin
+ENV REFERENCE_GENOME grch37
```
danking, line 5: unused

docker/vep/Dockerfile.GRCh37
```diff
@@ -0,0 +1,36 @@
...
+RUN sed -i 's/master/main/g' INSTALL.pl
+
+RUN perl INSTALL.pl -a a
+
+WORKDIR $VEP_DIR
+
+RUN echo "#!/bin/bash" > /vep
+RUN echo "export PERL5LIB=$VEP_DIR/loftee:\$PERL5LIB" >> /vep
+RUN echo "perl $VEP_DIR/ensembl-tools-release-$VEP_VERSION/scripts/variant_effect_predictor/variant_effect_predictor.pl \"\$@\"" >> /vep
+RUN chmod +x /vep
```
danking, line 32: Now that we're happy with this we should collapse the RUN lines into one line. Docker seems to really struggle with images with lots of small layers.

docker/vep/Dockerfile.GRCh37
```diff
@@ -0,0 +1,36 @@
+FROM konradjk/vep85_loftee:1.0.3
```
danking, line 1: I do not want a dependency on Konrad's DockerHub image.

Let's just copy [his Dockerfile](https://github.com/konradjk/loftee/blob/hail_grch37/hail_files/Dockerfile_vep) into this repo and replace the base with hail-ubuntu. Then we can use `hail-apt-get-install` to simplify some of the stuff below. Relatedly, I reject building Python from source. The hail-ubuntu image should have a reasonable version of Python.

hail/python/hail/backend/service_backend.py
```diff
@@ -325,6 +325,49 @@ def render(self, ir):
         assert len(r.jirs) == 0
         return r(finalize_randomness(ir))
 
+    async def _submit_batch(self, build_batch_f, name, *, progress: Optional[BatchProgressBar] = None, **batch_args):
```
danking, line 328: Hmm. I don't think this belongs here, it only references a handful of properties from ServiceBackend. It's totally legitimate for `_service_vep` to ask for things like `hl.current_backend().name_prefix` etc.

If we want to use some of the QoB error formatting information elsewhere, perhaps that should live in `hailtop.batch`?

hail/python/hail/backend/service_backend.py
```diff
@@ -325,6 +325,49 @@ def render(self, ir):
...
+            bb = self.async_bc.create_batch(token=token, attributes=batch_attributes)
+            build_batch_f(bb, self.flags)
+            b = await bb.submit(disable_progress_bar=True)
+
+        with timings.step("wait batch"):
+            try:
+                if self.disable_progress_bar is not True:
+                    deploy_config = get_deploy_config()
+                    url = deploy_config.external_url('batch', f'/batches/{b.id}/jobs/1')
+                    print(f'Submitted batch {b.id}, see {url}')
```
danking, line 345: Hmm, I think this should get nixed from here and `_rpc`. The new fancy progress bars already handle this really well. In fact, I never see this message printed which makes me think the if condition isn't right anyway.

hail/python/hail/backend/service_backend.py
danking: I think my inclination is to just stuff this in service_vep until we have another example of Batch-QoB interaction. A few more examples will enable us to generalize appropriately.

hail/python/hail/methods/qc.py
```diff
@@ -491,13 +504,128 @@ def n_discordant(counter):
     return glob, per_sample.cols(), per_variant.rows()
 
 
+def _service_vep(ht, config, block_size, csq, tolerate_parse_error):
+    token = secret_alnum_string(16)
+
+    backend = hl.context.current_backend()
+    assert isinstance(backend, hl.backend.service_backend.ServiceBackend)
```
danking, line 511: Take the backend as an argument with the right type. Then in the caller do this:
```
backend = hl.current_backend()
if isinstance(backend, hl.backend.ServiceBackend):
    _service_vep(..., backend)
else:
    ...
```

This way we do one is instance check. If we try to pass `hl.current_backend` then mypy will complain because the types don't match.

hail/python/hail/methods/qc.py
```diff
@@ -491,13 +504,128 @@ def n_discordant(counter):
     return glob, per_sample.cols(), per_variant.rows()
 
 
+def _service_vep(ht, config, block_size, csq, tolerate_parse_error):
```
danking, line 507: I know most of query doesn't have types, but new code should use types.

hail/python/hail/methods/qc.py
```diff
@@ -491,13 +504,128 @@ def n_discordant(counter):
...
+
+    backend = hl.context.current_backend()
+    assert isinstance(backend, hl.backend.service_backend.ServiceBackend)
+
+    tmp_dir = hl.backend.service_backend.tmp_dir()
+
+    vep_input_path = f'{tmp_dir}/vep/inputs/{token}.vcf'
+    vep_output_path = f'{tmp_dir}/vep/outputs/{token}'
+
+    hl.export_vcf(ht, vep_input_path, parallel='header_per_shard')
```
danking, line 518: Use `hl.TemporaryFile` or `hl.TemporaryDirectory`. They are context managers which will take care of deleting the temporary file or directory.

hail/python/hail/methods/qc.py
```diff
@@ -491,13 +504,128 @@ def n_discordant(counter):
...
+        image = local_config['image']
+        data_bucket = local_config['data_bucket']
+        data_mount = local_config['data_mount']
+        region = local_config['region']
+        env = local_config['env']
+
+    if csq:
+        vep_typ = hl.tstr
+    else:
+        vep_json_schema = local_config.get('vep_json_schema')
```
danking, line 531: @tpoterba What's the correct & forwards compatible way to serialize a Hail type as a string which we'll store somewhere and load in Python?

hail/python/hail/methods/qc.py
```diff
@@ -491,13 +504,128 @@ def n_discordant(counter):
...
+        if vep_json_schema is None:
+            raise ValueError("'vep_json_schema' not found in config.")
+        vep_typ = java_typ_to_dtyp(vep_json_schema)
+
+    def build_vep_batch(bb: bc.aioclient.BatchBuilder, flags):
+        requester_pays_project = flags.get('gcs_requester_pays_project')
```
danking, line 537: `hl._get_flags(...)` is the usual way to access this.

hail/python/hail/methods/qc.py
```diff
@@ -491,13 +504,128 @@ def n_discordant(counter):
...
+
+        if csq:
+            csq_command = local_config['csq_header_command']
+            local_output_file = '/io/output'
+
+            local_env = copy.deepcopy(env)
```
danking, line 543: Why deepcopy? AFAICT, no one else uses the env dict.

hail/python/hail/methods/qc.py
```diff
@@ -491,13 +504,128 @@ def n_discordant(counter):
...
+            if not part_name.startswith('part-'):
+                continue
+            part_id = int(part_name.split('-')[1])
+
+            run_vep_command = local_config['command']
+
+            local_input_file = '/io/input'
+            local_output_file = '/io/output.gz'
+
+            local_env = copy.deepcopy(env)
```
danking, line 575: same

hail/python/hail/methods/qc.py
```diff
@@ -491,13 +504,128 @@ def n_discordant(counter):
...
+                          output_files=[(local_output_file, f'{vep_output_path}/csq-header')],
+                          regions=[region],
+                          requester_pays_project=requester_pays_project,
+                          env=local_config['env'],
+                          )
+
+        for f in hl.hadoop_ls(vep_input_path):
+            path = f['path']
+            part_name = os.path.basename(path)
+            if not part_name.startswith('part-'):
```
danking, line 566: What other stuff shows up here?

hail/python/hail/methods/qc.py
```diff
@@ -491,13 +504,128 @@ def n_discordant(counter):
...
+    annotations = hl.import_table(f'{vep_output_path}/annotations/*',
+                                  key='variant',
+                                  types={'variant': hl.tstr,
+                                         'vep': vep_typ,
+                                         'vep_proc_id': hl.tstruct(part_id=hl.tint,
+                                                                   block_id=hl.tint)},
+                                  force=True)
+
+    reference_genome = ht.locus.dtype.reference_genome.name
+    annotations = annotations.key_by(**hl.parse_variant(annotations.variant, reference_genome=reference_genome))
```
danking, line 609: @tpoterba is it possible to avoid this key_by by specifying the right kind of type in `types`?

hail/python/hail/methods/qc.py
```diff
@@ -567,6 +695,32 @@ def vep(dataset: Union[Table, MatrixTable], config=None, block_size=1000, name='
...
+    The config file when using the service backend must contain the following fields:
+
+     - `command` (array of string) -- The command line to run for a VEP job for a partition.
+     - `csq_header_command` (array of string) -- The command line to run when generating the consequence header.
+     - `env` (object) -- A map of environment variables to values to add to the environment when invoking the command.  The value of each object member must be a string.
+     - `vep_json_schema` (string): The type of the VEP JSON schema (as produced by the VEP when invoked with the `--json` option).  Note: This is the old-style 'parseable' Hail type syntax.  This will change.
+     - `image` (string) -- The docker image to run VEP.
+     - `data_bucket` (string) -- The location where the VEP data is stored.
+     - `data_mount` (string) -- The location in the container where the data should be mounted.
+     - `region` (string) -- The cloud region in which to run VEP jobs.
```
danking, line 707: I think I'm totally convinced now that this should not be a URL to some remote config file. It should just be a dictionary (or maybe a bunch of arguments?). We never read this file in a Batch worker, AFAICT, it's just read here on the client-side.

hail/python/hail/methods/qc.py
```diff
@@ -567,6 +695,32 @@ def vep(dataset: Union[Table, MatrixTable], config=None, block_size=1000, name='
...
+     - `VEP_DATA_MOUNT` - location where the vep data is mounted (same as `data_mount` in the config)
+     - `VEP_CONSEQUENCE` - integer equal to 0 or 1 on whether `csq` is False or True
+     - `VEP_TOLERATE_PARSE_ERROR` - integer equal to 0 or 1 on whether `tolerate_parse_error` is False or True
+     - `VEP_JSON_SCHEMA` - type of the vep JSON schema (same as `vep_json_schema` in the config)
+
+    The VEP input VCF shard is available at `/io/input`. Hail expects an output TSV file with VEP data at `/io/output.gz`.
+
+    The following configuration files are available:
+
+     - (``GRCh37``, ``gcp``, ``us-central1``): ``gs://hail-common/qob-vep/qob-vep-config-grch37.json``
```
danking, line 722: My inclination is to just make this a dictionary which Hail uses to determine the right location for VEP data, e.g.:

```
vep_data_buckets = {
    ('GRCh37', 'gcp', 'us-central1'): 'hail-qob-vep-grch37-us-central1'
}
```
Add a parameter for `vep_data_bucket: Optional[str] = None`. If `vep_data_bucket` is unspecified, we figure it out from reference genome, cloud, and region. If specified, we ignore the dictionary and use the specified location. If we are forced to move the data, we announce it and users on old versions just have to explicitly specify `vep_data_bucket`.

docker/vep/vep.py
```diff
@@ -0,0 +1,227 @@
...
+        header = consume_header(inp)
+        for block_id, block in enumerate(grouped_iterator(block_size, inp)):
+            n_processed = len(block)
+            start_time = time.time()
+
+            proc_id = f'{{"part_id":{part_id},"block_id":{block_id}}}'
+            variants = [Variant.from_vcf_line(l.rstrip()) for l in block]
+            non_star_to_orig_variants = {str(v.strip_star_allele()): str(v) for v in variants}
+
+            with sp.Popen(vep_cmd, env=env, stdin=sp.PIPE, stdout=sp.PIPE, stderr=sp.PIPE, encoding='utf-8') as proc:
```
danking, line 122: Hmm, curious, tell me more about this. Why can't we just run VEP on the VCF shard directly?

danking (changes requested):

docker/vep/Dockerfile.GRCh37
jigold: The issue is that Konrad's image is built from an old ensembl image containing vep that has an old version of Python because it's Ubuntu 16.04. I don't want to mess with putting in a different version of VEP as the gnomad team is going to reject that. I don't want to be in the business of me personally having to validate any new version of VEP that we upgrade to and have to support. Also, who knows if Konrad's loftee code will still work.

I think that if we want to do this properly, then we should pull in someone else to either commit to helping out and testing the new image or build it themselves.

docker/vep/vep.py
jigold: I did the exact same thing as we do in the Scala implementation. Even if I had VEP call the variants from the VCF directly, we'd still need to parse the result in Python. We also keep a mapping of the original variant name with some preprocessing to match with the output from VEP.

docker/vep/Dockerfile.GRCh37
jigold: @konradjk Can you chime in on this conversation?

hail/python/hail/methods/qc.py
danking: I think we should just inline the type actually. this should do it:
```
hl.tstruct(
            assembly_name=hl.tstr,
            allele_string=hl.tstr,
            ancestral=hl.tstr,
            colocated_variants=hl.tarray(hl.tstruct(
                aa_allele=hl.tstr,
                aa_maf=hl.tfloat,
                afr_allele=hl.tstr,
                afr_maf=hl.tfloat,
                allele_string=hl.tstr,
                amr_allele=hl.tstr,
                amr_maf=hl.tfloat,
                clin_sig=hl.tarray(hl.tstr),
                end=hl.tint32,
                eas_allele=hl.tstr,
                eas_maf=hl.tfloat,
                ea_allele=hl.tstr,
                ea_maf=hl.tfloat,
                eur_allele=hl.tstr,
                eur_maf=hl.tfloat,
                exac_adj_allele=hl.tstr,
                exac_adj_maf=hl.tfloat,
                exac_allele=hl.tstr,
                exac_afr_allele=hl.tstr,
                exac_afr_maf=hl.tfloat,
                exac_amr_allele=hl.tstr,
                exac_amr_maf=hl.tfloat,
                exac_eas_allele=hl.tstr,
                exac_eas_maf=hl.tfloat,
                exac_fin_allele=hl.tstr,
                exac_fin_maf=hl.tfloat,
                exac_maf=hl.tfloat,
                exac_nfe_allele=hl.tstr,
                exac_nfe_maf=hl.tfloat,
                exac_oth_allele=hl.tstr,
                exac_oth_maf=hl.tfloat,
                exac_sas_allele=hl.tstr,
                exac_sas_maf=hl.tfloat,
                id=hl.tstr,
                minor_allele=hl.tstr,
                minor_allele_freq=hl.tfloat,
                phenotype_or_disease=hl.tint32,
                pubmed=hl.tarray(hl.tint32),
                sas_allele=hl.tstr,
                sas_maf=hl.tfloat,
                somatic=hl.tint32,
                start=hl.tint32,
                strand=hl.tint32)),
            context=hl.tstr,
            end=hl.tint32,
            id=hl.tstr,
            input=hl.tstr,
            intergenic_consequences=hl.tarray(hl.tstruct(allele_num=hl.tint32,
                                                         consequence_terms=hl.tarray(hl.tstr),
                                                         impact=hl.tstr,
                                                         minimised=hl.tint32,
                                                         variant_allele=hl.tstr)),
            most_severe_consequence=hl.tstr,
            motif_feature_consequences=hl.tarray(hl.tstruct(allele_num=hl.tint32,
                                                            consequence_terms=hl.tarray(hl.tstr),
                                                            high_inf_pos=hl.tstr,
                                                            impact=hl.tstr,
                                                            minimised=hl.tint32,
                                                            motif_feature_id=hl.tstr,
                                                            motif_name=hl.tstr,
                                                            motif_pos=hl.tint32,
                                                            motif_score_change=hl.tfloat,
                                                            strand=hl.tint32,
                                                            variant_allele=hl.tstr)),
            regulatory_feature_consequences=hl.tarray(hl.tstruct(allele_num=hl.tint32,
                                                                 biotype=hl.tstr,
                                                                 consequence_terms=hl.tarray(hl.tstr),
                                                                 impact=hl.tstr,
                                                                 minimised=hl.tint32,
                                                                 regulatory_feature_id=hl.tstr,
                                                                 variant_allele=hl.tstr)),
            seq_region_name=hl.tstr,
            start=hl.tint32,
            strand=hl.tint32,
            transcript_consequences=hl.tarray(hl.tstruct(allele_num=hl.tint32,
                                                         amino_acids=hl.tstr,
                                                         biotype=hl.tstr,
                                                         canonical=hl.tint32,
                                                         ccds=hl.tstr,
                                                         cdna_start=hl.tint32,
                                                         cdna_end=hl.tint32,
                                                         cds_end=hl.tint32,
                                                         cds_start=hl.tint32,
                                                         codons=hl.tstr,
                                                         consequence_terms=hl.tarray(hl.tstr),
                                                         distance=hl.tint32,
                                                         domains=hl.tarra(hl.tstruct(db=hl.tstr,
                                                                                     name=hl.tstr)),
                                                         exon=hl.tstr,
                                                         gene_id=hl.tstr,
                                                         gene_pheno=hl.tint32,
                                                         gene_symbol=hl.tstr,
                                                         gene_symbol_source=hl.tstr,
                                                         hgnc_id=hl.tstr,
                                                         hgvsc=hl.tstr,
                                                         hgvsp=hl.tstr,
                                                         hgvs_offset=hl.tint32,
                                                         impact=hl.tstr,
                                                         intron=hl.tstr,
                                                         lof=hl.tstr,
                                                         lof_flags=hl.tstr,
                                                         lof_filter=hl.tstr,
                                                         lof_info=hl.tstr,
                                                         minimised=hl.tint32,
                                                         polyphen_prediction=hl.tstr,
                                                         polyphen_score=hl.tfloat,
                                                         protein_end=hl.tint32,
                                                         protein_start=hl.tint32,
                                                         protein_id=hl.tstr,
                                                         sift_prediction=hl.tstr,
                                                         sift_score=hl.tfloat,
                                                         strand=hl.tint32,
                                                         swissprot=hl.tstr,
                                                         transcript_id=hl.tstr,
                                                         trembl=hl.tstr,
                                                         uniparc=hl.tstr,
                                                         variant_allele=hl.tstr)),
            variant_class=hl.tstr)
```

jigold:
@danking Can you give this another look from a **high level**. I haven't tested it at all and can flesh out the smaller stuff later on.

hail/python/hail/methods/qc.py
```diff
@@ -491,13 +507,335 @@ def n_discordant(counter):
...
+
+    @staticmethod
+    def from_dict(config: Dict[str, Any]):
+        return VEPConfig(
+            config['data_bucket'],
+            config['regions'],
+            config['image'],
+            config['data_mount'],
+            config['env'],
+            config['vep_json_typ'],
```
danking, line 643: Do we have a standard way to serialize types? If not, we should have one, but you'll need to do some post-processing here.

hail/python/hail/methods/qc.py
```diff
@@ -491,13 +507,335 @@ def n_discordant(counter):
...
+                 tolerate_parse_error: bool):
+    reference_genome = ht['locus'].dtype.reference_genome.name
+    if config is not None:
+        vep_config = VEPConfig.from_dict(config)
+    else:
+        vep_config = supported_vep_config(backend, reference_genome, regions=regions)
+
+    token = secret_alnum_string(16)
+    vep_input_path = hl.TemporaryDirectory(prefix=f'vep/inputs/{token}.vcf')
+    vep_output_path = hl.TemporaryDirectory(prefix=f'vep/outputs/{token}')
```
danking, line 713: The token is unnecessary, temporary directory will get you a unique location using secret_alnum_string.

danking (commented):
Yeah this seems like the right direction.

build.yaml
```diff
@@ -2172,8 +2194,14 @@ steps:
       export HAIL_TEST_STORAGE_URI={{ global.test_storage_uri }}/{{ token }}
       export HAIL_TEST_RESOURCES_DIR="{{ global.test_storage_uri }}/{{ upload_test_resources_to_blob_storage.token }}/test/resources/"
       export HAIL_DOCTEST_DATA_DIR="{{ global.test_storage_uri }}/{{ upload_test_resources_to_blob_storage.token }}/doctest/data/"
+      export HAIL_GENETICS_QOB_VEP_GRCH37_IMAGE={{ qob_vep_image_grch37.image }}
+      export HAIL_GENETICS_QOB_VEP_GRCH38_IMAGE={{ qob_vep_image_grch38.image }}
       export GOOGLE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
 
+      if [[ "$HAIL_CLOUD" = "gcp" ]] then
```
jigold, line 2201: @danking  Can I get some help here for why this code keeps failing?

https://ci.hail.is/batches/6813943/jobs/107

jigold: Nevermind. Figured it out.

jigold:
@danking This should be really close to having the tests passing. The only other thing left to do once it passes in GCP is to copy the data over to Azure and make a default Azure configuration.

hail/python/hail/methods/qc.py
tpoterba: not possible. However, it's incorrect to use `key='variant'` in import_table -- that should get optimized away, but as written this keys the table by the variant string, which is definitely not sorted.

hail/python/hail/methods/qc.py
```diff
@@ -492,13 +514,363 @@ def n_discordant(counter):
...
+                if 'error' in job_status['status']:
+                    job_status['status']['error'] = yamlx.dump(job_status['status']['error'].strip())
+            logs = failing_job.log()
+            for k in logs:
+                logs[k] = yamlx.dump(logs[k].strip())
+            message = {'batch_status': status,
+                       'job_status': job_status,
+                       'log': logs}
+            log.error(yamlx.dump(message))
+            raise FatalError(message)
```
danking, line 842: I think we should not log and raise the same message. It will appear twice and overwhelm the user. Let's just raise.

I believe I've fixed the issues with needing to `.strip()` strings. I think this whole thing can now be:
```
message = {
    'batch_status': status,
    'job_status': failing_job.status(),
    'log': failing_job.log()
}
raise FatalError(yamlx.dump(message))
```

hail/python/hail/methods/qc.py
```diff
@@ -492,13 +514,363 @@ def n_discordant(counter):
...
+    reference_genome = ht.locus.dtype.reference_genome.name
+    annotations = annotations.key_by(**hl.parse_variant(annotations.variant, reference_genome=reference_genome))
+
+    if csq:
+        with hl.hadoop_open(f'{vep_output_path.name}/csq-header') as f:
+            vep_csq_header = f.read().rstrip()
+    else:
+        vep_csq_header = ''
+
+    annotations = annotations.annotate_globals(vep_csq_header=vep_csq_header)
```
danking, line 861: This seems simpler:

```suggestion
    if csq:
        with hl.hadoop_open(f'{vep_output_path.name}/csq-header') as f:
            vep_csq_header = f.read().rstrip()
        annotations = annotations.annotate_globals(vep_csq_header=vep_csq_header)
```

hail/python/test/hail/methods/test_qc.py
```diff
@@ -285,3 +288,47 @@ def test_charr(self):
...
+
+    @skip_unless_service_backend()
+    @set_gcs_requester_pays_configuration(GCS_REQUESTER_PAYS_PROJECT)
+    def test_vep_grch37_consequence_false(self):
+        mt = hl.import_vcf(resource('sample.gnomad.exomes.r2.1.1.sites.chr1.vcf.gz'), reference_genome='GRCh37', force=True)
+        hail_vep_result = hl.vep(mt, csq=False)
+        ht = hail_vep_result.rows()
+        ht = ht.select(variant_class=ht.vep.variant_class)
+        result = ht.head(1).collect()[0]
+        assert result.variant_class == 'SNV', result
```
danking, line 324: We need to verify more than just this. At the very least we should write out the expected results as another file, manually inspect that it looks good, check that in, and assert `result._same(checked_in_results)`.

docker/vep/Dockerfile.GRCh37
```diff
@@ -0,0 +1,59 @@
...
+    libmysqlclient-dev \
+    libpng-dev \
+    libssl-dev \
+    locales \
+    manpages \
+    mysql-client \
+    openssl \
+    perl \
+    perl-base \
+    python3.8 \
```
danking, line 23: Why python3.8? This is gonna pull in a bunch of dependencies. hail_ubuntu already has python3.7.

docker/vep/Dockerfile.GRCh37
```diff
@@ -0,0 +1,59 @@
...
+    unzip ${VEP_VERSION}.zip && \
+    cd ensembl-tools-release-${VEP_VERSION}/scripts/variant_effect_predictor/ && \
+    perl INSTALL.pl -a a --NO_HTSLIB && \
+    echo "#!/bin/bash" > /vep/vep && \
+    echo "export PERL5LIB=\$PERL5LIB:${VEP_DIR}/loftee" >> /vep/vep && \
+    echo "exec perl ${VEP_DIR}/ensembl-tools-release-$VEP_VERSION/scripts/variant_effect_predictor/variant_effect_predictor.pl \"\$@\"" >> /vep/vep && \
+    chmod +x /vep /vep/vep && \
+    git clone -b v1.0.2 https://github.com/konradjk/loftee.git
+
+ENV PERL5LIB "${PERL5LIB}:${VEP_DIR}/loftee"
```
danking, line 57: We set this in the script above, why do we also set it here?

docker/vep/Dockerfile.GRCh38
```diff
@@ -0,0 +1,56 @@
...
+    perl INSTALL.pl -a a --NO_UPDATE && \
+    echo "#!/bin/bash" > /vep/vep && \
+    echo "export PERL5LIB=\$PERL5LIB:${VEP_DIR}/loftee" >> /vep/vep && \
+    echo "exec perl ${VEP_DIR}/ensembl-vep/vep \"\$@\"" >> /vep/vep && \
+    chmod +x /vep /vep/vep && \
+    git clone -b v1.0.2 https://github.com/konradjk/loftee.git && \
+    git clone https://github.com/Ensembl/VEP_plugins.git Plugins/ && \
+    mv loftee/* Plugins/
+
+ENV PERL5LIB "${PERL5LIB}:${VEP_DIR}/Plugins/loftee"
```
danking, line 54: same question here.

docker/vep/Dockerfile.GRCh38
```diff
@@ -0,0 +1,56 @@
+FROM {{ hail_ubuntu_image.image }}
+
+ENV VEP_VERSION 95
+ENV DEBIAN_FRONTEND noninteractive
```
danking, line 4: Hail ubuntu already sets this. Same for grch37

docker/vep/Dockerfile.GRCh37
jigold: Not sure. I grabbed this code from you. I'll try taking it out and see if it works.

hail/python/test/hail/methods/test_qc.py
jigold: My reasoning for this test being so simple was it's more to make sure the schemas are correct when csq=True. The other tests with csq=False tests the output is correct with the existing gnomAD data. If you feel we need both tests to test the actual VEP functionality then I can go ahead and add a file with whatever the current output is.

docker/vep/Dockerfile.GRCh38
jigold: See above.

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
```diff
@@ -707,7 +707,7 @@ class EmitClassBuilder[C](
     val literalsBc = if (hasLiterals)
```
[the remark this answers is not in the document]
jigold: I don't think so unless @tpoterba's fixes already went in for this error when exporting a VCF by partition.

build.yaml
```diff
@@ -2159,14 +2181,22 @@ steps:
...
+          export GCS_REQUESTER_PAYS_PROJECT=broad-ctsa
+      fi
+
       export HAIL_SHUFFLE_MAX_BRANCH=4
       export HAIL_SHUFFLE_CUTOFF=1000000
 
       hailctl config set query/backend batch
       hailctl config set batch/billing_project test
       hailctl config set batch/remote_tmpdir {{ global.test_storage_uri }}
+      hailctl config set domain {{ global.domain }}
```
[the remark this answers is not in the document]
jigold: I was using the `configuration_of` function to get the domain and I don't think this uses the deploy config. Is there something better to use?

hail/scripts/deploy.sh
```diff
@@ -18,11 +18,15 @@ WHEEL=$5
 GITHUB_OAUTH_HEADER_FILE=$6
 HAIL_GENETICS_HAIL_IMAGE=$7
 HAIL_GENETICS_HAILTOP_IMAGE=$8
-WHEEL_FOR_AZURE=$9
-WEBSITE_TAR=${10}
+QOB_VEP_GRCH37_IMAGE=$9
+QOB_VEP_GRCH38_IMAGE=$10
+WHEEL_FOR_AZURE=$11
+WEBSITE_TAR=${12}
```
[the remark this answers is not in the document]
jigold: My editor was showing an error for double digit numbers without curly braces. I fixed all the double digit numbers to have curly braces.

danking:
@jigold upon review of the flags PR, I don't see why this would depend on that; do you recall the issue?

jigold:
I was playing whack-a-mole with randomness tests and adding the GCS_REQUESTER_PAYS_BUCKET flag and the flags interface and testing infrastructure was making it impossible for the VEP tests to pass or the other tests for randomness that were failing. I'll revisit this PR again next week to see if your flags PR actually fixed the issue.

danking:
Agreed on dataproc. What do you mean by other things that need to be added? It seems to me that no review comments should affect VEP version / output so the dataproc "gold standard" output shouldn't change.

jigold:
Sorry -- I meant is comparing to dataproc sufficient or are there any other checks needed? I don't think so, but wanted to make sure as spinning up a dataproc cluster is a bit of work for me who's not a Hail expert anymore.

jigold:
I added the cleanup of old PR cached VEP images to the Archive Registry Cloud Run cleanup functions.

danking:
Ah, I understand. Yeah, I think comparing to dataproc is a sufficient test. A handful of variants is probably fine. It looks like we at one point had test data, but John C commented that out when he made VEP requester pays. I'm not sure why. Look in hail/python/cluster-tests.

jigold:
@danking Service tests all passed. I think this is ready for review!

danking:
In the good meaning of that phrase!!

danking:
AHH!
```
[gw1] PASSED test/hail/methods/test_qc.py::Tests::test_vep_grch37_against_dataproc 
```

hail/scripts/deploy.sh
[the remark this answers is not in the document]
danking: Huh, fascinating.

hail/scripts/deploy.sh
```diff
@@ -18,11 +18,15 @@ WHEEL=$5
 GITHUB_OAUTH_HEADER_FILE=$6
 HAIL_GENETICS_HAIL_IMAGE=$7
 HAIL_GENETICS_HAILTOP_IMAGE=$8
-WHEEL_FOR_AZURE=$9
-WEBSITE_TAR=${10}
+QOB_VEP_GRCH37_IMAGE=$9
+QOB_VEP_GRCH38_IMAGE=${10}
```
danking, line 22: Let's use HAIL_GENETICS_VEP_85_GRCH37_IMAGE and likewise for consistency with the other images.

hail/python/test/hail/methods/test_qc.py
```diff
@@ -285,3 +289,105 @@ def test_charr(self):
...
+        hail_vep_result = hail_vep_result.annotate_rows(vep=hail_vep_result.vep.annotate(
+            input=hl.str('\t').join([hail_vep_result.locus.contig, hl.str(hail_vep_result.locus.position), ".", hail_vep_result.alleles[0], hail_vep_result.alleles[1], ".", ".", "GT"])
+        ))
+        hail_vep_result = hail_vep_result.rows().select('vep')
+
+        def reorder_lof_info(ht):
+            return ht.annotate(vep=ht.vep.annotate(
+                transcript_consequences=hl.map(lambda csq: csq.annotate(
+                    lof_info=hl.if_else(csq.lof_info == 'null', hl.null(hl.tdict(hl.tstr, hl.tstr)), hl.dict(
+                        hl.map(lambda kv: hl.tuple([kv.split(":")[0], kv.split(":")[1]]), csq.lof_info.split(','))))),
```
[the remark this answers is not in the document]
danking: Don't duplicate the `kv.split(":")`. `hl.tuple` can accept a list, e.g.:
```python3
In [2]: import hail as hl
   ...: hl.eval(hl.tuple("a:b".split(":")))
Out[2]: ('a', 'b')
```
so we can just do:
```python3
hl.dict(csq.lof_info.split(',').map(lambda kv: hl.tuple(kv.split(':')[:2])))
```

hail/python/test/hail/methods/test_qc.py
```diff
@@ -285,3 +289,105 @@ def test_charr(self):
...
+                                               ht.vep.transcript_consequences)))
```
danking, line 354: Let's also get rid of this `hl.map` its too hard to see which argument matches function.

```python3
transcript_consequences=ht.vep.transcript_consequences.map(
    lambda csq: csq.annotate(
        lof_info=...
    )
)
```

danking (changes requested):

build.yaml
```diff
@@ -2258,8 +2280,15 @@ steps:
...
       export HAIL_TEST_RESOURCES_DIR="{{ global.test_storage_uri }}/{{ upload_test_resources_to_blob_storage.token }}/test/resources/"
       export HAIL_DOCTEST_DATA_DIR="{{ global.test_storage_uri }}/{{ upload_test_resources_to_blob_storage.token }}/doctest/data/"
+      export HAIL_GENETICS_QOB_VEP_GRCH37_IMAGE={{ qob_vep_image_grch37.image }}
+      export HAIL_GENETICS_QOB_VEP_GRCH38_IMAGE={{ qob_vep_image_grch38.image }}
       export GOOGLE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
 
+      if [[ "$HAIL_CLOUD" = "gcp" ]]
+      then
+          export GCS_REQUESTER_PAYS_PROJECT=broad-ctsa
+      fi
```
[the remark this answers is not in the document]
daniel-goldstein: Ya if it's something that the Hail System provides and not the Hail Team provides, I think it should be in our bootstrap process and part of the GCP project that runs the hail system. Worth looking into whether we can get this into terraform but an easier first step might just be adding a script as part of the bootstrap process that uploads this into an appropriate bucket (that is provisioned with terraform). It would be even better if our terraform could use the `http` data provider to do the copy, but I have a feeling that would involve storing the data in the terraform state which sounds not good.

> Regarding (2), I think AUS is using Pulumi. Is it gonna be a pain for them if we start managing infra using terraform but they're using Pulumi?

They run our terraform, but have started using Pulumi for other infrastructural things of theirs. FWIW, I think Pulumi uses terraform, so maybe they use a blend of both.

> In the short term, is our general process this?
Manually make the infra change.
Email AUS & MSFT with a description of the necessary change and the commit that relies on it.

Yes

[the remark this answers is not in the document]
jigold: > Worth looking into whether we can get this into terraform but an easier first step might just be adding a script as part of the bootstrap process that uploads this into an appropriate bucket (that is provisioned with terraform).

I agree with Daniel, but I think we should also consider the Terra use case as well. I'll continue to think about this some more.

hail/python/test/hail/methods/test_qc.py
```diff
@@ -285,3 +289,105 @@ def test_charr(self):
...
+    @set_gcs_requester_pays_configuration(GCS_REQUESTER_PAYS_PROJECT)
+    def test_vep_grch37_consequence_true(self):
+        gnomad_vep_result = hl.import_vcf(resource('sample.gnomad.exomes.r2.1.1.sites.chr1.vcf.gz'), reference_genome='GRCh37', force=True)
+        hail_vep_result = hl.vep(gnomad_vep_result, csq=True)
+
+        assert gnomad_vep_result.select_rows(vep=gnomad_vep_result.info.vep.map(lambda x: x.split('|')[:8])).rows()._same(
+            hail_vep_result.select_rows(vep=hail_vep_result.vep.map(lambda x: x.split('|')[:8])).rows().drop('vep_csq_header'))
+
+        vep_csq_header = hl.eval(hail_vep_result.vep_csq_header)
+        assert 'Consequence annotations from Ensembl VEP' in vep_csq_header, vep_csq_header
```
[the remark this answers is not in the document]
jigold: I'm not sure. I didn't really understand the purpose of csq=True. Maybe @konradjk knows why there's this second code path?

docker/vep/Dockerfile.GRCh38
```diff
@@ -0,0 +1,69 @@
...
+    wget https://github.com/ucscGenomeBrowser/kent/archive/v335_base.tar.gz && \
+    tar xzf v335_base.tar.gz && \
+    cd $KENT_SRC/lib && \
+    echo 'CFLAGS="-fPIC"' > ../inc/localEnvironment.mk && \
+    make clean && make && \
+    cd ../jkOwnLib && \
+    make clean && make && \
+    mkdir -p $VEP_DIR/cpanm && \
+    export PERL5LIB=\$PERL5LIB:$HOME/cpanm/lib/perl5 && \
+    cpanm Bio::DB::BigFile
```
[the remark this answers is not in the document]
jigold: I'm not sure. @konradjk do you know if we can simplify this at all?

docker/vep/vep.py
```diff
@@ -0,0 +1,220 @@
...
+                data = f'{header}{"".join(block)}'
+
+                stdout, stderr = proc.communicate(data)
+                if stderr:
+                    print(stderr)
```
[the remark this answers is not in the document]
jigold: I was trying to keep the exact same behavior as our current VEP implementation in Scala. I think we should check with @konradjk on what he would prefer we do when VEP prints stderr.

docker/vep/vep.py
```diff
@@ -0,0 +1,220 @@
...
+                        if consequence:
+                            vep_v = Variant.from_vcf_line(line)
+                            orig_v_str = non_star_to_orig_variants.get(str(vep_v))
+                            orig_v = Variant.from_string(orig_v_str)
+
+                            if orig_v is not None:
+                                x = CONSEQUENCE_REGEX.findall(line)
+                                if x:
+                                    first: str = x[0]
+                                    result = (orig_v, json.dumps(first[4:].split(',')), proc_id)
```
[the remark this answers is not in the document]
jigold: No idea. I copied it verbatim from the existing Scala implementation. https://github.com/hail-is/hail/blob/main/hail/src/main/scala/is/hail/methods/VEP.scala#L197

docker/vep/vep.py
```diff
@@ -0,0 +1,220 @@
...
+                                    continue
+                                raise Exception(msg) from e
+                            else:
+                                variant_string = jv.get('input')
+                                if variant_string is None:
+                                    raise ValueError(f'VEP generated null variant string\n'
+                                                     f'json: {line}\n'
+                                                     f'parsed: {jv}')
+                                v = Variant.from_vcf_line(variant_string)
+                                orig_v_str = non_star_to_orig_variants.get(str(v))
```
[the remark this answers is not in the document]
jigold: Not sure. Copied from the existing Scala implementation.

hail/python/hail/methods/qc.py
```diff
@@ -492,13 +513,362 @@ def n_discordant(counter):
...
+           config=oneof(nullable(str), nullable(dictof(str, anytype))),
            block_size=int,
            name=str,
            csq=bool,
-           tolerate_parse_error=bool)
-def vep(dataset: Union[Table, MatrixTable], config=None, block_size=1000, name='vep', csq=False, *, tolerate_parse_error=False):
+           tolerate_parse_error=bool,
+           regions=nullable(sequenceof(str)))
+def vep(dataset: Union[Table, MatrixTable], config=None, block_size=1000, name='vep', csq=False,
+        tolerate_parse_error=False, regions=None):
```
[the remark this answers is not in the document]
jigold: I think we should allow a user to specify multiple regions to be consistent with the rest of the batch interface and now QoB `hl.init()` setting. However, we can make the documentation better here to strongly encourage setting this to just the region where the data resides.

jigold:
I think the only two things I'm stuck on are:
(a) Do we want users to pass a VEPConfig instead of a `config` dictionary (and add documentation)?
(b) What is the best way to expose the VEP command interface so a user can customize it to their setup? I wanted to do something like this, but I don't see how to do this with the bash script being called with an argument `/bin/bash -c "...." csq` or `/bin/bash -c "..." vep`.

```python3
vep_85_grch37_command = '''
#!/bin/bash

if [ $VEP_CONSEQUENCE -ne 0 ]
then
  vcf_or_json="--vcf"
else
  vcf_or_json="--json"
fi

export VEP_COMMAND=/vep/vep \
${VEP_INPUT_FILE:+--input_file $VEP_INPUT_FILE} \
--format vcf \
${vcf_or_json} \
--everything \
--allele_number \
--no_stats \
--cache \
--offline \
--minimal \
--assembly GRCh37 \
--dir=${VEP_DATA_DIR} \
--plugin LoF,human_ancestor_fa:${VEP_DATA_DIR}/loftee_data/human_ancestor.fa.gz,filter_position:0.05,min_intron_size:15,conservation_file:${VEP_DATA_DIR}/loftee_data/phylocsf_gerp.sql,gerp_file:${VEP_DATA_DIR}/loftee_data/GERP_scores.final.sorted.txt.gz \
-o STDOUT

exec vep.py "$@"
'''

supported_vep_configs = {
    ('GRCh37', 'gcp', 'us-central1', 'hail.is'): VEPConfig(
        'hail-qob-vep-grch37-us-central1',
        ['us-central1'],
        HAIL_GENETICS_QOB_VEP_GRCH37_IMAGE,
        '/vep_data/',
        {},
        VEPConfig.default_vep_json_typ,
        ["/bin/bash", "-c", vep_85_grch37_command, "vep"],
        ["/bin/bash", "-c", vep_85_grch37_command, "csq_header"],
        True,
        'gcp',
    ),
```

daniel-goldstein:
Sure thing, I think you basically want the following in the top-level Makefile instead of docker/Makefile (not that it is not PHONY):

```make
vep-grch37-image: hail-ubuntu-image
        $(eval VEP_GRCH37_IMAGE := $(DOCKER_PREFIX)/hailgenetics/vep-85-grch37:$(TOKEN))
	python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"'$$(cat hail-ubuntu-image)'"}}' vep/Dockerfile.GRCh37 vep/Dockerfile.GRCh37.out
	../docker-build.sh docker/vep Dockerfile.GRCh37.out $(VEP_GRCH37_IMAGE)
	echo $(VEP_GRCH37_IMAGE) > $@
```

Side note: I do also agree with Dan's stylistic suggestion of having a directory structure like `docker/hailgenetics/vep/grch37/Dockerfile`

daniel-goldstein:
Additionally, you'll want to add those images to [this list](https://github.com/hail-is/hail/blob/92e2699a3fb6126736aaf0ac9f608dd1ba1f030c/docker/hailgenetics/mirror_images.sh#L44) to make sure that other batch instances (azure, AUS, etc.) can automatically mirror the images that we put in dockerhub

Makefile
```diff
@@ -177,3 +177,15 @@ batch-worker-image: batch/src/main/java/is/hail/JVMEntryway.class $(SERVICES_IMA
 	python3 ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"'$$(cat hail-ubuntu-image)'"},"global":{"cloud":"$(CLOUD)"}}' batch/Dockerfile.worker batch/Dockerfile.worker.out
 	./docker-build.sh . batch/Dockerfile.worker.out $(BATCH_WORKER_IMAGE)
 	echo $(BATCH_WORKER_IMAGE) > $@
+
+vep-grch37-image: hail-ubuntu-image
+        $(eval VEP_GRCH37_IMAGE := $(DOCKER_PREFIX)/hailgenetics/vep/grch37/85/vep:$(TOKEN))
+	python3 ../ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"'$$(cat hail-ubuntu-image)'"}}' vep/grch37/85/Dockerfile vep/grch37/85/Dockerfile.out
```
daniel-goldstein, line 183:
```suggestion
	python3 ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"'$$(cat hail-ubuntu-image)'"}}' docker/vep/grch37/85/Dockerfile vep/grch37/85/Dockerfile.out
```

docker/hailgenetics/mirror_images.sh
```diff
@@ -42,6 +42,8 @@ images=(
     "python-dill:3.10-slim"
     "hail:${HAIL_PIP_VERSION}"
     "hailtop:${HAIL_PIP_VERSION}"
+    "vep/grch37/85/vep:${HAIL_PIP_VERSION}"
```
daniel-goldstein, line 45: I might be wrong, but I think the directory structure was mostly so that all our dockerfiles could just be called `Dockerfile`, and that's ultimately separate from the name of the image. I personally think the image should still be named `vep-grch37-85`

jigold:
Tests are all passing. One question that remains is should we put a manifest file with the current Hail configurations somewhere in hail-common?

hail/python/hail/methods/qc.py
```diff
@@ -492,13 +512,548 @@ def n_discordant(counter):
...
+    def vep_command(self,
+                    consequence: bool,
+                    tolerate_parse_error: bool,
+                    part_id: int,
+                    input_file: Optional[str],
+                    output_file: str) -> List[str]:
+        raise NotImplementedError
+
+
+class HailVEPConfig(BaseVEPConfig, abc.ABC):
```
danking, line 597: I'm worried folks will find it confusing that there's a thing called `BaseVEPConfig` and a thing called `HailVEPConfig` and when configuring VEP for Hail they're *not* supposed to use the thing called `HailVEPConfig`.

Why not make `HailVEPConfig` the base class (and just not have `BaseVEPConfig`)? Then you can use the standard constructor documentation to describe each argument/field and their type. You can still leave `vep_command` as an abstract method.

Is `from_dict` used at all?

hail/python/hail/methods/qc.py
```diff
@@ -492,13 +512,548 @@ def n_discordant(counter):
     return glob, per_sample.cols(), per_variant.rows()
 
 
+class BaseVEPConfig(abc.ABC):
```
danking, line 515: Just call it `VEPConfig`, I don't think `Base` conveys useful information to the *user* (although it might mean something to us).

hail/python/hail/methods/qc.py
danking: Going this route, you'd need to make the `vep_json_type` something like `GRCh37_JSON_TYPE`. This also has the advantage of exposing `vep_json_typ` through a public class rather than as a field on an undocumented and unexported class. That helps a user discover it and use it as an example from which to define their own JSON type.

danking (changes requested):
I don't understand what the manifest file is. Is it referenced in this PR somewhere?

hail/python/hail/methods/qc.py
danking: Or you could push GRCh37_JSON_TYPE down into the GRCh37 class. That also seems fine.

hail/python/hail/methods/qc.py
```diff
@@ -492,13 +512,548 @@ def n_discordant(counter):
...
+    def vep_command(self,
+                    consequence: bool,
+                    tolerate_parse_error: bool,
+                    part_id: int,
+                    input_file: Optional[str],
+                    output_file: str) -> List[str]:
+        raise NotImplementedError
+
+
+class HailVEPConfig_GRCh37_85(HailVEPConfig):
```
danking, line 761: We're in the Hail library, I don't think we need to prefix this with `Hail`. Let's call it `VEPConfigGRCh37Version85`

hail/python/hail/methods/qc.py
```diff
@@ -492,13 +512,548 @@ def n_discordant(counter):
...
+
+
+supported_vep_configs = {
+    ('GRCh37', 'gcp', 'us-central1', 'hail.is'): HailVEPConfig_GRCh37_85(
+        data_bucket='hail-qob-vep-grch37-us-central1',
+        data_mount='/vep_data/',
+        image=HAIL_GENETICS_VEP_GRCH37_85_IMAGE,
+        regions=['us-central1'],
+        cloud='gcp',
+        reference_data_is_bucket_requester_pays=True,
```
danking, line 867: I think you want the `is` after `bucket`, yeah?

danking: Should it just be `data_bucket_is_requester_pays`? That's the name of the relevant config parameter.

hail/python/hail/methods/qc.py
```diff
@@ -492,13 +513,362 @@ def n_discordant(counter):
...
+                 config: Optional[Dict[str, Any]],
+                 regions: Optional[List[str]],
+                 block_size: int,
+                 csq: bool,
+                 tolerate_parse_error: bool):
+    reference_genome = ht['locus'].dtype.reference_genome.name
+    cloud = backend.bc.cloud()
+
+    if regions is None:
+        regions = backend.regions
```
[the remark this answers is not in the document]
jigold: Should we get rid of regions here all together here then and you have to specify the regions when you initialize the backend?

hail/scripts/deploy.sh
```diff
@@ -8,7 +8,7 @@ retry() {
         (sleep 5 && "$@");
 }
 
-[[ $# -eq 10 ]] || (echo "./deploy.sh HAIL_PIP_VERSION HAIL_VERSION GIT_VERSION REMOTE WHEEL GITHUB_OAUTH_HEADER_FILE HAIL_GENETICS_HAIL_IMAGE HAIL_GENETICS_HAILTOP_IMAGE WHEEL_FOR_AZURE WEBSITE_TAR" ; exit 1)
+[[ $# -eq 12 ]] || (echo "./deploy.sh HAIL_PIP_VERSION HAIL_VERSION GIT_VERSION REMOTE WHEEL GITHUB_OAUTH_HEADER_FILE HAIL_GENETICS_HAIL_IMAGE HAIL_GENETICS_HAILTOP_IMAGE QOB_VEP_GRCH37_IMAGE QOB_VEP_GRCH38_IMAGE WHEEL_FOR_AZURE WEBSITE_TAR" ; exit 1)
```
danking, line 11: update `hail/Makefile`'s deploy target as well

hail/scripts/deploy.sh
```diff
@@ -75,6 +79,10 @@ retry skopeo copy $HAIL_GENETICS_HAIL_IMAGE docker://docker.io/hailgenetics/hail
 retry skopeo copy $HAIL_GENETICS_HAIL_IMAGE docker://us-docker.pkg.dev/hail-vdc/hail/hailgenetics/hail:$HAIL_PIP_VERSION
 retry skopeo copy $HAIL_GENETICS_HAILTOP_IMAGE docker://docker.io/hailgenetics/hailtop:$HAIL_PIP_VERSION
 retry skopeo copy $HAIL_GENETICS_HAILTOP_IMAGE docker://us-docker.pkg.dev/hail-vdc/hail/hailgenetics/hailtop:$HAIL_PIP_VERSION
+retry skopeo copy $HAIL_GENETICS_VEP_GRCH37_85_IMAGE docker://docker.io/hailgenetics/vep-grch37-85:$HAIL_PIP_VERSION
+retry skopeo copy $HAIL_GENETICS_VEP_GRCH37_85_IMAGE docker://us-docker.pkg.dev/hail-vdc/hail/hailgenetics/vep-grch37-85:$HAIL_PIP_VERSION
+retry skopeo copy $HAIL_GENETICS_VEP_GRCH38_95_IMAGE docker://docker.io/hailgenetics/vep-grch38-95:$HAIL_PIP_VERSION
+retry skopeo copy $HAIL_GENETICS_VEP_GRCH38_95_IMAGE docker://us-docker.pkg.dev/hail-vdc/hail/hailgenetics/vep/grch38-95:$HAIL_PIP_VERSION
```
danking, line 85: Add `vep-grch38-95` and `vep-grch37-85` to HAIL_GENETICS_IMAGES in hail/python/hailtop/batch/hail_genetics_images.py

danking (changes requested):

jigold:
Tests are passing again! I know there were concerns about error handling in the `vep.py` code. I tried to make it exactly the same as the existing Scala code.

PR #12432, opened 2022-11-07
[ci] Remove hail-base image in favor of hailgenetics/hail

daniel-goldstein:
We should lean more heavily into using our published images for testing instead of having different images that install hail in different ways that are only used for tests.


build.yaml
```diff
@@ -1329,28 +1309,39 @@ steps:
...
+      - hail_run_image
+      - build_hail
+      - merge_code
     clouds:
       - gcp
   - kind: runImage
     name: make_docs
     image:
-      valueFrom: hail_base_image.image
+      valueFrom: hail_run_image.image
```
danking, line 1337: Hail is not installed in the `hail_run_image`, I think you can just use hailgenetics/hail here as well if you install the dev dependencies.

danking (changes requested):

daniel-goldstein:
This will no doubt be wrecked by the spicy meatball, will come back to these after that merges.

PR #12433, opened 2022-11-08
[hailgenetics] Use fully-pinned requirements in hailgenetics/hail

daniel-goldstein:
This installs our fully-pinned requirements deep in the docker image and then installs the hail wheel without dependencies on top. This will be a lot more consistent (and docker cache friendly) than the current approach, which will install all dependencies with the wheel and possibly upgrade some of them.

Also did the same to the hail-pip-installed images used for testing.


daniel-goldstein:
unassigning while I figure out dependency issues.

danking (approved):

PR #12444, opened 2022-11-09
[infra] Azure container registry cleanup script

jigold:
I have the dry-run flag on as default just to test it first. Then I just manually removed that line to actually delete things. I think we could potentially make this nicer with a flag, but I'm not sure how to do that easily.


daniel-goldstein (changes requested):
Seems fine, happy to add an argument later for dry run. Only nit is that I think I'd prefer this in `devbin` rather than the `infra` directory.

PR #12446, opened 2022-11-09
[hailgenetics] Make hailgenetics/hailtop image

daniel-goldstein:
This creates another hailgenetics image with just hailtop and its python dependencies installed (no hail query, no java, no gcloud), primarily useful for using the copy tool. Instead of 3GiB it is less than half a GiB.

This also is the start of separating out our python dependencies into those required by separate modules. The hail dependencies now includes the hailtop dependencies. Ultimately, I want to move the services' dependencies to include those of hailtop and not all of hail (getting pyspark out of our services images).


hail/scripts/deploy.sh
```diff
@@ -26,10 +27,12 @@ REMOTE=$4
 WHEEL=$5
 GITHUB_OAUTH_HEADER_FILE=$6
 HAIL_GENETICS_HAIL_IMAGE=$7
+HAIL_GENETICS_HAILTOP_IMAGE=$7
```
danking, line 30: Did you intend for this to also be $7?  If you're adding a new parameter, also update all the error lines above to include the new parameter.

I think we might want to change from the current:
```
[[ $1 ]] || err
[[ $2 ]] || err
...
```
to
```
[ $# -eq 10 ] || err
```

danking: I guess that doesn't verify all arguments are non-empty. Maybe we should do:
```
[[ -n $0 && -n $1 && -n $2 && ... ]] || err
```

danking: It seems a little unnecessary to defend against someone doing `./deploy.sh '' '' '' '' ...`

hail/scripts/deploy.sh
```diff
@@ -80,6 +83,8 @@ curl -XPOST -H @$GITHUB_OAUTH_HEADER_FILE https://api.github.com/repos/hail-is/h
 
 retry skopeo copy $HAIL_GENETICS_HAIL_IMAGE docker://docker.io/hailgenetics/hail:$HAIL_PIP_VERSION
 retry skopeo copy $HAIL_GENETICS_HAIL_IMAGE docker://us-docker.pkg.dev/hail-vdc/hail/hailgenetics/hail:$HAIL_PIP_VERSION
+retry skopeo copy $HAIL_GENETICS_HAILTOP_IMAGE docker://docker.io/hailgenetics/hailtop:$HAIL_PIP_VERSION
+retry skopeo copy $HAIL_GENETICS_HAILTOP_IMAGE docker://us-docker.pkg.dev/hail-vdc/hail/hailgenetics/hailtop:$HAIL_PIP_VERSION
```
danking, line 87: I am suddenly wondering: how do we get these images into Azure?

danking (dismissed):

hail/scripts/deploy.sh
daniel-goldstein: We don't. I want to run the `mirror_images.sh` script that I made recently as part of the CI process, maybe we could do that as part of #12378? I got stuck on the fact that we don't have a great way to synchronize when a release happens on GCP CI and azure CI grabbing the images from DockerHub, but every deploy should at least be much better than nothing.

hail/scripts/deploy.sh
daniel-goldstein: ya.. I honestly feel like `[ $# -eq 10 ] || err` hits a good balance. The boilerplate now feels more footgun-y than the last example you mention

daniel-goldstein: as evidenced by my totally messing this up.

hail/scripts/deploy.sh
danking: Or Azure could have a release process which also generates these common files? Why not do that?

hail/python/setup.py
```diff
@@ -19,12 +19,24 @@
...
+                continue
+            if stripped.startswith('-r'):
+                additional_requirements = stripped[len('-r'):].strip()
+                add_dependencies(additional_requirements)
+                continue
+            pkg = stripped
+            if pkg.startswith('pyspark') and os.path.exists('../env/SPARK_VERSION'):
+                with open('../env/SPARK_VERSION', 'r') as file:
+                    spark_version = file.read()
+                dependencies.append(f'pyspark=={spark_version}')
```
danking, line 36: why shouldn't we error if SPARK_VERSION is missing?

danking (dismissed):

hail/python/setup.py
daniel-goldstein: Great question, though this isn't a new change just moved code. Can I make a separate tiny pr to nix that check?

danking (approved):
separate PR seems good

hail/scripts/deploy.sh
daniel-goldstein: ah, duh, that is a good point

daniel-goldstein: For some reason I always think of it as strictly mirroring dockerhub and we only let one CI create the images that go to dockerhub..

hail/python/requirements.txt
```diff
@@ -1,35 +1,20 @@
-aiohttp>=3.8.1,<4
-aiohttp_session>=2.7,<2.8
+-r hailtop/requirements.txt
```
lgruen, line 1: @daniel-goldstein @danking This seems to break the `init_notebook.py` initialization action of `hailctl dataproc start`:
```
pip packages are ['setuptools', 'mkl<2020', 'lxml<5', 'google-cloud-storage==1.25.*', 'https://github.com/hail-is/jgscm/archive/v0.1.12+hail.zip', 'ipykernel==4.10.*', 'ipywidgets==7.5.*', 'jupyter-console==6.0.*', 'nbconvert==5.6.*', 'notebook==5.7.*', 'qtconsole==4.5.*', 'sklearn', 'slackclient', 'google', 'cpg-utils', '-r hailtop/requirements.txt', 'asyncinit>=0.2.4,<0.3', 'avro>=1.10,<1.12', 'bokeh==1.4.0', 'decorator<5', 'Deprecated>=1.2.10,<1.3', 'dill>=0.3.1.1,<0.4', 'frozenlist>=1.3.1,<2', 'hurry.filesize>=0.9,<1', 'Jinja2==3.0.3', 'nest_asyncio>=1.5.4,<2', 'numpy<2', 'pandas>=1.3.0,<1.5.0', 'parsimonious<0.9', 'plotly>=5.5.0,<5.11', 'protobuf==3.20.2', 'PyJWT', 'requests>=2.25.1,<3', 'scipy>1.2,<1.10', 'selenium', 'click', 'fsspec', 'gcloud', 'gnomad']
ERROR: Could not open requirements file: [Errno 2] No such file or directory: ' hailtop/requirements.txt'
```

vladsavelyev: Our proposed quick fix :) https://github.com/populationgenomics/hail/pull/250

vladsavelyev: Our proposed fix: https://github.com/populationgenomics/hail/pull/250/files

daniel-goldstein: Ah thanks for spotting this @lgruen and @vladsavelyev and sorry it broke the installation! I'll have to look into why the dataproc test passed.

Looks like you found the root cause, but I think a more proper fix would be to use `python/pinned-requirements.txt` instead of `python/requirements.txt`. The `pinned-requirements.txt` file should include the full set of requirements from hailtop and the rest of the hail package. Moreover, those will be fully pinned and tested versions that pass CI instead of some possibly new versions you get from a pip install. Does using that file instead of also including the hailtop requirements work?

daniel-goldstein: I've created a PR over at #12510 to try to get a fix in before an upcoming release.

PR #12447, opened 2022-11-10
[mypy] Upgrade mypy and make suggested fixes

daniel-goldstein:
These changes are mostly because mypy now needs all `None` parameters to be explicit optionals, whereas before it would let something like `def foo(arg: str = None)` slide. The `type: ignore` on the azure client call is an azure library bug, I looked into the source for that function and there's a comment saying that the input is an `Optional[int]` but the type stubs that mypy is picking up say `int`.

the `check_untyped_defs` config flag tells mypy to typecheck the bodies of functions that do not have a type signature, which apparently it wasn't doing on its own.

### DIFF
--- a/hail/python/Makefile
+++ b/hail/python/Makefile
@@ -1,4 +1,4 @@
-PYTHON := python3
+PYTHON := python
 
 .PHONY: check-hail
 check-hail:
--- a/hail/python/Makefile
+++ b/hail/python/Makefile
@@ -1,4 +1,4 @@
-PYTHON := python
+PYTHON := python3
 
 .PHONY: check-hail
 check-hail:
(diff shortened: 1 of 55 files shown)


daniel-goldstein:
Sorry jumped the gun there, unassigning because there's still more linting to fix

hail/python/Makefile
```diff
@@ -1,4 +1,4 @@
-PYTHON := python3
+PYTHON := python
```
jigold, line 1: fix.

jigold (approved):

jigold (changes requested):
python -> python3

jigold:
Looks like this needs a rebase now.

jigold:
Unfortunately, needs a rebase again.

jigold (approved):

jigold:
Approving in anticipation that the rebase is just fixing the pinned requirements. Can you re-request a review if you make other changes?

PR #12457, opened 2022-11-14
[query] use native pca on service, better default params

patrick-schultz:
CHANGELOG: pca and hwe_normalized_pca now supported in Query-on-Batch


danking:
Can you force push an edit to the commit message that includes a `CHANGELOG:` line indicating that pca and hwe_normalized_pca are now supported in Query-on-Batch?

danking (approved):

PR #12461, opened 2022-11-14
[batch] Avoid calling add_attempt_resources in MJC if MJS succeeded

daniel-goldstein:
Tested in a dev deploy'd load test that # of add_attempt_resources queries == # of jobs instead of double as you can currently observe in default.


danking (approved):
This seems correct to me.

I'm curious: if we pull this thread further would we find a simpler overall implementation? In particular, if we require the worker to always MJS (rather than at most once), does the implementation of MJC (in Python and SQL) get simpler? I think its worth considering the tradeoff of code complexity vs latency saved by avoiding an MJS on very fast jobs.

danking:
cc: @jigold

PR #12463, opened 2022-11-15
[query] fix flaky test, improve test infra

danking:
This test would fail because multiple test jobs could race with the exact same prefix. Instead, we use a fresh prefix for each job and each pytest thread in the job.

I also switched from unittest to pytest so I could use a fixture. I honestly do not understand why the local tests were passing the service backend.


danking:
Ah, hadoop_* is *supposed* to work with local paths by design. `hl.hadoop_open('/foo/bar')` refers to `/foo/bar` on the client's filesystem, which is actually very reasonable.

daniel-goldstein (approved):

PR #12464, opened 2022-11-15
[qob] Use one batch for the lifetime of a hl.ServiceBackend

daniel-goldstein:
CHANGELOG: When using Query-on-Batch, all jobs for a single hail session are inserted into the same batch instead of one batch per action.

### DIFF
--- a/hail/python/hail/backend/service_backend.py
+++ b/hail/python/hail/backend/service_backend.py
@@ -1,9 +1,11 @@
 from typing import Dict, Optional, Callable, Awaitable, Mapping, Any, List, Union, Tuple
 import abc
 import struct
+import os
 from hail.expr.expressions.base_expression import Expression
 import orjson
 import logging
+from pathlib import Path
 
 from hail.context import TemporaryDirectory, tmp_dir, TemporaryFilename, revision, _TemporaryFilenameManager
 from hail.utils import FatalError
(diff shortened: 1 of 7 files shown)


daniel-goldstein:
This is pseudo-stacked on #12490 in that I made a change here to serialize `_async_execute_many` but in #12490 I remove it altogether which takes care of the complications of multiple concurrent Query Drivers in the same batch.

hail/python/hail/backend/service_backend.py
```diff
@@ -325,43 +325,50 @@ async def _rpc(self,
...
                         deploy_config = get_deploy_config()
-                        url = deploy_config.external_url('batch', f'/batches/{b.id}/jobs/1')
-                        print(f'Submitted batch {b.id}, see {url}')
-
-                    await b.wait(description=name,
-                                 disable_progress_bar=self.disable_progress_bar,
-                                 progress=progress)
+                        # FIXME Should we not `print`?
+                        url = deploy_config.external_url('batch', f'/batches/{self._batch.id}/jobs/{j.job_id}')
+                        print(f'Submitted to batch {self._batch.id}, see {url}')
```
danking, line 360: I'm somewhat inclined to not print. The progress bar should provide all necessary information

danking (commented):
LGTM

hail/python/hail/backend/service_backend.py
daniel-goldstein: Happy to axe that.

PR #12465, opened 2022-11-15
[batch] Expose region variable to users

jigold:
This adds what region the job ran into both the job environment as well as a field in the job status.

### DIFF
--- a/batch/batch/cloud/azure/driver/create_instance.py
+++ b/batch/batch/cloud/azure/driver/create_instance.py
@@ -43,6 +43,8 @@ def create_vm_config(
 ) -> dict:
     _, cores = azure_machine_type_to_worker_type_and_cores(machine_type)
 
+    region = instance_config.region_for(location)
+
     if max_price is not None and not preemptible:
         raise ValueError(f'max price given for a nonpreemptible machine {max_price}')
 
@@ -188,6 +190,7 @@ NAME=$(curl -s -H Metadata:true --noproxy "*" "http://169.254.169.254/metadata/i
 BATCH_WORKER_IMAGE=$(jq -r '.batch_worker_image' userdata)
 DOCKER_ROOT_IMAGE=$(jq -r '.docker_root_image' userdata)
 DOCKER_PREFIX=$(jq -r '.docker_prefix' userdata)
+REGION=$(jq -r '.docker_prefix' region)
 
 INTERNAL_GATEWAY_IP=$(jq -r '.internal_ip' userdata)
 
@@ -235,6 +238,7 @@ docker run \
 -e SUBSCRIPTION_ID=$SUBSCRIPTION_ID \
 -e RESOURCE_GROUP=$RESOURCE_GROUP \
 -e LOCATION=$LOCATION \
+-e REGION=$REGION \
 -e DOCKER_PREFIX=$DOCKER_PREFIX \
 -e DOCKER_ROOT_IMAGE=$DOCKER_ROOT_IMAGE \
 -e INSTANCE_CONFIG=$INSTANCE_CONFIG \
@@ -288,6 +292,7 @@ done
         'instance_id': file_store.instance_id,
         'max_idle_time_msecs': max_idle_time_msecs,
         'instance_config': base64.b64encode(json.dumps(instance_config.to_dict()).encode()).decode(),
+        'region': region,
     }
     user_data_str = base64.b64encode(json.dumps(user_data).encode('utf-8')).decode('utf-8')
 
--- a/batch/batch/cloud/azure/driver/create_instance.py
+++ b/batch/batch/cloud/azure/driver/create_instance.py
@@ -190,6 +190,7 @@ DOCKER_ROOT_IMAGE=$(jq -r '.docker_root_image' userdata)
 DOCKER_PREFIX=$(jq -r '.docker_prefix' userdata)
 
 INTERNAL_GATEWAY_IP=$(jq -r '.internal_ip' userdata)
+REGION=$(jq -r '.region' userdata)
 
 # private job network = 172.20.0.0/16
 # public job network = 172.21.0.0/16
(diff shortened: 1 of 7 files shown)


batch/batch/worker/worker.py
```diff
@@ -1488,6 +1492,8 @@ def __init__(
 
         self.timings: Timings = Timings()
 
+        self.env['HAIL_REGION'] = REGION
```
daniel-goldstein, line 1495: I think this should be 
```suggestion
        self.env.append({'name': 'HAIL_REGION', 'value': REGION})
```

batch/batch/cloud/azure/driver/create_instance.py
```diff
@@ -190,6 +190,7 @@ def create_vm_config(
 DOCKER_PREFIX=$(jq -r '.docker_prefix' userdata)
 
 INTERNAL_GATEWAY_IP=$(jq -r '.internal_ip' userdata)
+REGION=$(jq -r '.region' userdata)
```
daniel-goldstein, line 193: Probably better to grab this from the instance config more like how we do in GCP. I could see this getting missed if we added more regions to azure.

batch/test/utils.py
```diff
@@ -18,3 +18,10 @@ def smallest_machine_type(cloud):
         return 'n1-standard-1'
     assert cloud == 'azure'
     return 'Standard_D2ds_v4'
+
+
+def nondefault_region(cloud):
```
daniel-goldstein, line 23: If this is only used in one test can we just keep it in that test? Also this is a nit because I don't see us changing our default region to `us-east1`, but it does bother me slightly that it's not necessarily correct. We could make it correct by defining two regions and picking one that is not the default region.

daniel-goldstein (changes requested):
Just a couple nits

batch/batch/cloud/gcp/instance_config.py
```diff
@@ -88,6 +88,8 @@ def worker_type(self) -> str:
 
     def region_for(self, location: str) -> str:
         # location = zone
+        if location.startswith('projects'):
```
daniel-goldstein, line 91: What's this "projects" about?

batch/batch/worker/worker.py
```diff
@@ -188,11 +189,13 @@ def compose(auth: Union[MutableMapping, str, bytes], registry_addr: str = None):
 log.info(f'INTERNET_INTERFACE {INTERNET_INTERFACE}')
 log.info(f'UNRESERVED_WORKER_DATA_DISK_SIZE_GB {UNRESERVED_WORKER_DATA_DISK_SIZE_GB}')
 log.info(f'ACCEPTABLE_QUERY_JAR_URL_PREFIX {ACCEPTABLE_QUERY_JAR_URL_PREFIX}')
+log.info(f'LOCATION {LOCATION}')
 
 instance_config = CLOUD_WORKER_API.instance_config_from_config_dict(INSTANCE_CONFIG)
 assert instance_config.cores == CORES
 assert instance_config.cloud == CLOUD
 
+REGION = instance_config.region_for(LOCATION)
```
daniel-goldstein, line 198: Can we just pass REGION as the env variable here instead?

batch/test/test_batch.py
```diff
@@ -1296,3 +1296,15 @@ def test_submit_update_to_deleted_batch(client: BatchClient):
         assert err.status == 404
     else:
         assert False
+
+
+@skip_in_azure
```
daniel-goldstein, line 1301: This should also work in azure, can we make a test fixture here that passes a us-east1 to the test method in gcp and eastus in azure? This PR is more testing that the region is provided in the status not necessarily that we can successfully schedule in different regions so it still makes sense as a test on azure.

daniel-goldstein (changes requested):

batch/test/test_batch.py
daniel-goldstein: Let's also have the job command print the env variable and assert that it shows up in the job log

batch/batch/cloud/gcp/instance_config.py
jigold: The issue is the return value from using Google's metadata server is a path and not just the zone name. It's like `/projects/hail-vdc/random-hash/us-central1-a`. I can either pass the zone directly that we know when creating the instance into the startup script or I can try and do a better job integrating the parsing with the other place in the code base where we have to parse the ZONE that starts with `projects`. Preferences?

batch/batch/worker/worker.py
jigold: I thought I had that previously, but you suggested using the instance config. I could be wrong though. This question is related to the comment above about how to pass the ZONE.

batch/test/test_batch.py
jigold: I originally wanted to test it creates a VM in a different region, but I was concerned that this will continue to add to test times as it's another VM that we have to wait on to create and startup. Do you have preferences on what the test should do? I can increase the timeout on the batch tests to account for this. Also, it would have to be a highcpu instance because we've capped the number of standard instances so it will probably never be able to create a new instance before it times out.

batch/batch/worker/worker.py
daniel-goldstein: Ah sorry that's not what I meant. I meant that in your original commit I liked the gcp `create_instance.py` changes, with the caveat that you could get the region by using the `instance_config.region_for` method instead of doing a string split.

What I intended was that the environment variable given to the worker be `REGION` and that we get that value in the first line of the `create_vm_config` function by doing

```python
region = instance_config.region_for(location)
```

batch/batch/cloud/gcp/instance_config.py
daniel-goldstein: ah ya let's just use the custom userdata for this like in your first commit and not use the metadata server's endpoint.

daniel-goldstein (dismissed):

daniel-goldstein (approved):

PR #12468, opened 2022-11-15
[batch/qob] retrieves stack trace for failed worker jobs

iris-garden:
Currently, when a QoB worker job fails in a way that prevents it from writing a result file with the error's stack trace in it, the Batch worker that started the job tries to get the result file anyway, and when it fails, raises a 404. This change retrieves the stack trace from the QoB job and raises an error with the stack trace included.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -2605,6 +2605,10 @@ class JVMUserError(Exception):
     pass
 
 
+class JVMQOBError(Exception):
+    pass
+
+
 class JVMProfiler:
     def __init__(self, container: JVMContainer, output_file: Optional[str]):
         self.container = container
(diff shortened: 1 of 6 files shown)


iris-garden:
With [this code](https://github.com/hail-is/hail/blob/81b0b20ea96278ca74c17b7618d53c330a5ee15f/hail/src/main/scala/is/hail/backend/service/Worker.scala#L182) commented out, error looks like [this](https://gist.github.com/iris-garden/33e7b0508f06f7e6aa4c17bd8c4d92b7)

daniel-goldstein:
@iris-garden do you have a sense of what work remains to be done here or whether this branch is worth reviving? Just had a support issue because of a `Hail Internal Error: …. 404 …` that turned out to be the worker jobs ran out of memory

PR #12474, opened 2022-11-16
[fs] Increase write buffer capacity

jigold:


hail/src/main/scala/is/hail/io/fs/FS.scala
```diff
@@ -174,7 +174,7 @@ abstract class FSSeekableInputStream extends InputStream with Seekable {
 
 abstract class FSPositionedOutputStream extends OutputStream with Positioned {
   protected[this] var closed: Boolean = false
-  protected[this] val bb: ByteBuffer = ByteBuffer.allocate(64 * 1024)
+  protected[this] val bb: ByteBuffer = ByteBuffer.allocate(8 * 1024 * 1024)
```
daniel-goldstein, line 177: This is used for the azure fs as well, so should this optionally be 4Mb?

daniel-goldstein (changes requested):

hail/src/main/scala/is/hail/io/fs/FS.scala
jigold: I was going to see if the tests all pass with 8 MB and then make a decision then.

daniel-goldstein: Ya looks like it's yelling about the chunk limit https://ci.azure.hail.is/batches/3209370/jobs/121

daniel-goldstein (approved):

PR #12478, opened 2022-11-16
[batch] make Batch.delete idempotent

danking:
`_delete` retries transient errors. The `.../delete` endpoints return 404 for batches that do not exist, which, I suppose, is reasonable. On the client-side, we need to ignore 404s to ensure idempotency.


danking:
Seen here: https://ci.hail.is/batches/6651758/jobs/117

jigold (approved):

PR #12479, opened 2022-11-17
[query] Read builtin reference genomes from python

daniel-goldstein:


hail/python/hail/backend/service_backend.py
```diff
@@ -574,32 +573,8 @@ async def inputs(infile, _):
...
-    async def _async_get_references(self, names, *, progress: Optional[BatchProgressBar] = None):
-        if progress is None:
-            with BatchProgressBar() as progress:
-                return await asyncio.gather(*[self._async_get_reference(name, progress=progress)
-                                              for name in names])
-        else:
-            return await asyncio.gather(*[self._async_get_reference(name, progress=progress)
-                                          for name in names])
+        _, reference, _ = await self._rpc('get_reference(...)', inputs, progress=progress)
+        return orjson.loads(reference)
```
danking, line 577: This now has no effect. The ServiceBackend currently lacks support for custom references, so the get_reference RPC can only return one of the default references. I think we should eliminate all this and just raise an error that no such reference exists.

This PR is a clear improvement that I want to land posthaste.

However, we should also pull this thread fully taught: references should be owned by Python. The Scala `ReferenceGenome.references` map should just be a cache for the Java Backends (Spark and Local). The Service Backend can't use `ReferenceGenome.references` for user references anyway because the `ReferenceGenome` class is in the JAR and thus shared across users.

References should live in a map on the Python `Backend` class. The Py4JBackend should also update the Java-side when a reference is added. The ServiceBackend should send along every custom reference genome with every RPC (they're not very large, but, yes, we should find a way to avoid sending them when not necessary; not necessary for MVP).

danking (changes requested):

hail/python/hail/backend/service_backend.py
danking: cc: @tpoterba I want to move references into Python, see plan above, curious if you have any more to add.

danking (approved):
We may want to restore the local cache dir, but we can revisit that when the time comes.

PR #12484, opened 2022-11-17
[query] extensive extract intervals tests

danking:


danking:
@tpoterba this actually fails tests, am I being overly optimistic about what extract intervals can achieve? The following are the failures and they're all about the number of partitions after a filter.
```
FAILED test/hail/extract_intervals/test_full_key.py::test_mt_lt[locus] - assert 20 == 15
FAILED test/hail/extract_intervals/test_full_key.py::test_mt_le[locus] - assert 20 == 15
FAILED test/hail/extract_intervals/test_full_key.py::test_mt_ge[locus] - assert 20 == 6
FAILED test/hail/extract_intervals/test_full_key.py::test_mt_gt[locus] - assert 20 == 6
FAILED test/hail/extract_intervals/test_full_key.py::test_ht_lt[locus] - assert 20 == 15
FAILED test/hail/extract_intervals/test_full_key.py::test_ht_le[locus] - assert 20 == 15
FAILED test/hail/extract_intervals/test_full_key.py::test_ht_ge[locus] - assert 20 == 6
FAILED test/hail/extract_intervals/test_full_key.py::test_ht_gt[locus] - assert 20 == 6
FAILED test/hail/extract_intervals/test_full_key.py::test_mt_lt[Locus] - assert 20 == 15
FAILED test/hail/extract_intervals/test_full_key.py::test_mt_le[Locus] - assert 20 == 15
FAILED test/hail/extract_intervals/test_full_key.py::test_mt_ge[Locus] - assert 20 == 6
FAILED test/hail/extract_intervals/test_full_key.py::test_mt_gt[Locus] - assert 20 == 6
FAILED test/hail/extract_intervals/test_full_key.py::test_ht_lt[Locus] - assert 20 == 15
FAILED test/hail/extract_intervals/test_full_key.py::test_ht_le[Locus] - assert 20 == 15
FAILED test/hail/extract_intervals/test_full_key.py::test_ht_ge[Locus] - assert 20 == 6
FAILED test/hail/extract_intervals/test_full_key.py::test_ht_gt[Locus] - assert 20 == 6
======================================================================================================================================= 16 failed, 34 passed in 19.17s =======================================================================================================================================
```

tpoterba:
I think we added general comparison on Expression to Python after extract intervals -- not hard to add that to the match.

danking:
Hmm. OK, maybe I'm misunderstanding you, but it seems like the issue is that arrays cannot generate open intervals. Is this a limitation of how we do comparisons? In particular, I can't have an interval like `( ["A", "G"] , ["D"] )` which matches `["A", "H"]` and `["A", "G", "A"]` but not `["A", "G"]`? cc: @patrick-schultz 
```
  def canGenerateOpenInterval(t: Type): Boolean = t match {
    case _: TNumeric => true
    case TBoolean => true
    case _: TLocus => true
    case ts: TBaseStruct => ts.fields.forall(f => canGenerateOpenInterval(f.typ))
    case _ => false
  }
```

tpoterba:
Ah! Yeah, I see. This isn't super easy to fix.

danking:
> It's more a question of what interval you generate if your key is an array `['A']` and you want to the set of all keys starting with A -- there's no way to fill in an endpoint.

Hmm. If you did this,
```
mt = mt.filter_rows(mt.row_key <= ["A"])
```

I think you want this:
```
[ [], ["A"] ]
```

Your sentence above seems to correspond most closely to this query?
```
mt = mt.filter_rows((mt.row_key > [""]) & (mt.row_key < ["B"]))
```
which has this interval
```
( [""], ["B"] )
```

I admit that ordering on nested sequences is pretty painful to my brain.

tpoterba:
The left endpoint is OK -- what do you do for the last key?

danking:
Sure, it seems like we need a way to indicate the maximum string.

danking:
In particular, we're entirely incapable of doing any interval queries on any string keyed things right now, right?

tpoterba:
we can't generate them for open comparisons, that's right.

tpoterba:
hmm...we sort null as the last value, don't we? wouldn't that work? :D

danking:
dastardly.

danking:
OK, well, I'll remove the variant inequality tests for now, but I think we should figure out how to interval query a string-keyed table.

danking:
OK, ready for review, the bug fix is a few characters in ExtractIntervalFilter.

tpoterba (approved):

PR #12487, opened 2022-11-18
[fs] Fix incorrect seeking semantics of GoogleStorageFS

tpoterba:


hail/src/main/scala/is/hail/io/fs/GoogleStorageFS.scala
```diff
@@ -209,8 +206,7 @@ class GoogleStorageFS(
         return n
       }
 
-      override def seek(newPos: Long): Unit = {
-        super.seek(newPos)
```
tpoterba, line 213: seeking the Google reader was incorrect if we ended up doing a "virtual" seek within the buffer in the superclass. Split it out for clarity and correctness.

PR #12490, opened 2022-11-21
[query] Remove unused _async and _many expression functions

daniel-goldstein:
Aside from removing these for being unused, I think it makes more sense when we want to bring something of the like back for the combiner just to have an async eval that the client (the combiner) can then use to evaluate many IRs concurrently. I don't much see the utility of the *_many methods.

cc: @tpoterba


danking (approved):
LGTM

PR #12495, opened 2022-11-22
[hailtop.utils] Add exception handling for KeyboardInterrupt

jigold:
KeyboardInterrupt inherits from BaseException and not Exception so we couldn't kill waiting on batches etc. in the terminal. I decided to not catch BaseException and instead add the separate handler case because I wasn't sure we want to actually catch BaseException rather than Exception.


daniel-goldstein (dismissed):

jigold:
Sorry! I got scared for a second that I missed a place where the rebase in my vep branch had BaseException. But it seems like main does not so it must have come from an earlier version of a different PR. Can you reapprove?

daniel-goldstein (approved):

PR #12496, opened 2022-11-22
[batch] Quiet down MJC and MJS logging exceptions

jigold:
This should quiet the logging situation when the driver goes offline.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -2732,7 +2732,7 @@ class Worker:
                 job.last_logged_mjc_attempt_failure = time_msecs()
 
             if time_msecs() - job.last_logged_mjc_attempt_failure >= 300 * 1000:
-                log.exception(f'error while marking {job} complete', stack_info=True)
+                log.exception(f'error while marking {job} complete since {time_msecs_str(job.last_logged_mjc_attempt_failure)}', stack_info=True)
                 job.last_logged_mjc_attempt_failure = time_msecs()
         finally:
             log.info(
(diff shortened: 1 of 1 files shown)


batch/batch/worker/worker.py
```diff
@@ -2726,7 +2728,12 @@ async def post_job_complete(self, job, full_status):
         except asyncio.CancelledError:
             raise
         except Exception:
-            log.exception(f'error while marking {job} complete', stack_info=True)
+            if job.last_logged_mjc_attempt_failure is None:
+                job.last_logged_mjc_attempt_failure = time_msecs()
+
+            if time_msecs() - job.last_logged_mjc_attempt_failure >= 300 * 1000:
+                log.exception(f'error while marking {job} complete', stack_info=True)
+                job.last_logged_mjc_attempt_failure = time_msecs()
```
daniel-goldstein, line 2736: Nit. This will print the first time we experience the error and not 5 minutes after the first time, and I feel like for the purposes of searching the logs in event of an outage it would be useful to see the first log attempt.
```suggestion
            if job.last_logged_mjc_attempt_failure is None or time_msecs() - job.last_logged_mjc_attempt_failure >= 300 * 1000:
                log.exception(f'error while marking {job} complete', stack_info=True)
                job.last_logged_mjc_attempt_failure = time_msecs()
```

daniel-goldstein (changes requested):

batch/batch/worker/worker.py
jigold: I thought we wanted to wait 5 minutes to log exceptions so we're not as trigger happy. I'd rather add the last time logged to the exception message. What do you think?

daniel-goldstein: Ah, right, forgot that this is exception and not info. Including it in the message sounds good.

daniel-goldstein (approved):

PR #12497, opened 2022-11-23
[query/ggplot] Updates faceting to respect legend grouping

iris-garden:
This change brings the `facet_wrap` interface more in line with [the `ggplot2` implementation](https://ggplot2-book.org/facet.html#facet-wrap) by defaulting to show axes (with tick marks) for all facets, as well as providing `scales`, `nrow` and `ncol` arguments that the user can use to specify whether the scales should be the same across facets and how many rows or columns to use.

It also updates the faceting code to avoid duplicating legend entries when [grouped legends](https://github.com/hail-is/hail/pull/12254) are used, and to disable interactivity accordingly.

With this update, running this code:

```python
import hail as hl
from hail.ggplot import aes, facet_wrap, geom_point, ggplot, vars
ht = hl.utils.range_table(10)
ht = ht.annotate(threeven=hl.if_else(ht.idx % 3 == 0, "good", "bad"))
ht = ht.annotate(fourven=hl.if_else(ht.idx % 4 == 0, "minor", "major"))
fig = (
    + geom_point(aes(shape=ht.even, color=ht.threeven))
    + facet_wrap(vars(ht.threeven, ht.fourven), nrow=1)
)
fig.show()
```

Produces the following plot:

Screen Shot 2022-11-23 at 12 23 21

### DIFF
--- a/hail/python/hail/ggplot/facets.py
+++ b/hail/python/hail/ggplot/facets.py
@@ -1,12 +1,16 @@
 import abc
 import math
 
+from typing import Dict, Tuple
+
 from .geoms import FigureAttribute
+from .utils import n_partitions
 
 import hail as hl
+from hail import Expression, StructExpression
 
 
-def vars(*args):
+def vars(*args: Expression) -> StructExpression:
     """
 
     Parameters
@@ -23,13 +27,19 @@ def vars(*args):
     return hl.struct(**{f"var_{i}": arg for i, arg in enumerate(args)})
 
 
-def facet_wrap(facets):
+def facet_wrap(facets: StructExpression, *, nrow: int = None, ncol: int = None, scales: str = "fixed") -> "FacetWrap":
     """Introduce a one dimensional faceting on specified fields.
 
     Parameters
     ----------
     facets: :class:`StructExpression` created by `hl.ggplot.vars` function.
         The fields to facet on.
+    nrow: :class:`int`
+        The number of rows into which the facets will be spread. Will be ignored if `ncol` is set.
+    ncol: :class:`int`
+        The number of columns into which the facets will be spread.
+    scales: :class:`str`
+        Whether the scales are the same across facets. For more information and a list of supported options, see `the ggplot documentation <https://ggplot2-book.org/facet.html#controlling-scales>`__.
 
     Returns
     -------
@@ -37,26 +47,65 @@ def facet_wrap(facets):
         The faceter.
 
     """
-    return FacetWrap(facets)
+    return FacetWrap(facets, nrow, ncol, scales)
 
 
 class Faceter(FigureAttribute):
 
     @abc.abstractmethod
-    def get_expr_to_group_by(self):
+    def get_expr_to_group_by(self) -> StructExpression:
         pass
 
 
 class FacetWrap(Faceter):
+    _base_scale_mappings = {
+        "shared_xaxes": "all",
+        "shared_yaxes": "all",
+    }
 
-    def __init__(self, facets):
+    _scale_mappings = {
+        "fixed": _base_scale_mappings,
+        "free_x": {
+            **_base_scale_mappings,
+            "shared_xaxes": False,
+        },
+        "free_y": {
+            **_base_scale_mappings,
+            "shared_yaxes": False,
+        },
+        "free": {
+            "shared_xaxes": False,
+            "shared_yaxes": False,
+        }
+    }
+
+    def __init__(self, facets: StructExpression, nrow: int = None, ncol: int = None, scales: str = "fixed"):
+        if nrow is not None and ncol is not None:
+            raise ValueError(
+                "Both `nrow` and `ncol` were specified. "
+                "Please specify only one of these values."
+            )
+        if scales not in self._scale_mappings:
+            raise ValueError(
+                f"An unsupported value ({scales}) was provided for `scales`. "
+                f"Supported values are: {[k for k in self._scale_mappings.keys()]}."
+            )
+        self.nrow = nrow
+        self.ncol = ncol
         self.facets = facets
+        self.scales = scales
 
-    def get_expr_to_group_by(self):
+    def get_expr_to_group_by(self) -> StructExpression:
         return self.facets
 
-    def get_facet_nrows_and_ncols(self, num_facet_values):
-        ncol = int(math.ceil(math.sqrt(num_facet_values)))
-        nrow = int(math.ceil(num_facet_values / ncol))
+    def get_facet_nrows_and_ncols(self, num_facet_values: int) -> Tuple[int, int]:
+        if self.ncol is not None:
+            return (n_partitions(num_facet_values, self.ncol), self.ncol)
+        elif self.nrow is not None:
+            return (self.nrow, n_partitions(num_facet_values, self.nrow))
+        else:
+            ncol = int(math.ceil(math.sqrt(num_facet_values)))
+            return (n_partitions(num_facet_values, ncol), ncol)
 
-        return (nrow, ncol)
+    def get_shared_axis_kwargs(self) -> Dict[str, str]:
+        return self._scale_mappings[self.scales]
--- a/hail/python/hail/ggplot/geoms.py
+++ b/hail/python/hail/ggplot/geoms.py
@@ -1,4 +1,4 @@
-from typing import Dict, Any
+from typing import Dict, Any, Optional
 import abc
 import numpy as np
 import plotly.graph_objects as go
@@ -18,7 +18,7 @@ class Geom(FigureAttribute):
         self.aes = aes
 
     @abc.abstractmethod
-    def apply_to_fig(self, agg_result, fig_so_far: go.Figure, precomputed, facet_row, facet_col, legend_cache) -> bool:
+    def apply_to_fig(self, agg_result, fig_so_far: go.Figure, precomputed, facet_row, facet_col, legend_cache, is_faceted) -> bool:
         """Add this geometry to the figure and indicate if this geometry demands a static figure."""
         pass
 
@@ -44,7 +44,7 @@ class Geom(FigureAttribute):
                 trace_args["showlegend"] = False
             else:
                 trace_args["showlegend"] = True
-                legend_cache.add(trace_args["name"])
+                legend_cache[trace_args["name"]] = {}
 
 
 class GeomLineBasic(Geom):
@@ -59,7 +59,7 @@ class GeomLineBasic(Geom):
         super().__init__(aes)
         self.color = color
 
-    def apply_to_fig(self, grouped_data, fig_so_far: go.Figure, precomputed, facet_row, facet_col, legend_cache) -> bool:
+    def apply_to_fig(self, grouped_data, fig_so_far: go.Figure, precomputed, facet_row, facet_col, legend_cache, is_faceted: bool) -> bool:
 
         def plot_group(df):
             trace_args = {
@@ -135,7 +135,7 @@ class GeomPoint(Geom):
                 values[aes_name] = value
         return values
 
-    def _add_trace(self, fig_so_far: go.Figure, df, facet_row, facet_col, values):
+    def _add_trace(self, fig_so_far: go.Figure, df, facet_row, facet_col, values, legend: Optional[str] = None):
         fig_so_far.add_scatter(
             **{
                 **{
@@ -144,7 +144,11 @@ class GeomPoint(Geom):
                     "mode": "markers",
                     "row": facet_row,
                     "col": facet_col,
-                    "showlegend": False
+                    **(
+                        {"showlegend": False}
+                        if legend is None else
+                        {"name": legend, "showlegend": True}
+                    )
                 },
                 **self._map_to_plotly(values)
             }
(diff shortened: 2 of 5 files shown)


hail/python/hail/ggplot/geoms.py
```diff
@@ -166,24 +170,43 @@ def _add_legend(self, fig_so_far: go.Figure, aes_name, category, value):
...
                     **legends.get(aes_name, {}),
-                    self._get_aes_value(df, f"{aes_name}_legend"): values[aes_name]
+                    category: values[aes_name]
                 })
-
-        number_of_displayed_legends = 0
-        for aes_name, legend_group in legends.items():
-            if len(legend_group) > 1:
-                number_of_displayed_legends += 1
+            traces = [*traces, [fig_so_far, df, facet_row, facet_col, values, trace_category]]
```
danking, line 186: I'm a bit confused about trace_category. Can you explain to me a bit more about what the goal is here? In particular, I'm generally suspicious of iterating through a group of values and arbitrarily taking the last one that is not None (what if there are multiple not None `category`s?).

hail/python/hail/ggplot/geoms.py
```diff
@@ -166,24 +170,43 @@ def _add_legend(self, fig_so_far: go.Figure, aes_name, category, value):
...
+            requires_static = True
+        else:
+            count = 0
+            for legend_group in legends.values():
+                count += 1 if len(legend_group) > 1 else 0
+            requires_static = count > 1
+
+        if requires_static:
+            for trace in traces:
+                self._add_trace(*trace[:-1])
```
danking, line 198: This is a bit too clever 😉 . What's the harm in sending the trace_category into `_add_trace` in the static case?

hail/python/hail/ggplot/ggplot.py
```diff
@@ -256,10 +258,15 @@ def get_aggregation_result(selected, mapping_per_geom, precomputed):
...
-            xaxis=dict(linecolor="black"),
-            yaxis=dict(linecolor="black"),
             font_family='Arial, "Open Sans", verdana, sans-serif',
-            title_font_size=26
+            title_font_size=26,
+            xaxis=dict(linecolor="black", showticklabels=True),
+            yaxis=dict(linecolor="black", showticklabels=True),
+            **{
+                f"{var}axis{idx}": {"linecolor": "black", "showticklabels": True}
+                for idx in range(2, n_facet_rows + n_facet_cols + 1)
```
danking, line 267: this is probably worthy of a short comment explaining the sillyness of the axis keyword argument names.

danking: Something like:
```
# xaxis stands in for xaxis1; xaxis2 and beyond are as expected
```

danking (changes requested):
Looks great, a few thoughts on tightening the code up

hail/python/hail/ggplot/geoms.py
```diff
@@ -166,24 +170,43 @@ def _add_legend(self, fig_so_far: go.Figure, aes_name, category, value):
...
-                number_of_displayed_legends += 1
+            traces = [*traces, [fig_so_far, df, facet_row, facet_col, values, trace_category]]
+
+        if is_faceted:
+            requires_static = True
+        else:
+            count = 0
+            for legend_group in legends.values():
+                count += 1 if len(legend_group) > 1 else 0
+            requires_static = count > 1
```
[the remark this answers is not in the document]
iris-garden: The original code is actually checking that there are at least *two* `legend_group`s with `len > 1`, so `any` wouldn't work here, as far as I can tell?

hail/python/hail/ggplot/geoms.py
iris-garden: `trace_category` is what will be used to label this group on the legend iff we are making an interactive plot, so in my example above, if we only had the `color` aesthetic and not the `shape` one, `trace_category` would end up being `"bad"` in one iteration, and `"good"` in the other. Because there's only one `shape` category and it was created by default instead of using a category from a user-provided aesthetic mapping, the value of `category` when we're looking at the `shape` aesthetic in the iteration is `None`. Since `trace_category` will only actually be used if we are making an interactive plot, and since we only support interactivity when there's one or fewer aesthetic mappings that will appear in the legend with multiple values, we can assume that if we have a `category` that is `None`, we have exactly one `category` that is `None`.

Agreed that this is suspicious haha, but hopefully we can ditch plotly and the way we currently do grouped legends soonish 😸

hail/python/hail/ggplot/facets.py
```diff
@@ -23,21 +23,27 @@ def vars(*args):
...
+        Whether the scales are the same across facets. For more information and a list of supported options, see `the ggplot documentation <https://ggplot2-book.org/facet.html#controlling-scales>`__.
 
     Returns
     -------
     :class:`FigureAttribute`
         The faceter.
 
     """
-    return FacetWrap(facets)
+    return FacetWrap(facets, nrow, ncol, scales)
```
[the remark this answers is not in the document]
iris-garden: Added an error!

hail/python/hail/ggplot/facets.py
```diff
@@ -49,14 +55,43 @@ def get_expr_to_group_by(self):
 
 class FacetWrap(Faceter):
 
-    def __init__(self, facets):
+    def __init__(self, facets, nrow, ncol, scales):
+        self.nrow = nrow
```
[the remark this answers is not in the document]
iris-garden: Added an error!

hail/python/hail/ggplot/facets.py
```diff
@@ -49,14 +55,43 @@ def get_expr_to_group_by(self):
...
+        self.scales = scales
 
     def get_expr_to_group_by(self):
         return self.facets
 
-    def get_facet_nrows_and_ncols(self, num_facet_values):
-        ncol = int(math.ceil(math.sqrt(num_facet_values)))
-        nrow = int(math.ceil(num_facet_values / ncol))
+    def _calc(self, num_facet_values, n):
+        return int(math.ceil(num_facet_values / n))
```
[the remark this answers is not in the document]
iris-garden: Replaced!

hail/python/hail/ggplot/facets.py
```diff
@@ -49,14 +55,43 @@ def get_expr_to_group_by(self):
...
 
-        return (nrow, ncol)
+    def get_facet_nrows_and_ncols(self, num_facet_values):
+        if self.ncol is not None:
+            return (self._calc(num_facet_values, self.ncol), self.ncol)
+        elif self.nrow is not None:
+            return (self.nrow, self._calc(num_facet_values, self.nrow))
+        else:
+            ncol = int(math.ceil(math.sqrt(num_facet_values)))
+            return (self._calc(num_facet_values, ncol), ncol)
```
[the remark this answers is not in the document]
iris-garden: That's kinda what I figured haha 🤷‍♀️

hail/python/hail/ggplot/facets.py
```diff
@@ -49,14 +55,43 @@ def get_expr_to_group_by(self):
 
 class FacetWrap(Faceter):
 
-    def __init__(self, facets):
+    def __init__(self, facets, nrow, ncol, scales):
```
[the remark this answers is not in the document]
iris-garden: Added!

hail/python/hail/ggplot/geoms.py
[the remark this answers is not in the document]
iris-garden: Replaced!

hail/python/hail/ggplot/ggplot.py
iris-garden: Added a comment!

hail/python/hail/ggplot/geoms.py
danking: OK, so, check that I understand this correctly:

`grouped_data` is `List[DataFrame]` where the DataFrame contains one column for each Hail field needed to render this geometry. I *think* each DataFrame represents a different "grouping" of the data. In particular, I *think* this comes from `Stat.listify`. We should add a type to that method. I think it returns `List[DataFrame]`. I also think there's a bug in StatNone. It should return `[pd.DataFrame({})]`.

IIUC, there's a "grouping" for each "visually manifested" scale. "visually manifested" means it's a discrete scale and not part of the tooltip. I think this should be called `data_by_visible_scale` or something like that.

Alright, so, IIUC, we're trying to collect all the possible values of these visually manifested scales. From that information, we can determine if we need a legend and, furthermore, if we need a static legend.

I think it's also the case that `color_legend` is the description *of the legend itself* for `color`s. Same for shape_legend and shape.

When we're looping through the `data_by_visible_scale`, it's not obvious that we don't get conflicting values for `category` across different data frames. Let's be concrete. In your example above there are four facets ({good, bad} cross {minor, major}) with two scales: color & shape. Color is perfectly correlated with {good, bad}. `data_by_visible_scale` will have a data frame for each combination of color and shape that appears in that particular facet. OK, so after the first iteration of this loop, legends should be this:
```
{
    "shape": {"shape": ["circle"]},
    "color": {"color": ["blue"]}
}
```
Then the next iteration, unless I'm totally misunderstanding here, would blow away these values and replace them with:
```
{
    "shape": {"shape": ["circle", "square"]},
    "color": {"color": ["red"]}
}
```

So clearly I'm misunderstanding something. I think it must be my understanding of `color_legend` / category, what is that?

danking (dismissed):
see above

hail/python/hail/ggplot/geoms.py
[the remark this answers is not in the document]
iris-garden: that's fair! updated based on your suggestion 😸

hail/python/hail/ggplot/geoms.py
iris-garden: good call, added return type annotations for `listify` and fixed the return value for `StatNone`!

for the example in the PR description, this is what `legends` looks like:
```
### `apply_to_fig` call ###
#### iteration of `grouped_data` loop ####
{'color': {'bad': 'rgb(0, 219, 207)'}, 'shape': {'no': 'square'}}
#### iteration of `grouped_data` loop ####
{'color': {'bad': 'rgb(0, 219, 207)'}, 'shape': {'no': 'square', 'yes': 'circle'}}
### `apply_to_fig` call ###
#### iteration of `grouped_data` loop ####
{'color': {'bad': 'rgb(0, 219, 207)'}, 'shape': {'yes': 'circle'}}
### `apply_to_fig` call ###
#### iteration of `grouped_data` loop ####
{'color': {'good': 'rgb(255, 0, 0)'}, 'shape': {'no': 'square'}}
#### iteration of `grouped_data` loop ####
{'color': {'good': 'rgb(255, 0, 0)'}, 'shape': {'no': 'square', 'yes': 'circle'}}
### `apply_to_fig` call ###
#### iteration of `grouped_data` loop ####
{'color': {'good': 'rgb(255, 0, 0)'}, 'shape': {'yes': 'circle'}}
```

so each aesthetic (shape, color) has a key in legends, which is mapped to a `dict` mapping the possible values for that aesthetic (the "categories" extracted from the corresponding column of the hail `Table` or `Expression`) to their value for that aesthetic.

in the `None` case i mentioned above, we get something like this:
```
{'color': {None: 'black'}, 'shape': {'no': 'circle', 'yes': 'square'}}
```

due to .... some other code somewhere 😭 (maybe somewhere in `ScaleDiscrete`?) ... the value for the category name when we don't actually have more than one category is `None`

does that actually address your question? sorry, i got a bit lost along the way

hail/python/hail/ggplot/geoms.py
```diff
@@ -44,7 +44,7 @@ def _update_legend_trace_args(self, trace_args, legend_cache):
                 trace_args["showlegend"] = False
             else:
                 trace_args["showlegend"] = True
-                legend_cache.add(trace_args["name"])
+                legend_cache[trace_args["name"]] = True
```
danking, line 47: On 197, you `**prev` where prev is a value from the legend_cache. `**True` will error. Should this be assigning `{}` instead?

danking (changes requested):

hail/python/hail/ggplot/stats.py
```diff
@@ -58,8 +62,8 @@ class StatNone(Stat):
     def make_agg(self, mapping, precomputed, scales):
         return hl.agg.take(hl.struct(), 0)
 
-    def listify(self, agg_result):
-        return pd.DataFrame({})
+    def listify(self, agg_result) -> List[DataFrame]:
+        return [pd.DataFrame({})]
```
danking, line 66: thanks!

hail/python/hail/ggplot/geoms.py
danking: Yeah, I think I kinda see what's happening now. I have a suspicion we could find a slightly simpler, if less efficient, solution by passing over the data twice: once to compute the legend and once to add the trace. No need to act on that now. Let's keep it in mind as we look to Altair though.

hail/python/hail/ggplot/geoms.py
iris-garden: this doesn't cause problems on 197 because `GeomPoint.apply_to_fig` never calls `_update_legend_trace_args`. i changed `legend_cache` to a `dict` instead of a `set` because i needed to put the whole `legend_group` in there rather than just the `aes_name`, but since all the other geoms still function the same as before, this was just the easiest way i could think of to have it basically still be a `set` (just the set of the `dict`'s keys, while ignoring the values)

that said, i think it's still a good idea to update it, so that the type of `legend_cache` is more consistent!

danking (approved):

PR #11879, opened 2022-06-01
[batch] Add pool labels

lgruen:
This adds a `label` column to pools, which can be used to select a subset of pools to consider when scheduling a job.

The label can be specified for each job by setting the `_pool_label` attribute, e.g. `job._pool_label = 'seqr'` will consider all pools that have the `seqr` label.

Note: this incurs a DB migration.

`batch/sql/add-seqr-pools.sql` is an example for adding a copy of the default preemptible pools, with an additional `seqr` label applied. CPG limits the number of instances in those dedicated pools to prevent long running seqr loading pipeline jobs from starving other batches for resources.

#assign services

PR #12501, opened 2022-11-23
[batch] Add instance collection labels

jigold:
@lgruen Can you take a look at these changes and let me know if they'll work easily with your current deployment. If yes, then I'll test it all works next week in my namespace.


jigold:
Modification of #11944 and #11879

jigold:
@lgruen bump

jigold:
I'm going to close this for now as we have enough db transformations in flight right now. Can reopen in the future.

PR #12505, opened 2022-11-28
[batch] Quiet warning log messages

jigold:
Open to suggestions if you have ideas for a better way of doing this.


daniel-goldstein (changes requested):
I'm just seeing now that the previous PR's exception handling was higher up the callstack than it maybe should have been. Looking at this now, it looks like the except block in `post_job_complete_1` would be where we expect to log errors/warnings from a down driver but if we get to the except block in `post_job_complete` then something has gone very wrong and we would want to log exceptions immediately.

Am I right that this is specifically so that we don't log errors when the driver is down and the workers are trying to talk to it? I feel like in that case we should only do this delayed logging when we get a 503 in `post_job_complete_1`, and everything else should log errors eagerly. Can we do that or am I missing another scenario where we want to prevent error logging?

batch/batch/worker/worker.py
```diff
@@ -1494,6 +1494,12 @@ def status(self):
     def done(self):
         return self.state in ('succeeded', 'error', 'failed')
 
+    def log_delayed_warning(self, message, exc_info):
+        last_logged_time_msecs, start_time_msecs = self.last_logged_message[message]
+        if time_msecs() - last_logged_time_msecs >= 300 * 1000:
+            log.warning(message + f' since {time_msecs_str(start_time_msecs)}', exc_info=exc_info)
+            self.last_logged_message[message] = (time_msecs(), start_time_msecs)
```
daniel-goldstein, line 1501: I don't really understand `start_time_msecs` role here. It's never reset, so we just keep printing out the first time this machine ever logged such a warning? I feel like if there's a burst of warnings, a quiet period where everything is fine, and then another burst, the warnings from the second burst have no reason to reference the previous burst.


I feel like this is starting to get pretty complicated. Instead of all this specific bookkeeping, can we instead add some flexibility to retry_transient_errors, where we say something along the lines of "only start logging warnings for X kind of error after X amount of time", and then the only thing we need in worker.py is to say that we should only log warnings for 503s after X minutes?

daniel-goldstein (dismissed):

batch/batch/worker/worker.py
daniel-goldstein: We also should not log 429s. We get 429s when being rate limited and 503s when batch is offline

daniel-goldstein (approved):

PR #12510, opened 2022-11-28
[hailctl] Fix dependency installation for hailctl dataproc clusters

daniel-goldstein:
I broke this when I separated out the dependencies for hail into two layers:

1. hailtop dependencies
2. hail dependencies, which builds on top of the hailtop dependencies.

This fix does two things:
- Use the full dependencies in 1 & 2
- Use fully pinned dependencies when installing on clusters which seems better than using our wide-range dependencies. I left the `install-deps` and `install-dev-deps` as the normal requirements files as those are meant for development (I think?) but am happy to take opinions on whether we should use fully pinned deps there as well. I have so far been going by the rule of thumb of fully-pinned for CI and production environments, more lax rules for dev environments.

See [here](https://github.com/hail-is/hail/pull/12446#discussion_r1030986069) for additional context.

cc: @tpoterba, any idea why the test dataproc test succeeded?


tpoterba:
test_dataproc only runs on deploys AFAIK

daniel-goldstein:
Like a release? I think the terminology is a little overloaded. In services code we consider "deploy" as the CI pipeline that runs whenever main is updated.

daniel-goldstein:
Ah I see, all these steps cut out early for non-release deploys. So we wouldn't have released a broken package which is good, but we'd still end up with a broken deploy.

tpoterba:
Sorry about the incorrect terminology! Your conclusion is my understanding, yes.

hail/Makefile
```diff
@@ -268,13 +268,13 @@ $(eval $(call ENV_VAR,cloud_base))
 $(eval $(call ENV_VAR,wheel_cloud_path))
 
 python/hailtop/hailctl/deploy.yaml: env/cloud_base env/wheel_cloud_path
-python/hailtop/hailctl/deploy.yaml: $(resources) python/requirements.txt
+python/hailtop/hailctl/deploy.yaml: $(resources) python/pinned-requirements.txt
```
danking, line 271: this target does not exist.

hail/Makefile
```diff
@@ -324,7 +324,7 @@ install: $(WHEEL)
 
 .PHONY: install-on-cluster
 install-on-cluster: $(WHEEL)
```
danking, line 326: this target can run without generated pinned requirements files.

danking (dismissed):

vladsavelyev:
Thank you!

hail/Makefile
daniel-goldstein: Not sure I understand entirely. Do you mean that `install-on-cluster` should use unpinned requirements or is this a concern about the target / dependencies of the target?

hail/Makefile
danking: This seems like a foot-gun to me. Suppose I'm trying to experiment with a modified version of Hail. I've modified `python/requirements.txt` but I've not run `generate-pip-lockfile`. I run `make install-hailctl` thinking that this will incorporate my changes into a new wheel file which I can use to start cluster. In reality, my changes are ignored and the old pinned requirements are used.

I think we need a rule like:

```
python/pinned-requirements.txt: python/requirements.txt
	docker run --rm -it \
		-v $(HAIL_HAIL_DIR):/hail \
		python:3.7-slim \
		/bin/bash -c "pip install pip-tools && cd /hail && \
	pip-compile --upgrade python/requirements.txt \
		--output-file=python/pinned-requirements.txt"
```

Or a rule like:

```
python/pinned-requirements.txt: generate-pip-lockfile
```

This latter rule will trigger `generate-pip-lockfile` to always be run if we depend on the pinned-requirements.txt which isn't great. I think the best option is to have a rule for each requirements file and to also have generate-pip-lockfile as a special fast path for developers who want to update all the lock files.

hail/Makefile
danking: This is similar to my concerns above, there's an implicit dependency between python/requirements.txt and python/pinned-requirements.txt. At the very least we should error if the pinned-requirements.txt is older than the requirements.txt

danking (dismissed):

hail/Makefile
daniel-goldstein: Thanks for the clarification, this makes a lot of sense! I will incorporate these changes.

daniel-goldstein:
Because of the unpredictable way that git clone might realize the requirements files, I removed the pinned-requirements file as a dependency of the changed targets. In both cases, regenerating that file (either in CI as part of the deploy.yaml target or on a cluster for `install-on-cluster`) could cause a dataproc cluster running with untested dependency versions even if the requirements.txt files are unchanged. I do, however, require that the pinned-requirements files be compatible using the same check we do in CI

I performed the following manual testing:
1. Creating a dataproc cluster through `hailctl dataproc start`
2. ssh'ing into said cluster, cloning this branch and running `make -C hail install-on-cluster` to completion
3. Updating the requirements.txt file to something incompatible and successfully installing on cluster again

danking (dismissed):

hail/Makefile
```diff
@@ -267,14 +267,35 @@ resources := $(wildcard python/hailtop/hailctl/dataproc/resources/*)
 $(eval $(call ENV_VAR,cloud_base))
 $(eval $(call ENV_VAR,wheel_cloud_path))
 
+
+
+python/hailtop/pinned-requirements.txt: python/hailtop/requirements.txt
+	HAIL_HAIL_DIR=$(HAIL_HAIL_DIR) \
+	../generate_pip_lockfile.sh python/hailtop/requirements.txt python/hailtop/pinned-requirements.txt
+
+python/pinned-requirements.txt: python/hailtop/requirements.txt python/requirements.txt
```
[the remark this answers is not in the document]
daniel-goldstein: `python/requirements.txt` includes those requirements from `python/hailtop/requirements.txt`, and the dev requirements are constrained by `python/requirements.txt`, so if any of these requirements files changed, it could affect the pinned-requirements.

hail/Makefile
```diff
@@ -267,14 +267,35 @@ resources := $(wildcard python/hailtop/hailctl/dataproc/resources/*)
...
-python/hailtop/hailctl/deploy.yaml: $(resources) python/requirements.txt
+python/hailtop/hailctl/deploy.yaml: $(resources)
 	rm -f $@
 	echo "dataproc:" >> $@
 	for FILE in $(notdir $(resources)); do \
 	  echo "  $$FILE: $(cloud_base)/$$FILE" >> $@ || exit 1; done
 	echo "  wheel: $(wheel_cloud_path)" >> $@
-	echo "  pip_dependencies: $(shell cat python/requirements.txt | sed '/^[[:blank:]]*#/d;s/#.*//' | grep -v pyspark | tr "\n" "|||")" >> $@
+	printf "  pip_dependencies: " >> $@
+	$(PIP) install pip-tools && ../check_pip_requirements.sh python/requirements.txt python/pinned-requirements.txt
```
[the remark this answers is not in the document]
daniel-goldstein: Yes

daniel-goldstein:
@danking Sorry for the delay on this. Trying to move this along faster now to unblock the release

danking (approved):
Please follow up with change to target naming of `check-pip-lockfiles`

PR #12511, opened 2022-11-28
[qob] Dont include the batch name in the worker name anymore

daniel-goldstein:
Now the worker job names look like this instead of the worker jobs also having the batch name as the prefix:

Screen Shot 2022-11-28 at 3 13 50 PM


danking (approved):
1 line PRs are my favorite

PR #12512, opened 2022-11-29
[query] Expose `hl.keyed_intersection` and `hl.keyed_union` in Python.

tpoterba:


chrisvittal (approved):

hail/python/hail/expr/expressions/typed_expressions.py
```diff
@@ -533,7 +533,7 @@ def first(self):
         None
         """
         # FIXME: this should generate short-circuiting IR when that is possible
-        return hl.rbind(self, lambda x: hl.or_missing(hl.len(x) > 0, x[0]))
+        return hl.fold(lambda acc, elt: hl.coalesce(acc, elt), hl.missing(self.dtype.element_type), self)
```
danking, line 536: is fold short-circuiting? we should nix the comment if so

hail/python/hail/expr/functions.py
```diff
@@ -4781,6 +4782,98 @@ def flatten(collection):
...
+
+@typecheck(arrays=expr_oneof(expr_stream(expr_any), expr_array(expr_any)), key=sequenceof(builtins.str))
+def keyed_intersection(*arrays, key):
+    """Compute the intersection of sorted arrays on a given key.
+
+    Requires sorted arrays with distinct keys.
+
+    Warning
+    -------
+    Experimental. Does not support downstream randomness.
```
danking, line 4832: Why doesn't this support downstream randomness whereas keyed_union does?

danking (dismissed):

hail/python/hail/expr/expressions/typed_expressions.py
tpoterba: we need a `foldWithEarlyTruncation` exposed

hail/python/hail/expr/functions.py
tpoterba: oops -- docs failure.

danking: I doubt our users will understand what this means. Is it difficult to properly support randomness? I'm surprised that something built in Python breaks randomness.

PR #12519, opened 2022-11-30
[qob] Invalidate batch cache on error

jigold:
Not sure this is the right fix, but I believe we can't update a batch once it's been cancelled still.


jigold:
Maybe this change isn't needed. Can we cancel a batch multiple times now?

daniel-goldstein (approved):
Good spot. I don't think you can do anything with a batch at the moment once it's been cancelled, I think this is the right change for the time-being.

PR #12520, opened 2022-11-30
[query] enable fs tests

danking:
The FS tests were accidentally disabled because pytest-asyncio was never included in dev/requirements.txt (as it should have been). As a result, all the `async def` tests were (silently) ignored by pytest.


daniel-goldstein (approved):
oh man

daniel-goldstein (approved):

PR #12522, opened 2022-11-30
[batch] Add hailctl config batch/backend

daniel-goldstein:
CHANGELOG: Added hailctl config setting 'batch/backend' to specify the default backend to use in batch scripts when not specified in code.

This is more consistent now with `HAIL_QUERY_BACKEND` and makes it less boilerplate-y to use the service.

cc: @danking

### DIFF
--- a/hail/python/hailtop/batch/batch.py
+++ b/hail/python/hailtop/batch/batch.py
@@ -8,6 +8,7 @@ import dill
 from hailtop.utils import secret_alnum_string, url_scheme, async_to_blocking
 from hailtop.aiotools import AsyncFS
 from hailtop.aiotools.router_fs import RouterAsyncFS
+from hailtop.config import configuration_of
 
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/batch/batch.py
```diff
@@ -121,7 +128,15 @@ def __init__(self,
...
-        self._backend = backend if backend else _backend.LocalBackend()
+        if backend:
+            self._backend = backend
+        else:
+            backend_config = configuration_of('batch', 'backend', None, 'local')
+            if backend_config == 'service':
+                self._backend = _backend.ServiceBackend()
+            else:
+                assert backend_config == 'local'
+                self._backend = _backend.LocalBackend()
```
danking, line 139: so delightfully simple.

danking (approved):

PR #12523, opened 2022-11-30
[query] address another rare Google error

danking:
For example: https://ci.hail.is/batches/6800777/jobs/108


daniel-goldstein (approved):
*sigh* google

PR #12526, opened 2022-12-01
[hailgenetics/hail] slim down to ~1.5GiB from 2.8GiB

danking:
The main offenders were:
1. `python3-pip` installs python3.8.
2. `gcloud` is like 1.2GiB.


docker/hailgenetics/hail/Dockerfile
```diff
@@ -1,32 +1,16 @@
...
+    vim
 
 COPY hail/python/pinned-requirements.txt requirements.txt
-RUN hail-pip-install -r requirements.txt \
-      ipython \
-      matplotlib \
-      numpy \
-      scikit-learn \
-      dill \
-      scipy
```
danking, line 29: The only ones that aren't picked up in requirements.txt are scikit-learn and matplotlib. Scikit-learn is kinda specialized, but I could be convinced to keep it. Matplotlib is similar in size but we already provide two plotting libraries: plotly and bokeh (bokeh is like 50MiB).

danking:
Still a totally unreasonable 42s to pull this image but better than >1minute.

danking:
The big opportunities to save now are things like Hail's JAR (108MB), JVM (123MB), pyspark (>200MB), the GCS connector (~30MB), bokeh (135 MB, we should kill it in favor of plotly or Altair cc: @iris-garden), vim (30 MB, maybe we should just get rid of it?).

danking:
But also, man, we're only pulling 308mbps inside the datacenter?

lgruen:
So this doesn't come with `gcloud` anymore, right? Don't you need that to activate service accounts? I'm asking because we'd also love to shave off 1.2 GiB from our images ;).

Generally I still think it would be awesome to investigate eStargz further, as that would make the image size itself much less of a problem: https://hail.zulipchat.com/#narrow/stream/300487-Hail-Batch-Dev/topic/Lazy.20container.20image.20pulls.20.2F.20stargz/near/301553710

danking:
@lgruen to address your question directly, AFAIK, if you set `GOOGLE_APPLICATION_CREDENTIALS`, tools will use those credentials. That's at least the intended behavior of all Hail tools. I think `gcloud auth activate-service-account` is primarily used to tell gcloud & gsutil about a key. `gcloud auth application-default login` likewise seems to be useful for getting ADC credentials if one lacks them but not necessary for using credentials with tools that check the env var.

danking:
FWIW, Hail Batch sets `GOOGLE_APPLICATION_CREDENTIALS` (resp. `AZURE_...`) appropriately if they're not explicitly specified by the user with `j.env(...)`.

daniel-goldstein (approved):
Looks like just nix the `gcloud` check from the test step and you're golden.

daniel-goldstein:
Oh yikes, looks like something in the LLVM dependencies requires/installs python3.8. I wonder whether that is necessary

danking:
Yeah. I think the main issue is that we need to run update-alternatives after we install a new version of Python. If query-build is just for native compiler stuff, then I'm fine with using 3.8 there. I'll see if I can get back to this PR later.

danking:
Slapping a WIP on while I debug this weirdness.

PR #12530, opened 2022-12-01
[hailtop.batch] Add ability to update a batch

jigold:
CHANGELOG: Added the ability to update an existing batch with additional jobs by calling `Batch.run()` more than once.

I wrote this quite awhile ago, so I don't remember if there were any issues as to why I didn't make a PR sooner or if I'm missing some complexity here.

### DIFF
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -329,9 +329,7 @@ class LocalBackend(Backend[None]):
                 code += ['\n']
 
                 run_code(code)
-
-                for job in batch._unsubmitted_jobs:
-                    job._submitted = True
+                job._submitted = True
         finally:
             if delete_scratch_on_exit:
                 sp.run(f'rm -rf {tmpdir}', shell=True, check=False)
@@ -629,7 +627,9 @@ class ServiceBackend(Backend[bc.Batch]):
                 jobs_to_command[j] = ' '.join(shq(x) for x in write_cmd)
                 n_jobs_submitted += 1
 
-        pyjobs = [j for j in batch._jobs if isinstance(j, _job.PythonJob)]
+        unsubmitted_jobs = batch._unsubmitted_jobs
+
+        pyjobs = [j for j in unsubmitted_jobs if isinstance(j, _job.PythonJob)]
         for pyjob in pyjobs:
             if pyjob._image is None:
                 version = sys.version_info
@@ -642,18 +642,18 @@ class ServiceBackend(Backend[bc.Batch]):
             batch_remote_tmpdir, dry_run=dry_run
         )
 
-        disable_setup_steps_progress_bar = disable_progress_bar or len(batch._jobs) < 10_000
-        with SimpleRichProgressBar(total=len(batch._jobs),
+        disable_setup_steps_progress_bar = disable_progress_bar or len(unsubmitted_jobs) < 10_000
+        with SimpleRichProgressBar(total=len(unsubmitted_jobs),
                                    description='upload code',
                                    disable=disable_setup_steps_progress_bar) as pbar:
             async def compile_job(job):
                 used_remote_tmpdir = await job._compile(local_tmpdir, batch_remote_tmpdir, dry_run=dry_run)
                 pbar.update(1)
                 return used_remote_tmpdir
-            used_remote_tmpdir_results = await bounded_gather(*[functools.partial(compile_job, j) for j in batch._jobs], parallelism=150)
+            used_remote_tmpdir_results = await bounded_gather(*[functools.partial(compile_job, j) for j in unsubmitted_jobs], parallelism=150)
             used_remote_tmpdir |= any(used_remote_tmpdir_results)
 
-        for job in track(batch._unsubmitted_jobs, description='create job objects', disable=disable_setup_steps_progress_bar):
+        for job in track(unsubmitted_jobs, description='create job objects', disable=disable_setup_steps_progress_bar):
             inputs = [x for r in job._inputs for x in copy_input(r)]
 
             outputs = [x for r in job._internal_outputs for x in copy_internal_output(r)]
@@ -788,7 +788,8 @@ class ServiceBackend(Backend[bc.Batch]):
         if wait:
             if verbose:
                 print(f'Waiting for batch {batch_handle.id}...')
-            status = batch_handle.wait(disable_progress_bar=disable_progress_bar)
+            starting_job_id = min(j._client_job.job_id for j in unsubmitted_jobs)
+            status = batch_handle.wait(disable_progress_bar=disable_progress_bar, starting_job=starting_job_id)
             print(f'batch {batch_handle.id} complete: {status["state"]}')
         return batch_handle
 
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -772,7 +772,7 @@ class ServiceBackend(Backend[bc.Batch]):
             batch._batch_handle = batch_handle
         else:
             new_batch_handle = bc_batch_builder.submit(disable_progress_bar=disable_progress_bar)
-            assert batch_handle.id == new_batch_handle.id
+            assert batch._batch_handle.id == new_batch_handle.id
             batch_handle = batch._batch_handle
 
         for job in batch._unsubmitted_jobs:
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -332,7 +332,6 @@ class LocalBackend(Backend[None]):
                 job._submitted = True
         finally:
             batch._python_function_defs.clear()
-            batch._python_function_files.clear()
             if delete_scratch_on_exit:
                 sp.run(f'rm -rf {tmpdir}', shell=True, check=False)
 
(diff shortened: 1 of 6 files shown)


daniel-goldstein:
Looks like it was opened and closed here #12381.

I haven't taken a look at this yet, but there are two footguns that should be avoided as much as possible:

- A job waiting on a batch that it is a part of. We add the batch id into the container so we should be able to throw an error here
- Waiting on a batch in general is really wonky when there is more than 1 entity controlling it. I don't know of a good way to control this, so it might just be a "be sure you know what you're doing thing", but worth thinking about.

jigold:
Thanks for the list of things to watch out for. I think the changes here have nothing to do with waiting, which was why I was concerned I was missing something.

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -1044,3 +1044,18 @@ def test_no_copy_output_on_failure(self):
...
+        res = b.run()
+
+        res_status = res.status()
+        assert res_status['state'] == 'success', str((res_status, res.debug_info()))
+
+        j2 = b.new_job()
+        j2.command('true')
+        res = b.run()
+        res_status = res.status()
+        assert res_status['state'] == 'success', str((res_status, res.debug_info()))
```
daniel-goldstein, line 1061: Can you add a test where there is a dependency on a job from a prior `run`?

daniel-goldstein (dismissed):
Ah ya you're right this change doesn't get into the pitfalls of waiting. I think the feature Michael was mentioning that would get into this territory is for a python job to be able to append to its own batch, not to `run` more than once. This is nevertheless useful though and I feel like Jonah / KC would need this.

hail/python/hailtop/batch/backend.py
```diff
@@ -329,6 +329,9 @@ def transfer_dicts_for_resource_file(res_file: Union[resource.ResourceFile, reso
                 code += ['\n']
 
                 run_code(code)
+
+                for job in batch._unsubmitted_jobs:
+                    job._submitted = True
```
daniel-goldstein, line 334: I think this is indented one too deep. Regardless, this is already in a loop over `batch._unsubmitted_jobs`, so can we just nix this loop and put `job._submitted = True` after `run_code(code)`?

hail/python/hailtop/batch/batch.py
```diff
@@ -343,6 +353,7 @@ def hello(name):
             j.timeout(self._default_timeout)
 
         self._jobs.append(j)
+        self._unsubmitted_jobs.append(j)
```
daniel-goldstein, line 356: I don't think this is meant to be here anymore since `_unsubmitted_jobs` is a computed property

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -535,7 +535,7 @@ def _create_job(self,
                     in_update_parent_ids.append(job._job_id)
             else:
                 assert isinstance(job, SubmittedJob)
-                if self._batch is None or job._batch != self._batch:
+                if self._batch is None or (job._batch != self._batch and job._batch.id != self._batch.id):
```
daniel-goldstein, line 538: I think the `job._batch != self._batch` part of this expression is redundant now and can be deleted.

daniel-goldstein (changes requested):
This looks great, just a couple tiny changes.

One additional thing to make sure of: how does the progress bar look for multiple `b.run`s? When making similar changes to the QoB service backend I noticed an annoying thing happened that I had to intentionally fix:

If I submitted 5 jobs I would get a normal progress bar 1-5. If I then submitted another 5 jobs I would start to see a normal progress bar 1-5 at first and then it would immediately jump to 5/10 because it would update to looking at the total number of jobs in the batch instead of this specific submission. I fixed this so the progress bar wouldn't jitter but am not sure if that fix would be picked up in this part of the code. Could you just confirm that the progress bar behavior seems consistent?

jigold:
I changed some things to try and avoid rewriting Python functions if the job has already been submitted previously.

jigold:
The one test failure was a transient blob Not Found on Azure.

daniel-goldstein (changes requested):

daniel-goldstein (dismissed):
So sorry, I had that comment pending since last week and just never hit "Submit Review"

daniel-goldstein (dismissed):

hail/python/hailtop/batch/backend.py
```diff
@@ -329,7 +329,10 @@ def transfer_dicts_for_resource_file(res_file: Union[resource.ResourceFile, reso
                 code += ['\n']
 
                 run_code(code)
+                job._submitted = True
         finally:
+            batch._python_function_defs.clear()
+            batch._python_function_files.clear()
```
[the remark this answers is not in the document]
jigold: The tests aren't passing now with this change. File not found error.

jigold:
I understand what is going on now. The issue is that the temp directory is getting removed after the first batch.run(), but we're assuming those input files are still there.

I think we should just clear the files and definitions cache after submission.

daniel-goldstein:
> I understand what is going on now. The issue is that the temp directory is getting removed after the first batch.run(), but we're assuming those input files are still there.
> 
> I think we should just clear the files and definitions cache after submission.

Ah I see, ya let's not go that far. In that case we don't need any of the changes in this PR to how python files work right? Other than to clear those dictionaries after submit?

hail/python/hailtop/batch/batch.py
```diff
@@ -194,10 +204,11 @@ async def _serialize_python_functions_to_input_files(
         self, path: str, dry_run: bool = False
     ) -> None:
         for function_id, function in self._python_function_defs.items():
-            file = await self._serialize_python_to_input_file(
-                path, "functions", function_id, function, dry_run
-            )
-            self._python_function_files[function_id] = file
+            if function_id not in self._python_function_files:
```
daniel-goldstein, line 207: I think you no longer need these changes. The if statement will always be true since you clear out the dictionary between submissions.

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -535,7 +535,7 @@ def _create_job(self,
                     in_update_parent_ids.append(job._job_id)
             else:
                 assert isinstance(job, SubmittedJob)
-                if self._batch is None or job._batch != self._batch:
+                if self._batch is None or job._batch.id != self._batch.id:
```
daniel-goldstein, line 538: I'm concerned that multiple `Batch` objects that represent the same batch floating around in memory could cause difficult to diagnose bugs surrounding equality. Is the reason for this change because the `_run` function accepts a `Batch` and then returns a different `Batch` to the user? Could we without much additional effort ensure that only one `Batch` object exists after resubmission? Having the number of batch objects be `O(updates)` feels like a memory leak (though not a big one).

daniel-goldstein (changes requested):
Sorry for the late turnaround on this one. I just have one concern the rest looks good.

hail/python/hailtop/batch_client/aioclient.py
jigold: Yes that was the reason. Hopefully, I fixed it in a reasonable way. I also added a static method to create a Batch from a batch id that I used for Siwei's use case.

jigold:
The tests are passing. The error on Azure was cleaning up the base image timed out.

hail/python/hailtop/batch/backend.py
```diff
@@ -759,7 +767,16 @@ async def compile_job(job):
             print(f'Built DAG with {n_jobs_submitted} jobs in {round(time.time() - build_dag_start, 3)} seconds.')
 
         submit_batch_start = time.time()
-        batch_handle = bc_batch.submit(disable_progress_bar=disable_progress_bar)
+        if batch._batch_handle is None:
+            batch_handle = bc_batch_builder.submit(disable_progress_bar=disable_progress_bar)
+            batch._batch_handle = batch_handle
+        else:
+            new_batch_handle = bc_batch_builder.submit(disable_progress_bar=disable_progress_bar)
+            assert batch._batch_handle.id == new_batch_handle.id
```
daniel-goldstein, line 775: I think if we make the above change where we call `update_batch` with the batch_handle, then we can actually make the following assertion:

```python
batch._batch_handle == new_batch_handle
```

which seems necessary since the aioclient Batch and Job classes hold references to each other and we don't want to run into situations where the aioclient Batch object doesn't have a reference to all past jobs.

hail/python/hailtop/batch_client/aioclient.py
daniel-goldstein: This should be reverted such that the assertion is `job._batch != self._batch`

daniel-goldstein (dismissed):

hail/python/hailtop/batch_client/client.py
```diff
@@ -315,7 +315,7 @@ def create_batch(self,
         return BatchBuilder.from_async_builder(builder)
 
     def update_batch(self, batch: Union[int, Batch]) -> 'BatchBuilder':
-        builder = async_to_blocking(self._async_client.update_batch(batch))
+        builder = async_to_blocking(self._async_client.update_batch(batch._async_batch))
```
daniel-goldstein, line 318: looks like this needs to be `batch._async_batch if isinstance(batch, Batch) else batch`

daniel-goldstein (approved):
Looks like just one bug fix to get things passing. This looks great, thanks for dealing with my pedantry on this one

jigold:
I dismissed the review because there was still an issue and I had to change more code that should probably be reviewed.

daniel-goldstein (approved):
hm there was some additional complexity I didn't anticipate with all this wrapping of optional state, but it's mostly a matter of readability and I don't have a better suggestion at the moment so no reason to block this important feature.

PR #12536, opened 2022-12-02
[query] Fix pytest-qob and fix test_import_bed

danking:
The environment variables should be affecting `pytest`, not `cd`.

`test_import_bed` just needs to use an `open` that can speak `gs://`.


daniel-goldstein (approved):
@iris-garden there ya go

PR #12538, opened 2022-12-02
[query] implement hl.grep for QoB

danking:


tpoterba (approved):

hail/python/test/hail/methods/test_impex.py
```diff
@@ -2098,7 +2098,6 @@ def test_error_with_context(self):
 
 
 class GrepTests(unittest.TestCase):
-    @fails_service_backend()
     @fails_local_backend()
     def test_grep_show_false(self):
         expected = {'sampleAnnotations.tsv': ['HG00120\tCASE\t19599', 'HG00121\tCASE\t4832'],
```
daniel-goldstein, line 2103: @danking Do these basically just need a `resource(…)` around them?

danking: basically. Spark-based grep doesn't include full paths so I switched on backend.

PR #12542, opened 2022-12-06
[batch] Storage & Network Usage Monitoring

jigold:
FYI: @danking 
Screen Shot 2022-12-06 at 4 18 55 PM

Screen Shot 2022-12-06 at 4 20 09 PM

### DIFF
--- a/batch/batch/resource_usage.py
+++ b/batch/batch/resource_usage.py
@@ -13,6 +13,9 @@ from hailtop.utils import check_shell_output, time_msecs, time_ns
 log = logging.getLogger('resource_usage')
 
 
+iptables_lock = asyncio.Lock()
+
+
 class ResourceUsageMonitor:
     VERSION = 2
     missing_value = None
@@ -39,7 +42,7 @@ class ResourceUsageMonitor:
             ('cpu_usage', '>f8'),
         ]
 
-        if version >= 1:
+        if version > 1:
             assert version == ResourceUsageMonitor.VERSION, version
             dtype += [
                 ('non_io_storage_in_bytes', '>i8'),
@@ -56,7 +59,7 @@ class ResourceUsageMonitor:
         container_name: str,
         container_overlay: str,
         io_volume_mount: Optional[str],
-        veth_host: Optional[str],
+        veth_host: str,
         output_file_path: str,
     ):
         self.container_name = container_name
@@ -64,6 +67,8 @@ class ResourceUsageMonitor:
         self.io_volume_mount = io_volume_mount
         self.veth_host = veth_host
         self.output_file_path = output_file_path
+
+        self.is_attached_disk = io_volume_mount is not None and os.path.ismount(io_volume_mount)
 
         self.last_time_ns: Optional[int] = None
         self.last_cpu_ns: Optional[int] = None
@@ -120,33 +125,35 @@ class ResourceUsageMonitor:
             return shutil.disk_usage(self.io_volume_mount).used
         return 0
 
-    def is_attached_disk(self) -> bool:
-        return self.io_volume_mount is not None and os.path.ismount(self.io_volume_mount)
-
     async def network_bandwidth(self) -> Tuple[Optional[float], Optional[float]]:
-        now_time_msecs = time_msecs()
+        async with iptables_lock:
+            now_time_msecs = time_msecs()
 
-        iptables_output, _ = await check_shell_output(
-            f'''
-iptables -t mangle -L -v -n -x | grep "{self.veth_host}" | awk '{{ if ($7 == "{self.veth_host}") print $2 }}'
+            iptables_output, stderr = await check_shell_output(
+                f'''
+iptables -t mangle -L -v -n -x -w | grep "{self.veth_host}" | awk '{{ if ($6 == "{self.veth_host}" || $7 == "{self.veth_host}") print $2, $6, $7 }}'
 '''
-        )
-        now_download_bytes = int(iptables_output.decode('utf-8').rstrip())
+            )
+        if stderr:
+            log.exception(stderr)
+        output = iptables_output.decode('utf-8').rstrip().splitlines()
+        assert len(output) == 2, str((output, self.veth_host))
 
-        iptables_output, _ = await check_shell_output(
-            f'''
-iptables -t mangle -L -v -n -x | grep "{self.veth_host}" | awk '{{ if ($6 == "{self.veth_host}") print $2 }}'
-'''
-        )
-        now_upload_bytes = int(iptables_output.decode('utf-8').rstrip())
+        now_upload_bytes = None
+        now_download_bytes = None
+        for line in output:
+            fields = line.split()
+            bytes_transmitted = int(fields[0])
 
-        if (
-            now_upload_bytes is None
-            or now_download_bytes is None
-            or self.last_upload_bytes is None
-            or self.last_download_bytes is None
-            or self.last_time_msecs is None
-        ):
+            if fields[1] == self.veth_host and fields[2] != self.veth_host:
+                now_upload_bytes = bytes_transmitted
+            else:
+                assert fields[1] != self.veth_host and fields[2] == self.veth_host, line
+                now_download_bytes = bytes_transmitted
+
+        assert now_upload_bytes is not None and now_download_bytes is not None, output
+
+        if self.last_upload_bytes is None or self.last_download_bytes is None or self.last_time_msecs is None:
             self.last_time_msecs = time_msecs()
             self.last_upload_bytes = now_upload_bytes
             self.last_download_bytes = now_download_bytes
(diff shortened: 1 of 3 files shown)


jigold:
I think this is much better!

Screen Shot 2022-12-08 at 9 38 27 AM
Screen Shot 2022-12-08 at 9 38 10 AM

jigold:
If you're okay with this, I need to test backwards compatibility with the older version of resource usage files.

batch/batch/resource_usage.py
```diff
@@ -82,6 +112,58 @@ def memory_usage_bytes(self) -> Optional[int]:
...
+
+        iptables_output, _ = await check_shell_output(
+            f'''
+iptables -t mangle -L -v -n -x | grep "{self.veth_host}" | awk '{{ if ($6 == "{self.veth_host}") print $2 }}'
+'''
+        )
+        now_upload_bytes = int(iptables_output.decode('utf-8').rstrip())
+
+        if (
+            now_upload_bytes is None
```
daniel-goldstein, line 144: How can casting the above to an `int` be None?

batch/batch/resource_usage.py
```diff
@@ -82,6 +112,58 @@ def memory_usage_bytes(self) -> Optional[int]:
...
+            f'''
+iptables -t mangle -L -v -n -x | grep "{self.veth_host}" | awk '{{ if ($7 == "{self.veth_host}") print $2 }}'
+'''
+        )
+        now_download_bytes = int(iptables_output.decode('utf-8').rstrip())
+
+        iptables_output, _ = await check_shell_output(
+            f'''
+iptables -t mangle -L -v -n -x | grep "{self.veth_host}" | awk '{{ if ($6 == "{self.veth_host}") print $2 }}'
+'''
```
daniel-goldstein, line 139: Could we possibly get both values from the same exec? Especially considering both iptables calls are the same it would be nice to do this in 1.

batch/batch/worker/worker.py
```diff
@@ -292,6 +293,14 @@ async def enable_iptables_forwarding(self):
 iptables -w {IPTABLES_WAIT_TIMEOUT_SECS} --append FORWARD --out-interface {self.veth_host} --in-interface {self.veth_host} --jump ACCEPT'''
         )
 
+    async def mark_packets(self):
+        await check_shell(
+            f'''
+iptables -t mangle -A PREROUTING --in-interface {self.veth_host} -j MARK --set-mark 10
+iptables -t mangle -A POSTROUTING --out-interface {self.veth_host} -j MARK --set-mark 11
```
daniel-goldstein, line 300: Do you know what happens to this table after the veths are deleted? I would hope that those rows get removed as part of a sort of cascading delete but would be good to verify we don't need to do any sort of cleanup here.

batch/batch/resource_usage.py
```diff
@@ -29,19 +31,47 @@ def decode_to_df(data: bytes) -> Optional[pd.DataFrame]:
...
+
         return pd.DataFrame.from_records(np_array)
 
-    def __init__(self, container_name: str, output_file_path: str):
+    def __init__(
+        self,
+        container_name: str,
+        container_overlay: str,
+        io_volume_mount: Optional[str],
+        veth_host: Optional[str],
```
daniel-goldstein, line 59: I don't think `veth_host` needs to be optional

batch/batch/resource_usage.py
```diff
@@ -82,6 +112,58 @@ def memory_usage_bytes(self) -> Optional[int]:
...
+    def overlay_storage_usage_bytes(self) -> int:
+        return shutil.disk_usage(self.container_overlay).used
+
+    def io_storage_usage_bytes(self) -> int:
+        if self.io_volume_mount is not None:
+            return shutil.disk_usage(self.io_volume_mount).used
+        return 0
+
+    def is_attached_disk(self) -> bool:
+        return self.io_volume_mount is not None and os.path.ismount(self.io_volume_mount)
```
daniel-goldstein, line 124: nit: could probably just make this a private field instead of a method and then just call `os.path.ismount` once (not that it would really matter for performance).

daniel-goldstein (dismissed):
These charts are sweet! Mostly just have a couple of nits and questions.

jigold:
This should be all set. I tested backwards compatibility and caught some errors that showed up in the worker logs with regards to iptables lock errors with simultaneous shell calls.

batch/batch/resource_usage.py
```diff
@@ -82,6 +117,60 @@ def memory_usage_bytes(self) -> Optional[int]:
...
+    def overlay_storage_usage_bytes(self) -> int:
+        return shutil.disk_usage(self.container_overlay).used
+
+    def io_storage_usage_bytes(self) -> int:
+        if self.io_volume_mount is not None:
+            return shutil.disk_usage(self.io_volume_mount).used
+        return 0
+
+    async def network_bandwidth(self) -> Tuple[Optional[float], Optional[float]]:
+        async with iptables_lock:
```
daniel-goldstein, line 129: How come you don't need this lock elsewhere in `worker.py` when invoking iptables?

daniel-goldstein (changes requested):
Just one question. Though it feels probably outside of the scope of this PR, it feels worth considering using a library with [python bindings to iptables](https://python-iptables.readthedocs.io/en/latest/intro.html) instead of all this shelling out and stdout parsing.

batch/batch/resource_usage.py
jigold: I think you solved this with `-w {MAX_IPTABLES_WAIT_SECONDS}` in worker.py. It blocks for 60 seconds waiting for the lock

I can switch to `-w {MAX_IPTABLES_WAIT_SECONDS}` in resource_usage.py. I think having a global iptables lock is more complicated because resource_usage.py is used by the front end as well.

Let me know what you prefer.

jigold:
We could use the library, but I thought (1) we were trying to be dependent on external libraries and (2) it's easier for me to see what's being run with the shell commands as it's easier to search for examples.

batch/batch/resource_usage.py
daniel-goldstein: Ah I see. Let's just stay consistent by using the wait then.

daniel-goldstein:
That's a fair argument, and I suppose we would probably face the locking issues either way.

danking:
I don't really put this in the same category as other external libraries. It's just bindings to the C API that `iptables` itself uses. In general, I think we should prefer just directly calling C libraries instead of shelling out and parsing strings, but I doubt this will have a major impact on worker speed.

I worry a bit about the reliability of parsing text, but `iptables` seems like the sort of core utility that would be machine-parsable.

danking:
Heh, `libiptc` was ["never meant as a public API"](https://www.netfilter.org/documentation/FAQ/netfilter-faq-4.html#ss4.5)

daniel-goldstein:
Interesting, looks like the [go](https://github.com/coreos/go-iptables/blob/d2b8608923d15b0800af7d9f4bb6dea90e03b7d5/iptables/iptables.go#L544) and [rust](https://github.com/yaa110/rust-iptables/blob/3a24f420fb08281fd3655e50820a0a1eee0213ac/src/lib.rs#L416) "bindings" (well, the ones that come up first in a google search) just shell out too

batch/batch/worker/worker.py
```diff
@@ -352,6 +364,9 @@ def free(self, netns: NetworkNamespace):
 
     async def _free(self, netns: NetworkNamespace):
         await netns.cleanup()
+        await netns.create_netns()
+        await netns.enable_iptables_forwarding()
+        await netns.mark_packets()
```
daniel-goldstein, line 369: These setup steps are duplicated in `init`. Let's pull them out into a separate function so it's not as easy to include a step in one place and not the other. I had initially tried to leave the iptables forwarding rules in place so we're not constantly deleting and recreating the same rules, but I might be misguided that that was a worthwhile performance shortcut. Do you need to delete and recreate the mangle rules for correctness in the bookkeeping?

daniel-goldstein (changes requested):

batch/batch/worker/worker.py
jigold: No - it's just a counter and we look at the difference between two time points rather than the absolute number. I added that because I thought it was a change you requested previously. Let me know what you prefer.

daniel-goldstein: Sorry for the confusion, I wanted to make sure that there was no garbage that would grow unbounded, and forgot about what I had done with the other iptables rules. Regardless though, leaving these rules in would not grow unbounded since we reuse the same veths and only need to create the rules once in the init. Given that we face all this contention with iptables, let's not add repeated work and just do the packet-marking once in `init`. Sorry!

jigold: ok. I reverted everything back to the old behavior and we just add the new mangle rules on init now.

daniel-goldstein (approved):

PR #12543, opened 2022-12-07
[vds] Fix local_to_global to support non-ascending local alleles

### LINKED ISSUE #12534: `hl.vds.local_to_global` should support non-increasing local alleles
Nothing in the VDS representation disallows unsorted LA.

tpoterba:
Fixes #12534


hail/python/hail/vds/functions.py
```diff
@@ -82,21 +82,18 @@ def local_to_global(array, local_alleles, n_alleles, fill_value, number):
     -------
```
danking, line 82: We should have an example that uses the fields of a VDS. Do we have a VDS in the tests we can use for that purpose?

hail/python/test/hail/vds/test_vds_functions.py
```diff
@@ -39,3 +39,13 @@ def test_local_to_global():
...
+
+def test_local_to_global_alleles_non_increasing():
+    local_alleles = [0, 3, 1]
+    lad = [1, 10, 9]
+    lpl = [1001, 1004, 0, 1002, 1003, 1005]
+
+    assert hl.eval(hl.vds.local_to_global(lad, local_alleles, 4, 0, number='R')) == [1, 9, 0, 10]
+    assert hl.eval(hl.vds.local_to_global(lpl, local_alleles, 4, 999, number='G')) == [1001, 1002, 1005, 999, 999, 999, 1004, 1003, 999, 0]
+
+    assert hl.eval(hl.vds.local_to_global([0, 1, 2, 3, 4, 5], [0, 2, 1], 3, 0, number='G')) == [0, 3, 5, 1, 4, 2]
```
danking, line 51: Hmm. I think I'm missing something. Do I have a step wrong here:
1. `1`, as a local GT triangle number, definitionally equivalent to
2. `0/1` in local alleles which is
3. `0/2` in global alleles (`LA[1] == 2` in this example)
4. `2` in global GT triangle number. 

---

local genotypes triangle numbers:

    0    1    2
         3    4
              5

local genotypes as local alleles:

    0/0  0/1  0/2
         1/1  1/2
              2/2

local genotypes as global alleles:

    0/0  0/2  0/1
         2/2  2/1
              1/1

local genotypes as global triangle numbers:

      0    2    1
           5    4
                3

tpoterba: The triangles go the other way:

```
0/0
0/1 1/1
0/2 1/2 2/2
```

```
0
1 2
3 4 5
```

The intuition here is that the addition of a `2` allele shouldn't affect any indices of calls involving only 0 and 1 alleles.

chrisvittal (approved):

PR #12544, opened 2022-12-07
[service-base] fix libsass by downgrading google-cloud-profiler

danking:


docker/service-base-requirements.txt
```diff
@@ -1 +1,3 @@
-google-cloud-profiler==4.0.0
+# https://github.com/sass/libsass-python/issues/413
+google-cloud-profiler<4.0.0
+protobuf<=3.20.0
```
daniel-goldstein, line 3: protobuf is already in the docker pinned requirements as a transitive dependency, so I'm guessing this will downgrade it. That should probably be pinned in `docker/requirements.txt` instead.

daniel-goldstein (dismissed):

daniel-goldstein (approved):

daniel-goldstein:
@danking I'd apply this patch. Not sure why we install those dependencies in the auth dockerfile, they're already in the docker requirements:

```diff
diff --git a/auth/Dockerfile b/auth/Dockerfile
index 0c2bfa4dad..a94928c697 100644
--- a/auth/Dockerfile
+++ b/auth/Dockerfile
@@ -1,9 +1,5 @@
 FROM {{ service_base_image.image }}
 
-RUN hail-pip-install \
-      google-auth-oauthlib==0.4.6 \
-      google-auth==1.25.0
-
 COPY auth/setup.py auth/MANIFEST.in /auth/
 COPY auth/auth /auth/auth/
 RUN hail-pip-install /auth && rm -rf /auth
```

PR #12552, opened 2022-12-09
[qob][query-local] fix custom reference writing

danking:


hail/python/test/hail/genetics/test_reference_genome.py
```diff
@@ -1,174 +1,186 @@
-import unittest
-
 import hail as hl
 from hail.genetics import *
 from ..helpers import *
 from hail.utils import FatalError
 
 
-class Tests(unittest.TestCase):
```
danking, line 9: Sorry for the diff. I'm on a mission to excise unittest.

chrisvittal: Why?

chrisvittal (approved):

hail/python/test/hail/genetics/test_reference_genome.py
danking: I dislike the extra four spaces in front of each test case. I'd rather isolate tests by file than class.

PR #12553, opened 2022-12-13
[vds] reorder to_merged_sparse_mt

tpoterba:
Putting reference fields first is better if you're going to export a SVCR VCF afterwards.


chrisvittal (approved):

danking:
Hopefully I resolved the conflict correctly.

PR #12554, opened 2022-12-13
[batch] Omit setup progress bars on small batch submission

daniel-goldstein:
I also noticed a bug in the process where the python client uses create-fast only if there is one *job* not one *bunch*. I fixed that and threw in a test.


danking (approved):
Excellent changes.

PR #12555, opened 2022-12-13
[query] update to latest google-cloud-storage

danking:


danking:
I am so glad I added those requester pays tests. They changed the exception type for requester pays failures and that broke our try-catch. The requester pays situation in GCP is so harebrained.

PR #12558, opened 2022-12-14
[website] Clean up use of scope instead of deploy

daniel-goldstein:
I noticed in #12526 that you added `scope` to the website Makefile but I don't think it's needed as using `deploy` is a little cleaner.


danking (approved):
Nice!

PR #12561, opened 2022-12-14
[batch] Fix JVM kill exit stack on worker shutdown

daniel-goldstein:
Since `jvm.kill` is async, using `callback` didn't do anything. This meant every worker that deactivated would spew a bunch of dangling tasks, wait for ten minutes, then exit. We probably don't notice this because the driver kills inactive workers pretty frequently, but I ran into this while doing terra work before I was having the driver delete workers.


danking (approved):
Good find.

PR #12566, opened 2022-12-16
[query] add a small & incomplete test suite for good index errors

danking:
CHANGELOG: Improve error message when combining incompatibly indexed fields in certain operations including array indexing.

See test cases for straightforward examples. In main, none of the test code triggers errors. You have to execute the table to actually trigger an error in IR serialization which will reference `"sa"` which is totally meaningless to the user (let alone many Hail engineers).


tpoterba (approved):
thanks!

PR #12572, opened 2023-01-03
[batch] Dont consider update time when deleting broken instances

daniel-goldstein:
The driver will regularly update the VM state from the azure API so including the update time can block the driver from deleting runaway instances, which can subsequently halt the system.


jigold (approved):
I agree your change is an improvement if we say any instance with more than 5 failed requests is always broken.

Just to summarize for future reference, the reason that the time `last_updated` error catching didn't work is because the instance was active and the VM was running, so the time `last_updated` at the end of the massive if-else statement resulted in the time `last_updated` being updated despite there being a problem. We might want to rethink this logic in the future if we run into another scenario where there is an edge case that we're missing.  I double checked and the failed requests increment only happens if the instance has previously been 'active'.

PR #12575, opened 2023-01-05
[batch] Make driver parameters configurable in the UI

jigold:
I tested this in my namespace and everything seems to be working. I tried really hard to keep the order of the parameters the same everywhere to avoid putting values in the wrong places (i.e. 7200 for max_instances instead of for standing worker idle time).


batch/batch/driver/instance_collection/job_private.py
```diff
@@ -109,28 +109,58 @@ def __init__(
         self.exceeded_shares_counter = ExceededSharesCounter()
 
         self.boot_disk_size_gb = config.boot_disk_size_gb
+        self.max_instances_per_autoscaler_loop = config.max_instances_per_autoscaler_loop
```
daniel-goldstein, line 112: I prefer `max_new_instances_per_autoscaler_loop`

batch/batch/driver/instance_collection/job_private.py
```diff
@@ -308,7 +338,10 @@ async def create_instances_loop_body(self):
             should_wait = True
             return should_wait
         user_share = {
-            user: max(int(300 * resources['n_allocated_jobs'] / total + 0.5), 20)
+            user: min(
+                int(self.max_instances_per_autoscaler_loop * resources['n_allocated_jobs'] / total + 0.5),
+                resources['n_allocated_jobs'],
+            )
```
daniel-goldstein, line 344: Why did you change the semantics here? I thought this was just a PR to make these values configurable.

batch/batch/driver/instance_collection/pool.py
```diff
@@ -428,7 +437,12 @@ async def create_instances(self):
             AUTOSCALER_HEAD_JOB_QUEUE_READY_CORES.labels(pool_name=self.name, region=region).set(ready_cores_mcpu)
 
     async def control_loop(self):
-        await periodically_call(AUTOSCALER_LOOP_PERIOD_SECONDS, self.create_instances)
```
daniel-goldstein, line 431: Without a clarifying comment, I could easily see myself changing this back to using `periodically_call` because I don't pick up on the subtlety that `self.autoscaler_loop_period_secs` is subject to change. I would either add an explicit comment here, write something like `periodically_call_dynamic_sleep(lambda: self.autoscaler_loop_period_secs, self.create_instances)`, or make `periodically_call` accept a `Union[int, Callable[[], int]]` and continue using `periodically_call`.

daniel-goldstein (changes requested):
> I tried really hard to keep the order of the parameters the same everywhere to avoid putting values in the wrong places (i.e. 7200 for max_instances instead of for standing worker idle time).

Can we find a way to not have to think this hard? I think these should be keyword-only arguments wherever possible so the only place you have to line things up properly is in sql.

batch/batch/driver/instance_collection/job_private.py
jigold: It was originally to make the parameters configurable. But the current behavior wasn't correct as it would spin up a constant number of instances that wasn't configurable (I believe 300 and/ or 20 per user). I can make the change separately, but felt that it was appropriate here in the context of the overall goal to make these parameters configurable. I didn't want it to look like this parameter was configurable for job-private, but in reality it does nothing.

Your call on whether to make a separate PR.

jigold: Actually, on second thought, the reason this code was like this in the first place was because the job private manager shouldn't care how quickly it ramps up as it's a job per worker. So 300 was the max for GCP rate limits. I think this change is still good though as it will be consistent with the pool.

batch/batch/inst_coll_config.py
```diff
@@ -220,7 +220,7 @@ def __init__(
         boot_disk_size_gb: int,
```
daniel-goldstein, line 220: Can we force these to be keyword-only?

daniel-goldstein (changes requested):
One last nit

jigold:
Please keep the WIP tag on so I can check the behavior after it merges in case there's a problem.

daniel-goldstein (approved):

PR #12576, opened 2023-01-06
[query] Support zstd compression in BGEN files

tpoterba:
CHANGELOG: `hl.import_bgen` and `hl.export_bgen` now support compression with Zstd.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
@@ -9,10 +9,11 @@ import is.hail.expr.ir.lowering.{LowererUnsupportedOperation, TableStage}
 import is.hail.expr.ir.streams.StreamProducer
 import is.hail.expr.{JSONAnnotationImpex, Nat}
 import is.hail.io._
+import is.hail.io.bgen.BgenSettings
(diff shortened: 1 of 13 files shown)


hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
```diff
@@ -972,8 +973,13 @@ case class MatrixBGENWriter(
     else
       path + ".bgen"
 
+    assert(compressionCodec == "zlib" || compressionCodec == "zstd")
     val writeHeader = exportType == ExportType.PARALLEL_HEADER_IN_SHARD
-    val partWriter = BGENPartitionWriter(tm, entriesFieldName, writeHeader)
+    val compressionInt = compressionCodec match {
```
chrisvittal, line 978: Is there any way to make this some kind of enum, or if that's too much, named constants?

chrisvittal (changes requested):
One small maintainability issue around the ids for compression.

hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
tpoterba: done.

chrisvittal (approved):

PR #12578, opened 2023-01-06
[docker] Trim down service images

daniel-goldstein:
This is a total overhaul of our docker images. Though very verbose, I tried to stick to these main tenets:

- Any docker image has exactly 1 layer in it (all the way down to ubuntu) that installs pip dependencies. This primarily aims to protect the cache for this particularly large layer and also avoids a later layer silently upgrading the version of a dependency installed in an earlier layer. This pairs nicely with the following goal
- We only ever use 1 version of a dependency across the monorepo. Liberal use of pip's [constraint files](https://pip.pypa.io/en/stable/user_guide/#constraints-files) to ensure that the dependencies for a service must be compatible with dependencies from hail. The `install-dev-dependencies` target which install all our pinned requirements files would tell you if there's any incompatible versions of transitive dependencies across the repo
- The image graph is shallow and images don't contain more than they need. In order to have a single layer with requirements and hail code on top, I moved the service images to just be based on hail-ubuntu. This shortens the critical path and therefore reduces total image building time by reducing the number of times our image data needs to be downloaded and re-uploaded to the registry. I also removed a lot of unnecessary cruft like gcloud in places it wasn't used anymore, some unused/unnecessary pip requirements, etc.


danking:
Just starting to explore this PR. I compared this in-progress CI run to 7126798 (from #12737). The time to service backend starting is ~7 minutes. In the other PR, its ~8 minutes. I suppose that's because this PR isn't hitting any caches, right?

Hmm, it also seems like the critical path to the service backend test is through `build_hail_jar_and_wheel_only`. I wonder if we double the cores, would the time halve? On my laptop a fresh build is like 3m.

Screen Shot 2023-03-07 at 10 27 05

Can you also get rid of hail_base_image?

danking:
This is a great PR, I'll dig in more later!

daniel-goldstein:
> Just starting to explore this PR. I compared this in-progress CI run to 7126798 (from #12737). The time to service backend starting is ~7 minutes. In the other PR, its ~8 minutes. I suppose that's because this PR isn't hitting any caches, right?
> 
> Hmm, it also seems like the critical path to the service backend test is through `build_hail_jar_and_wheel_only`. I wonder if we double the cores, would the time halve? On my laptop a fresh build is like 3m.
> 
> Screen Shot 2023-03-07 at 10 27 05

Ya I really only focus here on docker image steps, not on the overall critical path. But I like the idea of adding more cores I'll try that.

daniel-goldstein:
Looking at the auth image, it is down from 2.76GB in main to 675MB. The hail-ubuntu image underneath it has stayed basically the same at about half the auth image. Nearly all of the 674MB is split evenly between the layer that installs python in hail-ubuntu and the layer that installs the pip dependencies in the auth image. I've not yet inspected the hail-ubuntu layer, but for the pip dependencies the main offenders are:

```
77M	/usr/local/lib/python3.7/dist-packages/googleapiclient
76M	/usr/local/lib/python3.7/dist-packages/botocore
33M	/usr/local/lib/python3.7/dist-packages/_sass.abi3.so
29M	/usr/local/lib/python3.7/dist-packages/kubernetes_asyncio
20M	/usr/local/lib/python3.7/dist-packages/uvloop
14M	/usr/local/lib/python3.7/dist-packages/pip
14M	/usr/local/lib/python3.7/dist-packages/cryptography
8.9M	/usr/local/lib/python3.7/dist-packages/google
7.9M	/usr/local/lib/python3.7/dist-packages/pygments
7.0M	/usr/local/lib/python3.7/dist-packages/azure
5.0M	/usr/local/lib/python3.7/dist-packages/setuptools
4.2M	/usr/local/lib/python3.7/dist-packages/aiohttp
2.5M	/usr/local/lib/python3.7/dist-packages/googlecloudprofiler
2.2M	/usr/local/lib/python3.7/dist-packages/yaml
2.2M	/usr/local/lib/python3.7/dist-packages/hailtop
2.0M	/usr/local/lib/python3.7/dist-packages/rich
1.6M	/usr/local/lib/python3.7/dist-packages/pyasn1_modules
1.5M	/usr/local/lib/python3.7/dist-packages/boto3
1.4M	/usr/local/lib/python3.7/dist-packages/pkg_resources
1.4M	/usr/local/lib/python3.7/dist-packages/oauthlib
1.1M	/usr/local/lib/python3.7/dist-packages/pycparser
```

Most of a gigabyte still feels annoyingly bloated but maybe it's good enough for now and is certainly an improvement.

danking:
Stunning that google and amazon are each 10x the size of azure. And it's just, like, a bunch of classes to make HTTP requests and parse the responses.

daniel-goldstein:
Not even, apparently nearly all of the Amazon bloat is json files.

With Google they even recommend not to use that monolith library anymore and instead to use newer per-service libraries, but guess who didn't listen: googlecloudprofiler

ci/Dockerfile.ci-utils
```diff
@@ -1,3 +1,27 @@
+FROM {{ hail_ubuntu_image.image }} AS base
+
+# source: https://cloud.google.com/storage/docs/gsutil_install#linux
+RUN curl --remote-name https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-334.0.0-linux-x86_64.tar.gz && \
+    tar -xf google-cloud-sdk-334.0.0-linux-x86_64.tar.gz && \
+    curl --remote-name https://dl.k8s.io/release/v1.21.14/bin/linux/amd64/kubectl && \
+    install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+ENV PATH $PATH:/google-cloud-sdk/bin
+
+# https://bugs.mysql.com/bug.php?id=105288&thanks=sub
```
danking, line 10: This appears to now be fixed. We should update to latest mysql in a follow up PR. Can you do that?

ci/test/resources/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM {{ service_base_image.image }}
+FROM {{ ci_utils_image.image }}
```
danking, line 1: `hello` just needs an image with gear and hail top. I suppose building ci_utils and using it is fine though.

docker/Dockerfile.base
```diff
@@ -1,21 +1,13 @@
...
-    tar -xf google-cloud-sdk-334.0.0-linux-x86_64.tar.gz && \
-    curl --remote-name https://dl.k8s.io/release/v1.21.14/bin/linux/amd64/kubectl && \
-    install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
-ENV PATH $PATH:/google-cloud-sdk/bin
-
 RUN hail-apt-get-install \
     git \
     htop \
     unzip bzip2 zip tar \
-    wget curl \
```
danking, line 14: We only use `wget` in `Dockerfile.worker` and we ought to just use curl there. (We also install a bunch of deps in azure that seem unnecessary? cmake? pkg-config? TLS things? shouldn't libfuse-dev be sufficient? We should probably just delete all this and update to blobfuse2).

hail/Makefile
```diff
@@ -272,19 +272,9 @@ $(eval $(call ENV_VAR,cloud_base))
...
-	HAIL_HAIL_DIR=$(HAIL_HAIL_DIR) \
-	../generate_pip_lockfile.sh python/requirements.txt python/pinned-requirements.txt && \
-	HAIL_HAIL_DIR=$(HAIL_HAIL_DIR) \
-	../generate_pip_lockfile.sh python/dev/requirements.txt python/dev/pinned-requirements.txt
-
-
 .PHONY: check-pip-lockfile
 check-pip-lockfile:
-	$(PIP) install pip-tools && bash ../check_pip_requirements.sh python/requirements.txt python/pinned-requirements.txt
+	$(PIP) install pip-tools && bash ../check_pip_requirements.sh python
```
danking, line 277: So, `pytest` depends on INIT_SCRIPTS which depends on this which means I have to invoke `pip` to do a no-op install and then invoke pip-compile. Not great. It's been bothering me lately when I'm doing iterative development using the tests. Can we check the pip lock file once and be done with it?

hail/generate_splits.py
```diff
@@ -42,9 +41,17 @@ def partition(k, ls):
 
 splits = partition(n_splits, classes)
 
-with open('testng-splits.xml', 'r') as f:
-    template = jinja2.Template(f.read(), undefined=jinja2.StrictUndefined, trim_blocks=True, lstrip_blocks=True)
-
+newline = '\n'
```
danking, line 44: deadcode

hail/python/hailtop/batch/backend.py
```diff
@@ -30,6 +30,7 @@
 
 
 HAIL_GENETICS_HAILTOP_IMAGE = os.environ.get('HAIL_GENETICS_HAILTOP_IMAGE', f'hailgenetics/hailtop:{pip_version()}')
+HAIL_GENETICS_HAIL_IMAGE = os.environ.get('HAIL_GENETICS_HAIL_IMAGE', f'hailgenetics/hail:{pip_version()}')
```
danking, line 33: unused?

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -10,6 +10,7 @@
 import re
 
 from hailtop.batch import Batch, ServiceBackend, LocalBackend
+from hailtop.batch.backend import HAIL_GENETICS_HAIL_IMAGE
```
danking, line 13: this is redefined below

hail/python/hailtop/batch/backend.py
danking: ah, I see its used in the tests. any reason it needs to be here too?

tls/Dockerfile
```diff
@@ -1,17 +1,12 @@
 FROM {{ hail_ubuntu_image.image }}
 
-# source: https://cloud.google.com/storage/docs/gsutil_install#linux
 # re: RANDFILE, https://github.com/openssl/openssl/issues/7754#issuecomment-444063355
 # jdk not strictly necessary, but we want keytool
-RUN curl -sSLO https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-334.0.0-linux-x86_64.tar.gz && \
-    tar -xf google-cloud-sdk-334.0.0-linux-x86_64.tar.gz && \
-    curl -sSLO https://dl.k8s.io/release/v1.21.14/bin/linux/amd64/kubectl && \
+RUN curl -sSLO https://dl.k8s.io/release/v1.21.14/bin/linux/amd64/kubectl && \
```
danking, line 5: `-sS` is unnecessary, the curlrc defaults to these

danking (dismissed):
A tour de force, but a great change. A few comments.

hail/Makefile
daniel-goldstein: This has been bothering me too. Yes please.

batch/test/test_dag.py
```diff
@@ -116,34 +114,37 @@ def test_cancel_left_after_tail(client):
         assert node_status['state'] == 'Cancelled', str((node_status, batch.debug_info()))
 
 
-def test_callback(client):
-    from flask import Flask, request
+async def test_callback(client):
```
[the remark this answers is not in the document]
daniel-goldstein: The pytest upgrade we did a while back included [auto-recognizing async tests](https://github.com/hail-is/hail/blob/45c8993fd90156c14c1364fc24cdbc8c2f925655/hail/python/test/pytest.ini#L2). To be sure though you can see it ran successfully [here](https://ci.hail.is/batches/7134459/jobs/119). Getting rid of Flask here was especially satisfying.

hail/Makefile
daniel-goldstein: We primarily added this `check-pip-lockfile` target to make sure that someone running `install-on-cluster` doesn't add some requirements and then get surprised when they're not installed because they didn't change the lockfiles. This is already a dependency of `install-on-cluster` so I just removed it as a dependency of the `deploy.yaml`.

danking (approved):
serve me the spicy meatball!

PR #12581, opened 2023-01-06
[qob] Add regions parameter

jigold:
CHANGELOG: Added the regions parameter to `hl.init()` for specifying which regions Query on Batch jobs should run in.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2667,7 +2667,7 @@ steps:
 
       BATCH_ID=$(hailctl batch submit simple_hail.py --name=test-hailctl-batch-submit --files=foo -o json | jq '.id')
       STATUS=$(hailctl batch wait -o json $BATCH_ID)
-      STATE=$(jq -jr '.state' $STATUS)
+      STATE=$(echo $STATUS | jq -jr '.state')
       if [ "$STATE" == "success" ]; then
           exit 0;
       else
@@ -2688,7 +2688,7 @@ steps:
 
       BATCH_ID=$(hailctl batch submit --name=test-hailctl-batch-submit --files=foo -o json hail_with_args.py 100 | jq '.id')
       STATUS=$(hailctl batch wait -o json $BATCH_ID | jq -jr '.state')
-      STATE=$(jq -jr '.state' $STATUS)
+      STATE=$(echo $STATUS | jq -jr '.state')
       if [ "$STATE" == "success" ]; then
           exit 0;
       else
@@ -2705,7 +2705,7 @@ steps:
 
       BATCH_ID=$(hailctl batch submit --name=test-hailctl-batch-submit --files=foo -o json file.sh | jq '.id')
       STATUS=$(hailctl batch wait -o json $BATCH_ID)
-      STATE=$(jq -jr '.state' $STATUS)
+      STATE=$(echo $STATUS | jq -jr '.state')
       if [ "$STATE" == "success" ]; then
           exit 0;
       else
@@ -2724,7 +2724,7 @@ steps:
 
       BATCH_ID=$(hailctl batch submit --name=test-hailctl-batch-submit --files=foo -o json file-with-args.sh abc 123 | jq '.id')
       STATUS=$(hailctl batch wait -o json $BATCH_ID)
-      STATE=$(jq -jr '.state' $STATUS)
+      STATE=$(echo $STATUS | jq -jr '.state')
       if [ "$STATE" == "success" ]; then
           exit 0;
       else
--- a/build.yaml
+++ b/build.yaml
@@ -1227,6 +1227,7 @@ steps:
       export HAIL_TEST_RESOURCES_DIR=./resources
       export HAIL_DOCTEST_DATA_DIR=./data
       export HAIL_TEST_STORAGE_URI={{ global.test_storage_uri }}/{{ token }}
+      export HAIL_CLOUD={{ global.cloud }}
       export PYSPARK_SUBMIT_ARGS="--driver-memory 6g pyspark-shell"
       python3 -m pytest \
               -m unchecked_allocator \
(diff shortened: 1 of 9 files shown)


hail/python/hail/backend/service_backend.py
```diff
@@ -157,6 +157,8 @@ async def serialize(self, writer: afs.WritableStream):
 
 
 class ServiceBackend(Backend):
+    input_version = 2
```
daniel-goldstein, line 160: How come this versioning here? Seems arbitrary, is it ever checked? Are there instructions on how or when it should be updated?

daniel-goldstein (changes requested):

hail/python/hail/backend/service_backend.py
jigold: Sorry. It was old cruft from when I thought I had to make the spec backwards compatible.

hail/python/test/hail/backend/test_service_backend.py
```diff
@@ -1,8 +1,13 @@
+import os
+
 import hail as hl
 
 from ..helpers import skip_unless_service_backend
 from hail.backend.service_backend import ServiceBackend
 
+CLOUD = os.environ['HAIL_CLOUD']
```
daniel-goldstein, line 8: Does this impede running other query tests locally? Not sure if it does but it would be annoying to have to specify `HAIL_CLOUD` to run tests that have nothing to do with it. I'd prefer in that case to put the `os.environ` call in the method that actually needs it and remove the env variable from test steps that aren't testing the service backend.

daniel-goldstein (changes requested):
Sorry one more thing I noticed.

build.yaml
```diff
@@ -2764,9 +2776,11 @@ steps:
 
       BATCH_ID=$(hailctl batch submit --name=test-hailctl-batch-submit --files=foo -o json hail_with_args.py 100 | jq '.id')
       STATUS=$(hailctl batch wait -o json $BATCH_ID | jq -jr '.state')
-      if [ "$STATUS" == "success" ]; then
+      STATE=$(echo $STATUS | jq -jr '.state')
```
daniel-goldstein, line 2779: how come this filters through the state twice?

hail/python/test/hail/backend/test_service_backend.py
```diff
@@ -1,8 +1,13 @@
+import os
+
 import hail as hl
 
 from ..helpers import skip_unless_service_backend
 from hail.backend.service_backend import ServiceBackend
 
+CLOUD = os.environ.get('HAIL_CLOUD')
```
daniel-goldstein, line 8: This is going to treat `None` and Azure the same. Can we keep this as `os.environ['HAIL_CLOUD']` but just put this line inside the relevant test?

hail/python/hail/backend/service_backend.py
```diff
@@ -214,6 +215,12 @@ async def create(*,
         worker_memory = configuration_of('query', 'batch_worker_memory', worker_memory, None)
         name_prefix = configuration_of('query', 'name_prefix', name_prefix, '')
 
+        regions_str = configuration_of('batch', 'regions', regions, None)
+        if regions_str is not None:
+            regions = regions_str.split(',')
+        else:
+            regions = bc.supported_regions()
```
daniel-goldstein, line 222: The regions parameter below is `Optional[List[str]]`, but from this else case it looks like we don't exercise the `None` option. I don't have a strong opinion about whether to do one or the other, but I think we should either make `regions` not Optional, or keep this None instead of using `supported_regions`.

hail/python/hail/backend/service_backend.py
```diff
@@ -329,6 +340,9 @@ async def _rpc(self,
                         await write_str(infile, orjson.dumps(reference_config).decode('utf-8'))
                     await write_str(infile, str(self.worker_cores))
                     await write_str(infile, str(self.worker_memory))
+                    await write_int(infile, len(self.regions))
```
daniel-goldstein, line 343: This breaks if `self.regions` is None, so would need to change if we allow this to be optional

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -542,6 +548,19 @@ class ServiceBackendSocketAPI2(
...
+      val regionsArrayBuffer = mutable.ArrayBuffer[String]()
+      while (nRegions > 0) {
+        val region = readString()
+        regionsArrayBuffer += region
+        nRegions -= 1
+      }
+      Some(regionsArrayBuffer.toArray)
+    } else {
+      None
+    }
```
daniel-goldstein, line 562: When would the `service_backend` ever send an empty list?

daniel-goldstein (dismissed):

hail/python/hail/backend/service_backend.py
jigold: I think we should try and keep the same interface as `hailtop.batch` as much as possible for consistency reasons. This would mean we'd need to expose the `ServiceBackend.supported_regions` and the equivalent of `ServiceBackend.ANY_REGION`. I think they should be `hl.supported_regions()` that is implemented as a static method on the service backend when applicable and `hl.ANY_REGION` is just an object. Once we figure this out, then we can figure out what makes sense for the types further downstream.

daniel-goldstein: > I think we should try and keep the same interface as hailtop.batch as much as possible for consistency reasons.

Totally agree with this, but I see the constructor of the `ServiceBackend` object and its implementation as private, so I don't see why we can't currently write the type as it's being used -- if it's never `None` let's use that knowledge to cut down on None checks that are currently dead code.

daniel-goldstein: I agree that the user-facing API should be as close to batch as possible, but the `ServiceBackend` for query is not a user-facing API. Users only use `hl.init` which select the backend to use.

daniel-goldstein (dismissed):
Left further comments in the thread.

daniel-goldstein (dismissed):
Requesting changes so it's off my CI page. Looks fine there's just merge conflicts now.

jigold:
I know you've got a lot on your plate, but it would be awesome if we could get this in tomorrow or early next week. Of all my remaining PRs, I think this one is highest priority and should be close to being mergeable.

daniel-goldstein (approved):
Thanks for the reminder. This looks good

PR #12583, opened 2023-01-09
[query] Fix _to_relational for col_key

tpoterba:
Fixes the col_key.show() issue Dan unearthed.

### DIFF
--- a/hail/python/hail/expr/expressions/base_expression.py
+++ b/hail/python/hail/expr/expressions/base_expression.py
@@ -946,7 +946,7 @@ class Expression(object):
             if self is source.col:
                 return None, source.key_cols_by().cols()
             if self is source.col_key:
-                return None, source.cols().select()
+                return None, source.select_cols().key_cols_by().cols()
             if self is source.entry:
                 return None, source.select_rows().select_cols()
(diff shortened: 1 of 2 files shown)


hail/python/hail/expr/expressions/base_expression.py
```diff
@@ -946,7 +946,7 @@ def _to_relational_preserving_rows_and_cols(self, fallback_name):
             if self is source.col:
                 return None, source.key_cols_by().cols()
             if self is source.col_key:
-                return None, source.key_cols_by().cols().select()
```
danking, line 949: Won't this now produce a warning every time we call `_to_relational_...`, that seems bad. Can we somehow fix this without triggering the column table is sorted blah blah warning?

danking (changes requested):

hail/python/hail/expr/expressions/base_expression.py
tpoterba: initially disagreed, but this should produce an unkeyed thing the same as `col`. Fixed to `select_cols().key_cols_by().cols()`

danking (approved):

PR #12592, opened 2023-01-12
[qob] Better compiler stage names for QOB

tpoterba:


daniel-goldstein:
See [here](https://batch.azure.hail.is/batches/3423869) for an example of what this looks like

daniel-goldstein (approved):

PR #12596, opened 2023-01-13
[batch] Refresh credentials in docker retry loop

jigold:


batch/batch/worker/worker.py
```diff
@@ -443,9 +442,12 @@ async def _pull_image(self):
...
                 # FIXME improve the performance of this with a
                 # per-user image cache.
-                auth = self._current_user_access_token()
                 await docker_call_retry(
-                    MAX_DOCKER_IMAGE_PULL_SECS, str(self), docker.images.pull, self.image_ref_str, auth=auth
+                    MAX_DOCKER_IMAGE_PULL_SECS,
+                    str(self),
+                    docker.images.pull,
+                    self.image_ref_str,
+                    auth=self._current_user_access_token,
```
daniel-goldstein, line 450: I think this doesn't match the new signature because it's not async. But could just make `_current_user_access_token` async

daniel-goldstein (changes requested):
Great just one snag

daniel-goldstein (dismissed):

daniel-goldstein (approved):

PR #12602, opened 2023-01-17
[batch] Add job_id to UI search

jigold:
This is a feature Bob asked for to be able to search by job id in the UI. I added new functionality where if the "q" query string changes, we reset the last_job_id. Otherwise, something like "job_id=100" followed by "job_id=5" wouldn't work.


daniel-goldstein:
I'm not quite sure I understand the need for the last_query addition. The search bar doesn't send a last_job_id, only the Next page button. And there should never be a Next page button after a search for a particular job id. Could you explain more about what's broken here?

daniel-goldstein (dismissed):
See comment above.

jigold:
You're right. I forgot last_job_id isn't sent with the search form.

daniel-goldstein (approved):

PR #12607, opened 2023-01-18
[qob] Support custom references in Query-on-Batch

daniel-goldstein:


danking (approved):
Man, I love approving concise PRs that get the job done.

PR #12609, opened 2023-01-19
[batch] More graceful resource usage error handling

jigold:
This PR mitigates the 500 error Ben was getting on the UI page when a resource usage file was corrupt. It also implements a more graceful handling of an out of space error on the worker.

### DIFF
--- a/batch/batch/resource_usage.py
+++ b/batch/batch/resource_usage.py
@@ -34,6 +34,7 @@ class ResourceUsageMonitor:
         try:
             return ResourceUsageMonitor._decode_to_df(data)
         except Exception:
+            log.exception('corrupt resource usage file found', stack_info=True)
             return None
 
     @staticmethod
--- a/batch/batch/resource_usage.py
+++ b/batch/batch/resource_usage.py
@@ -1,4 +1,5 @@
 import asyncio
+import errno
 import logging
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


batch/batch/resource_usage.py
```diff
@@ -212,6 +220,11 @@ async def periodically_measure():
                 except asyncio.CancelledError:
                     cancelled = True
                     raise
+                except OSError as err:
+                    if err.errno == errno.ENOSPC:
+                        cancelled = True
```
jigold, line 225: I think this will kill the user's job. I'm not certain if we want that behavior or just to cancel the monitoring loop.

batch/batch/resource_usage.py
```diff
@@ -30,6 +31,13 @@ def version_to_bytes() -> bytes:
 
     @staticmethod
     def decode_to_df(data: bytes) -> Optional[pd.DataFrame]:
+        try:
+            return ResourceUsageMonitor._decode_to_df(data)
+        except Exception:
```
daniel-goldstein, line 36: Can we log the exception here so we're alerted?

batch/batch/resource_usage.py
daniel-goldstein: Why does this kill the user's job? Since this is happening in a separate task, I don't see anywhere where this exception is handled.

daniel-goldstein (changes requested):

jigold:
I think you're right.

daniel-goldstein (approved):

PR #12611, opened 2023-01-19
[batch] Dont use Batch identity on workers

daniel-goldstein:
- Use the worker VMs credentials for the FS and compute client instead of requesting credentials from the batch driver
- This allows creating the FS before worker activation, meaning that same FS can be passed to JVMContainers, instead of needing to create a new LocalAsyncFS for JVM objects
- Thread the task_manager through more of worker.py so the worker can properly cancel all tasks on shutdown

### DIFF
--- a/batch/batch/cloud/gcp/worker/disk.py
+++ b/batch/batch/cloud/gcp/worker/disk.py
@@ -11,22 +11,29 @@ log = logging.getLogger('disk')
 
 
 class GCPDisk(CloudDisk):
-    def __init__(self, name: str, zone: str, project: str, instance_name: str, size_in_gb: int, mount_path: str):
+    def __init__(
+        self,
+        name: str,
+        zone: str,
+        project: str,
+        instance_name: str,
+        size_in_gb: int,
+        mount_path: str,
+        compute_client: aiogoogle.GoogleComputeClient,  # BORROWED
+    ):
         assert size_in_gb >= 10
         # disk name must be 63 characters or less
         # https://cloud.google.com/compute/docs/reference/rest/v1/disks#resource:-disk
         # under the information for the name field
         assert len(name) <= 63
 
-        self.compute_client = aiogoogle.GoogleComputeClient(
-            project, credentials=aiogoogle.GoogleCredentials.from_file('/worker-key.json')
-        )
         self.name = name
         self.zone = zone
         self.project = project
         self.instance_name = instance_name
         self.size_in_gb = size_in_gb
         self.mount_path = mount_path
+        self.compute_client = compute_client
 
         self._created = False
         self._attached = False
@@ -48,9 +55,6 @@ class GCPDisk(CloudDisk):
                 await self._detach()
             finally:
                 await self._delete()
-
-    async def close(self):
-        await self.compute_client.close()
 
     async def _unmount(self):
         if self._attached:
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -15,20 +15,8 @@ import traceback
 import uuid
 import warnings
 from collections import defaultdict
-from contextlib import AsyncExitStack, ExitStack, contextmanager
-from typing import (
-    Any,
-    Awaitable,
-    Callable,
-    ContextManager,
-    Dict,
-    Iterator,
-    List,
-    MutableMapping,
-    Optional,
-    Tuple,
-    Union,
-)
+from contextlib import AsyncExitStack, ExitStack
+from typing import Any, Awaitable, Callable, ContextManager, Dict, List, MutableMapping, Optional, Tuple, Union
 
 import aiodocker  # type: ignore
 import aiodocker.images
@@ -41,7 +29,6 @@ from aiodocker.exceptions import DockerError  # type: ignore
 from aiohttp import web
 from sortedcontainers import SortedSet
 
-from gear.clients import get_cloud_async_fs, get_compute_client
 from hailtop import aiotools, httpx
 from hailtop.aiotools import AsyncFS, LocalAsyncFS
 from hailtop.aiotools.router_fs import RouterAsyncFS
@@ -79,6 +66,7 @@ from ..cloud.resource_utils import (
 )
 from ..file_store import FileStore
 from ..globals import HTTP_CLIENT_MAX_SIZE, RESERVED_STORAGE_GB_PER_CORE, STATUS_FORMAT_VERSION
+from ..instance_config import InstanceConfig
 from ..publicly_available_images import publicly_available_images
 from ..resource_usage import ResourceUsageMonitor
 from ..semaphore import FIFOWeightedSemaphore
@@ -173,7 +161,7 @@ assert UNRESERVED_WORKER_DATA_DISK_SIZE_GB >= 0
 ACCEPTABLE_QUERY_JAR_URL_PREFIX = os.environ['ACCEPTABLE_QUERY_JAR_URL_PREFIX']
 assert len(ACCEPTABLE_QUERY_JAR_URL_PREFIX) > 3  # x:// where x is one or more characters
 
-CLOUD_WORKER_API: CloudWorkerAPI = GCPWorkerAPI.from_env() if CLOUD == 'gcp' else AzureWorkerAPI.from_env()
+CLOUD_WORKER_API: Optional[CloudWorkerAPI] = None
 
 log.info(f'CLOUD {CLOUD}')
 log.info(f'CORES {CORES}')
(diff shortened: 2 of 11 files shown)


jigold:
Can you explain the changes you made for borrowing the compute client for creating the attached GCPDisks? I thought we did that for a reason. I can try and go back through the git history to find it.

batch/batch/cloud/gcp/worker/disk.py
```diff
@@ -11,22 +11,29 @@
...
-    def __init__(self, name: str, zone: str, project: str, instance_name: str, size_in_gb: int, mount_path: str):
+    def __init__(
+        self,
+        name: str,
+        zone: str,
+        project: str,
+        instance_name: str,
+        size_in_gb: int,
+        mount_path: str,
+        compute_client: aiogoogle.GoogleComputeClient,  # BORROWED
```
jigold, line 22: Can you explain the changes you made for borrowing the compute client for creating the attached GCPDisks? I thought we did that for a reason. I can try and go back through the git history to find it.

jigold (dismissed):

batch/batch/cloud/gcp/worker/disk.py
daniel-goldstein: I wasn't aware that there was a specific reason to create a new `ComputeClient` per disk. The reason I made this change was so that we wouldn't have O(disks) GoogleSessions each with their own access tokens. If you can find out the reason why I'll undo these change and add a comment.

jigold: I'm still looking, but the very original implementation had a borrowed compute_client:
https://github.com/hail-is/hail/pull/10090/files#diff-f8ba97f763395908a5b67f47a630c98e8d223ca5914f18d588f405d629d52197R958

jigold: Found it! https://github.com/hail-is/hail/pull/10811

daniel-goldstein: Oh look it's me!

daniel-goldstein: Huh, I can't quite understand what I was getting at in that PR. Do you see any reason for prematurely closing the session credentials that that PR would fix?

daniel-goldstein: I think this was happening because we at the time (and maybe until this PR), had a lot of job cleanup that wasn't getting waited on by the task manager during worker shutdown. So the worker object closed the compute client that was still being used by Disks. I think the proper fix here is to make sure that the Worker properly waits for usages of the compute client to finish up before deleting it, instead of just using more clients.

gear/gear/clients.py
```diff
@@ -23,21 +23,6 @@ def get_identity_client(credentials_file: Optional[str] = None):
     return aiogoogle.GoogleIAmClient(project, credentials_file=credentials_file)
 
 
-def get_compute_client(credentials_file: Optional[str] = None):
```
jigold, line 26: What is your rationale for refactoring this code? I thought it would be good to have all of the cloud abstractions for getting a compute client in the same place as the storage client. Otherwise, I thought it would be prone to error to have this same code called in multiple places.

batch/batch/worker/worker.py
```diff
@@ -2104,15 +2104,6 @@ def __str__(self):
         )
 
 
-@contextmanager
```
jigold, line 2107: Was this not used at all?

batch/batch/worker/worker.py
```diff
@@ -2915,16 +2915,22 @@ async def update():
...
 
     image_lock = aiorwlock.RWLock()
     docker = aiodocker.Docker()
 
+    CLOUD_WORKER_API = await GCPWorkerAPI.from_env() if CLOUD == 'gcp' else AzureWorkerAPI.from_env()
+    instance_config = CLOUD_WORKER_API.instance_config_from_config_dict(INSTANCE_CONFIG)
+    assert instance_config.cores == CORES
+    assert instance_config.cloud == CLOUD
+
+    task_manager = aiotools.BackgroundTaskManager()
```
jigold, line 2928: I don't see where the task manager is closed on shutdown.

batch/batch/cloud/gcp/worker/disk.py
jigold: I'm pretty sure you are correct that this change will be fine, but any changes to the Disk handling code gives me anxiety because of the potential financial costs if we mess up and overlook something. Can we make sure the disk monitoring code with an alert is up to date in Grafana? At one point, I had a chart where we were alerted if there was more than X amount of storage in disks being used or X number of disks. I think it's $0.17 / GB / month, so we probably want to be alerted if we are spending more than $20 per hour on disk costs which is 87 TB (please double check the math). So we probably can do an alert on 25 TB or 1k disks as that will catch slower accumulations over time. What do you think?

jigold (changes requested):

gear/gear/clients.py
daniel-goldstein: > I thought it would be good to have all of the cloud abstractions for getting a compute client in the same place as the storage client.

I think I personally don't see the benefit here. I'm not sure what you mean by "this same code", or why the storage and compute client getters need to be colocated. If there's some similarity between the two, I'd rather codify it in the types of the code rather than the position of the code. `get_compute_client` is only called in the worker process, so I don't see a reason at the moment for this code to be at a higher level of visibility than that.

However, my main motivations for moving this are the following two reasons:

1. I personally find this pattern of repeatedly checking `get_global_config()['cloud']` very noisy. It increases the surface area that this environment variable reaches and in the case of the worker it is mixing two design choices:
-  Having an object-oriented WorkerAPI class that knows cloud-specific implementation details
- A more function-based approach like this one that switches on global state to produce different implementations

I'd rather have only one of these approaches, and I like the former in the case of the worker because `worker.py` is already very object-oriented and because we only have to check the cloud once. When I trace the flow of the program from the worker main method through to some small detail, it is very noisy to me to be switching between cloud-specific and cloud-agnostic methods in the same callstack.

2. I wanted to share the cloud credentials object between the FS and the compute client. I think it would be awkward to introduce a `CloudSession` input to this method because it would involve an upcast followed by a downcast, which worsens my issue with multiple spots on the same codepath trying to figure out what cloud we're in. So at that point, I didn't see much value in having the function and stuck the appropriate parts of this code closer to where the corresponding inputs are.

batch/batch/cloud/gcp/worker/disk.py
daniel-goldstein: I'm all for making us feel confident in the change, so I'll do whatever metric we need to put in. Could you elaborate though on what could go wrong? I'd rather have an alert mean that something is wrong in our system and not that our scale is large.

batch/batch/worker/worker.py
daniel-goldstein: Not that I could find by grepping around.

gear/gear/clients.py
jigold: ok if this makes more sense to you. I think if we have multiple applications (not just batch) in the future that need a compute client, then we can revisit what the right abstraction is.

batch/batch/worker/worker.py
daniel-goldstein: Ah, it used to be owned by the worker so it's shutdown in `Worker.shutdown`, but now that I moved it out I should have the worker borrow it instead and shut it down in the main method.

batch/batch/cloud/gcp/worker/disk.py
jigold: What went wrong before is we were not cleaning up disks and they were slowly accumulating until we were paying for exabytes of disks -- I think we had over 100k 10Gi disks and didn't realize it. I'm worried that if the compute client suddenly doesn't work because the client session is closed (or some unknown issue with this change), then we're going to slowly accumulate orphaned disks. There's a garbage collection loop on the driver that also checks for orphaned disks and deletes them so there is redundancy here. Maybe we put an alert on how many orphaned disks we are cleaning up?

daniel-goldstein: > Maybe we put an alert on how many orphaned disks we are cleaning up?

This sounds great, I'll set up this metric.

batch/batch/worker/worker.py
daniel-goldstein: Ok so the reason I had initially moved the task_manager out of the Worker was so that I could give it to the `NetworkAllocator` as well. However, that created a mess because the TaskManager shutdown was hard to separate properly from the Worker shutdown because the order is so important. So I undid my change of moving the TaskManager out of the Worker and instead just created a separate task manager for the NetworkAllocator.

I'm going to follow this up with another PR that sorts out the whole shutdown situation and properly use 1 task manager with context managers, but that got pretty big to also be in this PR.

jigold (approved):

PR #12612, opened 2023-01-20
[batch] Restore printing of child batch URLs when `wait=False`

lgruen:
This is useful to find nested / child batches until the UI will group related batches together.

The previous print statement got removed in https://github.com/hail-is/hail/pull/12346/files#diff-40ac9d46c265cba91379598005c18322860f14a430de7cab77182d2083683be1.

#assign services

### DIFF
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -763,14 +763,15 @@ class ServiceBackend(Backend[bc.Batch]):
 
         jobs_to_command = {j.id: cmd for j, cmd in jobs_to_command.items()}
 
+        deploy_config = get_deploy_config()
+        url = deploy_config.external_url('batch', f'/batches/{batch_handle.id}')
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/batch/backend.py
```diff
@@ -763,15 +763,16 @@ async def compile_job(job):
 
         jobs_to_command = {j.id: cmd for j, cmd in jobs_to_command.items()}
 
+        deploy_config = get_deploy_config()
+        url = deploy_config.external_url('batch', f'/batches/{batch_handle.id}')
+        print(f'Submitted batch {batch_handle.id}, see {url}')
```
danking, line 768: Let's hide this behind an `if not wait` because the progress bar will include the URL when we're waiting

danking (changes requested):

danking:
Closing abandoned PRs. If someone wants to adopt please start a new PR.

PR #12617, opened 2023-01-24
[batch] Usability fixes to resource usage plots in the UI

jigold:
Based on @bw2 suggestions to make it clear where the limits are and more interpretable storage plots:

Screen Shot 2023-01-24 at 12 36 16 PM

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -73,7 +73,7 @@ from ..exceptions import (
     NonExistentBillingProjectError,
 )
 from ..file_store import FileStore
-from ..globals import BATCH_FORMAT_VERSION, HTTP_CLIENT_MAX_SIZE, complete_states
+from ..globals import BATCH_FORMAT_VERSION, HTTP_CLIENT_MAX_SIZE, RESERVED_STORAGE_GB_PER_CORE, complete_states
 from ..inst_coll_config import InstanceCollectionConfigs
 from ..resource_usage import ResourceUsageMonitor
 from ..spec_writer import SpecWriter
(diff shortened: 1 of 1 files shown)


batch/batch/front_end/front_end.py
```diff
@@ -1910,30 +1910,62 @@ def plot_job_durations(container_statuses: dict, batch_id: int, job_id: int):
...
+    memory_limit_bytes: Optional[int],
+    io_storage_limit_bytes: Optional[int],
+    non_io_storage_limit_bytes: Optional[int],
+) -> Optional[str]:
     if resource_usage is None:
         return None
 
+    if io_storage_limit_bytes is not None:
+        if io_storage_limit_bytes == 0:
+            io_storage_title = ''
```
daniel-goldstein, line 1924: In what situation would it be zero? I feel like we should always have some title.

daniel-goldstein (dismissed):

batch/batch/front_end/front_end.py
jigold: This is a placeholder for not having the 6th plot (it's empty) when someone requests `j.storage('0Gi')`. I decided in that case, it was better just to have one plot ("Storage (Container Overlay)") rather than trying to display a graph with 0 storage where `/io` is not really a "drive". Maybe I'm over thinking this.

jigold: See the Zulip thread for an example.

jigold: Screen-Shot-2023-01-20-at-2 41 07-PM

daniel-goldstein: Ah ok, this makes sense seeing the final result in the UI.

daniel-goldstein (approved):

PR #12622, opened 2023-01-26
[batch] Add prometheus metric for deletion of orphaned disks

daniel-goldstein:
I left the exception log statements in because it's not like these are driving up our logging costs, but let me know whether you want to keep them or not.

Can make a Grafana alert after this goes in based on this metric.


jigold (approved):
perfect! Thanks!

jigold:
Let's keep the log statements as it's helpful for debugging order of events and context from the logs.

PR #12624, opened 2023-01-26
[compiler] refactor BlockMatrix lowering to generate small IR

patrick-schultz:


tpoterba:
sometimes it takes a big change to make a little IR

hail/src/main/scala/is/hail/expr/ir/lowering/LowerBlockMatrixIR.scala
```diff
@@ -184,30 +186,695 @@ abstract class BlockMatrixStage(val letBindings: IndexedSeq[(String, IR)], val b
...
+              FastIndexedSeq(GetField(oldContext, "nRows"), GetField(oldContext, "nCols")),
+              False()),
+            blockIR(oldContext))
         }
+      }
+      BlockMatrixStage2(letBindings, broadcastVals, typ, newContexts, newBlock)
+  }
+
+  def withSparsity(rowPos: IR, rowIdx: IR, newType: BlockMatrixType, isSubset: Boolean = false): BlockMatrixStage2 = {
+    val (newLetBindings, rowPosRef, rowIdxRef) = {
```
tpoterba, line 661: can we add a bailout `return this` if the newType is the same as the old type? In practice we generate a lot of map2s where the old sparsity is the new, I think, and this can keep IR size down.

tpoterba (changes requested):
One small comment, structure looks great and the complexity I see is the complexity I expected. Nice work! I didn't read through everything closely (sparsity manipulations for instance) but that's what tests are for.

tpoterba (approved):

PR #12625, opened 2023-01-26
[fs] Reduce the memory footprint of the LocalAsyncFS

jigold:
I'm not 100% sure this will fix Ben's out of memory error when copying input files. It's possible there's a memory leak somewhere else. My guess on why this change could solve the issue is that because we open the files in "r+b" mode when writing to a local file in parallel, the kernel was caching the data as much as possible based on the available amount of memory to the container because it was anticipating sequential reads. I'd think the kernel is smart enough to evict data when needed, but maybe when using an attached disk with a network drive, the contention is high if the disk starts to lag and it couldn't write all of the data in the write buffer quickly enough and evict unneeded data to meet demand??? At the very least, hopefully this change will allow us to see if there is a true memory leak beyond the kernel using all of the available memory for caching potential reads.

Screen Shot 2023-01-26 at 3 52 38 PM

### DIFF
--- a/hail/python/hailtop/aiotools/fs/stream.py
+++ b/hail/python/hailtop/aiotools/fs/stream.py
@@ -151,6 +151,7 @@ class _WritableStreamFromBlocking(WritableStream):
         super().__init__()
         self._thread_pool = thread_pool
         self._f = f
+        self._start = f.tell()
 
     def writable(self) -> bool:
         return self._f.writable()
@@ -161,6 +162,7 @@ class _WritableStreamFromBlocking(WritableStream):
     async def _wait_closed(self) -> None:
         await blocking_to_async(self._thread_pool, self._f.flush)
         await blocking_to_async(self._thread_pool, os.fsync, self._f.fileno())
+        os.posix_fadvise(self._f.fileno(), self._start, self._f.tell(), os.POSIX_FADV_DONTNEED)
         await blocking_to_async(self._thread_pool, self._f.close)
         del self._f
 


danking:
What cgroups are you reading to get the memory graph? Kernel memory and user memory ought to be separate quantities with separate limits. If the graph is user memory, it seems like the issue is not the kernel but Python flushing buffers insufficiently often.

danking:
Hmm, I see this PR doesn't use flush

jigold:
I'm reading the memory tracked by the cgroup. I'd have to see if I can get more fine grained info on kernel vs user. The problem is not in Python flushing the buffer insufficiently often -- I tried flushing / fsync after every write and it didn't change the memory usage at all.

danking (approved):
Advising that we don't need the written data seems totally fine to me.

danking:
I am likewise pretty surprised this flag makes a difference to the kernel memory usage in this case.

hail/python/hailtop/aiotools/fs/stream.py
```diff
@@ -161,6 +162,7 @@ async def write(self, b: bytes) -> int:
     async def _wait_closed(self) -> None:
         await blocking_to_async(self._thread_pool, self._f.flush)
         await blocking_to_async(self._thread_pool, os.fsync, self._f.fileno())
+        os.posix_fadvise(self._f.fileno(), self._start, self._f.tell(), os.POSIX_FADV_DONTNEED)
         await blocking_to_async(self._thread_pool, self._f.close)
```
jigold, line 166: @danking Did this need to be async or does it not matter?

PR #12627, opened 2023-01-26
[fs] fix fadvise args

jigold:
I messed up that the third argument is length and not end position. I think the "+1" is needed, but not 100% sure.


danking (changes requested):
`tell` is the current position. If we have written nothing, `start` must equal `tell` and `tell - start + 1` must equal `1`, that seems wrong to me.

If we wrote one byte, `tell` is definitionally `start + 1`, and `tell - start` is `1`.

danking (approved):

PR #12629, opened 2023-01-27
[query] IBD implemented in terms of block matrices

jigold:
Patrick -- can you give this a careful look over? Once we're happy with it, I'll redo the benchmarking as a last sanity check.

### DIFF
--- a/hail/python/hail/methods/relatedness/identity_by_descent.py
+++ b/hail/python/hail/methods/relatedness/identity_by_descent.py
@@ -189,8 +189,7 @@ def identity_by_descent(dataset, maf=None, bounded=True, min=None, max=None) -> 
 
     def convert_to_table(bm, annotation_name):
         t = bm.entries()
-        t = t.annotate(**{annotation_name: t.entry})
-        t = t.drop('entry')
+        t = t.rename({'entry': annotation_name})
         return t
 
     z0 = convert_to_table(Z0, 'Z0')
--- a/hail/python/hail/methods/relatedness/identity_by_descent.py
+++ b/hail/python/hail/methods/relatedness/identity_by_descent.py
@@ -89,6 +89,11 @@ def identity_by_descent(dataset, maf=None, bounded=True, min=None, max=None) -> 
     :class:`.Table`
     """
 
+    require_col_key_str(dataset, 'identity_by_descent')
+
+    if not isinstance(dataset.GT, hl.CallExpression):
+        raise Exception('GT field must be of type Call')
+
     if maf is not None:
         analyze('identity_by_descent/maf', maf, dataset._row_indices)
         dataset = dataset.select_rows(__maf=maf)
@@ -108,18 +113,10 @@ def identity_by_descent(dataset, maf=None, bounded=True, min=None, max=None) -> 
             'max': max,
         })).persist()
 
-    require_col_key_str(dataset, 'identity_by_descent')
-
     min = min or 0
     max = max or 1
     if not 0 <= min <= max <= 1:
         raise Exception(f"invalid pi hat filters {min} {max}")
-
-    dataset = dataset.select_cols().select_globals().select_entries('GT')
-    dataset = require_biallelic(dataset, 'ibd')
-
-    if not isinstance(dataset.GT, hl.CallExpression):
-        raise Exception('GT field must be of type Call')
 
     sample_ids = dataset.s.collect()
     if len(sample_ids) != len(set(sample_ids)):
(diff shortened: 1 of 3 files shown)


hail/python/hail/methods/relatedness/identity_by_descent.py
```diff
@@ -85,20 +89,144 @@ def identity_by_descent(dataset, maf=None, bounded=True, min=None, max=None) ->
...
+        e20=hl.agg.sum(dataset._e20),
+        e11=hl.agg.sum(dataset._e11),
+        e21=hl.agg.sum(dataset._e21),
+        e22=hl.agg.sum(dataset._e22)
+    ))
+
+    IS_HOM_REF = BlockMatrix.from_entry_expr(dataset.is_hom_ref).checkpoint(hl.utils.new_temp_file())
+    IS_HET = BlockMatrix.from_entry_expr(dataset.is_het).checkpoint(hl.utils.new_temp_file())
+    IS_HOM_VAR = BlockMatrix.from_entry_expr(dataset.is_hom_var).checkpoint(hl.utils.new_temp_file())
+    NOT_MISSING = BlockMatrix.from_entry_expr(dataset.is_not_missing).checkpoint(hl.utils.new_temp_file())
```
patrick-schultz, line 176: Given the expense of `from_entry_expr`, I think it's worth doing the minimum necessary. These four are related by `not_missing = is_hom_ref + is_het + is_hom_var`, so any one can be computed from the other three.

patrick-schultz (changes requested):
This is very nice and clean. I only see a few little things to clean up.

jigold:
@patrick-schultz Are you okay with where the checkpoints are? If yes, then I'll do one last set of benchmarks and then this is ready to go!

hail/python/hail/methods/relatedness/identity_by_descent.py
```diff
@@ -87,18 +91,138 @@ def identity_by_descent(dataset, maf=None, bounded=True, min=None, max=None) ->
...
+            'min': min,
+            'max': max,
+        })).persist()
+
+    min = min or 0
+    max = max or 1
+    if not 0 <= min <= max <= 1:
+        raise Exception(f"invalid pi hat filters {min} {max}")
+
+    sample_ids = dataset.s.collect()
```
patrick-schultz, line 121: Since we're making multiple passes over `dataset`, we probably want to checkpoint it before this first pass.

hail/python/hail/methods/relatedness/identity_by_descent.py
```diff
@@ -87,18 +91,138 @@ def identity_by_descent(dataset, maf=None, bounded=True, min=None, max=None) ->
...
+        e21=hl.agg.sum(dataset._e21),
+        e22=hl.agg.sum(dataset._e22)
+    ))
+
+    IS_HOM_REF = BlockMatrix.from_entry_expr(dataset.is_hom_ref).checkpoint(hl.utils.new_temp_file())
+    IS_HET = BlockMatrix.from_entry_expr(dataset.is_het).checkpoint(hl.utils.new_temp_file())
+    IS_HOM_VAR = BlockMatrix.from_entry_expr(dataset.is_hom_var).checkpoint(hl.utils.new_temp_file())
+    NOT_MISSING = IS_HOM_REF + IS_HET + IS_HOM_VAR
+
+    total_possible_ibs = NOT_MISSING.T @ NOT_MISSING
```
patrick-schultz, line 175: Not sure if checkpointing NOT_MISSING before the matmul if worth the cost of the write. Might try benchmarking both ways.

hail/python/hail/methods/relatedness/identity_by_descent.py
```diff
@@ -87,18 +91,138 @@ def identity_by_descent(dataset, maf=None, bounded=True, min=None, max=None) ->
...
+
+    ibs0_pre = (IS_HOM_REF.T @ IS_HOM_VAR).checkpoint(hl.utils.new_temp_file())
+    ibs0 = ibs0_pre + ibs0_pre.T
+
+    is_not_het = IS_HOM_REF + IS_HOM_VAR
+    ibs1_pre = (IS_HET.T @ is_not_het).checkpoint(hl.utils.new_temp_file())
+    ibs1 = ibs1_pre + ibs1_pre.T
+
+    ibs2 = total_possible_ibs - ibs0 - ibs1
```
patrick-schultz, line 184: I *think* you want to checkpoint ibs0, ibs1, ibs2. But again, it's not clear to me if the io or the repeated computation is worse.

jigold:
```
# (n_samples, n_variants)

# Spark backend (1 core)
# (2500, 100000) =>   147.05s
# (5000, 10000) =>    113.28s

# Local backend
# (2500, 100000) =>   233.94s
# (5000, 10000) =>    489.46s
```

PR #12632, opened 2023-01-27
[query] coerce 1d ndarray into an array

danking:


danking:
fyi @tpoterba I want our type coercers to allow 1-d ndarrays when I ask for an array.

tpoterba (approved):

PR #12636, opened 2023-01-30
[query] better error messages for hstack and vstack

danking:
CHANGELOG: `hl.nd.vstack` and `hl.nd.hstack` provide clear error messages when used with incompatible matrices or with no arguments.


tpoterba:
let me noodle on this for the day and possibly propose an upstream solution

danking:
closing while you think

PR #12637, opened 2023-01-30
[query] Lower linear SKAT

danking:
CHANGELOG: Query-on-Batch now supports `hl.skat(..., logistic=False)`.

I also added actual tests for `hl.skat`, which were lost at some point.

I am somewhat not confident in my documentation and comments, because the SKAT paper is terse and unclear. I would really apprecaiate strong criticism of the documentation and the code comments.


danking:
These are the rendered docs. They are not included in the actual docs because I do not want people using this function yet.


Screen Shot 2023-01-30 at 11 31 24
Screen Shot 2023-01-30 at 11 31 29
Screen Shot 2023-01-30 at 11 31 33
Screen Shot 2023-01-30 at 11 31 39

hail/python/hail/methods/statgen.py
```diff
@@ -1509,6 +1509,456 @@ def linear_mixed_regression_rows(entry_expr,
...
+        G_take = hl.agg.take(ht.G_row, n=max_size + 1),
+        size = hl.agg.count()
+    )
+    ht = ht.annotate(
+        weight = hl.nd.array(hl.or_missing(hl.len(ht.weight_take) <= max_size, ht.weight_take)),
+        G = hl.nd.array(hl.or_missing(hl.len(ht.G_take) <= max_size, ht.G_take)).T
+    )
+    ht = ht.annotate(
+        Q = (((ht.y_residual @ ht.G) * ht.weight) @ ht.G.T) @ ht.y_residual.T
+    )
```
patrick-schultz, line 1841: More efficient:
```
((ht.y_residual @ ht.G).map(lambda x: x**2) * ht.weight).sum(0)
```

hail/python/hail/methods/statgen.py
```diff
@@ -1509,6 +1509,456 @@ def linear_mixed_regression_rows(entry_expr,
...
+    # Back to Z:
+    #
+    #     Z.T Z = s^2 (B0.T B0 - C0.T C0)
+    #
+    # Which we want the eigenvalues of. Hail lacks an eigenvalues method, but luckily the singular
+    # values of a symmetric semi-positive-definite matrix *are* the eigenvalues.
+    #
+    # We avoid the square root in order to avoid complex numbers.
+
+    Q, _ = hl.nd.qr(ht.covmat)
```
patrick-schultz, line 1911: Already computed this

hail/python/hail/methods/statgen.py
```diff
@@ -1509,6 +1509,456 @@ def linear_mixed_regression_rows(entry_expr,
...
+    C0 = Q.T @ ht.G
+    singular_values = hl.nd.svd(ht.G.T @ (ht.G * ht.weight) - C0.T @ (C0 * ht.weight), compute_uv=False)
```
patrick-schultz, line 1913: As discussed in zulip, I think a better way to compute this is:
```
R = nd.qr(ht.G - Q @ (Q.T @ G), mode="r")
singular_values = hl.nd.svd(R, compute_uv=False)
eigenvalues = singular_values.map(lambda x: x**2)
```
(replace `singular_values` with `eigenvalues` below)

patrick-schultz (changes requested):
Some notes after a first pass

danking:
Comments addressed. Still gotta work on the verbiage to clearly explain the present of P_0.

danking:
@patrick-schultz Alright. This passes tests locally. I'm happy with the docs verbiage. I am eager for your review!

Screen Shot 2023-02-01 at 17 54 00

Screen Shot 2023-02-01 at 17 54 10

patrick-schultz (dismissed):
A few docs comments. Everything else looks great!

hail/python/hail/methods/statgen.py
```diff
@@ -1509,6 +1509,483 @@ def linear_mixed_regression_rows(entry_expr,
...
+    # at most a sign.
+
+    weights_arr = ht.weight._data_array()
+    A = hl.case().when(
+        hl.all(weights_arr.map(lambda x: x >= 0)),
+        (ht.G - ht.covmat_Q @ (ht.covmat_Q.T @ ht.G)) * hl.sqrt(ht.weight)
+    ).or_error(hl.format('hl._linear_skat: every weight must be positive, in group %s, the weights were: %s',
+                         ht.group, weights_arr))
+    singular_values = hl.nd.svd(A, compute_uv=False)
+    # SVD(M) = U S V. U and V are unitary, therefore SVD(k M) = U (k S) V.
```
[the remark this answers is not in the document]
danking: In the next line, we scale the singular values by `\sigma^2` instead of multiplying A by `sqrt(\sigma^2)`.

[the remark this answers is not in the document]
danking: I added a blank line to clarify to which expression the comment applies.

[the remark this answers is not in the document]
patrick-schultz: Oh I see. That doesn't have anything to do with unitarity. Scalars commute with all matrices.

hail/python/hail/methods/statgen.py
```diff
@@ -1509,6 +1509,480 @@ def linear_mixed_regression_rows(entry_expr,
...
+        \end{align*}
+
+    This expression is a `"quadratic form" <https://en.wikipedia.org/wiki/Quadratic_form>`__ of the
+    vector :math:`h`. Because :math:`B` is a real symmetric matrix, we can eigendecompose it into an
+    orthogonal matrix and a diagonal matrix of eigenvalues:
+
+    .. math::
+
+        \begin{align*}
+        U \Lambda U &= B \quad\quad \Lambda \textrm{ diagonal } U \textrm{ orthogonal} \\
```
patrick-schultz, line 1606: Still missing a transpose: `U \Lambda U.T`

patrick-schultz (changes requested):

hail/python/hail/methods/statgen.py
danking: ah oops

patrick-schultz (approved):

PR #12642, opened 2023-02-01
[auth] Make HailCredentials for refreshable auth tokens

daniel-goldstein:
My motivation here is that in Terra we aren't going to be using hail authentication tokens, rather an authentication from gcloud or az. So I want our batch client to be able to use a `CloudCredentials` just as easily as it uses the hail auth token that we store on the user's machine. So I introduce HailCredentials which subclasses CloudCredentials but the behavior is the same. It also lets us re-use some of the retry logic that we have baked into `aiocloud.common.Session`.

### DIFF
--- a/batch/test/test_batch.py
+++ b/batch/test/test_batch.py
@@ -7,8 +7,8 @@ from typing import Set
 import pytest
 
 from hailtop import httpx
-from hailtop.auth import service_auth_headers
 from hailtop.batch.backend import HAIL_GENETICS_HAILTOP_IMAGE
+from hailtop.auth import hail_credentials
 from hailtop.batch_client.client import BatchClient
 from hailtop.config import get_deploy_config, get_user_config
 from hailtop.test_utils import skip_in_azure
@@ -690,7 +690,7 @@ def test_create_idempotence(client: BatchClient):
     assert b1.id == b2.id
 
 
-def test_batch_create_validation():
+async def test_batch_create_validation():
     bad_configs = [
         # unexpected field fleep
         {'billing_project': 'foo', 'n_jobs': 5, 'token': 'baz', 'fleep': 'quam'},
@@ -723,7 +723,7 @@ def test_batch_create_validation():
         {'attributes': {'k': None}, 'billing_project': 'foo', 'n_jobs': 5, 'token': 'baz'},
     ]
     url = deploy_config.url('batch', '/api/v1alpha/batches/create')
-    headers = service_auth_headers(deploy_config, 'batch')
+    headers = await hail_credentials().auth_headers()
     session = external_requests_client_session()
     for config in bad_configs:
         r = retry_response_returning_functions(session.post, url, json=config, allow_redirects=True, headers=headers)
--- a/batch/test/test_batch.py
+++ b/batch/test/test_batch.py
@@ -7,8 +7,8 @@ from typing import Set
 import pytest
 
 from hailtop import httpx
-from hailtop.batch.backend import HAIL_GENETICS_HAILTOP_IMAGE
 from hailtop.auth import hail_credentials
+from hailtop.batch.backend import HAIL_GENETICS_HAILTOP_IMAGE
 from hailtop.batch_client.client import BatchClient
 from hailtop.config import get_deploy_config, get_user_config
 from hailtop.test_utils import skip_in_azure
--- a/hail/python/hailtop/batch_client/aioclient.py
+++ b/hail/python/hailtop/batch_client/aioclient.py
@@ -10,8 +10,10 @@ import orjson
 import secrets
 
 from hailtop.config import get_deploy_config, DeployConfig
-from hailtop.auth import service_auth_headers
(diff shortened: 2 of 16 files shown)


daniel-goldstein:
Putting a WIP on here because I need to double check with the AUS folks if it's ok for me to remove the `token` input to the BatchClient. I'd really prefer not opening up our code to possibly putting tokens in there, but I need to understand their use-case better.

hail/python/hailtop/hailctl/dev/deploy/cli.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, deploy_config=None):
         self._session: Optional[httpx.ClientSession] = None
 
     async def __aenter__(self):
-        headers = service_auth_headers(self._deploy_config, 'ci')
+        headers = await hail_credentials().auth_headers()
```
jigold, line 37: What happened to `ci` here?

hail/python/hailtop/hailctl/auth/logout.py
```diff
@@ -18,7 +18,7 @@ async def async_main():
         print('Not logged in.')
         return
 
-    headers = service_auth_headers(deploy_config, 'auth')
+    headers = await hail_credentials().auth_headers()
```
jigold, line 21: where did `auth` go here?

hail/python/hailtop/config/deploy_config.py
```diff
@@ -128,28 +126,6 @@ async def get_metrics(request):  # pylint: disable=unused-argument,unused-variab
         log.info(f'serving paths at {base_path}')
         return root_app
 
-    async def addresses(self, service: str) -> List[Tuple[str, int]]:
```
jigold, line 131: Was this dead code?

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -828,49 +841,40 @@ async def create(billing_project: str,
...
-        self._session: Optional[httpx.ClientSession] = session
+        self._session: Session = session
         self._headers = headers
 
-    async def _get(self, path, params=None) -> aiohttp.client_reqrep.ClientResponse:
-        assert self._session
-        return await request_retry_transient_errors(
-            self._session, 'GET', self.url + path, params=params, headers=self._headers
-        )
+    async def _get(self, path, params=None) -> aiohttp.ClientResponse:
```
jigold, line 867: The return type here looks right from the code in main, but I thought we would have returned httpx.ClientResponse.

batch/test/test_batch.py
```diff
@@ -674,7 +674,7 @@ def test_create_idempotence(client: BatchClient):
     assert b1.id == b2.id
 
 
-def test_batch_create_validation():
+async def test_batch_create_validation():
```
jigold, line 677: Can you make this async? Not sure if we have async tests enabled here.

jigold (dismissed):
This looks great! Did you happen to test some of the hailctl functionality? I imagine we have enough hailctl uses in build.yaml that if there's a major problem, it will have been caught.

hail/python/hailtop/hailctl/dev/deploy/cli.py
daniel-goldstein: The service argument was basically unused. As long as I've been here all services lived in the same namespace for a given environment but maybe this is an artifact of some time before where that wasn't the case?

hail/python/hailtop/hailctl/auth/logout.py
daniel-goldstein: See `ci` answer above.

hail/python/hailtop/config/deploy_config.py
daniel-goldstein: Yes, it makes a request to the `address` service which no longer exists.

batch/test/test_batch.py
daniel-goldstein: I think it's on by default. It ran and passed [here](https://ci.hail.is/batches/7002970/jobs/116)

hail/python/hailtop/batch_client/aioclient.py
daniel-goldstein: hm ya I'll fix these types, good call

daniel-goldstein: Actually, weird, I can't find a single usage of `httpx.ClientResponse` in the codebase. Is that dead code? Either way it looks like this is the correct type from what I can see in `httpx.ClientSession._request`

jigold (changes requested):
This looks good. Error in Azure looks transient. Can you do one last sanity check for me which is to make sure `hailctl auth login` works locally for you or is covered by a test somewhere? I couldn't find one anywhere.

daniel-goldstein:
@jigold bump

jigold:
Sorry. Didn't see the dismissed review.

jigold (approved):

PR #12643, opened 2023-02-02
[query] lower logistic SKAT

danking:
CHANGELOG: `hl.skat(..., logistic=True)` now supported in the Batch backend.

### DIFF
--- a/hail/python/hail/methods/statgen.py
+++ b/hail/python/hail/methods/statgen.py
@@ -2030,8 +2030,7 @@ def _logistic_skat(group,
         X &: R^{N \times K} \\
         G &: \{0, 1, 2\}^{N \times M} \\
         \\
-        \varepsilon &\sim N(0, \sigma^2) \\
-        \textrm{logit}(P(y=1)) &= \beta_0 X + \beta_1 G + \varepsilon
+        \textrm{logit}(P(y=1)) &= \beta_0 X + \beta_1 G + \varepsilon \quad\quad \varepsilon \sim N(0, \sigma^2)
         \end{align*}
 
     The usual null hypothesis is :math:`\beta_1 = 0`. SKAT tests for an association, but does not
@@ -2040,7 +2039,14 @@ def _logistic_skat(group,
     Wu et al. argue that, under the null hypothesis, a particular value, :math:`Q`, is distributed
     according to a generalized chi-squared distribution with parameters determined by the genotypes,
     weights, and residual phenotypes. The SKAT p-value is the probability of drawing even larger
-    values of :math:`Q`. :math:`Q` is defined by Wu et al. as:
+    values of :math:`Q`. If :math:`\widehat{\beta_\textrm{null}}` is the best-fit beta under the
+    null model:
+
+    .. math::
+
+        \textrm{logit}(P(y=1)) = \beta_\textrm{null} X + \varepsilon \quad\quad \varepsilon \sim N(0, \sigma^2)
+
+    Then :math:`Q` is defined by Wu et al. as:
 
     .. math::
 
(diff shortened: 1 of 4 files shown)


hail/python/test/hail/methods/test_skat.py
```diff
@@ -105,7 +186,124 @@ def test_linear_skat_no_weights_R_truth():
     assert result.fault == 0
 
 
-def test_linear_skat_R_truth():
```
danking, line 108: this name was actually duplicated in current main, it's now called `test_linear_skat_no_weights_R_truth`

hail/python/hail/methods/statgen.py
```diff
@@ -2140,8 +2567,19 @@ def skat(key_expr,
         Table of SKAT results.
 
     """
-    if hl.current_backend().requires_lowering and not logistic:
-        ht = hl._linear_skat(key_expr, weight_expr, y, x, covariates, max_size, accuracy, iterations)
+    if hl.current_backend().requires_lowering:
```
danking, line 2570: In the second phase, we'll benchmark and address speed problems. Until we're certain it is fast enough, I don't want to enable it for all backends.

hail/python/test/hail/methods/test_skat.py
```diff
@@ -6,95 +6,176 @@
...
+    skat_name, skat = skat_model
     genotypes = [
-        [0, 1, 0, 0, 0],
-        [1, 0, 0, 0, 0],
-        [0, 1, 2, 0, 2],
-        [1, 0, 0, 2, 1]]
+        [2, 1, 1, 1, 0, 1, 1, 2, 1, 1, 2, 1, 0, 0, 1],
+        [1, 0, 1, 1, 1, 2, 0, 2, 1, 1, 0, 1, 1, 0, 0],
+        [0, 2, 0, 0, 2, 1, 1, 2, 2, 1, 1, 1, 0, 1, 1],
+        [1, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 0]]
```
danking, line 17: I needed to create a matrix on which the logistic model converged.

hail/python/test/hail/methods/test_skat.py
danking, line 108: This test is inadvertently duplicated in main. I just removed it entirely, but the diff is a bit awkward.a

danking:
Hmm. I trust the code now. I test against several R SKAT runs. I'm not sure I understand how we derive that Q is generalized chi-squared distributed. We use the residual phenotypes in the calculation of Q, but those are inverse-logit transformed normal variables. The derivation for the linear case doesn't apply, as far as I can tell. I assume the residuals are Bernoulli distributed? Maybe not. I guess the phenotypes are Bernoulli but the errors aren't? I'm not sure.

danking:
bump @patrick-schultz ideally we would merge this before Wednesday's meeting!

hail/python/hail/methods/statgen.py
```diff
@@ -1983,6 +1998,428 @@ def _linear_skat(group,
...
+    Therefore :math:`r`, the residual phenotype, is the portion of the phenotype unexplained by the
+    covariates alone. Also notice:
+
+    1. The residual phenotypes are Bernoulli distributed with mean :math:`p` and variance
+       :math:`\sigma^2 = p(1 - p)` where :math:`p` is the best-fit probability.
+
+    2. :math:`G W G^T`, is a symmetric positive-definite matrix when the weights are non-negative.
+
+    We can transform the residuals into standard normal variables by normalizing by their
+    variance.
```
patrick-schultz, line 2069: I think everything below this doesn't make much sense. The residuals are certainly not normal -- they can only take two possible values. I tried to look into how to justify the logistic case, and made a little progress ("Pearson residuals" are relevant), but I think it's best to not try to write a complete derivation here. It should be enough to summarize the math from the paper that we implement.

patrick-schultz (changes requested):

patrick-schultz:
> bump @patrick-schultz ideally we would merge this before Wednesday's meeting!

My bad, didn't realize I hadn't submitted my review.

danking:
@patrick-schultz , I've reworked the docs in a less wrong way. Let me know what you think!

Screen Shot 2023-02-15 at 18 32 21
Screen Shot 2023-02-15 at 18 32 26
Screen Shot 2023-02-15 at 18 32 29

hail/python/hail/methods/statgen.py
```diff
@@ -1983,6 +1997,427 @@ def _linear_skat(group,
...
+
+    .. math::
+
+        \textrm{logit}(P(y=1)) = \beta_\textrm{null} X + \varepsilon \quad\quad \varepsilon \sim N(0, \sigma^2)
+
+    Therefore :math:`r`, the residual phenotype, is the portion of the phenotype unexplained by the
+    covariates alone. Also notice:
+
+    1. The residual phenotypes are Bernoulli distributed with mean :math:`p` and variance
+       :math:`\sigma^2 = p(1 - p)` where :math:`p` is the best-fit probability.
```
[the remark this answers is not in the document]
danking: done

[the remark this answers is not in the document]
danking: I changed this to have a sample index since the sigmas are sample indexed in the logistic case.

hail/python/hail/methods/statgen.py
```diff
@@ -1983,6 +1998,446 @@ def _linear_skat(group,
...
+    `x`. For :math:`N` samples, in a group of :math:`M` variants, with :math:`K` covariates, the
+    model is given by:
+
+    .. math::
+
+        \begin{align*}
+        X &: R^{N \times K} \\
+        G &: \{0, 1, 2\}^{N \times M} \\
+        \\
+        \textrm{logit}(P(y=1)) &= \beta_0 X + \beta_1 G + \varepsilon \quad\quad \varepsilon \sim N(0, \sigma^2)
```
patrick-schultz, line 2033: This is still saying the residuals are normally distributed. I think the right way to express the logistic regression model is
```
y ~ Bernoulli(logit^{-1}(beta_0 X + beta_1 G))
```

hail/python/hail/methods/statgen.py
```diff
@@ -1983,6 +1998,446 @@ def _linear_skat(group,
...
+
+    Wu et al. argue that, under the null hypothesis, a particular value, :math:`Q`, is distributed
+    according to a generalized chi-squared distribution with parameters determined by the genotypes,
+    weights, and residual phenotypes. The SKAT p-value is the probability of drawing even larger
+    values of :math:`Q`. If :math:`\widehat{\beta_\textrm{null}}` is the best-fit beta under the
+    null model:
+
+    .. math::
+
+        \textrm{logit}(P(y=1)) = \beta_\textrm{null} X + \varepsilon \quad\quad \varepsilon \sim N(0, \sigma^2)
```
patrick-schultz, line 2047: Here too

hail/python/hail/methods/statgen.py
```diff
@@ -1983,6 +1998,446 @@ def _linear_skat(group,
...
+        \begin{align*}
+        V_{ii} &= \sigma^2_i \\
+        W_{ii} &= w_i \quad\quad \textrm{the weight for variant } i \\
+        \\
+        P_0 &= V - V X (X^T V X)^{-1} X^T V \\
+        Z Z^T &= P_0^{1/2} G W G^T P_0^{1/2}
+        \end{align*}
+
+    Recall that the eigenvalues of a symmetric matrix and its transpose are the same, so we can
+    instead consider :math:`Z^T Z` (note that we elide transpositions of symmetric matrices):
```
patrick-schultz, line 2085: This isn't saying what you mean (`Z Z^T` is symmetric, so its transpose is `Z Z^T`). The only way I know to say this is that the eigenvalues of both `Z Z^T` and `Z^T Z` are the squared singular values of `Z`.

hail/python/hail/methods/statgen.py
```diff
@@ -1983,6 +1997,427 @@ def _linear_skat(group,
...
+    )
+    ht = ht.annotate(
+        # Q=ht.y_residual @ (ht.G * ht.weight) @ ht.G.T @ ht.y_residual.T
+        Q=((ht.y_residual @ ht.G).map(lambda x: x**2) * ht.weight).sum(0)
+    )
+
+    # See linear SKAT code comment for an extensive description of the mathematics here.
+
+    sqrtv = hl.sqrt(ht.s2)
+    Q, _ = hl.nd.qr((ht.covmat.T * sqrtv).T)
```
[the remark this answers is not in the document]
patrick-schultz: Oops, yes, that's what I meant.

The -1 thing is a numpy behavior we mimic. From [their docs](https://numpy.org/doc/stable/reference/generated/numpy.reshape.html):
> One shape dimension can be -1. In this case, the value is inferred from the length of the array and remaining dimensions.

So `reshape(-1, 1)` is a universal "make this a column vector" command.

patrick-schultz (changes requested):
Looks great, just a couple small fixes

danking:
Thanks for clarifying these points and pointing out the errors!

Screen Shot 2023-02-16 at 18 31 01
Screen Shot 2023-02-16 at 18 31 04
Screen Shot 2023-02-16 at 18 31 09

PR #12650, opened 2023-02-03
[query] Stop silently exporting null bgens if no GP field exists

tpoterba:
could use a second look here -- am I crazy? 

If GP is missing from the MT and no expression is passed, there's no GP in the MT passed into the writer.

And if there's no GP in the writer, we export a file worth of nulls.

https://github.com/hail-is/hail/blob/eee0924c119d8aa2ba1cdaf4f30a4f2be4ba347f/hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala#L1123-L1149

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
@@ -884,7 +884,9 @@ final case class GenVariantWriter(typ: MatrixType, entriesFieldName: String, pre
     def writeC(code: Code[Int]) = _writeC(cb, code)
     def writeS(code: Code[String]) = _writeS(cb, code)
 
-    val hasGPField = typ.entryType.hasField("GP")
+
+    require(typ.entryType.hasField("GP") && typ.entryType.fieldType("GP") == TArray(TFloat64))
+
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
```diff
@@ -1126,28 +1127,24 @@ case class BGENPartitionWriter(typ: MatrixType, entriesFieldName: String, writeH
     entries.forEachDefinedOrMissing(cb)({ (cb, j) =>
       emitNullGP(cb)
     }, { case (cb, j, entry: SBaseStructValue) =>
-      if (entry.st.virtualType.fieldIdx.get("GP").isDefined) {
```
danking, line 1129: Couldn't the GP field still be missing even though the entries aren't? (Meaning we need this if)

danking (changes requested):
Indeed that seems not great. I think we need to keep the if. I also noticed on line 243/245 that we have `export_gen` where there should be an `export/bgen`.

hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
tpoterba: will do the same for GEN though!

danking (approved):

PR #12651, opened 2023-02-03
[batch] use specialized query for monitor_billing_limits

danking:
This should lighten the database load a bit by avoiding a couple joins and a CTE.


daniel-goldstein (approved):

danking:
I don't have time to sort out exactly what the issue is. Let's focus on addressing the root cause of too many resources instead.

PR #12654, opened 2023-02-03
[batch] Mitigate too many resources with same prices

jigold:
I left out the cloud specific check and just make it so we don't have this problem again on any cloud.


danking (dismissed):
let's check for equality

jigold:
I'm putting the WIP tag for Monday. I changed the code slightly from before. I checked and the test PR driver logs were at least clean.

danking (approved):
LGTM

jigold:
I forgot there was a WIP on here. Will take it off tomorrow morning.

PR #12659, opened 2023-02-06
[query] Construct MatrixTable from Constituent Parts

ehigham:
Add `MatrixTable.from_parts`.


hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,136 @@ class MatrixTable(ExprContainer):
...
+            else [[{} for _ in cols] for _ in rows]
+
+        entries_field_name = 'entry_structs'
+        for i, (row, row_entries) in enumerate(zip(rows, entries)):
+            row['row_idx'] = i
+            row[entries_field_name] = [hl.struct(**kvs) for kvs in row_entries]
+
+        return Table\
+            .parallelize(rows, key='row_idx', globals=hl.struct(**globals)) \
+            ._unlocalize_entries(entries_field_name, cols_field_name, col_key=[])
```
danking, line 670: In the spirit of consistency across the codebase, these kind of expressions are always written like this:
```
ht = Table.parallelize(...)
ht = ht._unlocalize_entries(...)
return ht
```
We're inconsistent on whether we `return ht` or just `return ht_.unlocalize_entries(...)`.

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,136 @@ class MatrixTable(ExprContainer):
...
+
+        cols_field_name = 'columns'
+        globals[cols_field_name] = cols
+
+        # `MatrixTable._unlocalize_entries` extracts matrix entries from the
+        # rows on the table, so these need to be defined too.
+        rows = invert(rows) if rows else [{} for _ in anyval(entries)]
+        entries = map(invert, invert(entries)) \
+            if entries \
+            else [[{} for _ in cols] for _ in rows]
```
danking, line 661: In the spirit of consistency across the codebase, this expression is usually written as:
```
if entries:
    entries = map(invert, invert(entries))
else:
    entries = [[{} for _ in cols] for _ in rows]
```
`mypy` should infer the type just fine.

We generally avoid line continuation in Python.

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,136 @@ class MatrixTable(ExprContainer):
...
         return MatrixTable(ir.JavaMatrix(jmir))
 
+    @staticmethod
+    @typecheck(
+        globals=nullable(dictof(str, anytype)),
+        rows=nullable(dictof(str, sequenceof(anytype))),
+        cols=nullable(dictof(str, sequenceof(anytype))),
+        entries=nullable(dictof(str, sequenceof(sequenceof(anytype)))),
+    )
+    def from_parts(globals=None, rows=None, cols=None, entries=None) -> 'MatrixTable':
```
danking, line 549: I'm not sure if this is team-wide standard, but I prefer that we include mypy type annotations in new code. I would like us to eventually actually enforce that mypy succeeds on the `hail/python/hail` code. Currently we only require that of the "services" code (ci, batch, etc.).

hail/python/hail/table.py
```diff
@@ -537,9 +541,13 @@ def parallelize(cls, rows, schema=None, key=None, n_partitions=None, *, partial_
...
-        table = Table(ir.TableParallelize(ir.MakeStruct([
-            ('rows', rows._ir),
-            ('global', ir.MakeStruct([]))]), n_partitions))
+        table = Table(ir.TableParallelize(
+            ir.MakeStruct([
+                ('rows', rows._ir),
+                ('global', (globals or hl.struct())._ir)
+            ]),
+            n_partitions
+        ))
```
danking, line 550: I appreciate this improved formatting.

danking (dismissed):
The interface looks like exactly what I had in mind! I made a few comments as I read through the impl, mostly arguing in favor of stylistic consistency across the code-base.

hail/python/hail/matrixtable.py
ehigham: I'll go with the `return ht. ...` since ht is no longer a hail table after `_unlocalize_entries`

hail/python/hail/matrixtable.py
ehigham: I have a bunch of ternaries in this function so putting this would be weird without changing those too. If the issue is avoiding the line continuation, I'll capitulate on 80 columns.

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+        ----------------------------------------
+        Entry fields:
+            'baz': int32
+        ----------------------------------------
+        Column key: None
+        Row key: ['row_idx']
+        ----------------------------------------
+        >>> mt.row.show()
+        +-------+---------+
+        |   foo | row_idx |
```
daniel-goldstein, line 586: Don't have much to say on the implementation here, and I'll let Dan re-review, but I'm curious here why `row_idx` comes second? If this is coming from evaluating this, how come the order would be different than the cols table?

ehigham: Oh yeah that is weird. It's as if dict orders its keys alphabetically. Fixed in latest commit.

ehigham: I still get the following in the unit tests, though I'm not sure if there's anything I can do about that
```python
>>> mt.entries().show()
+---------+-----+---------+-----+-------+
| row_idx | foo | col_idx | bar |   baz |
+---------+-----+---------+-----+-------+
|   int32 | str |   int32 | str | int32 |
+---------+-----+---------+-----+-------+
|       0 | "a" |       0 | "c" |     1 |
|       0 | "a" |       1 | "d" |     2 |
|       1 | "b" |       0 | "c" |     3 |
|       1 | "b" |       1 | "d" |     4 |
+---------+-----+---------+-----+-------+
```

ehigham: Interestingly, a `hl.Struct` equality comparison is independent of the order of fields. @tpoterba, is this expected?

daniel-goldstein: Starting in Python 3.7 [dict order](https://docs.python.org/3/library/stdtypes.html#mapping-types-dict) is guaranteed to be insertion order, so maybe we can force that so the row/col keys come first? Not sure how that plays with hail structs

tpoterba: I think field order doesn't matter for Hail structs in practice in Python.

ehigham: OK! So we're happy with the ordering then?

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+        # In the case rows or cols aren't specified, we need to infer the
+        # matrix dimensions from *an* entry. Which one isn't important as we
+        # won't enforce congruence among input dimensions.
+        assert not ((rows is None or cols is None) and (entries is None))
+        cols = invert(cols) if cols else [{} for _ in anyval(entries)[0]]
+        for i, _ in enumerate(cols):
+            cols[i] = hl.struct(col_idx=i, **cols[i])
+
+        # While globals aren't required, `MatrixTable._unlocalize_entries`
+        # extracts matrix columns from the the `Table`'s globals.
```
danking, line 654: I'm dubious on this comment's value.

The lines that follow it just set the default value for the `globals` parameter (since, thanks to Python's default value semantics. we can't just do `globals={}` in the parameter list). If it appeared immediately before line 658 or 659, I'm amenable to it.

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+        # `MatrixTable.localize_entries` and then call `_unlocalize_entries`.
+        def invert(kvs: Dict[str, Iterable[Any]]) -> List[Dict[str, Any]]:
+            return [dict(zip(kvs, vs)) for vs in zip(*kvs.values())]
+
+        def anyval(kvs):
+            return next(iter(kvs.values()))
+
+        # In the case rows or cols aren't specified, we need to infer the
+        # matrix dimensions from *an* entry. Which one isn't important as we
+        # won't enforce congruence among input dimensions.
```
danking, line 647: Ah, I commented below already, but I disagree, we should enforce a consistent number of rows and columns. A consistent design principle of the Hail API is that we try to avoid foot guns that silently ignore bad states.

Moreover, I think the code which verifies that is so obvious in function that it wouldn't need a comment anyway. Something like:
```
n_rows = None
n_cols = None

if rows:
    c = Counter(len(x) for x in rows.values())
    if len(c) > 1:
        raise ValueError(...)
    n_rows = next(iter(c.values()))
if cols:
    # ...
if entries:
    if n_rows and len(rows) != n_rows:
         raise ...
    # check/infer n_cols for/from entries
```

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+        Returns
+        -------
+        :class:`.MatrixTable`
+            A MatrixTable assembled from inputs keyed by `row_idx`
+
+        Notes
+        -----
+        - The number of rows and columns specified in `rows` and `cols` for all
+          fields must match the matrix dimensions in `entries`.
+        - This method is intended for educational and testing purposes only.
```
danking, line 635: I think we should prefer a statement of limitations. Something like: "This method assumes all the annotations are small enough to fit in memory."

Hmm. They clearly fit in RAM though since they're Python objects passed as arguments to this method. Is there something in particular you're trying to hedge against with this Note?

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+
+        entries: :class:`dict` from :class:`str` to :class:`list` of :class:`list` of :obj:`any`
+            Matrix entries by name in the form `entry[row_idx][col_idx]`
+
+        Returns
+        -------
+        :class:`.MatrixTable`
+            A MatrixTable assembled from inputs keyed by `row_idx`
+
+        Notes
```
danking, line 631: For better or worse, we break with NumPy convention and put the notes above the parameters in our methods. Let's stay consistent with Hail in this PR. I'm open to a PR that changes every Hail doc-string to consistently use NumPy convention if the team agrees to it.

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+        cols: :class:`dict` from :class:`str` to :class:`list` of :obj:`any`
+            Column fields by name
+
+        entries: :class:`dict` from :class:`str` to :class:`list` of :class:`list` of :obj:`any`
+            Matrix entries by name in the form `entry[row_idx][col_idx]`
+
+        Returns
+        -------
+        :class:`.MatrixTable`
+            A MatrixTable assembled from inputs keyed by `row_idx`
```
danking, line 629: Matrix tables have two keys: row-keys and column-keys. Let's be clear: "whose rows are keyed by `row_idx` and whose cols are keyed by `col_idx`"

danking: End every description of a return value with a period.

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+        |   int32 |       int32 |       int32 |
+        +---------+-------------+-------------+
+        |       0 |           1 |           2 |
+        |       1 |           3 |           4 |
+        +---------+-------------+-------------+
+
+        Parameters
+        ----------
+        globals : :class:`dict` from :class:`str` to :obj:`any`
+            Global fields by name
```
danking, line 615: End every description of a parameter with a period.

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+        cols = invert(cols) if cols else [{} for _ in anyval(entries)[0]]
+        for i, _ in enumerate(cols):
+            cols[i] = hl.struct(col_idx=i, **cols[i])
+
+        # While globals aren't required, `MatrixTable._unlocalize_entries`
+        # extracts matrix columns from the the `Table`'s globals.
+        if globals is None:
+            globals = {}
+
+        cols_field_name = 'columns'
```
danking, line 658: See Env.get_uid discussion below.

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+
+        # `MatrixTable._unlocalize_entries` extracts matrix entries from the
+        # rows on the table, so these need to be defined too.
+        rows = invert(rows) if rows else [{} for _ in anyval(entries)]
+        entries = map(invert, invert(entries)) if entries else [[{} for _ in cols] for _ in rows]
+
+        entries_field_name = 'entry_structs'
+        for i, (row_pros, entry_props) in enumerate(zip(rows, entries)):
+            row_entries = [hl.struct(**kvs) for kvs in entry_props]
+            rows[i] = hl.struct(row_idx=i, **row_pros, **{entries_field_name: row_entries})
```
danking, line 669: This is a bit of a technical issue, but I'm worried that this generates O(n_rows * n_cols) IR nodes. For example,
```
from_parts(rows=dict(x=[1,2],y=[3,4]), entries=dict(z=[[1], [2]]))
```
Will generate an IR like this:
```
(CastTableToMatrix
    ...
    (TableParallelize None
        (MakeStruct
            (rows 
                (MakeArray Array[Struct{row_idx:Int32,x:Int32,y:Int32,entry_structs:Array[Struct{z:Int32}]}] 
                    (MakeStruct 
                        (row_idx  (I32 0)) (x  (I32 1)) (y  (I32 3))
                        (entry_structs  (MakeArray Array[Struct{z:Int32}]  (MakeStruct  (z  (I32 1))))))
                    (MakeStruct
                        (row_idx  (I32 1)) (x  (I32 2)) (y  (I32 4))
                        (entry_structs  (MakeArray Array[Struct{z:Int32}]  (MakeStruct  (z  (I32 2))))))))
    ...)
```
If you instead use `hl.Struct` (which is just a Python object), Hail will encode this as a `Literal`, which should provide faster serialization, less compiler overhead, and, I think, faster runtime code (because the generated JVM bytecode doesn't reify our data as control flow). For example, just changing `hl.struct` to `hl.Struct` in this function has this effect:
```
...
(TableParallelize None
    (MakeStruct
        (rows
            (Literal Array[Struct{row_idx:Int32,x:Int32,y:Int32,entry_structs:Array[Struct{z:Int32}]}]
                "[{\\"row_idx\\": 0, \\"x\\": 1, \\"y\\": 3, \\"entry_structs\\": [{\\"z\\": 1}]}, {\\"row_idx\\": 1, \\"x\\": 2, \\"y\\": 4, \\"entry_structs\\": [{\\"z\\": 2}]}]"))
...
```
I'm a bit surprised to see this rendered as a JSON string. I thought these went through a fast binary format.

Not required for this pull request, but we could push this approach even further to create, I think, substantially faster code by instead serializing arrays as literals which we just index into with the row_idx and/or col_idx.

hail/python/hail/table.py
```diff
@@ -458,8 +458,12 @@ def _select_globals(self, caller, s) -> 'Table':
...
                       key=table_key_type,
                       n_partitions=nullable(int),
-                      partial_type=nullable(dict))
-    def parallelize(cls, rows, schema=None, key=None, n_partitions=None, *, partial_type=None) -> 'Table':
+                      partial_type=nullable(dict),
+                      globals=nullable(expr_struct()))
+    def parallelize(cls, rows, schema=None, key=None, n_partitions=None, *,
+                    partial_type=None,
+                    globals=None
+                    ) -> 'Table':
```
danking, line 466: Update the doc-string to include this new parameter.

hail/python/test/hail/matrixtable/test_matrix_table_from_parts.py
```diff
@@ -0,0 +1,130 @@
+import pytest
+import hail as hl
+
+class Inits():
+    globals={'hello': 'world'}
+    rows={'foo': ['a', 'b']}
+    cols={'bar': ['c', 'd']}
+    entries={'baz': [[1, 2], [3, 4]]}
```
danking, line 8: These seem like they should be file-level variables. Am I misunderstanding the value of making them class-level variables?

hail/python/test/hail/matrixtable/test_matrix_table_from_parts.py
```diff
@@ -0,0 +1,130 @@
...
+                   )
+        ]
+
+def test_from_parts():
+    mt = hl.MatrixTable.from_parts( globals=Inits.globals
+                                  , rows=Inits.rows
+                                  , cols=Inits.cols
+                                  , entries=Inits.entries
+                                  )
+    assert_matches_globals(mt)
```
danking, line 66: I'm dubious on the relative clarity of:
```
assert_matches_globals(mt)
```
compared to
```
assert hl.eval(mt.globals) == Inits.globals
```

The latter states clearly what the assertion is whereas the former sends me hunting to the definition to understand what's being asserted.

Is there an argument in favor of the former that I'm missing?

hail/python/test/hail/matrixtable/test_matrix_table_from_parts.py
```diff
@@ -0,0 +1,130 @@
...
+    entries={'baz': [[1, 2], [3, 4]]}
+
+def unless(test: bool, kvs):
+    return {} if test else kvs
+
+def assert_matches_globals(mt: 'hl.MatrixTable', no_props=False):
+    assert hl.eval(mt.globals) == hl.Struct(**unless(no_props, Inits.globals))
+
+def assert_matches_rows(mt: 'hl.MatrixTable', no_props=False):
+    assert mt.rows().order_by('row_idx').collect() == \
```
danking, line 17: The order_by is unnecessary, the rows table is always keyed by the row key which is, by construction in the method, row_idx.

danking: In particular, if a bug was introduced that changed the row key, I'd want the tests to fail.

hail/python/test/hail/matrixtable/test_matrix_table_from_parts.py
```diff
@@ -0,0 +1,130 @@
...
+        [ hl.Struct(row_idx=0, **unless(no_props, {'foo':'a'}))
+        , hl.Struct(row_idx=1, **unless(no_props, {'foo':'b'}))
+        ]
+
+def assert_matches_cols(mt: 'hl.MatrixTable', no_props=False):
+    assert mt.cols().order_by('col_idx').collect() == \
```
danking, line 23: same argument here about column table ordering vis-a-vis the column key.

hail/python/test/hail/matrixtable/test_matrix_table_from_parts.py
```diff
@@ -0,0 +1,130 @@
...
+
+def test_optional_entries():
+    mt = hl.MatrixTable.from_parts(rows=Inits.rows, cols=Inits.cols)
+    assert_matches_globals(mt, no_props=True)
+    assert_matches_rows(mt)
+    assert_matches_cols(mt)
+    assert_matches_entries(mt, no_entry_props=True)
+
+def assert_raises_when_no_rows_and_entries():
+    with pytest.raises(AssertionError):
```
danking, line 125: Ah, nice, I wasn't aware of the `pytest.raises`. I like how concise this is compared to what I've done elsewhere:
```
try:
    ...
except AssertionError err:
    assert ... err.args[0] ...
else:
    assert False
```

hail/python/test/hail/matrixtable/test_matrix_table_from_parts.py
```diff
@@ -0,0 +1,130 @@
...
+        [ hl.Struct(col_idx=0, **unless(no_props, {'bar': 'c'}))
+        , hl.Struct(col_idx=1, **unless(no_props, {'bar': 'd'}))
+        ]
+
+def assert_matches_entries( mt: 'hl.MatrixTable'
+                          , no_row_props=False
+                          , no_col_props=False
+                          , no_entry_props=False
+                          ):
+    assert mt.entries().order_by('row_idx', 'col_idx').collect() == \
```
danking, line 33: Since you're really trying to check that `from_parts` has the right entries (in particular, you're not testing the `entries` method itself) and since you separately check the row and column annotations, I think you're totally clear to do:

```
actual = mt.select_rows().select_cols().entries()
assert actual == [hl.Struct(row_idx=0, col_idx=0, ...), ...]
```

The "select" methods [1] keep only the relevant key and those fields provided as arguments.

[1] `Table.select_globals`, `Table.select`, `MatrixTable.select_rows`, `MatrixTable.select_cols`, `MatrixTable.select_globals`, `MatrixTable.select_entries`

hail/python/hail/matrixtable.py
danking: @tpoterba am I making a mountain of an anthill about this?

danking (changes requested):
Since Daniel yielded review to me, I went ahead and did a full review. Mostly minor style things but one point that I think is rather important: we should error if the annotations are inconsistent about the number of rows or cols.

danking:
Sorry I took so long to get to this review; it was a busy day of meetings.

hail/python/hail/matrixtable.py
ehigham: I didn't know numpy did this and I wasn't aware there was a convention. Thanks for the info

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+
+        Notes
+        -----
+        - The number of rows and columns specified in `rows` and `cols` for all
+          fields must match the matrix dimensions in `entries`.
+        - This method is intended for educational and testing purposes only.
+        """
+        # General idea: build a `Table` representation matching that returned by
+        # `MatrixTable.localize_entries` and then call `_unlocalize_entries`.
+        def invert(kvs: Dict[str, Iterable[Any]]) -> List[Dict[str, Any]]:
```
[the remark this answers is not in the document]
ehigham: I guess, I'll rename

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+        cols_field_name = 'columns'
+        globals[cols_field_name] = cols
+
+        # `MatrixTable._unlocalize_entries` extracts matrix entries from the
+        # rows on the table, so these need to be defined too.
+        rows = invert(rows) if rows else [{} for _ in anyval(entries)]
+        entries = map(invert, invert(entries)) if entries else [[{} for _ in cols] for _ in rows]
+
+        entries_field_name = 'entry_structs'
+        for i, (row_pros, entry_props) in enumerate(zip(rows, entries)):
```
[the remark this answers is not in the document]
ehigham: It should. I'm bad at typing

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+        # extracts matrix columns from the the `Table`'s globals.
+        if globals is None:
+            globals = {}
+
+        cols_field_name = 'columns'
+        globals[cols_field_name] = cols
+
+        # `MatrixTable._unlocalize_entries` extracts matrix entries from the
+        # rows on the table, so these need to be defined too.
+        rows = invert(rows) if rows else [{} for _ in anyval(entries)]
```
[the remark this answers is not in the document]
ehigham: Agreed, adding the check is easy but I expected this to be useful for testing only and so didn't think that extra complexity would be worth it. I'll add the check

hail/python/hail/matrixtable.py
ehigham: I was trying to comment the block rather than just the statement that immediately followed. Guess that didn't work so well. I'll think about how to best retarget it or just remove it.

hail/python/hail/matrixtable.py
ehigham: I'm trying to say it's the caller's responsibility to make sure matrix dimensions agree. Enforcing dimensions will make remove the need for it

hail/python/test/hail/matrixtable/test_matrix_table_from_parts.py
ehigham: I'm trying to group test data within some namespace though doesn't seem to have worked well. I'll just make them file level bindings

hail/python/test/hail/matrixtable/test_matrix_table_from_parts.py
ehigham: Agreed, bad function naming. I'm trying to limit the uses of the global data so that if things need to be changed, fewer uses need to be changed.

hail/python/test/hail/matrixtable/test_matrix_table_from_parts.py
ehigham: Ah good to know, nice to know this exists!

hail/python/hail/matrixtable.py
```diff
@@ -539,6 +539,138 @@ class MatrixTable(ExprContainer):
...
+        entries = map(invert, invert(entries)) if entries else [[{} for _ in cols] for _ in rows]
+
+        entries_field_name = 'entry_structs'
```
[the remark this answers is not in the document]
ehigham: > Sorry I took so long to get to this review; it was a busy day of meetings.

Thanks for the great review! I'll address your comments tomorrow

ehigham:
> Sorry I took so long to get to this review; it was a busy day of meetings.

Thanks for the great review! I'll address your comments tomorrow

hail/python/test/hail/matrixtable/test_matrix_table_from_parts.py
ehigham: On reflection, I prefer the grouping. I think making the functions static methods also helps group together logical units of code.

danking (approved):
Looks great!

hail/python/hail/matrixtable.py
danking: The ordering looks right to me now.

hail/python/hail/matrixtable.py
danking: We'll punt on this, it can always be a follow up PR.

PR #12246, opened 2022-09-29
[batch] Add batch-config to expose current batch id to user jobs

daniel-goldstein:

PR #12662, opened 2023-02-07
[batch] Add environment variable for batch id in worker

jigold:
This is needed for Siwei's recursive pipeline to work.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1555,7 +1555,11 @@ class DockerJob(Job):
 
         self.timings: Timings = Timings()
 
-        self.env.append({'name': 'HAIL_REGION', 'value': REGION})
+        hail_extra_env = [
+            {'name': 'HAIL_REGION', 'value': REGION},
+            {'name': 'HAIL_BATCH_ID', 'value': str(batch_id)},
+        ]
+        self.env += hail_extra_env
 
         if self.secrets:
             for secret in self.secrets:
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
I thought I'd proposed this env variable but then we ended up going with [this](https://github.com/hail-is/hail/pull/12246)?

daniel-goldstein (changes requested):
See comment above

jigold:
I thought the config file was better for QoB where versioning and backwards compatibility is important and there could be substantially more configuration in the future. I don't think we mount that config file in the worker regardless for DockerJobs, but I could be wrong.

daniel-goldstein (dismissed):
> was better for QoB where versioning and backwards compatibility is important

I think we have to fulfill the same backwards compatibility with env variables, though I agree that they are more ergonomic and favorable in this case to a config file. The config file was necessary for JVMJobs because the JVM containers are long-lived, so we cannot update their environment when a new job comes in. Looks like we do [write the config](https://github.com/hail-is/hail/blob/653eff0e1b3fc3fa9bea128a2db6d2bf30f32a1b/batch/batch/worker/worker.py#L1701) in DockerJobs but don't mount it, so let's add this env variable but do the following:

- Remove the use of `write_batch_config` in `DockerJob` and move `write_batch_config` from `Job` into `JVMJob`
- Document where `write_batch_config` is used in `JVMJob` why it's used instead of an env variable. I think I still don't understand the point of further configuration, since that's still possible with environment variables, we just can't change environment variables in JVMContainers like we can with the filesystem

daniel-goldstein (approved):
awesome, thanks

PR #12666, opened 2023-02-07
[batch] Support job logs that are not UTF-8 compatible

daniel-goldstein:
The crux of this PR is the 9 line change in `file_store.py` -- we shouldn't decode job logs to `str` before uploading because we can't trust jobs to output only valid UTF-8 to stdout. The FS can take bytes anyway so it is an unnecessary conversion. The rest of the complexity of this PR is figuring out how to communicate those logs over the Batch API in a backwards-compatible and straight-forward way. What makes that tricky is that when we say "job log" in the code we mostly talk about a dictionary or JSON object of container to log string. Valid JSON must be UTF-8, so JSON starts to make less and less sense for sending log files. Log files are easy to send though if we just want one, so I added an endpoint to request the log for a specific container in a job and we can easily fulfill that without unnecessary encoding/decoding between str and bytes.

Resolves #12666

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -617,7 +617,7 @@ async def get_job_log(request, userdata, batch_id):  # pylint: disable=unused-ar
     return web.json_response(job_log_strings)
 
 
-async def get_job_container_log(request, userdata, batch_id):
+async def get_job_container_log(request, batch_id):
     app = request.app
     job_id = int(request.match_info['job_id'])
     container = request.match_info['container']
@@ -632,7 +632,7 @@ async def get_job_container_log(request, userdata, batch_id):
 @routes.get('/api/v1alpha/batches/{batch_id}/jobs/{job_id}/log/{container}')
 @rest_billing_project_users_only
 async def rest_get_job_container_log(request, userdata, batch_id):  # pylint: disable=unused-argument
-    return await get_job_container_log(request, userdata, batch_id)
+    return await get_job_container_log(request, batch_id)
 
 
 async def _query_batches(request, user, q):
@@ -2207,7 +2207,7 @@ async def ui_get_job(request, userdata, batch_id):
 @web_billing_project_users_only()
 @catch_ui_error_in_dev
 async def ui_get_job_log(request, userdata, batch_id):  # pylint: disable=unused-argument
-    return await get_job_container_log(request, userdata, batch_id)
+    return await get_job_container_log(request, batch_id)
 
 
 @routes.get('/billing_limits')
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -606,14 +606,14 @@ async def _get_full_job_status(app, record):
 async def get_job_log(request, userdata, batch_id):  # pylint: disable=unused-argument
     job_id = int(request.match_info['job_id'])
     job_log_bytes = await _get_job_log(request.app, batch_id, job_id)
-    job_log_strings: Dict[str, Optional[str]] = dict()
+    job_log_strings: Dict[str, Optional[str]] = {}
     for container, log in job_log_bytes.items():
         try:
             job_log_strings[container] = log.decode('utf-8') if log else None
-        except UnicodeDecodeError:
+        except UnicodeDecodeError as e:
             raise web.HTTPBadRequest(
                 reason=f'log for container {container} is not valid UTF-8, upgrade your hail version to download the log'
-            )
+            ) from e
     return web.json_response(job_log_strings)
 
 
@@ -2171,7 +2171,7 @@ async def ui_get_job(request, userdata, batch_id):
 
     # Not all logs will be proper utf-8 but we attempt to show them as
     # str or else Jinja will present them surrounded by b''
-    job_log_strings_or_bytes = dict()
+    job_log_strings_or_bytes = {}
     for container, log in job_log_bytes.items():
         try:
             job_log_strings_or_bytes[container] = log.decode('utf-8') if log else None
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -408,47 +408,61 @@ def has_resource_available(record):
     return True
 
 
-def attempt_id_from_spec(record):
+def attempt_id_from_spec(record) -> Optional[str]:
     return record['attempt_id'] or record['last_cancelled_attempt_id']
 
 
-async def _get_job_log(app, batch_id, job_id):
-    record = await _get_job_record(app, batch_id, job_id)
+async def _get_job_container_log_from_worker(client_session, batch_id, job_id, container, ip_address) -> bytes:
+    try:
+        async with await request_retry_transient_errors(
+            client_session,
+            'GET',
+            f'http://{ip_address}:5000/api/v1alpha/batches/{batch_id}/jobs/{job_id}/log/{container}',
+        ) as resp:
+            return await resp.read()
+    except aiohttp.ClientResponseError:
+        log.exception(f'while getting log for {(batch_id, job_id)}')
+        return b'ERROR: encountered a problem while fetching the log'
 
-    client_session: httpx.ClientSession = app['client_session']
-    file_store: FileStore = app['file_store']
-    batch_format_version = BatchFormatVersion(record['format_version'])
 
-    state = record['state']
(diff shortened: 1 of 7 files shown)


daniel-goldstein:
An additional detail that I fixed in this change is that we no longer need to load the log file into memory on the worker in order to send it to the batch pod (by using `FileResponse`). It would also be nice to do this on the batch pod by chaining the `StreamResponse` from the worker or blob storage and forward that to the client that requested the log, but I left that out for now because it seemed to be a bigger change.

batch/batch/front_end/front_end.py
```diff
@@ -586,12 +600,35 @@ async def _get_full_job_status(app, record):
...
 @rest_billing_project_users_only
 async def get_job_log(request, userdata, batch_id):  # pylint: disable=unused-argument
     job_id = int(request.match_info['job_id'])
-    job_log = await _get_job_log(request.app, batch_id, job_id)
-    return web.json_response(job_log)
+    job_log_bytes = await _get_job_log(request.app, batch_id, job_id)
+    job_log_strings: Dict[str, Optional[str]] = {}
+    for container, log in job_log_bytes.items():
+        try:
+            job_log_strings[container] = log.decode('utf-8') if log else None
```
jigold, line 612: can we check that log is not None here. I'm worried about the empty string case.

batch/batch/front_end/front_end.py
```diff
@@ -2132,11 +2169,20 @@ async def ui_get_job(request, userdata, batch_id):
         resources['actual_cpu'] = cores
         del resources['cores_mcpu']
 
+    # Not all logs will be proper utf-8 but we attempt to show them as
+    # str or else Jinja will present them surrounded by b''
+    job_log_strings_or_bytes = {}
+    for container, log in job_log_bytes.items():
+        try:
+            job_log_strings_or_bytes[container] = log.decode('utf-8') if log else None
```
jigold, line 2177: same issue with not checking if log is None here for empty strings.

batch/batch/front_end/front_end.py
```diff
@@ -2152,17 +2198,20 @@ async def ui_get_job(request, userdata, batch_id):
     return await render_template('batch', request, userdata, 'job.html', page_context)
 
 
+# This should really be the exact same thing as the REST endpoint
 @routes.get('/batches/{batch_id}/jobs/{job_id}/log/{container}')
```
jigold, line 2202: I don't think this endpoint is even used at all. You can probably just delete it.

batch/batch/worker/worker.py
```diff
@@ -2265,7 +2247,8 @@ async def create_container_and_connect(
                 except (FileNotFoundError, ConnectionRefusedError) as err:
                     attempts += 1
                     if attempts == 240:
-                        jvm_output = await container.container.get_log() or ''
+                        # NOTE we assume that the JVM logs hail emits will be valid utf-8
+                        jvm_output = (await fs.read(container.container.log_path)).decode('utf-8')
```
jigold, line 2251: Can we guarantee there is always a file existing at `container.container.log_path`?

batch/batch/worker/worker.py
```diff
@@ -1202,13 +1202,6 @@ def container_is_running(self):
     def container_finished(self):
         return self.process is not None and self.process.returncode is not None
 
-    async def get_log(self, offset: Optional[int] = None):
```
jigold, line 1205: Did we never use the offset anywhere?

batch/batch/front_end/front_end.py
```diff
@@ -408,47 +408,61 @@ def has_resource_available(record):
...
-async def _get_job_log(app, batch_id, job_id):
-    record = await _get_job_record(app, batch_id, job_id)
+async def _get_job_container_log_from_worker(client_session, batch_id, job_id, container, ip_address) -> bytes:
+    try:
+        async with await request_retry_transient_errors(
+            client_session,
+            'GET',
+            f'http://{ip_address}:5000/api/v1alpha/batches/{batch_id}/jobs/{job_id}/log/{container}',
+        ) as resp:
+            return await resp.read()
```
jigold, line 422: Just so I understand, the FileResponse here doesn't deserve special treatment as it's just a special case of a regular Response?

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -323,6 +329,10 @@ async def wait(self):
             if i < 64:
                 i = i + 1
 
+    async def container_log(self, container_name: str) -> bytes:
+        async with await self._batch._client._get(f'/api/v1alpha/batches/{self.batch_id}/jobs/{self.job_id}/log/{container_name}') as resp:
+            return await resp.read()
```
jigold, line 334: Is this supposed to be a FileResponse as well?

jigold (dismissed):

batch/batch/front_end/front_end.py
daniel-goldstein: this is used for downloading logs from the UI. I meant to refactor this and remove the comment. I'll do that now

batch/batch/worker/worker.py
daniel-goldstein: No this was dead code

batch/batch/worker/worker.py
daniel-goldstein: Hm maybe not here, I'll protect against this

batch/batch/front_end/front_end.py
daniel-goldstein: Yes, I think it subclasses `Response`. The client can't tell, but it's an efficient way for the server to stream bytes directly from an open file handle instead of loading the whole thing into memory on the server and then sending those bytes.

hail/python/hailtop/batch_client/aioclient.py
daniel-goldstein: The client can't tell whether or not it's a `FileResponse`.

jigold:
Just so I understand correctly (and sorry if this is obvious), the current job logs interface is still the same. But if you want a container's logs, then you'll get bytes which the user will have to decode themselves. How does that affect the file download button in the UI and the hailctl batch logs functionality you have? Will you see text or a random byte string?

jigold (dismissed):
see question

daniel-goldstein:
> Just so I understand correctly (and sorry if this is obvious), the current job logs interface is still the same. But if you want a container's logs, then you'll get bytes which the user will have to decode themselves. How does that affect the file download button in the UI and the hailctl batch logs functionality you have? Will you see text or a random byte string?

Good question. For the download button, the file you download is still a normal text file and I've confirmed that I can download and view a log file as I would expect. For the `hailctl batch logs` functionality, I added logic to the CLI in this PR where I download the bytes of the log and if I can decode it as UTF-8 I do, so it prints exactly as before, and if I can't I just print the bytes.

jigold:
Ok. Last question -- do we need to do something for the aioclient for the new `container_log` method or are you assuming a client can make the utf-8 conversion themselves?

daniel-goldstein:
> Ok. Last question -- do we need to do something for the aioclient for the new `container_log` method or are you assuming a client can make the utf-8 conversion themselves?

I would leave it up to the caller of the `aioclient.BatchClient.container_log` method. At the end of the day it's returning a file which is more generally in bytes than str, so I feel like it makes sense for the method to return `bytes`.

batch/batch/front_end/front_end.py
```diff
@@ -586,12 +600,39 @@ async def _get_full_job_status(app, record):
...
+        try:
+            job_log_strings[container] = log.decode('utf-8') if log is not None else None
+        except UnicodeDecodeError as e:
+            raise web.HTTPBadRequest(
+                reason=f'log for container {container} is not valid UTF-8, upgrade your hail version to download the log'
+            ) from e
+    return web.json_response(job_log_strings)
+
+
+async def get_job_container_log(request, userdata, batch_id):
```
jigold, line 620: The checks are failing because `userdata` is not used. I think we don't need to use it because the decorators on the actual routes does the user validation check before it gets to this point. Would be great if you could double check that.

jigold (changes requested):

batch/batch/front_end/front_end.py
daniel-goldstein: Yup you're right it's used by the decorator and unused in the function I added.

jigold (approved):

PR #12671, opened 2023-02-08
[compiler] Redesign Repartition IR nodes to be naive coalesce only

tpoterba:
This changes the execution semantics for the Spark backend. Instead of dispatching either to (a) a two-pass algorithm that scans and coalesces (shuffle=False) or (b) a two-pass algorithm that shuffles and then rekeys (shuffle=True), we use write/read instead.

### DIFF
--- a/hail/python/test/hail/matrixtable/test_matrix_table.py
+++ b/hail/python/test/hail/matrixtable/test_matrix_table.py
@@ -2011,13 +2011,6 @@ def test_matrix_randomness():
     assert_contains_node(mt, ir.MatrixExplodeRows)
     assert_unique_uids(mt)
 
-    # test MatrixRepartition
-    if not hl.current_backend().requires_lowering:
-        rmt = hl.utils.range_matrix_table(20, 10, 3)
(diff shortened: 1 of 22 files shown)


hail/python/test/hail/matrixtable/test_matrix_table.py
```diff
@@ -2011,13 +2011,6 @@ def assert_contains_node(t, node):
     assert_contains_node(mt, ir.MatrixExplodeRows)
     assert_unique_uids(mt)
 
-    # test MatrixRepartition
-    if not hl.current_backend().requires_lowering:
-        rmt = hl.utils.range_matrix_table(20, 10, 3)
-        mt = rmt.repartition(5)
-        assert_contains_node(mt, ir.MatrixRepartition)
-        assert_unique_uids(mt)
-
```
patrick-schultz, line 2020: Can we rewrite this and the TableRepartition one to use naive_coalesce, so the handle_randomness implementations still have test coverage?

patrick-schultz (dismissed):

PR #12678, opened 2023-02-10
[query][scala-fs] more debugging information when memory fails

danking:


hail/src/main/scala/is/hail/io/fs/ServiceCacheableFS.scala
```diff
@@ -49,6 +49,7 @@ trait ServiceCacheableFS extends FS {
             using(createNoCompression(filename)) { os =>
               os.write(data)
             }
+            log.info(s"Sucessfully failed over to GCS")
```
tpoterba: typo, but who cares

daniel-goldstein (approved):

PR #12679, opened 2023-02-10
[query] add `hl.utils.genomic_range_table`

danking:
CHANGELOG: In Query on Batch, `hl.balding_nichols_model` is slightly faster. Also added `hl.utils.genomic_range_table` to quickly create a table keyed by locus.

It has grated on me for a while that `hl.balding_nichols_models` requires a whole pass to verify it is sorted even though it is plainly so. This change introduces the necessary infrastructure to convince Hail of that fact.

### DIFF
--- a/hail/python/hail/expr/types.py
+++ b/hail/python/hail/expr/types.py
@@ -16,7 +16,7 @@ from hailtop.hail_frozenlist import frozenlist
 from .nat import NatBase, NatLiteral
 from .type_parsing import type_grammar, type_node_visitor
 from .. import genetics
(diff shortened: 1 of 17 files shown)


chrisvittal (commented):
I have no major objections to this, but I am curious as to why you went the route of a new table IR, rather than implementing this with the unsafe key by on a normal range table?

danking:
AFAIK, unsafe key-by would still do an unnecessary scan across the data to compute partition bounds. This IR is exactly 1 pass (instead of 2).

PR #12681, opened 2023-02-10
[batch] Profile more services not just batch-driver

daniel-goldstein:


gear/gear/profiling.py
```diff
@@ -0,0 +1,30 @@
...
+        if profiler_tag == 'default':
+            profiler_tag = DEFAULT_NAMESPACE + f'-{HAIL_SHA[0:12]}'
+        googlecloudprofiler.start(
+            service=service,
+            service_version=profiler_tag,
+            # https://cloud.google.com/profiler/docs/profiling-python#agent_logging
+            verbose=3,
+        )
+
+    def ignore_failed_to_collect_and_upload_profile(record):
```
jigold, line 24: Does this block need to be indented to the left by 1 tab?

gear/gear/profiling.py
```diff
@@ -0,0 +1,30 @@
+import logging
+import os
+
+import googlecloudprofiler
+
+HAIL_SHA = os.environ['HAIL_SHA']
+HAIL_SHOULD_PROFILE = os.environ.get('HAIL_SHOULD_PROFILE', False)
```
jigold, line 7: You need to cast this to a boolean.

jigold (changes requested):

gear/gear/profiling.py
daniel-goldstein: Changed to an `in` check, since we only ever include that environment variable when we want to profile.

jigold (approved):

PR #12691, opened 2023-02-12
[query] improve the memory client

danking:
Currently, the memory client buffers the entire output in memory which is likely to cause OOMs. For reasons that are not entirely clear to me, sometimes these OOMs get muffled by our system and instead lead to non-termination. I vaguely remember this happening before with `using`. I suspect there is something somewhat subtle wrong with that method, but I am not certain.

Anyway, there are four big changes here:
1. Do not buffer the entire request body in memory when writing to memory.
2. Because of (1) we have to pull retry behavior all the way up to the top-level where we know how to recreate the body.
3. Because of (2) it is easier to provide a `write(url)(writerFunction)` style API, which I do here.
4. Again, because of (2), and because I want to preserve the file-object-like interface, I added a somewhat funky anonymous class which uses a second thread to facilitate the movement of data written into the OutputStream returned by `create` into the OutputStream of the HTTP connection.

Point (4) probably bears more explanation. The root issue is the bad Apache HTTP Client interface. Instead of `request` returning an OutputStream, it takes an "entity". An entity knows how to write itself into the OutputStream of an HTTP request. This works fine if the "writer" code is pased as a function (as in my new `write` method), but that does not work if the control flow looks like:

    f = create(...)
    f.write(...)
    r.close()

We avoid this limited API by initiating the request in a second thread which will eventually block waiting to receive data from a PipedInputStream. That PipedInputStream produces the data written to a PipedOutputStream. The `create` call returns a positioned OutputStream which just writes data into the PipedOutputStream and handles cleaning up the thread when it is closed.

In a multi-core system, network requests should proceed in parallel to the client code. In a single-core system, the written data will buffer until `close` is called which will definitely yield control to the other thread.

### DIFF
--- a/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
@@ -114,26 +114,24 @@ class ServiceBackend(
     val root = s"${ backendContext.remoteTmpDir }parallelizeAndComputeWithIndex/$token"
 
     // FIXME: HACK
-    val (open, create) = if (n <= 50) {
-      (fs.openCachedNoCompression _, fs.createCachedNoCompression _)
+    val (open, write) = if (n <= 50) {
+      (fs.openCachedNoCompression _, fs.writeCached _)
     } else {
-      ((x: String) => fs.openNoCompression(x), fs.createNoCompression _)
+      ((x: String) => fs.openNoCompression(x), fs.writePDOS _)
     }
 
     log.info(s"parallelizeAndComputeWithIndex: $token: nPartitions $n")
     log.info(s"parallelizeAndComputeWithIndex: $token: writing f and contexts")
 
     val uploadFunction = scalaConcurrent.Future {
-      retryTransientErrors {
-        using(new ObjectOutputStream(create(s"$root/f"))) { os =>
-          os.writeObject(f)
-        }
+      write(s"$root/f") { fos =>
+        using(new ObjectOutputStream(fos)) { oos => oos.writeObject(f) }
       }
     }
 
     val uploadContexts = scalaConcurrent.Future {
       retryTransientErrors {
-        using(create(s"$root/contexts")) { os =>
+        write(s"$root/contexts") { os =>
           var o = 12L * n
           var i = 0
           while (i < n) {
@@ -143,7 +141,6 @@ class ServiceBackend(
             i += 1
             o += len
           }
-          log.info(s"parallelizeAndComputeWithIndex: $token: writing contexts")
           collection.foreach { context =>
             os.write(context)
           }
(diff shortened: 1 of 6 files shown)


danking:
With these changes, I was not only able to get #12423 to pass but the test that was failing (`test_vcf_parser_golden_master__gvcf_GRCh37`) now uses the memory service! I think that test was actually OOM'ing when sending to memory and triggering a GCS retry.

danking:
With these changes, I was not only able to get #12423 to pass but the test that was failing (`test_vcf_parser_golden_master__gvcf_GRCh37`) now uses the memory service! I think that test was actually OOM'ing when sending to memory and triggering a GCS retry.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -115,26 +115,24 @@ class ServiceBackend(
...
 
     log.info(s"parallelizeAndComputeWithIndex: $token: nPartitions $n")
     log.info(s"parallelizeAndComputeWithIndex: $token: writing f and contexts")
 
     val uploadFunction = scalaConcurrent.Future {
-      retryTransientErrors {
-        using(new ObjectOutputStream(create(s"$root/f"))) { os =>
-          os.writeObject(f)
-        }
+      write(s"$root/f") { fos =>
```
daniel-goldstein, line 128: How come you removed the `retryTransientErrors` here?

daniel-goldstein (changes requested):
This seems like an overall good change, and the `ServiceBackend` and `Worker` already hold the computed result/data that they want to write to the service, so it doesn't seem like moving the retry up is a big deal, but it does make me wonder was this test already really close to the memory limit? The memory service has a really really small max payload (like 1Mib) so either the test still falls back to GCS or it OOM'd by buffering that <1Mib of extra data).

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
danking: whoops, I'll restore

danking:
How does the max payload interact with chunked transfer?

danking:
This change uses chunked transfers instead of sending it all in one go.

daniel-goldstein:
I believe that regardless of chunking, the total body of the request is not to exceed aiohttp's [client_max_size](https://docs.aiohttp.org/en/stable/web_reference.html?highlight=client_max_size#application). Internal-gateway will stream transparently to aiohttp which will buffer the chunks [here](https://github.com/hail-is/hail/blob/235d2bcba1d4594a27a3dea6947c91cc4043de72/memory/memory/memory.py#L61) and blow up.

It could be that adding chunking will help java to better see the 429 instead of erroring with `Connection closed` or hanging, but that is just wishful thinking, it could be happening at a deeper level of the network stack.

danking:
Hmm. Is it possible that gateway dropped large HTTP packets?

danking:
I still don’t have a satisfactory answer for why this makes my other PR pass but I think we should merge since it’s a generally good change.

daniel-goldstein (approved):
In the same boat as you

PR #12692, opened 2023-02-13
[hailtop] Keep strong references to tasks

daniel-goldstein:
The asyncio event loop only keeps weak references to tasks, so wherever we call `asyncio.create_task` we need to ensure that we keep a strong reference to its result. Specifically, `BackgroundTaskManager` needs to keep strong references not weak references to the tasks it creates. This is easy to do without accumulating garbage by using a done callback on the task to remove itself from the set. However, this felt iffy with the threadsafe futures, which were only used in sync.py anyway, so I pushed that functionality directly into sync.py and removed it from the `BackgroundTaskManager`.

To simplify the ownership story for tasks, this changes `BackgroundTaskManager` to *not* return the task and instead hold onto strong references. If a client wants a reference to the task it creates, it should call `asyncio.create_task` directly and manage the lifecycle of the spawned task. This required only a few small changes in worker.py since most of the codebase does not assign the result of `task_manager.ensure_future`. The only change that gave me pause was the handling of `mjs_fut`, whose lifetime is a little tricky since it is potentially passed to yet another task. I think this shows a general weakness in the handling of ownership and lifetimes in between the Job and Worker classes and think a larger refactor can make this less error-prone but is out of scope for this fix. So I'd appreciate an especially scrutinizing look at that piece of the code.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -16,7 +16,19 @@ import uuid
 import warnings
 from collections import defaultdict
 from contextlib import AsyncExitStack, ExitStack
-from typing import Any, Awaitable, Callable, ContextManager, Dict, List, MutableMapping, Optional, Tuple, Union
+from typing import (
+    Any,
+    Awaitable,
+    Callable,
+    ContextManager,
+    Coroutine,
+    Dict,
+    List,
+    MutableMapping,
+    Optional,
+    Tuple,
+    Union,
+)
 
 import aiodocker  # type: ignore
 import aiodocker.images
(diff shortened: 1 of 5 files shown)


batch/batch/worker/worker.py
```diff
@@ -1691,9 +1702,8 @@ async def run(self):
         async with self.worker.cpu_sem(self.cpu_in_mcpu):
             self.start_time = time_msecs()
 
+            mjs_fut = self.mark_started()
```
danking, line 1705: ah this is much nicer.

batch/batch/worker/worker.py
```diff
@@ -1793,8 +1803,16 @@ async def run(self):
...
+                        # but if it either is not run or is cancelled then we need
+                        # to cancel MJS
                         if not isinstance(exc, asyncio.CancelledError):
-                            await self.mark_complete()
+                            try:
+                                await self.mark_complete(mjs_fut)
+                            except asyncio.CancelledError:
+                                mjs_fut.cancel()
+                        else:
+                            mjs_fut.cancel()
```
danking, line 1815: Your change preserves semantics, so I'm totally good with it. However, it's worth noting that cancellation muffles exceptions. I think that's right in this case because we got interrupted so it doesn't matter what happened to MJS.

batch/batch/worker/worker.py
```diff
@@ -1964,13 +1982,12 @@ async def run(self):
             # file with explicit versioning to make sure we maintain backwards compatibility.
             self.write_batch_config()
 
+            mjs_fut = self.mark_started()
```
danking, line 1985: hmm. I wonder why we didn't do this before the JVM acquisition before. Seems better this way.

gear/gear/database.py
```diff
@@ -265,12 +268,14 @@ def __init__(self, rv):
 class Database:
     def __init__(self):
         self.pool = None
+        self.connection_release_task_manager = None
```
danking, line 271: lint complained about this line missing, I added it.

danking (approved):
Looks great.

danking:
I'm definitely here for a refactor that structures these things using an exit stack or just plain try-finallies.

PR #12696, opened 2023-02-13
[compiler] Iterative `DistinctlyKeyed` Analysis

ehigham:
Use iterative tree traversals to prevent exceeding stack size for large IRs.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/DistinctlyKeyed.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/DistinctlyKeyed.scala
@@ -1,71 +1,96 @@
 package is.hail.expr.ir
 
+import is.hail.utils.TreeTraversal
+
 object DistinctlyKeyed {
-  def apply(node: BaseIR): DistinctKeyedAnalysis = {
-    val distinctMemo = Memo.empty[Unit]
-    analyze(node, distinctMemo)
-    DistinctKeyedAnalysis(distinctMemo)
-  }
-  private def analyze(node:BaseIR, memo: Memo[Unit]): Unit = {
-    def basicChildrenCheck(children: IndexedSeq[BaseIR]): Unit = {
-      children.foreach(child => analyze(child, memo))
-      if (children.forall(child => memo.contains(child)))
-        memo.bind(node, ())
-    }
-    node match {
-      case TableLiteral(_,_,_,_) =>
-      case x@TableRead(_, _, _) =>
-        if(x.isDistinctlyKeyed)
-          memo.bind(node, ())
-      case TableParallelize(_,_) =>
-      case TableKeyBy(child, keys, _) =>
-        analyze(child, memo)
-        if (child.typ.key.forall(cKey => keys.contains(cKey)) && memo.contains(child))
-          memo.bind(node, ())
-      case TableRange(_,_) => memo.bind(node, ())
-      case TableFilter(child, _) => basicChildrenCheck(IndexedSeq(child))
-      case TableHead(child, _) => basicChildrenCheck(IndexedSeq(child))
-      case TableTail(child, _) => basicChildrenCheck(IndexedSeq(child))
-      case TableRepartition(child, _, _) => basicChildrenCheck(IndexedSeq(child))
-      case TableJoin(left, right, _, _) => basicChildrenCheck(IndexedSeq(left, right))
-      case TableIntervalJoin(left, right, _, _) => basicChildrenCheck(IndexedSeq(left, right))
-      case TableMultiWayZipJoin(children, _, _) => basicChildrenCheck(children)
-      case TableLeftJoinRightDistinct(left, right, _) => basicChildrenCheck(IndexedSeq(left, right))
-      case TableMapPartitions(child, _, _, _) => analyze(child, memo)
-      case TableMapRows(child, _) => basicChildrenCheck(IndexedSeq(child))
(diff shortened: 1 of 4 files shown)


hail/src/main/scala/is/hail/expr/ir/RefEquality.scala
```diff
@@ -31,6 +31,9 @@ class Memo[T] private(val m: mutable.HashMap[RefEquality[BaseIR], T]) {
     this
   }
 
+  def bindIf(test: Boolean, ir: BaseIR, t: T): Memo[T] =
```
tpoterba, line 34: This is certainly fine in the unit case, but we might not want to use this naively for every analysis if constructing `t` is not cheap (as it is for Unit), since we'll need to construct `t` even if we don't end up binding it.

An alternative is changing the signature to `def bindIf(test: Boolean, ir: BaseIR, t: => T)`, but this has other performance consequences -- Scala desugars this to a `() => T` and now a closure is allocated for every call to the function.

I think this is fine, we just won't be able to use it everywhere.

hail/src/main/scala/is/hail/expr/ir/DistinctlyKeyed.scala
```diff
@@ -1,71 +1,96 @@
...
+
+      case RelationalLetTable(_, _, body) =>
+        Iterator.single(body)
+
+      case t@(_: IR
+              | _: TableDistinct
+              | _: TableExplode
+              | _: TableFilterIntervals
+              | _: TableHead
+              | _: TableIntervalJoin
```
tpoterba, line 72: you didn't change this behavior, but TableIntervalJoin is actually inherits distinctness just from the left -- it joins intervals from the right onto the left without a cartesian product

hail/src/main/scala/is/hail/expr/ir/DistinctlyKeyed.scala
```diff
@@ -1,71 +1,96 @@
...
+              | _: TableJoin
+              | _: TableKeyBy
+              | _: TableLeftJoinRightDistinct
```
tpoterba, line 75: TableLeftJoinRightDistinct is also not strict enough here. Should just propagate distinctness from the left child.

hail/src/main/scala/is/hail/utils/TreeTraversal.scala
```diff
@@ -0,0 +1,67 @@
...
+// "Lightweight" (less-safe) implementations of tree traversal algorithms
+// inspired by those in Guava
+object TreeTraversal {
+
+  def postOrder[A](adj: A => Iterator[A])(root: A): Iterator[A] =
+    new Iterator[A] {
+      // Java (and Scala) iterators mutate on `next()` so it's convenient
+      // to hold on to a node and its children as we visit the node after
+      // its children.
+      private var stack = List((root, adj(root)))
```
tpoterba, line 14: Looks good.

tpoterba (dismissed):
great work -- couple tiny nits (mostly from looking at the existing analysis for the first time in a while :) )

hail/src/main/scala/is/hail/expr/ir/RefEquality.scala
ehigham: You're right, call by name makes much more sense here. Use it where appropriate I guess. Do you know if visualvm works with scala code?

hail/src/main/scala/is/hail/expr/ir/DistinctlyKeyed.scala
ehigham: Thanks! I'll update it

hail/src/main/scala/is/hail/expr/ir/RefEquality.scala
ehigham: BTW, really appreciate the explanations in comments. Thank you for taking the time.

tpoterba: we typically use either YourKit or async-profiler. async-profiler is free and lighter-weight, but the last time we looked a year or two ago it didn't provide timings at the level of line symbols within functions, and as such wasn't quite as useful for profiling our enormous functions inside generated code.

hail/src/main/scala/is/hail/expr/ir/DistinctlyKeyed.scala
tpoterba: it wasn't wrong, but it was unnecessarily loose. The semantic distinctness of the TableLeftJoinRightDistinct node doesn't care if the right side is distinct or not, but the previous implementation did.

ehigham: I'm now wondering if visiting a subset of the nodes in the tree is the right thing to do. What if there's a subtree in one of these unvisited nodes that might be distinctly keyed and we're not discovering it?
Is this analysis used top-down in a way that  making that unimportant?

tpoterba: Ah, you're totally right -- in the general case, we need to visit every TableIR. This isn't just used top-down, it's used in lowering joins to avoid grouping the right side if we know it's already distinct.

We want to visit all TableIRs, but the distinctness calculation for any given IR might not care about some inputs

ehigham: Sounds good!

tpoterba (approved):
great work!

PR #12697, opened 2023-02-14
[batch, WIP] Job Groups

jigold:


danking:
@jigold it looks like the full text search and something about duplicate cancellation (the op id?) is in here too; can you excise that and I'll take another look?

danking:
@jigold bump

jigold:
Sorry. I need to get my head back into this again and I need to do the billing fixes from the redundant resource prices first. Unassigning you for now until it's ready.

jigold:
The full text search is there because we don't know how long the path name will be and I thought this was better for looking for prefix matches with "LIKE". Maybe I'm wrong.

The cancellation_op_id is gone, but I'm still worried about the performance of some of these queries.

jigold:
I want cancel_after_n_failures to be on a job group. The things a job group doesn't have which maybe it should is:
- callback
- attributes
- updates

I think updates should be on a batch and not part of a job group. An update can add jobs to multiple job groups. Otherwise, the batches table should only have static fields that apply to the entire batch. I think we can do callbacks and attributes on a job group. I added a PATCH endpoint to be able to update a job group's cancel_after_n_attributes as the hailtop.batch interface was going to automatically generate job groups without any configuration settings.

As for the full text search, I think prefix searches are faster with full text search than with a regular index, but I could be wrong. We'd have to benchmark it.

> If we made batches simpler, does that ease complexity and decrease code duplication? In particular, what if batches didn't contain jobs at all? Instead, a batch contains exactly one job group. That job group contains zero or more job groups. Job groups manage: resource aggregation, cancellation, etc.

I believe my plan is basically already doing this. It might not be clear because I didn't put the migrations in. But basically all of the current batches tables are now indexed by batch_id, job_group_id where the current "batch" has job_group_id = 1.

jigold:
Closing for now while I incorporate the feedback and make a working prototype.

danking:
I agree cancel_after_n_failures should be on the group. That lets us match Spark semantics for QoB.

1. I agree, callback per group seems valuable.
2. I agree attributes seem useful on groups.
3. I agree, not much value in updates being at the job-group level. 

Depends what you mean by prefix search, if that means `LIKE "X%"`, I think that'll be quite fast on a normal index because you can jump directly to the first record whose prefix is X. I don't see how a fulltext index could do any better in that case. On the other hand, if you mean `LIKE "%X"` then I agree, a normal index is useless and MySQL will do a table scan. In that case, I expect a fulltext index to be a substantial improvement.

> I believe my plan is basically already doing this. It might not be clear because I didn't put the migrations in. But basically all of the current batches tables are now indexed by batch_id, job_group_id where the current "batch" has job_group_id = 1.

Ah, that sounds good. So the plan would be to drop, for example, `aggregated_batch_resources_v2` and the other tables which are now replaced with the job group ones? That's exactly what I had in mind.

PR #12699, opened 2023-02-15
[qob] Enable liftovers for Query-on-Batch

daniel-goldstein:
The bulk of this PR is plumbing to change `TLocus(ReferenceGenome) => TLocus(String)` and bring reference genomes off of global mutable state onto the Backend (both in scala and python). This allows mutable reference genomes in QoB and by extension liftovers. This also removes some duplication between local_backend and spark_backend in python and simplifies the initialization of reference genomes.

### DIFF
--- a/hail/python/hail/backend/local_backend.py
+++ b/hail/python/hail/backend/local_backend.py
@@ -249,21 +249,11 @@ class LocalBackend(Py4JBackend):
         jir = self._to_java_blockmatrix_ir(bmir)
         return tblockmatrix._from_java(jir.typ())
 
-    def add_reference(self, config):
-        self._jbackend.pyAddReference(json.dumps(config))
-
     def load_references_from_dataset(self, path):
         return json.loads(self._jbackend.pyLoadReferencesFromDataset(path))
 
     def from_fasta_file(self, name, fasta_file, index_file, x_contigs, y_contigs, mt_contigs, par):
         return json.loads(self._jbackend.pyFromFASTAFile(name, fasta_file, index_file, x_contigs, y_contigs, mt_contigs, par))
-
-    def remove_reference(self, name):
-        self._jbackend.pyRemoveReference(name)
-
-    def _get_non_builtin_reference(self, name):
-        from hail.genetics.reference_genome import ReferenceGenome
-        return ReferenceGenome._references[name]._config
 
     def add_sequence(self, name, fasta_file, index_file):
         self._jbackend.pyAddSequence(name, fasta_file, index_file)
--- a/hail/python/hail/backend/local_backend.py
+++ b/hail/python/hail/backend/local_backend.py
@@ -249,24 +249,6 @@ class LocalBackend(Py4JBackend):
         jir = self._to_java_blockmatrix_ir(bmir)
         return tblockmatrix._from_java(jir.typ())
 
-    def load_references_from_dataset(self, path):
-        return json.loads(self._jbackend.pyLoadReferencesFromDataset(path))
-
-    def from_fasta_file(self, name, fasta_file, index_file, x_contigs, y_contigs, mt_contigs, par):
-        return json.loads(self._jbackend.pyFromFASTAFile(name, fasta_file, index_file, x_contigs, y_contigs, mt_contigs, par))
-
-    def add_sequence(self, name, fasta_file, index_file):
-        self._jbackend.pyAddSequence(name, fasta_file, index_file)
-
-    def remove_sequence(self, name):
-        self._jbackend.pyRemoveSequence(name)
-
-    def add_liftover(self, name, chain_file, dest_reference_genome):
-        self._jbackend.pyAddLiftover(name, chain_file, dest_reference_genome)
-
-    def remove_liftover(self, name, dest_reference_genome):
-        self._jbackend.pyRemoveLiftover(name, dest_reference_genome)
-
     def parse_vcf_metadata(self, path):
         return json.loads(self._jhc.pyParseVCFMetadataJSON(self._jbackend.fs(), path))
 
--- a/hail/python/hail/backend/local_backend.py
+++ b/hail/python/hail/backend/local_backend.py
@@ -16,7 +16,7 @@ from hail.expr.table_type import ttable
 from hail.expr.types import dtype
 from hail.ir import finalize_randomness
 from hail.ir.renderer import CSERenderer
-from hail.utils.java import scala_package_object, scala_object
+from hail.utils.java import scala_package_object
 from .py4j_backend import Py4JBackend, handle_java_exception
 from ..fs.local_fs import LocalFS
 from ..hail_logging import Logger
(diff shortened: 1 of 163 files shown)


tpoterba:
I looked at all of these usages. Since the only value types that use the state manager are sets and dicts with locus in the elt/key, it's impossible to construct a failing example for most of these. It's possible for a usage of the RVDContext RVB in `orderedLeftJoinDistinctAndInsert`, but we can replace the implementation of `TableLeftJoinRightDistinct.execute` with a wrapped call to the lowered implementation (not necessary in this PR).

hail/python/hail/backend/local_backend.py
```diff
@@ -183,27 +180,31 @@ def hail_package(self):
     def utils_package_object(self):
         return self._utils_package_object
 
+    def register_ir_function(self,
```
tpoterba, line 183: move to py4jbackend? looks shared with spark

hail/src/main/scala/is/hail/backend/service/Worker.scala
```diff
@@ -45,6 +46,36 @@ class WorkerTimer() {
   }
 }
 
+object ExplicitClassLoaderInputStream {
```
tpoterba, line 49: can we add a comment here that links to the scala bug?

tpoterba: (and maybe a one-sentence description of the problem)

hail/src/main/scala/is/hail/compatibility/LegacyEncodedTypeParser.scala
```diff
@@ -1,15 +1,15 @@
...
+import is.hail.expr.ir.{IRParser, PunctuationToken, TokenIterator}
 import is.hail.types.encoded._
 import is.hail.types.virtual._
 import is.hail.rvd.RVDType
 import is.hail.utils.FastIndexedSeq
 
 object LegacyEncodedTypeParser {
 
-  def legacy_type_expr(env: TypeParserEnvironment)(it: TokenIterator): (Type, EType) = {
+  def legacy_type_expr(it: TokenIterator): (Type, EType) = {
```
tpoterba, line 12: love this.

hail/src/main/scala/is/hail/expr/ir/EmitClassBuilder.scala
```diff
@@ -472,21 +476,19 @@ class EmitClassBuilder[C](
   }
 
   def getPType[T <: PType : TypeInfo](t: T): Code[T] = {
-    val references = ReferenceGenome.getReferences(t.virtualType).toArray
-    val setup = Code(Code(references.map(addReferenceGenome)),
-      Code.checkcast[T](
+    ReferenceGenome.getReferences(t.virtualType).toArray.map(getReferenceGenome)
```
tpoterba, line 479: so I didn't realize this earlier, but the reason these methods serialize to strings and parse is that they don't want to serialize a RG inside a locus type and end up with multiple distinct RG objects for the same reference on the worker. That's not a problem anymore! You can just use `mb.getObject(t)` for both these implementations! 🎉

hail/src/main/scala/is/hail/linalg/BlockMatrix.scala
```diff
@@ -415,16 +415,16 @@ class BlockMatrix(val blocks: RDD[((Int, Int), BDM[Double])],
   val nCols: Long) extends Serializable {
 
   import BlockMatrix._
-  
+
```
tpoterba, line 418: if it's not a ton of trouble, would you mind reverting the whitespace changes here? Might cause rebase trouble with the other BM churn happening right now with lowering.

hail/src/main/scala/is/hail/linalg/RowMatrix.scala
```diff
@@ -16,18 +16,18 @@ import org.apache.spark.{Partition, Partitioner, SparkContext, TaskContext}
 object RowMatrix {
   def apply(rows: RDD[(Long, Array[Double])], nCols: Int): RowMatrix =
     new RowMatrix(rows, nCols, None, None)
-  
+
```
tpoterba, line 19: same as BM

hail/src/main/scala/is/hail/rvd/RVDType.scala
```diff
@@ -28,22 +29,25 @@ final case class RVDType(rowType: PStruct, key: IndexedSeq[String])
     .filter(i => !keySet.contains(rowType.fields(i).name))
     .toArray
 
-  val kOrd: UnsafeOrdering = kType.unsafeOrdering()
```
tpoterba, line 31: making these `def`s is causing some of the usages to now create orderings for each comparison. We should create an interface here that makes it impossible for these to be created multiple times. How about:

```
  private[this] var sm: HailStageManager = _
  def setStateManager(sm: HailStateManager): Unit = this.sm = sm
  
  lazy val kOrd: UnsafeOrdering = kType.unsafeOrdering(sm)
  lazy val kInRowOrd: UnsafeOrdering =
    RVDType.selectUnsafeOrdering(rowType, kFieldIdx, rowType, kFieldIdx, sm)
  lazy val kRowOrd: UnsafeOrdering =
    RVDType.selectUnsafeOrdering(kType, Array.range(0, kType.size), rowType, kFieldIdx, sm)
```

Now you just need to call `setStateManager` somewhere above, and you're guaranteed not to create multiple orderings per type.

hail/src/main/scala/is/hail/types/virtual/Type.scala
```diff
@@ -197,9 +197,9 @@ abstract class Type extends BaseType with Serializable {
 
   def canCompare(other: Type): Boolean = this == other
 
-  def mkOrdering(missingEqual: Boolean = true): ExtendedOrdering
+  def mkOrdering(sm: HailStateManager, missingEqual: Boolean = true): ExtendedOrdering
 
-  def ordering: ExtendedOrdering = mkOrdering()
+  def ordering(sm: HailStateManager): ExtendedOrdering = mkOrdering(sm)
```
tpoterba, line 202: This should cache the ordering:

```
@transient protected var ord: ExtendedOrdering = _
def ordering(sm: HailStateManager): ExtendedOrdering = {
  if (ord != null)
    ord
  else {
    ord =  mkOrdering(sm)
    ord
  }
}

```

hail/src/main/scala/is/hail/variant/ReferenceGenome.scala
```diff
@@ -646,40 +602,32 @@ object ReferenceGenome {
         val rgPath = fileSystem.getPath.toString
         val rg = using(fs.open(rgPath))(read)
         val name = rg.name
-        if (ReferenceGenome.hasReference(name) && ReferenceGenome.getReference(name) != rg)
-          fatal(s"'$name' already exists and is not identical to the imported reference from '$rgPath'.")
+        // if (ReferenceGenome.hasReference(name) && ReferenceGenome.getReference(name) != rg)
```
tpoterba, line 605: delete?

hail/src/main/scala/is/hail/variant/ReferenceGenome.scala
```diff
@@ -271,7 +269,7 @@ case class ReferenceGenome(name: String, contigs: Array[String], lengths: Map[St
       }
     }
 
-    if (!Interval.isValid(locusType.ordering, start, end, includesStart, includesEnd))
+    if (!Interval.isValid(locusType.ordering(HailStateManager(Map(this.name -> this))), start, end, includesStart, includesEnd))
```
tpoterba, line 272: don't love having to allocate a bunch of stuff here for this check (currently it's creating the ordering too, but once the Type.scala change is made that will return the cached ordering). How about we make `mkOrdering(rg)` / `ordering(rg)` methods on TLocus mirroring mkOrdering(sm) and ordering(sm), which `mkOrdering(sm)` calls through to after looking up the rg?

tpoterba (changes requested):
This is such a great change. Some comments about adding back ordering caching to avoid performance regressions, but overall a super straightforward read.

hail/python/hail/backend/local_backend.py
daniel-goldstein: Not sure if this is intentional, but this differs slightly from the spark backend. It was previously in py4j_backend and overridden in the spark_backend so I instead just moved this to local_backend. The difference is that spark_backend asserts the ir does not use randomness and does not call `finalize_randomness`. Happy to change if these should not in fact be different.

hail/src/main/scala/is/hail/variant/ReferenceGenome.scala
daniel-goldstein: Hm, I agree and it feels silly that we're constructing a map with one element just to take that one element out of it. But also a little funny that we're calling locusType.ordering(rg) that will just turn around and call rg.locusOrdering. I tried just moving the implementation over to ReferenceGenome so that instead of having 1 indirection in TLocus and 2 indirections in ReferenceGenome we have no indirections in ReferenceGenome and still 1 indirection in TLocus.

hail/python/hail/backend/local_backend.py
tpoterba: @patrick-schultz can you comment here in case you know so I don't have to dig in?

hail/src/main/scala/is/hail/variant/ReferenceGenome.scala
tpoterba: oh, yeah, great.

tpoterba (approved):

hail/python/hail/backend/local_backend.py
patrick-schultz: I think an easy fix here is to handle randomness in `define_function` by adding an extra parameter `__rng_state` (if the body uses randomness, but it doesn't hurt to do it all the time), then making the returned `f` pass in `Ref(__rng_state)` for the parameter, to capture the state from the caller's environment. Then `register_ir_function` should not call `finalize_randomness`.

hail/src/main/scala/is/hail/rvd/RVDPartitioner.scala
```diff
@@ -10,6 +11,7 @@ import org.apache.spark.sql.Row
 import org.apache.spark.{Partitioner, SparkContext}
 
 class RVDPartitioner(
+  val sm: HailStateManager,
```
ehigham, line 14: Got a friendly build failure from this :) Can you explain why partitioners need to know about reference genomes?

PR #12701, opened 2023-02-15
[query] Upgrade spark to 3.3.0 and dataproc to 2.1

daniel-goldstein:
CHANGELOG: Query on Spark now officially supports Spark 3.3.0 and Dataproc 2.1.x

Tested on dataproc via `make -C hail test-dataproc-37`. Updating the dependencies introduced a few new linting checks that I fixed here. Updating pyspark necessitated a couple of changes, namely a different py4j jar and they removed `SparkSession._wrapped` (but maybe we didn't need that anyway? not sure). Most importantly, the newer spark version brings with it a newer jackson version which is sufficient for the azure-storage-blob dependency, meaning we don't need to build against two different spark versions for spark and batch.

### DIFF
--- a/hail/python/hailtop/hailctl/auth/create_user.py
+++ b/hail/python/hailtop/hailctl/auth/create_user.py
@@ -4,6 +4,10 @@ from hailtop.utils import sleep_and_backoff
 from hailtop.auth import async_create_user, async_get_user
 
 
+class CreateUserException(Exception):
+    pass
+
+
 def init_parser(parser):
     parser.add_argument("username", type=str,
                         help="User name to create.")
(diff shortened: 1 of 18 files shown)


hail/python/hailtop/hailctl/auth/create_user.py
```diff
@@ -20,25 +20,22 @@ def init_parser(parser):
 
 
 async def async_main(args):
```
danking, line 22: why make these changes? Do we report the username in the underlying exception?

hail/python/requirements.txt
```diff
@@ -15,6 +15,6 @@ parsimonious<0.9
 plotly>=5.5.0,<5.11
 protobuf==3.20.2
 PyJWT
-pyspark>=3.1.1,<3.2.0
+pyspark==3.3.0
```
danking, line 18: Seems reasonable to do `pyspark>=3.3,<3.4`?

danking (changes requested):

hail/python/hailtop/hailctl/auth/create_user.py
daniel-goldstein: The linter complained about using too generic an error class, it wants us to subclass `Exception`. I removed the try except because I felt like it wasn't actually any more helpful than whatever stack trace was underneath, but I didn't think about the username not being in there. Can add that back in if you want.

hail/python/requirements.txt
daniel-goldstein: I guess I was distrusting of pyspark and wanted the exact same version that there is in dataproc, but maybe that is unfounded.

danking (approved):

PR #12702, opened 2023-02-16
[query] use a higher default p-value accuracy for _linear_skat

danking:
Konrad requested that we use a higher default. Above this some of the test case p-values do not converge


danking:
I don't think the Davies method can actually provide such high accuracy. Even after 100k iterations it fails to converge.

PR #12705, opened 2023-02-16
[query] Don't persist in table._filter_partitions

tpoterba:


tpoterba:
cc @patrick-schultz -- easy fix.

hail/python/hail/table.py
```diff
@@ -3750,7 +3750,7 @@ def summarize(self, handler=None):
 
     @typecheck_method(parts=sequenceof(int), keep=bool)
     def _filter_partitions(self, parts, keep=True) -> 'Table':
-        return Table(ir.TableToTableApply(self._tir, {'name': 'TableFilterPartitions', 'parts': parts, 'keep': keep})).persist()
+        return Table(ir.TableToTableApply(self._tir, {'name': 'TableFilterPartitions', 'parts': parts, 'keep': keep}))
```
danking, line 3753: this seems crazy?

hail/python/hail/ir/table_ir.py
```diff
@@ -934,8 +934,9 @@ def __init__(self, child, config):
 
     def _handle_randomness(self, uid_field_name):
         if uid_field_name is not None:
-            raise FatalError('TableToTableApply does not support randomness in consumers')
-        child = self.child.handle_randomness(None)
+            if self.config['name'] != 'TableFilterPartitions':
+                raise FatalError(f'TableToTableApply({self.config["name"]}) does not support randomness in consumers')
+        child = self.child.handle_randomness(uid_field_name)
```
danking, line 939: why did we switch to using uid_field_name from None?

hail/python/hail/ir/table_ir.py
```diff
@@ -934,8 +934,9 @@ def __init__(self, child, config):
 
     def _handle_randomness(self, uid_field_name):
         if uid_field_name is not None:
-            raise FatalError('TableToTableApply does not support randomness in consumers')
-        child = self.child.handle_randomness(None)
+            if self.config['name'] != 'TableFilterPartitions':
```
danking, line 937: why doers table filter partitions support random consumers?

danking (changes requested):

hail/python/hail/table.py
tpoterba: the persist was crazy, yes. No persist, no crazy.

hail/python/hail/ir/table_ir.py
tpoterba: TableFilterPartitions is the only TableToTableApply where we do support passing through random UIDs. Others trigger an error here.

hail/python/hail/ir/table_ir.py
tpoterba: why not? super easy to pass through row UIDs.

patrick-schultz:
This looks right to me. I just added persists after all TableToTableApply nodes, so they didn't have to implement uid handling. But Tim is right, FilterPartitions is trivially correct, because it doesn't modify rows.

danking (approved):

PR #12710, opened 2023-02-16
[batch] Cleanup old billing tables

jigold:
This is part 1 in mitigating the extra rows in the billing tables with redundant resources with the same prices.

I'm not sure how long it takes to drop a table of this size. Would prefer to merge in the morning in case there are any issues.

### DIFF
--- a/batch/sql/cleanup-old-billing-tables.sql
+++ b/batch/sql/cleanup-old-billing-tables.sql
@@ -1,0 +1,127 @@
+DELIMITER $$
+
+DROP TRIGGER IF EXISTS attempts_after_update $$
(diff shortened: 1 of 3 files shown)


daniel-goldstein (approved):
LGTM. WIP is on.

jigold:
@danking Can you confirm whether we can just drop the table in one shot or do we need to use truncate or delete rows in chunks first? I'm just worried about performance here and I can't find a solid answer on whether this operation is instant or could take days.

jigold:
Tested deletion in a separate database. 0 rows is instantaneous. 100,000,000 rows is 10 seconds. So at worst, it's probably on order of 10 minutes to drop a table since the table I had was very simple with not a lot of columns. Will rebase and then get this merged.

PR #12712, opened 2023-02-17
[batch] quiet down every kind of transient error

danking:


danking:
See for example https://cloudlogging.app.goo.gl/ziaRD9HKxxca8Nd3A

in which ~15 MJCs have to retry because of `ServerDisconnectedError` or `TimeoutError`. With this PR, I think we would have seen just the three "two errors observed" warning messages.

Here's a possible extension to this PR that fuses the thinking of both PRs (this one and #12505): use the total delay instead of `errors = 2`. We retry really quickly, so two errors could occur in ~500ms which really isn't enough time for batch driver to fix itself.

PR #12716, opened 2023-02-17
[ci] Run the local backend tests much more in parallel

daniel-goldstein:
The local backend tests only use 1 core, so this dedicates the same resources for much more parallelism.


danking (approved):

danking:
+1-4

Net 3 lines for a 1/2 reduction in test time is pretty good 😉

PR #12723, opened 2023-02-23
[batch] Add deduped id column to resources table

jigold:


daniel-goldstein (changes requested):
I'm a little confused, could you update the PR body with a description of what's getting updated here? How come the filter on `resource_id` > last_resource_id? Is `deduped_resource_id = NULL` not enough of a filter?

daniel-goldstein (approved):
Thanks for the explanation, makes more sense to me now. Looks good!

PR #12725, opened 2023-02-23
[compiler] Add TableGen TableIR Node

ehigham:
Context from Tim's scribblings:
"""
Why are we adding this node?
The hl.balding_nichols_model node is implemented as
>>> ht = hl.utils.range_table(N)
>>> ht = ht.key_by(locus = hl.locus('chr1', ht.idx + 1)) // or something

This requires us to do an extra pass over the data in order to key `ht` by this new key 'locus' we track information about keys statically -- need to know that tables are sorted by their key we also need to know partition intervals statically. So, when we implement key_by, we have to do extra work to compute this information before generating the rest of the query.

This is a super common tutorial example and it's currently slow because it an extra pass over the data! We need to implmement balding_nichols_model without a key_by, which means we need to inject the information we need to know about the key into the IR directly. This means passing the key that's sorted and the partition intervals.
"""

This change just implements the new TableIR node in Scala. Exposure to Python will come in a subsequent change.


hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
```diff
@@ -321,6 +321,9 @@ object PruneDeadFields {
     BindingEnv(e, Some(e), Some(e))
   }
 
+  def uses(name: String, env: Env[BoxedArrayBuilder[Type]]): Array[Type] =
+    env.lookupOption(name).map(_.result()).getOrElse(Array.empty)
+
```
ehigham, line 326: This pattern was used a lot in this file. I took the oportunity to define this function and substitute it in. It's made the code cleaner at the expense of the review. Plz forgive.

hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
```diff
@@ -344,6 +347,16 @@ object PruneDeadFields {
...
+      case TableGen(contexts, globals, cname, gname, body, _, _) =>
+        val bodyEnv = memoizeValueIR(ctx, body, TStream(requestedType.rowType), memo)
+        // Contexts are only used in the body so we only need to keep the fields used therein
+        val contextsElemType = unifySeq(TIterable.elementType(contexts.typ), uses(cname, bodyEnv.eval))
+        // Globals are exported and used in body, so keep the union of the used fields
+        val globalsType = unifySeq(globals.typ, uses(gname, bodyEnv.eval) :+ requestedType.globalType)
+        memoizeValueIR(ctx, contexts, TStream(contextsElemType), memo)
+        memoizeValueIR(ctx, globals, globalsType, memo)
+
```
ehigham, line 359: Drawing attention here

hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
```diff
@@ -246,9 +246,9 @@ object PruneDeadFields {
               }
               TTuple(subFields)
             case ta: TArray =>
-              TArray(unifySeq(ta.elementType, children.map(_.asInstanceOf[TArray].elementType)))
+              TArray(unifySeq(ta.elementType, children.map(TIterable.elementType)))
```
ehigham, line 249: Added `TIterable.elementType` because I got bored of writing `.asInstanceOf[TIterable].elementType`. This generated some noise but has made things slighly more concise. Sorry.

hail/src/main/scala/is/hail/expr/ir/PruneDeadFields.scala
```diff
@@ -1589,6 +1537,21 @@ object PruneDeadFields {
...
+        val bodyEnv = Env(cname -> TIterable.elementType(newContexts.typ), gname -> newGlobals.typ)
+        TableGen(
+          contexts = newContexts,
+          globals = newGlobals,
+          cname = cname,
+          gname = gname,
+          body = rebuildIR(ctx, body, BindingEnv(bodyEnv), memo),
+          partitioner.coarsen(requestedType.key.length),
+          errorId
+        )
```
ehigham, line 1553: Drawing attention to this

hail/src/main/scala/is/hail/types/virtual/TIterable.scala
```diff
@@ -1,10 +1,21 @@
...
 import is.hail.utils.FastSeq
 
 abstract class TIterable extends Type {
   def elementType: Type
 
   override def children = FastSeq(elementType)
 }
+
+object TIterable {
+  def elementType: BaseType => Type = {
```
tpoterba, line 13: This is a prime example of what I call Scala BS: something the language makes easy to write that seems fine, but when you think and look hard is horrifying. As an exercise for you, investigate why this will have worse performance than the method I've written below (it's not the match):
```
  def elementType(t: Type): Type = t.asInstanceOf[TIterable].elementType
```

tpoterba (dismissed):

tpoterba:
Excited to use this!

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -1794,6 +1791,65 @@ case class TableKeyBy(child: TableIR, keys: IndexedSeq[String], isSorted: Boolea
...
+ */
+case class TableGen(contexts: IR,
+                    globals: IR,
+                    cname: String,
+                    gname: String,
+                    body: IR,
+                    partitioner: RVDPartitioner,
+                    errorId: Int = ErrorIDs.NO_ERROR
+                   ) extends TableIR {
+  private def typeCheck[A <: Type](varname: String, typ: Type)(implicit tag: ClassTag[A]): Option[A] =
```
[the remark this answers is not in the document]
ehigham: I disagree. I hate the errors from class cast exceptions (along with pattern match failures), especially when these happen in a delayed execution context (like `StackFrame`). More information is always better.

hail/src/main/scala/is/hail/types/virtual/TIterable.scala
ehigham: Ok, though I really don't think we should be worrying about making micro-optimisations like this or a few extra allocs. If it's a problem and a profiler says it's a hotspot, then we should care, but not before.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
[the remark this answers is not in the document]
ehigham: The options are also allocated from the cast. I'm not sure how `asInstanceOf` is implemented, but a pattern match generates an `Option` (via `ClassTag.unapply`) anyway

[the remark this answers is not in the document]
ehigham: Rewritten without allocs.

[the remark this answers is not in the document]
tpoterba: What's the additional information here? I'm contending that this is a much longer way of writing an exception that doesn't offer any more context than the stack trace and mismatched classes in the default classcastexception will give.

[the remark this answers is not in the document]
ehigham: The assertion contains the name of the offending argument. This allows me to test that the right assertion has been encountered.
Following our discussion, I tried moving this into `TypeCheck` but the problem is the `typ` is requested when computing the bindings used by the type checker.
I want type errors to arise in a predicable place in a manner that's easily testable so I'm going to keep a slimmed down version of this so that `TableGen` will be type correct by construction.

hail/src/test/scala/is/hail/types/virtual/TStructSuite.scala
```diff
@@ -0,0 +1,29 @@
+package is.hail.types.virtual
+
+import is.hail.HailSuite
+import org.testng.Assert.{assertFalse, assertTrue}
+import org.testng.annotations.Test
+
+class TStructSuite extends HailSuite {
+  @Test
+  def testEmptyIsPrefixEmpty(): Unit =
+    assertTrue(TStruct().isPrefix(TStruct()), "empty")
```
ehigham, line 10: I didn't realise this was already implemented in `TBaseStruct`.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -1794,6 +1791,60 @@ case class TableKeyBy(child: TableIR, keys: IndexedSeq[String], isSorted: Boolea
...
+
+  private val globalType =
+    TypeCheck.requireInstance[TStruct]("globals", globals.typ)
+
+  private val rowType = {
+    val bodyType = TypeCheck.requireInstance[TStream]( "body", body.typ)
+    TypeCheck.requireInstance[TStruct]( "body.elementType", bodyType.elementType)
+  }
+
+  if (!partitioner.kType.isPrefix(rowType))
```
ehigham, line 1826: this is wrong, we need them to be subsets, not prefixes.

hail/src/test/scala/is/hail/types/virtual/TStructSuite.scala
```diff
@@ -0,0 +1,44 @@
...
+      // isPrefixOf ignores field names and compares the ordered sequence of types.
+      // Consider joins for example - we only care that the key fields have the same types
+      // so we compare the key types (which are structs) for equality ignoring field names.
+      // isPrefixOf is used in similar cases involving key types where we don't care about names.
+      Array(TStruct("b" -> TVoid), TStruct("a" -> TVoid, "b" -> TVoid), true)
+    )
+
+  @Test(dataProvider = "isPrefixOf")
+  def testIsPrefixOf(a: TStruct, b: TStruct, isPrefix: Boolean): Unit =
+    assert(a.isPrefixOf(b) == isPrefix, s"expected $a `isPrefixOf` $b == $isPrefix")
```
ehigham, line 26: I wasn't sure why we ignore field names but after Patrick's explaination (comment is lifted from a DM with him), I wanted to write this. Though trivial, the intent is to lock down behaviour for posterity that's decoupled from other logic.

hail/src/main/scala/is/hail/types/virtual/TIterable.scala
ehigham: For posterity, using a lambda generates an additional load and `invokedynamic`: https://godbolt.org/z/sx6TfKbeG

hail/src/main/scala/is/hail/types/virtual/TStruct.scala
```diff
@@ -387,4 +387,9 @@ final case class TStruct(fields: IndexedSeq[Field]) extends TBaseStruct {
       Row.fromSeq(subsetFields.map { case (i, subset) => subset(r.get(i)) })
     }
   }
+
+  def isSubsetOf(other: TStruct): Boolean = {
+    val dropIndex = (f: Field) => (f.name, f.typ)
```
tpoterba, line 392: Structs can get quite large, worth making this linear in the smaller struct:
```scala
  def isSubsetOf(other: TStruct): Boolean = {
    fields.forall(f => other.hasField(f.name) && other.field(f.name).typ ==f.typ )
  }
```

hail/src/test/scala/is/hail/types/virtual/TStructSuite.scala
tpoterba: that's weird behavior, we could probably rewrite all usages of isPrefixOf to do something clearer. Good to add a test, though.

tpoterba (dismissed):
2 more style/perf comments and good to go

hail/src/main/scala/is/hail/expr/ir/TypeCheck.scala
```diff
@@ -559,4 +561,13 @@ object TypeCheck {
       case _: BlockMatrixIR =>
     }
   }
+
+  def requireInstance[A <: Type](argname: String, typ: Type)(implicit tag: ClassTag[A]): A =
```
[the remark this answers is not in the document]
ehigham: You need the classtag to cast like that too: https://godbolt.org/z/YjPr77G77.
Happy to rename.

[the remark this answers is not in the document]
tpoterba: This works:

https://godbolt.org/z/har4dMdGj

tpoterba (approved):

hail/src/main/scala/is/hail/expr/ir/TypeCheck.scala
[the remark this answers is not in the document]
tpoterba: I'm also now super confused about what this is doing. The `A` is erased in my example above. It's casting A to a Type, though it already is a type? 

Anyway, approved. Can revisit this later if we care and find a better way.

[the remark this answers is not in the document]
ehigham: I think when you write constraints in the type signature, evidence for that is passed implicitly when the signature is desugared. I remember reading about this in cats documentation, but can't seem to find that now

PR #12726, opened 2023-02-23
[batch] Allow python jobs to have file resources too

daniel-goldstein:
I feel like `ofile` semantics should apply the same to bash and python jobs. I don't believe this is a breaking change since `j.ofile` didn't really mean anything usable until now for python jobs IIUC.

### DIFF
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -109,7 +109,12 @@ class Job:
         self._dirname = f'{safe_str(name)}-{self._token}' if name else self._token
 
     def _get_resource(self, item: str) -> '_resource.Resource':
-        raise NotImplementedError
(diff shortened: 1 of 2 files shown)


jigold:
I think I'm good with this change, but I just want to make sure my reasoning is the same as yours. In our "language" `j.ofile` is implied to be a temporary automatically localized / delocalized file. Doesn't matter whether it's from bash, c++, R, Python that wants to write / read from that file. I don't think there is anything special about PythonJobs that we'd want to reserve the `__getitem__` semantics for. Does this seem right?

jigold (approved):
Nice change!

jigold:
Can you add a CHANGELOG comment somewhere?

danking:
Ah this is really nice, I have wanted this for a while.

PR #12729, opened 2023-02-23
[batch] Add JVM async profiler capabilities

jigold:


batch/batch/file_store.py
```diff
@@ -39,6 +39,11 @@ def resource_usage_path(self, format_version, batch_id, job_id, attempt_id, task
             return f'{self.batch_log_dir(batch_id)}/{job_id}/{task}/resource_usage'
         return f'{self.batch_log_dir(batch_id)}/{job_id}/{attempt_id}/{task}/resource_usage'
 
+    def jvm_profile_path(self, format_version, batch_id, job_id, attempt_id, task):
+        if not format_version.has_attempt_in_log_path():
+            return f'{self.batch_log_dir(batch_id)}/{job_id}/{task}/jvm_profile'
```
daniel-goldstein, line 44: Suggestion: Could we end this in`.html` so that if someone wants to download from GCS it's more easily auto-opened in a browser?

batch/batch/worker/worker.py
```diff
@@ -2044,6 +2053,21 @@ async def cleanup(self):
...
+                else:
+                    profile_contents = b'profile not available'
+                await self.worker.file_store.write_jvm_profile(
+                    self.format_version,
+                    self.batch_id,
+                    self.job_id,
+                    self.attempt_id,
+                    'main',
+                    profile_contents,
+                )
```
daniel-goldstein, line 2069: Writing 'profile not available' feels a little wrong to me. I feel like if there's a profile file in GCS I expect it to have valid profiling data in it. What message is this trying to convey? Should we be logging an error or warning in this case?

jigold:
Note to self, there's a backwards compatibility issue with the validator.

danking:
@tpoterba bump

tpoterba:
Ah, sorry, looked good to me last time I read through but now there are a couple conflicts. @jigold if you rebase I'll look and approve immediately

tpoterba (approved):

PR #12733, opened 2023-02-24
[query] substantially speed up VCF writing

danking:
CHANGELOG: `hl.export_vcf` is now substantially faster on simple pipelines.

I fell into a bit of a rabbit hole. I started off using the balding nichols model as my test dataset. I improved random number generation a bit by making `rand_unif` into a method, but eventually bailed out to focus on other improvements. I also originally used `vcf.bgz` but bgzip is quite slow and it obscured the slow parts of what I could control, i.e. hail.

I finally settled on this test case run against the Spark backend:
```
import hail as hl
hl.init(master='local[1]')
hl._set_flags(write_ir_files='1')
mt = hl.utils.range_matrix_table(n_rows=1000_000, n_cols=4_000)
mt = mt.key_cols_by(s = hl.str(mt.col_idx))
mt = mt.key_rows_by(locus = hl.locus("1", mt.row_idx + 1), alleles = ['G', 'T'])
mt = mt.annotate_entries(GT = hl.call(mt.row_idx % 2, mt.col_idx % 2))
hl.export_vcf(mt, '/tmp/foo.vcf')
```
It generates a 15GiB file. My initial tests, which used the balding nichols model, had write times of ~8MiB/s. With all my changes, I once saw 177 MIB/s but I think that may have been a fluke. I see pretty consistent ~110MiB/s in the profiler's estimate of bandwidth to the FileOutputStream.

When measured by `time python3 test.py` this script writes at ~93MiB/s. Ideally we would hit 250MiB/s (1/8th of an n1-standard-8's network bandwidth), but, considering that we have to split that bandwidth with reading in most cases, ~91 MiB/s ain't so bad. On main, this pipeline writes at 32 MiB/s.

The wins in decreasing order of importance were:
1. Use buffered I/O. All of our exporters should now use buffered I/O because I changed it in the EmitMethodBuilder. I didn't change it in HadoopFS because (a) Hail's native I/O has buffering and (b) buffering and position tracking requires work.
2. Avoid String allocation, String to UTF8 conversion, and Array[Byte] allocation in VCF writing. In particular, for the most common types of Calls, I just return the UTF8 byte array in a switch statement.
3. Use a fast path for diploid genotypes. In the worst case, we did 5 branches and now we do 2 which should be well predicted.
4. Remove an allocation of a lambda in a hot method in Genotype.scala.

Future Work:
1. The randomness stuff still has a lot of low hanging fruit. NumPy can generate random numbers at bandwidths far above what we're managing here.
2. For VCFs with more entry fields, we should not write ints and floats by generating strings and getting their UTF8 encoding
3. Invert the writing control flow: serialization methods should take an OutputStream and write bytes directly into it. Contrast that with passing around arrays which are memcopied into a buffer.

---

Range table test:
```
import hail as hl
hl.init(master='local[1]')
hl._set_flags(write_ir_files='1')
mt = hl.utils.range_matrix_table(n_rows=1000_000, n_cols=4_000)
mt = mt.key_cols_by(s = hl.str(mt.col_idx))
mt = mt.key_rows_by(locus = hl.locus("1", mt.row_idx + 1), alleles = ['G', 'T'])
mt = mt.annotate_entries(GT = hl.call(mt.row_idx % 2, mt.col_idx % 2))
hl.export_vcf(mt, '/tmp/foo.vcf')
```

Balding Nichols Test:
```
import hail as hl
hl.init(master='local[1]')
hl._set_flags(write_ir_files='1')
mt = hl.balding_nichols_model(1, n_variants=100_000, n_samples=4_000)
mt = mt.key_cols_by(s = hl.str(mt.sample_idx))
hl.export_vcf(mt, '/tmp/foo.vcf')
```

Bandwidth is: `time python3 file.py` / `ls -l /tmp/foo.vcf`

| code | test  | size    | bandwidth (MiB/s) |
| ---- | ----- | ------- | ----------------- |
| main | range | 1Mx4K   | 32                |
| this | range | 1Mx4K   | 93 (~3x)          |
| main | BN    | 100Kx4K | 19                |
| this | BN    | 100Kx4K | 28 (~1.5x)        |

### DIFF
--- a/hail/src/main/java/is/hail/expr/ir/MatrixWriterConstants.java
+++ b/hail/src/main/java/is/hail/expr/ir/MatrixWriterConstants.java
@@ -1,0 +1,7 @@
+package is.hail.expr.ir;
+
+public class MatrixWriterConstants {
+  public static final byte[] dotUTF8 = ".".getBytes();
+  public static final byte[] missingUnphasedDiploidGTUTF8 = "./.".getBytes();
+  public static final byte[] passUTF8 = "PASS".getBytes();
+}
--- a/hail/src/main/scala/is/hail/expr/ir/Compile.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Compile.scala
@@ -72,7 +72,8 @@ object Compile {
 
       val emitContext = EmitContext.analyze(ctx, ir)
       val rt = Emit(emitContext, ir, fb, expectedCodeReturnType, params.length)
-      CompiledFunction(rt, fb.resultWithIndex(writeIRs, print))
+      val writeIRs2 = writeIRs || ctx.getFlag("write_ir_files") != null
+      CompiledFunction(rt, fb.resultWithIndex(writeIRs2, print))
     }).tuple
   }
 }
--- a/hail/src/main/scala/is/hail/expr/ir/Compile.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Compile.scala
@@ -72,8 +72,7 @@ object Compile {
 
       val emitContext = EmitContext.analyze(ctx, ir)
       val rt = Emit(emitContext, ir, fb, expectedCodeReturnType, params.length)
-      val writeIRs2 = writeIRs || ctx.getFlag("write_ir_files") != null
-      CompiledFunction(rt, fb.resultWithIndex(writeIRs2, print))
+      CompiledFunction(rt, fb.resultWithIndex(writeIRs, print))
     }).tuple
   }
 }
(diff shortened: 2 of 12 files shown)


danking:
Heh. I checked the range test with `vcf.bgz` and it's embarrassingly slow. In terms of compressed output bytes, it's 251 KiB/s. In terms of decompressed output bytes it's 80 MiB/s. I guess this makes sense though: we can't generate data fast enough to keep up with bgzip. I seem to recall bgzip writes at ~50 MiB/s. At those compression ratios we'd need to generate data at ~15GiB/s, which we are _so_ far from doing.

danking:
The profiler results are hard to interpret though. It seems like we're dominated by region manipulation.

Screen Shot 2023-02-24 at 17 27 08

tpoterba (approved):
🐎 💨

PR #12734, opened 2023-02-27
[batch] Error if PythonJob functions are called with incompatible signature

daniel-goldstein:
CHANGELOG: `PythonJob.call` now immediately errors when supplied arguments are incompatible with the called function instead of erroring only when the job is run.

inspect.Signature allows you to inspect the signature of some function, and [inspect.Signature.bind](https://docs.python.org/3/library/inspect.html#inspect.Signature.binds) takes arguments to the function and maps those arguments to method parameters. If the invocation is invalid, i.e. would raise a TypeError when *actually* invoking the function with those arguments, this method raises a similar error. This can help a user immediately catch something like not passing the correct number of arguments to a PythonJob call without having to submit and run it.

### DIFF
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -1077,8 +1077,13 @@ class PythonJob(Job):
             if isinstance(value, Job):
                 raise BatchException('arguments to a PythonJob cannot be other job objects.')
 
+        # Some builtins like `print` do not have signatures
         try:
             inspect.signature(unapplied).bind(*args, **kwargs)
+        except ValueError as e:
+            # Some builtins like `print` don't have a signature that inspect can read
+            if not 'no signature found for builtin' in e.args[0]:
+                raise e
         except TypeError as e:
             raise BatchException(f'Cannot call {unapplied.__name__} with the supplied arguments') from e
 
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -1082,7 +1082,7 @@ class PythonJob(Job):
             inspect.signature(unapplied).bind(*args, **kwargs)
         except ValueError as e:
             # Some builtins like `print` don't have a signature that inspect can read
-            if not 'no signature found for builtin' in e.args[0]:
+            if 'no signature found for builtin' not in e.args[0]:
                 raise e
         except TypeError as e:
             raise BatchException(f'Cannot call {unapplied.__name__} with the supplied arguments') from e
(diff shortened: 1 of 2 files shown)


hail/python/hailtop/batch/job.py
```diff
@@ -1077,6 +1077,16 @@ def run_async(*args, **kwargs):
...
 
+        # Some builtins like `print` do not have signatures
+        try:
+            inspect.signature(unapplied).bind(*args, **kwargs)
+        except ValueError as e:
+            # Some builtins like `print` don't have a signature that inspect can read
+            if 'no signature found for builtin' not in e.args[0]:
+                raise e
+        except TypeError as e:
+            raise BatchException(f'Cannot call {unapplied.__name__} with the supplied arguments') from e
```
jigold, line 1088: Can we add what the expected signature was and what was passed in the error message? If not, no worries.

jigold (dismissed):

hail/python/hailtop/batch/job.py
daniel-goldstein: The `TypeError` that I raise from here has that information, basically the same error that you'd get if you had just called the function. Here's an example:

```ipython
In [1]: import hailtop.batch as hb
b =
In [2]: b = hb.Batch()

In [3]: j = b.new_python_job()

In [4]: def foo(a, b):
   ...:     a + b
   ...:

In [5]: j.call(foo, 1)
TypeError                                 Traceback (most recent call last)
~/hail/hail/python/hailtop/batch/job.py in call(self, unapplied, *args, **kwargs)
   1081         try:
-> 1082             inspect.signature(unapplied).bind(*args, **kwargs)
   1083         except ValueError as e:

~/opt/miniconda3/envs/hail/lib/python3.7/inspect.py in bind(*args, **kwargs)
   3014         """
-> 3015         return args[0]._bind(args[1:], kwargs)
   3016

~/opt/miniconda3/envs/hail/lib/python3.7/inspect.py in _bind(self, args, kwargs, partial)
   2929                             msg = msg.format(arg=param.name)
-> 2930                             raise TypeError(msg) from None
   2931             else:

TypeError: missing a required argument: 'b'

The above exception was the direct cause of the following exception:

BatchException                            Traceback (most recent call last)
<ipython-input-5-63e0907b869d> in <module>
----> 1 j.call(foo, 1)

~/hail/hail/python/hailtop/batch/job.py in call(self, unapplied, *args, **kwargs)
   1086                 raise e
   1087         except TypeError as e:
-> 1088             raise BatchException(f'Cannot call {unapplied.__name__} with the supplied arguments') from e
   1089
   1090         def handle_arg(r):

BatchException: Cannot call foo with the supplied arguments
```

daniel-goldstein: Happy to put additional info in if that's not enough.

jigold (approved):

PR #12735, opened 2023-02-27
[query] Add to_pandas(types={}) argument to specify user-supplied pandas dtypes

### LINKED ISSUE #11738: hl.Table.to_pandas() generates a dtype=string dataframe which is still experimental
I encountered an error in an external package when I used a Hail-generated pandas data frame, which is due to an unsupported dtype [`pandas.StringDtype`](https://pandas.pydata.org/docs/reference/api/pandas.StringDtype.html#pandas.StringDtype).
https://github.com/biocore-ntnu/pyranges/pull/264

Given it's still experimental in pandas, can we have an option to generate a data frame that have `dtype=object` string columns? or maybe, we should make `dtype=object` default.
https://github.com/hail-is/hail/blob/c4b09953f62cea090c8ab2026bc81851b9f4d64a/hail/python/hail/table.py#L3345-L3346

mkanai:
This PR fixes #11738. Now users can specify arbitrary type conversation between Hail and Pandas dtypes via:
```
ht.to_pandas(types={"col1": "int32", "col2": np.float64, hl.tstring: "object"})
```
This maps `col1` and `col2` to `int32` and `np.float64`, respectively, and all `hl.tstring` fields to `object`.

One design question might be whether to have separate arguments for column name and Hail type specifications or not. Any thoughts? cc: @danking 

Also, I don't think the current type check would work for `np.float64`-like numpy dtype specifications...

### DIFF
--- a/hail/python/hail/table.py
+++ b/hail/python/hail/table.py
@@ -3475,15 +3475,18 @@ class Table(ExprContainer):
         """
         return Env.spark_backend('to_spark').to_spark(self, flatten)
 
-    @typecheck_method(flatten=bool)
-    def to_pandas(self, flatten=True):
+    @typecheck_method(flatten=bool, types=dictof(oneof(str, hail_type), str))
+    def to_pandas(self, flatten=True, types={}):
         """Converts this table to a Pandas DataFrame.
 
         Parameters
         ----------
         flatten : :obj:`bool`
             If ``True``, :meth:`flatten` before converting to Pandas DataFrame.
-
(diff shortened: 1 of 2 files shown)


danking:
Amazing! I'll look into this this week. Thank you Masa!

danking:
@patrick-schultz I added some tests and fixed behavior as a result. I think this is a valuable change, but since I've edited it directly seems like someone else should do review as well.

patrick-schultz (approved):
Looks good to me!

PR #12736, opened 2023-02-27
[qob] Support loading RGs from FASTA files

daniel-goldstein:
I left the changes to Query and Batch in separate commits for ease of review. I put these in the same PR because we don't really have standalone testing for JVM Jobs outside of Query-on-Batch so the FASTA use-case serves as a test here that cloudfuse is working properly for JVM Jobs. Would be great if Jackie you could review the batch commit and Tim could review the query commit.

## Hail Query
- Added support for the `FROM_FASTA_FILE` rpc and the service backend now passes sequence file information from RGs in every rpc
- Refactored the liftover handling in service_backend to not redundantly store liftover maps and just take them from the ReferenceGenome objects like I did for sequence files. This means that add/remove liftover/sequence functions on the Backend are just intended to sync up the backend with python, which is a no-op for the service backend.
- Don't localize the index file on fromFASTAFile/addSequence before creating the index object. `FastaSequenceIndex` just loads the whole file on construction so might as well stream it in from whatever storage it's in.
- FASTA caching is left alone because those files will be mounted and unmounted from the jvm container over the life of the job. JVM doesn't have to worry about disk usage because that's handled by Batch XFS quotas, so long as the service backend requests enough storage to fit the FASTA file. Batch will make sure that a given bucket (and therefore a given FASTA file) is mounted once per-user on a batch worker.

## Hail Batch
- Added support for read-only cloudfuse mounts for JVM jobs
- These mounts are shared between jobs on the same machine from the same user
- I did not change DockerJobs, but they could be very easily adapted to use this new mount-sharing code.


batch/batch/worker/worker.py
```diff
@@ -1019,6 +1082,7 @@ async def container_config(self):
                 },
             },
             'linux': {
+                'rootfsPropagation': 'slave',
```
jigold, line 1085: Why is this not `rslave`?

batch/batch/worker/worker.py
```diff
@@ -366,6 +367,68 @@ async def _free(self, netns: NetworkNamespace):
...
+
+    async def _bind_mount(self, src, dst):
+        await check_exec_output('mount', '--bind', src, dst)
+
+    async def _bind_unmount(self, dst):
+        await check_exec_output('umount', dst)
+
+    def _new_path(self):
+        path = f'{self.cloudfuse_dir}/{uuid.uuid4().hex}'
+        os.makedirs(path)
```
jigold, line 428: Does this need to be `blocking_to_async`? I think we use `check_shell` in other places in the worker.

batch/batch/worker/worker.py
```diff
@@ -366,6 +367,68 @@ async def _free(self, netns: NetworkNamespace):
...
+class FuseMount:
+    def __init__(self, path):
+        self.path = path
+        self.bind_mounts = set()
+
+
+# Mounts that can be shared across jobs by the same user
+# Only sharing within jobs of the same user ensures that
+# the user is authorized to access the bucket
+class ReadOnlyCloudfuseManager:
```
jigold, line 379: Can we add a comment about the credentials sharing and why this is okay?

batch/batch/worker/worker.py
```diff
@@ -366,6 +367,68 @@ async def _free(self, netns: NetworkNamespace):
...
+            tmp_path,
+            config,
+        )
+
+    async def _fuse_unmount(self, path: str):
+        assert CLOUD_WORKER_API
+        await CLOUD_WORKER_API.unmount_cloudfuse(path)
+
+    async def _bind_mount(self, src, dst):
+        await check_exec_output('mount', '--bind', src, dst)
```
jigold, line 421: Just to double check I understand why there is now an additional `mount` and `umount`, this is because we have the additional level of nesting where as with the regular DockerJob, the mounts were handled by `crun` when it created the container?

jigold (dismissed):
This looks great! Will give it another look on Monday.

batch/batch/worker/worker.py
daniel-goldstein: While it probably wouldn't hurt anything, we don't need `rslave` so I went with the more restrictive option. `slave` says that any mount points beneath the root mount will be propagated into the container, but only one level of mounts deep. `rslave` means this is applied recursively to mounts underneath mounts, e.g. if we were to bind mount an additional mount inside the directory tree of one of our fuse mounts, but we don't do this.

batch/batch/worker/worker.py
daniel-goldstein: Yes exactly! On `JVMContainer` creation, we create the `/cloudfuse` mountpoint and then these `mounts/umount`s are underneath that and get propagated into the running container.

daniel-goldstein:
@tpoterba I just realized I forgot to propagate the FUSE config through to worker jobs. Should I be and the singular test is just doing everything driver-side? Or is there a test we can write to ensure that worker jobs access the FASTA data?

danking:
BackendUtils.scala `collectDArray` has an optimization for single partition jobs.

danking:
Admittedly a debatable choice but worthwhile in the context of tests which generate tiny pipelines.

daniel-goldstein:
@tpoterba fixed the config issue and changed n_partitions to ensure workers are scheduled for the FASTA reading. I tested this on a single batch worker so the jobs overlapped and flexed the shared mount code, but we don't really have a guarantee in our test setup because batch has no way to force collocation of jobs (and even so we can't exactly force that the runtimes will overlap). I suppose if there's an issue here it will bubble up as a nondeterministic failure. Not great but perhaps good enough for now?

tpoterba:
Right now we run dataproc tests only on release, not on every commit, because they're too expensive/slow. That way we never release a version that can't pass. I wonder if that's also the right strategy here -- adding QoB release tests for things that only go wrong at scale.

That said, I don't want to block on that. Awesome change, thank you!

tpoterba (approved):

jigold (changes requested):
Give me a day to look this over one more time.

batch/batch/worker/worker.py
```diff
@@ -2056,6 +2188,15 @@ async def cleanup(self):
...
         assert self.worker.file_store is not None
         assert self.worker.fs
+        assert self.jvm
+
+        if self.cloudfuse:
+            for config in self.cloudfuse:
+                if config['mounted']:
+                    bucket = config['bucket']
+                    assert bucket
+                    mount_path = self.cloudfuse_data_path(bucket)
```
jigold, line 2198: Can we set config['mounted'] to False here?

batch/batch/worker/worker.py
```diff
@@ -2509,6 +2668,8 @@ def __init__(self, client_session: httpx.ClientSession):
 
         self.headers: Optional[Dict[str, str]] = None
 
+        self.cloudfuse_mount_manager = ReadOnlyCloudfuseManager()
```
jigold, line 2671: Do we need a shutdown method for the mount_manager?

jigold (dismissed):
Just some small things.

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -157,6 +159,9 @@ class ServiceBackend(
       if (backendContext.workerMemory != "None") {
         resources = resources.merge(JObject(("memory" -> JString(backendContext.workerMemory))))
       }
+      if (backendContext.storageRequirement != "0Gi") {
```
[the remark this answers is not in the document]
daniel-goldstein: Yes, this number is set in `service_backend.py`. There is also no need for any compatibility requirements between `service_backend.py` and `ServiceBackend.scala` so this can change as we wish as long as they agree with each other in the same commit.

hail/python/hail/backend/service_backend.py
```diff
@@ -553,22 +597,26 @@ async def inputs(infile, _):
         _, resp, _ = await self._rpc('load_references_from_dataset(...)', inputs, progress=progress)
         return orjson.loads(resp)
 
-    def add_sequence(self, name, fasta_file, index_file):
-        raise NotImplementedError("ServiceBackend does not support 'add_sequence'")
+    # Sequence and liftover information is stored on the ReferenceGenome
+    # and there is no persistent backend to keep in sync.
+    # Sequence and liftover information are passed on RPC
+    def add_sequence(self, name, fasta_file, index_file):  # pylint: disable=unused-argument
+        # FIXME Not only should this be in the cloud, it should be in the *right* cloud
```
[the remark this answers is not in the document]
daniel-goldstein: Ya I don't think we can do this until the client knows which cloud it is submitting jobs to, so I think best to come back to this after your VEP PR.

batch/batch/worker/worker.py
daniel-goldstein: I don't think so. When the worker exits, all jobs that were running on the worker should have gone through their `cleanup` step and unmounted their respective buckets, so there should be no remaining work for the `ReadOnlyCloudfuseManager` to do.

jigold (approved):

PR #12737, opened 2023-02-28
[query] Expose TableGen to Python

ehigham:
Add python classes analogous to those in Scala
Add printer/parsers to Scala and Python


hail/python/hail/table.py
```diff
@@ -554,6 +555,44 @@ def parallelize(cls, rows, schema=None, key=None, n_partitions=None, *,
             table = table.key_by(*key)
         return table
 
+    @staticmethod
+    @typecheck(
+        contexts=expr_stream(expr_any),
```
tpoterba, line 560: prefer expr_array here. We try not to expose streams much at the Python level -- just wrap the contexts IR in ToStream when creating the TableGen IR

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -796,23 +797,52 @@ object LowerTableIR {
...
-              s"${t.getClass.getSimpleName}: partitioner contains ${partitioner.numPartitions} partitions, got ",
-              invoke("str", TString, ArrayLen(ref))
-            )
-            ToStream(
-              If(ArrayLen(ref) ceq partitioner.numPartitions, ref, Die(dieMsg, ref.typ, traceId))
-            )
+          contexts = lowerIR {
+            bindIR(ToArray(contexts)) { ref =>
+              bindIR(ArrayLen(ref)) { len =>
+                // If contexts were an array we could assert this at compile time...
```
tpoterba, line 809: not true -- arrays aren't statically sized in the general case. We could write an analysis that tries to infer static length, but not worth it to eliminate this check.

tpoterba (dismissed):
2 nits -- nice work!

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
ehigham: But would those arrays not be fixed size generated from python?

tpoterba: you're right that if someone is passing in a partitioner, someone at some point probably is generating an array of a fixed size. But we don't know the runtime length of any particular IR of type TArray by default.

ehigham: fair enough!

hail/python/hail/table.py
ehigham: How about the body return type?

tpoterba (approved):

PR #12742, opened 2023-02-28
[batch] Allow more than one standing worker

jigold:
I added the configuration option for the minimum number of workers that should be present at any time. I tested this in my namespace. I'd like you to double check the logic is correct for the number of workers needed as I derived it by working through examples:

```python3
        n_live_instances = self.n_instances_by_state['pending'] + self.n_instances_by_state['active']
        n_standing_instances_needed = max(0, self.min_instances - self.n_instances)
        n_standing_instances_needed = min(
            n_standing_instances_needed,
            self.max_live_instances - n_live_instances,
            self.max_instances - self.n_instances,
            remaining_instances_per_autoscaler_loop,
            # 20 queries/s; our GCE long-run quota
            300,
        )
```

### DIFF
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -2,7 +2,7 @@ import asyncio
 import logging
 import random
 from collections import defaultdict
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Tuple
 
 import prometheus_client as pc
 import sortedcontainers
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -433,23 +433,28 @@ GROUP BY user;
                     break
 
         n_live_instances = self.n_instances_by_state['pending'] + self.n_instances_by_state['active']
-        n_standing_instances_needed = max(0, self.min_instances - self.n_instances)
-        n_standing_instances_needed = min(
-            n_standing_instances_needed,
-            self.max_live_instances - n_live_instances,
-            self.max_instances - self.n_instances,
-            remaining_instances_per_autoscaler_loop,
-            # 20 queries/s; our GCE long-run quota
-            300,
-        )
-        if n_standing_instances_needed > 0:
-            await self._create_instances(
-                n_instances=n_standing_instances_needed,
-                cores=self.standing_worker_cores,
-                data_disk_size_gb=self.data_disk_size_standing_gb,
-                regions=self.all_supported_regions,
-                max_idle_time_msecs=self.standing_worker_max_idle_time_secs * 1000,
+
+        capacity_for_live_instances = max(0, self.max_live_instances - n_live_instances)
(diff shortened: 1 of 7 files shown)


jigold:
This is ready for review. I keep getting unrelated transient errors in this PR.

batch/sql/estimated-current.sql
```diff
@@ -51,6 +51,7 @@ CREATE TABLE IF NOT EXISTS `pools` (
   `preemptible` BOOLEAN NOT NULL DEFAULT TRUE,
   `standing_worker_max_idle_time_secs` INT NOT NULL,
   `job_queue_scheduling_window_secs` INT NOT NULL,
+  `min_instances` BIGINT NOT NULL,
```
daniel-goldstein, line 54: Just one question, why is this on the `pools` and not the `inst_colls` like the `max_instances` setting? Is it because of job-private being an inst-coll but not a pool and this doesn't make sense for job-private?

daniel-goldstein (dismissed):

batch/sql/estimated-current.sql
jigold: exactly.

batch/batch/driver/instance_collection/pool.py
```diff
@@ -418,8 +433,18 @@ async def create_instances(self):
...
-            await self.create_instance(
+        n_standing_instances_needed = max(0, self.min_instances - self.n_instances)
+        n_standing_instances_needed = min(
+            n_standing_instances_needed,
+            self.max_live_instances - n_live_instances,
+            self.max_instances - self.n_instances,
+            remaining_instances_per_autoscaler_loop,
+            # 20 queries/s; our GCE long-run quota
+            300,
+        )
```
daniel-goldstein, line 444: I looked at this pretty carefully and think it's correct but since you asked me to double check I thought a bit to see if maybe we can write this in a clearer way. I am also always a bit foggy on live vs total instances so let me know if I'm misunderstanding something. Also, this is just an alternative. Feel free to take it if you think it's clearer or just dismiss and I'll approve.

```python3
capacity_for_live_instances = max(0, self.max_live_instances - n_live_instances)
capacity_for_any_instances = max(0, self.max_instances, self.n_instances)
capacity_for_new_instances = min(capacity_for_live_instances, capacity_for_any_instances)
if capacity_for_new_instances > 0:
    n_standing_instances_to_provision = min(capacity_for_new_instances, self.min_instances, remaining_instances_per_autoscaler_loop, 300)
    if n_standing_instances_to_provision > 0:
        await self._create_instances(…)
```

daniel-goldstein (dismissed):

batch/batch/driver/instance_collection/pool.py
jigold: This is much clearer! I had to modify it slightly though because you need `max(0, self.min_instances - self.n_instances)` in there to know how many you need to create to get to the minimum.

daniel-goldstein (approved):

PR #12745, opened 2023-03-01
[batch/ci] Address ServiceAccount token changes in Kubernetes 1.24

daniel-goldstein:
Prior to Kubernetes 1.24, when a `ServiceAccount` called `sa-foo` is created, a corresponding `Secret` containing a token for the service account is created call `sa-foo-token-xxxx`. The `ServiceAccount` resource contains the name of the corresponding secret, and to use the service account in Batch the batch-driver discovers the secret name from the service account resource.

In [Kubernetes >=1.24](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#urgent-upgrade-notes), creating a `ServiceAccount` no longer automatically creates a corresponding token secret. Our cluster auto-upgraded and PRs are failing because batch cannot find the secret field in the service account resource for SAs in test namespaces. From here on out, we need to make those token secrets ourselves. I explicitly added a token secret for the service accounts that need it and changed the batch-driver to handle both old and new service accounts.

I tested this in my own project since I already had an instance of Hail Batch / CI up and running. It was running on 1.23 so did not encounter this issue, but I:
1. Upgraded the cluster to 1.24
2. Dev deployed and received the same error that we're now seeing in [PRs](https://ci.hail.is/batches/7103889/jobs/14)
3. Manually redeployed batch and CI (from this branch)
4. Dev deployed successfully


batch/batch/driver/job.py
```diff
@@ -364,9 +364,13 @@ async def job_config(app, record, attempt_id):
...
-        assert len(sa.secrets) == 1
-
-        token_secret_name = sa.secrets[0].name
+        if sa.secrets is not None:
+            # ServiceAccounts created prior to Kubernetes 1.24 have autogenerated secrets
+            assert len(sa.secrets) == 1
+            token_secret_name = sa.secrets[0].name
+        else:
+            # ServiceAccounts post v1.24 don't have autogenerated secrets and we make those ourselves
+            token_secret_name = f'{name}-token'
```
danking, line 373: Hmm. So, you didn't change auth's create user stuff which makes me think that this stuff is only used by CI and only the service accounts you added below are relevant. Is that right?

danking (changes requested):

batch/batch/driver/job.py
daniel-goldstein: Yes, this only pertains to k8s service accounts not GSAs or Azure SPs

danking (approved):

PR #12748, opened 2023-03-02
[batch] Add new list_jobs query interface to REST calls

jigold:
This PR introduces Search v2.0 for querying jobs in a batch with the REST interface. The UI additions will come in a separate PR as well as the upgraded search for querying batches. We can decide whether to upgrade CI to v2. I think for backwards compatibility we need to have the default remain "1" in perpetuity in the batch client and you have to explicitly say you want version 2 as an argument if you want the new query language.

If we want to add tests for the SQL, I can do that once what I have currently is passing and we're happy with the code.


jigold:
I timed creating the new indices. Confirmed the operations are not blocking and they took 50 minutes on a 4 core database. We'd want to change the database size to 6 or 8 cores before doing this as it used about 70-80% of the available CPU on a 4 core machine.

daniel-goldstein:
@jigold Do you want me to review this or is this being passed off and can unassign?

jigold:
@danking Was this something you would like Iris to take on? If not, I can get to it later this month and will unassign Daniel.

jigold:
Also, to be more specific. The scope of work needed is 4 PRs. This is just 1 PR that exposes the REST API and the parser. There's still the list_batches query to implement as well as both UI search boxes with the help dropdown.

jigold:
I made a lot of progress on job groups this week so I think I can put this search PR on my active stack of work again. This PR isn't urgent, but would be great to have a review sometime within the next week or two.

batch/batch/front_end/operators.py
```diff
@@ -0,0 +1,99 @@
...
+    NotEqualPartialMatchOperator(),
+    EqualPartialMatchOperator(),
+    NotEqualExactMatchOperator(),
+    EqualExactMatchOperator(),
+]
+
+
+def parse_operator(term) -> Optional[Tuple[Operator, str, str]]:
+    for op in ordered_operators:
+        for symbol in op.user_symbols:
```
daniel-goldstein, line 95: I feel like instead of looping through all the operators for every term, we should just lay out a mapping from user-inputted tokens to the objects that represent that operator. We could probably get rid of the `user_symbols` class variable at that point too and just keep one `canonical_symbol` that is used in `to_sql`

batch/batch/front_end/operators.py
```diff
@@ -0,0 +1,99 @@
+import abc
+from typing import List, Optional, Tuple
+
+
+class Operator(abc.ABC):
+    user_symbols: List[str]
```
daniel-goldstein, line 6: Are these meant to be class fields or instance fields? I suppose it probably doesn't matter because these classes are used as singletons, but I would expect that the `user_symbols` here be somehow defined as abstract and that subclasses are forced to define it. I honestly don't know how class fields work in this regard but an abstract property should work.

batch/batch/front_end/query.py
```diff
@@ -0,0 +1,591 @@
...
+    MatchOperator,
+    Operator,
+    PartialMatchOperator,
+    parse_operator,
+)
+
+CURRENT_QUERY_VERSION = 1
+
+
+state_search_term_to_states = {
```
daniel-goldstein, line 22: Let's use an `Enum` instead of strings for these states. Some of these strings like `Running` are written multiple times and I imagine that would be an easy footgun if one of them were misspelled.

batch/batch/front_end/query.py
```diff
@@ -0,0 +1,591 @@
...
+
+
+def parse_search_by_operator_type(
+    expected_operator_type: Type[Operator], expected_field: Optional[str], term: str
+) -> Optional[Tuple[Operator, str, str]]:
+    result = parse_operator(term)
+    if result is None:
+        return None
+    operator, left, right = result
+    if expected_field and strip(left) != expected_field:
```
daniel-goldstein, line 74: If we split on whitespace in `parse_operator` we won't need to `strip` here.

batch/batch/front_end/query.py
```diff
@@ -0,0 +1,591 @@
...
+    return (operator, strip(left), strip(right))
+
+
+class Query(abc.ABC):
+    @staticmethod
+    def is_valid_query(term: str) -> bool:
+        raise NotImplementedError
+
+    def query(self) -> Tuple[str, List[Any]]:
+        raise NotImplementedError
```
daniel-goldstein, line 87: These should also use `abc.abstractmethod`

batch/batch/front_end/query.py
```diff
@@ -0,0 +1,591 @@
...
+    queries: List[Query] = []
+    q = request.query.get('q', '')
+
+    # logic to make time interval queries fast
+    min_start_gt_query: Optional[StartTimeQuery] = None
+    max_end_lt_query: Optional[EndTimeQuery] = None
+
+    if q:
+        terms = q.lstrip().rstrip().split('\n')
+        for term in terms:
```
daniel-goldstein, line 465: I think that, both for implementation and for documentation, this code could really use a comment with a [BNF grammar](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form). I think the classes you've already defined map to such a grammar very well, so it is mainly useful as a reference and a way to structure the parsing code. While perhaps not significant in practice, this approach spends O(n * m) time parsing query lines (where n is the number of operators and m is the number of query types). I think we can restructure things to take less of a trial-and-error approach that should also hopefully result in fewer lines of code. For example an incomplete grammar might look like:

```
<jobs-query-list> ::= <jobs-query> | <jobs-query> "\n" <jobs-query>

<jobs-query> ::= <cost-query> | <duration-query> | …

<cost-query> ::= "cost" <comparison-operator> <float>
… other query rules …

<comparison-operator> ::= "=" | "==" | "!=" | "<" | …
```

If this were the full grammar, and I'm given the text "cost < 1.0", I don't need to test all the query types to know how to parse this line. I can do the following:

```python
split = line.split()
if len(split) < 3:
    raise QueryError(…)
left, operator, right = split
if left == 'cost':
     # Will throw a QueryError if operator can't parse to a ComparisonOperator or right doesn't parse to a float.
     # It can also produce very precise error messages in those cases.
    queries.append(CostQuery.parse(operator, right))
elif … more rules that map to specific <*-query> rules …
```

batch/batch/front_end/query.py
```diff
@@ -0,0 +1,591 @@
...
+ (SELECT batch_id, job_id FROM job_attributes
+  WHERE `key` = %s AND `value` {op} %s))
+        '''
+        return (sql, [self.key, self.value])
+
+
+class TimeQuery(Query, abc.ABC):
+    @classmethod
+    def from_data(cls, field: str, operator: Operator, time_msecs: int):
+        q = object.__new__(cls)
```
daniel-goldstein, line 266: Using `object.__new__(cls)` is a smell to me. It looks to me like this method actually better suits the needs of a constructor, since the fields are "prepared" so to speak. The actual constructor needs to do work on its inputs in order to obtain values for fields, so the current constructor seems like it should be some static method that does those things and then calls the constructor with the prepared fields.

daniel-goldstein (changes requested):
This is an awesome feature! I have some ideas about the query string parsing but that's mainly it. It'll be really nice to start using this in the UI.

daniel-goldstein (changes requested):
This looks great, just a couple nits.

jigold:
Thanks! Still working on it :)

daniel-goldstein:
> Thanks! Still working on it :)

Oops! Didn't realize :)

batch/batch/front_end/query/operators.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class Operator(abc.ABC):
+    @abc.abstractmethod
+    def to_sql(self) -> str:
+        raise NotImplementedError
+
+
+class ComparisonOperator(Operator, abc.ABC):
+    pass
```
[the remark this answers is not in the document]
jigold: My editor complains if I don't do this...

jigold:
Ok. Actually ready for another look...

batch/batch/front_end/query/operators.py
```diff
@@ -0,0 +1,121 @@
...
+                raise QueryError(f'could not parse operator in query string "{s}"')
+            return ' '.join(groups)
+    raise QueryError(f'could not parse operator in query string "{s}"')
+
+
+def pad_maybe_operator(s: str) -> str:
+    for symbol in symbols_to_operator:
+        if symbol in s:
+            return pad_operator(s)
+    return s
```
daniel-goldstein, line 115: Ah I see this is for supporting operators without whitespace padding. That makes sense to suppor,t but it feels a bit complicated to find out the operator, then use a regular expression to pad the string, then parse it again later. What if instead of `return pad_operator(s)` we returned `[token.strip() for token in s.split(symbol)]` and this function instead of returning a `str` returned a `List[str]`. That basically cuts through straight to the point of having split on whitespace and removes the need for any regular expression parsing.

batch/batch/front_end/query/query.py
```diff
@@ -0,0 +1,321 @@
...
+    def query(self) -> Tuple[str, List[Any]]:
+        raise NotImplementedError
+
+
+class StateQuery(Query):
+    @staticmethod
+    def parse(op: str, state: str) -> 'StateQuery':
+        operator = get_operator(op)
+        if not isinstance(operator, ExactMatchOperator):
+            raise QueryError(f'unexpected operator "{op}" expected {ExactMatchOperator}')
```
daniel-goldstein, line 81: Curious, what does this look like with the actual class name interpolated into this string?

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -369,15 +369,19 @@ def __init__(self,
     async def cancel(self):
         await self._client._patch(f'/api/v1alpha/batches/{self.id}/cancel')
 
-    async def jobs(self, q=None):
+    async def jobs(self, q=None, version=None):
+        if version is None:
+            version = 1
+        if not (isinstance(version, int) and version >= 1):
```
daniel-goldstein, line 375: This feels a little out of character for the codebase. It seems to me that this sort of user-input validation belongs as close as possible to where the user would provide it instead of deep down in the implementation. I think inside these internal modules we can just add a type signature to this function and rely on that. Then if we expose this option to the user we should do this ValueError check there.

daniel-goldstein (changes requested):
Just a couple small points.

hail/python/hailtop/batch_client/aioclient.py
daniel-goldstein: I thought we considered `aioclient.py` an internal API, and the user API being the sync client

daniel-goldstein: I didn't realize that, I thought we make breaking changes to the aioclient interface.

daniel-goldstein: And it is undocumented.

jigold:
I fixed the SQL query. I had to use a feature available in MySQL v8.0.14 and later called lateral joins. I believe what this does is it scans the first table for rows that match, then scans the second lateral table for that matching row and applies the filter. So basically we have an iterator of the two tables joined together rather than a realized temporary table with the results of the entire subquery.

jigold:
See the last example on this page: https://dev.mysql.com/doc/refman/8.0/en/lateral-derived-tables.html

batch/batch/front_end/query/operators.py
jigold: I needed the actual operator in the result string which your example doesn't do. Not sure I ended up with a better solution, but probably nicer than what it was before.

hail/python/hailtop/batch_client/aioclient.py
daniel-goldstein: I think we should just add types to the signature of this function, so that way any internal uses should fail linting if they don't use integers, and then delete this if statement that's doing validation. The front end should validate the version anyway.

batch/batch/front_end/query/operators.py
daniel-goldstein: I'm ok with this. We should also be able to delete all the regular expression stuff right?

daniel-goldstein (dismissed):

jigold:
@daniel-goldstein Sorry for asking for another review, but there was a problem with the original query.

daniel-goldstein (approved):
LGTM

PR #12749, opened 2023-03-02
[batch] downgrade gcsfuse to avoid gcsfuse bug

danking:
https://github.com/GoogleCloudPlatform/gcsfuse/issues/982.


daniel-goldstein (approved):

daniel-goldstein:
Looks like it needs just one =

PR #12751, opened 2023-03-03
[ci] Add developers to test namespaces

daniel-goldstein:
The only noticeable change after this PR is that devs will be able to access `internal.hail.is/<PR-namespace>/<service>` while it is running and see jobs submitted by tests. Well, that and you can add another developer to a dev namespace as a developer without destroying the developer's existing namespace. Subsequent PRs will introduce on-demand dev namespaces and the ability to suspend the deletion of a test namespace. New context added to the CI pipeline is treated as optional to be backward compatible with the current CI. So devs won't be able to log in to test namespaces on *this* PR but will be able to once this PR becomes main.

### What has changed
- All developers from default are now added to all test namespaces using the `add_users` build.yaml step and removed at the end of the PR run through the `delete_users` step. These use the normal create and delete API instead of copying the user's gsa from the production namespace. This relies on / tests that the delete user endpoint is properly deleting cloud identities when the users are deleted (previously broken in GCP but fixed in this PR.
- The developer role no longer implicitly deletes and recreates a corresponding namespace. I wanted adding developers to test namespaces not to have side-effects that leaked out of the namespace. A follow-up PR will incorporate the ability for a developer to request an on-demand dev namespace, which should be made a lot easier after these changes. I think this also means that we can remove some permissions from the auth K8s ServiceAccount since it no longer needs the ability to create and delete namespaces.
- A fixed-but-sufficient number of oauth2 callbacks are hard-coded into the oauth2 secret from GCP/azure and then allocated to a given namespace. This is fairly self-contained, all that needs to happen is to tell `auth` what callback to use and rewrite those callback urls in gateway to route back to the appropriate auth. This is done only for test namespaces, production still just uses `auth.hail.is/oauth2callback`. This gets around a long-standing limitation of Google oauth2 clients where there is no programmatic way to change the oauth2 callbacks.

### What has stayed the same
- The lifecycle of dev and test namespaces has not changed (future PR).
- The semantics of a dev deploy has not changed (other than a different oauth2 callback which the user will not notice)

### Testing
I've tested this branch in my own google project both by deploying main and updating to this branch and by deploying this branch from the beginning. The CI in my project ran both some dev deploys and a test batch because I PR'd this branch against my fork. Tested that I could access both my dev namespace and the test namespace in that deployment. I manually verified that dev cloud identities in the test namespace were deleted after the namespace was cleaned up. I dev deployed this branch in Azure and made sure that create-user and delete-user created and deleted an SP.

cc: @jigold I know you have a lot on your plate with the migrations so no action required but cc'ing you here in case you have thoughts or concerns about this.

### DIFF
--- a/auth/auth/driver/driver.py
+++ b/auth/auth/driver/driver.py
@@ -4,8 +4,7 @@ import json
 import logging
 import os
 import random
-import secrets
-from typing import Any, Awaitable, Callable, Dict, List, Optional
+from typing import Any, Awaitable, Callable, Dict, List
 
 import aiohttp
 import kubernetes_asyncio.client
@@ -17,7 +16,6 @@ from gear.clients import get_identity_client
 from gear.cloud_config import get_gcp_config, get_global_config
 from hailtop import aiotools, httpx
 from hailtop import batch_client as bc
-from hailtop.auth.sql_config import SQLConfig, create_secret_data_from_config
 from hailtop.utils import secret_alnum_string, time_msecs
 
 log = logging.getLogger('auth.driver')
@@ -34,7 +32,7 @@ class DatabaseConflictError(Exception):
 
 
 class EventHandler:
-    def __init__(self, handler, event=None, bump_secs=60.0, min_delay_secs=0.1):
+    def __init__(self, handler, event=None, bump_secs=5.0, min_delay_secs=0.1):
         self.handler = handler
         if event is None:
             event = asyncio.Event()
@@ -166,7 +164,7 @@ class GSAResource:
 
     async def _delete(self, gsa_email):
         try:
-            await self.iam_client.delete(f'/serviceAccounts/{gsa_email}/keys')
+            await self.iam_client.delete(f'/serviceAccounts/{gsa_email}')
         except aiohttp.ClientResponseError as e:
             if e.status == 404:
                 pass
@@ -234,119 +232,6 @@ class AzureServicePrincipalResource:
         self.app_obj_id = None
 
 
-class DatabaseResource:
-    def __init__(self, db_instance, name=None):
-        self.db_instance = db_instance
-        self.name = name
-        self.password = None
-
-    async def create(self, name):
-        assert self.name is None
-
-        if is_test_deployment:
-            return
-
-        await self._delete(name)
-
-        self.password = secrets.token_urlsafe(16)
-        await self.db_instance.just_execute(
-            f'''
-CREATE DATABASE `{name}`;
-
-CREATE USER '{name}'@'%' IDENTIFIED BY '{self.password}';
-GRANT ALL ON `{name}`.* TO '{name}'@'%';
-'''
-        )
-        self.name = name
-
-    def secret_data(self):
-        with open('/database-server-config/sql-config.json', 'r', encoding='utf-8') as f:
-            server_config = SQLConfig.from_json(f.read())
-        with open('/database-server-config/server-ca.pem', 'r', encoding='utf-8') as f:
-            server_ca = f.read()
-        client_cert: Optional[str]
-        client_key: Optional[str]
-        if server_config.using_mtls():
-            with open('/database-server-config/client-cert.pem', 'r', encoding='utf-8') as f:
-                client_cert = f.read()
-            with open('/database-server-config/client-key.pem', 'r', encoding='utf-8') as f:
-                client_key = f.read()
-        else:
-            client_cert = None
-            client_key = None
-
-        if is_test_deployment:
-            return create_secret_data_from_config(server_config, server_ca, client_cert, client_key)
-
-        assert self.name is not None
-        assert self.password is not None
-
-        config = SQLConfig(
-            host=server_config.host,
-            port=server_config.port,
-            user=self.name,
-            password=self.password,
-            instance=server_config.instance,
-            connection_name=server_config.connection_name,
-            db=self.name,
-            ssl_ca='/sql-config/server-ca.pem',
-            ssl_cert='/sql-config/client-cert.pem' if client_cert is not None else None,
-            ssl_key='/sql-config/client-key.pem' if client_key is not None else None,
-            ssl_mode='VERIFY_CA',
-        )
-        return create_secret_data_from_config(config, server_ca, client_cert, client_key)
-
-    async def _delete(self, name):
-        if is_test_deployment:
-            return
-
-        # no DROP USER IF EXISTS in current db version
-        row = await self.db_instance.execute_and_fetchone('SELECT 1 FROM mysql.user WHERE User = %s;', (name,))
-        if row is not None:
-            await self.db_instance.just_execute(f"DROP USER '{name}';")
-
-        await self.db_instance.just_execute(f'DROP DATABASE IF EXISTS `{name}`;')
-
-    async def delete(self):
-        if self.name is None:
-            return
-        await self._delete(self.name)
-        self.name = None
-
-
-class K8sNamespaceResource:
-    def __init__(self, k8s_client, name=None):
-        self.k8s_client = k8s_client
-        self.name = name
-
-    async def create(self, name):
-        assert name not in ('default', DEFAULT_NAMESPACE)
-        assert self.name is None
-
-        await self._delete(name)
-
-        await self.k8s_client.create_namespace(
-            kubernetes_asyncio.client.V1Namespace(metadata=kubernetes_asyncio.client.V1ObjectMeta(name=name))
-        )
-        self.name = name
-
-    async def _delete(self, name):
-        assert name not in ('default', DEFAULT_NAMESPACE)
-        try:
-            await self.k8s_client.delete_namespace(name)
-        except kubernetes_asyncio.client.rest.ApiException as e:
-            if e.status == 404:
-                pass
-            else:
-                raise
-
-    async def delete(self):
-        if self.name is None:
-            return
-        await self._delete(self.name)
-        self.name = None
-
-
 class BillingProjectResource:
     def __init__(self, batch_client, user=None, billing_project=None):
         self.batch_client = batch_client
@@ -410,7 +295,6 @@ class BillingProjectResource:
 
 
 async def _create_user(app, user, skip_trial_bp, cleanup):
-    db_instance = app['db_instance']
     db = app['db']
     k8s_client = app['k8s_client']
     identity_client = app['identity_client']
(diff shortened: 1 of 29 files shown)


daniel-goldstein:
This PR contains some fixes to the GCP deployment scripts. LMK if you want me to pull those into a separate PR.

hail/python/hailtop/auth/auth.py
```diff
@@ -128,34 +136,59 @@ def create_user(username: str, login_id: str, is_developer: bool, is_service_acc
...
+                headers=headers) as session:
+            await request_retry_transient_errors(
+                session, 'DELETE', deploy_config.url('auth', f'/api/v1alpha/users/{username}')
+            )
+
+        if wait:
+            await _wait_user_state_transition(username, namespace, 'deleting', 'deleted')
+            print(f'Deleted user {username}')
+    except Exception as e:
+        raise DeleteUserException(f"Error while deleting user '{username}'") from e
```
daniel-goldstein, line 184: Most of this stuff is just refactoring. Pushing what was in the `{create|delete}_user.py` files into here so that I could have `wait=True` from python code and not just the CLI

danking (changes requested):
Can you update:
- `dev-docs/creating-a-developer-account.md`, and
- `dev-docs/development_process.md`
- ~~`infra/gcp/README.md`~~
either in this PR or in an appropriate later one to account for these changes?

danking:
Oops clicked submit review too soon

infra/install_bootstrap_dependencies.sh
```diff
@@ -8,8 +8,12 @@ umask 022
...
 curl -L https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/Release.key | sudo apt-key add -
 
+# For gcloud
+echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+
 sudo apt update
-sudo apt install -y docker.io python3-pip openjdk-8-jdk-headless jq skopeo
+sudo apt install -y docker.io python3-pip openjdk-8-jdk-headless jq skopeo google-cloud-cli google-cloud-cli-gke-gcloud-auth-plugin
 sudo snap install --classic kubectl
```
danking, line 17: What is snap and why don't we install kubectl using apt? https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl#apt Oddly, that page thinks the auth plugin should be called `google-cloud-sdk-gke-gcloud-auth-plugin`, note the sdk in place of cli.

infra/gcp/main.tf
```diff
@@ -471,7 +471,8 @@ module "test_gsa_secret" {
   project = var.gcp_project
   iam_roles = [
     "compute.instanceAdmin.v1",
-    "iam.serviceAccountUser",
+    "iam.serviceAccountAdmin",
+    "iam.serviceAccountKeyAdmin",
```
danking, line 475: Was this broken before?

infra/gcp/create_bootstrap_vm.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
danking, line 1: I see this is in parallel to Azure, but we don't have a markdown file describing steps that use this VM. Do you now have a series of steps for using terraform in GCP?

infra/gcp/bootstrap.sh
```diff
@@ -3,11 +3,11 @@
 source ../bootstrap_utils.sh
 
 function configure_gcloud() {
-    ZONE=${1:-"us-central1-a"}
+    REGION=${1:-"us-central1"}
+    ZONE=${2:-"us-central1-a"}
 
     gcloud -q auth configure-docker
```
danking, line 9: Is this line still relevant?

infra/azure/variables.tf
```diff
@@ -49,12 +49,14 @@ variable "ci_config" {
...
     watched_branches = list(tuple([string, bool, bool]))
+    test_oauth2_callback_urls = list(string)
     deploy_steps = list(string)
     github_context = string
   })
   default = null
 }
 
+# Deprecated
 variable oauth2_developer_redirect_uris {
```
danking, line 60: You'll remove this in a later PR, right?

build.yaml
```diff
@@ -1540,15 +1540,26 @@ steps:
       - default_ns
       - create_certs
   - kind: runImage
-    name: create_dummy_oauth2_client_secret
+    name: create_auth_config
```
danking, line 1543: What's the benefit of this as opposed to inlining the secret into the deployment.yaml as an envvar?

auth/auth/driver/driver.py
```diff
@@ -235,119 +232,6 @@ async def delete(self):
         self.app_obj_id = None
 
 
-class DatabaseResource:
```
danking, line 238: For posterity: functionality to create a database will be restored in a later PR that lets us create namespaces on-demand.

ci/ci/build.py
```diff
@@ -190,7 +184,7 @@ def __init__(self, params):
         self.clouds = json.get('clouds')
         self.run_if_requested = json.get('runIfRequested', False)
 
-        self.token = generate_token()
+        self.token = params.token
```
danking, line 187: Hmm. I *think* this is OK, but why do this? Doesn't seem bad for each step to get separate randomness.

ci/ci/ci.py
```diff
@@ -733,6 +739,15 @@ async def on_startup(app):
...
     )
 
+    async with await request_retry_transient_errors(
+        app['client_session'],
+        'GET',
+        deploy_config.url('auth', '/api/v1alpha/users'),
+        headers=service_auth_headers(deploy_config, 'auth'),
+    ) as resp:
+        users = await resp.json()
+        app['developers'] = [u for u in users if u['is_developer'] == 1 and u['username'] != 'test-dev']
```
danking, line 749: worth all of us remembering that the developers only get updated when CI restarts.

batch/gcp-create-worker-image.sh
```diff
@@ -18,11 +18,7 @@ create_build_image_instance() {
     python3 ../ci/jinja2_render.py '{"global":{"docker_root_image":"'${DOCKER_ROOT_IMAGE}'"}}' \
         build-batch-worker-image-startup-gcp.sh build-batch-worker-image-startup-gcp.sh.out
 
-    UBUNTU_IMAGE=$(gcloud compute images list \
-        --standard-images \
-        --filter 'family="ubuntu-minimal-2004-lts"' \
-        --format='value(name)')
-
+    UBUNTU_IMAGE=ubuntu-minimal-2004-focal-v20230216
```
danking, line 21: what's the thinking on this change?

ci/ci/envoy.py
```diff
@@ -101,22 +115,35 @@ def gateway_default_host(service: str) -> dict:
...
+    ]
+
+    oauth_callback_rewrites = [
+        {
+            'match': {'prefix': urlparse(namespace.oauth2_callback_url).path},
+            'route': {'cluster': f'{namespace.name}-auth', 'prefix_rewrite': f'/{namespace.name}/auth/oauth2callback'},
+        }
+        for namespace in internal_namespaces
+    ]
+    routes.extend(oauth_callback_rewrites)
```
danking, line 140: Do I understand correctly, that this is only necessary because oauth2 callback URLs are named by greek letters (due to the programmatic OAuth2 create/destroy limitations in GCP) and we need to direct the greek letter to the associated namespace (which has a developer-chosen name)?

ci/ci/github.py
```diff
@@ -896,21 +917,25 @@ async def _start_deploy(self, db: Database, batch_client):
...
             services.extend(test_services)
-            assert namespace is not None
-            await add_deployed_services(db, namespace, services, None)
+            await add_deployed_services(db, DEFAULT_NAMESPACE, services)
 
             log.info(f'creating deploy batch for {self.branch.short_str()}')
             deploy_batch = batch_client.create_batch(
                 attributes={
-                    'token': secrets.token_hex(16),
+                    'token': token,
```
danking, line 938: This is a nice change. What purpose did the token attribute serve before? It seems useless?

ci/ci/github.py
```diff
@@ -527,6 +546,7 @@ async def _start_build(self, db: Database, batch_client):
             self.set_build_state('error')
             self.source_sha_failed = True
             self.target_branch.state_changed = True
+            await release_namespace(db, namespace)
```
danking, line 549: Should we also release the namespace if there's a CancelledError? I think I don't fully understand how namespaces are resource managed.

danking (changes requested):

infra/install_bootstrap_dependencies.sh
daniel-goldstein: hm weird, I'll double check this by creating a bootstrap vm again and get gcloud + kubectl properly installed.

infra/gcp/create_bootstrap_vm.sh
daniel-goldstein: There is the `infra/gcp/README.md` which details the instructions of using terraform in GCP, but I will update it with a reference to this script.

infra/azure/variables.tf
daniel-goldstein: Ya removing this would immediately break current developer namespaces (I've already applied this terraform). So I'll remove it once our dev namespaces in azure are running this commit and no longer need those callback uris.

build.yaml
daniel-goldstein: I wanted to allow `make -C auth deploy` to keep working without adding additional complexity to the Makefile. By putting it in the deployment, the Makefile would need to somehow know which oauth2 callback CI assigned that namespace. But this way you can make deploy into your dev namespace without worrying about the callback. I did this with `ci-config` as well.

More philosophically, I feel like so far things that are templated into the deployment.yaml are things that the Makefile (or whoever is doing the templating) owns like SHA, images, etc. the oauth2 callback is not one of those things since it is a property of the namespace and follows the namespace's lifecycle. In that way it's more akin to the global-config, but it felt easier to manage in a separate secret than stuffing it into the global-config. Happy to take suggestions on simplifying things here though.

batch/gcp-create-worker-image.sh
daniel-goldstein: They added another image that made this query return two images and not one, which broke the invocation. More broadly, this is in the spirit of pinning our dependencies.

ci/ci/github.py
daniel-goldstein: Ya I guess so

danking:
Should I re-review?

daniel-goldstein:
I'm going to need to fix a lot of conflicts and I wanted to re-assess my approach just because it has been a while for me as well. I'll un-assign for a few days until I've freshened this up.

infra/install_bootstrap_dependencies.sh
daniel-goldstein: Still need to double check this.

daniel-goldstein:
A lot has changed and this is being replaced by smaller more focused PRs.

PR #12753, opened 2023-03-03
[query] Fix order of aggregate_cols, and make aggregate_cols do a loc…

tpoterba:
…al stream agg.

CHANGELOG: Fixed a longstanding bug where the columns were traversed in sorted order, rather than matrixtable order, in `MatrixTable.aggregate_cols`

### DIFF
--- a/hail/python/hail/matrixtable.py
+++ b/hail/python/hail/matrixtable.py
@@ -2223,7 +2223,7 @@ class MatrixTable(ExprContainer):
         """
         base, _ = self._process_joins(expr)
         analyze('MatrixTable.aggregate_cols', expr, self._global_indices, {self._col_axis})
-        cols_table = ir.MatrixColsTable(base._mir)
+        cols_table = ir.MatrixColsTable(ir.MatrixMapCols(base._mir, base.col._ir, []))
         subst_query = ir.subst(expr._ir, {}, {'sa': ir.Ref('row', cols_table.typ.row_type)})
 
         agg_ir = ir.TableAggregate(cols_table, subst_query)
--- a/hail/src/main/scala/is/hail/expr/ir/LowerMatrixIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/LowerMatrixIR.scala
@@ -4,7 +4,7 @@ import is.hail.backend.ExecuteContext
 import is.hail.expr.ir.functions.{WrappedMatrixToTableFunction, WrappedMatrixToValueFunction}
 import is.hail.expr.ir._
 import is.hail.types._
-import is.hail.types.virtual.{TArray, TBaseStruct, TDict, TInt32, TInterval, TString, TStruct}
+import is.hail.types.virtual.{TArray, TBaseStruct, TDict, TInt32, TInterval, TString, TStruct, Type}
 import is.hail.utils._
 
 object LowerMatrixIR {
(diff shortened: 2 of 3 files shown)


danking (changes requested):
Let's add a test. The regressionLogistic.vcf could be used as an example of out of order columns.

tpoterba:
oof, yeah, thanks.

hail/src/main/scala/is/hail/expr/ir/LowerMatrixIR.scala
```diff
@@ -478,10 +478,12 @@ object LowerMatrixIR {
         }
 
         loweredChild.mapGlobals('global.insertFields(colsField ->
-          aggOutsideTransformer(scanOutsideTransformer(irRange(0, 'global(colsField).len).map(idxSym ~> let(__cols_array = 'global(colsField), sa = '__cols_array(idxSym)) {
-            aggInsideTransformer(scanInsideTransformer(b0))
-          })))
-        ))
+          aggOutsideTransformer(scanOutsideTransformer(ToArray(StreamZip(
+            FastIndexedSeq(ToStream(GetField(Ref("global", loweredChild.typ.globalType), colsFieldName)), StreamIota(0, 0)),
```
danking, line 482: Is the second stream is meant to be 1, 2, 3? Why isn't step = 1?

It seems like the old code never had access to the index, so why bother to include it here at all?

hail/python/hail/matrixtable.py
```diff
@@ -2223,7 +2223,7 @@ def aggregate_cols(self, expr, _localize=True) -> Any:
         """
         base, _ = self._process_joins(expr)
         analyze('MatrixTable.aggregate_cols', expr, self._global_indices, {self._col_axis})
-        cols_table = ir.MatrixColsTable(base._mir)
+        cols_table = ir.MatrixColsTable(ir.MatrixMapCols(base._mir, base.col._ir, []))
```
danking, line 2226: Why is it kosher for map cols to drop the key? Shouldn't we need to use a key by for that?

danking (changes requested):

danking:
bump

hail/src/main/scala/is/hail/expr/ir/LowerMatrixIR.scala
tpoterba: totally a bug, cause of the test failures. The index is used in aggregations/scans to join the column stream with the agg/scan results.

hail/python/hail/matrixtable.py
tpoterba: don't actually have a MatrixKeyColsBy. MMC does both. We could separate it out, but don't have separate nodes because we (a) don't really care about optimizing around col keys since they're unordered, and (b) since the col key is totally ignored as we lower.

danking: Sorry if I'm being dense, but isn't ordering relevant to aggregation? I'm thinking of `hl.agg.collect` in particular. I thought it was true that the results came in key-order. If ordering is important, why is it safe to drop the key before `ir.TableAggregate`?

danking (changes requested):

hail/python/hail/matrixtable.py
tpoterba: This PR makes a semantic change in aggregation order (currently in sort order aka order of cols(), but desired behavior is  matrixtable column order (localize_entries order). This is what we agreed in team meeting the week before the IBG workshop, but let's revisit Wednesday to make sure we're still on the same page.

danking:
bump

danking:
@tpoterba bump!

tpoterba:
added to the team meeting agenda tomorrow.

PR #12756, opened 2023-03-06
[ci] Fix default namespace record for dev namespaces

daniel-goldstein:
Stacked on #12755, this shouldn't have an impact on production, but is just meant to replace `default` in test CI databases with the actual name of the namespace that the CI is in. Should mimic [the prior migration](https://github.com/hail-is/hail/blob/main/ci/sql/active-namespaces.sql) just parametrized by namespace.

### DIFF
--- a/ci/sql/fix-default-namespace.py
+++ b/ci/sql/fix-default-namespace.py
@@ -1,0 +1,28 @@
+import os
+import asyncio
+from gear import Database
+
+
+async def main():
+    default_namespace = os.environ['HAIL_NAMESPACE']
+
+    # This fix is only necessary for dev/test namespaces
+    if default_namespace == 'default':
+        return
+
+    db = Database()
+    await db.async_init()
+
+    await db.just_execute(
+        '''
+DELETE FROM `active_namespaces` WHERE `namespace` = 'default';
+INSERT INTO `active_namespaces` (`namespace`) VALUES (%s);
+INSERT INTO `deployed_services` (`namespace`, `service`) VALUES
+(%s, 'auth'), (%s, 'batch'), (%s, 'batch-driver'), (%s, 'ci');
+''',
+        (default_namespace,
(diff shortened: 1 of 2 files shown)


ci/sql/fix-default-namespace.py
```diff
@@ -0,0 +1,28 @@
...
+
+async def main():
+    default_namespace = os.environ['HAIL_NAMESPACE']
+
+    # This fix is only necessary for dev/test namespaces
+    if default_namespace == 'default':
+        return
+
+    db = Database()
+    await db.async_init()
```
jigold, line 14: Let's add an `await db.async_close()` at the end.

jigold (changes requested):

ci/sql/fix-default-namespace.py
```diff
@@ -0,0 +1,28 @@
+import os
+import asyncio
+from gear import Database
+
+
+async def main():
+    default_namespace = os.environ['HAIL_NAMESPACE']
+
```
jigold, line 8: I know this check is probably right, but is it worth explicitly making sure that we're not in the SCOPE = deploy as well just in case the default_namespace is always default. If you've tested that the namespace here is not always "default" (I remember there was an issue with this and that's why we have the variable `is_test_deployment`), then feel free to ignore.

jigold (dismissed):

ci/sql/fix-default-namespace.py
daniel-goldstein: I don't think the scope is what we want here -- a test CI can do a `deploy` into its own namespace, which is when `is_test_deployment` would be useful, but in either case it's the namespace that is relevant here not which scope we're in. This fixes all namespaces that are not named `default`. I'm confident that this variable represents the namespace the database is created for.

jigold (approved):

PR #12757, opened 2023-03-06
[batch] Setup all of the sql migration infra for deduping resource ids

jigold:
## Overview:
- The goal of this PR is to setup all of the SQL tables and triggers such that the next PR can iterate over the attempt resources table to copy over the correct deduped resource id as well as copy over all of the 4 billing tables (bp_users, bps, batches, jobs) to collapse rows with the same deduped resource id.
- I'll put up the subsequent PR as well after this so you can get an idea for why some of the functionality I added in this PR is there.

## Attempt Resources Table:
- Currently, the attempt resources table has a trigger after insert that creates entries in the billing tables if the msec_diff != 0.
- We will have merged the mitigation such that for all new attempt resources entries, resource_id == deduped_resource_id. Therefore, we can add a before insert trigger that sets deduped_resource_id to resource_id.
- We do **not** need a column `migrated` here because the only transformation we want to do with this table is set the deduped_resource_id to the correct value from the dedup resource id mapping. We cannot use the attempts or attempt resources table to populate the new billing tables because we have lost the intermediate update times and only have time_start and time_ended for each attempt.
- The next migration will populate the correct resource id in 100 row chunks as we cannot do one massive update (will take days and the transaction log will be enormous).
- The `attempt_resources_after_insert` table needs to only populate the new `v3` billing tables if the rows for that primary key have already been copied from the `v2` tables (i.e. migrated=1 on the row in the `v2` billing table). The issue here is that we need to account for the time in which we have not copied a row from the v2 billing table, but we are already adding it to the `v3` table. This would mean we'd end up double counting:

For example,

Say we have this row currently in the agg_bp_resources_v2 table:
(test_bp, compute, 100, migrated=0)

Say we insert this row into the attempt_resources table:
(1, 1, bp=test_bp, resource=compute, 50)

The agg_bp_resources_v2 table will be
(test_bp, compute, 150, migrated=0)

If we add the usage to the `v3` table (not what I am proposing to do):
(test_bp, compute, 50)

Then later on when we copy over rows with migrated=0 in the `v2` billing tables, we'll end up with this row in the `v3` table because we need a `+=` here to account for newly added records:
(test_bp, compute, 200)

It should be (test_bp, compute, 150).

Therefore, for the `attempt_resources_after_insert` trigger, we need to check whether we have migrated the row already (migrated = 1 on the corresponding billing table) before inserting / updating the billing table rows in `v3`.

## Attempts Table
- There's a trigger `attempts_after_update` here that updates the usage of all four `v2` billing tables based on the start and rollup times.
- Like for the attempt_resources table, we only want to add/update rows into `v3` billing tables if the corresponding rows for `v2` have already been migrated.
- This is where the extra join comes in as a filter:
```
    JOIN aggregated_job_resources_v2 ON
      aggregated_job_resources_v2.batch_id = attempt_resources.batch_id AND
      aggregated_job_resources_v2.job_id = attempt_resources.job_id AND
      aggregated_job_resources_v2.resource_id = attempt_resources.resource_id AND
      aggregated_job_resources_v2.token = rand_token
    WHERE batch_id = NEW.batch_id AND job_id = NEW.job_id AND attempt_id = NEW.attempt_id AND migrated = 1
```

## Aggregated Billing Tables v2
- We need to add a before insert trigger to set migrated=1 for all new rows that get inserted in each of the corresponding tables
- We need after update triggers to copy the rows from the v2 table into the corresponding v3 table if OLD.migrated = 0 and NEW.migrated = 1.

## Concerns
- The only qualm I have right now with the design I've described above is making sure we can change the attempt_resources table `deduped_resource_id` into the primary key of the table and eventually remove the current `resource_id` without it taking forever and blocking the table. I'll need a clone of the database to determine this.


batch/sql/estimated-current.sql
```diff
@@ -486,23 +537,63 @@ BEGIN
...
       resource_id,
       rand_token,
       msec_diff_rollup * quantity
     FROM attempt_resources
     WHERE batch_id = NEW.batch_id AND job_id = NEW.job_id AND attempt_id = NEW.attempt_id
     ON DUPLICATE KEY UPDATE `usage` = `usage` + msec_diff_rollup * quantity;
 
+    INSERT INTO aggregated_batch_resources_v3 (batch_id, resource_id, token, `usage`)
+    SELECT attempt_resources.batch_id,
+      attempt_resources.resource_id,
```
jigold, line 566: Note that in these triggers, we're using the resource_id instead of the deduped_resource_id. At this point, any new resource ids always equal the deduped resource id because we already performed the mitigation to not add dups to the database and there's a complete resource dedup map. When we eventually rename the columns from resource_id -> deduped_resource_id and deduped_resource_id to resource_id when we want to change the primary key to use the "deduped resource id" for the attempt resources table, the rename operation only changes the column references for indices and foreign key constraints. It does not change the name or reference to of the column in the triggers or stored procedures. So it's best to keep this to be the `resource_id` here so as to be able to drop the old "resource_id" column later on.

jigold:
This PR is instantaneous for the database migration.

batch/sql/estimated-current.sql
daniel-goldstein: So to parrot this back, use `resource_id` when the two are equivalent because it will be the permanent name in the long-run?

batch/sql/estimated-current.sql
```diff
@@ -723,20 +840,177 @@ BEGIN
...
+CREATE TRIGGER aggregated_bp_user_resources_v2_after_update AFTER UPDATE ON aggregated_billing_project_user_resources_v2
+FOR EACH ROW
+BEGIN
+  DECLARE new_resource_id INT;
+  DECLARE cur_n_tokens INT;
+  DECLARE rand_token INT;
+
+  IF OLD.migrated = 0 AND NEW.migrated = 1 THEN
+    SELECT n_tokens INTO cur_n_tokens FROM globals LOCK IN SHARE MODE;
+    SET rand_token = FLOOR(RAND() * cur_n_tokens);
```
daniel-goldstein, line 943: Why generate a rand token instead of using `NEW.rand_token`? It seems like a nice property that the resources for the same token would collapse into the same token on the other side. (same goes for the other triggers)

daniel-goldstein (changes requested):
I made sure to re-read it a couple times because it's easy for my eyes to glaze over with this kind of thing, but this all seems reasonable. I really only have one concern about the tokens making things a bit more complicated than needs to be.

batch/sql/estimated-current.sql
jigold: great suggestion!

batch/sql/estimated-current.sql
jigold: yes. exactly. It does make me uneasy to do this, but I'm worried about what happens when we try and cleanup the attempt resources table. This seemed better than having the triggers temporarily referencing a column that does not exist. But maybe I'm overthinking it.

jigold:
Thanks for the thorough review! I know it's not easy to do that for these types of PRs.

batch/sql/estimated-current.sql
jigold: Ok. I made things as clear as possible referencing `deduped_resource_id` where possible. For these specific queries, however, we cannot assume that the deduped resource id has already been populated as that only applies to new attempts that were generated after #12028 has fully redeployed. Therefore, we use the `resource_id` as that will never be NOT NULL and should be already deduped post-mitigation.

daniel-goldstein (approved):

PR #12760, opened 2023-03-07
[compiler] Use code tuple size to decide whether to create an InsertF…

tpoterba:
…ieldsStruct or pointer

### DIFF
--- a/hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SBaseStruct.scala
+++ b/hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SBaseStruct.scala
@@ -110,7 +110,7 @@ trait SBaseStructValue extends SValue {
   }
 
   def insert(cb: EmitCodeBuilder, region: Value[Region], newType: TStruct, fields: (String, EmitValue)*): SBaseStructValue = {
-    if (newType.size < 64 || fields.length < 16)
+    if (st.settableTupleTypes().length + fields.map(_._2.emitType.settableTupleTypes.length).sum < 64)
       return _insert(newType, fields: _*)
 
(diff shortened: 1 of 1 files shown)


hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SBaseStruct.scala
```diff
@@ -110,7 +110,7 @@ trait SBaseStructValue extends SValue {
   }
 
   def insert(cb: EmitCodeBuilder, region: Value[Region], newType: TStruct, fields: (String, EmitValue)*): SBaseStructValue = {
-    if (newType.size < 64 || fields.length < 16)
+    if (st.settableTupleTypes().length + fields.map(_._2.emitType.settableTupleTypes.length).sum < 64)
```
patrick-schultz, line 113: Besides changing to measure struct size by code tuple size, this also changes from "new struct is small, or added fields are small" to just "new struct is small". Can you explain that change? Do we not want to use an InsertFieldsStruct when adding a few fields to a big struct? (Actually, why does the size of the original struct matter at all?)

patrick-schultz (changes requested):

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SBaseStruct.scala
patrick-schultz: Makes sense, thanks!

tpoterba: good question, should have clarified the reason in the initial comment!

PR #12761, opened 2023-03-07
[batch] Populate v3 billing tables with deduped resource ids

jigold:
Stacked on #12757

- This PR gets the ranges of existing rows from the attempt_resources, aggregated_*_resources_v2 tables in bunches of 100 and then migrates each bunch by triggering an after update trigger for those rows that haven't been migrated. The triggers were added in #12757. 
- There's an audit at the end to make sure the new v3 tables give the same answer as the old v2 tables with duplicate resources.
- We use the same trick with a burn-in period to avoid the birthday problem with deadlocks.
- I added a function that generates the where statements programmatically based on looking at the where statement from previous migrations where we wrote out the where statement by hand. I think this way is less error-prone than writing out the where statement for each table, but it might be harder to reason about. Let me know if this way is too confusing.

### DIFF
--- a/batch/batch/driver/job.py
+++ b/batch/batch/driver/job.py
@@ -89,15 +89,16 @@ async def add_attempt_resources(app, db, batch_id, job_id, attempt_id, resources
             # This must be sorted in order to match the order of values in the actual SQL table!
             _resources = dict(sorted(_resources.items()))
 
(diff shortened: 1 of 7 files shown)


jigold:
Note to self that the audit should test for equality here.

jigold:
We decided not to combine all of these migrations into one PR. Instead they'll be separate PRs such as #12817, #12813

PR #12762, opened 2023-03-07
[batch] Cleanup infra from dedup resources migration

jigold:
Stacked on #12761

- This PR gets rid of the triggers needed for the migration as well as no longer needed columns
- There's some comments about the attempt resources `deduped_resource_id` column in the migration script. The goal is to swap the resource_id with the deduped_resource_id. This might be more complicated than just keeping both columns. I'm not sure. I also want to check the performance and blocking of the rename and generating the foreign key constraints.

### DIFF
--- a/batch/sql/cleanup-deduped-resource-ids-migration.sql
+++ b/batch/sql/cleanup-deduped-resource-ids-migration.sql
@@ -1,0 +1,210 @@
+DELIMITER $$
+
+DROP TRIGGER IF EXISTS attempts_after_update $$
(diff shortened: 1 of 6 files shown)


jigold:
Note to self to manually check the database to make sure the triggers and indices are all correct.

jigold:
I removed the complicated cleanup of the attempt resources table. For now, let's just make sure all of the deduped ids are not null and the foreign key constraint is there.

jigold:
Closing this PR for now. Will reopen or create a new PR when we're ready.

PR #12763, opened 2023-03-07
[batch] Drop remaining attempt_resources infra for deduping resource ids

jigold:
- This PR is stacked on #12762. It just removes the `deduped_resource_id` column from the attempt_resources table. Note that the trigger here is still correct even though it's backwards from the original column mapping (we renamed the columns in #12762) because resource_id == deduped_resource_id.


jigold:
I haven't decided whether it's worth keeping the old resource_id around rather than deleting it. Thoughts?

jigold:
We decided we'll address the extra "resource_id" column in the attempts_resources table separately from this stack at a later date once we're confident the new billing tables are correct.

PR #12768, opened 2023-03-07
[query] fix ggplot geom_bar, geom_col, and geom_point use of color

danking:
- `geom_point` did not correctly include color in the legend when the color was a continuous variable that had been mapped to a discrete one, e.g. an integer that the user specified as the color.

- `geom_bar` and `geom_col` suffered the same issue but additionally lacked the ability to name each color group in the legend.

### DIFF
--- a/hail/python/hail/ggplot/scale.py
+++ b/hail/python/hail/ggplot/scale.py
@@ -24,7 +24,7 @@ class Scale(FigureAttribute):
         return lambda x: x
 
     @abc.abstractmethod
-    def is_discrete(self):
+    def is_discrete_output(self):
         pass
 
     @abc.abstractmethod
@@ -80,7 +80,7 @@ class PositionScaleGenomic(PositionScale):
     def transform_data(self, field_expr):
         return field_expr.global_position()
 
-    def is_discrete(self):
+    def is_discrete_output(self):
         return False
 
     def is_continuous(self):
@@ -107,7 +107,7 @@ class PositionScaleContinuous(PositionScale):
     def transform_data(self, field_expr):
         return field_expr
 
-    def is_discrete(self):
+    def is_discrete_output(self):
         return False
 
     def is_continuous(self):
@@ -124,7 +124,7 @@ class PositionScaleDiscrete(PositionScale):
     def transform_data(self, field_expr):
         return field_expr
 
-    def is_discrete(self):
+    def is_discrete_output(self):
         return True
 
     def is_continuous(self):
@@ -138,7 +138,7 @@ class ScaleContinuous(Scale):
     def transform_data(self, field_expr):
         return field_expr
 
-    def is_discrete(self):
+    def is_discrete_output(self):
         return False
 
     def is_continuous(self):
@@ -158,7 +158,7 @@ class ScaleDiscrete(Scale):
     def transform_data(self, field_expr):
         return field_expr
 
-    def is_discrete(self):
+    def is_discrete_output(self):
         return True
 
     def is_continuous(self):
@@ -212,13 +212,28 @@ class ScaleDiscreteManual(ScaleDiscrete):
 
 
 class ScaleColorContinuous(ScaleContinuous):
+    def is_discrete_output(self):
+        return True
 
     def create_local_transformer(self, groups_of_dfs):
         overall_min = None
         overall_max = None
         for group_of_dfs in groups_of_dfs:
             for df in group_of_dfs:
-                if self.aesthetic_name in df.columns:
+                if self.aesthetic_name in df.attrs:
+                    assert self.aesthetic_name not in df.columns
+                    v = df.attrs[self.aesthetic_name]
+                    if overall_min is None:
+                        overall_min = v
+                    else:
+                        overall_min = min(v, overall_min)
+
+                    if overall_max is None:
+                        overall_max = v
+                    else:
+                        overall_max = max(v, overall_max)
+                elif self.aesthetic_name in df.columns:
+                    assert self.aesthetic_name not in df.attrs
                     series = df[self.aesthetic_name]
                     series_min = series.min()
                     series_max = series.max()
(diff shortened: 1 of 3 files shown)


danking:
I gotta figure out how to write tests for this, but I've pushed so that Ally can take a look.

danking:
Hmm. I'm perhaps a bit too clever with my image tests. They differ between linux and OS X and its not easy to see the difference given the way our CI works.

danking:
OK, in the interest of fixing these problems for people, let's merge without tests. Testing that two plots look the same will have to be future work. I think the environment/OS is influencing image generation in some subtle way such that the generated plots images are not the same.

hail/python/hail/ggplot/scale.py
```diff
@@ -212,13 +212,28 @@ def get_values(self, categories):
 
 
 class ScaleColorContinuous(ScaleContinuous):
+    def is_discrete_output(self):
+        return True
```
patrick-schultz, line 216: I'm not sure what distinction you're making with the `is_discrete`->`is_discrete_output` change. I think it's pretty fundamental in the ggplot model that we have discrete and continuous scales, and they behave differently. Making `ScaleColorContinuous` a discrete scale seems confusing at best. We do have `ScaleColorHue` as a discrete color scale. Can you explain the motivation for this change?

patrick-schultz (changes requested):

hail/python/hail/ggplot/scale.py
danking: Ahh, hmm. So this was a great comment. It sent me down a rabbit hole and I realized I misunderstand (and also our users misunderstand) what `scale_color_continuous` does.

We have two color scales:

- `scale_color_hue` (aliased by `scale_color_discrete`) which maps color name strings to colors (discrete to discrete).
- `scale_color_continuous` which linearly maps from some bounded continuous space to colors (continuous to continuous).

At least one user in at least one case was trying to use `scale_color_continuous` to convert a _categorical_ integer field into a discrete set of colors. This makes me think we're missing one more scale:

- `scale_color_categorical` which maps some subset of a (possibly continuous) space to evenly spaced colors. This scale should absolutely participate in legends and facets because we're ignoring the continuity of the type and treating these values as discrete/categorical.

Getting back to your question: I thought scale_color_continuous was scale_color_categorical, ergo all these changes and why I thought the output of scale_color_continuous is "discrete" (aka categorical).

I guess this means I should create a new scale?

danking: Ah, that's helpful context.

So, here's a concrete example of bad behavior:
```python3
import hail as hl
ht = hl.utils.range_table(10)
ht = ht.annotate(subpop = ht.idx % 3)
ht = ht.annotate(proj_title = hl.literal(['a', 'b'])[ht.idx % 2])
ht = ht.annotate(x = hl.rand_int32(0, 3))

p = hl.ggplot.ggplot(ht, hl.ggplot.aes(x=ht.x, color =  ht.subpop))
p += hl.ggplot.geom_bar()
p += hl.ggplot.xlab('x')
p += hl.ggplot.ylab('count')
p += hl.ggplot.ggtitle(" Subcontinental PCA pre-outlier removal")
p += hl.ggplot.labs(shape = 'Project', color = 'Population')

p
```
This should create a stacked bar plot with color distinguishing between the different populations. Instead, it gives me this black bar plot.

Screen Shot 2023-04-07 at 12 39 09

The trouble is that `geom_bar` with a continuous color scaling doesn't make sense.

danking: The tricker case is `geom_point`. `geom_point` *does* have a reasonable interpretation of continuous color, but:
1. We don't produce a continuous visual legend/scale for that color.
2. In reality, we want to treat this integer as a categorical and get a distinct legend entry for each value.
Screen Shot 2023-04-07 at 12 42 01

danking: I can't specify `scales` in the `ggplot` function. If I try to manually specify `scale={'color': hl.ggplot.scale_color_discrete()}` by directly calling the class constructor, Hail's ggplot complains:

```
ValueError: Invalid scale for aesthetic color of type int32
```

patrick-schultz: Ah, thanks, I understand now. I think the core issue is that we don't have any way to override the discrete/continuous inference from datatype. This works, but I agree we should find a way to do it more directly:
```
import hail as hl
ht = hl.utils.range_table(10)
ht = ht.annotate(subpop = hl.str(ht.idx % 3))
ht = ht.annotate(proj_title = hl.literal(['a', 'b'])[ht.idx % 2])
ht = ht.annotate(x = hl.rand_int32(0, 3))

p = hl.ggplot.ggplot(ht, hl.ggplot.aes(x=ht.x, fill =  ht.subpop))
p += hl.ggplot.geom_bar()
p += hl.ggplot.xlab('x')
p += hl.ggplot.ylab('count')
p += hl.ggplot.ggtitle(" Subcontinental PCA pre-outlier removal")
p += hl.ggplot.labs(shape = 'Project', color = 'Population')

p.show()
```

We should copy ggplot's interface for manually specifying continuous/discrete, but after a few minutes I haven't been able to figure out how to do it. I'll let you know if I find it.

patrick-schultz: It's possible ggplot lets you assign a discrete scale to any datatype, in which case we should too.

patrick-schultz: Ah, R has a generic way to force a continuous type to be used categorically, `factor`. So they would write this
```
aes(x=ht.x, fill = factor(ht.subpop))
```
Seems like a good pattern to emulate.

danking:
Iris is gonna add a `factor` or something to this effect to make this work as expected. In the mean time the user can `hl.str` the field they want to be a factor.

PR #12769, opened 2023-03-08
[workshop] Add necessary infrastructure for IBG 2023 workshop 

tpoterba:
* add qq_plot and manhattan built on plotly. It's a snarly problem to add support for all the customization necessary to build these out of ggplot directly, but this lets us use ggplot in the tutorial notebooks without having to use a separate plotting lib.
* Fix relatedness estimation for mating simulation

### DIFF
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -3905,7 +3905,7 @@ def group_by(f: Callable, collection) -> DictExpression:
 @typecheck(f=func_spec(2, expr_any),
            zero=expr_any,
            collection=expr_oneof(expr_set(), expr_array()))
-def fold(f: Callable, zero, collection) -> Expression:
+def fold(f: Callable, zero, collection):
     """Reduces a collection with the given function `f`, provided the initial value `zero`.
 
     Examples
--- a/hail/python/hail/ggplot/aes.py
+++ b/hail/python/hail/ggplot/aes.py
@@ -1,6 +1,7 @@
 from collections.abc import Mapping
 from hail.expr import Expression
 from hail import literal
+import hail
 
 
 class Aesthetic(Mapping):
@@ -44,6 +45,8 @@ def aes(**kwargs):
     hail_field_properties = {}
 
     for k, v in kwargs.items():
+        if isinstance(v, (tuple, dict)):
+            v = hail.str(v)
         if not isinstance(v, Expression):
             v = literal(v)
         hail_field_properties[k] = v
--- a/hail/python/hail/methods/relatedness/mating_simulation.py
+++ b/hail/python/hail/methods/relatedness/mating_simulation.py
@@ -1,70 +1,130 @@
 import hail as hl
-from hail.typecheck import typecheck, numeric
+from hail.typecheck import typecheck, numeric, nullable
+import random
+from hail.utils.java import info
 
 
-@typecheck(mt=hl.MatrixTable, n_rounds=int, generation_size_multiplier=numeric, keep_founders=bool)
-def simulate_random_mating(mt, n_rounds=1, generation_size_multiplier=1.0, keep_founders=True):
+@typecheck(mt=hl.MatrixTable,
+           n_rounds=int,
+           pairs_per_generation_multiplier=numeric,
+           children_per_pair=int,
+           seed=nullable(int))
+def simulate_random_mating(mt,
+                           n_rounds=1,
+                           pairs_per_generation_multiplier=0.5,
+                           children_per_pair=2,
+                           seed=None):
     """Simulate random diploid mating to produce new individuals.
+
+    .. include:: _templates/experimental.rst
+
+    Exmaples
+    --------
+
+    >>> dataset_sim = hl.simulate_random_mating(dataset, n_rounds=2, pairs_per_generation_multiplier=0.5)
 
     Parameters
     ----------
     mt
     n_rounds : :obj:`int`
         Number of rounds of mating.
-    generation_size_multiplier  : :obj:`float`
-        Ratio of number of offspring to current population for each round of mating.
-    keep_founders :obj:`bool`
-        If true, keep all founders and intermediate generations in the final sample list. If
-        false, keep only offspring in the last generation.
-
+    pairs_per_generation_multiplier  : :obj:`float`
+        Ratio of number of mating pairs to current population size for each round of mating.
+    children_per_pair  : :obj:`int`
+        Number of children per mating pair.
     Returns
     -------
     :class:`.MatrixTable`
     """
-    if generation_size_multiplier <= 0:
+    if pairs_per_generation_multiplier <= 0:
         raise ValueError(
-            f"simulate_random_mating: 'generation_size_multiplier' must be greater than zero: got {generation_size_multiplier}")
+            f"simulate_random_mating: 'generation_size_multiplier' must be greater than zero: got {pairs_per_generation_multiplier}")
     if n_rounds < 1:
         raise ValueError(f"simulate_random_mating: 'n_rounds' must be positive: got {n_rounds}")
 
-    ck = list(mt.col_key)[0]
-
     mt = mt.select_entries('GT')
-
     ht = mt.localize_entries('__entries', '__cols')
 
-    ht = ht.annotate_globals(
-        generation_0=hl.range(hl.len(ht.__cols)).map(lambda i: hl.struct(s=hl.str('generation_0_idx_') + hl.str(i),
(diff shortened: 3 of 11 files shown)


hail/python/hail/methods/relatedness/mating_simulation.py
```diff
@@ -1,70 +1,162 @@
...
+                    merged_parent_rel[k] = .5 * v
+
+            if mother in merged_parent_rel:
+                assert father in merged_parent_rel
+                merged_parent_rel[mother] += 0.25
+                merged_parent_rel[father] += 0.25
+            else:
+                merged_parent_rel[mother] = 0.25
+                merged_parent_rel[father] = 0.25
+            sibling_rel = merged_parent_rel[mother]  # here mother/father should be the same, since
```
danking, line 106: since?

hail/python/hail/methods/relatedness/mating_simulation.py
```diff
@@ -1,70 +1,162 @@
...
+            for k, v in mother_rel.items():
+                merged_parent_rel[k] = .5 * v
+            for k, v in father_rel.items():
+                if k in mother_rel:
+                    merged_parent_rel[k] += .5 * v
+                else:
+                    merged_parent_rel[k] = .5 * v
+
+            if mother in merged_parent_rel:
+                assert father in merged_parent_rel
```
danking, line 100: can you explain to me why this is the case?

hail/python/hail/methods/relatedness/mating_simulation.py
```diff
@@ -1,70 +1,162 @@
...
+
+    if seed is not None:
+        random.seed(seed)
+    for generation in range(n_rounds):
+        last_generation_end = len(samples)
+        mating_generation_size = last_generation_end - last_generation_start_idx
+
+        new_pairs = int(mating_generation_size * pairs_per_generation_multiplier)
+
+        curr_gen_fwd = defaultdict(dict)
```
danking, line 77: what's this for? It gets updated but, afaict, is not returned.

hail/python/hail/methods/relatedness/mating_simulation.py
```diff
@@ -1,70 +1,162 @@
...
+            sibling_rel = merged_parent_rel[mother]  # here mother/father should be the same, since
+
+            for child in range(children_per_pair):
+                samples.append(
+                    (curr_sample_idx, f'generation_{generation + 1}_pair_{pair}_child_{child}', mother, father))
+                for k, v in merged_parent_rel.items():
+                    set_rel(curr_sample_idx, k, v, curr_gen_fwd, last_generation_start_idx)
+
+                mother_sibs = parent_to_child[mother]
+                father_sibs = parent_to_child[father]
```
danking, line 115: nit, but these are more like `mothers_kids` and `fathers_kids`, right? I get that you mean siblings by way o mother, but I think `mothers_kids` is a bit easier to parse.

hail/python/hail/methods/relatedness/mating_simulation.py
```diff
@@ -1,70 +1,162 @@
...
+                for sib in mother_sibs:
+                    if sib in father_sibs:
+                        rel = sibling_rel
+                    else:
+                        _, _, sib_mom, sib_dad = samples[sib]
+                        other_parent = sib_mom if sib_mom != mother else sib_dad
```
danking, line 121: AFICT, there's no sex involved here, otherwise it would obviously be the sib's dad that is the other parent, right? Since we know this sib is the mother's child.

hail/python/test/hail/methods/test_simulation.py
```diff
@@ -8,10 +8,14 @@ def test_mating_simulation():
...
+    assert hl.simulate_random_mating(mt, n_rounds=3, pairs_per_generation_multiplier=1,
+                                     children_per_pair=2).count_cols() == n_samples + n_samples * 2 + n_samples * 4 + n_samples * 8
+    assert hl.simulate_random_mating(mt, n_rounds=2, pairs_per_generation_multiplier=0.5,
+                                     children_per_pair=1).count_cols() == n_samples * 1.75
 
-
-    hl.simulate_random_mating(mt, n_rounds=2, generation_size_multiplier=0.5, keep_founders=True)._force_count_rows()
\ No newline at end of file
+    hl.simulate_random_mating(mt, n_rounds=2, pairs_per_generation_multiplier=0.5,
+                              children_per_pair=2)._force_count_rows()
```
danking, line 21: Can we add a test where we simulate two rounds with just a few people and manually verify the relatednesses?

hail/python/hail/methods/relatedness/mating_simulation.py
```diff
@@ -1,70 +1,162 @@
...
+        last_generation_end = len(samples)
+        mating_generation_size = last_generation_end - last_generation_start_idx
+
+        new_pairs = int(mating_generation_size * pairs_per_generation_multiplier)
+
+        curr_gen_fwd = defaultdict(dict)
+
+        curr_sample_idx = len(samples)
+        parent_to_child = defaultdict(set)
+        for pair in range(new_pairs):
```
danking, line 81: not for this PR, but I think you've basically built a pedigree simulator. We should pull that out as a separately useful thing.

danking (changes requested):

hail/python/hail/expr/functions.py
```diff
@@ -3905,7 +3905,7 @@ def group_by(f: Callable, collection) -> DictExpression:
 @typecheck(f=func_spec(2, expr_any),
            zero=expr_any,
            collection=expr_oneof(expr_set(), expr_array()))
-def fold(f: Callable, zero, collection) -> Expression:
```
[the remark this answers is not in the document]
tpoterba: drive-by fix. Hinting the superclass of the returned value causes lint issues (in my IDE at least), leaving it out and using duck typing doesn't. Can remove if you want, but I've been stripping these out when I see them.

hail/python/hail/methods/relatedness/mating_simulation.py
tpoterba: Correct!

hail/python/hail/ggplot/aes.py
```diff
@@ -44,6 +45,8 @@ def aes(**kwargs):
     hail_field_properties = {}
 
     for k, v in kwargs.items():
+        if isinstance(v, (tuple, dict)):
+            v = hail.str(v)
         if not isinstance(v, Expression):
             v = literal(v)
```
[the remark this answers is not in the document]
tpoterba: I can remove this, I think I added this in the process of trying to build the manhattan plot in ggplot. The issue I was trying to solve is that passing `hl.tuple([something1, something2])` produces a tuple label, but if you pass `(something1, something2)`, you get an exception after it tries to take the literal path. This is a consequence of trying to support both expressions and non-expr sequences. This is sort of a hack, but catches more cases of things that are probably meant to go the expr route.

hail/python/hail/methods/relatedness/mating_simulation.py
tpoterba: the primary relatedness graph stores edges from higher-number sample to lower-number sample. However, we need to be able to look up the edges within a generation for any sample, and that requires a graph in the other direction. We create this graph for each generation and use it in the next, overwriting `last_gen_fwd` each generation.

hail/python/hail/methods/relatedness/mating_simulation.py
tpoterba: Yeah, curious what other applications might be.

hail/python/hail/expr/functions.py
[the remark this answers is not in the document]
danking: Hmm. Pyright complains about our use of `Callable`, but I don't understand the error message [1]. With or without this change, pyright says the type of
```
fold(lambda _: 3, literal(0), literal([1]))
```
is `Expression`. What type do you get?


Hmm, doesn't seem like there's integration for Pyright with IntelliJ unfortunately. The Python community seems to unifying on it since it's remarkably good.

[1]: `Illegal type annotation: variable not allowed unless it is a type alias (lsp)`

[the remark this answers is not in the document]
danking: Ohhhh, I bet this is because your IDE thinks the return type is the return type of the function after modification by `@typecheck`

danking (changes requested):

danking:
Closing abandoned PRs. If someone wants to adopt please start a new PR.

cc: @iris-garden this added som ggplot functionality that would be great to resurrect some day.

PR #12773, opened 2023-03-09
[query] Stage (matrix)table writes locally

ehigham:


hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
```diff
@@ -287,44 +329,25 @@ case class SplitPartitionNativeWriter(
...
       }
 
-      cb += ob1.writeByte(0.asInstanceOf[Byte])
-      cb += ob2.writeByte(0.asInstanceOf[Byte])
-      if (hasIndex)
-        indexWriter.close(cb)
-      cb += ob1.flush()
-      cb += ob2.flush()
-      cb += os1.invoke[Unit]("close")
-      cb += os2.invoke[Unit]("close")
```
ehigham, line 327: All `OutputBuffer` implementations seem to close their underlying output stream so there's no need to maintain a reference to them here

hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
```diff
@@ -210,30 +218,58 @@ case class SplitPartitionNativeWriter(
...
+    }
 
     context.toI(cb).map(cb) { pctx =>
-      val filename1 = mb.newLocal[String]("filename1")
-      val os1 = mb.newLocal[ByteTrackingOutputStream]("write_os1")
-      val ob1 = mb.newLocal[OutputBuffer]("write_ob1")
-      val filename2 = mb.newLocal[String]("filename2")
-      val os2 = mb.newLocal[ByteTrackingOutputStream]("write_os2")
-      val ob2 = mb.newLocal[OutputBuffer]("write_ob2")
-      val n = mb.newLocal[Long]("partition_count")
```
ehigham, line 236: This code duplication was crying out for parameterisation. Since I'm still using locals, there shouldn't be too much difference in the generated code

hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
```diff
@@ -210,30 +218,58 @@ case class SplitPartitionNativeWriter(
...
+                    stream: StreamProducer,
+                    context: EmitCode,
+                    region: Value[Region]
+                   ): IEmitCode = {
     val iAnnotationType = PCanonicalStruct(required = true, "entries_offset" -> PInt64())
     val mb = cb.emb
 
-    val indexWriter = ifIndexed { StagedIndexWriter.withDefaults(index.get._2, mb.ecb, annotationType = iAnnotationType,
-      branchingFactor = Option(mb.ctx.getFlag("index_branching_factor")).map(_.toInt).getOrElse(4096)) }
+    val writeIndexInfo = index.map { case (name, ktype) =>
```
ehigham, line 230: I'm allergic to `null` and have been bitten by them too many times. Keeping things within the `Option` allows us to avoid things like null refs and `ifIndexed`

hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala
```diff
@@ -287,44 +330,25 @@ case class SplitPartitionNativeWriter(
...
       }
 
-      cb += ob1.writeByte(0.asInstanceOf[Byte])
-      cb += ob2.writeByte(0.asInstanceOf[Byte])
-      if (hasIndex)
-        indexWriter.close(cb)
-      cb += ob1.flush()
-      cb += ob2.flush()
-      cb += os1.invoke[Unit]("close")
-      cb += os2.invoke[Unit]("close")
```
ehigham, line 327: There's a slight difference in ordering with respect to the index writer. I didn't think that mattered as the the output buffers and index writer seem to be quite separate.

hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
```diff
@@ -210,30 +213,30 @@ case class PartitionNativeWriter(spec: AbstractTypedCodecSpec, keyFields: Indexe
...
-
   class StreamConsumer(
-    _ctx: SValue,
-    private[this] val cb: EmitCodeBuilder,
-    private[this] val region: Value[Region]
+    ctx: SValue,
+    private val cb: EmitCodeBuilder,
+    private val region: Value[Region]
   ) {
-    private[this] val ctx = _ctx.asString
```
ehigham, line 222: `private[this]` was unnecessary. sorry for the noise

hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
```diff
@@ -654,7 +671,7 @@ case class TableNativeFanoutWriter(
           keyFields,
           s"$targetPath/rows/parts/",
           Some(s"$targetPath/index/" -> keyPType),
-          if (stageLocally) Some(ctx.localTmpdir) else None
+          if (stageLocally) Some(Path.of(ctx.localTmpdir, s"hail_staging_tmp_${UUID.randomUUID()}", "rows", "parts")) else None
```
ehigham, line 674: I've noticed we defer to `String`s for files and paths and so on. I have been bitten by this too, hence prefer to use stronger types to encode at the type level that this is indeed a path, not a generic string. Let me know your thoughts, realise this goes against the grain somewhat but if you like it perhaps we can adopt this stuff more.

ehigham:
Some strangeness going on with build... 
```

------------------------------------------------------------
Gradle 6.8.3
------------------------------------------------------------

Build time:   2021-02-22 16:13:28 UTC
Revision:     9e26b4a9ebb910eaa1b8da8ff8575e514bc61c78

Kotlin:       1.4.20
Groovy:       2.5.12
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          1.8.0_362 (Private Build 25.362-b09)
OS:           Linux 5.4.0-1042-gcp amd64


real	0m3.621s
user	0m4.448s
sys	0m0.623s
+ retry make jars wheel HAIL_DEBUG_MODE=1
BRANCH is set to "HEAD" which is different from old value ""
printf "HEAD" > env/BRANCH
SPARK_VERSION is set to "3.3.0" which is different from old value ""
printf "3.3.0" > env/SPARK_VERSION
echo '[Build Metadata]' > src/main/resources/build-info.properties
echo 'user=' >> src/main/resources/build-info.properties
echo 'revision=e1d86e1908f0911d45b03ef08a694d07e1c4627b' >> src/main/resources/build-info.properties
echo 'branch=HEAD' >> src/main/resources/build-info.properties
echo 'date=2023-03-09T23:23:56Z' >> src/main/resources/build-info.properties
echo 'sparkVersion=3.3.0' >> src/main/resources/build-info.properties
echo 'hailPipVersion=0.2.110' >> src/main/resources/build-info.properties
HAIL_DEBUG_MODE is set to "1" which is different from old value ""
printf "1" > env/HAIL_DEBUG_MODE
ELASTIC_MAJOR_VERSION is set to "7" which is different from old value ""
printf "7" > env/ELASTIC_MAJOR_VERSION
javac -d build/classes/scala/debug -Xlint:all -Werror -XDenableSunApiLintControl -XDignore.symbol.file src/debug/scala/is/hail/annotations/Memory.java
./gradlew shadowJar -Dscala.version=2.12.13 -Dspark.version=3.3.0 -Delasticsearch.major-version=7
Starting a Gradle Daemon (subsequent builds will be faster)
> Task :compileJava NO-SOURCE
> Task :compileScala
[Error] /io/repo/hail/src/main/scala/is/hail/expr/ir/MatrixWriter.scala:122: value of is not a member of object java.nio.file.Path
[Error] /io/repo/hail/src/main/scala/is/hail/expr/ir/TableWriter.scala:57: value of is not a member of object java.nio.file.Path
[Error] /io/repo/hail/src/main/scala/is/hail/expr/ir/TableWriter.scala:674: value of is not a member of object java.nio.file.Path
three errors found
```

hail/src/main/scala/is/hail/expr/ir/TableWriter.scala
chrisvittal: Honestly, I prefer this. We need to take some amount of care for the semantics of cloud storage object names vs filesystem paths, but that is again a way that types come in handy. Thanks for this.

chrisvittal (approved):
Great to finally have this!

PR #12775, opened 2023-03-10
[query] Fail fast when writing local output files with QoB

jigold:
I'm not sure this is the "query-way" of doing these checks, but having this feature would have saved Cal from running an expensive pipeline and the output file "disappearing" at the end.


jigold:
@tpoterba Feel free to assign to someone else. I've been giving too many PRs to Daniel lately and wasn't sure who was best to review this on the compilers side.

tpoterba (approved):

PR #12776, opened 2023-03-10
[compiler] IndexedSeq good, Seq bad

ehigham:


hail/src/main/scala/is/hail/types/TypeWithRequiredness.scala
```diff
@@ -417,15 +417,15 @@ sealed abstract class RBaseStruct extends TypeWithRequiredness {
 }
 
 object RStruct {
-  def apply(fields: Seq[(String, TypeWithRequiredness)]): RStruct =
+  def apply(fields: IndexedSeq[(String, TypeWithRequiredness)])(implicit dummy: DummyImplicit): RStruct =
```
tpoterba, line 420: prefer renaming when we get signature clashes like this. RStruct.fromNamesAndTypes?

hail/src/main/scala/is/hail/types/TypeWithRequiredness.scala
```diff
@@ -435,44 +435,44 @@ case class RStruct(fields: IndexedSeq[RField]) extends RBaseStruct {
 }
 
 object RTuple {
-  def apply(fields: Seq[TypeWithRequiredness]): RTuple =
+  def apply(fields: IndexedSeq[TypeWithRequiredness])(implicit dummy: DummyImplicit): RTuple =
```
tpoterba, line 438: RTuple.fromTypes?

tpoterba (changes requested):
happy days!

one comment about the signature overloading stuff.

tpoterba (approved):

PR #12780, opened 2023-03-14
[batch] Support always_run in local backend

daniel-goldstein:
See [Zulip thread](https://hail.zulipchat.com/#narrow/stream/223457-Hail-Batch-support/topic/.60always_on.60.20for.20local.20backend/near/341620608) for some discussion. The question to whether we can support `always_run` in the local backend depends on whether to make a breaking change in the local backend where a failure in a job does not immediately raise an exception. Will update this PR once consensus is reached in Zulip.


danking:
This seems like a good change to me.

jigold (approved):

PR #12781, opened 2023-03-14
[query][scripts][website][ukbb-rg] gsutil => gcloud

danking:
The only remaining references are in the datasets scripts, but those are meant as references of how we created the files, and in the `hailctl dataproc` command.

I chose not to change the latter because I fear some users might still have ancient versions of `gcloud`.

The reviewer should verify that I got the right version of the `gcloud` command in each case.

Hopefully this resolves the bizarre error we see in test-dataproc.


daniel-goldstein (changes requested):
Commands look good, just some cleanup needed

daniel-goldstein (approved):

PR #12782, opened 2023-03-14
[batch] Unique remote tmpdir path for each split

daniel-goldstein:
Different splits (`test_hailtop_batch_0`, `test_hailtop_batch_1`, …) can clobber each other's tmpdir.


hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -406,7 +406,7 @@ def setUp(self):
         remote_tmpdir = get_user_config().get('batch', 'remote_tmpdir')
         if not remote_tmpdir.endswith('/'):
             remote_tmpdir += '/'
-        self.remote_tmpdir = remote_tmpdir
+        self.remote_tmpdir = remote_tmpdir + str(uuid.uuid4())
```
danking, line 409: the above line suggests this is mean to end with `/`

danking (changes requested):

danking (approved):

PR #12783, opened 2023-03-15
[query] Fix console logging and logging through Python

tpoterba:
After spending a few hours digging through log4j and trying a bunch of approaches, I wasn't able to fix our current approach of adding appenders to the consoleLog after log4j has already been configured. Instead, we set up log4j in initial configuration to have the appenders we want.

Also moved logging config from HailContext to backend, where it should be. 

Storing the StringSocketAppender on the static object is definitely a bit funky, but it's being allocated inside log4j and I don't see a simpler way to store it for retrieval later.

### DIFF
--- a/hail/python/hail/backend/spark_backend.py
+++ b/hail/python/hail/backend/spark_backend.py
@@ -176,16 +176,16 @@ class SparkBackend(Py4JBackend):
 
         if idempotent:
             self._jbackend = hail_package.backend.spark.SparkBackend.getOrCreate(
-                jsc, app_name, master, local, True, min_block_size, tmpdir, local_tmpdir,
+                jsc, app_name, master, local, log, True, append, skip_logging_configuration, min_block_size, tmpdir, local_tmpdir,
                 gcs_requester_pays_project, gcs_requester_pays_buckets)
             self._jhc = hail_package.HailContext.getOrCreate(
-                self._jbackend, log, True, append, branching_factor, skip_logging_configuration, optimizer_iterations)
+                self._jbackend, branching_factor, optimizer_iterations)
         else:
             self._jbackend = hail_package.backend.spark.SparkBackend.apply(
-                jsc, app_name, master, local, True, min_block_size, tmpdir, local_tmpdir,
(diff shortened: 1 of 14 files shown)


hail/python/hail/backend/spark_backend.py
```diff
@@ -216,9 +216,6 @@ def __init__(self, idempotent, sc, spark_conf, app_name, master,
             if self._jsc.uiWebUrl().isDefined():
                 sys.stderr.write('SparkUI available at {}\n'.format(self._jsc.uiWebUrl().get()))
 
-            connect_logger(self._utils_package_object, 'localhost', 12888)
-
-            self._jbackend.startProgressBar()
```
danking, line 221: do we not need this at all?

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -433,7 +433,7 @@ object ServiceBackendSocketAPI2 {
       HailContext.get.backend = backend
       backend.addDefaultReferences()
     } else {
-      HailContext(backend, "hail.log", false, false, 50, skipLoggingConfiguration = true, 3)
+      HailContext(backend, 50, 3)
```
danking, line 436: did you confirm that service backend logs still appear?

danking (dismissed):
looks good one question

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
tpoterba: testing now, but I don't think I changed that since we were passing skip=True before

tpoterba: (pretty sure for service we just pull in the log4j.properties file)

tpoterba: logging works fine in QoB

danking (approved):
Approved *with gusto*.

PR #12784, opened 2023-03-15
[query] Dont use memory service

daniel-goldstein:


danking (approved):
Excellent. Let's release after this.

danking:
Always good to remove `// FIXME: HACK`, heh

PR #12786, opened 2023-03-15
[batch] Add deduped resource id column to attempt resources table

jigold:
This is now the first PR in the deduping resource ids stack. I decided that we should directly write the new "deduped_resource_id" value rather than in a trigger to make things simpler to reason about later when dropping unused columns. The implication of this is this PR must now be fully deployed before #12757. We need to announce this on Zulip once this PR merges.


daniel-goldstein (approved):
I'm all for avoiding more complexity in triggers.

PR #12788, opened 2023-03-15
Remove Jupyter/JupyterLab setup from hailctl

deflaux:
Remove the setup of Jupyter/JupyterLab from hailctl to enable use of Dataproc's [Component Gateway](https://cloud.google.com/dataproc/docs/concepts/accessing/dataproc-gateways) feature, which eliminates the need to use an ssh tunnel to reach the various web UIs on the Dataproc cluster.

To test this pull request:
1. run `hailctl dataproc start` as usual, but add parameters `--enable-component-gateway --optional-components JUPYTER --dry-run` to generate the `gcloud dataproc clusters create` command that will setup JupyterLab and eliminate the need for an ssh tunnel. For example:
    ```
    hailctl dataproc start my-cluster-name \
    --region us-central1 \
    --enable-component-gateway \
    --optional-components JUPYTER \
    --bucket name-of-my-staging-gcs-bucket-where-notebook-files-will-live \
    --temp-bucket name-of-my-gcs-bucket-with-a-lifecycle-rule-to-autodelete-cruft-after-two-weeks \
    --max-idle 60m \
    --dry-run
    ```
2. In the generated `gcloud dataproc clusters create` command, replace the value of `--initialization-actions` with the path of the GCS location to the script in this pull request. Also replace the value of `--temp-bucket`, since hailctl appears to stomp on the user specified value. Then run the command to create your cluster with component gateway enabled.
3. To obtain the URL to JupyterLab, run `gcloud dataproc clusters describe my-cluster-name --region=us-central1 --format="yaml(config.endpointConfig.httpPorts)"`
4. Run Hail notebooks to test the setup of JupyterLab provided by Dataproc!

### DIFF
--- a/hail/python/hailtop/hailctl/dataproc/resources/init_notebook.py
+++ b/hail/python/hailtop/hailctl/dataproc/resources/init_notebook.py
@@ -33,18 +33,14 @@ def mkdir_if_not_exists(path):
 role = get_metadata('dataproc-role')
 
 if role == 'Master':
-    # additional packages to install
+    # Additional packages to install.
+    # None of these packages should be related to Jupyter or JupyterLab because
+    # Dataproc Component Gateway takes care of properly setting up those services.
     pip_pkgs = [
         'setuptools',
         'mkl<2020',
         'lxml<5',
         'https://github.com/hail-is/jgscm/archive/v0.1.13+hail.zip',
-        'ipykernel==4.10.*',
-        'ipywidgets==7.5.*',
-        'jupyter-console==6.0.*',
-        'nbconvert==5.6.*',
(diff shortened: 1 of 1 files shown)


tpoterba:
Having run through a test myself, it looks like we'll actually need to add back all the deleted config aside from the last bit that starts up the jupyter server. Aside from the last block that starts up jupyter, the rest is important logic to configure extensions and content management.

danking:
We have insufficient bandwidth to complete review on this PR at this time.

danking:
We've tracked getting back to this PR at https://github.com/hail-is/hail/issues/14263

PR #12789, opened 2023-03-16
[devbin] updates to rotate_keys.py

danking:
A bunch of stuff:
1. grab disabled status, ignore them in most places, show disabledness when printing
2. Handle the case where a service account has no active key
3. Remove loop confirmation (I can just ctrl-c)

Future work:
We probably do need a bulk update thing but this remove enough annoyance for me for now.


daniel-goldstein (changes requested):
Looks good, just a nit

daniel-goldstein (approved):

PR #12794, opened 2023-03-16
[devbin] Automate cloning db and accessing it with a preconfigured VM

jigold:
This PR creates a clone of a database along with a companion VM where the SSL and MySQL user credentials are setup automatically.

Note that you'll have to run the script multiple times to get the final result as the "clone" step will time out. The script should be idempotent.

### DIFF
--- a/devbin/setup-vm-with-clone-db-access.sh
+++ b/devbin/setup-vm-with-clone-db-access.sh
@@ -12,6 +12,8 @@ git clone https://github.com/hail-is/hail.git
 cd hail
 
 pip3 install -r hail/python/requirements.txt
+pip3 install -r hail/python/dev/requirements.txt
+pip3 install -r hail/python/hailtop/requirements.txt
 
 mkdir /gear/
 chmod 777 /gear/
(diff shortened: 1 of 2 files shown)


danking:
I dismissed them as won’t fix because they’re leaking secrets as intended into config files.

danking:
@daniel-goldstein bump

devbin/setup-vm-with-clone-db-access.sh
```diff
@@ -0,0 +1,32 @@
...
+mkdir /sql-config/
+chmod 777 /sql-config/
+
+git clone https://github.com/hail-is/hail.git
+cd hail
+
+pip3 install -r hail/python/requirements.txt
+pip3 install -r hail/python/dev/requirements.txt
+pip3 install -r hail/python/hailtop/requirements.txt
+pip3 install -r gear/requirements.txt
```
daniel-goldstein, line 17: We try to use `pinned-requirements.txt` files as much as possible for reproducibility. You can instead do

```
pip install \
    -r hail/python/pinned-requirements.txt \
    -r hail/python/dev/pinned-requirements.txt \
    -r gear/pinned-requirements.txt
```

daniel-goldstein (changes requested):

jigold:
I'm going to close this for now and address your comments the next time I need to clone the database.

PR #12797, opened 2023-03-17
[lowering] Fix MatrixBlockMatrixWriter lowering when some partitions are empty

chrisvittal:
Using changePartitionerNoRepartition with a partitioner with a different
number of partitions cannot be correct and will result in dropped data.
Just construct the new partitioner (based on row index, so it will
comply with invariants) directly.

### DIFF
--- a/hail/python/test/hail/linalg/test_linalg.py
+++ b/hail/python/test/hail/linalg/test_linalg.py
@@ -169,6 +169,14 @@ class Tests(unittest.TestCase):
             a4 = hl.eval(BlockMatrix.read(path).to_ndarray())
             self._assert_eq(a1, a4)
 
+    def test_from_entry_expr_empty_parts(self):
+        with hl.TemporaryDirectory(ensure_exists=False) as path:
+            mt = hl.balding_nichols_model(n_populations=5, n_variants=2000, n_samples=20, n_partitions=200)
+            mt = mt.filter_rows((mt.locus.position <= 500) | (mt.locus.position > 1500)).checkpoint(path)
+            bm = BlockMatrix.from_entry_expr(mt.GT.n_alt_alleles())
+            nd = (bm @ bm.T).to_numpy()
+            assert nd.shape == (1000, 1000)
+
     def test_from_entry_expr_options(self):
         def build_mt(a):
             data = [{'v': 0, 's': 0, 'x': a[0]},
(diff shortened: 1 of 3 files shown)


chrisvittal:
resolves https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/.60pc_relate.60.20crash.20with.200.2E2.2E110

ehigham (commented):
Unsolicited feedback - is there no way to test this that doesn't involve an integration test and/or the need to check in so many files? I wonder if this can be captured in a scala unit test?

chrisvittal:
Great question. If I could create a table that could reproduce this bug in pure code, I would.

tpoterba:
shouldn't we be able to create a MatrixTable with empty partitions, write as BM, and trigger this?

tpoterba (changes requested):
This test seems unnecessarily integration-y for something that we should able to test directly. If we have the partition counts of the input to write_from_entry_expr, we should be able to just generate the MT with those counts and trigger the failure directly?

chrisvittal:
I'll try it.

hail/python/test/hail/methods/relatedness/test_pc_relate.py
```diff
@@ -1,7 +1,7 @@
 import hail as hl
 import hail.utils as utils
 
-from ...helpers import resource, skip_when_service_backend
+from ...helpers import resource, skip_when_service_backend, fails_service_backend, fails_local_backend
```
danking, line 4: unnecessary change

chrisvittal: thanks!

hail/python/test/hail/linalg/test_linalg.py
```diff
@@ -169,6 +169,14 @@ def test_from_entry_expr_simple(self):
             a4 = hl.eval(BlockMatrix.read(path).to_ndarray())
             self._assert_eq(a1, a4)
 
+    def test_from_entry_expr_empty_parts(self):
+        with hl.TemporaryDirectory(ensure_exists=False) as path:
+            mt = hl.balding_nichols_model(n_populations=5, n_variants=2000, n_samples=20, n_partitions=200)
+            mt = mt.filter_rows((mt.locus.position <= 500) | (mt.locus.position > 1500)).checkpoint(path)
```
chrisvittal, line 175: unfortunately, this is a bit of a load bearing checkpoint, the issue we've run into doesn't replicate otherwise

PR #12221, opened 2022-09-23
[batch] Add support for job regions

jigold:
CHANGELOG: Added a new method Job.regions() as well as a configurable parameter to the ServiceBackend to specify which cloud regions a job can run in. The default value is a job can run in any available region.

Stacked on #12212 

This PR threads through region requests from the user and feeds that information into the scheduler. The architecture of a pool per machine type has not changed. We explicitly chose not to have a new pool per region x machine_type. Instead, the control loop looks at the front of the job queue and tries to predict which jobs are likely to be scheduled. From those jobs, we then find which regions the jobs can run in and create the number of corresponding instances. We use the fair share calculation to estimate how many jobs per user can be scheduled in 2.5 minutes assuming the scheduling loop runs once per second. We then grab this many jobs from the queue for each user and estimate the "scheduling iteration" at which each iteration of the scheduler each chunk of user jobs would be scheduled. We sort the overall set of jobs that we've chosen by the "scheduling iteration". We also include the regions as part of the sorting queries with None (any region) being sorted last. This is to compact the free cores across jobs so as to avoid fragmentation of instances created and for jobs with no region specifications to fill in the remaining cores in any region.

For the hailtop.batch client, I added a new setting in `~/.config/hail` to set the default regions for all jobs in the ServiceBackend and a new method on `Job` that sets the list of regions to run in.

Things to double check once everything is working is the sort orders on the scheduling queries are correct. 

Once this PR goes in, then we can merge #11840 with some minor changes. There will also be a follow-up PR that gets rid of the CI-specific code in the scheduler.

PR #12799, opened 2023-03-20
[batch] Make ServiceBackend respect environment variable configuration

daniel-goldstein:
Looks like the `get_remote_tmpdir` function was made when the Query ServiceBackend needed this config and didn't want to copy code directly from the batch ServiceBackend. Howevr, the batch ServiceBackend was never changed to use the new functions so got left behind. Aside from deleting a lot of duplicate code, the only change is that now the batch ServiceBackend will pick up the following environment variables `HAIL_BATCH_REMOTE_TMPDIR` and `HAIL_BATCH_BILLING_PROJECT`

### DIFF
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -14,7 +14,7 @@ import warnings
 from rich.progress import track
 
 from hailtop import pip_version
-from hailtop.config import get_deploy_config, get_user_config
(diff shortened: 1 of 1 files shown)


jigold (changes requested):
I vaguely remember wanting to make this change before and Dan pushed back on it. I can try and dig up the PR -- I think it was the original PR that added regions to hailtop.batch.

jigold:
Found it. https://github.com/hail-is/hail/pull/12221#discussion_r988432928

jigold (approved):

PR #12801, opened 2023-03-20
[batch/qob] Ctrl-c when interactively running a batch cancels the batch

daniel-goldstein:
CHANGELOG: Hitting CTRL-C while interactively using Batch or Query-on-Batch cancels the underlying batch.

### DIFF
--- a/hail/python/hail/backend/service_backend.py
+++ b/hail/python/hail/backend/service_backend.py
@@ -443,8 +443,17 @@ class ServiceBackend(Backend):
                         raise reconstructed_error
                     raise reconstructed_error.maybe_user_error(ir)
 
+    def _cancel_on_ctrl_c(self, coro):
+        try:
+            async_to_blocking(coro)
+        except KeyboardInterrupt:
+            if self._batch is not None:
+                print("Received a keyboard interrupt, cancelling the batch...")
+                async_to_blocking(self._batch.cancel())
+            raise
+
     def execute(self, ir: BaseIR, timed: bool = False):
-        return async_to_blocking(self._async_execute(ir, timed=timed))
+        return self._cancel_on_ctrl_c(self._async_execute(ir, timed=timed))
 
     async def _async_execute(self,
                              ir: BaseIR,
@@ -475,7 +484,7 @@ class ServiceBackend(Backend):
         return converted_value
 
     def value_type(self, ir):
-        return async_to_blocking(self._async_value_type(ir))
+        return self._cancel_on_ctrl_c(self._async_value_type(ir))
 
     async def _async_value_type(self, ir, *, progress: Optional[BatchProgressBar] = None):
         async def inputs(infile, _):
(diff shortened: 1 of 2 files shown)


hail/python/hailtop/batch/backend.py
```diff
@@ -794,7 +800,7 @@ async def compile_job(job):
             if verbose:
                 print(f'Waiting for batch {batch_handle.id}...')
             starting_job_id = min(j._client_job.job_id for j in unsubmitted_jobs)
-            status = batch_handle.wait(disable_progress_bar=disable_progress_bar, starting_job=starting_job_id)
+            status = await batch_handle._async_batch.wait(disable_progress_bar=disable_progress_bar, starting_job=starting_job_id)
```
daniel-goldstein, line 803: This change was made because it is both silly to nest two `run_until_complete` and doing so also causes the KeyboardInterrupt stacktrace to be printed twice. This fixes so the stacktrace is only printed once and you don't get complaints about a task not handling the exception.

hail/python/hailtop/batch/backend.py
```diff
@@ -543,8 +543,14 @@ def _run(self,
...
-        return async_to_blocking(
-            self._async_run(batch, dry_run, verbose, delete_scratch_on_exit, wait, open, disable_progress_bar, callback, token, **backend_kwargs))
+        try:
+            return async_to_blocking(
+                self._async_run(batch, dry_run, verbose, delete_scratch_on_exit, wait, open, disable_progress_bar, callback, token, **backend_kwargs))
+        except KeyboardInterrupt:
+            if batch._batch_handle is not None:
+                print("Received a keyboard interrupt, cancelling the batch...")
+                batch._batch_handle.cancel()
+            raise
```
danking, line 553: I tried to make this change a long time ago and got a bunch of push back.

cc: @konradjk ; If you ctrl-c a `hailtop.batch` script that is waiting on a batch (e.g. `b.run(wait=True)`), should the batch be cancelled?

danking: For now, let's just make the QoB change and revisit this after some discussion.

danking (changes requested):

hail/python/hailtop/batch/backend.py
danking: for context: https://hail.zulipchat.com/#narrow/stream/127527-team/topic/0.2E2.2E78/near/257583824

daniel-goldstein: Interesting, I remember the argument but feel like these two behaviors are consistent with each other. Happy to remove the batch behavior for now though.

danking: I completely agree with you but, as the thread suggests, there was divergent opinions. I think people think of Batch as a submission sort of thing. IMHO, if you're trying to submit you should `wait=False`, but 🤷 . Let's take the win and fight this battle separately.

hail/python/hail/backend/service_backend.py
```diff
@@ -443,8 +443,17 @@ async def _rpc(self,
                         raise reconstructed_error
                     raise reconstructed_error.maybe_user_error(ir)
 
+    def _cancel_on_ctrl_c(self, coro):
+        try:
+            async_to_blocking(coro)
```
danking, line 448: `return async...`

danking (dismissed):

hail/python/hail/backend/service_backend.py
danking: also types!

danking (approved):

PR #12805, opened 2023-03-21
[qob] restart JVM if an OOM occurs

danking:
I often observe the JVM getting stuck in a high memory state unable to recover from after an OOM. Best to just kill the JVM.


danking:
Hmm, this might have been on me.

PR #12812, opened 2023-03-22
[query] Add reservoir sample aggregator

tpoterba:
CHANGELOG: Fixed bug where Table/MT._calculate_new_partitions returned unbalanced intervals with whole-stage code generation runtime.


tpoterba:
Suggestion from Patrick: https://en.wikipedia.org/wiki/Combinatorial_number_system could be used to transform the sample into a uniform single integer in `[0, n choose k)` for testing

patrick-schultz (changes requested):

patrick-schultz (approved):

PR #12814, opened 2023-03-22
[query] Extend `func_spec` to support functions with default arguments

ehigham:
Extend `FunctionChecker` to assert that at least the number of non-default parameters are specified, up to the maximum number of parameters.


ehigham:
Merged into https://github.com/hail-is/hail/pull/12779

PR #12815, opened 2023-03-22
[batch] Don't check invariants every 10 seconds in test PRs

jigold:


daniel-goldstein (changes requested):
Looks great, but if these queries can take several seconds we should probably set a generous (30 seconds?) client-side timeout in `test_invariants.py`

daniel-goldstein (approved):

PR #12818, opened 2023-03-24
[ci] Mirror third-party images and hailgenetics images on deploy

daniel-goldstein:
- On *deploys*, makes sure that whatever is in our third-party images is in our private registry before starting builds like hail-ubuntu that might depend on those images. This means that we can update our ubuntu base image without the australians needing to deploy any images by hand. However, this does not run in PRs because I 1) didn't want to add that kind of latency for PRs and 2) we don't do any kind of namespacing for our images so if we did include this for a PR that ultimately wasn't merged we would have to manually remove the image anyway so why not manually add it if you're going to PR it… I think point 2 is a little weak but I recall this being what we agreed on a couple months back when we discussed this. I'm wondering if we should just eat the minute or so latency at the beginning of PRs to be safe but it also feels like a shame for something that changes so infrequently. 

- Again on deploys, upload the hailgenetics/* images to the private registry if they don't already exist there. This way any deployments that aren't hail team's GCP deployment can get these images automatically when they deploy a new SHA instead of uploading them manually. It won't backfill skipped versions, but we decided that was ok. This seems less relevant for testing on PRs as it will get triggered on releases and we can easily dev deploy to rectify the image if this breaks.


danking:
To put it in words: I agree on point one. This doesn't improve or make worse the experience of Hail team devs adding new images. It does improve the experience of third parties tracking our main branch. They no longer need to worry about ensuring the third party image are present.

build.yaml
```diff
@@ -3147,6 +3185,57 @@ steps:
...
+        then
+          echo "$2 already exists, doing nothing"
+        else
+          echo "$2 does not exist, copying $1 to $2"
+          skopeo copy "docker://$1" "docker://$2"
+        fi
+      }
+
+      copy_if_not_present {{ hailgenetics_hail_image.image}} {{ global.docker_prefix }}/hailgenetics/hail:$(cat /io/hail_pip_version)
+      copy_if_not_present {{ hailgenetics_hailtop_image.image}} {{ global.docker_prefix }}/hailgenetics/hailtop:$(cat /io/hail_pip_version)
```
danking, line 3222: So, this is decoupled from the `deploy` step. We can fail to release 0.2.X but still push hailgenetics/hail:0.2.X. That docker image will have a different SHA from the PyPI release. I don't love that.

Moreover, if, say, AUS runs a deploy on a SHA that's not a tagged SHA, they'll get a hailgenetics/hail image with a tag like 0.2.X but whose SHA is not the SHA tagged with 0.2.X. I dislike that.

What do you think of cloning the images tagged with `/io/hail_pip_version` from docker hub, if it exists?

danking (changes requested):

build.yaml
daniel-goldstein: I would much prefer that. I think I went this approach because I was worried about say Azure racing GCP and getting to this step before GCP publishes the image to dockerhub. But ensuring that the same SHAs are in use everywhere seems worth Azure possibly not having the newest dockerhub image for a single deployment. If this is ever an actual issue we could just run a cronjob or webhook or something.

daniel-goldstein: Hm, actually this would apply to GCP too, as we don't want to necessarily condition it on running after the `deploy` step (AUS folks don't run that nor should they).

daniel-goldstein: @danking Wondering if the fact that this will also affect our GCP instance changes your perspective.

danking: Generally, I'd rather get it late than get it wrong.

That said, the `deploy` step already copies into us-docker.pkg.dev/hail-vdc, so I don't think this affects GCP.

danking (changes requested):
Answers in thread 😉

build.yaml
daniel-goldstein: Ah, right, great. I forgot we were keeping it in the `deploy` step. Fine with this then (and agreed on better late than wrong)

danking (approved):
You changed the permissions on copy_image.sh, but that setting is lost during our localize/unlocalize step. It doesn't look like you rely on the file being executable, so that's fine, but why did you make that change?

daniel-goldstein:
That was an accident I meant `mirror_images.sh`. I also came to this same realization but ended up just running it under `bash` instead of doing the `chmod` in the build.yaml step.

PR #12821, opened 2023-03-28
[compiler] Spill to temp storage when repartitioning is too expensive

ehigham:
Repartitioning can be prohibitively expensive when old partitions are recomputed multiple times to compute new partitions. Introduce a selector to estimate how much repartitioning costs relative to writing the table and re-reading partitions from the index based upon average number times old partitions are computed. Through benchmarking, a scaling factor of ~1/60 was a happy compromise. See attached image for top 20 benchmarks.

image


hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
```diff
@@ -1712,13 +1747,56 @@ object LowerTableIR {
...
+   *                     /     |          \
+   *                   *--*  *---*       *--*
+   *          out:     |  |  |   |  ...  |  |
+   *                   *--*  *---*       *--*
+   *
+   * We can estimate the relative cost of computing the new partitions vs
+   * spilling as being proportional to the mean number of old partitions
+   * used to compute new partitions.
+   */
+  def selectRepartitioning[A](original: RVDPartitioner, planned: RVDPartitioner)
```
tpoterba, line 1787: The inclusion of the recompute/reload args in the function seems to me like too much indirection. Can we just have this return a boolean and have the client branch itself?

An example of why I think this design is better is that we also want to gate the decision on an analysis that the consumer actually uses every partitions (no head/tail/filter_partitions) and there's actually no elegant way to do that with the function as-is. With a boolean, it's a simple `&&` in the client site.

tpoterba (changes requested):
one design comment, otherwise looks great. Can you point me to the file (gs path is fine) that includes the latest benchmarks for this branch?

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
ehigham: Happy to convert to a predicate, though you can just pass true and false to this function and achieve the same result.

ehigham:
> one design comment, otherwise looks great. Can you point me to the file (gs path is fine) that includes the latest benchmarks for this branch?

gs://hail-benchmarks-2/ehigham/compute-or-reload-partitions

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
tpoterba: how so? You mean add it that analysis as a parameter here? Sure, but seems the most self-contained to just make this return the decision from partitioning alignment. Let's convert to the boolean predicate if you're ok with that.

ehigham: It's equivalent as we're just encoding boolean algebra with function:
```
selectRepartitioning(old, new)(ifTrue)(ifFalse)
if (selectReparitioning(old, new)(true)(false)) ifTrue else ifFalse
if (selectReparitioning(old, new)(true)(false) && condition) ifTrue else ifFalse
if (selectReparitioning(old, new)(condition)(false)) ifTrue else ifFalse
```

tpoterba (approved):

hail/src/main/scala/is/hail/expr/ir/lowering/LowerTableIR.scala
tpoterba: agree it's semantically equivalent, I just find this one clearest:
```
if (selectReparitioning(old, new)(true)(false) && condition) ifTrue else ifFalse
```

PR #12823, opened 2023-03-28
[query][batch] fix warnings and add -Werror for pytest

danking:
My brain was a little slow today so I decided to fix some easy low hanging fruit.


daniel-goldstein (approved):
Love a good `Werror`

PR #12825, opened 2023-03-29
[query] fix Makefile & upload_qob_jar.sh

danking:
The TOKEN variable is defined in config.mk which is included into hail/Makefile.


tpoterba:
ahh, I see. Didn't see it defined in the makefile and assumed it was an env var from a parent.

tpoterba (approved):

danking:
Ah, I'm actually wrong, it was defined in the Makefile itself originally. We do have a `TOKEN` variable defined in `config.mk` but `config.mk` is only used by services Makefiles. I restored the definition of TOKEN into `upload_qob_jar.sh`

danking:
More issues discovered. Will resolve all and create a new PR.

PR #12826, opened 2023-03-29
[query] fix Makefile & upload_qob_jar.sh

danking:
OK, three issues:

1. The TOKEN envvar needs to be defined.
2. If statement indentation a bit confusing.
3. Hail Batch in the default namespace rejects JAR URLs that are not under `$QUERY_STORAGE_URI/jars`, ergo we need to use QUERY_STORAGE_URI in the default namespace branch. The official JARs have a retention policy and are at the top level (rather than under a `/$(whoami)/`), so there is little risk of corrupting them.


danking:
bump, tiny PR.

PR #12829, opened 2023-03-30
[benchmark] Get benchmarks running in qob again

tpoterba:


daniel-goldstein:
@tpoterba Instead of using `hailtop/requirements.txt` + `hail/requirements.txt` you can use `hail/pinned-requirements.txt`. It includes all the fully-pinned requirements to install hail (it is the union of the hailtop and hail requirements) and is the version of the requirements that we test with.

PR #12831, opened 2023-03-30
[vds] Support split sparse matrixtable to VDS conversion

tpoterba:


hail/python/hail/vds/variant_dataset.py
```diff
@@ -59,14 +59,17 @@ def _variants_path(base: str) -> str:
     def from_merged_representation(mt,
                                    *,
                                    ref_block_fields=(),
-                                   infer_ref_block_fields: bool = True):
+                                   infer_ref_block_fields: bool = True,
+                                   is_split=False):
```
danking, line 63:
```suggestion
                                   is_split: bool = False):
```

danking (changes requested):

danking (approved):
sorry for the latency

PR #12835, opened 2023-03-30
[db] encode our three database flags in TF & add two more

danking:
Both of these were noted by Bernick as a part of his security review.

1. `local_infile`: if this is on, a client could in theory read any file on the instance by loading it into a table.
2. `skip_show_database`: this disables `SHOW DATABASES` by default; we can still grant certain users (e.g. the admin-pod) the `SHOW DATABASES` privilege. A bit of security by obscurity, IMO, but it does not bother me much. I can always see the list of databases via the GCP console.

### DIFF
--- a/infra/gcp/main.tf
+++ b/infra/gcp/main.tf
@@ -237,6 +237,26 @@ resource "google_sql_database_instance" "db" {
       private_network = google_compute_network.default.id
       require_ssl = true
     }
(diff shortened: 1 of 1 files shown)


danking:
I've already made the changes to the DB.

daniel-goldstein (approved):

PR #12838, opened 2023-04-03
[benchmark] Tolerate tests that kill backend and mark as failed

tpoterba:


tpoterba:
bump on this. Still think it's an improvement that hedges against benchmark tests killing the JVM.

ehigham (approved):

ehigham:
OK, but I'm not sure it's the right change to make. Now some jobs will fail silently.
I think the right thing to do would be to change how benchmark jobs are run and always collect results, regardless of job outcome (making it resiliant to some benchmark files not being in their expected locations, which we'll have to do anyway). That way, you'd still see the failures in the batch.

tpoterba:
> OK, but I'm not sure it's the right change to make. Now some jobs will fail silently.

I don't think this lets anything fail silently -- the failure from run() is still tracked. Agree that the right solution is to additionally track information about dead JVMs to try to provide a better error message, but that juice isn't worth the squeeze at the moment.

ehigham:
> the failure from run() is still tracked
Is that true? When we get out-of-memory errors, doesn't `run` just return?

tpoterba:
There's the Python process running the benchmark driver Hail client in Python, and the java process running the hail backend. We want the benchmark driver to return 0 in all cases, with information about whether the hail command finished successfully and how long it took. In current main, this works well for successful runs as well as errors that don't kill the backend (assertion, HailException, etc). It doesn't work for an error that kills the backend (segfault / oom), because while the exception handler in run() works fine, we run stop() unhandled and assume it won't throw (it only throws if the JVM is dead).. In that case, this PR lets the benchmark driver return 0 after writing a file indicating the Hail function failed, just like in an assertion error.

PR #12840, opened 2023-04-03
[benchmark] actually set the env vars

danking:
cc: @ehigham


tpoterba:
oof

danking:
fwiw, `j.env(VAR, VAL)` is also available though that would set it for all the commands in the job.

danking:
(Also, I think you commented instead of approving)

tpoterba (approved):

PR #12841, opened 2023-04-04
[docker] Put Makefile image-building in the top-level Makefile

daniel-goldstein:
Got annoyed with the constant re-tagging of images that don't need to be rebuilt, and decided to play a little make golf along the way.

cc @jigold This should dramatically reduce the number of tags for hail-ubuntu from make-deployed images, though the number of layers in the container registry should not change.


Makefile
```diff
@@ -13,43 +28,17 @@ check-hail:
...
-.PHONY: check-website
-check-website:
-	$(MAKE) -C website check
+check-services: $(CHECK_SERVICES_MODULES)
+
+.PHONY: check-%
+$(CHECK_SERVICES_MODULES): check-%:
+	$(PYTHON) -m flake8  --config setup.cfg $*/$*
+	$(PYTHON) -m pylint --rcfile pylintrc $*/$* --score=n
+	$(PYTHON) -m mypy --config-file setup.cfg $*/$*
```
daniel-goldstein, line 37: Note to self as a followup: We should run these linters not just on `$*/$*` but on `$*` + excluding the `sql` directory

daniel-goldstein:
Turns out we weren't running mypy on auth, gear, monitoring and website, so this triggered a bunch of lints. I fixed them in the last commit.

Makefile
danking: we really just want to exclude extant SQL files, right?

Makefile
```diff
@@ -1,5 +1,20 @@
...
+SERVICES := auth batch ci memory notebook monitoring website
+SERVICES_IMAGES := $(patsubst %, %-image, $(SERVICES))
+SERVICES_MODULES := $(SERVICES) gear web_common
+CHECK_SERVICES_MODULES := $(patsubst %, check-%, $(SERVICES_MODULES))
+
+HAILTOP_VERSION := hail/python/hailtop/hail_version
+SERVICES_IMAGE_DEPS = hail-ubuntu-image $(HAILTOP_VERSION) $(shell git ls-files hail/python/hailtop gear web_common)
+
+EMPTY :=
+SPACE := $(EMPTY) $(EMPTY)
```
danking, line 14: does
```
SPACE := " "
```
not achieve what you're trying to achieve here?

Makefile
```diff
@@ -1,5 +1,20 @@
...
+EXTRA_PYTHONPATH := hail/python:$(subst $(SPACE),:,$(SERVICES_MODULES))
+PYTHON := PYTHONPATH=$${PYTHONPATH:+$${PYTHONPATH}:}$(EXTRA_PYTHONPATH) python3
```
danking, line 16: you're delaying the evaluation of PYTHONPATH until bash-evaluation-time, I don't think that's necessary? Admittedly I don't know immediately how to replicate bash `${X:+$X:}` in make.

Makefile
danking: Do the checkers care if we specify a folder that doesn't exist? Maybe a simple fix is `$*/$* $*/test`?

Makefile
```diff
@@ -113,3 +102,71 @@ generate-pip-lockfiles: auth/pinned-requirements.txt
...
+	echo $(HAIL_UBUNTU_IMAGE) > $@
+
+base-image: hail-ubuntu-image docker/Dockerfile.base
+	$(eval BASE_IMAGE := $(DOCKER_PREFIX)/base:$(TOKEN))
+	python3 ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"'$$(cat hail-ubuntu-image)'"}}' docker/Dockerfile.base docker/Dockerfile.base.out
+	./docker-build.sh . docker/Dockerfile.base.out $(BASE_IMAGE)
+	echo $(BASE_IMAGE) > $@
+
+private-repo-hailgenetics-hail-image: hail-ubuntu-image docker/hailgenetics/hail/Dockerfile $(shell git ls-files hail/src/main hail/python)
+	$(eval PRIVATE_REPO_HAILGENETICS_HAIL_IMAGE := $(DOCKER_PREFIX)/hailgenetics/hail:$(TOKEN))
```
danking, line 122: this is so that you don't have to smoosh the docker-build.sh, rm, and echo into one bash invocation, yeah?

danking (changes requested):
We should eventually lift the deploys up to the top-level too, right?

danking:
v. nice

daniel-goldstein:
> We should eventually lift the deploys up to the top-level too, right?

Ya, I decided just to stop just short of that for this PR for brevity and that I'll probably have to update dev docs (currently `make -C batch deploy` still works)

Makefile
```diff
@@ -1,5 +1,25 @@
...
+
+EMPTY :=
+SPACE := $(EMPTY) $(EMPTY)
+EXTRA_PYTHONPATH := hail/python:$(subst $(SPACE),:,$(SERVICES_MODULES))
+ifeq ($(PYTHONPATH), "")
+PYTHONPATH := $(EXTRA_PYTHONPATH)
+else
+PYTHONPATH := $(PYTHONPATH):$(EXTRA_PYTHONPATH)
+endif
+PYTHON := PYTHONPATH=$(PYTHONPATH) python3
```
danking, line 21: I think a `"$(PYTHONPATH)"` here (for bash's sake) is necessary in case someone has a space in their python path.

Makefile
```diff
@@ -113,3 +107,71 @@ generate-pip-lockfiles: auth/pinned-requirements.txt
 generate-pip-lockfiles: batch/pinned-requirements.txt
 generate-pip-lockfiles: ci/pinned-requirements.txt
 generate-pip-lockfiles: memory/pinned-requirements.txt
+
+$(HAILTOP_VERSION):
+	$(MAKE) -C hail python/hailtop/hail_version
+
+hail-ubuntu-image: docker/hail-ubuntu
```
danking, line 114: Make doesn't do what you'd want when you specify a directory as a dependency. You'll need to use something like `$(shell git ls-files docker/hail-ubuntu)`.

Makefile
```diff
@@ -113,3 +107,71 @@ generate-pip-lockfiles: auth/pinned-requirements.txt
...
+
+website-image: docs
+
+$(SERVICES_IMAGES): %-image: $(SERVICES_IMAGE_DEPS) $(shell git ls-files $$*)
+	$(eval IMAGE := $(DOCKER_PREFIX)/$*:$(TOKEN))
+	python3 ci/jinja2_render.py '{"hail_ubuntu_image":{"image":"'$$(cat hail-ubuntu-image)'"}}' $*/Dockerfile $*/Dockerfile.out
+	./docker-build.sh . $*/Dockerfile.out $(IMAGE)
+	echo $(IMAGE) > $@
+
+ci-utils-image: hail-ubuntu-image $(SERVICES_IMAGE_DEPS) ci/jinja2_render.py ci/wait-for.py ci/create_database.py
```
danking, line 154: I'm inclined to just use ls-files here for ci again. Seems like something we'd forget to fix.

danking (changes requested):
just some nits

Makefile
daniel-goldstein: Ah ya I missed this one

Makefile
```diff
@@ -113,3 +107,71 @@ generate-pip-lockfiles: auth/pinned-requirements.txt
...
+$(HAILTOP_VERSION):
+	$(MAKE) -C hail python/hailtop/hail_version
+
+hail-ubuntu-image: docker/hail-ubuntu
+	$(eval HAIL_UBUNTU_IMAGE := $(DOCKER_PREFIX)/hail-ubuntu:$(TOKEN))
+	python3 ci/jinja2_render.py '{"global":{"docker_prefix":"$(DOCKER_PREFIX)"}}' docker/hail-ubuntu/Dockerfile docker/hail-ubuntu/Dockerfile.out
+	./docker-build.sh docker/hail-ubuntu Dockerfile.out $(HAIL_UBUNTU_IMAGE)
+	echo $(HAIL_UBUNTU_IMAGE) > $@
+
+base-image: hail-ubuntu-image docker/Dockerfile.base
```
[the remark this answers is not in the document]
daniel-goldstein: Ya we'd still need some mapping in the Makefile for FROM images. I think it would be cleaner to use docker build-args and require all of our Dockerfiles start with `ARG BASE_IMAGE \ FROM ${BASE_IMAGE}` instead of jinja templating, but that doesn't solve the parametrization issue.

danking (approved):

PR #12843, opened 2023-04-04
[vds/combiner] Fix attribute assignment on combiner plan load

chrisvittal:


chrisvittal:
mypy on query would be nice

tpoterba (approved):

PR #12813, opened 2023-03-22
[batch] Populate dedup_resource_id in attempt_resources table

jigold:
Stacked on #12757. This is just the migration for attempt_resources. Will add new PRs for the other billing tables when you're happy with this PR. I haven't tested this migration on a clone yet.

PR #12848, opened 2023-04-05
[batch] Populate deduped aggregated batch resources table

jigold:
Stacked on #12813

PR #12849, opened 2023-04-05
[batch] Dedup aggregated job resources table

jigold:
Stacked on #12813

### DIFF
--- a/batch/sql/dedup_job_resources.py
+++ b/batch/sql/dedup_job_resources.py
@@ -128,12 +128,13 @@ FROM (
   LOCK IN SHARE MODE
 ) AS old
 LEFT JOIN (
-  SELECT batch_id, job_id, resource_id, CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
+  SELECT batch_id, job_id, deduped_resource_id, CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
   FROM aggregated_job_resources_v3
+  LEFT JOIN resources ON resources.resource_id = aggregated_job_resources_v3.resource_id
   {where_statement}
-  GROUP BY batch_id, job_id, resource_id
+  GROUP BY batch_id, job_id, deduped_resource_id
   LOCK IN SHARE MODE
-) AS new ON old.batch_id = new.batch_id AND old.job_id = new.job_id AND old.deduped_resource_id = new.resource_id
+) AS new ON old.batch_id = new.batch_id AND old.job_id = new.job_id AND old.deduped_resource_id = new.deduped_resource_id
 WHERE new.`usage` != old.`usage`
 LIMIT 100;
 ''',
--- a/batch/sql/dedup_job_resources.py
+++ b/batch/sql/dedup_job_resources.py
@@ -1,0 +1,212 @@
+import asyncio
+import functools
+import os
+import random
+import time
+from typing import List, Optional, Tuple
+
+from gear import Database, transaction
+from hailtop.utils import bounded_gather
+
+
+MYSQL_CONFIG_FILE = os.environ.get('MYSQL_CONFIG_FILE')
+
+
+class Counter:
+    def __init__(self):
+        self.n = 0
+
+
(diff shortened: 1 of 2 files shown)


jigold:
I'm debating whether this PR and #12848 are necessary right now. It's going to take a lot of effort to run the migrations with the number of rows involved and deduping these tables will not impact any query performance. The goal was to have a consistent set of tables with the billing tables that needed to be deduped.

jigold:
Closing for now until we're happy with the performance of the new v3 tables for billing project queries.

danking:
@jigold is this still stacked?

jigold:
No

danking:
Exciting!

jigold:
@iris-garden Thanks for volunteering to review! Basically, this PR is almost identical to #12848. You could even take the diff between the two migration files in this PR and that one to make sure the only differences are specific to a batch versus a job.

Background: There's 4 billing tables that keep track of aggregated resource usage by `billing_project_user`, `billing_project_user_by_date`, `batches`, and `jobs`. The currently used version of these tables are `aggregated_*_resources_v2` with the emphasis on "v2". We've created new "v3" tables already that are almost identical to "v2" except we're deduplicating resources with the exact same billing rate to save space in the database. We're currently double writing to both the "v2" and "v3" tables. When we added the new "v3" tables, data from all newly created batches since then have been correctly populating both tables. However, we need to back fill the tables with old records. We've currently finished back-populating 3 out of 4 of the "v3" tables. The last one that remains is `aggregated_job_resources_v3`. I estimate the time to back fill this table will be on the order of 12 hours.

How the backfilling works is we have a flag / Boolean Column on the equivalent `v2` table that says whether the rows have been migrated to the `v3` table already. There's then a trigger already in place that every time a row in the `v2` table is updated such as `update aggregated_job_resources_v2 set migrated = 1 where ...`, then we check whether the row has already been "migrated". If it hasn't then we insert new rows into the "v3" table with the deduped resource ID as the key. The exact trigger is 

```sql
DROP TRIGGER IF EXISTS aggregated_job_resources_v2_after_update $$
CREATE TRIGGER aggregated_job_resources_v2_after_update AFTER UPDATE ON aggregated_job_resources_v2
FOR EACH ROW
BEGIN
  DECLARE new_deduped_resource_id INT;

  IF OLD.migrated = 0 AND NEW.migrated = 1 THEN
    SELECT deduped_resource_id INTO new_deduped_resource_id FROM resources WHERE resource_id = OLD.resource_id;

    INSERT INTO aggregated_job_resources_v3 (batch_id, job_id, resource_id, `usage`)
    VALUES (NEW.batch_id, NEW.job_id, new_deduped_resource_id, NEW.usage)
    ON DUPLICATE KEY UPDATE
      `usage` = `usage` + NEW.usage;
  END IF;
END $$
```

What this PR does is find the keys of all rows in the `aggregated_jobs_resources_v2` table in intervals of 100 rows. This is a "chunk". The reason is because we want to keep the transactions small and fast. I optimized this and found 100 rows worked best for performance. We then want to set `migrated=1` for all rows in the given chunk which activates the trigger and also maintains idempotency so we only run the update for each chunk once. 

Most of the code in this PR is identifying the bounds of each chunk and then doing the update. We have a burn-in period at the beginning where we migrate chunks serially. Then we migrate the chunks in 10-way parallel. This is to get rid of deadlock errors due to row locks with the "birthday problem".

Lastly, once all of the updates are complete, we run an audit that makes sure the "v2" and "v3" tables are equivalent and have the same total aggregate resource usage. I believe I also run this audit in chunks here as these tables are massive and a single audit query would take hours. The bounds of the audit for these chunks are on the order of `(batch_id, job_id)` rather than `(batch_id, job_id, resource_id)` which was used for the actual updates. This is because the resource_ids can differ between "v2" and "v3", so we just check the overall job adds up to the same usage after deduplicating the resource IDs on both tables.

I recommend looking at the main function towards the bottom of the script and then working your way through it.

iris-garden (approved):
lgtm! diff'd with `batch/sql/dedup_agg_batch_resources.py` and this looks like it should do the same thing, but for jobs instead of batches.

PR #12856, opened 2023-04-07
[benchmark] add `visualize` + refactor hail-bench

ehigham:
I wanted to add a tool that can compare more than two benchmark results and hacked this together today.
I decided to use a means of "registering" sub-programs with the main parser to eliminate manual arg parsing.
Make sure you clean and build before using this otherwise you may get collisions with stale package directories.
```bash
$ hail-bench visualize --help
usage: hail-bench visualize [-h] [--metric {mean,median,stdev,max_memory}] [--head HEAD] [--abs] baseline runs [runs ...]

Visualize benchmark results

positional arguments:
  baseline              baseline benchmark results
  runs                  benchmarks to compare against baseline

options:
  -h, --help            show this help message and exit
  --metric {mean,median,stdev,max_memory}
  --head HEAD           number of most significant results to take
  --abs                 plot absolute differences
```
image


tpoterba:
Woah, this is awesome!!!

danking:
Love this! I think this graph would be easier to reason about if you made all the bars the positive percentage runtime relative to the mean. Then add a gray line at 100% to indicate whether you're faster or slower. In this representation, the relative sizes of the bar are easy to see and meaningful (if bar 1 is half the size of bar 2, then option 1 is 2x as fast).

ehigham:
> Love this! I think this graph would be easier to reason about if you made all the bars the positive percentage runtime (i.e. runtime_i / mean(runtime)). Then add a gray line at 100% to indicate whether you're faster or slower. In this representation, the relative sizes of the bar are easy to see and meaningful (if bar 1 is half the size of bar 2, then option 1 is 2x as fast).

I've taken a different approach. I've made the first argument the baseline against which all other runs are compared. the baseline result isn't plotted anymore (unless you don't give any other arguments). I think the plot is a lot clearer now. I also think negative values make more sense as we're comparing runs directly against a baseline. Let me know what you think!

tpoterba (approved):
👍 

Super awesome! sorry for the slow review.

PR #12857, opened 2023-04-07
[query] reduce wall time for local, scala, and spark tests

danking:
In practice, scalac appears to, at best, use 2.5 CPUs. The `build_hail` task compiles Hail twice, once in debug mode and once not. For reasons unclear to me, the second compilation does not appear to be able to reuse class files from the first one because it takes almost as long! This change runs the two in separate jobs in parallel. This should save ~5 minutes from the local, scala, and spark paths which have recently become slower, in wall-time terms, than service backend.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -661,7 +661,7 @@ steps:
       valueFrom: hail_build_image.image
     resources:
       memory: standard
-      cpu: '4'
+      cpu: '2'
     script: |
       set -ex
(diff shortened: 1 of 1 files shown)


danking:
CPU use evidence: https://batch.hail.is/batches/7233266/jobs/34

danking:
needs more thought

PR #12860, opened 2023-04-10
[query] Fix memory leak in buffered aggregation / shuffles on QoB

tpoterba:


tpoterba:
oh no.

daniel-goldstein:
@danking This was never actually approved

danking (approved):

PR #12862, opened 2023-04-10
[batch] Clean up environment variables for easier local execution of batch tests

daniel-goldstein:
Trying to make it more ergonomic to simply do `python3 -m pytest batch/test/test_batch.py::test_job` (now works without any extra environment variables or configuration).

This involved the following changes:
- Deleted of some env vars that are no longer used / can be easily consolidated into existing ones
- Gave defaults to those testing env variables for which there are reasonable defaults. E.g. `DOCKER_ROOT_IMAGE` and `HAIL_GENETICS_HAIL_IMAGE`.
- Pushed other environment variables for which there are not reasonable defaults into the tests that need them. If you run a test that requires `HAIL_CLOUD`, you'll still get an error that that env variable is unset and you should set it. But, if you just want to run a single test that doesn't need `HAIL_CLOUD` it won't get in the way.


danking (approved):
This is great!

PR #12864, opened 2023-04-10
[query] Refactor and fix Backend's persist and unpersist 

chrisvittal:
- There is now one persist method that accepts any of a Table,
  MatrixTable, or BlockMatrix.
- Backend.unpersist was incorrect with respect to unpersist docs, it was
  not accepting the persisted dataset, or returning the unpersisted
  dataset, or doing nothing if the dataset wasn't persisted.

### DIFF
--- a/hail/python/hail/backend/backend.py
+++ b/hail/python/hail/backend/backend.py
@@ -1,4 +1,4 @@
-from typing import Mapping, List, Union, Tuple, Dict, Optional, Any, AbstractSet
+from typing import Mapping, List, Union, TypeVar, Tuple, Dict, Optional, Any, AbstractSet
 import abc
 import orjson
 import pkg_resources
@@ -6,12 +6,18 @@ import zipfile
 
 from hailtop.config.user_config import configuration_of
 
-from ..fs.fs import FS
 from ..builtin_references import BUILTIN_REFERENCE_RESOURCE_PATHS
 from ..expr import Expression
 from ..expr.types import HailType
+from ..fs.fs import FS
 from ..ir import BaseIR
+from ..linalg.blockmatrix import BlockMatrix
+from ..matrixtable import MatrixTable
+from ..table import Table
 from ..utils.java import FatalError
+
+
+Dataset = TypeVar('Dataset', Table, MatrixTable, BlockMatrix)
 
 
 def fatal_error_from_java_error_triplet(short_message, expanded_message, error_id):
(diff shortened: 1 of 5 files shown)


ehigham (changes requested):
This is indeed consistent, though the code is more-or-less copy and paste. Could you refactor to eliminate the code duplication?

Another (perhaps minor) issue I have is the use of exceptions for control-flow in the `unpersist_` method and that the keys are not removed when they're unpersisted. Perhaps restructure to use `pop`, for example:
```python3
persisted = self._persisted_locations.pop(t_or_bm, None)

if not persisted:
    raise ValueError(f'{t_or_bm} was not persisted.')

return persisted.__exit__(None, None, None)
```

danking:
Drive by comment here as I review PRs, but this is The Python Way, for better or worse. People call it "easer to ask forgiveness than permission". I don't personally have strong feelings one way or the other but we do this in many places in our codebase.

ehigham:
> Drive by comment here as I review PRs, but this is The Python Way, for better or worse. People call it "easer to ask forgiveness than permission". I don't personally have strong feelings one way or the other but we do this in many places in our codebase.

Be that as it may, throwing if something does not exist is clearer and more robust that catching an error because some operation failed for what is arguably unexceptional, and just hoping that the `ValueError` that that operation threw is because the item did not exist, rather than another `ValueError` risen in any subsequent work.

ehigham:
> I agree this is the most compelling critique. I think the Pythonistas would make two points: a) KeyError is the one specific error you get in this case and b) it should be written like this:
> 
> ```python
> try:
>     persisted_bm = self._persisted_locations[bm]
> except KeyError as err:
>     raise ValueError(f'{bm} is not persisted') from err
> persisted_bm.__exit__(None, None, None)
> ```

Updated my comment - I think my argument still holds with KeyError

danking:
Agreed the lack of a pop might be a legitimate bug.

chrisvittal:
All of this is wrong compared to the docs anyways.
```python3
    def unpersist(self) -> 'MatrixTable':
        """
        Unpersists this dataset from memory/disk.

        Notes
        -----
        This function will have no effect on a dataset that was not previously
        persisted.

        Returns
        -------
        :class:`.MatrixTable`
            Unpersisted dataset.
        """
        return Env.backend().unpersist(self)
```

Whereas, we currently return `None` from `unpersist` and raise if the dataset was not persisted. This is blatantly wrong, and I'm going to fix that.

ehigham (approved):
Nice! Thanks for making the changes

PR #12868, opened 2023-04-10
[qob] retry `storage.writer` and `storage.reader`

danking:
I do not think we frequently get errors in `storage.reader`, but I think `storage.writer` was always flaky and we were protected by the `retryTransientErrors` on `createNoCompression`. My change to fix requester pays delayed the error until either the first `write` or the `close` which do not have a `retryTransientErrors` (and it is not obvious to me that it is safe to retry a `flush`).

PR #12867, opened 2023-04-10
[qob] retry transient errors reading the results file

danking:

PR #12869, opened 2023-04-10
[qob] Fix new transiencies

danking:
Merge of https://github.com/hail-is/hail/pull/12868 and https://github.com/hail-is/hail/pull/12867 because I expect each is likely to fail due to the other's error.

---

    [qob] retry `storage.writer` and `storage.reader`
    
    I do not think we frequently get errors in `storage.reader`, but I think `storage.writer` was
    always flaky and we were protected by the `retryTransientErrors` on `createNoCompression`. My
    change to fix requester pays delayed the error until either the first `write` or the `close`
    which do not have a `retryTransientErrors` (and it is not obvious to me that it is safe to retry
    a `flush`).

---

    [qob] retry transient errors reading the results file

---

### DIFF
--- a/hail/python/hail/backend/service_backend.py
+++ b/hail/python/hail/backend/service_backend.py
@@ -435,14 +435,7 @@ class ServiceBackend(Backend):
         async with driver_output as outfile:
             success = await read_bool(outfile)
             if success:
-                result_bytes = await read_bytes(outfile)
-                try:
-                    return result_bytes
-                except orjson.JSONDecodeError as err:
-                    raise FatalError('Hail internal error. Please contact the Hail team and provide the following information.\n\n' + yamlx.dump({
-                        'service_backend_debug_info': self.debug_info(),
-                        'batch_debug_info': await self._batch.debug_info()
-                    })) from err
+                return await read_bytes(outfile)
 
             short_message = await read_str(outfile)
             expanded_message = await read_str(outfile)
--- a/hail/python/hail/backend/service_backend.py
+++ b/hail/python/hail/backend/service_backend.py
@@ -18,7 +18,7 @@ from hail.ir.renderer import CSERenderer
 
 from hailtop import yamlx
 from hailtop.config import (configuration_of, get_remote_tmpdir)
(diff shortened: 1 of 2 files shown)


hail/python/hail/backend/service_backend.py
```diff
@@ -418,34 +418,40 @@ async def _rpc(self,
                     raise
 
             with timings.step("read output"):
```
danking, line 420: This diff is easier to read with whitespace differences disabled.

hail/python/hail/backend/service_backend.py
```diff
@@ -418,34 +418,40 @@ async def _rpc(self,
...
+
+        async with driver_output as outfile:
+            success = await read_bool(outfile)
+            if success:
+                result_bytes = await read_bytes(outfile)
                 try:
-                    driver_output = await self._async_fs.open(iodir + '/out')
-                except FileNotFoundError as exc:
+                    return result_bytes
+                except orjson.JSONDecodeError as err:
```
daniel-goldstein, line 441: Shouldn't the `read_bytes` line go in the try? Also, is the `orjson.JSONDecodeError` for if the HTTP response body is somehow malformed? How would that happen?

hail/python/hail/backend/service_backend.py
ehigham: much

hail/python/hail/backend/service_backend.py
```diff
@@ -418,34 +418,40 @@ async def _rpc(self,
...
+            })) from exc
+
+        async with driver_output as outfile:
+            success = await read_bool(outfile)
+            if success:
+                result_bytes = await read_bytes(outfile)
                 try:
-                    driver_output = await self._async_fs.open(iodir + '/out')
-                except FileNotFoundError as exc:
+                    return result_bytes
```
ehigham, line 440: can returning a value throw?

daniel-goldstein: It cannot

daniel-goldstein (approved):

PR #12877, opened 2023-04-11
Feature/sas token merge

gregsmi:
This PR is to enable `hail-az;` file references to contain SAS tokens to enable bearer-auth style file access to Azure storage. Basic summary of the changes:
	- Update `AzureAsyncFS` url parsing function to look for and separate out a SAS-token-like query string. Note: made fairly specific to SAS tokens - generic detection of query string syntax interferes with glob support and '?' characters in file names
	- Added `generate_sas_token` convenience function to `AzureAsyncFS`. Adds new `azure-mgmt-storage` package requirement.
	- Updated `AzureAsyncFS` to use `(account, credential)` tuple as internal `BlobServiceClient` cache key
	- Updated `AzureAsyncFSURL` and `AzureFileListEntry` to track the token separately from the name, and extend the base classes to allow returning url with or without a token 
	- Update `RouterFS.ls` function and associated `listfiles` function to allow for trailing query strings during path traversal 
	- Change to existing behavior: `LocalAsyncFSURL.__str__`no longer returns 'file:' prefix. Done to make `str()` output be appropriate for input to `fs` functions across all subclasses
	- Updated `inter_cloud/test_fs.py` to generically use query-string-friendly file path building functions
- Updated InputResource to not include the SAS token as part of the destination file name 

`test_fs.py` has been updated to respect the new model, where it is no longer safe to extend URLs by just appending new segments with + "/" because there may be a query string. But actually running those tests for the SAS case will require some new test variables to allow the test code to generate SAS tokens (`build.yaml/test_hail_python_fs`): 
```
      export HAIL_TEST_AZURE_ACCOUNT=hailtest
      export HAIL_TEST_AZURE_CONTAINER=hail-test-4nxei
      # Required for SAS testing on Azure
      export HAIL_TEST_AZURE_RESGRP=hailms02
      export HAIL_TEST_AZURE_SUBID=12ab51c6-da79-4a99-8dec-3d2decc97343
```
So the SAS case is disabled for now (`test_fs.py`):
```
@pytest.fixture(params=['file', 'gs', 's3', 'hail-az', 'router/file', 'router/gs', 'router/s3', 'router/hail-az']) # 'sas/hail-az'
```


hail/python/hailtop/aiocloud/aioazure/fs.py
```diff
@@ -398,10 +450,10 @@ async def isfile(self, url: str) -> bool:
         return await self.get_blob_client(url).exists()
 
     async def isdir(self, url: str) -> bool:
-        _, _, name = self.get_account_container_and_name(url)
-        assert not name or name.endswith('/'), name
+        _, _, name, _ = self.get_url_parts(url)
+        dir_name = name if not name or name.endswith('/') else name + '/'
```
daniel-goldstein, line 454: huh, what was here before looks like a bad assert. Is this a general bugfix or did you run into this specifically because of sas tokens?

hail/python/hailtop/aiocloud/aioazure/fs.py
```diff
@@ -500,7 +552,7 @@ async def close(self) -> None:
             self._credential = None
 
         if self._blob_service_clients:
-            await asyncio.wait([client.close() for client in self._blob_service_clients.values()])
+            await asyncio.wait([asyncio.create_task(client.close()) for client in self._blob_service_clients.values()])
```
daniel-goldstein, line 555: How come you've wrapped this in a task?

hail/python/hailtop/aiotools/fs/fs.py
```diff
@@ -46,9 +46,11 @@ def name(self) -> str:
...
         pass
 
-    @abc.abstractmethod
-    def url_maybe_trailing_slash(self) -> str:
-        pass
+    async def url_maybe_trailing_slash(self) -> str:
+        return await self.url()
+
+    async def url_with_query(self) -> str:
+        return await self.url()
```
daniel-goldstein, line 53: Huh, I feel like all three of these methods should not be `async`. Instead of making these async can you make `url` sync?

hail/python/hailtop/batch/batch.py
```diff
@@ -396,11 +397,12 @@ def _new_job_resource_file(self, source, value=None):
...
 
-    def _new_input_resource_file(self, input_path, value=None):
+    def _new_input_resource_file(self, input_path, root=None):
         self._backend.validate_file_scheme(input_path)
-        if value is None:
-            value = f'{secret_alnum_string(5)}/{os.path.basename(input_path.rstrip("/"))}'
-        irf = _resource.InputResourceFile(value)
+        root = root or secret_alnum_string(5)
+        # Take care not to include a SAS token query string in the local name.
+        file_name, _ = AzureAsyncFS.get_name_parts(input_path.rstrip("/"))
```
daniel-goldstein, line 404: For clarity, could we do something like (not married to the names)

```python
if RouterAsyncFS.blob_cloud(input_path) == 'azure':
    cleaned_path = AzureAsyncFS.strip_sas_token(input_path)
```

daniel-goldstein (changes requested):
Thanks @gregsmi for the contribution, this is great! I only got a brief chance to review so I still have to look at the tests but the implementation seems great. Just left a couple of style nits.

daniel-goldstein:
Sorry for leaving you with a mess of conflicts @gregsmi, but hopefully they're all pretty superficial stuff.

daniel-goldstein:
Also, you can use `haildev` for the RG and `22cd45fe-f996-4c51-af67-ef329d977519` for the subscription id for these tests. The changes to the tests you made seem totally reasonable.

hail/python/hailtop/aiocloud/aioazure/fs.py
```diff
@@ -346,21 +383,36 @@ def get_account_container_and_name(url: str) -> Tuple[str, str, str]:
...
-            self._blob_service_clients[account] = BlobServiceClient(f'https://{account}.blob.core.windows.net', credential=self._credential)
-        return self._blob_service_clients[account]
+    @staticmethod
+    def get_name_parts(name: str) -> Tuple[str, str]:
+        # Look for a terminating SAS token.
+        query_index = name.rfind('?')
+        if query_index != -1:
+            query_string = name[query_index + 1:]
+            # We will accept it as a token string if it begins with at least 1 key-value pair of the form 'k=v'.
+            if len(list(filter(str.strip, query_string.split('&')[0].split('=')))) == 2:
```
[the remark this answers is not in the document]
gregsmi: fixed

hail/python/hailtop/aiocloud/aioazure/fs.py
gregsmi: seemed like a bug - I hit it while doing some ad hoc testing

hail/python/hailtop/aiocloud/aioazure/fs.py
gregsmi: asyncio.wait needs a task - looks like Jackie already made this change upstream

hail/python/hailtop/batch/batch.py
gregsmi: added the cloud type check - would prefer to keep the `get_name_parts` signature as-is since there are other callers that need both parts, but let me know if you think it's better to have an additional wrapper method on `AzureAsyncFS`

gregsmi:
superseded by new [PR 13140](https://github.com/hail-is/hail/pull/13140)

PR #12878, opened 2023-04-11
[hailtop] our UnexpectedEOFError can be transient

danking:
I do not know how we have gone this long without handling this properly.


danking:
Hmm. That's not a good example because IncompleteReadError is not transient and I'm not sure if it should be. If a file was written with too few bytes, that's not transient, that's a bug that'll never get fixed.

PR #12880, opened 2023-04-12
[vds] Add function to patch old VDSes to store max reference block length

tpoterba:
CHANGELOG: Added `hl.vds.store_ref_block_max_len` to patch old VDSes to make interval filtering faster.


tpoterba:
cc @chrisvittal who might also want to look at this

danking (approved):
Seems fine to me

PR #12881, opened 2023-04-13
[hailtop.batch] Support nested resource files in PythonJobs

jigold:
CHANGELOG: Support nested resource files in PythonJobs.

### DIFF
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -1097,13 +1097,18 @@ class PythonJob(Job):
 
             self._mentioned.add(r)
 
-        for arg in args:
-            if isinstance(arg, _resource.Resource):
-                handle_arg(arg)
+        def handle_args(r):
+            if isinstance(r, _resource.Resource):
+                handle_arg(r)
+            elif isinstance(r, list):
+                for elt in r:
+                    handle_arg(elt)
+            elif isinstance(r, dict):
+                for v in r.values():
+                    handle_arg(v)
 
-        for value in kwargs.values():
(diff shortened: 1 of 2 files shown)


hail/python/hailtop/batch/job.py
```diff
@@ -1155,21 +1164,27 @@ def prepare_argument_for_serialization(arg):
...
+    if typ == 'value' and isinstance(val, (list, tuple)):
+        return [deserialize_argument(elt) for elt in val]
     if typ == 'py_path':
         return dill.load(open(val, 'rb'))
     if typ in ('path', 'dict_path'):
         return val
     assert typ == 'value'
     return val
 
+    return _deserialize(arg)
```
daniel-goldstein, line 1178: I think this is dead code

hail/python/test/hailtop/batch/test_batch.py
```diff
@@ -1229,3 +1229,45 @@ def test_update_batch_from_batch_id(self):
...
         res_status = res.status()
         assert res_status['state'] == 'success', str((res_status, res.debug_info()))
+
+    def test_list_recursive_resource_extraction_in_python_jobs(self):
+        b = self.batch(default_python_image=PYTHON_DILL_IMAGE)
+
+        def write(paths):
+            for path in paths:
+                with open(path, 'w') as f:
+                    f.write('foo')
```
daniel-goldstein, line 1239: nit: can we do `for i, path in enumerate(paths)` and then write `i` to the file so that the files' contents aren't the same?

daniel-goldstein (dismissed):
This is sweet, just a couple nits

PR #12882, opened 2023-04-13
[terraform] align gcp terraform with current state & create new terraform project to track true current state

danking:
I fixed some issues in `gcp` but there are enough difficult to resolve issues that I want to start a fresh terraform project to track reality as different from what we recommend a new user do.

The fresh attempt project is `gcp-broad`. I plan to slowly expand that until it captures everything. From this point forward, for the subset of infrastructure in `gcp-broad`, we should not make manual changes. Instead, we should propose changes as PRs, review them, merge them, and apply the terraform by hand once the changes are in `main`.


danking:
The terraform plan for `gcp-broad` has 4 changes after appropriately importing.

The bucket, ci_config, and hail_ci_0_1_github_oauth_token are actually unchanged. I do not know why they are listed as a change.

That secret currently has three files: hail-ci-0-1.key, oauth-token, user1, user2. AFAICT, only `user1` is used, so the new secret only has `user1`.

```
  # module.ci[0].google_storage_bucket.bucket will be updated in-place
  ~ resource "google_storage_bucket" "bucket" {
        id                          = "hail-ci-bpk3h"
      # Warning: this attribute value will be marked as sensitive and will not
      # display in UI output after applying this change. The value is unchanged.
      ~ location                    = (sensitive value)
        name                        = "hail-ci-bpk3h"
      # Warning: this attribute value will be marked as sensitive and will not
      # display in UI output after applying this change. The value is unchanged.
      ~ storage_class               = (sensitive value)
        # (8 unchanged attributes hidden)

        # (2 unchanged blocks hidden)
    }

  # module.ci[0].kubernetes_secret.ci_config will be updated in-place
  ~ resource "kubernetes_secret" "ci_config" {
        id        = "default/ci-config"
        # (3 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

  # module.ci[0].kubernetes_secret.hail_ci_0_1_github_oauth_token will be updated in-place
  ~ resource "kubernetes_secret" "hail_ci_0_1_github_oauth_token" {
        id        = "default/hail-ci-0-1-github-oauth-token"
        # (3 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

  # module.ci[0].kubernetes_secret.hail_ci_0_1_service_account_key will be updated in-place
  ~ resource "kubernetes_secret" "hail_ci_0_1_service_account_key" {
      ~ data      = (sensitive value)
        id        = "default/hail-ci-0-1-service-account-key"
        # (2 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```

infra/gcp-broad/bootstrap.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
daniel-goldstein, line 1: I don't think this file needs to be duplicated

infra/gcp/gcs_bucket/main.tf
```diff
@@ -1,9 +1,9 @@
-resource "random_id" "bucket_prefix" {
-  byte_length = 5
+resource "random_string" "bucket_suffix" {
+  length = 5
 }
 
 resource "google_storage_bucket" "bucket" {
-  name = "${var.short_name}-${random_id.bucket_prefix.hex}"
+  name = "${var.short_name}-${random_string.bucket_suffix.result}"
```
daniel-goldstein, line 6: This will involve some manual state management by the AUS folks. Probably easy, but we should maybe ask them/give them a heads up. This will try to delete and recreate their buckets if they don't manually create the `random_string` resource to match what their bucket is named.

infra/gcp/gsa_k8s_secret/main.tf
```diff
@@ -1,9 +1,21 @@
...
+      lower,
+    ]
+  }
 }
 
 resource "google_service_account" "service_account" {
-  account_id = "${var.name}-${random_id.name_suffix.hex}"
+  display_name = "${var.name}-${random_string.name_suffix.result}"
+  account_id = "${var.name}-${random_string.name_suffix.result}"
+  timeouts {}
```
daniel-goldstein, line 18: Same here about AUS importing.

infra/gcp/main.tf
```diff
@@ -209,20 +275,24 @@ resource "google_compute_global_address" "google_managed_services_default" {
 
 resource "google_service_networking_connection" "private_vpc_connection" {
   network = google_compute_network.default.id
-  service = "servicenetworking.googleapis.com"
-  reserved_peering_ranges = [google_compute_global_address.google_managed_services_default.name]
+  service = "services/servicenetworking.googleapis.com"
+  reserved_peering_ranges = [
+    "jg-test-clone-resource-id-ip-range",
```
daniel-goldstein, line 280: Can we make a note somewhere to remove this when Jackie no longer needs this. Or some lifecycle ignore where there can be more ranges added by hand?

infra/gcp/main.tf
```diff
@@ -237,37 +307,70 @@ resource "google_sql_database_instance" "db" {
...
+      hour = 16
+    }
   }
+
+  timeouts {}
 }
 
 resource "google_compute_address" "gateway" {
-  name = "gateway"
+  name = "site"
```
daniel-goldstein, line 367: Why this name change?

infra/gcp/main.tf
```diff
@@ -592,30 +679,54 @@ module "hail_query" {
...
 
   private_visibility_config {
     networks {
-      network_url = google_compute_network.default.id
+      network_url = google_compute_network.default.self_link
     }
   }
+
+  timeouts {}
 }
```
daniel-goldstein, line 723: For another PR, but this DNS zone and accompanying recordset is unused and can be deleted.

daniel-goldstein (changes requested):
Could you describe what the migration path would be to ultimately move off of this `gcp-broad` folder? Can we basically at some point point `infra/gcp` to our terraform state from `gcp-broad` and see that the plan is clean, then `rm -r gcp-broad`?

Also, it would be nice to set up remote state tracking so that others can apply this terraform before this change goes in.

infra/gcp/gcs_bucket/main.tf
danking: Agreed, worth an email. I found the random_id resource really difficult to import as. If you can give me the incantation that would correctly import our extant buckets, then I'd be happy to keep as is.

infra/gcp/main.tf
danking: `jg-test-clone-resource-id-ip-range` is currently not in used by anything.

It appears that db-jcaq4 was created in some special (perhaps outdated?) way. It doesn't have a range associated. 

It seems to me that the right fix is this:
1. Delete `jg-test-clone-resource-id-ip-range`.
2. Add `cloud-sql-ip-range`
3. Make `cloud-sql-ip-range` the reserved peering range for this VPC peering.

Do you agree? If yeah, I can talk to Jackie Monday and make sure we don't interrupt her workflow.

infra/gcp/main.tf
danking: That's the actual name of this address. see `gcloud compute addresses list`

danking (commented):
OK, here's a proposal:

Let's finalize the outstanding issues in this PR, then:
1. Merge it.
2. Import our infra
3. Store the state securely remotely
4. Migrate infra as much as possible so that it looks more like `infra/gcp/` (e.g. recreate the address with the right name and update our DNS accordingly)
5. Flip over from gcp-broad to gap and require a re-import from our downstream users but hopefully a small import thanks to 4.

infra/gcp/gcs_bucket/main.tf
daniel-goldstein: Have you tried [this](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/id#import)? Seems to suggest you could throw it through base64

infra/gcp/main.tf
daniel-goldstein: I think I'm confused as to why this change is in `infra/gcp/main.tf` instead of us having a different resource in `broad-gcp` and then trying to migrate that one to match what's here.

daniel-goldstein (changes requested):
Comments in previous threads. Still a bit wary about the amount of destroy-requiring changes in `infra/gcp`, but otherwise am fine with this approach.

infra/gcp/gcs_bucket/main.tf
danking: OK, so, this particular random id is used for buckets of which we only have two: hail_query and hail_batch. In both cases, we created them with the names `hail-query` and `hail-batch`. Those are not importable in either setting. I'll revert this change so that we don't change the gcp folder as much, but we must eventually confront the fact that our naming scheme and the one described in the gcp terraform folder are incompatible.

danking:
I reverted the `gcp` changes entirely.

Let's just get to work on having tracked infrastructure in GCP. We'll resolve the differences between `gcp` and `gcp-broad` (as much as possible) at some later date. In the short term, we need to get these security changes in.

daniel-goldstein (approved):
Thanks for reverting the changes to `gcp`, it will be much easier to resolve them individually.

PR #12884, opened 2023-04-13
[batch] Default to using the test method name for the name of the batch

daniel-goldstein:
very helpful when executing tests locally. Most of the lines here are noise because I added a function that wraps `client.create_batch` so needed to replace usages everywhere.


jigold:
I think this might break the one test where we query batches / jobs based on name. I'll take a look at the tests later, but otherwise, this seems fine to me.

daniel-goldstein:
Ya I hit that at first, so I made sure to allow the caller to override the name attribute, so this just sets the name for batches where the name isn't specified in the test

jigold (approved):

PR #12889, opened 2023-04-14
[auth] Add hail_identity to REST get users

illusional:
Context: https://hail.zulipchat.com/#narrow/stream/300487-Hail-Batch-Dev/topic/adding.20service.20account.20name.20to.20users.20list/near/349228510

It's useful for our automation to include this in the REST response get any get user call.

I don't know how testing works, if needed, could you guide me how to add tests to cover this?


daniel-goldstein (changes requested):
Thanks for the change!

Hm so I think we *should* have a proper pytest suite under `hail/python/test/hailtop/auth` that flexes the auth REST API like how `hail/python/test/hailtop/batch` does for the batch API, but we do not. I don't want to ask you to go out of your way for such a small change so feel free to pick any of the above options:

1. Confirm here that you've tested this manually and it works
2. Add a test in `hail/python/test/hailtop/batch/test_batch.py` that uses `get_userinfo` in `hailtop.auth` to verify that the json has the correct spec
3. Add that same test to a new test module under `hail/python/test/hailtop/auth/test_auth.py` and add a step to `build.yaml` that runs that test suite

These are in order of ascending preference but feel free to choose what you feel's most appropriate. I've made a note to flesh out a more comprehensive test suite when there's some free time.

illusional:
Thanks @daniel-goldstein! I tested the single user, but hadn't tested the multi user endpoint - with that test I have resolved a `TypeError: object async_generator can't be used in 'await' expression` in a follow-up commit.

illusional:
Hey @daniel-goldstein, the `ci-test` is marked as failing on GitHub, but the CI job looks like it finished in a complete state.

daniel-goldstein:
Looks like linter checks failed (trailing whitespace in auth.py). Try `make check-auth`? Everything else passed

daniel-goldstein (approved):

PR #12899, opened 2023-04-18
[hailctl] Improve startup time by removing big top-level imports

daniel-goldstein:
`google.api_core.exceptions` imports `pkg_resources` on import which is horrendously slow. Moving these imports out of the top level yields a 60% (almost half a second) improvement in startup time for `hailctl`.


daniel-goldstein:
Screenshot 2023-04-15 at 9 22 26 AM
For reference, this shows startup time for `hailctl` before this change.

jigold (approved):

PR #12900, opened 2023-04-18
[fs] Support reading from public containers in azure

daniel-goldstein:
CHANGELOG: Reading data from public blobs is now supported in Azure

On current main, the following test fails:
```python
def test_can_access_public_blobs():
    public_mt = 'hail-az://azureopendatastorage/gnomad/release/3.1/mt/genomes/gnomad.genomes.v3.1.hgdp_1kg_subset.mt'
    assert hl.hadoop_exists(public_mt)
    mt = hl.read_matrix_table(public_mt)
    mt.describe()
```

The `hadoop_exists` fails in the python fs with a `ClientAuthenticationError` and the QoB pipeline fails on the Query Driver with a 401. See [zulip](https://hail.zulipchat.com/#narrow/stream/300487-Hail-Batch-Dev/topic/Azure.20public.20blobs/near/284186815) for additional context. I contemplated whether we should have a public container in our infrastructure for running tests, but the lack of requester pays makes that feel not great. Azure's paying for the opendatastorage account so I figured it would be ok for us to test against it (and the test only reads metadata), but happy to take alternative suggestions.

### DIFF
--- a/hail/python/test/hail/fs/test_worker_driver_fs.py
+++ b/hail/python/test/hail/fs/test_worker_driver_fs.py
@@ -1,6 +1,6 @@
 import hail as hl
 from hailtop.utils import secret_alnum_string
-from hailtop.test_utils import skip_in_azure
+from hailtop.test_utils import skip_in_azure, run_if_azure
 
 from ..helpers import fails_local_backend, fails_service_backend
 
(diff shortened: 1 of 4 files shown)


hail/python/test/hail/fs/test_worker_driver_fs.py
```diff
@@ -116,3 +116,12 @@ def test_requester_pays_with_project_more_than_one_partition():
...
     assert hl.import_table('gs://hail-services-requester-pays/zero-to-nine', no_header=True, min_partitions=8).collect() == expected_file_contents
+
+
+@run_if_azure
+@fails_local_backend
+def test_can_access_public_blobs():
+    public_mt = 'hail-az://azureopendatastorage/gnomad/release/3.1/mt/genomes/gnomad.genomes.v3.1.hgdp_1kg_subset.mt'
+    assert hl.hadoop_exists(public_mt)
+    mt = hl.read_matrix_table(public_mt)
+    mt.describe()
```
danking, line 127: Can you add a test that does `hl.hadoop_open('.../README.txt')` and asserts that the content is what we expect. I'm not so concerned about the content as I am that the `read` path works as expected.

danking (changes requested):
Looks great, one comment!

PR #12901, opened 2023-04-18
[query] fix & improve pprint for hl.Struct

danking:
CHANGELOG: `hl.Struct` now has a correct and useful implementation of pprint.

For structs with keys that were not identifiers, we produced incorrect `repr` output. For `pprint`, we just `pprint`'ed a dictionary (so you cannot even tell that the object was an `hl.Struct`). This PR:

1. Fixes `hl.Struct.__str__` to use the kwargs or dictionary representation based on whether the keys are Python identifiers.
2. Teaches `StructPrettyPrinter` to first try to `repr` the struct (this is what the default pretty printer does)
3. Teaches `StructPrettyPrinter` to properly pretty print a struct as an `hl.Struct` preferring the kwarg representation when appropriate.
4. Teaches `_same` to use pretty printing when showing differing records.

### DIFF
--- a/hail/python/hail/table.py
+++ b/hail/python/hail/table.py
@@ -3670,7 +3670,7 @@ class Table(ExprContainer):
 
         if not hl.eval(_values_similar(t[left_global_value], t[right_global_value], tolerance, absolute)):
             g = hl.eval(t.globals)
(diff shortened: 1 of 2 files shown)


danking:
Lightening Tim's review load a bit.

chrisvittal (approved):

PR #12906, opened 2023-04-19
[query] permit pandas 2

danking:


daniel-goldstein (approved):

danking:
I feel like CI is taking a long time to merge things lately.

danking:
ah its frozen

jmarshall:
Note that pandas 2.0.0 [removes the deprecated `DataFrame.iteritems()`](https://pandas.pydata.org/docs/whatsnew/v2.0.0.html#removal-of-prior-version-deprecations-changes), which is used by bokeh-1.4.0. That particular version of bokeh is listed in _hail/python/requirements.txt_ but it is thus incompatible with pandas 2; so one or the other of these pinnings probably needs to be revisited.

(This incompatibility has caused the [large_cohort unit test failure](https://github.com/populationgenomics/production-pipelines/actions/runs/4782280056/jobs/8501466504?pr=354#step:5:134) in populationgenomics/production-pipelines#354.)

daniel-goldstein:
Thanks @jmarshall for bringing this to our attention. It looks like while we updated the upper bound here, we did not update our fully-pinned requirements which we use to test in CI, so it did not catch this incompatibility. That being said, I don't think that was necessarily a mistake, because by testing our minimum-compatible-version we make sure not to introduce incompatibilities on that end of the spectrum either.. I think I don't see a good way in which we can confidently support more than one major version of a dependency at a given point in time. Even without the bokeh issue, there could easily be place in our codebase where we use pandas 1.x functionality that has been removed in 2.0. @danking thoughts on moving the pandas pin to >= the 2.x.x version that we test with and <3?

daniel-goldstein:
> I haven't checked if new bokeh supports old pandas. Nor do I know if we have old pandas usage lurking in the codebase. Can we make our pinned-requirements.txt use pandas 2.0, fix whatever issues arise, but leave requirements.txt flexible for folks?

I think there are compromises either way, but I would be surprised if this just worked. It seems very easy to accidentally adopt new functionality so at that point why even have a lower-bound? I think that, while it's very hard to make sure that all our dependencies are compatible at all possible version combinations, and these things will happen, it just feels like an easily-avoidable lie to say we support 1.x and 2.x but then use functionality exclusive to 2.x. So I'm ok keeping the bounds more relaxed if we can guarantee that *our* usage of pandas is compatible with both.

FWIW, I think that our primary dependencies release major versions infrequently enough that it is reasonable to only support the most recent major version, in much the same way that we don't support python versions indefinitely.

daniel-goldstein:
I think in an ideal world, every project including ours, our dependencies, and our end users, maintain public wide-open requirements files and private fully-pinned lockfiles. Our users know we need X and Y and they pin whatever versions are compatible with the union of our code and theirs. If we don't want our end users to have to do that, we need to compromise by narrowing the window and accepting the faulted but practical middle-ground.

@jmarshall, out of curiosity does your group fully pin your dependencies / do you have any thoughts? It would be interesting to hear from the perspectives of end users but also third-party collaborators.

danking:
https://github.com/hail-is/hail/pull/12927 should jump us to bokeh 3.x.x which is compatible with pandas, afaik.

PR #12908, opened 2023-04-19
[all][query] update to Python3.8 which allows for numpy 1.24.2

danking:
CHANGELOG: Hail no longer officially supports Python 3.7.


daniel-goldstein:
Sorry if this is jumping the gun, but can you add a reference in the docs to numpy's versioning policy and say that we follow that now?

danking:
Haha, I had intentionally not assigned you until I sorted what I anticipate is a long tail of issues. Nonetheless, updated.

Makefile
```diff
@@ -75,13 +75,13 @@ install-dev-requirements:
 hail/python/hailtop/pinned-requirements.txt: hail/python/hailtop/requirements.txt
 	./generate-linux-pip-lockfile.sh hail/python/hailtop
 
-hail/python/pinned-requirements.txt: hail/python/requirements.txt hail/python/hailtop/pinned-requirements.txt
+hail/python/pinned-requirements.txt: hail/python/hailtop/pinned-requirements.txt hail/python/requirements.txt
 	./generate-linux-pip-lockfile.sh hail/python
 
-hail/python/dev/pinned-requirements.txt: hail/python/dev/requirements.txt hail/python/pinned-requirements.txt
+hail/python/dev/pinned-requirements.txt: hail/python/pinned-requirements.txt hail/python/dev/requirements.txt
```
danking, line 81: I flipped these so that the pinned-requirements.txt dependencies always come first

Makefile
```diff
@@ -75,13 +75,13 @@ install-dev-requirements:
...
 	./generate-linux-pip-lockfile.sh hail/python/dev
 
-gear/pinned-requirements.txt: hail/python/hailtop/pinned-requirements.txt gear/requirements.txt
+gear/pinned-requirements.txt: hail/python/pinned-requirements.txt hail/python/dev/pinned-requirements.txt gear/requirements.txt
```
danking, line 84: `gear/pinned-requirements.txt` has a `-c` dependency on `python/pinned-requirements.txt` and `python/dev/pinned-requirements.txt`. It was missing both of those dependencies in the makefile which lead me to have incompatible versions between `gear/pinned-requirements.txt` and `python/dev/pinned-requirements.txt` which lead to issues when both of those were `-c` required in auth.

daniel-goldstein:
[This](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table) would indicate that we're already in 3.9+ territory. Maybe a lot to drop two python versions at once, do you have any thoughts on when we should also drop 3.8?

danking:
Great point. I'm inclined to at least do separate PRs since this one is, as expected, becoming a bit of a compatibility nightmare.

danking:
I'm gonna submit a new PR which subsumes this and #12927 because they're both entangled in a bunch of the same issues and its increasingly apparent that I need them together to pass all the tests.

PR #12913, opened 2023-04-20
[query/ggplot] fixes legend behavior for single-category columns

iris-garden:
CHANGELOG: Fixed bug in `hail.ggplot` where all legend entries would have the same text if one column had exactly one value for all rows and was mapped to either the `shape` or the `color` aesthetic for `geom_point`.

This change fixes [an edge case that was encountered](https://hail.zulipchat.com/#narrow/stream/123000-general/topic/plotting.20legend.20bug/near/340198585) where, if the user creates a scatter plot and assigns one column with multiple values in it to the `color` aesthetic and one column with exactly one value in it to the `shape` aesthetic (or vice versa), the plot has a legend which contains the correct visual entries, but they are all labelled with the value from the single-value column, instead of the labels from the other column.

For a concrete example, this code:

```python3
import hail as hl

ht = hl.utils.range_table(10).annotate(x = hl.rand_norm(0, 1), y = hl.rand_norm(0, 1))

ht = ht.annotate(proj_title = 'constant')
ht = ht.annotate(subpop = hl.str(ht.idx % 3))

p = (
    hl.ggplot.ggplot(ht, hl.ggplot.aes(x=ht.x, y=ht.y, color=ht.subpop, shape=ht.proj_title))
    + hl.ggplot.geom_point()
)
p.show()
```

Produces a graph like this one:

Screen-Shot-2023-03-07-at-13 58 48

When it should instead produce one like this:

newplot

Further complicating matters, the first graph cannot be reproduced consistently due to the nature of the bug. See my comment on the code below for more information.

### DIFF
--- a/hail/python/hail/ggplot/geoms.py
+++ b/hail/python/hail/ggplot/geoms.py
@@ -173,15 +173,16 @@ class GeomPoint(Geom):
         legends = {}
         for df in grouped_data:
             values = self._get_aes_values(df)
-            trace_category = "trace1"
+            trace_categories = []
             for aes_name in self.aes_legend_groups:
                 category = self._get_aes_value(df, f"{aes_name}_legend")
-                trace_category = category if category is not None else trace_category
+                if category is not None:
+                    trace_categories.append(category)
                 legends[aes_name] = ({
                     **legends.get(aes_name, {}),
                     category: values[aes_name]
                 })
-            traces.append([fig_so_far, df, facet_row, facet_col, values, trace_category])
+            traces.append([fig_so_far, df, facet_row, facet_col, values, trace_categories])
 
         non_empty_legend_groups = [
             legend_group for legend_group in legends.values() if len(legend_group) > 1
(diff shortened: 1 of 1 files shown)


hail/python/hail/ggplot/geoms.py
```diff
@@ -173,21 +173,32 @@ def apply_to_fig(self, grouped_data, fig_so_far: go.Figure, precomputed, facet_r
         legends = {}
         for df in grouped_data:
             values = self._get_aes_values(df)
-            trace_category = "trace1"
+            trace_categories = []
             for aes_name in self.aes_legend_groups:
                 category = self._get_aes_value(df, f"{aes_name}_legend")
-                trace_category = category if category is not None else trace_category
```
iris-garden, line 179: When I tried to reproduce this issue locally, I was initially confused because the code example above (distilled from one of Dan's examples in the linked Zulip thread) produced the correct graph on my machine. However, another example that had worked fine for Dan evinced the issue for me. The example in question switched `shape` to be the aesthetic that had only one category (`"constant"`).

The reason for this is that this `for` loop is iterating through a `set`, not a `list`, so it does not have a deterministic order, and the root cause of the bug is that, in this line, if the case is encountered where there are two different aesthetics that each have categories, whichever one was seen most recently is the category used for the legend. This went unnoticed in my testing because it only shows up when there's a column with exactly one value in it mapped to either the `color` or `shape` aesthetic, not when one of the two is unmapped or both are mapped to columns with multiple values in them.

My change addresses the issue by storing all categories that are seen in this loop, and waiting until we have checked which column has multiple values to flatten them down to a single relevant category. In the case where multiple columns have multiple values, the category label we generate by this process is ignored in favor of the dummy legend introduced in https://github.com/hail-is/hail/pull/12254.

danking:
If ready for review, "assign" someone and it'll show up in their CI queue :)

patrick-schultz (approved):

PR #12917, opened 2023-04-21
[fs] Add support for azure https blob urls

daniel-goldstein:
CHANGELOG: ABS blob URIs in the format of `https://<ACCOUNT_NAME>.blob.core.windows.net/<CONTAINER_NAME>/<PATH>` are now supported. The `hail-az` scheme for referencing blobs in ABS is now deprecated and will be removed in an upcoming release.

This PR introduces the https addressing of blobs in ABS and phases out hail-az. The test suite converts completely to testing `https`, but both schemes are still supported. We can have confidence that this did not break completely break the `hail-az` scheme because our test bucket configuration is still using `hail-az` (and must until this PR is merged. So some of the test suite + all the service backend tests are flexing the `https` code path, and then the inter_cloud tests are flexing the `hail-az` code path. After this merges, we'll need the following PRs

- Update the azure terraform to use `https` instead of `hail-az` and apply the changes
- Remove support for the hail-az scheme. This will be a breaking change as the copy tool and batch worker will stop being able to transfer files and logs for that scheme. It seems like there is large support for dropping this scheme entirely though so I'd rather make this change while there is little use on azure.


danking (approved):
LGTM

PR #12918, opened 2023-04-21
[batch] Fix combinatorial explosion in cancellation calculation in LocalBackend

### LINKED ISSUE #12915: Combinatorial explosion in LocalBackend.add_dependents
### What happened?

One of our unit tests recently changed from taking around 20 seconds to being aborted by a time out after six hours — see populationgenomics/production-pipelines#352. This change turned out to coincide with the release of hail 0.2.113 and the unit test's `pip` selecting the new release.

PR #12780 added a recursive `add_dependents` function to `LocalBackend`, that appears to be used to compute the transitive dependencies of each job. Profiling our unit test indicates that it is spending six hours inside this function with no end in sight.

Running the job locally for a few seconds with more logging shows that it is calling `add_dependents` with the same `ancestor` and `child` millions of times. I'm not sure whether it's in an actual infinite loop or “merely” a combinatorial disaster than might terminate after a few months of runtime…

The following change, for example,

```diff
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -268,7 +268,7 @@ class LocalBackend(Backend[None]):
             def add_dependents(ancestor, child):
                 dependent_jobs[ancestor].add(child)
                 for ancestor_parent in ancestor._dependencies:
-                    add_dependents(ancestor_parent, child)
+                    if child not in dependent_jobs[ancestor_parent]: add_dependents(ancestor_parent, child)
 
             for j in jobs:
                 for parent in j._dependencies:
```

reduces it to calling it only once or twice for each `ancestor`/`child` combination, and returns the unit test to completing in ~20 seconds.

I am not familiar enough with the data structure to say if that is a correct fix, but something of this nature appears to be needed to return this transitive dependency computation to a sensible runtime.

### Version

0.2.113

### Relevant log output

_No response_

daniel-goldstein:
Fixes #12915

### DIFF
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -263,22 +263,23 @@ class LocalBackend(Backend[None]):
             )
 
             jobs = batch._unsubmitted_jobs
-            dependent_jobs = collections.defaultdict(set)
-
-            def add_dependents(ancestor, child):
-                dependent_jobs[ancestor].add(child)
-                for ancestor_parent in ancestor._dependencies:
-                    add_dependents(ancestor_parent, child)
+            child_jobs = collections.defaultdict(set)
 
             for j in jobs:
                 for parent in j._dependencies:
-                    add_dependents(parent, j)
+                    child_jobs[parent].add(j)
 
             cancelled_jobs = set()
+            def cancel_child_jobs(j):
+                for child in child_jobs[j]:
+                    if not child._always_run:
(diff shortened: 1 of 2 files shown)


danking (approved):

jmarshall:
Thanks, that returns our unit test to taking its expected 20 seconds runtime. The 454 jobs cause a total of 37844 entries to be added to `child_jobs`, which is both trivial and as expected for the number of edges.

PR #12927, opened 2023-04-24
[query] require bokeh 3.x.x & many fixes to plots.py

danking:
CHANGELOG: Bokeh 3.x.x is now required which allows for compatibility with pandas 2.x.x.

There were a lot of issues in this file that I tried to clean up. The mypy errors are now just things that are poorly typed by Bokeh. I also changed how manhattan plot hides irrelevant tooltip information. I made it an explicit argument to the `_get_scatter_plot_elements`.

I verified the plots in the GWAS Tutorial look the same by eye. I also verified the tooltips look the same with a few spot checks. I also built the docs to ensure those were fine.


danking:
I'm gonna merge this into https://github.com/hail-is/hail/pull/12908 because they're so entangled with the same issues and it also seems that updating to bokeh 3.x.x is the easiest way to resolve issues I'm seeing in https://github.com/hail-is/hail/pull/12908

PR #12929, opened 2023-04-24
[query] Python 3.8, Numpy 1.24.2, Bokeh 3.x.x, fixes to plots.py, fixes to req generation

danking:
CHANGELOG: Hail no longer officially supports Python 3.7.

Combines https://github.com/hail-is/hail/pull/12927 and https://github.com/hail-is/hail/pull/12908.

Many changes. All seem to be necessary together.

---

I fixed any new mypy errors or deprecation warnings. I also cleaned up plots.py (which isn't CI'd by mypy) because I was in there and it was a mess. I unified all our pip-tools versions. Require pandas 2 now. That requires Bokeh 3.x.x. Fix the pinned-requirements.txt dependencies so they reflect the actual necessary runtime harmony. Upgraded Sphinx. The method names lose their fixed-width styling but I think it looks fine. Version policy added. Updating everything means Jinja2 can jump to the latest version too. Numpy deprecated some of its types, using `_` gets rid of the warning.


daniel-goldstein (approved):
This all looks fine but I added WIP so that we can replicate Ben's issue with the same version of python/our dependencies. Once that's done merge at your leisure

hail/python/requirements.txt
```diff
@@ -1,13 +1,14 @@
...
-bokeh==1.4.0
+bokeh>=3,<4
 decorator<5
 Deprecated>=1.2.10,<1.3
 dill>=0.3.1.1,<0.4
 frozenlist>=1.3.1,<2
 hurry.filesize>=0.9,<1
-Jinja2==3.0.3
 numpy<2
 pandas>=1.3.0,<2.1
```
jmarshall, line 13: Does this requirement also need adjusting to `pandas>=2,<3`?

daniel-goldstein (changes requested):
Need to change pandas version in hall requirements

hail/python/requirements.txt
danking: Good catch thanks @jmarshall

daniel-goldstein (approved):

PR #12935, opened 2023-04-26
[query] Expand WriteValue beyond hail EType serialization

chrisvittal:
WriteValue has always been a weird bit of functionality. It takes a path and writes a single value to that path. This change does not change that quirk, however this change expands the possibilities of hail value serialization through the introduction of ValueWriter, analogous to PartitionWriter.

ValueWriters serialize hail values and are responsible for the allocation and cleanup of any resources required to do that.

One ValueWriter has been implemented as part of this change. The ETypeFileValueWriter, the minimum necessary functionality for ensuring WriteValue still works exactly as it used to.

The short term here is to implement the BlockMatrixRectanglesWriter with a value writer that dumps an NDArray in either binary or text. The longer term is to develop more faculties within the IR to (de)serialize values.


tpoterba (approved):
Looks good! simple change.

PR #12940, opened 2023-04-26
[qob] fix job logs

danking:
Spark 3.3.0 uses log4j2. Note the "2". I suspect log4j2, although it simulates the log4j1 interface, does not correctly support the log4j1 reconfiguration methods. As a result, we began losing QoB job logs.

This PR embraces log4j2. The JVMEntryway is now a proper gradle project with two files: JVMEntryway.java and QoBAppender.scala. The former now more extensively logs exceptions into the job logs where they can be seen by users.

`QoBAppender` and `QoBOutputStreamManager` work together to ensure QoB job logs go to a different file for each job.

In log4j2 there is a log manager and there are loggers, appenders, and appender managers. The log manager appears to be a singleton manager of the log system. Within a log manager there is a directed acyclic graph of loggers. A logger accepts new log messages from user code. It adds some information (like the source class) to the message. If it has an attached appenders, it sends the message to the appenders. It then sends the message up the DAG to its parent logger.

There is one root logger which is the ancestor of all other loggers. In this PR, the root logger has exactly one appender: `QoBAppender`. In Hail Query on Batch, we should always have exactly one appender, the QoBAppender, and it should be attached to the root logger.

Multiple appenders can share an appender manager. Appender managers allow multiple appenders to share an output file, for example. I think we probably did not need to use a manager in this case.


danking:
cc: @tpoterba @daniel-goldstein

danking:
I verified this by testing directly in my namespace

PR #12942, opened 2023-04-27
[batch] Fix bad insert attempt resources trigger

jigold:
There was an extra check in the where statement for the token matching on the jobs billing table. The jobs billing table is the only one not parameterized with a token.


jigold:
We should probably have SELECT FOR UPDATE on these queries as well.

daniel-goldstein (approved):

PR #12944, opened 2023-04-27
[batch] Add timeouts to ensure that post_job_complete is reached

daniel-goldstein:
We always ensure that `cleanup` is run, but in order to ensure that `post_job_complete` is run we need to ensure we get through any computation in cleanup and `mark_complete` that could potentially hang. So we add timeouts to any yield points in those functions and broadly catch exceptions so we always continue in the cleanup/mark_complete process regardless of failure.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1553,8 +1553,10 @@ class Job(abc.ABC):
                         self.attempt_id,
                         json.dumps(full_status),
                     )
+            except asyncio.CancelledError:
+                raise
             except Exception:
-                log.exception('Encountered error while writing status file')
+                log.exception(f'Encountered error while writing status file for job {self.id}')
 
         if not self.deleted:
             self.task_manager.ensure_future(self.worker.post_job_complete(self, mjs_fut, full_status))
@@ -1894,13 +1896,13 @@ class DockerJob(Job):
     async def cleanup(self):
         if self.disk:
             try:
-                async with async_timeout.timeout(120):
+                async with async_timeout.timeout(300):
                     await self.disk.delete()
                     log.info(f'deleted disk {self.disk.name} for {self.id}')
             except asyncio.CancelledError:
                 raise
             except Exception:
-                log.exception(f'while detaching and deleting disk {self.disk.name} for {self.id}')
+                log.exception(f'while detaching and deleting disk {self.disk.name} for job {self.id}')
         else:
             self.worker.data_disk_space_remaining.value += self.external_storage_in_gib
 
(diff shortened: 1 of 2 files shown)


batch/batch/worker/worker.py
```diff
@@ -1890,8 +1894,9 @@ async def run(self):
     async def cleanup(self):
         if self.disk:
             try:
-                await self.disk.delete()
-                log.info(f'deleted disk {self.disk.name} for {self.id}')
+                async with async_timeout.timeout(120):
```
jigold, line 1897: This timeout might be too short, but we'll find out if that's the case.

batch/batch/worker/worker.py
```diff
@@ -1908,18 +1913,24 @@ async def cleanup(self):
...
                     except asyncio.CancelledError:
                         raise
                     except Exception:
                         log.exception(f'while unmounting fuse blob storage {bucket} from {mount_path}')
 
-        await check_shell(f'xfs_quota -x -c "limit -p bsoft=0 bhard=0 {self.project_id}" /host')
+        try:
+            async with async_timeout.timeout(120):
+                await check_shell(f'xfs_quota -x -c "limit -p bsoft=0 bhard=0 {self.project_id}" /host')
+        except Exception:
```
jigold, line 1928: Do we need to catch `asyncio.CancelledError` here?

batch/batch/worker/worker.py
```diff
@@ -1544,13 +1544,17 @@ async def mark_complete(self, mjs_fut: asyncio.Task):
...
+            try:
+                async with async_timeout.timeout(120):
+                    await retry_transient_errors(
+                        self.worker.file_store.write_status_file,
+                        self.batch_id,
+                        self.job_id,
+                        self.attempt_id,
+                        json.dumps(full_status),
+                    )
+            except Exception:
```
jigold, line 1556: Do we need to catch `asyncio.CancelledError` here?

batch/batch/worker/worker.py
```diff
@@ -1908,18 +1913,24 @@ async def cleanup(self):
...
+        try:
+            async with async_timeout.timeout(120):
+                await check_shell(f'xfs_quota -x -c "limit -p bsoft=0 bhard=0 {self.project_id}" /host')
+        except Exception:
+            log.exception(f'while resetting xfs_quota project {self.project_id}')
 
         try:
-            await blocking_to_async(self.pool, shutil.rmtree, self.scratch, ignore_errors=True)
+            async with async_timeout.timeout(120):
+                await blocking_to_async(self.pool, shutil.rmtree, self.scratch, ignore_errors=True)
```
jigold, line 1933: Let's add the job id to the exception message below.

batch/batch/worker/worker.py
```diff
@@ -1908,18 +1913,24 @@ async def cleanup(self):
...
                         raise
                     except Exception:
                         log.exception(f'while unmounting fuse blob storage {bucket} from {mount_path}')
 
-        await check_shell(f'xfs_quota -x -c "limit -p bsoft=0 bhard=0 {self.project_id}" /host')
+        try:
+            async with async_timeout.timeout(120):
+                await check_shell(f'xfs_quota -x -c "limit -p bsoft=0 bhard=0 {self.project_id}" /host')
+        except Exception:
+            log.exception(f'while resetting xfs_quota project {self.project_id}')
```
jigold, line 1929: Let's add the job id to the exception message here as well.

jigold (changes requested):
I didn't put this comment everywhere. Let's make sure all of the log messages in this cleanup block have the job id in them.

batch/batch/worker/worker.py
danking: I agree with Jackie. we need a
```suggestion
            except asyncio.CancelledError:
                raise
            except Exception:
```

daniel-goldstein: Good points.

jigold (approved):

PR #12949, opened 2023-04-28
[batch] Turn off support for cloudfuse

jigold:
This is a mitigation for turning off cloudfuse until we understand why the unmount is not working properly.

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1065,7 +1065,7 @@ WHERE batch_updates.batch_id = %s AND batch_updates.update_id = %s AND user = %s
             spec['env'].append({'name': 'AZURE_APPLICATION_CREDENTIALS', 'value': '/gsa-key/key.json'})
 
         cloudfuse = spec.get('gcsfuse') or spec.get('cloudfuse')
-        if not is_developer and user not in ('ci', 'test', 'test-dev') and cloudfuse is not None:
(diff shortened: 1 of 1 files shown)


daniel-goldstein (approved):

daniel-goldstein (dismissed):

daniel-goldstein (dismissed):
I believe that QoB submits jobs with `cloudfuse` in the spec, it's just empty unless they're reading FASTA files. So we would need to check for a non-empty config instead of None

batch/batch/front_end/front_end.py
```diff
@@ -1063,6 +1064,10 @@ async def _create_jobs(userdata: dict, job_specs: dict, batch_id: int, update_id
         if cloud == 'azure' and all(envvar['name'] != 'AZURE_APPLICATION_CREDENTIALS' for envvar in spec['env']):
             spec['env'].append({'name': 'AZURE_APPLICATION_CREDENTIALS', 'value': '/gsa-key/key.json'})
 
+        cloudfuse = spec.get('gcsfuse') or spec.get('cloudfuse')
+        if not is_developer and user not in ('ci', 'test', 'test-dev') and cloudfuse is not None:
```
daniel-goldstein, line 1068: Should just need this

```suggestion
        if not is_developer and user not in ('ci', 'test', 'test-dev') and cloudfuse is not None and len(cloudfuse) > 0:
```

daniel-goldstein (approved):

PR #12953, opened 2023-05-01
[hailtop.batch] Fix 12924 -- waiting on a batch with zero unsubmitted jobs

### LINKED ISSUE #12924: `hailtop.batch` errs when a batch of zero jobs is submitted and waited on
### What happened?

Reported by Wenhan: https://hail.zulipchat.com/#narrow/stream/223457-Hail-Batch-support/topic/ValueError.3A.20min.28.29.20arg.20is.20an.20empty.20sequence/near/352319161

```
  File "/usr/local/Caskroom/miniconda/base/lib/python3.9/site-packages/hailtop/batch/backend.py", line 791, in _async_run
    starting_job_id = min(j._client_job.job_id for j in unsubmitted_jobs)
```
I suspect we do not handle waiting on an empty batch / batch update correctly. We should probably bail out immediately from the wait when there are no jobs on which to wait.

### Version

batch

### Relevant log output

_No response_

jigold:
Fixes #12924

### DIFF
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -786,7 +786,7 @@ class ServiceBackend(Backend[bc.Batch]):
 
         if open:
             webbrowser.open(url)
-        if wait:
+        if wait and len(unsubmitted_jobs) > 0:
             if verbose:
                 print(f'Waiting for batch {batch_handle.id}...')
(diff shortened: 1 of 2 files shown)


danking (changes requested):
Let's add a test for this behavior too

danking (approved):

PR #12954, opened 2023-05-01
[query] Call-Cache `CollectDistributedArray` (rfc-0000)

ehigham:
Adds the ability to rerun/retry queries from the nearest `CollectDistributedArray` (`CDA`) `IR` site.

Computes a "Semantic Hash" of the top-level `IR` which is used to generate a key for the various constituent `CDA` calls in a query. The implementation for CDA, `BackendUtils.collectDArray`, uses that key to look into an the execution cache for the results of each partition for that call and uses/updates the cache with successful partition computations.

The nature of the staged- lower and execute model means we don't know how many `CDA` calls that will be generated ahead of time. Thus we treat the "Semantic Hash" in a similar way to an RNG state variable and generate a key from the Semantic Hash every time every time we encounter a `CDA`.

The execution cache is implemented on-top of a local or remote filesystem (configurable via the `HAIL_CACHE_DIR` environment variable). This defaults to `{tmpdir}/hail/{pip-version}`.


hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
```diff
@@ -0,0 +1,409 @@
...
+    Some(buffer.result())
+  }
+
+  def getFileHash(fs: FS)(path: String): Array[Byte] =
+    fs.eTag(path) match {
+      case Some(etag) =>
+        etag.getBytes
+      case None =>
+        path.getBytes ++ Bytes.fromLong(fs.fileStatus(path).getModificationTime)
+    }
```
ehigham, line 352: Something to call out here:
I'm just using the file's etag if the filesystem supports them and NOT the path. 
I think that the etag is unique on azure (their doc is quite hard to navigate so I'm finding it hard to know for sure - welcome help!).
GCS has the nice property that copying the file preserves the etag and it only changes after modification.

hail/src/main/scala/is/hail/expr/ir/NormalizeNames.scala
```diff
@@ -241,10 +241,18 @@ class NormalizeNames(normFunction: Int => String, allowFreeVariables: Boolean =
...
         } yield CollectDistributedArray(newCtxs, newGlobals, newC, newG, newBody, newDynamicID, staticID, tsd)
       case RelationalLet(name, value, body) =>
+        val newName = gen()
         for {
-          newValue <- normalize(value, BindingEnv.empty)
-          newBody <- normalize(body)
-        } yield RelationalLet(name, newValue, newBody)
+          newValue <- normalize(value, env)
+          newBody <- normalize(body, env.bindRelational(name, newName))
+        } yield RelationalLet(newName, newValue, newBody)
```
patrick-schultz, line 248: This should match the binding structure defined in `Binds.scala`, which would be
```
          newValue <- normalize(value, env.onlyRelational)
          newBody <- normalize(body, env.noAgg.noScan.bindRelational(name, newName))
```

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
```diff
@@ -0,0 +1,409 @@
...
+    MurmurHash3.hash32x86(bytes, 0, bytes.length, x)
+
+
+  def apply(ctx: ExecuteContext)(root: BaseIR): Option[Type] =
+    ctx.timer.time("SemanticHash") {
+
+      // Running the algorithm on the name-normalised IR
+      // removes sensitivity to compiler-generated names
+      val nameNormalizedIR = ctx.timer.time("NormalizeNames") {
+        new NormalizeNames(_.toString, allowFreeVariables = true)(root)
```
patrick-schultz, line 31: Isn't this always run at the top level? What free variables could there be?

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+  private def encode(fs: FS, ir: BaseIR, trace: Int): Option[Array[Byte]] = {
+    val buffer =
+      Array.newBuilder[Byte] ++=
+        Bytes.fromClass(ir.getClass) ++=
+        Bytes.fromInt(trace)
+
+    ir match {
+      case a: AggExplode =>
```
patrick-schultz, line 64: I'm not sure about leaving bound names like `a.name` here out of the hash. *Maybe* it's safe because of the normalize names, but I'm not confident about that, and anyways that could easily break after some innocuous looking future change. I think it's safer to just always include them.

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
```diff
@@ -0,0 +1,409 @@
...
+        buffer ++= Bytes.fromClass(op.getClass)
+
+      case ApplyComparisonOp(op, _, _) =>
+        buffer ++= Bytes.fromClass(op.getClass)
+
+      case ApplyIR(fname, _, _, _) =>
+        buffer ++= fname.getBytes()
+
+      case ApplySeeded(fname, _, _, _, _) =>
+        buffer ++= fname.getBytes()
```
patrick-schultz, line 98: Include `staticUID` and `returnType`

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
```diff
@@ -0,0 +1,409 @@
...
+
+      case ApplySpecial(fname, _, _, _, _) =>
+        buffer ++= fname.getBytes()
```
patrick-schultz, line 101: `typeArgs`, `returnType`

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
```diff
@@ -0,0 +1,409 @@
...
+        bytes.ba.foreach(buffer ++= _)
+
+      case GetField(struct, name) =>
+        buffer ++= Bytes.fromInt(struct.typ.asInstanceOf[TStruct].fieldIdx(name))
+
+      case GetTupleElement(_, idx) =>
+        buffer ++= Bytes.fromInt(idx)
+
+      case Literal(typ, value) =>
+        buffer ++= typ.toJSON(value).toString.getBytes
```
patrick-schultz, line 147: Include `typ`

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
```diff
@@ -0,0 +1,409 @@
...
+      ByteBuffer.allocate(8).putDouble(x).array()
+
+    def fromClass(clz: Class[_]): Array[Byte] =
+      fromInt(clz.hashCode())
+  }
+
+}
+
+case object SemanticTypeName {
+  def apply(t: Type): Array[Byte] = {
```
patrick-schultz, line 386: This looks like it does about the same thing as `BaseType.parsableString`. Unless there's a good reason, should probably just use that.

hail/src/main/scala/is/hail/utils/package.scala
```diff
@@ -939,6 +941,76 @@ package object utils extends Logging
...
+
+    for (_ <- tasks.indices) {
+      completionService.take.get match {
+        case (Success(v), k) =>
+          buffer += ((v, k))
+
+        case (Failure(t), k) =>
+          err = accum(err, (t, k))
+      }
+    }
```
patrick-schultz, line 1007: I think this adds results to the buffer as they finish, is that right? We discussed that there doesn't seem to be any benefit over just blocking until all are finished.

patrick-schultz (commented):
Overall, this is really great work! I have some small requests, some of which we discussed last week. I haven't finished looking at all the hash cases yet, or the tests. With the hash cases, in general I think by default anything that we print, we should include in the hash. In some cases it might not be necessary, but I'd rather be conservative.

hail/src/main/scala/is/hail/backend/BackendUtils.scala
```diff
@@ -22,40 +26,88 @@ class BackendUtils(mods: Array[(String, (HailClassLoader, FS, HailTaskContext, R
...
+                }
+              (failureOpt, successes)
+          }
+
+        log.info(s"[collectDArray|$stageName]: executed ${remainingContexts.length} tasks " +
+          s"in ${formatTime(System.nanoTime() - t)}"
+        )
+
+        val results = merge[(Array[Byte], Int)](cachedResults, successes.sortBy(_._2), _._2 < _._2)
+        Option(semhash).foreach(s => backendContext.executionCache.put(s, results))
```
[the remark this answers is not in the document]
ehigham: My goal was to make the cache *really* dumb - what you give it is what you get out.
How would only putting the new results work?

hail/src/main/scala/is/hail/expr/ir/NormalizeNames.scala
ehigham: Thanks

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
ehigham: This is helpful for testing where I'm just using unbound `ref`s

ehigham: I plan to extract this and normalise all names once at the top

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
ehigham: Ah, I thought `name` was a bound local so it didn't matter (in the same way that the name bound in a let doesn't matter)

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
ehigham: The compiler inserts fieldnames that are not normalized

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -2501,14 +2501,52 @@ class Emit[C](
...
           emitI(dynamicID).consume(cb,
-            (),
+            ctx.executeContext.irMetadata.nextHash.foreach { hash =>
+              val boxed =
+                Code.invokeStatic[Integer](
+                  classOf[Integer],
+                  "valueOf",
+                  Array(classOf[Int]),
+                  Array(hash)
+                )
```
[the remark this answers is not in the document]
ehigham: Thanks. Using an `Option` now so this code is no more

hail/src/main/scala/is/hail/backend/BackendUtils.scala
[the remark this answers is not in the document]
patrick-schultz: You're right, this is fine.

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
patrick-schultz: The name doesn't matter, in the sense that if you change it, and also change all captured references, the program is the same (that's [alpha equivalence](https://en.wikipedia.org/wiki/Lambda_calculus#Alpha_equivalence)). But the hash isn't aware of the binding structure. So for example if you ignore names at binding sites, these two IRs would hash the same:
```
Let('x', ..., Let('y', ..., ... Ref('x')))
Let('y', ..., Let('x', ..., ... Ref('x')))
```
It's possible that the NormalizeNames invariant prevents examples like these, by forcing the outer name to be lexicographically less than the inner or something like that, but I wouldn't want to depend on that without thinking it through carefully, and documenting the properties of NormalizeNames that this depends on. Much simpler to just hash the names.

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
patrick-schultz: Ah, yup, forgot about field names.

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
ehigham: Thanks - this is a really good example

ehigham: Ok, after playing with this, I can't get an example to fail. I think normalizenames takes care of this for us.

NormalizeNames transforms your example to: 
```
Let('1', ..., Let('2', ..., ... Ref('1')))
Let('1', ..., Let('2', ..., ... Ref('2')))
```

For `Ref`, I include the bytes of the name being referenced in the semantic hash. I think the only time we'd see a hash collision with let/ref is if normalizenames produces a collision.

hail/src/main/scala/is/hail/expr/ir/analyses/SemanticHash.scala
```diff
@@ -0,0 +1,433 @@
...
+      ByteBuffer.allocate(8).putDouble(x).array()
+
+    def fromClass(clz: Class[_]): Array[Byte] =
+      fromInt(clz.hashCode())
+  }
+
+  object CodeGenSupport {
+    def lift(hash: SemanticHash.Type): Option[SemanticHash.Type] =
+      Some(hash)
+  }
```
ehigham, line 404: I was having a horrible time trying to generate code to lift an `Int` into an `Option` and kept getting `NoSuchMethodError`s when trying to call the constructor of `Some` or call `Option.apply`, presumably because `Option` is parameterised and thus needs a reference type in its jvm implementation.
I rage quit and wrote this instead.

hail/src/main/scala/is/hail/utils/richUtils/RichVal.scala
```diff
@@ -0,0 +1,6 @@
+package is.hail.utils.richUtils
+
+case class RichVal[A](a: A) extends AnyVal {
+  def &[B](f: A => B): B =
```
patrick-schultz, line 4: I'm not sure about this. We use `&` other places, like on `CodeBoolean` and `CodeInt`, that could clash with this. Where are you using this, and could it be a regular method instead of an operator?

patrick-schultz (changes requested):
I'm feeling good about merging this as an experimental flag, disabled by default. Just one last comment.

hail/src/main/scala/is/hail/utils/richUtils/RichVal.scala
ehigham: aaaand it's gone.

patrick-schultz (approved):

PR #12960, opened 2023-05-01
[batch] Restrict mount propagation for job container mounts

daniel-goldstein:
Applies the most restrictive bind and event propagation settings to job container mounts. While user jobs do not have the capabilities to create mount points, overlapping mount points in the container config can inadvertently trigger mount propagation back to the host which we just never want.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1035,10 +1035,21 @@ class Container:
 
         return False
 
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
Keeping this as a draft until after the post-mortem.

danking (approved):
I especially like the validation.

PR #12961, opened 2023-05-01
[query] xfail should always be strict

danking:
I noticed this while xfail-ing a different test. It is possible this test now passes. We will see!


patrick-schultz (changes requested):
Should we set this globally by adding `xfail_strict=true` to `pytest.ini`?
https://docs.pytest.org/en/7.1.x/how-to/skipping.html#strict-parameter

patrick-schultz (approved):

PR #12962, opened 2023-05-02
[batch] Simplify cloudfuse credential handling

daniel-goldstein:
Just a refactor, simplifies the interactions between `Worker`, `CloudWorkerAPI` and `CloudUserCredentials` and hopefully makes this code safer and easier to work with. Instead of the following occuring in worker.py:

1. get credentials string from `CloudUserCredentials`
2. tell `CloudWorkerAPI` to write credentials string to `path` owned by the job
3. tell `CloudWorkerAPI` to mount cloudfuse using the credentials stored at `path`

we instead just do

1. tell `CloudWorkerAPI` to mount cloudfuse using `CloudUserCredentials`

On its own I think this change makes the codepath simpler and easier to think about in terms of where credentials are stored, but this also gets rid of the requirement from `worker.py`'s point of view that credentials must be stored on the filesystem. This will make it easier to transition off of key files and over to metadata server tokens.

In order to make the new statement sound in terms of types, we can't have `CloudWorkerAPI.mount_cloudfuse` just accept a `CloudUserCredentials` argument, because that means `GCPWorkerAPI` would need to be able to support an argument of type `AzureUserCredentials`, which would never happen and doesn't make sense. What we can do here is make `CloudWorkerAPI` generic over the subtype of `CloudUserCredentials` that it both produces and consumes. This allows us to use stricter types like `GCPUserCredentials` and `AzureUserCredentials` inside of `GCPWorkerAPI` and `AzureWorkerAPI` respectively and now the type system is happy. It also relaxes the restriction that both of the `GCPUserCredentials` and `AzureUserCredentials` need to conform to the same `cloudfuse_credentials` interface.


daniel-goldstein:
Feel free to unassign if you'd rather deal with changes to cloudfuse-adjacent code at a later time, I just had this already primed last week.

batch/batch/cloud/gcp/worker/worker_api.py
```diff
@@ -100,8 +106,10 @@ def _mount_cloudfuse(
     {bucket} {mount_base_path_data}
 '''
 
-    def _unmount_cloudfuse(self, mount_base_path: str) -> str:
-        return f'fusermount -u {mount_base_path}'
+    async def unmount_cloudfuse(self, mount_base_path_data: str):
+        await check_exec_output('fusermount', '-u', mount_base_path_data)
```
danking, line 110: same here

danking (dismissed):

batch/batch/worker/worker_api.py
```diff
@@ -37,20 +39,24 @@ async def worker_access_token(self, session: httpx.ClientSession) -> Dict[str, s
...
     async def mount_cloudfuse(
-        self, credentials_path: str, mount_base_path_data: str, mount_base_path_tmp: str, config: dict
+        self,
+        credentials: CredsType,
+        mount_base_path_data: str,
+        mount_base_path_tmp: str,
+        config: dict,
     ) -> None:
-        mount_command = self._mount_cloudfuse(credentials_path, mount_base_path_data, mount_base_path_tmp, config)
+        mount_command = self._mount_cloudfuse(credentials, mount_base_path_data, mount_base_path_tmp, config)
```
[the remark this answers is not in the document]
daniel-goldstein: Gladly. I originally changed it to `exec` but undid that because it felt like a different change. but I will happily throw it in.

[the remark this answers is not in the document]
daniel-goldstein: I instead just made the abstract method run the operation instead of passing strings.

batch/batch/cloud/azure/worker/worker_api.py
```diff
@@ -55,41 +57,58 @@ async def worker_access_token(self, session: httpx.ClientSession) -> Dict[str, s
...
+            'blobfuse',
+            mount_base_path_data,
+            f'--tmp-path={mount_base_path_tmp}',
+            f'--config-file={fuse_credentials_path}',
+            '--pre-mount-validate=true',
+            '-o',
+            ','.join(options),
+            '-o',
+            'attr_timeout=240',
+            'entry_timeout=240',
```
danking, line 101: you're sure these don't need the `-o`?

danking (changes requested):

batch/batch/cloud/azure/worker/worker_api.py
daniel-goldstein: Nope, good catch, fixed.

danking (approved):

PR #12964, opened 2023-05-02
[batch] use regional bucket for requester pays tests

danking:
I updated terraform but

1. GCP Terraform state is still local on my laptop.

2. GCP Terraform appears to not configure global-config. As such, I cannot thread the name of the bucket through to the tests the way we do with TEST_STORAGE_URI. For now, I've hardcoded the name (which is what we were doing previously). When we eventually get to testing recreation of GCP in a new project we'll have to address the global config then.

### DIFF
--- a/hail/python/test/hail/fs/test_worker_driver_fs.py
+++ b/hail/python/test/hail/fs/test_worker_driver_fs.py
@@ -8,7 +8,7 @@ from ..helpers import fails_local_backend
 @skip_in_azure
 def test_requester_pays_no_settings():
     try:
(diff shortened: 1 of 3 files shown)


danking:
Ah, I deleted the global-config as a part of switching to gcp-broad. Regardless, we'll need to address that as terraform is iteratively improved.

daniel-goldstein (approved):
Ya agreed about the global-config. Since it's hardcoded right now it doesn't matter, but when we bring global-config ownership under terraform we will need to make sure that these resources get added to `gcp` terraform and subsequently threaded through to the global config.

PR #12967, opened 2023-05-02
[ci] Add ruff to linters to replace flake8 and isort

daniel-goldstein:
[ruff](https://beta.ruff.rs/docs/) is a *very* fast linter that we can use as a drop-in replacement for flake8 and isort. Fast enough to run in a pre-commit hook. While there are many pylint rules implemented in ruff, it is not at parity with pylint yet. This PR replaces flake8 and isort in favor of ruff but does not remove pylint yet. Nevertheless, from what I have seen so far ruff + mypy does catch a vast swath of everyday errors, and I have found that the 30+ seconds it can take to run pylint on any of our python modules is a deterrent to me linting often and catching lint errors early. So I added Makefile targets such as `check-batch-fast` that run all the linters except for pylint. The `check-batch` rule now does `check-batch-fast` and `pylint-batch`. So linter coverage should have strictly increased in CI, but there is a <5s linting target now available for devs in addition to the >30s it can take to also run pylint. You can also now just run `ruff .` in the root of the repo and it completes for me in 0.293 seconds.

For the most part in this PR, I added ruff, with the config enabling flake8 + isort + pylint rules, then disabled rules until there were no errors, save for a few rules that I thought to just fix immediately. These mostly line up with the flake8 rules we had already disabled. I also then added ruff's own ruleset ([RUF](https://beta.ruff.rs/docs/rules/#ruff-specific-rules-ruf)) particularly because I appreciated the `asyncio-dangling-task` and `unused-noqa` rules.


daniel-goldstein:
@danking bump

danking:
Yeah, fair bump. I scanned over this. I was a bit unsure how to convince myself it was OK.

I was a bit afraid that we're not listing `PLW2901`.

danking (approved):

danking (commented):
did you remove isort from python/dev/requirements.txt?

danking:
OK, I can see why you didn't enable PLW2901 but it did discover one bug in a CLI command. It's kind of annoying and invalidates reasonable programs but it seems really valuable for catching errors that are common amongst new programmers.

daniel-goldstein:
I didn't want to put too much in this PR but I dislike reassigning the loop variable, it only feels like a footgun. So will happily approve the follow-up PR that fixes violations of that rule. There a couple of rules that I would like to gradually follow more.

w.r.t. convincing it is ok, I suppose I just trust the ruff developers (and the big projects that have already converted) that given we don't use any niche flake8 plugins we are not dropping any rules by moving over.

daniel-goldstein:
> did you remove isort from python/dev/requirements.txt?

I did, but it is still in the pinned reqs because pylint pulls it in

danking (approved):

PR #12968, opened 2023-05-02
[docs] Clarify that developer dependencies require the JDK not just the JRE

daniel-goldstein:


danking (approved):

danking:
Huh. I saw this recently as well but the funny thing is that our streams set a closed bit so they’re idempotent. I don’t understand why the underlying stream is closed before we get to our close method.

PR #12974, opened 2023-05-03
[batch] Use same size VMs in dev and PR as we use in prod

daniel-goldstein:
Now that test databases are hosted on their own servers instead of the single cloud-hosted MySQL, we can ramp up the parallelism both in our tests and in the number of PRs that we run at once. I recall that even before we had this DB bottleneck we still restricted the number of PRs running at once for cost reasons, but if that's not the case we could remove that restriction entirely.

### DIFF
--- a/batch/deployment.yaml
+++ b/batch/deployment.yaml
@@ -319,11 +319,7 @@ spec:
   selector:
     matchLabels:
       app: batch
-{% if deploy %}
   replicas: 3
-{% else %}
-  replicas: 1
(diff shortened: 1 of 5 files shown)


danking:
Should we close?

daniel-goldstein:
Ya I'll revive this when it's time

danking (approved):
huzzah!

daniel-goldstein:
WIP until the migration fixes go in

PR #12984, opened 2023-05-04
[services] indicate how many errors we have seen

danking:
We are increasingly seeing errors from "Connection reset" which we switched from "transient" to "retry once". The current code makes it impossible to determine if we are correctly retrying this error once.

If we see that there are a lot of "Connection reset" errors that happen twice we should perhaps change "retry once" to "retry five times" or use a more generous delay.

### DIFF
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -799,7 +799,7 @@ async def retry_transient_errors_with_debug_string(debug_string: str, warning_de
             if errors == 1 and is_retry_once_error(e):
                 return await f(*args, **kwargs)
             if not is_transient_error(e):
-                raise
+                raise ValueError(f"errors={errors}, delay={delay}") from e
             log_warnings = (time_msecs() - start_time >= warning_delay_msecs) or not is_delayed_warning_error(e)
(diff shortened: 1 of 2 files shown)


daniel-goldstein (approved):

hail/python/hailtop/utils/utils.py
```diff
@@ -799,7 +799,7 @@ async def retry_transient_errors_with_debug_string(debug_string: str, warning_de
             if errors == 1 and is_retry_once_error(e):
                 return await f(*args, **kwargs)
             if not is_transient_error(e):
-                raise
+                raise ValueError(f"errors={errors}, delay={delay}") from e
```
daniel-goldstein, line 802: This smothers the real exception that the caller might expect, e.g. tests that test for a 404

daniel-goldstein (changes requested):

daniel-goldstein:
@danking Is this still relevant? Can we close it since now we're retrying frice?

daniel-goldstein (changes requested):
See comment

danking:
debatable. I think it's useful to know this but maybe better conceived as a log message. Let's close and reconsider when its a problem again.

PR #12985, opened 2023-05-04
[batch] Don't rmtree if any errors occur while unmounting

jigold:
I think this change is better than #12975

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -2230,10 +2230,10 @@ class JVMJob(Job):
                         config['mounted'] = False
                     except asyncio.CancelledError:
                         raise
-                    except Exception:
+                    except Exception as e:
                         raise IncompleteJVMCleanupError(
                             f'while unmounting fuse blob storage {bucket} from {mount_path} for {self.jvm_name} for job {self.id}'
-                        )
+                        ) from e
 
         if self.jvm is not None:
             self.worker.return_jvm(self.jvm)
@@ -2321,8 +2321,7 @@ class JVMCreationError(Exception):
 
 
 class IncompleteJVMCleanupError(Exception):
-    def __init__(self, msg):
-        self.msg = msg
+    pass
 
 
 class JVMUserCredentials:
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -2789,10 +2789,10 @@ class Worker:
             raise
         except JVMCreationError:
             self.stop_event.set()
-        except IncompleteJVMCleanupError as e:
+        except IncompleteJVMCleanupError:
             assert isinstance(job, JVMJob)
             await self.recreate_jvm(job.jvm)
-            log.exception(e.msg)
+            log.exception(f'while running {job}, ignoring')
         except Exception as e:
             if not user_error(e):
                 log.exception(f'while running {job}, ignoring')
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1936,6 +1936,7 @@ class DockerJob(Job):
                         log.exception(
                             f'while unmounting fuse blob storage {bucket} from {mount_path} for job {self.id}'
                         )
+                        raise
 
         try:
             async with async_timeout.timeout(120):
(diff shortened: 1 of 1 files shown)


jigold:
It would be great in a separate PR to check for sure there aren't any zombie fuse mounts before removing files.

batch/batch/worker/worker.py
```diff
@@ -2212,25 +2213,31 @@ async def cleanup(self):
...
                     mount_path = self.cloudfuse_data_path(bucket)
-                    await self.jvm.cloudfuse_mount_manager.unmount(mount_path, user=self.user, bucket=bucket)
-                    config['mounted'] = False
+                    try:
+                        await self.jvm.cloudfuse_mount_manager.unmount(mount_path, user=self.user, bucket=bucket)
+                        config['mounted'] = False
+                    except asyncio.CancelledError:
+                        raise
+                    except Exception:
+                        log.exception(f'error while unmounting cloudfuse for {self.jvm_name} for job {self.id}')
```
danking, line 2234: Let's include the same information we have above:
```suggestion
                        log.exception(
                            f'while unmounting fuse blob storage {bucket} from {mount_path} for {self.jvm_name} for job {self.id}'
                        )
```

batch/batch/worker/worker.py
```diff
@@ -2771,6 +2788,10 @@ async def run_job(self, job):
             raise
         except JVMCreationError:
             self.stop_event.set()
+        except IncompleteJVMCleanupError:
+            assert isinstance(job, JVMJob)
+            await self.recreate_jvm(job.jvm)
+            log.exception(f'while running {job}, incomplete cleanup from {job.jvm}')
```
danking, line 2794: This results in two exceptional logs for a single exception. Let's either not log here or not log near the raise.

danking (changes requested):

batch/batch/worker/worker.py
```diff
@@ -2231,8 +2231,9 @@ async def cleanup(self):
                     except asyncio.CancelledError:
                         raise
                     except Exception:
-                        log.exception(f'error while unmounting cloudfuse for {self.jvm_name} for job {self.id}')
-                        raise IncompleteJVMCleanupError
+                        raise IncompleteJVMCleanupError(
+                            f'while unmounting fuse blob storage {bucket} from {mount_path} for {self.jvm_name} for job {self.id}'
+                        )
```
danking, line 2236: Name the exception and add a `from e` so that we get the fulls tack trace in the log.exception later.

danking (changes requested):
x

batch/batch/worker/worker.py
```diff
@@ -2312,6 +2320,11 @@ class JVMCreationError(Exception):
     pass
 
 
+class IncompleteJVMCleanupError(Exception):
+    def __init__(self, msg):
+        self.msg = msg
```
danking, line 2325: This isn't necessary:
```
In [1]: class IncompleteJVMCleanupError(Exception):
   ...:     pass
   ...: 

In [2]: raise IncompleteJVMCleanupError('foo')
IncompleteJVMCleanupError                 Traceback (most recent call last)
<ipython-input-2-a8a0c94b0ce4> in <cell line: 1>()
----> 1 raise IncompleteJVMCleanupError('foo')

IncompleteJVMCleanupError: foo
```

danking (dismissed):

daniel-goldstein:
These changes look great

danking (approved):
LGTM!

batch/batch/worker/worker.py
```diff
@@ -2789,10 +2789,10 @@ async def run_job(self, job):
             raise
         except JVMCreationError:
             self.stop_event.set()
-        except IncompleteJVMCleanupError as e:
+        except IncompleteJVMCleanupError:
             assert isinstance(job, JVMJob)
             await self.recreate_jvm(job.jvm)
-            log.exception(e.msg)
+            log.exception(f'while running {job}, ignoring')
```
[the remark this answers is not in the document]
danking: Yeah, [`log.exception` does this](https://docs.python.org/3/library/logging.html#logging.Logger.exception).

danking (approved):

PR #12986, opened 2023-05-04
[batch] Check /proc/mounts for straggler cloudfuse mounts

jigold:


batch/batch/worker/worker.py
```diff
@@ -419,6 +419,9 @@ async def _fuse_mount(self, destination: str, *, credentials_path: str, tmp_path
     async def _fuse_unmount(self, path: str):
         assert CLOUD_WORKER_API
         await CLOUD_WORKER_API.unmount_cloudfuse(path)
+        proc_mounts_stdout, _ = await check_shell_output('cat /proc/mounts')
+        if path in str(proc_mounts_stdout):
+            raise IncompleteCloudFuseCleanup(f'incomplete cloudfuse unmounting: {proc_mounts_stdout}')
```
daniel-goldstein, line 424: Ah I see now. This looks good, but can we just use python `open` instead of `cat` to read in the contents of `/proc/mounts`?

daniel-goldstein (dismissed):

batch/batch/worker/worker.py
```diff
@@ -1937,6 +1945,11 @@ async def cleanup(self):
                             f'while unmounting fuse blob storage {bucket} from {mount_path} for job {self.id}'
                         )
 
+        with open('/proc/mounts', 'r') as f:
+            output = f.read()
+            if self.cloudfuse_base_path() in output:
+                raise IncompleteCloudFuseCleanup(f'incomplete cloudfuse unmounting: {output}')
```
danking, line 1951: I'm fairly sure exceptions raised in `DockerJob.cleanup` are not logged anywhere. Using the raise to skip the `xfs_quota` and `rmtree` seems fine but we need to log the exception here.

batch/batch/worker/worker.py
```diff
@@ -419,6 +419,10 @@ async def _fuse_mount(self, destination: str, *, credentials_path: str, tmp_path
     async def _fuse_unmount(self, path: str):
         assert CLOUD_WORKER_API
         await CLOUD_WORKER_API.unmount_cloudfuse(path)
+        with open('/proc/mounts', 'r') as f:
+            output = f.read()
+            if path in output:
+                raise IncompleteCloudFuseCleanup(f'incomplete cloudfuse unmounting: {output}')
```
danking, line 425: This would read this file for each mount rather than once after all attempts to unmount. Let's make JVMJob symmetric to DockerJob and check the mounts in `cleanup` in both Jobs. I think we should also make sure we attempt to upload the log before we check /proc/mounts.

danking (changes requested):

batch/batch/worker/worker.py
jigold: Why are we checking for paths in the scratch directory? Should I be checking for the scratch directory in JVM jobs as well?

batch/batch/worker/worker.py
```diff
@@ -2234,6 +2230,11 @@ async def cleanup(self):
                     await self.jvm.cloudfuse_mount_manager.unmount(mount_path, user=self.user, bucket=bucket)
                     config['mounted'] = False
 
+        with open('/proc/mounts', 'r') as f:
+            output = f.read()
+            if self.cloudfuse_base_path() in output:
```
jigold, line 2235: I think my question is won't this fuse path remain alive if there's other jobs reading the data? That's why I had the check after we actually call `fusermount`.

daniel-goldstein: This is not a fuse path, it's a bind mount path which is the thing that the JVMJob actually owns. Here's an example:

1. JVM 1 wants bucket `foo` at `/cloudfuse/JVM-1/foo`
2. FUSE cache manager makes two mounts:
  - FUSE mount at `/cloudfuse/read_only/abc123/foo`
  - Bind mounts `/cloudfuse/read_only/abc123/foo` to `/cloudfuse/JVM-1/foo`
3. JVM 2 wants bucket `foo` at `/cloudfuse/JVM-2/foo`
4. FUSE cache manager reuses the underlying FUSE mount and bind mounts `/cloudfuse/read_only/abc123/foo` to `/cloudfuse/JVM-2/foo`.
5. JVM 1 finishes and wants to clean up, it calls `cloudfuse_mount_manager.unmount`
6. FUSE cache manager unmounts the bind mount at `/cloudfuse/JVM-1/foo`. The underyling FUSE mount is still being used by another job, so it doesn't do anything else

At this point, the only thing JVM 1 needs to validate to make sure that it is properly cleaned up is that the bind mount that it owns (whatever is under `/cloudfuse/JVM-1`) is gone. Without that bind mount it has no connection to the underlying FUSE mount created by the cache manager.

jigold: Thanks! This was really helpful.

danking (approved):

PR #12988, opened 2023-05-04
[query] Enable use of aggregators on array expressions

tpoterba:
This method is useful for accessing functionality that exists in the aggregator library but not the basic expression library, for instance, `call_stats`.

### DIFF
--- a/hail/python/hail/expr/aggregators/aggregators.py
+++ b/hail/python/hail/expr/aggregators/aggregators.py
@@ -239,6 +239,18 @@ class AggFunc(object):
 
 
 def aggregate_local_array(array, f):
+    """Compute a summary of an array using aggregators. Useful for accessing
+    functionality that exists in `hl.agg` but not elsewhere, like `hl.agg.call_stats`.
+
+    Parameters
+    ----------
+    array
+    f
+
+    Returns
+    -------
+    Aggregation result.
+    """
     elt = array.dtype.element_type
 
     var = Env.get_uid(base='agg')
@@ -246,8 +258,8 @@ def aggregate_local_array(array, f):
     aggregated = f(ref)
 
     if not aggregated._aggregations:
-        raise ExpressionException(f"'hl.aggregate_local_array' "
-                                  f"must take mapping that contains aggregation expression.")
+        raise ExpressionException("'hl.aggregate_local_array' "
+                                  "must take a mapping that contains aggregation expression.")
 
     indices, _ = unify_all(array, aggregated)
     return construct_expr(ir.StreamAgg(ir.ToStream(array._ir), var, aggregated._ir),
--- a/hail/python/hail/expr/aggregators/aggregators.py
+++ b/hail/python/hail/expr/aggregators/aggregators.py
@@ -238,6 +238,24 @@ class AggFunc(object):
             return 'agg'
 
 
+def aggregate_local_array(array, f):
+    elt = array.dtype.element_type
+
+    var = Env.get_uid(base='agg')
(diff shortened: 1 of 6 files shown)


hail/python/hail/expr/aggregators/aggregators.py
```diff
@@ -238,6 +238,36 @@ def context(self):
             return 'agg'
 
 
+def aggregate_local_array(array, f):
```
patrick-schultz, line 241: We had discussed adding a `CollectionExpression.aggregate` method, similar to `fold`. Did you change your mind?

patrick-schultz (changes requested):

hail/python/hail/expr/aggregators/aggregators.py
tpoterba: nope, forgot over the weekend amid the presentation rush. Thanks for the reminder.

hail/python/hail/ir/ir.py
```diff
@@ -2074,6 +2074,78 @@ def renderable_bindings(self, i, default_value=None):
...
 
+class StreamAgg(IR):
+    @typecheck_method(a=IR, value_name=str, body=IR)
+    def __init__(self, a, value_name, body):
+        a = a.handle_randomness(body.uses_randomness)
+        if body.uses_randomness:
+            tuple, uid, elt = unpack_uid(a.typ)
+            body = AggLet(value_name, elt, body)
+            body = with_split_rng_state(body, uid)
+            value_name = tuple
```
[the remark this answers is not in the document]
patrick-schultz: Thanks Chris. That's almost right, but we still need to bind the `elt` (`body` expects `value_name` to be an element without uid).  And lets keep using `with_split_rng_state` to keep that logic in one place:
```
             tup, uid, elt = unpack_uid(a.typ)
             body = AggLet(value_name, elt, body)
             if body.uses_value_randomness:
                 body = Let('__rng_state', RNGStateLiteral(), body)
             if body.uses_agg_randomness(is_scan=False):
                 body = with_split_rng_state(body, is_scan=False)
             value_name = tup
```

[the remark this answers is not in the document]
chrisvittal: thank you, should be fixed now.

patrick-schultz (approved):

patrick-schultz (changes requested):
sorry, realized this still needs a test for the randomness handling, I can add that.

hail/python/hail/ir/ir.py
```diff
@@ -2077,14 +2077,11 @@ def renderable_bindings(self, i, default_value=None):
...
+        a = a.handle_randomness(body.uses_agg_randomness)
+        if body.uses_agg_randomness:
             tup, uid, elt = unpack_uid(a.typ)
-            body = AggLet(value_name, elt, body)
-            if body.uses_value_randomness:
-                body = Let('__rng_state', RNGStateLiteral(), body)
-            if body.uses_agg_randomness(is_scan=False):
-                body = with_split_rng_state(body, is_scan=False)
+            body = AggLet(value_name, elt, body, is_scan=False)
+            body = with_split_rng_state(body, uid, is_scan=False)
```
patrick-schultz: The test caught some problems with this. Most substantially, we shouldn't be making a new root `RNGStateLiteral` for any value randomness. That means if this occurs inside some other looping context, value randomness will be identical on each outer iteration. Instead, we should just use the rng state bound in the outer context, which means we don't need to do anything here.

chrisvittal: Thanks for explaining. Clears up why this was necessary.

chrisvittal (approved):

PR #12990, opened 2023-05-05
[ruff] maybe eliminate PLW2901

danking:
what do you think of it?

---

We should at least keep the fix to list_batches.py. I am unsure why I discovered blatant type errors in ci.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1182,7 +1182,7 @@ class Container:
                     return uid, gid
             raise ValueError("Container user not found in image's /etc/passwd")
 
-    def _mounts(self, uid: int, gid: int, cwd: str) -> List[MountSpecification]:
+    def _mounts(self, uid: int, gid: int) -> List[MountSpecification]:
         assert self.image.image_config
         assert self.netns
         # Only supports empty volumes
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1071,8 +1071,6 @@ class Container:
         uid, gid = await self._get_in_container_user()
         weight = worker_fraction_in_1024ths(self.cpu_in_mcpu)
         workdir = self.image.image_config['Config']['WorkingDir']
-        cwd = workdir if workdir != "" else "/"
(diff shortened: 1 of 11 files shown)


batch/batch/worker/worker.py
```diff
@@ -1177,9 +1177,11 @@ def _mounts(self, uid, gid):
...
         volumes = self.image.image_config['Config']['Volumes']
         if volumes:
-            for v_container_path in volumes:
-                if not v_container_path.startswith('/'):
-                    v_container_path = '/' + v_container_path
+            for _v_container_path in volumes:
+                if not _v_container_path.startswith('/'):
+                    v_container_path = '/' + _v_container_path
+                else:
+                    v_container_path = _v_container_path
```
daniel-goldstein, line 1184: This feels like a bug, no? the `destination` that we set in the container should be as specified by the image manifest, not altered to prepend a `/`. This prepending a slash is only for putting it somewhere easy to find on the host.

I think I only added `VOLUME` support to get buildkit working which I presume uses an absolute path which is why we never noticed it, but I presume there's nothing wrong with specifying a `VOLUME` that is relative to the `WORKDIR`.

daniel-goldstein (changes requested):
I just looked at your second commit, but you should be able to merge main now. Overall on board with attacking these ignored rules.

build.yaml
```diff
@@ -2431,6 +2431,8 @@ steps:
       inline: |
         FROM {{ hail_ubuntu_image.image }}
         VOLUME ["/data"]
+        WORKDIR "/tmp"
+        VOLUME ["relative_volume]
```
daniel-goldstein, line 2435:
```suggestion
        VOLUME ["relative_volume"]
```

daniel-goldstein (changes requested):
Just checking, does this use of a relative path replicate in docker? My stipulation earlier was what *seemed* reasonable, I do not remember off the top of my head the `VOLUME` spec.

danking:
Fixed and test updated.

daniel-goldstein (approved):

PR #12995, opened 2023-05-05
[query] Fix BAD loop deoptimization with StreamAgg

tpoterba:
CHANGELOG: Fixed bug causing poor performance and memory leaks for Matrix.annotate_rows aggregations

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/NestingDepth.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/NestingDepth.scala
@@ -45,6 +45,9 @@ object NestingDepth {
         case StreamMap(a, name, body) =>
           computeIR(a, depth)
           computeIR(body, depth.incrementEval)
+        case StreamAgg(a, name, body) =>
+          computeIR(a, depth)
+          computeIR(body, ScopedDepth(depth.eval, depth.eval + 1, depth.scan))
         case StreamZip(as, _, body, _, _) =>
           as.foreach(computeIR(_, depth))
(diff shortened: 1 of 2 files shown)


patrick-schultz (approved):

danking:
bump, has a bug

danking:
bump

danking:
@patrick-schultz can you adopt this PR?

patrick-schultz:
Yeah, I got it

patrick-schultz:
@tpoterba I think I understand what was happening. I think NestingDepth wasn't handling aggregation right at all. AggLet was getting its nesting depth from the eval scope, which got compared to the nesting depth of the ref's agg scope. So even if both were at the same level inside an aggregation, we weren't forwarding. We were also ignoring StreamAggScan.

Could you check if this looks right to you?

patrick-schultz:
I think there's another separate bug. The problem was that an ir fragment containing a StreamAgg was getting compiled, and it still had an AggLet after LowerArrayAggsToRunAggs. It's now getting inlined, but this should still work even if it isn't inlined. Maybe LowerArrayAggsToRunAggs needs to convert that to a normal Let.

patrick-schultz (approved):

danking:
@patrick-schultz can you confirm that you anticipated this to succeed now and it just needed to be retried or something? Just a bit concerning to see a merge after two weeks without changes.

patrick-schultz:
Yeah, I've been continually retrying. It was passing non-service tests when I approved two weeks ago.

PR #12997, opened 2023-05-05
[gear] use orjson everywhere

danking:


gear/requirements.txt
```diff
@@ -14,3 +14,4 @@ prometheus_async>=19.2.0,<20
 prometheus_client>=0.11.0,<1
 PyMySQL>=1,<2
 sortedcontainers>=2.4.0,<3
+orjson>=3.6.4,<4
```
daniel-goldstein, line 17: This should be inherited from hailtop requirements, why include it again here?

gear/gear/http_server_utils.py
```diff
@@ -0,0 +1,11 @@
+from typing import Any, Callable, Optional
+import orjson
+from aiohttp import web
+
+
+async def json_request(request: web.Request) -> Any:
+    return orjson.loads(await request.read())
```
daniel-goldstein, line 7: does `await request.json(loads=orjson.loads)` not work?

daniel-goldstein (changes requested):
Nice, just a couple comments.

gear/gear/http_server_utils.py
danking: The types don't match. `orjson` works with UTF-8 `bytes` not Python `str`. The [definition of `.json`](https://github.com/aio-libs/aiohttp/blob/b2a7983aed047e36c3418e788096d7f46cea7e1e/aiohttp/web_request.py#L666-L682) calls `.text` which unnecessarily creates a Python `str`. I couldn't figure out a way to avoid the allocation of the string without writing `json_request` myself.

gear/requirements.txt
daniel-goldstein: Ya I went back and forth for a bit on this but I think I made the wrong choice. `gear` imports `hailtop` so it *needs* all of `hailtop`'s requirements. But instead of using `-r` and having duplicative `pinned-requirements.txt` files for every service, I used `-c` and install a bunch of pinned-requirements.txt files in the Dockerfiles. But I think the better thing to do is

- have gear `-r` the hailtop requirements and `-c` the hail and dev requirements
- have the services `-r` the gear requirements
- in the service dockerfiles now we can just install the service's `pinned-requirements.txt`


I think I didn't want requirements to be duplicated across different pinned-requirements files with different versions, but this approach isn't immune to that problem because of transitive dependencies. And we have a more robust way of ensuring that all the requirements are in sync anyway.

gear/gear/http_server_utils.py
daniel-goldstein: Ah lame, ok.

daniel-goldstein (approved):
LGTM. Let's revisit the `-c` conversation but in a separate discussion.

PR #13000, opened 2023-05-05
not for merging just for test

daniel-goldstein:


daniel-goldstein:
Confirmed that test K8s SAs cannot list secrets outside their namespace

PR #13004, opened 2023-05-08
[ci] Run service backend tests on non-preemptibles

daniel-goldstein:
These tests spin up a lot of non-preemptible Query Driver jobs. This makes these tests feel not really preemptible safe, as a preemption of the test job will end up submitting duplicate Query Drivers.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2333,6 +2333,7 @@ steps:
       valueFrom: hail_run_image.image
     resources:
       cpu: '1'
+      preemptible: False
     script: |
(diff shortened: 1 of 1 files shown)


daniel-goldstein:
It is also just supremely annoying when one of these tests gets preempted 40 minutes in.

daniel-goldstein:
Ya I do. Haven't assigned that yet because I'm worried about db load, and I'm close enough to per-PR DBs anyway.

PR #13006, opened 2023-05-08
[git] ignore .metals and .bloop

danking:
These show up because I use metals and bloop for Emacs code completion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 .gradle/
 .idea/
 .vscode/
+.bloop
+.metals
 hail/libs/
 .pytest_cache/
 build/


daniel-goldstein (approved):

chrisvittal:
Not that this is a bad change per se, but this is once again a reminder that `$GIT_DIR/info/exclude` exists. 

https://git-scm.com/docs/gitignore
> Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are specific to one user’s workflow) should go into the `$GIT_DIR/info/exclude file.`

danking:
Thanks for the tip Chris! That would've addressed this in my case.

PR #13008, opened 2023-05-09
[ci] Use the ci-utils image from the new branch for db creation

daniel-goldstein:
Currently a change to `create_database.py` will only be tested once the change is in production, as opposed to in the PR where it was changed. This should fix that.


danking (approved):
Wow, how were we not already doing that? Great change.

PR #13011, opened 2023-05-09
[compiler] ensure CDA result Array[Array[Byte]] can be GCed after dec…

tpoterba:
…oding

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -2501,7 +2501,7 @@ class Emit[C](
             val eltTupled = bodySpec.encodedType.buildDecoder(bodySpec.encodedVirtualType, parentCB)
               .apply(cb, region, ib)
               .asBaseStruct
-            cb += (encRes.update(i, Code._null))
+            cb += (encRes.update(i, Code._null[Array[Byte]]))
             eltTupled.loadField(cb, 0)
           }
           cb.assign(encRes, Code._null)
--- a/hail/src/main/scala/is/hail/expr/ir/Emit.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Emit.scala
@@ -2444,7 +2444,6 @@ class Emit[C](
           val baos = mb.genFieldThisRef[ByteArrayOutputStream]()
           val buf = mb.genFieldThisRef[OutputBuffer]()
           val ctxab = mb.genFieldThisRef[ByteArrayArrayBuilder]()
-          val encRes = mb.genFieldThisRef[Array[Array[Byte]]]()
 
 
           def addContexts(cb: EmitCodeBuilder, ctxStream: StreamProducer): Unit = {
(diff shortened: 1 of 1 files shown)


hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -2498,6 +2482,9 @@ class Emit[C](
               cb.assign(stageName, stageName.concat("|").concat(dynamicID.asString.loadString(cb)))
             })
 
+          val encRes = cb.newLocal[Array[Array[Byte]]]("encRes")
+          val len = mb.newLocal[Int]("cda_result_length")
+          val ib = mb.newLocal[InputBuffer]("decode_ib")
```
danking, line 2487: why sometimes class builder sometimes method builder?

danking: More direct: How do you pick when to use mb vs cb?

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -2508,7 +2495,18 @@ class Emit[C](
...
+          cb.assign(len, encRes.length())
+          val pt = PCanonicalArray(bodySpec.encodedType.decodedSType(bodySpec.encodedVirtualType).asInstanceOf[SBaseStruct].fieldEmitTypes(0).storageType)
+          val resultArray = pt.constructFromElements(cb, region, len, deepCopy = false) { (cb, i) =>
+            cb.assign(ib, bodySpec.buildCodeInputBuffer(Code.newInstance[ByteArrayInputStream, Array[Byte]](encRes(i))))
+            val eltTupled = bodySpec.encodedType.buildDecoder(bodySpec.encodedVirtualType, parentCB)
+              .apply(cb, region, ib)
+              .asBaseStruct
+            eltTupled.loadField(cb, 0)
+          }
+          cb.assign(encRes, Code._null)
```
danking, line 2508: Is it not possible to also free each block after decoding it? As written we'd need twice the RAM to read a given result array, I think.

danking (dismissed):

hail/src/main/scala/is/hail/expr/ir/Emit.scala
tpoterba: These are aliased. Rewrote to use the more idiomatic "memoize" now and avoid explicit variable declarations.

The rule is basically to use the lowest-level builder that has the functionality you want. Module > Class > Method > Code, but codeBuilder doesn't duplicate the full functionality of everything above. So to create a field, use methodBuilder (it exists on MB and ClassB, not codeB).

hail/src/main/scala/is/hail/expr/ir/Emit.scala
tpoterba: nice

hail/src/main/scala/is/hail/expr/ir/Emit.scala
```diff
@@ -2498,6 +2482,7 @@ class Emit[C](
               cb.assign(stageName, stageName.concat("|").concat(dynamicID.asString.loadString(cb)))
             })
 
+          val encRes = cb.newLocal[Array[Array[Byte]]]("encRes")
           cb.assign(encRes, spark.invoke[BackendContext, HailClassLoader, FS, String, Array[Array[Byte]], Array[Byte], String, Option[TableStageDependency], Array[Array[Byte]]](
```
danking, line 2486: Approved, but is it possible to use `memoize` here as well or is this special in some way?

PR #13017, opened 2023-05-09
[infra][gke] disable gce metadata enable gke metadata

danking:


danking:
The workload identity config is already enabled. Using TF to do that would've destroyed the cluster for reasons unknown.

daniel-goldstein (approved):

PR #13020, opened 2023-05-09
[batch] without auto-create subnetworks, you must specify a subnet

danking:
This bug was introduced by https://github.com/hail-is/hail/pull/13016 .


danking:
See format for subnet here: https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert

daniel-goldstein (approved):

PR #13022, opened 2023-05-10
[ci] Add CreateDatabase2Step for in-namespace test and dev databases

daniel-goldstein:
These changes are a prerequisite for introducing a mysql DB pod for every test and dev namespace. The crux of such a change is any CreateDatabase steps should use the `database-server-config` from *their own namespace* (which will come in the PR that uses this step) and not from default. The other changes in this step bring the configuration for `dev` scopes closer to that of `test` scopes, because creation of test databases should really just be idempotent and there shouldn't be a difference between deploying a database in dev and test.


danking (approved):
LGTM.

danking (changes requested):
Why do we still create the db if we're not cleaning it up?

danking (approved):
If we did use this for a test/deploy/dev in which database-server-config is pointing at the main db, we would leave garbage. That is not the case for the test deployment that uses it here. Future PR reviews will need to carefully asses the use of CreateDatabase2Step.

PR #13030, opened 2023-05-10
[ci] Create a test MySQL server in test and dev namespaces

daniel-goldstein:
This change creates mysql pods for test and dev namespaces, instead of sharing the CloudSQL database server. The areas of change are as follows:

### Generation of the namespace's database-server-config
The current approach in main does a little trick. Since the current `createDatabase` step uses the `database-server-config` from default to generate admin/user sql configs, the CI pipeline creates a dummy database `test-database-instance` to create a `sql-test-instance-admin-config` that inherits the credentials from the production `database-server-config`, and then copies that within the test namespace to `database-server-config`.

In this change, since we are creating the server ourselves, we can just replace these with a step that creates a `database-server-config` from scratch, and then uses that for the DB pod.

Overall making these changes really gave me the heebie jeebies that the test and dev namespaces have all these credentials to the CloudSQL server. I'm glad this gets rid of that.

### Accessing the database server
We use the DB pod's service DNS name as the `host` so inside Kubernetes this Just Works. The one caveat is the CI pipeline in which we run migrations in batch jobs. Those jobs need a way to reach the DB pod. I achieve this with a NodePort and then use the job's K8s credentials to resolve the node and port that the DB is on. The code I've added to do this resolution feels a bit janky, wouldn't mind some feedback on that. In terms of security, if a user job was able to somehow resolve the address of a test db, they would still not have the credentials to access it, and this is currently also the case with the production database. Nevertheless, this does raise an action item that we should only allow traffic to the k8s and DB subnets for `network=private` jobs, but I think we should make that a separate PR.

### Database creation
In order to test this properly in a dev deploy, I needed to make some changes to `create_database.py`. In main, dev deploys can't create databases. I think they should be able to, and those operations should just be idempotent. When allowing dev deploys to create databases, I hit the `ALTER USER` lines in `create_database.py` which lock out any already-deployed application, which feels silly. Instead, I create the mysql user and create the corresponding secret iff the mysql username does not already exist.

### create_initial_account.py
When starting off with a fresh database, we encounter a bootstrapping problem where we need to add the developer's user. The current way to add the developer's user is through `create_initial_account.py` which assumes that the developer's gsa key secret is not already in the namespace, but now it could be, so I delete the key in the namespace if it exists before creating it. This could all change with my other PR to add devs to test namespaces. But this change to allow ephemeral dev databases will make testing that other PR *way* easier.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2674,6 +2674,7 @@ steps:
       cp -R /io/repo/ci/* ./ci/
       cp /io/repo/tls/Dockerfile ./ci/test/resources/Dockerfile.certs
       cp /io/repo/tls/create_certs.py ./ci/test/resources/
+      cp /io/repo/tls/create_test_db_config.sh ./ci/test/resources/
       cp /io/repo/pylintrc ./
       cp /io/repo/setup.cfg ./
       cp /io/repo/pyproject.toml ./
--- a/build.yaml
+++ b/build.yaml
@@ -1732,6 +1732,7 @@ steps:
       - delete_monitoring_tables
       - ci_utils_image
       - create_test_database_server_config
+      - deploy_test_db
   - kind: deploy
     name: deploy_monitoring
     namespace:
@@ -1978,6 +1979,7 @@ steps:
       - delete_ci_tables
       - ci_utils_image
       - create_test_database_server_config
+      - deploy_test_db
   - kind: createDatabase2
     name: batch_database
     databaseName: batch
@@ -2213,6 +2215,7 @@ steps:
       - delete_batch_tables
       - ci_utils_image
       - create_test_database_server_config
+      - deploy_test_db
   - kind: deploy
     name: deploy_batch
     namespace:
@@ -3041,6 +3044,7 @@ steps:
       - merge_code
       - ci_utils_image
       - create_test_database_server_config
+      - deploy_test_db
   - kind: deploy
     name: deploy_notebook
     namespace:
--- a/build.yaml
+++ b/build.yaml
@@ -255,35 +255,43 @@ steps:
     dependsOn:
       - base_image
       - merge_code
-  - kind: createDatabase
-    name: test_database_instance
-    databaseName: test-instance
-    image:
-      valueFrom: ci_utils_image.image
-    migrations: []
-    namespace:
-      valueFrom: default_ns.name
-    scopes:
-      - test
-    dependsOn:
-      - default_ns
-      - ci_utils_image
   - kind: runImage
-    name: create_database_server_config
+    name: create_test_database_server_config
     image:
-      valueFrom: ci_utils_image.image
+      valueFrom: create_certs_image.image
     script: |
-     kubectl -n {{ default_ns.name }} get -o json secret {{ test_database_instance.admin_secret_name }} | jq '{apiVersion, kind, type, data, metadata: {name: "database-server-config"}}' | kubectl -n {{ default_ns.name }} apply -f -
+      set -ex
+
+      if ! kubectl get secret -n {{ default_ns.name }} database-server-config;
+      then
+          NAMESPACE={{ default_ns.name }} bash /create_test_db_config.sh
+      fi
     serviceAccount:
       name: admin
       namespace:
         valueFrom: default_ns.name
     scopes:
+      - dev
       - test
     dependsOn:
       - default_ns
-      - ci_utils_image
-      - test_database_instance
+      - create_certs_image
+  - kind: deploy
+    name: deploy_test_db
+    namespace:
+      valueFrom: default_ns.name
+    config: docker/mysql/db.yaml
+    wait:
+      - kind: Service
+        name: db
+        for: alive
+        resource_type: statefulset
+    scopes:
+      - dev
+      - test
+    dependsOn:
+      - default_ns
+      - create_test_database_server_config
   - kind: buildImage2
     name: admin_pod_image
     dockerFile: /io/repo/admin-pod/Dockerfile
@@ -316,8 +324,8 @@ steps:
       - default_ns
       - admin_pod_image
       - merge_code
-      - create_database_server_config
-  - kind: createDatabase
+      - create_test_database_server_config
+  - kind: createDatabase2
     name: auth_database
     databaseName: auth
     image:
@@ -354,6 +362,8 @@ steps:
       - merge_code
       - delete_auth_tables
       - ci_utils_image
+      - create_test_database_server_config
+      - deploy_test_db
   - kind: runImage
     name: create_deploy_config
     image:
@@ -525,7 +535,7 @@ steps:
       - create_deploy_config
       - deploy_auth_driver_service_account
       - create_test_gsa_keys
-      - create_database_server_config
+      - create_test_database_server_config
   - kind: runImage
     name: create_initial_user
     runIfRequested: true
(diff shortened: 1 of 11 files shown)


build.yaml
```diff
@@ -255,35 +255,43 @@ steps:
...
-    name: create_database_server_config
+    name: create_test_database_server_config
     image:
-      valueFrom: ci_utils_image.image
+      valueFrom: create_certs_image.image
     script: |
-     kubectl -n {{ default_ns.name }} get -o json secret {{ test_database_instance.admin_secret_name }} | jq '{apiVersion, kind, type, data, metadata: {name: "database-server-config"}}' | kubectl -n {{ default_ns.name }} apply -f -
+      set -ex
+
+      if ! kubectl get secret -n {{ default_ns.name }} database-server-config;
```
daniel-goldstein, line 265: We should delete all sql configs in dev namespaces because they currently point to the production server.

danking: I think you're the only one using them this week so I say blow it all away as soon as this merged.

ci/create_database.py
```diff
@@ -89,79 +85,54 @@ async def create_database():
...
-
-    user_exists = await db.execute_and_fetchone("SELECT user FROM mysql.user WHERE user=%s", (user_username,))
-    user_exists = user_exists and user_exists.get('user') == user_username
-
-    create_admin_or_alter_password = (
-        f"CREATE USER '{admin_username}'@'%' IDENTIFIED BY '{admin_password}';"
-        if not admin_exists
-        else "ALTER USER '{admin_username}'@'%' IDENTIFIED BY '{admin_password}';"
-    )
+    async def create_user_idempotent(admin_or_user, mysql_username, mysql_password):
```
danking, line 88: This isn't really idempotent because someone could race between your SELECT and the CREATE USER. I'm not sure that matters though. There's only one person creating users at any time, right?

ci/create_database.py
```diff
@@ -89,79 +85,54 @@ async def create_database():
...
+        existing_user = await db.execute_and_fetchone('SELECT user FROM mysql.user WHERE user=%s', (mysql_username,))
+        if existing_user is not None and existing_user.get('user') == mysql_username:
```
danking, line 90: the second condition is always true if the first condition is true, right? I'm fine making it an assertion below.

ci/create_database.py
danking: Well, I guess what I mean to say is: it *is* single-writer idempotent but not *atomic* so with multiple writers its not idempotent.

gear/gear/database.py
```diff
@@ -77,6 +80,19 @@ async def aexit(acontext_manager, exc_type=None, exc_val=None, exc_tb=None):
     return await acontext_manager.__aexit__(exc_type, exc_val, exc_tb)
 
 
+async def resolve_test_db_endpoint(sql_config: SQLConfig) -> SQLConfig:
+    service_name, namespace = sql_config.host[: -len('.svc.cluster.local')].split('.', maxsplit=1)
```
danking, line 84: we're so close to `removesuffix`!

gear/gear/database.py
```diff
@@ -77,6 +80,19 @@ async def aexit(acontext_manager, exc_type=None, exc_val=None, exc_tb=None):
...
+    await kubernetes_asyncio.config.load_kube_config()
+    async with kubernetes_asyncio.client.ApiClient() as api:
+        client = kubernetes_asyncio.client.CoreV1Api(api)
+        db_service = await client.read_namespaced_service(service_name, namespace)
+        db_pod = await client.read_namespaced_pod(f'{db_service.spec.selector["app"]}-0', namespace)
+        sql_config_dict = sql_config.to_dict()
+        sql_config_dict['host'] = db_pod.status.host_ip
```
danking, line 91: this feels a bit flaky no? Are we not able to access this through the service's DNS name?

ci/create_initial_account.py
```diff
@@ -19,6 +19,14 @@ async def copy_identity_from_default(hail_credentials_secret_name: str) -> str:
 
     secret = await k8s_client.read_namespaced_secret(hail_credentials_secret_name, 'default')
 
+    try:
+        await k8s_client.delete_namespaced_secret(hail_credentials_secret_name, NAMESPACE)
```
danking, line 23: what's up with this?

gear/gear/database.py
danking: Ah. The workers. And exposing the service to the workers would require a private load balancer, etc. etc. annoying costly.

danking (changes requested):
Some questions, comments. The thing about the cores doesn't need to be resolved in this PR. I had two things to take action on: the change to create_initial_account.py and line 90 of create_database.py

gear/gear/database.py
daniel-goldstein: ya exactly. I didn't want to allocate a whole internal load balancer just to run a few couple minute migrations.

tls/create_test_db_config.sh
```diff
@@ -0,0 +1,71 @@
...
+dir=$(mktemp -d)
+cd $dir
+
+# Create the MySQL server CA
+openssl req -new -x509 \
+    -subj /CN=db-root -nodes -newkey rsa:4096 \
+    -keyout server-ca-key.pem -out server-ca.pem
+
+create_key_and_cert server db
+create_key_and_cert client client
```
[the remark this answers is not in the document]
daniel-goldstein: Yup this is cruft I forgot to erase

ci/ci/build.py
```diff
@@ -1164,3 +1166,164 @@ def cleanup(self, batch, scope, parents):
...
+            self.create_passwords_job = batch.create_job(
+                self.image,
+                command=['bash', '-c', create_passwords_script],
+                attributes={'name': self.name + "_create_passwords"},
+                output_files=[(x[1], x[0]) for x in password_files_input],
+                parents=self.deps_parents(),
+                regions=[REGION],
+            )
+
+        n_cores = 4 if scope == 'deploy' and not is_test_deployment else 1
```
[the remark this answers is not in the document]
daniel-goldstein: Ya these are good points. Let's talk about this in a follow-up when Jackie's back as she probably has the most insight on this.

ci/create_database.py
daniel-goldstein: Good point, I think I'm abusing idempotent a little bit here. What I really mean is just `create_user_if_doesnt_exist`.

ci/create_database.py
daniel-goldstein: Ya true I was just shifting code around. Just going to delete this check and `SELECT 1`.

ci/create_initial_account.py
daniel-goldstein: Because I was doing this in my namespace that already had a `dgoldste-gsa-key` secret but not a user in the database, the `create_namespaced_secret` failed. I needed a way to force create a secret even if one already existed and this is how we do it in auth/driver.py

gear/gear/database.py
daniel-goldstein: It does feel a bit flaky, which is why I specifically don't do this if you're running in k8s. I am implicitly assuming here that the DB pod is not going to be relocated while the create_database job is running, which I guess would pose a problem anyway. This seemed the most reasonable, albeit a little janky, approach in terms of cost and complexity given that these are just test databases.

danking (approved):
Such a good change!

danking:
Let's get this in so we can try scaling up the service backend tests.

danking:
ci test hung

PR #13035, opened 2023-05-10
[batch] Give cloud-specific API implementations more control over container registry credentials

daniel-goldstein:
Just some more shoving of implementation details into `CloudWorkerAPI` so that credential objects are more of a black box to `worker.py`. Can e.g. change up certain implementations to use oauth2tokens for users.


danking (approved):
looks great.

PR #13038, opened 2023-05-11
[ci] Remove redundancies between the two build_hail steps

daniel-goldstein:
- No longer build the normal jar and wheel twice. build_hail now only produces the debug jar+test-jar+wheel and the test jar+wheel

- Remove hail.zip in favor of installing from the wheel. For some reason this breaks the python doc tests because presumably spark can't find the hail jar. I don't understand entirely why this is the case. I'm installing from the wheel instead of what `Dockerfile.hail-base` used to do which was copying the jar under `SPARK_HOME`, but I hoped that pip installing the wheel meant I wouldn't have to deal with that.
- Remove other outputs from the build_hail job that weren't used or didn't need to be proxied through this step.


daniel-goldstein:
@danking Any idea why the new `hail-run` image would error on just the `test_python_docs` step with the error about spark not finding the hail jar?

build.yaml
```diff
@@ -1397,26 +1362,25 @@ steps:
               --doctest-glob='*.rst' \
               --ignore=docs/conf.py \
               --ignore=docs/doctest_write_data.py
```
danking, line 1364: @daniel-goldstein first guess: some shenanigans with the python path. The stack traces point at the `/io/hail/python/hail` directory:
```
/io/hail/python/hail/conftest.py:37: 
```
Instead of `cd`'ing can you just `python3 -m pytest hail/python/hail/ ...`? You should be able to replicate locally by installing (non editable) and then cd'ing and running pytest.

build.yaml
```diff
@@ -724,32 +695,39 @@ steps:
...
-      cpu: '4'
+      cpu: '2'
     script: |
       set -ex
       cd /io/repo/hail
       chmod 755 ./gradlew
       time retry ./gradlew --version
       time retry make shadowJar wheel
+
+      # Check wheel size is small enough for pypi (< 100 MB)
```
danking, line 706: message is out of date

danking (changes requested):

daniel-goldstein:
Closing in favor of #13076 🚀

PR #13040, opened 2023-05-11
[qob] only use highmem for one test

danking:
The service backend is shared among all the threads which means we were changing the resource requests of random other tests. This substantially delays certain tests when, say, 16 quick partitions require 8 highmem cores each.


danking:
cc: @tpoterba This is a bit of a hack for testing purposes, but I think we will eventually want to have a way to reconfigure Hail Query backend settings on a per-action basis.

tpoterba (approved):

daniel-goldstein:
@danking looks like the `Backend` isn't too happy about the extra `kwargs`

PR #13043, opened 2023-05-11
[transient-errors] will it ever end?

danking:
I also cleaned up our match statement and ensured we recur through every kind of exception just like Python.


daniel-goldstein (approved):

daniel-goldstein:
Throwing a WIP on until the migration fix merges

danking:
Somehow this change encountered several results files not existing or being size 0. That is super suspicious that either GCS is not consistent or we're returning before writes are flushed. Not sure why the changes in this PR would make that suddenly more common though.

PR #12987, opened 2023-05-04
updates changelog for 0.2.116

iris-garden:

PR #13047, opened 2023-05-14
Correct the 0.2.116 release date [very minor]

jmarshall:
For posterity, update the date in the change log to reflect when PR #12987 was merged and the release made.

Very minor, and probably already less important than it was last week, but for future readers it's useful for these to be aligned…


daniel-goldstein (approved):
Thanks!

PR #13051, opened 2023-05-15
[hailctl] Fix validation for remote_tmpdir to allow azure https scheme

### LINKED ISSUE #13049: https:// is considered invalid by hailctl config set batch/remote_tmpdir
### What happened?

> In regards to the https vs hail-az, I get an error for https:
> hailctl config set batch/remote_tmpdir https://kahlquisrefsa.blob.core.windows.net/test
> Error: bad value 'https://kahlquisrefsa.blob.core.windows.net/test' for parameter 'batch/remote_tmpdir' should be valid cloud storage URI such as gs://my-bucket/batch-tmp/

### Version

0.2.116

### Relevant log output

_No response_

daniel-goldstein:
Validation code was not updated when adding azure https support. Made this check a bit more robust and tested locally.

Resolves #13049

### DIFF
--- a/hail/python/hailtop/hailctl/config/cli.py
+++ b/hail/python/hailtop/hailctl/config/cli.py
@@ -5,11 +5,12 @@ import re
 import warnings
 
 from hailtop.config import get_user_config, get_user_config_path
+from hailtop.aiotools.router_fs import RouterAsyncFS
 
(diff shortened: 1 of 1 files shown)


danking (approved):
Thanks for picking this up!

PR #13053, opened 2023-05-15
[batch] include instance name in labels and label disks

danking:
AFAICT, labels are free. Without this, I cannot narrow the cost of, say, a Local SSD, down to one instance. https://cloud.google.com/resource-manager

Also, just, generally, I need to be able to know to which namespace a disk belongs.


daniel-goldstein (approved):

danking:
The instance names might actually be automatically added. Still trying to dig through all this but there's definitely some disk that has no namespace attached to it.

danking:
whoops this clearly broke batch somehow

PR #13055, opened 2023-05-15
[batch] Dont schedule on instances that dont match the current instance version

daniel-goldstein:
Long awaited, this change prompts the batch driver to only schedule jobs on workers with the most recent instance version, i.e. matches the `INSTANCE_VERSION` global variable. This way we can make backwards incompatible changes between the worker and driver without having to manually kill the whole fleet. This will allow pre-existing workers to finish gracefully, as they will just stop receiving work when the new batch driver is deployed and eventually die off.

### Scheduler changes
Just skips instances where the instance version doesn't match `INSTANCE_VERSION`

### Autoscaler changes
Cluster stats like free mcpu and live instances are tracked per instance version. The autoscaler now only looks at instances of the latest version when deciding whether it needs more workers. This way we don't get stuck unable to schedule new jobs until the old workers die off because there technically are enough cores available to meet demand but they are from old workers.

### DIFF
--- a/batch/batch/driver/templates/index.html
+++ b/batch/batch/driver/templates/index.html
@@ -48,12 +48,12 @@
       <td><a href="{{ base_path }}/inst_coll/pool/{{ pool.name }}">{{ pool.name }}</a></td>
       <td>{{ pool.worker_type }}</td>
       <td>{{ pool.preemptible }}</td>
-      <td class="numeric-cell">{{ pool.n_instances_by_state['pending'] }}</td>
-      <td class="numeric-cell">{{ pool.n_instances_by_state['active'] }}</td>
-      <td class="numeric-cell">{{ pool.n_instances_by_state['inactive'] }}</td>
-      <td class="numeric-cell">{{ pool.n_instances_by_state['deleted'] }}</td>
-      <td class="numeric-cell">{{ pool.live_total_cores_mcpu / 1000 }}</td>
-      <td class="numeric-cell">{{ pool.live_free_cores_mcpu / 1000 }}</td>
+      <td class="numeric-cell">{{ pool.current_worker_version_stats.n_instances_by_state['pending'] }}</td>
+      <td class="numeric-cell">{{ pool.current_worker_version_stats.n_instances_by_state['active'] }}</td>
+      <td class="numeric-cell">{{ pool.current_worker_version_stats.n_instances_by_state['inactive'] }}</td>
+      <td class="numeric-cell">{{ pool.current_worker_version_stats.n_instances_by_state['deleted'] }}</td>
+      <td class="numeric-cell">{{ pool.current_worker_version_stats.live_total_cores_mcpu / 1000 }}</td>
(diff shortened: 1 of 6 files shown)


daniel-goldstein:
I tested this by hand with the following schenario:

1. Deployed this branch and set the standard pool to use 4-core workers. Then submitted two jobs, with a total of 3 cores requested between them. The state looked as follows after a worker spun up:
Screenshot 2023-05-15 at 5 16 30 PM


2. I then bumped the instance version and redeployed. Then submitted a 1 core job. This should be able to fit on the existing worker, but batch instead spins up a new worker because the instance version has changed to 25 and the existing worker is version 24.
Screenshot 2023-05-15 at 5 19 41 PM

Screenshot 2023-05-15 at 5 20 59 PM

3. The following steps I consider additional verification. I submitted another job, and observed it falling on the new worker instead of the most utilized worker.

Screenshot 2023-05-15 at 5 21 36 PM


4. I then cancelled the first batch and submitted another job and observed it fell on the new worker instead of the least utilized worker. So this step and the previous step confirm that regardless of utilization the old worker is not considered. We also see that the old worker receives no more work and successfully deactivates.
Screenshot 2023-05-15 at 5 23 03 PM

batch/batch/driver/templates/index.html
```diff
@@ -48,12 +48,12 @@ <h2>Pools</h2>
...
-      <td class="numeric-cell">{{ pool.n_instances_by_state['inactive'] }}</td>
-      <td class="numeric-cell">{{ pool.n_instances_by_state['deleted'] }}</td>
-      <td class="numeric-cell">{{ pool.live_total_cores_mcpu / 1000 }}</td>
-      <td class="numeric-cell">{{ pool.live_free_cores_mcpu / 1000 }}</td>
+      <td class="numeric-cell">{{ pool.current_worker_version_stats.n_instances_by_state['pending'] }}</td>
+      <td class="numeric-cell">{{ pool.current_worker_version_stats.n_instances_by_state['active'] }}</td>
+      <td class="numeric-cell">{{ pool.current_worker_version_stats.n_instances_by_state['inactive'] }}</td>
+      <td class="numeric-cell">{{ pool.current_worker_version_stats.n_instances_by_state['deleted'] }}</td>
+      <td class="numeric-cell">{{ pool.current_worker_version_stats.live_total_cores_mcpu / 1000 }}</td>
+      <td class="numeric-cell">{{ pool.current_worker_version_stats.live_free_cores_mcpu / 1000 }}</td>
```
danking, line 56: Hmm. I'm a bit worried that this isn't clearly communicated in the UI. I think we should include the total over all versions somewhere.

Relatedly, if its simple to modify while we're here, it has *long* bothered me that we include the "activating" cores in the "live total". I feel like I want three notions:
- `{{free}} / {{total}}` for "schedule-able"
- `{{ total }}` "activating"
- `{{ total }}` not-inactive (so, anything we're paying for)

danking (dismissed):

daniel-goldstein:
Confirmed that the UI still works

danking (approved):

PR #13056, opened 2023-05-15
[batch] Let worker control job deploy-config instead of using k8s secret

daniel-goldstein:
In a world without gsa keys and hail auth tokens in secrets, the only other k8s secrets that we sometimes add to the job spec are the user ssl config which I believe to be unused (will address in a separate PR) and the `worker-deploy-config` secret. I don't really get why this is in a secret though, this is information already known to the worker and we can effortlessly add this to every job. Not only that, we can write the deploy-config once and mount it as read-only in all containers.


daniel-goldstein:
Marking as WIP until the PR to gracefully handle new instance versions merges.

PR #13058, opened 2023-05-16
Azure-redeploy-fixes-upstream

violetbrina:
We've had to do a redeploy of our hail batch instance on Azure. This PR resolves/clarifies two issues we encountered.

1) Storage Account Name Uniqueness

Due to Azure's restrictions on storage account naming (mainly that names must be globally unique) the redeploy did not succeed.

This is because the resource group name (we chose to reuse hail) is possible under a new subscription, but the generated storage account names were therefore identical to our previous stack.

I've added in an argument called `storage_account_suffix` to account for this issue. It can be set to any arbitrary string that complies with Azure's storage account naming scheme in order to avoid naming conflicts in the future.

While the option remains to simply choose a novel resource group name this is not enforced by Azure and anyone deploying a stack similarly named to someone else would not know until the `terraform apply` stage that the name would not work.

2) Mysql Flexible Server Zones

The only other issue is that the zone argument for the mysql flexible server is no longer always valid depending on your compute region. We needed to comment it out for a successful deploy in Australia East.

The comment that has been added we hope will be helpful for others in future.

### DIFF
--- a/infra/azure/main.tf
+++ b/infra/azure/main.tf
@@ -94,6 +94,7 @@ module "batch" {
   batch_test_user_storage_account_name = var.batch_test_user_storage_account_name
   resource_group        = data.azurerm_resource_group.rg
   container_registry_id = azurerm_container_registry.acr.id
+  storage_account_suffix = var.storage_account_suffix
 }
 
 module "global_config" {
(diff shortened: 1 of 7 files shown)


daniel-goldstein (approved):
Thank you for these fixes!

Issue #13061, opened 2023-05-16
In Query-on-Batch in Azure, a 1500 partition VCF fails to import


danking:
### What happened?

Example failure: https://batch.azure.hail.is/batches/3883658 (nb: driver job failed even though its marked success).

User report: https://hail.zulipchat.com/#narrow/stream/223457-Hail-Batch-support/topic/problem.20writing.20output.3F

### Version

0.2.116

PR #13065, opened 2023-05-16
[qob] remove resource leak in readNoCompression

danking:
CHANGELOG: In Azure Query-on-Batch, fix a resource leak that prevented running pipelines with >500 partitions and created flakiness with >250 partitions.

fixes: https://github.com/hail-is/hail/issues/13061


danking:
I verified this scales all the way to 50,000 partitions but the batch-driver can't schedule these fast enough to make the test fast. They take less than a second but more than ~300ms. We'd need like 64 8-core (512 cores) nodes to bring test time down to a reasonable amount. We should strive to get there but batch would need to schedule at 512 jobs per second for that to make sense.

daniel-goldstein (approved):

daniel-goldstein:
@danking Gotta regenerate the dev pinned requirements.

danking:
New ruff version found new issues which I fixed. Should pass now/

PR #13067, opened 2023-05-16
[qob] immediateFlush=true for QoB logger

danking:
CHANGELOG: In Query-on-Batch, driver and worker logs no longer buffer so messages should arrive in the UI after a fixed delay rather than proportional to the frequency of log messages.


tpoterba (approved):

tpoterba:
huh, thought that was the default. Guess not.

danking:
It is and I explicitly overrode the default. my bad!

PR #13068, opened 2023-05-16
[batch] Delete unused batch user code ssl config

daniel-goldstein:


daniel-goldstein:
Perhaps this is not unused…

PR #13071, opened 2023-05-17
[batch] Dont use the authorization header for worker tokens to the driver

daniel-goldstein:
Workers in terra batch need to use an actual azure bearer token in the authorization header to get through terra auth. I figured it's easiest to just not overload that header.

### DIFF
--- a/batch/batch/driver/main.py
+++ b/batch/batch/driver/main.py
@@ -87,6 +87,13 @@ def instance_from_request(request):
     return inst_coll_manager.get_instance(instance_name)
 
 
+# Old workers use the Authorization header for their identity token
+# but that can conflict with bearer tokens used when the driver is behind another
+# auth mechanism
+def instance_token(request):
+    return request.headers.get('X-Hail-Instance-Token') or authorization_token(request)
+
+
 def activating_instances_only(fun):
     @wraps(fun)
     async def wrapped(request):
@@ -100,7 +107,7 @@ def activating_instances_only(fun):
             log.info(f'instance {instance.name} not pending')
             raise web.HTTPUnauthorized()
 
-        activation_token = authorization_token(request)
+        activation_token = instance_token(request)
         if not activation_token:
             log.info(f'activation token not found for instance {instance.name}')
             raise web.HTTPUnauthorized()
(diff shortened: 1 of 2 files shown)


danking:
yesssss

batch/batch/driver/main.py
```diff
@@ -87,6 +87,13 @@ def instance_from_request(request):
     return inst_coll_manager.get_instance(instance_name)
 
 
+# Old workers use the Authorization header for their identity token
```
danking, line 90: After old workers die, we can remove this code, right?

danking (approved):
ah, I thought this was the access token change. 😢

batch/batch/driver/main.py
daniel-goldstein: Ya

daniel-goldstein:
@danking So would be curious to get your thoughts. I was initially going to make this change such that instead of these activation tokens the batch worker authenticates with a cloud access token. I had a minor pause though because this means that other workers (even from other namespaces) could potentially impersonate each other, whereas they cannot in our current token system. Is that a concern to you? I suppose we are already pretty compromised if someone gets control of the batch worker's identity, considering the buckets that the batch worker has access to.

danking:
Hmm. Good question. To impersonate an instance currently, you need its name and the activation token, and you need to make your request before the instance does. Names are slightly more public than activation tokens. They appear in the batch UI in various places (though, for end users, usually only after activation).

The name has fewer bits of randomness than the activation token, so the activation token makes it much harder to correctly guess both values.

I think I prefer keeping it. VM Metadata isn't secret, but we're very unlikely to accidentally leak something called "activation_token" whereas I think we should treat the names as basically public information.

Does that make sense to you?

daniel-goldstein:
Ya that's roughly the conclusion I came to. I think worker names are definitely public and should be treated as such, so it does feel a step backward in a way to drop these activation tokens. We can always also add bearer tokens using the worker identity, then you would need 2 secrets. But that also requires this change, which is why I made it standalone.

daniel-goldstein:
Adding bearer tokens would essentially add an element of expiration. our own issued tokens prove you are that particular instance, the bearer token proves you are the batch worker identity (at least that the batch worker identity requested an access token in the past X minutes)

PR #13073, opened 2023-05-17
[batch] Remove unnecessary get_compute_client method on CloudWorkerAPI

daniel-goldstein:
worker.py requests the compute client from the worker API just so it can close it at the end. Having a close method instead on the worker API means we don't need to create a compute client in environments that don't support disks (azure and azure-terra).


daniel-goldstein:
@jigold bump

jigold (approved):

PR #13082, opened 2023-05-18
[batch] Fix Python to add correct deduped resource id

jigold:


daniel-goldstein (approved):
A wonderfully simple fix

PR #13088, opened 2023-05-19
[batch] Exponential backoff of resource usage monitoring

jigold:


danking (approved):
omg amazing, yes please!

PR #13089, opened 2023-05-22
[query] fix requester pays in all Query backends

danking:
CHANGELOG: Requester pays buckets now work in `hailtop.fs` and `hl.hadoop_*`. This has been broken since at least 0.2.115.

I first check for an explicit argument or the standard hailctl configuration. If neither of those exist, I try to parse spark-defaults.conf with lots of error handling and warning.


hail/python/hailtop/batch/batch_pool_executor.py
```diff
@@ -132,7 +131,14 @@ def __init__(self, *,
...
         self.outputs = self.directory + 'outputs/'
-        self.fs = RouterAsyncFS(gcs_kwargs={'gcs_requester_pays_configuration': project})
+        if project is not None:
+            warnings.warn(
+                'The project parameter of BatchPoolExecutor is deprecated. Please '
+                'use the gcs_requester_pays_project parameter of ServiceBackend.'
+            )
+            self._fs = RouterAsyncFS(gcs_kwargs={'gcs_requester_pays_configuration': project})
+        else:
+            self._fs = None
```
danking, line 141: This just seemed like old cruft. If there's no project, we should just use the backend's configuration and FS.

hail/python/hailtop/config/user_config.py
```diff
@@ -117,3 +118,133 @@ def get_remote_tmpdir(caller_name: str,
     if remote_tmpdir[-1] != '/':
         remote_tmpdir += '/'
     return remote_tmpdir
+
+
+def get_gcs_requester_pays_configuration(
+        *,
+        gcs_requester_pays_configuration: Optional[Union[str, Tuple[str, List[str]]]] = None,
+) -> Optional[Union[str, Tuple[str, List[str]]]]:
```
daniel-goldstein, line 126: Shouldn't these use the `GCSRequesterPaysConfiguration` type?

daniel-goldstein (changes requested):
Just a few small things.

hail/python/hailtop/config/user_config.py
```diff
@@ -117,3 +118,133 @@ def get_remote_tmpdir(caller_name: str,
...
+    if gcs_requester_pays_configuration:
+        return gcs_requester_pays_configuration
+
+    project = configuration_of('gcs_requester_pays', 'project', None, None)
+    buckets = configuration_of('gcs_requester_pays', 'buckets', None, None)
```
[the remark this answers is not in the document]
danking: Python's parser doesn't seem to support sub-sections. https://docs.python.org/3.4/library/configparser.html#supported-ini-file-structure

I'm not exactly sure what to do here to be honest. Maybe its time to find a more sophisticated config file format / parser?

hail/python/hailtop/config/user_config.py
```diff
@@ -117,3 +120,146 @@ def get_remote_tmpdir(caller_name: str,
...
+                        mode = SparkConfGcsRequesterPaysMode.AUTO
+                    elif val == SparkConfGcsRequesterPaysMode.CUSTOM:
+                        mode = SparkConfGcsRequesterPaysMode.CUSTOM
+                    else:
+                        raise ValueError(
+                            f'When reading GCS requester pays configuration from '
+                            f'spark-defaults.conf ({path}) an unknown mode was '
+                            f'found: {val}. Expected ENABLED, AUTO, CUSTOM, or '
+                            f'DISABLED.'
+                        )
```
daniel-goldstein, line 258: I'm pretty sure you should just be able to do `mode = SparkConfGcsRequesterPaysMode(val)`

daniel-goldstein (changes requested):
Does my suggestion help?

hail/python/hailtop/config/user_config.py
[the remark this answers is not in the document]
daniel-goldstein: Hm, that's annoying. I guess that's where [TOML has the upper hand](https://toml.io/en/v1.0.0#table).

PR #13098, opened 2023-05-23
[batch] Add resource usage plots for JVM containers

jigold:


batch/batch/worker/worker.py
```diff
@@ -847,12 +846,15 @@ async def create(self):
                 raise ContainerCreateError from e
             raise
 
-    def start(self):
+    def start(self, monitor: Optional[ResourceUsageMonitor] = None):
+        if monitor is None:
+            monitor = ANullContextManager()
```
danking, line 851: This will fail types because this isn't a `ResourceUsageMonitor`.

batch/batch/worker/worker.py
```diff
@@ -1756,13 +1746,20 @@ def __init__(
             )
 
         self.containers = containers
+        self.monitors = {
+            task_name: c.new_resource_usage_monitor(self.resource_usage_path(task_name))
+            for task_name, c in containers.items()
+        }
```
danking, line 1752: I prefer we do not lift monitors up to the Job level. This feels like a violation of the abstraction boundary. The container should be responsible for monitoring its resources and returning the results. It does seem worthwhile for a container to hold on to the resource monitor class and use that to `read` the file.

batch/batch/resource_usage.py
```diff
@@ -1,22 +1,34 @@
...
+from typing import Optional, Tuple, TYPE_CHECKING
 
 import numpy as np
 import pandas as pd
 
 from hailtop.utils import check_shell_output, sleep_and_backoff, time_msecs, time_ns
+from hailtop.aiotools.fs import AsyncFS
+
+if TYPE_CHECKING:
+    from .worker.worker import Container  # pylint: disable=cyclic-import
```
danking, line 17: We should go to significant lengths to avoid introducing new circularities into our codebase. Let's see if we can avoid it.

batch/batch/resource_usage.py
```diff
@@ -65,19 +77,17 @@ def _decode_to_df(data: bytes) -> Optional[pd.DataFrame]:
 
     def __init__(
         self,
-        container_name: str,
-        container_overlay: str,
-        io_volume_mount: Optional[str],
-        veth_host: str,
+        container: 'Container',
```
danking, line 80: I detail this below, but this feels like an abstraction violation. The Container and ResourceUsageMonitor classes are now fully entangled.

batch/batch/worker/worker.py
```diff
@@ -1047,13 +1053,7 @@ async def _run_container(self) -> bool:
...
                     assert self.netns
 
-                    async with ResourceUsageMonitor(
-                        self.name,
-                        self.container_overlay_path,
-                        self.io_mount_path,
-                        self.netns.veth_host,
-                        self.resource_usage_path,
-                    ):
+                    async with monitor:
```
danking, line 1056: I don't think we should stop monitoring a JVM's resources. In fact, I think we should eventually have a per-worker page that includes information like JVM resource usage, JVM logs, and other worker metadata.

If we leave resource monitoring here, as it already is, I think it keeps the code simpler. Let's do that.

batch/batch/worker/worker.py
```diff
@@ -2302,7 +2321,7 @@ def get_container_log_path(self, container_name: str) -> str:
         return self.log_file
 
     async def get_resource_usage(self) -> Dict[str, bytes]:
-        return {'main': ResourceUsageMonitor.no_data()}
+        return {'main': await read_resource_usage(self.worker.fs, self.resource_usage_file)}
```
danking, line 2324: Hmm. I like what you did in the Bash job side where we only read this through the resource usage monitor. It does seem like there's no elegant way to achieve that here too.

danking (changes requested):

danking:
This is awesome! Really excited to give this visibility to everyone!

jigold:
@danking Can you give it a second look? If you're happy with this structure, then I can test it in my namespace.

danking (approved):
This looks great!

PR #13106, opened 2023-05-23
[query] Specialize export_blocks into it's own writer

chrisvittal:


chrisvittal:
@patrick-schultz this was too stale

PR #13113, opened 2023-05-24
[query/ggplot] Shows legend group for single-value columns

iris-garden:
CHANGELOG: `hail.ggplot.geom_point` now displays a legend group for a column even when it has only one value in it.

Currently, if a column with only one value in it is passed to either the `color` or the `shape` aesthetic for `geom_point`, the generated legend will not include a group with a single entry for that aesthetic. However, R's ggplot2 library in the same situation includes such a legend group.

This change brings our ggplot implementation into line with R's ggplot2 in that regard.

For example, this code:

```python
import hail as hl
ht = hl.utils.range_table(10)
ht = ht.annotate(threeven="good")
fig = (
    + geom_point(aes(color=ht.even, shape=ht.threeven))
)
fig.show()
```

generates a plot legend like this on `main`:

Screenshot 2023-05-24 at 13 19 49

but generates one like this on this branch:

Screenshot 2023-05-24 at 13 19 16

### DIFF
--- a/hail/python/hail/ggplot/geoms.py
+++ b/hail/python/hail/ggplot/geoms.py
@@ -185,7 +185,8 @@ class GeomPoint(Geom):
             traces.append([fig_so_far, df, facet_row, facet_col, values, trace_categories])
 
         non_empty_legend_groups = [
-            legend_group for legend_group in legends.values() if len(legend_group) > 1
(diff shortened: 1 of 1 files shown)


danking (approved):
LGTM

PR #13114, opened 2023-05-24
[docs] Expose hailtop.fs and hailtop.batch in query docs

jigold:
Screen Shot 2023-05-24 at 3 59 35 PM


danking:
I defer to Daniel’s thinking on the presentation of hailtop.fs. Otherwise looks good to me!

daniel-goldstein (changes requested):
This looks great, but I just have one question. Is this going to break existing links to our documentation? That seems inevitable to avoid in the general case if we do any larger restructure but if we can avoid that in this minor restructure that would be nice.

jigold:
Good point. I think I can just swap the file names of `query_api.html` and `api.html` and then make the current `api.html` into `root_api.html`?

jigold:
This is failing on a local backend test. I'm not going to retry it until you're able to review this.

jigold:
bump

daniel-goldstein (approved):

PR #13118, opened 2023-05-25
[deploy] assert sufficient space is available at PyPI in deploy

danking:


danking:
I've also asked for more space: https://github.com/pypi/support/issues/2857

daniel-goldstein (approved):

PR #13122, opened 2023-05-26
[local] Parameterize heap size of local backend

tpoterba:


hail/python/hail/context.py
```diff
@@ -552,10 +554,12 @@ def init_local(
     tmpdir = _get_tmpdir(tmpdir)
     optimizer_iterations = get_env_or_default(_optimizer_iterations, 'HAIL_OPTIMIZER_ITERATIONS', 3)
 
+    jvm_heap_size = get_env_or_default(jvm_heap_size, 'HAIL_LOCAL_BACKEND_HEAP_SIZE', None)
```
danking, line 557: Can you add an `export HAIL_LOCAL_BACKEND_HEAP_SIZE=3750M` to build.yaml for local tests?

danking (changes requested):

danking:
I pushed an env var change.

danking (approved):

danking:
I'm gonna push a change that puts a hard 2 minute limit on all tests, we'll see which ones timeout, then I'll mark the ones that are legitimately slow with a per-test timeout. Hopefully this will isolate us down to both the test and particular portion of code that's getting stuck.

danking:
I'm gonna pull these changes into a new PR.

danking:
https://github.com/hail-is/hail/pull/13128

PR #13124, opened 2023-05-30
[hailctl] Fix wrong temp-bucket setting in hailctl dataproc start

daniel-goldstein:
I'm surprised that no one has complained about this. Is this setting not used much? Should we alert people that they might have garbage in places they wouldn't expect?


tpoterba:
I think the default is some dataproc temp that gets cleaned up anyway. This isn't hail temp, I think, just the dataproc staging temp.

PR #13125, opened 2023-05-31
[infra] Add resource group to azure terraform state remote config

daniel-goldstein:
Unclear to me exactly why (perhaps a terraform update on my machine) but running terraform against azure was complaining that it didn't know the resource group that the terraform state storage account lives in. Unclear to me why this is necessary because the storage account is globally unique but it was easy enough to add and fixed my problem *shrug*. Let me know if you'd like me to look into this further.


danking (approved):
Plenty of other things I'd rather spend our time nailing down.

PR #13129, opened 2023-05-31
CHANGELOG Allow subnet to be passed through to gcloud in hailctl

tlangs:
`hailctl` allows the user to define a `network` for a new dataproc cluster, but not a `subnet`. The `network` flag in `gcloud` only works for `auto` and `legacy` project networks. If a Google Project has a custom network, the `subnet` must be supplied manually.

This PR adds that functionality to `hailctl`. Its a very simple passthrough, so not much effort was required.


daniel-goldstein (approved):
Thanks for the contribution @tlangs!

tlangs:
I don't believe I have access to look at the test failures. If you let me know what failed, I'll do my best to fix it!

daniel-goldstein:
@tlangs Nothing on your end, we've had some CI instability lately that's causing some unrelated failures. I'll make sure to retest this once that's fixed.

PR #13134, opened 2023-06-01
[query] Support VCF format float "nan"

tpoterba:


chrisvittal (changes requested):
I've been thinking about how to support VCF formatted floats properly and in a performant manner as well.

chrisvittal:
> I've been thinking about how to support VCF formatted floats properly and in a performant manner as well.

Sorry for the fast review, I just don't want this merged without sharing my thoughts.

hail/src/main/scala/is/hail/io/vcf/LoadVCF.scala
```diff
@@ -777,6 +777,16 @@ final class VCFLine(
...
+  def infoToFloat(s: String): Float = {
+    s match {
+      case "nan" => Float.NaN
+      case "-nan" => Float.NaN
+      case "inf" => Float.PositiveInfinity
+      case "-inf" => Float.NegativeInfinity
+      case _ => s.toFloat
+    }
+  }
+
```
chrisvittal, line 789: I don't think this is the best solution. The fast path here is that `toFloat` succeeds. And furthermore, this doesn't handle the full inf/nan regex `[-+]?(INF|INFINITY|NAN)`.

chrisvittal (changes requested):
So, in the vcf spec [link](https://samtools.github.io/hts-specs/VCFv4.4.pdf#subsection.1.3) a float is defined as:

> Float (32-bit IEEE-754, formatted to match one of the regular expressions `^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$` or `^[-+]?(INF|INFINITY|NAN)$` case insensitively)

There is a specific note that Java's `Double.valueOf` does not parse this correctly. Neither does this change.

hail/src/main/scala/is/hail/io/vcf/LoadVCF.scala
chrisvittal: Of course, what do I see but 4 year ago me, implementing exactly this to keep up with observed behavior of the old line parser.

chrisvittal: We remain dependent on htsjdk for several features like tabix, we should probably just use their double parser, it's public. We can use it for both floats and doubles since we should be able to convert to a float after parsing a double.

https://github.com/samtools/htsjdk/blob/6d3fc7bc1f613ecfce1c22d368f3ae17cb86823d/src/main/java/htsjdk/variant/vcf/VCFUtils.java#L260-L280

chrisvittal (approved):

PR #13137, opened 2023-06-01
[ci] Only alert on failed deploys once

jigold:
I did not test these changes at all. Let me know if you have suggestions for doing so.


daniel-goldstein (changes requested):
Should be testable with a dev deploy if you force merge into a watched branch and intentionally break `build.yaml`. But eh probably not worth the effort unless you want to try that out. If this typechecks I'm ok with it, but the new SQL needs to be added to `build.yaml`

PR #13144, opened 2023-06-05
[query] better error message in union_cols

danking:
Soyeon hit an ugly message from Scala.

### DIFF
--- a/hail/python/hail/matrixtable.py
+++ b/hail/python/hail/matrixtable.py
@@ -3960,6 +3960,19 @@ class MatrixTable(ExprContainer):
 
         if drop_right_row_fields:
             other = other.select_rows()
+        else:
+            shared_field_names = set(self.row_value) & set(other.row_value)
+            if shared_field_names:
+                field_infos = [
+                    f'  {f!r}: left type is {self.row_value.dtype[f]}, right type is {other.row_value.dtype[f]}.'
+                    for f in shared_field_names
+                ]
+                raise ValueError(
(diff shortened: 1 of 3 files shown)


patrick-schultz (changes requested):
I think it would be nicer and more consistent to copy the name mangling behavior from `Table.join`.

hail/python/hail/matrixtable.py
```diff
@@ -3960,6 +3960,17 @@ def union_cols(self, other: 'MatrixTable', row_join_type='inner', drop_right_row
...
+        else:
+            left_fields = set(self._fields)
+            other_fields = set(other._fields) - set(other.key)
+            renames, _ = deduplicate(
+                other_fields, max_attempts=100, already_used=left_fields)
+
+            if renames:
+                renames = dict(renames)
+                other = other.rename(renames)
+                info('Table.join: renamed the following fields on the right to avoid name conflicts:'
```
patrick-schultz, line 3972: fix info message

hail/python/test/hail/matrixtable/test_matrix_table.py
```diff
@@ -615,6 +615,17 @@ def test_union_cols_distinct(self):
         mt = mt.key_rows_by(x = mt.row_idx // 2)
         assert mt.union_cols(mt).count_rows() == 5
 
+    def test_union_cols_good_error(self):
```
patrick-schultz, line 618: need to update test

patrick-schultz (dismissed):

PR #13150, opened 2023-06-07
[batch] set min pool size to 1

danking:


jigold (changes requested):
Just to confirm, the intention of this PR is to make sure every pool has 1 core? I tried to see what a test PR has in the database currently, but there's no admin pod apparently in test namespaces to be able to see the database. I thought the timeout for test standing workers was 5 minutes which might defeat the purpose here.

daniel-goldstein:
FWIW, when I added the MySQL pods I made sure to install the client config in them so you don't need the admin-pod for test namespaces, just `kssh db <NAMESPACE>` and `mysql` should work

jigold:
Here's what's currently in the database:

```
mysql> select name, min_instances, standing_worker_max_idle_time_secs from pools;
+-------------+---------------+------------------------------------+
| name        | min_instances | standing_worker_max_idle_time_secs |
+-------------+---------------+------------------------------------+
| highcpu     |             0 |                                300 |
| highcpu-np  |             0 |                                300 |
| highmem     |             0 |                                300 |
| highmem-np  |             0 |                                300 |
| standard    |             1 |                                300 |
| standard-np |             1 |                                300 |
+-------------+---------------+------------------------------------+
6 rows in set (0.00 sec)
```

I think we also need to increase the timeouts here as well as make all min_instances = 1.

danking:
> the intention of this PR is to make sure every pool has 1 core?

Yes.

I will increase the timeout to 30 minutes to prevent an inopportune restart.

batch/sql/set-test-min-pool-size-to-1.py
```diff
@@ -10,7 +10,7 @@ async def main():
     await db.async_init()
     await db.execute_update(
         '''
-UPDATE pools SET min_instances = 1;
+UPDATE pools SET min_instances = 1 standing_worker_max_idle_time_secs = 1800;
```
jigold, line 13: I think you need a comma
```suggestion
UPDATE pools SET min_instances = 1, standing_worker_max_idle_time_secs = 1800;
```

jigold (changes requested):

jigold (approved):

PR #13163, opened 2023-06-09
[fs] rmtree empty dir, rmtree empty subdir, rmtree link, raise contents exc

danking:
This will at least give us more information about which directories are taking forever to delete.


jigold:
Just linting failures.

jigold (approved):

PR #13165, opened 2023-06-09
[ci] Grant create temporary tables privilege to test dbs

jigold:
I need this functionality for compacting the billing tables. I'm not sure how we get this privilege to `batch`, `batch-admin` etc., but this should be fine for me to at least measure any performance gains.

### DIFF
--- a/ci/create_database.py
+++ b/ci/create_database.py
@@ -81,20 +81,20 @@ async def create_database():
             return
 
     async def create_user_if_doesnt_exist(admin_or_user, mysql_username, mysql_password):
-        existing_user = await db.execute_and_fetchone('SELECT 1 FROM mysql.user WHERE user=%s', (mysql_username,))
-        if existing_user is not None:
-            await db.just_execute(
-f'''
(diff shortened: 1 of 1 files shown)


ci/create_database.py
```diff
@@ -89,7 +89,7 @@ async def create_user_if_doesnt_exist(admin_or_user, mysql_username, mysql_passw
             allowed_operations = 'ALL'
         else:
             assert admin_or_user == 'user'
-            allowed_operations = 'SELECT, INSERT, UPDATE, DELETE, EXECUTE'
+            allowed_operations = 'SELECT, INSERT, UPDATE, DELETE, EXECUTE, CREATE TEMPORARY TABLES'
```
daniel-goldstein, line 92: I think if would be fine to also execute the `GRANT` statement a few lines up in the `if existing_user is not None` block so that we can give already-created users this permission.

daniel-goldstein (changes requested):

jigold:
I double checked the grant operation is idempotent.

daniel-goldstein (changes requested):

daniel-goldstein (approved):

PR #13166, opened 2023-06-09
[query] add eigh method

patrick-schultz:
Add an `eigh` ndarray method, for finding eigenvalues of symmetric matrices ("h" is for Hermitian, the complex analogue of symmetric). This just dispatches to an [LAPACK method](https://netlib.org/lapack/explore-html/d2/d8a/group__double_s_yeigen_gaeed8a131adf56eaa2a9e5b1e0cce5718.html). The interface is compatible with [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.eigh.html#scipy.linalg.eigh).

### DIFF
--- a/hail/python/hail/nd/nd.py
+++ b/hail/python/hail/nd/nd.py
@@ -10,7 +10,7 @@ from hail.expr.expressions import (
     expr_numeric, Int64Expression, cast_expr, construct_expr, expr_bool,
     unify_all)
 from hail.expr.expressions.typed_expressions import NDArrayNumericExpression
-from hail.ir import NDArrayQR, NDArrayInv, NDArrayConcat, NDArraySVD, Apply
+from hail.ir import NDArrayQR, NDArrayInv, NDArrayConcat, NDArraySVD, NDArrayEigh, Apply
 
 
 tsequenceof_nd = oneof(sequenceof(expr_ndarray()), expr_array(expr_ndarray()))
@@ -423,6 +423,31 @@ def svd(nd, full_matrices=True, compute_uv=True):
     ir = NDArraySVD(float_nd._ir, full_matrices, compute_uv)
 
     return_type = ttuple(tndarray(tfloat64, 2), tndarray(tfloat64, 1), tndarray(tfloat64, 2)) if compute_uv else tndarray(tfloat64, 1)
+    return construct_expr(ir, return_type, nd._indices, nd._aggregations)
+
+
+@typecheck(nd=expr_ndarray(), eigvals_only=bool)
+def eigh(nd, eigvals_only=False):
+    """Performs an eigenvalue decomposition of a symmetric matrix.
+
+    Parameters
+    ----------
+    nd : :class:`.NDArrayNumericExpression`
+        A 2 dimensional ndarray, shape(N, N).
+    eigvals_only: :class:`.bool`
+        If False (default), compute the eigenvectors and eigenvalues. Otherwise, only compute eigenvalues.
+
+    Returns
+    -------
+    - w: :class:`.NDArrayNumericExpression`
+        The eigenvalues, shape(N).
+    - v: :class:`.NDArrayNumericExpression`
+        The eigenvectors, shape(N, N). Only returned if eigvals_only is false.
+    """
+    float_nd = nd.map(lambda x: hl.float64(x))
+    ir = NDArrayEigh(float_nd._ir, eigvals_only)
+
+    return_type = tndarray(tfloat64, 1) if eigvals_only else ttuple(tndarray(tfloat64, 1), tndarray(tfloat64, 2))
     return construct_expr(ir, return_type, nd._indices, nd._aggregations)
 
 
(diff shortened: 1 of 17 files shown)


hail/python/hail/nd/nd.py
```diff
@@ -426,6 +426,31 @@ def svd(nd, full_matrices=True, compute_uv=True):
...
+        A 2 dimensional ndarray, shape(N, N).
+    eigvals_only: :class:`.bool`
+        If False (default), compute the eigenvectors and eigenvalues. Otherwise, only compute eigenvalues.
+
+    Returns
+    -------
+    - w: :class:`.NDArrayNumericExpression`
+        The eigenvalues, shape(N).
+    - v: :class:`.NDArrayNumericExpression`
+        The eigenvectors, shape(N, N). Only returned if eigvals_only is false.
```
ehigham, line 445: I've had to think about this a little bit. It would seem to me that from this description of the return type, one can do this:
```python3
wexpr, vexpr = hl.nd.eigh(my_array)
v = hl.eval(vexpr)
```
but really, you need to do
```
v, w = hl.eval(hl.nd.eigh(my_array))
```

I'm not exactly sure how we go about saying "returns an expression that, when evaluated, yields the eigenvalues and optionally, the eigenvectors if eigenvals_only is truthy" (which is rather clunky), but I think we should say something along those lines to be absolutely clear.

hail/src/main/scala/is/hail/asm4s/Code.scala
```diff
@@ -342,6 +342,21 @@ object Code {
         a9ct.runtimeClass, a10ct.runtimeClass, a11ct.runtimeClass, a12ct.runtimeClass, a13ct.runtimeClass, a14ct.runtimeClass, a15ct.runtimeClass, a16ct.runtimeClass),
       Array(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16))
 
+  def invokeScalaObject19[A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12, A13, A14, A15, A16, A17, A18, A19, S](
```
ehigham, line 345: delightful

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
```diff
@@ -626,6 +627,63 @@ object SNDArray {
...
+
+  def syevr(cb: EmitCodeBuilder, uplo: String, A: SNDArrayValue, W: SNDArrayValue, Z: Option[(SNDArrayValue, SNDArrayValue)], Work: SNDArrayValue, IWork: SNDArrayValue): Unit = {
+    assertMatrix(A)
+    assertColMajor(cb, "orgqr", A)
+    assertVector(W, Work, IWork)
+    assert(IWork.pt.elementType.virtualType == TInt32)
+
+    val n = A.shapes(0)
+    A.assertHasShape(cb, Array(n, n), "syevr: A must be square")
+    W.assertHasShape(cb, Array(n), "syevr: W has wrong size")
```
ehigham, line 657: Is there a way to include what the shapes it got were? ie "expected square matrix, got 12-by-11".

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
```diff
@@ -626,6 +627,63 @@ object SNDArray {
       work.firstDataAddress, lwork.toI))
     cb.ifx(info.cne(0), cb._fatal(s"LAPACK error DGEQRF. Error code = ", info.toS))
   }
+
+  def syevr_query(cb: EmitCodeBuilder, jobz: String, uplo: String, n: Value[Int], region: Value[Region]): (SizeValue, SizeValue) = {
+    val WorkAddress = cb.memoize(region.allocate(8L, 8L))
```
ehigham, line 632: for my information, what's with all the memoization?

ehigham (dismissed):
Nice and clean - was pretty easy to understand without diving into the details of dsyevr. I'll assume that's wired up correctly.
Slight sticky point with docstring, otherwise comments for my reference/preference

hail/python/hail/nd/nd.py
patrick-schultz: The first thing should definitely work; I had a requiredness bug. Nice catch!

hail/src/main/scala/is/hail/asm4s/Code.scala
patrick-schultz: I know, right? Gotta love LAPACK.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
patrick-schultz: That's a fundamental part of our codegen abstraction. Would be happy to discuss in person. Short version is: `region.alllocate` returns a `Code`, which is a wrapper around some bytecode, and `cb.memoize` adds the bytecode to the `CodeBuilder` and returns a `Value`, which is just a reference to data on the stack.

hail/python/hail/nd/nd.py
ehigham: Oh interesting - that's really quite cool

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
ehigham: Thanks!

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SNDArray.scala
patrick-schultz: I fixed `assertHasShape` to print the shapes.

ehigham (approved):
Looks great, apologies for the delayed review.

PR #13169, opened 2023-06-09
[query] split yet another test

danking:


patrick-schultz (approved):
As a general comment, we don't have to check in a new resource every time we want to use something in multiple tests. Pytest's [fixtures](https://docs.pytest.org/en/latest/how-to/fixtures.html) are perfect for this.

For example:
```python
@pytest.fixture(scope="module")
def range_table_2000():
    t = hl.utils.range_table(2000, 10)
    f = new_temp_file(extension='ht')
    t.write(f)
    return f

def test_indexed_read_1(self, range_table_2000):
    t1 = hl.read_table(range_table_2000, _create_row_uids=True)
    ...
```

PR #13170, opened 2023-06-10
[transient-error] new transient error from GCS

danking:
https://ci.hail.is/batches/7525669/jobs/178

### DIFF
--- a/hail/src/main/scala/is/hail/services/package.scala
+++ b/hail/src/main/scala/is/hail/services/package.scala
@@ -84,7 +84,7 @@ package object services {
           if RETRYABLE_HTTP_STATUS_CODES.contains(e.getStatusCode()) =>
         true
       case e: HttpResponseException
-          if (e.getStatusCode() == 401 &&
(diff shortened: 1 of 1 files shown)


hail/src/main/scala/is/hail/services/package.scala
```diff
@@ -83,6 +83,28 @@ package object services {
       case e: HttpResponseException
           if RETRYABLE_HTTP_STATUS_CODES.contains(e.getStatusCode()) =>
         true
+      case e: HttpResponseException
+          if (e.getStatusCode() == 401 &&
```
jigold, line 87: Is this supposed to be 410 to match the output below?

jigold (changes requested):

hail/src/main/scala/is/hail/services/package.scala
danking: Yep, good catch!

jigold (approved):

PR #13173, opened 2023-06-12
[fs] Fix globbing in the scala blob storage FS

daniel-goldstein:
Using builtin URL parsing for object storage URIs has proven problematic as they do not always follow the same syntactic conventions. In particular, Java URI truncates the path at `?` and assigns the rest to a query string field (as it should), but `?` is a valid character in a blob storage object and is used as a glob pattern. We were forgetting to look at the `URI.query` parameter to see if the object name got truncated in parsing and therefore broke globbing. I added a few tests specifically that used single-character match pattern (`?`) and got rid of `URI` for the most dead-simple parsing approach I could think of to get ahead of any other potential footguns.


hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -66,53 +66,34 @@ class AzureStorageFSHttpsURL(
...
   def parseUrl(filename: String): AzureStorageFSURL = {
-    val uri = new URI(filename).normalize()
-
-    val scheme = uri.getScheme
-    if (scheme == null || !schemes.contains(scheme)) {
-      throw new IllegalArgumentException(s"Invalid scheme, expected hail-az or https: $scheme")
-    }
+    parseHttpsUrl(filename)
+      .orElse(parseHailAzUrl(filename))
+      .getOrElse(throw new IllegalArgumentException("ABS URI must be of the form https://<ACCOUNT>.blob.core.windows.net/<CONTAINER>/<PATH>"))
```
patrick-schultz, line 79: Error reporting has gotten a little worse. Before we gave specialized errors depending on the scheme. Maybe it's worthwhile to still match the scheme first, then match the rest with the appropriate regex?

patrick-schultz (changes requested):

patrick-schultz (approved):

PR #13175, opened 2023-06-12
[batch] Fix billing projects query to be fast with lateral joins

jigold:
This query is the slowest one we have right now and is making the billing projects pages to be slow. I saw substantial speed improvements using lateral joins.

- For querying all open billing projects for a given user, it went from 1.8 seconds to 0.87 seconds.
- For querying all open billing projects (i.e. billing projects page), it went from 5.5 seconds to 2.6 seconds.

The speed of the queries should go down further once the aggregated billing project user resources table has been deduped and further improvements after compaction.

### DIFF
--- a/batch/batch/utils.py
+++ b/batch/batch/utils.py
@@ -136,7 +136,7 @@ async def query_billing_projects(db, user=None, billing_project=None):
     sql = f'''
 SELECT billing_projects.name as billing_project,
   billing_projects.`status` as `status`,
-  users, `limit`, cost_t.accrued_cost
+  users, `limit`, COALESCE(cost_t.cost, 0) AS accrued_cost
 FROM billing_projects
 LEFT JOIN LATERAL (
   SELECT billing_project, JSON_ARRAYAGG(`user_cs`) as users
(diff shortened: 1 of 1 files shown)


batch/batch/utils.py
```diff
@@ -170,6 +167,8 @@ def record_to_dict(record):
             record['users'] = json.loads(record['users'])
         return record
 
+    log.info(f'sql {sql} args {args}')
```
daniel-goldstein, line 170: Was this for debugging?

daniel-goldstein (dismissed):
My apologies for letting this fall off my plate. This is a great improvement, just one log message that should get deleted

danking:
Neat! Where'd you learn about lateral joins? I had not heard about them until now.

daniel-goldstein (approved):

jigold:
I got frustrated with the query speeds, tried to figure out how I'd go and implement it efficiently if I were coding this from scratch, searched google for awhile and finally came across what I needed on either Stack Overflow or the MySQL docs. It's a pretty recent feature addition. v8.0.14.

PR #13176, opened 2023-06-13
[deploy] Add WHEEL to environment for assert_pypi_has_room.py

daniel-goldstein:


jigold (dismissed):
Just want to double check WHEEL_FOR_AZURE isn't relevant here.

danking (approved):

PR #13178, opened 2023-06-13
[fs] Add support for SAS tokens in Scala AzureStorageFS

daniel-goldstein:
This PR adds support for Azure SAS tokens in QoB. A SAS token is basically a blob storage URI with a short-lived credential to access that resource appended as a URL query string. In such a scenario where the FS receives a blob URI with a SAS token, that token should be used instead of the latent credentials on the system. Most of the changes to the `AzureStorageFS.scala` are to parse out a SAS token from blob names.

This change brings with it a couple caveats. Unfortunately it is not possible to truly disambiguate a SAS token from a glob pattern, or even just a normal blob filename. So we take what is probably a safe assumption and look to see if there exists a query-parameter style key-value pair after the last `?` in the blob name. If this is the case, we treat everything after the last `?` as a SAS token. If this condition is not satsified, we say there is no SAS token and treat the whole path as the blob name. This logic already exists in python, but I'm open to alternatives.

Introducing SAS tokens also breaks the way globbing is currently implemented, where it is deemed safe to iteratively append components to the end of a blob URI string. I added an abstract type member to `FS` and instead of a `String` have `globWithPrefix` accept that associated URL type that can properly handle path updates. I'm unclear on the best way to do this w.r.t. the type system, and wasn't quite sure what to put as the associated type for `RouterFS`, which ideally would accept a union of the URL types for the filesystems that it wraps, so some guidance on that would be great if you have any.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2437,11 +2437,14 @@ steps:
       export HAIL_DOCTEST_DATA_DIR="{{ global.test_storage_uri }}/{{ upload_test_resources_to_blob_storage.token }}/doctest/data/"
       export HAIL_GENETICS_VEP_GRCH37_85_IMAGE={{ hailgenetics_vep_grch37_85_image.image }}
       export GOOGLE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
+      export AZURE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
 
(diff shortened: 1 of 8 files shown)


patrick-schultz (approved):
I don't immediately see any better way to structure this, I think it's fine as is.

PR #13180, opened 2023-06-14
[auth] More easily add developers to developer namespaces

daniel-goldstein:
I decided to break off this chunk from another PR that has stalled. That PR will ultimately build on this to add all developers automatically to dev AND test namespaces, but this should be an improvement for now. A few things in here:

- Deleted all the `DatabaseResource` stuff in the auth driver. Since databases now are created and destroyed with the namespace and not the developer, this is basically dead code.
- Added the ability to add a user for an existing hail identity. This is only permitted in dev namespaces and serves as a way for developers to use the same hail identity across namespaces. There is one caveat here: `create_initial_account.py` tries to copy the `<dev-name>-gsa-key` secret from default into the developer namespace and this code will *not* do that anymore. For the developer to submit jobs to the namespace, they must first manually copy in the secret from `default` if it does not already exist inside the namespace. This is awkward, but IMO acceptable because:
    - the copying code in `create_initial_account.py` is already broken anyway because when that script is run in a dev deploy it does not have access to production secrets
    - I hope that when we eventually go keyless we can delete the gsa key secrets and this whole problem goes away.
    - I feel like it's not too bad to do this manual one time copy as opposed to maintaining code that is privileged enough to reach across namespaces. Seems error prone and like a security headache.
- Deletes `create_initial_account.py` in favor of using our actual API to create the dev user.


hail/python/hailtop/hailctl/dev/cli.py
```diff
@@ -53,7 +53,6 @@ async def _deploy(branch: str, steps: List[str], excluded_steps: List[str], extr
 
     deploy_config = get_deploy_config()
     steps = unpack_comma_delimited_inputs(steps)
-    print(steps)
```
daniel-goldstein, line 56: This accidentally snuck in when I was rewriting `hailctl`.

jigold (dismissed):
This looks great! Thank you for the comments inlined into the code. 

> For the developer to submit jobs to the namespace, they must first manually copy in the secret from default if it does not already exist inside the namespace.

Can we dummy proof this a bit more to have a nice error message if this is not the case?

auth/auth/driver/driver.py
```diff
@@ -481,21 +398,14 @@ async def _create_user(app, user, skip_trial_bp, cleanup):
         updates['hail_credentials_secret_name'] = hail_credentials_secret_name
 
     namespace_name = user['namespace_name']
-    if namespace_name is None and user['is_developer'] == 1:
+    # auth services in test namespaces cannot/should not be creating and deleting namespaces
```
jigold, line 401: I think I would want a check here if it is a dev deployment that the expected secret given by `hail_credentials_secret_name` exists in that namespace using the k8s client.

auth/auth/auth.py
```diff
@@ -367,8 +389,30 @@ async def create_user(request: web.Request, userdata):  # pylint: disable=unused
...
 
+    hail_identity = body.get('hail_identity')
+    hail_credentials_secret_name = body.get('hail_credentials_secret_name')
+    if (hail_identity or hail_credentials_secret_name) and not is_test_deployment:
+        raise web.HTTPBadRequest(text='Cannot specify an existing hail identity for a new user')
+    if hail_credentials_secret_name:
+        try:
+            k8s_cache: K8sCache = request.app['k8s_cache']
+            res = await k8s_cache.read_secret(hail_credentials_secret_name, DEFAULT_NAMESPACE)
+            log.info(res)
```
jigold, line 400: delete?

auth/auth/auth.py
```diff
@@ -367,8 +389,30 @@ async def create_user(request: web.Request, userdata):  # pylint: disable=unused
...
+        except kubernetes_asyncio.client.rest.ApiException as e:
+            raise web.HTTPBadRequest(
+                text=f'hail credentials secret name specified but was not found: {hail_credentials_secret_name}'
```
jigold, line 403: Can we add the namespace we're looking at as well in the error message?

jigold (changes requested):

PR #13181, opened 2023-06-15
[batch] Delay waiting on batch completion to avoid unnecessary requests

jigold:
https://hail.zulipchat.com/#narrow/stream/123011-Hail-Query-Dev/topic/simple.20job.20timings.20from.20batch.20perspective/near/366227291


daniel-goldstein (approved):
This seems worthwhile for now to reduce load during big tests, but long term I think we should really look into websockets/non-polling solutions instead.

PR #13184, opened 2023-06-15
[query/vds] Don't densify past the end of contigs

### LINKED ISSUE #13183: `hl.vds.to_dense_mt` incorrectly densifies the last reference block of a chromosome into the beginning of the next
### What happened?

The offending code is here:

https://github.com/hail-is/hail/blob/1883a71f741dda67d4d4d140c40f93532c7ab19f/hail/python/hail/vds/methods.py#L74

Checking END is insufficient unless we also check that the chromosome index is the same. However, I believe that the correct fix here is to not densify on END at all, but rather preprocess the END into the global position end:

```
mt = mt.annotate_rows(_locus_global_pos = mt.locus.global_position(), _locus_pos = mt.locus.position)
mt = mt.transmute_entries(END_GLOBAL = mt._locus_global_pos + (mt.END - mt._locus_pos))
```
then use that.

Alternatively, stop using END at all and use LEN, which will compress better anyway.

### Version

0.2.118

### Relevant log output

_No response_

chrisvittal:
fixes #13183

### DIFF
--- a/hail/python/hail/vds/methods.py
+++ b/hail/python/hail/vds/methods.py
@@ -41,7 +41,10 @@ def to_dense_mt(vds: 'VariantDataset') -> 'MatrixTable':
     ref = vds.reference_data
     # FIXME(chrisvittal) consider changing END semantics on VDS to make this better
     # see https://github.com/hail-is/hail/issues/13183 for why this is here and more discussion
-    ref = ref.transmute_entries(END_GLOBAL=hl.locus(ref.locus.contig, ref.END).global_position())
+    ref = ref.transmute_entries(END_GLOBAL=hl.locus(contig=ref.locus.contig,
+                                                    pos=ref.END,
+                                                    reference_genome=ref.locus.dtype.reference_genome
+                                                    ).global_position())
 
     ref = ref.drop(*(x for x in ('alleles', 'rsid', 'ref_allele') if x in ref.row))
     var = vds.variant_data
@@ -76,7 +79,7 @@ def to_dense_mt(vds: 'VariantDataset') -> 'MatrixTable':
                         lambda refs_at_this_row: hl.zip_with_index(hl.zip(dr._var_entries, dr.dense_ref)).map(
                             lambda tup: coalesce_join(hl.coalesce(refs_at_this_row[tup[0]],
                                                                   hl.or_missing(tup[1][1].END_GLOBAL >= dr.locus.global_position(),
-                                                                                tup[1][1])), tuple[1][0])
+                                                                                tup[1][1])), tup[1][0])
                         )),
     )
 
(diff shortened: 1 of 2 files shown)


tpoterba (dismissed):
can you add a test?

hail/python/hail/vds/methods.py
```diff
@@ -39,6 +39,13 @@ def to_dense_mt(vds: 'VariantDataset') -> 'MatrixTable':
         Dataset in dense MatrixTable representation.
     """
     ref = vds.reference_data
+    # FIXME(chrisvittal) consider changing END semantics on VDS to make this better
+    # see https://github.com/hail-is/hail/issues/13183 for why this is here and more discussion
+    ref = ref.transmute_entries(END_GLOBAL=hl.locus(contig=ref.locus.contig,
```
tpoterba, line 44: The function this calls is not super cheap, better to separate this out into the pieces I suggested on that issue

tpoterba (dismissed):

hail/python/hail/vds/methods.py
chrisvittal: Fine. I disagree though. This is simple, correct, and maintainable. It is still possible to densify past the end of a contig with your solution, `END` could be larger than the contig.

tpoterba: if END is greater than the max position of a contig? That's bad data, I'm OK having wrong results for densify there. If you want to catch that, can do so without calling the locus constructor per entry anyhow I think

patrick-schultz (changes requested):
lgtm, just one issue with field names

patrick-schultz (approved):

hail/python/hail/vds/methods.py
```diff
@@ -39,7 +39,14 @@ def to_dense_mt(vds: 'VariantDataset') -> 'MatrixTable':
         Dataset in dense MatrixTable representation.
     """
     ref = vds.reference_data
-    ref = ref.drop(*(x for x in ('alleles', 'rsid', 'ref_allele') if x in ref.row))
+    # FIXME(chrisvittal) consider changing END semantics on VDS to make this better
+    # see https://github.com/hail-is/hail/issues/13183 for why this is here and more discussion
+    # we assume that END <= contig.length
+    ref = ref.annotate_rows(_locus_global_pos=ref.locus.global_position(), _locus_pos=ref.locus.position)
+    ref = ref.transmute_entries(_END_GLOBAL=ref.locus_global_pos + (ref.END - ref.locus_pos))
```
danking, line 46: `_`

chrisvittal, line 46:
```suggestion
    ref = ref.transmute_entries(_END_GLOBAL=ref._locus_global_pos + (ref.END - ref._locus_pos))
```

chrisvittal: 🤦

PR #13187, opened 2023-06-16
[tls] make internal cert config optional

daniel-goldstein:
Pretty sure this is all dead code because we don't currently speak TLS between the worker and internal gateway. I think what's ultimately the right approach here is to start sending the hail root cert to workers and have them mount that at a well known location to all containers, not have it be part of the job spec. Then we should be able to talk https to internal gateway.

I'm not exactly sure what to do in this PR specifically. I want to get rid of this hard dependency on the ssl-config because
    1. Doesn't exist in terra as everything goes through the relay listener
    2. These ssl configs aren't used anyway
but I'd need to put in a bit more work to start sending the root cert to workers and add https to internal-gateway

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1821,10 +1821,6 @@ steps:
         namespace:
           valueFrom: default_ns.name
         mountPath: /user-tokens
-      - name: ssl-config-monitoring-tests
-        namespace:
-          valueFrom: default_ns.name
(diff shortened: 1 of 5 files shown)


danking (approved):
Seems fine to me. Fixing this up once internal-gateway speaks SSL seems fine too.

PR #13192, opened 2023-06-17
[batch] Add new ResourceGroup items to `_resources_inverse`

### LINKED ISSUE #13191: ResourceGroup issues modelling writing a bgzipped file and its TBI index via `write_output`
### What happened?

I have a job that runs a GATK command that produces a `foo.counts.tsv` file, compresses and indexes it, and should store the resulting `foo.counts.tsv.gz` and `foo.counts.tsv.gz.tbi` files. The indexing command requires the filename extension to be `.counts.tsv.gz` and there is clearly a strong relationship between the filenames of the three files, so I have been using a ResourceGroup to model this, essentially as follows:

```python
j = b.new_job(…)

j.declare_resource_group(counts={
    'tsv': '{root}.counts.tsv',
    'tsv.gz': '{root}.counts.tsv.gz',
    'tsv.gz.tbi': '{root}.counts.tsv.gz.tbi',
})

j.command(f"""
    gatk SubCommand … --output {j.counts['tsv']}
    bgzip {j.counts['tsv']}
    gatk IndexFeatureFile --input {j.counts['tsv.gz']}
    """)

b.write_output(j.counts['tsv.gz'], output_path)
b.write_output(j.counts['tsv.gz.tbi'], output_index_path)
```

(Trying to use `b.write_output(j.counts, output_base)` to write out the whole ResourceGroup fails because the `…/counts.counts.tsv` file no longer exists because it was removed by `bgzip`. In any case, we don't want to write that one to the final bucket anyway. Hence the two separate `write_output` invocations for the two desired output files.)

This fails in the second `write_output` with a fairly mysterious exception:

```
  File "…", line 92, in test
    b.write_output(j.counts['tsv.gz.tbi'], output_index_path)
  File "…/lib/python/site-packages/hailtop/batch/batch.py", line 595, in write_output
    name = resource._source._resources_inverse[resource]
KeyError: __RESOURCE_FILE__11
```

Checking that line of _batch.py_, it is failing while trying to print an error message because `_resources_inverse` is not set up for the JobResourceFiles within ResourceGroups. PR NNN is a suggested fix for this.

With that PR applied, this results in a more useful hail error message exception:

```
hailtop.batch.exceptions.BatchException: undefined resource 'counts["tsv.gz.tbi"]'
Hint: resources must be defined within the job methods 'command' or 'declare_resource_group'
```

This can be worked around by mentioning the filename in the commands to be run — in a comment, because none of the commands actually need to specify the `.tbi` output filename:

```python
…
j.command(f"""
    gatk SubCommand … --output {j.counts['tsv']}
    bgzip {j.counts['tsv']}
    gatk IndexFeatureFile --input {j.counts['tsv.gz']}
    : {j.counts['tsv.gz.tbi']}
    """)
…
```

This produces the desired two files — compressed data and the associated index — written to the final bucket. Is it kosher to use `write_output` on the individual items within a ResourceGroup like this?

However this resource **was** defined by a `declare_resource_group` invocation. So it seems to me that this too is a bug and it should not be necessary to work around the lack of `j.counts['tsv.gz.tbi']` being mentioned like this.

### Version

0.2.118

### Relevant log output

_No response_

jmarshall:
This is used to print the resource name in several error messages, which failed when the error occurred for an item within a ResourceGroup.

Proposed fix addressing part of #13191, for your consideration.

Fixes #13191

### DIFF
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -711,6 +711,7 @@ class BashJob(Job):
             self._resources[name] = rg
             for rname, r in rg._resources.items():
                 self._resources_inverse[r] = f'{name}["{rname}"]'
+                self._mentioned.add(r)
             _add_resource_to_set(self._valid, rg)
         return self
 
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -709,6 +709,8 @@ class BashJob(Job):
                 raise BatchException(f"value for name '{name}' is not a dict. Found '{type(d)}' instead.")
             rg = self._batch._new_resource_group(self, d, root=name)
             self._resources[name] = rg
+            for rname, r in rg._resources.items():
+                self._resources_inverse[r] = f'{name}["{rname}"]'
             _add_resource_to_set(self._valid, rg)
         return self
 


jmarshall:
I've added a second commit that fixes the remainder of #13191, marking the individual JobResourceFiles within the ResourceGroup as `_mentioned` and hence preventing the “undefined resource” BatchException previously observed. (Some of the tests in _hail/python/test/hailtop/batch/test_batch.py_ would also need adjusting to account for the re-imagined `_mentioned`.)

Having now studied f6fe19c085a9d9ebee23866961cb582a713cc1ad, which introduced `_mentioned` and this error message hint, IMHO this is a reasonable fix. Apart from the code in _backend.py_ to do with `symlink_input_resource_group`, which I haven't looked at, `_mentioned` is maintained solely to decide whether to emit this BatchException hinting to the user that the resource ought to be defined if you're going to use it in `write_output`. In this case, because the filenames are related, `foo.gz.tbi` may well have been created even though only `foo.gz` appears explicitly in the command text, so it may be a false positive (as in #13191's case) to raise the exception. So the conservative thing to do is to suppress the message in these `declare_resource_group` cases.

jigold:
I apologize for not responding sooner to this. I've been mulling over what to do here as it's been over 4 years since I wrote the first interface. I think your changes are fine, but I need to go through the tests again and figure out what `_mentioned` was originally intended for to make sure this change doesn't break anything subtle. I'm going to have our CI run this SHA so I can see what the failures are.

jigold:
Closing in favor of #14352

PR #13199, opened 2023-06-21
[batch] Add new list_batches query interface to REST calls

jigold:
This is stacked on #12748

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -88,7 +88,13 @@ from ..utils import (
     regions_to_bits_rep,
     unavailable_if_frozen,
 )
-from .query import CURRENT_QUERY_VERSION, build_batch_jobs_query, build_list_batches_query
+from .query import (
+    CURRENT_QUERY_VERSION,
+    parse_list_batches_query_v1,
+    parse_list_batches_query_v2,
+    parse_batch_jobs_query_v1,
+    parse_batch_jobs_query_v2,
+)
 from .validate import ValidationError, validate_and_clean_jobs, validate_batch, validate_batch_update
 
 uvloop.install()
@@ -234,7 +240,11 @@ async def _handle_api_error(f, *args, **kwargs):
 
 async def _query_batch_jobs(request, batch_id: int, version: int, q: str, last_job_id: Optional[int]):
     db: Database = request.app['db']
-    sql, sql_args = build_batch_jobs_query(batch_id, version, q, last_job_id)
+    if version == 1:
+        sql, sql_args = parse_batch_jobs_query_v1(batch_id, q, last_job_id)
+    else:
+        assert version == 2, version
+        sql, sql_args = parse_batch_jobs_query_v2(batch_id, q, last_job_id)
 
     jobs = [job_record_to_dict(record, record['name']) async for record in db.select_and_fetchall(sql, sql_args)]
 
@@ -602,7 +612,11 @@ async def rest_get_job_container_log(request, userdata, batch_id):  # pylint: di
 
 async def _query_batches(request, user: str, q: str, version: int, last_batch_id: Optional[int]):
     db: Database = request.app['db']
-    sql, sql_args = build_list_batches_query(user, version, q, last_batch_id)
+    if version == 1:
+        sql, sql_args = parse_list_batches_query_v1(user, q, last_batch_id)
+    else:
+        assert version == 2, version
+        sql, sql_args = parse_list_batches_query_v2(user, q, last_batch_id)
 
     batches = [batch_record_to_dict(record) async for record in db.select_and_fetchall(sql, sql_args)]
 
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -11,7 +11,7 @@ import signal
 import traceback
 from functools import wraps
 from numbers import Number
-from typing import Any, Dict, Optional, Tuple, Union
+from typing import Dict, Optional, Tuple, Union
 
 import aiohttp
 import aiohttp_session
@@ -210,22 +210,21 @@ async def rest_get_supported_regions(request, userdata):  # pylint: disable=unus
     return json_response(list(request.app['regions'].keys()))
 
 
-async def _handle_ui_error(session, f, *args, **kwargs) -> Tuple[bool, Any]:
+async def _handle_ui_error(session, f, *args, **kwargs):
     try:
-        result = await f(*args, **kwargs)
+        return await f(*args, **kwargs)
     except KeyError as e:
         set_message(session, str(e), 'error')
         log.info(f'ui error: KeyError {e}')
-        return (True, None)
+        raise
     except BatchOperationAlreadyCompletedError as e:
         set_message(session, e.message, e.ui_error_type)
         log.info(f'ui error: BatchOperationAlreadyCompletedError {e.message}')
-        return (True, None)
+        raise
     except BatchUserError as e:
         set_message(session, e.message, e.ui_error_type)
         log.info(f'ui error: BatchUserError {e.message}')
-        return (True, None)
-    return (False, result)
+        raise
 
 
 async def _handle_api_error(f, *args, **kwargs):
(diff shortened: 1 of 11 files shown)


batch/batch/front_end/query/query_v2.py
```diff
@@ -9,20 +9,157 @@
...
+  SELECT COALESCE(SUM(`usage` * rate), 0) AS cost
+  FROM (
+    SELECT batch_id, resource_id, CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
+    FROM aggregated_batch_resources_v2
+    WHERE batches.id = aggregated_batch_resources_v2.batch_id
+    GROUP BY batch_id, resource_id
+  ) AS usage_t
+  LEFT JOIN resources ON usage_t.resource_id = resources.resource_id
+  GROUP BY batch_id
+) AS cost_t ON TRUE
```
daniel-goldstein, line 152: I don't really know much about lateral joins, could you explain what this join `ON TRUE` means or point me to some docs?

batch/batch/front_end/query/query_v2.py
```diff
@@ -9,20 +9,157 @@
...
+WHERE {' AND '.join(where_conditions)}
+ORDER BY id DESC
+LIMIT 51;
+'''
+
+    sql_args = where_args
```
daniel-goldstein, line 158: nit: this line doesn't really do anything, I think we should just `return (sql, where_args)`

batch/batch/front_end/query/query.py
```diff
@@ -280,14 +281,14 @@ def query(self) -> Tuple[str, List[Any]]:
...
-class DurationQuery(Query):
+class JobDurationQuery(Query):
     @staticmethod
-    def parse(op: str, time: str) -> 'DurationQuery':
+    def parse(op: str, time: str) -> 'JobDurationQuery':
         operator = get_operator(op)
         if not isinstance(operator, ComparisonOperator):
             raise QueryError(f'unexpected operator "{op}" expected one of {ComparisonOperator.symbols}')
-        time_msecs = int(parse_float(time) * 1000 + 1)
-        return DurationQuery(operator, time_msecs)
```
daniel-goldstein, line 290: What was this + 1 for and why is it removed?

batch/test/test_batch.py
```diff
@@ -317,6 +317,173 @@ def assert_batch_ids(expected: Set[int], q=None):
...
+''',
+    )
+
+    assert_batch_ids(
+        set(),
+        '''
+start_time >= 2023-02-24T17:15:25Z
+end_time == 2023-02-24T17:15:25Z
+''',
+    )
```
daniel-goldstein, line 479: Can we add a test for `user`?

batch/test/test_batch.py
```diff
@@ -317,6 +317,173 @@ def assert_batch_ids(expected: Set[int], q=None):
...
+
+    assert_batch_ids(set(), 'duration > 50000')
+    assert_batch_ids(set(), 'cost > 1000')
+    assert_batch_ids({b1.id}, f'batch_id = {b1.id}')
+    assert_batch_ids({b1.id}, f'batch_id == {b1.id}')
+
```
daniel-goldstein, line 485: I think these tests are good, but it's hard to make them really thorough. Maybe not in this PR but as part of this project we can write some tests that just test the parser, and run it through some exhaustive tests to make sure that every possible grammatical construct parses without error (or with error for queries that don't support certain operators). Such tests would be super cheap and give me more confidence that all of the exception handling is working properly.

batch/test/test_batch.py
```diff
@@ -317,6 +317,173 @@ def assert_batch_ids(expected: Set[int], q=None):
...
+    assert_batch_ids(
+        set(),
+        '''
+start_time >= 2023-02-24T17:15:25Z
+end_time == 2023-02-24T17:15:25Z
+''',
+    )
+
+    assert_batch_ids(set(), 'duration > 50000')
```
daniel-goldstein, line 481: I think it would be nice to have at least a couple of tests with invalid query strings to make sure we get a 400 and not crashing the front-end pod.

daniel-goldstein (changes requested):
This all looks nice and straightforward. Mainly small things and some ideas moving forward around testing.

batch/batch/front_end/query/query_v2.py
jigold: I found this post to be helpful: https://jonmce.medium.com/what-the-heck-is-a-lateral-join-anyway-4c3345b94a63
See the section on syntactic quirks.

batch/batch/front_end/query/query.py
jigold: I wasn't sure which direction the value needed to be rounded. When I looked at the generated SQL statement for this query `duration > 5000`, then I got `((time_completed - time_started) > 5000001` (converting to msecs) which is not what I think we want. Maybe we can just keep the float comparison and MySQL will do the numeric comparison appropriately?

batch/batch/front_end/query/query.py
```diff
@@ -319,3 +320,249 @@ def __init__(self, operator: ComparisonOperator, cost: float):
...
+        if not isinstance(operator, ComparisonOperator):
+            raise QueryError(f'unexpected operator "{op}" expected one of {ComparisonOperator.symbols}')
+        batch_id = parse_int(maybe_batch_id)
+        return BatchIdQuery(batch_id, operator)
+
+    def __init__(self, batch_id: int, operator: ComparisonOperator):
+        self.batch_id = batch_id
+        self.operator = operator
+
+    def query(self) -> Tuple[str, List[Any]]:
```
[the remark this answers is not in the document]
jigold: I'm not sure exactly what you mean. They could be floats, int, str, dates, etc. Anything PyMySQL allows as SQL arguments.

batch/batch/front_end/query/query_v2.py
daniel-goldstein: thanks that was really helpful!

batch/batch/front_end/query/query.py
daniel-goldstein: Hm, I think what sticks out to me a little more is that we can't really guarantee users precision beyond the millisecond. If I query for `duration > 0.0015` and the job is recorded as 1ms, should I return that value or not? I don't think we can really know because the real time of the job *could* have been 1.9ms or it could have been 1.2ms. I think it would be best just to assume that these floats map to whole milliseconds (`int(duration * 1000)`) and maybe as a later project we can write a little client-side javascript that looks at the search box and shows a warning if a user tries to get finer precision than we can promise (I imagine this will basically never happen).

batch/batch/front_end/query/query.py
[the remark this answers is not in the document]
daniel-goldstein: > They could be floats, int, str, dates, etc. Anything PyMySQL allows as SQL arguments.

This is certainly true of this function on the `Query` class, i.e. the function interface. But looking at the `BatchIdQuery` class in isolation we can make a much stronger guarantee, specifically that this function is guaranteed to return `Tuple[str, List[int]]`. Being as specific as possible is helpful for type-checking and in-editor hints when dealing with objects that are known to be a specific subclass.

batch/batch/front_end/front_end.py
```diff
@@ -203,22 +209,22 @@ async def rest_get_supported_regions(request, userdata):  # pylint: disable=unus
...
         log.info(f'ui error: BatchOperationAlreadyCompletedError {e.message}')
-        return True
+        return (True, None)
     except BatchUserError as e:
         set_message(session, e.message, e.ui_error_type)
         log.info(f'ui error: BatchUserError {e.message}')
-        return True
-    return False
+        return (True, None)
+    return (False, result)
```
daniel-goldstein, line 227: This introduces another style of error handling into the codebase that I think harms consistency. While very idiomatic [go](https://go.dev/blog/error-handling-and-go), I think it would be better to stick with the pythonic style of error-handling where this function either returns the return value of `f` or raises an exception if `f` fails. Then callers should use this in a try-except block instead of using `if errored`.

daniel-goldstein (changes requested):
Looks great, just a couple small nits.

jigold:
This should be ready for review.

daniel-goldstein (approved):
LGTM! Really excited to start using this

PR #13202, opened 2023-06-21
[ci] Add test identities for services in test namespaces

daniel-goldstein:
Identities in test namespaces cannot share the same underlying cloud identity if we want to identify requests with cloud access tokens. This also means the `test` account does not need to have the union of roles of the other robot accounts, but pruning of the `test` account's roles is left until after this PR merges so we can properly assess which roles are still in use by the `test` account.


daniel-goldstein:
I think test-test-gsa-key is pretty awful sounding but I'm not sure how better to call a test user for test namespaces and it's consistent with the test-auth, test-batch, etc. names.

danking:
I don't mind it especially given the consistency.

infra/azure/modules/batch/main.tf
```diff
@@ -215,6 +247,42 @@ resource "azurerm_role_assignment" "test_test_container_contributor" {
   principal_id         = module.test_sp.principal_id
 }
 
+resource "azurerm_role_assignment" "test_registry_viewer" {
```
danking, line 250: maybe call this `test_sp_registry_viewer`? `test_registry_viewer` seems too much like test is an adjective.

build.yaml
```diff
@@ -477,11 +482,12 @@ steps:
...
+      kubectl -n {{ default_ns.name }} get -o json secret test-test-gsa-key | jq '{apiVersion, kind, type, data, metadata: {name: "test-gsa-key"}}' | kubectl -n {{ default_ns.name }} apply -f -
+      kubectl -n {{ default_ns.name }} get -o json secret test-test-dev-gsa-key | jq '{apiVersion, kind, type, data, metadata: {name: "test-dev-gsa-key"}}' | kubectl -n {{ default_ns.name }} apply -f -
+      kubectl -n {{ default_ns.name }} get -o json secret test-auth-gsa-key | jq '{apiVersion, kind, type, data, metadata: {name: "auth-gsa-key"}}' | kubectl -n {{ default_ns.name }} apply -f -
+      kubectl -n {{ default_ns.name }} get -o json secret test-batch-gsa-key | jq '{apiVersion, kind, type, data, metadata: {name: "batch-gsa-key"}}' | kubectl -n {{ default_ns.name }} apply -f -
+      kubectl -n {{ default_ns.name }} get -o json secret test-ci-gsa-key | jq '{apiVersion, kind, type, data, metadata: {name: "ci-gsa-key"}}' | kubectl -n {{ default_ns.name }} apply -f -
+      kubectl -n {{ default_ns.name }} get -o json secret test-grafana-gsa-key | jq '{apiVersion, kind, type, data, metadata: {name: "grafana-gsa-key"}}' | kubectl -n {{ default_ns.name }} apply -f -
```
danking, line 490: we could call them `testns-test`, `testns-auth`, etc.?

infra/azure/modules/auth/main.tf
```diff
@@ -21,6 +21,27 @@ module "auth_sp" {
...
+  display_name = "${var.resource_group_name}-test-auth"
+
+  required_resource_access {
+    resource_app_id = "00000003-0000-0000-c000-000000000000"
+
+    resource_access {
+      # Application.ReadWrite.All
+      id   = "1bfefb4e-e0b5-418b-a88f-73c46d2cc8e9"
+      type = "Role"
+    }
```
danking, line 34: Is the permissions/roles grant for auth? I was expecting an `azurerm_role_assignment` that granted it, I dunno, azure app creation/deletion/sp-creation.

infra/gcp-broad/main.tf
```diff
@@ -285,10 +277,6 @@ resource "google_container_node_pool" "vdc_nonpreemptible_pool" {
       enable_integrity_monitoring = true
       enable_secure_boot          = false
     }
-
-    workload_metadata_config {
-      mode = "GKE_METADATA"
-    }
```
danking, line 291: stacked on that other PR I presume, yeah?

infra/gcp-broad/main.tf
```diff
@@ -490,6 +487,23 @@ resource "google_storage_bucket_iam_member" "batch_hail_query_bucket_storage_vie
   member = "serviceAccount:${module.batch_gsa_secret.email}"
 }
 
+module "test_batch_gsa_secret" {
+  source = "./gsa"
+  name = "test-batch"
+  project = var.gcp_project
+  iam_roles = [
+    "compute.instanceAdmin.v1",
```
danking, line 495: totally unrelated to this PR, but at some point we should scope this to the test namespaced VMs.

danking (changes requested):
We have to apply the TF before we can test this right because the secrets are missing?

infra/azure/modules/auth/main.tf
daniel-goldstein: Ya there's a different way of assigning permissions within AAD vs permissions on other resources in Azure. This was the only way I could find at the time to create/read apps and service principals.

build.yaml
daniel-goldstein: I like this better.

infra/gcp-broad/main.tf
daniel-goldstein: yup

daniel-goldstein:
> We have to apply the TF before we can test this right because the secrets are missing?

I already applied this terraform since I didn't see another practical way to go about it and it doesn't mutate any existing resources.

infra/azure/modules/auth/main.tf
```diff
@@ -21,6 +21,27 @@ module "auth_sp" {
...
+  }
+}
+module "test_auth_sp" {
```
danking, line 37:
```suggestion
module "testns_auth_sp" {
```
yeah?

danking (changes requested):
@-me when you fix and I'll approve

daniel-goldstein:
@danking fixed

danking (approved):

danking:
You're gonna have a PR bonanza today.

PR #13203, opened 2023-06-22
[batch] Mount worker deploy config instead of using k8s secret

daniel-goldstein:
Once the current auth overhaul is through and we go "keyless", the only secret left that we mount into user jobs is the deploy config, which at that point feels kind of silly. It's also nearly always the same deploy config value that we serialize and write to a file for every job. It seems cleaner and simpler to me that we create one deploy config for the worker, and the worker readonly mounts that config into every job.

Note that this is overridable so that any pre-existing jobs that specify a deploy-config secret and all the special deploy-config stuff that we do in build.yaml should not be affected


daniel-goldstein:
I think `test-test-gsa-key` is pretty awful sounding but I'm not sure how better to call a test user for test namespaces and it's consistent with the `test-auth`, `test-batch`, etc. names.

batch/test/test_batch.py
```diff
@@ -931,12 +932,9 @@ def test_cant_submit_to_default_with_other_ns_creds(client: BatchClient):
     )
     b = bb.submit()
     status = j.wait()
-    if NAMESPACE == 'default':
-        assert status['state'] == 'Success', str((status, b.debug_info()))
-    else:
```
jigold, line 936: Do we not need this test anymore for not being able to log in?

jigold (dismissed):
I was having trouble figuring out exactly what the test was supposed to be doing.

batch/batch/cloud/azure/driver/create_instance.py
```diff
@@ -218,6 +219,11 @@ def create_vm_config(
 
 {make_global_config_str}
 
+mkdir /deploy-config
+cat >/deploy-config/deploy-config.json <<EOF
+{ json.dumps(get_deploy_config().with_location('gce').get_config()) }
```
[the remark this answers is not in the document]
daniel-goldstein: `azure`'s not a valid value here. One of those things that never got migrated, we mean `gce` to just mean "in a batch worker". Valid locations are `external`, `k8s` and `gce`.

batch/test/test_batch.py
daniel-goldstein: This test was meant to make sure that you couldn't submit jobs to default through various "adversarial" attempts at changing the deploy config. However, with this change to mounting the deploy config as readonly, the job can't `rm` or `mv` it anymore, and that's true for both test and production jobs. So this test is primarily to show that we correctly mount the deploy-config as readonly. The above test where I use the env variable is to show that you can't access default from a test namespace even if you override what is inside the deploy config in your code.

jigold (approved):

PR #13204, opened 2023-06-22
[ci] You can't get away that easily Tim

chrisvittal:
It will be way better for us to be able to ensure Tim's outstanding PRs continue to be tested. Also, welcome to the rotation @ehigham.


daniel-goldstein (approved):
Ya we should probably make it at some point such that whether a PR is tested depends not on the original author but the author of the latest commit.

ehigham:
In my defense, @danking suggested I don't put a team and instead leave blank... I haven't been skiving!

PR #13206, opened 2023-06-22
[query/vds] Rework VDS combiner so that it can run via Query on Batch

chrisvittal:
Key changes:

- Remove old VCF combiner
- Add StreamZipJoinProducers an IR that takes an array, and a function from array.elementType to stream and zip joins the result of calling that function on each member of the array.
- Combine Table._generate and this new stream zip operation to rewrite the gvcf merge stage of the vds combiner in O(1) IR


chrisvittal (commented):
The core of the change (not my work), looks good to me. 

The stream zip code duplication is present and will be handled in the future.

patrick-schultz (approved):

PR #13207, opened 2023-06-22
[auth] Add internal hail identity UIDs to the users table

daniel-goldstein:
In order to start using Google or AAD access tokens instead of hail-minted tokens, we need to be able to identify a service account in the system by its UID at their identity provider. In Google, this UID is the `uniqueId` field of the Service Account. In AAD, it is the Service Principal Object ID. In an upcoming change, we'll update the user creation process to add this ID upon user creation, at which point we will be able to safely remove this code that updates existing records.

I've marked this PR as `full-deploy` so the AUS folks can roll this commit out specifically before this column is relied on. This way we can safely remove the loop in a follow-up PR and know they will have run this code to populate the column.

### DIFF
--- a/auth/auth/driver/driver.py
+++ b/auth/auth/driver/driver.py
@@ -4,7 +4,7 @@ import json
 import logging
 import os
 import random
-from typing import Any, Awaitable, Callable, Dict, List
+from typing import Any, Awaitable, Callable, Dict, List, Optional
 
 import aiohttp
 import kubernetes_asyncio.client
@@ -16,6 +16,8 @@ from gear.clients import get_identity_client
 from gear.cloud_config import get_gcp_config, get_global_config
 from hailtop import aiotools, httpx
 from hailtop import batch_client as bc
+from hailtop.aiocloud.aioazure import AzureGraphClient
+from hailtop.aiocloud.aiogoogle import GoogleIAmClient
 from hailtop.utils import secret_alnum_string, time_msecs
 
 log = logging.getLogger('auth.driver')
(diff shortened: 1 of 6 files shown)


hail/python/hailtop/aiocloud/aioazure/client/graph_client.py
```diff
@@ -5,6 +5,12 @@
 
 
 class AzureGraphClient(AzureBaseClient):
+    required_scopes = ['https://graph.microsoft.com/.default']
+
     def __init__(self, session: Optional[AzureSession] = None, **kwargs):
+        if 'scopes' in kwargs:
+            kwargs['scopes'] += AzureGraphClient.required_scopes
+        else:
+            kwargs['scopes'] = AzureGraphClient.required_scopes
```
daniel-goldstein, line 14: I made this change because I was prototyping locally and `AzureGraphClient` is unusable without the graph scope.

daniel-goldstein:
#13131 is stacked on top of this. Tests passing in that PR means these are the correct IDs that can be used to identify identities in GCP IAM and Azure AD.

auth/auth/driver/driver.py
```diff
@@ -164,7 +170,7 @@ async def create(self, username):
 
     async def _delete(self, gsa_email):
         try:
-            await self.iam_client.delete(f'/serviceAccounts/{gsa_email}/keys')
+            await self.iam_client.delete(f'/serviceAccounts/{gsa_email}')
```
daniel-goldstein, line 173: I can move this out into a separate PR but noticed that deletion of GSAs was broken and this fixes it.

daniel-goldstein:
@danking bump

danking (approved):
LGTM

PR #13210, opened 2023-06-23
[batch] Avoid computing billing project cost when not needed

jigold:


daniel-goldstein (dismissed):
I think this is getting a little hard to follow with all these conditionals. Can we just make separate functions for these separate queries? I think it's easier to think of what the two pages need in isolation from one another.

batch/batch/utils.py
```diff
@@ -115,19 +115,34 @@ def __repr__(self):
...
 
     if user:
-        where_conditions.append("JSON_CONTAINS(users, JSON_QUOTE(%s))")
+        conditions.append("JSON_CONTAINS(users, JSON_QUOTE(%s))")
         args.append(user)
 
     if billing_project:
-        where_conditions.append('billing_projects.name_cs = %s')
+        conditions.append('billing_projects.name_cs = %s')
         args.append(billing_project)
```
daniel-goldstein, line 138: While this deduplicates some code, it also fragments the query, and I can't really understand what this function is doing unless I look at where it's used (where does this users json come from?). I think it's worth sacrificing the duplication to put this code in place in the functions that call this one instead of having this function.

batch/batch/utils.py
```diff
@@ -161,14 +176,36 @@ async def query_billing_projects(db, user=None, billing_project=None):
 LOCK IN SHARE MODE;
 '''
 
-    def record_to_dict(record):
-        if record['users'] is None:
-            record['users'] = []
-        else:
-            record['users'] = json.loads(record['users'])
-        return record
+    billing_projects = [users_record_to_dict(record) async for record in db.execute_and_fetchall(sql, tuple(args))]
```
daniel-goldstein, line 179: This isn't new in this PR so feel free to take or leave this critique, but I personally find functions that both mutate inputs hard to reason about, as I would expect a void function to mutate and a function with a return value to leave its inputs alone. I might prefer to write this the following way:

```python
billing_projects = []
async for record in db.execute_and_fetchall(sql, tuple(args)):
    record['users'] = json.loads(record['users']) if record['users'] is not None else []
    billing_projects.append(record)    
```

daniel-goldstein (dismissed):

daniel-goldstein (approved):

PR #13213, opened 2023-06-26
[batch] Compact aggregated_billing_*_resources_v3 tables periodically

jigold:


batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,139 @@ async def monitor_system(app):
...
+FROM {scratch_table_name}
+ORDER BY billing_project DESC, `user` DESC, resource_id DESC
+LIMIT 1;
+''',
+                args,
+            )
+        finally:
+            await db.just_execute(f'DROP TEMPORARY TABLE IF EXISTS {scratch_table_name};')
+
+        if next_key is None:
```
daniel-goldstein, line 1384: I think this is actually `last_processed_key` right?

batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,139 @@ async def monitor_system(app):
...
+INSERT INTO aggregated_billing_project_user_resources_by_date_v3 (billing_date, billing_project, `user`, resource_id, token, `usage`)
+SELECT billing_date, billing_project, `user`, resource_id, 0, `usage`
+FROM {scratch_table_name}
+ON DUPLICATE KEY UPDATE `usage` = aggregated_billing_project_user_resources_by_date_v3.`usage` + {scratch_table_name}.`usage`;
+
+SELECT billing_date, billing_project, `user`, resource_id
+FROM {scratch_table_name}
+ORDER BY billing_date DESC, billing_project DESC, `user` DESC, resource_id DESC
+LIMIT 1;
+''',
```
daniel-goldstein, line 1451: Could we possibly do this inside a transaction and assert that, for the locked rows, the total usage for a given key before and after the compaction is the same?

daniel-goldstein: I think an assertion of that sort would make me comfortable deploying this code, but it would be nice to follow up with some more observability. A couple of ideas that I think would be really illustrative:
- Sampling total usage for a billing project/user and see that usage is monotonically increasing over time
- Tracking a summary/histogram of how many rows/tokens exist for a particular key. On a long-running batch we should be able to see that rise and fall on a five-minute period.

Those two graphs together would give me a lot more confidence that this is working as expected, but I understand if you'd want to do that in parts.

batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,139 @@ async def monitor_system(app):
...
+        token = secret_alnum_string(5)
+        scratch_table_name = f'`scratch_{token}`'
+
+        try:
+            next_key = await db.execute_and_fetchone(
+                f'''
+CREATE TEMPORARY TABLE {scratch_table_name} ENGINE=MEMORY
+AS (SELECT billing_project, `user`, resource_id, CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
+    FROM aggregated_billing_project_user_resources_v3
+    WHERE token != 0 {where_condition}
```
daniel-goldstein, line 1356: This approach is conceptually very clear but has a bit of a stop-the-world effect for a particular batch. Maybe it's not a big deal so I'm happy to start with this solution, but if the blocking ever creates noticeable halts in throughput we might want to start randomly sampling subsets of the tokens so we don't lock all the rows for a particular resource at one point in time.

daniel-goldstein (changes requested):
So excited to see what this does for performance, and with such few lines changed!

danking:
Am I meant to do a full second review or something else here? It looks like Daniel is already doing a review.

jigold:
I want another review from you, but let me address Daniel's comments first as the exact query is going to change. I'll unassigned you for now and then when we're done iterating, you can do a final pass.

daniel-goldstein (dismissed):
I think this audit is really valuable! But frankly I'm having a very hard time digesting the SQL here, and it makes me question whether this specific approach of doing everything in one query is worth the tradeoff in complexity. The thing that I'm drawn back to is that right now we're not doing any compaction, so any compaction at all is a big win compared to what we currently have, especially considering we don't currently have many concurrent users of the system. What do you think of the following high-level approach?

On every invocation of the compaction loop we execute a query that finds the `(billing_project, user, resource_id)` tuple with the largest number of tokenized rows. This can be non-locking because I don't particularly care if it's not correct a moment later, it just gives us a good candidate to compact. We'll get the others later.

Then in one transaction do the following all as separate queries
1. Select (locking) the sum total usage for that `(billing_project, user, resource_id)` *across all tokens*
2. Perform an update that *sets* `(billing_project, user, resource_id, 0)` to that usage
3. Delete rows for `token != 0`
4. Select the sum total usage for all tokens and check in python that it equals the result of step 1. If it doesn't roll back the transaction. It's pretty obvious in this setup that selecting for all tokens is really just selecting the value we just inserted, but the point is to future-proof the fact that the total usage in this process never changes.


What do you think of this approach?

jigold:
I think your approach is better. I was over optimizing for the case where we have lots of active users.

jigold:
We should also assert the number of rows is equal to 1 (token == 0) at the end.

daniel-goldstein (changes requested):
This looks so much cleaner. Just one comment about the new approach.

jigold:
The query to get the tokens will take ~3 seconds initially and then should decrease over time.

jigold:
For posterity, we decided to make the code / SQL queries simpler here trading off efficiency and speed with having easy to verify correctness and readability. Once the initial compaction is done, this should run pretty quickly and we can scale back how aggressively we run the compaction. We may have to change this approach in the future if there's lots of simultaneous users.

daniel-goldstein:
> How quickly do you want to get through the initial compaction of the entire table? There's ~60K records to compact (billing_project, `user`, resource_id).
> 
> If we do one per second, it will take ~16 hours. 59058 / (3600) = 16

Ya I think it really doesn't matter at all how long this takes. This is fine

batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,150 @@ async def monitor_system(app):
...
+            '''
+DELETE FROM aggregated_billing_project_user_resources_v3
+WHERE billing_project = %s AND `user` = %s AND resource_id = %s AND token != 0;
+''',
+            (target['billing_project'], target['user'], target['resource_id']),
+        )
+
+        new_usage = await tx.execute_and_fetchone(
+            '''
+SELECT billing_project, `user`, resource_id, CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
```
daniel-goldstein, line 1384:
```suggestion
SELECT CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
```

batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,150 @@ async def monitor_system(app):
...
+FROM aggregated_billing_project_user_resources_v3
+WHERE billing_project = %s AND `user` = %s AND resource_id = %s
+GROUP BY billing_project, `user`, resource_id;
+''',
+            (target['billing_project'], target['user'], target['resource_id']),
+        )
+
+        if new_usage != original_usage:
```
daniel-goldstein, line 1392: are these dictionaries guaranteed to be the same? Should we instead check `new_usage['usage'] == original_usage['usage']`?

batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,150 @@ async def monitor_system(app):
...
+''',
+            (target['billing_project'], target['user'], target['resource_id']),
+        )
+
+        new_usage = await tx.execute_and_fetchone(
+            '''
+SELECT billing_project, `user`, resource_id, CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
+FROM aggregated_billing_project_user_resources_v3
+WHERE billing_project = %s AND `user` = %s AND resource_id = %s
+GROUP BY billing_project, `user`, resource_id;
```
daniel-goldstein, line 1387: I don't think we need this group by. Should be just like the first SELECT if I understand correctly.

daniel-goldstein (dismissed):
Just a few more observations.

batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,150 @@ async def monitor_system(app):
...
+            '''
+SELECT billing_project, `user`, resource_id
+FROM (
+  SELECT billing_project, `user`, resource_id, COUNT(*) AS n_tokens
+  FROM aggregated_billing_project_user_resources_v3
+  WHERE token != 0
+  GROUP BY billing_project, `user`, resource_id
+) AS t
+ORDER BY n_tokens DESC
+LIMIT 1;
```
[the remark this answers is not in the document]
jigold: Yes, if we want to sort by `n_tokens`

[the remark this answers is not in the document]
jigold: Actually, let me double check.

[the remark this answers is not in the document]
jigold: It does work! Thanks! I learned something new.

batch/batch/driver/main.py
jigold: We can either specify only to get the usage for `token=0` as that **should be** the only record. However, I think this is safer in case extra records are there. At least we'll know we haven't changed the database values even if it's not quite what we wanted.

jigold: The group by here shouldn't change the performance as the WHERE filter executes first.

jigold:
Assuming the tests are passing, this should be pretty close to being done. I propose we both sleep on it and relook at it tomorrow morning, triple check we're only modifying the `v3` tables and then merge it once we both feel good about it.

jigold:
Or should we wait for Dan to review next week and not wait for you. That might actually be better.

batch/batch/driver/main.py
jigold: We need both selects. We need to add a FOR UPDATE to the rows we want to lock. But if we add it to the first query, then we're going to lock the entire table as it reads all rows. So the second select is really a locking point query.

danking: This query must have a value we already know given the semantics of MySQL, AFAICT. We first run:
```
SELECT CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
FROM aggregated_billing_project_user_resources_v3
WHERE billing_project = %s AND `user` = %s AND resource_id = %s
FOR UPDATE;
```
This computes the total usage for a given bp, user, resource id *and also* prevents any other transaction from inserting, modifying, or deleting a record with this bp, this user, and this resource id.

Then, if we make my proposed simplification above, we run this:
```
DELETE FROM aggregated_billing_project_user_resources_v3
WHERE billing_project = %s AND `user` = %s AND resource_id = %s;
```
This could only delete the records which we just summed because we're holding an exclusive lock (including any necessary gap locks) on said records.

Finally we run
```
INSERT INTO aggregated_billing_project_user_resources_v3 (billing_project, `user`, resource_id, token, `usage`)
VALUES (%s, %s, %s, %s, %s)
```
Since we know there can be no records with this bp, user, and resource id (since we have an exclusive lock and also just deleted them all), this must insert a single record without failing. We now know for an absolute fact that we have exactly one record. We know its usage number too.

At this point, this query can only return the usage we know it to be.
```
SELECT CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
FROM aggregated_billing_project_user_resources_v3
WHERE billing_project = %s AND `user` = %s AND resource_id = %s
GROUP BY billing_project, `user`, resource_id;
```

batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,149 @@ async def monitor_system(app):
...
+WHERE billing_project = %s AND `user` = %s AND resource_id = %s
+GROUP BY billing_project, `user`, resource_id;
+''',
+            (target['billing_project'], target['user'], target['resource_id']),
+        )
+
+        if new_usage['usage'] != original_usage['usage']:
+            raise ValueError(
+                f'problem in audit for {target}. original usage = {original_usage} but new usage is {new_usage}. aborting'
+            )
```
danking, line 1399: Given the semantics of MySQL, this if will never trigger, I don't think we should include it.

batch/batch/driver/main.py
```diff
@@ -1422,6 +1573,10 @@ async def on_startup(app):
     task_manager.ensure_future(periodically_call(15, monitor_system, app))
     task_manager.ensure_future(periodically_call(5, refresh_globals_from_db, app, db))
 
+    # 3 seconds was chosen for the initial compaction time to get through all existing records in a reasonable amount of time
+    task_manager.ensure_future(periodically_call(1, compact_agg_billing_project_users_table, db))
```
danking, line 1577: Given that the query to find a record to compact takes 2s, does it make sense to try to run this every second? Maybe every 10s?

batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,149 @@ async def monitor_system(app):
...
+    @transaction(db)
+    async def compact(tx: Transaction):
+        target = await tx.execute_and_fetchone(
+            '''
+SELECT billing_date, billing_project, `user`, resource_id, COUNT(*) AS n_tokens
+FROM aggregated_billing_project_user_resources_by_date_v3
+WHERE token != 0
+GROUP BY billing_date, billing_project, `user`, resource_id
+ORDER BY n_tokens DESC
+LIMIT 1;
```
danking, line 1414: For future reference: this must scan the table, there's no way to do the order-by-limit without generating the full ordered-by result.

This table currently contains 2722023 records, 2704178 are token != 0. I ran this query a few times. The first time it takes 14s to execute this query. Subsequent runs take ~2s. The query should scale as O(BPs + USERS + TIME_IN_DAYS).

batch/batch/driver/main.py
```diff
@@ -1422,6 +1573,10 @@ async def on_startup(app):
     task_manager.ensure_future(periodically_call(15, monitor_system, app))
     task_manager.ensure_future(periodically_call(5, refresh_globals_from_db, app, db))
 
+    # 3 seconds was chosen for the initial compaction time to get through all existing records in a reasonable amount of time
+    task_manager.ensure_future(periodically_call(1, compact_agg_billing_project_users_table, db))
+    task_manager.ensure_future(periodically_call(1, compact_agg_billing_project_users_by_date_table, db))
```
danking, line 1578: Given that the query to find a record takes 2-15s, maybe we should do this every 60s?

batch/batch/driver/main.py
danking: Ah, was this comment what lead to the sum-and-if? Maybe we should wait for Daniel to return and talk through his concerns.

danking (dismissed):

jigold:
Thanks for the comments! I think we should wait for Daniel to return and have a meeting to discuss the best approach as I'm not fully convinced this is the optimal long term approach given the timings of the queries. I think the hope was with the index on `token`, after an initial long compaction period, the overhead of records with token != 0 should be small and that initial query should be fast.

Let me know if you want me to pick this back up sooner.

jigold:
Just to be sure, I checked the logs for errors and found none. I manually logged into the database and watched the compaction happen for both tables. I don't think we can do any more due diligence with this. Let's plan on merging on Tuesday when you're back from vacation.

```
mysql> select * from aggregated_billing_project_user_resources_v3 where resource_id = 6 limit 100;
+----------------------------------+------+-------------+-------+------------+
| billing_project                  | user | resource_id | token | usage      |
+----------------------------------+------+-------------+-------+------------+
| __testproject_iizhz61z7543_FUitX | test |           6 |     0 |    1817536 |
| __testproject_iizhz61z7543_uvxWn | test |           6 |     0 |   11331136 |
| ci                               | ci   |           6 |     0 |   79640784 |
| test                             | test |           6 |     0 | 4063028160 |
| test                             | test |           6 |     1 |     189760 |
| test                             | test |           6 |     3 |     607168 |
| test                             | test |           6 |     4 |     749952 |
| test                             | test |           6 |     5 |      46912 |
| test                             | test |           6 |     6 |     158336 |
| test                             | test |           6 |     7 |      70336 |
| test                             | test |           6 |     8 |     167680 |
| test                             | test |           6 |     9 |     523136 |
| test                             | test |           6 |    10 |      40640 |
| test                             | test |           6 |    11 |     616448 |
| test                             | test |           6 |    12 |     497024 |
| test                             | test |           6 |    14 |      87680 |
| test                             | test |           6 |    15 |     111104 |
| test                             | test |           6 |    16 |     120128 |
| test                             | test |           6 |    17 |      28736 |
| test                             | test |           6 |    19 |      42240 |
| test                             | test |           6 |    20 |     271232 |
| test                             | test |           6 |    21 |      88320 |
| test                             | test |           6 |    22 |     149760 |
| test                             | test |           6 |    23 |      47232 |
| test                             | test |           6 |    24 |      45888 |
| test                             | test |           6 |    25 |      41664 |
| test                             | test |           6 |    27 |      56704 |
| test                             | test |           6 |    28 |      36864 |
| test                             | test |           6 |    30 |      57792 |
| test                             | test |           6 |    31 |      62848 |
| test                             | test |           6 |    32 |      40320 |
| test                             | test |           6 |    33 |      61888 |
| test                             | test |           6 |    34 |      43520 |
| test                             | test |           6 |    35 |     219328 |
| test                             | test |           6 |    36 |     141760 |
| test                             | test |           6 |    38 |     157632 |
| test                             | test |           6 |    40 |      72064 |
| test                             | test |           6 |    41 |     317888 |
| test                             | test |           6 |    42 |      83648 |
| test                             | test |           6 |    43 |     123200 |
| test                             | test |           6 |    44 |     196032 |
| test                             | test |           6 |    45 |      62848 |
| test                             | test |           6 |    47 |     232768 |
| test                             | test |           6 |    48 |     937472 |
| test                             | test |           6 |    49 |      78272 |
| test                             | test |           6 |    50 |      88320 |
| test                             | test |           6 |    51 |     128320 |
| test                             | test |           6 |    52 |      36160 |
| test                             | test |           6 |    53 |    1268800 |
| test                             | test |           6 |    54 |      37568 |
| test                             | test |           6 |    55 |      53568 |
| test                             | test |           6 |    56 |     147520 |
| test                             | test |           6 |    57 |     102784 |
| test                             | test |           6 |    58 |      95360 |
| test                             | test |           6 |    60 |     156480 |
| test                             | test |           6 |    61 |      54976 |
| test                             | test |           6 |    62 |     123072 |
| test                             | test |           6 |    63 |      88128 |
| test                             | test |           6 |    64 |     202112 |
| test                             | test |           6 |    65 |      54848 |
| test                             | test |           6 |    67 |     715456 |
| test                             | test |           6 |    69 |      51264 |
| test                             | test |           6 |    70 |     127296 |
| test                             | test |           6 |    71 |      95040 |
| test                             | test |           6 |    72 |      72320 |
| test                             | test |           6 |    73 |      81856 |
| test                             | test |           6 |    74 |      28352 |
| test                             | test |           6 |    76 |     121536 |
| test                             | test |           6 |    77 |     132736 |
| test                             | test |           6 |    78 |      42432 |
| test                             | test |           6 |    79 |     563648 |
| test                             | test |           6 |    80 |     943616 |
| test                             | test |           6 |    81 |      34048 |
| test                             | test |           6 |    82 |      23424 |
| test                             | test |           6 |    83 |      37952 |
| test                             | test |           6 |    84 |     428352 |
| test                             | test |           6 |    86 |      40576 |
| test                             | test |           6 |    88 |     141312 |
| test                             | test |           6 |    89 |     203584 |
| test                             | test |           6 |    90 |     465536 |
| test                             | test |           6 |    91 |      47808 |
| test                             | test |           6 |    92 |      35456 |
| test                             | test |           6 |    93 |     768192 |
| test                             | test |           6 |    94 |     172416 |
| test                             | test |           6 |    95 |      48384 |
| test                             | test |           6 |    96 |     146816 |
| test                             | test |           6 |    97 |      69568 |
| test                             | test |           6 |    99 |     203520 |
| test                             | test |           6 |   100 |      44608 |
| test                             | test |           6 |   101 |     120512 |
| test                             | test |           6 |   102 |     233472 |
| test                             | test |           6 |   103 |     219584 |
| test                             | test |           6 |   105 |     399872 |
| test                             | test |           6 |   106 |     617600 |
| test                             | test |           6 |   107 |     114880 |
| test                             | test |           6 |   108 |      47616 |
| test                             | test |           6 |   109 |      98304 |
| test                             | test |           6 |   110 |      73472 |
| test                             | test |           6 |   111 |      44928 |
| test                             | test |           6 |   112 |      37696 |
+----------------------------------+------+-------------+-------+------------+
100 rows in set (0.00 sec)

mysql> select * from aggregated_billing_project_user_resources_v3 where resource_id = 6 limit 100;
+----------------------------------+------+-------------+-------+------------+
| billing_project                  | user | resource_id | token | usage      |
+----------------------------------+------+-------------+-------+------------+
| __testproject_iizhz61z7543_FUitX | test |           6 |     0 |    1817536 |
| __testproject_iizhz61z7543_uvxWn | test |           6 |     0 |   11331136 |
| ci                               | ci   |           6 |     0 |   79640784 |
| test                             | test |           6 |     0 | 4094463424 |
+----------------------------------+------+-------------+-------+------------+
4 rows in set (0.00 sec)

mysql> select * from aggregated_billing_project_user_resources_v3 where resource_id = 6 limit 100;
+----------------------------------+------+-------------+-------+------------+
| billing_project                  | user | resource_id | token | usage      |
+----------------------------------+------+-------------+-------+------------+
| __testproject_iizhz61z7543_FUitX | test |           6 |     0 |    1817536 |
| __testproject_iizhz61z7543_uvxWn | test |           6 |     0 |   11331136 |
| ci                               | ci   |           6 |     0 |   79640784 |
| test                             | test |           6 |     0 | 4094463424 |
| test                             | test |           6 |    15 |      54592 |
| test                             | test |           6 |    35 |      53568 |
| test                             | test |           6 |    78 |      39104 |
| test                             | test |           6 |   135 |      35712 |
| test                             | test |           6 |   164 |      54080 |
+----------------------------------+------+-------------+-------+------------+
9 rows in set (0.00 sec)

mysql> select * from aggregated_billing_project_user_resources_v3 where resource_id = 6 limit 100;
+----------------------------------+------+-------------+-------+------------+
| billing_project                  | user | resource_id | token | usage      |
+----------------------------------+------+-------------+-------+------------+
| __testproject_iizhz61z7543_FUitX | test |           6 |     0 |    1817536 |
| __testproject_iizhz61z7543_uvxWn | test |           6 |     0 |   11331136 |
| ci                               | ci   |           6 |     0 |   79640784 |
| test                             | test |           6 |     0 | 4094463424 |
| test                             | test |           6 |    10 |      49984 |
| test                             | test |           6 |    15 |      54592 |
| test                             | test |           6 |    24 |      49280 |
| test                             | test |           6 |    29 |      53056 |
| test                             | test |           6 |    31 |      52480 |
| test                             | test |           6 |    32 |      48000 |
| test                             | test |           6 |    35 |     107008 |
| test                             | test |           6 |    37 |      44160 |
| test                             | test |           6 |    75 |      51520 |
| test                             | test |           6 |    78 |      39104 |
| test                             | test |           6 |    81 |      50688 |
| test                             | test |           6 |    83 |     216064 |
| test                             | test |           6 |   102 |      48384 |
| test                             | test |           6 |   119 |      52544 |
| test                             | test |           6 |   126 |      56064 |
| test                             | test |           6 |   135 |      81728 |
| test                             | test |           6 |   142 |      57408 |
| test                             | test |           6 |   145 |      55488 |
| test                             | test |           6 |   152 |      48128 |
| test                             | test |           6 |   155 |      50176 |
| test                             | test |           6 |   160 |      54464 |
| test                             | test |           6 |   164 |      54080 |
| test                             | test |           6 |   173 |      46080 |
| test                             | test |           6 |   175 |      48704 |
| test                             | test |           6 |   176 |      45696 |
| test                             | test |           6 |   183 |      46080 |
| test                             | test |           6 |   189 |      47168 |
| test                             | test |           6 |   194 |      30912 |
| test                             | test |           6 |   199 |      47232 |
+----------------------------------+------+-------------+-------+------------+
33 rows in set (0.00 sec)
```

```
mysql> select * from aggregated_billing_project_user_resources_by_date_v3 where resource_id = 6 limit 100;
+--------------+----------------------------------+------+-------------+-------+------------+
| billing_date | billing_project                  | user | resource_id | token | usage      |
+--------------+----------------------------------+------+-------------+-------+------------+
| 2023-07-13   | __testproject_iizhz61z7543_FUitX | test |           6 |     0 |    1817536 |
| 2023-07-13   | __testproject_iizhz61z7543_uvxWn | test |           6 |     0 |   11331136 |
| 2023-07-13   | ci                               | ci   |           6 |     0 |   79640784 |
| 2023-07-13   | test                             | test |           6 |     0 | 4142226688 |
| 2023-07-13   | test                             | test |           6 |     1 |     108608 |
| 2023-07-13   | test                             | test |           6 |     2 |      80576 |
| 2023-07-13   | test                             | test |           6 |     5 |      35648 |
| 2023-07-13   | test                             | test |           6 |     7 |     578240 |
| 2023-07-13   | test                             | test |           6 |     9 |      33024 |
| 2023-07-13   | test                             | test |           6 |    10 |      33472 |
| 2023-07-13   | test                             | test |           6 |    11 |     110464 |
| 2023-07-13   | test                             | test |           6 |    14 |      47744 |
| 2023-07-13   | test                             | test |           6 |    15 |      45440 |
| 2023-07-13   | test                             | test |           6 |    16 |      79616 |
| 2023-07-13   | test                             | test |           6 |    17 |     604928 |
| 2023-07-13   | test                             | test |           6 |    18 |      78016 |
| 2023-07-13   | test                             | test |           6 |    19 |      87040 |
| 2023-07-13   | test                             | test |           6 |    20 |      89792 |
| 2023-07-13   | test                             | test |           6 |    21 |     188736 |
| 2023-07-13   | test                             | test |           6 |    22 |      75648 |
| 2023-07-13   | test                             | test |           6 |    24 |      96128 |
| 2023-07-13   | test                             | test |           6 |    25 |    2143680 |
| 2023-07-13   | test                             | test |           6 |    26 |      33728 |
| 2023-07-13   | test                             | test |           6 |    27 |      70400 |
| 2023-07-13   | test                             | test |           6 |    28 |      93312 |
| 2023-07-13   | test                             | test |           6 |    31 |      53696 |
| 2023-07-13   | test                             | test |           6 |    32 |      70976 |
| 2023-07-13   | test                             | test |           6 |    33 |      57280 |
| 2023-07-13   | test                             | test |           6 |    34 |      99712 |
| 2023-07-13   | test                             | test |           6 |    35 |      87360 |
| 2023-07-13   | test                             | test |           6 |    36 |     144384 |
| 2023-07-13   | test                             | test |           6 |    37 |     703616 |
| 2023-07-13   | test                             | test |           6 |    39 |     344832 |
| 2023-07-13   | test                             | test |           6 |    42 |     129152 |
| 2023-07-13   | test                             | test |           6 |    44 |      56000 |
| 2023-07-13   | test                             | test |           6 |    45 |     100800 |
| 2023-07-13   | test                             | test |           6 |    48 |      86592 |
| 2023-07-13   | test                             | test |           6 |    49 |      94336 |
| 2023-07-13   | test                             | test |           6 |    50 |      39744 |
| 2023-07-13   | test                             | test |           6 |    51 |      73344 |
| 2023-07-13   | test                             | test |           6 |    52 |     357248 |
| 2023-07-13   | test                             | test |           6 |    54 |      55872 |
| 2023-07-13   | test                             | test |           6 |    56 |      76864 |
| 2023-07-13   | test                             | test |           6 |    57 |      94016 |
| 2023-07-13   | test                             | test |           6 |    60 |      93504 |
| 2023-07-13   | test                             | test |           6 |    61 |     149952 |
| 2023-07-13   | test                             | test |           6 |    62 |      48064 |
| 2023-07-13   | test                             | test |           6 |    63 |     250560 |
| 2023-07-13   | test                             | test |           6 |    64 |      64704 |
| 2023-07-13   | test                             | test |           6 |    65 |      51968 |
| 2023-07-13   | test                             | test |           6 |    66 |      90752 |
| 2023-07-13   | test                             | test |           6 |    67 |      68928 |
| 2023-07-13   | test                             | test |           6 |    68 |      91712 |
| 2023-07-13   | test                             | test |           6 |    69 |      93632 |
| 2023-07-13   | test                             | test |           6 |    70 |     118784 |
| 2023-07-13   | test                             | test |           6 |    72 |      76544 |
| 2023-07-13   | test                             | test |           6 |    73 |     145472 |
| 2023-07-13   | test                             | test |           6 |    74 |      47040 |
| 2023-07-13   | test                             | test |           6 |    77 |     226944 |
| 2023-07-13   | test                             | test |           6 |    79 |     157568 |
| 2023-07-13   | test                             | test |           6 |    80 |      37120 |
| 2023-07-13   | test                             | test |           6 |    81 |     106496 |
| 2023-07-13   | test                             | test |           6 |    82 |     494464 |
| 2023-07-13   | test                             | test |           6 |    83 |      79424 |
| 2023-07-13   | test                             | test |           6 |    85 |     148608 |
| 2023-07-13   | test                             | test |           6 |    86 |      52288 |
| 2023-07-13   | test                             | test |           6 |    87 |     181696 |
| 2023-07-13   | test                             | test |           6 |    89 |      39232 |
| 2023-07-13   | test                             | test |           6 |    90 |      90624 |
| 2023-07-13   | test                             | test |           6 |    92 |     133504 |
| 2023-07-13   | test                             | test |           6 |    93 |     156672 |
| 2023-07-13   | test                             | test |           6 |    95 |     190528 |
| 2023-07-13   | test                             | test |           6 |    96 |      38016 |
| 2023-07-13   | test                             | test |           6 |    98 |      99392 |
| 2023-07-13   | test                             | test |           6 |    99 |     109568 |
| 2023-07-13   | test                             | test |           6 |   100 |      92928 |
| 2023-07-13   | test                             | test |           6 |   101 |      75712 |
| 2023-07-13   | test                             | test |           6 |   105 |      50048 |
| 2023-07-13   | test                             | test |           6 |   107 |      41472 |
| 2023-07-13   | test                             | test |           6 |   109 |      30208 |
| 2023-07-13   | test                             | test |           6 |   110 |      72064 |
| 2023-07-13   | test                             | test |           6 |   111 |      36672 |
| 2023-07-13   | test                             | test |           6 |   113 |      32832 |
| 2023-07-13   | test                             | test |           6 |   114 |      54400 |
| 2023-07-13   | test                             | test |           6 |   115 |     106240 |
| 2023-07-13   | test                             | test |           6 |   116 |      49216 |
| 2023-07-13   | test                             | test |           6 |   117 |      24640 |
| 2023-07-13   | test                             | test |           6 |   118 |     254080 |
| 2023-07-13   | test                             | test |           6 |   119 |      37056 |
| 2023-07-13   | test                             | test |           6 |   120 |     110784 |
| 2023-07-13   | test                             | test |           6 |   122 |      29888 |
| 2023-07-13   | test                             | test |           6 |   124 |      39616 |
| 2023-07-13   | test                             | test |           6 |   126 |      31232 |
| 2023-07-13   | test                             | test |           6 |   127 |      61696 |
| 2023-07-13   | test                             | test |           6 |   129 |      60096 |
| 2023-07-13   | test                             | test |           6 |   131 |     167040 |
| 2023-07-13   | test                             | test |           6 |   132 |     312768 |
| 2023-07-13   | test                             | test |           6 |   134 |      53312 |
| 2023-07-13   | test                             | test |           6 |   136 |      51840 |
| 2023-07-13   | test                             | test |           6 |   137 |      68224 |
+--------------+----------------------------------+------+-------------+-------+------------+
100 rows in set (0.00 sec)

mysql> select * from aggregated_billing_project_user_resources_by_date_v3 where resource_id = 6 limit 100;
+--------------+----------------------------------+------+-------------+-------+------------+
| billing_date | billing_project                  | user | resource_id | token | usage      |
+--------------+----------------------------------+------+-------------+-------+------------+
| 2023-07-13   | __testproject_iizhz61z7543_FUitX | test |           6 |     0 |    1817536 |
| 2023-07-13   | __testproject_iizhz61z7543_uvxWn | test |           6 |     0 |   11331136 |
| 2023-07-13   | ci                               | ci   |           6 |     0 |   79640784 |
| 2023-07-13   | test                             | test |           6 |     0 | 4160033472 |
+--------------+----------------------------------+------+-------------+-------+------------+
4 rows in set (0.00 sec)

mysql> select * from aggregated_billing_project_user_resources_by_date_v3 where resource_id = 6 limit 100;
+--------------+----------------------------------+------+-------------+-------+------------+
| billing_date | billing_project                  | user | resource_id | token | usage      |
+--------------+----------------------------------+------+-------------+-------+------------+
| 2023-07-13   | __testproject_iizhz61z7543_FUitX | test |           6 |     0 |    1817536 |
| 2023-07-13   | __testproject_iizhz61z7543_uvxWn | test |           6 |     0 |   11331136 |
| 2023-07-13   | ci                               | ci   |           6 |     0 |   79640784 |
| 2023-07-13   | test                             | test |           6 |     0 | 4160033472 |
| 2023-07-13   | test                             | test |           6 |    59 |      51584 |
| 2023-07-13   | test                             | test |           6 |    62 |      55616 |
| 2023-07-13   | test                             | test |           6 |   138 |     700160 |
| 2023-07-13   | test                             | test |           6 |   183 |      54976 |
+--------------+----------------------------------+------+-------------+-------+------------+
8 rows in set (0.00 sec)

mysql> select * from aggregated_billing_project_user_resources_by_date_v3 where resource_id = 6 limit 100;
+--------------+----------------------------------+------+-------------+-------+------------+
| billing_date | billing_project                  | user | resource_id | token | usage      |
+--------------+----------------------------------+------+-------------+-------+------------+
| 2023-07-13   | __testproject_iizhz61z7543_FUitX | test |           6 |     0 |    1817536 |
| 2023-07-13   | __testproject_iizhz61z7543_uvxWn | test |           6 |     0 |   11331136 |
| 2023-07-13   | ci                               | ci   |           6 |     0 |   79640784 |
| 2023-07-13   | test                             | test |           6 |     0 | 4160033472 |
| 2023-07-13   | test                             | test |           6 |    11 |     171840 |
| 2023-07-13   | test                             | test |           6 |    59 |      51584 |
| 2023-07-13   | test                             | test |           6 |    62 |      55616 |
| 2023-07-13   | test                             | test |           6 |   138 |     700160 |
| 2023-07-13   | test                             | test |           6 |   183 |      54976 |
+--------------+----------------------------------+------+-------------+-------+------------+
9 rows in set (0.00 sec)

mysql> select * from aggregated_billing_project_user_resources_by_date_v3 where resource_id = 6 limit 100;
+--------------+----------------------------------+------+-------------+-------+------------+
| billing_date | billing_project                  | user | resource_id | token | usage      |
+--------------+----------------------------------+------+-------------+-------+------------+
| 2023-07-13   | __testproject_iizhz61z7543_FUitX | test |           6 |     0 |    1817536 |
| 2023-07-13   | __testproject_iizhz61z7543_uvxWn | test |           6 |     0 |   11331136 |
| 2023-07-13   | ci                               | ci   |           6 |     0 |   79640784 |
| 2023-07-13   | test                             | test |           6 |     0 | 4160033472 |
| 2023-07-13   | test                             | test |           6 |    11 |     171840 |
| 2023-07-13   | test                             | test |           6 |    59 |      51584 |
| 2023-07-13   | test                             | test |           6 |    62 |      55616 |
| 2023-07-13   | test                             | test |           6 |   138 |     700160 |
| 2023-07-13   | test                             | test |           6 |   183 |      54976 |
+--------------+----------------------------------+------+-------------+-------+------------+
9 rows in set (0.01 sec)
```

batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,143 @@ async def monitor_system(app):
...
+WHERE billing_project = %s AND `user` = %s AND resource_id = %s AND token != 0;
+''',
+            (target['billing_project'], target['user'], target['resource_id']),
+        )
+
+        await tx.execute_update(
+            '''
+INSERT INTO aggregated_billing_project_user_resources_v3 (billing_project, `user`, resource_id, token, `usage`)
+VALUES (%s, %s, %s, %s, %s)
+ON DUPLICATE KEY UPDATE `usage` = %s
```
danking, line 1358: I prefer the delete everything and insert without duplicate key update. It just feels more straightforward to me. That record that you keep around is just gonna get changed by the insert-update anyway. What do you think?

batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,143 @@ async def monitor_system(app):
...
+            )
+
+    targets = db.execute_and_fetchall(
+        '''
+SELECT billing_project, `user`, resource_id, COUNT(*) AS n_tokens
+FROM aggregated_billing_project_user_resources_v3
+WHERE token != 0
+GROUP BY billing_project, `user`, resource_id
+ORDER BY n_tokens DESC
+LIMIT 10000;
```
danking, line 1392: This seems great. It'll amortize the cost of this query across lots of records.

batch/batch/driver/main.py
```diff
@@ -1322,6 +1330,143 @@ async def monitor_system(app):
...
+''',
+        query_name='find_agg_billing_project_user_resource_to_compact',
+    )
+
+    async for target in targets:
+        await compact(target)  # pylint: disable=no-value-for-parameter
```
danking, line 1398: I don't think we should hold the database connection open for the limit-10000 query. That's gonna stream the records 100 at a time from the DB. In the meantime, it keeps the query running and the connection open. I worry we are at risk of intermittent connection issues. I also worry about deadlocking since the DB needs to maintain for the limit-10000 query a consistent view of the aggregation table. That table is usually frequently mutated when batch is working! Finally, I feel a bit weird about mutating the table in one query that you're reading in a another query.

danking (dismissed):
just a couple thoughts!

batch/batch/driver/main.py
jigold: I changed it to get all of the targets in a list before iterating. The deadlocking isn't an issue because we're not holding locks to those tables, we don't care if the list of targets is outdated, and if the target doesn't need to be compacted, then it will just be a no-op when the function is called. But I agree we shouldn't hold the connection the entire time the compaction loop runs.

danking:
Some Spark tests timed out (e.g. in https://ci.hail.is/batches/7644244/jobs/74 it was `test_spectral_moments_4`) which often crashes the JVM leading to the other errors. Retry and post links into Query Dev to alert them that QoS can timeout on spectral moments.

danking (dismissed):
This all seems right to me.

jigold:
@danking Can you check the last two commits to make sure I didn't do anything stupid?

danking (commented):
seems fine to me

jigold:
Can you reapprove?

danking (approved):
oops, sorry, didn't notice the dismissal of the approval

danking:
I resolved the merge conflict.

PR #13214, opened 2023-06-26
[query] refactor BaseIR so children only exposes iterable interface

patrick-schultz:
This is a simple tidying refactoring to hide the `children` array from consumers of `BaseIR`. My main motivation is to make it easier to explore other ir data structure designs, and to migrate to a new design in the future, e.g. to allow for in-place mutation without requiring large-scale changes to every compiler pass, and to simplify how we encode binding structure.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/BaseIR.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/BaseIR.scala
@@ -1,23 +1,53 @@
 package is.hail.expr.ir
 
 import is.hail.types.BaseType
+import is.hail.utils.StackSafe._
+import is.hail.utils._
 
 abstract class BaseIR {
   def typ: BaseType
 
-  def children: IndexedSeq[BaseIR]
+  protected def childrenSeq: IndexedSeq[BaseIR]
 
-  def copy(newChildren: IndexedSeq[BaseIR]): BaseIR
+  def children: Iterable[BaseIR] = childrenSeq
 
-  def deepCopy(): this.type = copy(newChildren = children.map(_.deepCopy())).asInstanceOf[this.type]
+  protected def copy(newChildren: IndexedSeq[BaseIR]): BaseIR
+
+  def deepCopy(): this.type = copy(newChildren = childrenSeq.map(_.deepCopy())).asInstanceOf[this.type]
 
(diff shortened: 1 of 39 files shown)


hail/src/main/scala/is/hail/expr/ir/BaseIR.scala
```diff
@@ -1,23 +1,53 @@
...
+import is.hail.utils._
 
 abstract class BaseIR {
   def typ: BaseType
 
-  def children: IndexedSeq[BaseIR]
+  protected def childrenSeq: IndexedSeq[BaseIR]
 
-  def copy(newChildren: IndexedSeq[BaseIR]): BaseIR
+  def children: Iterable[BaseIR] = childrenSeq
```
ehigham, line 12: It's not clear to me why we need `children` and `childrenSeq` - can you explain?

ehigham:
This seems a little at-odds with what I'm doing with semhash-ing where i'm making direct use of the `children` of a `BaseIR` so I'm hesitant to approve it just now.

hail/src/main/scala/is/hail/expr/ir/BaseIR.scala
patrick-schultz: It's just separating implementation from interface: the implementation is an IndexedSeq of children, but this makes that fact protected, and only exposes an Iterable interface over children to external clients.

ehigham: Ah ok - I publicised and enforced that children should be an `IndexedSeq` on Tim's suggestion because apparently it was a bottleneck. I like the idea of not doing that.

ehigham:
> > This seems a little at-odds with what I'm doing with semhash-ing where i'm making direct use of the children of a BaseIR so I'm hesitant to approve it just now.
> 
> Ah, sure, we can hold off on this if it would cause any trouble. Are you doing anything more than iterating over the children?

I misunderstood the diff. On a second look happy to merge this.

ehigham (approved):

hail/src/main/scala/is/hail/expr/ir/BaseIR.scala
patrick-schultz: I tried to do this in a way that should't affect performance. Keep an eye out for anything that might.

PR #13219, opened 2023-06-27
[batch] Add identifying metadata to Batch requests

jigold:
I realize this looks like a lot of code changes, but it's mostly copying and pasting two SQL procedures and changing one line in each.

This adds 4 bits of metadata to requests that then can be queried as extra metadata:
- batch_id
- job_id
- batch_operation
- job_queue_time

Should be self-explanatory except job_queue time is the time in which the job is first set to ready to when it was scheduled on the worker (exact moment is when the job config is made to send to the worker).

Example logging query. Note that the search on "batch_id" is not optimized so you definitely want to add some kind of time limit that's short on the window to search. I can add my Python script that scrapes these logs and makes a Plotly figure in a separate PR once this goes in.

```
(
resource.labels.container_name="batch"
resource.labels.namespace_name="{namespace}"
) OR (
resource.labels.container_name="batch-driver"
resource.labels.namespace_name="{namespace}"
) OR (
resource.type="gce_instance"
logName:"worker.log"
labels."compute.googleapis.com/resource_name":"{namespace}"
)
jsonPayload.batch_id="{batch_id}"
timestamp >= "{start_timestamp}" {end_timestamp}
```

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -269,6 +269,8 @@ WHERE id = %s AND NOT deleted;
 @routes.get('/api/v1alpha/batches/{batch_id}/jobs')
 @rest_billing_project_users_only
 async def get_jobs_v1(request, userdata, batch_id):  # pylint: disable=unused-argument
+    request['batch_id'] = str(batch_id)
+    request['batch_operation'] = 'batch_list_jobs'
     q = request.query.get('q', '')
     last_job_id = request.query.get('last_job_id')
     if last_job_id is not None:
@@ -280,6 +282,8 @@ async def get_jobs_v1(request, userdata, batch_id):  # pylint: disable=unused-ar
 @routes.get('/api/v2alpha/batches/{batch_id}/jobs')
 @rest_billing_project_users_only
 async def get_jobs_v2(request, userdata, batch_id):  # pylint: disable=unused-argument
+    request['batch_id'] = str(batch_id)
+    request['batch_operation'] = 'batch_list_jobs'
     q = request.query.get('q', '')
     last_job_id = request.query.get('last_job_id')
     if last_job_id is not None:
@@ -735,6 +739,7 @@ async def get_batches(request, userdata):  # pylint: disable=unused-argument
     body = {'batches': batches}
     if last_batch_id is not None:
         body['last_batch_id'] = last_batch_id
+    request['batch_operation'] = 'batch_list_batches'
     return json_response(body)
 
 
@@ -755,9 +760,10 @@ def check_service_account_permissions(user, sa):
 @auth.rest_authenticated_users_only
 async def create_jobs(request: aiohttp.web.Request, userdata: dict):
     app = request.app
-
     batch_id = int(request.match_info['batch_id'])
     job_specs = await json_request(request)
+    request['batch_id'] = str(batch_id)
+    request['batch_operation'] = 'batch_create_jobs'
     return await _create_jobs(userdata, job_specs, batch_id, 1, app)
 
 
@@ -773,6 +779,8 @@ async def create_jobs_for_update(request: aiohttp.web.Request, userdata: dict):
     batch_id = int(request.match_info['batch_id'])
     update_id = int(request.match_info['update_id'])
     job_specs = await json_request(request)
+    request['batch_id'] = str(batch_id)
+    request['batch_operation'] = 'batch_create_jobs'
     return await _create_jobs(userdata, job_specs, batch_id, update_id, app)
 
 
@@ -1059,6 +1067,7 @@ WHERE batch_updates.batch_id = %s AND batch_updates.update_id = %s AND user = %s
         # of pending parents
         if update_id == 1 and len(parent_ids) == 0:
             state = 'Ready'
+            time_ready = time_msecs()
             icr['n_ready_jobs'] += 1
             icr['ready_cores_mcpu'] += cores_mcpu
             if not always_run:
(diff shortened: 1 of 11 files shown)


batch/batch/front_end/front_end.py
```diff
@@ -1673,6 +1702,8 @@ async def _commit_update(app: web.Application, batch_id: int, update_id: int, us
 @routes.delete('/api/v1alpha/batches/{batch_id}')
 @rest_billing_project_users_only
 async def delete_batch(request, userdata, batch_id):  # pylint: disable=unused-argument
+    request['batch_identifier'] = str(batch_id)
+    request['batch_operation'] = 'batch_delete_batch'
```
daniel-goldstein, line 1706: I think this would be better in a decorator like `rest_billing_project_users_only` that adds these fields to the request. Should be able to use `f.__name__` to get the name of the endpoint function.

batch/sql/estimated-current.sql
```diff
@@ -263,6 +263,7 @@ CREATE TABLE IF NOT EXISTS `jobs` (
   `inst_coll` VARCHAR(255),
   `n_regions` INT DEFAULT NULL,
   `regions_bits_rep` BIGINT DEFAULT NULL,
+  `time_ready` BIGINT DEFAULT NULL,
```
daniel-goldstein, line 266: I feel uneasy about adding this kind of metadata to the `jobs` table, it feels like a slippery slope for adding more and more telemetry and muddling the purpose of this table. Why not add a separate table for this?

Something that I think is important with telemetry is that as we track more and more things we might need to consider not turning tracking on for every request/resource in the system. I think we should design for eventually sampling or toggling telemetry instead of locking ourselves into tracking certain metadata forever. In this setup, I can't really tell the difference between a job that is not ready yet and a job for which we are not tracking this telemetry. With a separate table, we can just not add a row for a job we're not tracking.

hail/python/hailtop/hail_logging.py
```diff
@@ -51,10 +51,31 @@ def log(self, request, response, time):
...
+        if job_identifier:
+            extra['job_identifier'] = job_identifier
+
+        batch_operation = request.get('batch_operation')
+        if batch_operation:
+            extra['batch_operation'] = batch_operation
+
+        job_queue_time = request.get('job_queue_time')
+        if job_queue_time:
+            extra['job_queue_time'] = job_queue_time
```
daniel-goldstein, line 77: what if we add a layer of nesting, say `request['batch_telemetry']['batch_operation']` and then we have `extra={**extra, **request.get('batch_telemetry', {})}`

daniel-goldstein (changes requested):
A couple suggestions to make the changes a little less invasive but overall a highly useful change

batch/sql/estimated-current.sql
daniel-goldstein: Another note: I think it's worth thinking about this kind of telemetry in terms of state change and *event* tracking. I think it's worth considering whether instead of having a `time_ready` column we should have some table that looks like having `(prev_state, new_state, timestamp)`.

jigold: Good suggestion, but let's do that another time with a more fully fleshed out overall plan of monitoring the system?

jigold:
Note to self to look at the logs for this PR before merging to make sure the metadata is actually there.

batch/batch/front_end/front_end.py
```diff
@@ -1112,8 +1124,8 @@ async def insert_jobs_into_db(tx):
             try:
                 await tx.execute_many(
                     '''
-INSERT INTO jobs (batch_id, job_id, update_id, state, spec, always_run, cores_mcpu, n_pending_parents, inst_coll, n_regions, regions_bits_rep)
-VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s);
+INSERT INTO jobs (batch_id, job_id, update_id, state, spec, always_run, cores_mcpu, n_pending_parents, inst_coll, n_regions, regions_bits_rep, time_ready)
+VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s);
```
daniel-goldstein, line 1128: Looks like this was left over from before the new table.

batch/batch/utils.py
```diff
@@ -45,6 +45,24 @@ async def wrapped(request):
...
 
 
+def add_metadata_to_request(fun):
+    @wraps(fun)
+    async def wrapped(request, *args, **kwargs):
+        request['batch_telemetry'] = {'operation': fun.__name__}
+
+        batch_id = request.match_info.get('batch_id')
+        if batch_id is not None:
+            request['batch_id'] = batch_id
```
daniel-goldstein, line 55: I think this should be `request['batch_telemetry']['batch_id'] = batch_id`. (and same for job id below)

batch/batch/worker/worker.py
```diff
@@ -3005,6 +3008,10 @@ async def create_job_1(self, request):
         assert job_spec['job_id'] == job_id
         id = (batch_id, job_id)
 
+        request['batch_id'] = str(batch_id)
```
daniel-goldstein, line 3011: Same here with `'batch_telemetry'`

batch/sql/estimated-current.sql
```diff
@@ -1218,7 +1226,7 @@ BEGIN
       IF in_update_id != 1 THEN
         SELECT start_job_id INTO cur_update_start_job_id FROM batch_updates WHERE batch_id = in_batch_id AND update_id = in_update_id;
 
-        UPDATE jobs
+        UPDATE jobs, jobs_telemetry
```
daniel-goldstein, line 1229: Interesting, does this not need an additional join?

daniel-goldstein (dismissed):
This looks great, just a few small things that were left over

daniel-goldstein (approved):

PR #13224, opened 2023-06-30
[hailctl] Autocomplete for hailctl config {get,set,unset}

jigold:
To get this to work, I had to run the following command:
```
hailctl --install-completion zsh
```

And then add this line to the end of the ~/.zshrc file: `autoload -Uz compinit && compinit`


hail/python/hailtop/hailctl/config/cli.py
```diff
@@ -42,29 +142,29 @@ def get_section_key_path(parameter: str) -> Tuple[str, str, Tuple[str, ...]]:
...
-        ('email',): (lambda x: re.fullmatch(r'.+@.+', x) is not None, 'should be valid email address'),
-    }
+    validations = {name: validation for name, _, validation in config_variables()}
 
-    validation_func, msg = validations.get(path, (lambda _: True, ''))  # type: ignore
+    if parameter not in validations:
+        print(f"Error: unknown parameter {parameter}", file=sys.stderr)
+        sys.exit(1)
+
+    validation_func, msg = validations.get(parameter, (lambda _: True, ''))  # type: ignore
```
daniel-goldstein, line 167:
```suggestion
    validation_func, msg = validations[parameter]
```

We've now guaranteed that `parameter` will be present in the dictionary.

hail/python/hailtop/hailctl/config/cli.py
```diff
@@ -42,29 +142,29 @@ def get_section_key_path(parameter: str) -> Tuple[str, str, Tuple[str, ...]]:
     sys.exit(1)
 
 
+def complete_config_variable(incomplete: str):
+    for name, help_text, _ in config_variables():
+        if name.startswith(incomplete):
+            completion_item = (name, help_text)
+            yield completion_item
```
daniel-goldstein, line 149: Nit: for brevity I would personally forego the additional binding and just do `yield (name, help_text)`

daniel-goldstein (changes requested):
This is a super cool change! It doesn't have to be in this PR, but can you follow up with a small note in the documentation (if there's no hailctl documentation it can even be just a note in the `Installation` section of the docs) about installing autocomplete and the need to have `compinit` in the zshrc? I'm assuming that if this is not a bug in the CLI tool then it is something that zsh users are just expected to know, so we should link to some zsh documentation describing that.

hail/python/hailtop/hailctl/config/cli.py
```diff
@@ -9,6 +9,7 @@
 import typer
 from typer import Argument as Arg
 
+from hailtop.batch_client.parse import CPU_REGEXPAT, MEMORY_REGEXPAT
```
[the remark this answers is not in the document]
daniel-goldstein: Do you know how shell autocomplete is actually implemented? Is it just repeatedly invoking the `hailctl` executable or is there some long-lived `hailctl` process and it's communicating with the shell? This would be useful information to know for making the autocomplete snappier in the future as it's a bit laggy at the moment.

hail/python/hailtop/hailctl/config/cli.py
```diff
@@ -42,29 +142,29 @@ def get_section_key_path(parameter: str) -> Tuple[str, str, Tuple[str, ...]]:
...
+def complete_config_variable(incomplete: str):
+    for name, help_text, _ in config_variables():
+        if name.startswith(incomplete):
+            completion_item = (name, help_text)
+            yield completion_item
+
+
 @app.command()
-def set(parameter: str, value: str):
+def set(parameter: Ann[str, Arg(default=..., help="Configuration variable to set", autocompletion=complete_config_variable)], value: str):
```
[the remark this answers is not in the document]
jigold: I thought it was necessary to get a required argument from some resource I can't find anymore, but I don't think that's actually true. Removing.

hail/python/hailtop/hailctl/config/cli.py
jigold: Right, I was worried that if I missed some configuration in the code base or something we add later on, then we've basically made this version of hailctl unusable. So basically, my goal was just to make what we have more helpful rather than stricter. So the real question is whether lines 163-165 are correct.

hail/python/hailtop/hailctl/config/cli.py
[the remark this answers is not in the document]
jigold: It's much snappier when we don't import the regex patterns for cpu and memory as well as the RouterAsyncFS.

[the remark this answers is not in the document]
jigold: Let me know what you'd like to do here.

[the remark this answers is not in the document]
daniel-goldstein: So those imports are only needed to execute the validation lambdas, which aren't necessary for autocomplete. If we go with my other suggestion about turning these tuples into classes, and instead of those lambdas have a `validate` method on the classes, I think we should get best of both worlds because those imports can happen inside the `validate` method.

hail/python/hailtop/hailctl/config/cli.py
daniel-goldstein: I'm happy to pose this question to the larger team, but in my opinion it is a bug to accept arbitrary config values not a feature because if I'm allowed to set it I would expect it to do something. If we agree it's a bug, we are within our rights to revoke the ability to set arbitrary config values.

Regarding versions being unusable, that's an interesting question. I guess that would happen if someone does somewhere in the code `configuration_of(some_new_config)` without ever adding validations for it? And that would only make it through our tests if we only ever use the environment variable override instead of using `hailctl config set` which to be honest sounds very plausible. I suppose that for that problem I would propose the solution that `configuration_of(` should not accept some arbitrary string, but a variant of an `Enum` such that we can enforce in a test that every variant of that enum is represented here, such that you  *can't* introduce configuration without also adding support for it in `hailctl`. This would also give a lot more confidence that this PR caught all usages of `configuration_of`.

hail/python/hailtop/hailctl/config/cli.py
```diff
@@ -85,8 +176,30 @@ def set(parameter: str, value: str):
...
+            else:
+                path = f'{section_name}/{item_name}'
+            elements.append((path, value))
+
+    config_items = {config.name: config.help_msg for config in config_variables()}
+
+    for name, current_value in elements:
+        if name.startswith(incomplete):
+            help_msg = config_items.get(name)
+            yield (name, help_msg)
```
daniel-goldstein, line 198: Can this whole function be deleted now and we use `complete_config_variable` in `unset` and `get`?

daniel-goldstein (dismissed):
Left responses in earlier comments.

hail/python/hailtop/hailctl/config/cli.py
jigold: It depends on what we want the behavior to be for unsetting and getting variables that exist already that are not known in the validations. The way it's implemented now is we load all possibilities from the existing configuration. You're proposing to show all known possibilities even if they haven't been set before and don't exist in the configuration.

daniel-goldstein: Ah! I missed that distinction. In that case I'm good with listing what's already in the configuration

hail/python/hailtop/hailctl/config/cli.py
jigold: I forgot to address this comment.

jigold:
@iris-garden Would you be interested in collaborating on this set of changes? I know you were starting to think about environment variables and configuration as well.

jigold:
I haven't tested the refactored code yet -- would like to see if this was the refactoring you had in mind with Enums.

daniel-goldstein (changes requested):
Ya this is definitely what I was getting at. Just a couple small comments

hail/python/hailtop/hailctl/config/cli.py
```diff
@@ -42,29 +41,31 @@ def get_section_key_path(parameter: str) -> Tuple[str, str, Tuple[str, ...]]:
...
+    from hailtop.config import config_variables  # pylint: disable=import-outside-toplevel
+
+    for var, var_info in config_variables.items():
+        if var.value.startswith(incomplete):
+            yield (var.value, var_info.help_msg)
+
+
 @app.command()
-def set(parameter: str, value: str):
+def set(parameter: Ann[str, Arg(help="Configuration variable to set", autocompletion=complete_config_variable)], value: str):
```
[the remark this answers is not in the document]
jigold: I can't figure out how to do the import to be able to use that class annotation without the `from hailtop.config import ConfigVariable` at the top level.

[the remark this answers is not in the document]
daniel-goldstein: Ya it'll have to be at the top level. But we only need that single enum definition, we can put that in its own file that's cheap to import

hail/python/hailtop/hailctl/config/cli.py
```diff
@@ -85,7 +70,47 @@ def set(parameter: str, value: str):
...
+def set(parameter: Ann[ConfigVariable, Arg(help="Configuration variable to set", autocompletion=complete_config_variable)], value: str):
+    if parameter not in config_variables():
+        print(f"Error: unknown parameter {parameter!r}", file=sys.stderr)
+        sys.exit(1)
+
+    section, key, path = get_section_key_path(parameter.value)
+
+    validations = {var: var_info.validation for var, var_info in config_variables().items()}
+
+    validation_func, msg = validations.get(parameter, (lambda _: True, ''))  # type: ignore
```
daniel-goldstein, line 82: We should be able to assert now that `parameter` is in the map. We can make a test that asserts every `ConfigVariable` is in `config_variables` (the `ConfigVariable` class itself is iterable so you can loop over all the variants of the enum).

hail/python/test/hailtop/hailctl/config/test_cli.py
```diff
@@ -20,40 +20,90 @@ def test_config_list_empty_config(runner: CliRunner):
...
 
-    res = runner.invoke(cli.app, ['get', '/a/b/c'], catch_exceptions=False)
+
+# backwards compatibility
+def test_config_get_unknown_names(bc_runner: CliRunner):
+    res = bc_runner.invoke(cli.app, ['get', 'email'], catch_exceptions=False)
+    assert res.exit_code == 0
+    assert res.stdout.strip() == 'johndoe@gmail.com'
+
+    res = bc_runner.invoke(cli.app, ['get', 'batch/foo'], catch_exceptions=False)
```
daniel-goldstein, line 67: This is a good catch and I think we should proceed in this PR as you have it currently. I feel like at some point though we should somehow communicate to users that they have bogus settings in their config. Maybe something we can discuss at team meeting?

daniel-goldstein (dismissed):
This looks great, just a few small nits.

hail/python/test/hailtop/hailctl/config/test_cli.py
jigold: This could be part of a future `hailctl config check` where we also check the remote_tmpdir is correct for the regions and not cold storage.

jigold:
Should be all set except for the question in the comment.

hail/python/hailtop/config/user_config.py
```diff
@@ -69,6 +68,17 @@ def configuration_of(section: str,
...
 
 
+def configuration_of(config_variable: ConfigVariable,
+                     explicit_argument: Optional[T],
+                     fallback: T,
+                     *,
+                     deprecated_envvar: Optional[str] = None) -> Union[str, T]:
+    config_variable_info = config_variables()[config_variable]
+    section = config_variable_info.section
+    option = config_variable_info.option
```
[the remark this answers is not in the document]
daniel-goldstein: > If we went with your suggestion, we'd have to have something like this. Which do you prefer?
> 
> ```python
> if '/' in config_variable.value:
>     section, option = config_variable.value.split('/')
> else:
>     section = 'global'
>     option = config_variable.value
> ```


I would prefer this. If I'm correct it would allow us to delete the repeated `section` and `option` information in `config_variabels`

daniel-goldstein (dismissed):
Responded in comment

daniel-goldstein (approved):
Awesome change. Thank you for bearing with all the Enum changes, I think this will be very robust.

jigold:
A test failed because `hailctl config unset` now returns an error if the config variable does not exist. Let me know if you think we should maintain the current behavior -- otherwise, I slightly modified the tests.

hail/python/hailtop/hailctl/config/cli.py
```diff
@@ -96,10 +116,13 @@ def unset(parameter: str):
         del config[section][key]
         with open(config_file, 'w', encoding='utf-8') as f:
             config.write(f)
+    else:
+        print(f"Error: unknown parameter {parameter!r}", file=sys.stderr)
+        sys.exit(1)
```
daniel-goldstein, line 121: Ah, let's just allow this. I don't see a problem with `unset`ing a value that is not currently set.

daniel-goldstein (dismissed):

jigold:
Sorry - one more thing I need help with. There's a cyclical import with the RouterFS in `variables.py`. Should I just pylint ignore it?

daniel-goldstein:
> Sorry - one more thing I need help with. There's a cyclical import with the RouterFS in `variables.py`. Should I just pylint ignore it?

So it looks like that circular import is because of `config_variables`, but it looks like `config_variables` is only ever used in `hailctl` not elsewhere in `hailtop`:

```
hailctl/config/cli.py
10:from hailtop.config.variables import ConfigVariable, config_variables
46:    for var, var_info in config_variables().items():
56:    if parameter not in config_variables():
62:    config_variable_info = config_variables()[parameter]
86:    from hailtop.config import config_variables, get_user_config  # pylint: disable=import-outside-toplevel
99:    config_items = {var.name: var_info.help_msg for var, var_info in config_variables().items()}

config/__init__.py
4:from .variables import ConfigVariable, config_variables
14:    'config_variables',

config/variables.py
6:_config_variables = None
30:def config_variables():
34:    global _config_variables
36:    if _config_variables is None:
37:        _config_variables = {
112:    return _config_variables
```

Can you move `config_variables` into a file in `hailtop/hailctl/config` and keep the `ConfigVariable` enum in `hailtop/config/variables.py`? Then you can't have a circular reference because `hailtop` can't depend on `hailctl`.

jigold:
I think we're finally good on this. Thanks for the help!

Issue #13046, opened 2023-05-12
ClassCastException when comparing hl.Locus to row key


danking:
### What happened?

Details here: https://discuss.hail.is/t/subset-matrix-table-to-a-medium-sized-list-of-variants/3362/5

```
Java stack trace:
java.lang.ClassCastException: class org.apache.spark.sql.catalyst.expressions.GenericRow cannot be cast to class is.hail.variant.Locus (org.apache.spark.sql.catalyst.expressions.GenericRow is in unnamed module of loader 'app'; is.hail.variant.Locus is in unnamed module of loader org.apache.spark.util.MutableURLClassLoader @62435e70)
        at is.hail.expr.JSONAnnotationImpex$.exportAnnotation(AnnotationImpex.scala:124)
        at is.hail.expr.JSONAnnotationImpex$.$anonfun$exportAnnotation$5(AnnotationImpex.scala:129)
        at is.hail.expr.JSONAnnotationImpex$.$anonfun$exportAnnotation$5$adapted(AnnotationImpex.scala:128)
        at scala.collection.generic.GenTraversableFactory.tabulate(GenTraversableFactory.scala:150)
        at is.hail.expr.JSONAnnotationImpex$.exportAnnotation(AnnotationImpex.scala:128)
        at is.hail.types.virtual.Type.toJSON(Type.scala:184)
        at is.hail.expr.JSONAnnotationImpex$.$anonfun$exportAnnotation$4(AnnotationImpex.scala:125)
        at is.hail.utils.Interval.toJSON(Interval.scala:103)
        at is.hail.expr.JSONAnnotationImpex$.exportAnnotation(AnnotationImpex.scala:125)
        at is.hail.expr.JSONAnnotationImpex$.$anonfun$exportAnnotation$1(AnnotationImpex.scala:113)
        at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
        at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
        at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
        at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:38)
        at scala.collection.TraversableLike.map(TraversableLike.scala:238)
        at scala.collection.TraversableLike.map$(TraversableLike.scala:231)
        at scala.collection.AbstractTraversable.map(Traversable.scala:108)
[...]

PR #13226, opened 2023-07-06
Fix #13045 - Comparison of Structs and Struct Exprs

### LINKED ISSUE #13045: Hail Locus (and other Python data types) in __eq__, etc. should check if the RHS is a Hail Expression and delegate to the expression instead.
### What happened?

See confusing behavior here: 

https://discuss.hail.is/t/subset-matrix-table-to-a-medium-sized-list-of-variants/3362/6?u=danking

```
In [54]: vstruct
Out[54]: Struct(locus=Locus(contig=10, position=60515, reference_genome=GRCh37), alleles=['C', 'T'])

In [55]: (vstruct == minimt.row_key)
Out[55]: False
```

### Version

0.2.105

### Relevant log output

_No response_

ehigham:
RR: https://github.com/hail-is/hail/issues/13045
RR: https://github.com/hail-is/hail/issues/13046 
Support symmetric comparison of structs and struct expressions.
Provide better error messages when attempting to construct literals from expressions with free variables.

### DIFF
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -266,6 +266,16 @@ def literal(x: Any, dtype: Optional[Union[HailType, str]] = None):
                             .format(dtype)) from e
 
     if wrapper['has_expr']:
+        if ( x._ir.free_vars.__len__() > 0 or
+             x._ir.free_agg_vars.__len__() > 0 or
+             x._ir.free_scan_vars.__len__() > 0
+           ):
+            raise ValueError(
+                "'literal' cannot be used with hail expressions that depend "
+                "on other expressions. Use expression 'x' directly "
+                "instead of passing it to 'literal'."
+            )
+
         return literal(hl.eval(to_expr(x, dtype)), dtype)
 
     if x is None or x is pd.NA:
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -238,7 +238,7 @@ def literal(x: Any, dtype: Optional[Union[HailType, str]] = None):
     def typecheck_expr(t, x):
         if isinstance(x, Expression):
             wrapper['has_expr'] = True
-            wrapper['has_free_vars'] = (
+            wrapper['has_free_vars'] |= (
                 builtins.len(x._ir.free_vars) > 0 or
                 builtins.len(x._ir.free_agg_vars) > 0 or
                 builtins.len(x._ir.free_scan_vars) > 0
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -239,9 +239,9 @@ def literal(x: Any, dtype: Optional[Union[HailType, str]] = None):
         if isinstance(x, Expression):
             wrapper['has_expr'] = True
             wrapper['has_free_vars'] = (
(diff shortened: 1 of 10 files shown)


hail/python/hail/utils/struct.py
```diff
@@ -94,7 +94,9 @@ def __str__(self):
         )
 
     def __eq__(self, other):
-        return isinstance(other, Struct) and self._fields == other._fields
+        return self._fields == other._fields \
+            if isinstance(other, Struct) \
+            else NotImplemented
```
ehigham, line 99: See https://github.com/hail-is/hail/issues/13045

ehigham, line 99: See https://docs.python.org/3/reference/datamodel.html#object.__ge__

patrick-schultz (commented):
Thanks Ed, looks great. Would you mind making the same fix for Interval, Locus, and Cal for consistency?

ehigham:
> Thanks Ed, looks great. Would you mind making the same fix for Interval, Locus, and Cal for consistency?

as you wish

patrick-schultz (approved):
thanks!

hail/python/hail/expr/functions.py
```diff
@@ -233,11 +233,17 @@ def literal(x: Any, dtype: Optional[Union[HailType, str]] = None):
...
-    wrapper = {'has_expr': False}
+    wrapper = {'has_expr': False, 'has_free_vars': False}
 
     def typecheck_expr(t, x):
         if isinstance(x, Expression):
             wrapper['has_expr'] = True
+            wrapper['has_free_vars'] = (
+                x._ir.free_vars.__len__() > 0 or
+                x._ir.free_agg_vars.__len__() > 0 or
+                x._ir.free_scan_vars.__len__() > 0
```
patrick-schultz, line 244: This needs to be a `|=`, no?

hail/python/hail/expr/functions.py
```diff
@@ -266,6 +266,16 @@ def typecheck_expr(t, x):
                             .format(dtype)) from e
 
     if wrapper['has_expr']:
+        if ( x._ir.free_vars.__len__() > 0 or
+             x._ir.free_agg_vars.__len__() > 0 or
+             x._ir.free_scan_vars.__len__() > 0
```
patrick-schultz: Better form to use `len(...)`

patrick-schultz (changes requested):

hail/python/hail/expr/functions.py
ehigham: Can use len as it's aliased in the same file :(

ehigham: I could use `buildtins.len`, but not sure what advantage that has

hail/python/hail/expr/functions.py
ehigham: no difference?

patrick-schultz: I mean `wrapper['has_free_vars'] |= ...`. Otherwise, if there are multiple nested expressions, and the last has no free variables but others do, we will only remember the `False` from the last.

ehigham: Ah yes. Mutation. I forgot this can getbcalled more than once

patrick-schultz (approved):

PR #13229, opened 2023-07-07
[compiler] remove TableReader RDD implementations where possible

patrick-schultz:


hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -506,8 +509,20 @@ abstract class TableReader {
...
+      TableStage(
+        globals,
+        RVDPartitioner.empty(ctx, requestedType.keyType),
+        TableStageDependency.none,
+        MakeStream(FastIndexedSeq(), TStream(TStruct.empty)),
+        (_: Ref) => MakeStream(FastIndexedSeq(), TStream(requestedType.rowType)))
+    } else
+      _lower(ctx, requestedType)
+
+  def _lower(ctx: ExecuteContext, requestedType: TableType): TableStage
```
ehigham, line 525: I'm not sure what this change adds other than a template pattern which I generally avoid whenever possible. I preferred keeping this logic in `LowerTableIr`.

How is `_lower` different to `lower`? `_lower` seems like something that shouldn't be public and calling it directly seems like a mistake.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
```diff
@@ -481,7 +481,10 @@ trait TableReaderWithExtraUID extends TableReader {
 abstract class TableReader {
   def pathsUsed: Seq[String]
 
-  def apply(ctx: ExecuteContext, requestedType: TableType, dropRows: Boolean): TableValue
+  def toExecuteIntermediate(ctx: ExecuteContext, requestedType: TableType, dropRows: Boolean): TableExecuteIntermediate = {
+    assert(!dropRows)
+    TableExecuteIntermediate(_lower(ctx, requestedType))
```
ehigham, line 486: I like this interface change. One thing I'd suggest is to remove the `TableExecuteIntermediate` companion object and construct `TableStageIntermediate` and `TableValueIntermediate` directly.

ehigham (dismissed):

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
patrick-schultz: This is a pattern we use occasionally. `_lower` should really be protected, I'll make that change, and it is only called directly in `lower`, which is final. So `_lower` is the customization point for subclasses.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
patrick-schultz: That's not a bad change, but given that it will all be going away when everything is lowered, I don't think there's much benefit.

hail/src/main/scala/is/hail/expr/ir/TableIR.scala
ehigham: What I was trying to understand is why do we need `_lower` at all when `TableRead` seems to be the only place that uses it?

patrick-schultz: Ah, I understand now. That's fair. Addressed.

ehigham (approved):
:shipit: - this is a great change

PR #13235, opened 2023-07-11
[lint] runs black on hail/

iris-garden:
This change also updates the `make pylint-hail` target to include the `hail/python/(hail|test|cluster-tests)` directories, rather than the `hail/` directory.

This is because the latter would require adding `hail/__init__.py` and/or `hail/python/__init__.py` files, which creates issues because some modules in `hail/python/hail` and `hail/python/hailtop` have the same name but are at different paths, which breaks the package resolution for `ruff` and `mypy`.

For this reason, although `black` is run on the files in the `hail/` and `hail/python/` directories, `pylint` cannot be.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,17 +33,22 @@ check-all: check-hail check-services
 
 .PHONY: check-hail-fast
 check-hail-fast:
-	ruff check hail/python/hail
-	ruff check hail/python/hailtop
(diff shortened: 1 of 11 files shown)


iris-garden:
closing in favor of https://github.com/hail-is/hail/pull/14119

PR #13237, opened 2023-07-11
[batch] Fix list batches query and test

jigold:
The test that lists batches timed out. The main problem is the limit in the aioclient used by the test_batch tests was passing a string rather than an integer. I assumed downstream the function was passing an integer. Therefore, we were doing this:

batch_id < "137"
and not batch_id < 137.

So the query was running forever and scanning all batches from the test user.

I also was missing a tag annotation on the queries, but that was not causing the timeout.

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -11,7 +11,7 @@ import signal
 import traceback
 from functools import wraps
 from numbers import Number
-from typing import Dict, Optional, Tuple, Union
+from typing import Awaitable, Callable, Dict, Optional, ParamSpec, Tuple, TypeVar, Union
 
 import aiohttp
 import aiohttp_session
@@ -113,6 +113,10 @@ BATCH_JOB_DEFAULT_STORAGE = os.environ.get('HAIL_BATCH_JOB_DEFAULT_STORAGE', '0G
 BATCH_JOB_DEFAULT_PREEMPTIBLE = True
 
 
+T = TypeVar('T')
+P = ParamSpec('P')
+
+
 def rest_authenticated_developers_or_auth_only(fun):
     @auth.rest_authenticated_users_only
     @wraps(fun)
@@ -189,6 +193,12 @@ def web_billing_project_users_only(redirect=True):
     return wrap
 
 
+def cast_query_param_to_int(param: Optional[str]) -> Optional[int]:
+    if param is not None:
+        return int(param)
+    return None
+
+
 @routes.get('/healthcheck')
 async def get_healthcheck(request):  # pylint: disable=W0613
     return web.Response()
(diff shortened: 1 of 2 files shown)


danking (changes requested):
Can you describe the problem(s)? I'm not sure what this is fixing

jigold:
The test that lists batches timed out. The main problem is the limit in the aioclient used by the test_batch tests was passing a string rather than an integer. I assumed downstream the function was passing an integer. Therefore, we were doing this:

```
batch_id < "137"
```

and not `batch_id < 137`.

So the query was running forever and scanning all batches from the `test` user.

I also was missing a tag annotation on the queries, but that was not causing the timeout.

danking (dismissed):
Can you edit the PR message with your comment so that when the commit merges into main we have a record of why we made this change? Thanks!

jigold:
@danking 

I can keep digging, but it looks like the ParamSpec feature isn't available in the Python version we're running. I'll see if there's a workaround, but I'd really not like to hold up this change anymore than needed.

```
Module "typing" has no attribute "ParamSpec"; maybe "_ParamSpec"?
```

jigold:
It looks like ParamSpec was added in 3.10

danking:
Can we use the `ParamSpec` from `typing_extensions`?

jigold:
This looks like it's going to succeed.

batch/batch/front_end/front_end.py
```diff
@@ -227,17 +240,17 @@ async def _handle_ui_error(session, f, *args, **kwargs):
         raise
 
 
-async def _handle_api_error(f, *args, **kwargs):
+async def _handle_api_error(f: Callable[P, Awaitable[T]], *args: P.args, **kwargs: P.kwargs) -> Optional[T]:
```
danking, line 243: This should be a separate PR, but these new types and all those `assert ... is not None` reveal another issue.

This function conflates two distinct pieces of functionality:
1. Make functions returning no values idempotent
2. Convert `BatchUserError` into an HTTPResponse.

I'm a bit unsure if BatchOperationAlreadyCompletedError is the right thing to use. It seems a bit confusing if I am at the delete billing project form, then press submit, then I lose my WiFi, then refresh the page (clicking OK when it says "do you want to resubmit this form"), then I get "this billing project already is deleted". I guess that tells me that the first request actually did make it through, but as a user, all that matters is that the BP doesn't exist anymore, right? If I mistype the name I'd get a 404, so I can't accidentally think I deleted a different BP.

Regardless, I think we should have `_handle_api_error` which lacks the first except and `_handle_idempotence_and_api_error` which calls `_handle_api_error` wrapped in a try-except.

danking (approved):

danking:
Approved, but see my comment about _handle_api_error.

jigold:
Thank you! I agree these functions need to be rewritten.

PR #13255, opened 2023-07-18
[deps] removes dependency on hurry.filesize

iris-garden:


hail/python/hailtop/utils/filesize.py
```diff
@@ -0,0 +1,21 @@
+from typing import Any
+
+
+_PREFIXES = ["B", "K", "M", "G", "T", "P", "E", "Z", "Y"]
+
+
+def filesize(size: Any) -> str:
```
daniel-goldstein, line 7: This function would blow up on an input that is not coercible to an `int`, so it would be nice to narrow the signature just to those valid inputs. AFAIK, that's unfortunately not expressible in the python type system, but from the usage here it looks like we ever only use this function on `int`s anyway, so I think we should just enforce that the input be `int` and remove the cast.

hail/python/test/hailtop/utils/test_filesize.py
```diff
@@ -0,0 +1,28 @@
+from pytest import raises
+from hailtop.utils.filesize import filesize
+
+def test_filesize():
+    with raises(ValueError):
+        filesize(-1)
+        filesize(-1023)
+        filesize(-1024)
+        filesize(-1025)
+        filesize(-1024**3)
```
daniel-goldstein, line 10: I think only the first invocation is getting tested here. These need to go in separate `with`s.

hail/python/test/hailtop/utils/test_filesize.py
```diff
@@ -0,0 +1,28 @@
...
+    for exp in range(2, 9):
+        assert filesize(1024 ** exp - 1) == f"1023{prefixes[exp - 2]}iB"
+        assert filesize(1024 ** exp) == f"1{prefixes[exp - 1]}iB"
+        assert filesize(1024 ** exp + 1) == f"1{prefixes[exp - 1]}iB"
+        assert filesize(1024 ** exp * 2) == f"2{prefixes[exp - 1]}iB"
+    assert filesize(1024 ** 9 - 1) == "1023YiB"
+    assert filesize(1024 ** 9) == "1024YiB"
+    assert filesize(1024 ** 9 + 1) == "1024YiB"
+    assert filesize(1024 ** 9 * 2) == "2048YiB"
+    assert filesize(1024 ** 10) == "1048576YiB"
```
daniel-goldstein, line 27: I like these tests, and the name yobibyte even more

daniel-goldstein (changes requested):
Love when we can remove a dependency in exchange for 20 lines of code. Just a couple small comments.

hail/python/test/hailtop/utils/test_filesize.py
iris-garden: yobibyte moment

hail/python/test/hailtop/utils/test_filesize.py
iris-garden: good catch! fixed!

hail/python/hailtop/utils/filesize.py
iris-garden: good call, i think i just got caught up in trying to replicate what `hurry.filesize` does haha, fixed!

daniel-goldstein (approved):

PR #13262, opened 2023-07-19
[batch] Add feature switches

jigold:
Let me know if you think this is good and whether I need to test the UI with dev deploy.

### DIFF
--- a/batch/batch/driver/main.py
+++ b/batch/batch/driver/main.py
@@ -38,7 +38,14 @@ from gear.profiling import install_profiler_if_requested
 from hailtop import aiotools, httpx
 from hailtop.config import get_deploy_config
 from hailtop.hail_logging import AccessLogger
-from hailtop.utils import AsyncWorkerPool, Notice, dump_all_stacktraces, flatten, periodically_call, time_msecs
+from hailtop.utils import (
+    AsyncWorkerPool,
+    Notice,
+    dump_all_stacktraces,
+    flatten,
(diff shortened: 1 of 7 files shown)


batch/batch/driver/main.py
```diff
@@ -552,6 +559,27 @@ def validate_int(session, name, value, predicate, description):
...
+async def configure_feature_switches(request, userdata):  # pylint: disable=unused-argument
+    app = request.app
+    db: Database = app['db']
+    post = await request.post()
+
+    compact_billing_tables = 'compact_billing_tables' in post
+
+    await db.execute_update(
+        '''
+UPDATE feature_switches SET compact_billing_tables = %s;
```
danking, line 574: Let's consistently use the same name everywhere. `feature_flags` is the industry standard term for this functionality.

danking (changes requested):

danking:
It's dev only no need to check the UI, we can just do another PR if its broke

batch/batch/driver/templates/index.html
```diff
@@ -25,6 +25,19 @@ <h1>Globals</h1>
     {% endif %}
   </div>
 
+<h1>Feature Switches</h1>
```
danking, line 28:
```suggestion
<h1>Feature Flags</h1>
```

danking (changes requested):

danking (approved):

PR #13263, opened 2023-07-19
[query/vds] Fix local_to_global with missing fill

chrisvittal:
There was a logic error in constructFromIndicesUnsafe, if a missing value was pushed, pushing a present value with the same index would not clear the missing bit.


patrick-schultz (approved):

danking:

PR #13268, opened 2023-07-20
[batch] Fix autoscaler / scheduler deadlock with regions

jigold:
Ben submitted a pipeline where the first 85% of jobs run in us-central1 while the last 15% run in us-east1. The autoscaler only looks at the head of the job queue and then sorts the result set to figure out the regions to spin up instances in. The scheduler looks at the entire job queue and then sorts the result set to figure out the regions to spin up instances in. The sort order placed us-east1 before us-central1. Concretely, the autoscaler is spinning up instances in us-central1 only while the scheduler is trying to schedule jobs in us-east1.

This PR attempts to solve this problem by placing bounds on which jobs the scheduler can look at based on the records the autoscaler actually considered. This is a bit of a hack and I'm worried about the performance implications. On Ben's pipeline with 100K jobs, this will add 0.3 seconds per user considered by the autoscaler. However, the scheduler query got 5x faster with the bounds in place (0.05 seconds vs 0.25 seconds).

```
mysql> EXPLAIN SELECT jobs.job_id, spec, cores_mcpu, regions_bits_rep, time_ready
    -> FROM jobs FORCE INDEX(jobs_batch_id_state_always_run_cancelled)
    -> LEFT JOIN jobs_telemetry ON jobs.batch_id = jobs_telemetry.batch_id AND jobs.job_id = jobs_telemetry.job_id
    -> WHERE jobs.batch_id = BATCH_ID AND (jobs.batch_id < BATCH_ID OR (jobs.batch_id = BATCH_ID AND jobs.job_id <= 15000)) AND inst_coll = "standard" AND jobs.state = 'Ready' AND always_run = 0 AND cancelled = 0
    -> ORDER BY jobs.batch_id, inst_coll, state, always_run, -n_regions DESC, regions_bits_rep, jobs.job_id
    -> LIMIT 300;
+----+-------------+----------------+------------+--------+------------------------------------------+------------------------------------------+---------+-------------------------+-------+----------+----------------------------------------------------+
| id | select_type | table          | partitions | type   | possible_keys                            | key                                      | key_len | ref                     | rows  | filtered | Extra                                              |
+----+-------------+----------------+------------+--------+------------------------------------------+------------------------------------------+---------+-------------------------+-------+----------+----------------------------------------------------+
|  1 | SIMPLE      | jobs           | NULL       | range  | jobs_batch_id_state_always_run_cancelled | jobs_batch_id_state_always_run_cancelled | 136     | NULL                    | 24092 |    10.00 | Using index condition; Using where; Using filesort |
|  1 | SIMPLE      | jobs_telemetry | NULL       | eq_ref | PRIMARY                                  | PRIMARY                                  | 12      | const,batch.jobs.job_id |     1 |   100.00 | NULL                                               |
+----+-------------+----------------+------------+--------+------------------------------------------+------------------------------------------+---------+-------------------------+-------+----------+----------------------------------------------------+
2 rows in set, 1 warning (0.00 sec)
```

I'm open to other ideas for how to fix this. This change is a starting point for how to mitigate the issue in the short term. However, I think a longer term plan needs to be implemented which I already have considered with another design proposal.

### DIFF
--- a/batch/batch/driver/instance_collection/pool.py
+++ b/batch/batch/driver/instance_collection/pool.py
@@ -1,7 +1,7 @@
 import asyncio
 import logging
 import random
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 from typing import Dict, List, Tuple
 
 import prometheus_client as pc
@@ -59,6 +59,9 @@ AUTOSCALER_HEAD_JOB_QUEUE_READY_CORES = pc.Gauge(
     'Number of ready cores per control loop execution calculated from the head of the job queue',
     ['pool_name', 'region'],
 )
+
+
+UserBatchJobRecordCutoff = namedtuple('UserBatchJobRecordCutoff', ['batch_id', 'job_id'])
 
 
 class Pool(InstanceCollection):
@@ -141,6 +144,9 @@ WHERE removed = 0 AND inst_coll = %s;
         self.min_instances = config.min_instances
 
         self.all_supported_regions = self.inst_coll_manager.regions
+
+        self.user_batch_id_job_id_always_run_cutoffs: Dict[str, UserBatchJobRecordCutoff] = {}
+        self.user_batch_id_job_id_cancellable_cutoffs: Dict[str, UserBatchJobRecordCutoff] = {}
 
     @property
     def local_ssd_data_disk(self) -> bool:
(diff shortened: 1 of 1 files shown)


batch/batch/driver/instance_collection/pool.py
```diff
@@ -634,16 +702,28 @@ async def user_runnable_jobs(user):
                     record['format_version'] = batch['format_version']
                     yield record
                 if not batch['cancelled']:
+                    args = [batch['id'], self.pool.name]
+                    last_cancellable_job_id = self.pool.user_batch_id_job_id_cancellable_cutoffs[user]
+                    if last_cancellable_job_id is not None:
+                        bounds = 'AND (jobs.batch_id <= %s OR (jobs.batch_id = %s AND jobs.job_id <= %s))'
```
jigold, line 708: This should probably be `jobs.batch_id < %s`

PR #13275, opened 2023-07-20
Upgrade to Python 3.9

danking:


daniel-goldstein (approved):
LGTM. Something unrelated I noticed though, is the `ksync.md` doc even relevant anymore now that we have `sync.py`?

PR #13277, opened 2023-07-20
[k8s] update to avoid unsupported apiVersions

danking:


ehigham (approved):

danking:
Due to a noisy co-occurring batch from a user, we discovered several of our tests that assumed they could burst into mostly unused cpu. I backed off on their timeouts and also restricted them to their limited number of cores.

PR #13279, opened 2023-07-21
[hailctl] Initialize Hail Environment Interactively

jigold:
I wanted to get feedback on the code I've written thus far before I start testing everything. I'm worried it might be too complicated / brittle to maintain.

I also chose to blow away the entire existing environment rather than resetting the variables with new values. Not sure if that's what we want.


jigold:
In GCP, this code will check remote_tmpdir existence, compatibility with regions, and assign service account permissions. Same with the artifact registry.

jigold:
@daniel-goldstein When you get a chance, I'd appreciate your high level opinion on this addition before I start thoroughly testing everything. Goal is to have something before the ATGU Welcome Workshop September 5th. I know the import locations are an issue in `initialize.py`.

I mainly want feedback on whether the code passes the high level smell test and whether I've over designed this with too many checks and prompts or am forgetting something that would be nice to check for.

daniel-goldstein:
This is super cool! I am a big fan of the idea and the overall approach, particularly when it comes to setting up the tmp bucket and getting the permissions on it correct. Here's my high level thoughts. Sorry for the wall of text but I found these a little hard to articulate.

### Regarding number of prompts

I think this is my primary concern. There's a lot of great automation here, but it's a lot right off the bat. I think what this is aiming to do is make it quick and simple to start running batches and every time someone has to stop and ask someone a question as to how they should respond to some prompt that process gets longer and more complicated. I think it's worth considering what the first batch people should run might be and design for a minimal first experience. IMO, a temp bucket is an absolutely crucial piece of configuration before you can do anything interesting and configuring a temp bucket is something that `hailctl` can easily be very opinionated about. Container registry… I feel like there's harder questions there, and you can run a lot of cool batches before having to worry about provisioning your own. It's also not actually a part of the hailctl config (unless something has changed recently) so it feels a little unusual in this flow. I still think that it is helpful to set people up with an AR and keep them from footguns, but maybe that can go in a separate command that the initial init command points to once you're done? Something along the lines of "if you get to the point where you need to upload custom container images, you can use hailctl to set up a registry"?

Another thing that gives me a little pause is the wording around google projects. I get that you need one to create a bucket, but I think we should just make sure to steer clear of the implication that you are "selecting a GCP project to use for Hail Batch", because that implies some link or ownership that isn't there. But I think there's a quick fix here: for a given resource that we *are* creating for hail use, like the temp bucket, ask for the name first and then ask which project it should be created in, using the projects listed in gcloud as choices with the option to write in your own.

### Regarding number of checks
I think it'd be good to avoid warnings when possible. From looking at this I see a pattern of
1. Ask a leading question
2. Emit a warning if the user selects the alternative option instead of the suggested option

I think I would prefer instead to ask a leading question and in the prompt explain why the alternative option might be undesirable. Then when they make a decision just move on.

On a broader note, I think we should focus on having good documentation and linking to it over having perfectly thorough 
explanations in the CLI. At some point in an interactive setup if it gets longwinded I start spamming enter, but if it was quick and at the end it said something to the effect of: "Your current configuration could result in excess cloud cost. See the documentation <here> about common pitfalls and how to avoid them", I might decide to read through that FAQ with a more discerning eye. This is just my opinion though, I would be curious to see if other folks disagree regarding the UX.

jigold:
I'm thinking that I want to have this:


Basic setup for the domain, remote_tmpdir, billing_project, and only select one region
```
hailctl init
```

Which will prompt for:
1. GCP project
2. Hail domain
3. region
4. remote tmpdir location

And set:
- domain
- batch/billing_project
- batch/regions
- batch/remote_tmpdir
- batch/backend
- query/backend

And then print out a message with the default settings with instructions on how to change any of the settings and warnings about the configuration by using `hailctl config set ...`. I'll have the autocomplete PR merged (hopefully) by then so we can make that nice. The trial billing project is automatically used and we don't prompt for that.

Then we'll have optional flags for components to configure
```
hailctl init --container-registry --requester-pays-buckets --extra-query-settings --extra-batch-settings
```
This would prompt for the same as above as well as advanced regions settings, query settings, and create the container registry.

And lastly
```
hailctl config check
```
which prints warnings about misspecified regions and tmpdirs.

jigold:
Actually, we can just pick a random bucket name and create it automatically and the user doesn't even specify the tmp directory.

daniel-goldstein:
This seems good. I have one objection but I might have just misunderstood the following sentence.

> We can just pick a random bucket name and create it automatically and the user doesn't even specify the tmp directory.

I'm cool with having a default random name as an option but I think there should be a prompt for the bucket name so that the user is aware that we are creating resources on their behalf that will cost them money. I feel like they should always confirm any resource creation that `hailctl` does and the docs should inform them on when they might want to change or delete the resource later on.

jigold:
You're absolutely right. We should pick a default name for the prompt. Actually, I already have the code to do that!

danking (commented):
Cool stuff! A few comments. Seems broadly on the right track.

jigold:
Example with verbose turned on:
Screenshot 2023-07-28 at 1 50 41 PM

Example with mismatching regions:
Screenshot 2023-07-28 at 1 50 26 PM

danking:
Closing to ease CI burden

hail/python/hailtop/hailctl/config/initialize.py
```diff
@@ -0,0 +1,240 @@
...
+
+            remote_tmpdir = Prompt.ask(f'Enter a path to an existing remote temporary directory (ex: gs://my-bucket/batch/tmp)')
+            typer.secho(f'WARNING: You will need to grant read/write access to {remote_tmpdir} for account {hail_identity}', fg=typer.colors.MAGENTA)
+            warnings = True
+
+    config_file = get_user_config_path()
+
+    if not incremental:
+        if os.path.isfile(config_file):
+            os.remove(config_file)
```
daniel-goldstein, line 211: Now that I've gotten this far, I think we should just make a separate command for this. If someone really wants to start fresh they can run `hailctl config clear && hailctl config init`. What do you think?

hail/python/hailtop/hailctl/config/initialize.py
```diff
@@ -0,0 +1,240 @@
...
+
+    set_config('domain', domain)
```
daniel-goldstein, line 213: I think it's a mistake to have the domain in the user config. It's duplicated in the deploy config, not self-explanatory, and it's not something that a user can change without also re-authenticating.

Off the top of my head, I think it would be better if `hailctl auth login hail.is` set the domain in the deploy config and achieved both at once. But regardless of what we end up doing, that's not something that should get too entangled in this PR. I think we should not include setting it in this PR and make a GitHub issue describing how this config variable is duplicated and needs resolution.

daniel-goldstein: LMK if you want me to create the issue.

hail/python/hailtop/hailctl/config/initialize.py
```diff
@@ -0,0 +1,240 @@
...
+
+    if trial_bp_name:
+        set_config('batch/billing_project', trial_bp_name)
+
+    if remote_tmpdir:
+        set_config('batch/remote_tmpdir', remote_tmpdir)
+
+    set_config('batch/regions', region)
+    set_config('batch/backend', 'service')
+    set_config('query/backend', 'batch')
```
daniel-goldstein, line 223: I think this is a bit bold of a move at present. I feel like we should at least prompt the user to ask which backend they want to use for query.

hail/python/hailtop/hailctl/config/initialize.py
```diff
@@ -0,0 +1,240 @@
...
+
+    typer.secho('--------------------', fg=typer.colors.BLUE)
+    typer.secho('FINAL CONFIGURATION:', fg=typer.colors.BLUE)
+    typer.secho('--------------------', fg=typer.colors.BLUE)
+    list_config()
+
+    if warnings:
+        typer.secho('WARNING: Initialized Hail with warnings! The currently specified configuration causes excess costs when using Hail Batch.',
```
daniel-goldstein, line 231:
```suggestion
        typer.secho('WARNING: The currently specified configuration may cause excess costs when using Hail Batch.',
```

daniel-goldstein: I think this should link to the documentation. If we warn about a danger, we should provide a resolution. We should have a page devoted to these cost pitfalls, but for now we could point folks to this warning about [regions](https://hail.is/docs/batch/service.html#submitting-a-batch-to-the-service).

hail/python/hailtop/hailctl/config/utils.py
```diff
@@ -0,0 +1,149 @@
...
+        project = project.strip().decode('utf-8')
+        return project
+    except Exception:
+        return None
+
+
+def get_default_region(supported_regions: List[str], cloud: str) -> Optional[str]:
+    if cloud == 'gcp':
+        if 'us-central1' in supported_regions:
+            return 'us-central1'
```
daniel-goldstein, line 42: Is this supposed to represent the default region in the deploy config? This isn't necessarily true for e.g. the australians. How about another endpoint like the supported regions?

If this doesn't represent the default region of the hail batch service, then I think we should not have a default and force the user to think about which region they really want.

hail/python/hailtop/hailctl/config/utils.py
```diff
@@ -0,0 +1,149 @@
...
+                            labels: Optional[Dict[str, str]],
+                            verbose: bool):
+    from hailtop.utils import CalledProcessError, check_shell  # pylint: disable=import-outside-toplevel
+    if labels:
+        labels_str = ','.join(f'{k}={v}' for k, v in labels.items())
+    else:
+        labels_str = None
+
+    try:
+        await check_shell(f'gcloud --project {project} storage buckets create gs://{bucket} --location={location}', echo=verbose)
```
daniel-goldstein, line 76: To use `check_shell` we need to shell quote, but this isn't really using any shell functionality just the `gcloud` executable, so this should use `check_exec_output`.

hail/python/hailtop/hailctl/config/utils.py
```diff
@@ -0,0 +1,149 @@
...
+        service_account_member = f'serviceAccount:{service_account}'
+        await check_shell(f'gcloud {project} storage buckets add-iam-policy-binding gs://{bucket} --member {service_account_member} --role "roles/storage.objectCreator"',
+                          echo=verbose)
+    except CalledProcessError as e:
+        if 'does not have storage.buckets.getIamPolicy access to the Google Cloud Storage bucket' in e.stderr.decode('utf-8'):
+            msg = f'ERROR: You do not have the necessary permissions to set permissions for bucket {bucket} in project {project}. Ask a project administrator ' \
+                  f'to assign you the StorageIAMAdmin role in Google Cloud Storage or ask them to update the bucket {bucket} on your behalf by giving ' \
+                  f'service account {service_account} the role "roles/storage.objectCreator".'
+            raise InsufficientPermissions(msg) from e
+        raise
```
daniel-goldstein, line 149: We can replace these two functions with one function that also accepts a `role`.

hail/python/hailtop/hailctl/config/utils.py
```diff
@@ -0,0 +1,149 @@
...
+                f.write(json.dumps(lifecycle_policy))
+                f.flush()
+                await check_shell(f'gcloud --project {project} storage buckets update --lifecycle-file="{f.name}" gs://{bucket}', echo=verbose)
+
+        if labels_str:
+            await check_shell(f'gcloud --project {project} storage buckets update --update-labels={labels_str} gs://{bucket}', echo=verbose)
+    except CalledProcessError as e:
+        if 'does not have storage.buckets.get access to the Google Cloud Storage bucket' in e.stderr.decode('utf-8'):
+            msg = f'ERROR: You do not have the necessary permissions to update bucket {bucket} in project {project}. Ask a project administrator ' \
+                  f'to assign you the StorageAdmin role in Google Cloud Storage or ask them to update the bucket {bucket} on your behalf.'
```
daniel-goldstein, line 105: Users only need that role on that particular bucket, not project-wide.

hail/python/hailtop/hailctl/config/utils.py
```diff
@@ -0,0 +1,149 @@
...
+        labels_str = ','.join(f'{k}={v}' for k, v in labels.items())
+    else:
+        labels_str = None
+
+    try:
+        await check_shell(f'gcloud --project {project} storage buckets create gs://{bucket} --location={location}', echo=verbose)
+    except CalledProcessError as e:
+        if 'does not have storage.buckets.create access to the Google Cloud project' in e.stderr.decode('utf-8'):
+            msg = f'ERROR: You do not have the necessary permissions to create buckets in project {project}. Ask a project administrator ' \
+                  f'to assign you the StorageAdmin role in Google Cloud Storage or ask them to create the bucket {bucket} on your behalf.'
```
daniel-goldstein, line 80: Can we replace this with a narrower role that allows users to create buckets?

daniel-goldstein (changes requested):
I like this level of granularity in terms of the initial configuration. Added some more detailed comments, mostly around managing user expectations.

hail/python/hailtop/hailctl/config/utils.py
jigold: I find writing commands for `check_exec_output` to be significantly harder to read.

hail/python/hailtop/hailctl/config/initialize.py
jigold: I made a separate command, but I do think `hailctl config init` should always clear any existing environment. Otherwise, there might be parameters set that aren't intended.

hail/python/hailtop/hailctl/config/utils.py
daniel-goldstein: Fair point, but I think using the shell when we don't have to is a footgun because we could forget to properly quote inputs and shell invocations end up in bash history files (so we have to be careful not to `check_shell` with anything sensitive). What if we just made a wrapper for `check_exec_output` that takes a single string, splits on spaces, and passes it to `check_exec_output`?

hail/python/hailtop/hailctl/config/initialize.py
jigold: I cannot figure out how this is all supposed to work. The deploy config reads the domain value from the `get_user_config()` output if no domain is specified or the environment variable. But running `hailctl auth login hail.is` doesn't take the domain right now. It only takes the namespace.

Can you be more specific on what changes you want in this PR?

hail/python/hailtop/hailctl/config/utils.py
jigold: I don't like this option. It's too hard to get the splitting to be correct. I'll just make it check exec output.

jigold:
I think I addressed most of the comments. I haven't tested the new code -- I don't want to do that until we're happy with it. I don't know that I like how this is turning out. I think we're conflating what `hailctl config init` should be which is intitializing an environment configuration file versus a quick start to using Hail Batch and QoB. My intention for this feature was to idiot proof the latter especially for the ATGU workshop and make it as few commands as possible. I worry that needing to run `hailctl auth login` before this is not a just run this single command and then you can get going with little effort. I don't think `hailctl config init` is the right place for what I have written. Maybe it should be `hailctl batch quick-start`???

hail/python/hailtop/hailctl/config/initialize.py
daniel-goldstein: My apologies, you're right `hailctl auth login hail.is` is just not a thing. My main point here is that I think having `domain` in the user config is a mistake, and I would like not to pollute users configs with this if we are going to change it down the line. I'll make a separate issue detailing my concerns around `domain`, but for this PR I think we should:

1. Not set the domain anywhere. Most (or basically all?) users never set this and it defaults to `hail.is`, so let's just leave that out for now
2. Expect that before this command the user has already logged in.

danking: I've long been mystified by users who had `global/config` set, but now I see why. I apparently even approved the change back in [December 2021](https://github.com/hail-is/hail/pull/11113)! I agree with you Daniel, seems odd to store that in two places.

That said, I think this conversation is entirely independent of this PR. If we want to configure the domain in our setup command, that seems entirely reasonable, but let's mutate the deploy_config file, not the user config.

danking:
Agreed that `hailctl init` is maybe the wrong name b/c `hailctl` also controls Dataproc. `hailctl batch init` seems right to me. I agree that `hailctl config init` seems wrong if we're creating buckets.

On the issue of login, I think we can appease both points of view. If you're not logged in, have `hailctl batch init` do this:
```
You are not currently logged in to Hail. Redirecting you to a login screen.

... user does login flow ...

In the future, you can use hailctl auth login to login to Hail.

... continue with hailctl batch init ...
```

I think we should punt on addressing domain. Broad users don't need to interact with it at all.

What are the other concerns about how this is developing?

danking:
(and to be clear, I'm onboard with the idea of a single command which gets you from zero to simple batch pipelines; jury is out on the Artifact Registry. I think making sure that's configured correctly might be better upstreamed to Sam. Normal users might not have permission to enable/disable things like that.).

daniel-goldstein:
Hm, I guess I just don't personally see the value in one command instead of two. If the workshop is meant to teach users how to use batch, and one aspect of using batch is knowing how to log in, shouldn't they go through the motions?

danking:
That's a fair counter-point, but wouldn't it be annoying if you did:
```
hailctl batch init
```
And it was like:
```
Run hailctl auth login first.
```
Why not just *do* it for me?

daniel-goldstein:
I suppose this is where I started getting entangled with the domain issue. If the Australians run a workshop, what should happen if their users run `hailctl batch init`? Should they have to supply some additional argument so that _if_ they're not authenticated they get sent somewhere other than `hail.is`? Maybe that's ok, seems kind of awkward though.

While it doesn't work this way today, I imagine that we should ultimately configure `hailctl auth login` to accept a domain. I feel like that would make the AUS scenario slightly less awkward, and the tool more consistent, though I admit it is conceding some of our own convenience.

daniel-goldstein:
Heck not even the Australians, what about Azure?

danking:
I agree that we never carefully considered switching domains at the user-level. I'm mostly concerned with Broadies using the GCP instance for now. I think how to deal with domains really gets into the idea of having config profiles, which just makes all of this more complicated. Ergo, my gut is: if you're not a Broadie using Google, you have to do some extra work, but that set of users is very small.

danking:
But let's solve for Aus and Azure as a separate thread of work that doesn't block with PR.

I imagine whatever we decide will either make this command need to take a domain or enhance this command to create a profile per-domain.

jigold:
Could you please do one more broad look over this before I start testing again with the new changes?

daniel-goldstein:
This looks great, I love how straightforward `async_basic_initialize` is

hail/python/hailtop/hailctl/batch/initialize.py
```diff
@@ -0,0 +1,258 @@
...
+    try:
+        bucket_info = await get_gcp_bucket_information(bucket, verbose)
+    except InsufficientPermissions as e:
+        typer.secho(e.message, fg=typer.colors.RED)
+        raise Abort()
+
+    location = bucket_info['location'].lower()
+
+    if bucket_info['locationType'] != 'region':
+        typer.secho(f'WARNING: remote temporary directory {remote_tmpdir} is multi-regional. Using this bucket with the Batch Service will incur addtional ingress and egress fees.',
```
danking, line 34: You mention ingress fees in a few spots. AFAIK, there are no ingress fees for any network traffic in GCP. What's your thinking on mentioning ingress here?

danking (changes requested):

hail/python/hailtop/hailctl/batch/initialize.py
jigold: Aren't there inter-region network fees for reading data from us-east bucket into us-central?

hail/python/hailtop/hailctl/batch/cli.py
```diff
@@ -162,3 +163,11 @@ def submit(
 ):
     '''Submit a batch with a single job that runs SCRIPT with the arguments ARGUMENTS.'''
     asyncio.run(_submit.submit(name, image_name, files or [], output, script, arguments or []))
+
+
+@app.command('init', help='Initialize a Hail Batch environment.')
+def initialize(
+        domain: Ann[Optional[str], Opt('--domain', '-d', help='Domain to login to.')] = None,
```
daniel-goldstein, line 170: I thought we decided to punt on configuring `domain` in this PR and address that later?

hail/python/hailtop/hailctl/batch/initialize.py
daniel-goldstein: I thought those fees were interpreted as egress out of `us-east`, not sure though.

hail/python/hailtop/hailctl/batch/initialize.py
```diff
@@ -0,0 +1,252 @@
...
+
+    gcloud_installed = await check_for_gcloud()
+    if not gcloud_installed:
+        typer.secho('Have you installed gcloud? For directions see https://cloud.google.com/sdk/docs/install '
+                    'To log into gcloud run:\n'
+                    '> gcloud auth application-default login',
+                    fg=typer.colors.RED)
+        raise Abort()
+
+    create_remote_tmpdir = Confirm.ask('Do you want to create a new bucket in project for temporary files generated by Hail?')
```
daniel-goldstein, line 173:
```suggestion
    create_remote_tmpdir = Confirm.ask('Do you want to create a new bucket for temporary files generated by Hail?')
```

hail/python/hailtop/hailctl/batch/initialize.py
```diff
@@ -0,0 +1,252 @@
...
+    batch_client = await BatchClient.create(trial_bp_name)
+
+    async with batch_client:
+        cloud = await batch_client.cloud()
+        supported_regions = await batch_client.supported_regions()
+
+        if cloud == 'gcp':
+            remote_tmpdir, tmpdir_region, warnings = await initialize_gcp(username, hail_identity, supported_regions, verbose)
+        else:
+            remote_tmpdir = Prompt.ask('Enter a path to an existing remote temporary directory (ex: gs://my-bucket/batch/tmp)')
```
daniel-goldstein, line 218: This else block seems to assume azure, so let's use an azure dummy url instead of a GCS one

hail/python/hailtop/hailctl/batch/initialize.py
```diff
@@ -0,0 +1,252 @@
...
+    query_backend = Prompt.ask('Which backend do you want to use for Hail Query?', choices=['spark', 'batch', 'local'])
+    set_config(ConfigVariable.QUERY_BACKEND, query_backend)
+
+    typer.secho('--------------------', fg=typer.colors.BLUE)
+    typer.secho('FINAL CONFIGURATION:', fg=typer.colors.BLUE)
+    typer.secho('--------------------', fg=typer.colors.BLUE)
+    list_config()
+
+    if warnings:
+        typer.secho('WARNING: Initialized Hail with warnings! The currently specified configuration will result in additional ingress and egress fees when using Hail Batch.',
```
daniel-goldstein, line 250:
```suggestion
        typer.secho('WARNING: The currently specified configuration will result in additional ingress and egress fees when using Hail Batch.',
```

I don't think the first sentence is very necessary, and I'm not quite sure how to interpret it.

hail/python/hailtop/hailctl/config/cli.py
```diff
@@ -49,7 +49,7 @@ def complete_config_variable(incomplete: str):
             yield (var.value, var_info.help_msg)
 
 
-@app.command()
+@app.command(help='Set the value of a configuration parameter.')
```
daniel-goldstein, line 52: How come you needed to add these help texts? Is it not using the function's docstring?

daniel-goldstein (changes requested):
Looks good, just a few small nits and some confusion about where we ended up with this domain stuff.

hail/python/hailtop/hailctl/config/cli.py
jigold: Ah. Didn't realize that was what was happening. I removed them.

daniel-goldstein (approved):

PR #13281, opened 2023-07-21
[config] Add max_polling_delay environment variable for wait ops

jigold:


jigold:
I'm disappointed this PR didn't help a lot. Can easily reopen if we think this is a generally useful feature to have.

PR #13289, opened 2023-07-24
[batch/test] Wait for job to be running in list_jobs_v2 test

jigold:
EDIT: We changed the behavior of wait in this PR to use the same backoff we use in the rest of the code base.

Context: The list_jobs_v2 test failed in production because SQL treats NULL values as False and those records don't end up in the result set. For example, a ready job will not be in the result set for `start_time != 1970-01-01`. So we need to ensure the job is actually running before executing the tests.

### DIFF
--- a/batch/test/test_batch.py
+++ b/batch/test/test_batch.py
@@ -604,6 +604,12 @@ def test_list_jobs_v2(client: BatchClient):
         j_failure.wait()
         j_error.wait()
 
+        delay = 0.1
+        while True:
+            if j_running.status()['state'] == 'Running' or j_running.is_complete():
+                break
+            delay = sync_sleep_and_backoff(delay)
+
         assert_job_ids({j_success.job_id}, 'state = success')
         assert_job_ids({j_success.job_id}, 'state == success')
         assert_job_ids({j_success.job_id}, 'state=success')
--- a/batch/test/test_batch.py
+++ b/batch/test/test_batch.py
@@ -13,7 +13,7 @@ from hailtop.batch.backend import HAIL_GENETICS_HAILTOP_IMAGE
 from hailtop.batch_client.client import BatchClient
 from hailtop.config import get_deploy_config, get_user_config
 from hailtop.test_utils import skip_in_azure
-from hailtop.utils import external_requests_client_session, retry_response_returning_functions
+from hailtop.utils import external_requests_client_session, retry_response_returning_functions, sync_sleep_and_backoff
 from hailtop.utils.rich_progress_bar import BatchProgressBar
 
 from .failure_injecting_client_session import FailureInjectingClientSession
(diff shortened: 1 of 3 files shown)


batch/test/test_batch.py
```diff
@@ -604,6 +604,12 @@ def assert_job_ids(expected, q=None):
         j_failure.wait()
         j_error.wait()
 
+        delay = 0.1
+        while True:
+            if j_running.status()['state'] == 'Running' or j_running.is_complete():
+                break
+            delay = sync_sleep_and_backoff(delay)
```
daniel-goldstein, line 611: Can we pull this logic out into a function in the test utils? I see similar logic in `test_running_job_log_and_status`. That one should actually get updated to use this logic since it's a hot loop that probably causes a spike in requests.

batch/test/test_batch.py
```diff
@@ -604,6 +604,12 @@ def assert_job_ids(expected, q=None):
         j_failure.wait()
         j_error.wait()
 
+        delay = 0.1
+        while True:
+            if j_running.status()['state'] == 'Running' or j_running.is_complete():
```
daniel-goldstein, line 609:
```suggestion
            if j_running.is_running() or j_running.is_complete():
```

daniel-goldstein (changes requested):

jigold:
I might have made this more complicated....

batch/test/test_batch.py
```diff
@@ -1583,13 +1576,9 @@ def test_update_with_always_run(client: BatchClient):
...
 
-    delay = 0.1
-    while True:
-        if j1.is_complete():
-            assert False, str(j1.status(), b.debug_info())
-        if j1.is_running():
-            break
-        delay = sync_sleep_and_backoff(delay)
+    wait_status = j1._wait_for_states('Running')
+    if wait_status.prev_completed:
```
daniel-goldstein, line 1580: Can a job be previously complete but not currently complete?

daniel-goldstein (changes requested):
Good to deduplicate all the `while True` loops, just a couple questions.

batch/test/test_batch.py
jigold: No. This is to distinguish between whether the wait returned because the job was Running or because it had already completed (avoid infinite loops). I don't love this interface, but if you think it's an improvement, then let's go with it. Maybe `prev_completed` is a bad name.

daniel-goldstein: Ah ok I see. Instead of adding the extra `prev_completed` field then, can we look at the status dict to see if the state is complete? My initial comment was motivated more by deduplicating the loop-delay logic.

daniel-goldstein (approved):

danking:

PR #13306, opened 2023-07-26
[batch] Turn off oms_agent in test and dev

jigold:


daniel-goldstein (approved):

danking:
I retried Azure. It ran into the exact same problem with Azure Blob Storage flakiness https://ci.azure.hail.is/batches/4126780/jobs/136 .

We gotta fix our interaction with Azure Blob Storage, it's the only way to get CI to be stable again. It's possible https://github.com/hail-is/hail/pull/13325 will help, but it doesn't address the root cause of the SDK giving us these mysterious errors.

danking:

PR #13308, opened 2023-07-26
[gear] Add typing to the transaction decorator

daniel-goldstein:
Adding this signature appeases `pyright` in my editor. Unfortunately, `pylint` still [gets confused](https://github.com/pylint-dev/pylint/issues/259) so I disabled this particular check. I made sure that `mypy` still errors when forgetting to provide a required parameter to a function call.

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1170,7 +1170,12 @@ VALUES (%s, %s, %s);
         # must rollback. See https://github.com/hail-is/hail-production-issues/issues/9
         await asyncio.gather(write_spec_to_cloud(), insert_jobs_into_db(tx))
 
-    await write_and_insert()  # pylint: disable=no-value-for-parameter
+    await write_and_insert()
+
+    def foo(x):
+        return x
+
+    foo()
 
     return web.Response()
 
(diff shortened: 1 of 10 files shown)


batch/batch/front_end/front_end.py
```diff
@@ -1170,7 +1170,12 @@ async def write_and_insert(tx):
         # must rollback. See https://github.com/hail-is/hail-production-issues/issues/9
         await asyncio.gather(write_spec_to_cloud(), insert_jobs_into_db(tx))
 
-    await write_and_insert()  # pylint: disable=no-value-for-parameter
+    await write_and_insert()
+
+    def foo(x):
+        return x
+
+    foo()
```
danking, line 1178: Remove

gear/gear/database.py
```diff
@@ -59,13 +59,16 @@ async def wrapper(*args, **kwargs):
     return wrapper
 
 
-def transaction(db, **transaction_kwargs):
-    def transformer(fun):
+T = TypeVar("T")
+
+
+def transaction(db: 'Database', read_only: bool = False):
```
danking, line 65: Do you not need a return type here?

danking (changes requested):

danking:
Is there a way to add a comment to our ignores? It’s be good to note that we rely on mypy for this one

danking (approved):

PR #13317, opened 2023-07-27
[compiler] incremental hashing based on Threefry

patrick-schultz:


hail/src/main/scala/is/hail/expr/ir/Random.scala
```diff
@@ -269,6 +269,40 @@ object Threefry {
   }
 }
 
+class PMacHash() {
```
ehigham, line 272: This code looks very similar to what's above. Do you think it's worth refactoring `pmacHash` above to use your new code?

patrick-schultz: I think not in this case. This is low-level, performance sensitive code, and I don't see how to refactor without adding overhead. And besides, for hash functions, the only way to test it is to check that different implementations produce the same result, so I don't mind having it implemented two or three ways, and testing them against each other.

ehigham (approved):

PR #13319, opened 2023-07-27
[security] delete after 5 days, ignore system managed, consider two special secrets

danking:
A job running for five days *could* cause an issue here, so the operator should still be cognizant of that. The
move to access tokens will eliminate that concern anyway.

We should not consider system-managed keys because those are managed by Google and not even visible in the
UI. We are not obligated to rotate them (nor are we able, afaik).

There are two special secrets which do not conform to the `key.json` naming scheme for their
secret files. I added them as special cases.

### DIFF
--- a/devbin/rotate_keys.py
+++ b/devbin/rotate_keys.py
@@ -29,7 +29,7 @@ ENDC = '\033[0m'
 class ServiceAccountDict(TypedDict):
     name: str
     projectId: str
-    uniqueId:  str
+    uniqueId: str
     email: str
     displayName: str
     etag: str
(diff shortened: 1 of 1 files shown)


danking:
Take a look at https://github.com/hail-is/hail/pull/13319/commits/540f1486c9286f599670483f65d4796c4b39168f to ignore the black changes.

daniel-goldstein (approved):

danking:
Closing to make room for other PRs.

danking:
Closing to ease CI burden

PR #13320, opened 2023-07-27
[tests] test_skat should have default timeout of 10mins

jigold:
These tests were timing out consistently now. I think 3 minutes is too strict given what we know about the performance of our current system.

### DIFF
--- a/hail/python/test/hail/methods/test_skat.py
+++ b/hail/python/test/hail/methods/test_skat.py
@@ -510,7 +510,6 @@ def skat_dataset():
     return ds
 
 
-@test_timeout(3 * 60)
 def test_skat_1():
     ds = skat_dataset()
     hl.skat(key_expr=ds.gene,
@@ -521,7 +520,6 @@ def test_skat_1():
             logistic=False)._force_count()
 
 
-@test_timeout(3 * 60)
 def test_skat_2():
     ds = skat_dataset()
     hl.skat(key_expr=ds.gene,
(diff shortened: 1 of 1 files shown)


chrisvittal (approved):

hail/python/test/hail/methods/test_skat.py
```diff
@@ -562,7 +560,7 @@ def test_skat_5():
             logistic=(25, 1e-6))._force_count()
 
 
-@test_timeout(local=4 * 60)
```
danking, line 565: You have to preserve the spark and local timeouts for each of these methods

danking (changes requested):

hail/python/test/hail/methods/test_skat.py
danking: n.b. [default timeout is currently 2 minutes for spark](https://github.com/hail-is/hail/blob/main/build.yaml#L1178)

danking (dismissed):
I think this is actually a bug in the compiler. See discussion in Zulip: https://hail.zulipchat.com/#narrow/stream/123011-Hail-Query-Dev/topic/aggregating.20over.20samples.20triggers.2016-way.20parallelism.3F

Due to bad compilation, we're doing a distributed operation just to count the number of samples.

jigold:
do we still want to merge this?

danking:
Let's revisit after we put in the other mitigations to the underlying issues. In particular, lets let the new exponential back off algorithm land and then assess if we're seeing skat issues in GCP still.

PR #13322, opened 2023-07-27
[batch] Stream job logs in the batch front-end

### LINKED ISSUE #13328: Batch front-end pods cannot handle large job logs
### What happened?

When a job creates a log file in excess of about half a GB, loading the job page can cause the batch front-end pod to crash as it loads the log file into memory and interpolates it directly into the job page. The front-end should instead:

- Fully stream job logs in the log endpoint
- Show a truncated view of the log in the job page, with a pointer to download the full log if it's truncated.

### Version

0.2.120

### Relevant log output

_No response_

daniel-goldstein:
Fixes #13328. See that issue for context.

See [here](https://internal.hail.is/dgoldste/batch/batches/5/jobs/1) for an example of a job log that would currently OOM production pods.


batch/batch/front_end/front_end.py
```diff
@@ -609,8 +650,15 @@ async def get_job_container_log(request, batch_id):
     containers = job_tasks_from_spec(record)
     if container not in containers:
         raise web.HTTPBadRequest(reason=f'unknown container {container}')
-    job_log = await _get_job_container_log(app, batch_id, job_id, container, record)
-    return web.Response(body=job_log)
+    container_log = await _get_job_container_log(app, batch_id, job_id, container, record)
+
+    resp = web.StreamResponse()
+    await resp.prepare(request)
+    if isinstance(container_log, ReadableStream):
```
danking, line 657:
```suggestion
    if container_log is not None:
```
right?

batch/batch/front_end/front_end.py
```diff
@@ -2139,12 +2187,21 @@ async def ui_get_job(request, userdata, batch_id):
...
+    possibly_truncated_logs = set()
+    for container, log_stream in job_log_streams.items():
+        if log_stream is None:
+            job_log_strings_or_bytes[container] = None
+        else:
+            fifty_kib = 50 * 1024
+            async with log_stream:
+                log = await log_stream.read(fifty_kib)
+            if len(log) == fifty_kib:
+                possibly_truncated_logs.add(container)
```
danking, line 2200:
```suggestion
            fifty_kib = 50 * 1024
            async with log_stream:
                log = await log_stream.read(fifty_kib + 1)
            if len(log) == fifty_kib + 1:
                possibly_truncated_logs.add(container)
            log = log[:fifty_kib]
```

Let's give the user confidence.

batch/batch/front_end/templates/job.html
```diff
@@ -137,7 +137,14 @@ <h2>Logs</h2>
...
 <h3>Input</h3>
 {% if 'input' in job_log %}
-<h4>Log <a href="{{ base_path }}/batches/{{ batch_id }}/jobs/{{ job_id }}/log/input" download="batch-{{ batch_id }}-{{ job_id }}-input.log"><i class="fa fa-download" aria-hidden="true"></i></a></h4>
+<h4>Log
+    <a href="{{ base_path }}/batches/{{ batch_id }}/jobs/{{ job_id }}/log/input" download="batch-{{ batch_id }}-{{ job_id }}-input.log">
+        <i class="fa fa-download" aria-hidden="true"></i>
+    </a>
+</h4>
+{% if 'input' in possibly_truncated_logs %}
+<b>To see the full log, click the download button above</b>
```
danking, line 146: Don't make them go find the link.
```
<b>Log too long to display. <a href=...>Download the full log</a>.</b>
```

batch/batch/front_end/front_end.py
```diff
@@ -589,11 +626,15 @@ async def _get_full_job_status(app, record):
 @add_metadata_to_request
 async def get_job_log(request, userdata, batch_id):  # pylint: disable=unused-argument
     job_id = int(request.match_info['job_id'])
-    job_log_bytes = await _get_job_log(request.app, batch_id, job_id)
+    job_log_streams = await _get_job_log(request.app, batch_id, job_id)
     job_log_strings: Dict[str, Optional[str]] = {}
-    for container, log in job_log_bytes.items():
+    for container, log in job_log_streams.items():
```
danking, line 631: I dislike that we don't close these streams. Now that we don't read them in parallel, I'm not sure `_get_job_log` is pulling its weight. Can we refactor to something like `get_job_containers` which gives us the containers and then we can do:
```python3
for container in containers:
    maybe_log = await _get_job_container_log(app, batch_id, job_id, container, record)
    if maybe_log is not None:
        async with maybe_log as log_stream:
           ...
    else:
        ...
```

batch/batch/front_end/front_end.py
```diff
@@ -2139,12 +2187,21 @@ async def ui_get_job(request, userdata, batch_id):
...
     # str or else Jinja will present them surrounded by b''
-    job_log_strings_or_bytes = {}
-    for container, log in job_log_bytes.items():
-        try:
-            job_log_strings_or_bytes[container] = log.decode('utf-8') if log is not None else None
-        except UnicodeDecodeError:
-            job_log_strings_or_bytes[container] = log
+    job_log_strings_or_bytes: Dict[str, Union[str, bytes, None]] = {}
+    possibly_truncated_logs = set()
+    for container, log_stream in job_log_streams.items():
```
danking, line 2192: Same argument here. We're not loading the streams in parallel anymore so stuffing them all in that gather above is less compelling.

batch/batch/front_end/front_end.py
danking: Or: keep `_get_job_log` but make it `_get_truncated_job_logs` which returns 50kB plus a "has more" bit.

danking (changes requested):
This looks like exactly the right solution for the front end, just some nits about code organization.

batch/batch/front_end/front_end.py
danking: Or also see above

batch/batch/front_end/front_end.py
daniel-goldstein: Oop ya, this is cruft from a prior implementation where I was sometimes also returning `bytes`.

batch/batch/front_end/front_end.py
daniel-goldstein: Tried going the latter route

batch/batch/front_end/front_end.py
```diff
@@ -611,8 +667,16 @@ async def get_job_container_log(request, batch_id):
...
         raise web.HTTPBadRequest(reason=f'unknown container {container}')
-    job_log = await _get_job_container_log(app, batch_id, job_id, container, record)
-    return web.Response(body=job_log)
+    container_log = await _get_job_container_log(app, batch_id, job_id, container, record)
+
+    resp = web.StreamResponse()
+    await resp.prepare(request)
+    if container_log is not None:
+        async with container_log:
+            while b := await container_log.read(1024):
```
danking, line 676: I feel like 1KiB is a bit small, but we can fiddle with these parameters in future PRs.

batch/batch/front_end/front_end.py
```diff
@@ -2166,6 +2237,7 @@ async def ui_get_job(request, userdata, batch_id):
             resource_usage, memory_limit_bytes, io_storage_limit_bytes, non_io_storage_limit_bytes
         ),
         'has_jvm_profile': has_jvm_profile,
+        'possibly_truncated_logs': possibly_truncated_logs,
```
danking, line 2240: Sorry to pick nits but this is now `truncated_logs`, not possibly.

danking (changes requested):

danking (approved):
excellent, really good change.

PR #13331, opened 2023-07-28
[ci] Add easier way to add all production developers to a dev namespace

daniel-goldstein:
I broke my previous PR into littler pieces some of which has already merged. This PR adds developers from production to test namespaces (only programmatic access, browser OAuth flow does not work yet for test namespaces) and makes it easier to add developers to dev namespaces.

### DIFF
--- a/auth/auth/auth.py
+++ b/auth/auth/auth.py
@@ -427,14 +427,6 @@ async def create_user(request: web.Request, _) -> web.Response:
     hail_credentials_secret_name = body.get('hail_credentials_secret_name')
     if (hail_identity or hail_credentials_secret_name) and not is_test_deployment:
         raise web.HTTPBadRequest(text='Cannot specify an existing hail identity for a new user')
-    if hail_credentials_secret_name:
-        try:
-            k8s_cache: K8sCache = request.app['k8s_cache']
-            await k8s_cache.read_secret(hail_credentials_secret_name, DEFAULT_NAMESPACE)
-        except kubernetes_asyncio.client.rest.ApiException as e:
-            raise web.HTTPBadRequest(
-                text=f'hail credentials secret name specified but was not found in namespace {DEFAULT_NAMESPACE}: {hail_credentials_secret_name}'
-            ) from e
 
     try:
         await insert_new_user(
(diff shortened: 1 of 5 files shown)


auth/auth/auth.py
```diff
@@ -393,14 +393,6 @@ async def create_user(request: web.Request, userdata):  # pylint: disable=unused
...
     if (hail_identity or hail_credentials_secret_name) and not is_test_deployment:
         raise web.HTTPBadRequest(text='Cannot specify an existing hail identity for a new user')
-    if hail_credentials_secret_name:
-        try:
-            k8s_cache: K8sCache = request.app['k8s_cache']
-            await k8s_cache.read_secret(hail_credentials_secret_name, DEFAULT_NAMESPACE)
-        except kubernetes_asyncio.client.rest.ApiException as e:
-            raise web.HTTPBadRequest(
-                text=f'hail credentials secret name specified but was not found in namespace {DEFAULT_NAMESPACE}: {hail_credentials_secret_name}'
-            ) from e
```
daniel-goldstein, line 403: This was initially put in as a safeguard, because a user cannot submit jobs to the namespace if their gsa key secret does not exist in that namespace, but I feel like this hinders more than helps for the case when you just want other devs to be able to look at a batch in your namespace. This will also not be relevant once we phase out keys.

danking:
Closing to ease CI burden

auth/auth/auth.py
```diff
@@ -553,8 +545,15 @@ async def post_create_user(request, userdata):  # pylint: disable=unused-argumen
...
-@auth.rest_authenticated_developers_only
-async def rest_get_users(request, userdata):  # pylint: disable=unused-argument
+async def rest_get_users(request):
+    session_id = await get_session_id(request)
+    if not session_id:
+        raise web.HTTPUnauthorized()
+    userdata = await get_userinfo(request, session_id)
+    is_developer_or_sa = userdata is not None and (userdata['is_developer'] == 1 or userdata['is_service_account'] == 1)
+    if not is_developer_or_sa:
+        raise web.HTTPUnauthorized()
```
danking, line 555: Elsewhere we special case CI, can we continue to do that here? I don't want, e.g., my service account to be able to hit this endpoint.

danking (changes requested):

auth/auth/auth.py
daniel-goldstein: Fair point, I've restricted it further to just CI. For future reference though, `is_service_account` doesn't mean it could be a user service account, rather that the account is one of the accounts for our services, like `ci`, `batch`, `grafana`, etc.

danking: ah, OK, still better though.

danking (approved):

PR #13333, opened 2023-07-28
[query] allow inteval filters to work with indirect row key

patrick-schultz:
E.g. when the row key, and/or the row itself, in a predicate is hidden behind Refs, MakeStructs, SelectFields. For example, before matching the first key field "kf" only worked for exactly `GetField(Ref("row", ...), "kf")`. Now the following all work (meaning the analysis recognizes these all as being the first key field):
```
Let("foo",
  SelectFields(Ref("row"), [..., "kf", ...]),
  GetField(Ref("foo"), "kf"))

Let("foo",
  Ref("row"),
  Let("bar",
    MakeStruct("baz" -> GetField(Ref("foo"), "kf"), ...)
    GetField(Ref("bar"), "baz")))
```

Edit:
A note on how serious the issue was in practice: Any filter that uses more than one field (or the same field twice) was broken, where broken means it does no interval filter and reads all the data. E.g. `ht.filter((ht.locus >= hl.locus('20', 1)) & (ht.locus < hl.locus('20', 10200000)))`. This is because the repeated field is pulled out into a let by CSE. If it was two different fields, the underlying `Ref("row")` is pulled out.

Other restrictions this PR doesn't address:
* The predicate can't have a repeated sub-predicate, e.g. 
   ```
   subcond = ht.locus.contig == '20'
   ht.filter((subcond & (ht.locus.position < 10)) | (subcond & (ht.locus.position > 10200000)))
   ```
* The predicate can't use negation or `If`
* The key type being filtered can only be Locus or numeric (or structs of those), doesn't work for e.g. gene Strings.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/ExtractIntervalFilters.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/ExtractIntervalFilters.scala
@@ -2,6 +2,7 @@ package is.hail.expr.ir
 
 import is.hail.backend.ExecuteContext
 import is.hail.annotations.IntervalEndpointOrdering
+import is.hail.rvd.PartitionBoundOrdering
 import is.hail.types.virtual._
 import is.hail.utils.{FastSeq, Interval, IntervalEndpoint, _}
 import is.hail.variant.{Locus, ReferenceGenome}
@@ -32,7 +33,7 @@ object ExtractIntervalFilters {
     val rowType: TStruct = rowRef.typ.asInstanceOf[TStruct]
     val rowKeyType: TStruct = rowType.select(keyFields)._1
     val firstKeyType: Type = rowKeyType.types.head
-    val iOrd: IntervalEndpointOrdering = rowKeyType.ordering(ctx.stateManager).intervalEndpointOrdering
+    val iOrd: IntervalEndpointOrdering = PartitionBoundOrdering(ctx, rowKeyType).intervalEndpointOrdering
 
     def getReferenceGenome(rg: String): ReferenceGenome = ctx.stateManager.referenceGenomes(rg)
 
@@ -194,7 +195,6 @@ object ExtractIntervalFilters {
         val rr = extractAndRewrite(r, es)
         (ll, rr) match {
           case (Some((ir1, i1)), Some((ir2, i2))) =>
-            println(s"left: ${i1.mkString("(", ", ", ")")}, right: ${i2.mkString("(", ", ", ")")}")
             log.info(s"intersecting list of ${ i1.length } intervals with list of ${ i2.length } intervals")
             val intersection = Interval.intersection(i1, i2, es.iOrd)
             log.info(s"intersect generated ${ intersection.length } intersected intervals")
--- a/hail/src/main/scala/is/hail/expr/ir/ExtractIntervalFilters.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/ExtractIntervalFilters.scala
@@ -65,12 +65,14 @@ object ExtractIntervalFilters {
     def isFirstKey(ir: IR): Boolean = isKeyField(ir, rowKeyType.fieldNames.head)
 
     def isKeyStructPrefix(ir: IR): Boolean = ir match {
-      case MakeStruct(fields) => fields.view.zipWithIndex.forall { case ((_, fd), idx) =>
-        idx < rowKeyType.size && isKeyField(fd, rowKeyType.fieldNames(idx))
-      }
-      case SelectFields(old, fields) => fields.view.zipWithIndex.forall { case (fd, idx) =>
-        idx < rowKeyType.size && fieldIsKeyField(old, fd, rowKeyType.fieldNames(idx))
-      }
+      case MakeStruct(fields) =>
+        fields.size <= rowKeyType.size && fields.view.zipWithIndex.forall { case ((_, fd), idx) =>
+          isKeyField(fd, rowKeyType.fieldNames(idx))
+        }
+      case SelectFields(old, fields) =>
+        fields.size <= rowKeyType.size && fields.view.zipWithIndex.forall { case (fd, idx) =>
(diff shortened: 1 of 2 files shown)


hail/src/main/scala/is/hail/expr/ir/ExtractIntervalFilters.scala
```diff
@@ -267,10 +310,10 @@ object ExtractIntervalFilters {
             case _ => None
           }
         }
-      case Let(name, value, body) if name != es.rowRef.name =>
+      case Let(name, value, body) =>
         // TODO: thread key identity through values, since this will break when CSE arrives
         // TODO: thread predicates in `value` through `body` as a ref
```
danking, line 315: What are these TODOs about? What would it mean to do them?

hail/src/main/scala/is/hail/expr/ir/ExtractIntervalFilters.scala
```diff
@@ -2,32 +2,75 @@ package is.hail.expr.ir
...
+    def isFirstKey(ir: IR): Boolean = isKeyField(ir, rowKeyType.fieldNames.head)
 
     def isKeyStructPrefix(ir: IR): Boolean = ir match {
-      case MakeStruct(fields) => fields
-        .iterator
-        .map(_._2)
-        .zipWithIndex
-        .forall { case (fd, idx) => idx < rowKeyType.size && fd == GetField(rowRef, rowKeyType.fieldNames(idx)) }
-      case SelectFields(`rowRef`, fields) => keyFields.startsWith(fields)
+      case MakeStruct(fields) => fields.view.zipWithIndex.forall { case ((_, fd), idx) =>
```
danking, line 68: what's the `view` do?

patrick-schultz: It makes a lazy collection wrapping `fields`, so the `zipWithIndex` doesn't materialize. See https://docs.scala-lang.org/overviews/collections/views.html

hail/src/main/scala/is/hail/expr/ir/ExtractIntervalFilters.scala
patrick-schultz: Thanks, I forgot to update the todos.

The problem with the old `ExtractIntervalFilters` was that, like many of our analyses and optimizations, we couldn't see through Let nodes. This manifested in two ways, corresponding to the two comments:

1. We only recognized a key field when it was literaly `GetField(Ref("row"), keyField)`. If it was `Let(row, Ref("row"), ... GetField(row, keyField) ...)`, we didn't recognize it as a key field, and so wouldn't extract filters on the field to intervals of the key type.
2. If `value` is a predicate, we didn't try to extract intervals from it. For example, if the filter predicate is `Let(p, keyField < 10, ... p ... p)`, we didn't even try to lift the `keyField < 10` part. If `p` were inlined, however, we would analyze it correctly.

This PR addresses the first, but not the second. Over the weekend, I realized the right way to fix this, which addresses both restrictions, and I think fixes some other blind spots as well, and is a simpler design. I think the better change isn't much harder than this one was. I'm going to take a stab at it this morning, and make a PR if it doesn't take too long. But we can still merge this in the meantime since it fixes a blocking issue.

hail/src/main/scala/is/hail/expr/ir/ExtractIntervalFilters.scala
```diff
@@ -2,32 +2,75 @@ package is.hail.expr.ir
...
 
     def getReferenceGenome(rg: String): ReferenceGenome = ctx.stateManager.referenceGenomes(rg)
 
-    def isFirstKey(ir: IR): Boolean = ir == GetField(rowRef, rowKeyType.fieldNames.head)
+    def bind(name: String, v: IR): ExtractionState =
+      new ExtractionState(ctx, env.bind(name, v), rowRef, keyFields)
+
+    private def isRowRef(ir: IR): Boolean = ir match {
+      case ref: Ref =>
+        ref == rowRef || env.lookupOption(ref.name).exists(isRowRef)
```
ehigham, line 45: Something has puzzled me for a while and that's refs of refs.
Are they a thing (this code seems like it)?
Should they be a thing?
What do they mean?

hail/src/main/scala/is/hail/expr/ir/ExtractIntervalFilters.scala
```diff
@@ -2,32 +2,75 @@ package is.hail.expr.ir
...
-        .iterator
-        .map(_._2)
-        .zipWithIndex
-        .forall { case (fd, idx) => idx < rowKeyType.size && fd == GetField(rowRef, rowKeyType.fieldNames(idx)) }
-      case SelectFields(`rowRef`, fields) => keyFields.startsWith(fields)
+      case MakeStruct(fields) => fields.view.zipWithIndex.forall { case ((_, fd), idx) =>
+        idx < rowKeyType.size && isKeyField(fd, rowKeyType.fieldNames(idx))
+      }
+      case SelectFields(old, fields) => fields.view.zipWithIndex.forall { case (fd, idx) =>
+        idx < rowKeyType.size && fieldIsKeyField(old, fd, rowKeyType.fieldNames(idx))
```
ehigham, line 72: Do we not eliminate nested `SelectFields` nodes? This case analysis and that in `fieldIsKeyField` seems to indicate as much.

ehigham (changes requested):

hail/src/main/scala/is/hail/expr/ir/ExtractIntervalFilters.scala
patrick-schultz: Not sure what nesting you're referring to. A common case here would be `SelectFields(Ref("row"), ["locus", "alleles"])`. We want to recognize this as being the key (or possibly a prefix of the key).

ehigham (approved):
Nice. I look forward to your subsequent PR that removes this recursive pattern matching

PR #13336, opened 2023-07-28
[query] tableaggregate o tableparallelize should be local

danking:
For example, the following should not do a 16-way parallel operation.
```
In [1]:     import hail as hl
   ...:     mt = hl.utils.range_matrix_table(10, 10, n_partitions=3)
   ...:     mt = mt.annotate_globals(n=mt.aggregate_cols(hl.struct(
   ...:         n=hl.agg.count()
   ...:     ), _localize=False))
   ...:     mt.n.collect()
```

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Simplify.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Simplify.scala
@@ -616,19 +616,19 @@ object Simplify {
         AggLet(uid, newRow, Subst(query, BindingEnv(agg = Some(Env("row" -> Ref(uid, newRow.typ))))), isScan = false))
 
     // NOTE: The below rule should be reintroduced when it is possible to put an ArrayAgg inside a TableAggregate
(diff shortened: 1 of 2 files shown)


danking:
@patrick-schultz Do I understand the error message correctly to mean I need to implement an Interpret step for this? Is that like an hour of work, a day, a week?

danking (commented):
@patrick-schultz bump, and also, in light of recent events, is there a concern here that we sometimes might intentionally want to parallelize a computationally expensive operation on trivial data? It seems like the root issue here is that we generate TableAggregate(TableParallelize ...) for aggregating over sample metadata. That seems just incorrect, semantically, the following is understood to be a local operation in Hail
```
mt.aggregate_cols(hl.agg.mean(mt.pheno))
```

patrick-schultz:
I agree with all of that. We didn't have a way to express a local aggregation until recently, but now `aggregate_cols` really should be purely local. I'll PR that change.

patrick-schultz:
#13405

If you agree, we can close this PR.

PR #13342, opened 2023-07-31
[query] Add file info to Tabix VCF read exception

chrisvittal:


patrick-schultz (approved):

chrisvittal:
This doesn't seem to be working. Debugging now.

chrisvittal:
fixed it, need to have the checks all the way inside the innermost iterator since it's next method may get called by scala as part of the `filter`.

PR #13343, opened 2023-07-31
[ci] Remove unnecessary worker-deploy-config secret

daniel-goldstein:
Since #13211, all jobs by default have a deploy config mounted into the container. The `worker-deploy-config` secret is no longer necessary, so long as we properly configure the namespace that CI jobs need to talk to.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2470,10 +2470,6 @@ steps:
         namespace:
           valueFrom: default_ns.name
         mountPath: /test-gsa-key
-      - name: worker-deploy-config
-        namespace:
-          valueFrom: default_ns.name
-        mountPath: /deploy-config
       - name: test-tokens
         namespace:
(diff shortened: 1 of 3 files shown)


build.yaml
```diff
@@ -3551,10 +3501,6 @@ steps:
         namespace:
           valueFrom: default_ns.name
         mountPath: /user-tokens
-      - name: worker-deploy-config
```
jigold, line 3554: Do you need to set `HAIL_DEFAULT_NAMESPACE` for this job or was the secret not necessary?

jigold (changes requested):

build.yaml
daniel-goldstein: The namespace is already specified in the command: `hailctl curl {{ default_ns.name }}`

jigold (approved):

PR #13344, opened 2023-07-31
[azure] Add connection timeouts for Azure FS in both Python and Scala

jigold:
I hope this helps. The regular `timeout` parameter on the Python methods like `stage_block` is for the server-side timeout of the operation. We want the client side timeout to be shorter. I picked 5 seconds, but that was arbitrary. I'm not sure I implemented the Scala code correctly, but that's along the right lines of what needs to happen.

### DIFF
--- a/hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
+++ b/hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
@@ -6,7 +6,7 @@ import is.hail.shadedazure.com.azure.storage.blob.models.{BlobProperties, BlobRa
 import is.hail.shadedazure.com.azure.storage.blob.specialized.BlockBlobClient
 import is.hail.shadedazure.com.azure.storage.blob.{BlobClient, BlobContainerClient, BlobServiceClient, BlobServiceClientBuilder}
 import is.hail.shadedazure.com.azure.core.http.netty.NettyAsyncHttpClientBuilder
-import is.hail.shadedazure.reactor.netty.http.client.HttpClient
+import is.hail.shadedazure.com.azure.core.http.HttpClient
 import is.hail.services.retryTransientErrors
 import is.hail.io.fs.FSUtil.{containsWildcard, dropTrailingSlash}
 import org.apache.log4j.Logger
@@ -148,7 +148,7 @@ object AzureStorageFileStatus {
   }
 }
 
-class AzureBlobServiceClientCache(credential: TokenCredential) {
+class AzureBlobServiceClientCache(credential: TokenCredential, val httpClient: HttpClient) {
   private[this] lazy val clients = mutable.Map[(String, String, Option[String]), BlobServiceClient]()
 
   def getServiceClient(url: AzureStorageFSURL): BlobServiceClient = {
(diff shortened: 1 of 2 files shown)


daniel-goldstein (dismissed):
Commenting for the record: This change is to reduce the time we spend hanging on connection timeouts. For problems with established connections in the future, we would need to configure different timeout options.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -212,10 +214,12 @@ class AzureStorageFS(val credentialsJSON: Option[String] = None) extends FS {
     }
   }
 
+  private lazy val httpClient = new NettyAsyncHttpClientBuilder().connectTimeout(Duration.ofSeconds(5)).build()
```
danking, line 217: Let's use the AzureBlobServiceClientBuilder's "ClientOptions" interface for this instead. That way, we can still let Azure pick the best defaults for its HttpClient. We want to set an [`HTTPClientOptions`](https://learn.microsoft.com/en-us/java/api/com.azure.core.util.httpclientoptions?view=azure-java-stable) for [`clientOptions`](https://learn.microsoft.com/en-us/java/api/com.azure.storage.blob.blobserviceclientbuilder?view=azure-java-stable#com-azure-storage-blob-blobserviceclientbuilder-clientoptions(com-azure-core-util-clientoptions))


e.g.

```
    val httpClientOptions = new HttpClientOptions()
    httpClientOptions.setReadTimeout(...)
    httpClientOptions.setResponseTimeout(...)
    httpClientOptions.setConnectTimeout(...)
    httpClientOptions.setConnectionIdleTimeout(...)
    httpClientOptions.setWriteTimeout(...)
    val blobServiceClient = new BlobServiceClientBuilder()
      .endpoint(s"https://${url.account}.blob.core.windows.net")
      .clientOptions(httpClientOptions)
      .buildClient()
```

danking (changes requested):

jigold:
What do we want the timeouts to be?

```scala
  private lazy val httpClientOptions = new HttpClientOptions()
    .setReadTimeout(Duration.ofSeconds(5))
    .setConnectTimeout(Duration.ofSeconds(5))
    .setConnectionIdleTimeout(Duration.ofSeconds(5))
    .setWriteTimeout(Duration.ofSeconds(5))
```

Also, python has ``read_timeout`` as well. Should I add that and what should the timeout be? The default limit is 1 minute.

danking:
Indeed 5 seconds is the default timeout we use in httpx.py. Let's do 5s. If we would prefer a different default global timeout, let's do that as a separate PR. 15s also seems reasonable.

jigold:
Is 5 seconds for all of these types of timeouts?

danking:
In `httpx` the 5s is a "total timeout" which is applied to read and connect. I don't know what connection idle and write exactly refer to but setting to 5s and seeing how the tests react seems like a good start. In general, in a datacenter, I'd expect any ordinary network operation to be faster than 5s regardless of write vs read.

danking (approved):

PR #13347, opened 2023-07-31
[devdocs] Add instructions on giving a dev admin access to their dev namespace

daniel-goldstein:
As part of the security POAMs, we'll follow up on this with instructions of making operator devs viewers + ability to assign themselves roles. This is just what I did for Sophie to give access just to the dev namespace.


danking (approved):
Excellent.

PR #13348, opened 2023-07-31
[query] Parameterise BlockMatrix tests in test_linalg.py  (fixes #13261)

### LINKED ISSUE #13261: [query] test_linalg.py::test_slicing, serially executes 28 BM distributed operations
### What happened?

https://hail.zulipchat.com/#narrow/stream/127527-team/topic/CI.20Deploy.20Failure/near/376092745

> At first glance, it seems to me that the BatchedAsserts Python class places repeated calls to execute (which is good for trivial operations like 1 + 1) together in one IR (it creates a tuple of the results). This interacts badly with non-trivial IRs: it serially executes all these block matrix operations.
> 
> I think the right fix is to split up test_splicing and any other BM tests. BatchedAsserts should only be used for operations which don't invoke a CDA. For operations that invoke a CDA, they should be separate Python test functions so that they're actually run in parallel.

### Version

0.2.119

### Relevant log output

_No response_

ehigham:
RR: https://github.com/hail-is/hail/issues/13261

Grouping asserts of distributed `BlockMatrix` queries via `BatchAssert` lead to repeated timeout failures during tests that used the batch-service backend.
This change removes all `BatchAssert`s from `test_linalg.py`. It uses `pytest.mark.parameterize` to gain parallelism in test execution from the test driver.


chrisvittal (approved):
This is terrific. Nice work.

PR #13350, opened 2023-08-01
[batch] Fix rare bug in list batches v2 test

jigold:


batch/test/test_batch.py
```diff
@@ -314,14 +314,17 @@ def assert_batch_ids(expected: Set[int], q=None):
...
 
 def test_list_batches_v2(client: BatchClient):
-    # replace any occurrences of the substring "b2" in the tag to avoid collisions with the batch name for the partial
-    # match test
-    tag = secrets.token_urlsafe(64).replace("b2", "00")
-    bb1 = create_batch(client, attributes={'tag': tag, 'name': 'b1'})
+    tag = secrets.token_urlsafe(64)
+    partial_match_prefix = secrets.token_urlsafe(5)
+    bb1 = create_batch(
+        client, attributes={'tag': tag, 'name': 'b1', 'partial_match_name': f'{partial_match_prefix}-b1'}
```
iris-garden, line 320: it could be good to add some characters to the beginning and/or end of the `partial_match_name`s, to test that the match can be partial, rather than the exact value of an attribute

iris-garden (dismissed):
left a comment, otherwise LGTM!

iris-garden (approved):

Issue #13351, opened 2023-08-01
[query] In Azure, QoB sees elevated rates of weird errors from the Azure Blob Storage SDK 


danking:
### What happened?

We have reported to their GitHub, but we don't have a simple enough repro for them to make progress. https://github.com/Azure/azure-sdk-for-java/issues/35125

Personal correspondence with some MSFT researchers suggested there could be an issue with threading:
> It sort of reminds me of an issue we saw with Cromwell where their old akka pool code caused a bunch of unexpected network behavior that broke their API in certain cases.

I've asked if BlobServiceClient is thread-safe or not. We share an object of that class, but none of the things it produces (e.g. blobs).

We know that the java.io libraries can improperly drop an HTTP response if it is followed by a TCP RST. In particular, we've seen this happen when a server is load shedding and sends an HTTP "429 Too Many Requests" rapidly followed by a TCP RST. This might explain the "Connection reset" errors that we sometimes see.

We have fewer intuitions about the "Stream is already closed". That specific error was reported to Azure in the aforementioned GitHub issue. We treat both stream is closed and connection reset as "limited retry" errors.

We might retry too quickly. Our initial delay is `100ms * x` where `x` is drawn uniformly from `[0, 1]`. Perhaps we should try an initial delay of at least 1s? 

For example, [Azure gives as an example retrying after 2s, 4s, 10s, and 30s](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-performance-checklist#timeout-and-server-busy-errors). Google's [code examples](https://cloud.google.com/storage/docs/retry-strategy#client-libraries_1) suggest an initial delay of 1s with a multiplier of 2. AWS seems to use 500ms as the [default base backoff for "throttled" exceptions](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java#L39).

### Version

0.2.120

danking:
(1) implemented in  https://github.com/hail-is/hail/pull/13354

jigold:
In case this is helpful, these are the operations with the library version that did not succeed over the last 48 hours:

```
OperationName | UserAgentHeader | count_
-- | -- | --
GetBlobProperties | azsdk-java-azure-storage-blob/12.22.0 (1.8.0_362; Linux; 5.4.0-1056-azure) | 298484
[...]

PR #13353, opened 2023-08-01
[qob] disable QoB PR tests (keep deploy and dev)

danking:
For added context see: https://github.com/hail-is/hail/issues/13351

cc: @patrick-schultz @chrisvittal @daniel-goldstein @ehigham @iris-garden 

I actually don't think we need to notify anyone because we'll still get errors if something goes wrong in default. If main starts failing, we have the slightly annoying situation of needing to run tests in a one-off manner to verify and we might need to block a release until we can revert.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2469,7 +2469,7 @@ steps:
       - deploy_memory_sa
       - create_certs
   - kind: runImage
-    name: test_hail_python_service_backend
+    name: test_hail_python_service_backend_gcp
     numSplits: 16
     image:
       valueFrom: hail_run_image.image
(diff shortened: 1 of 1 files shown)


build.yaml
```diff
@@ -2554,6 +2554,99 @@ steps:
...
+      export HAIL_TEST_STORAGE_URI={{ global.test_storage_uri }}/{{ token }}
+      export HAIL_TEST_RESOURCES_DIR="{{ global.test_storage_uri }}/{{ upload_test_resources_to_blob_storage.token }}/test/resources/"
+      export HAIL_DOCTEST_DATA_DIR="{{ global.test_storage_uri }}/{{ upload_test_resources_to_blob_storage.token }}/doctest/data/"
+      export HAIL_GENETICS_VEP_GRCH37_85_IMAGE={{ hailgenetics_vep_grch37_85_image.image }}
+      export GOOGLE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
+      export AZURE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
+
+      {% if global.cloud == "gcp" %}
+      export GCS_REQUESTER_PAYS_PROJECT=broad-ctsa
+      {% elif global.cloud == "azure" %}
```
jigold, line 2584: You could theoretically get rid of this if-elif case. Same in the GCP tests.

build.yaml
```diff
@@ -2554,6 +2554,99 @@ steps:
...
+      export HAIL_AZURE_SUBSCRIPTION_ID={{ global.azure_subscription_id }}
+      export HAIL_AZURE_RESOURCE_GROUP={{ global.azure_resource_group }}
+      {% endif %}
+
+      export HAIL_SHUFFLE_MAX_BRANCH=4
+      export HAIL_SHUFFLE_CUTOFF=1000000
+      export HAIL_QUERY_BACKEND=batch
+      {% if global.cloud == "azure" %}
+      export HAIL_BATCH_REGIONS={{ global.azure_location }}
+      {% elif global.cloud == "gcp" %}
```
jigold, line 2594: You could get rid of the if-else here as well.

jigold (changes requested):

danking:
@jigold My inclination was to keep the jobs the same so that we could use `diff` between the two entries if we needed to verify our changes were the same across both. Does that change your opinion or do you prefer to distinct but similar steps?

jigold:
That's fine for the first two comments, but please address the last comment to have the `delete_batch_azure_instances` depend on the Azure tests.

danking:
Ah, done. Now delete azure and delete GCP depend on the exact same steps (except the image step).

danking:
Ah, and this isn't a bug in `main` because `test_batch_invariants` depends on the QoB tests.

jigold (approved):

PR #13354, opened 2023-08-01
[services] use AWS-inspired retry delays

danking:
The first major change is that the base delay is now never below 1s. Previously our base delay was often 100ms.

The second major change is that I implemented the "Equal Jitter" algorithm from AWS's page on [backoff and jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/).


danking:
GitHub had dropped the [7690873](https://github.com/hail-is/hail/pull/13354/commits/7690873031b053605999e6d14236a200c733b621) commit somehow.

batch/test/test_batch.py
```diff
@@ -1464,16 +1464,19 @@ def test_job_private_instance_cancel(client: BatchClient):
...
         if status['state'] == 'Creating':
             break
         now = time.time()
+
+        tries += 1
+        delay = delay_ms_for_try(tries)
         if now + delay - start > 60:
             assert False, str((status, b.debug_info()))
-        delay = sync_sleep_and_backoff(delay)
+        time.sleep(delay)
```
daniel-goldstein, line 1479: can you use `sync_sleep_before_try here` here?

daniel-goldstein (dismissed):
I think this will help loads with MJC rate limiting too. Can you link to the AWS post in the delay calculation?

batch/test/test_batch.py
danking: I can't, I need to know the delay so I can ensure we do no more than 60s of delay. This test is different from everywhere else that we use this stuff. It wants to ensure we wait no more than 60s for cancellation.

danking: I actually had to add an extra change to this test to ensure we at least wait 60s. This new code seems to back off so quickly that there's a decent chance that our cumulative delay is like ~30s and our next delay is ~30s so the `assert False` is trigged.

daniel-goldstein: Ah, missed that. Alright

daniel-goldstein (approved):

PR #13360, opened 2023-08-02
[query/vds] Teach read_vds to supress the 'no ref block max length warning'

chrisvittal:
Use it in the combiner as it's very noisy otherwise.


ehigham (approved):
:shipit:

PR #13364, opened 2023-08-02
[query/combiner] Fix import_gvcf_interval's default PGT handling

chrisvittal:
PGT is a call field by default for import_vcf, the now former import_gvcfs (old combiner start point). Make import_gvcf_interval also treat PGT as a call field.

### DIFF
--- a/hail/python/test/hail/vds/test_combiner.py
+++ b/hail/python/test/hail/vds/test_combiner.py
@@ -41,7 +41,7 @@ def test_combiner_works():
         comb = hl.vds.read_vds(out)
 
         # see https://github.com/hail-is/hail/issues/13367 for why these assertions are here
-        assert 'LPGT' in comb.variant_data
+        assert 'LPGT' in comb.variant_data.entry
         assert comb.variant_data.LPGT == hl.tcall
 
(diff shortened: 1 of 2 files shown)


danking (changes requested):
Do we have a GVCF with PGT? Can you test for this?

danking (approved):

hail/python/test/hail/vds/test_combiner.py
```diff
@@ -39,6 +39,11 @@ def test_combiner_works():
         out = os.path.join(tmpdir, 'out.vds')
         hl.vds.new_combiner(temp_path=tmpdir, output_path=out, gvcf_paths=paths, intervals=parts, reference_genome='GRCh38').run()
         comb = hl.vds.read_vds(out)
+
+        # see https://github.com/hail-is/hail/issues/13367 for why these assertions are here
+        assert 'LPGT' in comb.variant_data
```
danking, line 44: I think you need `variant_data.entry` or something

chrisvittal: 🤦 thanks!

PR #13368, opened 2023-08-02
[azure] Optimize file status operations

jigold:
I think this change will help the number of operations we're making substantially. My Scala skills are not great, so I don't know if this is written correctly.

Basically, we were making a call to list the blobs recursively to test if the path was a directory which was streaming through the first 5000 records. I made the page size equal to 1 record as we don't care about all records. The next thing I did was to just get the blob properties rather than calling exists + get blob properties. So that will cut the number of HTTP calls by half for every blob. Lastly, listing items in a directory which is used for globbing was making a call to get the metadata for each file and then it was making the 3 API calls above to check whether it's a directory, whether it exists, and what the blob properties are. All of this information is in the original result from listing the blobs in the hierarchy so I just use that information directly.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2605,7 +2605,6 @@ steps:
     scopes:
       - deploy
       - dev
-      - test
     clouds:
       - azure
   - kind: runImage
--- a/build.yaml
+++ b/build.yaml
@@ -2605,6 +2605,7 @@ steps:
     scopes:
       - deploy
       - dev
+      - test
     clouds:
       - azure
   - kind: runImage
--- a/hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
+++ b/hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
@@ -154,7 +154,7 @@ object AzureStorageFileStatus {
       new BlobStorageFileStatus(blobPath.getPath, null, 0, true)
     } else {
       val properties = blobItem.getProperties
-      new BlobStorageFileStatus(blobPath.getPath, properties.getLastModified.toEpochSecond, properties.getContentLength, true)
+      new BlobStorageFileStatus(blobPath.getPath, properties.getLastModified.toEpochSecond, properties.getContentLength, false)
     }
   }
 }
@@ -435,7 +435,9 @@ class AzureStorageFS(val credentialsJSON: Option[String] = None) extends FS {
             throw new FileNotFoundException(s"File not found: $filename")
           else
             throw e
-    } else null
+      }
+    } else
+      null
 
     AzureStorageFileStatus(filename, isDir, blobProperties)
   }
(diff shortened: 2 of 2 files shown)


build.yaml
```diff
@@ -2605,6 +2605,7 @@ steps:
     scopes:
       - deploy
       - dev
+      - test
```
danking, line 2608: note to ourselves to nix this before merging

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -389,8 +399,9 @@ class AzureStorageFS(val credentialsJSON: Option[String] = None) extends FS {
     val prefixMatches = blobContainerClient.listBlobsByHierarchy(prefix)
 
     prefixMatches.forEach(blobItem => {
-      statList += fileStatus(url.withPath(blobItem.getName))
+      statList += AzureStorageFileStatus(blobItem)
```
danking, line 402: holy yikes-a-molly! Great change, this is awful.

danking (changes requested):
Wow this is a killer find. Excellent work. I think we should hold off on enabling QoB tests. Let's do that as a separate PR and maybe run it 10-20 times to convince ourselves its finally reliable first.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
```diff
@@ -141,11 +141,21 @@ object AzureStorageFS {
...
+    if (isDir) {
+      new BlobStorageFileStatus(path, null, 0, true)
+    } else {
+      new BlobStorageFileStatus(path, blobProperties.getLastModified.toEpochSecond, blobProperties.getBlobSize, false)
+    }
+  }
 
-    new BlobStorageFileStatus(path, modificationTime, size, isDir)
+  def apply(blobItem: BlobItem): BlobStorageFileStatus = {
+    val properties = blobItem.getProperties
```
danking, line 153: While you're changing build.yaml, also push this into the else branch.

hail/src/main/scala/is/hail/io/fs/AzureStorageFS.scala
danking, line 402: Ah, I think this is actually a bit of a bug. The name of a blobItem is like `/foo/bar/baz` but we need the URL which is `hail-az://account/container/foo/bar/baz`.

danking (dismissed):

jigold:
I'm putting the WIP tag to remind ourselves to turn off the Azure tests before merging.

danking (approved):
LGTM. Excellent change.

jigold:
@danking I had to make some changes I'm not 100% confident in with stripping trailing "/".

danking (approved):
Yeah, this is right, the old code for directories did the same thing:
```scala
    val filename = dropTrailingSlash(url.toString)
    if (!isDir && !blobClient.exists()) {
      throw new FileNotFoundException(s"File not found: $filename")
    }

    if (isDir) {
      new BlobStorageFileStatus(path = filename, null, 0, isDir = true)
```

PR #13369, opened 2023-08-02
[query] ensure functions need by DSP are not moved

danking:
https://hail.zulipchat.com/#narrow/stream/379853-Hail.2FVariants/topic/handing.20off.20import_gvs/near/381266567


chrisvittal (approved):
for the record, I hate this. If we're going to do this, why not promote it out of experimental, and get it taken care of by our stability guarantees?

PR #13371, opened 2023-08-03
[query] GoogleFS: recreate reader on transient error

chrisvittal:
While running QoB jobs, we observed some errors relating to compression validation or unexpected end of line/file. It was suggested that after certain errors, the reader may be in an unsuable state, so we recreate it after an error.


chrisvittal:
I give up, every change I make seems to break something, and it's not clear that any of these changes actually are helping.

PR #13372, opened 2023-08-03
jobs after update simple

danking:


danking:
take a look and tell me if this helps simplify this trigger

danking:
bump for thoughts! I think this cuts through all the unnecessary complexity of SQL and isolates the essential complexity. This is just literally the state transitions for batch jobs:
```sql
  set delta_n_ready_cancellable_jobs        = (-1 * was_ready    *  was_cancellable  )     + (now_ready    *  now_cancellable  ) ;
  set delta_n_ready_jobs                    = (-1 * was_ready    * (not was_cancelled))    + (now_ready    * (not now_cancelled));
  set delta_n_cancelled_ready_jobs          = (-1 * was_ready    *  was_cancelled    )     + (now_ready    *  now_cancelled    ) ;

  set delta_n_running_cancellable_jobs      = (-1 * was_running  *  was_cancellable  )     + (now_running  *  now_cancellable  ) ;
  set delta_n_running_jobs                  = (-1 * was_running  * (not was_cancelled))    + (now_running  * (not now_cancelled));
  set delta_n_cancelled_running_jobs        = (-1 * was_running  *  was_cancelled    )     + (now_running  *  now_cancelled    ) ;

  set delta_n_creating_cancellable_jobs     = (-1 * was_creating *  was_cancellable  )     + (now_creating *  now_cancellable  ) ;
  set delta_n_creating_jobs                 = (-1 * was_creating * (not was_cancelled))    + (now_creating * (not now_cancelled));
  set delta_n_cancelled_creating_jobs       = (-1 * was_creating *  was_cancelled    )     + (now_creating *  now_cancelled    ) ;

  set delta_ready_cancellable_cores_mcpu    = delta_n_ready_cancellable_jobs * cores_mcpu;
  set delta_ready_cores_mcpu                = delta_n_ready_jobs * cores_mcpu;

  set delta_running_cancellable_cores_mcpu  = delta_n_running_cancellable_jobs * cores_mcpu;
  set delta_running_cores_mcpu              = delta_n_running_jobs * cores_mcpu;
```

jigold:
This is definitely much better! Do you want me to adopt this? I can put it on the lower priority stack I'm working on. I'm not sure if we would need to test this by hand. Thoughts on that?

danking:
I don't know the system or the test suite well any more. What parts of our system are not well tested by the automated tests?

No need for adoption, just review!

daniel-goldstein:
This is a great simplification. I'm on board

jigold:
I think we should add a new check for `check_invariants` that the counts all sum to 0 when the tests are all finished running in the `user_inst_coll_resources` table. I think we have enough tests with `always_run` and cancellation for pool jobs, but I don't think we test the interaction between always_run and job private ("creating" jobs).

jigold:
Also, you know I like CAPS rather than lower case for SQL keywords.

danking:
I find the SHOUTing is part of what makes the file feel so visually overwhelming, but there's much less SHOUTing now because there's just fewer keywords. Updated.

danking:
I added some where conditions to check_incremental

danking:
I added a test for always run job private cancellation.

danking:
I think that addresses the concerns. This is great, what other automated checks can we add to make us more confident in our tests?

batch/test/test_batch.py
```diff
@@ -1486,6 +1486,31 @@ def test_job_private_instance_cancel(client: BatchClient):
...
+    j = bb.create_job(DOCKER_ROOT_IMAGE, ['true'], resources=resources, always_run=True)
+    b = bb.submit()
+
+    tries = 0
+    start = time.time()
+    while True:
+        status = j.status()
+        if status['state'] == 'Creating':
+            break
+        now = time.time()
```
jigold, line 1501: Why do you need to poll here?

jigold (dismissed):

jigold:
Also, I'll check the SQL carefully once the tests pass.

batch/test/test_batch.py
danking: I copy pasted this test from `test_job_private_instance_cancel`. My guess is that we don't want to cancel until the job is actually creating (ergo we've actually created an instance).

jigold: It's always run though. Cancel shouldn't do anything.

batch/sql/jobs-after-update-simple.sql
```diff
@@ -0,0 +1,146 @@
...
+
+  SET was_marked_cancelled = old.cancelled OR cur_batch_cancelled;
+  SET was_cancelled        = NOT always_run AND was_marked_cancelled;
+  SET was_cancellable      = NOT always_run AND NOT was_marked_cancelled;
+
+  SET now_marked_cancelled = new.cancelled or cur_batch_cancelled;
+  SET now_cancelled        = NOT always_run AND now_marked_cancelled;
+  SET now_cancellable      = NOT always_run AND NOT now_marked_cancelled;
+
+  # NB: was_cancelled => now_cancelled b/c you cannot be uncancelled
```
jigold, line 69: I'm not sure what this comment was meant to say. Is it that if `was_cancelled = True`, then `now_cancelled` must be True?

batch/test/test_batch.py
jigold: Can you get rid of the polling? It's not necessary.

jigold (changes requested):

batch/sql/jobs-after-update-simple.sql
danking: I replaced `=>` with `implies`

danking:
bump

jigold (approved):

PR #13379, opened 2023-08-04
[batch] Raise instead of returning aiohttp objects that subclass Exception

daniel-goldstein:
Returning HTTP status objects that subclass `Exception` [is deprecated](https://docs.aiohttp.org/en/stable/web_quickstart.html#exceptions) and will at some point block an upgrade to a higher version of aiohttp.


danking (approved):
LGTM

PR #13385, opened 2023-08-07
[vds/combiner] Teach VDS combiner how to handle custom call fields

chrisvittal:
Add the call_fields argument.

Much like refrence entries, when passing a VDS to the combiner, we check it's call fields to ensure that there is a match between the passed argument and the VDS's call fields.

### DIFF
--- a/hail/python/hail/vds/combiner/variant_dataset_combiner.py
+++ b/hail/python/hail/vds/combiner/variant_dataset_combiner.py
@@ -4,6 +4,7 @@ import json
 import os
 import sys
 import uuid
+from itertools import chain
 from math import floor, log
 from typing import Collection, Dict, List, NamedTuple, Optional, Union
 
@@ -156,7 +157,7 @@ class VariantDatasetCombiner:  # pylint: disable=too-many-instance-attributes
         for all GVCF files. Finer partitioning yields more parallelism but less work per task.
     gvcf_info_to_keep : :class:`list` of :class:`str` or :obj:`None`
         GVCF ``INFO`` fields to keep in the ``gvcf_info`` entry field. By default, all fields are
-        kept except ``END`` and ``DP`` are kept.
+        except ``END`` and ``DP`` are kept.
     gvcf_reference_entry_fields_to_keep : :class:`list` of :class:`str` or :obj:`None`
         Genotype fields to keep in the reference table. If empty, the first 10,000 reference block
         rows of ``mt`` will be sampled and all fields found to be defined other than ``GT``, ``AD``,
@@ -196,6 +197,7 @@ class VariantDatasetCombiner:  # pylint: disable=too-many-instance-attributes
         '_gvcf_import_intervals',
         '_gvcf_info_to_keep',
         '_gvcf_reference_entry_fields_to_keep',
+        '_call_fields',
     ]
 
     __slots__ = tuple(__serialized_slots__ + ['_uuid', '_job_id', '__intervals_cache'])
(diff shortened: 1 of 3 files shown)


hail/python/hail/vds/combiner/variant_dataset_combiner.py
```diff
@@ -156,7 +157,7 @@ class VariantDatasetCombiner:  # pylint: disable=too-many-instance-attributes
         for all GVCF files. Finer partitioning yields more parallelism but less work per task.
     gvcf_info_to_keep : :class:`list` of :class:`str` or :obj:`None`
         GVCF ``INFO`` fields to keep in the ``gvcf_info`` entry field. By default, all fields are
```
danking, line 159: need to remove the `are` on the previous line

hail/python/hail/vds/combiner/variant_dataset_combiner.py
```diff
@@ -701,6 +709,16 @@ def maybe_load_from_saved_path(save_path: str) -> Optional[VariantDatasetCombine
...
         gvcf_reference_entry_fields_to_keep = ref_entry_tmp
 
+        all_entry_types = chain(vds.reference_data._type.entry.items(),
+                                vds.variant_data._type.entry.items())
+        call_fields_tmp = {name for name, typ in all_entry_types if typ == hl.tcall} - {'LGT', 'GT'}
+        if 'LPGT' in call_fields_tmp:
+            call_fields_tmp = (call_fields_tmp - {'LPGT'}) | {'PGT'}
+        if set(call_fields) != call_fields_tmp:
+            warning("Mismatch between 'call_fields' and VDS call fields. "
+                    "Overwriting with types from supplied VDS")
```
danking, line 719: Include the mismatch. In particular, the user should know the `call_fields` argument but might not know what's in the dataset.


What do you mean by "overwriting with types ..."? It's not the types that are taken from the VDS but the call fields that are taken from it, right?

hail/python/hail/vds/combiner/variant_dataset_combiner.py
```diff
@@ -701,6 +709,16 @@ def maybe_load_from_saved_path(save_path: str) -> Optional[VariantDatasetCombine
                     "entry types. Overwriting with types from supplied VDS.")
         gvcf_reference_entry_fields_to_keep = ref_entry_tmp
 
+        all_entry_types = chain(vds.reference_data._type.entry.items(),
+                                vds.variant_data._type.entry.items())
+        call_fields_tmp = {name for name, typ in all_entry_types if typ == hl.tcall} - {'LGT', 'GT'}
```
danking, line 714: These are the call fields from the vds, right? Let's use a descriptive name. Maybe `vds_call_fields`?

danking (changes requested):

danking:
Mostly me not knowing how this code works, so, my apologies for slowing you down.

hail/python/hail/vds/combiner/variant_dataset_combiner.py
chrisvittal: Will do. Also will clarify.

hail/python/hail/vds/combiner/variant_dataset_combiner.py
chrisvittal: 👍

hail/python/hail/vds/combiner/variant_dataset_combiner.py
```diff
@@ -701,6 +709,16 @@ def maybe_load_from_saved_path(save_path: str) -> Optional[VariantDatasetCombine
...
+        if 'LPGT' in call_fields_tmp:
+            call_fields_tmp = (call_fields_tmp - {'LPGT'}) | {'PGT'}
```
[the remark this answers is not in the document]
chrisvittal: We transform certain fields from their 'normal' gvcf versions to the VDS 'local' ones, PGT is one of them and is also the only VCF field we parse as a call field by default.

danking (approved):
Great!

PR #13396, opened 2023-08-08
[batch] expose Job.spot

danking:
Folks need this when preemption rates are too high for their jobs.

cc: @daniel-goldstein

### DIFF
--- a/hail/python/hailtop/batch/job.py
+++ b/hail/python/hailtop/batch/job.py
@@ -354,7 +354,7 @@ class Job:
         >>> b = Batch(backend=backend.ServiceBackend('test'))
         >>> j = b.new_job()
         >>> j = j.spot(False)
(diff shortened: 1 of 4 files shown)


jigold (approved):

hail/python/hailtop/batch/job.py
```diff
@@ -342,6 +342,32 @@ def always_run(self, always_run: bool = True) -> 'Job':
...
+
+        Examples
+        --------
+
+        Ensure a job only runs on non-spot instances:
+
+        >>> b = Batch(backend=backend.ServiceBackend('test'))
+        >>> j = b.new_job()
+        >>> j = j.spot(False)
+        >>> j = j.command(f'echo "hello"'))
```
jigold, line 357:
```suggestion
        >>> j = j.command(f'echo "hello"')
```

PR #13397, opened 2023-08-08
[build.yaml] scale dev namespaces to zero 9p-9a weekdays and weekends

danking:
cc: @jigold @iris-garden @sjparsa , since you all use dev namespaces.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -3741,6 +3741,7 @@ steps:
       ---
       {% endif %}
       {% endfor %}
+      EOF
       kubectl apply -f dev-namespace-scalers.yaml
     scopes:
       - deploy
--- a/build.yaml
+++ b/build.yaml
@@ -3679,3 +3679,73 @@ steps:
       - test_hail_python_service_backend_gcp
       - test_hail_python_service_backend_azure
       - cancel_all_running_test_batches
+  - kind: runImage
+    name: setup_dev_namespace_autoscalers
+    resources:
+      memory: standard
+      cpu: '0.25'
+    image:
+      valueFrom: ci_utils_image.image
+    script: |
+      set -ex
+      {% for user in code.get("developers", []) %}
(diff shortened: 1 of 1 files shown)


build.yaml
```diff
@@ -3679,3 +3679,73 @@ steps:
...
+      set -ex
+      {% for user in code.get("developers", []) %}
+      {% if user['username'] != 'test-dev' %}
+      apiVersion: batch/v1
+      kind: CronJob
+      metadata:
+        name: dev-namespace-scaledown-{{ user["username"] }}
+        namespace: {{ user["username"] }}
+      spec:
+        schedule: 0 22 * * 1,2,3,4,5"  # Weekdays at 8p
```
daniel-goldstein, line 3699: 10p, no?

daniel-goldstein (changes requested):
This is a great and easy change

build.yaml
```diff
@@ -3679,3 +3679,73 @@ steps:
...
+              spec:
+                serviceAccountName: admin
+                containers:
+                - name: dev-namespace-daytime-autoscaler
+                  image: bitnami/kubectl:latest
+                  command:
+                  - /bin/sh
+                  - -c
+                  - set -ex ; kubectl scale deployments --all -n {{ user["username"] }} --replicas=1 && kubectl scale statefulsets --all -n {{ user["username"] }} --replicas=1
+                restartPolicy: OnFailure
```
daniel-goldstein, line 3739: Don't you need a `---` delimeter at the end too since this is in a loop?

build.yaml
danking: Heh, yeah, I was experimenting with it at 10p. Fixed.

build.yaml
danking: yep.

daniel-goldstein (approved):

PR #13399, opened 2023-08-09
[batch] Mitigate siwei bug by removing cache

danking:
Before open batches, the `n_jobs` of a batch was a constant known before any jobs were added. Moreover, we did not start scheduling jobs until all the jobs were added to the database. Therefore, it was always safe to assume that the final "bunch" of jobs in the database was the last "bunch" ergo it spanned from its `start_job_id` to the job with id `n_jobs` (nb: job ids are 1-indexed).

When open batches were added, the `n_jobs` became a mutable value. Moreover, `n_jobs` includes jobs in bunches *which have not yet been added to the database*. In particular, suppose two clients are each submitting a bunch of size 10. Each client independently "reserves" 10 job slots by atomically incrementing `n_jobs` by ten. `n_jobs` is now 20. Further suppose that the first bunch is added to the database and begins scheduling before the second bunch is added to the database. In this case, when calculating the size of this bunch (for use in the bunch cache, and *only* in the bunch cache), we see that this is the last (and only) bunch in the database and assume that `n_jobs` is the last job id in this bunch. This is incorrect because `n_jobs` includes the not-yet-visible second bunch.

### DIFF
--- a/batch/batch/spec_writer.py
+++ b/batch/batch/spec_writer.py
@@ -8,12 +8,6 @@ from hailtop.utils import secret_alnum_string
 
 log = logging.getLogger('batch.spec_writer')
 
-JOB_TOKEN_CACHE: Dict[int, sortedcontainers.SortedSet] = collections.defaultdict(
-    lambda: sortedcontainers.SortedSet(key=lambda t: t[1])
-)
-JOB_TOKEN_CACHE_MAX_BATCHES = 100
-JOB_TOKEN_CACHE_MAX_BUNCHES_PER_BATCH = 100
-
 
 class SpecWriter:
     byteorder: Literal['little', 'big'] = 'little'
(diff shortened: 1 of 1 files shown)


jigold:
This is all correct including the description of the problem. I think we could add a test that does the following:

```python3
def test_update_batch_w_inprogress_job(client: BatchClient):
    bb = create_batch(client)
    bb.create_job(DOCKER_ROOT_IMAGE, ['sleep', '300'])
    bb.submit()

    j = bb.create_job(DOCKER_ROOT_IMAGE, ['true'])
    b = bb.submit()
    status = j.wait()

    assert status['state'] == 'Success', str((status, b.debug_info()))
    
    b.cancel()
```

jigold (dismissed):
see comment for test to add.

jigold:
Looks like something went wrong in the PR tests that seems unrelated to the PR. I haven't seen those tests fail before.

danking:
I have an error in the SQL

jigold (approved):

PR #13411, opened 2023-08-10
[query] changes in charr docstring

wlu04:
Hi all, I have made some quick edits in the docstring of the `hl.methods.qc.charr()` function. Please let me know if anything else is needed to clarify. Thanks!


danking:
@chrisvittal, you're more or less the keeper of VDS these days, and CHARR is pretty VDS-y.

chrisvittal (approved):
Thanks for the update. LGTM.

PR #13429, opened 2023-08-14
[ci] Make dev namespace scaling cron in UTC

daniel-goldstein:
GCP k8s crons are in UTC apparently. We could instead specify a timezone here, LMK which you prefer.


danking (approved):
Ah, yep, that makes total sense. My bad. Thanks for fixing for me!

PR #13430, opened 2023-08-14
Add Support for G2 Machines to Batch

sjparsa:
Currently, jobs in hail batch can only be run on n1 machines but with the rise of deep learning in bioinformatics, the ability to run jobs on g2 machines, as well as other GPU supported machines, is an important and exciting addition to hail batch. This PR highlights the steps needed to add new machine types into hail batch and could be used as a template for further development support. 

The changes in this PR can broadly be divided into additions to the job crun container and insertion of g2 resources (CPU, RAM, L4 Accelerator) into the resources table for billing. This PR uses the NVIDIA Container Toolkit, which allows the creation of GPU accelerated containers. This toolkit is integrated with docker via the parameters —runtime=nvidia and the specification of GPUs is made through —gpus all. The toolkit is installed in the batch worker VM startup script and the corresponding docker parameters are configured if the machine type is g2, so there is no change to the docker configuration for n1 machines. For the toolkit to work there is a nvidia hook that needs to be injected into the crun config. These modifications are also done based on machine type. On the billing side, the existing pricing setup was expanded to include g2 machines. The g2 instance cores and RAM are inserted into the database, and the SKUs are hard coded. For future machine type incorporation or updates, [https://cloud.google.com/skus/?currency=USD&filter=](https://cloud.google.com/skus/?currency=USD&filter=) may serve as a useful resource to identify relevant SKU ids. A new resource type was also added for the accelerator, including preemptible and non-preemtible. Finally, g2 machines mount the worker data disk under the name nvme0n2 so the code is updated to reflect this. Future work may want to investigate a way to automatically detect what the proper disk name is or make the disk naming logic more robust.


batch/batch/cloud/gcp/driver/create_instance.py
```diff
@@ -44,7 +44,7 @@ def create_vm_config(
 
     region = instance_config.region_for(zone)
 
-    if local_ssd_data_disk:
+    if local_ssd_data_disk or 'g2' in machine_type:
```
cseed, line 47: I think this logic is not quite right.  A pd-ssd disk attached to a g2 shows up as an NVMe drive, but as SCSI drive on a n1/n2.  But if local_ssd is not set, we should not switch the disk type to local-ssd.  I think you can check the machine type in the false branch to set the disk name.

batch/batch/cloud/gcp/driver/create_instance.py
```diff
@@ -43,7 +43,8 @@ def create_vm_config(
     _, cores = gcp_machine_type_to_worker_type_and_cores(machine_type)
 
     region = instance_config.region_for(zone)
-
+    nvidia_runtime = "--runtime=nvidia" if 'g2' in machine_type else ""
+    gpus_param = "--gpus all" if 'g2' in machine_type else ""
```
cseed, line 47: I think it will be useful to add a function (in batch/cloud/gcp/resource_utils.py, say) that returns whether a given machine_type has GPUs.  Then, when we add support for another machine type (like a2s), we only need to change it in one place.

You can also combine these options into a single variable, maybe something like `docker_run_gpu_args`.

batch/batch/cloud/gcp/driver/pricing.py
```diff
@@ -17,6 +18,58 @@
 log = logging.getLogger('pricing')
 
 
+G2_CPU_SKUS = [
```
cseed, line 21: These SKU numbers are a little opaque.  Can you change your script to include a comment with the description?

```
G2_CPU_SKUS = [
    # G2 Instance Core running in Iowa
    'D18E-3563-415F',
    ...
```

batch/batch/cloud/gcp/instance_config.py
```diff
@@ -49,15 +52,28 @@ def create(
...
-        resources = [
-            GCPComputeResource.create(product_versions, instance_family, preemptible, region),
-            GCPMemoryResource.create(product_versions, instance_family, preemptible, region),
-            GCPStaticSizedDiskResource.create(product_versions, 'pd-ssd', boot_disk_size_gb, region),
-            data_disk_resource,
-            GCPDynamicSizedDiskResource.create(product_versions, 'pd-ssd', region),
-            GCPIPFeeResource.create(product_versions, 1024),
-            GCPServiceFeeResource.create(product_versions),
-        ]
+        if GPU_VERSIONS.get(machine_type_parts.machine_family):
```
cseed, line 55: Better not to duplicate the resources list.  Just use `append` to add the accelerator resource to the list of resources if an accelerator is present:

```
  resources = [...]

  accelerator_family = GPU_VERSIONS.get(machine_type_parts.machine_family)
  if acclerator_family:
    resources.append(GCPAccleratorResource.create(...))
```

batch/batch/inst_coll_config.py
```diff
@@ -239,6 +239,7 @@ def __init__(
         self.worker_max_idle_time_secs = worker_max_idle_time_secs
 
     def convert_requests_to_resources(self, machine_type, storage_bytes):
+        # TODO: Can GPU resource request be added here?
```
cseed, line 242: Resolve this TODO.

batch/batch/worker/worker.py
```diff
@@ -1126,6 +1126,25 @@ async def container_config(self):
             'CAP_KILL',
             'CAP_AUDIT_WRITE',
         ]
+
+        nvidia_runtime_hook = []
+        if 'g2' in INSTANCE_CONFIG["machine_type"]:
+            log.info("g2-machine")
```
cseed, line 1132: Is this info necessary?

batch/batch/worker/worker.py
```diff
@@ -1126,6 +1126,25 @@ async def container_config(self):
             'CAP_KILL',
             'CAP_AUDIT_WRITE',
         ]
+
+        nvidia_runtime_hook = []
+        if 'g2' in INSTANCE_CONFIG["machine_type"]:
```
cseed, line 1131: Again, control this with the "has GPU" function.

batch/batch/worker/worker.py
```diff
@@ -1126,6 +1126,25 @@ async def container_config(self):
...
+                {
+                    "path": "/usr/bin/nvidia-container-runtime-hook",
+                    "args": ["nvidia-container-runtime-hook", "prestart"],
+                    "env": [
+                        "LANG=C.UTF-8",
+                        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin",
+                        "NOTIFY_SOCKET=/run/systemd/notify",
+                        "INVOCATION_ID=8b6eb73e784f439397beb9fd9b9aaba5",
+                        "JOURNAL_STREAM=8:17167",
+                        "SYSTEMD_EXEC_PID=761",
```
cseed, line 1143: I don't think the environment variables with hardcoded numbers make sense here.  I'd try removing INVOCATION_ID, JOURNAL_STREAM and SYSTEMD_EXEC_PID.

batch/batch/worker/worker.py
```diff
@@ -1324,7 +1345,11 @@ def _mounts(self, uid: int, gid: int) -> List[MountSpecification]:
 
     def _env(self):
         assert self.image.image_config
-        env = self.image.image_config['Config']['Env'] + self.env
+        gpu_env = ["NVIDIA_VISIBLE_DEVICES=all"]
+        if 'g2' in INSTANCE_CONFIG["machine_type"]:
+            env = self.image.image_config['Config']['Env'] + self.env + gpu_env
+        else:
+            env = self.image.image_config['Config']['Env'] + self.env
```
cseed, line 1352: Better to have one if and not duplicate the initial value of env:

```
   env = self.image.image_config['Config']['Env'] + self.env
   if machine_type_has_accelerator(INSTANCE_CONFIG['machine_type']:
       env += ["NVIDIA_VISIBLE_DEVICES=all"]
```

batch/build-batch-worker-image-startup-gcp.sh
```diff
@@ -41,12 +41,40 @@ curl -fsSL "https://github.com/GoogleCloudPlatform/docker-credential-gcr/release
...
 
+curl -fsSL https://nvidia.github.io/libnvidia-container/gpgkey \
+     | gpg --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg
+curl -s -L https://nvidia.github.io/libnvidia-container/ubuntu22.04/libnvidia-container.list | \
+    sed 's#deb https://#deb [signed-by=/usr/share/keyrings/nvidia-container-toolkit-keyring.gpg] https://#g' | \
+    tee /etc/apt/sources.list.d/nvidia-container-toolkit.list
+
+# don't use build-essentials, it pulls in gcc-11
+# the latest nvidia driver doesn't build with gcc-11
+#apt-get --yes install gcc-12 g++-12
```
cseed, line 52: Remove this line, the installation of gcc-12 is below.

batch/build-batch-worker-image-startup-gcp.sh
```diff
@@ -41,12 +41,40 @@ curl -fsSL "https://github.com/GoogleCloudPlatform/docker-credential-gcr/release
...
+apt-get update
```
cseed, line 53: You only need to run update once after installing the Nvidia key (the curl commands).

batch/sql/insert_gpu_resources.py
```diff
@@ -0,0 +1,28 @@
+import asyncio
```
cseed, line 1: I think this is no longer necessary since we're doing automatic resource price discovery now, right?

batch/test/test_batch.py
```diff
@@ -1110,7 +1110,8 @@ def test_verify_no_access_to_metadata_server(client: BatchClient):
     status = j.wait()
     assert status['state'] == 'Failed', str((status, b.debug_info()))
     job_log = j.log()
-    assert "Connection timed out" in job_log['main'], str((job_log, b.debug_info()))
+    # Curl output 'Connection timed out' changed to 'Connection timeout' with ubuntu image update
```
cseed, line 1113: This comment probably belongs in the PR, not the code.  The code is static and "ubuntu update" won't make senes over time.

batch/test/test_batch.py
```diff
@@ -1435,6 +1436,19 @@ def test_pool_standard_instance_cheapest(client: BatchClient):
     assert 'standard' in status['status']['worker'], str((status, b.debug_info()))
 
 
+def test_gpu_accesibility_g2(client: BatchClient):
+    bb = create_batch(client)
+    resources = {'machine_type': "g2-standard-4", 'storage': '100Gi'}
+    j = bb.create_job(
+        os.environ['HAIL_GPU_IMAGE'],
+        ['python', '-c', 'import torch; assert torch.cuda.is_available() == True'],
```
cseed, line 1444: You don't need `== True` here, it means the same thing as just `is_available()`.

batch_gpu_script.py
```diff
@@ -0,0 +1,10 @@
+import hailtop.batch as hb
```
cseed, line 1: This isn't used by the PR, is it?  Looks like it was a script you used for testing and has your hardcoded paths embedded.

pyega3_output.log
```diff
@@ -0,0 +1,4 @@
+[2023-08-21 10:05:19 -0400]
```
cseed, line 1: Remove this file?

test_batch.py
```diff
@@ -0,0 +1,9 @@
+import hailtop.batch as hb
```
cseed, line 1: Looks like this file also isn't used?

jigold:
Putting this here for posterity that the billing is correct. There's one job that's preemptible and the other nonpreemptible. Both jobs ran for an hour.

```
mysql> SELECT batch_id, job_id, resource, COALESCE(CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) * rate, 0) AS cost
    -> FROM aggregated_job_resources_v2
    -> LEFT JOIN resources ON aggregated_job_resources_v2.resource_id = resources.resource_id
    -> WHERE aggregated_job_resources_v2.batch_id = 5 AND aggregated_job_resources_v2.job_id = 1
    -> GROUP BY batch_id, job_id, resource;
+----------+--------+---------------------------------------------------------+----------------------+
| batch_id | job_id | resource                                                | cost                 |
+----------+--------+---------------------------------------------------------+----------------------+
|        5 |      1 | ip-fee/1024/1                                           | 0.004081841111111111 |
|        5 |      1 | service-fee/1                                           |  0.04081841111111112 |
|        5 |      1 | disk/pd-ssd/us-central1/1693067778861                   | 0.028497453342459508 |
|        5 |      1 | compute/g2-nonpreemptible/us-central1/1693067778861     |  0.10199791103476001 |
|        5 |      1 | memory/g2-nonpreemptible/us-central1/1693067778861      | 0.044810165988899996 |
|        5 |      1 | accelerator/l4-nonpreemptible/us-central1/1693110803273 |   0.5714988178566729 |
+----------+--------+---------------------------------------------------------+----------------------+
6 rows in set (0.00 sec)

mysql> SELECT batch_id, job_id, resource, COALESCE(CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) * rate, 0) AS cost
    -> FROM aggregated_job_resources_v2
    -> LEFT JOIN resources ON aggregated_job_resources_v2.resource_id = resources.resource_id
    -> WHERE aggregated_job_resources_v2.batch_id = 2 AND aggregated_job_resources_v2.job_id = 1
    -> GROUP BY batch_id, job_id, resource;
+----------+--------+------------------------------------------------------+----------------------+
| batch_id | job_id | resource                                             | cost                 |
+----------+--------+------------------------------------------------------+----------------------+
|        2 |      1 | ip-fee/1024/1                                        | 0.004279891111111111 |
|        2 |      1 | service-fee/1                                        | 0.042798911111111115 |
|        2 |      1 | compute/g2-preemptible/us-central1/1693067778861     | 0.042798911111111115 |
|        2 |      1 | disk/pd-ssd/us-central1/1693067778861                | 0.029880143280857863 |
|        2 |      1 | memory/g2-preemptible/us-central1/1693067778861      |  0.01879407184166667 |
|        2 |      1 | accelerator/l4-preemptible/us-central1/1693110803273 |   0.1797683423080672 |
+----------+--------+------------------------------------------------------+----------------------+
6 rows in set (0.00 sec)
```

```
mysql> SELECT jobs.batch_id, jobs.job_id, cost_t.*
    -> FROM jobs
    -> LEFT JOIN LATERAL (
    ->    SELECT COALESCE(SUM(`usage` * rate), 0) AS cost, JSON_OBJECTAGG(resources.resource, COALESCE(`usage` * rate, 0)) AS cost_breakdown
    ->    FROM (
    ->      SELECT batch_id, job_id, resource_id, CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
    ->      FROM aggregated_job_resources_v2
    ->      WHERE jobs.batch_id = aggregated_job_resources_v2.batch_id AND jobs.job_id = aggregated_job_resources_v2.job_id
    ->      GROUP BY batch_id, job_id, resource_id
    ->    ) AS usage_t
    ->    LEFT JOIN resources ON usage_t.resource_id = resources.resource_id
    ->    GROUP BY batch_id, job_id
    ->  ) AS cost_t ON TRUE
    -> WHERE jobs.batch_id = 2 AND jobs.job_id = 1;
+----------+--------+--------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| batch_id | job_id | cost               | cost_breakdown                                                                                                                                                                                                                                                                                                                                                              |
+----------+--------+--------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|        2 |      1 | 0.3183202707639251 | {"ip-fee/1024/1": 0.004279891111111111, "service-fee/1": 0.042798911111111115, "disk/pd-ssd/us-central1/1693067778861": 0.029880143280857863, "memory/g2-preemptible/us-central1/1693067778861": 0.01879407184166667, "compute/g2-preemptible/us-central1/1693067778861": 0.042798911111111115, "accelerator/l4-preemptible/us-central1/1693110803273": 0.1797683423080672} |
+----------+--------+--------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> SELECT jobs.batch_id, jobs.job_id, cost_t.*
    -> FROM jobs
    -> LEFT JOIN LATERAL (
    ->    SELECT COALESCE(SUM(`usage` * rate), 0) AS cost, JSON_OBJECTAGG(resources.resource, COALESCE(`usage` * rate, 0)) AS cost_breakdown
    ->    FROM (
    ->      SELECT batch_id, job_id, resource_id, CAST(COALESCE(SUM(`usage`), 0) AS SIGNED) AS `usage`
    ->      FROM aggregated_job_resources_v2
    ->      WHERE jobs.batch_id = aggregated_job_resources_v2.batch_id AND jobs.job_id = aggregated_job_resources_v2.job_id
    ->      GROUP BY batch_id, job_id, resource_id
    ->    ) AS usage_t
    ->    LEFT JOIN resources ON usage_t.resource_id = resources.resource_id
    ->    GROUP BY batch_id, job_id
    ->  ) AS cost_t ON TRUE
    -> WHERE jobs.batch_id = 5 AND jobs.job_id = 1;
+----------+--------+--------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| batch_id | job_id | cost               | cost_breakdown                                                                                                                                                                                                                                                                                                                                                             |
+----------+--------+--------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|        5 |      1 | 0.7269959944620336 | {"ip-fee/1024/1": 0.003748218888888889, "service-fee/1": 0.03748218888888889, "disk/pd-ssd/us-central1/1693067778861": 0.02616826328998403, "memory/g2-nonpreemptible/us-central1/1693067778861": 0.0411476845869, "compute/g2-nonpreemptible/us-central1/1693067778861": 0.09366128821796, "accelerator/l4-nonpreemptible/us-central1/1693110803273": 0.5247883505894118} |
+----------+--------+--------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

batch/batch/cloud/gcp/driver/create_instance.py
```diff
@@ -128,7 +129,7 @@ def scheduling() -> dict:
                 'boot': True,
                 'autoDelete': True,
                 'initializeParams': {
-                    'sourceImage': f'projects/{project}/global/images/batch-worker-13',
+                    'sourceImage': f'projects/{project}/global/images/batch-worker-parsa-100',
```
jigold, line 132: @daniel-goldstein or I will need to actually build `batch-worker-14` and you'll need to change this to `batch-worker-14` when we're done.

Also, reminder for Daniel and I that we need to let the Australians create their own boot disk image before merging this PR.

batch/batch/cloud/gcp/driver/create_instance.py
```diff
@@ -43,7 +43,8 @@ def create_vm_config(
     _, cores = gcp_machine_type_to_worker_type_and_cores(machine_type)
 
     region = instance_config.region_for(zone)
-
+    accelerator_family = machine_type.split("-")[0]
+    docker_run_gpu_args = "--runtime=nvidia --gpus all" if machine_type_to_gpu(accelerator_family) else ""
```
jigold, line 47: Can you use single quotes? Sorry for being nit picky. I think the entire rest of the code base is almost entirely single quotes.

batch/batch/cloud/gcp/driver/pricing.py
```diff
@@ -183,12 +318,53 @@ def process_compute_sku(sku: dict, regions: List[str]) -> List[GCPComputePrice]:
...
+        and pricing_expression['baseUnitConversionFactor'] == 3600
+    ), sku
+    cost_per_hour = pricing_expression_to_price_per_unit(pricing_expression)
+
+    compute_prices = []
+    service_regions = sku['serviceRegions']
+    for service_region in service_regions:
+        if service_region in regions:
+            if accelerator_family == 'l4':
+                log.info(f'SKUID {sku["skuId"]} with description {sku["description"]}')
```
jigold, line 358: I think you can remove this log statement.

batch/batch/cloud/gcp/driver/pricing.py
```diff
@@ -212,6 +388,9 @@ def process_memory_sku(sku: dict, regions: List[str]) -> List[GCPMemoryPrice]:
     service_regions = sku['serviceRegions']
     for service_region in service_regions:
         if service_region in regions:
+            if instance_family == 'g2':
+                log.info(f'SKUID {sku["skuId"]} with description {sku["description"]}')
```
jigold, line 392: remove this log statement.

batch/batch/cloud/gcp/driver/zones.py
```diff
@@ -89,11 +125,22 @@ def choose_location(
...
 
         zones = [zw.zone for zw in zone_weights]
 
+        machine_family = machine_type.split("-")[0]
+        if MACHINE_FAMILY_VALID_ZONES.get(machine_family):
+            valid_indices = []
+            for i, zone in enumerate(zones):
+                if zone in MACHINE_FAMILY_VALID_ZONES[machine_family]:
+                    valid_indices.append(i)
+                    log.info(f'adding {zone} as a valid zone for {machine_family}')
```
jigold, line 140: remove log statement.

batch/batch/cloud/gcp/driver/zones.py
```diff
@@ -11,6 +11,42 @@
 
 log = logging.getLogger('zones')
 
+MACHINE_FAMILY_VALID_ZONES = {
```
jigold, line 14: Is there a way to scrape this information automatically? if not, then I'd like to  (1) make an issue to figure out how to do this automatically and (2) remove the "n1" hard coding. Otherwise, we're going to break the Australian's deployment.

batch/batch/cloud/gcp/driver/zones.py
```diff
@@ -89,11 +125,22 @@ def choose_location(
...
+        machine_type,
     ) -> str:
         zone_weights = self.compute_zone_weights(cores, local_ssd_data_disk, data_disk_size_gb, preemptible, regions)
 
         zones = [zw.zone for zw in zone_weights]
 
+        machine_family = machine_type.split("-")[0]
+        if MACHINE_FAMILY_VALID_ZONES.get(machine_family):
+            valid_indices = []
+            for i, zone in enumerate(zones):
```
jigold, line 137: I think you can write this code more simply as 
```python3
if machine_family in MACHINE_FAMILY_VALID_ZONES:
    valid_zones = MACHINE_FAMILY_VALID_ZONES[machine_family]
    zones = [zw for zw in zone_weights if zw.zone in valid_zones]
```

batch/batch/cloud/gcp/resource_utils.py
```diff
@@ -8,6 +8,7 @@
 MACHINE_TYPE_REGEX = re.compile('(?P<machine_family>[^-]+)-(?P<machine_type>[^-]+)-(?P<cores>\\d+)')
 GCP_MACHINE_FAMILY = 'n1'
 
+GPU_VERSIONS = {"g2": "l4"}
```
jigold, line 11: Is this `MACHINE_TYPE_TO_ACCELERATOR_VERSIONS`? What is the difference between a gpu and an accelerator? If they're the same, can we choose one or the other?

batch/batch/cloud/gcp/driver/pricing.py
```diff
@@ -268,12 +447,43 @@ def process_disk_sku(sku: dict, regions: List[str]) -> List[GCPDiskPrice]:
...
+    params = {'currencyCode': currency_code}
+    g2_ram_total = set()
+    g2_cpu_total = set()
+    gpu_total = set()
+    async for sku in billing_client.list_skus('/6F81-5844-456A/skus', params=params):
+        category = sku['category']
+        if (
+            sku['description'].startswith("G2 Instance")
+            or sku['description'].startswith("Spot Preemptible G2 Instance")
+        ) and "us-" in sku['serviceRegions'][0]:
```
jigold, line 461: What is the reason for limiting to only the `us`? Applies below as well.

batch/batch/cloud/gcp/driver/pricing.py
```diff
@@ -17,6 +23,99 @@
...
+    # G2 Instance Core running in Dallas
+    'B911-29A9-DC06',
+    # G2 Instance Core running in Las Vegas
+    '29FF-9F02-F5CE',
+]
+
+G2_RAM_SKUS = [
+    # G2 Instance Ram running in Virginia
+    'A15F-B56F-97FE',
+    # G2 Instance Ram running in Iowa
```
jigold, line 60: Can you explain to me why you had to hard code these?

batch/batch/worker/worker.py
```diff
@@ -1171,6 +1189,7 @@ async def container_config(self):
                 'uidMappings': [],
                 'gidMappings': [],
                 'resources': {
+                    "devices": [{"allow": True, "type": "c", "major": 195, "minor": 0, "access": "rwm"}],
```
jigold, line 1192: Can you add a comment here on what this is doing or maybe a link to documentation?

batch/batch/worker/worker.py
```diff
@@ -1334,6 +1353,10 @@ def _env(self):
         env = (
             self.image.image_config['Config']['Env'] + self.env + CLOUD_WORKER_API.cloud_specific_env_vars_for_user_jobs
         )
+        gpu_env = ["NVIDIA_VISIBLE_DEVICES=all"]
+        accelerator_family = INSTANCE_CONFIG["machine_type"].split("-")[0]
+        if machine_type_to_gpu(accelerator_family):
```
jigold, line 1358: Can we add a new function to `resource_utils.py` and use that here and above:

```python3
def is_gpu(accelerator_family: str) -> bool:
    return machine_type_to_gpu(accelerator_family) is not None
```

batch/build-batch-worker-image-startup-gcp.sh
```diff
@@ -41,12 +41,35 @@ curl -fsSL "https://github.com/GoogleCloudPlatform/docker-credential-gcr/release
   | tar xz --to-stdout ./docker-credential-gcr \
 	> /usr/bin/docker-credential-gcr && chmod +x /usr/bin/docker-credential-gcr
 
+curl -fsSL https://nvidia.github.io/libnvidia-container/gpgkey \
+     | gpg --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg
+curl -s -L https://nvidia.github.io/libnvidia-container/ubuntu22.04/libnvidia-container.list | \
```
jigold, line 46: Can you explain to me why we need the `nvidia-container-toolkit` in both the host boot disk image and in the worker container?

infra/gcp-broad/gcp-ar-cleanup-policy.txt
```diff
@@ -114,6 +114,7 @@
                 "create_certs_image",
                 "curl",
                 "git-make-bash",
+                "gpu",
```
jigold, line 117: Reminder for me when this merges to update the policy with `gcloud`.

batch/batch/cloud/gcp/driver/pricing.py
```diff
@@ -268,12 +447,43 @@ def process_disk_sku(sku: dict, regions: List[str]) -> List[GCPDiskPrice]:
...
+                g2_cpu_total.add(sku["skuId"])
+            elif 'Ram' in sku['description']:
+                g2_ram_total.add(sku["skuId"])
+        if (
+            sku['description'].startswith("Nvidia L4 GPU")
+            and 'GPU' in category['resourceGroup']
+            and "us-" in sku['serviceRegions'][0]
+        ):
+            gpu_total.add(sku["skuId"])
+    log.info(f'g2_cpu_total {g2_cpu_total}')
```
jigold, line 472: Can you remove these log statements? Generally we try and keep the number of log messages to a minimum.

batch/batch/cloud/gcp/driver/pricing.py
```diff
@@ -43,6 +142,33 @@ def rate(self):
...
+        self.effective_start_date = effective_start_date
+        self.effective_end_date = effective_end_date
+
+    @property
+    def product(self):
+        return GCPAcceleratorResource.product_name(self.accelerator_family, self.preemptible, self.region)
+
+    @property
+    def rate(self):
+        # return rate_cpu_hour_to_mcpu_msec(self.cost_per_hour)
```
jigold, line 168: delete.

batch/Dockerfile.worker
```diff
@@ -19,6 +20,14 @@ RUN echo "APT::Acquire::Retries \"5\";" > /etc/apt/apt.conf.d/80-retries && \
     curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add - && \
     hail-apt-get-install fuse gcsfuse=0.41.12
 
+RUN curl -fsSL https://nvidia.github.io/libnvidia-container/gpgkey | gpg --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg && \
```
jigold, line 23: Can you add a comment explaining why the `nvidia-container-toolkit` is necessary here as well as on the host via the boot disk image?

jigold (changes requested):
Great job Sophie! I made a first pass with some comments. Please let me know if you have any questions.

jigold:
I can make a follow up PR, but it occurred to me that we need a check in `_create_jobs()` in the front end to make sure if a user requests a machine with a GPU, the requested regions are subsetted to only the regions that support the GPU.

sjparsa:
Currently the` nvidia-container-toolkit` is installed both in the vm startup script and in the worker docker image. The toolkit must be installed in the startup script to be able to configure docker with the command `nvidia-ctk runtime configure --runtime=docker`. This command cannot be run from Dockerfile.worker because it gets the error `"unable to flush config: unable to open /etc/docker/daemon.json for writing: open /etc/docker/daemon.json: no such file or directory"`.
The toolkit also has to be installed in Dockerfile.worker since that is where crun is invoked from. To execute the nvidia hook, the toolkit needs to be installed in that container. We could probably find a workaround to this if you would like but it only increased the worker image from 1.47Gb to 1.55Gb so it seems pretty small.

batch/batch/cloud/gcp/driver/pricing.py
jigold: With this assert, you'll know you're picking up thing that aren't what you intend. However, it's not going to help with start missing updates to those intended products. It will just not see an update and not update the database. If you want that behavior, then you'll want to keep track of the SKUs you've seen and make sure that list is the set of all SKUs you expected. If the list is different, then don't throw an exception. Write a new log message with the missing updates `log.exception(f'missing g2 skus {expected_g2_skus.difference(seen_g2_skus)}')`. Throwing an exception will just kill the batch driver entirely, which is not what we want. We just want the alerts to start firing with that error message and then we can respond appropriately.

batch/batch/cloud/gcp/driver/pricing.py
sjparsa: I think maybe we will only roll out GPUs in the US for now and not in the Australia deploy

batch/batch/cloud/gcp/resource_utils.py
sjparsa: They are the same I believe, I will update to use accelerator when referencing the GPU itself

batch/batch/worker/worker.py
cseed: The documentation of the `linux.resource.devices` field is here: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#devices.

Loosely, we need to specify in the config the (GPU) devices that we want to make available to the container.  Major mode 195 for Nvidia GPUs, see:

From https://www.kernel.org/doc/Documentation/admin-guide/devices.txt:

```
195 char	Nvidia graphics devices
		  0 = /dev/nvidia0		First Nvidia card
		  1 = /dev/nvidia1		Second Nvidia card
		    ...
		255 = /dev/nvidiactl		Nvidia card control device
```

However, there is still some behavior to the devices field we don't understand.  In particular, if we change this to device 5,0 /dev/tty, the container is still able to access the GPU!  (But if set `"devices": []` or omit the devices field, then the container cannot, as expected).

This appears to be related to cgroup device white/blacklists and default device permissions, which I'm still researching.

batch/build-batch-worker-image-startup-gcp.sh
```diff
@@ -5,6 +5,10 @@ set -ex
 curl --silent --show-error --remote-name --fail https://dl.google.com/cloudagents/add-logging-agent-repo.sh
 bash add-logging-agent-repo.sh
 
+"""
```
jigold, line 8: You need comments here "#". This is a sh script, not python.

batch/batch/cloud/gcp/driver/pricing.py
cseed: OK, this might be a non-issue, I don't totally understand the pricing/billing code.

GCP products are identified by matching on the description (and other fields).  For example, from the code below:

```
            sku['description'].startswith("G2 Instance")
            or sku['description'].startswith("Spot Preemptible G2 Instance")
```

What happens if Google adds a new product that satisfies this condition?  Then we will generate duplicate resources (potentially with different prices).  There are already GCP products that have similar descriptions (e.g. Sole Tenant G2 instances).  Therefore, we decided to hardcode the expected GCP SKUs.  These are only used in assertions.

`_refresh_resources_from_retail_prices` would log an error in this case.  I'd be slightly uncomfortable to have the driver running with wrong/corrupted pricing and I would probably prefer a failure and be forced to figure out what's wrong.

Actually, there might not even be an error: if two GCP SKUs collide with the same resource, and one of them has a newer version (start date), the product might be inserted as a new version with no errors (but it should generate an error on the next restart).

If you'd prefer we can back this part out and have the new code match the existing style.  If this is a real concern, it might make sense to address it in a separate PR since it isn't GPU specific.

batch/batch/cloud/gcp/driver/zones.py
cseed: Yeah, there's API calls to get all the machine types:

https://cloud.google.com/compute/docs/reference/rest/v1/machineTypes/aggregatedList?hl=en

which includes a zone field and a list of accelerators.  We can update the code to use that.

batch/batch/cloud/gcp/driver/pricing.py
jigold: I made a PR that will ensure that we don't update a product with the information of a different SKU then what was originally intended. Unfortunately, I am out on vacation the rest of the week. Feel free to adopt this and close https://github.com/hail-is/hail/pull/13607 if it will help you feel confident in your changes without needing the hard coded skus and the asserts. Otherwise, I can follow up with this change next week and you should keep the existing code that you are comfortable with including the hard coded skus and the asserts.

batch/batch/worker/worker.py
```diff
@@ -1171,6 +1189,7 @@ async def container_config(self):
                 'uidMappings': [],
                 'gidMappings': [],
                 'resources': {
+                    "devices": [{"allow": False, "access": "rwm"}],
```
cseed, line 1192: A short explanation for this:

 - The Nvidia container toolkit hook fails when devices is empty (or missing).  I'm not quite sure why.  I couldn't get crun or the hook to generate an error message (it just has exit code 1).
 - Devices seem to be managed by a allow/deny list.  This now denies all devices by default.  It isn't completely clear what the default behavior was.  The runtime adds some default, required devices (/dev/null, /dev/tty, etc.)  I believe the container runtime hook (based on NVIDIA_VISIBLE_DEVICES set below) finds and attaches the Nvidia GPUs.

batch/batch/worker/worker.py
```diff
@@ -1334,6 +1353,10 @@ def _env(self):
         env = (
             self.image.image_config['Config']['Env'] + self.env + CLOUD_WORKER_API.cloud_specific_env_vars_for_user_jobs
         )
+        gpu_env = ["NVIDIA_VISIBLE_DEVICES=all"]
+        machine_family = INSTANCE_CONFIG["machine_type"].split("-")[0]
+        if is_gpu(machine_family):
+            env += gpu_env
```
danking, line 1359: inline the definition of gpu_env on line 1359.

batch/build-batch-worker-image-startup-gcp.sh
```diff
@@ -41,12 +44,35 @@ curl -fsSL "https://github.com/GoogleCloudPlatform/docker-credential-gcr/release
...
+apt-get --yes install nvidia-container-toolkit
+nvidia-ctk runtime configure --runtime=docker
+
 # avoid "unable to get current user home directory: os/user lookup failed"
 export HOME=/root
 
 docker-credential-gcr configure-docker --include-artifact-registry
 docker pull {{ global.docker_root_image }}
 
+systemctl restart docker
```
danking, line 74: what's the motivation for this?

batch/gcp-create-worker-image.sh
```diff
@@ -36,7 +36,7 @@ create_build_image_instance() {
     gcloud -q compute instances create ${BUILDER} \
         --project ${PROJECT}  \
         --zone=${ZONE} \
-        --machine-type=n1-standard-1 \
+        --machine-type=n1-standard-4 \
```
danking, line 39: is this an intentional change? cc: @daniel-goldstein

danking (dismissed):
This looks great, a few nits.

batch/build-batch-worker-image-startup-gcp.sh
cseed: That was recommended by the Nvidia container toolkit installation instructions.  While it isn't strictly necessary since we're about to shut down, it seems better to keep it in case additional commands that use docker get added at the end of the startup script.

Either way, it should probably go after updating /etc/docker/daemon.json below.

batch/gcp-create-worker-image.sh
daniel-goldstein: Not that I'm aware of.

sjparsa: Oh yeah I had changed this when my startup script was taking a long time to build to see if it helped. The long build ended up being because of an error so I'll revert this

jigold:
Not sure where this PR is at. @danking I was originally hoping we could add a check on the front end during job creation to make sure the machine type is supported by the requested regions. Is it okay if that's a follow-up PR that I make later on akin to #13607?

danking:
@jigold Do I understand correctly that g2-standard-4 can only be created as a job private instance because there is no matching pool? If that's right, it looks like, unlike pool jobs, a JPIM job [will be correctly marked as error](https://github.com/hail-is/hail/blob/main/batch/batch/driver/instance_collection/job_private.py#L457-L467) if there are no available regions.

Assuming all that is correct, do I also understand correctly that the only reason to block incoming jobs at the front-end is for a better user experience, not to protect the system from bad data? If yes, then I agree that need not be part of this PR because it merely improves user experience rather than being critical for correct functioning of the system.

Are any of my assumptions or inferences wrong?

batch/batch/cloud/gcp/resource_utils.py
```diff
@@ -17,6 +18,7 @@
 
 
 gcp_valid_machine_types = []
+gcp_valid_machine_types.append("g2-standard-4")
```
danking, line 21: @sjparsa, I'll approve anyway but can you follow up with a nit PR that just changes this to `gcp_valid_machine_types = ['g2-standard-4']`?

infra/gcp-broad/gcp-ar-cleanup-policy.txt
danking: @jigold can you PR a change to add a build step that applies the gap ar cleanup policy on deploys?

danking (approved):

batch/batch/cloud/azure/driver/regions.py
```diff
@@ -24,6 +24,7 @@ def choose_location(
         data_disk_size_gb: int,  # pylint: disable=unused-argument
         preemptible: bool,  # pylint: disable=unused-argument
         regions: List[str],
+        machine_type,
```
jigold, line 27:
```suggestion
        machine_type: str,
```

batch/batch/cloud/gcp/driver/zones.py
```diff
@@ -89,10 +93,16 @@ def choose_location(
         data_disk_size_gb: int,
         preemptible: bool,
         regions: List[str],
+        machine_type,
```
jigold, line 96:
```suggestion
        machine_type: str,
```

batch/batch/cloud/gcp/resource_utils.py
```diff
@@ -8,6 +8,7 @@
 MACHINE_TYPE_REGEX = re.compile('(?P<machine_family>[^-]+)-(?P<machine_type>[^-]+)-(?P<cores>\\d+)')
 GCP_MACHINE_FAMILY = 'n1'
 
+MACHINE_FAMILY_TO_ACCELERATOR_VERSIONS = {"g2": "l4"}
```
jigold, line 11:
```suggestion
MACHINE_FAMILY_TO_ACCELERATOR_VERSIONS = {'g2': 'l4'}
```

jigold (commented):
Great job! Feel free to apply the nits in a different follow up PR with Dan's nit.

batch/batch/cloud/gcp/driver/zones.py
```diff
@@ -89,10 +93,16 @@ def choose_location(
...
     ) -> str:
         zone_weights = self.compute_zone_weights(cores, local_ssd_data_disk, data_disk_size_gb, preemptible, regions)
 
         zones = [zw.zone for zw in zone_weights]
+        machine_family = machine_type.split("-")[0]
+        if machine_family in self._machine_family_valid_zones:
```
jigold, line 102: Actually, we should change this behavior to always assume that the machine_family is always a key in `self._machine_family_valid_zones`. If it isn't, then we should set zones and zone_weights to be equal to `[]`.

jigold (dismissed):
See comment.

jigold:
@sjparsa This looks great! Go ahead and merge at your convenience.

PR #13434, opened 2023-08-15
[fs] fix bugs in glob and list in our GSFS

danking:
CHANGELOG: In QoB, Hail's file systems now correctly list all files in a directory, not just the first 1000. This could manifest in an import_table or import_vcf which used a glob expression. In such a case, only the first 1000 files would have been included in the resulting Table or MatrixTable.

I also moved two GSFS-only tests into the FSSuite. There should be very few tests that are cloud-specific.


jigold (changes requested):

jigold:
There's a bug somewhere in your code in the test suite.

jigold (approved):

jigold:
Can you add a CHANGELOG for a bug fix for this?

jigold:
I put the WIP tag on so it won't merge until you change the commit message.

danking:
Good catch. I did a bit of digging to confirm the default max results too.

PR #13435, opened 2023-08-15
[aiogoogle] Fix bucket name parsing

jigold:
This code was broken if the URL only had a bucket name like `gs://foo`. It truncated the last character of the bucket because the `find` operation returned -1 for not being able to find another `/` in the URL after `gs://`.

### DIFF
--- a/hail/python/hailtop/aiocloud/aiogoogle/client/storage_client.py
+++ b/hail/python/hailtop/aiocloud/aiogoogle/client/storage_client.py
@@ -617,9 +617,10 @@ class GoogleStorageAsyncFS(AsyncFS):
         end_of_bucket = rest.find('/', 2)
         if end_of_bucket != -1:
             bucket = rest[2:end_of_bucket]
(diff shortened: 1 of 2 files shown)


daniel-goldstein (approved):

danking:
@jigold how did you stumble across this?

jigold:
When I was writing `hailctl batch init`, I used that method to parse the remote_tmpdir to check if the bucket existed and to add permissions. I kept getting errors because my remote_tmpdir was `gs://hail-jigold-random-token` and I didn't specify a path name beyond the bucket name.

jigold:
The bucket name was truncated by 1 character. Took me a bit to figure out that was the issue.

danking:
Nice! I think we should add some error checking code that verifies the name is non-empty when we, for example, `create`. It's unlikely to happen but would create some confusing situations if a user creates a file with an empty name. I'm not even sure if it's allowed in the API.

PR #13437, opened 2023-08-15
[lint] Swap mypy for pyright

daniel-goldstein:
`pyright` is an alternative to `mypy` for type checking python. It's also a pretty solid language server. While they both adhere to the PEPs for type checking, they [have a few differences](https://github.com/microsoft/pyright/blob/main/docs/mypy-comparison.md#differences-between-pyright-and-mypy). That doc is worth a read, but the parts that I found most compelling are pyright's return-type inference that mypy doesn't do and that pyright differentiates between an `Unknown` type and an `Any` type. `mypy`, for the most part, treats variables as `Any` if it can't figure out what type they are, and as a result lets a lot of behaviors on `Any` variables slide. `pyright` tends to be much stricter (even when not on strict mode!) and while some of the assertions that I had to make here to appease `pyright` are a bit noisy, it's also pointed out a lot of areas where we can be stricter with our typing and then clean up code because some assertions are caught instead by the type system. It also did a good job informing of bad import practices like how `hl.utils` was not actually exported by the `hail` module but incidentally imported by some file in the repo.


daniel-goldstein:
The `pre-commit` hook is a little sticky because `pre-commit` installs each tool in its own isolated virtual env, which won't have the dependencies unless we tell `pre-commit` to also install all of our pinned dependencies into the pyright virtualenv. We can configure pyright to use a different virtualenv for all its dependencies, but that would require each developer specifying the name of their virtual environment in `pyproject.toml`.

daniel-goldstein:
Hm does `sphinx` not support `typing.Self`?

danking:
I'm not that surprised. are we on latest Sphinx?

daniel-goldstein:
Sphinx was trying to link to a class in the python docs for `typing.Self`. I'm not sure whether this is a 3.9 problem but I can't even find something that looks like a class definition [in the 3.11 docs](https://docs.python.org/3/library/typing.html). Given that, and that I don't think it's actually very helpful/desirable to link `typing` classes, I just ignore trying to find a link for `Self` and docs build successfully.

danking:
Agreed Self doesn't seem useful to link to. It does seem to exist: https://docs.python.org/3/library/typing.html#typing.Self . I'll plan to review today/tomorrow!

danking:
Ah you mean that it doesn't look like a *class* definition. Sure

danking (approved):
No objections from me, looks good from a quick scan.

PR #13440, opened 2023-08-15
[batch] update ubuntu image version to latest

sjparsa:
Upgrade to Ubuntu 22.04 everywhere including for the Batch Worker VMs in Azure and Google.

### DIFF
--- a/batch/Dockerfile.worker
+++ b/batch/Dockerfile.worker
@@ -25,7 +25,7 @@ RUN hail-apt-get-install ca-certificates pkg-config libfuse-dev cmake libcurl4-g
     curl -LO https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb && \
     dpkg -i packages-microsoft-prod.deb && \
     apt-get update && \
-    hail-apt-get-install blobfuse
+    hail-apt-get-install blobfuse2
 
 {% else %}
 RUN echo "!!! UNEXPECTED CLOUD {{global.cloud}} !!!" && exit 1
--- a/batch/Dockerfile.worker
+++ b/batch/Dockerfile.worker
@@ -22,7 +22,7 @@ RUN echo "APT::Acquire::Retries \"5\";" > /etc/apt/apt.conf.d/80-retries && \
 {% elif global.cloud == "azure" %}
 # https://github.com/Azure/azure-storage-fuse/issues/603
 RUN hail-apt-get-install ca-certificates pkg-config libfuse-dev cmake libcurl4-gnutls-dev libgnutls28-dev uuid-dev libgcrypt20-dev && \
-    curl -LO https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb && \
+    curl -LO https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb && \
     dpkg -i packages-microsoft-prod.deb && \
     apt-get update && \
     hail-apt-get-install blobfuse
--- a/batch/Dockerfile.worker
+++ b/batch/Dockerfile.worker
@@ -20,6 +20,9 @@ RUN echo "APT::Acquire::Retries \"5\";" > /etc/apt/apt.conf.d/80-retries && \
     hail-apt-get-install fuse gcsfuse=0.41.12
 
 {% elif global.cloud == "azure" %}
+RUN apt-get update && \
+    hail-apt-get-install libreadline8
+
 # https://github.com/Azure/azure-storage-fuse/issues/603
 RUN hail-apt-get-install ca-certificates pkg-config libfuse-dev cmake libcurl4-gnutls-dev libgnutls28-dev uuid-dev libgcrypt20-dev && \
     curl -LO https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb && \
--- a/batch/batch/cloud/azure/driver/create_instance.py
+++ b/batch/batch/cloud/azure/driver/create_instance.py
@@ -271,7 +271,6 @@ docker run \
 -v /var/run/docker.sock:/var/run/docker.sock \
 -v /var/run/netns:/var/run/netns:shared \
 -v /usr/bin/docker:/usr/bin/docker \
--v /usr/sbin/xfs_quota:/usr/sbin/xfs_quota \
 -v /batch:/batch:shared \
 -v /logs:/logs \
 -v /global-config:/global-config \
--- a/batch/batch/cloud/azure/driver/create_instance.py
+++ b/batch/batch/cloud/azure/driver/create_instance.py
@@ -277,7 +277,7 @@ docker run \
 -v /deploy-config:/deploy-config \
 -v /cloudfuse:/cloudfuse:shared \
 -v /etc/netns:/etc/netns \
--v /sys/fs/cgroup:/sys/fs/cgroup:shared \
+-v /sys/fs/cgroup:/sys/fs/cgroup \
 --mount type=bind,source=/mnt/disks/$WORKER_DATA_DISK_NAME,target=/host \
 --mount type=bind,source=/dev,target=/dev,bind-propagation=rshared \
 -p 5000:5000 \
@@ -288,6 +288,7 @@ docker run \
 --cap-add SYS_ADMIN \
 --security-opt apparmor:unconfined \
 --network host \
+--cgroupns host \
 $BATCH_WORKER_IMAGE \
 python3 -u -m batch.worker.worker >worker.log 2>&1
 
--- a/batch/batch/cloud/azure/driver/create_instance.py
+++ b/batch/batch/cloud/azure/driver/create_instance.py
@@ -288,7 +288,6 @@ docker run \
 --cap-add SYS_ADMIN \
 --security-opt apparmor:unconfined \
 --network host \
---cgroupns host \
 $BATCH_WORKER_IMAGE \
 python3 -u -m batch.worker.worker >worker.log 2>&1
 
--- a/batch/batch/cloud/azure/driver/create_instance.py
+++ b/batch/batch/cloud/azure/driver/create_instance.py
@@ -426,7 +426,7 @@ done
                 'imageReference': {
                     'value': {
                         'id': f'/subscriptions/{subscription_id}/resourceGroups/{resource_group}/providers/'
-                        f'Microsoft.Compute/galleries/{resource_group}_batch/images/batch-worker/versions/0.0.13'
+                        f'Microsoft.Compute/galleries/{resource_group}_batch/images/batch-worker-22-04/versions/0.0.13'
                     }
                 },
                 'workspaceName': {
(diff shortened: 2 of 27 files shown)


danking:
@daniel-goldstein should this PR update line 38 of main.tf which uses Ubuntu 20.04 for the batch worker? Do we need to take special action if that happens?

batch/gcp-create-worker-image.sh
```diff
@@ -24,7 +24,7 @@ else
     BUILDER=build-batch-worker-$NAMESPACE-image
```
danking, line 24: @sjparsa , I don't see any dev docs about how to update a worker image. Did someone explain how to do that to you?

I confirmed that batch-worker-13 already exists:
```
(base) dking@wm28c-761 hail % gcloud compute images list --project hail-vdc
NAME                                                  PROJECT              FAMILY                             DEPRECATED  STATUS
...
batch-worker-13                                       hail-vdc                                                            READY
```

docker/third-party/images.txt
```diff
@@ -14,6 +14,6 @@ python:3.10-slim
 redis:6.0.6-alpine
 ubuntu:18.04
 ubuntu:19.04
-ubuntu:20.04
+ubuntu:22.04
 ubuntu:bionic-20200921
 ubuntu:focal-20221019
```
danking, line 19: Remove focal here as well. I think we can nix all the old ubuntu's. @daniel-goldstein @jigold objections to that?

danking (changes requested):
Looks great, just a couple questions about Azure, a possible more complete cleanup, and a question about docs.

docker/third-party/images.txt
jigold: I think we should leave `20.04`. The others are fine to nix.

batch/gcp-create-worker-image.sh
sjparsa: Yes Daniel talked to me about it! I asked him to make batch-worker-13 with the update so I think that should be why it's showing up? @daniel-goldstein

daniel-goldstein:
> @daniel-goldstein should this PR update line 38 of main.tf which uses Ubuntu 20.04 for the batch worker? Do we need to take special action if that happens?
> 
> If we do this we also need to change `infra/azure/create_bootstrap_vm.sh` which references `focal`.

@jigold Is this terraform resource needed in the gallery in order for `az-create-worker-image.sh` to run? It's unclear to me whether `az-create-worker-image.sh` references this shared image.

batch/gcp-create-worker-image.sh
danking: @daniel-goldstein can you also create PR adding a dev-doc for how to do this? TY!

daniel-goldstein:
> > @daniel-goldstein should this PR update line 38 of main.tf which uses Ubuntu 20.04 for the batch worker? Do we need to take special action if that happens?
> > If we do this we also need to change `infra/azure/create_bootstrap_vm.sh` which references `focal`.
> 
> @jigold Is this terraform resource needed in the gallery in order for `az-create-worker-image.sh` to run? It's unclear to me whether `az-create-worker-image.sh` references this shared image.

@jigold bump

jigold:
Just to make sure I understand correctly, are you asking whether this resource `--image Canonical:0001-com-ubuntu-server-focal:20_04-lts:latest` needs to be in Terraform? This is an image from the Azure Marketplace. I think you can just pick a different one and it will be fine.

daniel-goldstein:
No sorry I'm asking if [this](https://github.com/hail-is/hail/blob/3a8be2b37ec387cbf0664354148a66e5930f2f88/infra/azure/modules/batch/main.tf#L27-L40) is unused and can be deleted or whether we are using it somewhere.

jigold:
I don't think you can delete it. If you look in the portal here, you'll see it contains all of the boot disk images:
https://portal.azure.com/#@haildev.onmicrosoft.com/resource/subscriptions/22cd45fe-f996-4c51-af67-ef329d977519/resourceGroups/haildev/providers/Microsoft.Compute/galleries/haildev_batch/images/batch-worker/overview

You might need to make a new one image definition for 22.04 -- otherwise, I think you're going to delete our existing images.

daniel-goldstein:
> I don't think you can delete it. If you look in the portal here, you'll see it contains all of the boot disk images: https://portal.azure.com/#@haildev.onmicrosoft.com/resource/subscriptions/22cd45fe-f996-4c51-af67-ef329d977519/resourceGroups/haildev/providers/Microsoft.Compute/galleries/haildev_batch/images/batch-worker/overview
> 
> You might need to make a new, separate image definition for 22.04 -- otherwise, I think you're going to delete our existing images.

Ah, great thanks, this is helpful. I wasn't sure exactly how the script played with that resource. I'll add a new resource for 22.04 and switch over to using that. Once we're fully moved over, we'll follow up with a separate PR that removes the old image.

batch/gcp-create-worker-image.sh
```diff
@@ -24,7 +24,7 @@ else
...
 fi
 
-UBUNTU_IMAGE=ubuntu-minimal-2004-focal-v20230725
+UBUNTU_IMAGE=ubuntu-minimal-2204-jammy-v20230726
```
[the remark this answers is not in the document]
sjparsa: Yes just did!

danking (changes requested):
Dismiss when you want me to review.

infra/azure/modules/batch/main.tf
```diff
@@ -39,6 +39,23 @@ resource "azurerm_shared_image" "batch_worker" {
...
+  specialized         = false
+
+  hyper_v_generation = "V2"
+
+  identifier {
+    publisher = "Hail"
+    offer     = "BatchWorker"
+    sku       = "Ubuntu-22.04-LTS"
+  }
+}
```
daniel-goldstein, line 57: I think we had a misunderstanding of what the shared image definition is doing. An image definition does not tie itself to a particular base VM image like Ubuntu 20.04 or Ubuntu 22.04, it just defines a [logical grouping of concrete image versions](https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries?tabs=azure-cli#image-definitions) with some broad characteristics defined like whether the image versions are Linux or Windows and which hypervisor generation is needed to run them.

This `identifier` block allows us to define a canonical name for the image definition that we are creating, and AFAICT is not intended to reference an existing marketplace identifier. So the existing definition doesn't actually give us a Ubuntu 20.04 base VM image.

Where the base VM image is actually defined is where we create the VM in `az-create-worker-image.sh` with `az vm create` (this makes sense in hindsight because we don't actually reference the image definition in the `az vm create` call so how could it know to pick 20.04 vs 22.04?). In that call we have been using `--image UbuntuTLS`. I think this is Ubuntu 18.04, because that's what I get when running `lsb_release -a` on a production Azure VM. So in this PR I've changed that to be `Ubuntu2204` which seems to have done the trick. I think we still need a different image definition here though because Ubuntu 18.04 and 22.04 need different `hyper_v_generation`. Trying to save an Ubuntu 22.04 VM under the existing definition resulted in an error saying that the VM was incompatible with a V1 image definition.

batch/Dockerfile.worker
```diff
@@ -20,12 +20,15 @@ RUN echo "APT::Acquire::Retries \"5\";" > /etc/apt/apt.conf.d/80-retries && \
...
+    hail-apt-get-install libreadline8
+
 # https://github.com/Azure/azure-storage-fuse/issues/603
 RUN hail-apt-get-install ca-certificates pkg-config libfuse-dev cmake libcurl4-gnutls-dev libgnutls28-dev uuid-dev libgcrypt20-dev && \
-    curl -LO https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb && \
+    curl -LO https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb && \
     dpkg -i packages-microsoft-prod.deb && \
     apt-get update && \
-    hail-apt-get-install blobfuse
+    hail-apt-get-install blobfuse2
```
danking, line 31: huzzah, at last!

docker/third-party/images.txt
danking: bump: do we need to keep 20.04? Presumably we can kill 18.04 and 19.04.

danking (dismissed):
This is an excellent change. I'm glad we cleaned up the situation in Azure a bit.

docker/third-party/images.txt
daniel-goldstein: See above message, we should keep `ubuntu:20.04` until we migrate the `global-config` then we can delete it.

batch/batch/cloud/azure/driver/create_instance.py
```diff
@@ -289,6 +288,7 @@ def create_vm_config(
 --cap-add SYS_ADMIN \
 --security-opt apparmor:unconfined \
 --network host \
+--cgroupns host \
```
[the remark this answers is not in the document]
daniel-goldstein: Related to the above error.

batch/batch/cloud/azure/driver/create_instance.py
```diff
@@ -271,14 +271,13 @@ def create_vm_config(
...
 -v /usr/bin/docker:/usr/bin/docker \
--v /usr/sbin/xfs_quota:/usr/sbin/xfs_quota \
 -v /batch:/batch:shared \
 -v /logs:/logs \
 -v /global-config:/global-config \
 -v /deploy-config:/deploy-config \
 -v /cloudfuse:/cloudfuse:shared \
 -v /etc/netns:/etc/netns \
--v /sys/fs/cgroup:/sys/fs/cgroup \
+-v /sys/fs/cgroup:/sys/fs/cgroup:shared \
```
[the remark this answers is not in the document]
sjparsa: Ah unfortunately no I've had to do several database wipes since then so I don't have any of my old stuff..... should I recreate a run without it?

batch/batch/cloud/azure/driver/create_instance.py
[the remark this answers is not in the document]
sjparsa: Removing this results in errors such as `writing file /sys/fs/cgroup/batch-7-job-1-main/cgroup.procs: No such file or directory`

batch/batch/cloud/azure/driver/create_instance.py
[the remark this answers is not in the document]
sjparsa: Looks like the :shared wasn't necessary and so I removed it

batch/batch/cloud/azure/driver/create_instance.py
```diff
@@ -271,14 +271,13 @@ def create_vm_config(
 -v /var/run/docker.sock:/var/run/docker.sock \
 -v /var/run/netns:/var/run/netns:shared \
 -v /usr/bin/docker:/usr/bin/docker \
--v /usr/sbin/xfs_quota:/usr/sbin/xfs_quota \
```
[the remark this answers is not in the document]
danking: @daniel-goldstein  Alright, let's PR that change separately.

batch/batch/cloud/azure/driver/create_instance.py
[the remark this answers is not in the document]
danking: Ah, right, right. This is the worker's VM exposing cgroups to the batch-worker container. Perfect.

docker/third-party/images.txt
danking: And we'll delete 19.04 then too, right?

danking (approved):
LGTM, two pending threads with Daniel that need not block this PR.

PR #13449, opened 2023-08-16
[dev-docs] Update dev docs for using dev namespaces

daniel-goldstein:
cc: @sjparsa


dev-docs/development_process.md
```diff
@@ -76,63 +76,141 @@ export DOCKER_BUILDKIT=1
...
+
+To examine the Kubernetes logs for the service in your namespace, use something like
+
+```bash
 kubectl -n <my_namespace> logs -l app=batch-driver --tail=999999 | less
 ```
 
-To check the MySQL database, you first need to find the name of the specific pod
-running an “admin-pod” and then log into that pod:
+To check the MySQL database in your namespace, you can log in to the database
```
jigold, line 166: I think either here or somewhere else or as a todo item, we should talk about how to delete a test database like `username-batch` so it will be recreated when testing migrations. I know this was a known wrinkle.

dev-docs/development_process.md
```diff
@@ -76,63 +76,141 @@ export DOCKER_BUILDKIT=1
...
+```bash
+make -C batch deploy NAMESPACE=<my_namespace>
+```
+
+If you only want to make a Python code change to an already-running service
+in Kubernetes, you can run `devbin/sync.py` like the following:
+
+```bash
+python3 devbin/sync.py \
+    --namespace <my_namespace> \
```
jigold, line 201: Can you make sure this works now with `batch-driver`? I was having problems with it because there's 2 containers in the batch-driver pod.

I also have `--path web_common/web_common /usr/local/lib/python3.9/dist-packages/` and `--path hail/python/hailtop /usr/local/lib/python3.9/dist-packages/`.

jigold (changes requested):
this is great! didn't know it existed.

dev-docs/development_process.md
daniel-goldstein: Great point, maybe I'll follow up with a development FAQ? I don't want to make this getting started section too long but I feel like there's multiple points like this one that would be nice to have in one place.

dev-docs/development_process.md
daniel-goldstein: Ah ya, added those paths too.

And ya I ran into the problem with `batch-driver` too but #13374 should have fixed it.

dev-docs/development_process.md
```diff
@@ -76,63 +76,143 @@ export DOCKER_BUILDKIT=1
...
+    --app batch --app batch-driver \
+    --path batch/batch /usr/local/lib/python3.9/dist-packages/ \
+    --path gear/gear /usr/local/lib/python3.9/dist-packages/
```
jigold, line 204:
```suggestion
    --path gear/gear /usr/local/lib/python3.9/dist-packages/ \
```

dev-docs/development_process.md
jigold: sounds good.

jigold (dismissed):

jigold (approved):

PR #13450, opened 2023-08-16

danking:

Inspired by @jigold 's file system improvement campaign, I fell down the rabbit hole of not issuing "list" operations unless we really need to know if a file is a directory. This should partly help reduce the flakiness in Azure (which is tracked in #13351) as well as high costs in Azure both of which are at least partly suspected to be caused by our frequent use of "list" operations.


danking:
For the record, our Python-side implementation already does the right thing distinguishing a `FileStatus` from a `FileListEntry`.

PR #13452, opened 2023-08-16

danking:

Inspired by @jigold 's file system improvement campaign, I pursued the avoidance of "list" operations. I anticipate this reduces flakiness in Azure (which is tracked in #13351) and cost in Azure.

I enforced aiotools.fs terminology on hail.fs and Scala:

1. `FileStatus`. Metadata about a blob or file. It does not know if a directory exists at this path.

2. `FileListEntry`. Metadata from a list operation. It knows if a directory exists at this path.

Variable names were updated to reflect this distinction:

1. `fileStatus` / `fileStatuses`

2. `fle`/ `fles` / `fileListEntry` / `fileListEntries`, respectively.

`listStatus` renamed to `listDirectory` for clarity.

In both Azure and Google, `fileStatus` does not use a list operation.

`getFileListEntry` can be used when we must know if a directory exists. I just rewrote this from first principles because:
1. In neither Google nor Azure did it check if the path was a directory and a file.
2. In Google, if the directory entry wasn't in the first page, it would fail (NB: there are fifteen non-control characters in ASCII before `/`, if the page size is 15 or fewer, we'd miss the first entry with a `/` at the end).
3. In Azure, we issued both a get and a list.

There are now unit tests for this method.

---

1. `copyMerge` and `concatenateFiles` previously used `O(N_FILES)` list operations, they now use `O(N_FILES)` get operations.
2. Writers that used `exists` to check for a _SUCCESS file now use a get operation.
3. Index readers, import BGEN, and import plink all now check file size with a get operation.

That said, overall, the bulk of our Class A Operations are probably writes.


danking:
Still needs more work.

danking:
Alright I think I squashed everything.

danking:
Wow, thanks for nothing azure https://github.com/Azure/azure-sdk-for-java/issues/36674.

PR #13453, opened 2023-08-17
[aiotools] Log exception is task manager task errors

daniel-goldstein:


hail/python/hailtop/aiotools/tasks.py
```diff
@@ -21,9 +21,16 @@ def ensure_future(self, coroutine):
...
             raise TaskManagerClosedError
         t = asyncio.create_task(coroutine)
-        t.add_done_callback(lambda _: self.tasks.remove(t))
+        t.add_done_callback(self.on_task_done(t))
         self.tasks.add(t)
 
+    def on_task_done(self, t: asyncio.Task) -> Callable[[asyncio.Future], None]:
+        def callback(fut: asyncio.Future):
+            self.tasks.remove(t)
+            if e := fut.exception():
```
danking, line 30: I'm pretty sure if `fut` is cancelled this will raise a cancelled error. We should probably catch that, log it, and raise it?

danking (changes requested):

hail/python/hailtop/aiotools/tasks.py
danking: I don't know what happens when a callback raises.

Hmm. I suppose the callback has no reason to raise. There's no one awaiting us who should be cancelled.

danking (changes requested):
Agreed we shouldn't raise.

hail/python/hailtop/aiotools/tasks.py
```diff
@@ -21,9 +21,19 @@ def ensure_future(self, coroutine):
...
+        t.add_done_callback(self.on_task_done(t))
         self.tasks.add(t)
 
+    def on_task_done(self, t: asyncio.Task) -> Callable[[asyncio.Future], None]:
+        def callback(fut: asyncio.Future):
+            self.tasks.remove(t)
+            try:
+                if e := fut.exception():
+                    log.exception(e)
+            except asyncio.CancelledError:
```
danking, line 33: log the cancellation too?

danking (dismissed):

hail/python/hailtop/aiotools/tasks.py
daniel-goldstein: It's ok to be cancelled on worker shutdown, so I've only logged if that's not the case.

danking (approved):

PR #13456, opened 2023-08-18
[dev-docs] Instructions for resetting a dev database

daniel-goldstein:
This also fixes the currently broken `delete_*_tables` steps.

In the past, all dev and test databases shared the same MySQL Server (with production), so instead of each service getting its own dedicated database, there was one database per dev/test namespace with all the tables for all the services. This made it difficult to reset the database state of a particular service -- you needed to explicitly delete only the tables for that particular service.

Nowadays, dev and test databases live on their own MySQL Servers, so each service gets its own database (like in production). This makes it a lot easier to reset a service's database, we just drop the MySQL database for that service. This PR makes that change, deletes all the now unused `delete-*-tables.sql` files, and adds a dev doc explaining how to reset a dev database.

The reason these steps were broken is that the sql configs in dev/test namespaces use K8s DNS for the `host`, which does not work out of the box in batch jobs because they are not in the K8s network. There's code in `database.py` that uses the K8s API to resolve the database host to an IP address that the batch jobs can access. This is why I wrote a python script instead of just using `mysql`.

I tested these with the following dev deploy, which scrapped everything and I was able to log in after it was done!

```
hailctl dev deploy -b daniel-goldstein/hail:dev-ns-delete-db -s delete_auth_tables,delete_batch_tables,deploy_batch,add_developers
```

cc: @sjparsa, @iris-garden given your recent dev namespace woes


build.yaml
```diff
@@ -1842,47 +1852,57 @@ steps:
...
+      name: admin
+      namespace:
+        valueFrom: default_ns.name
     runIfRequested: true
     scopes:
       - dev
     dependsOn:
       - default_ns
-      - admin_pod_image
+      - batch_image
```
jigold, line 1877: Why does this depend on the batch_image? And ci_image below and monitoring_image above?

jigold (changes requested):

build.yaml
daniel-goldstein: Ah right, I forgot to mention. `admin-pod` doesn't have `gear` installed, only the mysql CLI client. I could have used `ci-utils-image` but I figured you're most likely to run these steps like `delete_batch_tables,deploy_batch` so better to use an image that is probably already in the pipeline.

jigold (approved):

PR #13458, opened 2023-08-18
[batch_client.aioclient] Get rid of BatchBuilder

jigold:
This PR refactors the aioclient to merge the functionality of Batch and BatchBuilder into just a single Batch object. The reason for making this change is to make adding job groups simpler. I will follow up with a change to Jobs after this merges. I apologize for the number of line changes. Most are just renaming `bb -> b` in the tests.

### DIFF
--- a/batch/test/test_batch.py
+++ b/batch/test/test_batch.py
@@ -938,11 +938,6 @@ def test_authorized_users_only():
         (session.get, '/api/v1alpha/batches/0', 401),
         (session.delete, '/api/v1alpha/batches/0', 401),
         (session.patch, '/api/v1alpha/batches/0/close', 401),
-        (session.get, '/api/v1alpha/batches/0/job_groups', 401),
-        (session.patch, '/api/v1alpha/batches/0/job_groups/1/cancel', 401),
-        (session.post, '/api/v1alpha/batches/0/job_groups', 401),
-        (session.get, '/api/v1alpha/batches/0/job_groups/1', 401),
-        (session.get, '/api/v1alpha/batches/0/job_groups/resources', 401),
         # redirect to auth/login
         (session.get, '/batches', 302),
         (session.get, '/batches/0', 302),
@@ -1355,14 +1350,14 @@ async def test_old_clients_that_submit_mount_docker_socket_false_is_ok(client: B
         }
         spec = {'always_run': False, 'job_id': 1, 'parent_ids': [], 'process': process}
         with pbar.with_task('submitting jobs', total=1) as pbar_task:
-            await b._submit_jobs(b.id, update_id, [orjson.dumps(spec)], 1, pbar_task)
+            await b._submit_jobs(update_id, [orjson.dumps(spec)], 1, pbar_task)
 
 
 async def test_old_clients_that_submit_mount_docker_socket_true_is_rejected(client: BatchClient):
     b = create_batch(client)._async_batch
     await b._open_batch()
     b.create_job(DOCKER_ROOT_IMAGE, command=['sleep', '30'])
-    update_id = await b._create_update(b.id)
+    update_id = await b._create_update()
     with BatchProgressBar() as pbar:
         process = {
             'type': 'docker',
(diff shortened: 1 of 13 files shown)


batch/test/test_batch.py
```diff
@@ -942,6 +938,11 @@ def test_authorized_users_only():
         (session.get, '/api/v1alpha/batches/0', 401),
         (session.delete, '/api/v1alpha/batches/0', 401),
         (session.patch, '/api/v1alpha/batches/0/close', 401),
+        (session.get, '/api/v1alpha/batches/0/job_groups', 401),
+        (session.patch, '/api/v1alpha/batches/0/job_groups/1/cancel', 401),
+        (session.post, '/api/v1alpha/batches/0/job_groups', 401),
+        (session.get, '/api/v1alpha/batches/0/job_groups/1', 401),
+        (session.get, '/api/v1alpha/batches/0/job_groups/resources', 401),
```
daniel-goldstein, line 945: Guessing this snuck in from a different branch?

batch/test/test_batch.py
```diff
@@ -1623,90 +1622,88 @@ def test_update_jobs_are_not_serialized(client: BatchClient):
...
     try:
         for _ in range(4):
-            bb.create_job(DOCKER_ROOT_IMAGE, ['echo', 'a' * (900 * 1024)])
-        b = bb.submit()
+            b.create_job(DOCKER_ROOT_IMAGE, ['echo', 'a' * (900 * 1024)])
+        b.submit()
     except httpx.ClientResponseError as err:
         assert err.status == 400
-        assert 'Cannot submit new jobs to a cancelled batch' in err.body
+        assert 'bunch contains job where the job group has already been cancelled' in err.body
```
daniel-goldstein, line 1677: Same with these messages about job groups.

hail/python/hailtop/batch_client/client.py
```diff
@@ -104,14 +104,23 @@ def attempts(self):
 
 
 class Batch:
+    _async_batch: aioclient.Batch
+
     @classmethod
     def from_async_batch(cls, batch: aioclient.Batch):
         b = object.__new__(cls)
         b._async_batch = batch
         return b
```
daniel-goldstein, line 113: Adding `_async_batch` as a class variable in this way will [share the object between different Batch instances](https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables). I think `_async_batch` belongs as an instance variable. Instead of having this `from_async_batch` method, can we instead have the following constructor?

```python
def __init__(self, async_batch: aioclient.Batch):
    self._async_batch = batch
```

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -351,30 +351,80 @@ async def attempts(self):
...
+    pass
+
+
+def submitted_batch_only(fun):
+    @functools.wraps(fun)
+    async def wrapped(*args, **kwargs):
+        batch = args[0]
+        if not batch.submitted:
+            raise BatchNotSubmittedError
+        return await fun(*args, **kwargs)
```
daniel-goldstein, line 372: I really like how you've structured this to be explicit about what operations are valid in each state that a `Batch` object can be in. Unfortunately doing this as a decorator means you'll need different decorators for sync and async `fun`s. It also erases all type information from the signature of the method (I tried for a while to type this and was struggling). Maybe these can just be private methods in the class and instead of a decorator other methods should start with `self._raise_if_unsubmitted()`? It would still be the same number of lines, albeit perhaps slightly noisier.

hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -750,42 +781,42 @@ def _batch_spec(self):
...
-                     submission_info=BatchSubmissionInfo(False))
+        self._id = batch_json['id']
+        return int(batch_json['update_id'])
 
     def _update_spec(self) -> dict:
         update_token = secrets.token_urlsafe(32)
         return {'n_jobs': len(self._jobs), 'token': update_token}
 
-    async def _create_update(self, batch_id: int) -> int:
+    @submitted_batch_only
```
daniel-goldstein, line 801: If these are private methods, do they need these decorators? This feels like it would be dead code.

daniel-goldstein (dismissed):
Great simplification in the number of concepts involved in this module. Just have a few tiny nits.

daniel-goldstein (approved):

PR #13463, opened 2023-08-18
[batch/ui] adds advanced search box and help link

iris-garden:
The updated template renders the following in Firefox on my machine:

Screenshot 2023-08-18 at 19 56 27

### DIFF
--- a/batch/batch/front_end/templates/table_search.html
+++ b/batch/batch/front_end/templates/table_search.html
@@ -1,0 +1,67 @@
+{% macro table_search(component_name, endpoint) %}
+  <style>
+    #{{ component_name }} {
+      margin: 10px;
+      margin-left: 0px;
+    }
+    #{{ component_name }}-inner {
+      height: 15rem;
+      width: 65vw;
+      display: flex;
+      align-items: center;
+    }
+    #{{ component_name }}-input-box {
+      background-color: #efefef;
+      padding: 0.75rem;
+      border-radius: 3px;
+      font-size: 1rem;
+      resize: none;
+      width: 450px;
+      white-space: pre;
+      height: 10rem;
+    }
+    #{{ component_name}}-buttons {
+      margin-left: 1rem;
+      height: 100%;
+      display: flex;
+      flex-direction: column;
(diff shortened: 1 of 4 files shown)


batch/batch/front_end/templates/table_search.html
```diff
@@ -0,0 +1,67 @@
...
+      cursor: pointer;
+      color: #283870;
+    }
+  </style>
+  <div class="flex-col">
+    <div id="{{ component_name }}" class="flex-col-align-right">
+      <form method="GET" action="{{ endpoint }}">
+        <div id="{{ component_name }}-inner">
+          <div>
+            <textarea id="{{ component_name }}-input-box" autofocus name="q">{% if q %}{{ q }}{% endif %}</textarea>
```
jigold, line 52: I'm pretty sure we don't need a CSRF element in the form since it's a GET, but @danking can you confirm?

```
<input type="hidden" name="_csrf" value="{{ csrf_token }}"/>
```

jigold (changes requested):
This looks great! One question for Dan.

batch/batch/front_end/templates/table_search.html
jigold: I was being silly. We don't need this.

jigold (approved):

jigold:
I forgot we need to switch the `CURRENT_QUERY_VERSION = 1` to `CURRENT_QUERY_VERSION = 2` in batch/front_end/query/__init__.py

jigold (dismissed):
see comment

batch/batch/front_end/templates/table_search.html
danking: Yeah, CSRF is to protect the user from a malicious website that convinces them to take an action at hail.is. This is only bad if the action is bad. For example, submitting a batch could cost money. Viewing the batches page should generally be safe b/c if the user unexpectedly sees a list of their batches, that's weird, but doesn't cost money, delete data, etc.

jigold (approved):

PR #13467, opened 2023-08-21
[hailtop.batch_client] Merge Unsubmitted/Submitted Job into Job

jigold:
Follow up to #13458 for making adding job groups simpler.


daniel-goldstein (changes requested):
This looks good, just a couple small comments about internal state tracking.

daniel-goldstein (approved):

PR #13469, opened 2023-08-21
[query] implement MatrixTable.write_table_per_column

danking:


danking:
Let's close and stay focused on our immediate goals.

PR #13472, opened 2023-08-21
[fs] remove Azure _debug parameters

danking:
This was added in https://github.com/hail-is/hail/pull/12421 to help debug errors in `testSeekMoreThanMaxInt`. We have not seen that transient error in a while and I think the switch to https://github.com/hail-is/hail/pull/12590 might have fixed some underlying misuse of `AppendBlobClient` (by not using it).

### DIFF
--- a/hail/src/main/scala/is/hail/io/fs/RouterFS.scala
+++ b/hail/src/main/scala/is/hail/io/fs/RouterFS.scala
@@ -19,6 +19,8 @@ class RouterFS(fss: IndexedSeq[FS]) extends FS {
 
   override def createCachedNoCompression(filename: String): PositionedDataOutputStream = lookupFS(filename).createCachedNoCompression(filename)
 
+  def openNoCompression(filename: String): SeekableDataInputStream = lookupFS(filename).openNoCompression(filename)
+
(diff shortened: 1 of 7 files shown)


hail/src/main/scala/is/hail/io/fs/RouterFS.scala
```diff
@@ -19,8 +19,6 @@ class RouterFS(fss: IndexedSeq[FS]) extends FS {
 
   override def createCachedNoCompression(filename: String): PositionedDataOutputStream = lookupFS(filename).createCachedNoCompression(filename)
 
-  def openNoCompression(filename: String, _debug: Boolean = false): SeekableDataInputStream = lookupFS(filename).openNoCompression(filename, _debug)
```
daniel-goldstein, line 22: Was this just meant to delete the extra parameter?

daniel-goldstein (changes requested):

hail/src/main/scala/is/hail/io/fs/RouterFS.scala
danking: yep, fixed.

daniel-goldstein (approved):

PR #13474, opened 2023-08-22
[pre-commit] add ruff, check-sql checks

danking:
We can use this for pyright or mypy but I found both took ~1.5s even on a tiny file like `ci.py`.

The check is slow the first time because it has to install a new venv; however, subsequent executions can use the extant venv.


batch/test/utils.py
```diff
@@ -1,26 +1,13 @@
...
-def create_batch(client: bc.BatchClient, **kwargs) -> bc.Batch:
-    ...
-
-
-@overload
-def create_batch(client: aiobc.BatchClient, **kwargs) -> aiobc.Batch:
-    ...
-
-
-def create_batch(client: Union[bc.BatchClient, aiobc.BatchClient], **kwargs) -> Union[bc.Batch, aiobc.Batch]:
```
danking, line 23: I have no idea why mypy hates these types, but it does. It claims the parameter types are the same in both cases.

danking:
@daniel-goldstein curious for your thoughts here; particularly in light of your switch to pyright. We could use this to run pyright automatically.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,28 @@
...
+        language: system
+        types: [python]
+        require_serial: true
+      - id: check-auth
+        name: check-auth
+        entry: bash pre-commit-make-shim auth
+        files: auth/
+        language: system
+        types: [python]
+        require_serial: true
```
daniel-goldstein, line 25: Can we add `gear`, `web_common`, and `hailtop`?

Makefile
```diff
@@ -52,6 +52,13 @@ check-services: $(CHECK_SERVICES_MODULES)
 pylint-%:
 	$(PYTHON) -m pylint --rcfile pylintrc --recursive=y $* --score=n
 
+.PRECIOUS: $(patsubst %, .venvs/%/bin/activate, $(SERVICES))
+.venvs/%/bin/activate: %/pinned-requirements.txt hail/python/dev/pinned-requirements.txt Makefile
+	$(PYTHON) -m venv .venvs/$*
+	source .venvs/$*/bin/activate && \
+	    pip3 install -r $*/pinned-requirements.txt -r hail/python/dev/pinned-requirements.txt && \
+	    pip3 install --editable ./gear ./web_common
```
daniel-goldstein, line 60: How come this doesn't need to install `hailtop` as editable?

pre-commit-make-shim
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -ex
+
+service=$1
+shift
+make pre-commit-check-in-$service-venv FILES="$*"
```
daniel-goldstein, line 7: Won't `$*` include `service`? I feel like we don't want that, we want only the files pre-commit is passing.

daniel-goldstein (changes requested):
Can you make optional hooks with `pre-commit`? I feel like because I use `pyright` in-editor, I don't personally need it when committing, and unlike `ruff` it's not a no-brainer because it lags a little.

pre-commit-make-shim
danking: The shift drops the service from `$*`

Makefile
danking: Dunno why this didn't cause problems, but I added it.

.pre-commit-config.yaml
danking: Yes, done.

danking: I also added `hail` which, once the venv was set up takes 360ms.

danking:
@daniel-goldstein in the end this was a trivial change.

daniel-goldstein (approved):
Love to see it

PR #13475, opened 2023-08-22
[batch] Setup database infrastructure for job groups

jigold:
The conceptual change here is we want to parameterize all batch related tables to have a new job group ID that I've set to **0** for the root job group. We need to make sure all future inserts / updates into the batches table are propagated to the new job groups table. When we create a batch now, we also create the corresponding entries into the job groups and job group parents tables.

I chose the root job group to be 0 as I think conceptually, the client should start numbering job groups at 1 and not know there is a hidden root job group being created under the hood. I'm not wedded to this.

I tried to check for all the indices that would be needed in my prototype. It's possible I missed one or two, but it's not a big deal to add it later.

I don't think we need to test this on a populated database (dev deploy main, submit jobs, then run the migration), but let me know if you think that would be helpful.

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -1338,7 +1338,7 @@ VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s);
 
         await tx.execute_insertone(
             '''
-INSERT INTO job_group_self_and_ancestors (batch_id, job_group_id, parent_id, level)
+INSERT INTO job_group_self_and_ancestors (batch_id, job_group_id, ancestor_id, level)
 VALUES (%s, %s, %s, %s);
 ''',
             (
--- a/batch/sql/estimated-current.sql
+++ b/batch/sql/estimated-current.sql
@@ -214,13 +214,13 @@ DROP TABLE IF EXISTS `job_group_self_and_ancestors`;
 CREATE TABLE IF NOT EXISTS `job_group_self_and_ancestors` (
   `batch_id` BIGINT NOT NULL,
   `job_group_id` INT NOT NULL,
-  `parent_id` INT NOT NULL,
+  `ancestor_id` INT NOT NULL,
   `level` INT NOT NULL,
-  PRIMARY KEY (`batch_id`, `job_group_id`, `parent_id`),
+  PRIMARY KEY (`batch_id`, `job_group_id`, `ancestor_id`),
   FOREIGN KEY (`batch_id`, `job_group_id`) REFERENCES job_groups (`batch_id`, `job_group_id`) ON DELETE CASCADE,
-  FOREIGN KEY (`batch_id`, `parent_id`) REFERENCES job_groups (`batch_id`, `job_group_id`) ON DELETE CASCADE
+  FOREIGN KEY (`batch_id`, `ancestor_id`) REFERENCES job_groups (`batch_id`, `job_group_id`) ON DELETE CASCADE
 ) ENGINE = InnoDB;
-CREATE INDEX `job_group_self_and_ancestors_parent_id_level` ON `job_group_self_and_ancestors` (`batch_id`, `parent_id`, `level`);
+CREATE INDEX `job_group_self_and_ancestors_ancestor_id_level` ON `job_group_self_and_ancestors` (`batch_id`, `ancestor_id`, `level`);
 CREATE INDEX `job_group_self_and_ancestors_job_group_id_level` ON `job_group_self_and_ancestors` (`batch_id`, `job_group_id`, `level`);
 
 CREATE TABLE IF NOT EXISTS `batch_updates` (
@@ -702,7 +702,7 @@ BEGIN
     INSERT INTO job_groups (batch_id, job_group_id, `user`, cancel_after_n_failures, `state`, n_jobs, time_created, time_completed, callback, attributes)
     VALUES (NEW.id, 0, NEW.`user`, NEW.cancel_after_n_failures, NEW.state, NEW.n_jobs, NEW.time_created, NEW.time_completed, NEW.callback, NEW.attributes);
 
-    INSERT INTO job_group_self_and_ancestors (batch_id, job_group_id, parent_id, `level`)
+    INSERT INTO job_group_self_and_ancestors (batch_id, job_group_id, ancestor_id, `level`)
     VALUES (NEW.id, 0, 0, 0);
   END IF;
 END $$
@@ -1569,7 +1569,7 @@ BEGIN
   DECLARE cur_n_completed INT;
 
   DECLARE job_group_cursor CURSOR FOR
-  SELECT parent_id
+  SELECT ancestor_id
   FROM job_group_self_and_ancestors
   WHERE batch_id = in_batch_id AND job_group_id = in_job_group_id
   ORDER BY job_group_id ASC;
@@ -1695,11 +1695,11 @@ BEGIN
     # necessary for backwards compatibility
     UPDATE batches_n_jobs_in_complete_states
     INNER JOIN (
-      SELECT batch_id, parent_id
+      SELECT batch_id, ancestor_id
       FROM job_group_self_and_ancestors
       WHERE batch_id = in_batch_id AND job_group_id = cur_job_group_id AND job_group_id != 0
       ORDER BY job_group_id ASC
-    ) AS t ON batches_n_jobs_in_complete_states.id = t.batch_id AND batches_n_jobs_in_complete_states.job_group_id = t.parent_id
+    ) AS t ON batches_n_jobs_in_complete_states.id = t.batch_id AND batches_n_jobs_in_complete_states.job_group_id = t.ancestor_id
     SET n_completed = n_completed + 1,
         n_cancelled = n_cancelled + (new_state = 'Cancelled'),
         n_failed = n_failed + (new_state = 'Error' OR new_state = 'Failed'),
--- a/batch/sql/setup-job-groups.sql
+++ b/batch/sql/setup-job-groups.sql
@@ -23,13 +23,13 @@ DROP TABLE IF EXISTS `job_group_self_and_ancestors`;
 CREATE TABLE IF NOT EXISTS `job_group_self_and_ancestors` (
   `batch_id` BIGINT NOT NULL,
   `job_group_id` INT NOT NULL,
-  `parent_id` INT NOT NULL,
+  `ancestor_id` INT NOT NULL,
   `level` INT NOT NULL,
-  PRIMARY KEY (`batch_id`, `job_group_id`, `parent_id`),
+  PRIMARY KEY (`batch_id`, `job_group_id`, `ancestor_id`),
   FOREIGN KEY (`batch_id`, `job_group_id`) REFERENCES job_groups (`batch_id`, `job_group_id`) ON DELETE CASCADE,
-  FOREIGN KEY (`batch_id`, `parent_id`) REFERENCES job_groups (`batch_id`, `job_group_id`) ON DELETE CASCADE
+  FOREIGN KEY (`batch_id`, `ancestor_id`) REFERENCES job_groups (`batch_id`, `job_group_id`) ON DELETE CASCADE
 ) ENGINE = InnoDB;
-CREATE INDEX `job_group_self_and_ancestors_parent_id_level` ON `job_group_self_and_ancestors` (`batch_id`, `parent_id`, `level`);
+CREATE INDEX `job_group_self_and_ancestors_ancestor_id_level` ON `job_group_self_and_ancestors` (`batch_id`, `ancestor_id`, `level`);
 CREATE INDEX `job_group_self_and_ancestors_job_group_id_level` ON `job_group_self_and_ancestors` (`batch_id`, `job_group_id`, `level`);
 
 ALTER TABLE batches ADD COLUMN migrated_batch BOOLEAN NOT NULL DEFAULT 0, ALGORITHM=INSTANT;
@@ -65,7 +65,7 @@ BEGIN
     INSERT INTO job_groups (batch_id, job_group_id, `user`, cancel_after_n_failures, `state`, n_jobs, time_created, time_completed, callback, attributes)
     VALUES (NEW.id, 0, NEW.`user`, NEW.cancel_after_n_failures, NEW.state, NEW.n_jobs, NEW.time_created, NEW.time_completed, NEW.callback, NEW.attributes);
 
-    INSERT INTO job_group_self_and_ancestors (batch_id, job_group_id, parent_id, `level`)
+    INSERT INTO job_group_self_and_ancestors (batch_id, job_group_id, ancestor_id, `level`)
     VALUES (NEW.id, 0, 0, 0);
   END IF;
 END $$
@@ -250,7 +250,7 @@ BEGIN
   DECLARE cur_n_completed INT;
 
   DECLARE job_group_cursor CURSOR FOR
-  SELECT parent_id
+  SELECT ancestor_id
   FROM job_group_self_and_ancestors
   WHERE batch_id = in_batch_id AND job_group_id = in_job_group_id
   ORDER BY job_group_id ASC;
@@ -376,11 +376,11 @@ BEGIN
     # necessary for backwards compatibility
     UPDATE batches_n_jobs_in_complete_states
     INNER JOIN (
-      SELECT batch_id, parent_id
+      SELECT batch_id, ancestor_id
       FROM job_group_self_and_ancestors
       WHERE batch_id = in_batch_id AND job_group_id = cur_job_group_id AND job_group_id != 0
       ORDER BY job_group_id ASC
-    ) AS t ON batches_n_jobs_in_complete_states.id = t.batch_id AND batches_n_jobs_in_complete_states.job_group_id = t.parent_id
+    ) AS t ON batches_n_jobs_in_complete_states.id = t.batch_id AND batches_n_jobs_in_complete_states.job_group_id = t.ancestor_id
     SET n_completed = n_completed + 1,
         n_cancelled = n_cancelled + (new_state = 'Cancelled'),
         n_failed = n_failed + (new_state = 'Error' OR new_state = 'Failed'),


batch/batch/front_end/front_end.py
```diff
@@ -80,7 +80,13 @@
     QueryError,
 )
 from ..file_store import FileStore
-from ..globals import BATCH_FORMAT_VERSION, HTTP_CLIENT_MAX_SIZE, RESERVED_STORAGE_GB_PER_CORE, complete_states
+from ..globals import (
+    BATCH_FORMAT_VERSION,
+    HTTP_CLIENT_MAX_SIZE,
+    RESERVED_STORAGE_GB_PER_CORE,
+    ROOT_JOB_GROUP_ID,
```
daniel-goldstein, line 87: I'm all for making `ROOT_JOB_GROUP_ID` a constant, but let's move it out of globals because to me that gives the impression that it is configurable or might change in the future. IIUC it is basically never a good idea to change this constant.

batch/sql/setup-job-groups.sql
```diff
@@ -0,0 +1,149 @@
...
+    UPDATE job_groups
+    SET `user` = NEW.`user`,
+      cancel_after_n_failures = NEW.cancel_after_n_failures,
+      `state` = NEW.state,
+      n_jobs = NEW.n_jobs,
+      time_created = NEW.time_created,
+      time_completed = NEW.time_completed,
+      callback = NEW.callback,
+      attributes = NEW.attributes
+    WHERE batch_id = NEW.id AND job_group_id = 0;
```
daniel-goldstein, line 80: For clarification, are we going to need this part of the trigger forever to ensure that the duplicated state between job groups and batches is updated in unison?

daniel-goldstein: Do we want to allow things like the `user` to change?

batch/sql/estimated-current.sql
```diff
@@ -187,6 +188,40 @@ CREATE INDEX `batches_token` ON `batches` (`token`);
...
+CREATE TABLE IF NOT EXISTS `job_group_parents` (
+  `batch_id` BIGINT NOT NULL,
+  `job_group_id` INT NOT NULL,
+  `parent_id` INT NOT NULL,
+  `level` INT NOT NULL,
+  PRIMARY KEY (`batch_id`, `job_group_id`, `parent_id`),
+  FOREIGN KEY (`batch_id`, `job_group_id`) REFERENCES job_groups (`batch_id`, `job_group_id`) ON DELETE CASCADE,
+  FOREIGN KEY (`batch_id`, `parent_id`) REFERENCES job_groups (`batch_id`, `job_group_id`) ON DELETE CASCADE
+) ENGINE = InnoDB;
+CREATE INDEX `job_group_parents_p_id_level` ON `job_group_parents` (`batch_id`, `parent_id`, `level`);
```
daniel-goldstein, line 222: nit: I'd prefer to use the full name in the indexes ( `parent_id` )

batch/sql/estimated-current.sql
```diff
@@ -1362,7 +1433,7 @@ BEGIN
     # there are no cancellable jobs left, they have been cancelled
     DELETE FROM batch_inst_coll_cancellable_resources WHERE batch_id = in_batch_id;
 
-    INSERT INTO batches_cancelled VALUES (in_batch_id);
+    INSERT INTO batches_cancelled (id) VALUES (in_batch_id);
```
daniel-goldstein, line 1436: To confirm my understanding, cancelling the batch means a row that is `(batch_id, NULL)` and cancelling a job group means a row that is `(batch_id, job_group_id)`?

daniel-goldstein (changes requested):
Sorry for the delay on this, I wanted to make sure I could devote full attention to it. It looks good, I just have some clarifying questions.

batch/sql/estimated-current.sql
jigold: No. There's an implicit default value here of job_group_id = 0 (`INT DEFAULT 0`). I can explicitly add the job_group_id being 0 here for clarity.

batch/sql/setup-job-groups.sql
jigold: No. The `user` is for an index that is needed. It shouldn't change between job groups in the same batch, but I wanted to be complete here that any field updates are reflected in the job groups table.

As for needing the trigger forever, I think we have independent insert and update commands in the batch front end / driver that keep the state in sync and not in the trigger. At some point, I want most of these fields to be essentially deprecated on the batches table. Thoughts??

batch/batch/front_end/front_end.py
jigold: Yep, this constant can never change once we choose it.

batch/sql/setup-job-groups.sql
daniel-goldstein: ah, ya I appreciate wanting to be complete here but as a reader I interpret this less as being complete and more as "accounting for all of the fields that could change". Can we instead error if the new value differs from the old one?

jigold:
I'm putting this on hold until I get clarity on #13513

jigold:
@danking I'd like you to look over the new SQL in MJC as well.

jigold:
This is now passing tests in gcp. I didn't retry it in Azure, but I presume it would pass as well.

danking:
@daniel-goldstein are you owning review for this?

daniel-goldstein:
Apologies, I'll review this today

batch/sql/estimated-current.sql
```diff
@@ -278,6 +319,7 @@ CREATE INDEX `jobs_batch_id_state_always_run_cancelled` ON `jobs` (`batch_id`, `
 CREATE INDEX `jobs_batch_id_update_id` ON `jobs` (`batch_id`, `update_id`);
 CREATE INDEX `jobs_batch_id_always_run_n_regions_regions_bits_rep_job_id` ON `jobs` (`batch_id`, `always_run`, `n_regions`, `regions_bits_rep`, `job_id`);
 CREATE INDEX `jobs_batch_id_ic_state_ar_n_regions_bits_rep_job_id` ON `jobs` (`batch_id`, `inst_coll`, `state`, `always_run`, `n_regions`, `regions_bits_rep`, `job_id`);
```
daniel-goldstein, line 321: missing an added column to the jobs table.

daniel-goldstein (commented):
I think this looks good, but would appreciate a final review by Dan especially on the new procedure bit.

jigold:
@danking Can you please take a look?

danking:
I'm a bit flat out this week but will endeavor to look before Friday.

batch/sql/estimated-current.sql
```diff
@@ -268,6 +309,7 @@ CREATE TABLE IF NOT EXISTS `jobs` (
   `inst_coll` VARCHAR(255),
   `n_regions` INT DEFAULT NULL,
   `regions_bits_rep` BIGINT DEFAULT NULL,
+  `job_group_id` INT NOT NULL DEFAULT 0,
   PRIMARY KEY (`batch_id`, `job_id`),
```
danking, line 313: OK, so this is a really key and important aspect: `batch_id, job_id` is still the unique identifier of a job. Groups are merely some organizational structure atop all this. I think verbiage to this effect should make it into the RFC. What's the unique identifier for the relevant concepts. I think its:
1. Job: `batch_id, job_id`
2. Group: `batch_id, job_group_id`.
3. Batch: `batch_id`

batch/sql/estimated-current.sql
```diff
@@ -187,6 +188,40 @@ CREATE INDEX `batches_token` ON `batches` (`token`);
...
+  `parent_id` INT NOT NULL,
+  `level` INT NOT NULL,
+  PRIMARY KEY (`batch_id`, `job_group_id`, `parent_id`),
+  FOREIGN KEY (`batch_id`, `job_group_id`) REFERENCES job_groups (`batch_id`, `job_group_id`) ON DELETE CASCADE,
+  FOREIGN KEY (`batch_id`, `parent_id`) REFERENCES job_groups (`batch_id`, `job_group_id`) ON DELETE CASCADE
+) ENGINE = InnoDB;
+CREATE INDEX `job_group_parents_parent_id_level` ON `job_group_parents` (`batch_id`, `parent_id`, `level`);
+CREATE INDEX `job_group_parents_job_group_id_level` ON `job_group_parents` (`batch_id`, `job_group_id`, `level`);
+
 CREATE TABLE IF NOT EXISTS `batch_updates` (
```
danking, line 225: Do I understand correctly that, in this PR, we are setting up database structures for tracking states, cancelling, and billing but *we are not adding* structure of creation of new groups. Instead, in this PR, every batch has just one group (so we can piggy back on batch creation).

batch/sql/estimated-current.sql
danking: OK, so, a job is the member of exactly one group. Groups may or may not have extra structure, but every job is member of exactly one group.

batch/sql/estimated-current.sql
```diff
@@ -278,6 +320,7 @@ CREATE INDEX `jobs_batch_id_state_always_run_cancelled` ON `jobs` (`batch_id`, `
 CREATE INDEX `jobs_batch_id_update_id` ON `jobs` (`batch_id`, `update_id`);
 CREATE INDEX `jobs_batch_id_always_run_n_regions_regions_bits_rep_job_id` ON `jobs` (`batch_id`, `always_run`, `n_regions`, `regions_bits_rep`, `job_id`);
 CREATE INDEX `jobs_batch_id_ic_state_ar_n_regions_bits_rep_job_id` ON `jobs` (`batch_id`, `inst_coll`, `state`, `always_run`, `n_regions`, `regions_bits_rep`, `job_id`);
+CREATE INDEX `jobs_batch_id_job_group_id` ON `jobs` (`batch_id`, `job_group_id`);
 
 CREATE TABLE IF NOT EXISTS `jobs_telemetry` (
   `batch_id` BIGINT NOT NULL,
```
danking, line 326: Down in the unchanged code, `job_parents` indicates that jobs can have parents in different groups. Verify I understand correctly:

> This is fine because, by design, a job can be cancelled while its children are still Pending. The cancellation loop will eventually notice the cancelled parent and mark the children as cancelled. A cancelled job is only concerned with its member jobs (the jobs with its id as their job_group_id).

batch/sql/estimated-current.sql
```diff
@@ -351,13 +394,16 @@ CREATE TABLE IF NOT EXISTS `regions` (
 
 CREATE TABLE IF NOT EXISTS `batch_attributes` (
```
danking, line 395: OK, again, confirm for me:

> This is now a poorly named table: it's really `job_group_attributes`. A subsequent PR will rename this table.

And also confirm for me:

> The attributes on the root group id are displayed in the UI as the batch's attributes.

batch/sql/estimated-current.sql
```diff
@@ -391,6 +437,7 @@ CREATE INDEX aggregated_billing_project_user_resources_by_date_v2_user ON `aggre
 DROP TABLE IF EXISTS `aggregated_batch_resources_v2`;
 CREATE TABLE IF NOT EXISTS `aggregated_batch_resources_v2` (
```
danking, line 438: Note to ourselves, these tables should be dropped entirely soon anyway: https://github.com/hail-is/hail/issues/13656

batch/sql/estimated-current.sql
```diff
@@ -645,6 +693,19 @@ BEGIN
...
+DROP TRIGGER IF EXISTS batches_after_update $$
+CREATE TRIGGER batches_after_update AFTER UPDATE ON batches
+FOR EACH ROW
+BEGIN
+  IF OLD.migrated_batch = 0 AND NEW.migrated_batch = 1 THEN
+    INSERT INTO job_groups (batch_id, job_group_id, `user`, cancel_after_n_failures, `state`, n_jobs, time_created, time_completed, callback, attributes)
+    VALUES (NEW.id, 0, NEW.`user`, NEW.cancel_after_n_failures, NEW.state, NEW.n_jobs, NEW.time_created, NEW.time_completed, NEW.callback, NEW.attributes);
+
+    INSERT INTO job_group_parents (batch_id, job_group_id, parent_id, `level`)
+    VALUES (NEW.id, 0, 0, 0);
```
danking, line 705: I'm a bit surprised by this. Does this substantially simplify other code? Having a self-loop in the parents table is confusing because it violates the statement

> if `(_, y, z, _)` is in `job_groups_parents` then `z` is the parent of `y`.

batch/sql/estimated-current.sql
```diff
@@ -1114,11 +1176,21 @@ BEGIN
...
       WHERE id = in_batch_id;
 
+      UPDATE job_groups
+      INNER JOIN (
+        SELECT batch_id, job_group_id, CAST(COALESCE(SUM(n_jobs), 0) AS SIGNED) AS staged_n_jobs
+        FROM batches_inst_coll_staging
+        WHERE batch_id = in_batch_id AND update_id = in_update_id
+        GROUP BY batch_id, job_group_id
+      ) AS t ON job_groups.batch_id = t.batch_id AND job_groups.job_group_id = t.job_group_id
+      SET `state` = 'running', time_completed = NULL, n_jobs = n_jobs + t.staged_n_jobs;
```
danking, line 1186: Why the inner join? It has to be `expected_n_jobs`, right? This seems like natural code to be added by the subsequent PR that actually handles parallel creation of job groups.

batch/sql/estimated-current.sql
danking: (I made this note here because batch_updates has no job_group_id).

batch/sql/estimated-current.sql
```diff
@@ -1480,6 +1553,56 @@ BEGIN
...
+  DECLARE total_jobs_in_job_group INT;
+  DECLARE cur_n_completed INT;
+
+  DECLARE job_group_cursor CURSOR FOR
+  SELECT parent_id
+  FROM job_group_parents
+  WHERE batch_id = in_batch_id AND job_group_id = in_job_group_id
+  ORDER BY job_group_id ASC;
+
+  DECLARE CONTINUE HANDLER FOR NOT FOUND SET done = TRUE;
```
danking, line 1574: This is pretty heady stuff! Worth a link to https://dev.mysql.com/doc/refman/8.0/en/cursors.html

batch/sql/estimated-current.sql
```diff
@@ -1562,6 +1688,22 @@ BEGIN
...
+    INNER JOIN (
+      SELECT batch_id, parent_id
+      FROM job_group_parents
+      WHERE batch_id = in_batch_id AND job_group_id = cur_job_group_id AND job_group_id != 0
+      ORDER BY job_group_id ASC
+    ) AS t ON batches_n_jobs_in_complete_states.id = t.batch_id AND batches_n_jobs_in_complete_states.job_group_id = t.parent_id
+    SET n_completed = n_completed + 1,
+        n_cancelled = n_cancelled + (new_state = 'Cancelled'),
+        n_failed = n_failed + (new_state = 'Error' OR new_state = 'Failed'),
+        n_succeeded = n_succeeded + (new_state != 'Cancelled' AND new_state != 'Error' AND new_state != 'Failed');
```
danking, line 1703: Just to be clear: this has absolutely no effect in this PR because it is not possible to create job groups (other than the implicitly created root one)?

danking (dismissed):
Mostly questions

batch/sql/estimated-current.sql
jigold: Yes, you understand correctly. This is just to get job groups in the database and mirroring the correct state for the root job group and the batch.

`batch_updates` should never have a job group id as an update is a separate concept and an update can contain multiple job groups. The update is just the staged "transaction" of changes to be made to the batch rather than the job organization.

batch/sql/estimated-current.sql
jigold: Not sure who wrote that statement whether it was you or me. That is correct except the last statement should be 

> A cancelled job is only concerned with its child jobs (the jobs with its id as their parent_id). 

Job groups have nothing to do with individual job level cancellation. They are only necessary for a bulk cancellation operation in which case any cancelled jobs will propogate that cancellation event down to their children.

batch/sql/estimated-current.sql
jigold: Agreed the table name sucks (all of the new tables that have been updated have bad names). Figuring out how to do the rename online will require trying to write to both the old and new table name. I don't think we can just wait for the container to redeploy and I don't want to copy the tables completely. Writing to both the old and new doesn't seem so bad if we build some infrastructure for this. I just haven't thought about it deeply yet.

Nothing has changed about the UI. Since the job group id is always "0" and is not a primary key (yet), it's just as if there's a dummy extra column right now. Shouldn't change any existing queries.

batch/sql/estimated-current.sql
jigold: I believe if we don't do this, then we'll need to split the queries into two in a bunch of places or maybe use loops. Once to do the operation on the group of interest and then another to do it on all its parents. Here's an example of billing propagation up the tree and then cancellation down the tree:

```sql
    INSERT INTO aggregated_batch_resources_v3 (batch_id, job_group_id, resource_id, token, `usage`)
    SELECT attempt_resources.batch_id,
      job_group_parents.parent_id,
      attempt_resources.deduped_resource_id,
      rand_token,
      msec_diff_rollup * quantity
    FROM attempt_resources
    LEFT JOIN jobs ON attempt_resources.batch_id = jobs.batch_id AND attempt_resources.job_id = jobs.job_id
    LEFT JOIN job_group_parents ON jobs.batch_id = job_group_parents.batch_id AND jobs.job_group_id = job_group_parents.job_group_id
    JOIN aggregated_batch_resources_v2 ON
      aggregated_batch_resources_v2.batch_id = attempt_resources.batch_id AND
      aggregated_batch_resources_v2.resource_id = attempt_resources.resource_id AND
      aggregated_batch_resources_v2.token = rand_token
    WHERE attempt_resources.batch_id = NEW.batch_id AND attempt_resources.job_id = NEW.job_id AND attempt_resources.attempt_id = NEW.attempt_id AND migrated = 1
    ON DUPLICATE KEY UPDATE `usage` = aggregated_batch_resources_v3.`usage` + msec_diff_rollup * quantity;
```

```sql
  INSERT INTO batch_inst_coll_cancellable_resources (batch_id, update_id, job_group_id, inst_coll, token,
    n_ready_cancellable_jobs, ready_cancellable_cores_mcpu, n_creating_cancellable_jobs, n_running_cancellable_jobs,
    running_cancellable_cores_mcpu)
  SELECT batch_inst_coll_cancellable_resources.batch_id, batch_inst_coll_cancellable_resources.update_id, job_group_parents.parent_id, inst_coll, 0,
    -1 * (@n_ready_cancellable_jobs := COALESCE(SUM(n_ready_cancellable_jobs), 0)),
    -1 * (@ready_cancellable_cores_mcpu := COALESCE(SUM(ready_cancellable_cores_mcpu), 0)),
    -1 * (@n_creating_cancellable_jobs := COALESCE(SUM(n_creating_cancellable_jobs), 0)),
    -1 * (@n_running_cancellable_jobs := COALESCE(SUM(n_running_cancellable_jobs), 0)),
    -1 * (@running_cancellable_cores_mcpu := COALESCE(SUM(running_cancellable_cores_mcpu), 0))
  FROM batch_inst_coll_cancellable_resources
  INNER JOIN batch_updates ON batch_inst_coll_cancellable_resources.batch_id = batch_updates.batch_id AND
    batch_inst_coll_cancellable_resources.update_id = batch_updates.update_id
  INNER JOIN job_group_parents ON batch_inst_coll_cancellable_resources.batch_id = job_group_parents.batch_id AND
    batch_inst_coll_cancellable_resources.job_group_id = job_group_parents.job_group_id
  WHERE job_group_parents.batch_id = in_batch_id AND
    job_group_parents.job_group_id = in_job_group_id AND
    batch_updates.committed
  GROUP BY batch_inst_coll_cancellable_resources.batch_id, batch_inst_coll_cancellable_resources.update_id,
    job_group_parents.parent_id, inst_coll
  ON DUPLICATE KEY UPDATE
    n_ready_cancellable_jobs = n_ready_cancellable_jobs - @n_ready_cancellable_jobs,
    ready_cancellable_cores_mcpu = ready_cancellable_cores_mcpu - @ready_cancellable_cores_mcpu,
    n_creating_cancellable_jobs = n_creating_cancellable_jobs - @n_creating_cancellable_jobs,
    n_running_cancellable_jobs = n_running_cancellable_jobs - @n_running_cancellable_jobs,
    running_cancellable_cores_mcpu = running_cancellable_cores_mcpu - @running_cancellable_cores_mcpu;

  # delete all rows that are children of this job group
  DELETE batch_inst_coll_cancellable_resources FROM batch_inst_coll_cancellable_resources
  LEFT JOIN batch_updates ON batch_inst_coll_cancellable_resources.batch_id = batch_updates.batch_id AND
    batch_inst_coll_cancellable_resources.update_id = batch_updates.update_id
  INNER JOIN job_group_parents ON batch_inst_coll_cancellable_resources.batch_id = job_group_parents.batch_id AND
    batch_inst_coll_cancellable_resources.job_group_id = job_group_parents.job_group_id
  WHERE batch_inst_coll_cancellable_resources.batch_id = in_batch_id AND
    job_group_parents.parent_id = in_job_group_id AND
    batch_updates.committed;
```

jigold: There might be other queries in the client code that having the self loop is helpful.

batch/sql/estimated-current.sql
jigold: I don't remember why I used an inner join. I don't think it really matters. No, you need to add this query here. The whole point of this PR is to make sure the batches table and job groups table are in sync. If the batch is updated, then the corresponding job group's state must be updated.

batch/sql/estimated-current.sql
jigold: I believe that is correct. I believe I can move this into a different PR that will need to be done after the job groups tables are fully populated that allow for job_group_id != 0 as well as having explicit primary keys with job_group_id in the tables for `aggregated_batch_resources`, `batches_cancelled` etc.

jigold:
@danking I'm keeping the changes requested state on until I get the full scope of changes that need to be made. Would be good to get your feedback first after my responses so I can do all changes in one pass.

jigold:
@danking Would it be possible to get another review by the end of this week? Not urgent, but would like to keep making progress on this.

batch/sql/estimated-current.sql
jigold: This is I believe the major outstanding point of discussion.

danking: Ah, I think this table has the wrong name. Knowing only the name, I thought this was true:

> if `(_, y, z, _)` is in `job_groups_parents` then `z` is the parent of `y`.

But after seeing the self-loop I thought this was true:

> if `(_, y, z, _)` is in `job_groups_parents` then `z` is the parent of `y` or `z` == `y`.

But now I see that what this *actually* is is something like `job_groups_needing_to_be_updated_when_a_job_in_this_group_changes`. It's not the ancestors (because of the self loop) and it's not the parents (because it includes ancestors and a self loop).

We need a clear and correct name for this table, but I do not object to the table itself.

batch/sql/estimated-current.sql
danking: Just take batch offline long enough to do the renames. Renames should be near instantaneous. We have no SLAs and even if we did we wouldn't have 100% uptime! 😄

batch/sql/estimated-current.sql
danking: er, sorry I meant to say:
> A cancelled job group is only concerned with its member jobs (the jobs with its id as their job_group_id).

And maybe I should have added:
> A cancelled job group need not care about the descendants of its jobs. The job cancelling loop will handle them.

batch/sql/estimated-current.sql
danking: For this I meant we need a comment with that link

danking (dismissed):
responded to comments

batch/sql/estimated-current.sql
jigold: Is it okay if we do the renames in a separate PR? I don't want to do too much in one set of changes and there has to be a separate follow up PR to change the foreign key constraints and primary keys as well. We'll also want to rename the indices on those tables.

batch/sql/estimated-current.sql
jigold: How about `job_group_tree`?

jigold:
@danking I changed the name `job_group_parents` => `job_group_tree`.

jigold:
Can I assume this is basically done so I can move on to the next part of the stack?

danking:
I still dislike job_group_tree because trees are usually represented in terms of their ancestor or parent-child relationships, neither of which are what this is. That said, I don't think we should block this PR on a naming quibble.

We can do renames separately. We should update the RFC to be clear about the unique identifiers of the three things groups, jobs, and batches. (In particular, jobs are uniquely identifier by batch id and job id, group id is not part of it).

danking:
In Order Theory, this operation is called "the upper/downward closure of x" https://en.wikipedia.org/wiki/Upper_set . `groups_affected_by` maybe? `group_self_and_ancestors`?

jigold:
How about `job_group_directed_sets` or `job_group_lattice`? https://en.wikipedia.org/wiki/Directed_set https://en.wikipedia.org/wiki/Lattice_(order)

danking:
My issue with both of those names is the same as with `_tree`. It's not clear if you're storing the non-transitive or the transitive relation & its not clear if self-edges are included. I want a name that unambiguously says "the self-edge and all the ancestor edges are in here" or a name that is more domain-specific like "job groups that need to be updated when a job in this job group is changed". 

`job_group_self_and_ancestor` feels like the shortest name so far that satisfies my concerns, but I'm open to other ones that are clear about self-edge-ness and transitive-ness.

batch/sql/estimated-current.sql
jigold: See #13810 for table renames

jigold:
Ok. It's been renamed.

danking (approved):
I'm currently assigned, so I'll approve, but ping to @daniel-goldstein if you had remaining concerns.

daniel-goldstein:
I think this thread addressed everything, looks good to me!

jigold:
@daniel-goldstein Can you please check the last commit real quick? Should just be a simple rename, but want someone to double check I didn't do something bad.

batch/sql/setup-job-groups.sql
```diff
@@ -0,0 +1,423 @@
...
+      INNER JOIN `job_parents`
+        ON jobs.batch_id = `job_parents`.batch_id AND
+           jobs.job_id = `job_parents`.job_id
+      SET jobs.state = IF(jobs.n_pending_parents = 1, 'Ready', 'Pending'),
+          jobs.n_pending_parents = jobs.n_pending_parents - 1,
+          jobs.cancelled = IF(new_state = 'Success', jobs.cancelled, 1),
+          jobs_telemetry.time_ready = IF(jobs.n_pending_parents = 1, new_timestamp, jobs_telemetry.time_ready)
+      WHERE jobs.batch_id = in_batch_id AND
+            `job_parents`.batch_id = in_batch_id AND
+            `job_parents`.parent_id = in_job_id;
```
daniel-goldstein, line 402: Good rename but I'm seeing multiple occurrences of `parent_id` and `job_parents` in the migration itself. Feels like any old names in here should error in CI so I'm not worried once this is green.

daniel-goldstein: Oh yes oops, I fully read that as `job_group_parents`.

daniel-goldstein: Great rename, it'll be much easier to keep these column names straight now that there aren't two columns that mean different things across different tables

PR #13482, opened 2023-08-23
[pre-commit] add pyright to pre-push

danking:
IME, this runs in ~1.5s. Not so bad for a push.

I also added pyright to dev reqs.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,16 @@
 ---
+default_install_hook_types: [pre-commit, pre-push]
 repos:
+  - repo: local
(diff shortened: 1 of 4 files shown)


daniel-goldstein (changes requested):
I'm hesitant to introduce the bomb here. I get thousands of pyright errors on Hail Query code and hundreds on services code. What do you think of just adding the pre-push hook to the pyright PR? That already has the requirements changes. I approve of the code you have in the pre-commit yaml.

danking:
Sure, pull it into the pyright PR and maybe limit to non-hail.

PR #13487, opened 2023-08-24
[batch] Populate job groups in database

jigold:
This PR just populates the records for older batch into the `job_groups` and `job_group_self_and_ancestors` tables.

Stacked on #13475


jigold:
This PR is almost identical to #12849. This time we're back filling the job groups and job_groups_self_and_ancestors tables.

iris-garden (approved):

PR #13489, opened 2023-08-24
[devbin] Generate cleanup policies for GCP artifact registry

jigold:
Usage: `python3 devbin/generate_gcp_ar_cleanup_policy.py > my_policy_file.txt`
Reference: https://cloud.google.com/artifact-registry/docs/repositories/cleanup-policy
Note, we have a maximum of 10 policies available.


devbin/generate_gcp_ar_cleanup_policy.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+third_party_images_fp = 'docker/third-party/images.txt'
+third_party_packages = []
+with open(third_party_images_fp, 'r') as f:
+    for image in f:
+        image = image.strip()
+        package = image.split(':')[0]
+        if package not in third_party_packages:
+            third_party_packages.append(package)
```
danking, line 110: Can we keep only the tags listed there?

devbin/generate_gcp_ar_cleanup_policy.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+deploy_packages = []
+with open('build.yaml', 'r') as f:
```
danking, line 114: Also get the images from `ci/test/resources/build.yaml`

devbin/generate_gcp_ar_cleanup_policy.py
```diff
@@ -0,0 +1,140 @@
...
+        if self.package_name_prefixes is not None:
+            condition['packageNamePrefixes'] = self.package_name_prefixes
+        if self.older_than:
+            condition['olderThan'] = self.older_than
+        if self.newer_than:
+            condition['newerThan'] = self.newer_than
+        return data
+
+
+class ConditionalKeepPolicy(CleanupPolicy):
```
danking, line 48: This is just a `KeepPolicy`, right? Or, otherwise, the above is a `ConditionalDeletePolicy`?

devbin/generate_gcp_ar_cleanup_policy.py
```diff
@@ -0,0 +1,140 @@
...
+third_party_packages.sort()
+deploy_packages.sort()
+
+policies = [
+    DeletePolicy('delete_untagged', 'untagged'),
+    DeletePolicy('delete_dev', 'tagged', tag_prefixes=['dev-'], older_than='3d'),
+    DeletePolicy('delete_test_pr', 'tagged', tag_prefixes=['test-pr-'], older_than='3d'),
+    DeletePolicy('delete_test_deploy', 'tagged', tag_prefixes=['test-deploy-'], older_than='3d'),
+    DeletePolicy('delete_cache', 'tagged', tag_prefixes=['cache-pr-'], older_than='7d'),
+    DeletePolicy('delete_cache', 'tagged', tag_prefixes=['cache-'], older_than='30d'),
```
danking, line 133: Are the names required to be unique?

danking (changes requested):

devbin/generate_gcp_ar_cleanup_policy.py
danking: I guess that means we need #-of-package policies, which is currently 11.

devbin/generate_gcp_ar_cleanup_policy.py
jigold: I sort above.

devbin/generate_gcp_ar_cleanup_policy.py
jigold: I grabbed all of the tags if they exist, but I'm concerned because these are tag prefix matches so it's not 100% specific to the package.

devbin/generate_gcp_ar_cleanup_policy.py
danking: I'm referring to the names of the policies themselves; `delete_cache` appears twice.

devbin/generate_gcp_ar_cleanup_policy.py
jigold: I just used their nomenclature in the docs

devbin/generate_gcp_ar_cleanup_policy.py
danking: Yeah, I think if we can't do something exact we should do nothing at all. Better to not have a false sense of security.

danking:
I was thinking either:
1. `terraform apply` somehow executes a bash script which is then fed as input to the terraform provider for the cleanup polices.
2. The cleanup policy is a checked-in file which we verify is up-to-date in CI (much like how pinned-requirements.txt are derived from requirements.txt).

The latter seems better to me since we know the current state of the cleanup policy on any commit by looking at a file rather than executing some code.

daniel-goldstein:
Ah, I see. 1. is definitely possible, but I agree with you that I prefer 2 because it is checked by our system.

danking (changes requested):
Excellent, my final request is to:

1. Run this and check in the result (store it in `infra/gcp-broad/`).
2. Add a build.yaml step that runs this and `diff`s against the checked in file. Fail if they differ.
3. Install the policy in hail-vdc.
4. Create a bug issue that asks us to import this into terraform once https://github.com/hashicorp/terraform-provider-google-beta/commit/bc4aa512356891f78415d5f309bfe47b0697ac11 is publicly released (see also the [overarching tracking issue](https://github.com/hashicorp/terraform-provider-google/issues/13824)). It's not in 4.79.0 (see [what was added since then](https://github.com/hashicorp/terraform-provider-google-beta/compare/v4.79.0...main)). Releases appear to happen ~once a week, so we should be able to import into terraform in September.

jigold:
Once you've looked over the policy generated, I'll apply it. Otherwise, this is ready to go.

danking (approved):
lgtm

danking:
Once it merges, please apply it

PR #13490, opened 2023-08-24
[hailctl] Add a user to a billing project through the CLI

daniel-goldstein:
Trying to slowly add more reliability to the `hailctl` CLI. This adds some very basic tests for the `hailctl batch billing` subcommand that mocks the `BatchClient` so it's just testing that command line parsing and yaml dumping don't break. Most of the other noise in this PR is a refactor. I moved `hail/python/test/hailtop/hailctl/config/conftest.py` up a level so I could reuse its `CLIRunner` fixture across all `hailctl` test modules. That fixture sets a new config directory per test so if you use it in a test the test won't accidentally use or modify the user's actual hailctl config.


daniel-goldstein:
Something seems to have changed from when I wrote this to opened it. Unassigning until I fix the test

PR #13494, opened 2023-08-24
[query] clarify file format version vs library version

danking:
Addresses user confusion in https://discuss.hail.is/t/potential-outdated-error-statement-for-hail-version-incompatibility/3562.


matren395:
the note at the end `Note that the file format version and the Hail Python library version are not the same` is really helpful and would've answered my question when it popped up, thanks!

patrick-schultz (approved):

PR #13498, opened 2023-08-25
[query] fix shadowing of field names by methods

### LINKED ISSUE #13495: [query] do not allow the creation of a field which classes with a necessary builtin
### What happened?

`StructExpression` s are Dict-like and we rely on this behavior for methods like `.items()`. We should probably ban the use of names like `items` so that you can always iterate over a struct expressions field names and field value expressions.

```python3
ht = ht.group_by(
    'gene'
).aggregate(
    items=hl.agg.take(ht.key, max(list(n_dict.values())), ordering = ht.rand_id)
)
```
```
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_9677/3768472856.py in <cell line: 1>()
----> 1 ht = ht.group_by('gene').aggregate(items=hl.agg.take(ht.key, max(list(n_dict.values())), ordering = ht.rand_id))

<decorator-gen-1202> in aggregate(self, **named_exprs)

/opt/conda/miniconda3/lib/python3.10/site-packages/hail/typecheck/check.py in wrapper(__original_func, *args, **kwargs)
    582     def wrapper(__original_func, *args, **kwargs):
    583         args_, kwargs_ = check_all(__original_func, args, kwargs, checkers, is_method=is_method)
--> 584         return __original_func(*args_, **kwargs_)
    585
    586     return wrapper

/opt/conda/miniconda3/lib/python3.10/site-packages/hail/table.py in aggregate(self, **named_exprs)
    243
    244         key_struct = self._key_expr
--> 245         return Table(ir.TableKeyByAndAggregate(base._tir,
    246                                                hl.struct(**named_exprs)._ir,
    247                                                key_struct._ir,

/opt/conda/miniconda3/lib/python3.10/site-packages/hail/table.py in __init__(self, tir)
    365
    366         for k, v in itertools.chain(self._globals.items(),
--> 367                                     self._row.items()):
    368             self._set_field(k, v)
    369

TypeError: 'ArrayStructExpression' object is not callable
```

### Version

0.2.120

### Relevant log output

_No response_

patrick-schultz:
In particular, struct field names which clash with methods on `StructExpression`.

close #13495

CHANGELOG: Fix a bug where field names can shadow methods on the StructExpression class, e.g. "items", "keys", "values". Now the only way to access such fields is through the getitem syntax, e.g. "some_struct['items']". It's possible this could break existing code that uses such field names.

### DIFF
--- a/hail/python/hail/expr/expressions/typed_expressions.py
+++ b/hail/python/hail/expr/expressions/typed_expressions.py
@@ -1785,7 +1785,9 @@ class StructExpression(Mapping[Union[str, int], Expression], Expression):
         if key not in self._fields:
             if hasattr(self, key):
                 self._shadowed_fields.add(key)
-        self._fields[key] = value
+            else:
+                self.__dict__[key] = value
+            self._fields[key] = value
 
     def _get_field(self, item):
         if item in self._fields:
@@ -1802,10 +1804,7 @@ class StructExpression(Mapping[Union[str, int], Expression], Expression):
         return super().__getattribute__(item)
 
     def __getattr__(self, item):
-        if item in self._fields:
-            return self._fields[item]
-        else:
-            raise AttributeError(get_nice_attr_error(self, item))
+        raise AttributeError(get_nice_attr_error(self, item))
 
     def __len__(self):
         return len(self._fields)
--- a/hail/python/hail/expr/expressions/typed_expressions.py
+++ b/hail/python/hail/expr/expressions/typed_expressions.py
@@ -1743,6 +1743,8 @@ class StructExpression(Mapping[Union[str, int], Expression], Expression):
         x = ir.MakeStruct([(n, expr._ir) for (n, expr) in fields.items()])
         indices, aggregations = unify_all(*fields.values())
         s = StructExpression.__new__(cls)
+        s._warned_on_shadowed_name = False
+        s._shadowed_fields = set()
         s._fields = {}
         for k, v in fields.items():
             s._set_field(k, v)
@@ -1753,6 +1755,8 @@ class StructExpression(Mapping[Union[str, int], Expression], Expression):
     def __init__(self, x, type, indices=Indices(), aggregations=LinkedList(Aggregation)):
         super(StructExpression, self).__init__(x, type, indices, aggregations)
         self._fields: Dict[str, Expression] = {}
+        self._warned_on_shadowed_name = False
+        self._shadowed_fields = set()
 
         for i, (f, t) in enumerate(self.dtype.items()):
             if isinstance(self._ir, ir.MakeStruct):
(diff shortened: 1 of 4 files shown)


patrick-schultz:
@danking This doesn't forbid clashing field names, but accessing the field must use the `struct['items']` syntax. This appears to be what we meant to be doing, but we had a bug. Open to discussion on whether we should just prevent the clashing entirely (but what happens when importing data with a bad field name?).

danking:
Hmm. This also seems fine. I'm gonna assign Chris for a third perspective.

chrisvittal (approved):
LGTM. Just need to fix the parts of the tests where we overwrote items as well.

patrick-schultz:
Ugh, approx_cdf return a struct with a "values" field. No matter what we do to fix the shadowing issue is going to be a breaking change.

danking:
Does `s.values` currently give you the sampled values or the `.values()`?

patrick-schultz:
The sampled values. The method is shadowed and inaccesible.

danking (changes requested):
tests and a warning

patrick-schultz (commented):
Note that this fixes the behavior to aggree with the existing docs on `StructExpression`:

>    However, it is recommended to use square brackets to select fields:
    >>> hl.eval(struct['a'])
    5
>
>    The latter syntax is safer, because **fields that share their name with
    an existing attribute of :class:`.StructExpression` (`keys`, `values`,
    `annotate`, `drop`, etc.) will only be accessible using the
    :meth:`.StructExpression.__getitem__` syntax**. This is also the only way
    to access fields that are not valid Python identifiers, like fields with
    spaces or symbols.
```

File: hail/python/hail/expr/expressions/typed_expressions.py [hunk 1]
```diff
@@ -1778,21 +1780,30 @@ def __init__(self, x, type, indices=Indices(), aggregations=LinkedList(Aggregati
...
+            self._fields[key] = value
 
     def _get_field(self, item):
         if item in self._fields:
             return self._fields[item]
         else:
             raise KeyError(get_nice_field_error(self, item))
 
+    def __getattribute__(self, item):
+        if item in super().__getattribute__('_warn_on_shadowed_name'):
```
danking, line 1799: This avoids infinite recursion, right? And the super's getattribute is able to see the subclasses basic fields?

danking (changes requested):
This is excellent! Can you edit the PR message to include a `CHANGELOG: ` that describes this change to the scientists, including a note about the breaking change (which is OK b/c bug fix).

hail/python/hail/expr/expressions/typed_expressions.py
patrick-schultz: Yes, and yes. The super's getattribute is just what would have been called if we didn't override it (unless there's some special-method magic I don't understand).

PR #13500, opened 2023-08-25
[qob][batch] do not list all jobs on failure (plus: types!)

danking:
CHANGELOG: In Query-on-Batch, the client-side Python code will not try to list every job when a QoB batch fails. This could take hours for long-running pipelines or pipelines with many partitions.

I also added API types to the list jobs end point because I have to go hunting for this every time anyway. Seems better to have this information at our digital fingertips.

### DIFF
--- a/batch/batch/front_end/front_end.py
+++ b/batch/batch/front_end/front_end.py
@@ -44,6 +44,7 @@ from gear.clients import get_cloud_async_fs
 from gear.database import CallError
 from gear.profiling import install_profiler_if_requested
 from hailtop import aiotools, dictfix, httpx, version
+from hailtop.batch_client.types import JobListEntry, GetJobsResponse
 from hailtop.batch_client.parse import parse_cpu_in_mcpu, parse_memory_in_bytes, parse_storage_in_bytes
 from hailtop.config import get_deploy_config
 from hailtop.hail_logging import AccessLogger
@@ -252,7 +253,9 @@ async def _handle_api_error(f: Callable[P, Awaitable[T]], *args: P.args, **kwarg
         raise e.http_response()
 
 
-async def _query_batch_jobs(request: web.Request, batch_id: int, version: int, q: str, last_job_id: Optional[int]):
+async def _query_batch_jobs(
+    request: web.Request, batch_id: int, version: int, q: str, last_job_id: Optional[int]
+) -> Tuple[List[JobListEntry], Optional[int]]:
     db: Database = request.app['db']
     if version == 1:
         sql, sql_args = parse_batch_jobs_query_v1(batch_id, q, last_job_id)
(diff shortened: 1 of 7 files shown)


hail/python/hailtop/batch_client/aioclient.py
```diff
@@ -462,14 +474,21 @@ async def wait(self,
         with BatchProgressBar(disable=disable_progress_bar) as progress2:
             return await self._wait(description, progress2, disable_progress_bar, starting_job)
 
-    async def debug_info(self):
+    async def debug_info(self,
+                         _job_filter: Optional[Callable[[JobListEntry], bool]] = None,
```
daniel-goldstein, line 478: Just a drive by comment: You can use the query string `q` to search for failed jobs instead of filtering client-side

jigold (dismissed):

jigold:
I don't understand how the backwards compatibility is going to work in this case:

1. I have a V1 client, the server is sending me a V2 response with extra fields (or removed fields). What happens here?

danking:
A V1 client cannot hit the V2 endpoint because it doesn't know it exists.

hail/python/hailtop/batch_client/aioclient.py
danking: done

batch/batch/front_end/front_end.py
```diff
@@ -252,7 +253,9 @@ async def _handle_api_error(f: Callable[P, Awaitable[T]], *args: P.args, **kwarg
         raise e.http_response()
 
 
-async def _query_batch_jobs(request: web.Request, batch_id: int, version: int, q: str, last_job_id: Optional[int]):
+async def _query_batch_jobs(
```
jigold, line 256: You're missing the `V1Alpha` here.

jigold (changes requested):
Just so I understand correctly, we can add fields to the V1Alpha type later on, but not remove them or change the type?

danking:
I merged main and was forced to fix the pyright errors, which was great! I found two bugs with these new types:

1. It's possible for a spec file to be missing which means the spec is None which would fail our get job front end code when it tries to fix up the resources.
2. In the aioclient, we assumed the attributes was present and a dict but that key can be completely missing from the dict.

I also noticed `cost_str` has had the wrong type since always.

danking:
> we can add fields to the V1Alpha type later on, but not remove them or change the type?

We should not remove or change fields, the types of fields, or the meaning of fields of JSON dicts returned by our API endpoints because that would break backwards compatibility. 

The types on the other hand can change willy nilly, mypy and pyright have no way of knowing that we changed the definition of, for example, GetJobResponseV1Alpha. They *will* however, go report all the places in our codebase where we expected certain fields to exist if we remove those fields from GetJobResponseV1Alpha.

danking:
@jigold bump

jigold (approved):

PR #13501, opened 2023-08-25
[batch] Search area fixes

jigold:


batch/batch/front_end/templates/table_search.html
```diff
@@ -64,4 +64,13 @@
       </form>
     </div>
   </div>
+  <script>
+    document.getElementById("{{ component_name }}-input-box")
+      .addEventListener("keydown", function(e) {
+        if (e.which === 13 && e.metaKey) {
```
danking, line 70: I think windows users will find this surprising because meta either doesn't exist or is the windows key. Can we do `(e.metaKey || e.ctrlKey)`?

batch/batch/front_end/templates/table_search.html
```diff
@@ -45,8 +45,8 @@
     }
   </style>
   <div class="flex-col">
-    <div id="{{ component_name }}" class="flex-col-align-right">
-      <form method="GET" action="{{ endpoint }}">
+    <div id="{{ component_name }}" class="flex-col-align-left">
```
danking, line 48: What does the left vs right do?

danking (changes requested):

batch/batch/front_end/templates/table_search.html
iris-garden: what do you mean by jumping around? is it moving after the page has finished loading, or do you mean that it's in different places on the batches and jobs pages? if the goal is to get the search box to be aligned to the left side of the page consistently, i tried just changing this to `flex-col` and it did the trick on both the batches and jobs pages.

batch/batch/front_end/templates/table_search.html
```diff
@@ -64,4 +64,13 @@
       </form>
     </div>
   </div>
+  <script>
+    document.getElementById("{{ component_name }}-input-box")
+      .addEventListener("keydown", function(e) {
+        if (e.which === 13 && (e.metaKey || e.ctrlKey)) {
```
iris-garden, line 70: imo this would be a bit easier to read using the value of `e.key` (`"Enter"`) instead of `e.which` (`13`)

iris-garden (changes requested):

jigold:
Screenshot 2023-08-25 at 12 39 29 PM
Screenshot 2023-08-25 at 12 39 12 PM
Screenshot 2023-08-25 at 12 39 04 PM

jigold:
This is the same page. I just moved the window size around and the search box kept moving. I think it's better if it just stays on the left.

batch/batch/front_end/templates/table_search.html
iris-garden: ohh i see what you mean from the screenshots, `flex-col` or `flex-col-align-left` fixes that on my machine

iris-garden (approved):

PR #13503, opened 2023-08-25
[batch] Favor configuration_of and get_remote_tmpdir over get_user_config

daniel-goldstein:
Just some cleanup, all in test files.

### DIFF
--- a/batch/test/conftest.py
+++ b/batch/test/conftest.py
@@ -18,7 +18,7 @@ def log_before_after():
 
 @pytest.fixture(scope='module')
 def remote_tmpdir():
-    return get_remote_tmpdir('batch tests')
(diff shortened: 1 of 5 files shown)


batch/test/conftest.py
```diff
@@ -14,6 +16,11 @@ def log_before_after():
     log.info('ending test')
 
 
+@pytest.fixture(scope='module')
+def remote_tmpdir():
+    return get_remote_tmpdir('batch tests')
```
jigold, line 21:
```suggestion
    return get_remote_tmpdir('batch_tests')
```

jigold (dismissed):

jigold (approved):

PR #13505, opened 2023-08-25
[batch] re-allow sending kwargs to python jobs

danking:
CHANGELOG: Fix bug introduced in 0.2.117 by commit `c9de81108` which prevented the passing of keyword arguments to Python jobs. This manifested as "ValueError: too many values to unpack".

We also weren't preserving tuples. They become lists. I fixed that too. I also added some types and avoided an is instance by encoding the necessary knowledge.


hail/python/hailtop/batch/job.py
```diff
@@ -877,6 +877,19 @@ async def _compile(self, local_tmpdir, remote_tmpdir, *, dry_run=False):
         return True
 
 
+UnpreparedArg = '_resource.ResourceType' | List['UnpreparedArg'] | Tuple['UnpreparedArg', ...] | Dict[str, 'UnpreparedArg'] | Any
```
daniel-goldstein, line 880: I think we have to wait until [3.10](https://peps.python.org/pep-0604/) for this unfortunately.

daniel-goldstein (changes requested):

hail/python/hailtop/batch/resource.py
```diff
@@ -448,3 +448,6 @@ def __str__(self):
 
     def __repr__(self):
         return self._uid  # pylint: disable=no-member
+
+
+ResourceType = PythonResult | ResourceFile | ResourceGroup
```
daniel-goldstein, line 453: Still have pipes here. I didn't mean to get rid of the types altogether though, just replace the pipes with a `Union`

daniel-goldstein (dismissed):

danking:
Ah, my bad, I thought you mean the recursive types with strings. I didn't realize the pipes were new. That's what I get for not clicking links.

devbin/generate_gcp_ar_cleanup_policy.py
```diff
@@ -12,15 +12,17 @@ def to_dict(self):
```
danking, line 12: I don't entirely understand why pyright runs on everything but it does and it dislikes the formatting of this file.

hail/scripts/test_requester_pays_parsing.py
```diff
@@ -5,7 +5,7 @@
 
 from hailtop.aiocloud.aiogoogle import get_gcs_requester_pays_configuration
 from hailtop.aiocloud.aiogoogle.user_config import get_spark_conf_gcs_requester_pays_configuration, spark_conf_path
-from hailtop.config.user_config import ConfigVariable, configuration_of
+from hailtop.config import ConfigVariable, configuration_of
```
danking, line 8: pyright also disliked this

danking: and it's right: ConfigVariable is not part of user_config.py

daniel-goldstein (approved):

danking:
Blocked by https://github.com/hail-is/hail/pull/13536

PR #13506, opened 2023-08-25
[hailctl] Dont copy the user's config file into hailctl batch submit jobs

daniel-goldstein:
CHANGELOG: `hailctl batch submit` now propagates configuration environment variables to the submitted job.

Copying the user's config file into the job feels hacky and rude. Now that everything is guaranteed to be using the `configuration_of` mechanism we can set the entire config through environment variables. Also, if the intention is that the job's environment should magically reflect what the user set up locally, we should use `configuration_of` so that `HAIL_BILLING_PROJECT=foo hailctl batch submit` actually works.


danking:
Agreed, much better. Can we add some tests for some of this functionality?

danking (changes requested):
moving off my CI page

PR #13508, opened 2023-08-28
[query] MT.tail should prefer `n_rows` to `n`.

danking:
CHANGELOG: The `n` parameter of `MatrixTable.tail` is deprecated in favor of a new `n_rows` parameter.

MT.head already does this.


patrick-schultz (dismissed):
Should give a deprecation warning when `n` is used

danking:
bump

patrick-schultz (approved):

PR #13509, opened 2023-08-28
[batch] More debugging information for network namespace errors

daniel-goldstein:
In a previous PR we added exception logging when tasks on a `BackgroundTaskManager` fail. I'm not seeing any of those messages in the worker logs, suggesting that for some reason the Tasks to free a network namespace when a job is finished are not getting created in the first place. I don't see how in the code we could somehow *not* free the network namespace, but I'm hoping that some of these diagnostics shed some light. Also open to suggestions for where else might be a good spot to add more logging.

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1847,7 +1847,7 @@ class DockerJob(Job):
         except asyncio.CancelledError:
             raise
         except Exception:
-            pass
+            log.exception('While running container')
 
     async def run(self):
         async with self.worker.cpu_sem(self.cpu_in_mcpu):
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -1035,7 +1035,7 @@ class Container:
                 else:
                     assert self.network is None or self.network == 'public'
                     self.netns = await network_allocator.allocate_public()
-        except asyncio.CancelledError:
+        except asyncio.TimeoutError:
             log.exception(network_allocator.task_manager.tasks)
             raise
 
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -965,6 +965,7 @@ class Container:
                 self._killed = True
 
     async def _cleanup(self):
+        log.info(f'Cleaning up {self}')
         if self._cleaned_up:
             return
 
@@ -987,6 +988,7 @@ class Container:
             if self.netns:
                 assert network_allocator
                 network_allocator.free(self.netns)
+                log.info(f'Freed the network namespace for {self}')
                 self.netns = None
         finally:
             try:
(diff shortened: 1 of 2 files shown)


batch/batch/worker/worker.py
```diff
@@ -1026,12 +1028,16 @@ async def _setup_overlay(self):
...
-                assert self.network is None or self.network == 'public'
-                self.netns = await network_allocator.allocate_public()
+        try:
+            async with async_timeout.timeout(60):
+                if self.network == 'private':
+                    self.netns = await network_allocator.allocate_private()
+                else:
+                    assert self.network is None or self.network == 'public'
+                    self.netns = await network_allocator.allocate_public()
+        except asyncio.CancelledError:
```
jigold, line 1038: Are you trying to catch a TimeoutError? I think that's different than CancelledError.

jigold (changes requested):
My initial hypothesis was there's a deadlock in iptables somehow. We have the syslog that record all iptable operations. Maybe there's something there for the busted worker?

jigold:
```
resource.type="gce_instance"
logName:"syslog"
labels."compute.googleapis.com/resource_name":"batch-worker-default"
```

daniel-goldstein:
I thought #11986 disabled syslogs. When I try to look at the syslogs for a problematic worker it looks like I only get the syslog output from when the worker first starts up before we reset fluentd.

jigold:
Screenshot 2023-08-28 at 12 48 38 PM

jigold:
If it looks like it could help, I think we should turn it back on.

daniel-goldstein:
> Interesting. When I was helping Sophie earlier, I was able to see them for workers in her namespace??? https://console.cloud.google.com/logs/query;query=resource.type%3D%22gce_instance%22%0AlogName:%22syslog%22%0Alabels.%22compute.googleapis.com%2Fresource_name%22:%22batch-worker-parsa-job-private-39d9d%22;summaryFields=:false:32:beginning;cursorTimestamp=2023-08-27T13:45:49Z;duration=P2D?project=hail-vdc

Are those logs from after the worker container started?

batch/batch/worker/worker.py
daniel-goldstein: right good catch.

batch/batch/worker/worker.py
```diff
@@ -1847,7 +1853,7 @@ async def on_completion():
         except asyncio.CancelledError:
             raise
         except Exception:
-            pass
+            log.exception('While running container')
```
jigold, line 1856: You might want the container name in here.

jigold (changes requested):

jigold (approved):

PR #13510, opened 2023-08-28
[batch] Drop unused SQL functions

jigold:
I want to cleanup some of the old database functions as they're no longer used (`close_batch`) and not tested / not used in production / hard to maintain (`recompute_incremental`). The `recompute_incremental` was there in the early days of Batch when we needed to recompute the user_inst_coll_resources table etc. for a single migration to improve cancelling. We have not used it since.

I double checked `close_batch` isn't used in the Python code, but would appreciate it if you could also check as well.


daniel-goldstein (approved):
Confirmed that I only see these referenced in migration files and see no occurences of `CALL recompute_incremental` or `CALL close_batch` in application code.

PR #13513, opened 2023-08-28
[batch] Tokenize batches_n_complete_states

jigold:
This PR adds a new token and time_completed column to the `batches_n_jobs_in_complete_states` table. The reason for doing this is twofold:

1. This will get rid of the serialization in MJC where every job needs to update the same row in the `batches_n_jobs_in_complete_states` table.
2. We also push marking the batch complete to outside of the MJC stored procedure as a separate database query. The current code for marking a batch complete won't work for job groups (or at least not without a complicated query I don't know how to easily write) because the job group completion state check needs to be vectorized up the ancestors tree.

The mechanism for how this works is there is a new `mark_batch_complete` function in Python that is executed after every MJC. There's also a periodic loop that looks for batches that are complete (n_jobs == n_completed), but have not had state = "complete". This makes sure we eventually update the batch state when complete even if MJC is interrupted. The new `mark_batch_complete` function optimistically tries to update the batch state and time_completed if `n_jobs == n_complete AND state != "complete"`. If an update occurs, then the function issues a callback if specified; otherwise, it just returns. `time_completed` is tracked as a new column on the `batches_n_jobs_in_complete_states` table where we take the GREATEST timepoint when updating a row on duplicate key. Then the time_completed is just equal to the MAX of all entries in the table for that batch.

I'm still proving backwards compatibility to myself without manually copying time_completed to the `batches_n_jobs_in_complete_states` table from the `batches` table. I believe that if we have a running batch:

(A) MJC in SQL uses the new token column when inserting into `batches_n_jobs_in_complete_states`, but all new entries are always inserted at token=0 so the original SQL queries currently running on the server execute the same. However, we no longer mark the batch as complete or set the time_completed.
(B) The original server code will show the batch as not complete for batches that were running during and after the migration. So there is a delay, but it's not incorrect.
(C) The new server code deploys with the new `mark_batch_complete` code that runs periodically. Eventually the newly completed batches since the migration will get set to "complete".

Follow up: Have MJC insert rows at the random token (not hard coded to 0) to get the desired parallelism.

### DIFF
--- a/batch/batch/driver/job.py
+++ b/batch/batch/driver/job.py
@@ -33,12 +33,13 @@ async def mark_batch_complete(db: Database, client_session: httpx.ClientSession,
         '''
 UPDATE batches
 LEFT JOIN LATERAL (
-  SELECT CAST(COALESCE(SUM(n_completed), 0) AS SIGNED) AS n_completed
+  SELECT CAST(COALESCE(SUM(n_completed), 0) AS SIGNED) AS n_completed,
+    CAST(MAX(time_completed) AS SIGNED) AS time_completed
   FROM batches_n_jobs_in_complete_states
   WHERE batches.id = batches_n_jobs_in_complete_states.id
   GROUP BY id
 ) AS states ON TRUE
-SET state = 'complete'
+SET state = 'complete', time_completed = time_completed
 WHERE id = %s AND state != 'complete' AND n_jobs = n_completed;
 ''',
         (batch_id,),
--- a/batch/batch/driver/job.py
+++ b/batch/batch/driver/job.py
@@ -33,7 +33,7 @@ async def mark_batch_complete(db: Database, client_session: httpx.ClientSession,
         '''
 UPDATE batches
 LEFT JOIN LATERAL (
-  SELECT COALESCE(SUM(n_completed), 0) AS n_completed
+  SELECT CAST(COALESCE(SUM(n_completed), 0) AS SIGNED) AS n_completed
   FROM batches_n_jobs_in_complete_states
   WHERE batches.id = batches_n_jobs_in_complete_states.id
   GROUP BY id
@@ -54,8 +54,10 @@ WHERE id = %s AND state != 'complete' AND n_jobs = n_completed;
 SELECT batches.*, cost_t.*, batches_cancelled.id IS NOT NULL AS cancelled, states.*
 FROM batches
 LEFT JOIN LATERAL (
-  SELECT COALESCE(SUM(n_completed), 0) AS n_completed, COALESCE(SUM(n_succeeded), 0) AS n_succeeded,
-    COALESCE(SUM(n_failed), 0) AS n_failed, COALESCE(SUM(n_cancelled), 0) AS n_cancelled
+  SELECT CAST(COALESCE(SUM(n_completed), 0) AS SIGNED) AS n_completed,
+    CAST(COALESCE(SUM(n_succeeded), 0) AS SIGNED) AS n_succeeded,
+    CAST(COALESCE(SUM(n_failed), 0) AS SIGNED) AS n_failed,
+    CAST(COALESCE(SUM(n_cancelled), 0) AS SIGNED) AS n_cancelled
   FROM batches_n_jobs_in_complete_states
   WHERE batches.id = batches_n_jobs_in_complete_states.id
   GROUP BY id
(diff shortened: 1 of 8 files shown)


jigold:
@daniel-goldstein Can you give this a high level look to see if you agree with the overall approach?

jigold:
Most of the changes are duplicating MJC in SQL and changing the SQL queries to aggregate over `batches_n_jobs_in_complete_states` rather than use single point values.

jigold:
Thank you for taking a look!

> 1. I'm not opposed to adding tokens to the batches_n_jobs_in_complete_states table, but I'm not sure why this is related to the other pieces of this PR / job groups. Aren't tokens purely a performance optimization?

The issue is that I started with #13475 and after your insightful comment about keeping the batches and job groups tables in sync, I realized that rather than using the batch_after_update trigger to keep the job groups and batches table states identical, we should just go ahead and directly add a double update to the job groups and batches table wherever a batches update occurs in our current code base. Unfortunately, I got stuck with the MJC trigger with these lines of code:

```sql
    UPDATE batches_n_jobs_in_complete_states
      SET n_completed = (@new_n_completed := n_completed + 1),
          n_cancelled = n_cancelled + (new_state = 'Cancelled'),
          n_failed    = n_failed + (new_state = 'Error' OR new_state = 'Failed'),
          n_succeeded = n_succeeded + (new_state != 'Cancelled' AND new_state != 'Error' AND new_state != 'Failed')
      WHERE id = in_batch_id;

    # Grabbing an exclusive lock on batches here could deadlock,
    # but this IF should only execute for the last job
    IF @new_n_completed = total_jobs_in_batch THEN
      UPDATE batches
      SET time_completed = new_timestamp,
          `state` = 'complete'
      WHERE id = in_batch_id;
    END IF;
```

We can do the double update in the IF statement to both the job groups table for job_group_id = 0 and for the batches table in #13475. However, this SQL code / approach will eventually need to be changed for the full job group implementation. I don't know how to compute `@new_n_completed` grouped by job group and then `total_jobs_in_batch` would need to be computed per job group as well. I don't think you can use for loops in SQL. It might be possible to do this with temporary tables, but I thought it would be better to take a detour from adding job groups and get rid of how we currently do the batch update in MJC to allow for job groups in the future before putting in job groups tables so that I could slot in the appropriate state and time_completed updates to both batches and job_groups tables in the same place rather than relying on a trigger for the updates.

I can think about which set of changes should go first (I'm not wedded to either PR coming first -- just thought this way was conceptually easier to understand when there was just a batches table). I haven't 100% convinced myself this change to tokenize the `batches_n_jobs_in_complete_states` table is absolutely necessary for nested job groups, but it seemed like we would want the performance improvements regardless.

> 2. How come marking the batch as complete is moved into a separate transaction as marking the job complete? If it were in the same transaction wouldn't we not need this healing loop?

This is for performance reasons to avoid serialization and race conditions. If the update occurs in the same transaction, then each transaction will be serialized checking if `n_jobs == n_complete`. If we don't have a `SELECT ... FOR UPDATE`, I couldn't convince myself that there wouldn't be a race condition where a batch completion event isn't accidentally missed. 

I think the healing loop would only happen if the batch driver got restarted in between:
1. CALL mark_job_complete() in the database
2. mark_batch_complete()

If 1. errors, the operation is aborted entirely.

On second thought, it might be possible to use an optimistic locking strategy, but that's more complicated and I'd have to think about it some more.

> It seems to me like it would be preferable to instead first update application code to mark the batch complete if it is not complete, then remove the now redundant marking complete of the batch from the trigger. Then there is no delay after the migration where batches are not complete for some time.

Ah, good point!

jigold:
I talked to @danking about this earlier and he thinks this should be a separate change and the `IF` check for n_jobs = n_completed isn't buying us anything in preventing deadlocks or performance improvements. He's going to look over #11352 to make sure there's nothing we're missing.

danking:
MySQL supports looping. It looks like we could [use a cursor with loop](https://stackoverflow.com/a/16350693/6823256) to iterate through the ancestors of the job group. That seems to me like the simplest possible solution here.

We should leave any possible speed improvements to future work and not entangle that with job groups.

Jackie, my comment during our chat about deadlocks was incorrect. The if statement does prevent deadlocks because it ensures that only one MJC tries to grab an exclusive lock rather than more than one.

danking:
Also sorry for the high latency on a response, apparently Verizon has an outage in my neighborhood until tomorrow morning.

jigold:
Closing this in favor of adding a loop mechanism to #13475. We should revisit this PR at a later date for performance improvements in MJC.

PR #13524, opened 2023-08-30
[query] add hl.utils.table_iterator to iterate through values of a table

danking:
Not the fastest thing, but it makes the impossible possible.

### DIFF
--- a/hail/python/hail/utils/table_iterator.py
+++ b/hail/python/hail/utils/table_iterator.py
@@ -1,0 +1,38 @@
+from typing import Optional
+import hail as hl
+from concurrent.futures import ProcessPoolExecutor
+
+
+ht: Optional[hl.Table] = None
+
+
+def init_process(path):
+    global ht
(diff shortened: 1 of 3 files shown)


danking:
Relies on https://github.com/hail-is/hail/pulls

hail/python/hail/utils/table_iterator.py
```diff
@@ -0,0 +1,38 @@
...
+from concurrent.futures import ProcessPoolExecutor
+
+
+ht: Optional[hl.Table] = None
+
+
+def init_process(path):
+    global ht
+    assert ht is None
+    hl.init()
```
chrisvittal, line 12: Are there any non-default `init` parameters we'd want to consider here? I'm worried about log spam.

danking:
This is not a bad idea but there's too many other higher priority things on our plate; I don't want to distract from the bigger goals.

PR #13534, opened 2023-08-31
[query] All non-spark task context use in using blocks

chrisvittal:
This will ensure proper freeing of resources, no matter what.


danking (approved):
Very reasonable.

Issue #13535, opened 2023-09-01
[query] Hail Python jobs do not work when Python is version greater than 3.9.


danking:
### What happened?

This is a known issue in our underlying serialization library: https://github.com/uqfoundation/dill/issues/609.

I've discovered that if you have dill 0.3.7 (latest) and try to use our image (which has dill 0.3.5.1) you get this error:

Traceback (most recent call last):
  File "<string>", line 27, in <module>
  File "/usr/local/lib/python3.10/site-packages/dill/_dill.py", line 373, in load
    return Unpickler(file, ignore=ignore, **kwds).load()
  File "/usr/local/lib/python3.10/site-packages/dill/_dill.py", line 646, in load
    obj = StockUnpickler.load(self)
  File "/usr/local/lib/python3.10/site-packages/dill/_dill.py", line 805, in _create_code
    return CodeType(args[0], 0, 0, *args[1:])
TypeError: code expected at most 16 arguments, got 19
Switching local dill to the remote version eliminates the error.

Fix is to ensure the *de*-serializer is >0.3.5.1.

### Version

0.2.120

PR #13536, opened 2023-09-01
[python-dill][hailtop] update past the bad version of dill

danking:
Fixes https://github.com/hail-is/hail/issues/13535.

All the new hailgenetics/dill images are now up: https://hub.docker.com/r/hailgenetics/python-dill/tags


jigold:
Can you remind me how these new images get into the hail-vdc artifact registry and on Azure as well?

danking:
Sure! There's a CI build step that copies everything from hailgenetics into the local repo. [mirror_hailgenetics_images](https://github.com/hail-is/hail/blob/main/build.yaml#L3344C19-L3386). Then, in the worker, we notice a limited set of hailgenetics images and, instead of getting them from docker hub, we get them from our internal repositories.

jigold:
We only copy if the image isn't already present. I think this code is not what you want in mirror images. We should copy if the contents change as well.

```
copy_if_not_present() {
    src_image=$1
    dest_image=$2
    if ! skopeo inspect "docker://docker.io/$1";
    then
        echo "$1 does not exist yet, doing nothing"
    elif skopeo inspect "docker://$2";
    then
      echo "$2 already exists, doing nothing"
    else
      echo "$2 does not exist, copying $1 to $2"
      copy_image $1 $2
    fi
}
```

danking:
ugh, that used to just use copy_image and skopeo did The Right Thing (i.e. copy if it the tag had been updated).

danking:
I'll just manually do this right now.

danking:
Done.

danking:
And a PR to fix the root cause https://github.com/danking/hail/pull/new/fix-mirror-hailgenetics-images.

jigold (approved):

PR #13538, opened 2023-09-01
[docker] copy even if the tag already exists

danking:
This inspect command prevents us from updating a tag, for example, if we need to replace an image with a security problem or if there is a bug like the one fixed by https://github.com/hail-is/hail/pull/13536.

### DIFF
--- a/docker/hailgenetics/mirror_images.sh
+++ b/docker/hailgenetics/mirror_images.sh
@@ -11,7 +11,7 @@ copy_if_not_present() {
     then
         echo "$1 does not exist yet, doing nothing"
     else
-      echo "$2 does not exist, copying $1 to $2"
+      echo "copying $1 to $2"
       copy_image $1 $2
     fi
 }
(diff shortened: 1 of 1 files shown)


docker/hailgenetics/mirror_images.sh
```diff
@@ -10,9 +10,6 @@ copy_if_not_present() {
     if ! skopeo inspect "docker://docker.io/$1";
     then
         echo "$1 does not exist yet, doing nothing"
-    elif skopeo inspect "docker://$2";
-    then
-      echo "$2 already exists, doing nothing"
     else
       echo "$2 does not exist, copying $1 to $2"
```
daniel-goldstein, line 14: Sounds good, can you just update this `echo`? It's no longer correct

daniel-goldstein (changes requested):

daniel-goldstein (approved):

PR #13541, opened 2023-09-01
[batch] remove redundant super class

danking:
All these classes inherit from `Resource` which is an `abc.ABC`.


jigold (approved):

jigold:
@danking You pushed changes to the wrong PR.

jigold (approved):

Issue #13526, opened 2023-08-30
[batch] adjust Hail's cost structure to recover operating expenses


danking:
### What happened?

After sorting our costs into "cost of goods", "operating expenses", and "capital expenses", I realized there are four "operating expenses" that are not tracked and reported with the other expenses. I regressed these costs against the core-hours to estimate the cost per core-hour.

resource | intercept (USD) | cost (USD/core-hour)
-- | -- | --
GCP Support Variable fee | 3.46403 +- 0.49155 | 0.00123 +- 0.00007
System logs costs SKU#1 | 13.09991 +- 3.13991 | 0.00093 +- 0.00039
System logs costs SKU#2 | 7.87838 +- 0.81695 | 0.00027 +- 0.00012
Job specifications | 5.41150 +- 0.36608 | 0.00025 +- 0.00005
Firewall policy | 0.51216 +- 0.03185 | 0.00012 +- 0.00000

To fully recover the operating expenses at our current revenue, we need an additional 0.005 USD per core-hour (which is 0.002 more than the sum of intercepts).

This issue is complete after we add a new product:

resource | cost (USD/core-hour)
-- | --
support-logs-specs-and-firewall-fees/1 | 0.005

### Version

0.2.120

jigold:
Was there a reason you unassigned me? Happy to write a draft, but might be better to wait until we want to actually start billing for this new fee to avoid stale code.

danking:
I've already made the announcement. I'll draft the PR soon

PR #13539, opened 2023-09-01
[batch] fix misnamed resources

danking:
I am having trouble determining the effect of this mistake, but it seems like we would be substantially undercharging for the serivce fee if it was really being charged by worker_fraction_in_1024ths instead of core-hours.

PR #13542, opened 2023-09-01
[batch] introduce Support-Logs-Specs-and-Firewall fee

danking:
Fixes https://github.com/hail-is/hail/issues/13526

I verified I got the right number. I didn't import since I think we decided we shouldn't rely on the code base in the migrations.

```
In [1]: from hailtop.utils.rates import rate_instance_hour_to_fraction_msec, rate_cpu_hour_to_mcpu_msec
   ...: 
   ...: rate_cpu_hour_to_mcpu_msec(0.005) == 0.000000000001388888888888889
```


danking:
@jigold not actually WIP, but I want control over when it merges. Ready for review.

batch/batch/cloud/gcp/instance_config.py
```diff
@@ -151,6 +153,7 @@ def from_dict(data: dict) -> 'GCPSlimInstanceConfig':
                 GCPDynamicSizedDiskResource('disk/pd-ssd/1'),
                 GCPIPFeeResource('service-fee/1'),
                 GCPServiceFeeResource('ip-fee/1024/1'),
+                GCPSupportLogsSpecsAndFirewallFees('gcp-support-logs-specs-and-firewall-fees/1'),
```
jigold, line 156: I don't think you need this as it's for backwards compatibility. See comment in #13539.

jigold (changes requested):

batch/batch/cloud/gcp/instance_config.py
danking: Can you elaborate on what kind of backwards compatibility? Should we remove this branch entirely?

jigold: I think we can. It was for when the `instance_config` field was not in the database for old entries, so their value is NULL. We knew what the config should be at the time though. The only reason we'd need this code path is if we wanted to compute resources for old instances. I don't know why we would want to do that.

danking: https://github.com/hail-is/hail/pull/create?base=hail-is%3Amain&head=danking%3Aremove-unnecessary-backwards-compatibility-code&head_repo=hail

jigold (approved):

PR #13544, opened 2023-09-01
[batch] Do not mount the xfs_quota from the worker host

daniel-goldstein:


danking (approved):
See discussion for context: https://github.com/hail-is/hail/pull/13440#discussion_r1313089043

Issue #13242, opened 2023-07-13
[qob] log4j errors are showing up in JVM container logs


jigold:
### What happened?

I expected to see no error messages in the logs for a JVM container. Instead I found errors relating to assertion errors when setting up log4j.

https://console.cloud.google.com/logs/query;query=%2528%0Aresource.type%3D%22gce_instance%22%0AlogName:%22jvm-2%22%0Alabels.%22compute.googleapis.com%2Fresource_name%22:%22batch-worker-default-standard-np-xj3g%22%0A%2529;timeRange=PT1H;summaryFields=:false:32:beginning;cursorTimestamp=2023-07-13T18:49:03.668876534Z?project=hail-vdc

### Version

0.2.119

danking:
Here's a link with an absolute time window: https://cloudlogging.app.goo.gl/gXAWZpZtUiV8jphXA

This is the assertion's stack trace:
```
at scala.Predef$.assert(Predef.scala:208)
at is.hail.QoBOutputStreamManager.createOutputStream(QoBAppender.scala:38)
at org.apache.logging.log4j.core.appender.OutputStreamManager.getOutputStream(OutputStreamManager.java:165)
at org.apache.logging.log4j.core.appender.OutputStreamManager.writeToDestination(OutputStreamManager.java:250)
at org.apache.logging.log4j.core.appender.OutputStreamManager.flushBuffer(OutputStreamManager.java:283)
at org.apache.logging.log4j.core.appender.OutputStreamManager.flush(OutputStreamManager.java:294)
at org.apache.logging.log4j.core.appender.AbstractOutputStreamAppender.directEncodeEvent(AbstractOutputStreamAppender.java:217)
at org.apache.logging.log4j.core.appender.AbstractOutputStreamAppender.tryAppend(AbstractOutputStreamAppender.java:208)
at org.apache.logging.log4j.core.appender.AbstractOutputStreamAppender.append(AbstractOutputStreamAppender.java:199)
at org.apache.logging.log4j.core.config.AppenderControl.tryCallAppender(AppenderControl.java:161)
```

And the line of our code that triggers the logger appender:
```
is.hail.JVMEntryway$2.run(JVMEntryway.java:139)
```

On that line, we should have already evaluated line 97:
```
QoBOutputStreamManager.changeFileInAllAppenders(logFile);
```
Which updates the filename for all `QoBOutputStreamManager`s.

[...]

PR #13546, opened 2023-09-01
[batch] synchronize all access to QoBOutputStreamManager map

danking:
see also https://github.com/hail-is/hail/issues/13242


jigold:
Do you want me to adopt this and make sure it doesn't break anything / see if the error messages disappear?

danking:
Sorry I missed your message! The code as written now is plainly wrong: we access a mutable map from two threads without synchronization. We need this change regardless of how it affects error messages.

If the tests pass, I'm confident this is fine. Are there components of the system you don't think are well tested by our tests?

jigold (approved):

PR #13547, opened 2023-09-01
[website] make the navbar fixed again

danking:
Sep-01-2023 19-13-36

At somepoint we changed the class on the navbar and this broke. It is now fixed.


iris-garden (approved):
lgtm!

Issue #12950, opened 2023-04-28
Hail Batch "multiple clients uploading"


tpoterba:
### What happened?

Daniel M is seeing this bug running the script [here](https://github.com/broadinstitute/tgg_methods/blob/master/vrs/vrs_annotation_batch.py#L275) with the flag --annotate-original.

The Zulip thread [here](https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/Batch.20backend.20.22multiple.20clients.20uploading.22.3F/near/353337596) has more context as well.

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/batch/worker/worker.py", line 2150, in run
    await self.jvm.execute(local_jar_location, self.scratch, self.log_file, self.jar_url, self.argv)
  File "/usr/local/lib/python3.7/dist-packages/batch/worker/worker.py", line 2629, in execute
    raise JVMUserError(exception)
JVMUserError: java.util.concurrent.ExecutionException: java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
    at java.util.concurrent.FutureTask.report(FutureTask.java:122)
    at java.util.concurrent.FutureTask.get(FutureTask.java:192)
    at is.hail.JVMEntryway.retrieveException(JVMEntryway.java:224)
    at is.hail.JVMEntryway.finishFutures(JVMEntryway.java:186)
    at is.hail.JVMEntryway.main(JVMEntryway.java:156)
Caused by: java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
    at is.hail.JVMEntryway$1.run(JVMEntryway.java:107)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.run(FutureTask.java:266)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.run(FutureTask.java:266)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at java.lang.Thread.run(Thread.java:750)
Caused by: java.lang.reflect.InvocationTargetException
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #13551, opened 2023-09-03
[query] update to latest google storage APIs

danking:
It is possible this addresses #12950 for two reasons:
1. In 2.25.0, they rewrote the BlobWriteChannel entirely. [commit](https://github.com/googleapis/java-storage/commit/1b52a1053130620011515060787bada10c324c0b).
2. Also in 2.25.0, after the rewrite, they fixed in issue with tracking offsets after incremental flushes. Maybe the issue existed in the old code too?. [commit](https://github.com/googleapis/java-storage/commit/c099a2f4f8ea9afa6953270876653916b021fd9f).
3. In 2.22.4 they modified the BlobWriteChannel to use a different retry method. [commit](https://github.com/googleapis/java-storage/commit/1b52a1053130620011515060787bada10c324c0b).

Anyway, all circumstantial.

### DIFF
--- a/hail/build.gradle
+++ b/hail/build.gradle
@@ -332,4 +332,5 @@ shadowJar {
 
 task shadowTestJar(type: ShadowJar) {
     archiveClassifier = 'spark-test'
+    configurations = [project.configurations.testRuntimeClasspath]
 }
--- a/hail/build.gradle
+++ b/hail/build.gradle
@@ -85,8 +85,6 @@ compileScala {
 }
 
 configurations {
-    justSpark
-
     all {
         resolutionStrategy {
             eachDependency { DependencyResolveDetails details ->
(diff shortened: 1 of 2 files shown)


chrisvittal (dismissed):
fix the build, but it should be fine

danking:
Snyk is failing because it suddenly realized that we have dependencies (#security 🤦 ). I will fix those issue separately.

danking:
https://github.com/hail-is/hail/pull/13561

danking:
Yikes, this made the Hail wheel nearly 300 MB.

danking:
OK, there was a `justSpark` "configuration". I dunno what that means, it wasn't referenced by the other configurations. I initially made it `implementation` which includes it in the shadow JAR. I switched it to `shadow` b/c mllib (the only thing in the justSpark) should be provided by Spark at runtime anyway.

danking:
@ehigham Chris is out so reassigning to you.

ehigham:
Seems the `testImplementation` doesn't configure the classpath for tests correctly - javatests are currently failing with `Error: Could not find or load main class org.testng.TestNG`

danking:
Hmm. I’ll have to sort this out tomorrow. Not sure what’s going on with that. It seems like the shadowTestJar target is probably not correctly pulling in the testImolemebtation configuration.

danking:
Gah, OK, I think I have it now, but there was one more detail:

The gradle configuration `testCompileOnly` [1] *does not* inherit from the `shadow` configuration (as evidence see [this search](https://github.com/search?q=repo%3Ajohnrengelman%2Fshadow%20extendsFrom&type=code) of the shadow repo). We must explicitly request that `shadow` dependencies are included in the compile-time class path of the tests.

This is as it should be: the things in `shadow` are things which are provided to us by our runtime environment. That's true of both the *test* runtime environment and the normal runtime environment. The Gradle Shadow plugin takes a different perspective, it suggests that `shadow` dependencies shouldn't be used in the tests at all.


[1] NB: `testCompile` does not exist but you don't get an error if you try to use it, thanks for nothing gradle.

danking:
Alright alright jeez I can't program any more. I also needed to explicitly request that all the test and main classes from our project should be in the JAR.

danking:
bump @ehigham

ehigham (approved):
If it builds and the tests run I think it's probably ok

PR #13561, opened 2023-09-05
[gradle] update htsjdk and avro 

danking:
Snyk claims they have transitive vulnerabilities.


danking:
OK, the underlying PR merged. This is ready for a look!

danking:
There's one remaining snyk vulnerability which we cannot resolve because htsjdk is no longer publishing new versions compatible with Java 11.

chrisvittal (approved):

PR #13564, opened 2023-09-06
[qob] cleanup QoB tests without erroring

danking:
In main, we assumed that if `self._batch is not None` then there is a Batch that needs cancelling, but that is no longer true. Now, we must track whether we have called `submit` or not.

I made a few other minor cleanups of `ServiceBackend` while I was in there. There is no longer any need to have a `None` batch b/c there is no distinction between a builder and a batch now.

Example error:
```
INFO     hailtop.aiocloud.aiogoogle.credentials:credentials.py:92 using credentials file /test-gsa-key/key.json: GoogleServiceAccountCredentials for testns-test-418@hail-vdc.iam.gserviceaccount.com
_________ ERROR at teardown of Tests.test_loop_with_struct_of_strings __________

init_hail = None
request = <SubRequest 'set_query_name' for <TestCaseFunction test_loop_with_struct_of_strings>>

    @pytest.fixture(autouse=True)
    def set_query_name(init_hail, request):
        backend = current_backend()
        if isinstance(backend, ServiceBackend):
            backend.batch_attributes = dict(name=request.node.name)
            yield
            backend.batch_attributes = dict()
            references = list(backend._references.keys())
            for rg in references:
                backend.remove_reference(rg)
            backend.initialize_references()
            if backend._batch:
                report: Dict[str, CollectReport] = request.node.stash[test_results_key]
                if any(r.failed for r in report.values()):
>                   log.info(f'cancelling failed test batch {backend._batch.id}')

test/hail/conftest.py:81: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/local/lib/python3.9/dist-packages/hailtop/batch_client/aioclient.py:347: in id
    self._raise_if_not_created()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <hailtop.batch_client.aioclient.Batch object at 0x7ffae6f11bb0>

    def _raise_if_not_created(self):
        if not self.is_created:
>           raise BatchNotCreatedError
E           hailtop.batch_client.aioclient.BatchNotCreatedError

```

https://batch.hail.is/batches/7950601/jobs/156


hail/python/hail/backend/service_backend.py
```diff
@@ -203,7 +203,7 @@ async def create(*,
                      worker_cores: Optional[Union[int, str]] = None,
                      worker_memory: Optional[str] = None,
                      name_prefix: Optional[str] = None,
-                     token: Optional[str] = None,
+                     credentials_token: Optional[str] = None,
```
danking, line 206: ServiceBackend.create is a non-public API and the word token is overloaded in this class. Now `token` only means the idempotency token of a batch.

hail/python/hail/backend/service_backend.py
```diff
@@ -425,14 +429,12 @@ async def _rpc(self,
                         await write_str(infile, bucket)
                         await write_str(infile, mount_point)
                         await write_bool(infile, readonly)
-                    await inputs(infile, token)
+                    await inputs(infile, self._batch.token)
```
danking, line 432: This has been incorrect since we kept the Batch longer than the `_rpc` method's lifetime. The token would get regenerated repeatedly even though the batch never changed.

hail/python/hail/backend/service_backend.py
```diff
@@ -293,7 +293,7 @@ def __init__(self,
                  *,
                  billing_project: str,
                  sync_fs: FS,
-                 async_fs: AsyncFS,
+                 async_fs: RouterAsyncFS,
```
danking, line 296: this fixes a type error in `validate_file`. I'm not sure why mypy misses this.

danking:
bump

jigold (approved):

PR #13565, opened 2023-09-06
[batch] Allow hailgenetics/vep images to be public

jigold:


danking (changes requested):
Hmm. We need a new service account without GAR access.

The publicly available images are now the same as the HAIL_GENETICS_IMAGES. Let's change this to explicitly copy over all those images as publicly available ones.

jigold:
Can you be more specific on what you want the fix to be? Is there an alternate short-term fix here? What is wrong with this change?

danking (dismissed):

danking (approved):

danking:
removing high priority as I need to fix dataproc first.

PR #13578, opened 2023-09-06
[hailctl] Update hailctl auth logout to support OAuth2 credentials

daniel-goldstein:
Tested the following in GCP and Azure:
1. `hailctl auth login` on the current release (pre-OAuth2) and `hailctl auth logout` on `main`
2. Both login/logout on `main`.


daniel-goldstein:
It was while I fixed a bug but then forgot to reassign

daniel-goldstein:
@danking bump

danking (approved):
lgtm

PR #13579, opened 2023-09-06
[auth] Have services use OAuth2 acess tokens instead of their hail tokens

daniel-goldstein:


danking (approved):
lgtm, though I guess I was expecting an identity file rather than an env. Seems fine either way.

Issue #12936, opened 2023-04-26
hl.vep is flaky on dataproc, particularly when using more than two workers


danking:
### What happened?

Julia Sealock reported this https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/vep.20issue/near/352790173

We also saw it in test_dataproc. Cal also reported it.
```
org.apache.spark.SparkException: Job aborted due to stage failure: Task 56 in stage 4.0 failed 20 times, most recent failure: Lost task 56.19 in stage 4.0 (TID 48622) (jsealock-schema-sw-43bq.c.daly-neale-sczmeta.internal executor 3): is.hail.utils.HailException: VEP command '/vep --format vcf --json --everything --allele_number --no_stats --cache --offline --minimal --assembly GRCh38 --fasta /opt/vep/.vep/homo_sapiens/95_GRCh38/Homo_sapiens.GRCh38.dna.toplevel.fa.gz --plugin LoF,loftee_path:/opt/vep/Plugins/,gerp_bigwig:/opt/vep/.vep/gerp_conservation_scores.homo_sapiens.GRCh38.bw,human_ancestor_fa:/opt/vep/.vep/human_ancestor.fa.gz,conservation_file:/opt/vep/.vep/loftee.sql --dir_plugins /opt/vep/Plugins/ -o STDOUT' failed with non-zero exit status 125
  VEP Error output:
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
See 'docker run --help'.

	at is.hail.utils.ErrorHandling.fatal(ErrorHandling.scala:17)
	at is.hail.utils.ErrorHandling.fatal$(ErrorHandling.scala:17)
	at is.hail.utils.package$.fatal(package.scala:78)
	at is.hail.methods.VEP$.waitFor(VEP.scala:73)
	at is.hail.methods.VEP.$anonfun$execute$5(VEP.scala:231)
	at scala.collection.Iterator$$anon$11.nextCur(Iterator.scala:486)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:492)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:490)
	at is.hail.utils.richUtils.RichContextRDD$$anon$1.hasNext(RichContextRDD.scala:69)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:490)
[...]

PR #13580, opened 2023-09-06
[dataproc][vep] mitigate Docker bug

danking:
CHANGELOG: Mitigate #12936 in which VEP Dataproc clusters fail to start. The root cause is complex. Docker has a bug which prevents it from cleanly starting if it is *re* installed. Whatever Google is doing in Dataproc to configure their Docker "component" appears to trigger this bug.

See for details: https://github.com/hail-is/hail/issues/12936#issuecomment-1709120751

The basic fix is to sleep to allow the system to coalesce a bit and then to restart Docker.

### DIFF
--- a/hail/python/hailtop/hailctl/dataproc/resources/vep-GRCh37.sh
+++ b/hail/python/hailtop/hailctl/dataproc/resources/vep-GRCh37.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -x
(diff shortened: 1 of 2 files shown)


daniel-goldstein (approved):

danking:
I merged this into the other approved dataproc fix PR: https://github.com/hail-is/hail/pull/13573

PR #13581, opened 2023-09-06
[ci] Start update loop after watched branches are initialized

daniel-goldstein:
#13331 moved the initialization of the `WatchedBranch`s out of the top level in `ci.py` and into the end of `on_startup`. Notice that this new location is both after `app['task_manager'].ensure_future(update_loop(app))` is run *and* after an `await`-point. It's quite likely that the first time `update_loop` is run is *before* there are any watched branches. This loop then sleeps for five minutes before it runs again. I believe this is why sometimes on startup it can take multiple minutes before CI loads PR information.

### DIFF
--- a/ci/ci/ci.py
+++ b/ci/ci/ci.py
@@ -742,6 +742,8 @@ SELECT frozen_merge_deploy FROM globals;
 
     app['frozen_merge_deploy'] = row['frozen_merge_deploy']
 
+    app['task_manager'] = aiotools.BackgroundTaskManager()
+
     if DEFAULT_NAMESPACE == 'default':
         kubernetes_asyncio.config.load_incluster_config()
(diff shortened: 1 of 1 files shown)


ci/ci/ci.py
```diff
@@ -767,6 +764,9 @@ async def on_startup(app):
         )
     ]
 
+    app['task_manager'] = aiotools.BackgroundTaskManager()
```
danking, line 767: gotta keep task manager up above b/c of the update envoy configs and cleanup expired namespaces.

danking (dismissed):

ci/ci/ci.py
daniel-goldstein: right right

PR #13585, opened 2023-09-07
[hailtop] Fix defaulting to hail.is when the user has no deploy-config

daniel-goldstein:
or domain specified. The key bug here was a check for `'domain' not in config` instead of `config['domain'] is None`


danking (dismissed):
@daniel-goldstein can you just add the release change and change log to this PR?

danking (approved):

PR #13588, opened 2023-09-07
[WIP] QoB Saige Implementation

jigold:
@danking The latest version is the code in hail/methods/. I'm having trouble with all of the configs and how to instantiate that properly. After that, I need to figure out what inputs `run_saige` actually needs. Then I need to write util functions for creating the testing chunks and annotating the matrix table. Then I think after testing and cleaning it up, it will be sufficient for the workshop. To get it into main is going to be a lot more work to have helpful error messages, check MT is valid for this analysis, integrate it more carefully into QoB.

### DIFF
--- a/hail/python/hail/methods/saige/batch_utils.py
+++ b/hail/python/hail/methods/saige/batch_utils.py
@@ -1,0 +1,22 @@
+from typing import Dict
+
+import hailtop.batch as hb
+
+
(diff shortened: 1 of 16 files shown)


jigold:
Also, I forgot making sure all of the Docker images are valid and integrated into CI like the vep images.

PR #13594, opened 2023-09-08
[fs] consistent use of the term FileListEntry

danking:
In our asyncio FSes, we use the term FileListEntry for the package of information you receive from a cloud object store list operation. We reserve the term FileStatus for the metadata you get about a blob without a list operation. A major difference is that FLEs know if they refer to files or folders. In contrast, a FileStatus knows that it is a file *but does not know if a directory with the same name exists*.

### DIFF
--- a/hail/python/hail/fs/hadoop_fs.py
+++ b/hail/python/hail/fs/hadoop_fs.py
@@ -3,23 +3,26 @@ import json
 import time
 from typing import Dict, List, Union, Any
 
-import dateutil
+import dateutil.parser
 
 from hailtop.fs.fs import FS
-from hailtop.fs.stat_result import FileType, StatResult
+from hailtop.fs.stat_result import FileType, FileListEntry
 
 
-def _stat_dict_to_stat_result(stat: Dict[str, Any]) -> StatResult:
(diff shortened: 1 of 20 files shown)


hail/python/hail/fs/hadoop_fs.py
```diff
@@ -60,12 +69,23 @@ def is_file(self, path: str) -> bool:
...
+    def fast_stat(self, path: str) -> FileStatus:
+        '''Get information about a path other than its file/directory status.
+
+        In the cloud, determining if a given path is a file, a directory, or both is expensive. This
+        method simply returns file metadata if there is a file at this path. If there is no file at
+        this path, this operation will fail. The presence or absence of a directory at this path
+        does not affect the return value of this method at all.
+
+        '''
+        raise NotImplementedError
```
danking, line 81: Implementing this operation in Java is a bit complicated and this PR already has many lines. Trust that I will follow up with an implementation.

danking:
To be abundantly clear: in Scala this is strictly a rename. In Python, I created a FileStatus object that is effectively unused.

daniel-goldstein (approved):

PR #13596, opened 2023-09-08
[auth|batch|aiogoogle] Favor latent credentials over specifying /gsa-key/key.json in application code

daniel-goldstein:
This is mostly cleanup reducing the amount of the codebase that depends on GSA key files (which are not available in terra or a future keyless hail-vdc). The core bit is instead of threading `credentials_file="/gsa-key/key.json` through the batch and auth codebase we set `GOOGLE_APPLICATION_CREDENTIALS` in their deployments. I can't think of a scenario where `auth` or `batch` would need to use multiple identities so better to remove their ability to do so and always use the default credentials.

I also did a bit of tidying up, using `$GOOGLE_APPLICATION_CREDENTIALS` instead of the hard-coded path and removing the credentials endpoints on the batch-driver which have been unused by workers for many months now.

### DIFF
--- a/batch/batch/cloud/azure/driver/driver.py
+++ b/batch/batch/cloud/azure/driver/driver.py
@@ -28,7 +28,6 @@ class AzureDriver(CloudDriver):
         machine_name_prefix: str,
         namespace: str,
         inst_coll_configs: InstanceCollectionConfigs,
-        credentials_file: str,
         task_manager: aiotools.BackgroundTaskManager,  # BORROWED
     ) -> 'AzureDriver':
         azure_config = get_azure_config()
@@ -56,12 +55,10 @@ ON DUPLICATE KEY UPDATE region = region;
         with open(os.environ['HAIL_SSH_PUBLIC_KEY'], encoding='utf-8') as f:
             ssh_public_key = f.read()
 
-        arm_client = aioazure.AzureResourceManagerClient(
-            subscription_id, resource_group, credentials_file=credentials_file
-        )
-        compute_client = aioazure.AzureComputeClient(subscription_id, resource_group, credentials_file=credentials_file)
-        resources_client = aioazure.AzureResourcesClient(subscription_id, credentials_file=credentials_file)
-        network_client = aioazure.AzureNetworkClient(subscription_id, resource_group, credentials_file=credentials_file)
+        arm_client = aioazure.AzureResourceManagerClient(subscription_id, resource_group)
+        compute_client = aioazure.AzureComputeClient(subscription_id, resource_group)
+        resources_client = aioazure.AzureResourcesClient(subscription_id)
+        network_client = aioazure.AzureNetworkClient(subscription_id, resource_group)
         pricing_client = aioazure.AzurePricingClient()
 
         region_monitor = await RegionMonitor.create(region)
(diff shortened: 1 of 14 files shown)


batch/batch/cloud/azure/driver/driver.py
```diff
@@ -56,12 +55,10 @@ async def create(
         with open(os.environ['HAIL_SSH_PUBLIC_KEY'], encoding='utf-8') as f:
             ssh_public_key = f.read()
 
-        arm_client = aioazure.AzureResourceManagerClient(
-            subscription_id, resource_group, credentials_file=credentials_file
-        )
-        compute_client = aioazure.AzureComputeClient(subscription_id, resource_group, credentials_file=credentials_file)
-        resources_client = aioazure.AzureResourcesClient(subscription_id, credentials_file=credentials_file)
-        network_client = aioazure.AzureNetworkClient(subscription_id, resource_group, credentials_file=credentials_file)
```
daniel-goldstein, line 64: I think we don't *need* to specify `AZURE_APPLICATION_CREDENTIALS` because the `aioazure` credentials pick up on `/gsa-key/key.json` even though `aiogoogle` does not. It's unclear to me how to resolve this asymmetry. Better to specify `AZURE_APPLICATION_CREDENTIALS` for consistency?

datasets/extract/extract_dbSNP.py
```diff
@@ -20,7 +20,7 @@
     version = builds[build]["version"]
     j = batch.new_job(name=f"{name}_{version}_{build}")
     j.image("gcr.io/broad-ctsa/datasets:050521")
-    j.command("gcloud -q auth activate-service-account --key-file=/gsa-key/key.json")
+    j.command("gcloud -q auth activate-service-account --key-file=$GOOGLE_APPLICATION_CREDENTIALS")
```
danking, line 23: does gsutil not pick up GOOGLE_APPLICATION_CREDENTIALS automatically?

hail/python/hailtop/aiocloud/aiogoogle/credentials.py
```diff
@@ -79,13 +79,10 @@ def default_credentials(scopes: Optional[List[str]] = ..., *, anonymous_ok: Lite
...
-
-        if credentials_file is None:
-            if "HOME" in os.environ:
-                application_default_credentials_file = f'{os.environ["HOME"]}/.config/gcloud/application_default_credentials.json'
-                if os.path.exists(application_default_credentials_file):
-                    credentials_file = application_default_credentials_file
+        credentials_file = first_extant_file(
+            os.environ.get('GOOGLE_APPLICATION_CREDENTIALS'),
+            f'{os.environ["HOME"]}/.config/gcloud/application_default_credentials.json' if 'HOME' in os.environ else None,
+        )
```
danking, line 85: nice

batch/batch/cloud/azure/driver/driver.py
danking: My inclination is to exclude it rather than confuse folks in the future with "what was this env-var doing, actually?"

danking (changes requested):
very minor query

danking (approved):

danking:
Internet confirms that GOOGLE_APPLICATION_CREDENTIALS env-var is ignored by gsutil. 🤦

PR #13604, opened 2023-09-11
[gear] Turn check_csrf_token into middleware

daniel-goldstein:
This is half code cleanup and half guardrail. The key assertion here is: if the request authenticates using a cookie and is attempting a state-changing HTTP method, it should pass a CSRF check.

I tested this with the following:
1. dev deployed and loaded the billing projects page
2. Deleted the hidden csrf input from one of the forms
3. Submitted the form
4. Got a 401


gear/gear/csrf.py
```diff
@@ -1,25 +1,27 @@
...
 
 
-def check_csrf_token(fun):
-    @wraps(fun)
-    async def wrapped(request, *args, **kwargs):
+@web.middleware
+async def check_csrf_token(request: web.Request, handler: AIOHTTPHandler):
+    # CSRF prevention is only relevant to requests that use browser
+    # cookies for authentication.
+    if request.cookies and request.method in request.POST_METHODS:
```
daniel-goldstein, line 19: > request.POST_METHODS

`pyright` showed me this handy set but it doesn't seem to be documented. Perhaps we should just do `request.method == 'POST'`?

danking: let's list the "safe" methods instead.

danking (dismissed):

danking (approved):

PR #13607, opened 2023-09-11
[batch] Add SKU to latest_product_versions

jigold:
This change is to ensure that the SKU we think is associated with a product doesn't change in case the human readable name changes to a different SKU. This was raised as a good safety feature by #13430

### DIFF
--- a/batch/batch/driver/billing_manager.py
+++ b/batch/batch/driver/billing_manager.py
@@ -66,7 +66,9 @@ class ProductVersions:
         for product, new_product_info in new_data.items():
             if product not in self._product_versions:
                 valid_updates[product] = new_product_info
-                log.info(f'added product {product} with version {new_product_info.latest_version} and sku {new_product_info.sku}')
+                log.info(
+                    f'added product {product} with version {new_product_info.latest_version} and sku {new_product_info.sku}'
+                )
 
         self._product_versions.update(valid_updates)
 
--- a/batch/batch/driver/billing_manager.py
+++ b/batch/batch/driver/billing_manager.py
@@ -1,5 +1,6 @@
 import abc
 import logging
+from collections import namedtuple
 from typing import Dict, List, Optional
 
 from gear import Database, transaction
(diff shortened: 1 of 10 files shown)


jigold:
I checked the database and was surprised to see the SKUs weren't necessarily unique to a specific region. But it makes sense when I looked at their API here: https://cloud.google.com/billing/docs/reference/rest/v1/services.skus/list#sku

I think we should put this in and address what happens if they change the SKU of a particular region if that occurs in the future. We'll just get a bunch of error messages with no price updates and it shouldn't impact users.

I will also manually check this in Azure.

jigold:
bump

batch/batch/driver/billing_manager.py
```diff
@@ -9,40 +10,58 @@
 log = logging.getLogger('billing_manager')
 
 
+ProductVersionInfo = namedtuple('ProductVersionInfo', ['latest_version', 'sku'])
```
daniel-goldstein, line 13: Can we use a type that encodes that `sku` is an `Optional[str]`? A `dataclass` might be more appropriate here.

batch/batch/driver/pricing.py
```diff
@@ -9,6 +9,7 @@ class Price(abc.ABC):
     effective_start_date: int
     effective_end_date: Optional[int]
     time_updated: int
+    sku: str
```
daniel-goldstein, line 12: Does this cause a `pyright` error if you don't add a `sku` attribute to any of the subclasses? AFAIK forgetting to set `sku` would only error at runtime. Instead of making these declarations and require subclasses to set `self.sku = sku` in every single subclass, can you make a constructor here in `Price` that sets these and the subclasses can call `super().__init__(region, effective_start_date, …)`? That way there's an easy statically caught error that you didn't call the constructor with the right number of arguments.

Side note: Do you know if these are really *instance attribute* declarations and not class attribute declarations? I tried playing around in `IPython` and it looks like by not setting them they don't appear as attributes of the class, but they're not required to be attributes of the instance either, so I'm not really sure what they're doing in this case.

daniel-goldstein (changes requested):
Can you check my understanding here? The premise is that
1. We record GCP products that match certain product descriptions
2. Those descriptions are subject to change
3. Products have a 1:1 correspondence with GCP SKUs
4. We do not want a rename in a description to update product X with SKU A to now have SKU B

I agree with point 4 that our system should not have that property, since SKUs are, as I understand it, the ID for that product in GCP and should never change. So I'm in favor of this safeguard. But this makes me think that we are doing a bad thing with this description matching, as it's very wild west and if they are subject to change they are not meant to be relied upon. I feel like in the future we should change our system to instead just sync a pre-determined list of SKUs that we have committed to git. Then we don't have to worry about any of this.

jigold:
Your understanding is correct. I don't see how a hard coded list in the repo is any different than a compare and swap with the SKU stored in the database. I don't think we can get around parsing the description in some form when we initialize the product. But I'm open to suggestions on a better way to manage the automatic resource parsing.

daniel-goldstein (approved):

daniel-goldstein:
> I don't see how a hard coded list in the repo is any different than a compare and swap with the SKU stored in the database. I don't think we can get around parsing the description in some form when we initialize the product. 

I think I'm missing context here, especially surrounding what the source of truth is for a "product". Maybe we should have a meeting or discussion in a different forum at some point? I don't know how volatile the product descriptions are, but my current understanding is that this is all pretty low risk/priority.

For posterity, my main concern is around reproducibility. Assume a description change happens tomorrow and now however we're parsing the description ends up pointing at a different SKU. Now at any point in the future, if we have some data loss and have to rebuild these tables, the SKU will suddenly change, but only because we rebuilt the database not because of some actual change in GCP. Or if we or anyone else stands up a new batch cluster, they will get a SKU that differs from what we have currently in `hail-vdc`, just because they started up their cluster at a different point in time. With my current understanding that is an undesirable scenario, but I don't actually know what the implications of such a scenario are.

jigold:
Discussion in a different forum sounds good. This came up with the GPU branch as there was a question on whether we can trust what we are parsing as a resource or did we need to hardcode the SKUs explicitly. This was my proposed solution for us at least knowing that something has changed with lots of driver error messages and to not try and do any updates if the invariants of the "SKU" don't hold.

daniel-goldstein:
Very reasonable! I think that makes a lot of sense and am glad we'll have the extra security.

PR #13608, opened 2023-09-12
[query] add less biased unsmoothed pdf to ggplot

patrick-schultz:
… and make the default

Examples
Histogram without setting min/max. Requires two passes over data, has low resolution over interesting part of the distribution:
Screenshot 2023-09-12 at 8 26 54 AM

Histogram with manual min/max. Most accurate, but different choices of number of bins cause different artifacts. Requires knowledge of distribution beforehand.
Screenshot 2023-09-12 at 8 28 19 AM

Smoothed `approx_cdf` based pdf (old default). Smoothing causes large distortion.
Screenshot 2023-09-12 at 8 30 16 AM

New unsmoothed `approx_cdf_ based pdf. Single pass, works well for any distribution, needs no tuning or foreknowledge of distribution.
Screenshot 2023-09-12 at 8 31 20 AM

### DIFF
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -165,7 +165,10 @@ def _error_from_cdf_python(cdf, failure_prob, all_quantiles=False):
     def compute_single_error(s, failure_prob=failure_prob):
         return math.sqrt(math.log(2 / failure_prob) * s / 2)
 
-    if all_quantiles:
+    if s == 0:
+        # no compactions ergo no error
+        return 0
+    elif all_quantiles:
         p = compute_grid_size(s)
         return 1 / p + compute_single_error(s, failure_prob / p)
     else:
--- a/hail/python/hail/expr/functions.py
+++ b/hail/python/hail/expr/functions.py
@@ -128,6 +128,50 @@ def _error_from_cdf(cdf, failure_prob, all_quantiles=False):
         return hl.rbind(cdf, lambda cdf: hl.rbind(compute_sum(cdf), compute_single_error))
 
 
+def _error_from_cdf_python(cdf, failure_prob, all_quantiles=False):
+    """Estimates error of approx_cdf aggregator, using Hoeffding's inequality.
+
+    Parameters
+    ----------
+    cdf : :obj:`dict`
+        Result of :func:`.approx_cdf` aggregator, evaluated to a python dict
+    failure_prob: :obj:`float`
+        Upper bound on probability of true error being greater than estimated error.
+    all_quantiles: :obj:`bool`
+        If ``True``, with probability 1 - `failure_prob`, error estimate applies
+        to all quantiles simultaneously.
+
+    Returns
+    -------
+    :obj:`float`
+        Upper bound on error of quantile estimates.
+    """
+    import math
+
+    s = 0
+    for i in builtins.range(builtins.len(cdf._compaction_counts)):
+        s += cdf._compaction_counts[i] << (2 * i)
+    s = s / (cdf.ranks[-1] ** 2)
+
+    def update_grid_size(p):
+        return 4 * math.sqrt(math.log(2 * p / failure_prob) / (2 * s))
+
+    def compute_grid_size(s):
+        p = 1 / failure_prob
+        for _ in builtins.range(5):
+            p = update_grid_size(p)
+        return p
+
+    def compute_single_error(s, failure_prob=failure_prob):
+        return math.sqrt(math.log(2 / failure_prob) * s / 2)
+
+    if all_quantiles:
+        p = compute_grid_size(s)
+        return 1 / p + compute_single_error(s, failure_prob / p)
+    else:
+        return compute_single_error(s, failure_prob)
+
+
 @typecheck(t=hail_type)
 def missing(t: Union[HailType, str]):
     """Creates an expression representing a missing value of a specified type.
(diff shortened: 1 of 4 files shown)


hail/python/hail/ggplot/geoms.py
```diff
@@ -484,6 +485,102 @@ def geom_histogram(mapping=aes(), *, min_val=None, max_val=None, bins=None, fill
     return GeomHistogram(mapping, min_val=min_val, max_val=max_val, bins=bins, fill=fill, color=color, alpha=alpha, position=position, size=size)
 
 
+def _max_entropy_cdf(min_x, max_x, x, y, e):
```
ehigham, line 488: This function will be hard to maintain by anyone else without significant effort. Please can you split the code up with some functions and/or provide a link to whatever you're basing this implementation off.

ehigham (changes requested):

hail/python/hail/ggplot/geoms.py
patrick-schultz: You're right, of course. When I thought we'd be using this in the workshop, I just wanted to get the change in fast. This is a copy of the same function in the old plotting library, I'll find a better place for this to live so both can use the same function.

This is a completely novel method, so there are no references to link to. I don't know how to explain it without drawing pictures, but I'll do my best.

hail/python/hail/ggplot/geoms.py
```diff
@@ -493,46 +590,79 @@ class GeomDensity(Geom):
...
+
+            if self.smoothed:
+                n = data['ranks'][-1]
+                weights = np.diff(data['ranks'][1:-1])
+                min = data['values'][0]
+                max = data['values'][-1]
+                values = np.array(data['values'][1:-1])
+                slope = 1 / (max - min)
+
+                def f(x, prev):
```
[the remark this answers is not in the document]
patrick-schultz: That's for the smoothed plot. I wrote that (in the old plotting library) before I came up with the idea behind the unsmoothed max-entropy cdf based plot. I'm tempted to just delete the smoothed version. It's too easy to introduce serious artifacts, and the unsmoothed one works so well.

hail/python/hail/expr/functions.py
```diff
@@ -128,6 +128,50 @@ def compute_global_error(s):
...
+    def update_grid_size(p):
+        return 4 * math.sqrt(math.log(2 * p / failure_prob) / (2 * s))
+
+    def compute_grid_size(s):
+        p = 1 / failure_prob
+        for _ in builtins.range(5):
+            p = update_grid_size(p)
+        return p
+
+    def compute_single_error(s, failure_prob=failure_prob):
```
ehigham, line 165: Do you intend to pull these inner functions out or can we just get `failure_prob` from the enclosing scope (ie remove the parameter)?

hail/python/hail/ggplot/geoms.py
```diff
@@ -483,6 +484,117 @@ def geom_histogram(mapping=aes(), *, min_val=None, max_val=None, bins=None, fill
...
+            return max_x, 1
+        else:
+            yi = y[i] + e if upper else y[i+1] - e
+            return x[i], yi
+
+    def slope_from_fixed(i, upper):
+        xi, yi = point_on_bound(i, upper)
+        return (yi - fy) / (xi - fx)
+
+    def update_min_max_slopes():
```
ehigham, line 542: I think it would be clearer to return a tuple and unpack that at the call site rather than use side effects:
```python
min_slope, max_slop = min_max_slopes()
```

hail/python/hail/ggplot/geoms.py
```diff
@@ -483,6 +484,117 @@ def geom_histogram(mapping=aes(), *, min_val=None, max_val=None, bins=None, fill
...
+        nonlocal min_slope, max_slope
+        min_slope = slope_from_fixed(li, upper=False)
+        max_slope = slope_from_fixed(ui, upper=True)
+
+    def fix_point_on_result(i, upper):
+        nonlocal fx, fy, new_y, keep
```
ehigham, line 548: same here. I guess I'm not a fan of this pattern.

hail/python/hail/ggplot/geoms.py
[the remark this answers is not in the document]
ehigham: Oh i see this isn't new code, just indented with the `if` statement.

ehigham (dismissed):
Thanks for making such an effort to make this more readable. I have a couple of small suggestions then we can merge this.

hail/python/hail/expr/functions.py
patrick-schultz: The parameter is there because it gets called with different values. This function computes the error bound on an estimated rank for a single value, with a given probability of exceeding the bound. We sometimes want an error bound that applies to *all* possible values simultaneously, with a given probability of *any* rank estimate exceeding the bound. Computing that involves computing an error bound for a single value with an appropriately smaller failure probability.

hail/python/hail/ggplot/geoms.py
ehigham: As discussed, there's not need to change this one. You raised a good point that if this were a class and these variables were attributes of the class, then perhaps I wouldn't object - which is certainly true. I don't think it's worth re-writing this as a class. I think this is fine as you're indexing into and mutating state variables, just please define the variables before you use them in this function so it's clear what you're referencing.

PR #13610, opened 2023-09-12
[query/ggplot] runs plotting notebook setup on import

iris-garden:
CHANGELOG: Additional setup is no longer required when using `hail.plot` or `hail.ggplot` in a Jupyter notebook (calling `bokeh.io.output_notebook` or `hail.plot.output_notebook` and/or setting `plotly.io.renderers.default = 'iframe'` is no longer necessary).

I double-checked locally that `import hail as hl` and `from hail.ggplot import ggplot` both cause the added code to run, and that running them in a notebook does the setup we expect it to, but running it in iPython or the Python interpreter does not.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2965,7 +2965,6 @@ steps:
       export HAIL_DEFAULT_NAMESPACE={{ default_ns.name }}
       export GOOGLE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
       export AZURE_APPLICATION_CREDENTIALS=/test-gsa-key/key.json
-      cd /io/hailtop/batch
       hailctl config set batch/billing_project test
       hailctl config set batch/remote_tmpdir {{ global.test_storage_uri }}/test_batch_docs/{{ token }}/
       python3 -m pytest \
@@ -2980,7 +2979,8 @@ steps:
               --doctest-glob='*.rst' \
               --ignore=docs/cookbook/files/ \
               --timeout=120 \
-              --ignore=docs/conf.py
+              --ignore=docs/conf.py \
+              /io/hailtop/batch
     secrets:
       - name: test-gsa-key
         namespace:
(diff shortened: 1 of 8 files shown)


hail/python/hail/__init__.py
```diff
@@ -11,6 +11,18 @@
 del pkg_resources
 del sys
 
+try:
+    is_notebook = get_ipython().__class__.__name__ == 'ZMQInteractiveShell'
```
danking, line 15: One query, one comment:

Query: how quick is `time python3 -c 'import hail as hl'` on this branch and on `main`?

Comment: we should re-use this variable instead of calling the same code in hailtop.utils.rich_progress_bar

danking (changes requested):

hail/python/hail/__init__.py
iris-garden: in `zsh` on my machine with `n = 5`, the difference between the execution time on this branch and `main` was `-0.14s`; no reason it should be faster than `main`, so i assume that's just due to natural variation

in a jupyter notebook with the same parameters the difference was `0.12s`

good catch, didn't realize we already checked whether we were in a notebook somewhere; updated! i'm not super clear on how `hailtop.utils.rich_progress_bar.is_notebook` is currently used, though. what would be a good example to try out both locally and in a notebook to make sure that it still works as expected with the change?

danking (approved):

hail/python/hail/__init__.py
```diff
@@ -11,6 +11,14 @@
 del pkg_resources
 del sys
 
+
+try:
+    from IPython.core.getipython import get_ipython
+    IS_NOTEBOOK = get_ipython().__class__.__name__ == 'ZMQInteractiveShell'
+except (NameError, ModuleNotFoundError):
+    IS_NOTEBOOK = False
```
danking, line 19: @iris-garden  Put this in hailtop b/c hailtop isn't allowed to depend on hail (circularity!)

iris-garden: moved!

PR #13612, opened 2023-09-12
[ci] Support docker build args in buildImage steps

daniel-goldstein:
Curious for your thoughts on this. I think this will make it easier to build `hailgenetics/hail` images for different python versions without duplicating Dockerfiles with the nice side-effect of not needing to jinja our Dockerfiles.

### DIFF
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -38,7 +38,7 @@ class ServiceAccount(TypedDict):
     namespace: str
 
 
-def expand_value_from(value, config):
+def expand_value_from(value, config) -> str:
     if isinstance(value, str):
         return value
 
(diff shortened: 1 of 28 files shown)


ci/ci/build.py
```diff
@@ -373,15 +390,16 @@ def build(self, batch, code, scope):
...
-     --local context={shq(context)} \
-     --local dockerfile=/home/user \
-     --output 'type=image,"name={shq(self.image)},{shq(self.cache_repository)}",push=true' \
-     --export-cache type=inline \
-     --import-cache type=registry,ref={shq(self.cache_repository)} \
-     --import-cache type=registry,ref={shq(self.main_branch_cache_repository)} \
+retry buildctl-daemonless.sh \\
+     build \\
+     --frontend dockerfile.v0 \\
+     --local context={shq(context)} \\
```
daniel-goldstein, line 396: Every line here except for the added `{build_args_str}` line is just changing `\` to `\\` so it renders nicely with line breaks in the job page.

docker/hail-ubuntu/Dockerfile
```diff
@@ -1,12 +1,17 @@
-FROM {{ global.docker_prefix }}/ubuntu:jammy-20230624
+ARG DOCKER_PREFIX={{ global.docker_prefix }}
```
daniel-goldstein, line 1: This default is for backward compatibility with the current CI that doesn't know about `ARG` and just wants to render this file template. We could presumably replace this with `ARG DOCKER_PREFIX=docker.io` so someone can build this image without needing to set up a whole registry / copy over images / set up k8s secrets to get the `global-config`

ci/test/resources/build.yaml
```diff
@@ -46,6 +46,30 @@ steps:
         to: /io/hail-ubuntu
```
danking, line 46: Can you update the other images to user build args? The `git-make-bash` can take the docker root image as an arg and that serves as a test of build args in inline images.

danking (changes requested):

danking:
love it

danking (approved):
haha, yeah, I just meant the CI stuff but this is great too.

PR #13616, opened 2023-09-13
[gear] Support both cookie and auth token authn for web and rest endpoints

daniel-goldstein:
Felt a little silly that I can't run `hailctl curl default batch /` and see some HTML output. Perhaps more importantly this would allow a more sophisticated client-side UI to call `/api` methods if it wants to perform an action that isn't a page reload. This would also allow us to proxy the backend in local development environments (though I had written this prior to our recent discussion.

Stacking on the CSRF PR.


gear/gear/auth.py
```diff
@@ -53,48 +53,41 @@ def __init__(self):
...
-                userdata = await self._userdata_from_web_request(request)
+                userdata = await self._userdata_from_request(request)
                 if not userdata:
-                    rest_userdata = await self._userdata_from_rest_request(request)
-                    if rest_userdata:
-                        raise web.HTTPUnauthorized(reason="provided REST auth to web endpoint")
-                    _web_unauthenticated(request, redirect)
+                    if redirect is None:
+                        # Only web routes should redirect by default
+                        should_redirect = '/api/' not in request.path
```
danking, line 66: it should start with `/api/`.

gear/gear/auth.py
```diff
@@ -53,48 +53,41 @@ def __init__(self):
...
+                    else:
+                        should_redirect = redirect
+
+                    if should_redirect:
+                        raise login_redirect(request)
```
danking, line 71: The collapsed if isn't so bad to read, fwiw.
```python3
if redirect or redirect is None and not request.path.startswith('/api/'):
    raise login_redirect(request)
```

danking (approved):
an elegant simplification

gear/gear/auth.py
daniel-goldstein: Not exactly the case, it could also start with `/dgoldste/batch/api/`, but I see your point on looking at prefixes.

PR #13618, opened 2023-09-13
[auth] Dont load session_id in the UserData dict

daniel-goldstein:
The `UserData` dict is loaded from the `auth` database and passed around to different services, but mostly to look at the current user's username or some other metadata. Because it's in so many places I'm a little worried about it getting logged, which we can't do because it contains the user's `session_id`. But `userdata["session_id"]` is used in so few places that I removed `session_id` from the dict and retrieve it explicitly where it's needed.


iris-garden (approved):
lgtm!

PR #13625, opened 2023-09-14
[batch] clean up timestamp parsing

danking:
While reviewing Sophie's PR, I noticed that we repeat this timestamp parsing and conversion code in several places. It goes through floating point unnecessarily. I also noticed we were using `time.time` and floating point to compute milliseconds. None of this is necessary.

I use [the Python docs suggested code](https://docs.python.org/3/library/datetime.html?highlight=timezone#determining-if-an-object-is-aware-or-naive) to verify the datetime objects are timezone aware.

A timestamp is defined as the difference in (milli)seconds from the epoch. Indeed, Python [uses that definition](https://github.com/python/cpython/blob/main/Lib/_pydatetime.py#L1941) as well.

The docs for [`total_seconds()`](https://docs.python.org/3/library/datetime.html?highlight=timezone#datetime.timedelta.total_seconds) explicitly recommend using division by a timedelta to get durations in units other than seconds.


jigold (approved):
This needs a rebase.

PR #13627, opened 2023-09-15
[compiler] a couple bugfixes in ssa pretty-printer

patrick-schultz:
Blocks containing only a single `Ref` node weren't being handled correctly, and we were adding empty lines in some cases.

Examples:

sexpr:
```
(Let __iruid_2
  (I32 2)
  (Let __iruid_1
    (I32 1)
    (Let __iruid_0
      (True)
      (If
        (Ref __iruid_0)
        (ApplyUnaryPrimOp Negate (Ref __iruid_1))
        (Ref __iruid_2)))))
```
ssa before:
```
%c2 = I32 [2]

%c1 = I32 [1]

%true = True

If %true then {
  ApplyUnaryPrimOp(%c1) [Negate]
} else {
  Ref [__iruid_2]
}
```
ssa after:
```
%c2 = I32 [2]
%c1 = I32 [1]
%true = True
If %true then {
  ApplyUnaryPrimOp(%c1) [Negate]
} else { %c2 }
```

sexpr:
```
(StreamMap __iruid_3
  (StreamRange -1 False (I32 0) (I32 5) (I32 1))
  (Ref __iruid_3))
```
ssa before:
```
!c0 = I32 [0] 
!c5 = I32 [5] 
!c1 = I32 [1]
!s = StreamRange(!c0, !c5, !c1) [-1, False]
StreamMap(!s) { (%elt) =>
  Ref [__iruid_3]
}
```
ssa after:
```
!c0 = I32 [0] 
!c5 = I32 [5] 
!c1 = I32 [1]
!s = StreamRange(!c0, !c5, !c1) [-1, False]
StreamMap(!s) { (%elt) => %elt }
```

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Pretty.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Pretty.scala
@@ -652,6 +652,9 @@ class Pretty(width: Int, ribbonWidth: Int, elideLiterals: Boolean, maxLen: Int, 
         case Ref(name, _) =>
           val body = blockBindings.lookupOption(name).getOrElse(uniqueify("%undefined_ref"))
           concat(openBlock, group(nest(2, concat(line, body, line)), "}"))
+        case RelationalRef(name, _) =>
+          val body = blockBindings.lookupOption(name).getOrElse(uniqueify("%undefined_relational_ref"))
+          concat(openBlock, group(nest(2, concat(line, body, line)), "}"))
         case _ =>
           val (pre, body) = pretty(ir, blockBindings)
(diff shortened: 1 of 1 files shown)


ehigham (commented):
Nice. Are we missing this code for other ref nodes (like `RelationalLet` and so on)?

patrick-schultz:
Good call, done.  It would be nice to factor out the common parts of handling refs, whether they're normal children or the only node in a block, but it's not clear how to do that without a pretty big structural change.

ehigham (approved):

PR #13634, opened 2023-09-15
[qob] don’t use unbounded threads reading results

### LINKED ISSUE #13533: [query] Hail uses more than 2.7 gigabytes of RAM to load ~100,000 export VCF results
### What happened?

These results should just be the filenames, so that's ~27kB of space per result. We should be able to fit ~100,000 cloud file path stings in about 100 MB.

Immediate next steps:

1. Run some smaller export VCF CDAs in QoB to assess how memory relates to partition count.
2. Inspect the CDA code and the VCFPartitionWriter code to ensure that the collected values and the result path are what we expect (specifically: an array of strings and a single file path string).
3. Add more log statements to clarify how far we get before blowing RAM.

Larger steps:

1. Set up memory allocation profiling in QoB so that we can carefully examine the memory use.
2. Assess if the Spark path also uses >27kB of RAM per partition.
3. Consider using SpillingCollectIterator to spill large results to disk.

Reported by Julia S: https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/vep.20issue
Discussed: https://hail.zulipchat.com/#narrow/stream/123011-Hail-Query-Dev/topic/vep.20issue
Job: https://batch.hail.is/batches/7917394/jobs/3 (bp: daly-neale-sczmeta)

### Version

0.2.120

### Relevant log output

_No response_

patrick-schultz:
CHANGELOG: fix a bug which caused jobs with a large number of partitions (close to 100k) to run out of memory on the driver after all partitions finish.

Cap number of threads allocated when reading results of a cda on the driver.

~Still waiting on a test to confirm this fixes the bad scaling with number of partitions.~

We were reading cda results with a threadpool containing one thread per partition, but kept all but 1000 threads blocked. This changes to use a threadpool of 1000 threads.

We still use a surprising amount of memory when reading results, but this change will likely allow jobs with ~100k partitions to complete.

fixes #13533

### DIFF
--- a/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
+++ b/hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
@@ -67,9 +67,8 @@ class ServiceBackend(
   import ServiceBackend.log
 
   private[this] var stageCount = 0
-  private[this] val executor = Executors.newCachedThreadPool()
   private[this] val MAX_AVAILABLE_GCS_CONNECTIONS = 1000
-  private[this] val availableGCSConnections = new Semaphore(MAX_AVAILABLE_GCS_CONNECTIONS, true)
+  private[this] val executor = Executors.newFixedThreadPool(MAX_AVAILABLE_GCS_CONNECTIONS)
 
   override def shouldCacheQueryInfo: Boolean = false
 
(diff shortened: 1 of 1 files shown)


danking (approved):
Approved with gusto!

PR #13639, opened 2023-09-15
[query] non-shadow-jar for pytest, doctest, and install-editable

danking:
Apparently, creating a zip file of all our dependencies and the Hail code takes ~30s. This change skips the shadow JAR for local development. We still need a shadow JAR to produce a shareable wheel file; however, when doing local development, we can simply tell Py4J (and the JVM) where to find our dependencies class files.

Also notice that I made install-editable non-PHONY. It need not be PHONY as long as we can reliably determine if `hail/python` is the pip-installed version. To do so, we simply check if the `__init__.py` at the root of the pip package is newer than when we last installed. If its newer, then either:
1. We edited `__init__.py`, or
2. The pip location of Hail has changed since we last ran install-editable

I also deleted eggs because nobody uses eggs.

### DIFF
--- a/hail/python/hail/backend/backend.py
+++ b/hail/python/hail/backend/backend.py
@@ -1,5 +1,6 @@
 from typing import Mapping, List, Union, TypeVar, Tuple, Dict, Optional, Any, AbstractSet
 import abc
+import warnings
 import orjson
 import pkg_resources
 import zipfile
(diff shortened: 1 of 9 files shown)


danking:
With this change `time make -C hail pytest PYTEST_ARGS='-k test_get_flags'` on an already compiled Hail takes 9s, most of which is initializing the JVM & Spark.

If I make a trivial change to X.scala (change a string constant), the same command takes 21s, 11s of which is compiling & creating a skinny JAR.

Reverting my edit and running the same command again takes 19s, 8s of which is compiling & creating a skinny JAR.

hail/python/hail/backend/backend.py
```diff
@@ -33,6 +34,42 @@ def fatal_error_from_java_error_triplet(short_message, expanded_message, error_i
...
+
+
+def resource_filename(f: str) -> str:
+    return pkg_resources.resource_filename(__name__, f)
+
+
+def local_jar_information() -> LocalJarInformation:
+    if resource_exists('hail.jar'):
+        warnings.warn('!!! THIS IS A DEVELOPMENT VERSION OF HAIL !!!')
+        with open(pkg_resources.resource_filename(__name__, 'extra_classpath')) as fobj:
```
daniel-goldstein, line 55:
```suggestion
        with open(resource_filename(__name__, 'extra_classpath')) as fobj:
```

hail/python/hail/backend/backend.py
```diff
@@ -33,6 +34,42 @@ def fatal_error_from_java_error_triplet(short_message, expanded_message, error_i
...
+    return pkg_resources.resource_exists(__name__, f)
+
+
+def resource_filename(f: str) -> str:
+    return pkg_resources.resource_filename(__name__, f)
+
+
+def local_jar_information() -> LocalJarInformation:
+    if resource_exists('hail.jar'):
+        warnings.warn('!!! THIS IS A DEVELOPMENT VERSION OF HAIL !!!')
```
daniel-goldstein, line 54: Is this going to warn every time we invoke hail like this?

daniel-goldstein (approved):
seems good. A couple comments but nothing to block on.

PR #13640, opened 2023-09-15
[make] Building images should not depend on k8s deployment files

daniel-goldstein:
This way if I change something in `batch.deployment.yaml` and then make deploy it only takes a couple seconds because no images need rebuilding and it just retemplates the yaml and applies it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,7 +181,7 @@ docs:
 
 website-image: docs
 
-$(SERVICES_IMAGES): %-image: $(SERVICES_IMAGE_DEPS) $(shell git ls-files $$*)
+$(SERVICES_IMAGES): %-image: $(SERVICES_IMAGE_DEPS) $(shell git ls-files $$* -- . ':!:deployment.yaml')
 	$(eval IMAGE := $(DOCKER_PREFIX)/$*:$(TOKEN))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -181,7 +181,7 @@ docs:
 
 website-image: docs
 
-$(SERVICES_IMAGES): %-image: $(SERVICES_IMAGE_DEPS) $(shell git ls-files $$*)
+$(SERVICES_IMAGES): %-image: $(SERVICES_IMAGE_DEPS) $(shell git ls-files $$* -- . ':!:deployment.yaml')
```
danking, line 184: I can't get this to work on my laptop, are you sure this works? I'd expect you to need to use grep or something to remove things.

danking (changes requested):

Makefile
daniel-goldstein: I was missing some `**/` but yes you can use git's [pathspec](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec) here to exclude (`!`) some patterns.

daniel-goldstein: This works for me now, `make deploy` auth then `touch auth/deployment.yaml` doesn't rebuild the image

danking (approved):

PR #13641, opened 2023-09-15
[fs] deduplicate gzip checking code & rename globAllStatuses

danking:
It is the wrong default to *throw away* the structured data about the files. There was only one case where we needed just the file names anyway, so I inlined the `.map(_.getPath)` there.


danking:
bump @patrick-schultz

patrick-schultz (approved):

PR #13642, opened 2023-09-15
[batch] Use normal credentials instead of internal_token for communicating between batch and batch-driver

daniel-goldstein:
In the spirit of getting everyone talking with the same authentication mechanism. The communication between batch and batch-driver isn't really a big deal, but we are basically storing a password as plaintext with this `internal_token`.

### DIFF
--- a/ci/bootstrap_create_accounts.py
+++ b/ci/bootstrap_create_accounts.py
@@ -74,6 +74,7 @@ async def main():
     users = [
         # username, login_id, is_developer, is_service_account
         ('auth', None, 0, 1),
+        ('batch', None, 0, 1),
         ('ci', None, 0, 1),
         ('test', None, 0, 0),
         ('test-dev', None, 1, 0),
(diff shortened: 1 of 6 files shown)


daniel-goldstein:
QOL side-note: It was kinda sweet that the following Just Worked

```
kssh batch dgoldste
hailctl curl dgoldste batch-driver /api/v1alpha/batches/cancel -X POST -k -vvv
```

ci/bootstrap_create_accounts.py
```diff
@@ -74,6 +74,7 @@ async def main():
     users = [
         # username, login_id, is_developer, is_service_account
         ('auth', None, 0, 1),
+        ('batch', None, 0, 1),
```
daniel-goldstein, line 77: Looks like batch was added manually in production? Feels like a mistake that it wasn't in here, and this is necessary for batch to use its hail credentials within the system.

jigold (approved):

PR #13643, opened 2023-09-18
[docker] Build hailgenetics/hail images with all supported python versions

### LINKED ISSUE #13558: [batch] hailgenetics/hail needs a version for each supported Python version
### What happened?

See, for example, a segfault Wenhan encountered when submitting a Python job that uses Hail Query: https://hail.zulipchat.com/#narrow/stream/223457-Hail-Batch-support/topic/Batch.20error/near/389248723.

I suspect this arises due to mismatch between Python versions.

### Version

0.2.120

### Relevant log output

_No response_

daniel-goldstein:
CHANGELOG: Python jobs in Hail Batch that use the default image now support all supported python versions and include the `hail` python package.

The `-py3.X` tag suffix is copied from dask's DockerHub images. Resolves #13558

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -120,6 +120,42 @@ steps:
     dependsOn:
       - merge_code
       - copy_third_party_images
+  - kind: buildImage2
+    name: hail_ubuntu_image_python_3_10
+    dockerFile: /io/hail-ubuntu/Dockerfile
+    contextPath: /io/hail-ubuntu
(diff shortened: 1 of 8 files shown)


danking (changes requested):
We need a changelog and maybe some new docs in new_python_job to clarify that the default image now supports hail

danking (approved):

PR #13650, opened 2023-09-18
[jvm-fs] simplify use of URLs

danking:
This change enforces that `String` accepting methods just convert to `URL` and do nothing more. All the logic is now in `URL` accepting methods. I think this is a bit easier to understand. It also avoids the `val url = XXXX.parseUrl(filename)` at the start of every method. I had to remove the type parameter because I would have needed a kind of existential types that I could not easily encode in Scala.

I suspect there's a slight speed benefit because anywhere that we work with URLs we can avoid re-parsing them for each method call.


danking:
@daniel-goldstein I resolved the conflict. Bump for Tuesday!

daniel-goldstein (approved):
Thanks for cleaning this up. All the URL parsing was indeed pretty ridiculous.

PR #13653, opened 2023-09-18
[batch] Remove batch worker ssh key

daniel-goldstein:
This just seems like a liability and was always awkward because we don't have a good place to keep common keys like this.

### DIFF
--- a/infra/azure/modules/batch/main.tf
+++ b/infra/azure/modules/batch/main.tf
@@ -56,16 +56,6 @@ resource "azurerm_shared_image" "batch_worker_22_04" {
   }
 }
 
-resource "kubernetes_secret" "batch_worker_ssh_public_key" {
-  metadata {
-    name = "batch-worker-ssh-public-key"
-  }
-
-  data = {
-    "ssh_rsa.pub" = file("~/.ssh/batch_worker_ssh_rsa.pub")
-  }
-}
-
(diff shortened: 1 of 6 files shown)


infra/azure/modules/batch/main.tf
```diff
@@ -56,16 +56,6 @@ resource "azurerm_shared_image" "batch_worker_22_04" {
   }
 }
 
-resource "kubernetes_secret" "batch_worker_ssh_public_key" {
-  metadata {
-    name = "batch-worker-ssh-public-key"
-  }
-
-  data = {
```
jigold, line 64: Did we have to generate this key file ourselves? Where were those instructions / commands if yes?

jigold (changes requested):

infra/azure/modules/batch/main.tf
daniel-goldstein: Good catch, I removed the instruction to generate the key.

jigold:
This all seems fine to me, but all of the test jobs failed spectacularly on Azure -- timed out. There must be another place the key is used?

jigold (approved):

danking:
WIP until I sort the release.

daniel-goldstein:
Turns out you need a key or a password. Closing this for now

PR #13659, opened 2023-09-18
[release] 0.2.123

daniel-goldstein:
I also amended the previous changelog with some PRs that I missed.


danking (approved):
Wait for https://github.com/hail-is/hail/pull/13648 to merge

danking:
#13648 is merged, just check if there are any new changelog entries and then we're good to go

danking:
```
manifest for us-docker.pkg.dev/hail-vdc/hail/hailgenetics/hail:0.2.123 not found: manifest unknown: Failed to fetch "0.2.123"
```

daniel-goldstein:
@danking That's because we weren't actually respecting the `HAIL_GENETICS_HAIL_IMAGE` for the default python image and always referencing Dockerhub. This didn't manifest until now because we were using the `python-dill` images which do not change every release

PR #13670, opened 2023-09-19
[devbin] Add a script for running a database locally

daniel-goldstein:
I picked this refactor off of my terra branch and took it the last ten yards such that now you can run `make batch-db` and get a full local instance of the batch database that you can access through `mysql -h 127.0.0.1 -u root -ppw` or

```ipython
(hail) dgoldste@wmce3-cb7 hail % HAIL_SQL_DATABASE=local-batch ipython
Python 3.9.17 (main, Jul  5 2023, 16:17:03)
Type 'copyright', 'credits' or 'license' for more information
IPython 8.15.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from gear import Database
   ...: db = Database()
   ...: await db.async_init()
   ...: await db.select_and_fetchone('SELECT * FROM globals')
   ...:
   ...:
Out[1]:
{'instance_id': 'XXXXXXXX',
 'internal_token': 'XXXXXXX',
 'n_tokens': 200,
 'frozen': 0}
```

If you add a migration, just run `make batch-db` again.

### DIFF
--- a/ci/create_local_database.py
+++ b/ci/create_local_database.py
@@ -1,0 +1,65 @@
+import asyncio
+import os
+import tempfile
+from typing import List
+
+import typer
+import yaml
+from create_database import create_migration_tables, migrate
+
+from gear import Database
+
+
+def main(service: str, database_name: str):
+    asyncio.run(async_main(service, database_name))
+
+
+async def async_main(service: str, database_name: str):
+    migrations = read_migrations_from_build_yaml(service)
+
+    db = Database()
+    await db.async_init()
+    await db.just_execute(f'CREATE DATABASE IF NOT EXISTS `{database_name}`;')
+
+    os.environ['HAIL_SQL_DATABASE'] = database_name
+    os.environ['HAIL_SCOPE'] = 'dev'
+    os.environ['HAIL_NAMESPACE'] = 'local'
+
+    if 'HAIL_CLOUD' not in os.environ:
+        os.environ['HAIL_CLOUD'] = 'gcp'
+
+    # Pick up the `HAIL_SQL_DATABASE` change
+    await db.async_close()
+    db = Database()
+    await db.async_init()
+
+    await create_migration_tables(db, database_name)
+    with tempfile.NamedTemporaryFile() as mysql_cnf:
+        mysql_cnf.write(
+            f'''
+[client]
+host = 127.0.0.1
+user = root
+password = pw
+database = {database_name}
+'''.encode()
+        )
+        mysql_cnf.flush()
+        for i, m in enumerate(migrations):
+            await migrate(database_name, db, mysql_cnf.name, i, m)
+
+
+def read_migrations_from_build_yaml(service: str) -> List[dict]:
+    with open('build.yaml', 'r', encoding='utf-8') as f:
+        build = yaml.safe_load(f)
+
+    for step in build['steps']:
+        if step['name'] == f'{service}_database':
+            return [
+                {**m, 'script': m['script'].replace('/io/', f'{service}/'), 'online': True} for m in step['migrations']
+            ]
+    raise ValueError(f'No database for service {service}')
+
+
+if __name__ == '__main__':
+    typer.run(main)
--- a/hail/python/hailtop/auth/sql_config.py
+++ b/hail/python/hailtop/auth/sql_config.py
@@ -56,6 +56,8 @@ ssl-mode={self.ssl_mode}
         assert self.port is not None
         assert self.user is not None
         assert self.password is not None
+        assert self.instance is not None
+        assert self.connection_name is not None
         if self.ssl_cert is not None:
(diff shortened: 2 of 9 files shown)


daniel-goldstein:
This can be thought of as a step toward a fully local development version of batch. I tried to keep the not-ci-tested `create_local_database.py` script as small as possible and reuse components of the tested `create_database.py`, but it's still adding yet-another-random-script into the mix.

@jigold @danking Do you think this is worthwhile on its own?

danking:
This is super cool!

daniel-goldstein:
Assigning because I want to stack my terra PR on this

daniel-goldstein:
Rebalancing my PR assigns

ci/create_local_database.py
```diff
@@ -0,0 +1,65 @@
...
+async def async_main(service: str, database_name: str):
+    migrations = read_migrations_from_build_yaml(service)
+
+    db = Database()
+    await db.async_init()
+    await db.just_execute(f'CREATE DATABASE IF NOT EXISTS `{database_name}`;')
+
+    os.environ['HAIL_SQL_DATABASE'] = database_name
+    os.environ['HAIL_SCOPE'] = 'dev'
+    os.environ['HAIL_NAMESPACE'] = 'local'
```
jigold, line 26: Can you explain why you have to modify these environment variables after creating the database instead of before?

jigold (changes requested):

hail/python/hailtop/auth/sql_config.py
```diff
@@ -56,8 +56,6 @@ def check(self):
         assert self.port is not None
         assert self.user is not None
         assert self.password is not None
-        assert self.instance is not None
```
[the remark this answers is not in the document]
daniel-goldstein: I'll change it up so we just don't `sql_config.check()` when using local mode. That way we know if `sql_config.check()` passes we are sure to be using TLS and all these are not None.

FWIW, `instance` and `connection_name` have been unused for a while, I just had a hell of a time taking them out last time I tried.

ci/create_local_database.py
```diff
@@ -0,0 +1,65 @@
+import asyncio
+import os
+import tempfile
+from typing import List
+
+import typer
+import yaml
+from create_database import create_migration_tables, migrate
```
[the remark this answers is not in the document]
daniel-goldstein: I just made the minimal changes I needed to to reuse the functionality that I wanted. You're right that's a bit of a mess. Which of the following options sounds better?
- Create a proper `create_database_utils` module to put these functions in, then pip install that in the CI utils image (I would probably want to do that in a separate PR)
- Get rid of the all the hard-coded paths in `create_database.py` and move those into the `if __name__ == '__main__'`, so basically turning the file into a utility module with a short section at the end that allows you to run it as a single file. Fewer files, but the code is still all jumbled together.

I don't have much of a preference at the moment so whatever you'd prefer to have going forward.

ci/create_local_database.py
daniel-goldstein: `'HAIL_SQL_DATABASE'` cannot be set at first because the database in the server might not exist. We then need to recreate the `Database` object to pick up that change in configuration once the database *does* exist. It's an unfortunate aspect of the `Database` API as it currently stands.

The rest of the environment variables are used by the migrations so we don't need them for the earlier lines where we're just creating the MySQL database. I can move them up, but just thought to keep all the environment variable setting together.

jigold: Where is `HAIL_SQL_DATABASE` in the code base? I couldn't find it at all.

daniel-goldstein: It's new, I added it to pick up what database to use in the `local_insecure_config` because `local_insecure_config` doesn't have a config on the filesystem.

ci/create_local_database.py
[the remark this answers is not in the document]
jigold: I think part of my confusion was there was generic functions after the database-specific code like `create_migration_tables`. If you could move all of the dual-use functions to the top of the file and then name the main function something like `create_remote_database` that would help a lot.

ci/create_local_database.py
jigold: ah, I get it now!

ci/create_local_database.py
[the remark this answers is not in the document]
daniel-goldstein: Sorry for the diff, the last commit just does the following:

1. Moves the "public" functions `migrate` and `create_migration_tables` to the top of the file
2. Then `async_main`
3. Then "private" functions that are now prefixed with an underscore.

jigold (dismissed):
This is all great. Sorry I didn't think of this before, can you either add instructions somewhere or some functionality to remove the local database (example: when you need to start on a new set of conflicting migrations)?

jigold (approved):

PR #13672, opened 2023-09-20
[auth] Reuse auth decorators without making additional requests

daniel-goldstein:
Currently the auth service uses its own authentication decorators instead of using those that the other services do (because those make a request to auth which feels a bit circular), but now we have two sets of decorators which we want to mostly keep in step aside from where they get their `UserData` from. This PR creates a small abstraction with an `Authenticator` base class and a `AuthServiceAuthenticator` subclass that extends it and fetches userdata from the auth service. The auth service instead uses a `LocalAuthenticator` which instead of making an rpc directly fetches the userdata from its database.

(I would recommend using the split diff)


auth/auth/auth.py
```diff
@@ -68,39 +68,15 @@
 routes = web.RouteTableDef()
 
 
-def authenticated_users_only(fun: AuthenticatedAIOHTTPHandler) -> AIOHTTPHandler:
-    @wraps(fun)
-    async def wrapped(request: web.Request) -> web.StreamResponse:
+class LocalAuthenticator(Authenticator):
+    async def _fetch_userdata(self, request: web.Request) -> Optional[UserData]:
         session_id = await get_session_id(request)
```
danking, line 73: During review of this I noticed that `get_session_id` and `_get_session_id` are pretty similar but have a different preference order for cookie vs header. Can we make these methods more similar? I understand that services other than auth won't look at X-Hail-Internal-Authorization, but it seems like the auth-only get_session_id should delegate to a common get_session_id if we're not using X-Hail-Internal-Authorization, right?

danking (changes requested):

auth/auth/auth.py
daniel-goldstein: Good point, I've moved `auth` to check for a `X-Hail-Internal-Authorization` header when appropriate and fall back on the `get_session_id` in gear.

danking (approved):

PR #13673, opened 2023-09-20
[hailtop] replace humanize.naturaldelta

danking:
Specifically for time, humanize does not respect significant figures. Consider this:

```
In [7]: humanize.naturaldelta(120-1)
Out[7]: 'a minute'

In [8]: humanize.naturaldelta(120)
Out[8]: '2 minutes'
```

While the difference between one minute and two minutes is minor, the difference between one hour and two hours is rather large:

```
In [6]: humanize.naturaldelta(2*60*60-1)
Out[6]: 'an hour'

In [7]: humanize.naturaldelta(2*60*60)
Out[7]: '2 hours'
```

This leads to user confusion because the cost of a two hour job seems to be substantially more than "an hour" job.


danking:
I'm also partial to `5d12h5m6s` as a format. I'm not sure how milli and micro fit into that though.

hail/python/hailtop/humanizex.py
```diff
@@ -0,0 +1,56 @@
...
+    (_MILLISECOND, 'millisecond'),
+    (_MICROSECOND, 'microsecond'),
+]
+
+
+def naturaldelta(seconds: Union[int, float]):
+    return _naturaldelta(seconds, value_unit = _SECOND, value_name = 'second')
+
+
+def naturaldelta_msec(milliseconds: Union[int, float]):
```
jigold, line 37: Can you add return types here?

jigold (changes requested):

jigold (approved):

PR #13674, opened 2023-09-20
[compiler] Fix Appending Code to Class Init

ehigham:
A returnX statement at the end of the original init block made all code appended to the constructor dead. Move this statement to the end of the constructor when emitting the class.


patrick-schultz (changes requested):
Can you throw an exception [here](https://github.com/hail-is/hail/blob/92e4776161e732d6e6b18c0889a5c2901fc2f94a/hail/src/main/scala/is/hail/expr/ir/Compile.scala#L206)? That would have caught this bug.

ehigham:
> Can you throw an exception [here](https://github.com/hail-is/hail/blob/92e4776161e732d6e6b18c0889a5c2901fc2f94a/hail/src/main/scala/is/hail/expr/ir/Compile.scala#L206)? That would have caught this bug.

I'm not quite sure what you mean? On what condition should I throw?

patrick-schultz:
Whoops, sorry clipboard malfunction. I meant [here](https://github.com/hail-is/hail/blob/e6728b812e73a8bd636046bed22f97c5b47a5f12/hail/src/main/scala/is/hail/lir/X.scala#L416), when trying to add code to an lir block after a terminating control op.

patrick-schultz (approved):

PR #13677, opened 2023-09-20
[hailtop] Apply nest_asyncio on hailtop import

daniel-goldstein:
Currently on `main`, Hail Query calls that check for cold storage use `asyncio.run_until_complete` which fails in a jupyter notebook without the application of `nest_asyncio.apply()` (because you are already running inside jupyter's event loop). This is a blanket solution that just always applies `nest_asyncio.apply()` on hailtop import to avoid these edge cases.

I don't think there is any downside to applying nest_asyncio, but something that I'm unsure about is whether I should delete the application from all these other callsites? There's now an implicit requirement that for our codebase to work as expected you must import `hailtop` before you do anything meaningful. Perhaps we should instead alter our call sites that use `run_until_complete` and apply there if we haven't already?

### DIFF
--- a/hail/python/hail/context.py
+++ b/hail/python/hail/context.py
@@ -335,8 +335,6 @@ def init(sc=None,
         backend = 'batch'
 
     if backend == 'batch':
-        import nest_asyncio
-        nest_asyncio.apply()
         import asyncio
         return asyncio.get_event_loop().run_until_complete(init_batch(
             log=log,
--- a/hail/python/hailtop/__init__.py
+++ b/hail/python/hailtop/__init__.py
@@ -1,3 +1,6 @@
+import nest_asyncio
+nest_asyncio.apply()
+
 _VERSION = None
 
 
--- a/batch/test/test_dag.py
+++ b/batch/test/test_dag.py
@@ -117,10 +117,6 @@ def test_cancel_left_after_tail(client):
 
 
 async def test_callback(client):
-    import nest_asyncio  # pylint: disable=import-outside-toplevel
-
-    nest_asyncio.apply()
-
     app = web.Application()
     callback_bodies = []
     callback_event = asyncio.Event()
--- a/hail/python/hailtop/batch/backend.py
+++ b/hail/python/hailtop/batch/backend.py
@@ -53,8 +53,6 @@ class Backend(abc.ABC, Generic[RunningBatchType]):
 
     def __init__(self):
         self._requester_pays_fses: Dict[GCSRequesterPaysConfiguration, RouterAsyncFS] = {}
-        import nest_asyncio  # pylint: disable=import-outside-toplevel
-        nest_asyncio.apply()
 
     def requester_pays_fs(self, requester_pays_config: GCSRequesterPaysConfiguration) -> RouterAsyncFS:
         try:
--- a/hail/python/hailtop/fs/router_fs.py
+++ b/hail/python/hailtop/fs/router_fs.py
@@ -1,7 +1,6 @@
 from typing import List, AsyncContextManager, BinaryIO, Optional, Tuple, Dict, Any
 import asyncio
 import io
-import nest_asyncio
 import os
 import functools
 import glob
@@ -175,7 +174,6 @@ class RouterFS(FS):
                  gcs_kwargs: Optional[Dict[str, Any]] = None,
                  azure_kwargs: Optional[Dict[str, Any]] = None,
                  s3_kwargs: Optional[Dict[str, Any]] = None):
-        nest_asyncio.apply()
         if afs and (local_kwargs or gcs_kwargs or azure_kwargs or s3_kwargs):
             raise ValueError(
                 f'If afs is specified, no other arguments may be specified: {afs=}, {local_kwargs=}, {gcs_kwargs=}, {azure_kwargs=}, {s3_kwargs=}'


danking (changes requested):
I think this won't work with uvloop. I suspect it has a minor negative impact on speed.

I think because of uvloop we probably shouldn't apply in hailtop unless we're already inside a loop, otherwise this would break our services that use uvloop.

The main place this is a problem is when we're in a notebook, right? In our own code, we should never do the async-sync-async layering. Is it sufficient to check in `hailtop/__init__.py` if we're in a loop and only apply if that's the case?

daniel-goldstein:
Good point! Didn't realize that `uvloop` didn't support nesting. I think it's sufficient to only apply if we're already in a loop. It's useless if we're not and any service that applies `uvloop` should only do that after imports which will include importing hailtop.

danking:
We do apparently layer async and sync. For example, TemporaryDirectory uses the file systems which might be async.

danking:
It seems the test failures are due to:
1. TemporaryDirectory (and TemporaryFilename)
2. `hailtop.batch.backend.ServiceBackend` absolutely should not use sync `BatchClient`, the async one is right there!
3. `hailctl batch submit` is broken because of (2)
4. `test_callback` should use async `BatchClient` b/c it is async.

TemporaryDirectory & TemporaryFilename use `hailtop.fs`, which is sync. This is nearly the async FS API except:
1. `isfile` vs `is_file`
2. `isdir` vs `is_dir`
3. `stat` returns a `FileListEntry` instead of a `FileStatus`.
4. `listfiles` vs `ls`

`hailtop.fs.router_fs.RouterFS` is a sync shim between these APIs. So there's basically sync-vs-async and Python-vs-Hail FS APIs. We have:
1. sync, Python: `hailtop.fs.FS`.
2. async, Python: does not exist.
3. async, Hail: `hailtop.aiotools.fs.FS`.
4. sync, Hail: `hailtop.fs.router_fs.RouterFS`.

If we had (2), we could write an async version of TemporaryDirectory and TemporaryFilename and use those in async methods (in particular, in `hail.backend.ServiceBackend`).

The high-level need is that we gotta be careful about not interleaving async-sync-async. Your PR reveals that we were inadvertently violating that rule. It seems best to follow the rule and only use `nest_asyncio` when we're in a Jupyter Notebook.

daniel-goldstein:
@danking I agree we shouldn't do this whole interleaving.

> If we had (2), we could write an async version of TemporaryDirectory and TemporaryFilename and use those in async methods (in particular, in hail.backend.ServiceBackend).

Do you mean that `FS` should have async variants for its methods? I could write `__aenter__` and `__aexit__` methods for `TemporaryDirectory` or `TemporaryFilename`. I think that would require restricting their `FS` field to a `RouterFS` and using that to grab the `RouterAsyncFS` to get at async implementations for the methods. Kinda gross.

danking:
Yeah, adding an `async` copy of each method of `FS` seems right to me. RouterFS is pretty close to having that anyway. `HadoopFS` can just delegate to its sync methods.

danking:
And I guess TemporaryDirectory and TemporaryFilename can have both `exit` and `aexit` and do the right thing in each. Pretty elegant actually!

hail/python/hail/context.py
```diff
@@ -687,6 +686,20 @@ def __exit__(self, type, value, traceback):
...
+        return self.name
+
+    async def __aexit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ):
+        try:
+            return await self.fs.aremove(self.name)
```
danking, line 699: I don't think you can return in `exit`; it actually means something. [Python docs](https://docs.python.org/3/reference/datamodel.html#object.__exit__):
> If an exception is supplied, and the method wishes to suppress the exception (i.e., prevent it from being propagated), it should return a true value. Otherwise, the exception will be processed normally upon exit from this method.

danking (approved):
You've a bug b/c of the return, but otherwise this looks good to me!

danking (dismissed):
Actually, please add a changelog line

hail/python/hail/context.py
daniel-goldstein: oops, thanks for the find!

danking (dismissed):
approved with gusto!

danking:
I think the tests are interfering with one another in a bad way? Notice that batch id 85 is created by test_split_multi_shuffle but then cancelled by the next test. More importantly the next test fails because apparently the loop hasn't been patched?
```
[2023-10-16 17:08:13] test/hail/methods/test_statgen.py::Tests::test_split_multi_shuffle 
-------------------------------- live log setup --------------------------------
INFO     hailtop.aiocloud.aiogoogle.credentials:credentials.py:89 using credentials file /test-gsa-key/key.json: GoogleServiceAccountCredentials for testns-test-418@hail-vdc.iam.gserviceaccount.com
Welcome to
     __  __     <>__
    / /_/ /__  __/ /
   / __  / _ `/ / /
  /_/ /_/\_,_/_/_/   version 0.2.124-3dcb6c4cac1b
LOGGING: writing to /io/repo/hail/python/hail-20231016-1708-0.2.124-3dcb6c4cac1b.log
-------------------------------- live log call ---------------------------------
INFO     hailtop.aiocloud.aiogoogle.credentials:credentials.py:89 using credentials file /test-gsa-key/key.json: GoogleServiceAccountCredentials for testns-test-418@hail-vdc.iam.gserviceaccount.com
INFO     batch_client.aioclient:aioclient.py:806 created batch 85


INFO     batch_client.aioclient:aioclient.py:818 updated batch 85


INFO     batch_client.aioclient:aioclient.py:818 updated batch 85


PASSED
[2023-10-16 17:09:01] test/hail/methods/test_statgen.py::Tests::test_split_multi_table SKIPPED
[2023-10-16 17:09:01] test/hail/methods/relatedness/test_identity_by_descent.py::test_ibd_default_arguments 
-------------------------------- live log setup --------------------------------
INFO     hailtop.aiocloud.aiogoogle.credentials:credentials.py:89 using credentials file /test-gsa-key/key.json: GoogleServiceAccountCredentials for testns-test-418@hail-vdc.iam.gserviceaccount.com
Welcome to
     __  __     <>__
    / /_/ /__  __/ /
   / __  / _ `/ / /
  /_/ /_/\_,_/_/_/   version 0.2.124-3dcb6c4cac1b
LOGGING: writing to /io/repo/hail/python/hail-20231016-1709-0.2.124-3dcb6c4cac1b.log
-------------------------------- live log call ---------------------------------
INFO     hailtop.aiocloud.aiogoogle.credentials:credentials.py:89 using credentials file /test-gsa-key/key.json: GoogleServiceAccountCredentials for testns-test-418@hail-vdc.iam.gserviceaccount.com
INFO     batch_client.aioclient:aioclient.py:806 created batch 100


INFO     batch_client.aioclient:aioclient.py:818 updated batch 100


WARNING  backend.service_backend:java.py:186 export_vcf: ignored the following fields:
    'a_index' (row)
    'was_split' (row)
INFO     batch_client.aioclient:aioclient.py:818 updated batch 100


INFO     batch_client.aioclient:aioclient.py:818 updated batch 100


INFO     batch_client.aioclient:aioclient.py:818 updated batch 100


INFO     batch_client.aioclient:aioclient.py:818 updated batch 100


INFO     batch_client.aioclient:aioclient.py:818 updated batch 100


FAILED
------------------------------ live log teardown -------------------------------
INFO     test.hail.conftest:conftest.py:77 cancelling failed test batch 85

```

danking:
@daniel-goldstein I think these changes resolve the problems in hailtop.batch. My guiding philosophy there was sync methods are always shims around async methods. https://github.com/daniel-goldstein/hail/compare/fix-nest-asyncio-apply...danking:hail:fix-nest-asyncio-apply

The issue in Query is deeper. We lazily compute the type of IRs, but to know the type of reading a Table, MatrixTable, or BlockMatrix, we must make a network request to read the metadata from that object. I'm not exactly sure what to do here.

danking:
I suppose the most correct answer is that type resolution could potentially need to do many parallel network requests (consider union'ing tens of tables). Ergo, type resolution should be async.

danking:
OK, I think this fixes the stream issue by not trying to compute the type of a `Let(..., WriteBlockMatrix)` by checking if the body could ever be a stream.

```
diff --git a/hail/python/hail/ir/ir.py b/hail/python/hail/ir/ir.py
index 85bc52aff5..76614dea0d 100644
--- a/hail/python/hail/ir/ir.py
+++ b/hail/python/hail/ir/ir.py
@@ -291,6 +291,10 @@ class Let(IR):
     def _handle_randomness(self, create_uids):
         return Let(self.name, self.value, self.body.handle_randomness(create_uids))
 
+    @property
+    def might_be_stream(self):
+        return self.body.might_be_stream
+
     @typecheck_method(value=IR, body=IR)
     def copy(self, value, body):
         return Let(self.name, value, body)
```

hail/python/hail/backend/service_backend.py
```diff
@@ -362,7 +360,7 @@ def logger(self):
 
     def stop(self):
         async_to_blocking(self._async_fs.close())
-        async_to_blocking(self.async_bc.close())
+        async_to_blocking(self._batch_client.close())
```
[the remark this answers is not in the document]
daniel-goldstein: When is that ever not null? I don't see a way to specify it from `init_batch`, can I just remove it as an argument in `create` and for sure own the client?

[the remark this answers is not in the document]
danking: Yeah that seems fine. I suspect I just did it as a matter of course.

daniel-goldstein:
@danking Do I need to set the event loop inside a pytest fixture or something?

danking:
I will look at this

hail/python/hailtop/__init__.py
```diff
@@ -1,3 +1,13 @@
+import asyncio
+
+try:
+    asyncio.get_running_loop()
+    import nest_asyncio
+    nest_asyncio.apply()
+except RuntimeError as err:
+    assert err.args[0] == "no running event loop"
+    asyncio.set_event_loop(asyncio.new_event_loop())
```
danking, line 9: I'm still chasing down exactly why this is a problem but I think this creates an event loop in a thread that is not the Main Thread. Asyncio wants there to be one single event loop in the entire program, not one per thread, so we error later when a test, running in the Main Thread, calls `asyncio.get_event_loop()`. That function refuses to create a new loop if `asyncio.set_event_loop` was already called.

High level: why create the event loop here? If its not running, just do nothing, yeah?

danking: OK, I think I've made sense of this.

You *really* don't want to have two event loops. `asyncio.run` creates an event loop, runs the coro, then destroys the event loop. `aiohttp` (and perhaps other libraries) stash a reference to the event loop. They cannot be used across multiple calls to `asyncio.run`.

For this reason, asyncio works to prevent you from creating two event loops. Pytest appears to use multiple threads to `import` all the test modules. This change forces initialization of an event loop in some non-Main-Thread. That causes `self._local._set_called` to be set to `True` on the `_UnixDefaultEventLoopPolicy`. That prevents `get_event_loop` from creating new threads.

So we're stuck: the event loop doesn't exist in the main thread and we're not allowed to create a new one.

danking:
I think this PR still has good changes. We should avoid nesting when possible; I fear it leads to confusing situations. I'm gonna take it off the release 0.2.125 checklist though

hail/python/test/hail/matrixtable/test_file_formats.py
```diff
@@ -45,35 +47,26 @@ def all_values_table_fixture(init_hail):
...
+            try:
+                with os.scandir(Path(resource_dir, version_dir, dirname)) as old_files:
+                    for file in old_files:
+                        file_paths.append(file.path)
+            except FileNotFoundError:
+                pass
+    return file_paths
+
+ht_paths = add_paths('table')
+mt_paths = add_paths('matrix_table')
```
daniel-goldstein, line 64: Using async here was quite problematic, I *think* because `pytest` uses non-main threads to run some imports so this was creating an event loop outside of the main thread. Since we don't run these tests in QoB, I just used normal synchronous python FS primitives.

daniel-goldstein:
@danking This is passing now. This is now primarily a PR that avoids async / sync / async.

hail/python/test/hail/matrixtable/test_file_formats.py
danking: I agree with your diagnosis. Solution seems fine to me.

danking (approved):

PR #13678, opened 2023-09-20
[batch] Account for JVM containers when allocating network namespaces

### LINKED ISSUE #13402: [batch] Requesting a network namespace can sometimes hang indefinitely
### What happened?

On startup, a Batch Worker pre-allocates network namespaces equal to the maximum number of slots on the worker (64 for a 16 core worker). When a job finishes, its network namespace is deleted and a task is created to replenish the namespace. This way, jobs seldom need to wait on the creation of a network namespace and there *should* always be one available or creating. However, sometimes the creation of a new network namespace is disrupted, causing new jobs to hang indefinitely waiting on one.

### Version

0.2.120

### Relevant log output

_No response_

daniel-goldstein:
Creating network namespaces can often take hundreds of milliseconds (and sometimes seconds with `iptables` contention), so Batch takes this off the job hot path by pre-allocating namespaces. All job namespaces are configured identically and there is a fixed number of "slots" on any batch worker (`CORES * 4`), so pre-allocation and asynchronous recycling of namespaces is fairly straight-forward so long as we never attempt to run more containers on a worker than the number of slots (which the scheduling system should prohibit).

However, since we started running long-lived JVM containers (#11397), the number of containers running on a given worker can easily be *greater* than `N_SLOTS`. On a 16-core machine, we create 30 JVMs that sit idle waiting for JVMJobs all the while occupying a precious network namespace.

I thought for the longest time that #13402 was a race condition so was trying to trigger it through a barrage of quick jobs. Turns out all it took was running >34 long-running jobs on a single 16-core worker. In a dev deploy of `main`, running a batch with 35 quarter-core `sleep 150` jobs fails with a single job timing out waiting for a network. On this branch, I am able to run the same 35 job batch as well as a batch with 64 quarter-core jobs.

Unfortunately, we don't have a great way to test "run all these jobs at once on the same worker".

Resolves #13402

### DIFF
--- a/batch/batch/worker/worker.py
+++ b/batch/batch/worker/worker.py
@@ -203,6 +203,8 @@ instance_config: Optional[InstanceConfig] = None
 
 N_SLOTS = 4 * CORES  # Jobs are allowed at minimum a quarter core
 
+N_JVM_CONTAINERS = sum(1 for jvm_cores in (1, 2, 4, 8) for _ in range(CORES // jvm_cores))
+
 deploy_config = get_deploy_config()
 
 docker: Optional[aiodocker.Docker] = None
@@ -345,11 +347,12 @@ class NetworkAllocator:
         self.internet_interface = INTERNET_INTERFACE
 
     async def reserve(self):
-        for subnet_index in range(N_SLOTS):
+        for subnet_index in range(N_SLOTS + N_JVM_CONTAINERS):
             public = NetworkNamespace(subnet_index, private=False, internet_interface=self.internet_interface)
             await public.init()
             self.public_networks.put_nowait(public)
 
+        for subnet_index in range(N_SLOTS):
             private = NetworkNamespace(subnet_index, private=True, internet_interface=self.internet_interface)
 
             await private.init()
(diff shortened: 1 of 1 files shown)


batch/batch/worker/worker.py
```diff
@@ -345,11 +347,12 @@ def __init__(self, task_manager: aiotools.BackgroundTaskManager):
         self.internet_interface = INTERNET_INTERFACE
 
     async def reserve(self):
-        for subnet_index in range(N_SLOTS):
+        for subnet_index in range(N_SLOTS + N_JVM_CONTAINERS):
             public = NetworkNamespace(subnet_index, private=False, internet_interface=self.internet_interface)
             await public.init()
             self.public_networks.put_nowait(public)
 
+        for subnet_index in range(N_SLOTS):
```
danking, line 355: b/c JVMJobs can't be on the private network?

danking (changes requested):

batch/batch/worker/worker.py
daniel-goldstein: Yes I don't see a reason for JVMJobs to use the private network. Private basically just means you can talk to the metadata server.

danking (approved):

PR #13680, opened 2023-09-20
[batch] Support container images with nonexistent users

daniel-goldstein:


danking:
I think our service backend debug stuff should search for “bad” not “failed” jobs. Looks like the worker is having trouble making jobs

daniel-goldstein:
We only ever got one image with this problem and the user had a workaround. Low enough priority that we're closing and leaving this PR for future reference if it becomes an issue in the future.

PR #13681, opened 2023-09-21
[hailctl] Fixes for first time init

daniel-goldstein:
Two things:
- I broke `hailctl batch init` after my changes to auth in the case where you are starting from no config (`~/.config/hail` does not exist)
- The service account needs delete permissions on the temp bucket to run `remove_tmpdir` jobs.

Cutting a release here so we can use this in today's workshop.


daniel-goldstein:
The change regarding the copy tool progress bar is because on non-verbose mode (like in the local backend) it was generating some weird whitespace by just creating the progress bar not disabled even though none of the tasks were visible. In jupyter notebooks, whitespace rendered as `5l` in the jupyter output.

PR #13682, opened 2023-09-21
[query] Better error messages for unsupported types in export_vcf

chrisvittal:


hail/src/main/scala/is/hail/io/vcf/ExportVCF.scala
```diff
@@ -82,13 +111,19 @@ object ExportVCF {
...
-      if (!valid)
-        fatal(s"Invalid type for format field '${ fd.name }'. Found '${ fd.typ }'.")
+      if (valid) {
+        None
+      } else {
+        Some(s"Invalid type for format field '${ fd.name }'. Found '${ fd.typ }'.")
+      }
+    }
+    if (!invalid.isEmpty) {
+      fatal(invalid.mkString("\n"))
```
danking, line 126: nice improvement!

danking (dismissed):
This looks great but let's add a couple python tests asserting the error message is better.

danking (approved):

PR #13691, opened 2023-09-22
[query] revert install-editable speedup

danking:


danking:
bump @iris-garden let's merge this and I'll revisit at a later date.

danking:
We fixed all the known issues and install-editable now seems reliable and fast.

PR #13694, opened 2023-09-22
[query] use parameterization for test_blanczos_flags

danking:


patrick-schultz (approved):
We should probably pull out the common setup into fixtures at some point, but that shouldn't be too significant timewise.

PR #13709, opened 2023-09-26
[website] Make better use of make when building docs locally for the website

daniel-goldstein:
So you don't need to rebuild the docs if you only touched code in `website`.

### DIFF
--- a/website/Makefile
+++ b/website/Makefile
@@ -2,22 +2,19 @@ include ../config.mk
 
 .PHONY: build run run-docker deploy clean
 
-WEBSITE_IMAGE := $(DOCKER_PREFIX)/website:$(TOKEN)
-
 build:
 	$(MAKE) -C .. website-image
 
 run:
-	$(MAKE) -C .. docs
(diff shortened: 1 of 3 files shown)


website/Makefile
```diff
@@ -2,22 +2,19 @@ include ../config.mk
...
-WEBSITE_IMAGE := $(DOCKER_PREFIX)/website:$(TOKEN)
-
 build:
 	$(MAKE) -C .. website-image
 
 run:
-	$(MAKE) -C .. docs
-	cd website && tar -xvzf ../../docs.tar.gz
-	HAIL_DOMAIN=localhost:5000 python3 -m website local
+	$(MAKE) -C .. docs.tar.gz
```
danking, line 9: you remove the extraction of the docs tar; how does this work now?

danking (changes requested):

danking (approved):
Another transient `test_gpu_accesibility_g2` error. I retried. LGTM.

PR #13711, opened 2023-09-26
[services] Use AsyncExitStack on app cleanup

daniel-goldstein:


danking (approved):
I love it. Thoughts on pushing it back to the allocation points so its more obvious that we've remembered to close everything we open?

daniel-goldstein:
Like creating the exit stack in `on_startup` and using it `on_cleanup`? I think that makes sense, though a little awkward because we'll have to invert the closing order.

Issue #13554, opened 2023-09-05
[batch] instances leaked by PR 13509 and 13458


danking:
### What happened?

This namespace leaked at least two instances: pr-13509-default-p2aogbaogrsp. This namespace appears to have leaked at least one instance: pr-13458-default-eac7piapxept.

```
batch-worker-pr-13509-default-p2aogbaogrsp-highmem-np-zx6w4   us-central1-b  n1-highmem-16                10.128.0.52   34.68.68.147     RUNNING
batch-worker-pr-13509-default-p2aogbaogrsp-standard-np-3e6i5  us-central1-c  n1-standard-16               10.128.0.202  34.132.64.121    RUNNING
batch-worker-pr-13458-default-eac7piapxept-standard-bb67h     us-east1-d     n1-standard-16  true         10.142.0.51   34.75.118.53     RUNNING
```

### Version

0.2.120

jigold:
I'm still looking, but I could only find the logs for PR 13509 as PR 13458 is too old. There are no batch worker logs at all for these two instances, but there are a bunch of sys logs. I didn't see an obvious error message, but there's 1000s of sys log messages in there.

https://console.cloud.google.com/logs/query;query=resource.type%3D%22gce_instance%22%0Alabels.%22compute.googleapis.com%2Fresource_name%22:%22batch-worker-pr-13509-default-p2aogbaogrsp-highmem-np-zx6w4%22;summaryFields=:false:32:beginning;cursorTimestamp=2023-08-29T20:39:28Z;aroundTime=2023-08-29T20:16:33.950Z;duration=PT24H?project=hail-vdc

jigold:
It looks like permissions for deleting disks and VMs are broken for the `delete_batch_instances` CI step. This job also hung for a long time and then got restarted. There's some other wonky things about this PR, but it just seems like the main issue was the Batch deployment was cancelled mid-run and the driver didn't have time to cleanup those 2 VMs that weren't responding before being shut off. Then the cleanup step isn't actually working so they didn't get cleaned up. The only remaining question I have is why these VMs weren't starting up correctly. There were at least 5 in this one PR that didn't start up in time before the driver was shut down.

https://batch.hail.is/batches/7908998/jobs/207

jigold:
@daniel-goldstein Does this command in `gcloud -q auth activate-service-account --key-file=/test-gsa-key/key.json` in a `delete_gcp_batch_instances` CI job still work after your auth changes?

[...]

PR #13719, opened 2023-09-27
[build.yaml] Give delete_gcp_batch_instances job the proper service account

jigold:
See #13554. This is the first fix that needs to happen such that the service account has permissions to delete VMs and disks. Note that the correct service account was already being used in Azure.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -3674,7 +3674,7 @@ steps:
     alwaysRun: true
     script: |
       set -ex
-      gcloud -q auth activate-service-account --key-file=/test-gsa-key/key.json
+      gcloud -q auth activate-service-account --key-file=/batch-gsa-key/key.json
       set +e
       kubectl -n {{ default_ns.name }} scale deployment batch-driver --replicas=0
       gcloud -q compute instances list \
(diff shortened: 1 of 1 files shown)


daniel-goldstein (approved):

jigold:
There's still an issue here with deleting disks, but I'll debug separately. Let's get the VM deletion fix in.

```
+ gcloud -q compute disks list --filter 'labels.batch=1 AND labels.namespace=pr-13719-default-fb8ih36hcona' '--format=table[no-heading](zone.basename(), name)' --project hail-vdc
+ xargs -P8 -n2 -r sh -c 'gcloud -q compute disks delete --zone "$1" --project hail-vdc "$2" || true' argv0
WARNING: The following filter keys were not present in any resource : labels.batch, labels.namespace
```

jigold:
Hmmm... we do give these labels to disks. Maybe that's the error when no orphaned disks are found with any labels

daniel-goldstein:
> Maybe that's the error when no orphaned disks are found with any labels

Ya I interpret this as the query just returned no results.

PR #13725, opened 2023-09-27
[batch] Fix missing exact match and partial match search symbols

jigold:


batch/batch/front_end/query/operators.py
```diff
@@ -19,11 +19,11 @@ class MatchOperator(Operator, abc.ABC):
 
 
 class PartialMatchOperator(MatchOperator, abc.ABC):
-    pass
+    symbols: ClassVar[Set[str]] = {'!~', '=~'}
 
 
 class ExactMatchOperator(MatchOperator, abc.ABC):
-    pass
+    symbols: ClassVar[Set[str]] = {'=', '!='}
```
daniel-goldstein, line 26: I've forgotten some details here. How come there's overlap between the symbols in `ComparisonOperator` and `MatchOperator`? Is that allowed or problematic?

There also appears to be a redundancy that's not clear to me between `==` and `=`.

daniel-goldstein (changes requested):
A couple questions

batch/batch/front_end/query/operators.py
jigold: That's allowed. One is supposed to be for numeric values (Comparison) and the other is for string matches. I put the `==` and `=` because it always drove me nuts for other tools trying to remember if something should be `=` or `==`. Happy to reconsider if you think Comparison should always be `==` and Match should be `=` or some combination of that.

daniel-goldstein: Ah, great, thanks for the explanation. I'm happy to accept either `==` or `=` to be a little more user-friendly (I also often get it wrong). Could I ask for a follow-up that just renames `ComparisonOperator` to `NumericComparisonOperator`? I think that's all that's needed to clear this up in the future.

daniel-goldstein (approved):

PR #13727, opened 2023-09-27
[batch] Only write logs and resource usage for started containers

jigold:
I'm not sure this is the only reason why we're getting worker log errors when a user deletes jobs, but this code is definitely wrong in the case a container hasn't been started. I'm conflicted on whether we should do nothing or write empty files though.


daniel-goldstein (approved):
I think this is good behavior. An empty file could be mistaken for a job that wrote nothing to stdout, and the front-end expects such a scenario where it can't find a log file and alerts the user that it wasn't found.

PR #13728, opened 2023-09-27
[batch] Update gcsfuse version to 1.2.0

jigold:
Let's see if this "just works".

### DIFF
--- a/batch/Dockerfile.worker
+++ b/batch/Dockerfile.worker
@@ -16,10 +16,10 @@ RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.9 1
 {% if global.cloud == "gcp" %}
 RUN echo "APT::Acquire::Retries \"5\";" > /etc/apt/apt.conf.d/80-retries && \
     hail-apt-get-install curl gnupg && \
-    export GCSFUSE_REPO=gcsfuse-`lsb_release -c -s` && \
+    export GCSFUSE_REPO=gcsfuse-jammy && \
     echo "deb http://packages.cloud.google.com/apt $GCSFUSE_REPO main" | tee /etc/apt/sources.list.d/gcsfuse.list && \
     curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add - && \
-    hail-apt-get-install fuse gcsfuse=1.2.0
+    hail-apt-get-install fuse gcsfuse=1.1.0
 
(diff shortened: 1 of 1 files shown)


daniel-goldstein (approved):

danking:
We don’t have lab_release in our Ubuntu image. We could add it to hail Ubuntu or just in-line the name. We had bionic here but that’s 18.04 we really should use jammy.

danking:
Also looks like we might need to go to 1.1.0 b/c that’s all that’s been released to Apt https://packages.cloud.google.com/apt/dists/gcsfuse-jammy/main/binary-amd64/Packages

jmarshall:
Looking at the `…/Packages` URL in the previous comment, 1.2.0 is now available (and 1.1.0 does not appear to be there).

In our recent local hail update deployment, this batch-worker-image-creation job failed several times due to GoogleCloudPlatform/gcsfuse#1424. We worked around this as initially suggested on that issue with populationgenomics/hail@607408bee752dabca48d9a2732b14d32813ace9f, but later comments on the issue suggest that the better approach would be this PR with an additional change to access the apt repo via https:

```diff
-    echo "deb http://packages.cloud.google.com/apt $GCSFUSE_REPO main" | tee /etc/apt/sources.list.d/gcsfuse.list && \
+    echo "deb https://packages.cloud.google.com/apt $GCSFUSE_REPO main" | tee /etc/apt/sources.list.d/gcsfuse.list && \
```

danking:
@jigold can I remove the WIP tag? I'd like this to make it into the 0.2.125 release so that AUS can check out a tagged release.

jigold:
Yes with the caveat that I was going to look at the worker logs in the PR test namespace just to make sure there were no hidden errors that would be problematic. I broke the logging query generator -- fixed in #13813 -- and didn't think this was urgent. I'll take a look at the logs now without that fix in.

jigold:
I forgot I turned off the syslog so we won't see errors there from while the worker is running.

Up to you on whether you think this change needs more scrutiny.

jmarshall:
It's not so much so that **we** can check out a tagged release, as we have already worked around the problem. But I would expect that you and any other installations will also run into the same `batch_worker_image` failure.

We have been running our production instance using gcsfuse 1.2.0 for about a week now, and I think @illusional will agree with me that we haven't seen any problems from it.

PR #13738, opened 2023-09-27
[query] Deprecate `collect_all` in `hail.plots.*` (use `n_divisions` instead)

### LINKED ISSUE #13696: [query] hl.plot.manhattan uses `collect_all=True` to mean "do not downsample
### What happened?

We should use a more obvious name like `downsample=False`.

### Version

0.2.124

### Relevant log output

_No response_

ehigham:
Fixes #13696.
`n_divisions` is a downsampling factor. It conflicts with `collect_all` when `collect_all=True` is specified.
Deprecate `collect_all` and issue warning when users supply a non-`None` `collect_all` argument.
Use `n_divisions=None` to not downsample (implies `collect_all=True`).
Throw when `collect_all=True` and `n_divisions` is not `None` - this is a conflict.

### DIFF
--- a/hail/python/hail/plot/plots.py
+++ b/hail/python/hail/plot/plots.py
@@ -6,6 +6,7 @@ import pandas as pd
 import bokeh
 import bokeh.io
 import bokeh.models
+import warnings
 from bokeh.models import (HoverTool, ColorBar, LogTicker, LogColorMapper, LinearColorMapper,
                           CategoricalColorMapper, ColumnDataSource, BasicTicker, Plot, CDSView,
                           GroupFilter, IntersectionFilter, Legend, LegendItem, Renderer, CustomJS,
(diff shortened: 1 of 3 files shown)


hail/python/hail/plot/plots.py
```diff
@@ -1491,11 +1498,15 @@ def manhattan(pvals, locus=None, title=None, size=4, hover_fields=None, collect_
 
     pvals = -hail.log10(pvals)
 
+    if collect_all:
+        warnings.warn('manhattan: `collect_all` has been deprecated. Use `downsample` instead.')
+        downsample = False
```
danking, line 1503: If a user specifies `manhattan(collect_all=False, downsample=False)` they should get an error. As written, I think that would not downsample (violating the `collect_all` setting). I think your best bet is to change the default argument for `collect_all` to `None` and then:
1. Verify only one of the two is not `None`.
2. Warn if `collect_all` is not `None`.

I know this adds a fair bit of code bloat, but it seems worthwhile if it saves users from getting confused by doubly-specified parameters.

danking (changes requested):
Awesome! thanks for taking this on

hail/python/hail/plot/plots.py
ehigham: I had considered doing that, but questioned the benefit. I'll add that in now

ehigham: Seems to me we already have double specificity by the existance of `n_divisions`.

hail/python/hail/plot/plots.py
```diff
@@ -1450,27 +1451,44 @@ def qq(
...
+           significance_line=nullable(numeric)
+           )
+def manhattan(pvals: 'Float64Expression',
+              locus: 'Optional[LocusExpression]' = None,
+              title: 'Optional[str]' = None,
+              size: int = 4,
+              hover_fields: 'Optional[Dict[str, Expression]]' = None,
+              collect_all: 'Optional[bool]' = None,
+              n_divisions: 'Optional[int]' = 500,
+              significance_line: 'Optional[Union[int, float]]' = 5e-8
```
danking, line 1470: These types don't need to be in strings unless there's a typing circularity, which, in this case, there isn't. Doesn't affect correctness or anything, just a style thing.

danking (approved):
Awesome, this significantly improves the usability of this method.

ehigham:
Made this change to all plots in `hail.plots.*` to be consistent. Fixed docs.

danking:
huzzah!

PR #13741, opened 2023-09-28
[batch] Ignore client_job attribute in test_callback test

daniel-goldstein:
I think the `client_job` attribute is doing its intended job but we need to ignore uncontrollable attributes in this test in particular.

### DIFF
--- a/batch/test/test_dag.py
+++ b/batch/test/test_dag.py
@@ -162,6 +162,7 @@ async def test_callback(client):
         callback_body.pop('duration')
         callback_body.pop('duration_ms')
         callback_body.pop('cost_breakdown')
+        callback_body['attributes'].pop('client_job')
         assert callback_body == {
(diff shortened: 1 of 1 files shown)


jigold (approved):

batch/test/test_dag.py
```diff
@@ -162,6 +162,7 @@ async def callback(request):
         callback_body.pop('duration')
         callback_body.pop('duration_ms')
         callback_body.pop('cost_breakdown')
+        callback_body['attributes'].pop('client_job')
```
danking, line 165: there's also `'name'`

danking (changes requested):

batch/test/test_dag.py
daniel-goldstein: @danking name is controlled for and tested down below though

PR #13743, opened 2023-09-28
[humanizex] Run humanize tests and fix microseconds bug

daniel-goldstein:
This is two independent commits but I put them in one PR, happy to split if you want.

1. These tests weren't being run in CI I think because of the filename not starting with `test_*`. The microseconds test currently fails on main. I believe this is because the fallback case at the end uses the units of the input and not the smallest possible units (on main you get naturaldelta(0.000001) == 0 seconds not 1 microsecond)

2. Changed second(s) -> s, millisecond(s) -> ms, microsecond(s) -> µs


hail/python/test/hailtop/test_humanizex.py
```diff
@@ -0,0 +1,43 @@
+from hailtop.humanizex import naturaldelta, naturaldelta_msec
```
daniel-goldstein, line 1: hm I swear I `git mv`'d this file…

hail/python/hailtop/humanizex.py
```diff
@@ -13,36 +13,39 @@
...
     (_WEEK, 'week'),
     (_DAY, 'day'),
     (_HOUR, 'hour'),
     (_MINUTE, 'minute'),
-    (_SECOND, 'second'),
-    (_MILLISECOND, 'millisecond'),
-    (_MICROSECOND, 'microsecond'),
+    (_SECOND, 's'),
+    (_MILLISECOND, 'ms'),
+    (_MICROSECOND, 'μs'),
```
danking, line 32: this won't display correctly on some browsers unless we set `<meta charset="utf-8">`. It also must appear in the first 1024 bytes. [details](https://www.w3.org/International/questions/qa-html-encoding-declarations). It appears to be set for the website, but we need to set it for web_common. I think that's it though?

danking (changes requested):
x

hail/python/test/hailtop/test_humanizex.py
danking: I'm pretty sure git doesn't store moves, it just deduces them from the diff. If you changed it enough, you fool the logic.

hail/python/hailtop/humanizex.py
daniel-goldstein: Interesting, didn't know you could be non-utf-8 these days.

danking (approved):

danking:
Yeah, when there's no encoding set, Safari defaults to IOS Latin 1

Screenshot 2023-09-28 at 10 14 46

Mu is two bytes in UTF-8, so you get two characters

Screenshot 2023-09-28 at 10 17 09


It happens to me fairly often (like a couple times a year), usually on personal websites. Chrome and Firefox must default to UTF-8 or something because I seem to be in the minority of noticing these things.

daniel-goldstein:
Ya my chromium-based browser shows a µ with no charset. Interesting!

PR #13744, opened 2023-09-28
[ci] Generate PR specific logging queries

### LINKED ISSUE #13556: [ci] CI should link to the various logs (worker, batch, driver, etc.) from the CI page
### What happened?

x

### Version

0.2.120

### Relevant log output

_No response_

jigold:
Fixes #13556. I haven't tested these changes -- would like to get initial feedback first.

### DIFF
--- a/ci/ci/ci.py
+++ b/ci/ci/ci.py
@@ -190,6 +190,8 @@ async def get_pr(request: web.Request, userdata: UserData) -> web.Response:
             if CLOUD == 'gcp':
                 start_time = status['time_created']
                 end_time = status['time_completed']
+                assert start_time is not None
+                assert PROJECT is not None
 
                 page_context['logging_queries'] = {
                     'batch-k8s-error-warning': generate_gcp_service_logging_url(
--- a/ci/ci/ci.py
+++ b/ci/ci/ci.py
@@ -191,7 +191,7 @@ async def get_pr(request: web.Request, userdata: UserData) -> web.Response:
                 start_time = status['time_created']
                 end_time = status['time_completed']
                 assert start_time is not None
-                page_context['logging_queries'] = gcp_logging_queries(DEFAULT_NAMESPACE, start_time, end_time)
+                page_context['logging_queries'] = gcp_logging_queries(start_time, end_time)
             else:
                 page_context['logging_queries'] = None
         else:
--- a/ci/ci/ci.py
+++ b/ci/ci/ci.py
@@ -40,10 +40,10 @@ from hailtop.utils import collect_aiter, humanize_timedelta_msecs, periodically_
 from web_common import render_template, set_message, setup_aiohttp_jinja2, setup_common_static_routes
 
 from .constants import AUTHORIZED_USERS, TEAMS
-from .environment import CLOUD, DEFAULT_NAMESPACE, PROJECT, STORAGE_URI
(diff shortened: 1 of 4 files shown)


daniel-goldstein:
What kind of feedback would you like? At a high-level, I think constructing these URLs and embedding links like this into the PR page is a great idea. Love how little code it takes to do this.

jigold:
This confirmation is good enough. Thanks!

daniel-goldstein (changes requested):
Requesting changes so this leaves my CI queue. Feel free to dismiss when you want a review

ci/ci/ci.py
```diff
@@ -185,6 +186,34 @@ async def get_pr(request: web.Request, userdata: UserData) -> web.Response:
...
+                    ),
+                    'batch-workers-error-warning': generate_gcp_worker_logging_url(
+                        PROJECT, DEFAULT_NAMESPACE, start_time, end_time, ['ERROR', 'WARNING']
+                    ),
+                    'ci-k8s-error-warning': generate_gcp_service_logging_url(
+                        PROJECT, ['ci'], DEFAULT_NAMESPACE, start_time, end_time, ['ERROR', 'WARNING']
+                    ),
+                    'auth-k8s-error-warning': generate_gcp_service_logging_url(
+                        PROJECT, ['auth', 'auth-driver'], DEFAULT_NAMESPACE, start_time, end_time, ['ERROR', 'WARNING']
+                    ),
```
daniel-goldstein, line 213: I find this a bit noisy for this function, can we package these lines in a function that's also in the utils file and then change this to just be `page_context['logging_queries'] = gcp_log_urls(DEFAULT_NAMESPACE)` (or something to that effect)?

ci/ci/environment.py
```diff
@@ -21,6 +21,8 @@
 
 if CLOUD == 'gcp':
     REGION = get_gcp_config().region
+    PROJECT = get_gcp_config().project
 else:
     assert CLOUD == 'azure', CLOUD
     REGION = get_azure_config().region
+    PROJECT = None
```
daniel-goldstein, line 28: nit: Since project can never be not-None in azure, I would rather the variable just not exist in that environment at all. Since `PROJECT` is basically never going to take on multiple values (and therefore is not super useful as a function parameter), can you instead access the project config variable inside the GCP-specific function in utils that uses it? Then you also don't need these asserts that it's not None.

ci/ci/utils.py
```diff
@@ -36,3 +37,62 @@ async def add_deployed_services(
...
+    if severity is not None:
+        severity_queries = []
+        for level in severity:
+            severity_queries.append(f'severity={level}')
+        query += ' OR '.join(severity_queries)
+
+    timestamp_query = f';startTime={start_time}'
+    if end_time is not None:
+        timestamp_query += f';endTime={end_time}'
+
```
daniel-goldstein, line 73: Nit: the severity, timestamp and URL construction I believe are the same between these two functions. Can you separate those out into a helper?

daniel-goldstein (dismissed):
This is awesome! Just have a couple organizational nits. This is a big QOL improvement

daniel-goldstein (dismissed):
Oops accidentally hit approve instead of request changes. I just had a couple small organizational comments.

daniel-goldstein (approved):

PR #13745, opened 2023-09-28
[linting] Catch unintended errors in check-sql.sh

daniel-goldstein:
I really borked our SQL linting. This PR is short but it catches a few critical problems.

1. The point of `check-sql.sh` is to detect modifications or deletions of SQL files in PRs and fail if such a change occurs. Currently on `main` it does not detect modifications. In #13456, I removed the `delete-<service>-tables.sql` files (intentionally), so added the `^D` to the `grep` regex to indicate that it is OK to have a deletion. What I inadvertently did though is change the rule from "It's ok to have Additions of any file OR  Modifications of estimated-current.sql / delete-<service>-tables.sql" to "It's ok to have Additions OR Modifications OR Deletions of estimated-current.sql / delete-<service>-tables.sql". Really this should have been "It's ok to have Additions OR Modifications of estimated-current.sql OR Deletions of delete-<service>-tables.sql". I've changed it to reflect that rule.

2. Rules currently silently *pass* in CI with an error message that git is not installed. In #13437 I changed the image used to run the linters and inadvertently didn't include `git` which `check-sql.sh` needs to run. Here's how it failed in a sneaky way:
- Since `git` is not installed, all calls to `git` fail, but the script is not run with `set -e` so every line of the script is executed
- Since `git` lines fail, `modified_sql_file_list` remains empty
- Since `modified_sql_file_list` remains empty, it appears to the check at the end that everything checked out
- The if statement runs successfully and the script returns with error code 0

To fix this I do a few things:
- installed `git` in the linting image
- `set -e` by default and only enable `set +e` later on when necessary (because we don't want a failed `git diff` to immediately exit)
- Do away with the file checking and instead check the error code of the grep. If nothing survives the grep filter, which means there were no illegal changes made, grep will return with exit code 1. So we treat that exit code as a success.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,10 +55,10 @@ pylint-%:
 
 .PHONY: check-%-fast
 check-%-fast:
-	# ruff check $*
-	# $(PYTHON) -m pyright $*
-	# $(PYTHON) -m black $* --check --diff
-	# curlylint $*
+	ruff check $*
+	$(PYTHON) -m pyright $*
+	$(PYTHON) -m black $* --check --diff
+	curlylint $*
 	cd $* && bash ../check-sql.sh
 
 .PHONY: check-%
--- a/Makefile
+++ b/Makefile
@@ -55,10 +55,10 @@ pylint-%:
 
 .PHONY: check-%-fast
 check-%-fast:
-	ruff check $*
-	$(PYTHON) -m pyright $*
-	$(PYTHON) -m black $* --check --diff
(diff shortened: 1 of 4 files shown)


daniel-goldstein:
Unassigning until I fix errors in CI, apologies for the noise

daniel-goldstein:
Ok this looks to be working now

jigold (approved):

PR #13747, opened 2023-09-28
[pytest] Stop and reinitialize hail between QoB tests

daniel-goldstein:
Stopping and re-initing is cheap and easy with QoB (compared to spark) and is a cleaner way to reset service backend state. Also note that we pass in the test name as `app_name` in init so batches are guaranteed to be created with the name of the test.


danking (approved):
Ah this is much nicer

PR #13749, opened 2023-09-28
[vds] for filter_intervals, do not serialize the intervals to Python

### LINKED ISSUE #13748: [qob] Since 0.2.117 vds.filter_intervals unnecessarily uses a lot of RAM
### What happened?

Consider this Python code

```python3
# ...
bed_intervals = hl.import_bed(bed_url, reference_genome='GRCh38')
callset = hl.vds.filter_intervals(vds1, bed_intervals, split_reference_blocks=True)
# ...
```

`filter_intervals` delegates to `_parameterized_filter_intervals`. In that method, the intervals are collected into a local array:

```python3
intervals = hl.literal(intervals.aggregate(hl.agg.collect(intervals.key[0]), _localize=False))
```

Unfortunately, `hl.literal` will evaluate a Hail expression in order to produce a `Literal` IR node. As a result, this serializes a potentially large array of intervals from Scala to Python. This is usually fine, but when this array is serialized back to Java as a `Literal`, it is serialized as JSON. Our JSON representation of an Interval of Locus is particularly verbose:

```ipython3
In [5]: import hail as hl
   ...:
   ...: hl.literal(hl.Interval(start=hl.Locus(contig='chr1', position=10001, reference_genome='GRCh38'), end=hl.Locus(contig='chr1', position=11001, reference_genome='GRCh38'), includes_start=True, includes_end=False))._ir.head_str()
Out[5]: 'Interval[Locus(GRCh38)] "{\\"start\\": {\\"contig\\": \\"chr1\\", \\"position\\": 10001}, \\"end\\": {\\"contig\\": \\"chr1\\", \\"position\\": 11001}, \\"includeStart\\": true, \\"includeEnd\\": false}"'

In [6]: len(Out[5])
Out[6]: 183
```

183 bytes per locus interval. The binary representation is just a string and a 32-bit integer, both required. I suspect this fits in no more than 16 bytes.

There are four more fundamental problems:
1. `hl.literal` should probably be more careful about converting a Scala value into a Python value only to stick it inside a literal again.
2. The Scala-Python communication should not use Py4J, a plain-text protocol. https://github.com/hail-is/hail/issues/13756
3. Python->Scala should use a a compact binary encoding rather than the JSON encoding for literals. (Scala->Python already does this). https://github.com/hail-is/hail/issues/13757
4. SemanticHash should not generate a JSON string just to compute a hash. (This becomes irrelevant if large literals are always `EncodedLiteral`, see (3)).

### Version

0.2.117

### Relevant log output

_No response_

danking:
CHANGELOG: Fix #13748, which, since 0.2.117, required at least 10x the RAM to represent a list of intervals.

### DIFF
--- a/hail/python/hail/vds/methods.py
+++ b/hail/python/hail/vds/methods.py
@@ -606,7 +606,7 @@ def _parameterized_filter_intervals(vds: 'VariantDataset',
                 f"'filter_intervals': expect a table with a single key of type {expected}; "
                 f"found {list(intervals.key.dtype.values())}")
         intervals_table = intervals
-        intervals = hl.literal(intervals.aggregate(hl.agg.collect(intervals.key[0]), _localize=False))
+        intervals = intervals.aggregate(hl.agg.collect(intervals.key[0]), _localize=False)
 
     if mode == 'unchecked_filter_both':
         return VariantDataset(hl.filter_intervals(vds.reference_data, intervals, keep),


patrick-schultz (changes requested):
I think this will cause the collect to be recomputed each time `intervals` is used (filtering both variants and reference).

Ideally, we should be doing an interval join here instead of collecting the table. But until then, I'm not sure this is an improvement.

danking:
Interval join won't use the index though, right? I don't think it's necessarily better.

patrick-schultz:
Ah, you're right.

danking:
I suppose there's two root causes here:
1. We use the Py4J socket which inflates a byte array by ~5x (and stores it twice, so 10x really).
2. For Python->Scala we use JSON, which also radically inflates the data size.

danking:
Is there no way to bind the result of the collect and reference in the IR twice?

patrick-schultz:
That's what the literal was doing. The problem is the result is being used in two completely separate pipelines---the variants and reference. The right thing is probably to collect with `localize=False`, and use the result to annotate globals on both sides, but we still don't support referencing the local environment in MapGlobals.

danking:
Ah, right, OK, I forget that we have two parallel pipelines here.

PR #13750, opened 2023-09-29
[ci] Use private base image for GPU testing image

daniel-goldstein:
Trying to rely on our private registry instead of dockerhub where we can.


jigold (approved):

danking:
Looks like the image lacks pythoN?

danking:
WIP until I sort the release.

PR #13751, opened 2023-09-29
[compiler] fix ndarray broadcasting with 0 sizes

patrick-schultz:
Really bad example:
```python
In [2]: a = hl.nd.array(2)

In [3]: empty = np.array([], np.int32).reshape((0, 2, 2))

In [4]: empty = hl.nd.array(empty)

In [5]: hl.eval(empty)
Out[5]: array([], shape=(0, 2, 2), dtype=int32)

In [6]: hl.eval(a + empty)
Out[6]:
array([[[       2, 55622722],
        [       2,        3]]], dtype=int32)
```

Result of sum of shape `()` with shape `(0, 2, 2)` should be shape `(0, 2, 2)` by standard numpy broadcasting rules. Instead, we try to produce a result with shape `(1, 2, 2)`, which contains garbage data.

### DIFF
--- a/hail/python/test/hail/expr/test_ndarrays.py
+++ b/hail/python/test/hail/expr/test_ndarrays.py
@@ -425,6 +425,7 @@ def test_ndarray_map2():
                        [11, 12]],
                       [[13, 14],
                        [15, 16]]])
+    empty = np.array([], np.int32).reshape((0, 2, 2))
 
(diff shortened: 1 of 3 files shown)


ehigham (approved):
looks great. glad to see tests enabled!

danking:
WIP until I sort the release.

PR #13752, opened 2023-09-29
[compiler] Don't append StmtX after ControlX

ehigham:
This change grew out of https://github.com/hail-is/hail/pull/13674.
The idea is simple - we shouldn't be appending code after control statements as such statements are redundant. That idea opened pandora's box, but now we're not generating and dropping dead code anymore.

Main changes that rose form fixing fallout from adding assert in `Block.append`:
- Implement basic control-flow structures (if, while, for, switch) in `CodeBuilderLike` and remove the older implementations from `Code`.
  - main difference is these are built from sequencing `Code` operations rather than being defined from LIR
  - allows for a higher-level implementation that I think is simpler to read.
- Use the type-system to prevent foot-guns like `cb.ifx(cond, label.goto)`.

Other changes:
- rename `ifx`, `forLoop` and `whileLoop` to just `if_`, `for_` and `while_`, respectively.
- Implement loops in-terms of one-another to remove code duplication.
- Fix logic for when to write IRs as some default value behaviour was broken when `HAIL_WRITE_IR_FILES` was set in tests

### DIFF
--- a/hail/src/main/scala/is/hail/asm4s/Code.scala
+++ b/hail/src/main/scala/is/hail/asm4s/Code.scala
@@ -2,7 +2,7 @@ package is.hail.asm4s
 
 import is.hail.expr.ir.EmitCodeBuilder
 import is.hail.lir
-import is.hail.lir.ValueX
+import is.hail.lir.{Block, ControlX, StmtX, ValueX}
 import is.hail.utils._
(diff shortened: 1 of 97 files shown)


chrisvittal (approved):
Nice work!

PR #13755, opened 2023-10-02
[query] fix ndarray concat with size 0 dims

patrick-schultz:
ndarray concat was broken when the first input has size 0 along the concat axis. For example
```
In [3]: hl.eval(hl.nd.hstack([hl.nd.zeros((2, 0)), hl.nd.array([[1.0, 2.0], [3.0, 4.0]])]))
Out[3]:
array([[0., 2.],
       [0., 4.]])
```
The zeros matrix is 2 by 0, so horizontal concatenation should just return the other matrix.
(I once saw the first column filled with random numbers, presumably from a buffer overflow)

I did some cleaning up in the concat implementation, but the functional change is to record the index of the first input which is non-empty along the concat axis, and when resetting to the start of the axis, reset to that non-empty index. Other size 0 inputs are correctly handled when incrementing the index, the problem was that the first read happens before an increment.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/ndarrays/EmitNDArray.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/ndarrays/EmitNDArray.scala
@@ -275,56 +275,51 @@ object EmitNDArray {
                 cb._fatal("need at least one ndarray to concatenate")
               })
 
-              val missing: Code[Boolean] = {
-                if (ndsArraySValue.st.elementEmitType.required)
-                  const(false)
-                else {
-                  val missing = cb.newLocal[Boolean]("ndarray_concat_result_missing")
-                  cb.assign(missing, false)
-                  // Need to check if the any of the ndarrays are missing.
-                  val missingCheckLoopIdx = cb.newLocal[Int]("ndarray_concat_missing_check_idx")
-                  cb.forLoop(cb.assign(missingCheckLoopIdx, 0), missingCheckLoopIdx < arrLength, cb.assign(missingCheckLoopIdx, missingCheckLoopIdx + 1),
-                    cb.assign(missing, missing | ndsArraySValue.isElementMissing(cb, missingCheckLoopIdx))
-                  )
-                  missing
+              IEmitCode(cb, ndsArraySValue.hasMissingValues(cb), {
+                val firstND = ndsArraySValue.loadElement(cb, 0).get(cb).asNDArray
+
+                // compute array of sizes along concat axis, and total size of concat axis
+                val arrayLongPType = PCanonicalArray(PInt64(), true)
+                val (pushSize, finishSizes) = arrayLongPType.constructFromFunctions(cb, region, arrLength, false)
+
+                val concatAxisSize = cb.newLocal[Long](s"ndarray_concat_axis_size", 0L)
+
+                ndsArraySValue.forEachDefined(cb) { (cb, i, nd) =>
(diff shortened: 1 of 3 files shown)


hail/src/main/scala/is/hail/expr/ir/ndarrays/EmitNDArray.scala
```diff
@@ -275,56 +275,51 @@ object EmitNDArray {
...
+                  cb.assign(concatAxisSize, concatAxisSize + dimLength)
+                }
+                val stagedArrayOfSizes = finishSizes(cb)
 
-                  cb.forLoop(cb.assign(loopIdx, 1), loopIdx < arrLength, cb.assign(loopIdx, loopIdx + 1), {
-                    val shapeOfNDAtIdx = ndsArraySValue.loadElement(cb, loopIdx).map(cb) { sCode => sCode.asNDArray }.get(cb).shapeStruct(cb)
-                    val dimLength = cb.newLocal[Long]("dimLength", shapeOfNDAtIdx.loadField(cb, dimIdx).get(cb).asInt64.value)
+                // compute index of first input which has non-zero concat axis size
+                val firstNonEmpty = cb.newLocal[Int]("ndarray_concat_first_nonempty", 0)
+                cb.whileLoop(stagedArrayOfSizes.loadElement(cb, firstNonEmpty).get(cb).asInt64.value.ceq(0L), {
```
ehigham, line 296: is this checking is null? don't we have a `isna` function to make this clearer?

hail/src/main/scala/is/hail/expr/ir/ndarrays/EmitNDArray.scala
```diff
@@ -275,56 +275,51 @@ object EmitNDArray {
...
-                      )
-                    }
-                  })
-                  localDim
+                // check that all sizes along other axes are consistent
+                ndsArraySValue.forEachDefined(cb) { case (cb, _, nd: SNDArrayValue) =>
+                  val mismatchedDim = cb.newLocal[Int]("ndarray_concat_mismatched_dim", -1)
+                  val expected = cb.newLocal[Long]("ndarray_concat_expected_size")
+                  val found = cb.newLocal[Long]("ndarray_concat_found_size")
+                  for (i <- (0 until firstND.st.nDims).reverse if i != axis) {
```
ehigham, line 305: I think you can do something like `first.st.nDims - 1 to 0 by -1`, though `.reverse` is fine too

ehigham (changes requested):

hail/src/main/scala/is/hail/expr/ir/ndarrays/EmitNDArray.scala
patrick-schultz: No, this is checking for size 0 axes.

hail/src/main/scala/is/hail/expr/ir/ndarrays/EmitNDArray.scala
patrick-schultz: I find it tricky to think about endpoint inclusivity using the -1 step, reverse is harder to get wrong.

hail/src/main/scala/is/hail/expr/ir/ndarrays/EmitNDArray.scala
ehigham: Oh yes, arrayOfSizes makes this kinda obvious. duh.

ehigham (approved):
Thanks!

Issue #13690, opened 2023-09-22
[query] show and checkpoint are very slow in Jupyter Notebooks; they are not slow outside Jupyter


danking:
### What happened?

This started happening in 0.2.123. It does not happen in 0.2.120

### Version

0.2.123

danking:
I can't reproduce with
```
import hail as hl
hl.utils.range_table(10).show()
```

Must be something more complex.

iris-garden:
local notebook works fine for me as well, looks to be just dataproc that's not working as expected. submitting that test command as a script finished in 36.2s. notebook is currently still hanging with this output (it's been 11 minutes):
```
BokehJS 3.2.2 successfully loaded.
Initializing Hail with default parameters...
Setting default log level to "WARN".
To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).
SPARKMONITOR_LISTENER: Started SparkListener for Jupyter Notebook
SPARKMONITOR_LISTENER: Port obtained from environment: 55989
SPARKMONITOR_LISTENER: Application Started: application_1695402030462_0001 ...Start Time: 1695402594764
Running on Apache Spark version 3.3.0
SparkUI available at http://notebook-slowdown-repro-m.c.broad-ctsa.internal:43055/
Welcome to
     __  __     <>__
    / /_/ /__  __/ /
   / __  / _ `/ / /
  /_/ /_/\_,_/_/_/   version 0.2.124-ee7fef6fc40d
LOGGING: writing to /home/hail/hail-20230922-1709-0.2.124-ee7fef6fc40d.log
[Stage 0:>                                                          (0 + 2) / 2]
```

danking:
@iris-garden can you grab that log file and upload here? it should live on the `CLUSTER_NAME-m` machine.

iris-garden:
yep, [here it is](https://gist.github.com/iris-garden/bfacd8246bb8b29cd67caea063273c16)

danking:
Nothing suspicious there. Something is going wrong in the executors. I think the only way we're gonna solve this is by running a pipeline and looking at the executor logs. I'm at a complete loss for how Jupyter could affect what happens on the executors.

iris-garden:
[...]

PR #13759, opened 2023-10-02
[query] fixes hail jar for dataproc notebooks

iris-garden:
closes https://github.com/hail-is/hail/issues/13690.

to test that this works, i've been running these commands from the root of my clone of the hail repo:

```bash
make -C hail install-editable
make -C hail install-hailctl
hailctl dataproc start notebook-slowdown-repro --region us-central1
hailctl dataproc connect notebook-slowdown-repro notebook
```

and then running this minimal example in the notebook:

```python
import hail
hail.utils.range_table(10).show()
```

and making sure it outputs a visual of the table, instead of getting stuck displaying `Stage 0:> (0+X)/Y` and not progressing.

### DIFF
--- a/hail/build.gradle
+++ b/hail/build.gradle
@@ -85,6 +85,8 @@ compileScala {
 }
 
 configurations {
+    justSpark
+
     all {
         resolutionStrategy {
             eachDependency { DependencyResolveDetails details ->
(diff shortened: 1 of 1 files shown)


iris-garden:
my strategy here was just to undo everything added in [this commit](https://github.com/hail-is/hail/commit/12e0f497db0f3e5453f870495e48e44191b315f4), except the version upgrades, so i'm not sure if there are some changes i'm making here that are unnecessary or produce weird results as far as what all ends up in the jar or anything

from running `jar -tf` on the jar produced by the current `main` and the one produced by the commit prior to the one i'm partially reverting, it looked like the updates to the config only added things to the jar, rather than removing any, so hopefully that should be fine

danking:
Wow, really excellent find @iris-garden . I must admit, I don't really understand how this helps. I'm not sure what `justSpark` is doing, it seems like it's completely unused?

It makes sense to me that we want to exclude Spark & Scala, but I don't understand how we're pulling those in? Anyway, I'm glad it works.

PR #13713, opened 2023-09-26
[query] fix backoff code

danking:
CHANGELOG: Fixes #13704, in which Hail could encounter an IllegalArgumentException if there are too many transient errors.

I need to do the multiplication in 64-bits so that it does not wrap around to a large negative value. Then I can use `math.min` with the maxDelayMs to get us back into 32-bits.

I'm just pushing through a bunch of bugs to get Wenhan unblocked today.

PR #13776, opened 2023-10-04
[query] substantially reduce single core latency for force-count

danking:
CHANGELOG: Reduce latency on simple pipelines by as much as 50% by reducing decoding time.

Force count essentially tests decoding because it forces decoding but then just increments a counter by one. Analysis of profile results indicates that the array inplace decoder was perhaps 50% of time, but exactly what part of decoding was unclear.

I attempted many different things. I eventually settled on loop unrolling as the primary benefit. After team meeting, I applied @patrick-schultz 's advice to use bit twiddling to further improve the speed.

---

I assessed the latency using `time python3` on this file:

```python
import hail as hl
hl.init(master='local[1]')
hl._set_flags(write_ir_files='1')
hl.read_matrix_table('/Users/dking/projects/hail-data/foo.mt')._force_count_rows()
```

`foo.mt` is a subset of the `variant_data` from a VDS with ~80k samples, ~300k variants, stored in ~1.6GiB.

1. This PR: 34s, 33s
2. no twiddling: 43s, 43s https://github.com/hail-is/hail/compare/main...danking:hail:unroll-64
3. no twiddling & 8 element blocks: 37s, 38s https://github.com/hail-is/hail/compare/main...danking:hail:unroll-8
4. `main` (`481cfc201b [query] fix backoff code (#13713)`): 68s, 69s

In YourKit, I observe that (1) reads 50-70MB/s with one core whereas (4) reads 15-35MB/s.

I also assessed the 10-core latency and JIT effects:

- (1) starts at ~12s, warms to ~6s (+- 0.5s). Peak bandwidth 490MB/s.
- (4) starts at ~17s and warms up to ~11s (+- 2s). Peak bandwidth ~250MB/s.

I suspect, with this PR, the multi-core speed is fast enough to saturate any of our file stores (including my laptop, which I think taps out just around ~500MB/s).

Big thanks to everyone who contributed, particularly @patrick-schultz, whose suggestion to use bit-twiddling, squeezeed another 10% off the 8 element blocks.

### DIFF
--- a/hail/src/main/scala/is/hail/types/encoded/EArray.scala
+++ b/hail/src/main/scala/is/hail/types/encoded/EArray.scala
@@ -100,22 +100,69 @@ final case class EArray(val elementType: EType, override val required: Boolean =
 
     val len = cb.newLocal[Int]("len", in.readInt())
     val array = cb.newLocal[Long]("array", arrayType.allocate(region, len))
+    val elemAddr = cb.newLocal[Long]("elemAddr", arrayType.firstElementOffset(array, len))
     arrayType.storeLength(cb, array, len)
 
     val i = cb.newLocal[Int]("i")
+    val mbyteOffset = cb.newLocal[Long]("mbyteOffset", array + arrayType.lengthHeaderBytes)
+    val mbyte = cb.newLocal[Byte]("mbyte", 0.toByte)
+    val presentBitsLong = cb.newLocal[Long]("presentBitsLong", 0)
+    val inBlockIndexToPresentValue = cb.newLocal[Int]("inBlockIndexToPresentValue", 0)
     val readElemF = elementType.buildInplaceDecoder(arrayType.elementType, cb.emb.ecb)
 
     if (!elementType.required)
       cb += in.readBytes(region, array + const(arrayType.lengthHeaderBytes), arrayType.nMissingBytes(len))
 
-    cb.forLoop(cb.assign(i, 0), i < len, cb.assign(i, i + 1), {
-      val elemAddr = cb.memoize(arrayType.elementOffset(array, len, i))
-      if (elementType.required)
-        readElemF(cb, region, elemAddr, in)
-      else
-        cb.ifx(arrayType.isElementDefined(array, i),
-          readElemF(cb, region, elemAddr, in))
-    })
+    val LONG_MASK_ALL_BUT_FIRST = const(0x7fffffffffffffffL)
+
+    def numberOfLeadingZeros(l: Code[Long]): Code[Int] =
+      Code.invokeStatic1[java.lang.Long, Long, Int]("numberOfLeadingZeros", l)
+
+    cb.assign(i, 0)
+    if (elementType.required) {
+      cb.forLoop({}, i + 64 < len, cb.assign(i, i + 64), {
+        for (_ <- 0 to 63) {
+          readElemF(cb, region, elemAddr, in)
+          cb.assign(elemAddr, arrayType.nextElementAddress(elemAddr))
+        }
+      })
+    } else {
+      cb.forLoop({}, i + 64 < len, cb.assign(i, i + 64), {
+        cb.assign(presentBitsLong, ~Region.loadLong(mbyteOffset))
+        cb.assign(mbyteOffset, mbyteOffset + 8)
+        cb.assign(inBlockIndexToPresentValue, numberOfLeadingZeros(presentBitsLong))
+
+        cb.whileLoop(inBlockIndexToPresentValue < 64, {
+          cb.assign(elemAddr,
+            arrayType.elementOffset(array, len, i + inBlockIndexToPresentValue))
+          readElemF(cb, region, elemAddr, in)
+
+          cb.assign(inBlockIndexToPresentValue,
+            numberOfLeadingZeros(presentBitsLong & (LONG_MASK_ALL_BUT_FIRST >>> inBlockIndexToPresentValue)))
+        })
+      })
+      cb.assign(elemAddr, arrayType.elementOffset(array, len, i))
+    }
+    cb.forLoop({}, i < len,
+      {
+        cb.assign(i, i + 1)
+        cb.assign(elemAddr, arrayType.nextElementAddress(elemAddr))
+      },
+      {
+        if (elementType.required) {
+          readElemF(cb, region, elemAddr, in)
+        } else {
+          cb.ifx((i % 8).ceq(0),
+            {
+              cb.assign(mbyte, Region.loadByte(mbyteOffset))
+              cb.assign(mbyteOffset, mbyteOffset + 1)
+            }
+          )
+          cb.ifx((mbyte.load() & (const(1) << (i & 7))).ceq(0),
+            readElemF(cb, region, elemAddr, in))
+        }
+      }
+    )
 
     new SIndexablePointerValue(st, array, len, cb.memoize(arrayType.firstElementOffset(array, len)))
   }
--- a/hail/src/main/scala/is/hail/types/encoded/EArray.scala
+++ b/hail/src/main/scala/is/hail/types/encoded/EArray.scala
@@ -115,9 +115,6 @@ final case class EArray(val elementType: EType, override val required: Boolean =
 
     val LONG_MASK_ALL_BUT_FIRST = const(0x7fffffffffffffffL)
 
-    def numberOfLeadingZeros(l: Code[Long]): Code[Int] =
-      Code.invokeStatic1[java.lang.Long, Long, Int]("numberOfLeadingZeros", l)
-
     cb.assign(i, 0)
     if (elementType.required) {
       cb.forLoop({}, i + 64 < len, cb.assign(i, i + 64), {
@@ -130,7 +127,7 @@ final case class EArray(val elementType: EType, override val required: Boolean =
       cb.forLoop({}, i + 64 < len, cb.assign(i, i + 64), {
         cb.assign(presentBitsLong, ~Region.loadLong(mbyteOffset))
         cb.assign(mbyteOffset, mbyteOffset + 8)
-        cb.assign(inBlockIndexToPresentValue, numberOfLeadingZeros(presentBitsLong))
+        cb.assign(inBlockIndexToPresentValue, presentBitsLong.numberOfLeadingZeros)
 
         cb.whileLoop(inBlockIndexToPresentValue < 64, {
           cb.assign(elemAddr,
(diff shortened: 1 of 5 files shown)


hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -100,22 +100,69 @@ final case class EArray(val elementType: EType, override val required: Boolean =
...
-      if (elementType.required)
-        readElemF(cb, region, elemAddr, in)
-      else
-        cb.ifx(arrayType.isElementDefined(array, i),
-          readElemF(cb, region, elemAddr, in))
-    })
+    val LONG_MASK_ALL_BUT_FIRST = const(0x7fffffffffffffffL)
+
+    def numberOfLeadingZeros(l: Code[Long]): Code[Int] =
+      Code.invokeStatic1[java.lang.Long, Long, Int]("numberOfLeadingZeros", l)
```
patrick-schultz, line 119: Btw, we already have `Code[Long].numberOfLeadingZeros` (and trailing too)

patrick-schultz:
One concern that shouldn't be hard to fix: we probably don't want to unroll loops with non-primitive element types that require significant decoding code themselves. That could lead to exponential generated code sizes for deeply nested types. And the loop unrolling shouldn't have any benefit in that case anyways.

hail/src/main/scala/is/hail/types/encoded/EArray.scala
danking: I am crushed to learn this is completely unused elsewhere.

danking:
The per-element code seems to end up in a separate method. Is that not standard? For example, the pipeline I'm using is an array of struct and the struct is always in a separate method

hail/src/main/scala/is/hail/types/encoded/EArray.scala
danking: Fixed.

hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -100,22 +100,69 @@ final case class EArray(val elementType: EType, override val required: Boolean =
...
+        cb.assign(inBlockIndexToPresentValue, numberOfLeadingZeros(presentBitsLong))
+
+        cb.whileLoop(inBlockIndexToPresentValue < 64, {
+          cb.assign(elemAddr,
+            arrayType.elementOffset(array, len, i + inBlockIndexToPresentValue))
+          readElemF(cb, region, elemAddr, in)
+
+          cb.assign(inBlockIndexToPresentValue,
+            numberOfLeadingZeros(presentBitsLong & (LONG_MASK_ALL_BUT_FIRST >>> inBlockIndexToPresentValue)))
+        })
```
patrick-schultz: I think this saves a few instructions by avoiding adding the length of the missing bits to the base address every time, and avoiding the masking
```suggestion
        cb.assign(presentBitsLong, ~Region.loadLong(mbyteOffset))
        cb.assign(mbyteOffset, mbyteOffset + 8)
        cb.assign(elemAddr,
          arrayType.elementOffset(array, len, i))

        cb.whileLoop(presentBitsLong != 0, {
          cb.assign(numToSkip, numberOfLeadingZeros(presentBitsLong))
          cb.assign(elemAddr, elemAddr + arrayType.elementByteSize * numToSkip)
          readElemF(cb, region, elemAddr, in)
          cb.assign(presentBitsLong, presentBitsLong << (numToSkip + 1))
        })
```

patrick-schultz (changes requested):
A possible optimization, ~but more importantly the code size concern above~. Nm, if nested elements decoding is a separate method, that's perfect.

hail/src/main/scala/is/hail/types/encoded/EArray.scala
danking: hehe, tricksy!

I was worried something like this wasn't safe in general. I'm not really sure what kinds of contracts we have on the PTypes. Are you suggesting that the contract is that the PArrays are always dense and laid out in order?

hail/src/main/scala/is/hail/types/encoded/EArray.scala
patrick-schultz: I’m using pop count, I think in the rng. Don’t remember if I used leading zeroes at some point, or if I just added them at the same time as pop count out of thoroughness.

hail/src/main/scala/is/hail/types/encoded/EArray.scala
patrick-schultz: Oh that’s true. Should probably add a method to PArray like nextElementAddress that takes a relative offset.

danking:
Looks like I have a bug. I'll have to dig into this.

hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -100,22 +100,66 @@ final case class EArray(val elementType: EType, override val required: Boolean =
...
+        cb.whileLoop(inBlockIndexToPresentValue < 64, {
+          cb.assign(elemAddr,
+            arrayType.elementOffset(array, len, i + inBlockIndexToPresentValue))
+          readElemF(cb, region, elemAddr, in)
+
+          cb.assign(inBlockIndexToPresentValue,
+            (presentBitsLong & (LONG_MASK_ALL_BUT_FIRST >>> inBlockIndexToPresentValue)).numberOfLeadingZeros)
+        })
+      })
+      cb.assign(elemAddr, arrayType.elementOffset(array, len, i))
```
patrick-schultz, line 141: Ah, I think you have the bit ordering wrong. We arrange bytes left-to-right, and bits within a byte right-to-left (which I think is right; the index 0 bit should be the least-significant bit). It might be worth changing PCanonicalArray to group missing bits by 64 rather than 8, to make tricks like this faster. I don't expect reading a byte is any faster than reading a long. But in the meantime, we can just use `Long.reverseBytes`.

Note that the right-to-left bit order allows a faster way to do your original approach:

```scala
    def reverseBytes(l: Code[Long]): Code[Long] =
      Code.invokeStatic1[java.lang.Long, Long, Long]("reverseBytes", l)
```
```suggestion
      cb.assign(blockAddr, arrayType.firstElementOffset(array, len))
      val endPtr = cb.memoize(arrayType.elementOffset(array, len, len - 64))
      // need to make nextElementAddress(addr, offset) or something
      cb.forLoop({}, elemAddr < endPtr, cb.assign(blockAddr, arrayType.nextElementAddress(array, offset=64)), {
        cb.assign(presentBitsLong, ~reverseBytes(Region.loadLong(mbyteOffset)))
        cb.assign(mbyteOffset, mbyteOffset + 8)

        cb.whileLoop(presentBitsLong != 0, {
          cb.assign(inBlockIndexToPresentValue, presentBitsLong.numberOfTrailingZeros)
          val cb.assign(elemAddr,
            arrayType.nextElementOffset(blockAddr, inBlockIndexToPresentValue))
          readElemF(cb, region, elemAddr, in)

          // unset rightmost set bit
          cb.assign(presentBitsLong, presentBitsLong & (presentBitsLong - 1))
        })
      })
```

patrick-schultz (changes requested):

hail/src/main/scala/is/hail/types/encoded/EArray.scala
danking: killer find, this would've taken me quite a while to sort out.

danking:
I'm gonna see if this code works on the full test suite. I'm finding it pretty tricky to work with this stuff. The code I wrote for the last remaining elements in the non-required case seems to be wrong.

danking:
This is still broken. I think my plan now is to revert back to my first attempt and just fix that to traverse the bytes in the right order. Doing that while also changing our iteration pattern to be based on the pointers rather than an index is introducing too many bugs.

Doing C++ iterator style pointer does seem cleaner and easier to read than the sort of jumbled mess I had before.

danking:
https://github.com/hail-is/hail/pull/13786

danking:
oops

danking:
https://github.com/hail-is/hail/pull/13786

patrick-schultz (commented):
I'm totally fine with getting a simple version in that still has a big performance improvement. Just leaving this here for future reference.

And re. making assumptions about the array layout: we currently only have one array ptype, and the docoder asserts that the requested ptype is a `PCanonicalArray`, so it's really not an issue. If we add other array ptypes, they'll probably need completely separate decoders.

hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -102,20 +102,72 @@ final case class EArray(val elementType: EType, override val required: Boolean =
...
+          cb.whileLoop(presentBits.cne(0L), {
+            cb.assign(inBlockIndexToPresentValue, presentBits.numberOfTrailingZeros)
+            val elemOff = cb.memoize(
+              arrayType.incrementElementOffset(blockOff, inBlockIndexToPresentValue))
+            readElemF(cb, region, elemOff, in)
+            cb.assign(presentBits, unsetRightMostBit(presentBits))
+          })
+        }
+      )
+      val i = cb.newLocal[Int]("i", len - (len % 64))
```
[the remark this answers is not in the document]
patrick-schultz: It should also be possible to avoid the separate loop for the last block. Here's an untested attempt. The idea is to compute a mask which usually leaves all bits unchanged, but on the last incomplete block, zeroes bits which are past the end. Since the inner loop only writes to positions with a set bit, that ensures we handle the last block correctly.
```scala
      val partialBlockSize = cb.memoize(len % 64)
      val partialBlockStart = cb.memoize(arrayType.elementOffset(array, len, len - partialBlockSize)
      val partialBlockMask = cb.memoize(-1 << partialBlockSize)
      val mask = cb.newLocal[Long]("mask", 0L)
      val presentBits = cb.newLocal[Long]("presentBits", 0L)
      val mbyteOffset = cb.newLocal[Long]("mbyteOffset", array + arrayType.lengthHeaderBytes)
      val blockOff = cb.newLocal[Long]("blockOff", arrayType.firstElementOffset(array, len))
      val inBlockIndexToPresentValue = cb.newLocal[Int]("inBlockIndexToPresentValue", 0)

      cb.forLoop(
        {},
        blockOff < pastLastOff,
        cb.assign(blockOff, arrayType.incrementElementOffset(blockOff, 64)),
        {
          // hopefully generates a conditional move, could also try blockOff.ceq(partialBlockStart).toL * partialBlockMask
          cb.ifx(blockOff.ceq(partialBlockStart), cb.assign(mask, partialBlockMask, 0L))
          cb.assign(presentBits, ~(Region.loadLong(mbyteOffset).reverseBytes | mask))
          cb.assign(mbyteOffset, mbyteOffset + 8)
          cb.whileLoop(presentBits.cne(0L), {
            cb.assign(inBlockIndexToPresentValue, presentBits.numberOfTrailingZeros)
            val elemOff = cb.memoize(
              arrayType.incrementElementOffset(blockOff, inBlockIndexToPresentValue))
            readElemF(cb, region, elemOff, in)
            cb.assign(presentBits, unsetRightMostBit(presentBits))
          })
        }
      )
```

PR #13778, opened 2023-10-04
[query] do not reinitialize zstd repeatedly

danking:
In the end, this did not seem to have any significant effect on decoding, but it does show up in the profiler as a hotspot.

### DIFF
--- a/hail/src/main/scala/is/hail/io/InputBuffers.scala
+++ b/hail/src/main/scala/is/hail/io/InputBuffers.scala
@@ -3,12 +3,13 @@ package is.hail.io
 import java.io._
 import java.util
 import java.util.UUID
+import java.util.function.Supplier
 
 import is.hail.annotations.{Memory, Region}
 import is.hail.io.compress.LZ4
 import is.hail.utils._
 
-import com.github.luben.zstd.Zstd
+import com.github.luben.zstd.{Zstd, ZstdDecompressCtx}
 
 trait InputBuffer extends Closeable {
   def close(): Unit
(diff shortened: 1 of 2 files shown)


hail/src/main/scala/is/hail/io/InputBuffers.scala
```diff
@@ -646,16 +652,15 @@ final class ZstdInputBlockBuffer(blockSize: Int, in: InputBlockBuffer) extends I
     } else {
       val compLen = blockLen - 4
       val decompLen = Memory.loadInt(comp, 0)
-      val ret = Zstd.decompressByteArray(buf, 0, decompLen, comp, 4, compLen)
-      if (Zstd.isError(ret))
-        throw new com.github.luben.zstd.ZstdException(ret)
+      val ret = zstd.decompressByteArray(buf, 0, decompLen, comp, 4, compLen)
```
chrisvittal, line 655: If we trust these methods to throw then we don't need to bind their return values.

hail/src/main/scala/is/hail/io/InputBuffers.scala
```diff
@@ -675,9 +680,7 @@ final class ZstdSizedBasedInputBlockBuffer(blockSize: Int, in: InputBlockBuffer)
         compLen
       } else {
         val decompLen = decomp >>> 1
-        val ret = Zstd.decompressByteArray(buf, 0, decompLen, comp, 4, compLen)
-        if (Zstd.isError(ret))
-          throw new com.github.luben.zstd.ZstdException(ret)
+        val ret = zstd.decompressByteArray(buf, 0, decompLen, comp, 4, compLen)
```
chrisvittal, line 683: unused

chrisvittal (changes requested):
Looks good. Thanks for making this change.

hail/src/main/scala/is/hail/io/InputBuffers.scala
danking: good point! For the record, they can fail, but the `throw` the exception. The return value is always the number of bytes, never an error code.

chrisvittal (approved):

PR #13787, opened 2023-10-10
[query] faster array decoder

patrick-schultz:
Picking up where #13776 left off.

CHANGELOG: improved speed of reading hail format datasets from disk

This PR speeds up decoding arrays in two main ways:
* instead of calling `arrayType.isElementDefined(array, i)` on every single array element, which expands to
  ```scala
  val b = aoff + lengthHeaderBytes + (i >> 3)
  !((Memory.loadByte(b) & (1 << (i & 7).toInt)) != 0)
  ```
  process elements in groups of 64, and load the corresponding long of missing bits once
* once we have a whole long of missing bits, we can be smarter than branching on each bit. After flipping to get `presentBits`, we use the following psuedocode to extract the positions of the set bits, with time proportional to the number of set bits:
  ```
  while (presentBits != 0) {
    val idx = java.lang.Long.numberOfTrailingZeroes(presentBits)
    // do something with idx
    presentBits = presentBits & (presentBits - 1) // unsets the rightmost set bit
  }
  ```

To avoid needing to handle the last block of 64 elements differently, this PR changes the layout of `PCanonicalArray` to ensure the missing bits are always padded out to a multiple of 64 bits. They were already padded to a multiple of 32, and I don't expect this change to have much of an effect. But if needed, blocking by 32 elements instead had very similar performance in my benchmarks.

I also experimented with unrolling loops. In the non-missing case, this is easy. In the missing case, I tried using `if (presentBits.bitCount >= 8)` to guard an unrolled inner loop. In both cases, unrolling was if anything slower.

Dan observed benefit from unrolling, but that was combined with the first optimization above (not loading a bit from memory every element), which I beleive was the real source of improvement.

### DIFF
--- a/hail/src/main/scala/is/hail/types/encoded/EArray.scala
+++ b/hail/src/main/scala/is/hail/types/encoded/EArray.scala
@@ -109,12 +109,21 @@ final case class EArray(val elementType: EType, override val required: Boolean =
       val elemOff = cb.newLocal[Long]("elemOff", arrayType.firstElementOffset(array, len))
       if (arrayType.trivialElements) {
         // elements have 0 size, so all elements have the same address
-        // still need to read `len` elements from the input stream
+        // still need to read `len` elements from the input stream, as they may have non-zero size
         val i = cb.newLocal[Int]("i", 0)
-        cb.forLoop({}, i < len, cb.assign(i, i+1), readElemF(cb, region, elemOff, in))
+        cb.forLoop({}, i < (len & -8), cb.assign(i, i + 8), {
+          for (k <- 0 until 8)
+            readElemF(cb, region, cb.memoize(arrayType.incrementElementOffset(elemOff, k)), in)
+        })
+        cb.forLoop({}, i < len, cb.assign(i, i + 1), readElemF(cb, region, elemOff, in))
       } else {
+        val lastBlockAddr = cb.memoize(arrayType.elementOffset(array, len, len & -8))
+        cb.forLoop({}, elemOff < lastBlockAddr, cb.assign(elemOff, arrayType.incrementElementOffset(elemOff, 8)), {
+          for (i <- 0 until 8)
+            readElemF(cb, region, cb.memoize(arrayType.incrementElementOffset(elemOff, i)), in)
+        })
         cb.forLoop({}, elemOff < pastLastOff, cb.assign(elemOff, arrayType.nextElementAddress(elemOff)), {
-            readElemF(cb, region, elemOff, in)
+          readElemF(cb, region, elemOff, in)
         })
       }
     } else {
--- a/hail/src/main/scala/is/hail/types/encoded/EArray.scala
+++ b/hail/src/main/scala/is/hail/types/encoded/EArray.scala
@@ -98,8 +98,8 @@ final case class EArray(val elementType: EType, override val required: Boolean =
       case t: PCanonicalDict => t.arrayRep
     }
 
-    val len = cb.memoize(in.readInt())
-    val array = cb.memoize(arrayType.allocate(region, len))
+    val len = cb.memoize(in.readInt(), "len")
+    val array = cb.memoize(arrayType.allocate(region, len), "array")
     arrayType.storeLength(cb, array, len)
 
     val readElemF = elementType.buildInplaceDecoder(arrayType.elementType, cb.emb.ecb)
--- a/hail/src/main/scala/is/hail/types/encoded/EArray.scala
+++ b/hail/src/main/scala/is/hail/types/encoded/EArray.scala
@@ -90,7 +90,6 @@ final case class EArray(val elementType: EType, override val required: Boolean =
   }
 
   override def _buildDecoder(cb: EmitCodeBuilder, t: Type, region: Value[Region], in: Value[InputBuffer]): SValue = {
-    cb.println("in decoder")
     val st = decodedSType(t).asInstanceOf[SIndexablePointer]
 
     val arrayType: PCanonicalArray = st.pType match {
@@ -106,26 +105,27 @@ final case class EArray(val elementType: EType, override val required: Boolean =
     val readElemF = elementType.buildInplaceDecoder(arrayType.elementType, cb.emb.ecb)
 
     val pastLastOff = cb.memoize(arrayType.pastLastElementOffset(array, len))
-
     if (elementType.required) {
       val elemOff = cb.newLocal[Long]("elemOff", arrayType.firstElementOffset(array, len))
-      cb.forLoop(
-        {}, elemOff < pastLastOff, cb.assign(elemOff, arrayType.nextElementAddress(elemOff)), {
-          readElemF(cb, region, elemOff, in)
+      if (arrayType.trivialElements) {
+        // elements have 0 size, so all elements have the same address
+        // still need to read `len` elements from the input stream
+        val i = cb.newLocal[Int]("i", 0)
+        cb.forLoop({}, i < len, cb.assign(i, i+1), readElemF(cb, region, elemOff, in))
+      } else {
+        cb.forLoop({}, elemOff < pastLastOff, cb.assign(elemOff, arrayType.nextElementAddress(elemOff)), {
+            readElemF(cb, region, elemOff, in)
         })
+      }
     } else {
       cb += in.readBytes(region, array + const(arrayType.missingBytesOffset), arrayType.nMissingBytes(len))
-      cb.println("len = ", len.toS)
-      arrayType.printDebug(cb, array)
 
-//      cb.ifx((len % 64).cne(0), {
-//        // ensure that the last missing block has all missing bits set past the last element
-////        arrayType.printDebug(cb, array)
-//        val lastMissingBlockAddr = cb.memoize(arrayType.firstElementOffset(array) - 8)
-//        val lastMissingBlock = cb.memoize(Region.loadLong(lastMissingBlockAddr))
-//        cb += Region.storeLong(lastMissingBlockAddr, (lastMissingBlock | (const(-1L) << len)))
-////        arrayType.printDebug(cb, array)
-//      })
+      cb.ifx((len % 64).cne(0), {
+        // ensure that the last missing block has all missing bits set past the last element
+        val lastMissingBlockAddr = cb.memoize((arrayType.pastLastMissingByteAddr(array, len) - 1) & -8)
+        val lastMissingBlock = cb.memoize(Region.loadLong(lastMissingBlockAddr))
+        cb += Region.storeLong(lastMissingBlockAddr, lastMissingBlock | (const(-1L) << len))
+      })
 
       def unsetRightMostBit(x: Value[Long]): Code[Long] =
         x & (x - 1)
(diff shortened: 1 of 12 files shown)


hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -98,24 +98,58 @@ final case class EArray(val elementType: EType, override val required: Boolean =
       case t: PCanonicalDict => t.arrayRep
     }
 
-    val len = cb.newLocal[Int]("len", in.readInt())
-    val array = cb.newLocal[Long]("array", arrayType.allocate(region, len))
+    val len = cb.memoize(in.readInt())
+    val array = cb.memoize(arrayType.allocate(region, len))
```
danking, line 102: preserve the name hints

hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -98,24 +98,58 @@ final case class EArray(val elementType: EType, override val required: Boolean =
...
-      if (elementType.required)
-        readElemF(cb, region, elemAddr, in)
-      else
-        cb.ifx(arrayType.isElementDefined(array, i),
-          readElemF(cb, region, elemAddr, in))
-    })
+      cb.ifx((len % 64).cne(0), {
+        // ensure that the last missing block has all missing bits set past the last element
+//        arrayType.printDebug(cb, array)
+        val lastMissingBlockAddr = cb.memoize(arrayType.firstElementOffset(array) - 8)
```
danking, line 121: If the array is short, the last missing block might be smaller than 8 bytes, right?

hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -98,24 +98,58 @@ final case class EArray(val elementType: EType, override val required: Boolean =
...
+      val blockOff = cb.newLocal[Long]("blockOff", arrayType.firstElementOffset(array, len))
+      val inBlockIndexToPresentValue = cb.newLocal[Int]("inBlockIndexToPresentValue", 0)
+      val blockOffIncrement = arrayType.elementByteSize * 64
+
+      cb.forLoop(
+        {},
+          blockOff < pastLastOff,
+          cb.assign(blockOff, blockOff + blockOffIncrement),
+        {
+          cb.assign(presentBits, ~Region.loadLong(mbyteOffset))
```
danking, line 141: Reverse the bytes, right?

danking (changes requested):

hail/src/main/scala/is/hail/types/encoded/EArray.scala
patrick-schultz: No, I was wrong about that. `sun.unsafe` uses the system endianness, so little endian, which means no reverse needed.

hail/src/main/scala/is/hail/types/encoded/EArray.scala
patrick-schultz: The idea I'm trying here is to ensure that missing bits are always padded with 1s to a multiple of 64 bits. So we can treat the last block the same as any other, as long as we only try to access present elements.

danking: With the alignment change below, I think we're safe to read the whole long.

patrick-schultz: exactly

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SContainer.scala
```diff
@@ -81,13 +81,14 @@ trait SIndexableValue extends SValue {
   override def sizeToStoreInBytes(cb: EmitCodeBuilder): SInt64Value = {
     val storageType = this.st.storageType().asInstanceOf[PContainer]
     val length = this.loadLength()
-    val totalSize = cb.newLocal[Long]("sindexableptr_size_in_bytes", storageType.elementsOffset(length).toL)
+    val totalSize = cb.newLocal[Long]("sindexableptr_size_in_bytes")
     if (this.st.elementType.containsPointers) {
       this.forEachDefined(cb) { (cb, _, element) =>
+        cb.assign(totalSize, storageType.elementsOffset(length).toL)
         cb.assign(totalSize, totalSize + element.sizeToStoreInBytes(cb).value)
```
danking, line 88: This is wrong, no? You add the `elementsOffset` (meaning the number of missing bytes) for each defined element of the array rather than just once. I'm a bit surprised/concerned this didn't fail any tests.

danking (changes requested):

danking:
Also: need a CHANGELOG and can you write a little note for our future selves in the PR description?

danking:
FWIW, I built this branch and ran it twice as described in https://github.com/hail-is/hail/pull/13776 and got 35s, 36s.

danking:
Seems to average 60MB/s. No clear culprits. Zstd decoding is the top hit right now.

The hottest generated code is inplace decoding of an optional array of optional int32. Really sucks because things like `LA` are somehow getting written as element optional, even though, by construction their elements are not optional.

```
+EBaseStruct{
  `the entries! [877f12a8827e18f61222c6c8c5fb04a8]`:
    +EArray[EBaseStruct{
       LA:EArray[EInt32]
      ,LGT:EInt32
      ,LAD:EArray[EInt32]
      ,LPGT:EInt32
      ,LPL:EArray[EInt32]
      ,RGQ:EInt32,
      gvcf_info: EBaseStruct{
         AC:EArray[EInt32]
        ,AF:EArray[EFloat64]
        ,AN:EInt32,AS_BaseQRankSum:EArray[EFloat64]
        ,AS_FS:EArray[EFloat64]
        ,AS_InbreedingCoeff:EArray[EFloat64]
        ,AS_MQ:EArray[EFloat64]
        ,AS_MQRankSum:EArray[EFloat64]
        ,AS_QD:EArray[EFloat64]
        ,AS_QUALapprox:EArray[EInt32]
        ,AS_RAW_BaseQRankSum:EBinary,AS_RAW_MQ:EArray[EFloat64]
        ,AS_RAW_MQRankSum:EArray[EBaseStruct{`0`:EFloat64,`1`:EInt32}]
        ,AS_RAW_ReadPosRankSum:EArray[EBaseStruct{`0`:EFloat64,`1`:EInt32}]
        ,AS_ReadPosRankSum:EArray[EFloat64]
        ,AS_SB_TABLE:EArray[EArray[EInt32]]
        ,AS_SOR:EArray[EFloat64]
        ,AS_VarDP:EArray[EInt32]
        ,BaseQRankSum:EFloat64,ExcessHet:EFloat64,FS:EFloat64,InbreedingCoeff:EFloat64,MQ:EFloat64,MQRankSum:EFloat64,MQ_DP:EInt32,QD:EFloat64,QUALapprox:EInt32,RAW_GT_COUNT:EArray[EInt32]
        ,RAW_MQandDP:EArray[EInt32]
        ,ReadPosRankSum:EFloat64,SOR:EFloat64,VarDP:EInt32}
      ,DP:EInt32
      ,GQ:EInt32
      ,MIN_DP:EInt32
      ,PID:EBinary
      ,PS:EInt32
      ,SB:EArray[EInt32]
    }
  ]
}
```

Async profiler periodic sampling:
Screenshot 2023-10-10 at 18 06 38
Screenshot 2023-10-10 at 18 07 10
Sync profiler (note safe point bias)
Screenshot 2023-10-10 at 18 32 14

hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
```diff
@@ -85,6 +99,11 @@ final case class PCanonicalArray(elementType: PType, required: Boolean = false)
     _elementsOffset(length)
   }
 
+  def missingBytesOffset: Long = lengthHeaderBytes
```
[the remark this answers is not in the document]
patrick-schultz: > This is a misleading name. We pervasively use the word offset to mean address (we should stop doing that).

Agreed, but this is unfortunately also consistent with `elementsOffset`. I'm very tempted to just go through and fix all the naming, at least in `PCanonicalArray`, but maybe after this PR.

What I'm trying to do here is draw some boundaries around what details of the memory layout consumers may depend on. So far I've done that just by making some things private, but it needs to be described in text too. My current thought is: there is an integer length, a byte array of missingness bits padded to a multiple of 8 bytes, and an elements array. The start of each of the three sections is 8-byte aligned. Each section occupies a contiguous slice of memory. Consumers may access the start address and length of each section, but may not depend on their relative position, such as their order or the distance in between them.

This line was because the length of the header shouldn't be accessible, but the start of the missing bytes should. The naming was to be consistent with the other section start, `elementsOffset`, which I agree is terrible and needs to be changed.

hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -46,8 +46,8 @@ final case class EArray(val elementType: EType, override val required: Boolean =
         if (!elementType.required) {
           val nMissingLocal = cb.newLocal[Int]("nMissingBytes", pArray.nMissingBytes(prefixLen))
           cb.ifx(nMissingLocal > 0, {
-            cb += out.writeBytes(array + const(pArray.lengthHeaderBytes), nMissingLocal - 1)
-            cb += out.writeByte((Region.loadByte(array + const(pArray.lengthHeaderBytes)
+            cb += out.writeBytes(array + pArray.missingBytesOffset, nMissingLocal - 1)
+            cb += out.writeByte((Region.loadByte(array + pArray.missingBytesOffset
               + (nMissingLocal - 1).toL) & EType.lowBitMask(prefixLen)).toB)
```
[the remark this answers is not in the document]
patrick-schultz: The issue is that the encoded representation stores the missingness bits in the minimum number of bytes, i.e. it pads out to a multiple of 8 bits, whereas the in-memory representation pads them more (with this PR, always to a multiple of 64 bits).

I'm not sure if by "creation time", you mean the creation of the encoded or in-memory representation. This line is creating the encoded representation, so there are no unspecified bits. I do want to enforce an invariant that all padding bits are set (missing) in the in-memory representation, but I haven't done that here. I could try, but I'm not sure how many places would need to change.

hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -98,24 +98,72 @@ final case class EArray(val elementType: EType, override val required: Boolean =
...
     val readElemF = elementType.buildInplaceDecoder(arrayType.elementType, cb.emb.ecb)
 
-    if (!elementType.required)
-      cb += in.readBytes(region, array + const(arrayType.lengthHeaderBytes), arrayType.nMissingBytes(len))
+    val pastLastOff = cb.memoize(arrayType.pastLastElementOffset(array, len))
+    if (elementType.required) {
+      val elemOff = cb.newLocal[Long]("elemOff", arrayType.firstElementOffset(array, len))
+      if (arrayType.trivialElements) {
+        // elements have 0 size, so all elements have the same address
+        // still need to read `len` elements from the input stream, as they may have non-zero size
```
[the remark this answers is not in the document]
patrick-schultz: Agree about the name, but I think we still need the comment here. It's not obvious that reading 0-size elements from an input stream has any effect.

hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -98,24 +98,72 @@ final case class EArray(val elementType: EType, override val required: Boolean =
...
+        val i = cb.newLocal[Int]("i", 0)
+        cb.forLoop({}, i < (len & -8), cb.assign(i, i + 8), {
```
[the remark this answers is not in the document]
patrick-schultz: I'd prefer a method. I'll just add `roundDownAlignment` next to `UnsafeUtils.roundUpAlignment`. Normally I'd agree about hex literals, but in this case it's much easier for me to understand "8 byte alignment" than "0xfffffff8 alignment".

hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -98,24 +98,72 @@ final case class EArray(val elementType: EType, override val required: Boolean =
...
+          for (k <- 0 until 8)
+            readElemF(cb, region, cb.memoize(arrayType.incrementElementOffset(elemOff, k)), in)
```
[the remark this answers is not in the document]
patrick-schultz: Whoops, copy paste error.

hail/src/main/scala/is/hail/types/encoded/EArray.scala
```diff
@@ -98,24 +98,72 @@ final case class EArray(val elementType: EType, override val required: Boolean =
...
-      else
-        cb.ifx(arrayType.isElementDefined(array, i),
-          readElemF(cb, region, elemAddr, in))
-    })
+      cb.ifx((len % 64).cne(0), {
+        // ensure that the last missing block has all missing bits set past the last element
+        val lastMissingBlockAddr = cb.memoize((arrayType.pastLastMissingByteAddr(array, len) - 1) & -8)
+        val lastMissingBlock = cb.memoize(Region.loadLong(lastMissingBlockAddr))
+        cb += Region.storeLong(lastMissingBlockAddr, lastMissingBlock | (const(-1L) << len))
+      })
```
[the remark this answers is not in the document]
patrick-schultz: This is creation of the in memory representation, which has more padding on the missingness bits than the encoded representation, so we need to set the padding bits here.

hail/src/main/scala/is/hail/types/physical/PCanonicalArray.scala
```diff
@@ -542,4 +583,22 @@ final case class PCanonicalArray(elementType: PType, required: Boolean = false)
     else
       PCanonicalArray(copiedElement, required)
   }
+
+  def forEachDefined(cb: EmitCodeBuilder, aoff: Value[Long])(f: (EmitCodeBuilder, Value[Int], SValue) => Unit) {
```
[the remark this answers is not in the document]
patrick-schultz: Absolutely! That was one reason I thought it worth changing the PCanonicalArray layout to not have any partial last block of missingness bits - so we can easily use blocked missingness logic elsewhere too.

hail/src/main/scala/is/hail/types/physical/stypes/interfaces/SContainer.scala
patrick-schultz: Yeah, no idea where that came from.

This is only used in the shuffler, and only as a heuristic, to try to balance resulting partitions by total size, not just number of rows. I agree it should be tested better, but since it's not trying to be exact, it's not obvious how to do that.

danking (approved):
awesome!

Issue #13567, opened 2023-09-06
[fs] hailtop.fs makes it impossible to explicitly configure the GCS requester pays project in code


danking:
### What happened?

See: https://hail.zulipchat.com/#narrow/stream/128581-Cloud-support/topic/requester.20pays.20in.20batch

We need a way to configure the RouterFS in a Python file. Maybe `hfs.init`? I'm not sure.

### Version

0.2.120

danking:
Where do I write requester pays information?

```python3
import hailtop.fs as hfs
with hfs.open('gs://bucket/path/to/file') as fobj:
    ....

```

danking:
Probably need an `hfs.init`?

danking:
Options:
1. `hfs.init`. Pros: applies to all uses of open/close/etc. Cons: global state.
2. `hfs.open(..., requester_pays_configuration=...)`. Pros: allows use-site overriding. Cons: user has to specify at every open.
3. both? Pros: easy way to set a default, and allows use-site overrides. Cons: committing ourselves to another global state.

Let's do option (2) for now.

PR #13795, opened 2023-10-11
[batch] adds requester pays config to `hailtop.fs.open`

iris-garden:
Closes https://github.com/hail-is/hail/issues/13567.


hail/python/hailtop/fs/fs_utils.py
```diff
@@ -31,6 +38,25 @@ def open(path: str, mode: str = 'r', buffer_size: int = 8192) -> io.IOBase:
...
+
+    Specify multiple Requester Pays Buckets within a project that are acceptable
+    to access:
+
+    >>> with hfs.open( # doctest: +SKIP
+    ...     'gs://my-bucket/notes.txt',
+    ...     requester_pays_config=('my-project', ['my-bucket', 'bucket-2'])
+    ... ) as f:
+    ...     for line in f:
+    ...         print(line.strip())
```
daniel-goldstein, line 58: I feel like this should only support `requester_pays_project`, since you can't `open` from more than one bucket.

hail/python/hailtop/utils/gcs_requester_pays.py
```diff
@@ -0,0 +1,25 @@
...
+
+def get_fs_by_requester_pays_config(
+    cls: Type[FS],
+    requester_pays_fses: Dict[Optional[str], FS],
+    requester_pays_config: Optional[GCSRequesterPaysConfiguration] = None
+) -> FS:
+    hashable_config = (
+        requester_pays_config
+        if requester_pays_config is None or isinstance(requester_pays_config, str)
+        else f"{requester_pays_config[0]}:{','.join(requester_pays_config[1])}"
```
daniel-goldstein, line 17: I think this will produce different keys for `("project", ["bucket1", "bucket2"])` vs `("project", ["bucket2", "bucket1"])`. If doing this interpolation is stemming from lists not being hashable, you can construct a `tuple(str, frozenset)` which should be hashable.

hail/python/hailtop/fs/fs_utils.py
daniel-goldstein: AFAICT the only method where it would be possible to specify multiple buckets would be `copy`. It's perhaps worth asking though if anyone would ever want to specify two different requester pays projects when copying between two buckets. Seems kind of weird and like it's leading to some more annoying bookkeeping than if we just had a `Dict[str, FS]`.

hail/python/hailtop/utils/gcs_requester_pays.py
```diff
@@ -0,0 +1,25 @@
...
+    )
+    try:
+        return requester_pays_fses[hashable_config]
+    except KeyError:
+        requester_pays_fses[hashable_config] = cls(
+            gcs_kwargs={"gcs_requester_pays_configuration": requester_pays_config}
+        )
```
daniel-goldstein, line 24: Maybe its the introduction of the hashable stuff, but the bookkeeping around the `requester_pays_fses` is starting to feel a little messy. I wonder if instead of keeping a `dict` around + using this utility method, we could instead [subclass defaultddict](https://stackoverflow.com/questions/43268306/can-defaultdict-accept-callables-dependent-on-the-given-missing-key)?

daniel-goldstein (dismissed):
Looking good, just a few organizational comments

hail/python/hailtop/fs/fs_utils.py
iris-garden: good point; checked in with @danking and he said it's not inconceivable that someone would want to put together one big requester pays config and just pass the same object into multiple invocations of `open`, so hopefully this latest change reduces the bookkeeping enough that it's not too big of an issue to support the `Tuple[str, List[str]]` case of the type

hail/python/hailtop/utils/gcs_requester_pays.py
iris-garden: ooo good catch, updated to use a `frozenset`

hail/python/hailtop/utils/gcs_requester_pays.py
iris-garden: i definitely like the approach of subclassing `DefaultDict` better haha, updated to do that

hail/python/hailtop/batch/backend.py
```diff
@@ -160,7 +157,12 @@ def __init__(self,
             flags += f' -v {gsa_key_file}:/gsa-key/key.json'
 
         self._extra_docker_run_flags = flags
-        self.__fs = RouterAsyncFS()
+        self.__requester_pays_fses = gcs_requester_pays_fs_cache(fs_constructor=RouterAsyncFS)
```
daniel-goldstein, line 160: Instead of having this field in both subclasses and requiring an abstract property, you can pass a `GCSRequesterPaysFSCache` to `super().__init__`

hail/python/hailtop/utils/gcs_requester_pays.py
```diff
@@ -0,0 +1,34 @@
...
+from hailtop.aiocloud.aiogoogle import GCSRequesterPaysConfiguration
+
+
+FS = TypeVar("FS")
+MaybeGCSRequesterPaysConfiguration = Optional[GCSRequesterPaysConfiguration]
+GCSRequesterPaysKey = Optional[Union[str, Tuple[str, FrozenSet[str]]]]
+
+
+class GCSRequesterPaysFSCache(DefaultDict[MaybeGCSRequesterPaysConfiguration, FS]):
+    def __init__(self, *args: Any, fs_constructor: Type[FS], default_kwargs: Dict[str, Any] = {}, **kwargs: Any) -> None:
```
daniel-goldstein, line 12: Is the only use the `*args` here the default factory in the `gcs_requester_pays_fs_cache` function? If so can we restrict to just that argument and add more as necessary down the line? Makes for nicer type checking.

daniel-goldstein: On a similar note, I don't see `**kwargs` used. Can we remove it until we need it?

hail/python/hailtop/utils/gcs_requester_pays.py
```diff
@@ -0,0 +1,34 @@
...
+        self._fs_constructor = fs_constructor
+        self[None] = fs_constructor(**default_kwargs)
```
daniel-goldstein, line 14: I feel like a cache shouldn't eagerly construct elements. Can we delete this and let the normal behavior deal with this case?

hail/python/test/hailtop/utils/test_gcs_requester_pays.py
```diff
@@ -0,0 +1,27 @@
...
+
+def test_get_fs_by_requester_pays_config():
+    def test_with_cls(cls, kwargs_field):
+        config_1 = "foo"
+        config_2 = ("foo", ["bar", "baz", "bat"])
+        kwargs_1 = {"gcs_requester_pays_configuration": config_1}
+        kwargs_2 = {"gcs_requester_pays_configuration": config_2}
+        fses = gcs_requester_pays_fs_cache(cls)
+        assert attrgetter(kwargs_field)(fses[None]) == {}
+        assert attrgetter(kwargs_field)(fses[config_1]) == kwargs_1
```
daniel-goldstein, line 18: Interesting, I've never seen `attrgetter` before. I think I would prefer though adding a `_gcs_kwargs` property to both `RouterFS` and `RouterAsyncFS` and accessing that property directly, so that if I mess something up the linters will tell me that these attr accesses will fail without me having to run this test.

hail/python/test/hailtop/utils/test_gcs_requester_pays.py
```diff
@@ -0,0 +1,27 @@
...
+        set_kwargs = attrgetter(kwargs_field)(fses[config_2])
+        assert set_kwargs["gcs_requester_pays_configuration"][0] == config_2[0]
+        assert len(set_kwargs["gcs_requester_pays_configuration"][1]) == len(config_2[1])
+        assert set(set_kwargs["gcs_requester_pays_configuration"][1]) == set(config_2[1])
+        default_kwargs_fses = gcs_requester_pays_fs_cache(cls, {"gcs_kwargs": kwargs_1})
+        assert attrgetter(kwargs_field)(default_kwargs_fses[None]) == kwargs_1
+
+    test_with_cls(RouterFS, "afs._gcs_kwargs")
+    test_with_cls(RouterAsyncFS, "_gcs_kwargs")
```
daniel-goldstein, line 27: I feel like we can instead make `test_with_cls` the actual test method and use `pytest.mark.parametrize` with the two classes.

daniel-goldstein (changes requested):
Sorry for the delay on this, this looks great! Just a couple more organizational nits.

hail/python/hailtop/batch/backend.py
iris-garden: good call, updated!

hail/python/hailtop/utils/gcs_requester_pays.py
iris-garden: removed!

hail/python/hailtop/utils/gcs_requester_pays.py
iris-garden: updated!

hail/python/test/hailtop/utils/test_gcs_requester_pays.py
iris-garden: added!

hail/python/test/hailtop/utils/test_gcs_requester_pays.py
iris-garden: updated!

hail/python/hailtop/batch/backend.py
```diff
@@ -53,24 +54,16 @@ class Backend(abc.ABC, Generic[RunningBatchType]):
     """
     _closed = False
 
-    def __init__(self):
-        self._requester_pays_fses: Dict[GCSRequesterPaysConfiguration, RouterAsyncFS] = {}
+    def __init__(self, requester_pays_fses: GCSRequesterPaysFSCache):
+        import nest_asyncio  # pylint: disable=import-outside-toplevel
+        nest_asyncio.apply()
```
daniel-goldstein, line 59: I think this `nest_asyncio` was deleted in main. Is this still necessary?

daniel-goldstein (changes requested):
One last thing. We should no longer need to apply nest_asyncio in this constructor.

hail/python/hailtop/batch/backend.py
iris-garden: ooo good catch, thank you!

PR #13796, opened 2023-10-11
[query] Dont compile code cache function three times

daniel-goldstein:
`f` is a thunk so it is currently being evaluated thrice before inserted into the code cache. The `compiledFunction` variable was unused so I think this is what was originally intended.


patrick-schultz (approved):
🤦 good catch

PR #13797, opened 2023-10-11
[query] Avoid py4j for python-backend interactions

### LINKED ISSUE #13756: [query] Don't use py4j for Backend operations
### What happened?

The spark and local backends use `py4j` to execute methods on java backends. `py4j` uses a TCP socket and a text-based protocol to communicate between python and the jvm and handles marshaling of data between the two processes. Unfortunately it has poor memory performance with large byte arrays, as the text protocol requires base64 encoding byte arrays and it uses Java `String`s which, being UTF-16, more than double the size of the original data in memory.

Hail should not use `py4j` for these operations and just open its own connection to the java backend. This gives us the control to not use more memory than is necessary to just ship bytes back and forth. This also provides an opportunity to deduplicate some code as the `ServiceBackend` already communicates writes its inputs over a socket instead of using `py4j` (there is no live JVM to communicate to in the `ServiceBackend` case, so it must serialize the requested operation to be run at a later time on a different machine).

### Version

0.2.124

### Relevant log output

_No response_

daniel-goldstein:
CHANGELOG: Fixes #13756: operations that collect large results such as `to_pandas` may require up to 3x less memory.

This turns all "actions", i.e. backend methods supported by QoB into HTTP endpoints on the spark and local backends. This intentionally avoids py4j because py4j was really designed to pass function names and references around and does not handle large payloads well (such as results from a `collect`). Specifically, py4j uses a text-based protocol on top of TCP that substantially inflates the memory requirement for communicating large byte arrays. On the Java side, py4j serializes every binary payload as a Base64-encoded `java.lang.String`, which between the Base64 encoding and `String`'s use of UTF-16 results in a memory footprint of the `String` being `4/3 * 2 = 8/3` nearly three times the size of the byte array on either side of the py4j pipe. py4j also appears to do an entire copy of this payload, which means nearly a 6x memory requirement for sending back bytes. Using our own socket means we can directly send back the response bytes to python without any of this overhead, even going so far as to encode results directly into the TCP output stream. Formalizing the API between python and java also allows us to reuse the same payload schema across all three backends.

### DIFF
--- a/hail/python/hail/backend/backend.py
+++ b/hail/python/hail/backend/backend.py
@@ -1,5 +1,7 @@
 from typing import Mapping, List, Union, TypeVar, Tuple, Dict, Optional, Any, AbstractSet
 import abc
+from enum import Enum
+from dataclasses import dataclass
 import warnings
 import orjson
 import pkg_resources
@@ -10,8 +12,12 @@ from hailtop.fs.fs import FS
 
 from ..builtin_references import BUILTIN_REFERENCE_RESOURCE_PATHS
 from ..expr import Expression
-from ..expr.types import HailType
-from ..ir import BaseIR
+from ..expr.table_type import ttable
+from ..expr.matrix_type import tmatrix
+from ..expr.blockmatrix_type import tblockmatrix
+from ..expr.types import HailType, dtype
+from ..ir import BaseIR, finalize_randomness
+from ..ir.renderer import CSERenderer
 from ..linalg.blockmatrix import BlockMatrix
 from ..matrixtable import MatrixTable
 from ..table import Table
@@ -70,6 +76,56 @@ def local_jar_information() -> LocalJarInformation:
         )
 
 
+class ActionTag(Enum):
+    LOAD_REFERENCES_FROM_DATASET = 1
+    VALUE_TYPE = 2
+    TABLE_TYPE = 3
+    MATRIX_TABLE_TYPE = 4
+    BLOCK_MATRIX_TYPE = 5
+    EXECUTE = 6
+    PARSE_VCF_METADATA = 7
+    IMPORT_FAM = 8
+    FROM_FASTA_FILE = 9
+
+@dataclass
+class ActionPayload:
+    pass
+
+@dataclass
+class IRTypePayload(ActionPayload):
+    ir: str
+
+@dataclass
+class ExecutePayload(ActionPayload):
+    ir: str
+    stream_codec: str
+
+@dataclass
+class LoadReferencesFromDatasetPayload(ActionPayload):
+    path: str
+
+@dataclass
+class ParseVCFMetadataPayload(ActionPayload):
+    path: str
+
+@dataclass
+class ImportFamPayload(ActionPayload):
+    path: str
+    quant_pheno: bool
+    delimiter: str
+    missing: str
+
+@dataclass
+class FromFASTAFilePayload(ActionPayload):
+    name: str
+    fasta_file: str
+    index_file: str
+    x_contigs: List[str]
+    y_contigs: List[str]
+    mt_contigs: List[str]
+    par: List[str]
+
+
 class Backend(abc.ABC):
     # Must match knownFlags in HailFeatureFlags.scala
     _flags_env_vars_and_defaults: Dict[str, Tuple[str, Optional[str]]] = {
@@ -115,33 +171,52 @@ class Backend(abc.ABC):
     def stop(self):
         pass
 
-    @abc.abstractmethod
     def execute(self, ir: BaseIR, timed: bool = False) -> Any:
-        pass
+        payload = ExecutePayload(self._render_ir(ir), '{"name":"StreamBufferSpec"}')
+        try:
+            result, timings = self._rpc(ActionTag.EXECUTE, payload)
+        except FatalError as e:
+            raise e.maybe_user_error(ir) from None
+        value = ir.typ._from_encoding(result)
+        return (value, timings) if timed else value
 
     @abc.abstractmethod
-    async def _async_execute(self, ir, timed=False):
+    def _rpc(self, action: ActionTag, payload: ActionPayload) -> Tuple[bytes, str]:
         pass
 
-    @abc.abstractmethod
+    def _render_ir(self, ir):
+        r = CSERenderer()
+        return r(finalize_randomness(ir))
+
     def value_type(self, ir):
-        pass
+        payload = IRTypePayload(ir=self._render_ir(ir))
+        type_bytes, _ = self._rpc(ActionTag.VALUE_TYPE, payload)
+        return dtype(type_bytes.decode('utf-8'))
 
-    @abc.abstractmethod
     def table_type(self, tir):
-        pass
+        payload = IRTypePayload(ir=self._render_ir(tir))
+        type_bytes, _ = self._rpc(ActionTag.TABLE_TYPE, payload)
+        return ttable._from_json(orjson.loads(type_bytes))
 
-    @abc.abstractmethod
     def matrix_type(self, mir):
-        pass
+        payload = IRTypePayload(ir=self._render_ir(mir))
+        type_bytes, _ = self._rpc(ActionTag.MATRIX_TABLE_TYPE, payload)
+        return tmatrix._from_json(orjson.loads(type_bytes))
 
-    @abc.abstractmethod
+    def blockmatrix_type(self, bmir):
+        payload = IRTypePayload(ir=self._render_ir(bmir))
+        type_bytes, _ = self._rpc(ActionTag.BLOCK_MATRIX_TYPE, payload)
+        return tblockmatrix._from_json(orjson.loads(type_bytes))
+
     def load_references_from_dataset(self, path):
-        pass
+        payload = LoadReferencesFromDatasetPayload(path=path)
+        references_json_bytes, _ = self._rpc(ActionTag.LOAD_REFERENCES_FROM_DATASET, payload)
+        return orjson.loads(references_json_bytes)
 
-    @abc.abstractmethod
     def from_fasta_file(self, name, fasta_file, index_file, x_contigs, y_contigs, mt_contigs, par):
-        pass
+        payload = FromFASTAFilePayload(name, fasta_file, index_file, x_contigs, y_contigs, mt_contigs, par)
+        rg_json_bytes, _ = self._rpc(ActionTag.FROM_FASTA_FILE, payload)
+        return orjson.loads(rg_json_bytes)
 
     def add_reference(self, rg):
         self._references[rg.name] = rg
@@ -184,9 +259,10 @@ class Backend(abc.ABC):
     def remove_liftover(self, name, dest_reference_genome):
         pass
 
-    @abc.abstractmethod
     def parse_vcf_metadata(self, path):
-        pass
+        payload = ParseVCFMetadataPayload(path)
+        metadata_json_bytes, _ = self._rpc(ActionTag.PARSE_VCF_METADATA, payload)
+        return orjson.loads(metadata_json_bytes)
 
     @property
     @abc.abstractmethod
@@ -198,9 +274,10 @@ class Backend(abc.ABC):
     def fs(self) -> FS:
         pass
 
-    @abc.abstractmethod
     def import_fam(self, path: str, quant_pheno: bool, delimiter: str, missing: str):
-        pass
+        payload = ImportFamPayload(path, quant_pheno, delimiter, missing)
+        fam_json_bytes, _ = self._rpc(ActionTag.IMPORT_FAM, payload)
+        return orjson.loads(fam_json_bytes)
 
     def persist(self, dataset: Dataset) -> Dataset:
         from hail.context import TemporaryFilename
@@ -242,7 +319,7 @@ class Backend(abc.ABC):
         }, **initial_flags)
 
     @abc.abstractmethod
-    def set_flags(self, **flags: Mapping[str, str]):
+    def set_flags(self, **flags: str):
         """Set Hail flags."""
         pass
 
--- a/hail/python/hail/backend/backend.py
+++ b/hail/python/hail/backend/backend.py
@@ -99,6 +99,7 @@ class IRTypePayload(ActionPayload):
 class ExecutePayload(ActionPayload):
     ir: str
     stream_codec: str
+    timed: bool
 
 @dataclass
 class LoadReferencesFromDatasetPayload(ActionPayload):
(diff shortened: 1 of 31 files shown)


danking:
I can start looking at this as soon as you give me the OK

daniel-goldstein:
@danking Looks like I'm still failing to configure a couple of settings related to references on the `ServiceBackend` but you can feel free to start looking. You'll notice that I made quite a substantial refactor in `ServiceBackend.scala` in an attempt to harmonize the scala backends a bit more. The rationale behind the refactor is I was having a hard time working with the various thunks passed around there. I saw them as a bit of poor-man's-object way to capture some state from the input file while keeping the `ServiceBackend` stateless. IMO there's no harm in keeping the `ServiceBackend` just as stateful as the other backends since it is single use. So I lifted a lot of that state into backend-creation time and created a harder delineation between which part of the input is for configuring the backend and which part is for the action being performed. This made it easier to reuse a couple of methods like `tableType` and such.

hail/python/hail/backend/backend.py
```diff
@@ -242,7 +320,7 @@ def _initialize_flags(self, initial_flags: Dict[str, str]) -> None:
         }, **initial_flags)
 
     @abc.abstractmethod
-    def set_flags(self, **flags: Mapping[str, str]):
+    def set_flags(self, **flags: str):
```
danking, line 323: good catch

hail/python/hail/backend/spark_backend.py
```diff
@@ -268,7 +273,18 @@ def hail_package(self):
     def utils_package_object(self):
         return self._utils_package_object
 
+    def _rpc(self, action, payload) -> Tuple[bytes, str]:
```
danking, line 276: This can live on `Py4JBackend`, right?

danking: I see that you depend on _backend_server. You could pass the `_jbackend` into `Py4JBackend`'s `__init__` and then you can construct `_backend_server` in `Py4JBackend`.

hail/python/hail/backend/py4j_backend.py
```diff
@@ -2,20 +2,20 @@
...
-from hail.expr.blockmatrix_type import tblockmatrix
-from hail.expr.matrix_type import tmatrix
-from hail.expr.table_type import ttable
-from hail.expr.types import dtype
 
-from .backend import Backend, fatal_error_from_java_error_triplet
+from .backend import ActionTag, Backend, fatal_error_from_java_error_triplet
+
+import http.client
+http.client._MAXLINE = 2 ** 20
```
danking, line 18: what's this about?

hail/python/hail/backend/service_backend.py
```diff
@@ -485,10 +430,10 @@ async def _rpc(self,
                     raise
 
             with timings.step("read output"):
-                result_bytes = await retry_transient_errors(self._read_output, ir, iodir + '/out', iodir + '/in')
-                return result_bytes, timings
+                result_bytes = await retry_transient_errors(self._read_output, iodir + '/out', iodir + '/in')
+                return result_bytes, str(timings.to_dict())
```
danking, line 434: why the `str(...to_dict())`?

hail/python/hail/backend/service_backend.py
```diff
@@ -537,127 +479,43 @@ def _cancel_on_ctrl_c(self, coro: Awaitable[T]) -> T:
...
-            inputs,
-            ir=ir,
-            progress=progress,
-            **kwargs
+    def _rpc(self, action: ActionTag, payload: ActionPayload) -> Tuple[bytes, str]:
+        return self._cancel_on_ctrl_c(self._async_rpc(action, payload))
+
+    async def _async_rpc(self, action: ActionTag, payload: ActionPayload):
+        if action == ActionTag.EXECUTE:
+            assert isinstance(payload, ExecutePayload)
```
danking, line 487: This kind of pattern makes me wonder if the tag should just hang off of the payload e.g. `.tag`?

hail/python/test/hail/backend/test_service_backend.py
```diff
@@ -52,14 +52,13 @@ def test_tiny_worker_has_tiny_memory():
...
+    backend.worker_memory = 'highmem'
     t = hl.utils.range_table(2, n_partitions=2).annotate(nd=hl.nd.ones((30_000, 30_000)))
     t = t.annotate(nd_sum=t.nd.sum())
     # We only eval the small thing so that we trigger an OOM on the worker
     # but not the driver or client
-    hl.eval(
-        t.aggregate(hl.agg.sum(t.nd_sum), _localize=False),
-        _execute_kwargs={'worker_cores': 8, 'worker_memory': 'highmem'}
-    )
+    hl.eval(t.aggregate(hl.agg.sum(t.nd_sum), _localize=False))
```
danking, line 61: This is a bit nicer actually.

hail/src/main/scala/is/hail/backend/Backend.scala
```diff
@@ -1,16 +1,120 @@
...
+case class ExecutePayload(ir: String, stream_codec: String, timed: Boolean)
+
+object BackendServer {
+  def apply(backend: Backend) = new BackendServer(backend)
+}
+
+class BackendServer(backend: Backend) {
+  // 0 => let the OS pick an available port
+  private val httpServer = HttpServer.create(new InetSocketAddress(0), 10)
+  private val handler = new BackendHttpHandler(backend)
```
danking, line 49: `private[this]` will get JVM fields, `private` is Scala horribleness.

hail/src/main/scala/is/hail/backend/Backend.scala
```diff
@@ -1,16 +1,120 @@
...
+    par: Array[String])
+case class ParseVCFMetadataPayload(path: String)
+case class ImportFamPayload(path: String, quant_pheno: Boolean, delimiter: String, missing: String)
+case class ExecutePayload(ir: String, stream_codec: String, timed: Boolean)
+
+object BackendServer {
+  def apply(backend: Backend) = new BackendServer(backend)
+}
+
+class BackendServer(backend: Backend) {
```
danking, line 46: I feel like this file is getting rather large; does it make sense to put the HTTP stuff in a different file?

hail/src/main/scala/is/hail/backend/Backend.scala
```diff
@@ -1,16 +1,120 @@
...
+        case "/fam/import" =>
+          val config = body.extract[ImportFamPayload]
+          backend.importFam(config.path, config.quant_pheno, config.delimiter, config.missing)
+      }
+
+      exchange.sendResponseHeaders(200, response.length)
+      using(exchange.getResponseBody())(_.write(response))
+    } catch {
+      case t: Throwable =>
+        error(t.getMessage)
```
danking, line 105: Is this a log statement? Can we use `log.error` instead?

hail/src/main/scala/is/hail/backend/spark/SparkBackend.scala
```diff
@@ -519,18 +531,25 @@ class SparkBackend(
           case Right((pt, addr)) => GetFieldByIdx(EncodedLiteral.fromPTypeAndAddress(pt, addr, ctx), 0)
         }
         log.info(s"finished execution of query $queryID")
-        literalIR
+        val id = UUID4().id
```
danking, line 534: UUIDs are kinda big and weird. Can we use a random long? Or an incrementing index?

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -545,10 +494,43 @@ private class HailSocketAPIOutputStream(
...
+case class SerializedIRFunction(
+  name: String,
+  type_parameters: Array[String],
+  value_parameter_names: Array[String],
+  value_parameter_types: Array[String],
+  return_type: String,
+  rendered_body: String,
+)
+
 class ServiceBackendSocketAPI2(
```
danking, line 531: We should give this a better name now. It's not a socket api at all, it's just an API I guess?

danking (changes requested):
A few thoughts; I'll have to do another closer read but this looks awesome so far. I'm really glad we moved away from my silly binary protocol to a JSON based one.

hail/python/hail/backend/spark_backend.py
daniel-goldstein: Good call. I took it several steps further and also lifted a bunch of other duplication into `Py4JBackend`. Should be limited to just commit `aa3ffcc` if you want to look at that separately.

hail/python/hail/backend/py4j_backend.py
daniel-goldstein: I left a comment explaining, and here's an [SO post](https://stackoverflow.com/questions/63157046/python-http-request-exception-linetoolong). I'm not really sure what to do here. It felt reasonable to send back timings in a response header instead of smushing it into the response body, as it is metadata, but I was left with a decision of how to raise the max header size and I went with something overly generous. AFAIK there is no actual limit in HTTP.

hail/python/hail/backend/service_backend.py
daniel-goldstein: Ya I kind of agree, except on the server I switch on the tag to determine how I should deserialize the payload, so it didn't make sense to me for that to be part of the payload.

hail/src/main/scala/is/hail/backend/Backend.scala
daniel-goldstein: Moved

hail/src/main/scala/is/hail/backend/Backend.scala
daniel-goldstein: This was debugging cruft, the error will get logged in python anyway. Removed

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
daniel-goldstein: Renamed to `ServiceBackendAPI`. I could be convinced to not have this class at all and just put it on the `object ServiceBackend`

hail/python/hail/backend/service_backend.py
danking: Hmm. I leave your comment up to you. I'm OK with this all as it is and don't have my head deep enough in this to have strong opinions about whether ir functions and tokens are in execute or outside.

hail/python/hail/backend/service_backend.py
danking: Hmm. Let's not touch in this PR. I don't really recall how timings works.

hail/python/hail/backend/py4j_backend.py
danking: I was going to say I was worried about proxies but I remember this entire connection is controlled by us. Yeah this seems fine.

hail/python/hail/backend/service_backend.py
daniel-goldstein: I think I'll leave as is now and we can revisit in a separate PR

hail/python/hail/backend/local_backend.py
```diff
@@ -219,22 +101,14 @@ def register_ir_function(self,
             return_type._parsable_string(),
             jbody)
 
-    def _is_registered_ir_function_name(self, name: str) -> bool:
-        return name in self._registered_ir_function_names
-
     def stop(self):
+        self._backend_server.stop()
```
danking, line 105: You need to stop this in the SparkBackend as well. I think it really should live on the Py4JBackend and both subclasses should call their super methods.

hail/python/hail/ir/ir.py
```diff
@@ -3705,23 +3705,36 @@ def _compute_type(self, env, agg_env, deep_typecheck):
...
+
+
 class JavaIR(IR):
-    def __init__(self, jir):
+    def __init__(self, hail_type, ir_id, ref: Optional[JavaIRSharedReference] = None):
         super(JavaIR, self).__init__()
-        self._jir = jir
-        super().__init__()
+        self._type: HailType = hail_type
+        self._id: int = ir_id
```
danking, line 3724: move the types to the parameters so that callers get type info

hail/src/main/scala/is/hail/backend/service/ServiceBackend.scala
```diff
@@ -305,19 +323,20 @@ class ServiceBackend(
...
       val off = ctx.scopedExecution((hcl, fs, htc, r) => f(hcl, fs, htc, r).apply(r))
       val codec = TypedCodecSpec(
-        EType.fromTypeAllOptional(retPType.virtualType),
-        retPType.virtualType,
+        EType.fromTypeAllOptional(elementType.virtualType),
+        elementType.virtualType,
         BufferSpec.parseOrDefault(bufferSpecString)
       )
-      codec.encode(ctx, retPType, off)
+      codec.encode(ctx, elementType, pt.loadField(off, 0))
```
danking, line 339: add an assert that the field is not missing

hail/src/main/scala/is/hail/HailContext.scala
```diff
@@ -124,7 +125,9 @@ object HailContext {
     info(s"Running Hail version ${ theContext.version }")
 
     // needs to be after `theContext` is set, since this creates broadcasts
-    backend.addDefaultReferences()
+    if (addDefaultReferences) {
+      backend.addDefaultReferences()
+    }
```
danking, line 130: I think you can get rid of this entirely and just let the backends always add the default references. ReferenceGenome use lazy val for their broadcasted values. I don't see why we would start a Backend without immediately starting a HailContext.

hail/src/main/scala/is/hail/backend/service/Worker.scala
```diff
@@ -158,11 +158,11 @@ object Worker {
     timer.start("executeFunction")
 
     if (HailContext.isInitialized) {
-      HailContext.get.backend = new ServiceBackend(null, null, new HailClassLoader(getClass().getClassLoader()), null, None)
+      HailContext.get.backend = new ServiceBackend(null, null, new HailClassLoader(getClass().getClassLoader()), null, None, null, null, null, null)
```
danking, line 161: This is feeling increasingly wrong. I think after this PR lands I'll PR something that creates a `QoBWorkerBackend` which is just the class loader and raises UnsupportedOperationException for everything else.

danking (dismissed):
minor things

hail/src/main/scala/is/hail/backend/service/Worker.scala
daniel-goldstein: that would be great

hail/python/hail/backend/spark_backend.py
```diff
@@ -146,13 +146,10 @@ def validate_file(self, uri: str) -> None:
         validate_file(uri, self._router_async_fs)
 
     def stop(self):
+        super().stop()
         self._jbackend.close()
```
danking, line 150: stopping the hail context calls this, so you can just remove it.

danking: But it's idempotent so you don't need to. I'll approve but let's nix it.

danking (approved):

PR #13799, opened 2023-10-11
[query] fix local backend for install-editable

danking:
I neglected to include the extra classpath necessary when using a skinny JAR.


danking:
See also: https://github.com/hail-is/hail/pull/13639

patrick-schultz (approved):

PR #13804, opened 2023-10-12
[hailtop.saige] Implement SAIGE in QoB

jigold:
@danking I have a mostly completed draft for SAIGE in QoB. Can you take a look? I'm mainly looking for enough feedback to get a green light to actually start testing this end to end, fill in the remaining not implemented components, add documentation, add verbosity and possibly a dry run feature, and support VEP annotations natively.

There are a couple of core concepts:
1. Phenotypes - Set of phenotypes to test. I support the ability to group phenotypes together. This is in anticipation of a new version of SAIGE that Wei is going to release soon.
2. VariantChunks - The set of variant intervals of data to test per job. If it's SAIGE-GENE, then there's also the "groups" to actually test within that interval.
3. io - There's a bunch of wrappers that handle input and output files so all of that logic combined with the checkpointing logic is abstracted away from what is actually going on.
4. steps - These are the SAIGE modules to run. They are all dataclasses with configuration options
5. saige - There's a class that can be instantiated in Python or I started writing the framework for a CLI. This has the code that builds the DAG end to end.

All configuration happens with a yaml file that can overwrite default parameters for each step such as whether to checkpoint or where the results should be written to. For the CLI, I envision you can either give a config file and/or specify `--overrides step1_null_glmm.use_checkpoint=true`. For every Saige run, I write out the configuration used to a file in the output directory as well as information about the input data and variant chunks and the batch information.


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.black]
 line-length = 120
 skip-string-normalization = true
-force-exclude = 'hail|datasets|sql|\.mypy'
+force-exclude = 'datasets|sql|\.mypy'
```
jigold, line 4: This is just so I can run black for now on my new code.

danking:
Very exciting!

danking:
Apologies for delay; I will look this afternoon.

hail/python/hailtop/saige/saige.py
```diff
@@ -0,0 +1,388 @@
...
+        covariates: List[str],
+        group_col: str,
+        output_dir: str,
+        b: Optional[hb.Batch] = None,
+        checkpoint_dir: Optional[str] = None,
+        variant_chunks: Optional[VariantChunks] = None,
+        run_kwargs: Optional[dict] = None,
+    ):
+        if checkpoint_dir is None:
+            checkpoint_dir = hl.TemporaryDirectory()
```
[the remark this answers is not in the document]
danking: It seems like the user should be forced to explicit provide a checkpoint directory if they want to create/restore-from checkpoints and otherwise we should just not use checkpoints.

hail/python/hailtop/saige/saige.py
```diff
@@ -0,0 +1,388 @@
...
+    variant_chunks: Ann[
+        Optional[str], Opt(help='Path to variant interval file in either bed or interval list format.')
+    ] = None,
+    overrides: Ann[
+        Optional[List[str]],
+        Opt(help='Override configuration parameters in the default configuration. Example: sparse_grm.checkpoint=true'),
+    ] = None,
+):
+    if config is not None:
+        config = SaigeConfig.from_yaml_file(config)
```
danking, line 360: I'm a bit skeptical of YAML files for configuration. Why is this easier than using Python to configure it?

I think the YAML would be:
```
version=1
name='saige'
attributes:
    foo: bar
    ...
prepare_inputs:
    ...
...
```
And the Python would be:
```
from ... import run_saige

config = SaigeConfig(
    version=1,
    name='Saige',
    attributes={
        'foo': 'bar',
        ...
    },
    prepare_inputs=...
)

run_saige(config)
```

In the YAML case, I don't have an IDE that can give me hints about types and parameter names. If I get a parameter name wrong in YAML, I probably don't find out until we parse it? Or maybe it just gets ignored?

In Python, the IDE gives me hints and can give me red squiggles exactly where there's an error. Also, if I want to do some kind of preprocessing, automation, etc., I can just use Python to do that.

hail/python/hailtop/saige/saige.py
```diff
@@ -0,0 +1,388 @@
...
+
+            with hfs.open(f'{output_dir}/batch_info.yaml', 'w') as f:
+                info = yaml.safe_dump(
+                    {
+                        'batch_id': b_handle.id,
+                        'name': b.name,
+                        'attributes': b.attributes,
+                        'timestamp': str(datetime.datetime.now()),
+                    }
+                )
```
danking, line 331: This kind of dump seems more appropriate for a log statement

hail/python/hailtop/saige/variant_chunk.py
```diff
@@ -0,0 +1,170 @@
...
+    require_row_key_variant(dataset, 'saige')
+
+    variants = dataset.rows()
+
+    if group is None:
+        return create_chunks_by_contig(variants, max_count_per_chunk, max_span_per_chunk)
+    return create_chunks_by_group(variants, group, max_count_per_chunk, max_span_per_chunk)
+
+
+class VariantChunks:
```
danking, line 143: This class is just a `List[VariantChunk]`, I think it's better to just pass around the List itself

hail/python/hailtop/saige/steps.py
```diff
@@ -0,0 +1,495 @@
...
+    --LOCO=FALSE 2>&1 | tee {stdout}
+'''
+        return command
+
+
+@dataclass
+class Step2SPAStep(CheckpointConfigMixin, JobConfigMixin):
+    base_name: str = '{{ phenotype["name"] }}-{{ chunk["idx"] }}'
+    save_stdout: bool = True
+    base_output_root: str = '{{ output_dir }}/results/{{ phenotype["name"] }}/{{ chunk["idx"] }}'
```
danking, line 228: I don't think we should use templates for this; these are just functions:
```
def base_name(phenotype: str, chunk_idx: int):
    return f'{phenotype}-{chunk_idx}'
```

hail/python/hailtop/saige/utils.py
```diff
@@ -0,0 +1,13 @@
...
+
+def cast_dataclass_attributes(dc):
+    for f in dataclasses.fields(dc):
+        value = getattr(dc, f.name)
+        if not isinstance(value, f.type):
+            if isinstance(f.type, bool):
+                if value in ('0', 'false', 'False'):
+                    value = False
+                elif value in ('1', 'true', 'True'):
+                    value = True
```
danking, line 12: Doesn't the YAML parser do this?

danking (changes requested):
Some high-level thoughts on structure. I think I'd generally bias towards just writing Python code rather than YAMLs, it feels like that adds complexity that doesn't justify itself with added value, but let me know if I'm off-base on that.

hail/python/hailtop/saige/saige.py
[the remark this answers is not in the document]
jigold: I think I conflated working directory with checkpoint directory here. If the user wants checkpointing, then we should do the "work" in the checkpoint directory. Else, we should just blow away the working directory when we're done. I can rename to make this clearer.

hail/python/hailtop/saige/saige.py
jigold: I think it depends on how we want `hailtop.saige` (or similar) to be used. Do we want to support a command line interface or does the user have to directly use the Python functions and write their own wrappers and use the Python config classes? If we do support a command line interface, think it's a pain to have to use a bunch of command line options especially with all of the different SAIGE steps having their own parameters so I thought having a configuration file that was easy to specify the configuration would be easier. I was also inspired by the machine learning library I've been playing with where the configuration is done this way since there's just so many parameters to specify. Maybe it's overkill though for this use case. I can think about it some more.

jigold:
@danking Can you take another look? The only thing I didn't address is the `Phenotypes -> List[Phenotype], VariantChunks -> List[VariantChunk]`. I don't want to rip it out yet in case Wei comes out with the new SAIGE implementation with phenotype groupings. Again, just looking for a green light to start testing it. We can figure out the question about the CLI later on.

hail/python/hailtop/saige/io.py
```diff
@@ -0,0 +1,307 @@
...
+
+class SaigeSparseGRMResourceGroup(BaseSaigeResource):
+    @staticmethod
+    def inputs(root: str) -> ResourceGroupInputs:
+        return ResourceGroupInputs('grm', grm=f'{root}.sparseGRM.mtx', sample_ids=f'{root}.sparseGRM.mtx.sampleIDs.txt')
+
+    @staticmethod
+    async def use_checkpoint_if_exists_and_requested(
+        fs: AsyncFS, pool: AsyncWorkerPool, b: hb.Batch, config: CheckpointConfigMixin, grm_root: str
+    ) -> Optional['SaigeSparseGRMResourceGroup']:
```
jigold, line 166: Actually, I can probably consolidate this even more with more classmethods. Will do that in the next round.

jigold:
Just realized that I didn't entirely make this parallel enough yet. Will do tomorrow.

hail/python/hailtop/saige/saige.py
danking: I agree a CLI is a nice way to invoke a simple tool. I'd rather to `cp x y` than `python3 -c 'import cp; cp("x", "y")'. However, once the interface is more complex than a CLI, the choice is a configuration file or general purpose programming language. If the choice is between those two, my preference is for the latter, but maybe I'm missing some clear win of the former. To be really specific it feels like the cli+config is:

```
./cli-tool --config config.yaml --maybe-another-option
```

And the python script is:

```
python3 config.py
```

Admittedly, if you want to pass extra command line options to config.py you have to use argparse or  `sys.argv`.

hail/python/hailtop/saige/saige.py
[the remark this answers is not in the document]
danking: Are you thinking about the LocalBackend case? I don't understand what you mean about doing work. Intermediate files should get cleaned up by hailtop.batch's infrastructure. And it seems the rest of your infrastructure here avoids `write_output` if you're not checkpointing. What would we need to blow away? Also, what does a working directory mean when you're in the cloud?

hail/python/hailtop/saige/steps.py
```diff
@@ -0,0 +1,530 @@
...
+        b: hb.Batch,
+        mt: hl.MatrixTable,
+        phenotypes: Phenotypes,
+        working_dir: str,
+    ) -> Tuple[TextFile, PlinkResourceGroup]:
+        phenotypes_output = await self._write_phenotype_data(
+            fs, pool, b, mt, f'{working_dir}/inputs/phenotypes.tsv', phenotypes.phenotype_names
+        )
+        input_data = await self._convert_mt_to_plink(fs, pool, b, mt, f'{working_dir}/inputs/input_data')
+        return (phenotypes_output, input_data)
```
danking, line 65: I still feel like there are more classes and lines of code than we need to express the logic here. You're ultimately trying to get some `hailtop.batch.Resource` objects that you can use with a `job.command`, right? What do things look like if you go in this kind of direction:

```python3
async def write_plink(
    fs: AsyncFS,
    cfg: CheckpointConfigMixin,
    b: Batch,
    mt: MatrixTable,
    fname: str
) -> Resource:
    files = dict(bed=f'{fname}.bed', bim=f'{fname}.bim', fam=f'{fname}.fam')
    if cfg.overwrite or not all(await asyncio.gather(fs.exists(x) for x in files.values())):
        hl.export_plink(mt, fname)
    return b.read_input_group(**files)


async def export_tsv(
    fs: AsyncFS,
    cfg: CheckpointConfigMixin,
    b: Batch,
    ht: Table,
    fname: str
) -> Resource:
    if cfg.overwrite or not await fs.exists(fname):
        ht.export(fname, delimiter='\t')
    return b.read_input(fname)


async def prepare_inputs(
    fs: AsyncFS,
    cfg: CheckpointConfigMixin,
    b: hb.Batch,
    mt: hl.MatrixTable,
    working_dir: str,
    pheno_cols: List[str]
):
    pheno_file = f'{working_dir}/inputs/phenotypes.tsv'
    plink_file = f'{working_dir}/inputs/input_data'

    ht = mt.cols().select(pheno_cols)

    return asyncio.gather(
        export_tsv(fs, cfg, b, ht, pheno_file),
        write_plink(fs, cfg, b, mt, plink_file)
    )
```

In fact, restructuring this way reveals that we probably should have a class that contains an FS, a CheckpointConfig, and a Batch. That would save us a couple lines of code and some noise every time we call a function.

danking (changes requested):

hail/python/hailtop/saige/steps.py
jigold: Again, I can't quite tell what you are advocating for. Is the issue that you don't like the `PrepareInputs` step class? My intention was to make this as customizable as possible. Each "Step" in the pipeline is a class. This class must inherit from the CheckpointConfigMixin. The config for each step is present in the SaigeConfig. If a user wanted to prepare their inputs differently with custom matrix table filter functions etc., then they can inherit from the `PrepareInputs` class and override the appropriate function call (`call`). Same for all of the other steps.

hail/python/hailtop/saige/io.py
```diff
@@ -0,0 +1,307 @@
...
+        inputs = VCFResourceGroup.inputs(root)
+        return VCFResourceGroup._from_input_files(b, inputs)
+
+    @staticmethod
+    def from_job_intermediate(j: hb.Job) -> 'VCFResourceGroup':
+        inputs = VCFResourceGroup.inputs('{root}')
+        return VCFResourceGroup._from_job_intermediate(j, inputs)
+
+
+class BgenResourceGroup(BaseSaigeResource):
```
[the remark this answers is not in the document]
jigold: Actually, you can't do this. There's a chicken egg problem where we have to create a new job for this function hence defeating the purpose of using the checkpointed files and not running this job.

danking (changes requested):
I think we covered the comments in our conversation; if you want me too look again, dismiss!

jigold:
I don't think I addressed all of your comments, but I'd like you to take another look. I rewrote the `io.py` component and fixed the parallelism and context managers. I also figure out where to write files to based on the checkpoint_dir argument. We need to `results.checkpoint()` at the end because I decided to treat the raw SAIGE output as temp files (or checkpoints) and the results as a hail table is the thing people want to save to an output directory. I can make the output directory optional as well and not checkpoint that. The issue is that if we write the raw SAIGE results to a temp dir, when the context manager exits, the results HailTable won't be able to read the data once it's returned back to the user.

jigold:
I did not fix the variant chunks or phenotypes classes to not be actual classes. But I would like your feedback on the rewritten `io.py`. It still might be overkill, but getting closer.

danking:
This all seems fine to me now. I'm skeptical the new types will catch non-trivial bugs, but they're also not syntactically obtrusive.

I'm strongly opposed to unnecessary writing and reading of data. In much the same way that `import_table` doesn't immediately write to a table and then read that, I don't think we should be writing the results. If you really feel strongly that the results should be saved to a file, then do not return a table/matrix-table. Just accept a file path and write to that file path.

danking (changes requested):
dismiss if I should look again

hail/python/hailtop/saige/steps.py
```diff
@@ -0,0 +1,610 @@
...
+cat > compile_results.py <<EOF
+import hail as hl
+mt = hl.read_matrix_table("{mt_path}")
+reference_genome=mt.locus.dtype.reference_genome
+ht = hl.import_table("{results_path}", impute=True)
+ht = ht.annotate(locus=hl.locus(hl.str(ht.CHR), ht.POS, reference_genome=reference_genome), alleles=hl.array([ht.Allele1, ht.Allele2]))
+ht = ht.key_by(ht.locus, ht.alleles, ht.phenotype)
+ht = ht.drop('CHR', 'POS', 'Allele1', 'Allele2')
+ht.write("{results_ht}")
+ht.describe()
```
jigold, line 579: note to self to remove this line.

jigold:
For reference, Wei's documentation for SAIGE is here: https://saigegit.github.io/SAIGE-doc/
The code is here: https://github.com/saigegit/SAIGE
Example files are here: https://github.com/saigegit/SAIGE/tree/main/extdata/input
Dockerhub image is here: https://hub.docker.com/r/wzhou88/saige/tags

PR #13807, opened 2023-10-12
[query] replace BlockMatrixStage with BlockMatrixStage2 in lowerToTableStage (O(N*M) IR => O(1) IR).

danking:


danking:
@patrick-schultz I wrote this up when debugging some other stuff.

With this, my other stuff encounters hard to debug array index issues. That makes me a bit apprehensive to merge it b/c I can't tell if those issues are in my other stuff or real BM issues. 

My specific question: how comfortable are you with the tests of BlockMatrix. Do you think it's well tested enough to just trust this PR is correct?

patrick-schultz:
Between `BlockMatrixIRSuite.scala`, `test_linalg.py`, and the methods that use BlockMatrix, I think the BlockMatrix lowering is decently tested. And this change is certainly one we have to make. I can give it a careful review to look for potential issues.

patrick-schultz:
Actually, this does appear to be a gap in our test coverage. I put in a failing assertion, and both `BlockMatrixIRSuite.scala` and `test_linalg.py` passed.

I didn't notice any obvious errors in your code, but I don't think this is the right way to do it, so I tried writing it myself. I'll try to write a test that exercises this path and test both implementations.

patrick-schultz:
#13839 is I think the right way to implement this. It passes `test_block_matrix_entries`, which is the only test I could find that exercises this path. Others that should, like `test_to_table`, can't yet be fully lowered.

Maybe try rebasing on my branch and see if you still see the indexing errors?

danking:
I still have the index error, so it must be inherent to my branch. Let's close and go with your implementation instead.

PR #13810, opened 2023-10-13
[batch] Rename batches tables to job groups

jigold:
Stacked on #13475. This PR renames the following tables to have job groups in the name instead of batches. Note that this PR needs to shutdown the batch deployment (offline migration). I'm not 100% sure this is necessary, but I want to avoid a case where MJC of the database migration job cannot happen thus deadlocking the system.

```sql
RENAME TABLE batch_attributes TO job_group_attributes,
             batches_cancelled TO job_groups_cancelled,
             batches_inst_coll_staging TO job_groups_inst_coll_staging, 
             batch_inst_coll_cancellable_resources TO job_group_inst_coll_cancellable_resources, 
             aggregated_batch_resources_v2 TO aggregated_job_group_resources_v2,
             aggregated_batch_resources_v3 TO aggregated_job_group_resources_v3, 
             batches_n_jobs_in_complete_states TO job_groups_n_jobs_in_complete_states;
```


jigold:
Closing for now as there's too many merge conflicts with existing PRs.

jigold:
@daniel-goldstein FYI: #13893 is going to conflict with this one. Let me know how you'd like to proceed. Should I stack one on top of the other now? Or which ever one goes in first, I'll have to fix the rebase conflicts.

daniel-goldstein (approved):
Everything checks out to me, feels like we're so close!

jigold:
I looked at this again and I think we can do this with `online: true`. It's a quick enough migration where it shouldn't impact the driver for too long that it's trying to query the long tables. If there's a problem and the driver can't make forward progress once the database migration is done, we can just shut down the deployment to 0 replicas. Thoughts?

daniel-goldstein:
Doesn't the batch driver need to schedule the `deploy_batch` job to update itself? How can it eventually get to deploying the new / compatible code?

jigold:
I think we need it to be offline unless we're willing to tolerate up to 5-10 mins of not being able to cancel a batch and some alerts. The only parts that would be referencing the wrong tables are in the `Canceller` and `notify_job_complete`. I think scheduling and MJC would just work because we update those stored procedures and don't change the child code.

We can shut batch down though for the migration. Seems safest although more of a pain.

daniel-goldstein:
> need it to be offline unless we're willing to tolerate up to 5-10 mins of not being able to cancel a batch and some alerts

I want as much as possible for alerts to mean something unexpected is going wrong. If we can I think this should just be done offline.

PR #13812, opened 2023-10-13
[hailctl] Fix uploading files from parent directories of cwd

### LINKED ISSUE #13785: hailctl batch submit tries to copy bad directory files
### What happened?

We tried to copy and symlink the directory file `..` when trying to use the `--files` argument with a directory.

### Version

0.2.123

### Relevant log output

```shell
"input_files": [
      {
        "from": "gs://my_bucket//jwTsgaP9Q7zbQJVgO20Z84/sample_qc.py",
        "to": "/io/batch/7d4f36/inputs/xzMZj/sample_qc.py"
      },
      {
        "from": "gs://my_bucket//jwTsgaP9Q7zbQJVgO20Z84/..",
        "to": "/io/batch/7d4f36/inputs/bukrz/.."
      },
      {
        "from": "gs://my_bucket//jwTsgaP9Q7zbQJVgO20Z84//Users/me/.config/hail/config.ini",
        "to": "/io/batch/7d4f36/inputs/Rjsy0/config.ini"
      }
    ],
```

```
ln: failed to create symbolic link '../..': File exists
```
```

### PR DESCRIPTION
**Pull request by jigold**
The bug in #13785 was that we used a `relpath`, but the files that were to be included was the parent directory of the current working directory of the script. To solve this, I added a new option `--mounts` that uses the Docker syntax for `-v src:dest` to specify where the contents of the source should be mounted into the container.

Fixes #13785

### DIFF
--- a/hail/python/hailtop/hailctl/batch/cli.py
+++ b/hail/python/hailtop/hailctl/batch/cli.py
@@ -4,7 +4,7 @@ import typer
 from typer import Option as Opt, Argument as Arg
 import json
 
-from typing import Optional, List, Annotated as Ann, cast, Dict, Any, Tuple
+from typing import Optional, List, Annotated as Ann, cast, Dict, Any
 
 from . import list_batches
 from . import billing
--- a/hail/python/hailtop/hailctl/batch/cli.py
+++ b/hail/python/hailtop/hailctl/batch/cli.py
@@ -165,9 +165,6 @@ def submit(
     files: Ann[
         Optional[List[str]], Opt(help='Files or directories to add to the working directory of the job.')
     ] = None,
-    mounts: Ann[
-        Optional[List[str]], Opt(help='Files or directories to add to a specific mount location in the job. Equivalent to -v src:dest in Docker.')
-    ] = None,
     name: Ann[str, Opt(help='The name of the batch.')] = '',
     image_name: Ann[Optional[str], Opt(help='Name of Docker image for the job (default: hailgenetics/hail)')] = None,
     output: StructuredFormatPlusTextOption = StructuredFormatPlusText.TEXT,
@@ -180,7 +177,7 @@ def submit(
 
     $ hailctl batch submit --image-name docker.io/image my_script.py -- some-argument --animal dog
     '''
-    asyncio.run(_submit.submit(name, image_name, files or [], mounts or [], output, script, [*(arguments or []), *ctx.args]))
+    asyncio.run(_submit.submit(name, image_name, files or [], output, script, [*(arguments or []), *ctx.args]))
 
 
 @app.command('init', help='Initialize a Hail Batch environment.')
--- a/hail/python/hailtop/hailctl/batch/cli.py
+++ b/hail/python/hailtop/hailctl/batch/cli.py
@@ -4,7 +4,7 @@ import typer
 from typer import Option as Opt, Argument as Arg
 import json
 
-from typing import Optional, List, Annotated as Ann, cast, Dict, Any
+from typing import Optional, List, Annotated as Ann, cast, Dict, Any, Tuple
 
 from . import list_batches
 from . import billing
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by jigold** [comment]
I think we might need a different name than "mount". In Docker, you can only use the same destination mount once. i.e. you can't have this. `--mount file1:/ --mount file2:/`. 

@danking Do you have any thoughts on the interface for this?

**Comment by jigold** [comment]
Fixes https://github.com/hail-is/hail/issues/13785

File: hail/python/hailtop/hailctl/batch/cli.py [hunk 1]
```diff
@@ -164,6 +164,9 @@ def submit(
     files: Ann[
         Optional[List[str]], Opt(help='Files or directories to add to the working directory of the job.')
     ] = None,
+    mounts: Ann[
+        Optional[List[str]], Opt(help='Files or directories to add to a specific mount location in the job. Equivalent to -v src:dest in Docker.')
+    ] = None,
```
danking, line 169: I don't like the term `mount` b/c this isn't mounting the users hard drive into the cloud.

I think it's a bit confusing that there are two arguments with nearly identical descriptions.

How about this:
1. Replace `--file` with the functionality provided by `--mount`
2. Allow the new `--file` to elide the destination. When the destination is elided, it is assumed to be the same as the source (so: `..` means `..:..` both relative to the current working directory)
3. By default, make the working directory of the job match the working directory of the user's laptop. 

These three things together allow `--file .. --file ../../reference_data:/reference_data --file ../../other_project/thing.py` to work as expected. You'll get a container with this file structure:
```
/Users
  /dking
    /projects
      /foo
        thing.py
        ... everything else in the working directory of this script on the user's laptop ...
/reference_data
   ... everything in /Users/dking/projects/reference_data on the user's laptop ...
```

NB: if you do `--file .. --file foo.py` you're implicitly copying foo.py twice. Unnecessary, but should be fine if foo.py is not mutated while we are submitting.

danking (changes requested):

jigold:
bump. Test failure was unrelated.

iris-garden (changes requested):
lgtm! happy to hit approve once the conflicts are fixed

jigold:
I rebased this @iris-garden

iris-garden (approved):

danking:
Supporting `file://` seems to suggest that we would support URIs in general (e.g. `--files gs://foo/bar:/baz`); AFAICT we don't support the latter, right? What's the motivation to support `file://`?

jigold:
> Supporting `file://` seems to suggest that we would support URIs in general (e.g. `--files gs://foo/bar:/baz`); AFAICT we don't support the latter, right? What's the motivation to support `file://`?

I didn't know what we currently support with regards to file paths, so I just made it that we're resilient to someone adding file://

jigold:
I'm not sure what to do about the invalid block id transient error...

jigold:
This took me an incredibly long time to figure out, but I'm pretty sure the issue is that my last test copies the same file twice. Once explicitly as a single file and the second time as part of the directory to copy recursively. I don't think our copier code interacts correctly with Azure Blob Storage in this case and we end up with invalid block ID list errors.

I'm still stewing over which option is the best to address this. I could either not have `submit.py` ask to copy the same file twice or I can modify the copier to have a lock on the create blob writer stream so that we don't have this issue. However, I'm worried about deadlocks with our extensive use of semaphores and parallelism. Thoughts?? I think the right thing to do is fix it for now in `submit.py` so we can get this in and then make an issue to make a more durable fix to the copier.

danking:
@jigold https://azure.github.io/Storage/docs/application-and-user-data/code-samples/concurrent-uploads-with-versioning

Looks like this is the expected behavior from Azure Blob Storage when concurrently uploading blobs. It seems like *all* blocks are purged when any single upload succeeds. Unfortunately, it doesn't seem possible to distinguish between a truly invalid block list and this transiently invalid block list. I dislike this interface, but it is what we have.

It seems to me that the right fix is to deduplicate the file list before uploading. Multiple files uploading to the same target should be an error if they're different source files.

jigold:
@iris-garden Can you review again when you get a chance? Tests are all passing now

PR #13817, opened 2023-10-16
[hailtop] yet another transient error

danking:
CHANGELOG: Mitigate new transient error from Google Cloud Storage which manifests as `aiohttp.client_exceptions.ClientOSError: [Errno 1] [SSL: SSLV3_ALERT_BAD_RECORD_MAC] sslv3 alert bad record mac (_ssl.c:2548)`.

As of around 1500 ET 2023-10-16, this exception happens whenever we issue a lot of requests to GCS.

See [Zulip thread](https://hail.zulipchat.com/#narrow/stream/300487-Hail-Batch-Dev/topic/cluster.20size/near/396777320).

### DIFF
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -687,6 +687,14 @@ def is_transient_error(e):
     if (isinstance(e, aiohttp.ClientPayloadError)
             and e.args[0] == "Response payload is not completed"):
         return True
+    if (isinstance(e, aiohttp.ClientOSError)
+            and len(e.args) >= 2
+            and e.args[0] == 1
+            and 'sslv3 alert bad record mac' in e.args[1]):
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/utils/utils.py
```diff
@@ -687,6 +687,14 @@ def is_transient_error(e):
     if (isinstance(e, aiohttp.ClientPayloadError)
             and e.args[0] == "Response payload is not completed"):
         return True
+    if (isinstance(e, aiohttp.ClientOSError)
+            and len(e.args) >= 2
+            and e.args[0] == 1
```
jigold, line 692: This is hard to parse. Should it be `e.args[0] == '1'` as is `e.args[0]` a string or int? Can we just check for the message in `e.args[1]`?

jigold (changes requested):

PR #13818, opened 2023-10-16
[hailtop] types, new exceptions, better task exception handling

danking:
Besides the types and some transient exceptions, I think I fixed our task exception handling in several spots. Two things:

1. We do not need to wait on a cancelled task. If it was not done, then it could not possibly have an exception to retrieve. Moreover, now that it is cancelled, there is nothing else to do. Cancellation is immediate.

2. If a task is done, we *must* always retrieve the exception, otherwise we might not see an exception.

### DIFF
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -17,7 +17,6 @@ import asyncio
 import aiohttp
 import urllib.parse
 import urllib3.exceptions
-import google.api_core.exceptions
 import secrets
 import socket
 import requests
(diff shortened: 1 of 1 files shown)


jigold (dismissed):

hail/python/hailtop/utils/utils.py
```diff
@@ -678,9 +690,16 @@ def is_transient_error(e):
     if isinstance(e, asyncio.TimeoutError):
         return True
     if (isinstance(e, aiohttp.ClientConnectorError)
-            and hasattr(e, 'os_error')
             and is_transient_error(e.os_error)):
         return True
+    if (isinstance(e, aiohttp.ClientOSError)
+            and len(e.args) >= 2
+            and e.args[0] == 1
```
jigold, line 697: I thought the change that added this already went in. Must have been in a different location?

jigold (changes requested):

hail/python/hailtop/utils/utils.py
danking: You are correct, good catch. I screwed up a rebase. Removed. Originally added here: https://github.com/hail-is/hail/pull/13817/files

hail/python/hailtop/utils/utils.py
```diff
@@ -681,7 +693,6 @@ def is_transient_error(e):
     if isinstance(e, asyncio.TimeoutError):
         return True
     if (isinstance(e, aiohttp.ClientConnectorError)
-            and hasattr(e, 'os_error')
```
danking, line 684: ClientConnectorError always has an os_error attribute: https://github.com/aio-libs/aiohttp/blob/a7bc5e9eeae7c5c90898411962e9a74bf10a9cef/aiohttp/client_exceptions.py#L124-L139

jigold (approved):

PR #13820, opened 2023-10-16
[copier] minor clean up to copier

danking:
1. File rate is more interesting for small files.
2. The source_report controls the progress bar. By updating it eagerly while we are listing a directory, the progress bar is more accurate sooner. We currently wait until we get a semaphore for a particular file to update the progress bar.

### DIFF
--- a/hail/python/hailtop/aiotools/fs/copier.py
+++ b/hail/python/hailtop/aiotools/fs/copier.py
@@ -405,7 +405,10 @@ class SourceCopier:
             try:
                 copy_thunks = []
                 async for srcentry in srcentries:
-                    bytes_to_copy += await (await srcentry.status()).size()  # this is almost never a syscall/net-request (afaik: only local symlinks)
+                    # In cloud FSes, status and size never make a network request. In local FS, they
+                    # can make system calls on symlinks. This line will be fairly expensive if
+                    # copying a tree with a lot of symlinks.
+                    bytes_to_copy += await (await srcentry.status()).size()
                     copy_thunks.append(functools.partial(copy_source, srcentry))
                 return (copy_thunks, bytes_to_copy)
             finally:
--- a/hail/python/hailtop/aiotools/fs/copier.py
+++ b/hail/python/hailtop/aiotools/fs/copier.py
@@ -1,4 +1,4 @@
-from typing import Any, AsyncIterator, Awaitable, Optional, List, Union, Dict, Callable
+from typing import Any, AsyncIterator, Awaitable, Optional, List, Union, Dict, Callable, Tuple
 import os
 import os.path
 import asyncio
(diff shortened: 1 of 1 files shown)


hail/python/hailtop/aiotools/fs/copier.py
```diff
@@ -163,7 +163,10 @@ def add_source_reports(transfer_report):
         print(f'  Time: {humanize_timedelta_msecs(self._duration)}')
         assert self._duration is not None
         if self._duration > 0:
-            print(f'  Average transfer rate: {humanize.naturalsize(total_bytes / (self._duration / 1000))}/s')
+            bandwidth = humanize.naturalsize(total_bytes / (self._duration / 1000))
+            print(f'  Average bandwidth: {bandwidth}/s')
+            file_rate = total_files / (self._duration / 1000)
+            print(f'  Average file rate: {file_rate:,.1f}/s')
```
jigold, line 169:
```suggestion
            print(f'  Average file rate: {file_rate:.1f}/s')
```

hail/python/hailtop/aiotools/fs/copier.py
```diff
@@ -394,19 +397,23 @@ async def copy_source(srcentry: FileListEntry) -> None:
...
+            bytes_to_copy = 0
             if srcentries is None:
                 srcentries = await files_iterator()
             try:
-                return [
-                    functools.partial(copy_source, srcentry)
-                    async for srcentry in srcentries]
+                copy_thunks = []
+                async for srcentry in srcentries:
+                    bytes_to_copy += await (await srcentry.status()).size()  # this is almost never a syscall/net-request (afaik: only local symlinks)
```
jigold, line 408: I don't understand the end of this comment. What do you mean this does a symlink?

jigold (changes requested):

hail/python/hailtop/aiotools/fs/copier.py
danking: Getting the status and the size of an entry of a directory listing, AFAICT, only requires a system call or network request in one special case:

1. You have a LocalFS
2. The file in question is a symlink.

Linux doesn't eagerly retrieve the size of the target of a symlink (I think because it could be an arbitrary number of symlinks away).

danking: That's why status and size are `async`. If it wasn't for the local fs, they could be sync.

Maybe a better comment is:
> In cloud FSes, status and size never make a network request. In local FS, they can make system calls on symlinks. This line will be fairly expensive if copying a tree with a lot of symlinks.

hail/python/hailtop/aiotools/fs/copier.py
jigold: Ah. Thought it was a typo. Should have dug into the docs for formatting floats more carefully.

hail/python/hailtop/aiotools/fs/copier.py
jigold: That's a better comment.

danking: Updated, thanks!

jigold (approved):

PR #13821, opened 2023-10-16
[fs] delete in batches, accept stdin

danking:
Some quality-of-life stuff for `hailtop.aiotools.delete`.

Without deleting in batches, I found it impossible to delete very large lists of files because we create too many asyncio tasks.

### DIFF
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -21,6 +21,7 @@ import secrets
 import socket
 import requests
 import botocore.exceptions
+import itertools
 import time
 from requests.adapters import HTTPAdapter
 from urllib3.poolmanager import PoolManager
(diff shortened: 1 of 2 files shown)


hail/python/hailtop/utils/utils.py
```diff
@@ -111,12 +112,14 @@ def digits_needed(i: int) -> int:
     return 1 + digits_needed(i // 10)
 
 
-def grouped(n: int, ls: List[T]) -> Iterable[List[T]]:
+def grouped(n: int, ls: Iterable[T]) -> Iterable[List[T]]:
```
daniel-goldstein, line 115: Based on the above usage, shouldn't this be `Iterator`?

daniel-goldstein, line 115: Could you leave a comment to replace this with [itertools.batched](https://docs.python.org/3/library/itertools.html#itertools.batched) when we (one day hopefully) hit 3.12?

daniel-goldstein (changes requested):

hail/python/hailtop/utils/utils.py
danking: We use `grouped` in other places where we pass a `List`. Those would be type failures if this was `Iterator[T]`.

danking: done

daniel-goldstein (approved):

PR #13822, opened 2023-10-16
[ci] Bump number of cores to 16 for db deployment jobs

jigold:
I wanted to be safe rather than sorry for #13487 to make sure we have enough memory to fit all of the 100 row chunk endpoints in memory. We can switch it back to 4 cores after that PR is merged.

### DIFF
--- a/ci/ci/build.py
+++ b/ci/ci/build.py
@@ -1328,7 +1328,7 @@ EOF
                 regions=[REGION],
             )
 
-        n_cores = 4 if scope == 'deploy' and not is_test_deployment else 1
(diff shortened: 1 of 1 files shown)


daniel-goldstein (approved):

danking:
Let's make this a config parameter in the future so we can specify which migrations need more cores. This also makes it self-documenting for our future selves.

jigold:
This is actually challenging since all migrations run in the same job. You need the max of requested cores across all migrations.

jigold:
FWIW, the intervals ended up being 6.2GB on GCP. So I gave it 10x more than needed. Based on that info, we can reassess the best parameter and maybe just keep it at 4.

PR #13829, opened 2023-10-16
[vds/combiner] use get_lgt for PGT handling as well

### LINKED ISSUE #13828: [query] VDS combiner converts PGT to LPGT *but does not actually locally index it*
### What happened?

In both [`make_variants_matrix_table`](https://github.com/hail-is/hail/blob/main/hail/python/hail/vds/combiner/combine.py#L59-L60) and [`make_variants_stream`](https://github.com/hail-is/hail/blob/main/hail/python/hail/vds/combiner/combine.py#L189-L190), we simply copy the `PGT` to `LPGT` without fixing the indexing. We must properly fix the indexing.

### Version

0.2.142

### Relevant log output

_No response_

chrisvittal:
resolves #13828

### DIFF
--- a/hail/python/hail/vds/combiner/combine.py
+++ b/hail/python/hail/vds/combiner/combine.py
@@ -31,7 +31,7 @@ def make_variants_matrix_table(mt: MatrixTable,
     transform_row = _transform_variant_function_map.get((mt.row.dtype, info_key))
     if transform_row is None or not hl.current_backend()._is_registered_ir_function_name(transform_row._name):
         def get_lgt(gt, n_alleles, has_non_ref, row):
-            index = gt.unphased_diploid_gt_index()
+            index = gt.unphase().unphased_diploid_gt_index()
             n_no_nonref = n_alleles - hl.int(has_non_ref)
             triangle_without_nonref = hl.triangle(n_no_nonref)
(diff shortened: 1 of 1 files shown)


hail/python/hail/vds/combiner/combine.py
```diff
@@ -30,16 +30,16 @@ def make_variants_matrix_table(mt: MatrixTable,
 
     transform_row = _transform_variant_function_map.get((mt.row.dtype, info_key))
     if transform_row is None or not hl.current_backend()._is_registered_ir_function_name(transform_row._name):
-        def get_lgt(e, n_alleles, has_non_ref, row):
-            index = e.GT.unphased_diploid_gt_index()
+        def get_lgt(gt, n_alleles, has_non_ref, row):
+            index = gt.unphased_diploid_gt_index()
```
danking, line 34: Is this function valid on phased GTs? Do we need to unphase it first?

danking (changes requested):

hail/python/hail/vds/combiner/combine.py
chrisvittal: we do thanks for checking

danking (approved):

PR #13832, opened 2023-10-16
[fs] improve copy tool progress bars

danking:
I renamed RichProgressBar and SimpleRichProgressBar to ...CopyToolProgressBar because that is more accurate. I enhanced both to now include a count and a rate with the right units based on the description. It is a bit flaky because I need the descriptions to be exactly "files" or exactly "bytes" to pick the right units, but this seems fine for the specific case of th CopyToolProgressBar.

There is probably a better way to build these UIs. I am sure we will start to figure that out as we use rich more.

Before:
Screenshot 2023-10-16 at 18 28 14

After:
Screenshot 2023-10-16 at 18 27 53

### DIFF
--- a/hail/python/hailtop/aiotools/copy.py
+++ b/hail/python/hailtop/aiotools/copy.py
@@ -7,7 +7,7 @@ import sys
 
 from concurrent.futures import ThreadPoolExecutor
 
-from ..utils.rich_progress_bar import RichProgressBar, make_listener
+from ..utils.rich_progress_bar import CopyToolProgressBar, make_listener
 from . import Transfer, Copier
 from .router_fs import RouterAsyncFS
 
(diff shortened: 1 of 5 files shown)


danking:
bump; there's a few more FS / copy-tool changes I need to get in and this is at the base of that.

daniel-goldstein (approved):

daniel-goldstein:
Making way for the release fix

jmarshall:
There is occasional use of this in other projects, e.g., gnomad_methods (`SimpleRichProgressBar` in this case).

Do you consider these classes to be part of the API? Was it intended to rename these without any compatibility shim, e.g., having the old names as aliases for a while?

It's not the end of the world and gnomad_methods has already updated accordingly. It does however mean that older gnomad_methods is only compatible with hail ≤ 0.2.125 and newer gnomad_methods is only compatible with hail ≥ 0.2.126, which is an otherwise unnecessary lock-step restriction.

PR #13867, opened 2023-10-20
[aiogoogle] finally squash an open http session

danking:
:facepalm:

I grepped for other hasattr that do not check `self`


daniel-goldstein (approved):
Oof

PR #13868, opened 2023-10-20
[batch] wait long enough for aiohttp to clean itself up

danking:
We already do this in httpx.ClientSession. I just copied from there.

### DIFF
--- a/batch/batch/driver/main.py
+++ b/batch/batch/driver/main.py
@@ -1624,11 +1624,13 @@ async def on_cleanup(app):
             cleanup.push_async_callback(app['db'].async_close)
 
             k8s: kubernetes_asyncio.client.CoreV1Api = app['k8s_client']
+
             async def close_and_wait():
                 # - Following warning mitigation described here: https://github.com/aio-libs/aiohttp/pull/2045
                 # - Fixed in aiohttp 4.0.0: https://github.com/aio-libs/aiohttp/issues/1925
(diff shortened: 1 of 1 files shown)


batch/batch/driver/main.py
```diff
@@ -1622,7 +1622,14 @@ async def on_cleanup(app):
             cleanup.push_async_callback(app['client_session'].close)
             cleanup.callback(app['async_worker_pool'].shutdown)
             cleanup.push_async_callback(app['db'].async_close)
-            cleanup.push_async_callback(app['k8s_client'].api_client.rest_client.pool_manager.close)
+
+            k8s: kubernetes_asyncio.client.CoreV1Api = app['k8s_client']
+            async def close_and_wait():
+                # - Following warning mitigation described here: https://github.com/aio-libs/aiohttp/pull/2045
+                # - Fixed in aiohttp 4.0.0: https://github.com/aio-libs/aiohttp/issues/1925
+                await k8s.api_client.close()
```
danking, line 1630: Worth double checking, but as of our current version of async_kubernetes, there are now close methods that recursively close down to the http session. They also close thread pools.

daniel-goldstein (approved):

batch/batch/driver/main.py
daniel-goldstein: Could you make an issue to find and replace all the hacky k8s client closing?

PR #13870, opened 2023-10-20
[batch] maybe reduce average JVMJob "connecting to jvm" time

danking:


batch/batch/worker/worker.py
```diff
@@ -2942,39 +2951,50 @@ def __init__(self, client_session: httpx.ClientSession):
...
+            self._jvm_waiters.put_nowait(n_cores)
+            return await jvmqueue.queue.get()
 
     def return_jvm(self, jvm: JVM):
         jvm.reset()
-        self._jvms.add(jvm)
+        self._jvms_by_cores[jvm.n_cores].queue.put_nowait(jvm)
 
     async def recreate_jvm(self, jvm: JVM):
-        self._jvms.remove(jvm)
```
danking, line 2974: I think this is a latent bug; the JVM is still owned by the job when recreate_jvm is called, it won't be in this array. If this every happened it would fail.

batch/batch/worker/worker.py
```diff
@@ -2984,8 +3004,9 @@ async def shutdown(self):
         log.info('Worker.shutdown')
         self._jvm_initializer_task.cancel()
         async with AsyncExitStack() as cleanup:
-            for jvm in self._jvms:
```
danking, line 2987: this seems a bit odd; the JVMs might still be held by jobs, right?

batch/batch/worker/worker.py
```diff
@@ -2942,39 +2953,54 @@ def __init__(self, client_session: httpx.ClientSession):
...
-            assert len(jvms) == N_JVM_CONTAINERS
-            self._jvms.update(await asyncio.gather(*jvms))
-        log.info(f'JVMs initialized {self._jvms}')
+            global_jvm_index = 0
+            while True:
+                try:
+                    n_cores = self._jvm_waiters.get_nowait()
+                    jvmqueue = self._jvms_by_cores[n_cores]
+                    jvmqueue.queue.put_nowait(await JVM.create(global_jvm_index, n_cores, self))
+                    jvmqueue.total += 1
```
jigold, line 2971: I'm having a hard time following this code and why the outer queue of `jvm_waiters` is necessary.

jigold (changes requested):

batch/batch/worker/worker.py
```diff
@@ -2942,39 +2983,54 @@ def __init__(self, client_session: httpx.ClientSession):
...
+                        next_unfull_jvmpool = jvmpool
+                        break
+
+                if next_unfull_jvmpool is None:
+                    break
+                await next_unfull_jvmpool.create_jvm()
+
+        assert self._waiting_for_jvm_with_n_cores.empty()
+        assert all(jvmpool.full() for jvmpool in self._jvmpools_by_cores.values())
+        log.info(f'JVMs initialized {self._jvmpools_by_cores}')
```
danking, line 3014: @jigold Thanks for pushing back! What do you think of it now?

To directly answer your question: one queue (`jvmpool.queue`) is a place for a consumer to borrow a JVM, the other queue (`waiting_for_jvm_with_n_cores`) is a place for a producer to learn that a consumer is waiting.

Without `waiting_for_jvm_with_n_cores`, `_initialize_jvms` has no way to be told that someone is waiting for a JVM. `asyncio.Queue` doesn't expose a method like `has_waiters()`.

jigold:
Much better! I understand what's going on now. Just to make sure I understand where the performance improvements are, we don't wait for all JVMs to be intitialized before accepting JVM jobs and the queue is FIFO so we reuse the same JVMs that are warm already?

jigold (dismissed):
See comment.

danking:
> we don't wait for all JVMs to be intitialized before accepting JVM jobs and the queue is FIFO so we reuse the same JVMs that are warm already?

No. We have no way to accept only JVM jobs or only Batch jobs, so we either accept all jobs or no jobs. In main, we accept jobs before the JVMs have initialized. We wait for all JVMs to initialize before giving JVMs to any JVM Job.

There are two improvements in this PR:
1. If a JVM with the requested number of cores is available, allow the requesting JVMJob to start before the remaining JVMs are initialized.
2. Rather than starting all the JVMs in parallel, start JVMs serially *and also* start them in the order that they are requested. If we have three waiting JVM Jobs two requesting 1 core and one requesting 4 cores, prefer to start JVMs with 1 and 4 cores to JVMs with 2 or 8 cores.


(2) might sound slower (why start serially when we can stat in parallel?) but it appears that 30 JVMs competing for CPU time dramatically slows down average start up time. In both main and this PR it takes about ~25s for all JVMs to be ready; however, in this PR, some jobs can start much sooner than 25s b/c their JVMs are started first.

jigold (approved):

PR #13880, opened 2023-10-20
[benchmark] Benchmarks on Ubuntu 22.04

ehigham:
benchmark make targets are now made directly through the make targets in `benchmark/Makefile`. Previous targets in `hail/Makefile` have been deleted.
To submit benchmarks in batch from the root directory, use
```
$ BENCHMARK_LABEL=label NAMESPACE=default BENCHMARK_PROJECT=your-google-project make -C benchmark submit
```
The benchmark docker image is now based of `hail-ubuntu` with a minimal set of requirements installed on top.
The hail wheel is added after the benchmark wheel in attempt to optimise build times for changes to hail code.


hail/python/dev/requirements.txt
```diff
@@ -21,6 +21,9 @@ sphinx_rtd_theme>=1.0.0,<2
 jupyter>=1.0.0,<2
 sphinxcontrib.katex>=0.9.0,<1
 fswatch>=0.1.1,<1
+
+# benchmark requirements
+wheel>=0.41,<0.42
```
daniel-goldstein, line 26: Can you explain why you just now needed to add wheel to the dev requirements?

benchmark/Dockerfile
```diff
@@ -27,17 +29,18 @@ RUN wget -O - https://packages.adoptium.net/artifactory/api/gpg/key/public | apt
...
 COPY requirements.txt requirements.txt
 RUN python3 -m pip install --upgrade --no-cache-dir --upgrade pip && \
-  python3 -m pip install --upgrade --no-cache-dir setuptools && \
-  python3 -m pip install --upgrade --no-cache-dir -r requirements.txt && \
-  python3 -m pip install --upgrade --no-cache-dir aiomysql && \
-  python3 -m pip check
+    python3 -m pip install --upgrade --no-cache-dir setuptools && \
+    python3 -m pip install --upgrade --no-cache-dir pycairo && \
+    python3 -m pip install --upgrade --no-cache-dir -r requirements.txt && \
+    python3 -m pip install --upgrade --no-cache-dir aiomysql && \
```
daniel-goldstein, line 35: I know this isn't really to do with your PR, but do you know if there's any reason these all can't be combined into the same pip install? That way any shared transitive dependency versions are resolved only once and we can be fully confident that the pinned versions in the requirements files are actually what gets installed. If I'm not mistaken the separate install for aiomysql could force an upgrade of a previously pinned dependency. I also am confused why aiomysql is even needed in this image.

benchmark/Dockerfile
```diff
@@ -27,17 +29,18 @@ RUN wget -O - https://packages.adoptium.net/artifactory/api/gpg/key/public | apt
...
+    python3 -m pip check
 
 ARG HAIL_WHEEL
 COPY $HAIL_WHEEL .
 RUN python3 -m pip install --upgrade --no-cache-dir --quiet $HAIL_WHEEL && \
-  python3 -m pip check
+    python3 -m pip check
```
daniel-goldstein, line 41: On a similar note to my above comment, we expect at this point that we have installed all dependencies in a previous layer, so `--upgrade` shouldn't be doing anything. I would go so far as to replace it with `--no-deps` to ensure that we aren't accidentally installing pip dependencies in this layer because that would happen every time we rebuild with changes to the hail wheel.

daniel-goldstein (changes requested):

hail/python/dev/requirements.txt
ehigham: I can't build the benchmark wheel when I create a new venv because `wheel` has been missing from the dev requirements. I've been too lazy to PR this fix until now lol

benchmark/Dockerfile
ehigham: smarts

benchmark/Dockerfile
ehigham: There are a number of things about this I don't understand.
I'm not sure why we need `pycairo` - it's not installed on my system and I don't need it to run benchmarks. Yet without it:
```
42.46 pygobject 3.42.1 requires pycairo, which is not installed.
```
I also don't have pygobject `installed`. If I move this into requirements.txt, I still get this error. I'll keep digging.

daniel-goldstein (dismissed):

daniel-goldstein (approved):

PR #13885, opened 2023-10-23

danking:

Inspired by @jigold 's file system improvement campaign, I pursued the avoidance of "list" operations. I anticipate this reduces flakiness in Azure (which is tracked in #13351) and cost in Azure.

I enforced aiotools.fs terminology on hail.fs and Scala:

1. `FileStatus`. Metadata about a blob or file. It does not know if a directory exists at this path.

2. `FileListEntry`. Metadata from a list operation. It knows if a directory exists at this path.

Variable names were updated to reflect this distinction:

1. `fileStatus` / `fileStatuses`

2. `fle`/ `fles` / `fileListEntry` / `fileListEntries`, respectively.

`listStatus` renamed to `listDirectory` for clarity.

In both Azure and Google, `fileStatus` does not use a list operation.

`fileListEntry` can be used when we must know if a directory exists. I just rewrote this from first principles because:
1. In neither Google nor Azure did it check if the path was a directory and a file.
2. In Google, if the directory entry wasn't in the first page, it would fail (NB: there are fifteen non-control characters in ASCII before `/`, if the page size is 15 or fewer, we'd miss the first entry with a `/` at the end).
3. In Azure, we issued both a get and a list.

There are now unit tests for this method.

---

1. `copyMerge` and `concatenateFiles` previously used `O(N_FILES)` list operations, they now use `O(N_FILES)` get operations.
2. Writers that used `exists` to check for a _SUCCESS file now use a get operation.
3. Index readers, import BGEN, and import plink all now check file size with a get operation.

That said, overall, the bulk of our Class A Operations are probably writes.


danking:
OK, I think this is finally ready for review. It can wait until January if necessary.

hail/python/test/hail/utils/test_utils.py
```diff
@@ -170,7 +170,7 @@ def test_hadoop_ls_file_that_does_not_exist(self):
         except FileNotFoundError:
             pass
         except FatalError as err:
-            assert 'FileNotFoundException: a_file_that_does_not_exist' in err.args[0]
+            assert 'FileNotFoundException: file:/io/a_file_that_does_not_exist' in err.args[0]
```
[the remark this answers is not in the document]
danking: I admittedly was getting frustrated with Hadoop at this point. I could change the HadoopFSURL.toString to elide `file:`, but that would be confusing when you're actually in an HDFS cluster. On the other hand, including file: is confusing if you're not in an HDFS cluster.


On the specific question of `file:` versus `file:/`, `file:` is the correct choice. The [URL grammar](https://en.wikipedia.org/wiki/URL#/media/File:URI_syntax_diagram.svg) calls the stuff before the `:` the "scheme". The reason why web URLs are always `http://` is because they have a "host" portion (the domain name). File systems don't have a host because they're assumed to be local. The `/` in `file:/` is just the name of the root directory of the file system: `/`.

hail/src/main/scala/is/hail/io/fs/GoogleStorageFS.scala
```diff
@@ -64,17 +64,19 @@ object GoogleStorageFileListEntry {
   def apply(blob: Blob): BlobStorageFileListEntry = {
     val isDir = blob.isDirectory
 
-    val name = dropTrailingSlash(blob.getName)
-
     new BlobStorageFileListEntry(
-      s"gs://${ blob.getBucket }/$name",
+      s"gs://${ blob.getBucket }/${blob.getName}",
```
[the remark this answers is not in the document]
danking: I've made this change.

We do use `${url}` pervasively. Should that be `${ url }`?

hail/src/main/scala/is/hail/io/fs/GoogleStorageFS.scala
```diff
@@ -449,31 +451,47 @@ class GoogleStorageFS(
...
+      storage.get(url.bucket, url.path, options:_*),
+      BlobGetOption.userProject _,
+      url.bucket
+    )
+  }
 
+  override def fileStatus(url: URL): FileStatus = retryTransientErrors {
     if (url.path == "")
-      return new BlobStorageFileListEntry(s"gs://${url.bucket}", null, 0, true)
+      return GoogleStorageFileListEntry.dir(url)
```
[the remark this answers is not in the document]
danking: A FileListEntry is a subtype of a FileStatus.

A FileStatus is a thing about which we know the name, the size, the creation time, etc. A FileListEntry is something we know all of that *and also* we know for sure if there is a directory with this name.

hail/src/main/scala/is/hail/io/fs/HadoopFS.scala
```diff
@@ -171,6 +174,14 @@ class HadoopFS(private[this] var conf: SerializableHadoopConfiguration) extends
     files.map(fileListEntry => new HadoopFileListEntry(fileListEntry))
   }
 
+  override def fileStatus(url: URL): FileStatus = {
```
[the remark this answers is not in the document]
danking: See above, FileListEntry is a sub-type of FileStatus. The compiler wouldn't compile this code otherwise.

hail/src/test/scala/is/hail/io/fs/FSSuite.scala
```diff
@@ -42,45 +44,56 @@ trait FSSuite extends TestNGSuite {
...
+    assert(s.getLen == 12)
+  }
+
   @Test def testFileListEntryOnFile(): Unit = {
     // file
     val f = r("/a")
     val s = fs.fileListEntry(f)
     assert(s.getPath == f)
-    assert(s.isFile)
-    assert(!s.isDirectory)
```
jigold, line 64: Why did you get rid of these tests here? They seem valuable to me.

hail/src/test/scala/is/hail/io/fs/FSSuite.scala
```diff
@@ -42,45 +44,56 @@ trait FSSuite extends TestNGSuite {
...
     assert(s.getLen == 12)
   }
 
+  @Test def testFileStatusOnDirIsFailure(): Unit = {
+    val f = r("/dir")
+    TestUtils.interceptException[FileNotFoundException](r("/dir"))(
+      fs.fileStatus(r("/dir"))
```
jigold, line 81: I thought you could create a fileStatus object for a directory. It just doesn't have the directory information as a FLE would have.

hail/src/test/scala/is/hail/io/fs/FSSuite.scala
```diff
@@ -42,45 +44,56 @@ trait FSSuite extends TestNGSuite {
...
+    )
+  }
+
   @Test def testFileListEntryOnDir(): Unit = {
     // file
     val f = r("/dir")
     val s = fs.fileListEntry(f)
     assert(s.getPath == f)
-    assert(!s.isFile)
```
jigold, line 73: Same comment about removing these tests and whether these are still valuable. Same below.

jigold (changes requested):

hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
```diff
@@ -252,15 +252,15 @@ object GenericLines {
...
+    fileStatuses0: IndexedSeq[_ <: FileStatus],
     nPartitions: Option[Int],
     blockSizeInMB: Option[Int],
     minPartitions: Option[Int],
     gzAsBGZ: Boolean,
     allowSerialRead: Boolean,
     filePerPartition: Boolean = false
   ): GenericLines = {
-    val fileListEntries = fileListEntries0.zipWithIndex.filter(_._1.getLen > 0)
+    val fileListEntries = fileStatuses0.zipWithIndex.filter(_._1.getLen > 0)
```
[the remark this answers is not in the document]
danking: done.

hail/src/main/scala/is/hail/expr/ir/AbstractMatrixTableSpec.scala
```diff
@@ -25,15 +25,21 @@ object RelationalSpec {
...
-        fatal(s"MatrixTable and Table files are directories; path '$path' is not a directory")
-      }
-    }
     val metadataFile = path + "/metadata.json.gz"
-    val jv = using(fs.open(metadataFile)) { in => parse(in) }
+    val jv = try {
+      using(fs.open(metadataFile)) { in => parse(in) }
+    } catch {
+      case exc: FileNotFoundException =>
+        if (fs.isFile(path)) {
```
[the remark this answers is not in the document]
danking: Notice that we're catching FileNotFoundException for the metadataFile, but then checking if the path (i.e. the matrix table path) is a file or a directory.

hail/src/test/scala/is/hail/io/fs/FSSuite.scala
danking: mistake, restored.

hail/src/test/scala/is/hail/io/fs/FSSuite.scala
danking: You can only create a fileStatus for a file/object not a directory/prefix. In particular, no single Class B Operation can reliably determine if a path is a directory. For example, if a GET for the object `gs://bucket/dir/a` returns 200, I know that `gs://bucket/dir` is a directory but if it returns 404, I don't know if that directory does or does not exist.

hail/src/test/scala/is/hail/io/fs/FSSuite.scala
danking: restored and restored in the next test.

hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
```diff
@@ -260,8 +260,8 @@ object GenericLines {
     allowSerialRead: Boolean,
     filePerPartition: Boolean = false
   ): GenericLines = {
-    val fileListEntries = fileStatuses0.zipWithIndex.filter(_._1.getLen > 0)
-    val totalSize = fileListEntries.map(_._1.getLen).sum
+    val fileSTatuses = fileStatuses0.zipWithIndex.filter(_._1.getLen > 0)
```
jigold, line 263:
```suggestion
    val fileStatuses = fileStatuses0.zipWithIndex.filter(_._1.getLen > 0)
```

hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
```diff
@@ -260,8 +260,8 @@ object GenericLines {
...
+    val totalSize = fileSTatuses.map(_._1.getLen).sum
```
jigold, line 264:
```suggestion
    val totalSize = fileStatuses.map(_._1.getLen).sum
```

hail/src/main/scala/is/hail/expr/ir/GenericLines.scala
```diff
@@ -276,7 +276,7 @@ object GenericLines {
       case None =>
     }
 
-    val contexts = fileListEntries.flatMap { case (fileListEntry, fileNum) =>
+    val contexts = fileSTatuses.flatMap { case (fileListEntry, fileNum) =>
```
jigold, line 279:
```suggestion
    val contexts = fileStatuses.flatMap { case (fileListEntry, fileNum) =>
```

jigold (changes requested):

jigold (approved):

PR #13888, opened 2023-10-23
[query] Teach hl.default_reference to set the default reference genome

chrisvittal:
CHANGELOG: hl.default_reference can now be passed an argument to change the default reference genome.

Now, when hl.default_reference is passed a reference genome argument, it sets the default reference to that reference genome and returns None.


danking (changes requested):
Excellent! Can you add a `CHANGELOG` line we can use to inform users of this new functionality? Otherwise I'm ready to approve.

danking (approved):

PR #13890, opened 2023-10-24
[make] Separate image building from image pushing

daniel-goldstein:
This came to mind yesterday during our pairing. This PR introduces the following properties that our image building targets do not currently have:
1. If your intention is only to build images, you shouldn't need `kubectl`. When `DOCKER_PREFIX` is used as a docker build arg it is because we mirror some dockerhub images inside our registry (for reliability/rate limiting reasons). But for local building there's no reason you can't use the dockerhub image. Also, other people should be able to build the hail image if they want to!
2. One should *only* need to use `kubectl` if they are intending to use an image in a kubernetes deployment. In other words, you should only need the private registry `DOCKER_PREFIX` for pushing images.
3. One should not need to endure image pushing if the only goal is to build the image locally
4. No intermediate tags should end up in the private registry. If we push on every image build, the private docker registry will accumulate a lot of `hail-ubuntu:dev-xxxxxx` tags that are never used again because `hail-ubuntu` is just an intermediate used to build other images. This does *not* change the number of layers that end up in the registry, but reduces a bit of the work that the registry cleanup job needs to do to untag and delete images and just seems cleaner.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,10 @@ SERVICES_DATABASES := $(patsubst %, %-db, $(SERVICES))
 SERVICES_MODULES := $(SERVICES) gear web_common
 CHECK_SERVICES_MODULES := $(patsubst %, check-%, $(SERVICES_MODULES))
 
+HAILGENETICS_IMAGES = $(foreach img,hail vep-grch37-85 vep-grch38-95,hailgenetics-$(img))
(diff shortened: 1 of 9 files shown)


ehigham (approved):
This is great! Thank you

PR #13892, opened 2023-10-24
[batch] Stop writing to v2 billing tables

jigold:
This PR modifies the billing triggers to stop writing to the v2 billing tables as well as remove the check for whether the equivalent v2 rows have been "migrated" when writing to the v3 tables. Stacked on #13891.

### DIFF
--- a/batch/sql/estimated-current.sql
+++ b/batch/sql/estimated-current.sql
@@ -597,7 +597,7 @@ BEGIN
   IF msec_diff_rollup != 0 THEN
     INSERT INTO aggregated_billing_project_user_resources_v3 (billing_project, user, resource_id, token, `usage`)
     SELECT batches.billing_project, batches.`user`,
-      attempt_resources.resource_id,
+      attempt_resources.deduped_resource_id,
       rand_token,
       msec_diff_rollup * quantity
     FROM attempt_resources
@@ -607,7 +607,7 @@ BEGIN
 
     INSERT INTO aggregated_job_group_resources_v3 (batch_id, resource_id, token, `usage`)
     SELECT attempt_resources.batch_id,
-      attempt_resources.resource_id,
+      attempt_resources.deduped_resource_id,
       rand_token,
       msec_diff_rollup * quantity
     FROM attempt_resources
@@ -616,7 +616,7 @@ BEGIN
 
     INSERT INTO aggregated_job_resources_v3 (batch_id, job_id, resource_id, `usage`)
     SELECT attempt_resources.batch_id, attempt_resources.job_id,
-      attempt_resources.resource_id,
+      attempt_resources.deduped_resource_id,
       msec_diff_rollup * quantity
     FROM attempt_resources
     WHERE attempt_resources.batch_id = NEW.batch_id AND attempt_resources.job_id = NEW.job_id AND attempt_id = NEW.attempt_id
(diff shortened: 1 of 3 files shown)


batch/sql/estimated-current.sql
```diff
@@ -595,101 +595,43 @@ BEGIN
...
-      msec_diff_rollup * quantity
-    FROM attempt_resources
-    JOIN batches ON batches.id = attempt_resources.batch_id
-    WHERE batch_id = NEW.batch_id AND job_id = NEW.job_id AND attempt_id = NEW.attempt_id
-    ON DUPLICATE KEY UPDATE `usage` = `usage` + msec_diff_rollup * quantity;
-
     INSERT INTO aggregated_billing_project_user_resources_v3 (billing_project, user, resource_id, token, `usage`)
     SELECT batches.billing_project, batches.`user`,
-      attempt_resources.deduped_resource_id,
+      attempt_resources.resource_id,
```
daniel-goldstein, line 600: Can you remind me why this column name is changing?

batch/sql/estimated-current.sql
```diff
@@ -883,74 +825,25 @@ BEGIN
   SET cur_billing_date = CAST(UTC_DATE() AS DATE);
 
   IF msec_diff_rollup != 0 THEN
-    INSERT INTO aggregated_billing_project_user_resources_v2 (billing_project, user, resource_id, token, `usage`)
+    INSERT INTO aggregated_billing_project_user_resources_v3 (billing_project, user, resource_id, token, `usage`)
```
daniel-goldstein, line 828: And just checking here to make absolutely sure, `resource_id` and `deduped_resource_id` are the same in this table?

daniel-goldstein (dismissed):
This is awesome! Just a couple sanity check questions. I'll give this a full second pass as well just to be extra careful.

batch/sql/estimated-current.sql
jigold: At this point `resource_id` **should** always equal `deduped_resource_id` with the previous fixes to how we insert resources to not be duplicated. I made this change here in case we do decide to swap the columns "deduped_resource_id" -> "resource_id" and drop the original resource id column (has duplicates in it). We can save that for a separate step if that's easier.

batch/sql/estimated-current.sql
jigold: In the billing tables, there is no deduped_resource_id. The v2 table is essentially the "resource_id" in attempt_resources and the v3 table "resource_id" is equivalent to the "deduped_resource_id" in the attempt resources table.

batch/sql/estimated-current.sql
daniel-goldstein: Ok, I think that checks out but ya I think I works prefer just doing that separately if that's alright.

batch/sql/estimated-current.sql
daniel-goldstein: Awesome, thanks!

daniel-goldstein (dismissed):
Awesome! I think I would rather separate out the rename but otherwise all good to go

daniel-goldstein (approved):

jigold:
@daniel-goldstein Can you re-review this? I rebased on latest main.

daniel-goldstein (approved):
@jigold Everything checks out to me

PR #13896, opened 2023-10-24
[ci] include migration output in job logs

danking:
cc: @daniel-goldstein 

For y'all's consideration.

### DIFF
--- a/hail/python/hailtop/utils/process.py
+++ b/hail/python/hailtop/utils/process.py
@@ -1,6 +1,6 @@
 from typing import Tuple, List
 import asyncio
-ekimport subprocess
+import subprocess
 
 from .utils import async_to_blocking
 
--- a/hail/python/hailtop/utils/process.py
+++ b/hail/python/hailtop/utils/process.py
@@ -1,6 +1,6 @@
 from typing import Tuple, List
 import asyncio
-import subprocess
+ekimport subprocess
 
 from .utils import async_to_blocking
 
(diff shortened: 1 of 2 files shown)


hail/python/hailtop/utils/process.py
```diff
@@ -1,6 +1,6 @@
 from typing import Tuple, List
 import asyncio
-import subprocess
+ekimport subprocess
```
jigold, line 3:
```suggestion
import subprocess
```

hail/python/hailtop/utils/process.py
```diff
@@ -36,7 +46,7 @@ async def check_exec_output(command: str,
     return outerr
 
 
-async def check_shell_output(script: str, echo: bool = False) -> Tuple[bytes, bytes]:
+async def check_shell_output(script: str, echo: bool = False,) -> Tuple[bytes, bytes]:
```
jigold, line 49:
```suggestion
async def check_shell_output(script: str, echo: bool = False) -> Tuple[bytes, bytes]:
```

jigold (dismissed):

jigold:
I can't tell from your code whether you're trying to suppress the output entirely or it's supposed to be printed with no buffering by the kernel. My apologies if this is obvious.

hail/python/hailtop/utils/process.py
```diff
@@ -59,3 +69,14 @@ def sync_check_exec(*command_args: str, echo: bool = False, capture_output: bool
         subprocess.run(command_args, check=True, capture_output=capture_output)
     except subprocess.CalledProcessError as e:
         raise CalledProcessError(list(command_args), e.returncode, (e.stdout, e.stderr)) from e
+
+
+async def check_exec_no_buffering(command: str, *args: str):
```
danking, line 74: I'm not sure what to call this. I think the extant methods have somewhat confusing names because it's not clear that they buffer the entire stdout and stderr in memory. This new function just runs the subprocess with stdout and stderr connected to *this* process's stdout and stderr.

I'm very open to name suggestions.

As to how to understand what this code is doing:
- The [`asyncio.create_subprocess_exec`](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.create_subprocess_exec) docs point us at `loop.subprocess_exec`.
- The [`loop.subprocess_exec`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.subprocess_exec) docs describe the meaning of the `stdin` and `stdout` parameters:
  > the value None which will make the subprocess inherit the file descriptor from this process

jigold (changes requested):
Will rereview once bugs are fixed.

danking:
no time

PR #13901, opened 2023-10-25
[query] also prevent reading from cold storage

danking:
In Hail Query, we need to prevent reading from cold storage as well as writing to cold storage (which we already appear to be preventing).


danking:
lemme fix the bugs first

PR #13910, opened 2023-10-25
[make] Take the rest of docker-build.sh arguments as docker build arguments

daniel-goldstein:
#13890 removed an optional argument to `docker-build.sh`. Now that there are no optional arguments I think it is cleaner to claim any additional arguments as docker args instead of stuffing docker args into an environment variable.


danking (approved):
Nice

daniel-goldstein:
Waiting for higher prio PRs to merge

PR #13912, opened 2023-10-25
[ci] Dont tar the hail wheel in CI pipelines

daniel-goldstein:
IIUC the reason we `tar` the hail wheel to move it between jobs in the CI pipeline is because the wheel name must contain the pip version and that is not known statically in `build.yaml`. However, it would be just as effective to copy the wheel around inside a directory, and then we don't have to do all this tar'ing and untar'ing.

cc @ehigham I'm happy to hold off on this if it would bork your branch


danking (approved):
Sure. Trades the conceptual overhead of a tar file for a Class A Operation (to list the directory), but that seems fine.

danking:
WIP until I sort the release.

PR #13916, opened 2023-10-26
[query] avoid hanging the JVM in Dataproc

danking:
Non-daemon threads [keep a JVM alive](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html):

> When a Java Virtual Machine starts up, there is usually a single non-daemon thread (which typically calls the method named main of some designated class). The Java Virtual Machine continues to execute threads until either of the following occurs:
>
> The exit method of class Runtime has been called and the security manager has permitted the exit operation to take place.
>
> All threads that are not daemon threads have died, either by returning from the call to the run method or by throwing an exception that propagates beyond the run method.

Spark appears to wait for the JVM to terminate before it considers a job complete.

### DIFF
--- a/hail/python/hail/docs/change_log.md
+++ b/hail/python/hail/docs/change_log.md
@@ -54,7 +54,7 @@ critically depend on experimental functionality.**
 
 ## Version 0.2.125
 
-Released 2023-10-25
+Released 2023-10-26
 
 ### New Features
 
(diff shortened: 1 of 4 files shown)


daniel-goldstein (dismissed):

danking:
WIP until https://github.com/hail-is/hail/pull/13919 merges.

daniel-goldstein (approved):

PR #13917, opened 2023-10-26
[ci] Run the dataproc tests on release PRs

daniel-goldstein:
These should basically run as no-ops on every PR but then actually test dataproc if we dare to bump the pip version. Can read these CI tests as "every version of hail has passed tests on dataproc".

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -3225,9 +3225,6 @@ steps:
         namespace:
           valueFrom: default_ns.name
         mountPath: /test-dataproc-service-account-key
-    scopes:
-      - deploy
-      - dev
     clouds:
       - gcp
   - kind: runImage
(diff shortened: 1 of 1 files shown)


danking:
Failed as expected, nice!

danking (changes requested):
There's a race condition:
1. Main merges
2. CI starts testing a PR against new main
3. CI runs test_dataproc-37
4. Main deploy finally pushes the tag.

I'm not exactly sure what to do in this case. If the release fails and the tag isn't pushed you *do* want to run test_dataproc because you want to know if you've fixed the release. You don't really want to run these tests if a release is in progress though.

daniel-goldstein:
Closing for now

PR #13921, opened 2023-10-26
[hailtop] avoid errors on rare transient errors

danking:
`aiohttp.ClientOSError` inherits from `OSError`, so we can just use `errno` or `strerror` directly.

We should not directly use the `args` because one of the subclasses of `ClientOSError` sets them to *its* arguments after initializing its super classes with the expected arguments:

```python3
class ClientConnectorError(ClientOSError):
    """Client connector error.

    Raised in :class:`aiohttp.connector.TCPConnector` if
        a connection can not be established.
    """

    def __init__(self, connection_key: ConnectionKey, os_error: OSError) -> None:
        self._conn_key = connection_key
        self._os_error = os_error
        super().__init__(os_error.errno, os_error.strerror)
        self.args = (connection_key, os_error)
```

I also tried to remove `e.args` from the `ClientPayloadError` case (the one right above this, and the only one still using `e.args`), but neither that class nor any super class sets a field with the error message (in fact, no fields are ever set so we can only use `e.args`).

### DIFF
--- a/hail/python/hailtop/utils/utils.py
+++ b/hail/python/hailtop/utils/utils.py
@@ -638,8 +638,7 @@ def is_transient_error(e: BaseException) -> bool:
             and e.args[0] == "Response payload is not completed"):
         return True
     if (isinstance(e, aiohttp.ClientOSError)
-            and len(e.args) >= 2
-            and 'sslv3 alert bad record mac' in e.args[1]):
+            and 'sslv3 alert bad record mac' in e.strerror):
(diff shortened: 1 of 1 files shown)


danking:
See [`ClientConnectorError`](https://github.com/aio-libs/aiohttp/blob/v3.8.6/aiohttp/client_exceptions.py#L153-L164), [`ClientOSError`](https://github.com/aio-libs/aiohttp/blob/v3.8.6/aiohttp/client_exceptions.py#L149-L150), and Python's [`OSError`](https://github.com/python/typeshed/blob/main/stdlib/builtins.pyi#L1951-L1958) (I link typeshed b/c it is implemented in C)

daniel-goldstein (approved):

danking:
For expediency's sake merged into https://github.com/hail-is/hail/pull/13916

PR #13928, opened 2023-10-26
[query] actually eliminate daemon threads

danking:


daniel-goldstein (approved):
Very nice. Thanks for digging into this

PR #13929, opened 2023-10-26
[batch] Support running the test_regions test in non-US deployments

daniel-goldstein:


daniel-goldstein:
This needs a fix in CI to make `batch_gcp_regions` a set and not the string representation of a json array

PR #13930, opened 2023-10-26
[ci] Upgrade buildkit to 0.12.3

daniel-goldstein:
also fixed the `copy-images` target which has been broken a while because without a namespace it can't get `DOCKER_PREFIX`.


daniel-goldstein:
Don't have bandwidth at the moment, this is tracked as part of the necessary changes for supporting undefined container users

PR #13934, opened 2023-10-27
[hailctl] Add hailctl auth command to print a hail access token

daniel-goldstein:
Add a `hailctl auth` subcommand that prints an access token for the hail service, mirroring the behavior of `gcloud auth application-default print-access-token`.

### DIFF
--- a/hail/python/hailtop/hailctl/auth/cli.py
+++ b/hail/python/hailtop/hailctl/auth/cli.py
@@ -41,6 +41,18 @@ def copy_paste_login(copy_paste_token: str, namespace: NamespaceOption = None):
 
 
 @app.command()
+def print_access_token():
+    '''Obtain temporary Hail credentials.'''
+    asyncio.run(async_print_access_token())
+
+
+async def async_print_access_token():
+    from hailtop.auth import hail_credentials
+    async with hail_credentials() as credentials:
+        print(await credentials.access_token())
+
+
+@app.command()
 def logout():
     '''Revoke Hail credentials.'''
     from hailtop.auth import async_logout  # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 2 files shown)


daniel-goldstein:
I'm not sure whether we should add this proactively or not, but this is the best solution I can think of so far for @illusional's question about what to do when we have removed support for the old hail access tokens but users still are forced to run a pipeline on a very old hail version. Old hail access tokens are stored in JSON in `~/.hail/tokens.json`, so I believe (though have not yet tested, that the following should allow a user to use an old version of hail against a version of batch that only supports cloud access tokens:

On the *new* version of hail, run

```
hailctl auth login
hailctl auth print-access-token | jq -R -c '{ default: . }' > ~/.hail/tokens.json
```

Then switch to an old version and proceed as usual (but don't run `hailctl auth login`!)

daniel-goldstein:
Hrm, actually, this only solves the problem for the user, not for the Hail Query driver which would also stop working once we stop mounting the old token secrets. Going to close for now but might reopen if I come up with more thoughts on this.

daniel-goldstein:
Somewhat unrelatedly though, this seems like it would be a natural replacement for our copy-paste tokens, as you could do the following:

1. `hailctl auth print-access-token | pbcopy`
2. On some notebook where you don't have access to your hail identity for some reason, do `hl.init_batch(token=…)` or `hl.ServiceBackend(token=…)`

danking:
I don't think the last comment checks out. Access tokens aren't refreshable, right? So you'd have very short-lived access.

daniel-goldstein:
True. I thought at least for the copy-paste tokens that this would be intentional. Looks like you can get access tokens from GCP that last up to 12 hours, but that could be insufficient for large workloads.

PR #13939, opened 2023-10-27
[query] fix bug in new dict decoder

patrick-schultz:
The decoder uses a `StagedArrayBuilder` to hold elements while being sorted. The array builder is stored in a class field. When the same decoder function is called more than once, that array builder is reused.

Before this fix, the array builder was never cleared, so if the decoder function was called more than once, the array builder would still contain the elements from previously decoded dicts.

Since it's highly non-obvious that you need to call `clear` immediately after `new StagedCodeBuilder`, this PR makes the constructor take a CodeBuilder, and always inserts a clear at the call site. I also took the opportunity to CodeBuilderify the rest of the interface.


danking:
It still seems like a foot gun that I can’t use two stages code builders at once. Like, if I was implementing a “partition” function that produced two arrays from one, I might naively try to use two stages array builders. Are you saying they’ll actually be the same array builder?

patrick-schultz:
Using multiple staged array builders is fine. It's using a staged array builder in a function (or loop) which is called multiple times that's the issue.

Calling `new StagedArrayBuilder(...)` before only took a method builder, not a code builder. It created a new field on the containing class to hold the runtime array builder, and added code to the class constructor to initialize it. So if the generated code using that array builder is executed multiple times, it will be reusing the same array builder object in the same class field. But calling `new StagedArrayBuilder` multiple times will create multiple array builder objects in multiple fields.

danking:
@patrick-schultz OK, I understand now. Can you ping Ed or find someone else to review? Let's get this in post-haste.

hail/src/main/scala/is/hail/expr/ir/SpecializedArrayBuilders.scala
```diff
@@ -9,8 +9,9 @@ import is.hail.utils.BoxedArrayBuilder
 
 import scala.reflect.ClassTag
 
-class StagedArrayBuilder(val elt: SingleCodeType, val eltRequired: Boolean, mb: EmitMethodBuilder[_], len: Code[Int]) {
+class StagedArrayBuilder(cb: EmitCodeBuilder, val elt: SingleCodeType, val eltRequired: Boolean, len: Int) {
```
ehigham, line 12: I don't think it's the right move for this class to keep hold of a code-builder. It should just need a class builder to define the layout and methods of an array. Really, it should just need a module builder.
Seems the normal plan is to thread one through method calls.

ehigham:
I'm a little puzzled by this solution. I feel that we shouldn't be modifying class composition in method calls; the structure of our code seems somewhat upside down. If anything the StagedArrayBuilder should only take a class builder.

PR #13943, opened 2023-10-30
[batch] Make instance status statistics easier to understand

### LINKED ISSUE #13815: [batch] driver pool stats are confusing/wrong
### What happened?

1. On the pool.html page, the "Schedulable Cores" numerator appears to include the "pending" (aka not schedulable) cores.
2. The denominator of the "Schedulable Cores" appears to just be the "Provisioned Cores". Either we should not include a "Provisioned Cores" column, or the "Schedulable Cores" denominator should indicate the total active cores, not the total provisioned cores.

I think the intention is for "Schedulable Cores" to be a quick measure of the efficiency of the scheduler. If the scheduler can keep up with job turn over, then the schedulable cores should be 0/N_ACTIVE_CORES. The provisioned cores just gives us a sense of the total cost of the cluster at the current time.

The driver's index.html has the correct values for "Schedulable Cores" but suffers from Provisioned == Scheduled Cores denominator issue.

### Version

0.2.124

### Relevant log output

_No response_

jigold:
Fixes #13815. I tried to simplify the concepts here to be unified across all instance collection types. I renamed "provisioned" to "live". Schedulable means only workers that are active and from the latest instance version.

I think the example figures are self explanatory.

 
Screenshot 2023-10-30 at 11 33 39 AM

Screenshot 2023-10-30 at 11 35 29 AM


danking:
I like this table much better! However, it's too wide. I don't know exactly the best way to shrink it down, but here's a few off the cuff thoughts:

I don't think I care about "Live", I can do that math myself (it's pending + active, right?).

Can we shorten Instances to "I" and Cores to "C" with abbr tags a la `<abbr title="Instances">`?

I don't think I care about schedulable instances, for scheduling I really care about cores. I don't think I care about the cores column, right? Isn't that a synonym for "active cores"? If it's just "Free Cores" and "Percent Free Cores" I think we could just do "Free" and "%" as sub-headings.

I think the most important super-heading is "Schedulable", what do you think of putting that at the far left of the table?

If we swap "Spot" for "Preemptible" that will also shrink the width of the table.

danking:
I also think we should think carefully about what we want from this table and only present the information necessary for those activities. A few things I want to do with this page:
1. Core efficiency: what percent of cores that could have a job don't?
2. Pending core count: is something going wrong with activation and lots of things stuck in pending?

daniel-goldstein (changes requested):
I largely agree with Dan's big points here so I'll just leave a few smaller stylistic points/suggestions:

- I can see the argument for having both cores and instances but I rarely care about both at the same time. Maybe a toggle button could switch between showing cores and instances?

- one way to get rid of a Boolean column (Spot) is to create two tables. Separating spot and non-spot into two tables (and also leaving job private as its own thing) would allow you to get rid of the worker type column as it is already covered by the name (or vice versa). It doesn't even need to be two totally separate tables, you can reuse the same headers but then have blank rows separating the two groups and denoting which group is spot and which is not.

- I think `Total` columns should go at the end

- nit: I would personally forego the color on the Total row. I think it won't be as necessary if we can shrink down the size of the table. Also, and this is probably just me, it looks quite similar to the default text highlighting color which is a bit confusing.

- I would use blank cells instead of NA. A little less noisy.

jigold:
Do you like this better? I abbreviated Percent to % after taking the pictures.
Screenshot 2023-11-03 at 2 10 25 PM
Screenshot 2023-11-03 at 2 10 57 PM

Screenshot 2023-11-03 at 2 12 37 PM

jigold:
Also, ignore getting the schedulable cores backwards.

jigold:
Screenshot 2023-11-09 at 10 54 04 AM
Screenshot 2023-11-09 at 10 53 46 AM
Screenshot 2023-11-09 at 10 53 35 AM

daniel-goldstein (approved):

PR #13944, opened 2023-10-30
[batch] Support deployment mode as a Terra on Azure app

daniel-goldstein:
First draft of a helm chart for packaging up Hail Batch as a Terra on Azure App. This will likely need numerous bug fixes as we set up a proper testing strategy, but the rough shape of everything should be pretty stable.


daniel-goldstein:
I don't love what I've had to do with the deploy config stuff. That's in my opinion the most finicky part of this and it's mostly our fault, because we overload the `namespace` parameter with both identifying the namespace in Kubernetes and signifying whether the environment is prod or not. All I want really is to change the `domain` to a domain and path prefix, and not have the namespace have such an impact on routing. Like what if `namespace` didn't affect routing, but if the deploy config only gave a domain with no path e.g. `example.com`, we use subdomains so `batch.example.com`, but if we provided a domain with a path prefix like `example.com/dgoldste`, we make the batch root `example.com/dgoldste/batch`?

hail/python/hailtop/config/deploy_config.py
```diff
@@ -133,6 +135,58 @@ async def get_metrics(_):
...
+        return external_client_ssl_context()
+
+    def server_ssl_context(self) -> Optional[ssl.SSLContext]:
+        if self._location == 'k8s':
+            return internal_server_ssl_context()
+        # local mode does not have access to self-signed certs
+        return None
+
+
+class TerraDeployConfig(DeployConfig):
```
danking, line 151: It seems to me that a namespace is fully identified by its domain & subpath. If you replace the explicit namespace name with a pair of domain and subpath everywhere it's used, would that work? You'll need to migrate the tokens files or just force everyone to log in again so that the pair is used as the key. Instead of saying "Logged in to namespace XXX" you can just say "Logged in to domain/subpath". I think your suggestion in the main thread comment works: if subpath is empty, you put the service in the domain name otherwise you append it.

hail/src/main/scala/is/hail/services/DeployConfig.scala
```diff
@@ -122,3 +127,23 @@ class DeployConfig(
...
+    this.domain
+  }
+
+  override def basePath(service: String): String = {
+    s"$subpath/$service"
+  }
+
+  override def scheme(baseScheme: String = "http"): String = {
+    "https"
+  }
```
danking, line 148: Hmm. We always communicate over https? What about driver<->front-end comms, does that not use the local http connection?

hail/src/main/scala/is/hail/io/fs/TerraAzureStorageFS.scala
```diff
@@ -0,0 +1,93 @@
...
+  private val storageContainerUrl = sys.env.get("WORKSPACE_STORAGE_CONTAINER_URL")
+  private val log = Logger.getLogger(getClass.getName)
+}
+
+class TerraAzureStorageFS() extends FS {
+  type URL = AzureStorageFSURL
+  import TerraAzureStorageFS.{log, workspaceManagerUrl, workspaceId, containerResourceId, storageContainerUrl}
+
+  private[this] val azureFS = new AzureStorageFS()
+  private[this] val credential: DefaultAzureCredential = new DefaultAzureCredentialBuilder().build()
```
danking, line 34: prefer no types on the vals

hail/src/main/scala/is/hail/io/fs/TerraAzureStorageFS.scala
```diff
@@ -0,0 +1,93 @@
...
+
+object TerraAzureStorageFS {
+  private val workspaceManagerUrl = sys.env.get("WORKSPACE_MANAGER_URL")
+  private val workspaceId = sys.env.get("WORKSPACE_ID")
+  private val containerResourceId = sys.env.get("WORKSPACE_STORAGE_CONTAINER_ID")
+  private val storageContainerUrl = sys.env.get("WORKSPACE_STORAGE_CONTAINER_URL")
+  private val log = Logger.getLogger(getClass.getName)
+}
+
+class TerraAzureStorageFS() extends FS {
```
danking, line 29: prefer no parens

hail/src/main/scala/is/hail/io/fs/TerraAzureStorageFS.scala
```diff
@@ -0,0 +1,93 @@
...
+  type URL = AzureStorageFSURL
+  import TerraAzureStorageFS.{log, workspaceManagerUrl, workspaceId, containerResourceId, storageContainerUrl}
+
+  private[this] val azureFS = new AzureStorageFS()
+  private[this] val credential: DefaultAzureCredential = new DefaultAzureCredentialBuilder().build()
+  private[this] val httpClient = HttpClients.custom().build()
+  private[this] val sasTokenCache = mutable.Map[String, (String, Long)]()
+
+  def parseUrl(filename: String): AzureStorageFSURL = {
+    val urlStr = storageContainerUrl match {
```
danking, line 39: What does it mean to not have a storage container in Azure?

hail/src/main/scala/is/hail/io/fs/TerraAzureStorageFS.scala
```diff
@@ -0,0 +1,93 @@
...
+      case Some(url) if filename.startsWith(url) =>
+        sasTokenCache.get(filename) match {
+          case Some((sasTokenUrl, expiration)) if expiration > System.currentTimeMillis + 10 * 60 * 1000 => sasTokenUrl
```
danking, line 42: Mild preference for a named constant here like `TEN_SECONDS_IN_MS`

hail/src/main/scala/is/hail/io/fs/TerraAzureStorageFS.scala
```diff
@@ -0,0 +1,93 @@
...
+
+    val context = new TokenRequestContext()
+    context.addScopes("https://management.azure.com/.default")
+    val token = credential.getToken(context).block().getToken()
+
+    val url = s"${workspaceManagerUrl}/api/workspaces/v1/${workspaceId}/resources/controlled/azure/storageContainer/${containerResourceId}/getSasToken"
+    val req = new HttpPost(url)
+    req.addHeader("Authorization", s"Bearer $token")
+
+    val expiresInSeconds = 3600
```
danking, line 62: Are we allowed to set this to, like, 10 hours?

hail/src/main/scala/is/hail/io/fs/TerraAzureStorageFS.scala
```diff
@@ -0,0 +1,93 @@
...
+  def setConfiguration(config: Any): Unit = azureFS.setConfiguration(config)
+  def openNoCompression(url: URL): SeekableDataInputStream = azureFS.openNoCompression(url)
+  def createNoCompression(url: URL): PositionedDataOutputStream = azureFS.createNoCompression(url)
+  def delete(url: URL, recursive: Boolean): Unit = azureFS.delete(url, recursive)
+  def glob(url: URL): Array[FileListEntry] = azureFS.glob(url)
+  def fileListEntry(url: URL): FileListEntry = azureFS.fileListEntry(url)
+  def makeQualified(filename: String): String = azureFS.makeQualified(filename)
+  def eTag(url: URL): Some[String] = azureFS.eTag(url)
+  def listDirectory(url: URL): Array[FileListEntry] = azureFS.listDirectory(url)
+  def urlAddPathComponent(url: URL, component: String): URL = url.addPathComponent(component)
```
danking, line 92: This all works out rather nicely.

hail/python/hailtop/aiotools/router_fs.py
```diff
@@ -55,10 +57,14 @@ def _load_fs(self, uri: str):
                 **self._gcs_kwargs,
                 bucket_allow_list = self._gcs_bucket_allow_list.copy()
             )
-        elif aioazure.AzureAsyncFS.valid_url(uri):
-            fs = aioazure.AzureAsyncFS(**self._azure_kwargs)
         elif aioaws.S3AsyncFS.valid_url(uri):
             fs = aioaws.S3AsyncFS(**self._s3_kwargs)
+        # Azure URLs are also valid for Terra on Azure
+        elif aioazure.AzureAsyncFS.valid_url(uri):
+            if aioterra_azure.TerraAzureAsyncFS.configured():
```
danking, line 64: I don't love configured? Maybe "enabled"? Or just "in_terra_azure"?

hail/python/hailtop/aiocloud/aioterra/azure/fs.py
```diff
@@ -0,0 +1,120 @@
...
+    def valid_url(url: str) -> bool:
+        return AzureAsyncFS.valid_url(url)
+
+    def parse_url(self, url: str) -> AzureAsyncFSURL:
+        return AzureAsyncFS.parse_url(url)
+
+    async def _to_azure_url(self, url: str) -> str:
+        if url in self._sas_token_cache:
+            sas_token, expiration = self._sas_token_cache[url]
+            ten_minutes_from_now = time_msecs() + 10 * 60
```
danking, line 42: same comment here about preferring a longer duration. I'm thinking: what if we're streaming data for a few minutes but hit a couple retries along the way?

hail/python/hailtop/aiocloud/aioterra/azure/fs.py
```diff
@@ -0,0 +1,120 @@
...
+            if expiration > ten_minutes_from_now:
+                return sas_token
+
+        assert WORKSPACE_STORAGE_CONTAINER_URL is not None
```
danking, line 46: Assert this in the constructor too so that we get an error as soon as possible.

hail/python/hailtop/aiocloud/aioterra/azure/fs.py
```diff
@@ -0,0 +1,120 @@
...
+        if url.startswith(WORKSPACE_STORAGE_CONTAINER_URL):
+            sas_token, expiration = await self._create_terra_sas_token(url)
```
danking, line 48: I think in Scala you update the cache in the creation script; either seems fine but lets be consistent across langs.

hail/python/hailtop/aiocloud/aioterra/azure/fs.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @staticmethod
+    def valid_url(url: str) -> bool:
+        return AzureAsyncFS.valid_url(url)
+
+    def parse_url(self, url: str) -> AzureAsyncFSURL:
+        return AzureAsyncFS.parse_url(url)
+
+    async def _to_azure_url(self, url: str) -> str:
+        if url in self._sas_token_cache:
```
danking, line 40: can we use the walrus operator now?

gear/gear/clients.py
```diff
@@ -18,6 +21,8 @@ def get_cloud_async_fs() -> AsyncFS:
     cloud = get_global_config()['cloud']
 
     if cloud == 'azure':
+        if os.environ.get('HAIL_TERRA'):
```
danking, line 24: Should we just use the presence of the Terra workspace env vars as evidence that we're in azure terra?

gear/gear/auth.py
```diff
@@ -119,6 +120,19 @@ async def _fetch_userdata_from_auth_service(session_id_and_session: Tuple[str, h
             raise web.HTTPInternalServerError() from e
 
 
+class TrustedSingleTenantAuthenticator(Authenticator):
+    async def _fetch_userdata(self, _: web.Request) -> Optional[UserData]:
+        return cast(
+            UserData,
+            {
+                'is_developer': True,
+                'username': 'test-dev',
```
danking, line 129: We should maybe use `user`?

batch/batch/cloud/driver.py
```diff
@@ -16,6 +19,9 @@ async def get_cloud_driver(
 ) -> CloudDriver:
     cloud = get_global_config()['cloud']
 
+    if os.environ.get('HAIL_TERRA'):
+        return await TerraAzureDriver.create(app, db, machine_name_prefix, namespace, inst_coll_configs)
+
     if cloud == 'azure':
```
danking, line 25: seems best for consistency to put the env get under the cloud if

batch/batch/worker/worker.py
```diff
@@ -3031,9 +3035,11 @@ def return_jvm(self, jvm: JVM):
     async def return_broken_jvm(self, jvm: JVM):
         return await self._jvmpools_by_cores[jvm.n_cores].return_broken_jvm(jvm)
 
-    @property
-    def headers(self) -> Dict[str, str]:
-        return {'X-Hail-Instance-Name': NAME, 'X-Hail-Instance-Token': self.instance_token}
+    async def headers(self):
+        headers = {'X-Hail-Instance-Name': NAME, 'X-Hail-Instance-Token': self.instance_token}
+        if isinstance(CLOUD_WORKER_API, TerraAzureWorkerAPI):
+            headers['Authorization'] = 'Bearer: ' + await CLOUD_WORKER_API.vm_identity_token()
```
danking, line 3041: So, this is for talking back to the driver right? And this is because the comms from worker to driver has to go through some external load balancer rather than within the local VPC's network?

danking: Regardless, this should probably be a `worker_api.extra_headers()` method rather than an if-isinstance.

batch/batch/worker/worker.py
```diff
@@ -2273,8 +2274,11 @@ async def run(self):
 
                 self.state = 'running'
 
-                with self.step('downloading_jar'):
-                    local_jar_location = await self.download_jar()
+                if os.environ.get('HAIL_TERRA'):
+                    local_jar_location = '/hail-jars/hail-all-spark.jar'
+                else:
+                    with self.step('downloading_jar'):
+                        local_jar_location = await self.download_jar()
```
danking, line 2281: this also feels like a worker API level thing. Like `get_local_jar()` or w/e.

danking: Ehh. bleh, it depend on the worker's jar locks and the pool. You could pass that in I guess but it does start to feel a bit weird that we moved all that functionality to a different class. Let's keep it as is.

batch/batch/front_end/front_end.py
```diff
@@ -966,7 +965,7 @@ async def _create_jobs(
                         reason=f'Cloudfuse requests with mount_path=/io are not supported. Found {config}'
                     )
 
-        if spec.get('mount_tokens', False):
+        if spec.get('mount_tokens', False) and not os.environ.get('HAIL_TERRA', False):
```
danking, line 968: should this be an assertion instead? It seems wrong to ask to mount tokens if we don't have them.

batch/batch/driver/job.py
```diff
@@ -355,22 +356,22 @@ async def job_config(app, record, attempt_id):
 
     userdata = json.loads(record['userdata'])
 
-    secrets = job_spec.get('secrets', [])
-    k8s_secrets = await asyncio.gather(
-        *[k8s_cache.read_secret(secret['name'], secret['namespace']) for secret in secrets]
-    )
-
     gsa_key = None
+    if not os.environ.get('HAIL_TERRA'):
```
danking, line 360: same question here: should it just be an assertion that there are no secrets when we're in hail terra?

batch/batch/driver/instance.py
```diff
@@ -200,6 +200,9 @@ async def make_request():
                 await self.client_session.post(
                     f'http://{self.ip_address}:5000/api/v1alpha/kill', timeout=aiohttp.ClientTimeout(total=30)
                 )
+            except ConnectionRefusedError:
+                await self.deactivate('killed')
```
danking, line 204: say more?

batch/batch/cloud/terra/azure/worker/worker_api.py
```diff
@@ -0,0 +1,95 @@
...
+    def mount_path(self) -> str:
+        return ''
+
+    @property
+    def identity_provider_json(self):
+        return {'idp': IdentityProvider.MICROSOFT.value}
+
+
+class TerraAzureWorkerAPI(CloudWorkerAPI[AzureManagedIdentityCredentials]):
+    nameserver_ip = '168.63.129.16'
```
danking, line 31: unused perhaps?

batch/batch/cloud/terra/azure/driver/driver.py
```diff
@@ -0,0 +1,554 @@
...
+                    'jobControl': {'id': str(uuid.uuid4())},
+                },
+            )
+        except aiohttp.ClientResponseError as e:
+            if e.status == 404:
+                raise VMDoesNotExist() from e
+            raise
+
+    async def get_vm_state(self, instance: Instance) -> VMState:
+        # TODO This should look at the response and use all applicable lifecycle types
```
danking, line 354: say more?

batch/batch/cloud/terra/azure/driver/driver.py
```diff
@@ -0,0 +1,554 @@
...
+            'vmSize': machine_type,
+            'vmImage': {
+                'publisher': 'Canonical',
+                'offer': '0001-com-ubuntu-server-focal',
+                'sku': '20_04-lts-gen2',
+                'version': '20.04.202305150',
+            },
+            'vmUser': {
+                'name': 'hail-admin',
+                'password': 'StrongP@ssword123',  # TODO Change
```
danking, line 303: just randomly generate one, right? We don't actually need ssh access

batch/batch/cloud/terra/azure/driver/driver.py
```diff
@@ -0,0 +1,554 @@
...
+          set +x
+          sleep 1000
+          token=$(az account get-access-token --query accessToken --output tsv)
+
+          VM_RESOURCE_ID={ instance_config._resource_id }
+          curl -X POST "{ WORKSPACE_MANAGER_URL }/api/workspaces/v1/$WORKSPACE_ID/resources/controlled/azure/vm/$VM_RESOURCE_ID" \
+              -H  "accept: */*" \
+              -H  "Authorization: Bearer $token" \
+              -H  "Content-Type: application/json" \
+              -d "{{\"jobControl\":{{\"id\":\"$VM_RESOURCE_ID\"}}}}"
```
danking, line 135: presumably this deletes the vm but I'm surprised Therese no -XDELETE or `/delete` or `"action":"delete"`?

danking (changes requested):
Pretty straightforward review! Still letting it all sink in, so some of my comments might be naive; tell me if they are!

batch/terra-chart/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: hail-batch-terra-azure
+description: A chart for deploying Hail Batch into an Azure Terra workspace
+type: application
+version: "0.1.9"
```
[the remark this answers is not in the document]
daniel-goldstein: Yes, updating hail batch in terra basically just requires a PR to Leonardo that updates the chart version they use. We could just use the pip version across the board, but haven't given that much thought.

batch/terra-chart/Dockerfile.batch
```diff
@@ -0,0 +1,12 @@
+ARG BASE_IMAGE
+FROM $BASE_IMAGE
+
+# https://bugs.mysql.com/bug.php?id=105288&thanks=sub
+RUN hail-apt-get-install xz-utils libncurses5 && \
+    curl --remote-name https://downloads.mysql.com/archives/get/p/23/file/mysql-8.0.26-linux-glibc2.17-x86_64-minimal-rebuild.tar.xz && \
+    mkdir -p /opt && \
+    tar -vx -C /opt -f mysql-8.0.26-linux-glibc2.17-x86_64-minimal-rebuild.tar.xz && \
+    ln -s /opt/mysql-8.0.26-linux-glibc2.17-x86_64-minimal-rebuild/bin/* /usr/bin/
```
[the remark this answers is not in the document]
daniel-goldstein: Unfortunately do need the `mysql` CLI to run certain migrations.

batch/terra-chart/Makefile
```diff
@@ -0,0 +1,49 @@
...
+	docker build \
+		-t $(TERRA_BATCH_WORKER_IMAGE) \
+		-f Dockerfile.worker \
+		--build-arg BASE_IMAGE=$(shell cat ../../batch-worker-image)
+	echo $(TERRA_BATCH_WORKER_IMAGE) > $@
+
+terra-acr-login:
+	# ACR login is short-lived. Log in again if it's been more than an hour
+	test -n "$$(find $@-stmp -cmin -60)" || \
+	{ az acr login --name terradevacrpublic; touch $@-stmp; }
```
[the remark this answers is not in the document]
daniel-goldstein: I use that to track the last time docker was logged into ACR. There's a several second latency to the login, enough to be annoying when iterating but also only lasts a few hours unlike GAR. I did this to automatically log in if it's been "a while". I'm not sure how to conditionally run a target based on an expiration time in Make so this was the best I could come up with. Could also just not make it a target and leave a comment to log in if you want to push.

batch/terra-chart/Makefile
```diff
@@ -0,0 +1,49 @@
...
+$(TERRA_IMAGES): pushed-%: % terra-acr-login
+	docker push $(shell cat $*)
+	echo $(shell cat $*) > $@
+
+values.yaml: $(TERRA_IMAGES)
+	python3 ../../ci/jinja2_render.py '{"code":{"sha":"$(shell git rev-parse --short=12 HEAD)"},"batch_image":{"image":"$(shell cat pushed-terra-batch-image)"},"batch_worker_image":{"image":"$(shell cat pushed-terra-batch-worker-image)"}}' values.yaml.tmpl values.yaml
+
+publish-terra-helm-chart: values.yaml
+	helm package .
+	helm push hail-batch-terra-azure-*.tgz oci://terradevacrpublic.azurecr.io/hail
```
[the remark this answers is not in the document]
daniel-goldstein: Yes Rob gave me push access to the repository. Though I'm not sure I want push access long term.

hail/src/main/scala/is/hail/services/DeployConfig.scala
daniel-goldstein: Currently the only way to communicate with k8s services in Azure Terra is through Azure Relay. I would hope that Azure Relay's smart enough to not have to send that over the internet if you're already in Azure, but I'm not sure. There was a conversation a while back about adding an internal load balancer but the conclusion at the time was they preferred traffic to just go through Relay.

hail/python/hailtop/config/deploy_config.py
daniel-goldstein: I'll try this out. I'm slightly worried about backward compatibility but I think that's only a concern for dev namespaces, which I don't care about being backwards compatible with old clients. For production users, the only change should be the added field of an empty subpath.

hail/src/main/scala/is/hail/io/fs/TerraAzureStorageFS.scala
daniel-goldstein: Should be able to. I do worry that this is going to be a pain if we don't make this configurable somehow.

hail/src/main/scala/is/hail/io/fs/TerraAzureStorageFS.scala
daniel-goldstein: We should always have one if we're getting this far. I'll turn this into a non-optional field of the class.

hail/python/hailtop/aiotools/router_fs.py
daniel-goldstein: Changed to enabled.

hail/python/hailtop/aiocloud/aioterra/azure/fs.py
daniel-goldstein: Do you mean that the token should live longer or we should renew the token earlier? What's here now is that we renew the token if it is within ten minutes of its expiration.

hail/python/hailtop/aiocloud/aioterra/azure/fs.py
daniel-goldstein: Unfortunately I don't believe that tuple deconstruction works with the walrus operator, so I'm not sure I see much value.

gear/gear/clients.py
daniel-goldstein: ya we can do that.

gear/gear/auth.py
daniel-goldstein: Ya, we might want to add a small script after the migrations that, say, creates a billing project and adds them to it though.

batch/batch/worker/worker.py
daniel-goldstein: First comment is correct. Regarding the second comment, I'll do the `extra_headers` method.

batch/batch/front_end/front_end.py
daniel-goldstein: True I'll adjust the client not to request them

batch/batch/cloud/terra/azure/driver/driver.py
daniel-goldstein: Apparently this is a restriction imposed by what why use to generate their [APIs](https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/ControlledAzureResource/deleteAzureVm)

daniel-goldstein: > Delete a controlled Azure VM resource. This is async, because the operation can take a few minutes. OpenAPI does not support request body in DELETE, but async state requires it. Hence this is a POST.

hail/python/hailtop/config/deploy_config.py
daniel-goldstein: I don't think we can immediately remove `default_namespace`, old clients still expect it to exist. But we can basically stop using it in new code.

batch/batch/driver/instance.py
daniel-goldstein: I noticed before I got the workers online the driver got into a loop treating a connection refused as a transient error. I think this was before we introduced limited retries though, so I'll take this out.

batch/batch/cloud/terra/azure/worker/worker_api.py
daniel-goldstein: This is a class variable that is used by `worker.py`. It probably is inconsequential in terra but I copied this value from the Azure worker API.

batch/batch/cloud/terra/azure/driver/driver.py
daniel-goldstein: Lost track of this in all the LoC in this change. In the [WSM API](https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/ControlledAzureResource/getAzureVm) there's a `state` field in the response of the following enum `[ BROKEN, CREATING, DELETING, READY, UPDATING ]`. I'm hoping, but am not certain, that we can at least map `READY` to our notion of ready. Interestingly, this functionality doesn't seem entirely necessary, as a VM is marked ready regardless when it comes online and contacts the driver.

batch/batch/cloud/terra/azure/driver/driver.py
danking: Heh. I really hate that GET XXX gets metadata about the resource but POST XXX deletes it. That just feels dangerous, but, it is what it is.

danking (commented):
I'll do a more thorough review tomorrow; today has too many meetings sadly.

hail/python/hailtop/aiocloud/aioterra/azure/fs.py
daniel-goldstein: Ya I agree this is a problem, and one that we should probably try to solve in the terra case, but I worry we won't be able to do this in the general case, possibly even in Terra. You can be given a SAS token without having the permissions to create one. They've made credential expiration an unrecoverable error. So maybe we will have to track whether we were *originally* given a SAS token or not, which should inform whether we are able to request a new one upon expiration ☹️

batch/terra-chart/Chart.yaml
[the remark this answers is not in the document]
danking: I like pip version. The fewer distinct versions we have to deal with the better.

hail/src/main/scala/is/hail/io/fs/TerraAzureStorageFS.scala
danking: It seems to me we should set it to the longest permissible time. It's a bit antagonistic, but it gives the best outcome to the scientists and the least chance of bugs reported to us

hail/python/hailtop/aiocloud/aioterra/azure/fs.py
danking: Ah, sorry, I think I'm crossing some lines here between signed URLs and DRS.

Am I correct in now realizing, that your changes in the Scala-side are to support using Hail Query to read/write to the workspace storage container in particular?

---

I think there's two kinds of access patterns:

1. Batch-like. Cloud storage for relatively small objects. Object fits in RAM. Time between `open` and `close` is short (seconds).
2. Query-like. Cloud storage for analytical data. Object may not fit in RAM. Time between `open` and `close` is long (tens of minutes). Time between job specification and job execution could be >24h. Credentials must be long-lived or refreshable.

I think there's three kinds of URIs:

1. Cloud storage URLs. It has two sub-kinds.
   1. Normal cloud storage URLs. Authorization is handled by cloud IAM.
   2. Terra App workspace's container. The user will refer to this by its https:// URL. Batch service principal lacks permission to access this container. Terra Workspace Manager will mint a signed URL on demand.
3. DRS URIs. I *think* this is both an alias and authorization system. A DRS URI can decode into a cloud storage URL *and may be* a signed URL.
4. Signed URLs. These can come from decoding a DRS URI, accessing the Terra workspace manager API, or manually by a user. These have a lifetime. I'm pretty sure we cannot extend/refresh the expiration time.

---

So, it seems to me that for a Batch-like access pattern, constructing a signed URL at time-of-open is fine because the streams are always short-lived.

The expiration time is an issue for Query-like access.

>  You can be given a SAS token without having the permissions to create one.

I think my conclusion is: we must reject this. The user must provide one of: a normal cloud storage URL (which the Batch service principal is authorized to access), the Terra App's workspace container, or a DRS URI (which the Batch service principal is authorized to decode). We may also support signed URLs, but the user risks credential expiration. For the other three options:
1. Storage URLs. Easy. No changes to our code.
2. Terra App workspace container. Construct signed URL at the time of authn header creation.
3. DRS URIs. Same as (2): decode at time of authn header creation.

danking: And to be clear, I think for this PR we don't need to address DRS URIs. As for signed URLs, ideally we support these but, as you stated, there's nothing we can do about expiration. I think we *do* need to move URL signing to the point of authn header creation.

daniel-goldstein: Yes I agree with all of this. So in this PR the only thing we use signed URLs for are for the workspace storage container for short-lived things.

> I think we do need to move URL signing to the point of authn header creation.

Ok so we need to push down SAS token creation to any operations that issue reads/writes from the storage client. This should work so long as the storage client does not do transient error handling or issue multiple requests per client library function call…

danking: If we observe those issues, we can set a tight timeout on azure and hopefully handle retries ourselves.

danking (dismissed):
dismiss when the signed URL fiddliness is ready for another look

daniel-goldstein:
@danking , I'm a bit stuck on how to proceed with the credential refreshing. Here's the layout of the problem:

1. In normal Azure, we accept user-provided SAS tokens. Since they are user-provided, we have no way of obtaining new ones and the onus is on the user to obtain a SAS token for however long they expect to need to use it.
2. This current design in Terra is to not make the user have to do that, because that seems annoying, and for terra-controlled ABS containers we have an endpoint we can hit to get a SAS token. Ok, but now we need to update our Azure FS infrastructure to refresh a credential if it expires. But, we use the azure client lib and don't control all http requests. For example, for `AzureStorageFS.open`, we call `downloader.readall()` if we want to load the whole file into memory. I went spelunking through their source and looks like `readall` mostly wraps a sequence of range reads, but regardless if we were to use that method we would have to catch credential expiration errors, reset credentials on the blob client and retry hoping that we didn't break any invariants -- I don't want to do that as I wouldn't trust a stream that encountered a non-transient error like that. It could be that getting rid of `downloader.readall` is the only thing we have to worry about, but it makes me uneasy not having control of the http requests we're making to ABS.

Do you see a solution other than raking through our `aioazure.fs` and making sure that we only use "quick" methods and possibly retrying 401s? It just seems to me like we're going against the grain and even though it feels user-hostile the intention of SAS tokens are to have users own credential expiration.

danking (changes requested):
boop

danking (changes requested):
some thoughts, haven't gone through everything yet

batch/batch/cloud/terra/azure/driver/driver.py
```diff
@@ -0,0 +1,556 @@
...
+
+        if not local_ssd_data_disk:
+            disk_name = f'{machine_name}-data'
+            disk_config = create_disk_config(disk_name, instance_config._disk_resource_id, data_disk_size_gb)
+            try:
+                res = await self.terra_client.post('/disks', json=disk_config)
+                log.info(f'Terra response creating disk {disk_name}: {res}')
+            except Exception:
+                log.exception(f'error while creating disk {disk_name}')
+                return total_resources_on_instance
```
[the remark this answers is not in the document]
daniel-goldstein: Turns out this is how the other cloud drivers operate when they can't create an instance. We're going to keep this as is but this should be scrutinized more broadly to consider whether we should immediately remove the instance when it fails to be created.

batch/batch/cloud/terra/azure/driver/driver.py
```diff
@@ -0,0 +1,556 @@
...
+                json={
+                    'jobControl': {'id': str(uuid.uuid4())},
+                },
+            )
+
+        try:
+            await self.terra_client.post(
+                f'/vm/{terra_vm_resource_id}',
+                json={
+                    'jobControl': {'id': str(uuid.uuid4())},
```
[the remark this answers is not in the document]
daniel-goldstein: This is fair, I was simply going by the recommendation in the [WSM API docs](https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/ControlledAzureResource/createAzureVm). I don't have a big preference, happy to change it to `base64.b64encode(secrets.token_bytes(16))`

batch/batch/cloud/terra/azure/driver/driver.py
```diff
@@ -0,0 +1,556 @@
...
+            'accessScope': 'PRIVATE_ACCESS',
+            'managedBy': 'USER',
+            'resourceId': disk_resource_id,
+            'properties': [],
+        },
+        'azureDisk': {
+            'name': disk_name,
+            'size': disk_size_gb,
+        },
+    }
```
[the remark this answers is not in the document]
daniel-goldstein: We've decided to remove the data disk from this PR and only accept VMs with a local ssd.

daniel-goldstein:
> Hmm. Is the TODO suggesting that we assert the state field is READY in the response?

I handled the other states and removed the TODO

batch/batch/cloud/terra/azure/driver/driver.py
```diff
@@ -0,0 +1,530 @@
...
+
+      curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
+
+      az login --identity --allow-no-subscription
+
+      # avoid "unable to get current user home directory: os/user lookup failed"
+      export HOME=/root
+
+      UNRESERVED_WORKER_DATA_DISK_SIZE_GB=50
+      ACCEPTABLE_QUERY_JAR_URL_PREFIX={ ACCEPTABLE_QUERY_JAR_URL_PREFIX }
```
danking, line 162: I'd shq this too but its probably fine

batch/batch/cloud/terra/azure/driver/driver.py
```diff
@@ -0,0 +1,530 @@
...
+
+      sudo mkdir -p /host/batch
+      sudo mkdir -p /host/logs
+      sudo mkdir -p /host/cloudfuse
+
+      sudo mkdir -p /etc/netns
+
+      sudo mkdir /deploy-config
+      sudo cat >/deploy-config/deploy-config.json <<EOF
```
danking, line 171: use `<<'EOF'` to prevent substitution of variables in the heredoc

danking (approved):
One nit, but this seems broadly right to me!

danking:
WIP'ed in case you want to add the last shq

PR #13945, opened 2023-10-30
[batch] Check whether user exists before adding to billing project

### LINKED ISSUE #13858: [batch] billing project page should validate user names and reject invalid usernames
### What happened?

I added "Huang" to a billing project. This user name does not exist. It should have given me an error.

### Version

0.2.124

### Relevant log output

_No response_

jigold:
Fixes #13858.

Screenshot 2023-10-30 at 12 23 06 PM

### DIFF
--- a/batch/test/test_accounts.py
+++ b/batch/test/test_accounts.py
@@ -8,7 +8,7 @@ import orjson
 import pytest
 
 from hailtop import httpx
-from hailtop.auth import session_id_encode_to_str
+from hailtop.auth import async_get_user, session_id_encode_to_str
 from hailtop.batch_client.aioclient import Batch, BatchClient
 from hailtop.utils import secret_alnum_string
 from hailtop.utils.rich_progress_bar import BatchProgressBar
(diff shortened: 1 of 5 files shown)


batch/test/test_accounts.py
```diff
@@ -279,6 +279,16 @@ async def test_add_and_delete_user(dev_client: BatchClient, new_billing_project:
     assert r['user'] not in bp['users']
 
 
+async def test_error_adding_nonexistent_user(dev_client: BatchClient, new_billing_project: str):
+    project = new_billing_project
```
daniel-goldstein, line 283: nit: I don't feel like we need this variable `project`.

batch/test/test_accounts.py
```diff
@@ -279,6 +279,16 @@ async def test_add_and_delete_user(dev_client: BatchClient, new_billing_project:
...
+    try:
+        await dev_client.add_user('foobar', project)
+    except httpx.ClientResponseError as e:
+        assert e.status == 403, e
+    else:
+        assert False, 'expected error'
```
daniel-goldstein, line 289: I think you can write this instead as (note the indentation, it's a little funky):

```python
with pytest.raises(httpx.ClientResponseError) as e_info:
    await dev_client.add_user('foobar', project)
assert e_info.value.status == 403
```

batch/test/test_accounts.py
```diff
@@ -279,6 +279,16 @@ async def test_add_and_delete_user(dev_client: BatchClient, new_billing_project:
     assert r['user'] not in bp['users']
 
 
+async def test_error_adding_nonexistent_user(dev_client: BatchClient, new_billing_project: str):
+    project = new_billing_project
+    try:
+        await dev_client.add_user('foobar', project)
```
daniel-goldstein, line 285: I know this is very unlikely, but I think for clarity it's worth adding an assert before this line where you `get_user` and assert that it doesn't exist.

daniel-goldstein (changes requested):
Wonderful, just a couple really small nits

daniel-goldstein (approved):

PR #13949, opened 2023-10-30
[batch] Install the Ops Agent in GCP on instance startup

### LINKED ISSUE #13903: [batch] instance RAM and disk usage is not available in GCP Monitoring
### What happened?

All batch workers should have the Ops Agent so that we have RAM and disk usage available in GCP Monitoring. This is critical for diagnosing issues on workers.

### Version

0.2.124

### Relevant log output

_No response_

jigold:
Fixes #13903

Screenshot 2023-10-30 at 3 47 30 PM

### DIFF
--- a/batch/batch/cloud/gcp/driver/create_instance.py
+++ b/batch/batch/cloud/gcp/driver/create_instance.py
@@ -318,6 +318,8 @@ cat >/deploy-config/deploy-config.json <<EOF
 { json.dumps(get_deploy_config().with_location('gce').get_config()) }
 EOF
 
+curl -sSO https://dl.google.com/cloudagents/add-google-cloud-ops-agent-repo.sh
+bash add-google-cloud-ops-agent-repo.sh --also-install --version=1.*.*
 
 # retry once
 docker pull $BATCH_WORKER_IMAGE || \
--- a/batch/batch/cloud/gcp/driver/create_instance.py
+++ b/batch/batch/cloud/gcp/driver/create_instance.py
@@ -129,7 +129,7 @@ tag jvm-{idx}.log
                 'boot': True,
                 'autoDelete': True,
                 'initializeParams': {
-                    'sourceImage': f'projects/{project}/global/images/batch-worker-14',
+                    'sourceImage': f'projects/{project}/global/images/batch-worker-15',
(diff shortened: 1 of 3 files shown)


jigold:
Tests failed unrelatedly on `test_ci` due to a timeout on pulling an image.

danking:
Should we do this in create_instance.py or in the image itself? It seems like it will unnecessarily add to the startup time of the VMs

danking (changes requested):
see comment

daniel-goldstein:
Looks like the ops agent [also does logging](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent) in addition to monitoring. The logging agent we're using now is considered legacy, we should probably switch everything over to this new agent.

jigold:
We need to notify @jmarshall and @illusional when this goes in.

batch/batch/cloud/gcp/driver/create_instance.py
```diff
@@ -234,62 +208,60 @@ def scheduling() -> dict:
...
+        receivers: [hostmetrics]
+        processors: [metrics_filter]
 EOF
-rm /etc/google-fluentd/google-fluentd.conf.bak
 
-mkdir -p /batch/jvm-container-logs/
-{jvm_touch_command}
-
-sudo service google-fluentd restart
+sudo systemctl restart google-cloud-ops-agent
```
danking, line 264: Since we restart google-fluentd currently, I'll approve this, but I'm gonna create an issue to (1) characterize worker startup time and (2) reduce it.

danking (approved):
Great change; I'm excited to get CPU, RAM, disk, and network metrics!!

danking:
Heh. I didn't apply the WIP tag fast enough. Sorry @jigold I figured this might be the kind of change you wanted to control the merge time of.

batch/batch/cloud/gcp/driver/create_instance.py
jigold: The reason for doing the restart is we can't know the namespace and instance name ahead of time and have to modify the config anyways to insert these labels.

danking: The issue in question: https://github.com/hail-is/hail/issues/13976

PR #13956, opened 2023-10-31
[benchmark] fix combiner benchmarks

ehigham:
The combiner benchmarks broke following the deletion of the `experimental.vcf_combiner` python package. Re-implement them in terms of the `vds` package.


benchmark/python/benchmark_hail/__main__.py
```diff
@@ -1,9 +1,11 @@
+import sys
+
 from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
 
 from .run import cli as run
 from . import compare, create_resources, combine, summarize, visualize
 
-def main():
+def main(argv=sys.argv[1:]):
```
ehigham, line 8: This change makes it easier for me to launch the benchmark suite in a debugger and change which benchmark i want to debug from within the debugger.

ehigham:
Hi Chris, would you mind taking a look when you get a chance? I believe you might know the most about the functionality these are exercising. Do the new implementations look ok? Are they benchmarking equivalent functionality?

chrisvittal (approved):

benchmark/python/benchmark_hail/run/combiner_benchmarks.py
```diff
@@ -43,56 +46,51 @@ def compile_2k_merge(path):
...
-                            use_exome_default_intervals=True)
+        combiner = new_combiner(
+            gvcf_paths=list(paths),
+            output_path=tmpdir,
+            temp_path='/tmp',
+            branch_factor=16,
+            reference_genome='GRCh38',
+            use_exome_default_intervals=True
+        )
+        combiner.run()
```
chrisvittal, line 81: I believe this is now redundant.

chrisvittal (changes requested):
This is indeed what we would like to be checking.

benchmark/python/benchmark_hail/run/combiner_benchmarks.py
ehigham: Great! I'll remove it. Thanks.

chrisvittal (approved):

PR #13963, opened 2023-11-01
[compiler] Add `Switch` to `IR`

ehigham:
We have no high-level IR analogue to `CodeBuilderLike.switch`. Such a node is useful for flattening the IR in deeply-nested `If` nodes, predicated on integer equality.
This partially addresses the stack-overflow error on the `matrix_muluti_write_nothing` benchmark, which currently has a stack-overflow error when computing the type of the CDA.

### DIFF
--- a/hail/src/main/scala/is/hail/expr/ir/Children.scala
+++ b/hail/src/main/scala/is/hail/expr/ir/Children.scala
@@ -28,6 +28,12 @@ object Children {
     case Consume(value) => FastSeq(value)
     case If(cond, cnsq, altr) =>
       Array(cond, cnsq, altr)
+    case s@Switch(x, default, cases) =>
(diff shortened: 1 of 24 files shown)


danking (commented):
Needn't be in this PR, but it'd be nice if `hl.switch` from Python generated an `ir.Switch` when the discriminant is an int32!

ehigham:
Would you mind taking a look, Chris?

ehigham:
> Needn't be in this PR, but it'd be nice if `hl.switch` from Python generated an `ir.Switch` when the discriminant is an int32!

Nice idea. I hadn't intended to expose this to python just yet. Using a table to index the cases would be much more powerful!

chrisvittal (approved):

PR #13966, opened 2023-11-02
[query/plot] removes "chr" prefix for manhattan plot x-axis labels

### LINKED ISSUE #13952: [query] In GRCh38, the Manhattan plot x axis is unreadable
### What happened?

lsa_manhattan_plot

Notice that GRCh37 lacks the `chr` prefix and is thus readable.

We should either nix the `chr` prefix from `GRCh38` ticks or we should rotate ticks to vertical:
```
p.xaxis.major_label_orientation = "vertical"
```

### Version

0.2.126

### Relevant log output

_No response_

iris-garden:
Fixes #13952.

### DIFF
--- a/hail/python/hail/plot/plots.py
+++ b/hail/python/hail/plot/plots.py
@@ -1594,7 +1594,7 @@ def manhattan(pvals: 'Float64Expression',
     assert legend is not None
     legend.visible = False
     p.xaxis.ticker = contig_ticks
-    p.xaxis.major_label_overrides = dict(zip(contig_ticks, observed_contigs))
+    p.xaxis.major_label_overrides = dict(zip(contig_ticks, [contig.replace("chr", "") for contig in observed_contigs]))
 
     if significance_line is not None:
         p.renderers.append(Span(location=-math.log10(significance_line),


hail/python/hail/plot/plots.py
```diff
@@ -1573,7 +1573,7 @@ def manhattan(pvals: 'Float64Expression',
     source_pd['_contig'] = [locus.split(":")[0] for locus in source_pd['locus']]
 
     observed_contigs = [
-        contig for contig in ref.contigs.copy()
+        contig.replace("chr", "") for contig in ref.contigs.copy()
         if contig in set(source_pd['_contig'])
     ]
```
danking, line 1579: this next line is gonna fail b/c the contains aren't valid anymore. I think you'll have to replace this at the very latest moment: when you construct the major_label_overrides.

iris-garden:
tested the latest version of the change by copying and running [the plotting tutorial notebook](https://hail.is/docs/0.2/tutorials/08-plotting.html); verified that the code as is generates a visually identical manhattan plot, and then added this line before the manhattan plot is displayed:

```python
gwas = gwas.key_by(locus = hl.locus("chr" + gwas.locus.contig, hl.int(gwas.locus.position / 2), reference_genome="GRCh38"))
```

and the output looked like this (the big gaps between the points for each chromosome come from dividing by 2 above, which is a quick-and-dirty trick to make sure none of the loci have position values that are too large for GRCh38):

Screenshot 2023-11-03 at 15 01 17

danking (dismissed):
Do you have unpushed changes? 

This fails:
```
In [4]: gwas = hl.balding_nichols_model(1,100, 1, reference_genome='GRCh38').rows().annotate(p_value=10 **
   ...: -hl.rand_int32(1, 8))
   ...: hl.plot.manhattan(gwas.p_value, gwas.locus).show()
```

iris-garden:
@danking oh yep, sorry! github was down but it said the changes had pushed from the command line, now it's fixed

danking (approved):

PR #13967, opened 2023-11-02
[ci] Remove dead code from CI configs

daniel-goldstein:
- `CreateNamespaceStep.public` was entirely unused
- `adminServiceAccount` is not used in `build.yaml` so `CreateNamespaceStep.admin_service_account` is always `None` meaning it has no effect.
- The three environment variables that I deleted from the `deployment.yaml` are as far as I can tell entirely unused (they are now grabbed from the global config)

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -37,7 +37,6 @@ steps:
   - kind: createNamespace
     name: default_ns
     namespaceName: default
-    public: true
     secrets:
       - name: auth-oauth2-client-secret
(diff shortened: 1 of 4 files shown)


jigold (approved):
I wouldn't be surprised if the public and service account options on CreateNamespace were artifacts from when we were running batch pods inside k8s in a secondary namespace. But I'm not 100% sure.

